PR #1460, opened 2015-05-06
tls: force 'raw' reassembly after each record

inliniac:
Trigger raw reassembly after each record and after the handshake.


inliniac:
Replaced by #1461

PR #1455, opened 2015-04-30
Fix bug #1435 (data loss when dumping payloads to JSON) v2

gozzy:
Slightly fixed #1423.

This pull request tries to fix bug #1435. Data should not be lost any more.

PR #1464, opened 2015-05-07
Fix bug #1435 (data loss when dumping payloads to JSON) v3

gozzy:
Fixed #1455 - use MemBufferWriteRaw() wrapper instead of memcpy. Also fixed MemBufferWriteRaw() - there was unneeded return stetement.

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -93,12 +93,10 @@ typedef struct JsonAlertLogThread_ {
 
 /* Callback function to pack payload contents from a stream into a buffer
  * so we can report them in JSON output. */
-static int AlertJsonPrintStreamSegmentCallback(const Packet *p, void *data, uint8_t *buf, uint32_t buflen)
+static int AlertJsonDumpStreamSegmentCallback(const Packet *p, void *data, uint8_t *buf, uint32_t buflen)
 {
     MemBuffer *payload = (MemBuffer *)data;
-
-    PrintStringsToBuffer(payload->buffer, &payload->offset, payload->size,
-                         buf, buflen);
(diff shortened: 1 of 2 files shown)


src/output-json-alert.c
```diff
@@ -287,27 +285,30 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
...
-                PrintStringsToBuffer(packet_buf, &offset,
-                                     p->payload_len + 1,
-                                     p->payload, p->payload_len);
-
                 if (json_output_ctx->flags & LOG_JSON_PAYLOAD_BASE64) {
-                    unsigned long len = sizeof(packet_buf) * 2;
+                    size_t len = p->payload_len * 2 + 1;
                     unsigned char encoded[len];
-                    Base64Encode(packet_buf, offset, encoded, &len);
+                    Base64Encode(p->payload, p->payload_len, encoded, &len);
```
inliniac: output-json-alert.c: In function ‘AlertJson’:
output-json-alert.c:301:21: error: passing argument 4 of ‘Base64Encode’ from incompatible pointer type [-Werror]
util-crypt.h:82:5: note: expected ‘long unsigned int *’ but argument is of type ‘size_t *’

gozzy: It's a bit strange - my compiler doesn't see anything suspicious.
Maybe we should return unsigned long here? For instance, LogTlsLogPem() uses unsigned long when calling Base64Encode.

inliniac: Probably yeah. size_t is 4 bytes for 32bit args, 8bytes for 64bit args.

gozzy:
Replaced with #1468

PR #1466, opened 2015-05-07
Threads rename v6

Maxtors:
Renaming of runmodes threads and flow threads to new naming standard.


Maxtors:
Hi Victor, cant remember, but was there any work here missing / faulty?

PR #1423, opened 2015-04-16
Fix bug #1435 (data loss when dumping payloads to JSON)

gozzy:
This pull request tries to fix bug #1435. Data should not be lost any more.

PR #1467, opened 2015-05-07
Support for reconnecting unix domain socket log files (Issue #1423)

jsbarber:
Issue #1423

This patch changes the behavior of the code that writes log records to a unix domain socket:
If suricata cannot immediately connect to the receiver, the receiver socket path name and socket type (stream vs dgram) will be remembered in the LogFileCtx. When log records are written later, the connect will be attempted.

Also, if a connection to the receiver is broken -- for example, if the receiver program is restarted -- suricata will now clean up the old connection and automatically reconnect.

Since creating the unix domain socket then unsuccessfully attempting to connect then tearing it down requires relatively significant kernel resources, the reconnection is only retried occasionally (currently set to 1/2 second). This moderation scheme uses Posix clock_gettime with MONOTONIC_CLOCK and is ifdef'd appropriately since I'm not sure on what systems suricata runs. If the system doesn't support the posix time, it will simply attempt to reconnect every time an alert log record needs to be sent.

Tested with both unix_stream and unix_dgram socket types. Only tested with fast-alert log, but should work for any caller of SCConfLogOpenGeneric.

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -36,11 +36,12 @@
 
 /** \brief connect to the indicated local stream socket, logging any errors
  *  \param path filesystem path to connect to
+ *  \param log_err, non-zero if connect failure should be logged.
  *  \retval FILE* on success (fdopen'd wrapper of underlying socket)
  *  \retval NULL on error
  */
 static FILE *
-SCLogOpenUnixSocketFp(const char *path, int sock_type)
+SCLogOpenUnixSocketFp(const char *path, int sock_type, int log_err)
 {
     struct sockaddr_un sun;
     int s = -1;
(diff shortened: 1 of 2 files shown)


src/util-logopenfile.c
```diff
@@ -73,6 +76,49 @@ SCLogOpenUnixSocketFp(const char *path, int sock_type)
...
+        SCLogError(SC_ERR_SOCKET,
+            "Write error on Unix socket \"%s\": %s; reconnecting...",
+            log_ctx->filename, strerror(errno));
+        fclose(log_ctx->fp);
+        log_ctx->fp = NULL;
+        log_ctx->reconn_timer = 0;
+        disconnected = 1;
+    }
+
+#if defined(_POSIX_TIMERS) && _POSIX_TIMERS > 0 && _POSIX_MONOTONIC_CLOCK
```
inliniac: why not simply use gettimeofday here? Seems simpler and more portable.

src/util-logopenfile.c
```diff
@@ -64,8 +65,10 @@ SCLogOpenUnixSocketFp(const char *path, int sock_type)
     return ret;
 
 err:
-    SCLogError(SC_ERR_SOCKET, "Error connecting to socket \"%s\": %s",
-               path, strerror(errno));
+    if (log_err)
+        SCLogError(SC_ERR_SOCKET,
```
inliniac: I guess this could be a SCLogWarning instead?

src/util-logopenfile.c
```diff
@@ -83,9 +129,21 @@ static int SCLogFileWrite(const char *buffer, int buffer_len, LogFileCtx *log_ct
 
     int ret = 0;
 
+    if (log_ctx->fp == NULL && log_ctx->is_sock)
+        ret = SCLogUnixSocketReconnect(log_ctx);
```
inliniac: I think it's a bit unclear now what SCLogFileWrite returns: it can either return the retval of SCLogUnixSocketReconnect, or of fwrite.

inliniac:
It seems currently the logs will be lost if connection is lost. Even if the SCLogFileWrite returns error, the errors are ignored by the callers. Wonder if it would make sense to create a buffering mechanism, to at least overcome short outages. There would need to be limits and all of course. Thoughts?

src/util-logopenfile.c
jsbarber: I tend to avoid gettimeofday due to time adjustment hassles (time going backward and so forth). I suppose that issue could be ignored here.

inliniac: On 07/16/2015 01:57 PM, jsbarber wrote:
> I tend to avoid gettimeofday due to time adjustment hassles (time going
> backward and so forth). I suppose that issue could be ignored here.
> 

I think so ya, we use it in other places as well :)

src/util-logopenfile.c
jsbarber: It could be, but it seems like an important notification: since an alert (or whatever) may not have been recorded due to the connect failure, error seemed like the right characterization.

src/util-logopenfile.c
jsbarber: SCLogUnixSocketReconnect returns zero iff log_ctx->fp is NULL. Hence, it returns the same value as before: zero on failure; otherwise the result of fwrite.

jsbarber:
I think a buffering mechanism would be a useful addition. However, this diff was primarily motivated by a desire just to make suricata tolerant of the alert receiver being restarted; or of suricata being started slightly before the alert receiver.

A buffering mechanism would also probably require some kind of asynchronous wakeup to periodically retry the connect and flush any accumulated log messages.

src/util-logopenfile.c
inliniac: Ok cool. Could you add a comment above the function to explain this?

inliniac: Hah this was meant to go below.

src/util-logopenfile.c
inliniac: Ok cool. Could you add a comment above the function to explain this?

src/util-logopenfile.c
inliniac: In new code I'm really doing errors only for unrecoverable events, while warnings are used for recoverable events. So here a warning would make sense.

inliniac:
Yeah, maybe @regit has some ideas about it. He implemented something similar in ulogd2.

Anyway, if you submit an updated branch+pr I'll get it merged quickly. Thanks!

inliniac:
Oh and rebased against the latest master plz :)

inliniac:
Replaced by #1597

PR #1468, opened 2015-05-07
Fix bug #1435 (data loss when dumping payloads to JSON) v4

gozzy:
Replaces #1464


inliniac:
Merge through #1470, thanks!

PR #1474, opened 2015-05-08
Bug 1231 - SSL State Negation - v3

jasonish:
Accidentally deleted the previous branch, so rebasing and resubmitting.

Previous PR: #1147 

Redmine Issue: https://redmine.openinfosecfoundation.org/issues/1231

First this required the current state being passed to the ssl_state match function instead of the cumulative state, otherwise ssl_state:client_hello would also match after while in the state server_hello. Note that this patch is a bit naive as I don't know the SSL protocol that well, so its more of a preview or RFC.

This patch also changes the state separator from "|" to "," to be compatible with Snort.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/79
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/80

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -142,20 +142,19 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
 
     switch (ssl_state->curr_connp->handshake_type) {
         case SSLV3_HS_CLIENT_HELLO:
-            ssl_state->flags |= SSL_AL_FLAG_STATE_CLIENT_HELLO;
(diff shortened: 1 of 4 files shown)


inliniac:
"This patch also changes the state separator from "|" to "," to be compatible with Snort." It would probably be better to keep supporting the old way as well, just in case ppl use it.

jasonish:
> "This patch also changes the state separator from "|" to "," to be compatible with Snort." It would probably be better to keep supporting the old way as well, just in case ppl use it.

Yeah, makes sense. Thats the simple part. The rebase not so much.

inliniac:
Seems this was already replaced with #2044

PR #1392, opened 2015-03-21
unix socket: protocol v0.2

regit:
This patch updates the unix socket protocol. Messages send from
the server have now a '\n' at the end. This allows the client to
detect easily the end of a command.

Backward compatibility is preserved as a client with the older
version of the protocol can still connect to a Suricata with
version 2 of the protocol.

PR builds:
- PR build: https://buildbot.openinfosecfoundation.org/builders/regit/builds/58
- PR pcaps: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/56

PR #1476, opened 2015-05-09
unix socket: protocol v0.2

regit:
Rework of #1392 introducing the '\n' termination for client message.

This patch updates the unix socket protocol. Messages send from
the server and the client have now a '\n' at the end. This allows
both sides to detect easily the end of a command.

Backward compatility is preserved as a client with the older
version of the protocol can still connect to a Suricata with
version 2 of the protocol.

PR builds
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/65
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/63

### DIFF
--- a/scripts/suricatasc/src/suricatasc.py
+++ b/scripts/suricatasc/src/suricatasc.py
@@ -25,10 +25,11 @@ from time import sleep
 import select
 import sys
 
-SURICATASC_VERSION = "0.9"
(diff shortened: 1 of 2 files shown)


regit:
Closing and rebasing.

PR #1480, opened 2015-05-11
radix-tree - prevent out of bounds array access - v1

jasonish:
An IPv6 entry specified before an IPv4 entry on the host-os-policy
table can cause the stream byte array to be access one byte after
the end of the allocated memory at util-radix-tree.c:578.

Redmine bug: https://redmine.openinfosecfoundation.org/issues/1467

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/82
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/83


inliniac:
Merged after rebase, thanks!

PR #1482, opened 2015-05-11
Fix rcode parsing, as noticed by Coverity.

edeca:
Without support for OPT RR from RFC6891 (Extension mechanisms for DNS)
values of RCODE above 15 are not possible.  Remove dead code which will
never match.


inliniac:
Merged after rebase, thanks!

PR #1440, opened 2015-04-18
Bug #1441: add timezone to timestamp in JSON logs v4

gozzy:
Fixed #1439 - that one was not fully ISO conformant (missed 'T' between date and time).

PR #1485, opened 2015-05-13
Feature #1440: support wildcards in rule filenames v1

gozzy:
Attempt to support wildcards in rule filenames. It's a draft, but seems to work. Any remarks or suggestions are welcome.

### DIFF
--- a/src/detect.c
+++ b/src/detect.c
@@ -199,6 +199,8 @@
 
 #include "runmodes.h"
 
+#include <glob.h>
+
 extern int rule_reload;
 
 extern int engine_analysis;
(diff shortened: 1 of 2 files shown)


src/detect.c
```diff
@@ -393,20 +439,14 @@ int SigLoadSignatures(DetectEngineCtx *de_ctx, char *sig_file, int sig_file_excl
...
-    int badfiles = 0;
-
-    int goodtotal = 0;
-    int badtotal = 0;
-
     char varname[128] = "rule-files";
+    int good_sigs = 0;
+    int bad_sigs = 0;
+
+    bzero(&sig_stat, sizeof(SigFileLoaderStat));
```
inliniac: bzero is deprecated, please use memset

src/detect.c
```diff
@@ -199,6 +199,8 @@
 
 #include "runmodes.h"
 
+#include <glob.h>
```
inliniac: might be good to test for this header file in configure.ac

inliniac:
If I read the code correctly this patch pushes all detect file loads through the new 'glob' function, where for exact matches it will work as before, and for patterns it loads multiple files. Correct?

gozzy:
@inliniac, yes, for exact matches it works like it did before as glob() returns one match in this case.

src/detect.c
gozzy: Ok, will try to handle it. If header is not found, suricata should compile anyway, but without wildcard support?

inliniac: Lets not bother with a 'not found'-path at first. I will run the new PR through my QA and if that passes we can assume glob.h is available everywhere.

gozzy:
Replaced with #1500.

PR #1491, opened 2015-05-19
Allow colon in SSH version string

atonkyra:
Simple patch for regex that checks SSH version string (ssh.softwareversion).


inliniac:
Merged through #1608, thanks!

PR #1492, opened 2015-05-19
[preview] hsbd: inspect buffer depending on the engine mode

glongo:

To test it, you need
this rule: drop http any any -> any any (msg:"alert"; file_data; content:"<?xml"; depth:5; sid:1000; rev:1;)
and set
response-body-inspect-window: 1kb

### DIFF
--- a/src/detect-engine-hsbd.c
+++ b/src/detect-engine-hsbd.c
@@ -89,6 +89,119 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
     return 0;
 }
 
+static void HSBDGetBufferForTXInIDSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    int first = 1;
+    while (cur != NULL) {
+        /* see if we can filter out chunks */
+        if (htud->response_body.body_inspected > 0) {
+            if (cur->stream_offset < htud->response_body.body_inspected) {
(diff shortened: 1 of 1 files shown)


src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,119 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+                return;
+            }
+            det_ctx->hsbd[index].buffer = ptmp;
+        }
+
+        /* see if we can filter out chunks */
+        if (htud->response_body.content_len_so_far > window_size) {
+            left_edge = htud->response_body.content_len_so_far - window_size;
+
+            if (cur->len < left_edge) {
```
inliniac: I think you need to include stream_offset in this evaluation. The len only works well for the first chunk.

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,119 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+static void HSBDGetBufferForTXInIPSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    SCEnter();
+    uint32_t left_edge = 0;
+    uint32_t window_size = 0;
+    void *ptmp;
+
+    if ((ptmp = SCMalloc(htp_state->cfg->response_inspect_window)) == NULL) {
```
inliniac: This seems unnecessary, as the while loop below also does this allocation.

PR #1493, opened 2015-05-20
Eve fix 2015 v1.2

regit:
This patchset mainly update the JSON output submodules to get rid of global variable.

It could be used as a base for #1449 code rework if @inliniac is ok with this basis.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/66
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/64

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -78,14 +78,13 @@
 #define JSON_STREAM_BUFFER_SIZE 4096
 
 typedef struct AlertJsonOutputCtx_ {
-    LogFileCtx* file_ctx;
+    OutputJsonCtx* json_ctx;
     uint8_t flags;
     HttpXFFCfg *xff_cfg;
 } AlertJsonOutputCtx;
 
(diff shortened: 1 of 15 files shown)


inliniac:
I still don't see the need for a separate OutputJsonCtx. I would like things to move into LogFileCtx if possible.

regit:
OK so that was the point why the previous PR was refused :) I did not remember if it was this part of my redis patchset that was the innocent victim.

My main concern regarding the move is that do we want potential options for a EVE output to be set on LogFileCtx as this is our only reference here. I'm giving a try to that approach.

inliniac:
I may actually change my mind on this. I see the point of the Eve settings that don't belong in LogFileCtx, and a generic OutputJsonCtx would not be much different from LogHttpFileCtx for example.

regit:
OK, let me know. I've done #1494 so we have a comparison.

PR #1494, opened 2015-05-20
Eve fix 2015 v1.3

regit:
As suggested by @inliniac, this is a rework of #1493 using the LogFileCtx to store the file type information. And it is indeed shorter.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/67
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/65

### DIFF
--- a/src/util-logopenfile.h
+++ b/src/util-logopenfile.h
@@ -45,6 +45,9 @@ typedef struct LogFileCtx_ {
      * record cannot be written to the file in one call */
     SCMutex fp_mutex;
 
+    /** the type of file */
+    uint8_t type;
+
     /** The name of the file */
     char *filename;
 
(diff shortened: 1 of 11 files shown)


src/util-logopenfile.h
```diff
@@ -45,6 +45,9 @@ typedef struct LogFileCtx_ {
      * record cannot be written to the file in one call */
     SCMutex fp_mutex;
 
+    /** the type of file */
+    uint8_t type;
```
inliniac: This should probably be a enum like 'enum JsonOutput' from output-json.h.

regit: Are you ok to move JsonOutput to OutputType and have it defined in util-logfile.h ?

PR #1495, opened 2015-05-20
smtp file_data: permit to use flow keyword in a rule

glongo:
Currently the following rule can't be loaded:
alert tcp any any -> any 25 (msg:"SMTP file_data test"; flow:to_server,established; file_data; content:"abc";sid:1;)
and produces the error output:
"Can't use file_data with flow:to_server or from_client with http or smtp."

This checks if the tcp protocol is specified in a signature,
so permits to use flow keyword also.

Issue reported by rmkml.

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/31
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/30

### DIFF
--- a/src/detect-file-data.c
+++ b/src/detect-file-data.c
@@ -84,7 +84,7 @@ static int DetectFiledataSetup (DetectEngineCtx *de_ctx, Signature *s, char *str
         return -1;
     }
 
-    if ((s->init_flags & SIG_FLAG_INIT_FLOW) &&
+    if ((s->init_flags & SIG_FLAG_INIT_FLOW) && !DetectProtoContainsProto(&s->proto, IPPROTO_TCP) &&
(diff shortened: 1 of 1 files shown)


src/detect-file-data.c
```diff
@@ -84,7 +84,7 @@ static int DetectFiledataSetup (DetectEngineCtx *de_ctx, Signature *s, char *str
         return -1;
     }
 
-    if ((s->init_flags & SIG_FLAG_INIT_FLOW) &&
+    if ((s->init_flags & SIG_FLAG_INIT_FLOW) && !DetectProtoContainsProto(&s->proto, IPPROTO_TCP) &&
```
inliniac: I don't see how this fixes anything. It just effectively disables this whole test, as all valid sigs that use 'alert http', 'alert smtp' or 'alert tcp' will fail to meet this new condition.

PR #1496, opened 2015-05-20
--set fixup - v1

jasonish:
Spoted by @regit .

Handle spaces around the '=' in --set.

The first commit just fixes up style to make all function declarations the same.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/93
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/94

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -61,8 +61,7 @@ static ConfNode *root_backup = NULL;
  * \retval The existing configuration node if it exists, or a newly
  *   created node for the provided name.  On error, NULL will be returned.
  */
-static ConfNode *
-ConfGetNodeOrCreate(const char *name, int final)
+static ConfNode *ConfGetNodeOrCreate(const char *name, int final)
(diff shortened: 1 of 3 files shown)


inliniac:
For some strange reason this doesn't build for me, but it passed prscript:
<pre>
suricata.c: In function ‘ParseCommandLine’:
suricata.c:1486:21: error: passing argument 2 of ‘ConfSetFromString’ makes integer from pointer without a cast [-Werror]
                     if (!ConfSetFromString(optarg, val)) {
                     ^
In file included from suricata.c:72:0:
conf.h:65:5: note: expected ‘int’ but argument is of type ‘char *’
 int ConfSetFromString(const char *input, int final);
     ^

</pre>

jasonish:
See #1503 .

PR #1497, opened 2015-05-20
Suri graphite improvements

regit:
A fix and two new features for suri-graphite:
- fix the capability to set port
- add daemon feature
- add output to file to be able to inject the data in a graphite instance via logstash


inliniac:
Merged through #1502, thanks Eric.

PR #1500, opened 2015-05-21
Feature #1440: support wildcards in rule filenames

gozzy:
Rebased and slightly fixed #1485.
Adds support for wildcards in rule filenames.

### DIFF
--- a/src/detect.c
+++ b/src/detect.c
@@ -199,6 +199,10 @@
 
 #include "runmodes.h"
 
+#include <glob.h>
+
+extern int rule_reload;
+
 extern int engine_analysis;
 static int fp_engine_analysis_set = 0;
 static int rule_engine_analysis_set = 0;
@@ -287,8 +291,9 @@ char *DetectLoadCompleteSigPath(const DetectEngineCtx *de_ctx, char *sig_file)
  *  \brief Load a file with signatures
  *  \param de_ctx Pointer to the detection engine context
  *  \param sig_file Filename to load signatures from
- *  \param sigs_tot Will store number of signatures processed in the file
- *  \retval Number of rules loaded successfully, -1 on error
+ *  \param goodsigs_tot Will store number of valid signatures in the file
+ *  \param badsigs_tot Will store number of invalid signatures in the file
+ *  \retval 0 on success, -1 on error
  */
 static int DetectLoadSigFile(DetectEngineCtx *de_ctx, char *sig_file,
         int *goodsigs, int *badsigs)
@@ -301,11 +306,6 @@ static int DetectLoadSigFile(DetectEngineCtx *de_ctx, char *sig_file,
 
     (*goodsigs) = 0;
     (*badsigs) = 0;
-
-    if (sig_file == NULL) {
-        SCLogError(SC_ERR_INVALID_ARGUMENT, "opening rule file null");
-        return -1;
-    }
 
     FILE *fp = fopen(sig_file, "r");
     if (fp == NULL) {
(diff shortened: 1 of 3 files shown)


src/detect.c
```diff
@@ -373,15 +373,65 @@ static int DetectLoadSigFile(DetectEngineCtx *de_ctx, char *sig_file,
...
+        SCLogWarning(SC_ERR_NO_RULES, "No rule files match the pattern %s", pattern);
+        ++(st->bad_files);
+        return -1;
+    } else if (r != 0) {
+        SCLogError(SC_ERR_OPENING_RULE_FILE, "error expanding template %s: %s",
+                 pattern, strerror(errno));
+        return -1;
+    }
+
+    for (size_t i = 0; i < files.gl_pathc; i++) {
```
inliniac: <pre>
detect.c: In function 'ProcessSigFiles':
detect.c:408:26: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     for (size_t i = 0; i < files.gl_pathc; i++) 
</pre>

gozzy: Seems to be a buggy header, because gl_pathc in struct glob_t should be size_t. Maybe, type cast on gl_pathc will do the trick?

gozzy:
Replaced with #1501

src/detect.c
inliniac: On 05/21/2015 06:19 PM, Alexander Gozman wrote:
> Seems to be a buggy header, because gl_pathc in struct glob_t should be
> size_t. Maybe, type cast on gl_pathc will do the trick?

This was on OpenBSD 5.4. The gl_pathc member is a regular int there.

pevma:
What is the appropriate way for this to be run/work?
I tried the flowing - 
suricata -c /etc/suricata/suricata.yaml -S /etc/suricata/rules/emerging* -v -i eth0
suricata -c /etc/suricata/suricata.yaml  -v -i eth0 -s /etc/suricata/rules/*
suricata -c /etc/suricata/suricata.yaml  -v -i eth0 -S /etc/suricata/rules/*.rules
suricata -c /etc/suricata/suricata.yaml  -v -i eth0 -S /etc/suricata/rules/emerging*.rules

and also tried similar in suricata.yaml - in all the cases it Suricata errs out and does not start.

gozzy:
@pevma, all cases should be valid. What kind of errors do you get? I develop/test it on Fedora 21 and everything seems to work. However, if rules are specified with -s or -S, the argument should be quoted, otherwise suricata blames on wrong bpf filter (and stock suricata does the same).

pevma:
@gozzy 
Example (Ubuntu) - 
suricata -c /etc/suricata/suricata.yaml  -v -i eth0 -S "/etc/suricata/rules/*.rules"

[18173] 21/5/2015 -- 20:06:35 - (detect.c:453) <Info> (SigLoadSignatures) -- Loading rule file: /etc/suricata/rules/*.rules 
[18173] 21/5/2015 -- 20:06:35 - (detect.c:313) <Error> (DetectLoadSigFile) -- [ERRCODE: SC_ERR_OPENING_RULE_FILE(41)] - opening rule file /etc/suricata/rules/*.rules : No such file or directory.

gozzy:
I can't understand why asterisk disappears from pattern... I'll try to test it on Ubuntu virtual machine. What ubuntu version?

pevma:
I used Trusty.
The asterisk appears - it is just that in this comment it makes it disappear.

gozzy:
@pevma, just checked on Ubuntu Trusty (14.04) - nothing suspicious. Did you start my version of suricata? Only stock version says 'No such file or directory' because it doesn't support wildcards and tries to open a file "as is". Modified version says that no files match the pattern, if there were no appropriate files

pevma:
Which ver do you refer to exactly?

gozzy:
@pevma I mean version cloned from my repo (either rules-wildcard2 branch, or, which is better, rules-wildcard3)

gozzy:
Or version with glob patch (from this PR) applied.

gozzy:
From what I see now it seems that stock version was run - no leading dot and slash in command line, different error message...

pevma:
You stand correct. Tested - looks good !
Thank  you

PR #1501, opened 2015-05-21
Feature #1440: support wildcards in rule filenames v3

gozzy:
Fixed #1500.
Adds support for wildcards in rule filenames.


gozzy:
Probably there's a problem with glob. I'll test it on Ubuntu.

inliniac:
Thanks Alexander!

PR #1504, opened 2015-05-24
Add Feature #1454. CEE/Lumberjack syslog support v2

zrasmor:
Rework addressing comments from PR #1449 and subsequent refactoring of JSON output module.

@inliniac @regit - I elected to add a new flag to the LogFileCtx structure instead of adding the @cee: cookie to the 'prefix' field, but let me know if you'd prefer it done differently.

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -341,9 +341,18 @@ int OutputJSONBuffer(json_t *js, LogFileCtx *file_ctx, MemBuffer *buffer)
         return TM_ECODE_OK;
 
     SCMutexLock(&file_ctx->fp_mutex);
-    if (file_ctx->type == LOGFILE_TYPE_SYSLOG) {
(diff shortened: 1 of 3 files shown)


inliniac:
I'm wondering if it would make sense to support adding a generic prefix instead. Or is '@cee:' really the only conceivable prefix?

eg:
prefix: "@cee"

zrasmor:
Yes, that's what I was thinking when i saw the 'prefix' field in the LogFileCtx structure. We could make the yaml field more generic to support any Eve log event prefix that the user wants. 

As for possible prefixes, "@cee: " is still the only one i can think of - in general I'm not sure how common it is to prepend strings to JSON output.

But it might make sense to make it generic to account for compatibility with other tools or use cases that might need a prepended string that we may not know about.

inliniac:
Yeah lets make it generic, but default "@cee:" in the yaml.

PR #1505, opened 2015-05-25
hsbd: inspect buffer depending on the engine mode

glongo:

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/34
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/33

The first build has failed with the following error:
"Test SCLogTestInit02: Error opening file /var/log/suricata/boo.txt"

### DIFF
--- a/src/detect-engine-hsbd.c
+++ b/src/detect-engine-hsbd.c
@@ -89,6 +89,111 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
     return 0;
 }
 
+static void HSBDGetBufferForTXInIDSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    int first = 1;
+    while (cur != NULL) {
+        /* see if we can filter out chunks */
+        if (htud->response_body.body_inspected > 0) {
+            if (cur->stream_offset < htud->response_body.body_inspected) {
+                if ((htud->response_body.body_inspected - cur->stream_offset) > htp_state->cfg->response_inspect_window) {
+                    cur = cur->next;
+                    continue;
+                } else {
+                    /* include this one */
+                }
+            } else {
+                /* include this one */
+            }
+        }
+
+        if (first) {
+            det_ctx->hsbd[index].offset = cur->stream_offset;
+            first = 0;
+        }
+
+        /* see if we need to grow the buffer */
+        if (det_ctx->hsbd[index].buffer == NULL || (det_ctx->hsbd[index].buffer_len + cur->len) > det_ctx->hsbd[index].buffer_size) {
+            void *ptmp;
+            det_ctx->hsbd[index].buffer_size += cur->len * 2;
+
+            if ((ptmp = SCRealloc(det_ctx->hsbd[index].buffer, det_ctx->hsbd[index].buffer_size)) == NULL) {
+                SCFree(det_ctx->hsbd[index].buffer);
+                det_ctx->hsbd[index].buffer = NULL;
+                det_ctx->hsbd[index].buffer_size = 0;
+                det_ctx->hsbd[index].buffer_len = 0;
+                return;
+            }
+            det_ctx->hsbd[index].buffer = ptmp;
+        }
+        memcpy(det_ctx->hsbd[index].buffer + det_ctx->hsbd[index].buffer_len, cur->data, cur->len);
+        det_ctx->hsbd[index].buffer_len += cur->len;
+
+        cur = cur->next;
+    }
+
+    /* update inspected tracker */
+    htud->response_body.body_inspected = htud->response_body.last->stream_offset + htud->response_body.last->len;
+}
+
+static void HSBDGetBufferForTXInIPSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    SCEnter();
+    uint32_t left_edge = 0;
+    uint32_t window_size = 0;
+
+    /* maybe it's not safe to change the htp_state->cfg->response_inspect_window
+       value since it can be used somewhere else. */
+    window_size = htp_state->cfg->response_inspect_window;
+
+    int first = 1;
+    while (cur != NULL) {
+        /* see if we need to grow the buffer */
+        if (det_ctx->hsbd[index].buffer == NULL) {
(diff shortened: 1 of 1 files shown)


src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,111 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+       value since it can be used somewhere else. */
+    window_size = htp_state->cfg->response_inspect_window;
+
+    int first = 1;
+    while (cur != NULL) {
+        /* see if we need to grow the buffer */
+        if (det_ctx->hsbd[index].buffer == NULL) {
+            void *ptmp;
+
+            if ((htp_state->cfg->response_inspect_window + cur->len) > window_size)
```
inliniac: This is not called unless it's about the first chunk when this condition is still true: if (det_ctx->hsbd[index].buffer == NULL) {

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,111 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+
+static void HSBDGetBufferForTXInIPSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    SCEnter();
+    uint32_t left_edge = 0;
+    uint32_t window_size = 0;
+
+    /* maybe it's not safe to change the htp_state->cfg->response_inspect_window
```
inliniac: It is safe to use htp_state->cfg->response_inspect_window directly.

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,111 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+                /* include this one */
+            }
+        }
+
+        if (first) {
+            det_ctx->hsbd[index].offset = cur->stream_offset;
+            first = 0;
+        }
+
+        memcpy(det_ctx->hsbd[index].buffer + det_ctx->hsbd[index].buffer_len, cur->data, cur->len);
```
inliniac: If the 2nd or later chunk is bigger than window_size, the buffer is not expanded (see comment above). The chunk is used completely, so it will lead to a buffer overflow here.

src/detect-engine-hsbd.c
inliniac: What is the meaning of the (htp_state->cfg->response_inspect_window + cur->len) > window_size check? Does it just say: "if window+something > window" (well duh of course) then do something?

src/detect-engine-hsbd.c
```diff
@@ -158,63 +263,24 @@ static uint8_t *DetectEngineHSBDGetBufferForTX(htp_tx_t *tx, uint64_t tx_id,
...
-
-            if ((ptmp = SCRealloc(det_ctx->hsbd[index].buffer, det_ctx->hsbd[index].buffer_size)) == NULL) {
-                SCFree(det_ctx->hsbd[index].buffer);
-                det_ctx->hsbd[index].buffer = NULL;
-                det_ctx->hsbd[index].buffer_size = 0;
-                det_ctx->hsbd[index].buffer_len = 0;
-                goto end;
-            }
-            det_ctx->hsbd[index].buffer = ptmp;
+    if (EngineModeIsIDS()) {
```
inliniac: I would like this to be a setting instead, like the 'stream.inline' option for the stream engine. The default value should be 'auto', where it means that in IPS mode we use the IPS logic, in IDS mode the IDS logic.

src/detect-engine-hsbd.c
```diff
@@ -3631,6 +3697,419 @@ static int DetectEngineHttpServerBodyFileDataTest03(void)
...
+    if (de_ctx != NULL)
+        SigCleanSignatures(de_ctx);
+    if (de_ctx != NULL)
+        DetectEngineCtxFree(de_ctx);
+
+    StreamTcpFreeConfig(TRUE);
+    FLOW_DESTROY(&f);
+    UTHFreePackets(&p1, 1);
+    UTHFreePackets(&p2, 1);
+    return result;
```
inliniac: Don't forget to reset the 'Engine mode'

src/detect-engine-hsbd.c
```diff
@@ -3631,6 +3697,419 @@ static int DetectEngineHttpServerBodyFileDataTest03(void)
...
```
inliniac: Reset engine mode

src/detect-engine-hsbd.c
```diff
@@ -3631,6 +3697,419 @@ static int DetectEngineHttpServerBodyFileDataTest03(void)
...
+
+    if (!(DetectEngineAppendSig(de_ctx, "alert http any any -> any any "
+                               "(msg:\"http server body test\"; "
+                               "file_data; content:\"abcd\"; sid:2;)")))
+        goto end;
+    if (!(DetectEngineAppendSig(de_ctx, "alert http any any -> any any "
+                              "(msg:\"http server body test\"; "
+                              "file_data; content:\"12345678\"; sid:3;)")))
+        goto end;
+
```
inliniac: Can you add a sig for: content:"cd12"

glongo: This sig fails, since the buffer contains the last chunk '12345678', and not 'cd123456'

inliniac: On 05/26/2015 10:31 AM, Giuseppe Longo wrote:
> In src/detect-engine-hsbd.c
> <https://github.com/inliniac/suricata/pull/1505#discussion_r31013539>:
> 
>> +    de_ctx->flags |= DE_QUIET;
>> +
>> +    if (!(DetectEngineAppendSig(de_ctx, "alert http any any -> any any "
>> +                               "(msg:\"http server body test\"; "
>> +                               "file_data; content:\"ab\"; sid:1;)")))
>> +        goto end;
>> +
>> +    if (!(DetectEngineAppendSig(de_ctx, "alert http any any -> any any "
>> +                               "(msg:\"http server body test\"; "
>> +                               "file_data; content:\"abcd\"; sid:2;)")))
>> +        goto end;
>> +    if (!(DetectEngineAppendSig(de_ctx, "alert http any any -> any any "
>> +                              "(msg:\"http server body test\"; "
>> +                              "file_data; content:\"12345678\"; sid:3;)")))
>> +        goto end;
>> +
> 
> This sig fails, since the buffer contains the last chunk '12345678', and
> not 'cd123456'

That will need to be addressed then :)

PR #1506, opened 2015-05-26
Redisdead v1.4

regit:
This patchset adds redis output to EVE. redis can now be alternatively selected as output mode. It implements:
- LIST mode: software like logstash can fetch events from the LIST and process them
- SUBSCRIPTION mode: this is a multicast for events which are sent to multiple listeners.
- pipelining: it could also be called batch mode. Events are sent by block. This lower the number of write and read on the TCP socket

Sadly, hiredis is not threadsafe so suricata is making all operations inside the usual lock. An update to this part of the code could be to have one file/socket per threads to allow non mutex-protected writing.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/70
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/68

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -332,8 +332,17 @@ json_t *CreateJSONHeader(Packet *p, int direction_sensitive, char *event_type)
 
 int OutputJSONBuffer(json_t *js, LogFileCtx *file_ctx, MemBuffer *buffer)
 {
-    char *js_s = json_dumps(js,
-                            JSON_PRESERVE_ORDER|JSON_COMPACT|JSON_ENSURE_ASCII|
+    char *js_s = NULL;
+
+#ifdef HAVE_LIBHIREDIS
+    if (file_ctx->type == LOGFILE_TYPE_REDIS) {
+        json_object_set_new(js, "host",
+                            json_string(file_ctx->redis_setup.sensor_name));
+    }
+#endif
+
+    js_s = json_dumps(js,
+            JSON_PRESERVE_ORDER|JSON_COMPACT|JSON_ENSURE_ASCII|
 #ifdef JSON_ESCAPE_SLASH
                             JSON_ESCAPE_SLASH
 #else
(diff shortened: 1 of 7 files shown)


src/output-json.c
```diff
@@ -509,6 +515,25 @@ OutputCtx *OutputJsonInitCtx(ConfNode *conf)
...
+#if HAVE_LIBHIREDIS
+        else if (json_ctx->json_out == LOGFILE_TYPE_REDIS) {
+            ConfNode *redis_node = ConfNodeLookupChild(conf, "redis");
+            const char *sensor_name = ConfNodeLookupChildValue(conf, "sensor-name");
+            if (!sensor_name) {
+                char hostname[1024];
+                gethostname(hostname, 1023);
+                sensor_name = hostname;
+            }
+            json_ctx->file_ctx->redis_setup.sensor_name = SCStrdup(sensor_name);
```
inliniac: error check?

src/util-logopenfile.c
```diff
@@ -262,6 +262,132 @@ int SCConfLogReopen(LogFileCtx *log_ctx)
     return 0;
 }
 
+
+#if HAVE_LIBHIREDIS
```
inliniac: should this be #ifdef?

src/util-logopenfile.c
```diff
@@ -311,3 +450,100 @@ int LogFileFreeCtx(LogFileCtx *lf_ctx)
 
     SCReturnInt(1);
 }
+
+int LogFileWrite(LogFileCtx *file_ctx, MemBuffer *buffer, char *string, size_t string_len)
+{
+    SCMutexLock(&file_ctx->fp_mutex);
+    if (file_ctx->type == LOGFILE_TYPE_SYSLOG) {
+        syslog(file_ctx->syslog_setup.alert_syslog_level, "%s", string);
```
inliniac: wonder if this needs the lock?

src/util-logopenfile.c
```diff
@@ -311,3 +450,100 @@ int LogFileFreeCtx(LogFileCtx *lf_ctx)
...
+    } else if (file_ctx->type == LOGFILE_TYPE_FILE ||
+               file_ctx->type == LOGFILE_TYPE_UNIX_DGRAM ||
+               file_ctx->type == LOGFILE_TYPE_UNIX_STREAM)
+    {
+        MemBufferWriteString(buffer, "%s\n", string);
+        file_ctx->Write((const char *)MEMBUFFER_BUFFER(buffer),
+            MEMBUFFER_OFFSET(buffer), file_ctx);
+    }
+#if HAVE_LIBHIREDIS
```
inliniac: Think it would be nicer to move this redis block into a util func

inliniac: that would also make the locking more clear, as that could just live outside of the redis call

inliniac:
Nice work Eric!

src/util-logopenfile.c
regit: You mean for syslog ?

src/output-json.c
regit: Oups.

src/util-logopenfile.c
inliniac: ya

PR #1507, opened 2015-05-26
Redisdead v1.6

regit:
Same code as #1506 with the update discussed in this previous PR. Almost no rewrite in the existing patches but some supplementary ones.

An example YAML EVE configuration could look like:
```
   - eve-log:
        enabled: yes
        filetype: redis
        redis:
          server: 127.0.0.1
          port: 6379
          mode: list ## possible values: list (default), channel
          key: suricata ## key or channel to use (default to suricata)
          pipelining:
            enabled: yes ## set enable to yes to enable query pipelining
            batch-size: 10 ## number of entry to keep in buffer
```

The pipelining is meant for busy network as suricata will only send events when batch-size events have been queued.

To fetch data generated by this input yu can use the following logstash input configuration:
```
input {
       redis {
          host => "127.0.0.1"
          codec => json
          key => "suricata"
          data_type => "list"
          type => "json-log" #optional
       }
  }
```

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/74
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/72

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ compiler:
   - gcc
   - clang
 # Change this to your needs
-script: sh autogen.sh && ./configure --enable-nfqueue --enable-unittests && make && make check
+script: sh autogen.sh && ./configure --enable-nfqueue --enable-unittests --enable-hiredis && make && make check
 before_install:
   - sudo add-apt-repository -y ppa:npalix/coccinelle
   - sudo apt-get update -qq
-  - sudo apt-get install -y libpcre3 libpcre3-dbg libpcre3-dev build-essential autoconf automake libtool libpcap-dev libnet1-dev libyaml-0-2 libyaml-dev zlib1g zlib1g-dev libcap-ng-dev libcap-ng0 make libmagic-dev libnetfilter-queue-dev libnetfilter-queue1 libnfnetlink-dev libnfnetlink0 coccinelle
+  - sudo apt-get install -y libpcre3 libpcre3-dbg libpcre3-dev build-essential autoconf automake libtool libpcap-dev libnet1-dev libyaml-0-2 libyaml-dev zlib1g zlib1g-dev libcap-ng-dev libcap-ng0 make libmagic-dev libnetfilter-queue-dev libnetfilter-queue1 libnfnetlink-dev libnfnetlink0 coccinelle libjansson-dev libhiredis-dev
   - ./qa/travis-libhtp.sh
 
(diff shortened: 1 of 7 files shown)


inliniac:
Can you give a general explanation of a suricata->redis->elk setup? I don't really understand yet how the setup should function.

Also, can you explain the batch mode?

regit:
Hi @inliniac 

Regarding the setup of the logging chain:
- Suricata log to redis
- Logstash read redis
- Logstash write to Elasticsearch

So basically, the conf snippet in the PR are what you need to get Suricata events treated by Logstash.

Regarding the batch mode, it is currently dedicated to people that have a continuous flow of event. The algorithm is simple:
- we append the event to the command buffer
- if we have batch-size events we ask for the replies which trigger the real request to redis

So instead of sending TCP data for each event we are sending a bigger alout of data every batch-size events. This method is not containing any flush so Suricata will only write every batch-size events.

inliniac:
On 05/29/2015 08:37 PM, Eric Leblond wrote:
> Regarding the setup of the logging chain:
> 
>   * Suricata log to redis
>   * Logstash read redis
>   * Logstash write to Elasticsearch
> 
> So basically, the conf snippet in the PR are what you need to get
> Suricata events treated by Logstash.

So is there a separate redis 'server' process in between Suricata and
Logstash?

> Regarding the batch mode, it is currently dedicated to people that have
> a continuous flow of event. The algorithm is simple:
> 
>   * we append the event to the command buffer
>   * if we have batch-size events we ask for the replies which trigger
>     the real request to redis
> 
> So instead of sending TCP data for each event we are sending a bigger
> alout of data every batch-size events. This method is not containing any
> flush so Suricata will only write every batch-size events.

What happens when Suricata shuts down while there are still queued events?

regit:
Yes, there is a redis server. A basic package install is enough to get a working setup.

I did not code the flushing of queued events. Maybe we could do it a in the Free function. But even if it will be better to do it, I think that regarding the target setup it will be the same for the user:
- batch mode is targeted to live trafic
- batch-count should remain below the TCP window size to minimize packet exchanges (so we have only a few events queued)
So it will be for the user almost the same as if he has shutdown  suricata a few ms sooner.

If you consider this is not ok to loose those few events, I can add the needed code.

inliniac:
Thanks for the explanation. Yeah I think it's not acceptable to loose events. Perhaps it's as easy as flushing out the queue at shutdown, but I think it's necessary to cover this case.

regit:
OK, closing the PR and doing a new one with the flush at exit code.

PR #1236, opened 2014-12-04
[preview] stats log json v0.5

inliniac:
Rebased https://github.com/inliniac/suricata/pull/1202

Still not happy with the format.

Prscript:
- PR build: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/67
- PR pcaps: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/67

PR #1508, opened 2015-05-27
counters rewrite + json stats v14.9

inliniac:
Complete (and messy*) rewrite/refactor of the stats API as it's now called.
* Introduces concept of 'global' counters for use with 'memuse'
* Output by default shows aggregated stats for all the threads (so a total)
* Optionally, per thread counters can also be printed.
* The old grouping by 'thread module' is removed.
* JSON output with Eve support was added. Based on @decanio 's work in #1236. Supports both global (default) and per thread output. Or both at the same time.

Includes https://github.com/inliniac/suricata/pull/1236

Tickets:
https://redmine.openinfosecfoundation.org/issues/1374
https://redmine.openinfosecfoundation.org/issues/1228
https://redmine.openinfosecfoundation.org/issues/1210

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/226
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/228

*) there really is no clean way to do something like this, so I've decided to just keep all the commits. This way the whole process can be followed.


inliniac:
One thing I forgot to mention: stats over unix socket are not implemented. It would probably make sense to unify the JSON output with this feature. Thoughts @regit?

regit:
Yes, unification really make senses here.

PR #1488, opened 2015-05-15
DNP3 Support - v4.

jasonish:
DNP3 support is reach a state of completion.  In this current state it should be ready for review and testing.

Not all objects are decoded, but that is a matter of going through the documentation and filling out the object size table.

The new keywords are documented here:
https://redmine.openinfosecfoundation.org/projects/suricata/wiki/DNP3_keyword

Example output of the DNP3 JSON logging:
https://gist.github.com/jasonish/71d4df870977b68c5db9

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/91
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/92

PR #1510, opened 2015-05-27
DNP3 Support - v5

jasonish:
Replaces #1488.

Updates from previous PR:
- All static size objects decoded.
- Drop payload logging for now.

DNP3 support includes:
- Application layer decoder.
- Snort compatible keywords:
- Request / response logging.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/98
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/99


rules/dnp3-events.rules
```diff
@@ -0,0 +1,17 @@
+# Flooded.
+alert dnp3 any any -> any any (msg:"SURICATA DNP3 Request flood detected"; \
+      app-layer-event:dnp3.flooded; sid:5000000; rev:1;)
```
inliniac: can you use the 2260000 range here?

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+    uint8_t qualifier;
+} __attribute__((packed)) DNP3ObjHeader;
+
+/**
+ * A table to hold object sizes for all object groups and their
+ * variations.
+ *
+ * Sizes are in bits.
+ */
+static int GROUP_VAR_SIZE_MAP[DNP3_OBJ_GROUP_MAX][DNP3_OBJ_VAR_MAX];
```
inliniac: uppercase naming is reserved for macro's

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+    0x0a6b, 0x3c35, 0x66d7, 0x5089, 0xd313, 0xe54d, 0xbfaf, 0x89f1,
+    0x4789, 0x71d7, 0x2b35, 0x1d6b, 0x9ef1, 0xa8af, 0xf24d, 0xc413,
+    0xb800, 0x8e5e, 0xd4bc, 0xe2e2, 0x6178, 0x5726, 0x0dc4, 0x3b9a,
+    0xdc4d, 0xea13, 0xb0f1, 0x86af, 0x0535, 0x336b, 0x6989, 0x5fd7,
+    0x23c4, 0x159a, 0x4f78, 0x7926, 0xfabc, 0xcce2, 0x9600, 0xa05e,
+    0x6e26, 0x5878, 0x029a, 0x34c4, 0xb75e, 0x8100, 0xdbe2, 0xedbc,
+    0x91af, 0xa7f1, 0xfd13, 0xcb4d, 0x48d7, 0x7e89, 0x246b, 0x1235
+};
+
+static uint16_t DNP3ComputeCRC(uint8_t *buf, uint32_t len)
```
inliniac: buf can be const?

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+
+/**
+ * \brief Check the CRC of a block.
+ *
+ * \param block The block of data with CRC to be checked.
+ * \param len The size of the data block.
+ *
+ * \retval 1 if CRC is OK, otherwise 0.
+ */
+static int DNP3CheckCRC(uint8_t *block, uint32_t len)
```
inliniac: block can be const

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+    uint16_t crc = DNP3ComputeCRC(block, len - DNP3_CRC_LEN);
+    if (((crc & 0xff) == block[crc_offset]) &&
+        ((crc >> 8) == block[crc_offset + 1])) {
+        return 1;
+    }
+
+    return 0;
+}
+
+static int DNP3CheckLinkHeaderCRC(DNP3LinkHeader *header)
```
inliniac: header can be const

Please review all function calls for making arguments const. It helps in review and it helps the compiler.

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+        /* The remainder must be at least one byte plus a CRC. */
+        if (rem < 3) {
+            return NULL;
+        }
+        len += rem + DNP3_CRC_LEN;
+    }
+
+    /* Remove one byte for the transport header which won't be
+     * included in the output. */
+    len--;
```
inliniac: if function is called with input_len 0, we'd underflow here

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+    int offset = 0, block_size;
+    while ((uint32_t)offset < input_len) {
+        if (input_len - offset > DNP3_BLOCK_SIZE + DNP3_CRC_LEN) {
+            block_size = DNP3_BLOCK_SIZE + DNP3_CRC_LEN;
+        }
+        else {
+            block_size = input_len - offset;
+        }
+
+        /* For the first block we trim the leading transport header. */
```
inliniac: Where is it checked that block_size is big enough for the subtractions?

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+    SCEnter();
+    DNP3State *dnp3;
+
+    dnp3 = (DNP3State *)SCCalloc(1, sizeof(DNP3State));
+    if (unlikely(dnp3 == NULL)) {
+        return NULL;
+    }
+    TAILQ_INIT(&dnp3->tx_list);
+
+    dnp3->sessions = HashListTableInit(256, DNP3SessionHash,
```
inliniac: This will lead to a lot of memory use. The struct itself is 48 bytes, but the 256 means we're doing 256 * 8 bytes of allocations. If the number of session won't be high this isn't a problem necessarily, but it's something to keep in mind.

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+ * \brief Calculate the length of a frame with CRCs added.
+ */
+static uint32_t DNP3CalculateUnassembledLength(uint8_t length)
+{
+    uint32_t frame_len = 0;
+    int rem;
+
+    /* Subtract the 5 bytes of the header that are included in the
+     * length. */
+    length -= DNP3_LINK_HDR_LEN;
```
inliniac: is there an underflow risk?

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+            SCLogInfo(
+                "Insufficient data: not enough data for object header (%d).",
+                len);
+            SCLogInfo("%02x", buf[2]);
+            goto done;
+        }
+        DNP3ObjHeader *header = (DNP3ObjHeader *)buf;
+        offset += sizeof(DNP3ObjHeader);
+
+        DNP3Object *object = SCCalloc(1, sizeof(*object));
```
inliniac: if this fails object will be NULL

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+        }
+
+        SCLogDebug("Group: 0x%02x (%d); Variation: 0x%02x (%d); "
+            "Qualifier: 0x%02x (%d)",
+            header->group, header->group,
+            header->variation, header->variation,
+            header->qualifier, header->qualifier);
+        uint8_t prefix_code = DNP3_OBJ_PREFIX(header->qualifier);
+        uint8_t range_code = DNP3_OBJ_RANGE(header->qualifier);
+        object->prefix = DNP3_OBJ_PREFIX(header->qualifier);
```
inliniac: then this is a null-deref

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+        int prefix_len = DNP3_OBJ_PREFIX_CODE_MAP[prefix_code];
+        if (prefix_len < 0) {
+            goto unknown;
+        }
+
+        uint32_t count = 0;
+
+        /* IEEE 1815-2012, Table 4-5. */
+        switch (range_code) {
+            case 0x00: {
```
inliniac: does it make sense to use named values here?

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+                uint8_t start = buf[offset++];
```
inliniac: Do we properly check the size of buf? If so, where?

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+                SCLogInfo("Range code 0x%02x is reserved.", range_code);
+                goto done;
+        }
+
+        if (count == 0) {
+            goto next;
+        }
+
+        /* Get the size of the object - in bits. */
+        int object_size = GROUP_VAR_SIZE_MAP[header->group][header->variation];
```
inliniac: group and variation are both uint8_t right? Since GROUP_VAR_SIZE_MAP is [256][256], it should be safe

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+        if (offset + object_len > len) {
+            SCLogInfo("Insufficient data to decode object %d:%d", header->group,
+                header->variation);
+            goto not_enough_data;
+        }
+
+        offset += object_len;
+    next:
+
+        object->data = SCCalloc(1, offset);
```
inliniac: object can still be null here

src/app-layer-dnp3.c
```diff
@@ -0,0 +1,2616 @@
...
+            break;
+    }
+
+    /* Reassemble the request buffer. */
+    tx->request_buffer = DNP3ReassembleApplicationLayer(
+        input + th_offset, // + sizeof(DNP3TransportHeader),
+            input_len - th_offset, // - sizeof(DNP3TransportHeader),
+            &tx->request_buffer, &tx->request_buffer_len);
+    tx->request_done = 1;
+
```
inliniac: error handling seems missing

rules/dnp3-events.rules
jasonish: I see that modbus uses 225XXXX, 226XXXX is the app-layer stuff.  Should DNP3 be with the app-layer in 226XXXX? Or something like 227XXXX?

Another thought - should the rules stay in their own file are get merged into app-layer-events.rules?

src/app-layer-dnp3.c
jasonish: Addressed.  Will be in next PR.

src/app-layer-dnp3.c
jasonish: Yes it can.  Adressed in next PR.

src/app-layer-dnp3.c
jasonish: Addressed in next PR.

rules/dnp3-events.rules
inliniac: Right then it should be 227XXXXX

jasonish: Will be addressed in next PR.

src/app-layer-dnp3.c
jasonish: Yes, I think this is the ideal candidate for a malformed frame event.  We should only attempt to re-assemble "user data" frames, and a "user data" frame without data is not valid.

src/app-layer-dnp3.c
jasonish: So I have not actually seen more than one DNP3 address pair per TCP session, but support is there in the protocol.  I expect it to be just a few DNP3 sessions over a TCP session, so maybe I could lower this value, to 8?

src/app-layer-dnp3.c
jasonish: Yes, thats correct.  With the further object decoding, this will likely become a lookup function though.

src/app-layer-dnp3.c
jasonish: In next PR, the return code of DNP3ReassembleApplicationLayer will be checked, and a malformed event will be raised as needed.

src/app-layer-dnp3.c
jasonish: Inverted the logic, bail if its NULL, continue otherwise.

src/app-layer-dnp3.c
jasonish: Addressed by failing earlier on NULL.

src/app-layer-dnp3.c
jasonish: Normally I'd say yes, but even in the spec, the hex value is the "name" of the field, unless we wanted to do something like RANGE_CODE_1, ..., RANGE_CODE_9?

src/app-layer-dnp3.c
jasonish: Addressed by failing earlier on NULL.

src/app-layer-dnp3.c
jasonish: Length is now checked for 0 prior to subtraction, then checked for minimum valid length after subtraction.

src/app-layer-dnp3.c
jasonish: Addressed in next PR.

src/app-layer-dnp3.c
jasonish: The value is verified earlier, when the check for enough data for a user data frame, but I'll add the check here, and validate the response to callers of this function.

src/app-layer-dnp3.c
jasonish: Not, it is not checked as we go.  Will be in next PR.

PR #1511, opened 2015-05-27
logging: fix modules ordering during logging

regit:
Here's a fixing invalid value for some keys in the EVE fileinfo output.

With the previous code the order of the logging modules in the
YAML were determining which module was run first. This was not
wished and a consequences was that the EVE fileinfo module was
not correctly displaying the key 'stored' because it was
depending on a flag set alter by the filestore module.

This patch adds a priority file to the TmModule structure. The
higher the priority is set, the sooner the module is run in the
logging process. The RunModeOutput structure has also been
updated to contain the name of the original TmModule. Thus allowing
to define a priority for a RunModeOutput.

Currently only the filestore module has a priority set. The rest of them is
set to the default value of zero.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/75
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/73


inliniac:
Merged through #1523, thanks!

PR #1513, opened 2015-05-28
hsbd v1.1: inspect buffer depending on the engine mode

glongo:

This addresses the comments in the last PR: https://github.com/inliniac/suricata/pull/1505

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/36
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/35

### DIFF
--- a/src/detect-engine-hsbd.c
+++ b/src/detect-engine-hsbd.c
@@ -89,6 +89,127 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
     return 0;
 }
 
+static void HSBDGetBufferForTXInIDSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    int first = 1;
+    while (cur != NULL) {
+        /* see if we can filter out chunks */
+        if (htud->response_body.body_inspected > 0) {
+            if (cur->stream_offset < htud->response_body.body_inspected) {
+                if ((htud->response_body.body_inspected - cur->stream_offset) > htp_state->cfg->response_inspect_window) {
+                    cur = cur->next;
+                    continue;
+                } else {
+                    /* include this one */
+                }
+            } else {
+                /* include this one */
+            }
+        }
+
+        if (first) {
+            det_ctx->hsbd[index].offset = cur->stream_offset;
+            first = 0;
+        }
+
+        /* see if we need to grow the buffer */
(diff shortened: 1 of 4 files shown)


src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,127 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+                prev = cur;
+                cur = cur->next;
+                continue;
+            } else {
+                /* include this one */
+            }
+        }
+
+        if (first) {
+            det_ctx->hsbd[index].offset = cur->stream_offset;
```
inliniac: this should account for the offset as well I think

inliniac: I mean the offset into the first chunk as you use below

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,127 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+    while (cur != NULL) {
+        /* see if we need to grow the buffer */
+        if (cur->len > window_size) {
+            extra_size =  window_size * 0.25;
+            window_size = cur->len;
+        } else if (cur->len == window_size) {
+            window_size = window_size * 1.25;
+        }
+
+        if (det_ctx->hsbd[index].buffer == NULL) {
```
inliniac: if only the 2nd chunk is >= window_size, expanding window_size, we won't grow the buffer because det_ctx->hsbd[index].buffer != NULL, correct?

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,127 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+                /* include this one */
+            }
+        }
+
+        if (first) {
+            det_ctx->hsbd[index].offset = cur->stream_offset;
+            first = 0;
+        }
+
+        if (prev != NULL && cur->len <= window_size) {
```
inliniac: I think this means you're copying the chunk before the current on in as well. This definitely needs an explanation in a comment.

src/suricata.c
```diff
@@ -257,6 +260,27 @@ void EngineModeSetIDS(void)
     g_engine_mode = ENGINE_MODE_IDS;
 }
 
+void InlineModeSetIDS(void)
```
inliniac: I think this is too broad, lets make something specific for the HTTP body tracking.

Perhaps a setting in the detection engine ctx.

PR #1515, opened 2015-05-29
hsbd v1.2: inspect buffer depending on the engine mode

glongo:
Currently, data is buffered up to response-body-minimal size and response-body-inspect-window before being inspected.
With this, in IPS mode, inspect data as it comes in up.

This addresses the comments in the last PR: https://github.com/inliniac/suricata/pull/1513

PR #1516, opened 2015-05-29
hsbd v1.3: inspect buffer depending on the engine mode

glongo:

This addresses the comments in the last PR: #1515 

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/38
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/37

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -2467,6 +2467,21 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
                 exit(EXIT_FAILURE);
             }
             cfg_prec->randomize_range = range;
+        } else if (strcasecmp("http-body-inline", p->name) == 0) {
+            if (!strcmp("yes", p->val)) {
+                cfg_prec->http_body_inline = 1;
+            } else if (!strcmp("no", p->val)) {
+                cfg_prec->http_body_inline = 0;
+            } else {
+                if (strcmp("auto", p->val) != 0) {
+                    WarnInvalidConfEntry("http_body_inline", "%s", "auto");
+                }
+                if (EngineModeIsIPS()) {
+                    cfg_prec->http_body_inline = 1;
+                } else {
+                    cfg_prec->http_body_inline = 0;
+                }
+            }
         } else {
             SCLogWarning(SC_ERR_UNKNOWN_VALUE, "LIBHTP Ignoring unknown "
                          "default config: %s", p->name);
--- a/src/detect-engine-hsbd.c
+++ b/src/detect-engine-hsbd.c
@@ -89,6 +89,117 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
     return 0;
 }
 
+static void HSBDGetBufferForTXInIDSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    int first = 1;
+    while (cur != NULL) {
+        /* see if we can filter out chunks */
+        if (htud->response_body.body_inspected > 0) {
+            if (cur->stream_offset < htud->response_body.body_inspected) {
+                if ((htud->response_body.body_inspected - cur->stream_offset) > htp_state->cfg->response_inspect_window) {
+                    cur = cur->next;
+                    continue;
+                } else {
+                    /* include this one */
+                }
+            } else {
+                /* include this one */
+            }
+        }
+
+        if (first) {
+            det_ctx->hsbd[index].offset = cur->stream_offset;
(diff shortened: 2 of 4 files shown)


inliniac:
This commit: "hsbd: inspect buffer depending on the engine mode" will need a proper message, explaining the goal and high level implementation.

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,117 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+
+        /*
+         * Copy a part of the previous buffer before the current chunk on.
+         * For example, let's have the chunks [123], [456].
+         * When the current chunk is [456], this copy [3] before.
+         */
+        if (det_ctx->hsbd[index].buffer_len + cur->len > window_size) {
+            uint32_t offset = (det_ctx->hsbd[index].buffer_len + cur->len) - (window_size + extra_size);
+            /* we use memmove to avoid memory overlap */
+            memmove(det_ctx->hsbd[index].buffer, det_ctx->hsbd[index].buffer + offset, det_ctx->hsbd[index].buffer_len - offset);
```
inliniac: Why is memmove needed?

glongo: With memcpy, DetectEngineHttpServerBodyFileDataTest07 produces this issue:
==31217== ERROR: AddressSanitizer: memcpy-param-overlap: memory ranges [0x60040001b3b0,0x60040001b3b2) and [0x60040001b3b1, 0x60040001b3b3) overlap
memmove prevents memory overlap, so fixes it.

inliniac: Can you add an explanation (to the code) of why the memmove step is necessary? What it's purpose is?

inliniac:
In IDS mode we respect a response-body-minimal-inspect-size setting (defaults to 32kb). IPS mode doesn't seem to have this. Think we need this as well.

src/app-layer-htp.c
```diff
@@ -2467,6 +2467,21 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
                 exit(EXIT_FAILURE);
             }
             cfg_prec->randomize_range = range;
+        } else if (strcasecmp("http-body-inline", p->name) == 0) {
+            if (!strcmp("yes", p->val)) {
```
inliniac: Use ConfValIsTrue

src/app-layer-htp.c
```diff
@@ -2467,6 +2467,21 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
...
+                cfg_prec->http_body_inline = 1;
+            } else if (!strcmp("no", p->val)) {
```
inliniac: Use ConfValIsFalse

glongo:
Do we need only response-body-minimal-inspect-size setting, or even response-body-limit?

inliniac:
Only response-body-minimal-inspect-size. The response-body-limit parameter is enforced elsewhere.

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,117 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+            first = 0;
+        }
+
+        /*
+         * Copy a part of the previous buffer before the current chunk on.
+         * For example, let's have the chunks [123], [456].
+         * When the current chunk is [456], this copy [3] before.
+         */
+        if (det_ctx->hsbd[index].buffer_len + cur->len > window_size) {
+            uint32_t offset = (det_ctx->hsbd[index].buffer_len + cur->len) - (window_size + extra_size);
```
inliniac: This doesn't look correct. The extra size is not part of the 'if' condition above. Thus window_size + extra_size may be bigger than det_ctx->hsbd[index].buffer_len + cur->len, leading to a negative value, which in a uint will lead to a very large value.

glongo: See the following example:
window=4, chunks [ab][cd][12345678]
when cur=[12345678], the buffer is [abcd].
Without 'extra_size', the offset is 4, it means I can't copy a part of buffer ('d').

src/app-layer-htp.c
glongo: ack

src/app-layer-htp.c
glongo: ack

src/detect-engine-hsbd.c
inliniac: You're missing the point. Offset underflows (I have confirmed this), this leads to out of bounds reads/writes. If the logic depends on that, it's simply wrong.

glongo: ack

PR #1517, opened 2015-05-29
hsbd v1.4: inspect buffer depending on the engine mode

glongo:

This addresses the comments in the last PR: #1516

### DIFF
--- a/src/detect-engine-hsbd.c
+++ b/src/detect-engine-hsbd.c
@@ -89,6 +89,117 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
     return 0;
 }
 
+static void HSBDGetBufferForTXInIDSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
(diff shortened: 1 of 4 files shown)


src/detect-engine-hsbd.c
```diff
@@ -158,63 +269,28 @@ static uint8_t *DetectEngineHSBDGetBufferForTX(htp_tx_t *tx, uint64_t tx_id,
...
-                SCFree(det_ctx->hsbd[index].buffer);
-                det_ctx->hsbd[index].buffer = NULL;
-                det_ctx->hsbd[index].buffer_size = 0;
-                det_ctx->hsbd[index].buffer_len = 0;
-                goto end;
-            }
-            det_ctx->hsbd[index].buffer = ptmp;
+        HSBDGetBufferForTXInIDSMode(det_ctx, htp_state, cur, htud, index);
+    } else {
+        if (htud->response_body.content_len_so_far < htp_state->cfg->response_inspect_min_size) {
```
inliniac: this isn't correct, we shouldn't wait until we have this amount of data. Instead, we need to make our initial window size response_inspect_min_size. Then after this has been reached, the rest of the window will be response_inspect_window

PR #1521, opened 2015-05-29
Redisdead v1.7

regit:
A rework of #1507. It adds a patch fixing the flush issue discussed in the previous PR. Other change is a white space fix in a patch.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/78
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/76

PR #1522, opened 2015-05-30
Redisdead v1.8

regit:
Update of #1521:
- Fix coccinelle check
- Add sensor-name in yaml

I'm almost ok with @jasonish suggestion on updating the other eve event. But we need to discuss the details a bit so maybe a later PR.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/79
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/77

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -267,8 +267,18 @@ int SCConfLogReopen(LogFileCtx *log_ctx)
 
 static void SCLogFileCloseRedis(LogFileCtx *log_ctx)
 {
-    if (log_ctx->redis)
+    if (log_ctx->redis) {
+        redisReply *reply;
+        int i, count;
+        count = SC_ATOMIC_GET(log_ctx->redis_setup.batch_count);
+        for (i = 0; i < count; i++) {
+            redisGetReply(log_ctx->redis, (void **)&reply);
+            if (reply)
+                freeReplyObject(reply);
+        }
         redisFree(log_ctx->redis);
+        log_ctx->redis = NULL;
+    }
     log_ctx->redis_setup.tried = 0;
     SC_ATOMIC_SET(log_ctx->redis_setup.batch_count, 0);
 }
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -303,7 +303,6 @@ int SCConfLogOpenRedis(ConfNode *redis_node, LogFileCtx *log_ctx)
         exit(EXIT_FAILURE);
     }
 
-    log_ctx->redis_setup.batch_timeout = 0;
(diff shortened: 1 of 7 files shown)


src/util-logopenfile.c
```diff
@@ -311,3 +463,108 @@ int LogFileFreeCtx(LogFileCtx *lf_ctx)
...
     SCReturnInt(1);
 }
+
+#ifdef HAVE_LIBHIREDIS
+static int  LogFileWriteRedis(LogFileCtx *file_ctx, char *string, size_t string_len)
+{
+    if (file_ctx->redis == NULL) {
+        SCConfLogReopenRedis(file_ctx);
+        if (file_ctx->redis == NULL) {
+            SCMutexUnlock(&file_ctx->fp_mutex);
```
inliniac: where is the lock?

regit: Rhaaa, I should not code at 1am after a big day of work.

src/util-logopenfile.c
```diff
@@ -311,3 +463,108 @@ int LogFileFreeCtx(LogFileCtx *lf_ctx)
...
+            SCLogInfo("Reconnected to redis server");
+        }
+    }
+    /* TODO go async here ? */
+    if (file_ctx->redis_setup.batch_size) {
+        redisAppendCommand(file_ctx->redis, "%s %s %s",
+                file_ctx->redis_setup.command,
+                file_ctx->redis_setup.key,
+                string);
+        if (SC_ATOMIC_CAS(&file_ctx->redis_setup.batch_count, file_ctx->redis_setup.batch_size, 0)) {
```
inliniac: I don't understand this CAS call. Actually, I think we have the file_ctx lock here right? So why do we need atomics?

inliniac:
Can "util-logopenfile: suppress unused batch timeout" be squashed into the commit that adds the code?

src/util-logopenfile.c
regit: OK, fixed.

src/util-logopenfile.c
regit: Yes, my initial code was without that so I needed a lock but as hiredis is not thread safe I added back the lock. So yes, it is mainly useless now. I will try without it.

regit:
Humm, yes :) easier than thought and done.

inliniac:
On 06/01/2015 10:59 PM, Eric Leblond wrote:
> Humm, yes :) easier than thought and done.

Often in such cases it's easiest to edit the original commit instead of
squashing another into it.

regit:
I was lucky on this one: direct rebase was working :)

PR #1525, opened 2015-05-30
hsbd v1.6: inspect buffer depending on the engine mode

glongo:
This addresses a memory corruption buf which occur in the last PR. [https://github.com/inliniac/suricata/pull/1519]

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/39
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/38

### DIFF
--- a/src/detect-engine-hsbd.c
+++ b/src/detect-engine-hsbd.c
@@ -89,6 +89,123 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
     return 0;
 }
 
+static void HSBDGetBufferForTXInIDSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    int first = 1;
+    while (cur != NULL) {
+        /* see if we can filter out chunks */
+        if (htud->response_body.body_inspected > 0) {
+            if (cur->stream_offset < htud->response_body.body_inspected) {
+                if ((htud->response_body.body_inspected - cur->stream_offset) > htp_state->cfg->response_inspect_window) {
+                    cur = cur->next;
+                    continue;
+                } else {
+                    /* include this one */
+                }
+            } else {
+                /* include this one */
+            }
+        }
+
+        if (first) {
+            det_ctx->hsbd[index].offset = cur->stream_offset;
+            first = 0;
+        }
+
+        /* see if we need to grow the buffer */
+        if (det_ctx->hsbd[index].buffer == NULL || (det_ctx->hsbd[index].buffer_len + cur->len) > det_ctx->hsbd[index].buffer_size) {
+            void *ptmp;
+            det_ctx->hsbd[index].buffer_size += cur->len * 2;
+
+            if ((ptmp = SCRealloc(det_ctx->hsbd[index].buffer, det_ctx->hsbd[index].buffer_size)) == NULL) {
+                SCFree(det_ctx->hsbd[index].buffer);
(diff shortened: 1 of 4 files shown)


src/detect-engine-hsbd.c
```diff
@@ -3631,6 +3709,1227 @@ static int DetectEngineHttpServerBodyFileDataTest03(void)
...
+
+    char input[] = "\
+%YAML 1.1\n\
+---\n\
+libhtp:\n\
+\n\
+  default-config:\n\
+\n\
+    http-body-inline: yes\n\
+    response-body-minimal-inspect-size: 1\n\
```
inliniac: these tests are odd: response-body-minimal-inspect-size will normally be quite a bit bigger than response-body-inspect-window, but all tests use the opposite.

inliniac:
Can you run prscript as well?

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,123 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+    htud->response_body.body_inspected = htud->response_body.last->stream_offset + htud->response_body.last->len;
+}
+
+static void HSBDGetBufferForTXInIPSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
+                                        HtpTxUserData *htud, int index)
+{
+    uint32_t window_size = 0;
+    uint32_t extra_size = 0;
+    int resize = 0;
```
inliniac: I think we are resizing too much (every time this func is called). It can be done much simpler: if you actually set the buffer_size like the IDS func does, you can simply check for the case where the size we need is > buffer_size, and only resize then. It's not a problem if the buffer size stays big.

src/detect-engine-hsbd.c
```diff
@@ -89,6 +89,123 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
...
+            det_ctx->hsbd[index].offset = cur->stream_offset;
+            first = 0;
+        }
+
+        /* see if we need to grow the buffer */
+        if (det_ctx->hsbd[index].buffer == NULL || (det_ctx->hsbd[index].buffer_len + cur->len) > det_ctx->hsbd[index].buffer_size) {
+            void *ptmp;
+            det_ctx->hsbd[index].buffer_size += cur->len * 2;
+
+            if ((ptmp = SCRealloc(det_ctx->hsbd[index].buffer, det_ctx->hsbd[index].buffer_size)) == NULL) {
```
glongo: do you mean I should do the same as here?

inliniac: Not exactly, but more like it.

src/detect-engine-hsbd.c
glongo: Does make sense to remove response-body-minimal-inspect-size?

inliniac: How would it be tested then?

glongo: the value is loaded from yaml:
Test DetectEngineHttpServerBodyFileDataTest04                     : [31276] 1/6/2015 -- 13:56:29 - (app-layer-htp.c:2235) <Info> (HTPConfigSetDefaultsPhase2) -- 'default' server has 'request-body-minimal-inspect-size' set to 34116 and 'request-body-inspect-window' set to 3973 after randomization.

inliniac: if the default value is used with the small bodies used in the test, the regular window size setting won't be used. You need to test both.

PR #1526, opened 2015-05-31
hsbd v1.7: inspect buffer depending on the engine mode

glongo:
This should address the slow issue, UTs aren't fixed yet.
Last PR: https://github.com/inliniac/suricata/pull/1525


inliniac:
Another case that is not working well (observed):
- total body ~40kb
- chunks 8kb (so bigger than window_size of ~4kb)

In this case the window is resized to ~11kb, leading to cutting off the first part of the data.

inliniac:
Will share pcap+rule offline

PR #1530, opened 2015-06-03
Add Feature #1454. Generic eve-log prefix support v3

zrasmor:
Support for adding a prefix to eve-log entries. Targeted use case is CEE/Lumberjack rsyslog compatibility ("@cee: " cookie), but any string can be specified as a prefix.

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -341,13 +341,29 @@ int OutputJSONBuffer(json_t *js, LogFileCtx *file_ctx, MemBuffer *buffer)
         return TM_ECODE_OK;
 
     SCMutexLock(&file_ctx->fp_mutex);
-    if (file_ctx->type == LOGFILE_TYPE_SYSLOG) {
-        syslog(alert_syslog_level, "%s", js_s);
-    } else if (file_ctx->type == LOGFILE_TYPE_FILE ||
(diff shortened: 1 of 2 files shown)


src/output-json.c
```diff
@@ -456,6 +472,12 @@ OutputCtx *OutputJsonInitCtx(ConfNode *conf)
             }
         }
 
+        const char *prefix = ConfNodeLookupChildValue(conf, "prefix");
+        if (prefix != NULL)
+        {
+            json_ctx->file_ctx->prefix = SCStrdup(prefix);
```
inliniac: This is going to fail our qa (coccinelle check) because it's a memory allocation that isn't checked. It shouldn't lead to problems in this specific case, but I need a fix anyway :)

Also, I don't see where this memory is freed.

inliniac: I 'fix' would be to just check the result, and issue a warning or something.

inliniac:
As predicted:
<pre>
Testing cocci file: /home/travis/build/inliniac/suricata/qa/coccinelle/malloc-error-check.cocci
Structure malloced at /home/travis/build/inliniac/suricata/src/output-json.c:478 but error is not checked.
</pre>

PR #1532, opened 2015-06-03
Add Feature #1454. Generic eve-log prefix support v4

zrasmor:
Checking that memory is allocated successfully. I elected to log an error and exit because an inability to allocate memory indicates a greater issue, even if it's not fatal in this particular case.

Also, memory is freed on line 305 of util-logopenfile.c in the LogFileFreeCtx function, which is called from the OutputJsonDeInitCtx function.


inliniac:
Merged through #1540, thanks!

PR #1534, opened 2015-06-05
file_data: check for signature alproto and flow

glongo:
Currently the following rule can't be loaded:
alert tcp any any -> any 25 (msg:"SMTP file_data test"; flow:to_server,established; file_data; content:"abc";sid:1;)
and produces the error output:
"Can't use file_data with flow:to_server or from_client with http or smtp."

This checks if the alproto is not http in a signature,
so permits to use flow keyword also.

Issue reported by rmkml.

Prscript:

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/41
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/40

Last PR: https://github.com/inliniac/suricata/pull/1495

### DIFF
--- a/src/detect-file-data.c
+++ b/src/detect-file-data.c
@@ -84,7 +84,7 @@ static int DetectFiledataSetup (DetectEngineCtx *de_ctx, Signature *s, char *str
         return -1;
     }
 
-    if ((s->init_flags & SIG_FLAG_INIT_FLOW) &&
+    if (s->alproto == ALPROTO_HTTP && (s->init_flags & SIG_FLAG_INIT_FLOW) &&
         (s->flags & SIG_FLAG_TOSERVER) && !(s->flags & SIG_FLAG_TOCLIENT)) {
             SCLogError(SC_ERR_INVALID_SIGNATURE, "Can't use file_data with flow:to_server or from_client with http or smtp.");
             return -1;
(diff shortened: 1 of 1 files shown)


src/detect-file-data.c
```diff
@@ -84,7 +84,7 @@ static int DetectFiledataSetup (DetectEngineCtx *de_ctx, Signature *s, char *str
         return -1;
     }
 
-    if ((s->init_flags & SIG_FLAG_INIT_FLOW) &&
+    if (s->alproto == ALPROTO_HTTP && (s->init_flags & SIG_FLAG_INIT_FLOW) &&
```
inliniac: I think it would be good to add a similar test for SMTP as well.

glongo: why? If I add this test for SMTP, the rule will fail again

inliniac: For SMTP only to_server is valid, right? Where is this checked?

glongo: ok, I see. thanks

inliniac: On 06/05/2015 10:31 AM, Giuseppe Longo wrote:
> In src/detect-file-data.c
> <https://github.com/inliniac/suricata/pull/1534#discussion_r31795476>:
> 
>> @@ -84,7 +84,7 @@ static int DetectFiledataSetup (DetectEngineCtx *de_ctx, Signature *s, char *str
>>          return -1;
>>      }
>>  
>> -    if ((s->init_flags & SIG_FLAG_INIT_FLOW) &&
>> +    if (s->alproto == ALPROTO_HTTP && (s->init_flags & SIG_FLAG_INIT_FLOW) &&
> 
> ok, I see. thanks

This is also an indication of insufficient testing, so make sure you add
tests that deal with both valid and invalid rules.

PR #1544, opened 2015-06-11
json stats reorg - v1

jasonish:
Puts thread stats under a "threads" object.  Moves the "Totals" up a level out of the totals object.
Also converts uptime to an integer (seconds) instead of a string (less human readable, but easier to parse).

Example without threads:
https://gist.github.com/ffcc03d6fa540ebe597a

Example with threads:
https://gist.github.com/24b8e338118895f77346

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/102
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/103


inliniac:
Merged through #1549, thanks Jason.

PR #1545, opened 2015-06-11
http body ips v7

glongo:
This addresses the comments in https://github.com/inliniac/suricata/pull/1537

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/45
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/44

### DIFF
--- a/src/detect-engine-hsbd.c
+++ b/src/detect-engine-hsbd.c
@@ -89,6 +89,123 @@ static inline int HSBDCreateSpace(DetectEngineThreadCtx *det_ctx, uint16_t size)
     return 0;
 }
 
+static void HSBDGetBufferForTXInIDSMode(DetectEngineThreadCtx *det_ctx,
+                                        HtpState *htp_state, HtpBodyChunk *cur,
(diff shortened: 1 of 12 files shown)


src/detect-engine-hsbd.c
```diff
@@ -3631,6 +3847,308 @@ static int DetectEngineHttpServerBodyFileDataTest03(void)
...
+            "ab",
+            0, STREAM_TOCLIENT, 1, 1 },
+        {   (const uint8_t *)"cd",
+            0, STREAM_TOCLIENT, 2, 1 },
+        {   (const uint8_t *)"123456789",
+            0, STREAM_TOCLIENT, 3, 1 },
+        {   NULL, 0, 0, 0, 0},
+    };
+
+    const char *sigs[] = {
```
inliniac: I don't like that we're setting an expected result for just one sig per chunk, but we actually have 3 sigs in each test. I would prefer a method of passing expectations for each sig, perhaps in a separate array.

PR #1546, opened 2015-06-11
file-json: add file_id to message

regit:
This will allow to get the filename and by consequence the file
after a parsing of the EVE log file.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/82
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/80

### DIFF
--- a/src/output-json-file.c
+++ b/src/output-json-file.c
@@ -241,6 +241,9 @@ static void FileWriteJsonRecord(JsonFileLogThread *aft, const Packet *p, const F
     }
     json_object_set_new(fjs, "stored",
                         (ff->flags & FILE_STORED) ? json_true() : json_false());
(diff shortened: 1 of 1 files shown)


src/output-json-file.c
```diff
@@ -241,6 +241,9 @@ static void FileWriteJsonRecord(JsonFileLogThread *aft, const Packet *p, const F
     }
     json_object_set_new(fjs, "stored",
                         (ff->flags & FILE_STORED) ? json_true() : json_false());
+    if (ff->flags & FILE_STORED) {
```
inliniac: Why not add it unconditionally? An internal id is always assigned I think.

inliniac:
Merged through #1607.

PR #1551, opened 2015-06-16
suricatasc: remove "u" prefix when printing JSON output.

gozzy:
If we want to parse suricatasc's output, python's unicode prefix should be removed to make JSON parsers happy.

### DIFF
--- a/scripts/suricatasc/suricatasc.in
+++ b/scripts/suricatasc/suricatasc.in
@@ -42,7 +42,7 @@ except SuricataReturnException as err:
 if args.command:
     (command, arguments) = sc.parse_command(args.command)
     res = sc.send_command(command, arguments)
-    print(res)
+    print(json.dumps(res))
     sc.close()
     if res['return'] == 'OK':
         sys.exit(0)


inliniac:
@regit can you look at this? cc @jasonish

Zopieux:
Well, that makes sense. Cmdline docs says:
> and return JSON

Yet what is returned is the representation of a Python dict, which happens to looks like JSON, but is not:

```python
>>> print({u'hey "mémé"': None})
{u'hey "m\xe9m\xe9"': None}
>>> print(json.dumps({u'hey "mémé"': None}))
{"hey \"m\u00e9m\u00e9\"": null}
```

@gozzy Also I would suggest addind to your PR: printing to `stderr` for all the error messages (all the `print`s except the output one).

gozzy:
@Zopieux I think it can be done, but I don't know what's the right way to do it - print behaves differently in python 2 and 3. Will it be ok if I implement printing to stderr like below?

from __future__ import print_function
...
print("Error message", file=sys.stderr)

jasonish:
> @Zopieux I think it can be done, but I don't know what's the right way to do it - print behaves differently in python 2 and 3. Will it be ok if I implement printing to stderr like below? Or use sys.stderr.write?
> 
> from future import print_function
> ...
> print("Error message", file=sys.stderr)

In my own Python scripts, that I usually write to work with Python 2.6, and often check for 3+ compatibility, I always use from __future__ import print_function.  So I'd suggest using it.

Zopieux:
Indeed, the easiest and most portable way to do this is:

```python
from __future__ import print_function
import sys

print("foo", "bar", file=sys.stderr)
```

PR #1552, opened 2015-06-16
http body ips v8

glongo:
This addresses the comments in https://github.com/inliniac/suricata/pull/1545

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/46
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/45


inliniac:
The goal in the unittests is not just to pass a list of sigs and see if they match when they should, but also to validate that they _don't_ match when then should not.

PR #1554, opened 2015-06-16
af-packet: handle per interface config in different orders

jasonish:
Handles the case where the order of items in the mappings in the af-packet sequence have their ordered change, which can happen when the YAML is generated from another tool, as maps generally do not have an order.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1487

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/108
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/109

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -65,6 +65,27 @@ const char *RunModeAFPGetDefaultMode(void)
     return default_mode_workers;
 }
 
+/**
+ * \brief Get the af-packet configuration node for a device.
+ *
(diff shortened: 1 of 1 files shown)


inliniac:
The other methods will need the same fix, so it might make sense to generalize a bit?

inliniac:
@jasonish could you revisit this? I would like to improve this for 3.1. If we can have a more generic solution it would be great.

inliniac:
Replaced by #2062

PR #1555, opened 2015-06-16
af-packet: use maps instead of sequences.

jasonish:
Builds on PR #1554 .

This is an idea of a new way to configure af-packet interfaces which is probably a bit more true to YAML.  Some discussion here: https://redmine.openinfosecfoundation.org/issues/1487

Includes fallback to old config with a deprecation warning.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/109
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/110

Consider this an RFC, as we'd probably want to convert all input configurations in one go.

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -65,6 +65,27 @@ const char *RunModeAFPGetDefaultMode(void)
     return default_mode_workers;
 }
 
+/**
+ * \brief Get the af-packet configuration node for a device.
+ *
+ * Basically hunts through the list of maps for the first one with a
+ * key of "interface", and a value of the provided interface.
+ */
+static ConfNode *AFPFindDeviceConfig(ConfNode *root, const char *iface)
+{
+    ConfNode *if_root, *item;
+    const char key[] = "interface";
+    TAILQ_FOREACH(if_root, &root->head, next) {
+        TAILQ_FOREACH(item, &if_root->head, next) {
+            if (!(strcmp(item->name, key) && strcmp(item->val, iface))) {
+                return if_root;
+            }
+        }
+    }
+
+    return NULL;
+}
+
 void RunModeIdsAFPRegister(void)
 {
     RunModeRegisterNewRunMode(RUNMODE_AFP_DEV, "single",
@@ -156,7 +177,7 @@ void *ParseAFPConfig(const char *iface)
         return aconf;
     }
 
-    if_root = ConfNodeLookupKeyValue(af_packet_node, "interface", iface);
+    if_root = AFPFindDeviceConfig(af_packet_node, iface);
 
     if_default = ConfNodeLookupKeyValue(af_packet_node, "interface", "default");
 
@@ -380,7 +401,7 @@ int AFPRunModeIsIPS()
             return 0;
         }
         char *copymodestr = NULL;
-        if_root = ConfNodeLookupKeyValue(af_packet_node, "interface", live_dev);
+        if_root = AFPFindDeviceConfig(af_packet_node, live_dev);
 
         if (if_root == NULL) {
             if (if_default == NULL) {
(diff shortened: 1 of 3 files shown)


inliniac:
When overriding settings from the commandline, would this result in "--set af-packet.interfaces.eth0.threads=4"? If so, would it be possible to skip the 'interfaces'? Think "--set af-packet.eth0.threads=4" is neater.

jasonish:
For that to work, without special casing we would have to change the config to:
```
af-packet:
  defaults:
    use-mmap: yes
  eth0:
    threads: auto
    cluster-id: 99
  eth1:
    threads: atuo
    cluster-id: 98
```

while doable it comes with the following downsides:
- When iterating over the interfaces (the keys of af-packet) you will have to explicitly ignore "defaults".
- If a new key is added, for some other form of af-packet configuration, it would have to be explicitly ignored when iterating for interfaces.

By lumping all the interfaces under the "interfaces" key, you know that any child is an interface configuration, without having to explicitly ignore some keys.

That being said, I'm more or less indifferent here.

inliniac:
On 06/17/2015 06:22 PM, Jason Ish wrote:
> For that to work, without special casing we would have to change the
> config to:
> 
> |af-packet:
>   defaults:
>     use-mmap: yes
>   eth0:
>     threads: auto
>     cluster-id: 99
>   eth1:
>     threads: atuo
>     cluster-id: 98
> |
> 
> while doable it comes with the following downsides:
> 
>   * When iterating over the interfaces (the keys of af-packet) you will
>     have to explicitly ignore "defaults".
>   * If a new key is added, for some other form of af-packet
>     configuration, it would have to be explicitly ignored when iterating
>     for interfaces.
> 
> By lumping all the interfaces under the "interfaces" key, you know that
> any child is an interface configuration, without having to explicitly
> ignore some keys.
> 
> That being said, I'm more or less indifferent here.


I guess we could reverse it as well:
<pre>
af-packet:
  settings:
    default:
      setting1: value1
  eth0:
    threads: auto
</pre>

Then if we need to add global sections it could be under the 'settings'
banner.

Not sure yet either, but I think indentation depth is another factor to
consider. The libhtp section is a nice example how that can get out of
control, even if the structure makes sense.

jasonish:
> I guess we could reverse it as well:
> <pre>
> af-packet:
>   settings:
>     default:
>       setting1: value1
>   eth0:
>     threads: auto
> </pre>
> 
> Then if we need to add global sections it could be under the 'settings'
> banner.
> 
> Not sure yet either, but I think indentation depth is another factor to
> consider. The libhtp section is a nice example how that can get out of
> control, even if the structure makes sense.

This would still require the explicit ignore for "settings".  My main intention here is was to avoid have to explicitly ignore anything, knowing that you can iterate  a node knowing what kind of "type" they were. 

I also think the settings is kind of superfluous?  To me, af-packet is the settings, and the interfaces to listen on are a child of settings.

How often does one change the threads on the command line?

src/runmode-af-packet.c
```diff
@@ -65,6 +67,49 @@ const char *RunModeAFPGetDefaultMode(void)
...
+    const char key[] = "interface";
+
+    ConfNode *interfaces = ConfNodeLookupChild(root, "interfaces");
+    if (interfaces != NULL) {
+        return ConfNodeLookupChild(interfaces, iface);
+    }
+
+    SCLogWarning(SC_WARN_DEPRECATED,
+        "Found deprecated af-packet interface configuration. "
+        "Please switch to map based interface configuration.");
```
inliniac: I think it would be good to reference to a document/link here, as this warning is too technical to make sense to most people.

PR #1560, opened 2015-06-18
stream_size operator comparison (fix issue #1488)

Zopieux:
Funny bug.

`DetectStreamSizeParse` was first checking if `mode[0] == '<'`, which is true for both `<` and `<=`, thus `<=` (resp. `>=`) is never matched. This patch does the `strcmp` to `<=` (resp. `>=`) within the if block of `<` (resp. `>`) to fix [#1488](https://redmine.openinfosecfoundation.org/issues/1488).


inliniac:
Merged through #1580, thanks!

PR #1563, opened 2015-06-18
Suricatasc fix v2

gozzy:
Updated #1551 - changed printing errors in non-interactive mode (stderr instead of stdout).


inliniac:
Merged through #1580, thanks!

PR #1564, opened 2015-06-19
Af packet new modes

regit:
Implement the new features of AF_PACKET. Last patch of the series should dramatically reduce packets drop and increase treated bandwidth on single bandwidth intensive flow.

 See http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=77f65ebdca506870d99bfabe52bde222511022ec for information of rollover.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/83
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/81

PR #1565, opened 2015-06-20
Af packet new modes v1.1

regit:
Same code as #1564 but the configure test is done on another variable that appear to be undefined at least on some CentOS.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/84
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/82

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1054,8 +1054,8 @@
             [enable_af_packet="no"],
             [[#include <sys/socket.h>
               #include <linux/if_packet.h>]])
-        AC_CHECK_DECL([PACKET_FANOUT],
-            AC_DEFINE([HAVE_PACKET_FANOUT],[1],[Packet fanout support is available]),
+        AC_CHECK_DECL([PACKET_FANOUT_QM],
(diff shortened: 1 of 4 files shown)


inliniac:
I tried the rollover option in my setup, and it leads to a massive reduction of alerts in workers mode. (10k vs 20k in cluster_flow). Is any additional setup required?

regit:
Hi,
Le 21 juin 2015 10:52 AM, Victor Julien <notifications@github.com> a écrit :
>
> I tried the rollover option in my setup, and it leads to a massive reduction of alerts in workers mode. (10k vs 20k in cluster_flow). Is any additional setup required?
Did you set cluster mode to cluster_rollover ? If this is the case then go back to cluster_flow and add 'rollover: yes' to your iface configuration.
On my test, I was running cluster_cpu and I've seen an increase in logging as well as a quasi disappearance of packets drop.

pevma:
running on a live box  with cluster_cpu (and variations) and rollover set to yes - will update in  a few days.

pevma:
It looks good to me so far. Better performance as well.

regit:
Just proposed a rebased version.

PR #1566, opened 2015-06-25
Updated CUDA to use sm_50 and sm_52, deprecated sm_10, sm_11, sm_12, sm_13

robopt:
sm_1x architectures were deprecated in CUDA Toolkit 6.0. Meaning that these architectures can only run on CUDA Toolkit 4.0 -> 5.5. This adds sm_50 and sm_52 for the most recent cards and allows suricata to be compiled with CUDA Toolkit 6.x and 7.0.

The CUDA Toolkit Documentation states "When in doubt about the compute capability of the hardware that will be present at runtime, it is best to assume a compute capability of 2.0 as defined in the CUDA C Programming Guide section on Technical and Feature Specifications." which means it suggests sm_20 being the baseline (not sm_10).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -422,35 +422,19 @@ util-mpm-ac-cuda-kernel.cu
 NVCCFLAGS=-O2
 
 SUFFIXES = \
-.ptx_sm_10 \
-.ptx_sm_11 \
-.ptx_sm_12 \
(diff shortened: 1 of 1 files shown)


inliniac:
I fixed CUDA in a similar but more complete way in #1883, thanks anyway!

PR #1567, opened 2015-06-26
Issue 1491: fix capabilities for pf_ring mode when running under non-root account

gozzy:
CAP_NET_ADMIN is insufficient to turn promiscuous mode on, so add CAP_NET_RAW.


inliniac:
Merged, thanks!

PR #1575, opened 2015-07-01
output-json-email: fix to: and cc: lists when there is a comma within an

decanio:
email address.  for example:

"Julien, Victor" <victor@example.net>

This had been being split at the comma into two items on the list.

### DIFF
--- a/src/output-json-email-common.c
+++ b/src/output-json-email-common.c
@@ -55,6 +55,35 @@
 #ifdef HAVE_LIBJANSSON
 #include <jansson.h>
 
+char *emailaddrtok_r(char *str, char **saveptr)
+{
+    char *p = str;;
+
+    if (p != NULL) {
+        if (*p == '\"') {
+            /* scan to closing quote */
+            p = strtok_r(p, "\"", saveptr);
+            p = *saveptr;
+            if (p != NULL) {
+                p[-1] = '\"';
+            }
+            p = NULL;
+        }
+        p = strtok_r(p, ",", saveptr);
+        return str;
+    } else {
+        if (**saveptr == '\"') {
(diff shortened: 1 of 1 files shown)


decanio:
Small eve-log bug that had been splitting email to: and cc: lines incorrectly

- PR decanio: https://buildbot.openinfosecfoundation.org/builders/decanio/builds/10
- PR decanio-pcap: https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/10

src/output-json-email-common.c
```diff
@@ -55,6 +55,35 @@
 #ifdef HAVE_LIBJANSSON
 #include <jansson.h>
 
+char *emailaddrtok_r(char *str, char **saveptr)
```
inliniac: Can you add a comment about the intent of this func?

inliniac: Func name not matching our func naming std.

src/output-json-email-common.c
```diff
@@ -55,6 +55,35 @@
...
+{
+    char *p = str;;
```
inliniac: minor typo

src/output-json-email-common.c
```diff
@@ -55,6 +55,35 @@
...
+        if (*p == '\"') {
+            /* scan to closing quote */
+            p = strtok_r(p, "\"", saveptr);
+            p = *saveptr;
+            if (p != NULL) {
+                p[-1] = '\"';
+            }
+            p = NULL;
+        }
+        p = strtok_r(p, ",", saveptr);
```
inliniac: p is set, but on the next line str is returned. Is that correct?

PR #1576, opened 2015-07-03
Fix compile bad dereferences

sfd:
The src/source-erf-dag.c code was not compiling. It looks like some stats counters were added but not tested as the dereferences are incorrect.


sfd:
Looks like a minor error in structure dereferencing for added stats counters.

inliniac:
Thanks @sfd! @jasonish can you check this?

jasonish:
Yup, its the same that I have in my tree.

inliniac:
Merged through #1580, thanks!

PR #1586, opened 2015-07-10
[rfc] dns lua v8.1

inliniac:
@regit @jasonish could you check https://github.com/inliniac/suricata/commit/0f69115360b951180fada6a83c0b43828cfd3665 to see if you think this approach makes sense?


inliniac:
Updated PR at #1591

PR #1587, opened 2015-07-12
PF_RING IPS/TAP support

cardigliano:
This patch adds support for IPS/TAP mode in the Suricata PF_RING module.
Please note that original code from https://github.com/glongo/suricata.git has been reworked
mainly with fixes to socket capture direction, pfring_send parameters, and a few others.


inliniac:
Thanks @cardigliano, can you rework this so that @glongo 's work is preserved and your changes are incremental to that? Your commit message should then contain explanation of the changes/fixes. Listing them just in the PR doesn't add them to the git history.

PR #1588, opened 2015-07-12
netmap: support non-equal count of Rx and Tx rings on interface

gureedo:
Removed restriction on interfaces with non-equal count of Rx and Tx rings.

### DIFF
--- a/src/source-netmap.c
+++ b/src/source-netmap.c
@@ -127,6 +127,8 @@ TmEcode NoNetmapSupportExit(ThreadVars *tv, void *initdata, void **data)
 
 #else /* We have NETMAP support */
 
+#define max(a,b) (((a) > (b)) ? (a) : (b))
+
 #define POLL_TIMEOUT 100
 
 #if defined(__linux__)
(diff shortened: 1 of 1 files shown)


inliniac:
Could you expand on what this means? Is this addressing some of the things Oliver ran into?

gureedo:
No, it's absolutely different improvement.
Just want to remove this limitation for future.
Netmap does not guarantees that NIC will have equal number of transmit and receive rings.
Current code code can't work with such NIC.

I've found that my code will not work in autofp mode. Sorry for that, I'll make another pull request to fix it in couple of days.
Also I will implement ability to specify OS network stack as copy endpoint, there will be one more pull request for this.

inliniac:
I see, thanks. Could you add explanations like this to the commits themselves? That makes it easier to trace back changes later.

gureedo:
Amended commit with new comments.

inliniac:
Please don't update existing PR-branches, instead create a new branch, update it and create a new PR. Thanks.

gureedo:
Ok, will open new PR. Just interesting why it's should not be updated?

PR #1592, opened 2015-07-14
Af packet new modes v1.2

regit:
A rebase of #1565.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/85
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/83

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1054,8 +1054,8 @@
             [enable_af_packet="no"],
             [[#include <sys/socket.h>
               #include <linux/if_packet.h>]])
-        AC_CHECK_DECL([PACKET_FANOUT],
-            AC_DEFINE([HAVE_PACKET_FANOUT],[1],[Packet fanout support is available]),
+        AC_CHECK_DECL([PACKET_FANOUT_QM],
+            AC_DEFINE([HAVE_PACKET_FANOUT],[1],[Recent packet fanout support is available]),
             [],
             [[#include <linux/if_packet.h>]])
     ])
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -282,6 +282,19 @@ void *ParseAFPConfig(const char *iface)
         SCLogInfo("Using cpu cluster mode for AF_PACKET (iface %s)",
                 aconf->iface);
         aconf->cluster_type = PACKET_FANOUT_CPU;
+    } else if (strcmp(tmpctype, "cluster_qm") == 0) {
+        SCLogInfo("Using queue based cluster mode for AF_PACKET (iface %s)",
+                aconf->iface);
+        aconf->cluster_type = PACKET_FANOUT_QM;
+    } else if (strcmp(tmpctype, "cluster_random") == 0) {
+        SCLogInfo("Using random based cluster mode for AF_PACKET (iface %s)",
+                aconf->iface);
+        aconf->cluster_type = PACKET_FANOUT_RND;
+    } else if (strcmp(tmpctype, "cluster_rollover") == 0) {
+        SCLogInfo("Using rollover based cluster mode for AF_PACKET (iface %s)",
+                aconf->iface);
+        aconf->cluster_type = PACKET_FANOUT_ROLLOVER;
+
     } else {
         SCLogError(SC_ERR_INVALID_CLUSTER_TYPE,"invalid cluster-type %s",tmpctype);
         SCFree(aconf);
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -301,6 +301,14 @@ void *ParseAFPConfig(const char *iface)
         return NULL;
     }
 
+    int conf_val = 0;
+    ConfGetChildValueBoolWithDefault(if_root, if_default, "rollover", &conf_val);
+    if (conf_val) {
+        SCLogInfo("Using rollover kernel functionality for AF_PACKET (iface %s)",
+                aconf->iface);
+        aconf->cluster_type |= PACKET_FANOUT_FLAG_ROLLOVER;
+    }
+
     /*load af_packet bpf filter*/
     /* command line value has precedence */
     if (ConfGet("bpf-filter", &bpf_filter) != 1) {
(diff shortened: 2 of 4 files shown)


configure.ac
```diff
@@ -1054,8 +1054,8 @@
             [enable_af_packet="no"],
             [[#include <sys/socket.h>
               #include <linux/if_packet.h>]])
-        AC_CHECK_DECL([PACKET_FANOUT],
```
inliniac: Will this disable our FANOUT support for any kernel not supporting FANOUT_QM?

regit: No, see https://github.com/regit/suricata/blob/af-packet-new-modes-v1.2/src/source-af-packet.h#L27, we are forcing the declaration if the headers do not have the define.

inliniac: I see. So really the whole test is sort of useless?

regit: Hi,
Yes, no functional change and for now no impact that can be foreseen.
Le 21 juil. 2015 11:38 PM, Victor Julien <notifications@github.com> a écrit :In configure.ac:
> @@ -1054,8 +1054,8 @@
>              [enable_af_packet="no"],
>              [[#include <sys/socket.h>
>                #include <linux/if_packet.h>]])
> -        AC_CHECK_DECL([PACKET_FANOUT],

I see. So really the whole test is sort of useless?

inliniac:
I think some of the new options are quite confusing. How I understand them:

* cluster_random: random distribution of packets over the sockets? In other words, useless for any use case where stateful protocols like TCP are used as timing issues will break the order in which we process the packets.

* cluster_qm: balances by rss queue, so the balancing is done by the nic and not the kernel

* cluster_rollover: I have no idea what this does

* rollover option: modifies all existing cluster modes to spill over packets from one socket to the next, so that we don't drop them. However due to the timing issues also present in 'cluster_random', this will lead to massive TCP (and other stateful protocol) tracking issues. This is what I see in my replay test as well (cluster_flow+rollover). Drops are way down, but so are the alerts. So despite inspecting more packets we're missing more alerts.

Can you confirm and fill in the blanks?

regit:
Hi,

- cluster_random: useless for us indeed
- cluster_qm: your analysis is perfect
- cluster_rollover: send packet to first socket, when ring is full send them to next socket with room available (I think it is useless too)
- rollover: analysis correct. Strange you see less alerts. We may find something to fix or improve here.

inliniac:
Is the rollover option enforced per packet or is a flow 'migrated' to a new socket completely? Assuming the first.

It does make sense to me that alert rates would be lower. My test measures drop rates, and the base line is about 30-50% drop (depending on runmode and options). This means that many flows will be affected.

What will happen is that for many flows multiple threads will be processing the packets at approximately the same time. Due to threading timing uncertainty, we will see many cases where ordering is off. For example process a SYN/ACK before the SYN. All kinds of issues will happen: packets considered to be out of window, bad RST's, bad ACK's, etc, etc.

In my test, I see a slight reduction in drop rates, but almost 50% less alerts.

regit:
Yes, rollover is per-packet.

Regarding your alert rate, you are testing a very bad case (30% drop). I understand that in that case it can just help to get thing a bit worse. I will try to check with lower drop rate that appear to be fixed by the setup.

inliniac:
Rerunning my tests at lower speeds as well.

I think it may be good to hide all the 'useless' options from the yaml and just point to a wiki page where the various options and their effects are explained. There may be a case for each of them, but most users shouldn't even consider cluster_random and cluster_rollover for example.

regit:
Yes, make sense.

inliniac:
Ok cool, I'll be merging this PR today. Such changes can be incremental.

inliniac:
rolloverstats

Ok there seems to be a slight advantage when drop rate is close to 0, but in all other cases I see massive alert reduction, even when drop rates align. The red line is key here.

Left Y-axis is alert count. Right Y-axis drop rate in %. X-axis replay speed in Gbps.

inliniac:
Btw drop rate is calculated by taking number of sent packets (known quantity) and the known processed packets (stats.log). The drop stats suri reports are ignored.

regit:
Really interesting. What kind of alerts are generated (you've got heavy number here) ? Do you the stats for the metadata ?

inliniac:
This based on etpro-all.rules from a while back, on a 118GiB pcap with much HTTP traffic. I'm not yet tracking HTTP sessions/logs or TCP session in my testing. Will see if I can add that.

What kind of metadata would you be interested in?

regit:
Basically, I would like to see if we have the same ratio in metadata generation. HTTP for example would be nice.

inliniac:
rolloverstats2
Speed vs drop rate vs http tx log count. Pretty clear that the rollover leads to massive issues.

inliniac:
I tested the same branch with the rollover opt disabled and then it behaves like master. So it's not caused by some other diff between the branches.

poona:
Is there a difference between cluster_qm and cluster_cpu?

My understanding was that cluster_cpu depended on rss, and hence the nic to do the balancing, which is the same thing as cluster_qm?

regit:
@poona: It is sort of the same but it is possible to pin some flow to a given qm in the network card. So in that case that can be really different.

pevma:
@inliniac  - do you get the exact same packet stats with both runs using the same pcap ? (ex kernel.packets/bytes..)

inliniac:
My tool isn't tracking the stats.log yet, it uses the stats dumped at shutdown instead. From there it takes pkts seen and pkts dropped. Since I know exactly how many packets were sent (using the same pcap each time) I can with certainty tell what the drop rate is. Alerts and http counts are simple 'wc -l' on the http.log and fast.log.

pevma:
I did some live traffic tests - on one of our test boxes. I did not see such a big discrepancy in the runs with and without this PR. Could this be related to some particularity in a pcap replay technique?

inliniac:
Merged through #1607. We'll continue to look at that rollover issue in the coming time.

PR #1593, opened 2015-07-14
Tls eve log fix v1.0

regit:
This patchset contains one patch introducing a new 'tls-store' module and a second restoring backward compatibility.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1150

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/86
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/84

### DIFF
--- a/src/log-tlsstore.c
+++ b/src/log-tlsstore.c
@@ -1,0 +1,424 @@
+/* Copyright (C) 2014 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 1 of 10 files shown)


src/log-tlsstore.c
```diff
@@ -0,0 +1,424 @@
...
+/**
+ *  \internal
+ *
+ *  \brief deinit the log ctx and write out the waldo
+ *
+ *  \param output_ctx output context to deinit
+ */
+static void LogTlsStoreLogDeInitCtx(OutputCtx *output_ctx)
+{
+    free(output_ctx);
```
inliniac: SCFree?

src/log-tlsstore.c
```diff
@@ -0,0 +1,424 @@
...
+    SCLogWarning(SC_ERR_FWRITE, "Unable to write certificate");
+end_fwrite_fpmeta:
+    if (fpmeta) {
+        fclose(fpmeta);
+        SCLogWarning(SC_ERR_FWRITE, "Unable to write certificate metafile");
+    }
+    SCReturn;
+end_fp:
+    fclose(fp);
+}
```
inliniac: minor: missing return, while we have it everywhere else in the func

inliniac:
Right, so ultimately this just adds another flag to the state for keeping the eve cert logging and this modules logging independent, correct?

regit:
Yes, exactly.

regit:
Going to propose a updated PR.

inliniac:
btw I noticed that we log errors that we could encounter at packet runtime, meaning we could be flooded with messages. This isn't new in the pr, but still something to consider maybe?

PR #1595, opened 2015-07-15
Enip feature 1495 v3

kwong-solana:
rebased to master.  Add SCADA EtherNet/IP(ENIP) and Common Industrial Support (CIP )

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -202,6 +202,9 @@ detect-window.c detect-window.h \
 detect-within.c detect-within.h \
 detect-modbus.c detect-modbus.h \
 detect-xbits.c detect-xbits.h \
+detect-cipservice.c detect-cipservice.h \
+decode-enip.c decode-enip.h \
+decode-cip.c decode-cip.h \
 flow-bit.c flow-bit.h \
 flow.c flow.h \
 flow-hash.c flow-hash.h \
--- a/src/decode-cip.c
+++ b/src/decode-cip.c
@@ -1,0 +1,575 @@
+/* Copyright (C) 2007-2014 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Kevin Wong <kwong@solananetworks.com>
+ *
+ * Decode CIP
+ */
+
+#include "suricata-common.h"
+#include "util-unittest.h"
+#include "util-unittest-helper.h"
+#include "detect-parse.h"
+#include "detect-engine.h"
+#include "util-byte.h"
+#include "pkt-var.h"
+#include "util-profiling.h"
+#include "decode-cip.h"
+
+
+/**
+ * entry Point for CIP parsing - Decode the Common Industrial Protocol
+ */
+int DecodeCIP(Packet *p, ENIP_DATA *enip_data, uint16_t offset)
+{
+	int ret = 1;
+
+	if (enip_data->encap_data_item.length == 0)
+	{
+		SCLogDebug("DecodeCIP: No CIP Data\n");
+		return 0;
+	}
+
+	if (offset > p->payload_len)
+	{
+		SCLogDebug("DecodeCIP: Parsing beyond payload length\n");
+		return 0;
+	}
+
+	uint8_t service = 0;
+	service = *(p->payload + offset);
+
+	SCLogDebug("CIP Service 0x%x\n",service);
+
+	//use service code first bit to determine request/response, no need to save or push offset
+	if (service >> 7)
+	{
+		ret = DecodeCIPResponse(p, enip_data, offset);
+	} else {
+		ret = DecodeCIPRequest(p, enip_data, offset);
+	}
+
+	return ret;
+}
+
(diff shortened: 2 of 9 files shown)


hubayirp:
+1

inliniac:
The approach here seems to be to create a decoder module that doesn't actually run in our decoding stage, but instead in the detection context.

I think an approach like our DNS UDP (app-layer) parsing would make a lot of sense here, as it hands you the API's to more complex detection logic, plus it makes logging possible. I think it would be great to add CIP and EINP logging to our eve-log output.

It's unclear to me if this should inspect TCP as well? If so, the app-layer approach will be unavoidable to deal with stream reassembly.

inliniac:
The PR/branch will have to be improved wrt style and such btw:
- commits and commit messages: https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria
- coding style: https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Coding_Style

kwong-solana:
Thanks very much for the feedback.  Can you please give an example of issues in the coding style? We used other code in src and your guide as an examples.

jasonish:
I took a quick look at ENIP/CIP.  It looks likes it is always carried over TCP or UDP, in which case it should really go into the app-layer.  To start with, you should probably break this out into 2 areas, the app-layer and the detection.

Per @inliniac 's suggestion, the DNS app-layer is probably a good point of reference to start with.  In app-layer-dns-tcp.c you will see RegisterDNSTCPParsers where you can register  your application layer decoding logic on the necessary ports.  You then register request and response handlers, and Suricata will pass you the reassembled stream buffers.

Your app-layer will then decode ethernet/ip, and cip and record the data into a transaction object, which will then be accessible to the detection layer.  You will probably also want to register a protocol, such as "enip" so you can write rules like "alert enip...".

Another example to look at is the modbus support (app-layer-modbus.*, detect-modbus.*).  And if you want to see an in progress app-layer, detection and logging checkout the ongoing DNP3 work at https://github.com/jasonish/suricata/tree/ish-dnp3 - it might make it easier to pull out whats needed to hook it all up by diffing that against master.

I have yet to take a closer look at the code, but I think getting it organized and hooked in properly would be a good start.

kwong-solana:
Thanks again we will look into this.  We started this by using the Packet Inspection Module example as a base and started from there.  We will look into doing it app-layer style.   BTW, the other links for extending suricata on the website are dead.

kwong-solana:
I was looking at the example rules you have for modbus and dns and was wondering how you would be able to deal with vendor specific type ENIP/CIP rules.  CIP packets can have Service, Class and Attribute defined by numeric value.  Some of these are defined by ODVA but for the most part, the manufacturer specifies them and builds equipment  using their own additions.

jasonish:
> I was looking at the example rules you have for modbus and dns and was wondering how you would be able to deal with vendor specific type ENIP/CIP rules. CIP packets can have Service, Class and Attribute defined by numeric value. Some of these are defined by ODVA but for the most part, the manufacturer specifies them and builds equipment using their own additions.

I don't think thats too much of a problem.  So you have a keyword, cip_server:<numeric_value>...  Obviously we can't ship vendor specific ones as we don't know what they are.  So it would be up the end user to write those rules - if thats what you are asking?

kwong-solana:
To clarify, our rules have keyword "cip_service" with a comma delimited numeric parameter string for the Service, Class(optional) and Attribute (optional).  For example:  "cip_service:4,300,16"  With detects a CIP packet with Service=4, Class=300 and Attribute=16.  These services,classes,and attribute are unknown to us and can exist out in the field, but we need to support them anyways

inliniac:
@kwong-solana wrt style, things that I noticed: indent issues (should be 4 spaces), structure names (are all caps), function comments (should use doxygen format).

kwong-solana:
Hi All, have you gotten a chance to determine whether it is possible to do an app-layer implementation with the types of rules we want to support?  I have looked at the modbus and dns rules and can see that they deal with defined events whereas ENIP/CIP allows vendors to decide what is possible.

jasonish:
> Hi All, have you gotten a chance to determine whether it is possible to do an app-layer implementation with the types of rules we want to support? I have looked at the modbus and dns rules and can see that they deal with defined events whereas ENIP/CIP allows vendors to decide what is possible.

Honestly I think this is up to you.  The app-layer approach doesn't put any more limitations on what you've tried to do already, its just a better place to do the protocol decoding as you will get a normalized stream data.

I'm still not sure I understand what you are after.  Basically you will have to hard code some keywords for detection, but the contents of those keywords could give you the flexibility of what you need.

Modbus is one example, eg) "modbus: access write coils, address > 500;" is almost like a small language within the value of the keyword.  The DNP3 approach is a little more strict where the keyword denotes what data is being looked for, and the value of the keyword is just match parameters.

So you'll have to find what you need to do to provide the flexibility required to look at vendor specific events.  It might be as simple as just allowing unknown values to be used.  For example, I think in DNP3, vendors can use vendor specific function codes (I may be wrong here), but the DNP3 layer doesn't care that its vendor specific, it will just look for it.

But if the vendor is going to send vendor specific data objects over the wire, then there is little you can do to provide support for that within Suricata, short of hooking into Lua and writing Lua scripts that would do the custom decoding required.

Hope that helps.

kwong-solana:
Ok, I think I understand what you are saying.  It's not that I should scrap all the work I did with keyword style rules, I should be adding this to app-layer handling so it can handle stream data (seen in detect-modbus.c).  Also future-wise, I can add protocol specific events to my rules (seen in applayer-mobus.c).  ex. app-layer-event:enip-event:invalid-path

kwong-solana:
Hi All, our team was wondering if you can pull in the ENIP/CIP feature as is into master.  This would be  a first cut so that in some point in the future we can proceed with the app layer implementation.  More features should follow afterwards.  We have done extensive testing using various security tools and it has been successful.

inliniac:
The issue I see is that the current code has a fundamental weakness: it's trivial to evade it using very low tech tricks (tcp session splicing). This won't be addressed in a simple incremental fix/update, but will require significant re-factoring which will essentially remove most of the current implementation.

By merging it now it could end up in a stable release before you've been able to improve it, which is simply not acceptable. That would force me to rip it out again, etc. So it makes much more sense to leave this out of tree until it's in better shape.

Conclusion: I won't merge this iteration, sorry.

PR #1594, opened 2015-07-15
Tls eve log fix v1.1

regit:
A rework of #1593 with only changes being fixes for remark and an incremental commit with a free->SCFree mods.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/87
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/85

PR #1596, opened 2015-07-15
Tls eve log fix v1.2

regit:
Update of #1594 with some spacing issue fix in initial commit and a additional commit that fixes the log flooding possibility.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/88
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/86

### DIFF
--- a/src/log-tlsstore.c
+++ b/src/log-tlsstore.c
@@ -1,0 +1,421 @@
+/* Copyright (C) 2014 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


src/log-tlsstore.c
```diff
@@ -0,0 +1,439 @@
...
+
+#include "util-logopenfile.h"
+#include "util-crypt.h"
+#include "util-time.h"
+
+#define MODULE_NAME "LogTlsStoreLog"
+
+static char tls_logfile_base_dir[PATH_MAX] = "/tmp";
+SC_ATOMIC_DECLARE(unsigned int, cert_id);
+static char logging_dir_not_writable;
```
inliniac: As there are multiple threads using/updating this var at the same time, I wondering if we need to use an atomic here. It's not really critical, but for consistency it may be better.

regit: We could indeed have more than 6 messages in case multiple threads are used at the same time. You have to never underestimate users but I can not think about someone counting these messages.

inliniac:
Merged through #1601, thanks!

PR #1597, opened 2015-07-16
Support for reconnecting unix domain socket log files

jsbarber:
https://redmine.openinfosecfoundation.org/issues/1423

Updated as discussed in #1467:
- Changed clock_gettime to more portable gettimeofday
- Changed SCLogError to SCLogWarning for recoverable failures
- Made return value of SCLogFileWrite clearer
- Rebased to current master

This patch changes the behavior of the code that writes log records to a unix domain socket:
If suricata cannot immediately connect to the receiver, the receiver socket path name and socket type (stream vs dgram) will be remembered in the LogFileCtx. When log records are written later, the connect will be attempted.

Also, if a connection to the receiver is broken -- for example, if the receiver program is restarted -- suricata will now clean up the old connection and automatically reconnect.

Since creating the unix domain socket then unsuccessfully attempting to connect then tearing it down requires relatively significant kernel resources, the reconnection is only retried occasionally (currently set to 1/2 second). 

Tested with both unix_stream and unix_dgram socket types. Only tested with fast-alert log, but should work for any caller of SCConfLogOpenGeneric.

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -36,11 +36,12 @@
 
 /** \brief connect to the indicated local stream socket, logging any errors
  *  \param path filesystem path to connect to
+ *  \param log_err, non-zero if connect failure should be logged.
  *  \retval FILE* on success (fdopen'd wrapper of underlying socket)
  *  \retval NULL on error
  */
 static FILE *
-SCLogOpenUnixSocketFp(const char *path, int sock_type)
+SCLogOpenUnixSocketFp(const char *path, int sock_type, int log_err)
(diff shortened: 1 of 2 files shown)


regit:
Regarding the buffering system, it is implemented in ulogd2 by http://git.netfilter.org/ulogd2/commit/?id=6e4d8af18923ec4873d96228ccf24eda3da4577e. The idea is simple, we allocate a buffer and store line event to it and then flush.

inliniac:
Merged through #1601, thanks!

PR #1604, opened 2015-07-20
netmap: bunch of features

gureedo:
Support non-equal count of Rx and Tx rings on interface.
Netmap does not guarantees that NIC will have equal number of transmit and receive rings.
Fixed autofp mode for netmap backend. Previous implementation was lack of it.
Added support to place suricata between network card and OS network stack.
Now it's possible to run suricata as firewall and web server on same machine or forward traffic to OS for routing after inspecting.

### DIFF
--- a/src/runmode-netmap.c
+++ b/src/runmode-netmap.c
@@ -103,7 +103,7 @@ static void NetmapDerefConfig(void *conf)
 *
 * \return a NetmapIfaceConfig corresponding to the interface name
 */
-static void *ParseNetmapConfig(const char *iface)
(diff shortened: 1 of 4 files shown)


inliniac:
Merged through #1607, thanks!

In future patches, can you improve the layout of your commit messages? See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits

Example: https://github.com/inliniac/suricata/commit/1cd97713c2945a1e55ea6e3b8df804ab3f06fca4
Thanks!

PR #1605, opened 2015-07-21
xff: support ports and more ipv6 notations

inliniac:
It's not uncommon to see an header like:
X-Forwarded-For: 1.2.3.4:56789

This patch recognizes this case and ignores the port. It also supports
this for IPv6 if the address has the following notation:
X-Forwarded-For: [12::34]:1234

This patch also adds unittests.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/268
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/270

### DIFF
--- a/src/app-layer-htp-xff.c
+++ b/src/app-layer-htp-xff.c
@@ -31,6 +31,7 @@
 
 #include "util-misc.h"
 #include "util-memrchr.h"
+#include "util-unittest.h"
 
 /** XFF header value minimal length */
 #define XFF_CHAIN_MINLEN 7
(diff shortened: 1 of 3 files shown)


inliniac:
@jasonish @serializingme could you review this?

src/app-layer-htp-xff.c
```diff
@@ -39,6 +40,70 @@
...
+    } else {
+        /* lets see if the xff string ends in a port */
+        int c = 0;
+        int d = 0;
+        char *p = input;
+        while (*p != '\0') {
+            if (*p == ':')
+                c++;
+            if (*p == '.')
+                d++;
```
jasonish: Perhaps a bit of a nit pick, but make this an else if, with the IPv4 case first as its the most common (at least for now)?

serializingme:
Seems OK to me, it will also be able to handle the IPv6 addresses without ports and brackets which is good. Maybe some unit tests should be added for those type of cases.

inliniac:
Merged through #1608 with some extra unittests.

PR #1610, opened 2015-07-29
Add option to omit payload in unified2 output

alessandro-guido:
Add a boolean option named "payload" to the unified2-alert output type.
When enabled, suricata omits the payload in the resulting unified2
file. The default value is true in order to preserve the current behaviour.

### DIFF
--- a/src/alert-unified2-alert.c
+++ b/src/alert-unified2-alert.c
@@ -186,7 +186,10 @@ typedef struct AlertUnified2Packet_ {
 typedef struct Unified2AlertFileCtx_ {
     LogFileCtx *file_ctx;
     HttpXFFCfg *xff_cfg;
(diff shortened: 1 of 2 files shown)


inliniac:
Hi Alessandro, can you add the option you added to the suricata.yaml.in as well with a small explanation. In this branch it's only added to the code which means most users won't find it.

alessandro-guido:
Sure! I've added some comments in the suricata.yaml.in file. you can find them in the same, updated, branch.

inliniac:
@jasonish any thoughts on this?

jasonish:
Seems fine. I question the why, but it doesn't hurt.  And its configurable with eve, so why not unified2 as well.

inliniac:
Merged through #1675, thanks!

PR #1617, opened 2015-08-28
templates v8

inliniac:
Detect and decode templates, plus setup scripts.

Since #1616:
- add detect setup script
- improve detect template

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -60,6 +60,7 @@ decode-teredo.c decode-teredo.h \
 decode-udp.c decode-udp.h \
 decode-vlan.c decode-vlan.h \
 decode-mpls.c decode-mpls.h \
+decode-template.c decode-template.h \
 defrag-config.c defrag-config.h \
 defrag.c defrag.h \
 defrag-hash.c defrag-hash.h \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -193,6 +193,7 @@ detect-ssl-state.c detect-ssl-state.h \
 detect-ssl-version.c detect-ssl-version.h \
 detect-stream_size.c detect-stream_size.h \
 detect-tag.c detect-tag.h \
+detect-template.c detect-template.h \
 detect-threshold.c detect-threshold.h \
 detect-tls.c detect-tls.h \
 detect-tls-version.c detect-tls-version.h \
--- a/src/detect-template.c
+++ b/src/detect-template.c
@@ -1,0 +1,302 @@
+/* Copyright (C) 2015 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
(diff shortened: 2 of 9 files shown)


src/Makefile.am
```diff
@@ -60,6 +60,7 @@ decode-teredo.c decode-teredo.h \
 decode-udp.c decode-udp.h \
 decode-vlan.c decode-vlan.h \
 decode-mpls.c decode-mpls.h \
+decode-template.c decode-template.h \
```
regit: Do we really want to have it here? Should it not be only in DIST_FILES ?

src/detect-template.c
```diff
@@ -0,0 +1,302 @@
...
+    return templated;
+
+error:
+    if (templated)
+        SCFree(templated);
+    return NULL;
+}
+
+/**
+ * \brief this function is used to atemplated the parsed template data into the current signature
```
regit: I don't understand atemplated here.

src/detect.c
```diff
@@ -5236,6 +5237,7 @@ void SigTableSetup(void)
     DetectDnsQueryRegister();
     DetectModbusRegister();
     DetectAppLayerProtocolRegister();
+    DetectTemplateRegister();
```
regit: Maybe we could use coccinelle patches for code modification ?

src/Makefile.am
inliniac: A bit torn here. I would like this to be part of the regular build testing. Planning to create a QA test that also tests the scripts themselves, but that would exclude it from static code analysis unless we do a lot of trickery first. So think I'll keep it for now and we can revisit later.

src/detect-template.c
inliniac: Me neither :) Will reword.

regit:
I really like this work except on one point: we are adding "template" code into the effective code. Maybe we could put some banner inside comments at the correct place and do some multiline sed. This should work well for C file.

Regarding the Makefile.am, if I remember correctly it would be better to follow alphabetical order. Maybe we could try some python thingy to do that.

src/detect.c
inliniac: Maybe, although this case is simple enough to work like this as well.

src/Makefile.am
regit: Ok, got the reason reading the rest of the patchset.

regit:
Humm, just seen the Makefile.am file list was not really fully ordered.

src/detect.c
regit: Yes, make sense to stay like that for now.

src/Makefile.am
inliniac: Guess we could hide it behind a configure option, but that would complicate the code search and replace a bit since we'd have to rename/remove #ifdefs too.

inliniac:
Replaced by #1624

PR #1621, opened 2015-09-01
RFC - Application layer templates - v3

jasonish:
This is a template that can be used for creating an application layer detector/parser, logger, and content match keywords.

The commits are layered:
1. First we create the application layer detector and parser and hook it into the build.
2. Then a JSON logger is created for the application layer transactions and hooked into eve as a sub-module.
3. Lastly a keyword is created for performing content inspections on decoded application layer buffers.

The protocol used by the templates is simply an echo.  The request is the data sent to the server, the response is the data echo'd back.  There is little we can do to further decode this data, but its about as simple as we can get for a protocol that transitions transaction states.

TODO: Another keyword for detecting some parameter decoded by the application layer.

The last commit is 2 scripts for helping a developer get started with a new application layer.

- ./scripts/setup-app-layer.sh - This script will copy the application detector/parser templates and hook them into the build.
- ./scripts/setup-app-layer-logger.sh - This script will copy the logging templates and hook them into the build and depends on setup-app-layer.sh having already been run.

TODO: a script to provision the content match modifier

I thought separate scripts might be ideal so the developer could then work on their app layer, then provision their logger. But I'm not sure if it might be better to do it all at one.  Or perhaps have command line flags:

./scripts/setup-app-layer.sh --with-logger --with-content-match --with-detect

Where --with-logger would include the logger, --with-content-match would setup a noopt, content match modifier and --with-detect would setup a detect keyword for looking at some decoded value in the transaction.


src/detect-template-buffer.h
```diff
@@ -0,0 +1,27 @@
...
+ * 02110-1301, USA.
+ */
+
+#ifndef __DETECT_TEMPLATE_BUFFER_H__
+#define __DETECT_TEMPLATE_BUFFER_H__
+
+#include "app-layer-template.h"
+
+void DetectTemplateBufferRegister(void);
+uint32_t DetectTemplateBufferInspectMpm(DetectEngineThreadCtx *, Flow *,
```
inliniac: Not implemented here, so shouldn't be part of this patch series.

src/output-json-template.c
```diff
@@ -0,0 +1,186 @@
...
+void TmModuleJsonTemplateLogRegister(void)
+{
+    tmm_modules[TMM_JSONTEMPLATELOG].name = "JsonTemplateLog";
+    tmm_modules[TMM_JSONTEMPLATELOG].ThreadInit = JsonTemplateLogThreadInit;
+    tmm_modules[TMM_JSONTEMPLATELOG].ThreadDeinit = JsonTemplateLogThreadDeinit;
+    tmm_modules[TMM_JSONTEMPLATELOG].RegisterTests = NULL;
+    tmm_modules[TMM_JSONTEMPLATELOG].cap_flags = 0;
+    tmm_modules[TMM_JSONTEMPLATELOG].flags = TM_FLAG_LOGAPI_TM;
+
+    /* Register as en eve sub-module. */
```
inliniac: Minor: typo 'en'

src/detect-template-buffer.c
```diff
@@ -0,0 +1,58 @@
...
+{
+    s->list = DETECT_SM_LIST_TEMPLATE_BUFFER_MATCH;
+    s->alproto = ALPROTO_TEMPLATE;
+    return 0;
+}
+
+static void DetectTemplateBufferRegisterTests(void)
+{
+#ifdef UNITTESTS
+#endif /* UNITTESTS */
```
inliniac: Lets add at least one sig parse test here. "alert tcp any any -> any any (template_buffer; content:"test"; sid:1;) or something.

src/app-layer-template.c
```diff
@@ -0,0 +1,413 @@
...
+    TemplateTransaction *tx = SCCalloc(1, sizeof(TemplateTransaction));
+    if (unlikely(tx == NULL)) {
+        return NULL;
+    }
+
+    /* Increment the states transaction count before assigning the
+     * transaction ID. This means the lowest transaction ID will be 1
+     * instead of 0.
+     *
+     * VictorJ: Why? - Other protocols do this, but it doesn't seem
```
inliniac: I don't remember, I guess we can remove it in other protocols and see if it still works :)

src/app-layer-template.c
```diff
@@ -0,0 +1,413 @@
...
+        return NULL;
+    }
+    TAILQ_INIT(&state->tx_list);
+    return state;
+}
+
+static void TemplateStateFree(void *state)
+{
+    SCLogNotice("Freeing template state.");
+    SCFree(state);
```
inliniac: We should free the transactions here as well to be sure.

src/app-layer-template.c
jasonish: Ok, leaving the first one at 0, and removing this comment from the template.

jasonish:
See https://github.com/inliniac/suricata/pull/1626

PR #1622, opened 2015-09-02
travis: move off legacy support

inliniac:
http://docs.travis-ci.com/user/migrating-from-legacy

Currently fails to build:

<pre>
Disallowing packages: libpcre3-dbg, libnet1-dev, libcap-ng-dev, libcap-ng0, libnetfilter-queue-dev, libnetfilter-queue1, coccinelle

If you require these packages, please review the package approval process at: https://github.com/travis-ci/apt-package-whitelist#package-approval-process
</pre>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,34 @@ compiler:
   - clang
 # Change this to your needs
 script: sh autogen.sh && ./configure --enable-nfqueue --enable-unittests && make && make check
+addons:
+  apt:
+    packages:
+    - libpcre3
+    - libpcre3-dbg
+    - libpcre3-dev
+    - build-essential
+    - autoconf
+    - automake
+    - libtool
+    - libpcap-dev
+    - libnet1-dev
+    - libyaml-0-2
+    - libyaml-dev
+    - zlib1g
+    - zlib1g-dev
+    - libcap-ng-dev
+    - libcap-ng0
+    - make
+    - libmagic-dev
+    - libnetfilter-queue-dev
+    - libnetfilter-queue1
+    - libnfnetlink-dev
+    - libnfnetlink0
+    - coccinelle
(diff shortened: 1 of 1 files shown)


inliniac:
Requested support for some of the packages: https://github.com/travis-ci/apt-package-whitelist/issues/1095, https://github.com/travis-ci/apt-package-whitelist/issues/1096, https://github.com/travis-ci/apt-package-whitelist/issues/1097

inliniac:
The coccinelle one will have to be addressed differently, as the Ubuntu package is too old for us.

We'll have to do something like: http://docs.travis-ci.com/user/migrating-from-legacy/#How-Do-I-Install-Custom-Software%3F

inliniac:
It now completes, although the coccinelle checks aren't done yet.

inliniac:
@regit any thoughts on the coccinelle side?

regit:
Coccinelle people have changed syntax between the version. Maybe we can try autodetect in configure and generate smpl patches via .in system?

inliniac:
Seems a bit like overkill. We should be able to install a new cocci in the build env, just not through a ppa and not as 'root'

inliniac:
I guess we could even consider bundling spatch for qa purposes.

regit:
Agree on overkill. Building it would be far better.

inliniac:
@jasonish any thoughts on this? You've done a big overhaul after this PR, wonder were we currently are wrt the legacy support.

jasonish:
I'd like to move off legacy support.  I can relook at this if you like?

inliniac:
If you want, please. I think the spatch/cocci issue is still valid, so we'll have to think of a way to get the command.

jasonish:
I think this can be closed now. See #2635

PR #1626, opened 2015-09-07
Application layer templates - v4

jasonish:
Fixes all comments made in the previous PR https://github.com/inliniac/suricata/pull/1621 including:

- detect unit test
- remove "template" sections for eve-log and app-layer
- script to provision a keyword for content inspection on application layer created buffers
- fix scripts to work with BSD sed and ed.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/122
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/123

While most app-layers default to enabled when not found in the configuration file, I wonder if this one should default to disabled, but a big comment in the generated template stating so (or even removing those few lines of codes during provisioning)?  Of course, this only applies if we keep this in the build, which does have its pros and cons - a big pro of course being active in the build so issues are caught early.


scripts/setup-app-layer-logger.sh
```diff
@@ -0,0 +1,154 @@
...
+patch_makefile_am
+patch_suricata_c
+patch_tm_modules_c
+patch_tm_threads_common_h
+patch_suricata_yaml_in
+
+cat <<EOF
+
+A JSON application layer transaction logger for the protocol
+${protoname} has not been set in the files:
```
inliniac: s/not/now/g ?

src/app-layer-template.c
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    SCLogNotice("Transaction ID %"PRIu64" not found.", tx_id);
+    return NULL;
+}
+
+/**
+ * \brief Called by the application layer.
+ *
+ * Victor: Please comment on how these relates to GetAlstateProgress.
```
inliniac: this will need to be updated, see:
https://github.com/inliniac/suricata/commit/a798cf9518192bb78432cddb43c848f31c22ba3f and
https://redmine.openinfosecfoundation.org/projects/suricata/wiki/AppLayer_Decoder#Transaction-handling

src/app-layer-template.c
```diff
@@ -0,0 +1,418 @@
...
+    SCLogNotice("Transaction progress requested for tx ID %"PRIu64
+        ", direction=0x%02x", echotx->tx_id, direction);
+
+    if (direction & STREAM_TOCLIENT) {
+        if (echotx->response_done) {
+            SCLogNotice("Response is done, returning 2.");
+            return 2;
+        }
+        else {
+            /* Victor: Should this return 1 or 0? */
```
inliniac: see https://github.com/inliniac/suricata/pull/1626/files#r39039481

inliniac:
Can you fix the travis build error?

inliniac:
Think it would be good to add a small pcap to use with this code.

inliniac:
> While most app-layers default to enabled when not found in the configuration file, I wonder if this one should default to disabled, but a big comment in the generated template stating so (or even removing those few lines of codes during provisioning)? Of course, this only applies if we keep this in the build, which does have its pros and cons - a big pro of course being active in the build so issues are caught early.
Disabled by default makes most sense. Maybe we can add a simple commandline option to enable it? The provision script should make the new stuff enabled by default though.

jasonish:
See: https://github.com/inliniac/suricata/pull/1633

PR #1630, opened 2015-09-09
Store previous value to pvalue variable per thread

jmlich:
Storing of previous value is done just for "Totals" counters. I believe that it should be stored also per thread.

### DIFF
--- a/src/counters.c
+++ b/src/counters.c
@@ -706,6 +706,7 @@ static int StatsOutput(ThreadVars *tv)
             StatsRecord *r = &stats_table.tstats[offset];
             r->name = table[c].name;
             r->tm_name = sts->name;
+            r->pvalue = r->value;
 
             switch (e->type) {
                 case STATS_TYPE_AVERAGE:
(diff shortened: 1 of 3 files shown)


jmlich:
In other words, the pvalue is currently uninitialized which is in my opinion wrong.

inliniac:
Thanks @jmlich. @jasonish can you check this? Thanks!

jmlich:
I am sorry, but I added next two commits into that pull request. It wasn't intentional.

I am not sure how to split this into two separate issues.

In second issue, I would like to pass "type" of record into the StatsRecord.

inliniac:
This contains a lot of helpful info: https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing

jasonish:
@jmlich Please read up on the URL @inliniac provided.  In short, create a new branch from master, and just include the first commit then submit a new PR.

I also find it helps if there is a bit of explanation of why the PR is being submitted.  In this case, this PR fixes the delta outputs in the JSON stats logging.  Is that what made you create this patch?

inliniac:
Needs some cleanup.

PR #1633, opened 2015-09-10
App layer templates - v5

jasonish:
Obsoletes: https://github.com/inliniac/suricata/pull/1626

- Fix travis-ci build error.
- Typos and comment updates.

Still need to have disabled by default.

I have a pcap here: https://www.dropbox.com/s/screla8lgh2a03t/template-echo.pcap?dl=0

Just noticed that one side of the checksum is bad.  I'll fix that up and provide a new pcap.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/125
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/126


src/app-layer-template.c
```diff
@@ -0,0 +1,419 @@
...
+        SCLogNotice("Failed to find transaction for response on echo state %p.",
+            echo);
+        goto end;
+    }
+
+    SCLogNotice("Found transaction %"PRIu64" for response on echo state %p.",
+        tx->tx_id, echo);
+
+    /* Make a copy of the response. */
+    tx->response_buffer = SCCalloc(1, input_len);
```
inliniac: We should check for this buffer to be already set. Today I used the template for a slightly more complex protocol and it took a bit for me to realize this pointer was overwritten for data coming in in chunks.

src/app-layer-template.c
```diff
@@ -0,0 +1,419 @@
...
+    /* Allocate a transaction. */
+    TemplateTransaction *tx = TemplateTxAlloc(echo);
+    if (unlikely(tx == NULL)) {
+        SCLogNotice("Failed to allocate new Template tx.");
+        goto end;
+    }
+    SCLogNotice("Allocated Template tx %"PRIu64".", tx->tx_id);
+    
+    /* Make a copy of the request. */
+    tx->request_buffer = SCCalloc(1, input_len);
```
inliniac: Same as with https://github.com/inliniac/suricata/pull/1633/files#r39186975

src/app-layer-template.c
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    SCLogNotice("Transaction ID %"PRIu64" not found.", tx_id);
+    return NULL;
+}
+
+/**
+ * \brief Called by the application layer.
+ *
+ * In mose cases 1 can be returned here.
```
inliniac: typo: mose

src/output-json-template.c
```diff
@@ -0,0 +1,202 @@
...
+        return TM_ECODE_FAILED;
+    }
+
+    templatejs = json_object();
+    if (unlikely(templatejs == NULL)) {
+        goto error;
+    }
+
+    json_object_set_new(templatejs, "request",
+        json_string((char *)templatetx->request_buffer));
```
inliniac: treating the buffers as strings is very poor practice, and the first thing that broke when I tried another proto. It also assumes the string is terminated on the wire, which is highly dangerous.

Think we should switch to this approach: https://github.com/inliniac/suricata/blob/master/src/output-json-dns.c#L96

src/app-layer-template.c
```diff
@@ -0,0 +1,419 @@
...
+    return ALPROTO_UNKNOWN;
+}
+
+static int TemplateParseRequest(Flow *f, void *state, AppLayerParserState *pstate,
+    uint8_t *input, uint32_t input_len, void *local_data)
+{
+    TemplateState *echo = state;
+
+    SCLogNotice("Parsing echo request.");
+
```
inliniac: We can get called with this condition being true:
<pre>
if ((input == NULL || input_len == 0) && AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF)) {
</pre>
So no new data, instead a 'wrap up' signal. See for example: https://github.com/inliniac/suricata/blob/master/src/app-layer-htp.c#L784 where it is used to trigger closing of the htp state.

In this simple example we can just bail if we see the condition. We shouldn't create a tx for it for sure.

scripts/setup-app-layer-detect.sh
```diff
@@ -0,0 +1,233 @@
...
+
+    detect-engine-${protoname_lower}.h detect-engine-${protoname_lower}.c
+
+        The content inspection engine.
+
+Please fix in src/detect-engine-state.h the values for:
+    DE_STATE_FLAG_${protoname_upper}_BUFFER_INSPECT
+    DE_STATE_FLAG_TEMPLATE_BUFFER_INSPECT
+
+Please fix in src/detect-h the values for:
```
inliniac: typo: detect-h should be detect.h

src/app-layer-template.c
jasonish: I think that would make more sense if we perhaps framed each message on a new line or something. But it does take away from the simplicity that is this protocol parser - which is one tx for each chunk of data each way. 

Perhaps a comment - but I think it doesn't match the super trivial protocol that is being implemented here - it's not multi chunk.

I guess I'm open to comments, but it was intentionally like this.

inliniac: I think it should be checked. This template is meant to be copied and used as a basis for a "real" parser. It's not going to be obvious that this buffer is overwritten. So a simple check with a log message will safe everyone a lot of time and trouble.

src/app-layer-template.c
jasonish: Ok, adding this condition.

Can we expect input == NULL || input_len == 0 as well, without the state flag being set?

src/app-layer-template.c
jasonish: Its a little out of scope here, as we just allocated the transaction, instead of looked one up.  Will comment on the matter though.

src/app-layer-template.c
jasonish: Ok, changed the code and added a comment about how you might want to realloc the data instead.

But now I'm thinking that maybe the echo protocol parser should have a little more meat to it.  We could buffer data till a new line is seen and use that as our framing.  But this brings up another question.

I am more inclined to create a buffer on my state object and wait for a complete PDU before allocating a transaction.  I do this in DNP3 for example.  DNP3, and I imagine other protocols then have sub-framing over TCP, where I may need multiple PDUs to complete the request side of a transaction, that makes sense to buffer in the TX as one may seen overlapping "frames" on a TCP session.  For echo we can do the former, the latter is a little harder.

jasonish:
See https://github.com/inliniac/suricata/pull/1634

PR #1634, opened 2015-09-11
App layer template - v7

jasonish:
Addresses issues in https://github.com/inliniac/suricata/pull/1633.

Adds a app-layer decoder level event.  Its a bit bogus, but shows hot to create an app-layer event. In this case an event is raised on an empty echo request - that is a request with just \r\n or \n and no actual content.

The pcap at https://www.dropbox.com/s/screla8lgh2a03t/template-echo.pcap?dl=0 has been updated where the 3rd "tx" is an empty message.

The 4th "tx" is crafted such that the response mismatches the request.  This could be an exercise for the developer to create a new alert or something.  Or perhaps is better for a template alert.

Sorry about the lack of a prscript output this time, wanted to get this out but didn't have time to wait for it at this time.


inliniac:
Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/281
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/283

scripts/setup-app-layer-logger.sh
```diff
@@ -0,0 +1,154 @@
...
+
+    copy_template_file ${src_h} ${dst_h}
+    copy_template_file ${src_c} ${dst_c}
+}
+
+function patch_makefile_am() {
+    filename="src/Makefile.am"
+    echo "Patching ${filename}."
+    ed -s ${filename} > /dev/null <<EOF
+/output-json-template.c
```
inliniac: This seems to miss the .h file, so we get a line: output-json-xxx.c output-json-template.h \
which makes make distcheck fail.

jasonish: https://github.com/inliniac/suricata/pull/1643

src/app-layer-template.c
```diff
@@ -0,0 +1,533 @@
...
+    DetectEngineState *s)
+{
+    TemplateTransaction *tx = vtx;
+    tx->de_state = s;
+    return 0;
+}
+
+void RegisterTemplateParsers(void)
+{
+    char *proto_name = "template";
```
inliniac: We saw at least one case there this wasn't updated, can you double check the setup script logic for it?

jasonish: I'm surprised, this one is handled by very basic sed usage:

    sed -e "s/TEMPLATE/${protoname_upper}/g" \
	-e "s/template/${protoname_lower}/g" \
	-e "s/Template/${protoname}/g" > ${dst} < ${src}

However, I have found that different versions of sed, tr and some of the other tools to produce different results.  But this usage should be sane on Linux at least.

PR #1635, opened 2015-09-11
Flowvar eve dump

andersjak68:
This is slightly a work in progress, but it is (to me) useful in its current state as it allows to use lua to extract parts of the payload data and inspect them in the eve log.  My next aim is to improve what is done here by adding flowvar names also into the eve log,  but this seems a bit involved as the variable names exist only in the detect engine context, and not in the flow->flowvar list.

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -64,6 +64,9 @@
 #include "util-logopenfile.h"
 #include "util-crypt.h"
 
+#include "flow-var.h"
+#include "pkt-var.h"
+
 #define MODULE_NAME "JsonAlertLog"
 
 #ifdef HAVE_LIBJANSSON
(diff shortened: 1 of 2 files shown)


inliniac:
Check fails because of "Banned function 'sprintf' used at /home/travis/build/inliniac/suricata/src/output-json-alert.c:180"

inliniac:
What you're trying to do here is a bit tricky because of the missing names of flowvars. These are only available in the detection engine (de_ctx), which is really not available during logging. There is a hacky and expensive workaround in detect.c::AlertDebugLogModeSyncFlowbitsNamesToPacketStruct().

A better solution would be to detach the varname <-> varid store from the detect engine, but that would be tricky when updating it during rule reloads.

inliniac:
In general, please have a look at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing

PR #1637, opened 2015-09-14
configure: use pkg_config for libhtp

regit:
It was not possible to simply specify PKG_CONFIG_PATH to build
with an non bundled libhtp. With this patch we don't need anymore
the htp lib and include configure options.

This patch will work if https://github.com/OISF/libhtp/pull/113 is merged in libhtp.


regit:
Some default in PR. Resubmitting.

PR #1640, opened 2015-09-14
app-layer template: add mpm for buffer -v1

jasonish:
Adds MPM for decoded application layer buffer.

NOTE: The changes to detect-fast-pattern.c were required
for the MPM to be called, but the fast_pattern keyword
errors out at the moment. I'm not sure whats going on here right now.

Is the fast pattern stuff required for the MPM?

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/127
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/128

### DIFF
--- a/src/suricata-common.h
+++ b/src/suricata-common.h
@@ -314,6 +314,7 @@ typedef enum PacketProfileDetectId_ {
     PROF_DETECT_CLEANUP,
     PROF_DETECT_GETSGH,
     PROF_DETECT_MPM_FD_SMTP,
+    PROF_DETECT_MPM_TEMPLATE_BUFFER,
 
     PROF_DETECT_SIZE,
 } PacketProfileDetectId;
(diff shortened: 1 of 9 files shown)


src/suricata-common.h
```diff
@@ -314,6 +314,7 @@ typedef enum PacketProfileDetectId_ {
     PROF_DETECT_CLEANUP,
     PROF_DETECT_GETSGH,
     PROF_DETECT_MPM_FD_SMTP,
+    PROF_DETECT_MPM_TEMPLATE_BUFFER,
```
inliniac: This should have an accompanying util-profile.c update in PacketProfileDetectIdToString(). Otherwise it looks good. Was very helpful!

PR #1646, opened 2015-09-17
app-layer detect template: move more setup into detect module

jasonish:
Move the setting of some masks into the detect setup to limit code changes outside of the detect module.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/131
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/132

### DIFF
--- a/src/detect-template-buffer.c
+++ b/src/detect-template-buffer.c
@@ -48,6 +48,9 @@ static int DetectTemplateBufferSetup(DetectEngineCtx *de_ctx, Signature *s,
 {
     s->list = DETECT_SM_LIST_TEMPLATE_BUFFER_MATCH;
     s->alproto = ALPROTO_TEMPLATE;
+    s->flags |= SIG_FLAG_STATE_MATCH;
+    s->mask |= SIG_MASK_REQUIRE_TEMPLATE_STATE;
+    s->mask |= SIG_MASK_REQUIRE_FLOW;
     return 0;
 }
 
--- a/scripts/setup-app-layer-detect.sh
+++ b/scripts/setup-app-layer-detect.sh
@@ -129,18 +129,6 @@ w
 EOF
 }
 
-function patch_detect_parse_c() {
-    filename="src/detect-parse.c"
-    echo "Patching ${filename}."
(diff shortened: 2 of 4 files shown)


src/detect-template-buffer.c
```diff
@@ -48,6 +48,9 @@ static int DetectTemplateBufferSetup(DetectEngineCtx *de_ctx, Signature *s,
 {
     s->list = DETECT_SM_LIST_TEMPLATE_BUFFER_MATCH;
     s->alproto = ALPROTO_TEMPLATE;
+    s->flags |= SIG_FLAG_STATE_MATCH;
+    s->mask |= SIG_MASK_REQUIRE_TEMPLATE_STATE;
```
inliniac: I'm not convinced this is a good idea. The whole logic is currently all in src/detect.c, and scattering it over the keywords would not necessarily make it easier. Besides, I think if we decide to do it, we should do it for all and not just for this one. But for now I vote against it.

src/detect-template-buffer.c
```diff
@@ -48,6 +48,9 @@ static int DetectTemplateBufferSetup(DetectEngineCtx *de_ctx, Signature *s,
 {
     s->list = DETECT_SM_LIST_TEMPLATE_BUFFER_MATCH;
     s->alproto = ALPROTO_TEMPLATE;
+    s->flags |= SIG_FLAG_STATE_MATCH;
```
inliniac: I think if we do this, it should be done globally, so for the other keywords as well.

inliniac: One way would be to make it part of the register function, but I guess there are some corner cases to consider. There are some keywords that depending on how and where they are used (e.g. lua) should have this flag or not.

src/detect-template-buffer.c
jasonish: Just keep in mind that to make a detect module dynamically loadable, its setup code will need to consolidated. Not to mention, its easier to document how to make a new detect module.

Yes, it would be best to do for all at once, this was more to show the idea.

PR #1524, opened 2015-05-30
erspan v2

inliniac:
As https://github.com/inliniac/suricata/pull/1498, but with added counter.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/231
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/233

PR #1648, opened 2015-09-21
Threads rename v7 - Bug #1524

zrasmor:
IPS thread names may need a little re-work, but wanted to submit this as a first cut. 
IDS thread names should be pretty consistent with what was decided.

### DIFF
--- a/src/flow-manager.c
+++ b/src/flow-manager.c
@@ -119,7 +119,7 @@ void FlowDisableFlowManagerThread(void)
     tv = tv_root[TVT_MGMT];
 
     while (tv != NULL) {
-        if (strcasecmp(tv->name, "FlowManagerThread") == 0) {
+        if (strncasecmp(tv->name, "FM-", 3) == 0) {
             TmThreadsSetFlag(tv, THV_KILL);
             cnt++;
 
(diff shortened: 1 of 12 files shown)


inliniac:
Finally had some time to look at this, sorry for the delay.

Couple of things I noticed:
- offline (pcap file) modes not yet done
- 'single' doesn't seem to add the iface: "thread "W01" closed on initialization" (when using -i eth0). This should be W01-eth0, right?
- in IPS we use RX/TX in some cases, I think I like RX more than 'C' used in IDS mode elsewhere.
- maybe we can have the names be globally defined instead of hardcoded into all the runmodes, that would make changing it easier.
- Minor: originally I had in mind to use something like 'C#01', adding the #. Thats just cosmetics though.

zrasmor:
Replies to each bullet: 
* Oops, I missed offline pcap mode, i'll add it.
* Yes, this was a question I had - do we want the interface name in every thread name? I adopted whatever the current convention was. I don't believe workers/single appended the interface name previously.
* So I'm clear, are you suggesting that we do away with C completely, replacing it with RX?
* Totally agree about global definitions. Do you have a suggestion for while file to add the definitions to? If not, I can look around and make a suggestion.
* Sure, I can add the '#', that's no problem

inliniac:
* "do we want the interface name in every thread name" -> yeah, where it makes sense. So for workers and single it does, since they are always attached to a iface. The autofp workers are not.

* "do away with C completely, replacing it with RX" -> yeah, but the global defines make it trivial

* "suggestion for while file to add the definitions" -> maybe runmodes.h?

inliniac:
Please close a PR once you've submitted a new one to replace it.

Replaced by #1735

PR #1649, opened 2015-09-21
fix util-host-os-info unit tests.

jasonish:
I had a PR for this, but it got closed when I accidentally deleted the remote branch: https://github.com/inliniac/suricata/pull/1441
 
- Some tests always passed due to the result value being
  initialized to 1.

- Fix tests that now fail.  Looks like just the test were wrong
  and that the code does the right thing.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/133
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/134

### DIFF
--- a/src/util-host-os-info.c
+++ b/src/util-host-os-info.c
@@ -450,12 +450,12 @@ int SCHInfoTestInvalidIPV4Address02(void)
 {
     SCHInfoCreateContextBackup();
 
-    int result = 1;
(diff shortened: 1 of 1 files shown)


inliniac:
Merged through #1675, thanks!

PR #1650, opened 2015-09-21
magic: prefer brew magic on OS X if exists

jasonish:
I'm not sure if this is the best way to do this, but on OS X, magic.h doesn't exist by default.  One way to add it is libmagic from homebrew, however, the path to magic has to be updated as Suricata will fail if it loads the magic file included by OS X.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/134
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/135

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -231,6 +231,9 @@
             CFLAGS="${CFLAGS} -DOS_DARWIN"
             CPPFLAGS="${CPPFLAGS} -I/opt/local/include"
             LDFLAGS="${LDFLAGS} -L/opt/local/lib"
+	    if test -e /usr/local/share/misc/magic.mgc; then
+	       e_magic_file=/usr/local/share/misc/magic
+	    fi
             ;;
         *-*-linux*)
             #for now do nothing


configure.ac
```diff
@@ -231,6 +231,9 @@
             CFLAGS="${CFLAGS} -DOS_DARWIN"
             CPPFLAGS="${CPPFLAGS} -I/opt/local/include"
             LDFLAGS="${LDFLAGS} -L/opt/local/lib"
+	    if test -e /usr/local/share/misc/magic.mgc; then
```
inliniac: indent looks one-off in github?

jasonish:
> indent looks one-off in github?

Will fix.  Otherwise, is this an OK thing for configure.ac?

inliniac:
it's not pretty, so if you know a cleaner way... does brew have a commandline tool that you can use to get to the same info?

jasonish:
Something like this perhaps:

            AC_CHECK_PROG(HAVE_BREW, brew, "yes")
            if test "$HAVE_BREW" = "yes"; then
                AC_CHECK_FILE($(brew --prefix)/share/misc/magic.mgc,
                    e_magic_file=$(brew --prefix)/share/misc/magic.mgc)
            fi

But running it in the same spot as the current checke messes with the ok output for "checking host os".

inliniac:
Not terribly concerned about that output. Doesn't brew have a quiet mode or something similar?

jasonish:
> Not terribly concerned about that output. Doesn't brew have a quiet mode or something similar?

Its not brew output, its autoconf output:

checking host os... -n installation for x86_64-apple-darwin15.0.0 OS...
checking for brew... yes
checking for /usr/local/share/misc/magic.mgc... yes
ok

inliniac:
AC_CHECK_PROG(HAVE_BREW, brew, "yes") should probably elsewhere, but the other output... meh. Not very important.

jasonish:
#1823 provides a better solution for this.

PR #1651, opened 2015-09-21
GPL license sync with official gpl-2.0.txt

jasonish:
Synced up with:
http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

I think these should be consistent with each other, and neither were 100% consistent with the official copy.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/1540


inliniac:
Merged through #1675, thanks!

PR #1653, opened 2015-09-22
detect-pcre: allow upper cases metachars for host (/W) - v1

jasonish:
Redmine issue https://redmine.openinfosecfoundation.org/issues/1490

Probably not exhaustive, but will allow upper case meta characters where the current implementation won't.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/136
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/137

### DIFF
--- a/src/detect-pcre.c
+++ b/src/detect-pcre.c
@@ -274,6 +274,26 @@ static int DetectPcreSetList(int list, int set)
     return set;
 }
 
+static int DetectPcreHasUpperCase(const char *re)
+{
+    size_t len = strlen(re);
+    int is_meta = 0;
+
+    for (size_t i = 0; i < len; i++) {
+        if (is_meta) {
+            is_meta = 0;
+        }
+        else if (re[i] == '\\') {
+            is_meta = 1;
+        }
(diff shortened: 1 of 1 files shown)


src/detect-pcre.c
```diff
@@ -274,6 +274,26 @@ static int DetectPcreSetList(int list, int set)
     return set;
 }
 
+static int DetectPcreHasUpperCase(const char *re)
+{
+    size_t len = strlen(re);
+    int is_meta = 0;
+
+    for (size_t i = 0; i < len; i++) {
```
inliniac: using i for a unsigned var... not liking :)

jasonish: Honestly I've never thought of i conveying any information about its type, but instead I use it to convey info about what it is, an index ;)

inliniac:
Merged into 2.0.x through #1659

inliniac:
Merged into master through #1675, thanks!

PR #1658, opened 2015-09-24
util-debug: don't colorize if a redirect is used

regit:
It is better to disable the color mode when a redirect of stderr
is done to avoid getting colorized output in the generated file.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/101
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/99

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/1562


inliniac:
Merged through #1675, thanks!

PR #1660, opened 2015-09-25
[preview] mixed-mode

glongo:
The idea is to add a new runmode named "netfilter",
which takes care to start nflog and nfq runmodes.

Test case, before to run suri:
iptables -A INPUT -j NFLOG --nflog-group 2

then:
suricata --mixed-mode -q 0 --nflog -v

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/66
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/65


src/decode.h
```diff
@@ -351,6 +351,8 @@ typedef struct PktProfiling_ {
 /* forward declartion since Packet struct definition requires this */
 struct PacketQueue_;
 
+enum mode {IDS, IPS, AUTO};
```
inliniac: this enum will need a proper name and explanation of it's purpose

src/runmode-netfilter.c
```diff
@@ -0,0 +1,139 @@
...
+                              "Workers netfilter mode",
+                              RunModeNetfilterWorkers);
+    return;
+}
+
+int RunModeNetfilterAutoFp(void)
+{
+    SCEnter();
+
+#if defined(HAVE_NFLOG) || defined(NFQ)
```
inliniac: should be AND. The mixed mode makes no sense if only one part is supported

src/stream-tcp.c
```diff
@@ -120,7 +120,7 @@ uint64_t StreamTcpReassembleMemuseGlobalCounter(void);
 SC_ATOMIC_DECLARE(uint64_t, st_memuse);
 
 /* stream engine running in "inline" mode. */
-int stream_inline = 0;
+enum mode stream_inline;
```
inliniac: here you can see why the name of the enum is important

src/util-device.c
```diff
@@ -54,6 +54,11 @@ int LiveRegisterDevice(char *dev)
         SCFree(pd);
         return -1;
     }
+    pd->runmode = SCStrdup(runmode);
```
inliniac: why use a copy? it's a constant hardcoded string that is passed around

src/util-device.c
```diff
@@ -42,7 +42,7 @@ static int live_devices_stats = 1;
  *  \retval 0 on success.
  *  \retval -1 on failure.
  */
-int LiveRegisterDevice(char *dev)
+int LiveRegisterDevice(char *dev, char *runmode)
```
inliniac: runmode should be a const

inliniac:
Not sure I like the '--mixed-mode' option. It seems redundant. Just specifying the -q and --nflog options could accomplish the same?

inliniac:
Think it would be good to support a AF_PACKET mixed setup as well. That should make more clear which parts should be generalized.

PR #1662, opened 2015-09-27
unix-manager: log client (dis)connection with DEBUG level.

gozzy:
Also select/receive errors are logged as ERROR.


gozzy:
I think that (dis)connection should be logged with DEBUG level. If one polls suricata periodically, a log will be filled with unneeded messages. At the same time some critical errors were really INFOrmational - not good.

inliniac:
Merged through #1675, thanks!

PR #1663, opened 2015-09-28
Capture full session on alert - Feature #120

meatchicken:
In most signature based IDS's, when a signature fires, they have the
ability to dump the packet data for which the signature fired upon
and subsequent packet data, usually within the same network stream.
Unfortunately, this can mean important information may not be
available to a user if the information occured prior to the packet
containing the signature.  For example, if a signature fires on an
HTTP response, it is possible that important information in an
HTTP request may not be available but could be useful in
investigations.

TimeMachine is a fast, memory-efficient solution for storing network
traffic for retrieval after an alarm fires and dumping all stored
captured traffic to disk.

When TimeMachine is enabled, each detection thread creates one or more
pre-allocated in-memory heaps based on a configurable packet size and
configurable amount of memory to use.

When a new packet is processed by a given thread, several actions may
take place:

  -  In cases where TimeMachine does not know of an existing flow
     associated with a given packet, TimeMachine will generate a
     new minimal flow stored within a splay tree, store the packet
     data in one of the pre-allocated in-memory heaps and associate
     the packet to the given flow.

  -  In cases where TimeMachine already knows of an existing flow,
     it once again stores the data in the pre-allocated in-memory
     heap and associates the packet to the given flow.  This process
     also rebalances the splay tree ensuring the most recent flows
     should take less time to find within the tree structure

  -  In cases where the packet is associated with a given alert
     generated by the detection engine, a new output file will be
     opened and all data associated with the packet's flow will be
     dumped to disk in chronological order.

  -  In cases where an output file already exists indicating the
     flow was previously associated with an alert, packets are not
     stored within the in-memory heap and instead are written
     directly to disk.

As the heaps are fixed size, when a given heap is full, TimeMachine
will purge the oldest packet first.  In cases where the packet is
the last item associated with a given flow, TimeMachine will also
clean up the given flow information.

TimeMachine is an effective alternative in cases where full packet
capture is not available due to disk or IOP's constraints.  The
overall effectiveness of TimeMachine is dependent upon the overall
network speed, size of heaps configured for TimeMachine and total
memory available.

This is an implementation to address feature request #120, capture
full session on alert:

https://redmine.openinfosecfoundation.org/issues/120


inliniac:
@jasonish could you do an initial review of this?

configure.ac
```diff
@@ -951,6 +951,30 @@
...
+        echo "   ERROR!  libbsd-dev files not found, go get them"
+        echo
+        echo "   Ubuntu: apt-get install libbsd-dev"
+        echo "   Fedora: yum install libbsd-devel"
+        echo
+        exit 1
+    fi   
+     
+    AC_DEFINE([HAVE_SYS_TREE_H],[1],[bsd/sys/tree.h available])
+
```
jasonish: I wonder if we should just include tree.h with Suricata like we do queue.h...  Or if depending on libbsd, we could drop queue.h and use the one from libbsd.

src/timemachine.c
```diff
@@ -0,0 +1,613 @@
...
+        return TM_ECODE_OK;
+    } 
+    
+    /* find the queue the packet should fall in */
+    TAILQ_FOREACH(heap, &td->heaps, next) {
+        if (pkthdr.caplen <= heap->conf->max_packet_size) {
+            break;
+        }
+    }
+
```
jasonish: If pkthdr.caplen is > 1500, heap will be undefined here causing a segv at line 160.

src/timemachine.c
```diff
@@ -0,0 +1,613 @@
...
+        SCMutexLock(&tm->tm_lock);   
+        if (TimeMachineHeapCanExpand(tm)) {
+            TimeMachineHeapExpand(tm, heap);
+            SCMutexUnlock(&tm->tm_lock);
+        }
+        /* couldn't expand, need to remove the first entry */
+        else {
+            SCMutexUnlock(&tm->tm_lock);
+           
+            TimeMachineFlow* rem_flow;
```
jasonish: Not shown in the diff on github, but there are a few lines of extra spaces here.

configure.ac
meatchicken: bsd/sys/queue.h is a subset of sys/queue.h though the only difference is
the CIRCLEQ data structure.  Otherwise I believe it would be just a drop in
replacement.

inliniac:
Can you say something about how the concept relates to the threading? A quick glance suggests the use of a global lock. If true, I fear this is going to be a major contention point especially considering the relatively high cost of copying/storing data.

meatchicken:
The timemachine heaps are per thread and can dynamically grow until some
maximum amount of system memory you define within the config file.  The
lock you see within performance critical section is around this check +
growth.  I was actually thinking a few days ago whether it would make sense
to remove this.  The tradeoffs are performance for locking vs not being
able to dynamically grow heaps to match the number of packets that are
received by a given thread.  Ultimately, each thread would have a different
amount of time it could go back to dump traffic.  Removing the lock and
having static allocation of heap sizes may be a good tradeoff in the sense
that it would simplify the code and make this completely lockless.  I'd
probably add more metrics around how far back each thread could go to
validate whether your heap allocation sizes are correct.  Thoughts?

src/timemachine.c
```diff
@@ -0,0 +1,613 @@
...
+    TimeGet(&current_time);
+    
+    TimeMachineThreadData *td = (TimeMachineThreadData *)thread_data;
+    TimeMachineData *tm = td->tm_data;
+    
+    /* make sure packet has real data */
+    if (GET_PKT_PAYLOAD_LEN(p) == 0) {
+        return TM_ECODE_OK;
+    }
+
```
jasonish: In TCP, doesn't this suppress the handshake setup?  I did run it for a while, and noticed my output pcaps were missing session setup. Is that by design?

src/timemachine.c
```diff
@@ -0,0 +1,613 @@
...
+    TimeMachineData *tm = td->tm_data;
+    
+    /* make sure packet has real data */
+    if (GET_PKT_PAYLOAD_LEN(p) == 0) {
+        return TM_ECODE_OK;
+    }
+    
+    /* check whether a corresponding flow already exists */
+    if (!(flow = TimeMachineFlowLookup(td->flows, p))) {
+        flow = TimeMachineFlowNew(td->flows, p);
```
jasonish: TimeMachineFlowNew will exit instead of return NULL on memory allocation failure. Exits should not occur in the packet path.  Instead, TimeMachineProcess should return TM_ECODE_FAILED.

jasonish:
Why does TimeMachine create its own flash hash? Could the already computed flow hash be reused?

Any consideration to pseudo-packets, or tunneled packets, and if the original packet should be logged, or the inner packet?

src/timemachine.c
meatchicken: Yes, that's by design.  We could add the handshake though at the cost of
the amount of packets that can be saved.  For very short lived connections,
this can cause a lot of pollution within the heap that I didn't see the
value in having and ultimately takes up space from the heap of more
interesting traffic.

meatchicken:
Initially it was to save a bit of space.  Though, at the cost of
re-computation, it may be an ok tradeoff.  I'll update the patch to use the
existing flow hash information vs specific to TM.  I didn't do any
consideration for pseudo-packets, tunneled packets, etc. Generally my
assumption would be the original packet would be logged.

src/timemachine.c
meatchicken: Cool, will update this as well.  Thanks!

configure.ac
inliniac: Think we have added one extra macro to the TAILQ implementation. But we could probably put that elsewhere.

inliniac:
Would you be interested in making this more generic? Storing pcaps is a great use case, but I would really love to see this be usable to other parts of the engine as well. E.g. in our JSON logging where we would want to go back and log as much as possible from the time machine. Thoughts?

inliniac:
I'm not a great fan of having a parallel flow tracking engine for this purpose, so I would like to consider how the existing flow engine could be used here. It may make managing the resources more tricky?

meatchicken:
I'm totally fine making this more generic.  I was originally thinking of adding this directly into the flow tracking engine but decided not to.  Basically this related to the way flow manager handled timeouts for flows I was thinking there could be a case where flow manager may time out or reach capacity limits but time machine may still want packets to log for a longer period of time.  Think like a very long session (maybe multiple hours?) where time machine would still want to dump all packets from that session for the entire length of the session.  This may not be an issue or there may be a way around this by stating "this flow is tagged as using time machine, keep dumping and keep the flow in memory regardless of timeout".  Feel free to correct me if I'm wrong in this analysis.

Resource tracking does become more tricky but I don't think it's significantly different from some of the changes I've already made (not committed).  In addition to the above paragraph, there are items around time machine heap resource management per thread.  Instead of tracking TM memory usage globally (reference above comments), I moved the resource memory limitations of each heap to be per thread instead of global so there is no more locking.  What this means is each thread would have to manage how much memory it has per all heaps it contains in it's thread and then be able to dynamically adjust accordingly (e.g. move memory from a heap with large packets to a heap with smaller packets).  I think this is ok but this makes the assumption that packets are pretty evenly spread across all threads instead of one flow having significantly more than all others skewing results.  

I'm in the process of adding some instrumentation to determine how far back a single thread can go via the data contained in the TimeMachine heaps (looks like all stats are aggregated and no easy way to do per thread at the moment?), which should help give better visibility into the assumptions above to better size memory allocation for TM.

meatchicken:
Didn't mean to merge, will re-issue a pull request.

PR #1664, opened 2015-09-28
Feature 1527: ability to compile as a position independent executable

gozzy:
Adds corresponding configure option which enables proper CPPFLAGS and LDFLAGS.

Seems to compile well, readelf and hardening-check script report that the binary is position independent.


inliniac:
Merged through #1675, thanks!

PR #1665, opened 2015-09-30
lua: TLS lua output support

thus:
```lua
function init (args)
    local needs = {}
    needs["protocol"] = "tls"
    return needs
end

function setup (args)
end

function log (args)
    ts = SCPacketTimeString()

    if version == nil then
        return 0
    end

    file:flush()
end

function deinit (args)
    file:close(file)
end
```

### DIFF
--- a/src/output-lua.c
+++ b/src/output-lua.c
@@ -40,6 +40,7 @@
 #include "output.h"
 #include "app-layer-htp.h"
 #include "app-layer.h"
+#include "app-layer-ssl.h"
 #include "app-layer-parser.h"
 #include "util-privs.h"
 #include "util-buffer.h"
@@ -57,6 +58,7 @@
 #include "util-lua-common.h"
 #include "util-lua-http.h"
 #include "util-lua-dns.h"
+#include "util-lua-tls.h"
 
 #define MODULE_NAME "LuaLog"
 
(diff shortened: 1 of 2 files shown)


src/output-lua.c
```diff
@@ -229,6 +231,106 @@ static int LuaPacketConditionAlerts(ThreadVars *tv, const Packet *p)
...
+
+    if (!(PKT_IS_IPV4(p)) && !(PKT_IS_IPV6(p))) {
+        goto end;
+    }
+
+    uint16_t proto = FlowGetAppProtocol(p->flow);
+    if (proto != ALPROTO_TLS)
+        goto end;
+
+    SSLState *ssl_state = (SSLState *)FlowGetAppState(p->flow);
```
inliniac: Flow access is done w/o lock throughout this function.

src/output-lua.c
```diff
@@ -229,6 +231,106 @@ static int LuaPacketConditionAlerts(ThreadVars *tv, const Packet *p)
...
+ *  The script is called once for each packet.
+ */
+static int LuaPacketLoggerTls(ThreadVars *tv, void *thread_data, const Packet *p)
+{
+    LogLuaThreadCtx *td = (LogLuaThreadCtx *)thread_data;
+
+    char timebuf[64];
+    CreateTimeString(&p->ts, timebuf, sizeof(timebuf));
+
+    SCMutexLock(&td->lua_ctx->m);
```
inliniac: Think you can take this lock later. To just above line 283.

IIRC we need it since the lua states are global. Might be good to change that at some point as well.

thus:
Updated pull request with lock changes suggested.

src/output-lua.c
inliniac: Still an issue. Any access into the flow must be locked.

PR #1666, opened 2015-10-02
lua: SSH support

thus:

Example detection script:
```lua
function init (args)
    local needs = {}
    needs["ssh"] = tostring(true)
    return needs
end

function match (args)

        return 0
    end

    SCLogInfo(str);
    return 1
end
```

Example output script:
```lua
function init (args)
    local needs = {}
    needs["protocol"] = "ssh" 
    return needs
end

function setup (args)
end

function log (args)
    ts = SCPacketTimeString()

        return
    end 

               ss .. "'\n")
    file:flush()
end

function deinit (args)
    file:close(file)
end
```

### DIFF
--- a/src/output-lua.c
+++ b/src/output-lua.c
@@ -40,6 +40,7 @@
 #include "output.h"
 #include "app-layer-htp.h"
 #include "app-layer.h"
+#include "app-layer-ssh.h"
 #include "app-layer-parser.h"
 #include "util-privs.h"
 #include "util-buffer.h"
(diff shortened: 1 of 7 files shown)


src/output-lua.c
```diff
@@ -229,6 +231,97 @@ static int LuaPacketConditionAlerts(ThreadVars *tv, const Packet *p)
...
+    LogLuaThreadCtx *td = (LogLuaThreadCtx *)thread_data;
+
+    char timebuf[64];
+    CreateTimeString(&p->ts, timebuf, sizeof(timebuf));
+
+    if (!(PKT_IS_IPV4(p)) && !(PKT_IS_IPV6(p))) {
+        goto end;
+    }
+
+    uint16_t proto = FlowGetAppProtocol(p->flow);
```
inliniac: Goes into the flow w/o lock.

inliniac:
Needs some locking improvements. Looks good otherwise!

PR #1668, opened 2015-10-05
lua: SSH support

thus:

Example detection script:
```lua
function init (args)
    local needs = {}
    needs["ssh"] = tostring(true)
    return needs
end

function match (args)

        return 0
    end

    SCLogInfo(str);
    return 1
end
```

Example output script:
```lua
function init (args)
    local needs = {}
    needs["protocol"] = "ssh" 
    return needs
end

function setup (args)
end

function log (args)
    ts = SCPacketTimeString()

        return
    end 

               ss .. "'\n")
    file:flush()
end

function deinit (args)
    file:close(file)
end
```

### DIFF
--- a/src/output-lua.c
+++ b/src/output-lua.c
@@ -40,6 +40,7 @@
 #include "output.h"
 #include "app-layer-htp.h"
 #include "app-layer.h"
+#include "app-layer-ssh.h"
 #include "app-layer-parser.h"
 #include "util-privs.h"
 #include "util-buffer.h"
@@ -57,6 +58,7 @@
 #include "util-lua-common.h"
 #include "util-lua-http.h"
 #include "util-lua-dns.h"
+#include "util-lua-ssh.h"
 
 #define MODULE_NAME "LuaLog"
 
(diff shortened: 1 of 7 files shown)


thus:
It wouldn't log if I held both the flow lock and the mutex lock at the same time, so I had to change the code a bit.

src/output-lua.c
```diff
@@ -229,6 +231,99 @@ static int LuaPacketConditionAlerts(ThreadVars *tv, const Packet *p)
...
+    if (unlikely(ssh_state == NULL)) {
+        goto dontlog;
+    }
+
+    if (ssh_state->cli_hdr.software_version == NULL ||
+            ssh_state->srv_hdr.software_version == NULL)
+        goto dontlog;
+
+    /* We only log the state once */
+    if (ssh_state->cli_hdr.flags & SSH_FLAG_STATE_LOGGED_LUA)
```
inliniac: Shouldn't this be done in the Condition function below? That's where we check if we need to log at all.

src/output-lua.c
```diff
@@ -229,6 +231,99 @@ static int LuaPacketConditionAlerts(ThreadVars *tv, const Packet *p)
...
+    ssh_state->cli_hdr.flags |= SSH_FLAG_STATE_LOGGED_LUA;
+    FLOWLOCK_UNLOCK(p->flow);
+
+    SCMutexLock(&td->lua_ctx->m);
+
+    lua_getglobal(td->lua_ctx->luastate, "log");
+
+    LuaStateSetThreadVars(td->lua_ctx->luastate, tv);
+    LuaStateSetPacket(td->lua_ctx->luastate, (Packet *)p);
+    LuaStateSetFlow(td->lua_ctx->luastate, p->flow, /* unlocked */LUA_FLOW_NOT_LOCKED_BY_PARENT);
```
inliniac: When you say "It wouldn't log if I held both the flow lock and the mutex lock at the same time, so I had to change the code a bit.", did you update this "/* unlocked */LUA_FLOW_NOT_LOCKED_BY_PARENT" too?

src/output-lua.c
thus: This was how it was done in "src/output-json-ssh.c", but I agree that it would be better suited to put all the checks in the condition function. Maybe we could avoid the flow lock in this function then?

src/output-lua.c
thus: I missed that, thanks! I should know that if it doesn't work, it's probably user error :)

src/output-lua.c
thus: I've tried implementing what I meant in #1670.

PR #1669, opened 2015-10-05
lua: TLS output support

thus:

```lua
function init (args)
    local needs = {}
    needs["protocol"] = "tls"
    return needs
end

function setup (args)
end

function log (args)
    ts = SCPacketTimeString()

    if version == nil then
        return 0
    end

    file:flush()
end

function deinit (args)
    file:close(file)
end
```

### DIFF
--- a/src/app-layer-ssl.h
+++ b/src/app-layer-ssl.h
@@ -81,6 +81,8 @@ enum {
 /* flags for file storage */
 #define SSL_AL_FLAG_STATE_STORED                0x40000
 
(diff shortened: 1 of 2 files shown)


thus:
I had to change the code a bit (as in #1668), since it wouldn't log when I held the flow lock and the mutex lock at the same time.

thus:
This needs to be changed to fix the same errors as in #1668. Closing the pull request while I await comments on #1670.

PR #1670, opened 2015-10-05
lua: SSH support (v2)

thus:

Example detection script:
```lua
function init (args)
    local needs = {}
    needs["ssh"] = tostring(true)
    return needs
end

function match (args)

        return 0
    end

    SCLogInfo(str);
    return 1
end
```

Example output script:
```lua
function init (args)
    local needs = {}
    needs["protocol"] = "ssh" 
    return needs
end

function setup (args)
end

function log (args)
    ts = SCPacketTimeString()

        return
    end 

               ss .. "'\n")
    file:flush()
end

function deinit (args)
    file:close(file)
end
```

### DIFF
--- a/src/output-lua.c
+++ b/src/output-lua.c
@@ -40,6 +40,7 @@
 #include "output.h"
 #include "app-layer-htp.h"
 #include "app-layer.h"
+#include "app-layer-ssh.h"
 #include "app-layer-parser.h"
 #include "util-privs.h"
 #include "util-buffer.h"
@@ -57,6 +58,7 @@
 #include "util-lua-common.h"
 #include "util-lua-http.h"
 #include "util-lua-dns.h"
+#include "util-lua-ssh.h"
 
 #define MODULE_NAME "LuaLog"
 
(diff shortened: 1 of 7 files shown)


src/output-lua.c
```diff
@@ -229,6 +231,81 @@ static int LuaPacketConditionAlerts(ThreadVars *tv, const Packet *p)
...
+
+    if (ssh_state->cli_hdr.software_version == NULL ||
+        ssh_state->srv_hdr.software_version == NULL)
+        goto dontlog;
+
+    /* We only log the state once */
+    if (ssh_state->cli_hdr.flags & SSH_FLAG_STATE_LOGGED_LUA)
+        goto dontlog;
+
+    ssh_state->cli_hdr.flags |= SSH_FLAG_STATE_LOGGED_LUA;
```
inliniac: The idea of the Condition function was to just check things, not modify things. I do like the clean logic of the flags handling this way though.

I guess one thing to consider is that the Condition function now always takes a write lock, even though it's very likely that it's only going to return 'true' in a small percentage of the cases. So I think I would like the flag 'set' to move to the LuaPacketLoggerSsh function, where we can wrap it in a WRLOCK while keeping the condition func at the RDLOCK.

thus: Good point, Victor! I'll change it and submit a new pull request.

PR #1672, opened 2015-10-07
lua: SSH support (v3)

thus:

Example detection script:
```lua
function init (args)
    local needs = {}
    needs["ssh"] = tostring(true)
    return needs
end

function match (args)

        return 0
    end

    SCLogInfo(str);
    return 1
end
```

Example output script:
```lua
function init (args)
    local needs = {}
    needs["protocol"] = "ssh" 
    return needs
end

function setup (args)
end

function log (args)
    ts = SCPacketTimeString()

        return
    end 

               ss .. "'\n")
    file:flush()
end

function deinit (args)
    file:close(file)
end
```

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -356,6 +356,7 @@ util-lua-common.c util-lua-common.h \
 util-lua-dns.c util-lua-dns.h \
 util-lua-http.c util-lua-http.h \
 util-lua-tls.c util-lua-tls.h \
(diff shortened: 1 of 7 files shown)


inliniac:
Looks good to me.

inliniac:
Can you rebase this to the current master? I got some ugly conflicts when trying to merge both TLS and SSH support.

thus:
That's quite understandable. I'll do it first thing tomorrow.

PR #1674, opened 2015-10-08
lua: TLS output support

thus:

```lua
function init (args)
    local needs = {}
    needs["protocol"] = "tls"
    return needs
end

function setup (args)
end

function log (args)
    ts = SCPacketTimeString()

    if version == nil then
        return 0
    end

    file:flush()
end

function deinit (args)
    file:close(file)
end
```

### DIFF
--- a/src/app-layer-ssl.h
+++ b/src/app-layer-ssl.h
@@ -81,6 +81,8 @@ enum {
 /* flags for file storage */
 #define SSL_AL_FLAG_STATE_STORED                0x40000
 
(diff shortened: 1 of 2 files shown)


thus:
Testing using my test rig with live corporate traffic since yesterday gives the exact same amount of log lines as the tls log.

inliniac:
Merged through #1675, thanks!

thus:
Thanks, Victor!

PR #1676, opened 2015-10-08
app-layer-template: disable by default - v1

jasonish:
Disables the app-layer templates by default if the app-layer template protocol configuration node doesn't exist.

The template setup scripts will remove this code on setup so new app-layer's are enabled by default like most other app-layer protocols.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/145
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/146

### DIFF
--- a/scripts/setup-app-layer-detect.sh
+++ b/scripts/setup-app-layer-detect.sh
@@ -34,9 +34,11 @@ function copy_template_file() {
 
     echo "Creating ${dst}."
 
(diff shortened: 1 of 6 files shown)


inliniac:
Merged through #1699, thanks!

PR #1678, opened 2015-10-09
lua: rename DNS field names

thus:
Rename lua DNS field names to conform to RFC 1035.

In my opinion, all response data should be available through the "rdata" field. I think it's strange to use "addr" as a field name for this for other types (that are not IP addresses) than "A" and "AAAA".

This would break all existing DNS lua scripts, but I think it's better to change this while it's still in beta.

### DIFF
--- a/src/util-lua-dns.c
+++ b/src/util-lua-dns.c
@@ -165,7 +165,7 @@ static int DnsGetQueryTable(lua_State *luastate)
                     SCFree(c);
                     return LuaCallbackError(luastate, "invalid length");
                 }
-                lua_pushstring(luastate, "rrname");
+                lua_pushstring(luastate, "name");
                 LuaPushStringBuffer(luastate, (uint8_t *)c, input_len);
                 lua_settable(luastate, -3);
                 SCFree(c);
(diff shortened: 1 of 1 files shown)


inliniac:
cc: @regit @jasonish

src/util-lua-dns.c
```diff
@@ -165,7 +165,7 @@ static int DnsGetQueryTable(lua_State *luastate)
                     SCFree(c);
                     return LuaCallbackError(luastate, "invalid length");
                 }
-                lua_pushstring(luastate, "rrname");
+                lua_pushstring(luastate, "name");
```
regit: It seems the RFC is using qname for name in the query. So we could use this. As a side effect it will avoid ambiguity  with the 'name' in the response.

regit:
I'm really fine with the idea modulo the comment in the commit.

src/util-lua-dns.c
thus: Good catch, regit! We should definitely use qname for the 'name' in the query.

inliniac:
On a related note: we've been trying to stay close to the common information model where it fits. http://docs.splunk.com/Documentation/CIM/latest/User/Overview

jasonish:
Should eve field names be in sync with the field names exposed to Lua?

PR #1679, opened 2015-10-10
mpm ac-ks v11

inliniac:
Port of the mpm ac-tile by @ken-tilera to other platforms. Named it 'ac-ks' => Kenneth Steele.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/288
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/290


inliniac:
@pevma can you test this for perf? Seeing promising results in pcap tests, less so in replay tests.

pevma:
I have started a perf-test sequence on live traffic.

It seems there are some troubles in general with live rule reloads that i am also looking into - however much more vivid when employing this mpm algorithm. It will not execute any additional reload after the first (USR2) reload in  a short sequence.

pevma:
Observations after running for some time:
- live rule reload is a problem as previously explained
- pf-ring performance is steady/the same
- af-packet performance is improved

inliniac:
Replaced by #1697

PR #1682, opened 2015-10-13

thus:

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -143,11 +143,97 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
     switch (ssl_state->curr_connp->handshake_type) {
         case SSLV3_HS_CLIENT_HELLO:
             ssl_state->flags |= SSL_AL_FLAG_STATE_CLIENT_HELLO;
+
+            /* skip version */
+            initial_input += 2;
+
+            /* skip random */
+            initial_input += 32;
+
+            /* skip session id */
+            uint8_t session_id_length = *(initial_input++);
+            initial_input += session_id_length;
+
+            if ((initial_input - input) > input_len)
+                goto end;
+
+            /* skip cipher suites */
+            uint16_t cipher_suites_length = *(uint16_t *)initial_input;
+            initial_input += ntohs(cipher_suites_length) + 2;
+
+            if ((initial_input - input) > input_len)
+                goto end;
+
+            /* skip compression methods */
+            uint8_t compression_methods_length = *(initial_input++);
+            initial_input += compression_methods_length;
+
+            if ((initial_input - input) > input_len)
+                goto end;
+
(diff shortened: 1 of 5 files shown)


src/app-layer-ssl.c
```diff
@@ -143,11 +143,97 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
     switch (ssl_state->curr_connp->handshake_type) {
         case SSLV3_HS_CLIENT_HELLO:
             ssl_state->flags |= SSL_AL_FLAG_STATE_CLIENT_HELLO;
+
+            /* skip version */
+            initial_input += 2;
```
inliniac: the moving forward of initial_input is opposite of what I'd expect. I'd expect input to move forward, and initial_input to be a reference to the original input, so that we can validate how far we are.

thus: Woops, I actually didn't catch that it said "initial". I'll change it the other way around.

src/app-layer-ssl.c
```diff
@@ -143,11 +143,97 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+                if ((initial_input - input) > input_len)
+                    goto end;
+
+                uint16_t ext_len = *(uint16_t *)initial_input;
+                initial_input += 2;
+
+                if ((initial_input + ntohs(ext_len) - input) > input_len)
+                    goto end;
+
+                switch (ntohs(ext_type)) {
```
inliniac: Think I'd prefer doing the ntohs when assigning the value to ext_type a few lines up.

src/app-layer-ssl.c
```diff
@@ -143,11 +143,97 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+                    case SSL_EXTENSION_SNI:
+                    {
+                        /* skip sni_list_length and sni_type */
+                        initial_input += 3;
+
+                        uint16_t sni_len = *(uint16_t *)initial_input;
+                        initial_input += 2;
+
+                        if ((initial_input + ntohs(sni_len) - input) > input_len)
```
inliniac: Same here for ntohs, shall we put that at the initial assignment?

src/app-layer-ssl.c
```diff
@@ -143,11 +143,97 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+                goto end;
+
+            uint16_t extensions_len = *(uint16_t *)initial_input;
+            initial_input += 2;
+
+            if ((initial_input - input) > input_len)
+                goto end;
+
+            /* return if no extensions are found */
+            if (ntohs(extensions_len) <= 0)
```
inliniac: minor: unsigned to < in <= doesn't make sense.

thus: That's true. After taking a closer look I see that the whole if check is probably redundant, since it won't enter the loop anyway if it's zero.

PR #1683, opened 2015-10-13
base64_decode, base64_data: decode and match base64 - v1

jasonish:
I realize I left some extra debug in there, but I'd like to use the PR to continue the discussion about the placement of base64_data execution for stateless and non-stateless rules.

This appears to do the job, but I'm still looking into potential side effects of the placement, particularly when there may be multiple sticky keywords in place.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/147
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/148

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -71,6 +71,8 @@ detect-ack.c detect-ack.h \
 detect-app-layer-event.c detect-app-layer-event.h \
 detect-app-layer-protocol.c detect-app-layer-protocol.h \
 detect-asn1.c detect-asn1.h \
+detect-base64-data.c detect-base64-data.h \
+detect-base64-decode.c detect-base64-decode.h \
 detect-byte-extract.c detect-byte-extract.h \
 detect-bytejump.c detect-bytejump.h \
 detect-bytetest.c detect-bytetest.h \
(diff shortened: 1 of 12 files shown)


inliniac:
I'm wondering about how the base64 buffer inspection works. Right now it's separate from the DetectEngineContentInspection run that created the base64 buffer. So that run completes, then we check if we have a base64 list and buffer, and inspect that.

What may be cleaner, is to have it inspected in-place. So right after the buffer is created, invoke DetectEngineContentInspection with the base64 list on the newly created base64 buffer. Then we don't have to worry about when & where to invoke the base64 list anymore, it's instant and in-place.

Make sense?

jasonish:
> What may be cleaner, is to have it inspected in-place. So right after the buffer is created, invoke DetectEngineContentInspection with the base64 list on the newly created base64 buffer. Then we don't have to worry about when & where to invoke the base64 list anymore, it's instant and in-place.
> 
> Make sense?

Thats too simple! Initial test and it works well as this PR, but much simpler.

PR #1684, opened 2015-10-14

thus:

Fixed changes suggested in #1682.

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -143,11 +143,93 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
     switch (ssl_state->curr_connp->handshake_type) {
         case SSLV3_HS_CLIENT_HELLO:
             ssl_state->flags |= SSL_AL_FLAG_STATE_CLIENT_HELLO;
+
+            /* skip version */
+            input += 2;
+
+            /* skip random */
+            input += 32;
+
+            /* skip session id */
+            uint8_t session_id_length = *(input++);
+            input += session_id_length;
+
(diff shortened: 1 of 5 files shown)


inliniac:
@jasonish can you review the parsing logic. Need a 2nd opinion :)

thus:
I'll add a detection module for this as mentioned in Feature 1134 (redmine).

src/app-layer-ssl.c
```diff
@@ -143,11 +143,93 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+
+            /* skip version */
+            input += 2;
+
+            /* skip random */
+            input += 32;
+
+            /* skip session id */
+            uint8_t session_id_length = *(input++);
+            input += session_id_length;
```
jasonish: Its possible to get here and dereference input with much less than 34 bytes. Check the bounds of input as you progress, and perhaps create a #define for random length, and version length.

src/app-layer-ssl.c
```diff
@@ -143,11 +143,93 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+
+            if ((input - initial_input) > input_len)
+                goto end;
+
+            /* skip cipher suites */
+            uint16_t cipher_suites_length = ntohs(*(uint16_t *)input);
+            input += cipher_suites_length + 2;
+
```
jasonish: Similarly here, the bounds are checked at line 157 to make sure there is enough input for the data already processed.  But we should check to make sure there is enough input data left before dereferencing at 161.

thus:
Thanks for the comments, Jason :)

PR #1689, opened 2015-10-14
Issue 1282: base64_decode and base64_data - v4

jasonish:
Snort compatible base64_decode and base64_data keywords.

util-base64 was updated to return what it has decoded when it hits an invalid character. This was required to match Snort's behaviour of base64_decode.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/150
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/151


inliniac:
One thing about the current approach that we should be careful about is that the recursive calls might modify 'state' in the det_ctx. I'm wondering if we might mess up relative matches because of the base64 list DetectEngineContentInspection call happening in the middle of another run. To what extend would this be a problem? To what extend is it possible that we'd continue on the original list after we've jumped into the base64 list? Things like this.

jasonish:
See https://github.com/inliniac/suricata/pull/1690

PR #1690, opened 2015-10-16
Issue 1282 - base64_decode and base64_data - v5

jasonish:
Snort compatible base64_decode and base64_data keywords.

util-base64 was updated to return what it has decoded when it hits an invalid character. This was required to match Snort's behaviour of base64_decode.

Updates from PR #1689 
- Make the base64_data keyword terminal. Any change of list after base64_data will cause the rule to fail.
- Checked against all VRT and ET Pro Snort rules.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/151
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/152


src/detect-base64-decode.c
```diff
@@ -0,0 +1,723 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#undef DEBUG_BASE64
```
inliniac: What is this? Why is it needed?

src/detect-base64-decode.c
```diff
@@ -0,0 +1,723 @@
...
+    sigmatch_table[DETECT_BASE64_DECODE].flags |= SIGMATCH_PAYLOAD;
+    sigmatch_table[DETECT_BASE64_DECODE].flags |= SIGMATCH_OPTIONAL_OPT;
+
+    decode_re = pcre_compile(decode_pattern, 0, &pcre_errptr, &pcre_erroffset,
+        NULL);
+    if (decode_re == NULL) {
+        SCLogError(SC_ERR_PCRE_COMPILE, "Failed to compile pattern \"%s\" at"
+            " offset %d: %s", decode_pattern, pcre_erroffset, pcre_errptr);
+        exit(EXIT_FAILURE);
+    }
```
inliniac: any reason for leaving out the study part that is used elsewhere?

src/detect-base64-decode.c
```diff
@@ -0,0 +1,723 @@
...
+    if (relative_str != NULL) {
+        pcre_free_substring(relative_str);
+    }
+    return retval;
+}
+
+static int DetectBase64DecodeSetup(DetectEngineCtx *de_ctx, Signature *s,
+    char *str)
+{
+    unsigned long bytes = 0;
```
inliniac: please use more explicit types like uint32_t and uint64_t

src/detect-base64-decode.h
```diff
@@ -0,0 +1,33 @@
...
+ * 02110-1301, USA.
+ */
+
+#ifndef __DETECT_BASE64_DECODE_H__
+#define __DETECT_BASE64_DECODE_H__
+
+#include "app-layer-template.h"
+
+typedef struct DetectBase64Decode_ {
+    unsigned long bytes;
```
inliniac: use explicit types

src/detect-engine-content-inspection.c
```diff
@@ -551,6 +553,15 @@ int DetectEngineContentInspection(DetectEngineCtx *de_ctx, DetectEngineThreadCtx
         SCLogDebug("lua match");
         goto match;
 #endif /* HAVE_LUA */
+    } else if (sm->type == DETECT_BASE64_DECODE) {
+        if (DetectBase64DecodeDoMatch(det_ctx, s, sm, buffer, buffer_len)) {
+            if (s->sm_arrays[DETECT_SM_LIST_BASE64_DATA] != NULL) {
+                KEYWORD_PROFILING_END(det_ctx, sm->type, 1);
+                if (DetectBase64DataDoMatch(de_ctx, det_ctx, s, f)) {
+                    goto match;
```
inliniac: The rule setup code should have made sure there is no SigMatch to check after this, right? Wonder if we should return directly, or jump to a new 'final_match' or similar.

Alternatively, a comment about this plus a DEBUG_VALIDATE_BUG_ON(sm->next != NULL) could also make it clear to anyone it's final.

src/detect-parse.c
```diff
@@ -855,14 +856,28 @@ int SigParse(DetectEngineCtx *de_ctx, Signature *s, char *sigstr, uint8_t addrs_
          * output variable. */
         do {
             memset(output, 0x00, buffer_size);
+            plist = s->list;
             ret = SigParseOptions(de_ctx, s, input, output, buffer_size);
             if (ret == 1) {
+                /* Check for a transition out of a terminal state. */
```
inliniac: I think I would prefer having this check go elsewhere, like in SigValidate. It runs post-parsing, but it keeps the core parsing function cleaner.

src/detect.h
```diff
@@ -687,6 +689,9 @@ typedef struct DetectEngineCtx_ {
      *  we can't lookup by proto, address, port as we don't have these */
     struct SigGroupHead_ *decoder_event_sgh;
 
+    /* Maximum size of the buffer for decoded base64 data. */
+    unsigned long base64_decode_max_len;
```
inliniac: explicit type

src/util-base64.c
```diff
@@ -103,7 +103,6 @@ uint32_t DecodeBase64(uint8_t *dest, const uint8_t *src, uint32_t len) {
             /* Invalid character found, so decoding fails */
             if (src[i] != '=') {
                 valid = 0;
-                numDecoded = 0;
```
inliniac: What effects (if any) does this have on other users of DecodeBase64? Seems SMTP decoder uses it.

inliniac:
Needs some more work.

src/detect-base64-decode.c
jasonish: Thought it was a bit better than some if 0's or uncommenting some extra debug info.  The debug info is only useful in dev, so I'll if 0 for now, and perhaps remove it for the next pr.

src/detect-base64-decode.c
jasonish: It couldn't optimize the query - it returns NULL, with an err or NULL.  I've added it back for consistency.

src/detect-base64-decode.c
jasonish: Fixed.

src/detect-base64-decode.h
jasonish: Fixed.

src/detect.h
jasonish: Fixed.

src/util-base64.c
jasonish: Yeah, it might have an effect as the mime decoder is using this to detect an anomaly in the base64 data, whereas base64_decode sometimes will just want to decode as much base64 as it can until a bad byte.

Toughts on adding a new flag, strict to determine if it should fail, or return what its decoded so far?

src/detect-parse.c
jasonish: But does SigValidate have the context to know if something like file_data came before or after base64_data?

src/util-base64.c
inliniac: I wonder if it would make sense to support the partial decoding there as well. Any thoughts @regit ?

src/detect-parse.c
inliniac: Not easily no

inliniac: maybe the SigMatch::idx field can help there

src/detect-engine-content-inspection.c
jasonish: Something like: https://github.com/jasonish/suricata/commit/a2c651c931aba15e3b0d038571153518e451d9ae

src/detect-parse.c
jasonish: Yes, it did. Thanks. Validation moved to SigValidate and test added.

jasonish:
Ready for a new PR, but will await further discussion on modification to util-base64.c.

PR #1692, opened 2015-10-19
lua: rename DNS field names

thus:
Rename lua DNS field names to conform to RFC 1035.

Changed query name to "qname" as commented on in #1678. Also changed the query type to "qtype", since this is also mentioned in the RFC.


inliniac:
Should have merged this before 3.0, but that didn't happen. So now we need to reconsider this. Scripts that work for 3.0 should continue to work with 3.1. So I guess we need a fallback of some type.

inliniac:
Closing as we'll need an approach that includes backwards compatibility.

PR #1695, opened 2015-10-20
New PktAcqBreakLoop callback in TmModule

cardigliano:
This patch adds a new callback PktAcqBreakLoop() in TmModule to let packet acquisition modules define "break-loop" functions to terminate the capture loop. This is useful in case of blocking functions that need special actions to take place in order to stop the execution.


inliniac:
Can you do a rebase to my master to get rid of the merge commit?

@regit any other thoughts?

regit:
@inliniac I don't think this is too intrusive so I'm ok with the change modulo a fix on my comment.

PR #1698, opened 2015-10-20
[preview] mixed-mode v2

glongo:

The idea is to replace single run_mode with a structure named RunModesList,
which represents the runmodes that will be run in an array.
Currently it's possibile to run two runmodes of the following:
AF_PACKET, Nfqueue, Nflog.

It means that if you try to run, for example, PF_RING and AF_PACKET, it won't work.

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/71
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/70

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -351,6 +351,8 @@ typedef struct PktProfiling_ {
 /* forward declartion since Packet struct definition requires this */
 struct PacketQueue_;
 
+enum mode {IDS, IPS, AUTO};
+
 /* sizes of the members:
  * src: 17 bytes
  * dst: 17 bytes
@@ -400,6 +402,8 @@ typedef struct Packet_
     struct Flow_ *flow;
 
     struct timeval ts;
+
+    enum mode pkt_mode;
 
     union {
         /* nfq stuff */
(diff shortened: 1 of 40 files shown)


inliniac:
Please add a link to the PR it replaces and also mention what changed since that PR.

src/decode.h
```diff
@@ -351,6 +351,8 @@ typedef struct PktProfiling_ {
 /* forward declartion since Packet struct definition requires this */
 struct PacketQueue_;
 
+enum mode {IDS, IPS, AUTO};
```
inliniac: This is not addressed: https://github.com/inliniac/suricata/pull/1660#discussion_r41411237

Please make sure all comments are addressed.

src/util-device.h
```diff
@@ -32,14 +33,14 @@ typedef struct LiveDevice_ {
 } LiveDevice;
 
 
-int LiveRegisterDevice(char *dev);
-int LiveGetDeviceCount(void);
-char *LiveGetDeviceName(int number);
-LiveDevice *LiveGetDevice(char *dev);
-int LiveBuildDeviceList(char * base);
+int LiveRegisterDevice(char *dev, int runmode);
```
inliniac: Instead of 'int runmode' I would like this to take the runmodes enum. Thats currently anonymous, so that will have to change as well.

inliniac:
I'll have a look at the next one that addresses *all* my comments.

src/util-device.h
glongo: fixed

src/decode.h
glongo: fixed

PR #1700, opened 2015-10-20

thus:


inliniac:
Wouldn't it be nicer to use a human readable time format in the rules and outputs?

thus:
Yes, it would be nicer. Any suggestions on time format? :)

inliniac:
That's a good question. Don't think we use time in the rules or yaml anywhere yet. We do use a time format in the json output both for the log timestamp and the flow logging (flow start/end). Perhaps use that?

Might be a good idea to send a rfc mail to oisf-devel.

thus:
What about implementing something like GNU getdate, but let it get patterns from a string array instead of reading patterns from file (defined by an environment variable, blah..)? That would enable us to easily parse several different time formats, instead of just one.

It would be a good idea to standardize on one format, though, to avoid that everyone use different time formats for their rules.

thus:
Created ticket for pull request (Feature 1583).

thus:
What I'm thinking of, is to add a function that parses date formats based on a string array of patterns (something similar to what GNU C library function getdate does).

Then we could do something like this whenever we want to parse a date:
```c
struct tm time;
char *string = "2015-10-22 23:59";

const char *patterns[] = {
            "%s",           // epoch (Unix time)
            "%Y-%m-%d %T",  // e.g 2015-10-22 19:25:30
            "%Y-%m-%d %R",  // e.g 2015-10-22 19:25
            "%Y-%m-%d",     // e.g 2015-10-22
            "%D"            // e.g 10/22/2015
};

int r = SCStringPatternToTime(string, patterns, &time);
```
(These might not be the patterns we want for these keywords, they were just added as an example)

The reason why I don't think we should use "getdate" directly, but implement something similar, is because "getdate" needs to read the patterns from a file. It also doesn't look like it's present in FreeBSD and OpenBSD.

It might be easier to understand if I just implement it. Code usually speaks for itself :)

jasonish:
I like dates consistent with https://en.wikipedia.org/wiki/ISO_8601.

eve-log is mostly consistent with this format, but has microseconds added.

thus:
I agree, ISO 8601 dates are quite nice. However, ISO 8601 specifies a lot of possible date combinations, so we should probably define a subset to support. I think we at least should support the following formats for the rule keywords (tls_notbefore and tls_notafter):
* 2015-10-22
* 2015-10-22 23:59:59
* 2015-10-22 23:59
* 2015-10-22 23
* 2015-10-22T23:59:59
* 2015-10-22T23:59
* 2015-10-22T23

It's probably not worth adding formats for specifying UTC time (2015-10-22 23:59Z), since the validity field in TLS certificates doesn't allow time zones anyway.

I'm not sure if we should support the following (or we might since it's simple if we add something similar to GNU getdate):
* 2015-295  (day of year)
* 2015-W43 (week number 43)
* 2015-W43-4 (week number 43, day four)

I think we should force the separators ("-" and ":") to be able to have support for epoch (Unix time) in addition to ISO 8601.

I also think ISO 8601 works well for logging. I think we should go for the format "2015-10-22 23:59:59", and I think we should add seconds even when they are not specified. Then we would get the same date length for every log entry, which I feel would add readability to the log.

Any more thoughts? Are my suggestions making sense at all? :)

thus:
Opened new pull request (#1719) that replaces this one.

PR #1701, opened 2015-10-20
workers runmode: allow multiple input devices

cardigliano:
This fix allows multiple input devices when running in 'workers' mode.

### DIFF
--- a/src/util-runmodes.c
+++ b/src/util-runmodes.c
@@ -420,7 +420,7 @@ int RunModeSetLiveCaptureWorkers(ConfigIfaceParserFunc ConfigParser,
 
     for (ldev = 0; ldev < nlive; ldev++) {
         char *live_dev_c = NULL;
-        if (live_dev != NULL) {
+        if ((nlive <= 1) && (live_dev != NULL)) {
             aconf = ConfigParser(live_dev);
             live_dev_c = SCStrdup(live_dev);
             if (unlikely(live_dev_c == NULL)) {


inliniac:
What are you trying to do that needs this change? AFAICT many ppl successfully use multiple worker threads with multiple nics

cardigliano:
Probably I just didn't understand how to configure workers mode with multiple interfaces then. What I was not able to do is:
suricata --pfring-int=zc:ethX --pfring-int=zc:ethY -c /etc/suricata/suricata.yaml
All threads were trying to open zc:ethY. 
What is the correct configuration for this?

inliniac:
Normally this is done by just using --pfring and adding the interfaces to the list in the yaml. I do think it would make sense to support it all from the command line as well.

@regit any thoughts?

lucaderi:
@inliniac Victor the commit is one line, I think you should include it as it hardens the code without any side effect.

regit:
@inliniac this patch is making the code in sync with autofp one that is using the same logic. So this will lead the two multithread capture modes to the same level. So I think it is acceptable.

But looking at the code handling multiple devices, I'm not really proud of it. o it would also be great to fix  it later.

lucaderi:
To be honest 100% it is not clear to me why you do not have a network abstraction layer such as DAQ on snort, so that you could avoid all these issues and greatly simplify Suricata design.

inliniac:

@lucaderi we can discuss such ideas elsewhere.

cardigliano:
Thank you!

lucaderi:
@inliniac Barcelona looks like a good place for discussing.

PR #1702, opened 2015-10-21
Dev mixed mode v3

glongo:

The idea is to replace single run_mode with a structure named RunModesList,
which represents the runmodes that will be run in an array.
Currently it's possibile to run two runmodes of the following:
AF_PACKET, Nfqueue, Nflog.

It means that if you try to run, for example, PF_RING and AF_PACKET, it won't work.

Updates:
- declare runmode as enum and not int
- rename enum mode to enum PktMode

Last PR: https://github.com/inliniac/suricata/pull/1698

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/72
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/71

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -351,6 +351,14 @@ typedef struct PktProfiling_ {
 /* forward declartion since Packet struct definition requires this */
 struct PacketQueue_;
 
+/*
+   since a pkt could come from different runmodes, we need to set
+   a mode to be able to do an action later.
+   (e.g. if a pkt come from NFQ, pkt_mode will be set to IPS and
+         it means that we'll be able to drop it.)
+*/
+enum PktMode {IDS, IPS, AUTO};
+
 /* sizes of the members:
  * src: 17 bytes
  * dst: 17 bytes
@@ -400,6 +408,8 @@ typedef struct Packet_
     struct Flow_ *flow;
 
     struct timeval ts;
+
+    enum PktMode pkt_mode;
 
     union {
         /* nfq stuff */
@@ -874,6 +884,8 @@ int PacketCallocExtPkt(Packet *p, int datalen);
 int PacketCopyData(Packet *p, uint8_t *pktdata, int pktlen);
 int PacketSetData(Packet *p, uint8_t *pktdata, int pktlen);
 int PacketCopyDataOffset(Packet *p, int offset, uint8_t *data, int datalen);
+int PacketModeIsIPS(const Packet *p);
+int PacketModeIsIDS(const Packet *p);
 const char *PktSrcToString(enum PktSrcEnum pkt_src);
 
 DecodeThreadVars *DecodeThreadVarsAlloc(ThreadVars *);
--- a/src/stream-tcp-inline.c
+++ b/src/stream-tcp-inline.c
@@ -41,9 +41,14 @@ extern int stream_inline;
  *  \retval 0 no
  *  \retval 1 yes
  */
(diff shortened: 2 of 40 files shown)


src/stream-tcp-inline.c
```diff
@@ -41,9 +41,14 @@ extern int stream_inline;
  *  \retval 0 no
  *  \retval 1 yes
  */
-int StreamTcpInlineMode(void)
+int StreamTcpInlineMode(Packet *p)
```
inliniac: const Packet

src/stream-tcp.c
```diff
@@ -430,13 +430,10 @@ void StreamTcpInitConfig(char quiet)
...
         if (strcmp(temp_stream_inline_str, "auto") == 0) {
-            if (EngineModeIsIPS()) {
-                stream_inline = 1;
-            } else {
-                stream_inline = 0;
-            }
+            stream_inline = AUTO;
         } else if (ConfGetBool("stream.inline", &inl) == 1) {
-            stream_inline = inl;
+            if (inl == IPS)
```
inliniac: This doesn't look right. ConfGetBool will not set IPS to inl. Only by chance (IPS == 1). Also, if stream.inline is true, it should set IPS to stream_inline unconditionally.

src/suricata.c
```diff
@@ -262,17 +262,30 @@ void EngineModeSetIDS(void)
...
 }
 
-int RunmodeGetCurrent(void)
+int RunmodeGetCurrent(int index)
 {
-    return run_mode;
+    return runmodeslist.run_mode[index];
 }
 
+static int RunmodeIsAlreadyRan(RunModesList *runmodes, int run_mode)
```
inliniac: int run_mode should be enum RunModes? Also, what is the purpose of this func? Add comment to it plz.

src/suricata.h
```diff
@@ -126,8 +128,14 @@ PacketQueue trans_q[256];
 
 SCDQDataQueue data_queues[256];
 
+typedef struct RunModesList_ {
+    int run_mode[RUNMODES_MAX];
+    int runmodes_cnt;
+    int enable_mixed_mode;
```
inliniac: Why do we need this field? Wouldn't it be enough to test for runmodes_cnt >= 2?

src/decode.h
```diff
@@ -351,6 +351,14 @@ typedef struct PktProfiling_ {
 /* forward declartion since Packet struct definition requires this */
 struct PacketQueue_;
 
+/*
+   since a pkt could come from different runmodes, we need to set
+   a mode to be able to do an action later.
+   (e.g. if a pkt come from NFQ, pkt_mode will be set to IPS and
+         it means that we'll be able to drop it.)
+*/
+enum PktMode {IDS, IPS, AUTO};
```
inliniac: I think it's better to call the members something like PKT_MODE_IDS, PKT_MODE_IPS, PKT_MODE_AUTO

inliniac:
I think the overall approach makes sense, however I would like to hide some of the internals. Most this:
runmodeslist.run_mode[0] == ...
As we currently just support 2 modes, I think it's fine to use 2 functions like RunmodeGetPrimary(), RunmodeGetSecondary().

glongo:
Replaced by https://github.com/inliniac/suricata/pull/1726

PR #1706, opened 2015-10-21
pfring pkt acq: use zero-copy recv in workers runmode

cardigliano:
This patch removes packet copy when suricata is running in workers runmode,
packet copy is not needed in this case since packets are processed in sequence.


inliniac:

cardigliano:
Thank you!

PR #1707, opened 2015-10-21
pfring pkt acq: removed reentrant flag

cardigliano:
PF_RING_REENTRANT is not needed as each pfring socket is used by a single thread.


inliniac:

cardigliano:
Thank you!

PR #1708, opened 2015-10-21
pfring pkt acq: capture loop optimisation

cardigliano:
For each packet the capture module checks whether it is time to dump stats calling
TimeGet(). TimeGet() is an expensive function using gettimeofday() or SCSpinLock()
which affect performance. Since gettimeofday() is already called for setting packet
timestamp, it is more efficient to use the packet timestamp directly.


inliniac:
In the future, could you do changes like this and #1706 and #1707 in a single pull request? Different commits like you did now, but a single PR works better. Esp since #1706 and #1708 lead to a minor merge conflict.

inliniac:

cardigliano:
Thank you!

PR #1709, opened 2015-10-22

thus:

Fixed changes suggested by @jasonish in #1684.

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -119,6 +119,9 @@ SslConfig ssl_config;
 #define SSLV3_RECORD_HDR_LEN 5
 #define SSLV3_MESSAGE_HDR_LEN 4
 
+#define SSLV3_CLIENT_HELLO_VERSION_LEN 2
+#define SSLV3_CLIENT_HELLO_RANDOM_LEN 32
+
 /* TLS heartbeat protocol types */
 #define TLS_HB_REQUEST              1
 #define TLS_HB_RESPONSE             2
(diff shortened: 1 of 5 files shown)


src/app-layer-ssl.c
```diff
@@ -143,11 +146,97 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
             ssl_state->flags |= SSL_AL_FLAG_STATE_CLIENT_HELLO;
+
+            /* skip version */
+            input += SSLV3_CLIENT_HELLO_VERSION_LEN;
+
+            /* skip random */
+            input += SSLV3_CLIENT_HELLO_RANDOM_LEN;
+
+            /* skip session id */
+            uint8_t session_id_length = *(input++);
```
inliniac: Here we use the input that we just moved forward, _before_ it is checked 2 lines later. Seems more sane to do it the other way around?

src/app-layer-ssl.c
```diff
@@ -143,11 +146,97 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+
+            if ((input - initial_input) > input_len)
+                goto end;
+
+            input += session_id_length;
+
+            /* skip cipher suites */
+            uint16_t cipher_suites_length = ntohs(*(uint16_t *)input);
```
inliniac: same here

src/app-layer-ssl.c
```diff
@@ -143,11 +146,97 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+            input += 2;
+
+            if ((input - initial_input) > input_len)
+                goto end;
+
+            input += cipher_suites_length;
+
+            /* skip compression methods */
+            uint8_t compression_methods_length = *(input++);
```
inliniac: and here

inliniac:
Think the pointer logic is wrong. The pattern is now: move ptr fwd, use it, then check it. Should be move fwd, check it, then use it.

thus:
When you mention it, I see what you mean. The logic (or the opposite of logic) in my head was thinking that I were dealing with pointers, and that I had to check them before I used them, but I'm dealing with integers here.

Coffee first, then new pull request! :)
(should always be applied in this order)

PR #1711, opened 2015-10-22

thus:

Fixed changes suggested in #1709.

https://redmine.openinfosecfoundation.org/issues/1134

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -119,6 +119,9 @@ SslConfig ssl_config;
 #define SSLV3_RECORD_HDR_LEN 5
 #define SSLV3_MESSAGE_HDR_LEN 4
 
+#define SSLV3_CLIENT_HELLO_VERSION_LEN 2
+#define SSLV3_CLIENT_HELLO_RANDOM_LEN 32
+
 /* TLS heartbeat protocol types */
 #define TLS_HB_REQUEST              1
 #define TLS_HB_RESPONSE             2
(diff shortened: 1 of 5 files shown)


thus:
This PR already has a ticket (Feature 1134)

src/app-layer-ssl.c
```diff
@@ -143,11 +146,101 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+
+            if ((input - initial_input) > input_len)
+                goto end;
+
+            /* skip session id */
+            uint8_t session_id_length = *(input++);
+
+            input += session_id_length;
+
+            if ((input + 1 - initial_input) > input_len)
```
inliniac: Still not quite happy with the pattern here. Really the > (opposed to >= ) would guarantee at least 1 byte, then the +1 another. As we know we will need 2 bytes, this works. However I think the code readability would much improve from making the '2' explicit.

Thinking more, I think it might be good to use a (local) macro. And then instead of this line, use something like:
<pre>
if (!(HAS_SPACE(2)))
    goto end;
</pre>

thus: That would make the code look nicer :)

Do you mean a macro looking something like this?

```c
#define HAS_SPACE(n) (((input) + (n) - (initial_input)) >= (input_len)) ?  0 : 1
```

thus:
Replaced by #1729.

PR #1713, opened 2015-10-22
pfring pkt acq: keep running on 'pfring_set_cluster' failure when clu…

cardigliano:
…ster is not required

Suricata creates a pfring cluster with a default ID = 1 when not explicitly configured,
unless the device has prefix 'dna' or 'zc'. Since pf_ring also supports other cards
implementing kernel-bypass (cluster not supported), this is preventing those cards from
running on top of this module. This patch stops suricata on 'pfring_set_cluster' failure
only when error code != PF_RING_ERROR_NOT_SUPPORTED or cluster ID has not been explicitly
configured.

### DIFF
--- a/src/runmode-pfring.c
+++ b/src/runmode-pfring.c
@@ -109,6 +109,7 @@ void *OldParsePfringConfig(const char *iface)
     }
 
     strlcpy(pfconf->iface, iface, sizeof(pfconf->iface));
+    pfconf->flags = 0;
     pfconf->threads = 1;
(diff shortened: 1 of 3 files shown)


inliniac:
This PR shows nicely why the 'subject' of a commit message should be cut at 72 chars :) See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits

inliniac:
Merged, thanks @cardigliano!

PR #1715, opened 2015-10-22
rules-reload: fix reload with -s or -S

regit:
When using the -S or -s option, the reload was causing the specified
rules file to be forgotten and the default rules to be loaded at
reload time.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/115
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/113


inliniac:

PR #1719, opened 2015-10-26

thus:

Changes:


inliniac:
Massive QA fail. Unittests:
<pre>
==== TEST RESULTS ====
PASSED: 3841
FAILED: 20
======================
</pre>

thus:
Strange, when I run the unittests, I get:
```
==== TEST RESULTS ====
PASSED: 3878
FAILED: 0
======================
```
Which tests are failing on your end?

inliniac:
See https://s3.amazonaws.com/archive.travis-ci.org/jobs/87438588/log.txt

thus:
It's still strange that none of the tests fail on my test machine.

It looks like the tests failing, are the tests I created for testing the parsing of ISO 8601 time strings. They all fail, while the tests checking Unix time strings does not, even though they use the same function for parsing the date.

Any ideas of why this happen, or hints on where I should start looking to fix this? :)

thus:
I managed to reproduce the failed tests using a machine with a 3.6 kernel, so disregard my last question :)

thus:
\*facepalm\*

This error was because of my use of "mktime". I didn't realize that it use the time zone from the executing environment, instead of using "tm_gmtoff". Because of this, it worked on one machine, but failed on another machine with different time zone settings.

I'll fix this and send a new pull release :)

inliniac:
Ah, good catch. Time handling is definitely non-trivial...

thus:
This pull request is replaced by #1722.

PR #1721, opened 2015-10-26
Issue 1282: base64_decode, base64_data - v7

jasonish:
Fixes issues in last PR: https://github.com/inliniac/suricata/pull/1690

For util-base64, intoduce "strict" mode so other uses keep the same behaviour.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/154
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/155


src/detect-base64-decode.c
```diff
@@ -0,0 +1,733 @@
...
+        printf("Decoded data:\n");
+        PrintRawDataFp(stdout, det_ctx->base64_decoded,
+            det_ctx->base64_decoded_len);
+    }
+#endif
+
+    return det_ctx->base64_decoded_len > 0;
+}
+
+static int ParseUint32(const char *str, uint32_t *value)
```
inliniac: We can use ByteExtractUint32 instead?

src/detect-engine-content-inspection.c
```diff
@@ -551,6 +553,16 @@ int DetectEngineContentInspection(DetectEngineCtx *de_ctx, DetectEngineThreadCtx
         SCLogDebug("lua match");
         goto match;
 #endif /* HAVE_LUA */
+    } else if (sm->type == DETECT_BASE64_DECODE) {
+        if (DetectBase64DecodeDoMatch(det_ctx, s, sm, buffer, buffer_len)) {
+            if (s->sm_arrays[DETECT_SM_LIST_BASE64_DATA] != NULL) {
+                KEYWORD_PROFILING_END(det_ctx, sm->type, 1);
```
inliniac: Seems this is called elsewhere regardless of the result here. Do we need it?

src/detect-base64-decode.c
jasonish: > We can use ByteExtractUint32 instead?

Isn't that for extracting a uint32 from packed bytes? ParseUint32 is to parse an integer from text. A lot of other code uses atoi, but I'm not a fan of that function. Perhaps we need some generic integer parsing functions so we can do away with atoi?

src/detect-engine-content-inspection.c
jasonish: Yes, we need it before the call to DetectBase64DataDoMatch as that will reenter DetectEngineContentInspection with profiling on, causing Suricata to bail at line 110 in this file.

src/detect-base64-decode.c
inliniac: ByteExtractStringUint32 is for strings, so perhaps that? Would need to consider how errors are handled.

src/detect-engine-content-inspection.c
inliniac: ack

src/detect-base64-decode.c
jasonish: That works almost as a drop in here. prscript running now.

jasonish:
Closed. See #1746.

PR #1722, opened 2015-10-27

thus:

Changes:
* Fixed time zone handling


src/detect-tls-validity.c
```diff
@@ -0,0 +1,1008 @@
...
+        cert_epoch = connp->cert0_not_before;
+    else if (dd->type == DETECT_TLS_TYPE_NOTAFTER)
+        cert_epoch = connp->cert0_not_after;
+
+    if (cert_epoch == 0)
+        SCReturnInt(0);
+
+    if (dd->mode == DETECT_TLS_VALIDITY_EQ && dd->epoch == cert_epoch)
+        ret = 1;
+    else if (dd->mode == DETECT_TLS_VALIDITY_LT && dd->epoch > cert_epoch)
```
inliniac: _LT followed by > is a bit confusing. Also, it seems that by using 'keyword epoch > cert epoch', it really is 'cert epoch <= keyword epoch'. So LEQ instead of LT.

src/detect-tls-validity.c
```diff
@@ -0,0 +1,1008 @@
...
+        ret = 1;
+    else if (dd->mode == DETECT_TLS_VALIDITY_GT && dd->epoch < cert_epoch)
+        ret = 1;
+    else if (dd->mode == DETECT_TLS_VALIDITY_RA &&
+            dd->epoch < cert_epoch && dd->epoch2 > cert_epoch)
```
inliniac: I think <= and >= would make more sense here. See also https://redmine.openinfosecfoundation.org/issues/1416

src/detect-tls-validity.c
```diff
@@ -0,0 +1,1008 @@
...
+    int gmtoff;
+    char *tz;
+
+    gmtoff = tm.tm_gmtoff;
+
+    /* from timegm */
+    tz = getenv("TZ");
+    if (tz)
+        tz = strdup(tz);
+    setenv("TZ", "", 1);
```
inliniac: Is the env being modified here? That seems wrong.

src/detect-tls-validity.c
```diff
@@ -0,0 +1,1008 @@
...
+
+    s->flags |= SIG_FLAG_APPLAYER;
+    s->alproto = ALPROTO_TLS;
+
+    SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_AMATCH);
+
+    return 0;
+
+error:
+    return -1;
```
inliniac: sm and dd are lost here

src/detect-tls-validity.c
```diff
@@ -0,0 +1,1008 @@
...
+    UtRegisterTest("ValidityTestParse22", ValidityTestParse22, 1);
+    UtRegisterTest("ValidityTestParse23", ValidityTestParse23, 1);
+    UtRegisterTest("ValidityTestParse24", ValidityTestParse24, 1);
+#endif /* UNITTESTS */
+}
+
+/**
+ * \brief Register unit tests for tls_notafter.
+ */
+void TlsNotAfterRegisterTests(void)
```
inliniac: Seems we just register the same tests twice, that is not useful.

src/detect-tls-validity.c
thus: This is how this is proposed done in "man timegm". I agree that modifying environment variables is not preferable. I'll look for alternative ways to implement this.

src/util-decode-der-get.c
```diff
@@ -72,6 +73,67 @@ static const char *Oid2ShortStr(const char *oid)
     return "unknown";
 }
 
+static time_t GentimeToTime(char *gentime, size_t length)
```
inliniac: Seems all callers pass valid c strings, and pass the result of strlen() as length. That means we can ditch the argument and call strlen in the function itself

src/util-decode-der-get.c
```diff
@@ -72,6 +73,67 @@ static const char *Oid2ShortStr(const char *oid)
...
+     * therefore be on the format YYmmddHHMMSSZ. */
+    if (length != 13)
+        goto error;
+
+    /* UTCTime use two digits to represent the year. The year field (YY)
+     * should be interpreted as 19YY when it is greater than or equal to
+     * 50. If it is less than 50 it should be interpreted as 20YY.
+     * Because of this, GeneralizedTime must be used for dates in the
+     * year 2050 or later. */
+    strlcpy(yy, utctime, 2);
```
inliniac: As a general pattern: I prefer using sizeof(yy) as that will autmatically work if the size of yy is changed.

src/util-time.c
```diff
@@ -302,3 +316,106 @@ void CreateTimeString (const struct timeval *ts, char *str, size_t size)
...
+    oldlen = inlen = strlen(string);
+
+    /* Skip trailing whitespace */
+    while (inlen > 0 && isspace(string[inlen - 1]))
+        inlen--;
+
+    char *instr = NULL;
+
+    if (inlen < oldlen) {
+        instr = malloc(inlen + 1);
```
inliniac: SCMalloc

src/util-time.c
```diff
@@ -302,3 +316,106 @@ void CreateTimeString (const struct timeval *ts, char *str, size_t size)
...
+        tp->tm_gmtoff = 0;
+        tp->tm_zone = NULL;
+        result = strptime(string, patterns[i], tp);
+
+        if (result && *result == '\0')
+            break;
+    }
+
+    if (instr != NULL)
+        free(instr);
```
inliniac: SCFree

src/util-time.c
```diff
@@ -302,3 +316,106 @@ void CreateTimeString (const struct timeval *ts, char *str, size_t size)
...
+        tp->tm_mday = 1;
+
+    /* January 1 is assumed, if only year is given */
+    if (tp->tm_year != INT_MIN && tp->tm_mon == INT_MIN &&
+            tp->tm_mday == INT_MIN) {
+        tp->tm_mon = 0;
+        tp->tm_mday = 1;
+    }
+
+    /* Today is assumed, if no date is given */
```
inliniac: In what case might this happen? Using the system time might have side effects when we're processing pcaps and we're considering the 'pcap time' to be our time.

src/detect-tls-validity.c
thus: Note to self.

Replacing line 195 - 216 with:
```c
return mktime(&tm) + tm.tm_gmtoff;
```
looks like it might do the same without touching any environment variables...

src/util-time.c
thus: In our case, we can probably just assume that a date must be specified. Should probably return 1 (error) instead of setting date to today if a date is not specified.

src/detect-tls-validity.c
thus: I agree that it makes more sense for these keywords to be inclusive, instead of just between.

thus:
Replaced by #1728.

PR #1726, opened 2015-10-28
Dev mixed mode v4

glongo:
This addressed the comments in last PR: https://github.com/inliniac/suricata/pull/1702
Updates:
- Changed name of enum PktMode members
- Hidden RunModeList internals
- Add utility function RunmodeIsUnknown
- Fixed enabling inline setting
- Removed enable_mixed_mode member from RunModesList
 
Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/78
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/77

### DIFF
--- a/src/alert-fastlog.c
+++ b/src/alert-fastlog.c
@@ -147,7 +147,7 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
         }
 
         char *action = "";
-        if ((pa->action & ACTION_DROP) && EngineModeIsIPS()) {
+        if ((pa->action & ACTION_DROP) && PacketModeIsIPS(p)) {
             action = "[Drop] ";
         } else if (pa->action & ACTION_DROP) {
             action = "[wDrop] ";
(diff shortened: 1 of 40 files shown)


inliniac:
"It's only a preview, so it will need more work." -> explain what.
"This addressed the comments in last PR: #1702" -> what did you change? Provide a list.

glongo:
I think we couldn't stop to consider it as preview since you think the approach is good,
and the code is working.
Of course there will be some minor comment to address, but in general should be good.

inliniac:
"Hidden RunModeList internals" -> seems this is only done for suricata.c, not globally. I think it should be done globally.

glongo:
I agree.

inliniac:
Are you seeing this work normally wrt reloads and shutdown? Haven't tried yet, but would like to hear about your experiences.

PR #1728, opened 2015-10-30

thus:

Changes:
* Fixed time zone handling

Fixed changes suggested in #1722.


src/util-decode-der-get.c
```diff
@@ -128,6 +191,57 @@ const Asn1Generic * Asn1DerGet(const Asn1Generic *top, const uint8_t *seq_index,
     return node;
 }
 
+int Asn1DerGetValidity(const Asn1Generic *cert, time_t *not_before,
+                       time_t *not_after, uint32_t *errcode)
+{
+    const Asn1Generic *node, *it;
+    int rc = -1;
+
+    if (errcode)
```
inliniac: Seems like a no-op since it's directly followed by setting it to 0

src/util-time.c
```diff
@@ -138,6 +138,20 @@ void CreateIsoTimeString (const struct timeval *ts, char *str, size_t size)
     }
 }
 
+void TimeToIsoTimeString (time_t time, char *str, size_t size)
```
inliniac: Is the result here the same as (or at least compatible to) the output by CreateIsoTimeString?

src/util-decode-der-get.c
thus: You are right. 

In the other ASN1 functions "errcode" is set to 0 right before returning when no error is encountered. It probably makes most sense to just remove it, since "errcode" is not used any way when there is no error.

src/util-time.c
thus: The output is the same, but without microseconds. It probably makes more sense to convert from time_t to timeval and use CreateIsoTimeString instead. What do you think?

inliniac: That makes sense to me. Keeps things simpler.

thus: Then I will do that. Thanks :)

thus:
Proposed changes fixed in #1849.

PR #1729, opened 2015-10-30

thus:

Fixed changes suggested in #1711.

https://redmine.openinfosecfoundation.org/issues/1134


inliniac:
Merged, thanks a lot Mats!

PR #1732, opened 2015-11-02
lua: expose TLS certificate chain to lua

thus:
Expose the entire TLS certificate chain to lua through TlsGetCertChain().

Makes it possible to use lua scripts to play with raw TLS certificates and certificate chains.

Example:
```lua
function init (args)
    local needs = {}
    needs["protocol"] = "tls"
    return needs
end

function setup (args)
    x509 = require('openssl.x509')
    filename = SCLogPath() .. "/" .. "lua_tls_cert.log"
end

function log (args)
    chain = TlsGetCertChain()

    if chain == nil then
        return
    end

    for n, t in pairs(chain) do
        cert = x509.new(t["data"], "DER")
        file:write(cert:text() .. "\n")
    end

    file:flush()
end

function deinit (args)
    file:close(file)
end
```

The simple example above uses the luaossl library (https://github.com/wahern/luaossl) to log all certificates to file as text.

https://redmine.openinfosecfoundation.org/issues/1535

### DIFF
--- a/src/util-lua-tls.c
+++ b/src/util-lua-tls.c
@@ -133,12 +133,76 @@ static int TlsGetCertInfo(lua_State *luastate)
     return r;
 }
 
+static int GetCertChain(lua_State *luastate, const Flow *f, int direction)
+{
+    void *state = FlowGetAppState(f);
+    if (state == NULL)
+        return LuaCallbackError(luastate, "error: no app layer state");
+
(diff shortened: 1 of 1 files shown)


inliniac:
Please do a rebase and resubmit. Thanks!

thus:
Replaced by #1761.

PR #1733, opened 2015-11-03
handle MTU discovery in multi iface case

regit:
This patch adds handling of multiple interface by using as
default-packet-size the maximum MTU + header length among the
live iface.

This fixes the issue reported in https://redmine.openinfosecfoundation.org/issues/1556

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/116
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/114


inliniac:

PR #1735, opened 2015-11-05
Threads rename v8 - Bug #1524

zrasmor:
Addressed feedback from #1648:
 * Enforced standard for offline pcap
 * Added interface name to single mode
 * Replace C with RX
 * Created global thread names
 * Added "#" in front of thread numbers

### DIFF
--- a/src/flow-manager.c
+++ b/src/flow-manager.c
@@ -118,8 +118,11 @@ void FlowDisableFlowManagerThread(void)
     /* flow manager thread(s) is/are a part of mgmt threads */
     tv = tv_root[TVT_MGMT];
 
-    while (tv != NULL) {
-        if (strncasecmp(tv->name, "FM", 2) == 0) {
+    while (tv != NULL) 
+    {
+        if (strncasecmp(tv->name, thread_name_flow_mgr,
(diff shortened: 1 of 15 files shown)


inliniac:
This doesn't pass my QA yet. The reason is that tv->name is now freed, however there are still quite a few places where it's set to a static string: counters.c does this, detect-engine-loader.c & unix-manager.c as well. This will lead to crashes when trying to free those.

Side note: I think the commit count is getting rather high. Perhaps we can squash them a bit. E.g. all the "Update thread names for $name based on new naming standard" into one, etc.

inliniac:
I guess it means that we need short names for the counter threads, unix socket and detect engine loaders as well.

counters: CS CW
unix socket: US
detect loader: DL#xx

Thoughts?

zrasmor:
Those names are fine with me. 

Another alternative, if you're not sold on shortening those names right now, is to change 'name' from a char* to a char name[16]. This would slightly change some of the logic, but then we don't have to worry about allocating or freeing memory. I'm fine with either approach.

And I will squash some of these commits in the next PR.

inliniac:
I like the name[16] suggestion. Wrt the names, I think it's good to switch over completely in one step.

PR #1736, opened 2015-11-06
lua: functions to force magic and md5 lookup on files

thus:
Added lua functions "SCFileForceMagic" and "SCFileForceMd5" to force magic and md5 lookup on all files. These functions are equivalent to setting "force-magic: yes" and "force-md5: yes" in suricata.yaml.

https://redmine.openinfosecfoundation.org/issues/1590

Example:
```lua
function init (args)
    local needs = {}
    needs['type'] = 'file'
    return needs
end

function setup (args)
    filename = SCLogPath() .. "/" .. "lua_file.log"
    SCFileForceMagic()
    SCFileForceMd5()
end

function log (args)
    fileid, txid, name, size, magic, md5 = SCFileInfo()
    file:write("filename='" .. name .. "' magic='" .. magic ..
               "' md5='" .. md5 .. "'\n")
    file:flush()
end

function deinit (args)
    file:close()
end
```

### DIFF
--- a/src/util-lua-common.c
+++ b/src/util-lua-common.c
@@ -666,6 +666,17 @@ static int LuaCallbackFileState(lua_State *luastate)
 }
 
 /** \internal
+ *  \brief wrapper to force magic lookup on all files from lua script
+    \retval cnt number of items placed on the stack
+ */
+static int LuaCallbackFileForceMagic(lua_State *luastate)
+{
+    FileForceMagicEnable();
+    lua_pushinteger(luastate, 1);
+    return 1;
+}
+
+/** \internal
  *  \brief fill lua stack with thread info
  *  \param luastate the lua state
  *  \param pa pointer to packet alert struct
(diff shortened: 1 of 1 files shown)


src/util-lua-common.c
```diff
@@ -666,6 +666,28 @@ static int LuaCallbackFileState(lua_State *luastate)
 }
 
 /** \internal
+ *  \brief wrapper to force magic lookup on all files from lua script
+    \retval cnt number of items placed on the stack
+ */
+static int LuaCallbackFileForceMagic(lua_State *luastate)
+{
+    FileForceMagicEnable();
```
inliniac: This is really not meant to be run at packet runtime, as it modifies a global variable that doesn't use locking or atomics. It *should* be fairly harmless though.

src/util-lua-common.c
```diff
@@ -666,6 +666,28 @@ static int LuaCallbackFileState(lua_State *luastate)
...
+    lua_pushinteger(luastate, 1);
```
inliniac: Why is this push done?

PR #1737, opened 2015-11-08
Misc improvements

inliniac:
Some misc things:
- when using -i, use afpacket if available
- afpacket output cleanups
- afpacket: no more warnings for missing cluster values, use sane defaults
- stats.log: option to suppress 0 values, enabled by default

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/303
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/305

Btw, https://redmine.openinfosecfoundation.org/issues/1591 affects this, -i on unsupported linktypes now fails. Need to check if afpacket supports a link first maybe.

cc: @regit @pevma @jasonish

### DIFF
--- a/src/log-stats.c
+++ b/src/log-stats.c
@@ -51,6 +51,7 @@
 
 #define LOG_STATS_TOTALS  (1<<0)
 #define LOG_STATS_THREADS (1<<1)
+#define LOG_STATS_NULLS   (1<<2)
 
 TmEcode LogStatsLogThreadInit(ThreadVars *, void *, void **);
 TmEcode LogStatsLogThreadDeinit(ThreadVars *, void *);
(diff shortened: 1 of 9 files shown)


regit:
All seems really sane for me but the point already mentioned. We should check if af_packet supports the link type. In current state we may loose support for some link type and support for card using a pcap modified version.

inliniac:
On 17-11-15 14:14, Eric Leblond wrote:
> All seems really sane for me but the point already mentioned. We should
> check if af_packet supports the link type. In current state we may loose

See the linked ticket. It seems the issue is linktype detection, not
support for the linktype. At least in the case I ran into.

> support for some link type and support for card using a pcap modified
> version.

Yeah, wonder if we should somehow do a 'try open' and fallback to
straight pcap on failure.

inliniac:
Partly merged through #1784, will resubmit the rest later.

PR #1740, opened 2015-11-17
Config fix

regit:
Two patches related to configuration. First one is cleaning dump-config and the next one is using some built at configure path instead of using hardcoded one.

PR scripts:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/117
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/115


inliniac:

PR #1741, opened 2015-11-17
Fix the comment and explanation for random-chunk-size

norg:


inliniac:
Merged through #1845, thanks!

PR #1745, opened 2015-11-18
app-layer-smtp: support for multiline response

maurizioabba:
Multiline response support is provided but not enforced. This patch
allow parsing multiline response when a reply is processed


inliniac:
Do you have a test case to show what the effect is?

inliniac:
Thanks for the test case. Merged this through #1765

PR #1749, opened 2015-11-20
conf: fail if sequence node set from command line - v1

jasonish:
One possible solution - better than a segv anyways:

It is currently not possible to set a sequence node from the
command line with --set. If done, the node will be setup as a map
with sequence nodes added causing issues at run time. For now
its best to just bail during initialization when this is seen.

### DIFF
--- a/src/conf-yaml-loader.c
+++ b/src/conf-yaml-loader.c
@@ -308,6 +308,12 @@ ConfYamlParse(yaml_parser_t *parser, ConfNode *parent, int inseq)
             SCLogDebug("event.type=YAML_SEQUENCE_START_EVENT; state=%d", state);
             if (ConfYamlParse(parser, node, 1) != 0)
                 goto fail;
+            if (node->final && !node->is_seq) {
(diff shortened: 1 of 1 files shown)


jasonish:
Just a thought... Applying --set values after loading, and overriding the already set values.  We could then determine if it was a sequence.  I think it could still lead to confusing outcomes though, if the configuration file changed by startup scripts didn't and so on.

inliniac:
Is this still needed now that #1816 is merged?

jasonish:
Yes, they solve different issues. #1816 just protects a NULL access introduced when you --set an option in potentially the wrong place, it doesn't actually help support lists.

This PR prevents conflicts between setting a list item on the command line, and the item not being a list in the configuration.

But maybe we should close this for now. I'm working on something that would supersede this and provide an overlay config instead of these "final" parameters. The idea being that setting list items on the command line would lead to predictable behaviour.

inliniac:
Ok, sounds good. Closing.

PR #1750, opened 2015-11-21
dns: response delta logging

inliniac:
Log the response time in microseconds as 'response_time' in JSON.

UDP only for now.

Generally I wonder if this should be done in suri (as there is a cost in memory use). Maybe it belongs in postprocessing, e.g. logstash. @regit @jasonish

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/308
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/310

### DIFF
--- a/src/app-layer-dns-common.h
+++ b/src/app-layer-dns-common.h
@@ -172,6 +172,9 @@ typedef struct DNSTransaction_ {
 
     TAILQ_ENTRY(DNSTransaction_) next;
     DetectEngineState *de_state;
+
+    struct timeval request_ts;
+    struct timeval response_ts;
 } DNSTransaction;
 
 /** \brief Per flow DNS state container */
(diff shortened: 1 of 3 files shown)


jasonish:
> Generally I wonder if this should be done in suri (as there is a cost in memory use). Maybe it belongs in postprocessing, e.g. logstash.

Is the cost that big in terms of the overall transaction structures?  For something like DNP3 its not too much, but that would have a lot less memory in use than something like DNS.

Something else I've seen is that the requests and responses are logged with the same timestamp (the timestamp of the ack completing the tx) making it impossible to do in post-processing.

inliniac:
Cost was 32bytes per transaction, but brought that down to 12 in PR #1751

inliniac:
Replaced by #1751

PR #1751, opened 2015-11-21
dns: response delta logging

inliniac:
Log the response time in microseconds as 'response_time' in JSON.

As #1750, but with more efficient storage. Instead of 32 bytes (2x struct timeval), this uses 12bytes per transaction.

Still only UDP.

PRscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/309
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/311


inliniac:
Replaced by #1764

PR #1766, opened 2015-11-26
socket-permission: chmod for socket permission under linux

maurizioabba:
linux uses chmod (with the name of the socket as parameter) in order to
set the permissions of a socket.


inliniac:
QA failure https://s3.amazonaws.com/archive.travis-ci.org/jobs/93359377/log.txt

PR #1767, opened 2015-11-26
filestore-call: forcing a call to FileStore instead of manual update

maurizioabba:
Avoid to touch directly the flag in order to have a single point where we actually
touch the File structure


inliniac:
Merged through #1845, thanks!

PR #1768, opened 2015-11-26
pcap-timestamp: set time of pseudopacket to the last one of the flow

maurizioabba:
The pseudopacket get created with the current time. This is suboptimal,
as the packet is not a real one and the time it reports should not
depend on a configuration parameter. Instead, assign to the packet the
timestamp of the last actual packet seen.
Perform calculation with abs in order to comply with pcaps in different
timezone


inliniac:
QA failure: https://s3.amazonaws.com/archive.travis-ci.org/jobs/93359654/log.txt

PR #1770, opened 2015-11-26
pcap-timestamp: set time of pseudopacket to the last one of the flow

maurizioabba:
The pseudopacket get currently created with the current time. This is suboptimal, as the packet is not a real one and the time it reports should not depend on a configuration parameter (specifically, the flow timeout). Instead, assign to the packet the timestamp of the last actual packet seen. Perform calculation with abs in order to comply with pcaps in different timezone.

NOTE: created v2 as the first version had a compilation error. Sorry about that.

### DIFF
--- a/src/detect-engine-threshold.c
+++ b/src/detect-engine-threshold.c
@@ -150,7 +150,7 @@ int ThresholdTimeoutCheck(Host *host, struct timeval *tv)
 
     prev = NULL;
     while (tmp != NULL) {
-        if ((tv->tv_sec - tmp->tv_sec1) <= tmp->seconds) {
+        if (abs(tv->tv_sec - tmp->tv_sec1) <= tmp->seconds) {
             prev = tmp;
             tmp = tmp->next;
             retval = 0;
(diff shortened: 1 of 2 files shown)


src/detect-engine-threshold.c
```diff
@@ -150,7 +150,7 @@ int ThresholdTimeoutCheck(Host *host, struct timeval *tv)
 
     prev = NULL;
     while (tmp != NULL) {
-        if ((tv->tv_sec - tmp->tv_sec1) <= tmp->seconds) {
+        if (abs(tv->tv_sec - tmp->tv_sec1) <= tmp->seconds) {
```
inliniac: AFAICS this threshold change has nothing to do with the flow timestamp change. Thus it should not be part of the same commit.

inliniac: Ah misread things, scratch that comment.

inliniac: It does make me a bit worried about the effect on time handling in general.

axelrodGunnarson: what do you mean "effect on time handling in general"? Is there anything you would like to be modified in a different way?

PR #1771, opened 2015-11-26
socket-permission: chmod for socket permission under linux

maurizioabba:
linux uses chmod (with the name of the socket as parameter) in order to set the permissions of a socket.

Branch created after first branch (fix-socket-permissions) failed to build, sorry about that


inliniac:
@regit can you have a look?

regit:
Hello, first of all the commit message is wrongly formatted. You need one short description then a more complete explanation.

PR #1776, opened 2015-11-30
Midstream fix v2

regit:
A work from @glongo and me, aiming at improving midstream. When working with modbus traffic, suricata is not able to picked up a large part of sessions when midstream is enable. This is due to the fact, it does not manage to pick all sessions starting with data in to_client side. This is a real problem for modbus protocol where session last forever and you can not really restart an industrial system when you restart suricata.

This patchset fixes this and it has shown some results on both real (test environment) and simulated (manually altered pcap) modbus traffic.

Regarding the main patch, I tend to think we should maybe completely invert the src and dest in the flow when the application layer detected it is reverted. This may lead to bigger, intrusive changes so I've decided to propose this code and discuss if now as a later solution.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/123
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/121

### DIFF
--- a/src/flow.h
+++ b/src/flow.h
@@ -338,6 +338,8 @@ typedef struct Flow_
 
     uint32_t flags;
 
+    uint8_t reverted;
+
     /* time stamp of last update (last packet). Set/updated under the
      * flow and flow hash row locks, safe to read under either the
      * flow lock or flow hash row lock. */
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -345,7 +345,8 @@ static AppProto AppLayerProtoDetectPPGetProto(Flow *f,
             SCLogDebug("toserver - Probing parser found for source port %"PRIu16, f->sp);
 
             /* found based on source port, so use sp registration */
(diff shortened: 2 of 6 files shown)


inliniac:
I think this needs a better approach. Probably closer to what StreamTcpPacketSwitchDir does, and more integrated with that logic. I don't like the extra logic that only applies to the app-layer, but does amend the flow structure.

src/flow.h
```diff
@@ -338,6 +338,8 @@ typedef struct Flow_
 
     uint32_t flags;
 
+    uint8_t reverted;
```
inliniac: adding a new (undocumented!) var for setting a flag seems rather excessive. We have f->flags for this.

regit: Yes, forgot to tag RFC this patchset.

regit:
OK, good. I'm gonna look at StreamTcpPacketSwitchDir  and mimic that.

PR #1777, opened 2015-11-30
Modbus update

regit:
A work on Modbus protocol. It mainly adds logging of Modbus transactions.

I would appreciate feedback from @daviddiallo on the choice I've made in this patchset. Mostly on latest patch where protocol/implementation logic has to be checked.

EVE events display most information available:
```JSON
{
  "timestamp": "1970-01-01T02:14:38.827724+0100",
  "flow_id": 49631440,
  "pcap_cnt": 84139,
  "event_type": "modbus",
  "src_ip": "13.17.10.7",
  "src_port": 1121,
  "dest_ip": "13.17.10.14",
  "dest_port": 502,
  "proto": "TCP",
  "modbus": {
    "transaction_id": 15079,
    "length": 4,
    "function": "readcoils",
    "read": {
      "address": 23,
      "quantity": 3,
      "value": {
        "0": 0,
        "1": 0,
        "2": 0
      }
    },
    "category": "assigned",
    "replied": 1
  }
}
```
In the case of a multiple read like this one, it is possible to output values as it is shown here or as a simple array of value. The dictionary approach is making integration with ELK direct so it is the default. 

This is a PR on top of #1776 as there is a light imbrication of code.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/121
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/119


daviddiallo:
Nice Work. Thanfs for it.

Regarding to EVE events display, I think it would be better to get information like following (if it possible):
```
{
  "timestamp": "1970-01-01T02:14:38.827724+0100",
  "flow_id": 49631440,
  "pcap_cnt": 84139,
  "event_type": "modbus",
  "src_ip": "13.17.10.7",
  "src_port": 1121,
  "dest_ip": "13.17.10.14",
  "dest_port": 502,
  "proto": "TCP",
  "modbus": {
    "transaction_id": 15079,
    "length": 4,
    "function": "readcoils",
    "read": {
      "address": 23,
      "quantity": 3,
      "value": {
(address + 0) ->        "23": 0,
(address + 1) ->        "24": 0,
(address + 2) ->        "25": 0
      }
    },
    "category": "assigned",
    "replied": 1
  }
}
```

Otherwise, if this JSON output works on regular file, it does not work with redis output.

daviddiallo:
ModbusParserTest08 is failed because of commit "app-layer-modbus: add missing exception" (145cb62).

According to MODBUS Application Protocol Specification V1.1b3, all "Exception code" are not possible for each "Function code", for example, only "Exception code" 0x1, 0x2, 0x3 and 0x4 are possible for "Read Coils" (Function code 0x1), or "Exception code" 0x1 and 0x4 are available for "Read Exception status" (Function code 0x7).

ModbusExceptionResponse function (and ModbusParserTest08) checks "Exception code" is available according to "Function code" thanks to a switch case and "Fallthrough" to default case.

In reverting C file of the commit, ModbusParserTest08 will pass.

I think I should add a comment to explain it in source code.

regit:
OK, I have missed the function + exception check. But I think reverting is not the answer, we should instead add the new exceptions in the correct place as I have seen them on a simulated network.

daviddiallo:
In that case, in adding the new exceptions in correct place, I propose:
             /* Fallthrough */
+        case MODBUS_ERROR_CODE_ACKNOWLEDGE:
+        case MODBUS_ERROR_CODE_SERVER_DEVICE_BUSY:
+        case MODBUS_ERROR_CODE_GATEWAY_PATH_UNAVAILABLE:
+        case MODBUS_ERROR_CODE_GATEWAY_FAIL_TO_RESPOND:
+            if (tx->category != MODBUS_CAT_PUBLIC_ASSIGNED)
+                break;
           default:

regit:
@daviddiallo it is hard to read bullet point code but I think it should do the job.

PR #1780, opened 2015-12-03
output-json: add app_proto key in root

regit:
By adding the key in the root of *flow and fileinfo  events it
will be possible to get all events for one application layer by
using a 'event_type:proto OR app_proto:proto' filter. This will
permit to the analyst to get a good view of events related to
one protocol.

This patch also fixes a regression in file logging where app_proto
was available before 94dbd303e4744a40f3761265be7c73a7a4754764 create
the regression.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/125
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/123

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/1614

### DIFF
--- a/src/output-json-file.c
+++ b/src/output-json-file.c
@@ -108,6 +108,8 @@ static void FileWriteJsonRecord(JsonFileLogThread *aft, const Packet *p, const F
             break;
     }
 
+    json_object_set_new(js, "app_proto",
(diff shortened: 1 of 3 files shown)


inliniac:
Merged through #1784, thanks Eric.

PR #1781, opened 2015-12-03
modbus: disable by default - v1

jasonish:
Disable modbus by default, including the case where no modbus configuration is provided.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1615

Possible issue: Without this PR, Modbus configures itself with sane defaults (assuming you want modbus enabled) if the configuration is not provided.  Now there is no way to say "yes, enable modbus with a default sane configuration" - the user must enable it and make sure the detection-ports is correct.  This is probably not specific to Modbus, but to any app-layer protocol we may want to ship by default.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/160
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/161

### DIFF
--- a/src/app-layer-modbus.c
+++ b/src/app-layer-modbus.c
@@ -1427,21 +1427,13 @@ void RegisterModbusParsers(void)
                                           STREAM_TOSERVER,
                                           ModbusProbingParser);
         } else {
-            /* if we have no config, we enable the default port 502 */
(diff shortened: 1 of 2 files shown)


inliniac:
Merged through #1784, thanks Jason.

PR #1782, opened 2015-12-06
Threads rename v9 - Bug #1524

zrasmor:
Update for #1735 

 * Change ThreadVars name from char * name to char name[16] to eliminate need for memory allocation
 * Update counters, erf-file, detect engine, unix socket to use global thread names
 * clean up commit history


inliniac:
QA failure https://travis-ci.org/inliniac/suricata/jobs/95126194

PR #1783, opened 2015-12-06
Threads rename v10 - Bug #1524

zrasmor:
Update for #1782 

Add missed commit updating unit test to account for char * to char[16] shift.

As a reminder, here's what was added in the last PR:

 * Change ThreadVars name from char * name to char name[16] to eliminate need for memory allocation
 * Update counters, erf-file, detect engine, unix socket to use global thread names
 * clean up commit history


inliniac:
Replaced by #1891

PR #1792, opened 2015-12-14
Unix manager fix rc2

regit:
This is a set of fixes for unix manager. First one fixes a race condition causing a segfault occuring at suricata start on some low profile and loaded systems. Third one causes suricata to exit when incompatible settings are used. Effect of using -l and unix socket runmode at the same time was the running mode not to be working at all without any warning to the user.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/128
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/120

Redmine ticket for 1st patch: https://redmine.openinfosecfoundation.org/issues/1645

### DIFF
--- a/src/unix-manager.c
+++ b/src/unix-manager.c
@@ -851,6 +851,37 @@ static TmEcode UnixManagerThreadInit(ThreadVars *t, void *initdata, void **data)
     if (utd == NULL)
         return TM_ECODE_FAILED;
 
+    if (UnixNew(&command) == 0) {
+        int failure_fatal = 0;
+        SCLogError(SC_ERR_INITIALIZATION,
+                   "Unable to create unix command socket");
+        if (ConfGetBool("engine.init-failure-fatal", &failure_fatal) != 1) {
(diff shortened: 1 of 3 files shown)


src/unix-manager.c
```diff
@@ -905,6 +882,28 @@ static TmEcode UnixManager(ThreadVars *th_v, void *thread_data)
...
+}
+
+static TmEcode UnixManager(ThreadVars *th_v, void *thread_data)
+{
+    int ret;
+
+    /* set the thread name */
+    SCLogDebug("%s started...", th_v->name);
+
+    StatsSetupPrivate(th_v);
```
inliniac: these really shouldn't be here, as UnixManager is called by TmThreadsManagement which calls this and SCDropCaps too. I see it's wrong in the unpatched code as well btw, just caught my eye now.

regit: Do you want an updated patchset with that included ?

inliniac:
Merged through #1799, thanks Eric.

PR #1794, opened 2015-12-17
logging: per direction tx logging - v1

jasonish:
Add per-direction directional transaction logging. A logging module can request per-directional logging by setting the directional flag when registering. A value of 0 (non-directional) should result in the same behaviour as now. A value of 1 will cause the callback to be called on completion of each direction, so the callback will need to be direction aware to avoid logging 2 times.

I have verified an eve log file containing DNS and HTTP requests before and after this patch and the output was identical.

Directional logging has been tested in a DNP3 branch.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/168
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/170

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -133,10 +133,9 @@ struct AppLayerParserState_ {
     /* Indicates the current transaction that is being inspected.
      * We have a var per direction. */
     uint64_t inspect_id[2];
-    /* Indicates the current transaction being logged.  Unlike inspect_id,
-     * we don't need a var per direction since we don't log a transaction
-     * unless we have the entire transaction. */
-    uint64_t log_id;
+    /* Indicates the current transaction being logged.  
+     * We have a var per direction. */
+    uint64_t log_id[2];
 
     /* Used to store decoder events. */
     AppLayerDecoderEvents *decoder_events;
@@ -519,19 +518,27 @@ void AppLayerParserDestroyProtocolParserLocalStorage(uint8_t ipproto, AppProto a
     SCReturn;
 }
 
-uint64_t AppLayerParserGetTransactionLogId(AppLayerParserState *pstate)
+uint64_t AppLayerParserGetTransactionLogId(AppLayerParserState *pstate,
+    uint8_t direction)
 {
     SCEnter();
+    uint64_t retval = 0;
 
-    SCReturnCT((pstate == NULL) ? 0 : pstate->log_id, "uint64_t");
+    if (pstate != NULL) {
+        retval = pstate->log_id[direction & STREAM_TOSERVER ? 0 : 1];
+    }
+
+    SCReturnCT(retval, "uint64_t");
 }
 
-void AppLayerParserSetTransactionLogId(AppLayerParserState *pstate)
+void AppLayerParserSetTransactionLogId(AppLayerParserState *pstate,
+    uint8_t direction)
 {
     SCEnter();
 
-    if (pstate != NULL)
-        pstate->log_id++;
+    if (pstate != NULL) {
+        pstate->log_id[direction & STREAM_TOSERVER ? 0 : 1]++;
+    }
 
     SCReturn;
 }
@@ -633,7 +640,7 @@ FileContainer *AppLayerParserGetFiles(uint8_t ipproto, AppProto alproto,
 uint64_t AppLayerTransactionGetActiveDetectLog(Flow *f, uint8_t flags)
 {
     AppLayerParserProtoCtx *p = &alp_ctx.ctxs[FlowGetProtoMapping(f->proto)][f->alproto];
-    uint64_t log_id = f->alparser->log_id;
+    uint64_t log_id = f->alparser->log_id[flags & STREAM_TOSERVER ? 0 : 1];
     uint64_t inspect_id = f->alparser->inspect_id[flags & STREAM_TOSERVER ? 0 : 1];
     if (p->logger == TRUE) {
         return (log_id < inspect_id) ? log_id : inspect_id;
@@ -655,7 +662,7 @@ uint64_t AppLayerTransactionGetActiveLogOnly(Flow *f, uint8_t flags)
     AppLayerParserProtoCtx *p = &alp_ctx.ctxs[f->protomap][f->alproto];
 
     if (p->logger == TRUE) {
-        uint64_t log_id = f->alparser->log_id;
+        uint64_t log_id = f->alparser->log_id[flags & STREAM_TOSERVER ? 0 : 1];
         SCLogDebug("returning %"PRIu64, log_id);
         return log_id;
     }
@@ -802,7 +809,7 @@ uint64_t AppLayerParserGetTransactionActive(uint8_t ipproto, AppProto alproto,
 
     uint64_t active_id;
 
-    uint64_t log_id = pstate->log_id;
+    uint64_t log_id = pstate->log_id[direction & STREAM_TOSERVER ? 0 : 1];
     uint64_t inspect_id = pstate->inspect_id[direction & STREAM_TOSERVER ? 0 : 1];
     if (alp_ctx.ctxs[FlowGetProtoMapping(ipproto)][alproto].logger == TRUE) {
         active_id = (log_id < inspect_id) ? log_id : inspect_id;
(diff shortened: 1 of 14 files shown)


src/app-layer-parser.c
```diff
@@ -655,7 +662,7 @@ uint64_t AppLayerTransactionGetActiveLogOnly(Flow *f, uint8_t flags)
     AppLayerParserProtoCtx *p = &alp_ctx.ctxs[f->protomap][f->alproto];
 
     if (p->logger == TRUE) {
```
inliniac: I think it's important that we now also track whether we have loggers per direction. If we have a logger for one direction, is the log_id for the other dir also updated somehow? Otherwise this func will always return tx id 0

src/output-tx.c
```diff
@@ -138,35 +148,87 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
...
+                tc_ready = 1;
             }
         }
 
+        if (!(ts_ready || tc_ready)) {
+            SCLogNotice("progress not for enough, not logging");
+            break;
+        }
+
+        /* Prevent the processing of one direction getting farther
```
inliniac: Whats the reasoning here? In for example HTTP pipelining we can have many requests before we see a response. It would make sense to log them in the order that we get them.

src/app-layer-parser.c
jasonish: The logger is for both directions, you can't control that. If you want it for one direction you have to ignore a certain direction in your log callback.  So nothing has really changed here.  On transaction complete, both log IDs will be incremented and we'll move forward.

src/output-tx.c
jasonish: All this does is wait for one transaction to be complete before moving on to the other.  I thought that was current behaviour.  For example, transaction can go like this:

1. TX 0 Request
2. TX 1 Request
3. TX 1 Response
4. TX 0 Response

Now that we check per direction, without enforcing order, the response for tx 0 will never get logged as the transaction ID tracker is incremented after step 3, so we never check the state of of TX 0 again.

I believe this was implicit before, now that we check per direction we have to enforce it explicitly to avoid the transaction counters being incremented to soon.

Regarding HTTP - is that something that happens now? Or is that something we could do with per direction logging.  I think that will also require a different way of tracking complete transactions like my overlapping TX case.

jasonish: I guess pipelining is different than the case I present, as everything should be order, so it could be handled by the current tracking.  I think this would be pretty easy to extend to that case when we have a logger that would make use of it - like logging HTTP requests prior to the reply.

src/app-layer-parser.c
inliniac: On 17-12-15 16:36, Jason Ish wrote:
>> @@ -655,7 +662,7 @@ uint64_t AppLayerTransactionGetActiveLogOnly(Flow *f, uint8_t flags)
>>      AppLayerParserProtoCtx *p = &alp_ctx.ctxs[f->protomap][f->alproto];
>>  
>>      if (p->logger == TRUE) {
> 
> The logger is for both directions, you can't control that. If you want
> it for one direction you have to ignore a certain direction in your log
> callback. So nothing has really changed here. On transaction complete,
> both log IDs will be incremented and we'll move forward.

The problem is that both log_id's will now be used. Are both updated in
all possible configurations?

src/output-tx.c
inliniac: On 17-12-15 16:42, Jason Ish wrote:
> In src/output-tx.c
> <https://github.com/inliniac/suricata/pull/1794#discussion_r47920316>:
> 
>>              }
>>          }
>>  
>> +        if (!(ts_ready || tc_ready)) {
>> +            SCLogNotice("progress not for enough, not logging");
>> +            break;
>> +        }
>> +
>> +        /* Prevent the processing of one direction getting farther
> 
> All this does is wait for one transaction to be complete before moving
> on to the other. I thought that was current behaviour. For example,
> transaction can go like this:
> 
>  1. TX 0 Request
>  2. TX 1 Request
>  3. TX 1 Response
>  4. TX 0 Response
> 
> Now that we check per direction, without enforcing order, the response
> for tx 0 will never get logged as the transaction ID tracker is
> incremented after step 3, so we never check the state of of TX 0 again.
> 
> I believe this was implicit before, now that we check per direction we
> have to enforce it explicitly to avoid the transaction counters being
> incremented to soon.
> 
> Regarding HTTP - is that something that happens now? Or is that
> something we could do with per direction logging. I think that will also
> require a different way of tracking complete transactions like my
> overlapping TX case.

In HTTP the transactions are implicit, so the first response is mapped
to the first request. So we can't have a case there the response for a
TX 1 comes in before TX 0.

inliniac: On 17-12-15 16:53, Jason Ish wrote:
> In src/output-tx.c
> <https://github.com/inliniac/suricata/pull/1794#discussion_r47921644>:
> 
>>              }
>>          }
>>  
>> +        if (!(ts_ready || tc_ready)) {
>> +            SCLogNotice("progress not for enough, not logging");
>> +            break;
>> +        }
>> +
>> +        /* Prevent the processing of one direction getting farther
> 
> I guess pipelining is different than the case I present, as everything
> should be order, so it could be handled by the current tracking. I think
> this would be pretty easy to extend to that case when we have a logger
> that would make use of it - like logging HTTP requests prior to the reply.

If we have:

REQ 1
REQ 2
REQ 3
RESP 1
RESP 2
RESP 3

Could we in the current patch log all the requests before the responses
are received?

src/app-layer-parser.c
jasonish: Currently yes.  They will be updated as each direction reaches completion, whether or not anything was logged (as thats up to the logging callback, but the transition is based on readyness). Then we move forward to the next one. So you are getting the same behaviour as before which waits for both to become ready before anything is logged.

So it should behave exactly the same as now.  Unless I'm missing some special case on the current setup?

src/output-tx.c
jasonish: No, this patch is concerned with the more narrow scope of just logging the request before the response if available.  This would preserve what I believe is the current behaviour or not logging until the response is received, correct?

inliniac: On 17-12-15 17:49, Jason Ish wrote:
> No, this patch is concerned with the more narrow scope of just logging
> the request before the response if available. This would preserve what I
> believe is the current behaviour or not logging until the response is
> received, correct?

The current behaviour is to log on 'tx complete', which means both
request and response are done.

If we move to a model where one callback is 'request complete' and the
other 'tx complete' (implying 'response complete') then I think the
request logging really should happen as soon as the request is complete.

Other transactions shouldn't affect this.

Perhaps the implied 'tx complete == response complete' doesn't always
make sense, and we need a separate 'response complete'. Not sure what
protocol would have a response before the request though :)

src/app-layer-parser.c
inliniac: On 17-12-15 17:46, Jason Ish wrote:
> Currently yes. They will be updated as each direction reaches
> completion, whether or not anything was logged (as thats up to the
> logging callback, but the transition is based on readyness). Then we
> move forward to the next one. So you are getting the same behaviour as
> before which waits for both to become ready before anything is logged.
> 
> So it should behave exactly the same as now. Unless I'm missing some
> special case on the current setup?

The id tracking is complex is master already, but gets really hard to
follow now. It's not very clear to me when/how the 2 log id's are
updated now. I would like to see a cleaner implementation of this, esp
in the main logging loop.

src/output-tx.c
jasonish: I think we're both after the same thing, but expanding the scope of what I'm trying to attempt in this fix.

For logging we have request complete, and response complete. So tx complete is request_complete && response_complete.  It would be nice to log the requests as they are seen, prior to the responses, but we don't do that now, and I need to retain that behaviour until we can fix the scenario of overlapping transactions, which is another task.  But this gets us half way there by within the scope of transaction, logging each part on receipt if requested to do so.

src/app-layer-parser.c
jasonish: Ok, I'll see what I can do. I don't think it can get any simpler though - only more complex if we have multiple outstanding transactions partially logged.

But can we agree on the scope here.  Being that the problem we have right now is we want to log the request and response as seen, rather than on TX completion, but within the scope of a transaction. And not the issue of logging multiple requests before the responses are seen, which I don't think we have any loggers attempting to do right now - which is going to require even more invasive changes in the case of an overlapping TX.

jasonish:
See #1807.

PR #1795, opened 2015-12-18
RFC Only - Per direction TX logging with "pipelining" support.

jasonish:
I took a look into what it would take to log sequential requests before responses were seen and retain support for overlapping transactions. Some extra book-keeping is required for the counters. For instance, we don't want to increment a counter until the earliest transaction is finished. Overlapping transactions messed this up as we'd never iterate over the lower transactions again.

So to make this happen we delay incrementing the log_id counters, but need to also set a flag on a transaction to flag if its been logged or not. This will allow us to handle something like:

REQ 1
REQ 2
RSP 2
REQ 1

Or (as can be in async messaging protocols):

REQ 1
RSP 2
REQ 2
RSP 1

Or even,

REQ 1
REQ 2
REQ 3
RSP 1
RSP 2
RSP 3

(http pipelining for example).

This is due for some refactoring, but I was after thoughts on the extra bookkeeping required to make this happen.  I didn't see another way than creating another set of callbacks into the protocol specific app-layer to retrieve and set the extra per tx data.

It would probably be a bit cleaner if we didn't have to worry about directional vs non-directional loggers, and let the logger choose to ignore the direction. For example, the current DNS and HTTP loggers could ignore on direction TO_SERVER and just handle the TO_CLIENT case.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/169
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/171

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -68,6 +68,8 @@
 #include "util-unittest-helper.h"
 #include "util-validate.h"
 
+#include "output-tx.h"
+
 #include "runmodes.h"
 
 static GetActiveTxIdFunc AppLayerGetActiveTxIdFuncPtr = NULL;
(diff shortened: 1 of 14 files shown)


inliniac:
Couple of thoughts:
1. I think we can simplify the OutputTx API logic to calls that simply have something like "IsLogged" and "SetLogged", so that we can leave it to the actual implementation to see how it stores this per tx. Most likely it would be something like 2 bit flags in an already existing flags field.

2. I think we risk confusing the relation between the log_id's and the per tx flags. I guess we wouldn't need the per direction log_id anymore in case the flags are used. Then we can simply move the single log_id field forward if both direction flags are set.

e.g.
if (...IsLogged(tx, TOSERVER) && ...IsLogged(tx, TOCLIENT)) { ... increment log_id ... }

jasonish:
See #1807.

PR #1796, opened 2015-12-18
Decode errors stats

maurizioabba:
We create one counter per each specific decode error detected by suricata. Whenever a decode errors is detected by the decode engine, we update not only the invalid counter but also the specific decode_error_type counter in order to have a more precise view over faulty packets in the network.
The pull is composed of two commits, the first one is related to counters, allowing the copy of the counter name to the internal StatsCounter structure (so that we can create counters on the fly and we do not have to statically hardcode them) and freeing it while release the StatsCounter resource. The second one is the actual implementation of the decode error stats.


src/decode.c
```diff
@@ -419,7 +425,14 @@ void DecodeRegisterPerfCounters(DecodeThreadVars *dtv, ThreadVars *tv)
         StatsRegisterCounter("defrag.ipv6.timeouts", tv);
     dtv->counter_defrag_max_hit =
         StatsRegisterCounter("defrag.max_frag_hits", tv);
-
+    
+    int i = 0;
+    for (i = 0; i < DECODE_EVENT_PACKET_MAX; i++) {
+        char counter_name[256];
+        snprintf(counter_name, 256, "decoder.%s", DEvents[i].event_name);
```
inliniac: I don't like how this makes dynamically allocated counter names necessary again (had just removed that recently).

I'm thinking that we should sync the decoder and stream names. That would mean we prepend 'decoder.' to each decoder name string so we can use those directly in the counter.

It would require an update to DetectDecodeEventSetup, similar to how DetectStreamEventSetup handles it.

inliniac:
Any thoughts about (optionally) also adding the counters for the stream events? Probably also more useful than using the rules...

inliniac:
Closing because of QA fail. See https://s3.amazonaws.com/archive.travis-ci.org/jobs/97652252/log.txt

For improving the work, please see https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Incorporating-feedback, esp make changes into new branch, new pr. Thanks!

src/decode.c
axelrodGunnarson: Yes sure I will update it and do a new pull request

PR #1798, opened 2015-12-18
util-device: fix iface stats output

glongo:
This permits to output stats per interface only if it exists.

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/91
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/90

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1648

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -1589,6 +1589,7 @@ static int AFPCreateSocket(AFPThreadVars *ptv, char *devname, int verbose)
 
     /* Init is ok */
     AFPSwitchState(ptv, AFP_STATE_UP);
+    LiveDeviceIfaceWasSeen(ptv->livedev);
     return 0;
 
 frame_err:
--- a/src/source-erf-dag.c
+++ b/src/source-erf-dag.c
@@ -309,6 +309,7 @@ ReceiveErfDagThreadInit(ThreadVars *tv, void *initdata, void **data)
 
     ewtn->tv = tv;
     *data = (void *)ewtn;
+    LiveDeviceIfaceWasSeen(ptv->livedev);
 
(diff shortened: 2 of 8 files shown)


inliniac:
Why not just check for SC_ATOMIC_GET(pd->pkts) > 0?

glongo:
Because we may have an interface with no traffic, in that case the notice message is wrong

inliniac:
Then the message could just include that there is no traffic?

glongo:
It depends how you want to handle that.
My idea was to make sure that an interface was seen, because in suri 2.1 was possible to use iface which doesn't exist, look at these two log lines (bit old):

    [10792] 1/12/2015 -- 10:54:15 - (util-device.c:191) <Notice> (LiveDeviceListClean) -- Stats for 'eth0':  pkts: 669, drop: 0 (0.00%), invalid chksum: 0
    [10792] 1/12/2015 -- 10:54:15 - (util-device.c:191) <Notice> (LiveDeviceListClean) -- Stats for 'eth1':  pkts: 0, drop: 0 (-nan%), invalid chksum: 0

In that case, eth1 iface doesn't exist, but we can see stats.
If we check that SC_ATOMIC_GET(pd->pkts) > 0, we're not sure if an iface exists or not.

@regit, @pevma any thoughts?

regit:
I remember some reports stating suricata was not capturing because people did not get the iface was not usable. Idea of this patch is to have an explicit message in this case.

pevma:
Similar to the ideas/feedback expressed here  https://redmine.openinfosecfoundation.org/issues/1595

inliniac:
I'm still not convinced this is the way to go.

PR #1803, opened 2015-12-23
unix manager: log client's version with debug level

gozzy:
As (dis)connects are already logged as a debug events, this one
should do the same.


inliniac:
Merged through #1821, thanks Alexander!

PR #1806, opened 2015-12-31
json-stats: add per app-layer stats

glongo:
This adds adds the number of application layer transactions handled
by the supported protocols.

In some application layer, like ssh, we don't have a transaction,
so we count the number of the states.

The output added in EVE will be:
"app-layer": { "http": 234, "stmp": 12, "ssh": 22, "dns": 2000 }

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/92
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/91


inliniac:
I would prefer this to be implemented as regular counters, not using global atomics and then global counters.

PR #1807, opened 2016-01-04
Per direction logging  - v3

jasonish:
Another cut on per-direction logging based on feedback on the previous PRs: #1794, #1795.

- correct request timestamps
- correct ordering

Prscript output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/175
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/177

### DIFF
--- a/src/app-layer-dns-common.h
+++ b/src/app-layer-dns-common.h
@@ -172,6 +172,8 @@ typedef struct DNSTransaction_ {
 
     TAILQ_ENTRY(DNSTransaction_) next;
     DetectEngineState *de_state;
+    uint8_t request_logged;
+    uint8_t response_logged;
 } DNSTransaction;
 
 /** \brief Per flow DNS state container */
@@ -235,6 +237,9 @@ void *DNSStateAlloc(void);
 void DNSStateFree(void *s);
 AppLayerDecoderEvents *DNSGetEvents(void *state, uint64_t id);
 int DNSHasEvents(void *state);
+
+int DNSGetTxIsLogged(void *vtx, uint8_t direction);
+void DNSSetTxIsLogged(void *vtx, uint8_t direction);
 
 int DNSValidateRequestHeader(DNSState *, const DNSHeader *dns_header);
 int DNSValidateResponseHeader(DNSState *, const DNSHeader *dns_header);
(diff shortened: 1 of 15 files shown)


src/app-layer-dns-common.h
```diff
@@ -172,6 +172,8 @@ typedef struct DNSTransaction_ {
 
     TAILQ_ENTRY(DNSTransaction_) next;
     DetectEngineState *de_state;
+    uint8_t request_logged;
```
inliniac: minor: this grows the data structure. Maybe it would be better to add a flags field that can replace this and most of the uint8_t's above.

src/output-tx.c
```diff
@@ -137,19 +143,21 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
...
+        }
 
-            tx_progress = AppLayerParserGetStateProgress(p->proto, alproto,
-                                                         tx, FlowGetDisruptionFlags(f, STREAM_TOCLIENT));
-            if (tx_progress < tx_progress_done_value_tc) {
-                SCLogDebug("progress not far enough, not logging");
-                break;
-            }
+        if (!(ts_ready || tc_ready)) {
+            SCLogNotice("progress not for enough, not logging");
```
inliniac: don't forget to remove/convert

inliniac:
I think it would be good to document this on the wiki.

inliniac:
I think the overall design makes sense.

jasonish:
> I think it would be good to document this on the wiki.

Developer docs? User docs?

inliniac:
API/dev docs

src/app-layer-dns-common.h
jasonish: I noticed we don't make much use of bitfields..  Is there anything wrong with:

`
    uint8_t replied:1;
    uint8_t reply_lost:1;
    uint8_t recursion_desired:1;
    uint8_t request_logged:1;
    uint8_t response_logged:1;
    uint8_t padding:3;
`

jasonish:
> API/dev docs

I'm wondering if this is a good idea to have in the Wiki or not? It will tend to go out of date, not just with respect to this, but with any API docs that are in the wiki and not auto generated.

Perhaps effort should be made so that Doxygen produces nice output that can be used to document the API and serves as the official reference.  The Wiki could then have a guide of sorts for implementing an app-layer, but still references API documentation for the source of truth.

IMO, out of date/inaccurate documentation is almost worse than no documentation.

PR #1809, opened 2016-01-04
json-stats: add per app-layer stats v2

glongo:
This adds adds the number of application layer transactions handled
by the supported protocols.

In some application layer, like ssh, we don't have a transaction,
so we count the number of the states.

The output added in EVE will be:
"app-layer": { "http": 234, "stmp": 12, "ssh": 22, "dns": 2000 }

Last PR: #1806

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/94
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/93

### DIFF
--- a/src/app-layer-dns-common.c
+++ b/src/app-layer-dns-common.c
@@ -283,7 +283,7 @@ static DNSTransaction *DNSTransactionAlloc(DNSState *state, const uint16_t tx_id
 /** \internal
  *  \brief Free a DNS TX
  *  \param tx DNS TX to free */
-static void DNSTransactionFree(DNSTransaction *tx, DNSState *state)
+void DNSTransactionFree(DNSTransaction *tx, DNSState *state)
 {
     SCEnter();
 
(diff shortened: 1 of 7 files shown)


inliniac:
What's different? I still see the globals based approach here.

glongo:
I've removed the use of atomic definition and use directly the counter.
But if it's still not correct, I'd really appreciate if you explain me what
do you mean with "regular counters"

inliniac:
Counters like those used in the decoders. Global atomics are expensive and should be used rarely.

regit:
@glongo some mistakes in this PR so previous is better but we need to address Victor's comment

@inliniac I've just checked the code and as the patchset increments the counter in HTPStateTransactionFree it seems we don't have access to any thread context. This place seems to be a good one but it would be great if you already see another place where where we could access a *tv. If not @glongo and I can try to look.

inliniac:
Yeah we'd need to add the tv pointer to the app layer API.

PR #1813, opened 2016-01-15
Copy the improved strip of double quotes code to every old section

norg:
This fixes also the issue 1638

prscript output:
- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/0
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/0

### DIFF
--- a/src/detect-depth.c
+++ b/src/detect-depth.c
@@ -62,12 +62,14 @@ static int DetectDepthSetup (DetectEngineCtx *de_ctx, Signature *s, char *depths
     SigMatch *pm = NULL;
     int ret = -1;
 
-    /* strip "'s */
-    if (depthstr[0] == '\"' && depthstr[strlen(depthstr) - 1] == '\"') {
+    /* Strip leading and trailing "s. */
+    if (depthstr[0] == '\"') {
         str = SCStrdup(depthstr + 1);
         if (unlikely(str == NULL))
             goto end;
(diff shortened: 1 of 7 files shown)


inliniac:
Couple of comments:
- please add a link to the ticket
- I think a utility function is needed here. We have same code in too many places.
- commit message. We have a convention that we prefix the msg with the sub-system. E.g. "rule parsing: fix quotation parsing issues".

inliniac:
Next step: https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Incorporating-feedback :)

norg:
Will do, before i add it to the wrong place, where should the utility function be placed, since we have quite a lot util-* :)

inliniac:
In this case detect-parse.c would probably be the best file.

norg:
Hmm i would have rather put it in util-*.h similar to SCStrdup concept and there are some trade offs depending on the solution you prefer. We either mess with the pointer for "str" (saves memory) or have to work on a copy. Or did you have something totally else in mind when talking about utility function in regard to detect-parse.c?

inliniac:
I think Jason was suggesting we strip the quotes before the per keyword parse functions are called. That may be the best idea so far. That would have to be done in SigParseOptions(). I worry there might be unforeseen side effects though, but it's worth a try.

PR #1815, opened 2016-01-20
Magic mgc loadfix 2

norg:
This fixes the Issue 873 as it copies the way magic_load checks for the .mgc file (by adding .mgc itself if the file is without .mgc in the config).

- Issue: https://redmine.openinfosecfoundation.org/issues/873
- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/4
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/4

### DIFF
--- a/src/detect-filemagic.c
+++ b/src/detect-filemagic.c
@@ -303,6 +303,7 @@ error:
 static void *DetectFilemagicThreadInit(void *data)
 {
     char *filename = NULL;
+    char *filename_mgc = NULL;
     FILE *fd = NULL;
     DetectFilemagicData *filemagic = (DetectFilemagicData *)data;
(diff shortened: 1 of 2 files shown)


inliniac:
Yikes, did you just copy paste the same code in 2 places? Util func?

norg:
Shall i put/keep the util func into util-magic.c and use that within detect-filemagic.c (which already includes util-magic)?

inliniac:
On 20-01-16 16:17, Andi wrote:
> Shall i put/keep the util func into util-magic.c and use that within
> detect-filemagic.c (which already includes util-magic)?
>

Yeah sounds good.

PR #1820, opened 2016-01-28
file-magic: improve libmagic handling on *nix systems

norg:
This fixes https://redmine.openinfosecfoundation.org/issues/873 without changing the code too much.
So for *nix systems we can skip setting magic-file since it will load the system default which might differ between distributions and libmagic (file) package versions. We just need to set it properly on Windows.

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/5
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/5

### DIFF
--- a/src/detect-filemagic.c
+++ b/src/detect-filemagic.c
@@ -322,13 +322,20 @@ static void *DetectFilemagicThreadInit(void *data)
 
     (void)ConfGet("magic-file", &filename);
     if (filename != NULL) {
-        SCLogInfo("using magic-file %s", filename);
-
-        if ( (fd = fopen(filename, "r")) == NULL) {
-            SCLogWarning(SC_ERR_FOPEN, "Error opening file: \"%s\": %s", filename, strerror(errno));
-            goto error;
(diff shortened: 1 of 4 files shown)


src/detect-filemagic.c
```diff
@@ -322,13 +322,20 @@ static void *DetectFilemagicThreadInit(void *data)
 
     (void)ConfGet("magic-file", &filename);
     if (filename != NULL) {
-        SCLogInfo("using magic-file %s", filename);
+        if (strncmp(filename,"",strlen(filename)) == 0) {
```
inliniac: this could be written simpler: strcmp(filename,"") == 0, or just strlen(filename) == 0

src/detect-filemagic.c
```diff
@@ -322,13 +322,20 @@ static void *DetectFilemagicThreadInit(void *data)
...
+            // set filename to NULL on *nix systems so magic_load uses system default path (see man libmagic)
```
inliniac: we use /* */ style comments, even though you'll see the c++ // variant lingering around as well

PR #1822, opened 2016-01-29
file-magic: improve libmagic handling on *nix systems

norg:
This fixes https://redmine.openinfosecfoundation.org/issues/873 without changing the code too much.
So for *nix systems we can skip setting magic-file since it will load the system default which might differ between distributions and libmagic (file) package versions. We just need to set it properly on Windows, which still works (tested by pevma).

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/6
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/6

This PR fixes the notes from the previous PR https://github.com/inliniac/suricata/pull/1820

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -208,7 +208,8 @@
     fi
     echo -n "installation for $host OS... "
 
-    e_magic_file="/usr/share/file/magic"
+    e_magic_file=""
+    e_magic_file_comment="#"
     case "$host" in
         *-*-*freebsd*)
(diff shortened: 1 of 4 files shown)


norg:
Sorry for the merge pull, if you want a new PR without that i will create one.

inliniac:
Yup I want it w/o that :)

PR #1824, opened 2016-01-31
configure: remove unnecessary braces

norg:
They don't harm, since they just show up when using ./configure -h but are unnecessary :)

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/7
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/7


inliniac:
Merged through #1842, thanks.

PR #1828, opened 2016-02-02
Correct the hiredis-devel package name recommeded by configure script…

btbytes:
… for CentOS


inliniac:
Implemented in #2343 somewhat differently.

PR #1829, opened 2016-02-03
Fix typo of trailing ] in configure --help

bladeswords:
It is the small things that count.  This is an example of the fix:

Before
`--disable-threading-tls Disable TLS (thread local storage)]`

After
`--disable-threading-tls Disable TLS (thread local storage)`


inliniac:
Could you email me at victor@inliniac.net? Thanks!

inliniac:
Merged through #1845, thanks!

PR #1830, opened 2016-02-03
Fix stack overflow when parsing variables

gozzy:
For instance, a crash occured when HOME_NET is specified as $HOME_NET


gozzy:
Maybe there's more tidy way to do this, but the checks allow to handle the error and notify user about misconfiguration.

PR #1831, opened 2016-02-03
Fix stack overflow when parsing variables

gozzy:
For instance, a crash occured when HOME_NET is specified as $HOME_NET

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -1053,6 +1053,11 @@ static int DetectAddressParse2(const DetectEngineCtx *de_ctx,
                     snprintf(temp_rule_var_address, strlen(rule_var_address) + 3,
                              "[%s]", rule_var_address);
                 }
+
+                if (!strcmp(s, temp_rule_var_address) || strstr(temp_rule_var_address, s)) {
+                    goto recursive_error;
+                }
+
                 DetectAddressParse2(de_ctx, gh, ghn, temp_rule_var_address,
                                     (negate + n_set) % 2);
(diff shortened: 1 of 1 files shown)


inliniac:
Can you explain the problem in a bit more detail? Create some example rules that crash for you?

gozzy:
Actually it occurs in case of misconfiguration. For instance, if in yaml-file we specify the following:

vars:
    HOME_NET: "!$HOME_NET"

It's definitely wrong configuration, but suricata recursively tries to resolve a variable until stack is exhausted and segmentation fault occurs. Valgrind shows this:

==11392== Stack overflow in thread #1: can't grow stack to 0xffe801000
==11392== 
==11392== Process terminating with default action of signal 11 (SIGSEGV)

I'm not sure if it's the best way to handle this misconfiguration... Furthermore, we can do something like this:

vars:
    HOME_NET: "$EXTERNAL_NET"
    EXTERNAL_NET: "$HOME_NET"

This is more complicated case, and it also leads to segmentation fault.

inliniac:
I see, thanks. Could you open a redmine ticket for this bug as well? Thanks!

gozzy:
Done: https://redmine.openinfosecfoundation.org/issues/1689

I'll try to look at the second case (when A is defined as B and vice versa).

gozzy:
@inliniac probably we should just limit the recursion depth without tricky strings comparison.

gozzy:
Got an idea about possible 'loop detection', gonna check it.

PR #1832, opened 2016-02-03

gozzy:
Reworked #1831.

Suricata crashed when variable referred to itself. For instance, if one declared HOME_NET as !$HOME_NET or created a looped chain of variables.

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -892,6 +892,18 @@ error:
 }
 
 /**
+ * A list of variables we try to resolve.
+ * Used by DetectAddressParse2() to detect loops in variables declaration
+ */
+typedef struct ResolvedVariable_ {
+    char *var_name;
+    TAILQ_ENTRY(ResolvedVariable_) next;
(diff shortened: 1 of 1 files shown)


src/detect-engine-address.c
```diff
@@ -892,6 +892,18 @@ int DetectAddressSetup(DetectAddressHead *gh, char *s)
...
 /**
+ * A list of variables we try to resolve.
+ * Used by DetectAddressParse2() to detect loops in variables declaration
+ */
+typedef struct ResolvedVariable_ {
+    char *var_name;
+    TAILQ_ENTRY(ResolvedVariable_) next;
+} ResolvedVariable;
+
+static TAILQ_HEAD(, ResolvedVariable_) resolved_variables_list =
```
inliniac: if no one outside of DetectAddressParse2 accesses this list, can we move it into DetectAddressParse2? Saves us a global.

gozzy: Ok, I'll fix this.

PR #1834, opened 2016-02-04
In configuration test mode, check signatures if 'delayed-detect' is enabled

gozzy:
When 'detection-engine.delayed-detect' option was set to 'yes', suricata didn't check signatures validity in configuration test mode. IMHO, it looks like a bug.


inliniac:
Merged through #1842, thanks a lot Alexander!

PR #1835, opened 2016-02-05
lua: set thread vars in DetectLuaMatch

p1-nico:
Fix internal error when calling SCThreadInfo from Lua detection modules.


inliniac:
Merged through #1845, thanks!

PR #1836, opened 2016-02-05
Fix install permissions for configuration files

gozzy:
When suricata runs under non-root account with 'delayed-detect' enabled,
it can not load signatures properly due to insufficient access rights.
With disabled 'delayed-detect' everything works fine because privileges
are dropped after signatures loading.

So let's change configuration dir/files access rights a bit.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -15,11 +15,11 @@ install-data-am:
 install-full: install install-conf install-rules
 
 install-conf:
-	install -d "$(DESTDIR)$(e_sysconfdir)"
-	@test -e "$(DESTDIR)$(e_sysconfdir)/suricata.yaml" || install -m 600 "$(top_srcdir)/suricata.yaml" "$(DESTDIR)$(e_sysconfdir)"
-	@test -e "$(DESTDIR)$(e_sysconfdir)/classification.config" || install -m 600 "$(top_srcdir)/classification.config" "$(DESTDIR)$(e_sysconfdir)"
-	@test -e "$(DESTDIR)$(e_sysconfdir)/reference.config" || install -m 600 "$(top_srcdir)/reference.config" "$(DESTDIR)$(e_sysconfdir)"
-	@test -e "$(DESTDIR)$(e_sysconfdir)/threshold.config" || install -m 600 "$(top_srcdir)/threshold.config" "$(DESTDIR)$(e_sysconfdir)"
+	install -m 755 -d "$(DESTDIR)$(e_sysconfdir)"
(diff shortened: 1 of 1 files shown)


regit:
I disagree with the following modifications. In the case of suricata running as non-root, the only thing we want is the `suricata` user to be able to read the file. So the permission changes should be to 740 with a group set to be one `suricata` user belong to.

Problem is this `suricata` user can not be created by the installation script. So we can't set it up at installation. By consequence we can't change the group permission.

The issue is triggered by `delayed-detect` but could also be triggered by rule reload. So this is a user permission issue. As it is not enable by default we could just have a comment in the YAML explaining the  issue and telling that admin should change permission of files to be readable by the user suricata will switch to.

gozzy:
There must be a reason. I don't see anything 'top secret' in these dir and files that may result in 740 or 600 permissions. If we install an RPM package, we may expect that permissions are sane. Commenting the YAML is good, but I really don't see a reason for such a limited permissions.
For instance, snort sets 755 for configuration directory and 644 for configuration files. What are we trying to hide and for what reason?

regit:
Suricata configuration is the security policy of the IDS and it should not be read by non-privileged user. For instance, a suricata can be deployed to watch over traffic by user on the box (it may look weird but there is some example of it). In that case, you don't want the users to read the configuration. Furthermore I've seen some security recommendation stating that such config file must not be read by common user.

If the RPM package don't handle thing correctly then just change distro ;) Correct permission setup is the work of distribution. On snort comparison, I would not check suricata permission against decade old snort install system.

If we comment the YAML given some precise indication and pointing to a doc, then any user changing the default value will see the potential problem and take correct action. If it does not do so then we will RTFM him. This applies too to the package maintainers.

inliniac:
I'm leaning in the opposite direction for this issue. While I think @regit's points are all valid, there are other considerations that need to be considered.

- the make install-full config is not meant replace a full installed/packaging, instead it's meant for getting started quickly
- the following things will break in non-obvious (to the user) ways: dropping privs in combination with delayed detect, rule reloads, adding/removing tenants through unix socket
- pcap processing from the commandline will not work as a regular user, forcing users to use root/sudo which is not recommended for security
- unix socket runmode won't work in the recommended non-root use
- I suspect that by far most deployments of Suricata are on single user systems

I think we can use more liberal permissions here. Packaging can go much further in doing things like setting up a suricata user and group, setting restrictive permissions on input and output

I think regardless of this issue, it would be good to write a user document about deployment best practices with a big section about managing privileges, and other security measures.

gozzy:
Should I reopen this request or there're some more things to discuss? Actually, I don't think there're any security flaws related to these permissions, and everything will work out of the box.

PR #1838, opened 2016-02-06
fix nfq_get_timestamp

notnyt:
Handle case when nfq_get_timestamp returns 0 for success, but timestamp is empty.


inliniac:
Merged through #1845, thanks!

PR #1791, opened 2015-12-14
DNP3 Support - v6

jasonish:
This PR contains:
- An application decoder for DNP3.
- Snort compatible DNP3 keywords:
  - dnp3_data
  - dnp3_func
  - dnp3_ind
  - dnp3_obj
- A DNP3 output module for EVE (JSON logging).
- DNP3 support for Lua detection.

Documentation for the DNP3 rule keywords can be found here:
https://redmine.openinfosecfoundation.org/projects/suricata/wiki/DNP3_keyword
OR in newer format:
http://jasonish-suricata.readthedocs.org/en/latest/rules/dnp3-keywords.html

NOTES:
- Not all objects are supported due to lack of traffic samples. I
  suspect this will remain work in progress for some time.
- I started to use code generation for the Lua detection support. If
  this looks OK, I'd like to add generation for the JSON logging as
  well. I've decided to check the generated code into VC so builders
  don't have to worry about code generation. It is also available for
  static analysis.

Buildbot output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/167
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/169

PR #1844, opened 2016-02-09
DNP3 Support - v7

jasonish:
Replaces #1791.

This PR contains:
An application decoder for DNP3.
* Snort compatible DNP3 keywords:
  * dnp3_data
  * dnp3_func
  * dnp3_ind
  * dnp3_obj
* A DNP3 output module for EVE (JSON logging).
* DNP3 support for Lua detection.

Documentation for the DNP3 rule keywords can be found here:
OR in newer format:
http://jasonish-suricata.readthedocs.org/en/latest/rules/dnp3-keywords.html

Changes from the last PR include:
* Misc cleanups.
* More objects.
* Generating object logging code.

PRScript output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/184
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/186


rules/dnp3-events.rules
```diff
@@ -0,0 +1,21 @@
+# Flooded.
```
inliniac: File misses a header like http-events.rules has

src/util-error.h
```diff
@@ -271,6 +271,7 @@ typedef enum {
     SC_ERR_THRESHOLD_SETUP,
     SC_ERR_DNS_CONFIG,
     SC_ERR_MODBUS_CONFIG,
+    SC_ERR_DNP3_CONFIG,
```
inliniac: new errors should be appended to keep the error codes stable

src/output-json-dnp3.c
```diff
@@ -0,0 +1,464 @@
...
+    json_t *jsitems;
+
+    if (unlikely((jsitems = json_array()) == NULL)) {
+        return NULL;
+    }
+
+    TAILQ_FOREACH(item, object->points, next) {
+        json_t *js = json_object();
+        if (unlikely(js == NULL)) {
+            break;
```
inliniac: jsitems is lost here

src/output-json-dnp3.c
```diff
@@ -0,0 +1,464 @@
...
+#include "detect-dnp3.h"
+
+#include "output.h"
+#include "output-json.h"
+#include "output-json-dnp3-objects.h"
+
+#ifdef HAVE_LIBJANSSON
+#include <jansson.h>
+
+/* Appears not all current distros have jansson that defines this. */
```
inliniac: move this to suricata-common.h?

src/output-json-dnp3.c
jasonish: Its still returned even in this case, which means it gets referenced by the parent object, even if empty, so will get free'd with the complete object.

Perhaps that would be more clear if we allocated the json_t object in the caller and passed it in?

inliniac: Looks like I misread the code.

inliniac:
This will need AFL entry points similar to https://github.com/inliniac/suricata/pull/1999/files#diff-f5e72d82198f985531c162cd7fca0b54R1430

Then we'll hammer it with some fuzz testing.

PR #1847, opened 2016-02-10
lua: add an SCPacketTimeNumber function

p1-nico:
The SCPacketTimeNumber returns the packet timestamp as a real number (where integral part = tv_sec & fractional part = tv_usec)

### DIFF
--- a/src/util-lua-common.c
+++ b/src/util-lua-common.c
@@ -159,6 +159,34 @@ static int LuaCallbackPacketPayload(lua_State *luastate)
 }
 
 /** \internal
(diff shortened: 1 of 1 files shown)


inliniac:
What about pushing 2 numbers back instead?
sec, usec = SCPacketTimeXXX()

Seems a lot simpler on the C code side and to use as well.

p1-nico:
Sounds good to me, should I update the PR?

inliniac:
Please do a new branch and new PR. See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Incorporating-feedback

Thanks!

PR #1849, opened 2016-02-11

thus:

Changes:
* Fixed time zone handling

Fixed changes suggested in #1728.


thus:
When using "CreateIsoTimeString" instead of the other function I added, we add timezone difference to the "notBefore" and "notAfter" timestamps. I don't think that's a good solution. What do you think?

This means that if I log on a server with GMT+1, then all timestamps would be changed one hour, instead of being printed as UTC.

thus:
What about adding a "CreateUtcTimeString" function?

thus:
Hmf, "strftime" adds timezone difference, so my previous code was also wrong. I must admit I really don't like time handling in C..

It's actually better to print the timestamp in local time and adding timezone info (like we do now), than printing a wrong timestamp pretending it's UTC.

thus:
Replaced by #1930

PR #1850, opened 2016-02-11
json: use top-level sensor-name if provided - v0

jasonish:
Currently the default configuration file contains a "sensor-name"
at the root of the configuration file, however, eve-log will only
use it if its specified under eve-log.

Now we will look for it at the eve-log, if present we'll use it
but log a deprecation warning, if its not present we'll look
for sensor-name at the root of the configuration.

Related Redmine issues:
https://redmine.openinfosecfoundation.org/issues/1679
https://redmine.openinfosecfoundation.org/issues/1680

Still unsure about the output in json being "host". IMO "sensor_name" would be better, but I don't want to break existing integrations.

Prscript output:
- https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/186
- https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/188

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -428,7 +428,17 @@ OutputCtx *OutputJsonInitCtx(ConfNode *conf)
 {
     OutputJsonCtx *json_ctx = SCCalloc(1, sizeof(OutputJsonCtx));;
 
+    /* First lookup a sensor-name value in this outputs configuration
+     * node (deprecated). If that fails, lookup the global one. */
     const char *sensor_name = ConfNodeLookupChildValue(conf, "sensor-name");
+    if (sensor_name != NULL) {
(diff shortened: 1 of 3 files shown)


inliniac:
Wonder how this made it through QA...
<pre>
  CC       output-json.o
output-json.c: In function ‘OutputJsonInitCtx’:
output-json.c:447:9: error: passing argument 2 of ‘ConfGet’ from incompatible pointer type [-Werror]
         ConfGet("sensor-name", &sensor_name);
         ^
In file included from output-json.c:31:0:
conf.h:59:5: note: expected ‘char **’ but argument is of type ‘const char **’
 int ConfGet(const char *name, char **vptr);
     ^
</pre>

inliniac:
Updated the buildbot (prscript) to kick this kind of error out.

PR #1585, opened 2015-07-09
tls lua v0

inliniac:
Detection only for now: https://redmine.openinfosecfoundation.org/issues/1502

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/260
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/262

PR #1853, opened 2016-02-16
Ruleset stats v4

glongo:

First feature is an update on stats JSON with information added about the ruleset so user can know and monitor how much rules are loaded and failed.

    "detect":{"engine":{"last_reload":"2015-10-13T09:59:48.044996+0200","rules_loaded":17184,"rules_failed":0},"alert":28}

Other changes are related to unix socket with some new commands:

- last-reload
- ruleset-stats: display information about number of loaded and failed rules
- show-failed-rules: display a list of failed rules (dc53401 for more details)

This PR also contains a patch implementing a non blocking reload command in unix socket. As we can know with this patchset when reload is done we can also have an asynchronous command.

This also provides ruleset information through unix-socket when multi-tenancy is enabled.

Last PR: https://github.com/inliniac/suricata/pull/1778
Ticket: [#1585] (https://redmine.openinfosecfoundation.org/issues/1585)

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/97
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/96

### DIFF
--- a/src/detect.c
+++ b/src/detect.c
@@ -447,14 +447,12 @@ int SigLoadSignatures(DetectEngineCtx *de_ctx, char *sig_file, int sig_file_excl
 
     ConfNode *rule_files;
     ConfNode *file = NULL;
-    SigFileLoaderStat sig_stat;
+    SigFileLoaderStat *sig_stat = &de_ctx->sig_stat;
     int ret = 0;
     char *sfile = NULL;
     char varname[128] = "rule-files";
     int good_sigs = 0;
     int bad_sigs = 0;
-
-    memset(&sig_stat, 0, sizeof(SigFileLoaderStat));
 
     if (strlen(de_ctx->config_prefix) > 0) {
         snprintf(varname, sizeof(varname), "%s.rule-files",
@@ -479,7 +477,7 @@ int SigLoadSignatures(DetectEngineCtx *de_ctx, char *sig_file, int sig_file_excl
                 TAILQ_FOREACH(file, &rule_files->head, next) {
                     sfile = DetectLoadCompleteSigPath(de_ctx, file->val);
                     good_sigs = bad_sigs = 0;
-                    ret = ProcessSigFiles(de_ctx, sfile, &sig_stat, &good_sigs, &bad_sigs);
+                    ret = ProcessSigFiles(de_ctx, sfile, sig_stat, &good_sigs, &bad_sigs);
                     SCFree(sfile);
 
                     if (ret != 0 || good_sigs == 0) {
@@ -494,7 +492,7 @@ int SigLoadSignatures(DetectEngineCtx *de_ctx, char *sig_file, int sig_file_excl
 
     /* If a Signature file is specified from commandline, parse it too */
     if (sig_file != NULL) {
-        ret = ProcessSigFiles(de_ctx, sig_file, &sig_stat, &good_sigs, &bad_sigs);
+        ret = ProcessSigFiles(de_ctx, sig_file, sig_stat, &good_sigs, &bad_sigs);
 
         if (ret != 0) {
             if (de_ctx->failure_fatal == 1) {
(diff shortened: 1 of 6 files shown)


src/detect.c
```diff
@@ -292,6 +292,70 @@ char *DetectLoadCompleteSigPath(const DetectEngineCtx *de_ctx, char *sig_file)
     return path;
 }
 
+static SigString *SigStringAlloc(void)
+{
+    SigString *sigstr = SCCalloc(2, sizeof(SigString));
```
inliniac: why 2? seems a typo

src/output-json-stats.c
```diff
@@ -55,6 +57,14 @@
 #define JSON_STATS_THREADS (1<<1)
 #define JSON_STATS_DELTAS  (1<<2)
 
+/**
+ * specify which engine info will be printed in stats log.
+ * ALL means both last reload and ruleset stats.
+ */
+#define JSON_ENGINE_LAST_RELOAD    (1<<0)
```
inliniac: these are defined as bits but not used as such

inliniac: given how it's used, an enum might make more sense

src/output-json-stats.c
```diff
@@ -65,6 +75,111 @@ typedef struct JsonStatsLogThread_ {
...
+static json_t *OutputTenancyStats2Json(int output)
+{
+    json_t *jdata = json_object();
+    if (jdata == NULL) {
+        return NULL;
+    }
+
+    DetectEngineCtx *de_ctx = DetectEngineGetCurrent();
+    if (de_ctx == NULL) {
+        SCFree(jdata);
```
inliniac: wrong free func

src/output-json-stats.c
```diff
@@ -65,6 +75,111 @@ typedef struct JsonStatsLogThread_ {
...
+        json_array_append_new(js_tenant_list, js_tenant);
+        list = list->next;
+    }
+
+    DetectEngineDeReference(&de_ctx);
+    return js_tenant_list;
+
+error:
+    DetectEngineDeReference(&de_ctx);
+    return NULL;
```
inliniac: if we get here it seems jdata is lost

src/unix-manager.c
```diff
@@ -647,18 +649,107 @@ TmEcode UnixManagerCaptureModeCommand(json_t *cmd,
...
+}
+
+TmEcode UnixManagerLastReloadCommand(json_t *cmd,
+                                     json_t *server_msg, void *data)
+{
+    SCEnter();
+    TmEcode retval = TM_ECODE_OK;
+    json_t *jdata = NULL;
+
+    retval = OutputTenancyStatsLastReload(&jdata);
```
inliniac: why not just have OutputTenancyStatsLastReload return the jdata object directly? And use NULL as error? Seems simpler.

src/unix-manager.c
```diff
@@ -647,18 +649,107 @@ TmEcode UnixManagerCaptureModeCommand(json_t *cmd,
...
+        }
+        list = list->next;
+    }
+
+    json_object_set_new(server_msg, "message", js_sigs_array);
+    DetectEngineDeReference(&de_ctx);
+    SCReturnInt(TM_ECODE_OK);
+
+error:
+    DetectEngineDeReference(&de_ctx);
```
inliniac: seems we can loose js_sigs_array here

src/unix-manager.c
```diff
@@ -877,6 +968,10 @@ static TmEcode UnixManagerThreadInit(ThreadVars *t, void *initdata, void **data)
     UnixManagerRegisterCommand("conf-get", UnixManagerConfGetCommand, &command, UNIX_CMD_TAKE_ARGS);
     UnixManagerRegisterCommand("dump-counters", StatsOutputCounterSocket, NULL, 0);
     UnixManagerRegisterCommand("reload-rules", UnixManagerReloadRules, NULL, 0);
+    UnixManagerRegisterCommand("non-blocking-reload-rules", UnixManagerNonBlockingReloadRules, NULL, 0);
```
inliniac: think I like reload-rules-non-blocking better

src/detect.c
```diff
@@ -292,6 +292,70 @@ char *DetectLoadCompleteSigPath(const DetectEngineCtx *de_ctx, char *sig_file)
     return path;
 }
 
+static SigString *SigStringAlloc(void)
```
inliniac: Thinking we should maybe put this in a new/different file. Trying to reduce the size of detect.c and this seems more like a util than a core detect feature.

src/unix-manager.c
glongo: OutputTenancyStatsLastReload can also set an error message like "Detect engine is not yet ready", and I need also a TM_ECODE_FAILED value in retval (unix-socket).
In this way it's possible to display the message above as error.

PR #1854, opened 2016-02-16
decode-error-stats updated version

axelrodGunnarson:
Following the suggestion, we synced the decode and stream events in order to remove dynamic allocation of counter names.
This also corrected qa failure (as we do not have dynamic allocation any longer).

For now, I left stream events out. That's because while every decoder event is related to a single packet and has a single point of failure (after decoding), stream events can be related to as single packet or to a reassembly stream. This means we can do one of the two:
1. separate stream events related to reassembly/stream sequence from the single tcp events and put the counters in the different context (single packet => StreamTcpThread structure, reassembly: TcpReassemblyThreadCtx) and put the different counters in different place according to its type.
2. put everything together and change some of the functions declaration in order to always have a pointer to StreamTcpThread and be able to update the counters


inliniac:
What is the relation between the @axelrodGunnarson and @maurizioabba accounts?

inliniac:
Rebased version passes prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/366
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/369

inliniac:
Commit message will have to be improved. It will need to explain the code, not the changes since the last pull request. Changes since the last PR should be in the PR itself.

Also, the formatting of the commit message needs to be fixed. See point 5 in https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits  As an example, check https://github.com/inliniac/suricata/commit/6b37985c666db978af42cc095b4719af46cf7a43

inliniac:
I see the risk for side effects as being low, would you all agree? If so, I would like to take this into the 3.0 branch.

inliniac:
Replaced by #1921

PR #1855, opened 2016-02-16
detect-tls-version: fix small resource leak

norg:
That fixes the leak from https://redmine.openinfosecfoundation.org/issues/1166 as in this case we jump into error and skip the SCFree(orig) which should still be called as orig was != NULL.

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/10
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/10


inliniac:
Merged through #1868

PR #1857, opened 2016-02-17
Mixed mode v7

glongo:
A rebased patchset of https://github.com/inliniac/suricata/pull/1773

There is also a small change in ca9feda:
In the previous code we set a flag only if a packet is not in ips mode:

	if (!PacketModeIsIPS(p)) {
	    if (p->flowflags & FLOW_PKT_TOSERVER) {
		    flag = FLOW_PKT_TOCLIENT;
	    } else {
		    flag = FLOW_PKT_TOSERVER;
	    }
	}

The first test has been removed since doesn't look correct and it also fires up the following warning:

    output-json-alert.c: In function ‘JsonAlertLogger’:
    output-json-alert.c:286:17: warning: ‘flag’ may be used uninitialized in this function [-Wmaybe-uninitialized]
                 StreamSegmentForEach((const Packet *)p, flag,
                 ^
    output-json-alert.c:274:25: note: ‘flag’ was declared here
                 uint8_t flag;

[Old code https://github.com/glongo/suricata/commit/2f447d6dfe53a84f66218a9aa3d60de8ac6084b2#diff-e2a291d82d2cff36c51007fafaa79b86R277]:

Ticket: [#1604] (https://redmine.openinfosecfoundation.org/issues/1604)

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/98
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/97

### DIFF
--- a/src/runmodes.c
+++ b/src/runmodes.c
@@ -49,6 +49,8 @@
 
 #include "source-pfring.h"
 
+extern RunModesList runmodeslist;
+
 int debuglog_enabled = 0;
 
 /**
@@ -188,9 +190,9 @@ char *RunmodeGetActive(void)
  *
  * \return a string containing the current running mode
  */
-const char *RunModeGetMainMode(void)
+const char *RunModeGetMainMode(int index)
 {
-    int mainmode = RunmodeGetCurrent();
+    int mainmode = RunmodeGetCurrent(&runmodeslist, index);
 
     return RunModeTranslateModeToName(mainmode);
 }
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -782,6 +782,7 @@ int AFPReadFromRing(AFPThreadVars *ptv)
             SCReturnInt(AFP_FAILURE);
         }
         PKT_SET_SRC(p, PKT_SRC_WIRE);
+        p->pkt_mode = PKT_MODE_AUTO;
 
         /* Suricata will treat packet so telling it is busy, this
          * status will be reset to 0 (ie TP_STATUS_KERNEL) in the release
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -1682,7 +1682,7 @@ TmEcode ReceiveAFPThreadInit(ThreadVars *tv, void *initdata, void **data)
     strlcpy(ptv->iface, afpconfig->iface, AFP_IFACE_NAME_LENGTH);
     ptv->iface[AFP_IFACE_NAME_LENGTH - 1]= '\0';
 
-    ptv->livedev = LiveGetDevice(ptv->iface);
+    ptv->livedev = LiveGetDevice(ptv->iface, RUNMODE_AFP_DEV);
     if (ptv->livedev == NULL) {
         SCLogError(SC_ERR_INVALID_VALUE, "Unable to find Live device");
         SCFree(ptv);
--- a/src/stream-tcp.c
+++ b/src/stream-tcp.c
@@ -120,7 +120,7 @@ uint64_t StreamTcpReassembleMemuseGlobalCounter(void);
 SC_ATOMIC_DECLARE(uint64_t, st_memuse);
 
 /* stream engine running in "inline" mode. */
-int stream_inline = 0;
+enum PktMode stream_inline;
 
 void TmModuleStreamTcpRegister (void)
 {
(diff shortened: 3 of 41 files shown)


src/source-af-packet.c
```diff
@@ -782,6 +782,7 @@ int AFPReadFromRing(AFPThreadVars *ptv)
             SCReturnInt(AFP_FAILURE);
         }
         PKT_SET_SRC(p, PKT_SRC_WIRE);
+        p->pkt_mode = PKT_MODE_AUTO;
```
inliniac: What does this do? Both PacketModeIsIPS and PacketModeIsIDS will return false on this on AFAICS

src/util-runmodes.c
```diff
@@ -104,15 +104,15 @@ int RunModeSetLiveCaptureAutoFp(ConfigIfaceParserFunc ConfigParser,
                               ConfigIfaceThreadsCountFunc ModThreadsCount,
                               char *recv_mod_name,
                               char *decode_mod_name, char *thread_name,
-                              const char *live_dev)
+                              const char *live_dev, int runmode)
```
inliniac: Should this be enum Runmode as well instead of 'int'?

src/runmodes.c
```diff
@@ -188,9 +190,9 @@ char *RunmodeGetActive(void)
  *
  * \return a string containing the current running mode
  */
-const char *RunModeGetMainMode(void)
+const char *RunModeGetMainMode(int index)
```
inliniac: I'm a bit confused by the use of 'RunModeGetMainMode' and 'RunmodeGetPrimary'. Do we need both?

glongo: RunModeGetMainMode is used to output the running mode through unix-socket, instead RunmodeGetPrimary returns the runmode id. 
So, yes we need both currently.

src/util-runmodes.c
glongo: yes.

src/source-af-packet.c
glongo: Since afpacket can work both in ips and ids mode, setting this means that it's allowed to do ips or ids on the packet if stream.inline is enabled or not.

glongo: PKT_MODE_AUTO works with StreamTcpInlineMode().

inliniac: I don't think it works. StreamTcpInlineMode() checks explicitly for IDS or IPS, or "return (stream_inline && PacketModeIsIPS(p));" PacketModeIsIPS again, checks for IPS. So where does 'AUTO' get handled?

glongo: I think that this function should be rework as below:

    if (PacketModeIsIDS(p))
        return 0;
    else if (PacketModeIsIPS(p))
        return 1;
    else /* implied PKT_MODE_AUTO */
        return stream_inline;

So what we do here is to decide if we do IPS or IDS on the packet, but if the pkt_mode is 'AUTO' (like in af-packet)
we check if stream.inline is enabled or not.

inliniac: On 12-04-16 08:44, Giuseppe Longo wrote:
> I think that this function should be rework as below:
> 
> |if (PacketModeIsIDS(p)) return 0; else if (PacketModeIsIPS(p)) return 1;
> else /* implied PKT_MODE_AUTO */ return stream_inline; |
> 
> So what we do here is to decide if we do IPS or IDS on the packet, but
> if the pkt_mode is 'AUTO' (like in af-packet)
> we check if stream.inline is enabled or not.
> 

What if 'stream.inline' is set to AUTO itself? Would that work?

src/stream-tcp.c
```diff
@@ -424,19 +424,14 @@ void StreamTcpInitConfig(char quiet)
...
     if (ConfGet("stream.inline", &temp_stream_inline_str) == 1) {
         /* checking for "auto" and falling back to boolean to provide
          * backward compatibility */
         if (strcmp(temp_stream_inline_str, "auto") == 0) {
-            if (EngineModeIsIPS()) {
-                stream_inline = 1;
-            } else {
-                stream_inline = 0;
-            }
+            stream_inline = PKT_MODE_AUTO;
```
glongo: I think we can't set AUTO to stream_inline here. It should be only IPS or IDS.

src/stream-tcp.c
```diff
@@ -424,19 +424,14 @@ void StreamTcpInitConfig(char quiet)
...
         } else if (ConfGetBool("stream.inline", &inl) == 1) {
-            stream_inline = inl;
+            stream_inline = PKT_MODE_IPS;
```
glongo: Also here PKT_MODE_IDS is not set in any case.

src/source-af-packet.c
inliniac: I think the capture method has to tell the rest of the engine what mode it is in. So NFQ tells us we are inline, and AF_PACKET should tell us a packet is either IDS or IPS. It knows this, and the rest of the engine has no way of figuring this out without the help of the capture method.

glongo: Exactly, PKT_MODE_AUTO means it can be either IDS or IPS, but I think stream.inline can't be set to AUTO.
Otherwise, another way to set the pkt mode could be looking at p->afp_v.copy_mode.

inliniac: On 12-04-16 08:54, Giuseppe Longo wrote:
> Exactly, PKT_MODE_AUTO means it can be either IDS or IPS, but I think
> stream.inline can't be set to AUTO.
> Otherwise, another way to set the pkt mode could be looking at
> p->afp_v.copy_mode.

The capture method will have to set IDS or IPS on a packet, nothing
else. Who else can know what the mode of a packet is?

PR #1858, opened 2016-02-17
unix socket stats v0

inliniac:
Attempt at getting a new dump-counters working. Feedback welcome.

PR #1859, opened 2016-02-17
unix socket stats v1

inliniac:
Attempt at getting a new dump-counters working. Feedback welcome.

As #1858, fixes:
- comment by @regit 
- dump-counters in case no stats loggers are active

### DIFF
--- a/src/counters.c
+++ b/src/counters.c
@@ -36,7 +36,10 @@
 #include "util-privs.h"
 #include "util-signal.h"
 #include "unix-manager.h"
+
 #include "output.h"
+#include "output-stats.h"
+#include "output-json-stats.h"
 
 /* Time interval for syncing the local counters with the global ones */
 #define STATS_WUT_TTS 3
@@ -101,6 +104,7 @@ void StatsReleaseCounters(StatsCounter *head);
 /** stats table is filled each interval and passed to the
  *  loggers. Initialized at first use. */
 static StatsTable stats_table = { NULL, NULL, 0, 0, 0, {0 , 0}};
+static SCMutex stats_table_mutex = SCMUTEX_INITIALIZER;
 
 static uint16_t counters_global_id = 0;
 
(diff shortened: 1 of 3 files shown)


inliniac:
@regit this PR just does the 'global' stats, not the per thread ones. Any thoughts on how to expose those? Can do different command of course, 'dump-thread-counters' or something.

inliniac:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/350
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/353

regit:
I'm good with a specific command like 'dump-thread-counters'. Most people will be good with the total command.

duanehoward:
fwiw I do use the per thread stats, and collect them currently. It's usually helpful in identifying issues where one thread gets pegged and starts dropping, etc. It would be nice to have some sort of guidelines around this, as I was planning to upstream our collectd module at some point, but if the output of the command keeps changing the logic will get really complex (version check, then parse different things, etc.) A per thread command is probably fine too, though I personally preferred the 'dump everything and let the person/tool parsing the json decide what they want to use'

regit:
@duanehoward that make sense. @inliniac maybe the 'dump-counters' could stay as it was in 2.0 being per thread and we could have 'dump-global-counters' to get the global one.

inliniac:
Replaced by #1861

inliniac:
The reason I'm against restoring the 2.0 output is that it was pretty much broken. It had a bad and confusing mix of per thread and global stats. Getting rid of these problems (which also affected the regular stats.log) was one of the reasons for the redesign.

I understand it makes the upgrade painful, which is something I try to avoid normally.

The new structure is much better and shouldn't change in major ways in the future.

duanehoward:
I completely understand that logic Victor, and am supportive of it. I think it'll require some strange hacky bits on our side for a bit, but if we don't see major changes to this for a while that'll ease the process of cleaning up my hacks to get a supportable module for 3.0 opened up =)

inliniac:
@duanehoward the latest iteration is at #1862, it contains both totals (accumulated stats) and per thread stats.

duanehoward:
Yep, I've pulled that down, and will be using this in testing this week or next... My calendar has gotten nuts though. sigh.

PR #1863, opened 2016-02-18
build-info: workaround special _FORTIFY_SOURCE defines

norg:
This works around the issue with distributions defining _FORTIFY_SOURCE by default, like Gentoo does (#define _FORTIFY_SOURCE=((defined __OPTIMIZE__ && __OPTIMIZE__ > 0) ? 2 : 0)) and resolves the issue https://redmine.openinfosecfoundation.org/issues/1676

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/12
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/12

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -779,8 +779,12 @@ void SCPrintBuildInfo(void)
 #if __SSP_ALL__ == 2
     printf("compiled with -fstack-protector-all\n");
 #endif
-#ifdef _FORTIFY_SOURCE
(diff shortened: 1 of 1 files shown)


inliniac:
The explanation really belongs in a code comment in this case. This is one of those things that someone might 'optimize' after looking at the code.

In general, explanations should not just be in pull requests. Make them part of commit messages or code comments or both, whatever makes most sense in specific cases.

Can you do a new PR with the explanation as a code comment?

PR #1864, opened 2016-02-19
output-json-dns: add logging of NS answer record content.

decanio:
Added logging of rdata portion of NS DNS answer records.  These had been empty leading to what looked like duplicate json logs without the meaningful part of the data.

prscript results are here.
https://buildbot.openinfosecfoundation.org/builders/decanio/builds/12
https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/12


inliniac:
Merged through #1868, thanks Tom!

PR #1866, opened 2016-02-21
configure: warn if libpcre 8.35 is used

norg:
To warn (debian) user with libpcre 8.35 we should add at least some warning that with taht libpcre versions segfaults can occur, see https://redmine.openinfosecfoundation.org/issues/1693

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/13
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/13

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -479,6 +479,19 @@
         exit 1
     fi
 
+    # libpcre 8.35 (especially on debian) has a known issue that results in segfaults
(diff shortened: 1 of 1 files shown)


inliniac:
Merged through #1868, thanks Andreas!

PR #1870, opened 2016-02-23
nfqueue: fix wrong return value check in error cases

norg:
The check for the return value was wrong, we have 0 for success and 1 (and 2) for the error cases like 
TM_ECODE_FAILED.

This fixes also https://redmine.openinfosecfoundation.org/issues/1411

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/14
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/14


inliniac:
More explanation in the PR than the commit message...

PR #1874, opened 2016-02-24
output-json-dns: dns output filtering.

decanio:
Add the capability to filter DNS eve-log output by query/answer and RR type.  Configured in suricata.yaml as follows:

        - dns:
            # control logging of queries and answers
            # default yes, no to disable
            query: yes     # enable logging of DNS queries
            answer: yes    # enable logging of DNS answers
            # control which RR tyeps are logged
            # all enabled if custom not specified
            #custom: [a, aaaa, cname, mx, ns, ptr, txt]

- PR decanio: https://buildbot.openinfosecfoundation.org/builders/decanio/builds/13
- PR decanio-pcap: https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/13

### DIFF
--- a/src/output-json-dns.c
+++ b/src/output-json-dns.c
@@ -56,6 +56,69 @@
  * TX id handling doesn't expect it */
 #define QUERY 0
 
+#define LOG_QUERIES   (1<<0)
+#define LOG_ANSWERS  (1<<1)
+
+#define LOG_A         (1<<2)
+#define LOG_NS        (1<<3)
+#define LOG_CNAME     (1<<4)
+#define LOG_SOA       (1<<5)
+#define LOG_PTR       (1<<6)
+#define LOG_MX        (1<<7)
+#define LOG_TXT       (1<<8)
+#define LOG_AAAA      (1<<9)
+#define LOG_SRV       (1<<10)
+#define LOG_NAPTR     (1<<11)
+#define LOG_DS        (1<<12)
+#define LOG_RRSIG     (1<<13)
+#define LOG_NSEC      (1<<14)
+#define LOG_NSEC3     (1<<15)
+#define LOG_TKEY      (1<<16)
+#define LOG_TSIG      (1<<17)
+
+#define LOG_ALL_RRTYPES (~(LOG_QUERIES|LOG_ANSWERS))
+
+typedef enum {
+    DNS_RRTYPE_A = 0,
+    DNS_RRTYPE_NS,
+    DNS_RRTYPE_CNAME,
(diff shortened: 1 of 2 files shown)


src/output-json-dns.c
```diff
@@ -69,6 +132,46 @@ typedef struct LogDnsLogThread_ {
     MemBuffer *buffer;
 } LogDnsLogThread;
 
+static int DNSRRTypeEnabled(uint16_t type, uint32_t flags)
```
inliniac: would it make sense to start this func with a 'if (likely(flags == ~0)) return 1;' ? The common & default case will be that the filtering is disabled.

mcm:
This functionality would be awesome, are there any plans to include it in an upcoming release? It would also be awesome to have the option to log all of the answers in a single record... for very large sensors we're seeing DNS logging account for a huge part of the daily log size, and I think it would be far more efficient if we had the option to log them as a single event - that is, a single "type=answer" record, with an array of answers rather than just one.

I'd be happy to provide testing (and feedback) on this stuff if you need some.

decanio:
I owe Victor another PR.  We discovered in testing that there is an issue with this when the DNS logs are split out from the usual eve-log stream.  I've already got a fix that is known to work. I just need to get it to Victor.

decanio:
In addition I've got code lying around here someplace that combines each DNS transaction into a single log.  The answers are placed in an array.  We implemented DNS logs this was originally, but moved away from it as Kibana didn't play well with arrays, however there are a lot of other use cases that could make good use of this format.

mcm:
Awesome! I'd be happy to test these things out if you need, we use Splunk
and we've got a customer where the DNS logs account for more than 2/3 of
the suricata logs, bytes-wise, and being able to combine those would be
phenomenal.

On Thu, May 19, 2016 at 11:04 AM, Tom DeCanio <notifications@github.com>
wrote:

> In addition I've got code lying around here someplace that combines each
> DNS transaction into a single log. The answers are placed in an array. We
> implemented DNS logs this was originally, but moved away from it as Kibana
> didn't play well with arrays, however there are a lot of other use cases
> that could make good use of this format.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/inliniac/suricata/pull/1874#issuecomment-220352453>
>



-- 
Steve McMaster
Security Operations Team Lead
Hurricane Labs, LLC
Direct: 216.923.1330 x108

inliniac:
Hoping for a new version of this :)

PR #1875, opened 2016-02-25
output-json-alert: fix segfault when alerts separated out from eve-log.

decanio:
Fix a segfault encountered when json alerts output is separated from the rest of the eve-log output.

- PR decanio: https://buildbot.openinfosecfoundation.org/builders/decanio/builds/14
- PR decanio-pcap: https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/14


inliniac:
Merged through #1883, thanks Tom!

PR #1879, opened 2016-02-25
Remove Modbus compiler warnings on Fedora 23 when compiling 3.0RC3 fr…

daviddiallo:
All variables are initialized thanks to ModbusExtractUint8 or ModbusExtractUint16 function that extracts 8bits or 16bits data from pointer the received input data. In case of extracting error (because of length), ModbusExtractUint8 or ModbusExtractUint16 returns an error that is managed by the caller function.

All variables are now initialized to zero when they are declared. It does not change anything functionally but it will remove Modbus warnings.

### DIFF
--- a/src/app-layer-modbus.c
+++ b/src/app-layer-modbus.c
@@ -490,7 +490,7 @@ static void ModbusExceptionResponse(ModbusTransaction   *tx,
                                     uint16_t            *offset)
 {
     SCEnter();
-    uint8_t exception;
+    uint8_t exception = 0;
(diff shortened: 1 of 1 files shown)


inliniac:
Couple of things:
1. please rebase to master before you submit
2. the commit message is a bit weird. I would suggest something like: "modbus: fix compiler warnings with -Wmaybe-uninitialized"
3. PR contains more explanation than commit message, please add the explanation to the body of the commit msg.

PR #1881, opened 2016-02-25
modbus: fix compiler uninitialized warnings with -Wmaybe-uninitialized

daviddiallo:
All variables are initialized thanks to ModbusExtractUint8 or ModbusExtractUint16
function that extracts 8bits or 16bits data from pointer the received input data.
In case of extracting error (because of length), ModbusExtractUint8 or
ModbusExtractUint16 returns an error that is managed by the caller function.

All variables are now initialized to zero when they are declared. It does not
change anything functionally but it removes Modbus warnings.

Ticket: #1670
Last PR: #1879


inliniac:
Merged through #1883, thanks David!

PR #1882, opened 2016-02-25
nfqueue: fix wrong return value check in error cases

norg:
The check for the return value was wrong, we have 0 for success and 1
(and 2) for the error cases like TM_ECODE_FAILED, so we should quit
unless TM_ECODE_OK (0) is returned for NFQInitThread. This fixes #1870 

This is the updated version from https://github.com/inliniac/suricata/pull/1870

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/16
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/16


inliniac:
Merged through #1883, thanks Andreas!

PR #1884, opened 2016-02-28
rule-parsing: quick fix for rules with wrong double quotes

norg:
The stripping of leading and trailing "s has issues with rules like the
ones described in issue #1638 thus resulted in crashing the rule parser.
So for now this is a quick fix which approaches this issue directly by
stripping those "s correctly and handling error cases. A
more general approach should be done in the future.

This is a quick fix for https://redmine.openinfosecfoundation.org/issues/1638 since the general approach needs more testing and brainstorming, so we decided to do quick fix for 3.0.x and the big one for 3.1.

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/17
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/17

PR #1886, opened 2016-02-29
rule-parsing: quick fix for rules with wrong double quotes

norg:
The stripping of leading and trailing "s has issues with rules like the
ones described in issue 1638 thus resulted in crashing the rule parser.
So for now this is a quick fix which approaches this issue directly by
stripping those "s correctly and handling error cases. It also adds the
skip for leading spaces at the msg keyword and worksaround a possible
null pointer dereference (that should never occur though).
A more general approach should be done in the future.

This is a quick fix for https://redmine.openinfosecfoundation.org/issues/1638 and included the comments from the former PR #1884 and IRC suggestions.

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/19
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/19

### DIFF
--- a/src/detect-depth.c
+++ b/src/detect-depth.c
@@ -62,12 +62,14 @@ static int DetectDepthSetup (DetectEngineCtx *de_ctx, Signature *s, char *depths
     SigMatch *pm = NULL;
     int ret = -1;
 
-    /* strip "'s */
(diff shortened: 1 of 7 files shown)


inliniac:
Merged into master, thanks Andreas!

PR #1888, opened 2016-03-01
Remove unused cpu affinity config settings in suricata.yaml

pevma:


inliniac:
I think it would be good to combine changes to the affinity sets with the work in #1783. We can perhaps use the new unified thread names for the affinity sets. @zrasmor any interest in looking at that?

pevma:
Can be interesting - let me know when ready if you need QA-ing/testing - I have plenty of systems to do so.

zrasmor:
@inliniac I will take a look and get back to you shortly

inliniac:
Needs to be part of a more integrated approach wrt the affinity settings.

PR #1902, opened 2016-03-04

gozzy:
Reworked #1832

Suricata crashed when variable (either address or port) referred to itself. For instance, if one declared HOME_NET as !$HOME_NET or created a looped chain of variables.

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -45,6 +45,7 @@
 
 #include "util-debug.h"
 #include "util-print.h"
+#include "util-var.h"
 
 /* prototypes */
 void DetectAddressPrint(DetectAddress *);
@@ -915,7 +916,7 @@ error:
  */
 static int DetectAddressParse2(const DetectEngineCtx *de_ctx,
         DetectAddressHead *gh, DetectAddressHead *ghn,
-        char *s, int negate)
+        char *s, int negate, ResolvedVariablesList *var_list)
 {
     size_t x = 0;
     size_t u = 0;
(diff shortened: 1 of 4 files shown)


src/detect-engine-address.c
```diff
@@ -1453,7 +1477,8 @@ int DetectAddressParse(const DetectEngineCtx *de_ctx,
         goto error;
     }
 
-    r = DetectAddressParse2(de_ctx, gh, ghn, str, /* start with negate no */0);
+    printf("CHECK PORTS\n");
```
inliniac: debug left over

src/detect-engine-port.c
```diff
@@ -1037,6 +1038,13 @@ static int DetectPortParseDo(const DetectEngineCtx *de_ctx,
     char *rule_var_port = NULL;
     int r = 0;
 
+    printf("port var %s\n", s);
```
inliniac: debug left over

src/util-var.h
```diff
@@ -57,9 +57,21 @@ typedef struct XBit_ {
     uint32_t expire;
 } XBit;
 
+// A list of variables we try to resolve.
+// Helps to detect recursive declarations.
+typedef struct ResolvedVariable_ {
+    char var_name[255];
```
inliniac: super minor, but shouldn't this be 256? Due to alignment the struct will still have the same size now :)

inliniac:
I like this implementation a lot more, nice work.

src/detect-engine-address.c
gozzy: Damn it :)

src/util-var.h
gozzy: Thought about that, but the habit die hard :)

PR #1903, opened 2016-03-04

gozzy:
Reworked #1902 

Suricata crashed when variable (either address or port) referred to itself or if one created a looped chain of variables. For instance:

HOME_NET:     "!$EXTERNAL_NET"
EXTERNAL_NET: "!$HOME_NET"

Or:

Var1: "$Var2"
Var2: "$Var3"
Var3: "$Var1"

### DIFF
--- a/src/util-var.h
+++ b/src/util-var.h
@@ -57,9 +57,21 @@ typedef struct XBit_ {
     uint32_t expire;
 } XBit;
 
+// A list of variables we try to resolve while parsing configuration file.
+// Helps to detect recursive declarations.
+typedef struct ResolvedVariable_ {
+    char var_name[256];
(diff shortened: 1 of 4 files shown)


src/util-var.h
```diff
@@ -57,9 +57,21 @@ typedef struct XBit_ {
     uint32_t expire;
 } XBit;
 
+// A list of variables we try to resolve while parsing configuration file.
+// Helps to detect recursive declarations.
+typedef struct ResolvedVariable_ {
+    char var_name[256];
```
inliniac: :)

inliniac:
Merged through #1905, thanks a lot Alexander!

inliniac:
https://redmine.openinfosecfoundation.org/issues/1689

PR #1904, opened 2016-03-04
rule-parsing: remove unnecessary code

norg:
Those lines were included former to prevent possible null pointer
dereference but that won't happen anymore with the rest of the rework
done in the code. The code even results in a control flow issue reported
by coverity scan, so just remove it.

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/20
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/20


inliniac:
Merged through #1905, thanks Andreas!

PR #1909, opened 2016-03-07
configure: bypass libpcre 8.35 check

norg:
When --with-libpcre-libraries is used we skip the libpcre 8.35 check
since pkg-config might still point to the 8.35 version installed
although newer version was passed with --with-libpcre-libraries.

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/21
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/21


inliniac:
Merged into master, thanks Andreas!

PR #1911, opened 2016-03-08
app-layer-ssl: fix memleak

thus:
An invalid TLS client hello packet where the SNI extensions length field is set to a smaller value than the SNI server name length field, causes memory to be allocated multiple times.


thus:
Replaced by #1932

PR #1925, opened 2016-03-11
flow: record the flow hash for use as the output flow id - v0

jasonish:
Provides a consistent hash for a flow, as well as a better
distribution than using a memory address.

Potential solution to https://redmine.openinfosecfoundation.org/issues/1696

Prscript output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/196
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/199

### DIFF
--- a/src/flow.h
+++ b/src/flow.h
@@ -329,6 +329,9 @@ typedef struct Flow_
     /** flow queue id, used with autofp */
     SC_ATOMIC_DECLARE(int16_t, autofp_tmqh_flow_qid);
 
+    /** flow hash - the flow hash before hash table size mod. */
+    uint32_t flow_hash;
+
     /** flow tenant id, used to setup flow timeout and stream pseudo
      *  packets with the correct tenant id set */
     uint32_t tenant_id;
--- a/src/flow-hash.c
+++ b/src/flow-hash.c
@@ -218,9 +218,9 @@ typedef struct FlowHashKey6_ {
  *
  *  For ICMP we only consider UNREACHABLE errors atm.
  */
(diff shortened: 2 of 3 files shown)


inliniac:
I was thinking it would make more sense to store it in the Packet structure. That way we don't have to deref the flow pointer and it reduces the size of the Flow.

jasonish:
> I was thinking it would make more sense to store it in the Packet structure. That way we don't have to deref the flow pointer and it reduces the size of the Flow.

I was going to do that, but it was a more invasive patch as there is no packet passed into the flow/netflow loggers.  Will take another look at doing this.

inliniac:
Didn't think of the flow logger case. Your approach may make the most sense after all.

src/flow.h
```diff
@@ -329,6 +329,9 @@ typedef struct Flow_
     /** flow queue id, used with autofp */
     SC_ATOMIC_DECLARE(int16_t, autofp_tmqh_flow_qid);
 
+    /** flow hash - the flow hash before hash table size mod. */
+    uint32_t flow_hash;
```
inliniac: Since this is static after initialization, I think it would be better to move it up in the struct a bit. Comments denote a 'header' part which contains all fields that don't change after flow creation.

inliniac: it may actually be a good idea to move the startts field up there as well, I think that would make the header fit perfectly into it's own cache line.

jasonish:
Closed. See https://github.com/inliniac/suricata/pull/1953

PR #1930, opened 2016-03-14

thus:

Changes:
* Fixed time zone handling

Fixed changes suggested in #1849 .


thus:
Replaced by #1979

PR #1932, opened 2016-03-14
app-layer-ssl: fix memleak

thus:

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -205,8 +205,31 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
                 switch (ext_type) {
                     case SSL_EXTENSION_SNI:
                     {
-                        /* skip sni_list_length and sni_type */
-                        input += 3;
+                        /* there must not be more than one extension of the same
+                           type (RFC5246 section 7.4.1.4) */
+                        if (ssl_state->curr_connp->sni) {
+                            SCLogDebug("Multiple SNI extensions");
+                            AppLayerDecoderEventsSetEvent(ssl_state->f,
+                                    TLS_DECODER_EVENT_INVALID_HANDSHAKE_MESSAGE);
+                            return -1;
+                        }
+
+                        /* skip sni_list_length */
+                        input += 2;
+
+                        if (!(HAS_SPACE(1)))
+                            goto end;
+
+                        uint8_t sni_type = *(input++);
+
+                        /* currently the only type allowed is host_name
+                           (RFC6066 section 3) */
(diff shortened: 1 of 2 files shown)


src/app-layer-ssl.c
```diff
@@ -205,8 +205,31 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
                     case SSL_EXTENSION_SNI:
                     {
-                        /* skip sni_list_length and sni_type */
-                        input += 3;
+                        /* there must not be more than one extension of the same
+                           type (RFC5246 section 7.4.1.4) */
+                        if (ssl_state->curr_connp->sni) {
+                            SCLogDebug("Multiple SNI extensions");
+                            AppLayerDecoderEventsSetEvent(ssl_state->f,
+                                    TLS_DECODER_EVENT_INVALID_HANDSHAKE_MESSAGE);
```
inliniac: Any objections to making this more specific? ..._MULTIPLE_SNI

src/app-layer-ssl.c
```diff
@@ -205,8 +205,31 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
...
+                            goto end;
+
+                        uint8_t sni_type = *(input++);
+
+                        /* currently the only type allowed is host_name
+                           (RFC6066 section 3) */
+                        if (sni_type != SSL_SNI_TYPE_HOST_NAME) {
+                            SCLogDebug("Unknown SNI type");
+                            AppLayerDecoderEventsSetEvent(ssl_state->f,
+                                    TLS_DECODER_EVENT_INVALID_HANDSHAKE_MESSAGE);
```
inliniac: Same here, something like _INVALID_SNI_TYPE perhaps?

src/app-layer-ssl.c
```diff
@@ -217,6 +240,16 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
                         if (!(HAS_SPACE(sni_len)))
                             goto end;
 
+                        /* host_name contains the fully qualified domain name,
+                           and should therefore be limited by the maximum domain
+                           name length */
+                        if (sni_len > 255) {
+                            SCLogDebug("SNI length >255");
+                            AppLayerDecoderEventsSetEvent(ssl_state->f,
+                                    TLS_DECODER_EVENT_INVALID_HANDSHAKE_MESSAGE);
```
inliniac: Something similar here.

src/app-layer-ssl.c
thus: No objections at all. I'll fix and submit a new PR.

thus:
Replaced by #1934

PR #1933, opened 2016-03-14
detect: denote the max detection list; fix issue 1674 - v2

jasonish:
Denotes the max detection list so that rule validation can
allow post-detection lists to come after base64_data, but
disallow detection lists to come after it.

Change from last PR:
- Use \test.
- Do not increase size of array.

Other comments: I think the only other way is to either list the the allowed or disallowed lists in the sanity check. Both have a maintenance cost moving forward but I believe this is the lower cost option.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1674

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/197
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/200

### DIFF
--- a/src/detect-base64-data.c
+++ b/src/detect-base64-data.c
@@ -186,6 +186,10 @@ end:
     return retval;
 }
 
+/**
+ * \test Test that the rule fails to load if the detection list is
+ *     changed after base64_data.
+ */
 static int DetectBase64DataSetupTest03(void)
 {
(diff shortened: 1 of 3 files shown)


inliniac:
Merged through #1943, thanks Jason!

PR #1934, opened 2016-03-15
app-layer-ssl: fix memleak

thus:


src/app-layer-ssl.c
```diff
@@ -65,6 +65,9 @@ SCEnumCharMap tls_decoder_event_table[ ] = {
     { "INVALID_HEARTBEAT_MESSAGE",   TLS_DECODER_EVENT_INVALID_HEARTBEAT },
     { "OVERFLOW_HEARTBEAT_MESSAGE",  TLS_DECODER_EVENT_OVERFLOW_HEARTBEAT },
     { "DATALEAK_HEARTBEAT_MISMATCH", TLS_DECODER_EVENT_DATALEAK_HEARTBEAT_MISMATCH },
+    { "MULTIPLE SNI EXTENSIONS",     TLS_DECODER_EVENT_MULTIPLE_SNI_EXTENSIONS },
```
inliniac: These should not contain spaces but underscores. They are used as 'tls.multiple_sni_extensions' in the app-layer-events rule keyword.

inliniac: See https://github.com/inliniac/suricata/blob/master/rules/tls-events.rules for how they are used.

inliniac: Can you also add rules to rules/tls-events.rules for each of the new events?

thus: Note to self: coffee first, then pull requests :)

Thanks for the comments. I'll fix!

thus:
Replaced by #1935

PR #1935, opened 2016-03-15
app-layer-ssl: fix memleak

thus:


rules/tls-events.rules
```diff
@@ -18,10 +18,12 @@ alert tls any any -> any any (msg:"SURICATA TLS certificate invalid length"; flo
...
 alert tls any any -> any any (msg:"SURICATA TLS overflow heartbeat encountered, possible exploit attempt (heartbleed)"; flow:established; app-layer-event:tls.overflow_heartbeat_message; flowint:tls.anomaly.count,+,1; classtype:protocol-command-decode; reference:cve,2014-0160; sid:2230012; rev:1;)
 alert tls any any -> any any (msg:"SURICATA TLS invalid heartbeat encountered, possible exploit attempt (heartbleed)"; flow:established; app-layer-event:tls.invalid_heartbeat_message; flowint:tls.anomaly.count,+,1; classtype:protocol-command-decode; reference:cve,2014-0160; sid:2230013; rev:1;)
 alert tls any any -> any any (msg:"SURICATA TLS invalid encrypted heartbeat encountered, possible exploit attempt (heartbleed)"; flow:established; app-layer-event:tls.dataleak_heartbeat_mismatch; flowint:tls.anomaly.count,+,1; classtype:protocol-command-decode; reference:cve,2014-0160; sid:2230014; rev:1;)
+alert tls any any -> any any (msg:"SURICATA TLS multiple SNI extensions"; flow:established; app-layer-event:tls.multiple_sni_extensions; flowint:tls.anomaly.count,+,1; classtype:protocol-command-decode; sid:2230016; rev:1;)
```
inliniac: Can SNI appear in both sides of the conversation, or only in the to_server direction?

thus: We are only looking at SNI in the client hello handshake, so it probably makes sense to limit the rule to traffic flowing towards the server (flow:established,to_server).

inliniac: Yeah, lets do that.

thus:
Replaced by #1936

PR #1939, opened 2016-03-15
Modbus update v2

glongo:
This patchset reworks commit 88d7c8d,
because it breaks a UT (ModbusParserTest08) in the previous version.

Last PR: https://github.com/inliniac/suricata/pull/1777


src/app-layer-modbus.c
```diff
@@ -598,6 +598,11 @@ static void ModbusParseReadResponse(ModbusTransaction   *tx,
             goto error;
     }
 
+    tx->data = SCMalloc(count);
```
inliniac: I see multiple places in the code where tx->data is alloc'd, all unconditionally. I worry about the case of overwriting existing allocations and thus memory leaks.

src/output-json-modbus.c
```diff
@@ -0,0 +1,355 @@
...
+    json_object_set_new(modbusjs, "function",
+                        json_string(ModbusGetFunctionName(modbustx->function))
+                       );
+    if (modbustx->function == MODBUS_FUNC_DIAGNOSTIC) {
+        json_object_set_new(modbusjs, "subfunction",
+                            json_string(ModbusGetSubFunctionName(modbustx->subFunction))
+                           );
+    }
+    if (modbustx->type & MODBUS_TYP_READ) {
+        json_t *read = json_object();
```
inliniac: read (and write below) are common C function names, please use another name here.

src/app-layer-modbus.c
```diff
@@ -208,6 +209,15 @@ int ModbusGetAlstateProgress(void *modbus_tx, uint8_t direction) {
         ((modbus->transaction_max - tx->tx_num) > request_flood))
         return 1;
 
+    /* Check if we have a hole: next transaction is over but this one not.
```
inliniac: what if you miss 2? Or is that not possible?

inliniac:
Commit " json-modbus: log modbus fields in alerts" is lying. It doesn't add anything to alerts.

src/app-layer-detect-proto.c
```diff
@@ -345,7 +345,9 @@ static AppProto AppLayerProtoDetectPPGetProto(Flow *f,
             SCLogDebug("toserver - Probing parser found for source port %"PRIu16, f->sp);
 
             /* found based on source port, so use sp registration */
-            pe2 = pp_port_sp->sp;
+            pe2 = pp_port_sp->dp;
+            StreamTcpPacketSwitchDir(NULL, p);
```
inliniac: this use of StreamTcpPacketSwitchDir doesn't make sense. It flips the direction flags for this one packet, but does nothing for the rest of the flow.

src/flow.h
```diff
@@ -43,67 +43,70 @@ typedef struct AppLayerParserState_ AppLayerParserState;
 /* per flow flags */
 
 /** At least on packet from the source address was seen */
-#define FLOW_TO_SRC_SEEN                  0x00000001
```
inliniac: This flow renumbering should really be in it's own commit.

src/app-layer-detect-proto.c
inliniac: I see the problem of the applayer recognizing that the flow's direction is reversed, but we'll need a better solution for dealing with that.

src/app-layer-modbus.c
regit: Right, we should iterate here.  @glongo can you give a look ?

PR #1941, opened 2016-03-16
Hyperscan MPM integration v2

jviiret:
(version 2; a small reworking to add fix printf format issues that Coccinelle warned about)


jviiret:
Superseded by #1945.

PR #1945, opened 2016-03-16
Hyperscan MPM integration v3

jviiret:
> (Version 3: fixes the "no scratch space prototype" issue when running with no MPM patterns reported on oisf-devel. The previous PR was #1941.)

Caseful/caseless matching is done by Hyperscan natively.

Offset and depth are also checked by Hyperscan using its extended parameter support (this meant that these parameters had to be passed through to the add functions in 87bd04a).

Hyperscan MPM structures are cached and deduped. We found during testing that some configurations generate many duplicate copies of the same MPM matcher, and this has a very significant memory cost. It might be that this can be addressed, or that deduplication can be hoisted up to apply generically to any MPM implementation.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -585,6 +585,36 @@
         AC_MSG_RESULT(no)
     fi
 
+  # libhs
+    enable_hyperscan="no"
+    AC_ARG_WITH(libhs_includes,
+            [  --with-libhs-includes=DIR  libhs include directory],
+            [with_libhs_includes="$withval"],[with_libhs_includes=no])
+    AC_ARG_WITH(libhs_libraries,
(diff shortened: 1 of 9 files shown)


configure.ac
```diff
@@ -585,6 +585,36 @@
         AC_MSG_RESULT(no)
     fi
 
+  # libhs
```
inliniac: I noticed that hyperscan ships with a pkg-config file. I think it would make sense to use that here?

jviiret: Quite right -- I'll add it in and prep v4 of this PR.

jviiret:
Superseded by #1949.

PR #1947, opened 2016-03-18
detect grouping v189

inliniac:
Rebase of #1895

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/390
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/393

PR #1949, opened 2016-03-20
Hyperscan MPM integration v4

jviiret:
> (Version 4: tries pkg-config to find Hyperscan at configure time first. The previous PR was #1945.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -585,6 +585,44 @@
         AC_MSG_RESULT(no)
     fi
 
+  # libhs
+    enable_hyperscan="no"
+
+    # Try pkg-config first:
+    PKG_CHECK_MODULES([libhs], libhs,, [with_pkgconfig_libhs=no])
+    if test "$with_pkgconfig_libhs" != "no"; then
+        CPPFLAGS="${CPPFLAGS} ${libhs_CFLAGS}"
+        LIBS="${LIBS} ${libhs_LIBS}"
+    fi
+
+    AC_ARG_WITH(libhs_includes,
+            [  --with-libhs-includes=DIR  libhs include directory],
+            [with_libhs_includes="$withval"],[with_libhs_includes=no])
+    AC_ARG_WITH(libhs_libraries,
+            [  --with-libhs-libraries=DIR    libhs library directory],
+            [with_libhs_libraries="$withval"],[with_libhs_libraries="no"])
+
+    if test "$with_libhs_includes" != "no"; then
+        CPPFLAGS="${CPPFLAGS} -I${with_libhs_includes}"
+    fi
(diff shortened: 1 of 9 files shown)


inliniac:
Is it expected that initialization can be very slow? 30 mins+
<pre>
[2286] 21/3/2016 -- 09:16:08 - (detect.c:525) <Info> (SigLoadSignatures) -- 49 rule files processed. 30758 rules successfully loaded, 12 rules failed
[2286] 21/3/2016 -- 09:16:10 - (detect.c:2991) <Info> (SigAddressPrepareStage1) -- 30766 signatures processed. 1008 are IP-only rules, 11220 are inspecting packet payload, 22219 inspect application layer, 100 are decoder event only
[2286] 21/3/2016 -- 09:16:10 - (detect.c:2994) <Info> (SigAddressPrepareStage1) -- building signature grouping structure, stage 1: preprocessing rules... complete
[2286] 21/3/2016 -- 09:16:11 - (detect.c:3627) <Info> (SigAddressPrepareStage2) -- building signature grouping structure, stage 2: building source address list... complete
[2286] 21/3/2016 -- 09:47:14 - (detect.c:4152) <Info> (SigAddressPrepareStage3) -- building signature grouping structure, stage 3: building destination address lists... complete
</pre>
The same config using the 'ac' mpm implementation loads in just under a minute.

jviiret:
Can you describe (or email) the config so I can reproduce this locally, if you're able to share it?

I think that there are a number of reasons for the difference here, and some of them are default policy decisions in Suricata that we can tweak:

* The AC matcher interprets 'sgh-mpm-context: auto' as meaning 'single', rather than 'full'. This means that it is building vastly fewer matchers. On my ET-Open test case, which builds in 190 sec using Hyperscan and 'sgh-mpm-context: full', switching it to 'single' improves compile time to 8 sec.

* Hyperscan is able to build optimized engines for two cases specifically that result in some more expensive compile-time analysis inside the library:
    1. Pattern sets containing patterns with offset/depth information;
    2. Small pattern sets, for which an alternative engine that scans small blocks efficiently can be built.

What do you think of the idea of adding Hyperscan to the list of MPM matchers that set sgh-mpm-context to 'single' if 'auto' is specified? Does it improve init time on the case you were testing?

inliniac:
Hah, now I feel stupid for not thinking of this ;)

Indeed forcing it to 'single' makes it load nicely in 14s instead of 30mins. I'm using the ETpro ruleset.

So yes, let's make 'single' the default. This whole part of the engine is getting a big overhaul in 3.1. See #1947 if you're interested. It should lead to much fewer rule groups.

jviiret:
OK, I have made the change that interprets 'auto' as single' for sgh-mpm-context. Superseded by v5 of this PR in #1955.

PR #1951, opened 2016-03-21
Ruleset stats v5

glongo:

Updates:
- Fixed free funcs and lost objects deletion
- Rename of the unix-socket command
- Add a new file "util-detect" with detection engine helper functions

Small note:
The OutputTenancyStats2Json could set an error message in jdata object, which will be displayed through unix-socket, but to indicate it as "error message", the TM_ECODE_FAILED value is needed.

Last PR: #1853 
Ticket: [#1585] (https://redmine.openinfosecfoundation.org/issues/1585)

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/104
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/103

### DIFF
--- a/src/detect.c
+++ b/src/detect.c
@@ -449,14 +449,12 @@ int SigLoadSignatures(DetectEngineCtx *de_ctx, char *sig_file, int sig_file_excl
 
     ConfNode *rule_files;
     ConfNode *file = NULL;
-    SigFileLoaderStat sig_stat;
+    SigFileLoaderStat *sig_stat = &de_ctx->sig_stat;
     int ret = 0;
     char *sfile = NULL;
     char varname[128] = "rule-files";
     int good_sigs = 0;
     int bad_sigs = 0;
-
-    memset(&sig_stat, 0, sizeof(SigFileLoaderStat));
 
     if (strlen(de_ctx->config_prefix) > 0) {
         snprintf(varname, sizeof(varname), "%s.rule-files",
@@ -481,7 +479,7 @@ int SigLoadSignatures(DetectEngineCtx *de_ctx, char *sig_file, int sig_file_excl
                 TAILQ_FOREACH(file, &rule_files->head, next) {
                     sfile = DetectLoadCompleteSigPath(de_ctx, file->val);
                     good_sigs = bad_sigs = 0;
-                    ret = ProcessSigFiles(de_ctx, sfile, &sig_stat, &good_sigs, &bad_sigs);
+                    ret = ProcessSigFiles(de_ctx, sfile, sig_stat, &good_sigs, &bad_sigs);
                     SCFree(sfile);
 
                     if (ret != 0 || good_sigs == 0) {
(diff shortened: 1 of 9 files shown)


src/unix-manager.c
```diff
@@ -702,18 +702,30 @@ TmEcode UnixManagerCaptureModeCommand(json_t *cmd,
     SCReturnInt(TM_ECODE_OK);
 }
 
-TmEcode UnixManagerReloadRules(json_t *cmd, json_t *server_msg, void *data)
+TmEcode UnixManagerReloadRulesWrapper(json_t *cmd, json_t *server_msg, void *data, int wait)
```
inliniac: As this function does not wait for the reload complete, it could be used to (attempt to) start another while the first is still active. DetectEngineReloadStart() will return -1 in this case, which should be handled.

It's something that should be handled in general I think, as we can also have mixes of SIGUSR2 and unix socket leading to this in master.

src/util-detect.c
```diff
@@ -0,0 +1,118 @@
...
+ * \retval 1 on success 0 on failure
+ */
+int SigStringAddSig(SigString *sig, const char *sig_file,
+                    const char *sig_str, int line)
+{
+    if (sig_file == NULL || sig_str == NULL) {
+        return 0;
+    }
+
+    sig->filename = SCStrdup(sig_file);
```
inliniac: so for every rule, we make a memory allocation for the filename, which for many users will be just one or a handful. Yet on 30k rules we alloc it 30k times... there must be a better way to do that.

inliniac: ah except this is about failed rules, there shouldn't be that many

src/detect.c
```diff
@@ -375,6 +376,10 @@ static int DetectLoadSigFile(DetectEngineCtx *de_ctx, char *sig_file,
                 EngineAnalysisRulesFailure(line, sig_file, lineno - multiline);
             }
             bad++;
+            if (!SigStringAppend(&de_ctx->sig_stat.failed_sigs, sig_file, line, (lineno - multiline))) {
+                SCLogError(SC_ERR_MEM_ALLOC, "Error adding sig \"%s\" from "
```
inliniac: wonder what the value is of providing the line no? We already logged the parse error with that info, right?

src/unix-manager.c
```diff
@@ -757,6 +757,52 @@ TmEcode UnixManagerRulesetStatsCommand(json_t *cmd,
...
+        SigString *sigs_str = list->sig_stat.failed_sigs;
+        while (sigs_str != NULL) {
+            json_t *jdata = json_object();
+            if (jdata == NULL) {
+                json_object_set_new(server_msg, "message", json_string("Unable to get the sig"));
+                json_object_clear(js_sigs_array);
+                goto error;
+            }
+
+            json_object_set_new(jdata, "id", json_integer(list->tenant_id));
```
inliniac: This id is confusing, I assumed that it would be rule id. So rename it to tenant_id or something similar.

PR #1955, opened 2016-03-22
Hyperscan MPM integration v5

jviiret:
> (Version 5: makes the HS matcher intepret 'sgh-mpm-context: auto' as 'single', as per the AC matchers, to keep init time tractable. The previous PR was #1949.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -585,6 +585,44 @@
         AC_MSG_RESULT(no)
     fi
 
+  # libhs
+    enable_hyperscan="no"
+
+    # Try pkg-config first:
+    PKG_CHECK_MODULES([libhs], libhs,, [with_pkgconfig_libhs=no])
+    if test "$with_pkgconfig_libhs" != "no"; then
+        CPPFLAGS="${CPPFLAGS} ${libhs_CFLAGS}"
+        LIBS="${LIBS} ${libhs_LIBS}"
+    fi
+
+    AC_ARG_WITH(libhs_includes,
+            [  --with-libhs-includes=DIR  libhs include directory],
+            [with_libhs_includes="$withval"],[with_libhs_includes=no])
+    AC_ARG_WITH(libhs_libraries,
+            [  --with-libhs-libraries=DIR    libhs library directory],
+            [with_libhs_libraries="$withval"],[with_libhs_libraries="no"])
+
+    if test "$with_libhs_includes" != "no"; then
+        CPPFLAGS="${CPPFLAGS} -I${with_libhs_includes}"
+    fi
+    AC_CHECK_HEADER(hs.h,HYPERSCAN="yes",HYPERSCAN="no")
+    if test "$HYPERSCAN" = "yes"; then
+        if test "$with_libhs_libraries" != "no"; then
+            LDFLAGS="${LDFLAGS}  -L${with_libhs_libraries}"
+        fi
+
+        AC_CHECK_LIB(hs,hs_compile,,HYPERSCAN="no")
+        enable_hyperscan="yes"
+        if test "$HYPERSCAN" = "no"; then
+            echo
+            echo "   Hyperscan headers are present, but link test failed."
+            echo "   Check that you have a shared library and C++ linkage available."
+            echo
+            enable_hyperscan="no"
+        fi
+    fi
+    AS_IF([test "x$enable_hyperscan" = "xyes"], [AC_DEFINE([BUILD_HYPERSCAN], [1], [Intel Hyperscan support enabled])])
+
   # libyaml
     AC_ARG_WITH(libyaml_includes,
             [  --with-libyaml-includes=DIR  libyaml include directory],
@@ -1896,6 +1934,7 @@ SURICATA_BUILD_CONF="Suricata Configuration:
   Non-bundled htp:                         ${enable_non_bundled_htp}
   Old barnyard2 support:                   ${enable_old_barnyard2}
   CUDA enabled:                            ${enable_cuda}
+  Hyperscan support:                       ${enable_hyperscan}
 
   Suricatasc install:                      ${enable_python}
 
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -372,6 +372,7 @@ util-mpm-ac-tile.c util-mpm-ac-tile.h \
 util-mpm-ac-tile-small.c \
 util-mpm-b2g.c util-mpm-b2g.h \
 util-mpm-b3g.c util-mpm-b3g.h \
+util-mpm-hs.c util-mpm-hs.h \
 util-mpm.c util-mpm.h \
 util-mpm-wumanber.c util-mpm-wumanber.h \
 util-optimize.h \
--- a/src/detect-engine-mpm.c
+++ b/src/detect-engine-mpm.c
@@ -1211,28 +1211,40 @@ static void PopulateMpmHelperAddPatternToPktCtx(MpmCtx *mpm_ctx,
                                                 Signature *s, uint8_t flags,
                                                 int chop)
 {
+    uint16_t pat_offset = cd->offset;
+    uint16_t pat_depth = cd->depth;
+
+    /* recompute offset/depth to cope with chop */
+    if (chop && (pat_depth || pat_offset)) {
+        pat_offset += cd->fp_chop_offset;
+        if (pat_depth) {
+            pat_depth -= cd->content_len;
+            pat_depth += cd->fp_chop_offset + cd->fp_chop_len;
+        }
+    }
+
     if (cd->flags & DETECT_CONTENT_NOCASE) {
         if (chop) {
             MpmAddPatternCI(mpm_ctx,
                             cd->content + cd->fp_chop_offset, cd->fp_chop_len,
-                            0, 0,
+                            pat_offset, pat_depth,
                             cd->id, s->num, flags);
         } else {
             MpmAddPatternCI(mpm_ctx,
                             cd->content, cd->content_len,
-                            0, 0,
+                            pat_offset, pat_depth,
                             cd->id, s->num, flags);
         }
     } else {
         if (chop) {
             MpmAddPatternCS(mpm_ctx,
                             cd->content + cd->fp_chop_offset, cd->fp_chop_len,
-                            0, 0,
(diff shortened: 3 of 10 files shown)


inliniac:
Address Sanitizer doesn't like the way patterns are added:

<pre>
[5829] 23/3/2016 -- 10:48:24 - (detect.c:4154) <Info> (SigAddressPrepareStage3) -- building signature grouping structure, stage 3: building destination address lists... complete
=================================================================
==5829==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020006bb834 at pc 0x00000113d19f bp 0x7ffd45476eb0 sp 0x7ffd45476ea8
READ of size 4 at 0x6020006bb834 thread T0 (Suricata-Main)
    #0 0x113d19e in hashlittle /home/victor/dev/suricata/src/util-hash-lookup3.c:325:14
    #1 0x127513c in SCHSPatternHash /home/victor/dev/suricata/src/util-mpm-hs.c:483:12
    #2 0x125ed04 in PatternDatabaseHash /home/victor/dev/suricata/src/util-mpm-hs.c:519:16
    #3 0x112e8a8 in HashTableLookup /home/victor/dev/suricata/src/util-hash.c:193:12
    #4 0x1256297 in SCHSPreparePatterns /home/victor/dev/suricata/src/util-mpm-hs.c:656:34
    #5 0x817da9 in SigGroupBuild /home/victor/dev/suricata/src/detect.c:4711:13
    #6 0x81495f in SigLoadSignatures /home/victor/dev/suricata/src/detect.c:540:9
    #7 0xfe432d in LoadSignatures /home/victor/dev/suricata/src/suricata.c:1979:9
    #8 0xfcd84e in main /home/victor/dev/suricata/src/suricata.c:2345:17
    #9 0x7fd6f327eec4 in __libc_start_main /build/eglibc-3GlaMS/eglibc-2.19/csu/libc-start.c:287
    #10 0x4e3d8c in _start (/home/victor/dev/suricata/src/suricata+0x4e3d8c)

0x6020006bb836 is located 0 bytes to the right of 6-byte region [0x6020006bb830,0x6020006bb836)
allocated by thread T0 (Suricata-Main) here:
    #0 0x4c670b in __interceptor_malloc (/home/victor/dev/suricata/src/suricata+0x4c670b)
    #1 0x12693f6 in SCHSAddPattern /home/victor/dev/suricata/src/util-mpm-hs.c:331:27
    #2 0x126c965 in SCHSAddPatternCS /home/victor/dev/suricata/src/util-mpm-hs.c:1018:12
    #3 0x128cae6 in MpmAddPatternCS /home/victor/dev/suricata/src/util-mpm.c:746:12
    #4 0xa1cc11 in PopulateMpmHelperAddPatternToPktCtx /home/victor/dev/suricata/src/detect-engine-mpm.c:1245:13
    #5 0xa11b31 in PopulateMpmAddPatternToMpm /home/victor/dev/suricata/src/detect-engine-mpm.c:1375:29
    #6 0xa0428c in PatternMatchPreparePopulateMpm /home/victor/dev/suricata/src/detect-engine-mpm.c:1935:9
    #7 0x9f9199 in PatternMatchPrepareGroup /home/victor/dev/suricata/src/detect-engine-mpm.c:2358:9
    #8 0x85b3d2 in BuildDestinationAddressHeadsWithBothPorts /home/victor/dev/suricata/src/detect.c:3962:37
    #9 0x85d2eb in SigAddressPrepareStage3 /home/victor/dev/suricata/src/detect.c:4064:13
    #10 0x817514 in SigGroupBuild /home/victor/dev/suricata/src/detect.c:4678:9
    #11 0x81495f in SigLoadSignatures /home/victor/dev/suricata/src/detect.c:540:9
    #12 0xfe432d in LoadSignatures /home/victor/dev/suricata/src/suricata.c:1979:9
    #13 0xfcd84e in main /home/victor/dev/suricata/src/suricata.c:2345:17
    #14 0x7fd6f327eec4 in __libc_start_main /build/eglibc-3GlaMS/eglibc-2.19/csu/libc-start.c:287

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/victor/dev/suricata/src/util-hash-lookup3.c:325 hashlittle
Shadow bytes around the buggy address:
  0x0c04800cf6b0: fa fa fa fa fa fa fa fa fa fa 00 fa fa fa 00 fa
  0x0c04800cf6c0: fa fa 00 fa fa fa 00 fa fa fa 04 fa fa fa 00 fa
  0x0c04800cf6d0: fa fa 04 fa fa fa 00 fa fa fa 04 fa fa fa 00 fa
  0x0c04800cf6e0: fa fa 04 fa fa fa 00 fa fa fa fa fa fa fa 06 fa
  0x0c04800cf6f0: fa fa fa fa fa fa 06 fa fa fa fa fa fa fa 06 fa
=>0x0c04800cf700: fa fa fa fa fa fa[06]fa fa fa 04 fa fa fa 00 06
  0x0c04800cf710: fa fa 04 fa fa fa fa fa fa fa fa fa fa fa 04 fa
  0x0c04800cf720: fa fa 00 02 fa fa 04 fa fa fa 04 fa fa fa 00 fa
  0x0c04800cf730: fa fa 04 fa fa fa 00 fa fa fa 00 fa fa fa fd fa
  0x0c04800cf740: fa fa 04 fa fa fa 04 fa fa fa 04 fa fa fa fd fa
  0x0c04800cf750: fa fa 04 fa fa fa 04 fa fa fa fa fa fa fa 04 fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Heap right redzone:      fb
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack partial redzone:   f4
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  ASan internal:           fe
==5829==ABORTING
</pre>

Suricata was compiled as follows:
<pre>
$ ac_cv_func_malloc_0_nonnull=yes ac_cv_func_realloc_0_nonnull=yes CC=clang-3.5 CFLAGS="-O0 -ggdb3 -Werror -Wchar-subscripts -fno-strict-aliasing -fstack-protector-all -fsanitize=address -fno-omit-frame-pointer -Wno-unused-parameter -Wno-unused-function" ./configure --prefix=/usr --sysconfdir=/etc --localstatedir=/var --disable-shared && make clean && make -j
</pre>

Then I ran it as:
<pre>
ASAN_OPTIONS="external_symbolizer_path=/usr/bin/llvm-symbolizer-3.5" ./src/suricata -c suricata.yaml -l /tmp/ -S ~/etrules/open-1458726337.rules -v --set mpm-algo=hs -T
</pre>
Where ~/etrules/open-1458726337.rules is today's ET open ruleset (https://rules.emergingthreatspro.com/open/suricata-edge/emerging-all.rules)

This is on Ubuntu 14.04 on x86_64. I used Hyperscan 4.0.1 and Boost 1.60 headers.

jviiret:
I can reproduce this here, and I've tracked it down.

This appears to be because I use [hashlittle()](https://github.com/jviiret/suricata/blob/f17ef5e46b5fbcebf0027604a5b8f5fb8e5f2219/src/util-hash-lookup3.c#L282) from Suricata's copy in src/util-hash-lookup3.c (the Jenkins hash) and that plays a bit fast-and-loose with memory, at least by ASan's definition.

If you define VALGRIND in that file (see the comment on lines [305-314](https://github.com/jviiret/suricata/blob/f17ef5e46b5fbcebf0027604a5b8f5fb8e5f2219/src/util-hash-lookup3.c#L305-314)), you'll see these errors go away.

It looks like I'm the only user of hashlittle() in the Suricata codebase, and I'm using it to hash the strings in the pre-compile pattern database to do deduplication, which is only really useful I think when sgh-mpm-context is 'full'. Anyone else who starts using that function on strings will trigger the same issue.

I could work around this by:

* Defining VALGRIND in util-hash-lookup3.c to use the slower/safer variant;
* Use atribute [no_sanitize_address](http://llvm.org/releases/3.5.0/tools/clang/docs/AttributeReference.html#no-sanitize-address-no-address-safety-analysis-gnu-no-address-safety-analysis-gnu-no-sanitize-address) (supported by both gcc and clang, it looks like) to instruct asan not to instrument hashlittle();
* Roll my own hash function in util-mpm-hs.c.

Which would you prefer?

inliniac:
I guess a 4th option would be to add padding to the memory chunks so that they are always multiples of 8. However due to the large number of vars passed to hashlittle this is not really practical.

I honestly don't like any of the options very much. I do prefer to keep the number of exceptions for QA to an absolute minimum, so I prefer the solution to be in the regular code path. So that would leave your third suggestion. I doubt that the hash function will be the most critical part of the performance btw, so perhaps we can simply make a copy of hashlittle to hashlittle_safe and use the VALGRIND path in there?

Thinking about it, we can likely simply "#define VALGRIND 1" before the hashlittle calls and "#undef VALGRIND" after. I was thinking that this might need a configure option or similar, but I guess we can just hardcode it like this.

So, at the end of my thinking out loud, I may prefer option 1 after all :)

inliniac:
Ok, after a bit of testing I think I prefer my hashlittle_safe idea. The defining of VALGRIND doesn't work the way I thought as we're talking about a function and not a macro. So then it'd be a global switch inside lookup3.c which I don't really like.

I tested with doing the copy, and making the VALGRIND path default in my hashlittle_safe. That fixes the ASAN issue.

jviiret:
Seems like a reasonable solution.

Shall I create a v6 then, with a separate (earlier) commit that adds hashlittle_safe() as per your test, and a Hyperscan MPM integration that uses that interface instead?

inliniac:
On 27-03-16 12:10, Justin Viiret wrote:
> Seems like a reasonable solution.
> 
> Shall I create a v6 then, with a separate (earlier) commit that adds
> hashlittle_safe() as per your test, and a Hyperscan MPM integration that
> uses that interface instead?

That would be great! Thanks.

jviiret:
Done in v6. This PR is superseded by #1965.

PR #1956, opened 2016-03-23
pcap-file: reset PCAP stats for each new PCAP

browner87:
PCAP stats not reset between files in Unix socket mode. Added a memset to the Global Init function to clear these stats.

### DIFF
--- a/src/source-pcap-file.c
+++ b/src/source-pcap-file.c
@@ -136,6 +136,7 @@ void TmModuleDecodePcapFileRegister (void)
 
 void PcapFileGlobalInit()
 {
+    memset(&pcap_g, 0x00, sizeof(pcap_g));
     SC_ATOMIC_INIT(pcap_g.invalid_checksums);
 }
 


src/source-pcap-file.c
```diff
@@ -136,6 +136,7 @@ void TmModuleDecodePcapFileRegister (void)
 
 void PcapFileGlobalInit()
 {
+    memset(&pcap_g, 0x00, sizeof(pcap_g));
```
regit: Good catch. There is also a call to memset on pcap_g in TmModuleReceivePcapFileRegister(). It would be better to remove it as it is now useless.

regit:
Thanks a lot for the PR. Can you update the commit message please. First line should not reference the issue but instead document what is done in the patch. This will help when grepping the git history. Something like 'pcap-file: $(FIX A PROBLEM)' would be perfect.

browner87:
As far as I know I can't update a commit message without force overwriting the existing commit. Also according to those rules I shouldn't add any new commits after making the pull request. So should I close the PR, remove the now-redundant line and give a better commit comment, and then open a new PR?

regit:
Hello,
Le 24 mars 2016 4:54 PM, browner87 <notifications@github.com> a écrit :
>
> As far as I know I can't update a commit message without force overwriting the existing commit. Also according to those rules I shouldn't add any new commits after making the pull request. So should I close the PR, remove the now-redundant line and give a better commit comment, and then open a new PR?
That's exactly what's needed.
BR,

PR #1957, opened 2016-03-23
detect reload on pfring in workers v1

inliniac:
When PF_RING is seeing no traffic, just using BreakLoop doesn't help us with reloads, as so packet is going through the system.

This PR makes the reload code first inject packets, then call BreakLoop, then wait. In the PF_RING code we 'catch' this BreakLoop and inject and empty packet into the engine. This allows for completion of the reload.

Should fix https://redmine.openinfosecfoundation.org/issues/1716 for PF_RING, however other capture methods will probably have the same issue.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/392
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/395


inliniac:
cc @regit @jasonish @gureedo

inliniac:
Replaced by #1961

PR #1958, opened 2016-03-23
Add "ippair" scheduler for autofp to split traffic based on source an…

cdwakelin:
…d destination IP only (not ports).

- This is useful when using the "xbits" feature to track events that occur between the same hosts but not necessarily the same flow (such as exploit kit landings/expoits/payloads)
- The disadvantage is that traffic may be balanced very unevenly between threads if some host pairs are much more frequently seen than others, so it may be only practicable for sandbox or pcap analysis
- not tested for IPv6

See https://redmine.openinfosecfoundation.org/issues/1661

### DIFF
--- a/src/tmqh-flow.c
+++ b/src/tmqh-flow.c
@@ -41,6 +41,7 @@
 
 Packet *TmqhInputFlow(ThreadVars *t);
 void TmqhOutputFlowHash(ThreadVars *t, Packet *p);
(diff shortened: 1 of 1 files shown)


inliniac:
Thanks Chris, I merged this through https://github.com/inliniac/suricata/pull/1968. I did have to fix the commit message formatting. Please have a look at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits for future contributions.

PR #1959, opened 2016-03-24
doxygen: extract information from static

regit:
Some internal (static) functions are documented but this did not show in doxygen generated doc. Fixing this to help understand internal from doxy doc.


inliniac:
Need to reconsider this after the last unittest update. See https://github.com/inliniac/suricata/pull/1987#issuecomment-207267000

PR #1960, opened 2016-03-24
source-pcap-file: fix stats not being reset

browner87:
Fix for redmine #1737. Resets PCAP stats before every 
file in Unix socket mode.


inliniac:
Merged through #1970, thanks!

PR #1961, opened 2016-03-24
 reload on pfring in workers v7

inliniac:
When PF_RING is seeing no traffic, just using BreakLoop doesn't help us with reloads, as so packet is going through the system.

This PR makes the reload code first inject packets, then call BreakLoop, then wait. In the PF_RING code we 'catch' this BreakLoop and inject and empty packet into the engine. This allows for completion of the reload.

Should fix https://redmine.openinfosecfoundation.org/issues/1716 for PF_RING, however other capture methods will probably have the same issue.

Changes since #1957:
- fix hanging at interrupted reload
- redo signal handling to avoid deadlocks (see commits)
- cleanups

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/393
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/396

### DIFF
--- a/src/detect-engine.c
+++ b/src/detect-engine.c
@@ -571,6 +571,34 @@ static void BreakCapture(void)
 }
 
 /** \internal
+ *  \brief inject a pseudo packet into each detect thread that doesn't use the
+ *         new det_ctx yet
+ */
+static void InjectPackets(ThreadVars **detect_tvs,
+                          DetectEngineThreadCtx **new_det_ctx,
+                          int no_of_detect_tvs)
+{
+    int i;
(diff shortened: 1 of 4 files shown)


inliniac:
@regit @gureedo I think it's likely that afpacket and netmap need something similar to https://github.com/inliniac/suricata/pull/1961/commits/ec35ee359df3d6962aa7bda32382e9fb5687887c. The idea is that we don't just interrupt the poll/recv/read func, but also make sure a (fake) packet is injected into the system so that the detect threads can complete the reload.

inliniac:
Replaced by #1964

PR #1276, opened 2014-12-21
[discussion] Feature 509. Add SHA1 and SHA256 checksum support for files

serializingme:
Need some brainstorming over this. Warning, not the cleanest code and an awful amount  of copy and paste.

PR #1963, opened 2016-03-27
Dev feat509 hashing v0.2

serializingme:
New version of #1276. Created a single file for filemd5, filesha1 and filesha256 keywords. If it doesn't work for you lads, I can always change to four different files  (one with common code, and the rest with the specific code of the hashing algorithms).


src/util-file.c
```diff
@@ -493,13 +529,25 @@ int FileAppendData(FileContainer *ffc, const uint8_t *data, uint32_t data_len)
 
     if (FileStoreNoStoreCheck(ffc->tail) == 1) {
 #ifdef HAVE_NSS
-        /* no storage but forced md5 */
+        /* no storage but forced md5, sha1 or/and sha256 */
         if (ffc->tail->md5_ctx) {
             if (ffc->tail->md5_ctx)
```
serializingme: I have replicated this bellow, but I'm unsure why it has two if statements with the same condition. Shouldn't this be only one if?

inliniac: Looks like it ya.

serializingme: Ok, fix will go in the next version of the pull request.

inliniac:
Replaced by #2025

PR #1967, opened 2016-03-29
Per direction TX logging - v4

jasonish:
Another cut on per-direction logging based on feedback on the previous PRs: #1794, #1795, #1807.

- correct request timestamps
- correct ordering

Changes from last PR:
- More compact DNS data structure.
- Doxygen docs on setting up per direction TX logging.

Prscript output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/201
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/204


jasonish:
See #1992.

PR #1973, opened 2016-03-31
Fix heap-buffer-overflows

thus:
Fix heap-buffer-overflows in app-layer-tls-handshake.c and util-decode-der.c.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/2
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/2


thus:
Replaced by #1975

PR #1974, opened 2016-03-31
util-decode-der: fix NULL dereference bug

thus:
Make sure that the length is not longer than the size of the buffer
provided.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/3
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/3


thus:
Replaced by #1975

PR #1979, opened 2016-04-05
app-layer-tls: validity dates from TLS certificate v7

thus:

New detection keywords (tls_notbefore and tls_notafter), added validity dates to extended JSON log and extended TLS log output, and new lua functions TlsGetCertNotBefore() and TlsGetCertNotAfter().


thus:
Changed lua function "TlsGetCertValidity" to "TlsGetCertNotBefore" and "TlsGetCertNotAfter" instead.

thus:
Replaced by #1981

PR #1982, opened 2016-04-05
tcp: reduce TCP options storage in packets

inliniac:
Until now, the TCP options would all be stored in the Packet structure.
The commonly used ones (wscale, ts, sack, sackok and mss*) then had a
pointer to the position in the option array. Overall this option array
was large. About 360 bytes on 64bit systems. Since no part of the engine
would every access this array other than through the common short cuts,
this was actually just wasteful.

This patch changes the approach. It stores just the common ones in the
packet. The rest is gone. This shrinks the packet structure with almost
300 bytes.

* even though mss wasn't actually used

https://redmine.openinfosecfoundation.org/issues/1037

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/406
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/411

### DIFF
--- a/src/decode-tcp.c
+++ b/src/decode-tcp.c
@@ -42,8 +42,16 @@
 #include "pkt-var.h"
 #include "host.h"
 
+#define SET_OPTS(dst, src) \
+    (dst).type = (src).type; \
+    (dst).len  = (src).len; \
+    (dst).data = (src).data
+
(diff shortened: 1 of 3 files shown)


inliniac:
Merged through #1989.

PR #1985, opened 2016-04-07
Support sending rejects via libnet when running under non-root.

gozzy:
Since version 1.1.6 libnet handles capabilities correctly.
So changing libnet's version checking a little bit should do the trick.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -969,13 +969,15 @@
 
     AC_MSG_CHECKING(for libnet.h version 1.1.x)
     if test "$LIBNET_INC_DIR" != ""; then
-        if eval "grep LIBNET_VERSION $LIBNET_INC_DIR/libnet.h | grep -v '1.[[12]]' >/dev/null"; then
-            AC_MSG_RESULT(no)
-            LIBNET_DETECT_FAIL="yes"
-            LIBNET_FAIL_WARN($libnet_dir)
-        else
-            AC_MSG_RESULT(yes)
-        fi
+        LIBNET_VER=`grep LIBNET_VERSION $LIBNET_INC_DIR/libnet.h | sed 's/[[^"]]*"\([[^"]]*\).*/\1/'`
+
+        AS_VERSION_COMPARE([LIBNET_VER], [1.1.0],
+            [
+                AC_MSG_RESULT(no)
+                LIBNET_DETECT_FAIL="yes"
+                LIBNET_FAIL_WARN($libnet_dir)
+            ],
+            [AC_MSG_RESULT(yes)], [AC_MSG_RESULT(yes)])
 
         #CentOS, Fedora, Ubuntu-LTS, Ubuntu all set defines to the same values. libnet-config seems
         #to have been depreciated but all distro's seem to include it as part of the package.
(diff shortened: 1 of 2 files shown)


inliniac:
Looking at https://github.com/sam-github/libnet/commit/671742244c20804c9e935326e3649dac3382f17a I see no alternative way to test it than the version number. So I think your patch makes sense.

inliniac:
This breaks CentOS 5
<pre>
Found libtoolize
Remember to add `AC_PROG_LIBTOOL' to `configure.ac'.
libtoolize: `config.guess' exists: use `--force' to overwrite
libtoolize: `config.sub' exists: use `--force' to overwrite
libtoolize: `ltmain.sh' exists: use `--force' to overwrite
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
autoreconf: configure.ac: tracing
autoreconf: configure.ac: subdirectory libhtp to autoreconf
autoreconf: Entering directory `libhtp'
autoreconf: running: libtoolize --copy --force
Putting files in AC_CONFIG_AUX_DIR, `..'.
autoreconf: running: /usr/bin/autoconf --force
autoreconf: running: /usr/bin/autoheader --force
autoreconf: running: automake --add-missing --copy --force-missing
Makefile.am: installing `./INSTALL'
autoreconf: Leaving directory `libhtp'
Remember to add `AC_PROG_LIBTOOL' to `configure.ac'.
configure.ac:56: error: possibly undefined macro: AC_MSG_RESULT
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:65: error: possibly undefined macro: AC_DEFINE
configure.ac:974: error: possibly undefined macro: AS_VERSION_COMPARE
autoreconf: /usr/bin/autoconf failed with exit status: 1
program finished with exit code 1
elapsedTime=15.877499
</pre>

gozzy:
I don't know how to fix it for such an old distros :( Maybe, new configure option should be added, but I don't like this solution.

inliniac:
Looks like it works to add
<pre>
m4_ifndef([AS_VERSION_COMPARE], [AC_DEFUN([AS_VERSION_COMPARE],[false])])
</pre>
to the top of the file (directly below the AM_SILENT_RULES line).

gozzy:
Ok, I'll add it shortly.

configure.ac
```diff
@@ -969,13 +969,15 @@
...
-        if eval "grep LIBNET_VERSION $LIBNET_INC_DIR/libnet.h | grep -v '1.[[12]]' >/dev/null"; then
-            AC_MSG_RESULT(no)
-            LIBNET_DETECT_FAIL="yes"
-            LIBNET_FAIL_WARN($libnet_dir)
-        else
-            AC_MSG_RESULT(yes)
-        fi
+        LIBNET_VER=`grep LIBNET_VERSION $LIBNET_INC_DIR/libnet.h | sed 's/[[^"]]*"\([[^"]]*\).*/\1/'`
+
+        AS_VERSION_COMPARE([LIBNET_VER], [1.1.0],
```
inliniac: I think it's best if this one is restored to the old check, since this is the generic 'do we have 1.1+' check.

inliniac:
Thanks! Please note that I edited my own comments a few times, not sure if you get notifications for that as well :)

configure.ac
gozzy: I'll try to handle this case.

PR #1986, opened 2016-04-07
Support sending rejects via libnet when running under non-root.

gozzy:
Replaces #1985.

Since version 1.1.6 libnet handles capabilities correctly.
So changing libnet's version checking a little bit should do the trick.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,6 @@
     AC_INIT(suricata, 3.1dev)
     m4_ifndef([AM_SILENT_RULES], [m4_define([AM_SILENT_RULES],[])])AM_SILENT_RULES([yes])
+    m4_ifndef([AS_VERSION_COMPARE], [AC_DEFUN([AS_VERSION_COMPARE],[false])])
     AC_CONFIG_HEADERS([config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -969,7 +970,9 @@
 
     AC_MSG_CHECKING(for libnet.h version 1.1.x)
     if test "$LIBNET_INC_DIR" != ""; then
-        if eval "grep LIBNET_VERSION $LIBNET_INC_DIR/libnet.h | grep -v '1.[[12]]' >/dev/null"; then
+        LIBNET_VER=`grep LIBNET_VERSION $LIBNET_INC_DIR/libnet.h | grep '1.[[12]]' | sed 's/[[^"]]*"\([[^"]]*\).*/\1/'`
```
gozzy: This one should preserve compatibility with old systems.

inliniac:
I made slight modification to the workaround here https://github.com/inliniac/suricata/pull/1997/commits/e27ad81a43c9780f0756ffba4eb5c2ed97d0bfb0. It removes the 'false'. More importantly, I wanted it to be in it's own commit so that 'git blame' makes it clear why this strange line even exists :)

inliniac:
Merged through #1997, thanks Alexander!

PR #1987, opened 2016-04-07
unit testing improvements - v2

jasonish:
Redmine issue: https://redmine.openinfosecfoundation.org/issues/1758
Previous PR: https://github.com/inliniac/suricata/pull/1984

- First we convert all tests to be 1 for pass, 0 for fail.
- Then we refactor the registration code to not take a success value, as 1 is pass, 2 is fail.
- Introduce new macros for unit testing: FAIL_IF, FAIL_IF_NOT, FAIL_IF_NULL, FAIL_IF_NOT_NULL, PASS_IF and PASS
- Convert conf.c to use new macros.
- Some other minor unit test updates to make use of macros.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/208
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/213


inliniac:
Wrt doxygen and the unittests framework. If I apply @regit #1959, it will document not just the framework, but _all_ the the tests as well. I think that would be a bit excessive. Maybe we can use 2 different defines for the tests and the framework.

inliniac:
Merged, thanks Jason!

jasonish:
> Wrt doxygen and the unittests framework. If I apply @regit #1959, it will document not just the framework, but all the the tests as well. I think that would be a bit excessive. Maybe we can use 2 different defines for the tests and the framework.

Two thoughts:
- Build the unit test support code no matter what. Only wrap the actual unit tests in UNITTESTS.
- Or as you suggests, keep UNITTESTS as is, and wrap the support code in UNITTESTS_SUPPORT?

PR #1992, opened 2016-04-08
Per direction TX logging - v5

jasonish:
Previous PR: https://github.com/inliniac/suricata/pull/1967

This PR is just a rebase.

- correct request timestamps
- correct ordering

Changes addressed in previous PR comments:
- More compact DNS data structure.
- Doxygen docs on setting up per direction TX logging.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/210
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/215

### DIFF
--- a/src/app-layer-dns-common.h
+++ b/src/app-layer-dns-common.h
@@ -158,11 +158,16 @@ typedef struct DNSAnswerEntry_ {
 typedef struct DNSTransaction_ {
     uint16_t tx_num;                                /**< internal: id */
     uint16_t tx_id;                                 /**< transaction id */
-    uint8_t replied;                                /**< bool indicating request is
-                                                         replied to. */
-    uint8_t reply_lost;
     uint8_t rcode;                                  /**< response code (e.g. "no error" / "no such name") */
-    uint8_t recursion_desired;                      /**< server said "recursion desired" */
+
+    uint8_t replied:1;          /**< Flag indicating request is replied to. */
+    uint8_t reply_lost:1;       /**< Flag indicating if reply was lost .*/
+    uint8_t recursion_desired:1; /**< Flag indicating server said
+                                  * "recursion desired". */
+    uint8_t request_logged:1; /**< Set after request has been logged. */
+    uint8_t response_logged:1; /**< Set after response has been
+                                * logged. */
+    uint8_t padding:3;
 
     TAILQ_HEAD(, DNSQueryEntry_) query_list;        /**< list for query/queries */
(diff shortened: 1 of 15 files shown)


src/app-layer-dns-common.h
```diff
@@ -172,6 +177,8 @@ typedef struct DNSTransaction_ {
 
     TAILQ_ENTRY(DNSTransaction_) next;
     DetectEngineState *de_state;
+
```
inliniac: white space issue

src/output-tx.c
```diff
@@ -137,47 +142,88 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
 
         if (!(AppLayerParserStateIssetFlag(f->alparser, APP_LAYER_PARSER_EOF)))
         {
-            int tx_progress = AppLayerParserGetStateProgress(p->proto, alproto,
-                                                             tx, FlowGetDisruptionFlags(f, STREAM_TOSERVER));
-            if (tx_progress < tx_progress_done_value_ts) {
-                SCLogDebug("progress not far enough, not logging");
-                break;
-            }
+                tx_progress = AppLayerParserGetStateProgress(p->proto, alproto,
```
inliniac: indent is off

src/output-tx.c
```diff
@@ -137,47 +142,88 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
...
             SCLogDebug("logger %p", logger);
-            if (logger->alproto == alproto) {
-                SCLogDebug("alproto match, logging tx_id %ju", tx_id);
+
+            if (logger->alproto != alproto) {
+                goto next;
+            }
+            SCLogDebug("alproto match, logging tx_id %ju", tx_id);
+
+            if (AppLayerParserSupportsTxLogState(p->proto, alproto)) {
```
inliniac: I'm doubting a bit about the readability of this function. Perhaps it would be good to split between the 2 types earlier, as I think the func is quite complex right now. Lots of local vars to track about the state for example. Maybe things would be cleaner if we have 2 versions, or at least 2 versions of the inner loop?

inliniac:
I'm seeing significant reduction in http logs with this branch.

jasonish:
> I'm seeing significant reduction in http logs with this branch.

Can't say I've seen significant loss, but in all my tests I'm consistently seeing 1 less.  Refactoring for 2 versions, for readability does gracefully deal with this though.

PR #1994, opened 2016-04-10
util-decode-der: fix hang detected by AFL

thus:
Fix hang that occurs when child length is zero, resulting in an endless loop.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/10
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/10


inliniac:
Merged through #1997, thanks Mats!

PR #1995, opened 2016-04-11
Lock hunting 2

regit:
When playing with perf top on a busy box, I've realized the lock and unlock function of pthread library were taking some time. By looking at the call graph, we could see that the logging condition functions were partly responsible of that. This PR is updating the code to cache some flow related information in the packet. This result in locking the flow to be less common in the condition functions.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/158
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/154


regit:
Never, never do a PR at 2:30am

PR #1996, opened 2016-04-11
Lock hunting 3

regit:
Update of #1995 fixing the commit on log drop (now 7b3dff0).

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/159
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/155

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -396,15 +396,12 @@ typedef struct Packet_
     uint8_t flowflags;
     /* coccinelle: Packet:flowflags:FLOW_PKT_ */
 
-    uint8_t cachedflags;
-    uint16_t alproto;
-
     /* Pkt Flags */
     uint32_t flags;
 
-    struct Flow_ *flow;
-
     struct timeval ts;
+
+    struct Flow_ *flow;
 
     union {
         /* nfq stuff */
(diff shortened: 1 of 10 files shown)


inliniac:
I think this needs a lot more documentation, both in pr/commits and in the code itself. It will need to contain some analysis of what it means for the autofp runmode, when flow engine and workers run in different threads as well.

src/decode.h
```diff
@@ -274,6 +274,10 @@ typedef struct PacketAlert_ {
 /** alert is in a tx, tx_id set */
 #define PACKET_ALERT_FLAG_TX            0x08
 
+#define PACKET_APPLAYER_LOGGED          (1<<0)
```
inliniac: I think these names are likely to confuse ppl. We also have PKT_* flags.

src/decode.h
```diff
@@ -546,6 +550,9 @@ typedef struct Packet_
      */
     struct PktPool_ *pool;
 
+    uint16_t alproto;
```
inliniac: Type AppProto would probably be better.

src/decode.h
```diff
@@ -546,6 +550,9 @@ typedef struct Packet_
...
+    uint8_t cachedflags;
```
inliniac: Something like 'cached_flow_flags'? Although it's not a pure cache/copy of the flow flags of course.

inliniac: Comment similar to /* coccinelle: Packet:flowflags:FLOW_PKT_ */ is missing.

PR #1998, opened 2016-04-12
Ruleset stats v6

glongo:
Updates:
-  unix-socket: add reload-rules-non-blocking command (5e4c4fe)
Handle the case if DetectEngineReloadStart returns -1, it simply just output a message.

- unix-manager: print failed rules (3133aa0)
Rename "id" string to "tenant_id"

Last PR: #1951 
Ticket: https://redmine.openinfosecfoundation.org/issues/1585

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/108
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/107

### DIFF
--- a/src/unix-manager.c
+++ b/src/unix-manager.c
@@ -762,6 +762,52 @@ TmEcode UnixManagerRulesetStatsCommand(json_t *cmd,
     json_object_set_new(server_msg, "message", jdata);
     SCReturnInt(retval);
 }
+
+TmEcode UnixManagerShowFailedRules(json_t *cmd,
+                                   json_t *server_msg, void *data)
+{
+    SCEnter();
+    DetectEngineCtx *de_ctx = DetectEngineGetCurrent();
+    /* Since we need to deference de_ctx, we don't want to lost it. */
+    DetectEngineCtx *list = de_ctx;
+    json_t *js_sigs_array = json_array();
+
+    if (js_sigs_array == NULL) {
+        json_object_set_new(server_msg, "message", json_string("Unable to get info"));
+        goto error;
+    }
+    while (list) {
+        SigString *sigs_str = list->sig_stat.failed_sigs;
+        while (sigs_str != NULL) {
+            json_t *jdata = json_object();
+            if (jdata == NULL) {
+                json_object_set_new(server_msg, "message", json_string("Unable to get the sig"));
+                json_object_clear(js_sigs_array);
+                goto error;
+            }
+
+            json_object_set_new(jdata, "tenant_id", json_integer(list->tenant_id));
+            json_object_set_new(jdata, "rule", json_string(sigs_str->sig_str));
+            json_object_set_new(jdata, "filename", json_string(sigs_str->filename));
+            json_object_set_new(jdata, "line", json_integer(sigs_str->line));
+            json_array_append_new(js_sigs_array, jdata);
+
+            sigs_str = sigs_str->next;
+        }
+        list = list->next;
+    }
+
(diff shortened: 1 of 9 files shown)


src/unix-manager.c
```diff
@@ -700,18 +702,112 @@ TmEcode UnixManagerCaptureModeCommand(json_t *cmd,
...
-    DetectEngineReloadStart();
+    int r = DetectEngineReloadStart();
 
-    while (DetectEngineReloadIsDone() == 0)
-        usleep(100);
+    if (wait) {
+        while (DetectEngineReloadIsDone() == 0)
+            usleep(100);
+    } else {
+        if (DetectEngineReloadIsDone() == 0 && r == -1) {
```
inliniac: can you explain this logic? Why is the DetectEngineReloadIsDone == 0 check added?

glongo: If the function is called by 'reload-rules-non-blocking' command, we won't be able to do a rule reload again because DetectEngineReloadIsDone() is not called (detect_sync.state never set to IDLE).
We should be in error case only if DetectEngineReloadIsDone() returns 0 and DetectEngineReloadStart returns -1 (live reload is already occurring)

inliniac: I don't follow. If the r == -1 we didn't start a new reload as one is in progress. That should be reason enough to return error, right?

glongo: When the reload is completed, you can't restart a new one again because detect_sync.state is not set IDLE. 
To see that try to remove DetectEngineReloadIsDone() from the else statement, so it should be called.

inliniac: On 12-04-16 14:33, Giuseppe Longo wrote:
> In src/unix-manager.c
> <https://github.com/inliniac/suricata/pull/1998#discussion_r59367085>:
> 
>>  
>> -    while (DetectEngineReloadIsDone() == 0)
>> -        usleep(100);
>> +    if (wait) {
>> +        while (DetectEngineReloadIsDone() == 0)
>> +            usleep(100);
>> +    } else {
>> +        if (DetectEngineReloadIsDone() == 0 && r == -1) {
> 
> When the reload is completed, you can't restart a new one again because
> detect_sync.state is not set IDLE.
> To see that try to remove DetectEngineReloadIsDone() from the else
> statement, so it should be called.

This function calls DetectEngineReloadStart(), which returns -1. This
means it didn't touch the detect_sync.state which was already set to RELOAD.

wait == 0, so we're not waiting. Why are we not just returning an error
message?

If the detect_sync.state is wrong somehow, that should be fixed in the
proper place, not here.

glongo: Ok that's right, but when we're not waiting detect_sync.state will stay to RELOAD and never set to IDLE (DetectEngineReloadIsDone).
It won't be possible to redo another rule reload.

inliniac: https://github.com/inliniac/suricata/blob/master/src/suricata.c#L2641 should make sure the status gets to 'DONE', but I can see it's not getting to IDLE by itself. I think we can probably remove the 'DONE' state completely and just use DetectEngineReloadIsDone to set it to IDLE.

Would need some good testing though.

src/unix-manager.c
```diff
@@ -700,18 +702,112 @@ TmEcode UnixManagerCaptureModeCommand(json_t *cmd,
...
+            json_object_set_new(server_msg, "message", json_string("Reload in progress"));
```
regit: "Reload already in progress" may be more accurate here

PR #1999, opened 2016-04-12
afl: add --afl-der=file and AFL PERSISTANT_MODE

thus:
Add support for fuzzing SSL/TLS certificates (DER) through --afl-der=file.

Add support for AFL PERSISTANT_MODE when Suricata is compiled with a supported compiler (only afl-clang-fast for now). This gives a ~10x performance boost when fuzzing.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/11
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/11


thus:
Replaced by #2002

PR #2000, opened 2016-04-12
runmode-pcap-file: don't force flow reassembly

thus:
Don't force flow reassembly when stopping Suricata when reading from PCAP file.

Forcing flow reassembly when Suricata stops, some times messes up app-layer parsing, because APP_LAYER_EOF is set too early on some sessions.

- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/12


thus:
This didn't work as intended. Closing.

PR #2002, opened 2016-04-12
afl: add --afl-der=file and AFL PERSISTANT_MODE

thus:
Add support for fuzzing SSL/TLS certificates (DER) through --afl-der=file.

Add support for AFL PERSISTANT_MODE when Suricata is compiled with a supported compiler (only afl-clang-fast for now). This gives a substantial performance boost when fuzzing.


inliniac:
Merged with some configure fixes, thanks Mats!

PR #1976, opened 2016-04-01
Dev app layer stats v3

glongo:
This patchset adds per app-layer stats in EVE json output, couting precisally the number
of transactions and flows handled by the supported protocol.

To manage the tx counter,  a new API GetTxCounter is added to app-layer-parser. For flows, the counters are stored into an array, and a single counter is incremented when the alproto is recognized.

The output in EVE is:
"app-layer":{
  "tx":{"http":8,"smtp":0,"dns_tcp":0,"dns_udp":0},
  "flow":{"http":12,"ftp":0,"smtp":0,"tls":6,"ssh":0,"imap":0,"msn":0,"smb":0,"dcerpc":0,"dns":24}
}

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1657
Last PR: #1809 

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/105
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/104

PR #2005, opened 2016-04-13
App layer stats v4

glongo:
This version of the patchset replaces atomic counters by stats API.

Last PR: #1976 

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/109
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/108

### DIFF
--- a/src/app-layer-dcerpc-udp.c
+++ b/src/app-layer-dcerpc-udp.c
@@ -1062,7 +1062,9 @@ int DCERPCUDPParserTest01(void)
 	StreamTcpInitConfig(TRUE);
 
 	SCMutexLock(&f.m);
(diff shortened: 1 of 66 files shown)


inliniac:
I wondering if we can completely hide the flow counter from the actual implementations. After registration we already know the name of the registered protocols, we know its AppProto id. The actual increment of the counter is done inside the app-layer API. So I think we can completely hide this part of the counters from the protocol implementations.

inliniac:
The tx counters should only be registered for protocols that are actually tx-aware. For protocols that are not, the counter counts something else. E.g. request count. It's not the same.

inliniac:
Please add some API documentation too.

inliniac:
Will still need some work, but getting a lot closer.

PR #2008, opened 2016-04-14
[preview] reassembly buffer v101

inliniac:
This PR changes the way HTTP bodies and the File API store their data. Instead of a list of fragments that is reassembled on demand, a single reassembled buffer is kept and used.

https://github.com/inliniac/suricata/commit/32b902ec4090291b9ed540943b6766e6e8fe1ed7 introduces the underlying API.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/414
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/419

### DIFF
--- a/src/util-streaming-buffer.c
+++ b/src/util-streaming-buffer.c
@@ -1,0 +1,689 @@
+/* Copyright (C) 2015-2016 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#include "suricata-common.h"
+#include "util-streaming-buffer.h"
+#include "util-unittest.h"
+#include "util-print.h"
+
+/**
+ * \file
+ *
+ * \author Victor Julien <victor@inliniac.net>
+ *
+ *  \brief Streaming Buffer API
+ */
+
+/* memory handling wrappers. If config doesn't define it's own set of
+ * functions, use the defaults */
+#define MALLOC(cfg, s) \
+    (cfg)->Malloc ? (cfg)->Malloc((s)) : SCMalloc((s))
+#define CALLOC(cfg, n, s) \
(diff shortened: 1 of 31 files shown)


inliniac:
@jasonish could you do a review of this work?

inliniac:
Btw, I'm also working on moving over the TCP reassembly to this API, but that is a lot more involved and is not ready yet.

src/util-streaming-buffer.c
```diff
@@ -0,0 +1,689 @@
...
+#include "util-print.h"
+
+/**
+ * \file
+ *
+ * \author Victor Julien <victor@inliniac.net>
+ *
+ *  \brief Streaming Buffer API
+ */
+
```
jasonish: My first thought might be a little silly.  I think "StreamBuffer" flows better.  This buffer isn't streaming itself, but tracks a stream, and its a little less typing in for a utility that I think will also be heavily used.

A little more usage doc as well.  For example, I think I could use this to replace my very simple buffer's I'm using in DNP3 (https://github.com/jasonish/suricata/blob/dnp3/src/app-layer-dnp3.c#L579), but I'm not 100% sure yet without reading all the code.

inliniac: Ack on the docs.

Wrt other use. It's mostly meant for cases where the data we're tracking is (potentially) too large to buffer completely. We're sliding through it. This we already do in HTTP, FILE and TCP currently, but with lots of fragmentation and reassembly overhead. This tries to streamline that.

src/util-streaming-buffer.c
```diff
@@ -0,0 +1,689 @@
...
+        memcpy(sb->buf + sb->buf_offset, data, data_len);
+        seg->stream_offset = sb->stream_offset + sb->buf_offset;
+        seg->segment_len = data_len;
+        sb->buf_offset += data_len;
+        return seg;
+    }
+    return NULL;
+}
+
+void StreamingBufferAppend(StreamingBuffer *sb, StreamingBufferSegment *seg,
```
jasonish: Maybe this should be StreamingBufferAppendSegment? While the one below can be just StreamBufferAppend instead of NoTrack - better describes the function and the arguments required.

Regarding my comment above on documentation: What do you see as the common use case? With or without segments.  For the DNP3 buffers, I'd probably not use the segments, unless there was a compelling reason to do so?

inliniac: I really only see TCP as a use case for segments, as there we need to track out of order inserts and data overlaps. In the FILE implementation I already got rid of the segments. In HTTP I kept them, IIRC because we have an API interacting with HTTP body chunks. But we can probably get rid of it there as well.

jasonish: Ok, so do you think this could also be used in DNP3? Where I'm really only buffering the case where the data provided to the app-layer only contains a partial message, or perhaps one or more mesages but ending with just a partial message?

This begins to make sense now, if this is mainly targetted for things like large files, and its perhaps not the best fit for the use case I thought it might handle as well.

inliniac:
Merged a newer version through #2091

PR #2013, opened 2016-04-15
app-layer-ssl: make tx aware

thus:

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/14
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/14

### DIFF
--- a/src/app-layer-ssl.h
+++ b/src/app-layer-ssl.h
@@ -26,6 +26,8 @@
 #ifndef __APP_LAYER_SSL_H__
 #define __APP_LAYER_SSL_H__
 
+#include "app-layer-protos.h"
+#include "app-layer-parser.h"
 #include "decode-events.h"
 #include "queue.h"
 
@@ -176,9 +178,17 @@ typedef struct SSLState_ {
 
     /* there might be a better place to store this*/
     uint16_t hb_record_len;
+
+    uint8_t finished;
+
+    DetectEngineState *de_state;
+
+    uint16_t events;
+    AppLayerDecoderEvents *decoder_events;
 } SSLState;
 
 void RegisterSSLParsers(void);
 void SSLParserRegisterTests(void);
+void SSLSetEvent(SSLState *ssl_state, uint8_t event);
 
 #endif /* __APP_LAYER_SSL_H__ */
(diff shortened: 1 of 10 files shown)


src/app-layer-ssl.h
```diff
@@ -176,9 +178,17 @@ typedef struct SSLState_ {
 
     /* there might be a better place to store this*/
     uint16_t hb_record_len;
+
+    uint8_t finished;
+
+    DetectEngineState *de_state;
+
+    uint16_t events;
```
inliniac: the layout here can be made more efficient, this will waste 6 bytes between the 2 pointers

inliniac:
Not sure about https://github.com/inliniac/suricata/pull/2013/commits/789d18c3f83fa441d48b85982375ee066526849f. I like the goal of it, but I think the added code (duplication) is a bit unfortunate. Perhaps a common util function would be useful instead?

thus:
Almost all the loggers use the same code for this, so it could profit from being a util function instead. I will drop it from this PR and take a look at it separately.

src/app-layer-ssl.h
inliniac: <pre>
struct SSLState_ {
public:

	Flow *                     f;                    /*     0     8 */
	uint32_t                   flags;                /*     8     4 */

	/* XXX 4 bytes hole, try to pack */

	SSLStateConnp *            curr_connp;           /*    16     8 */
	SSLStateConnp              client_connp;         /*    24   120 */
	/* --- cacheline 2 boundary (128 bytes) was 16 bytes ago --- */
	SSLStateConnp              server_connp;         /*   144   120 */
	/* --- cacheline 4 boundary (256 bytes) was 8 bytes ago --- */
	uint16_t                   hb_record_len;        /*   264     2 */
	uint8_t                    finished;             /*   266     1 */

	/* XXX 5 bytes hole, try to pack */

	DetectEngineState *        de_state;             /*   272     8 */
	uint16_t                   events;               /*   280     2 */

	/* XXX 6 bytes hole, try to pack */

	AppLayerDecoderEvents *    decoder_events;       /*   288     8 */

	/* size: 296, cachelines: 5, members: 10 */
	/* sum members: 281, holes: 3, sum holes: 15 */
	/* last cacheline: 40 bytes */
};
</pre>

inliniac: Looking at this, I think 'finished' could be a flag instead. Then events and hb_record_len could fill the gap right after the flags.

inliniac:
The detection keywords have not yet been moved over. Do you have that planned?

thus:
Yes, moving the detection keywords is also something I'm planning on doing.

thus:
Replaced by #2016

PR #2014, opened 2016-04-16
Tpacket v3 4

regit:
Update of #1993 addressing comments. It adds two patches:
* 35194cd is forcing mmap to be used by default, it must now be explicitly disabled
* 386525e is a big improvement on the initial packet loss issue in af_packet.

Redmine issue: https://redmine.openinfosecfoundation.org/issues/681

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/160
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/156


regit:
Just realized that last commit was not accurate enough.

PR #2015, opened 2016-04-17
Tpacket v3 5

regit:
Update of #2014 fixing the initial packet loss fix.

Compare to #1993, it is addressing comments and adds two patches:
* 35194cd is forcing mmap to be used by default, it must now be explicitly disabled
* 7c6db20 is a big improvement on the initial packet loss issue in af_packet.

Redmine issue: https://redmine.openinfosecfoundation.org/issues/681

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/161
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/157

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -361,6 +361,12 @@ void *ParseAFPConfig(const char *iface)
         aconf->ring_size = max_pending_packets * 2 / aconf->threads;
     }
 
+    if ((ConfGetChildValueIntWithDefault(if_root, if_default, "block-size", &value)) == 1) {
+        aconf->block_size = value;
+    } else {
+        aconf->block_size = getpagesize() << AFP_BLOCK_SIZE_DEFAULT_ORDER;
+    }
+
     (void)ConfGetChildValueBoolWithDefault(if_root, if_default, "disable-promisc", (int *)&boolval);
     if (boolval) {
         SCLogInfo("Disabling promiscuous mode on iface %s",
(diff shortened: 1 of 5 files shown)


src/runmode-af-packet.c
```diff
@@ -216,11 +216,40 @@ void *ParseAFPConfig(const char *iface)
...
+            aconf->flags |= AFP_RING_MODE;
+        } else {
+            SCLogInfo("Disabling mmaped capture on iface %s",
+                    aconf->iface);
+        }
+    } else {
+        aconf->flags |= AFP_RING_MODE;
+    }
+
+    (void)ConfGetChildValueBoolWithDefault(if_root, if_default, "mmap-locked", (int *)&boolval);
```
inliniac: should we only enable this if AFP_RING_MODE is set?

regit: Option is only used on the mmap call so there is no harm of parsing it unconditionally.

inliniac: It will enable a info message that will confuse users: SCLogInfo("Enabling locked memory for mmap on iface %s",
They will think mmap is enabled even if it's not.

regit: Indeed, you made a good comment for l.232 ;) I will fix that.

inliniac: On 18-04-16 14:19, Eric Leblond wrote:
> Indeed, you made a good comment for l.232 ;) I will fix that.
> 

I'm getting fuzzy from all the fuzzing I guess ;)

regit:
@inliniac are you done with review ? I can push a updated PR if you want.

inliniac:
On 18-04-16 14:58, Eric Leblond wrote:
> @inliniac <https://github.com/inliniac> are you done with review ? I can
> push a updated PR if you want.

For now, yes. Will do some replay testing on your next PR.

PR #2016, opened 2016-04-18
app-layer-ssl: make tx aware - v2

thus:

Fixed changes suggested in #2013 

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/15
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/15


thus:
Replaced by #2038

PR #2023, opened 2016-04-24
Log debug messages fix v1

maxtors:


src/util-error.h
```diff
@@ -124,6 +124,12 @@ typedef enum {
     SC_ERR_UNIFIED_LOG_GENERIC,
     SC_ERR_HTTP_LOG_GENERIC,
     SC_ERR_TLS_LOG_GENERIC,
+    SC_ERR_STATS_LOG_GENERIC,
```
inliniac: these should be appended to keep the numbers stable

maxtors: Ah, sorry. Will fix.

maxtors:
Fixed feedback in #2023.

PR #2024, opened 2016-04-24
Log debug messages fix v2

maxtors:
Fixed error code placement per feedback from last PR #2023.


inliniac:
Merged through #2028, thanks!

PR #2025, opened 2016-04-24
Dev feat509 hashing v0.3

serializingme:
New version of #1963. Decided to split the code into four different files making it easier to read (especially if more hashing algorithms are added). Fixed the repeated `if` (`util-file.c`, `FileAppendData` function) and added the configuration options to Suricata. Also broken down the commits to make it easier to review.


src/detect.h
```diff
@@ -298,13 +298,15 @@ typedef struct DetectPort_ {
...
-#define FILE_SIG_NEED_FILE          0x01
-#define FILE_SIG_NEED_FILENAME      0x02
-#define FILE_SIG_NEED_TYPE          0x04
-#define FILE_SIG_NEED_MAGIC         0x08    /**< need the start of the file */
-#define FILE_SIG_NEED_FILECONTENT   0x10
-#define FILE_SIG_NEED_MD5           0x20
-#define FILE_SIG_NEED_SIZE          0x40
+#define FILE_SIG_NEED_FILE          0x0001
+#define FILE_SIG_NEED_FILENAME      0x0002
+#define FILE_SIG_NEED_TYPE          0x0004
```
inliniac: looks like this is unused, so it can be removed. Then the 8 bits are enough again :)

src/detect.h
```diff
@@ -364,7 +366,7 @@ typedef struct Signature_ {
 
     /** inline -- action */
     uint8_t action;
-    uint8_t file_flags;
+    uint16_t file_flags;
```
inliniac: and then this can be avoided

src/util-detect-file-hash.c
```diff
@@ -0,0 +1,377 @@
...
+ * \param de_ctx pointer to the Detection Engine Context
+ * \param s pointer to the Current Signature
+ * \param str pointer to the user provided "filemd5", "filesha1" or "filesha256" option
+ * \param type type of file hash to setup
+ *
+ * \retval 0 on Success
+ * \retval -1 on Failure
+ */
+int DetectFileHashSetup (DetectEngineCtx *de_ctx, Signature *s, char *str,
+        uint8_t type)
```
inliniac: probably safer to use 'int' here, as the number of supported keywords is growing. Would be a hard to find bug later ;)

src/util-error.h
```diff
@@ -237,8 +237,10 @@ typedef enum {
     SC_ERR_FLOW_INIT,
     SC_ERR_HOST_INIT,
     SC_ERR_MEM_BUFFER_API,
-    SC_ERR_INVALID_MD5,
+    SC_ERR_INVALID_HASH,
     SC_ERR_NO_MD5_SUPPORT,
+    SC_ERR_NO_SHA1_SUPPORT,
```
inliniac: please append new errors so the codes stay stable

inliniac:
The commit " Preparation for SHA-1 and SHA-256" is a bit weird. I don't see the point of adding empty files. A few commits later the files are filled, so it would make sense to just introduce them there as well.

inliniac:
Looks good overall. Could you address my comments, do a rebase and submit a new PR? Will push that through QA then.

Thanks!

inliniac:
I suspect rebase will have small conflicts because of https://github.com/inliniac/suricata/commit/8cd4405c213bc5f4344f2d5d37cf7a3dacbeeec5

src/detect.h
serializingme: Good catch! I should have double checked that :)

PR #2026, opened 2016-04-24
Removed duplicate include statements.

maxtors:
Looked through *.c files and removed duplicated include statements that i could find.


inliniac:
Merged through #2028, thanks!

PR #2031, opened 2016-04-25
Preview: DNP3 Support - v8

jasonish:
Replaces: #1844 
Remine ticket: https://redmine.openinfosecfoundation.org/issues/1745

This PR contains:
An application decoder for DNP3.
* Snort compatible DNP3 keywords:
  * dnp3_data
  * dnp3_func
  * dnp3_ind
  * dnp3_obj
* A DNP3 output module for EVE (JSON logging).
* DNP3 support for Lua detection.

Documentation for the DNP3 rule keywords can be found here:
OR in newer format:
http://jasonish-suricata.readthedocs.org/en/latest/rules/dnp3-keywords.html

Changes from the last PR include:
* Fixups from last PR comments.
* Many more objects supported.
* Fixups found during fuzzing.
* Other cleanups.

PRScript output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/215
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/220


scripts/dnp3-gen/dnp3-gen.py
```diff
@@ -0,0 +1,707 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * DO NOT EDIT. THIS FILE IS AUTO-GENERATED.
```
inliniac: can this include the exact command used to generate the file?

src/detect.h
```diff
@@ -127,6 +127,10 @@ enum DetectSigmatchListEnum {
 
     DETECT_SM_LIST_BASE64_DATA,
 
+    DETECT_SM_LIST_DNP3_DATA_MATCH,
+    DETECT_SM_LIST_DNP3_MATCH,
```
inliniac: can DETECT_SM_LIST_DNP3_MATCH and DETECT_SM_LIST_DNP3_LUA_MATCH be merged into a single list?

src/app-layer-dnp3-decoders.c
```diff
@@ -0,0 +1,1930 @@
+
```
inliniac: missing copyright header

inliniac:
Part of the code is generated by a script, but it's not done like that by the build system correct? Is that how it should be? When & how often would we need to run this script?

jasonish:
> Part of the code is generated by a script, but it's not done like that by the build system correct? Is that how it should be? When & how often would we need to run this script?

You would need to regenerate after editing scripts/dnp3-gen/dnp3-objects.yaml which you would edit when adding a new object, or fixing an old one.

You would also need to regenerate if fixing the code in one of the generated code sections, say refactoring.

I thought about doing it at part of the build, but I didn't based on some code gen I have worked with in the past with BSD kernels where the generation is done once, committed.  Then only the developer tweaking the definitions at a later date need worry about regenerating.  I think each method has its pros and cons, but I like the code to be readable after a checkout before build.

src/app-layer-dnp3-decoders.c
jasonish: Oops, that file shouldn't be there.

scripts/dnp3-gen/dnp3-gen.py
jasonish: Ok, all generated files, or segments have a comment, right now it looks like:

/* Code generated by:
 *     ./scripts/dnp3-gen/dnp3-gen.py
 */

src/detect.h
jasonish: Yes, with a bit of reworking it looks possible.

jasonish: Looks like this would involve converting the detection functions to AppLayerTxMatch variants. Is that what you had in mind?

jasonish: See https://github.com/inliniac/suricata/pull/2104/commits/dba95ad3e2a91fdd0bb068a5ff988d7100fa2184

PR #2018, opened 2016-04-18
Tpacket v3 6

regit:
Address comments made on #2015.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/162
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/158

PR #2032, opened 2016-04-26
Tpacket v3 9

regit:
A rework of #2018 addressing the comments. Only open point may be the fact that a new yaml is needed to have tpacket_v3 enabled by default.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/165
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/161

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -727,7 +727,6 @@ int AFPReadFromRing(AFPThreadVars *ptv)
 {
     Packet *p = NULL;
     union thdr h;
-    struct sockaddr_ll *from;
     uint8_t emergency_flush = 0;
     int read_pkts = 0;
     int loop_start = -1;
@@ -789,20 +788,11 @@ int AFPReadFromRing(AFPThreadVars *ptv)
          * function. */
         h.h2->tp_status |= TP_STATUS_USER_BUSY;
 
-        from = (void *)h.raw + TPACKET_ALIGN(ptv->tp_hdrlen);
-
         ptv->pkts++;
         ptv->bytes += h.h2->tp_len;
         p->livedev = ptv->livedev;
-
-        /* add forged header */
-        if (ptv->cooked) {
-            SllHdr * hdrp = (SllHdr *)ptv->data;
-            /* XXX this is minimalist, but this seems enough */
-            hdrp->sll_protocol = from->sll_protocol;
-        }
-
         p->datalink = ptv->datalink;
+
         if (h.h2->tp_len > h.h2->tp_snaplen) {
             SCLogDebug("Packet length (%d) > snaplen (%d), truncating",
                     h.h2->tp_len, h.h2->tp_snaplen);
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -1575,17 +1575,154 @@ static int AFPComputeRingParamsV3(AFPThreadVars *ptv)
     return 1;
 }
 
+static int AFPSetupRing(AFPThreadVars *ptv, char *devname)
+{
+    int val;
+    unsigned int len = sizeof(val), i;
+    unsigned int ring_buflen;
+    uint8_t * ring_buf;
+    int order;
+    int r;
+
+    if (ptv->flags & AFP_TPACKET_V3) {
+        val = TPACKET_V3;
+    } else {
+        val = TPACKET_V2;
+    }
+    if (getsockopt(ptv->socket, SOL_PACKET, PACKET_HDRLEN, &val, &len) < 0) {
+        if (errno == ENOPROTOOPT) {
+            if (ptv->flags & AFP_TPACKET_V3) {
+                SCLogError(SC_ERR_AFP_CREATE,
+                        "Too old kernel giving up (need 3.2 for TPACKET_V3)");
+            } else {
+                SCLogError(SC_ERR_AFP_CREATE,
+                        "Too old kernel giving up (need 2.6.27 at least)");
+            }
+        }
+        SCLogError(SC_ERR_AFP_CREATE, "Error when retrieving packet header len");
+        return AFP_FATAL_ERROR;
+    }
+
+    if (ptv->flags & AFP_TPACKET_V3) {
+        val = TPACKET_V3;
+    } else {
+        val = TPACKET_V2;
+    }
+    if (setsockopt(ptv->socket, SOL_PACKET, PACKET_VERSION, &val,
+                sizeof(val)) < 0) {
+        SCLogError(SC_ERR_AFP_CREATE,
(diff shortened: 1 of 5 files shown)


src/source-af-packet.c
```diff
@@ -94,6 +94,8 @@
 #include <sys/mman.h>
 #endif
 
+#include <linux/net_tstamp.h>
```
inliniac: think it would be good to add a AC_CHECK_HEADERS test for this and include it conditionally

src/source-af-packet.c
```diff
@@ -1211,13 +1413,16 @@ TmEcode ReceiveAFPLoop(ThreadVars *tv, void *data, void *slot)
...
-                r = AFPReadFromRing(ptv);
-            } else {
-                /* AFPRead will call TmThreadsSlotProcessPkt on read packets */
-                r = AFPRead(ptv);
-            }
+            r = AFPReadFunc(ptv);
             switch (r) {
+                case AFP_READ_OK:
+                    /* Trigger one dump of stats every second */
+                    TimeGet(&current_time);
```
inliniac: might be interesting to optimize this. TimeGet in live mode is essentially 'gettimeofday'. Each packet will have a timestamp, so maybe that can be used?

src/source-af-packet.c
```diff
@@ -1388,14 +1585,205 @@ frame size: TPACKET_ALIGN(snaplen + TPACKET_ALIGN(TPACKET_ALIGN(tp_hdrlen) + siz
...
 }
 
+#ifdef HAVE_TPACKET_V3
+static int AFPComputeRingParamsV3(AFPThreadVars *ptv)
+{
+    ptv->req3.tp_block_size = ptv->block_size;
+    ptv->req3.tp_frame_size = 2048;
+    int frames_per_block = ptv->req3.tp_block_size / ptv->req3.tp_frame_size;
+    if (frames_per_block == 0) {
+        SCLogInfo("frame size to big");
```
inliniac: https://github.com/inliniac/suricata/pull/2018#discussion_r60220823

src/source-af-packet.c
```diff
@@ -1388,14 +1585,205 @@ frame size: TPACKET_ALIGN(snaplen + TPACKET_ALIGN(TPACKET_ALIGN(tp_hdrlen) + siz
...
+                strerror(errno));
+    }
+#endif
+
+    /* Allocate RX ring */
+#ifdef HAVE_TPACKET_V3
+    if (ptv->flags & AFP_TPACKET_V3) {
+        if (AFPComputeRingParamsV3(ptv) != 1) {
+            SCLogInfo("Ring parameter are incorrect. Please correct the devel");
+            return -AFP_FATAL_ERROR;
```
inliniac: is the -AFP_FATAL_ERROR a typo? In other places it's used without the minus in front

src/source-af-packet.c
```diff
@@ -1388,14 +1585,205 @@ frame size: TPACKET_ALIGN(snaplen + TPACKET_ALIGN(TPACKET_ALIGN(tp_hdrlen) + siz
...
+#ifdef HAVE_TPACKET_V3
+    }
+#endif
+    mmap_flag = MAP_SHARED;
+    if (ptv->flags & AFP_MMAP_LOCKED)
+        mmap_flag |= MAP_LOCKED;
+    ring_buf = mmap(0, ring_buflen, PROT_READ|PROT_WRITE,
+            mmap_flag, ptv->socket, 0);
+    if (ring_buf == MAP_FAILED) {
+        SCLogError(SC_ERR_MEM_ALLOC, "Unable to mmap");
```
inliniac: mmap sets errno on error, so we can log a more detailed error with strerror(errno)

src/source-af-packet.h
```diff
@@ -56,6 +59,8 @@
 #define AFP_FILE_MAX_PKTS 256
 #define AFP_IFACE_NAME_LENGTH 48
 
+#define AFP_BLOCK_SIZE_DEFAULT_ORDER 3
```
inliniac: what does this mean? Comment would be nice here.

src/source-af-packet.h
```diff
@@ -65,6 +70,10 @@ typedef struct AFPIfaceConfig_
     int buffer_size;
     /* ring size in number of packets */
     int ring_size;
+    /* block size for tpacket_v3 */
+    int block_size;
+    /* block timeout for tpacket_v3 */
+    int block_timeout;
```
inliniac: timeout in what unit? would be good to have that in comment

src/source-af-packet.c
regit: Size does matter for @inliniac. Sorry to have missed that.

src/source-af-packet.c
regit: Yep, fixed thanks.

src/source-af-packet.h
regit: milliseconds, comment added.

src/source-af-packet.h
regit: Done.

src/source-af-packet.c
regit: Done.

src/source-af-packet.c
regit: I've just used the time() function instead as we only need second. It is said to be just a few cycles.

src/source-af-packet.c
regit: Done. configure.ac check was here https://github.com/inliniac/suricata/pull/2032/commits/d908891eea1db2dd8b6fe04a29624e83c94cf8d7#diff-67e997bcfdac55191033d57a16d1408aR1203 so I've just made inclusion conditional.

regit:
OK, closing and resubmitting a new one.

src/source-af-packet.c
inliniac: Do I read it correctly that with a block size of 4k and a frame size of 2k, AFPReadFromRingV3 will read 2 packets before returning? In that case any call to a time func may be expensive, esp if we can just keep the seconds from the packets we just processed in a local var.

regit: Yes, you read it correctly but we are not supposed to be in this case. Bigger block are supposed to be better 32 which should contains more than 32 packets is the default. time() is supposed to be 3 cycles (if http://stackoverflow.com/questions/6498972/faster-equivalent-of-gettimeofday is correct).

Anyway, easiest solution I see if you want to use get packet info is to add a time_t to PacketThreadVars and pass it to the timestamp of the last packet. If we don't do that on last packet then we have 32 useless affectation. Problem is to get the last packet without doing any test for each packet. If we do a test then we are far over the 3 cycles.

PR #2022, opened 2016-04-23
thread names v9

zrasmor:
#1898 Rebase per your request.

Let me know if there is anything else you need.

PR #2037, opened 2016-04-27
thread names v10

zrasmor:
Fixes issues in #2022


inliniac:
Merged through #2050, thanks!

PR #2038, opened 2016-04-27
app-layer-ssl: make tx aware - v4

thus:

Fixed changes suggested in #2013 

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/19
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/19

### DIFF
--- a/src/output-tx.c
+++ b/src/output-tx.c
@@ -28,6 +28,7 @@
 #include "output-tx.h"
 #include "app-layer.h"
 #include "app-layer-parser.h"
+#include "app-layer-ssl.h"
 #include "util-profiling.h"
 
 typedef struct OutputLoggerThreadStore_ {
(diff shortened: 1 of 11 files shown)


src/output-tx.c
```diff
@@ -119,12 +120,24 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
...
-    int tx_progress_done_value_tc =
-        AppLayerParserGetStateProgressCompletionStatus(p->proto, alproto,
-                                                       STREAM_TOCLIENT);
+
+    int tx_progress_done_value_ts = 0;
+    int tx_progress_done_value_tc = 0;
+
+    /* we want to log TLS once the handshake is done */
+    if (alproto == ALPROTO_TLS) {
+        tx_progress_done_value_ts = TLS_HANDSHAKE_DONE;
```
inliniac: I think we should make this a configurable setting when we register the logger. Register a value for when the tx is 'done enough' to log. Make sense?

thus: That makes sense :)

inliniac: On 27-04-16 11:42, Mats Klepsland wrote:
> In src/output-tx.c
> <https://github.com/inliniac/suricata/pull/2038#discussion_r61230156>:
> 
>> @@ -119,12 +120,24 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
>>  
>>      uint64_t total_txs = AppLayerParserGetTxCnt(p->proto, alproto, alstate);
>>      uint64_t tx_id = AppLayerParserGetTransactionLogId(f->alparser);
>> -    int tx_progress_done_value_ts =
>> -        AppLayerParserGetStateProgressCompletionStatus(p->proto, alproto,
>> -                                                       STREAM_TOSERVER);
>> -    int tx_progress_done_value_tc =
>> -        AppLayerParserGetStateProgressCompletionStatus(p->proto, alproto,
>> -                                                       STREAM_TOCLIENT);
>> +
>> +    int tx_progress_done_value_ts = 0;
>> +    int tx_progress_done_value_tc = 0;
>> +
>> +    /* we want to log TLS once the handshake is done */
>> +    if (alproto == ALPROTO_TLS) {
>> +        tx_progress_done_value_ts = TLS_HANDSHAKE_DONE;
> 
> That makes sense :)

I'm thinking that perhaps the cleanest way would be to have another reg
function like 'OutputRegisterTxModule' that either allows you to
register an additional callback for getting the 'ready to log'
'completion' value per direction.

It would likely result in another issue though. If we register 2 loggers
with different callbacks it may mess up the 'log id' tracking.

e.g. have a logger that logs as soon as http_uri is available and one
that logs when tx is complete. This would probably not work in the
suggested scheme.

thus:
Replaced by #2039

PR #2041, opened 2016-04-28
Added more DNS Resource Record Types.

maxtors:
Added some more DNS RRTYPE's. Some are obsolete and or experimental, but having detection for these could help. Saw a few "in the wild" that should be obsolete, but they are still in use.

### DIFF
--- a/src/app-layer-dns-common.c
+++ b/src/app-layer-dns-common.c
@@ -1016,11 +1016,11 @@ void DNSCreateTypeString(uint16_t type, char *str, size_t str_size)
         case DNS_RECORD_TYPE_NS:
             snprintf(str, str_size, "NS");
             break;
-        case DNS_RECORD_TYPE_AAAA:
-            snprintf(str, str_size, "AAAA");
+        case DNS_RECORD_TYPE_MD:
+            snprintf(str, str_size, "MD");
             break;
-        case DNS_RECORD_TYPE_TXT:
-            snprintf(str, str_size, "TXT");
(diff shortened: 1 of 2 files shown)


src/app-layer-dns-common.c
```diff
@@ -1016,51 +1016,174 @@ void DNSCreateTypeString(uint16_t type, char *str, size_t str_size)
         case DNS_RECORD_TYPE_NS:
             snprintf(str, str_size, "NS");
             break;
-        case DNS_RECORD_TYPE_AAAA:
```
inliniac: I think the most used types should be on top for efficiency.

maxtors: That sounds smart, will update and submit a new PR soon.

inliniac: On 28-04-16 14:52, Andreas Moe wrote:
> In src/app-layer-dns-common.c
> <https://github.com/inliniac/suricata/pull/2041#discussion_r61421248>:
> 
>> @@ -1016,51 +1016,174 @@ void DNSCreateTypeString(uint16_t type, char *str, size_t str_size)
>>          case DNS_RECORD_TYPE_NS:
>>              snprintf(str, str_size, "NS");
>>              break;
>> -        case DNS_RECORD_TYPE_AAAA:
> 
> That sounds smart, will update and submit a new PR soon.

Thanks!

maxtors:
Opened new PR #2042 .

PR #2043, opened 2016-04-28
Issue 1231 - SSL Negation - v4.1

jasonish:
Previous PR: #1474 
Redmine issue: https://redmine.openinfosecfoundation.org/issues/1231

First this required the current state being passed to the ssl_state match function instead of the cumulative state, otherwise ssl_state:client_hello would also match while in the state server_hello. Note that this patch is a bit naive as I don't know the SSL protocol that well, so its more of a preview or RFC.

This patch accepts "," as a state separator to be compatible with Snort, but keeps "|" as a separator for compatibility with existing Suricata rules.

PRscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/216
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/221

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -147,7 +147,10 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
 
     switch (ssl_state->curr_connp->handshake_type) {
         case SSLV3_HS_CLIENT_HELLO:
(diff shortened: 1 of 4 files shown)


jasonish:
Left debugging code.

PR #2049, opened 2016-05-02
app-layer-ssl: make tx aware - v8

thus:

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/23
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/23

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -130,9 +130,15 @@ struct AppLayerParserState_ {
 
     /* State version, incremented for each update.  Can wrap around. */
     uint8_t version;
+
+    /* Indicates which loggers that have already logged the transaction,
+       using the logger id (power of two) as a bitwise flag. */
+    uint32_t logged;
+
     /* Indicates the current transaction that is being inspected.
      * We have a var per direction. */
     uint64_t inspect_id[2];
+
     /* Indicates the current transaction being logged.  Unlike inspect_id,
      * we don't need a var per direction since we don't log a transaction
      * unless we have the entire transaction. */
@@ -532,6 +538,27 @@ void AppLayerParserSetTransactionLogId(AppLayerParserState *pstate)
 
     if (pstate != NULL)
         pstate->log_id++;
+
+    SCReturn;
+}
+
+uint8_t AppLayerParserGetTransactionLogged(AppLayerParserState *pstate, int logger_id)
+{
+    SCEnter();
+
+    if (pstate != NULL)
+        if (pstate->logged & logger_id)
+            SCReturnInt(1);
+
+    SCReturnInt(0);
+}
+
+void AppLayerParserSetTransactionLogged(AppLayerParserState *pstate, int logger_id)
+{
+    SCEnter();
+
+    if (pstate != NULL)
+        pstate->logged |= logger_id;
 
     SCReturn;
 }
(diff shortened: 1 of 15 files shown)


src/output-tx.c
```diff
@@ -71,20 +74,60 @@ int OutputRegisterTxLogger(const char *name, AppProto alproto, TxLogger LogFunc,
...
             t = t->next;
+        op->id = t->id * 2;
         t->next = op;
     }
 
     SCLogDebug("OutputRegisterTxLogger happy");
     return 0;
 }
 
+int OutputRegisterTxLoggerWithProgress(const char *name, AppProto alproto,
```
inliniac: instead of duplicating OutputRegisterTxLogger I think it's cleaner to make OutputRegisterTxLogger a wrapper to this function with preset values for int tc_log_progress, int ts_log_progress. I think it's possible to simply set it to the result AppLayerParserGetStateProgressCompletionStatus.

src/app-layer-parser.c
```diff
@@ -130,9 +130,15 @@ struct AppLayerParserState_ {
 
     /* State version, incremented for each update.  Can wrap around. */
     uint8_t version;
+
+    /* Indicates which loggers that have already logged the transaction,
+       using the logger id (power of two) as a bitwise flag. */
+    uint32_t logged;
```
inliniac: this doesn't track things per tx, which I think it should. The AppLayerParserState is per flow, not per tx.

src/output-tx.c
```diff
@@ -71,20 +74,60 @@ int OutputRegisterTxLogger(const char *name, AppProto alproto, TxLogger LogFunc,
...
-    if (list == NULL)
+    if (list == NULL) {
+        op->id = 1;
         list = op;
-    else {
+    } else {
         OutputTxLogger *t = list;
         while (t->next)
             t = t->next;
+        op->id = t->id * 2;
```
inliniac: we need a check to make sure this value is within the limits of the bitarray. I think you've used 32 bits so we should make sure it doesn't exceed that.

src/app-layer-parser.c
thus: Was this a comment on my comment? Or should this be implemented some other way?

src/output-tx.c
thus: Agreed, good idea!

src/output-tx.c
thus: Will do!

src/app-layer-parser.c
inliniac: in this implementation the flags are store in AppLayerParserState which a per-flow structure, not per tx. This means that logging TX 1 would disable logging for TX 2. The logged field should be stored in the tx instead I think.

thus: Oh, bummer! That should probably be changed then :)

inliniac: I think the way to go is a variant on @jasonish https://github.com/inliniac/suricata/pull/1992/commits/4de928f61287978e79ae4bc57c2ee96c691dc94d patch. The patch is boolean, in our use case we're looking at storing a uint16_t or uin32_t per tx. Not sure if we'd need per direction flags. Not if the logger id's are unique I think.

thus:
Replaced by #2056

PR #2036, opened 2016-04-26
Tpacket v3 10

regit:
Rework of #2032.

Modifications:
* Merge commit on block-size.
* Add a patch to use time() instead of GetTime()
* Add a patch to print mmap error
* Address comments of #2032

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/167
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/163

PR #2052, opened 2016-05-03
Tpacket v3 11

regit:
Update of #2036 

Rebased on latest master and fix message if block size is too small.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/170
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/166

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1192,6 +1192,11 @@
             AC_DEFINE([HAVE_PACKET_FANOUT],[1],[Recent packet fanout support is available]),
             [],
             [[#include <linux/if_packet.h>]])
+        AC_CHECK_DECL([TPACKET_V3],
+            AC_DEFINE([HAVE_TPACKET_V3],[1],[AF_PACKET tpcket_v3 support is available]),
+            [],
+            [[#include <sys/socket.h>
+              #include <linux/if_packet.h>]])
     ])
 
   # Netmap support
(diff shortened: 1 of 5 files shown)


inliniac:
On an interface with mtu 9000 block_size 8192 doesn't work:
<pre>
[4380] 3/5/2016 -- 09:19:25 - (source-af-packet.c:1614) <Error> (AFPComputeRingParamsV3) -- [ERRCODE: SC_ERR_INVALID_VALUE(130)] - Block size is too small, it should be at least 9104
</pre>
This makes sense, however I think you said somewhere the block_size for V3 should be a power of 2? If so this warning is inaccurate.

regit:
Do you want an update of text to say "Block size is too small, it should be a power of 2 superior to 9104"

inliniac:
On 03-05-16 09:41, Eric Leblond wrote:
> Do you want an update of text to say "Block size is too small, it should
> be a power of 2 superior to 9104"

Thats the lazy approach. If only we had a computer to actually calculate
the correct value ;)

regit:
Something like:
```
Block size is too small, it should be at least 16384 to contain a frame of length 9088
``` </lazy>

inliniac:
Merged through #2091, thanks Eric!

PR #2054, opened 2016-05-04
modbus: fix AddressSanitizer error (segmentation fault)

daviddiallo:
In case of Mask Write register or Write single register request with
no data (malformed packet), app-layer-modbus checks response content
(data) with the none stored request content. That causes the segmentation
fault.

Before accessing to request content, app-layer-modbus checks now if
content has been previously stored. 4 unitests have been adding, 2 of them
to test the management of Mask Write register and Write single register requests,
and the 2 others to check invalid Mask Write register and Write single register
requests.


inliniac:
Merged, thanks David!

PR #2051, opened 2016-05-02
Lua: SMTP support; Addresses feature ticket #1775; - v1

tobiass1:
Implemented support for matching on SMTP traffic from lua scripts.
This enables availibility to MIME fields, Attachement information,
MailFrom attribute, recipients list.
API documentation can be found at: http://pastebin.com/wtE0SgSt
Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1775

PR #2055, opened 2016-05-04
Lua: Smtp lua feature 1775 v2

tobiass1:
Implemented support for matching on SMTP traffic from lua scripts.
This enables availibility to MIME fields, Attachement information,
MailFrom attribute, recipients list.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1775

Fixes changes suggested in #2051

### DIFF
--- a/src/util-lua-smtp.c
+++ b/src/util-lua-smtp.c
@@ -15,7 +15,6 @@
  * 02110-1301, USA.
  */
 
-
 /**
  *  \file
  *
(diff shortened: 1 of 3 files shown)


inliniac:
Please squash the commits. See point 2 under 'commits' here https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits

src/util-lua-smtp.c
```diff
@@ -0,0 +1,477 @@
...
+    lua_newtable(luastate);
+    while (field != NULL) {
+        if(field->name != NULL) {
+            lua_pushinteger(luastate,num++);
+            LuaPushStringBuffer(luastate, field->name, field->name_len);
+            lua_settable(luastate,-3);
+        }
+        field = field->next;
+    }
+    return 1;
```
inliniac: is this correct if we just had one 'field' and it's name was NULL? In this case we'd not push anything to the stack.

src/util-lua-smtp.c
```diff
@@ -0,0 +1,477 @@
...
+    SMTPTransaction *smtp_tx = state->curr_tx;
+    if(smtp_tx == NULL) {
+        return LuaCallbackError(luastate, "Internal Error: no SMTP transaction");
+    }
+    if(smtp_tx->mail_from == NULL || smtp_tx->mail_from_len == 0) {
+        return LuaCallbackError(luastate, "MailFrom not found");
+    }
+    LuaPushStringBuffer(luastate, smtp_tx->mail_from, smtp_tx->mail_from_len);
+    /* Returns 1 because we never push more then 1 item to the lua stack */
+    return 1;
```
inliniac: we can probably have 'return LuaPushStringBuffer(luastate, smtp_tx->mail_from, smtp_tx->mail_from_len);' instead?

src/util-lua-smtp.c
```diff
@@ -0,0 +1,477 @@
...
+    /* point to start of list for iterating trough */
+    File *file = file_con->head;
+    if(file == NULL) {
+        return LuaCallbackError(luastate, "error: no file(s) in container");
+    }
+    int u = 1;
+    /* create new table for placement of findings */
+    lua_newtable(luastate);
+    /* loop through and push filename to luastate table on stack */
+    while(file != NULL) {
```
inliniac: files are stored per smtp state, not per 'transaction'. Here simply all files are returned. I think it would make more sense to act only on the files belong to the current tx?

inliniac:
@regit could you have a look as well?

PR #2056, opened 2016-05-04
app-layer-ssl: make tx aware - v10

thus:

Fixed changes suggested in #2049.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/25
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/25

### DIFF
--- a/src/app-layer-ssl.h
+++ b/src/app-layer-ssl.h
@@ -98,8 +98,6 @@ enum ssl_tx_progress_t {
 /* flags for file storage */
 #define SSL_AL_FLAG_STATE_STORED                0x40000
 
-#define SSL_AL_FLAG_STATE_LOGGED_LUA            0x80000
-
 /* config flags */
 #define SSL_TLS_LOG_PEM                         (1 << 0)
 
--- a/src/app-layer-ssl.h
+++ b/src/app-layer-ssl.h
@@ -26,6 +26,8 @@
 #ifndef __APP_LAYER_SSL_H__
 #define __APP_LAYER_SSL_H__
 
+#include "app-layer-protos.h"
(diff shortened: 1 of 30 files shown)


src/app-layer-ssl.h
```diff
@@ -53,6 +55,12 @@ enum {
     TLS_DECODER_EVENT_INVALID_SSL_RECORD,
 };
 
+enum ssl_tx_progress_t {
```
inliniac: can you update the name to be compliant to our naming style?

src/output-tx.c
```diff
@@ -161,20 +172,46 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
...
+                        goto next;
+                    }
+                }
+
                 PACKET_PROFILING_TMM_START(p, logger->module_id);
                 logger->LogFunc(tv, store->thread_data, p, f, alstate, tx, tx_id);
                 PACKET_PROFILING_TMM_END(p, logger->module_id);
-                proto_logged = 1;
+
+                AppLayerParserSetTxLogged(p->proto, alproto, alstate, tx_id,
```
inliniac: I think this is quite inefficient: even though at this point we have the tx ptr, we just pass the tx_id to the AppLayerParserSetTxLogged call. This in turn means that the underlying implementation has to look up the tx ptr from the alstate based on the tx_id. Unless I'm missing a side effect, I think it's much more efficient to simply pass on the tx ptr here. Makes all the protocol callbacks simpler as well.

thus:
Replaced by #2071

PR #2058, opened 2016-05-05
Fix wrong boolean conf read v1

maxtors:
Multiple places there were checks for "yes" or "no" with regards to some functionality being enabled or not. Refactored these to use the ConfValIsTrue / ConfValIsFalse so that it is more accepting of valid YAML input (true being 1,yes,true,on. false being 0,no,false,off).


inliniac:
Thanks @maxtors !

PR #2060, opened 2016-05-05
Cppcheck cleanup v1

maxtors:
Looked for variables where scope could be altered, also looked for codestyle fixing. Not all source files have been processed though.

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -346,11 +346,10 @@ void HTPStateFree(void *state)
     if (s->connp != NULL) {
         SCLogDebug("freeing HTP state");
 
-        uint64_t tx_id;
         uint64_t total_txs = HTPStateGetTxCnt(state);
         /* free the list of body chunks */
         if (s->conn != NULL) {
-            for (tx_id = 0; tx_id < total_txs; tx_id++) {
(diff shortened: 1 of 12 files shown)


src/app-layer-htp.c
```diff
@@ -824,8 +822,7 @@ static int HTPHandleResponseData(Flow *f, void *htp_state,
 
     htp_time_t ts = { f->lastts.tv_sec, f->lastts.tv_usec };
     if (input_len > 0) {
-        r = htp_connp_res_data(hstate->connp, &ts, input, input_len);
-        switch(r) {
+        switch(htp_connp_res_data(hstate->connp, &ts, input, input_len)) {
```
inliniac: I don't like this, the compiler can optimize this and while debugging it's good to have an explicit variable to inspect.

src/app-layer-smb.c
```diff
@@ -98,160 +98,132 @@ static uint32_t SMBParseWriteAndX(Flow *f, void *smb_state,
                 if (!(--input_len))
                     break;
             }
-            /* fall through */
```
inliniac: why would you remove this? it's important to tell devs AND code checkers that we fall through intentionally

maxtors:
Incorporated feedback into #2061.

PR #2061, opened 2016-05-06
Cppcheck cleanup v2

maxtors:
Looked for variables where scope could be altered, also looked for codestyle fixing. Not all source files have been processed though.

Incorporated feedback from #2060.


maxtors:
More work done in #2064.

PR #2062, opened 2016-05-06
sources: allow interface definitions to be reordered - v2

jasonish:
Previous PR: https://github.com/inliniac/suricata/pull/1554
Related ticket: https://redmine.openinfosecfoundation.org/issues/1487

For af-packet, pf-ring, netmap, and pcap use a generic
lookup function to find the configuration node for an
interface.

The new lookup function does not depend on the ordering
of the items inside the device configuration.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/222
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/227

### DIFF
--- a/src/util-conf.c
+++ b/src/util-conf.c
@@ -25,6 +25,7 @@
 #include "suricata-common.h"
 #include "config.h"
 #include "conf.h"
+#include "util-conf.h"
 
 TmEcode ConfigSetLogDirectory(char *name)
 {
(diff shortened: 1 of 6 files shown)


src/util-conf.c
```diff
@@ -63,3 +64,30 @@ TmEcode ConfigCheckLogDirectory(char *log_dir)
...
+ *
+ * \param iface The name of the interface to find the config for.
+ */
+ConfNode *ConfFindDeviceConfig(ConfNode *node, const char *iface)
+{
+    ConfNode *if_node, *item;
+    const char key[] = "interface";
+    TAILQ_FOREACH(if_node, &node->head, next) {
+        TAILQ_FOREACH(item, &if_node->head, next) {
+            if (!(strcmp(item->name, key) && strcmp(item->val, iface))) {
```
inliniac: this line confuses me, strcmp() == true (which is implied), means no match. So !( no match && no match) ?

jasonish: Ah, yes. This is flawed. It works with the expected YAML, but there is YAML that could make it fail.

jasonish:
See https://github.com/inliniac/suricata/pull/2063

PR #2065, opened 2016-05-07
cppcheck / codestyle cleanup v4

maxtors:
Fixes for codestyle based on suricata project codestyle, and cppcheck.


src/alert-unified2-alert.c
```diff
@@ -254,14 +254,11 @@ void TmModuleUnified2AlertRegister(void)
  *  \param t Thread Variable containing  input/output queue, cpu affinity etc.
  *  \param aun Unified2 thread variable.
  */
-
 int Unified2AlertCloseFile(ThreadVars *t, Unified2AlertThread *aun)
 {
-    if (aun->unified2alert_ctx->file_ctx->fp != NULL) {
```
inliniac: I don't want these 'fixes' that are removing braces. Keeping the braces makes the code more robust against later insertion errors.

PR #2067, opened 2016-05-09
config-test: fix memory leak detected by ASAN

regit:
NSS library was not deinit at exit resulting in memory leak. As
it is useless for a config test, the patch updates the code so it
is not initialized.

Patch also calls MagicDeinit to free memory used by libmagic.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/171
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/167


inliniac:
Merged, thanks Eric!

PR #2072, opened 2016-05-12
app-layer-ssl: make tx aware - v13

thus:

Fixed changes suggested in #2056.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/27
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/27

### DIFF
--- a/src/output-tx.c
+++ b/src/output-tx.c
@@ -47,6 +47,7 @@ typedef struct OutputLoggerThreadData_ {
 typedef struct OutputTxLogger_ {
     AppProto alproto;
     TxLogger LogFunc;
+    TxLoggerCondition LogCondition;
     OutputCtx *output_ctx;
     struct OutputTxLogger_ *next;
     const char *name;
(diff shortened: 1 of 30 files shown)


src/output-tx.c
```diff
@@ -47,15 +47,21 @@ typedef struct OutputLoggerThreadData_ {
 typedef struct OutputTxLogger_ {
     AppProto alproto;
     TxLogger LogFunc;
+    TxLoggerCondition LogCondition;
```
inliniac: Wondering if it makes sense to have conditions per direction as well.

src/output-tx.h
```diff
@@ -34,9 +34,11 @@ typedef int (*TxLogger)(ThreadVars *, void *thread_data, const Packet *, Flow *f
 /** packet logger condition function pointer type,
  *  must return true for packets that should be logged
  */
-//typedef int (*TxLogCondition)(ThreadVars *, const Packet *);
+typedef int (*TxLoggerCondition)(ThreadVars *, const Packet *, void *state);
```
inliniac: I think it's a bit strange that the Tx 'Condition' doesn't get a pointer to the, you know, tx ;)

thus: I didn't need it for my one condition, so I thought it could be added later, but yeah ;)

thus:
Replaced by #2074

PR #2074, opened 2016-05-12
app-layer-ssl: make tx aware - v14

thus:

Fixed changes suggested in #2056.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/30
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/30


thus:
Added tx pointer and tx_id to condition callback.

inliniac:
Merged through #2081. Thanks a lot @thus!

PR #2075, opened 2016-05-13
Hyperscan SPM integration v1

jviiret:

The algorithm is selectable with the "spm-algo" yaml option, and defaults to "bm".

### DIFF
--- a/src/detect-content.c
+++ b/src/detect-content.c
@@ -217,7 +217,8 @@ error:
  * \brief DetectContentParse
  * \initonly
  */
-DetectContentData *DetectContentParse (char *contentstr)
+DetectContentData *DetectContentParse(SpmThreadCtx *spm_thread_ctx,
+                                      char *contentstr)
 {
     DetectContentData *cd = NULL;
     uint8_t *content = NULL;
@@ -247,7 +248,8 @@ DetectContentData *DetectContentParse (char *contentstr)
 
     /* Prepare Boyer Moore context for searching faster */
     cd->bm_ctx = BoyerMooreCtxInit(cd->content, cd->content_len);
-    cd->spm_ctx = SpmInitCtx(cd->content, cd->content_len, 0, NULL, SPM_BM);
+    cd->spm_ctx = SpmInitCtx(cd->content, cd->content_len, 0, spm_thread_ctx,
+                             spm_thread_ctx->matcher);
     cd->depth = 0;
     cd->offset = 0;
     cd->within = 0;
@@ -258,7 +260,8 @@ DetectContentData *DetectContentParse (char *contentstr)
 
 }
 
-DetectContentData *DetectContentParseEncloseQuotes(char *contentstr)
+DetectContentData *DetectContentParseEncloseQuotes(SpmThreadCtx *spm_thread_ctx,
+                                                   char *contentstr)
 {
     char str[strlen(contentstr) + 3]; // 2 for quotes, 1 for \0
 
@@ -267,7 +270,7 @@ DetectContentData *DetectContentParseEncloseQuotes(char *contentstr)
     str[strlen(contentstr) + 1] = '\"';
     str[strlen(contentstr) + 2] = '\0';
 
-    return DetectContentParse(str);
+    return DetectContentParse(spm_thread_ctx, str);
 }
 
 /**
@@ -371,7 +374,7 @@ int DetectContentSetup(DetectEngineCtx *de_ctx, Signature *s, char *contentstr)
     DetectContentData *cd = NULL;
     SigMatch *sm = NULL;
 
-    cd = DetectContentParse(contentstr);
+    cd = DetectContentParse(de_ctx->spm_thread_ctx, contentstr);
     if (cd == NULL)
         goto error;
     DetectContentPrint(cd);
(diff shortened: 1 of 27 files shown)


src/util-spm-hs.c
```diff
@@ -0,0 +1,201 @@
...
+                       const uint8_t *haystack, uint16_t haystack_len)
+{
+    const SpmHsCtx *sctx = ctx->ctx;
+    hs_scratch_t *scratch = thread_ctx->ctx;
+
+    uint64_t match_offset = UINT64_MAX;
+    hs_error_t err = hs_scan(sctx->db, (const char *)haystack, haystack_len, 0,
+                             scratch, MatchEvent, &match_offset);
+    if (err != HS_SUCCESS && err != HS_SCAN_TERMINATED) {
+        SCLogError(SC_ERR_FATAL, "Scan with Hyperscan returned error %d.", err);
```
inliniac: in what kind of scenario could this happen?

src/util-spm-hs.c
```diff
@@ -0,0 +1,201 @@
...
+    uint16_t needle_len;
+} SpmHsCtx;
+
+static SpmCtx *HSInitCtx(const uint8_t *needle, uint16_t needle_len, int nocase,
+                         SpmThreadCtx *thread_ctx)
+{
+    char *expr = HSRenderPattern(needle, needle_len);
+    if (expr == NULL) {
+        SCLogError(SC_ERR_FATAL, "Unable to alloc string. Exiting.");
+        exit(EXIT_FAILURE);
```
inliniac: When we have rule reloads I would like to be able to survive running out of memory. So would it be feasible to do this more gracefully? I do realize the existing BM code does the same thing currently.

src/detect-content.c
```diff
@@ -435,7 +438,14 @@ int DetectContentParseTest01 (void)
     char *teststring = "\"abc\\:def\"";
     char *teststringparsed = "abc:def";
 
-    cd = DetectContentParse(teststring);
+    uint16_t spm_matcher = SinglePatternMatchDefaultMatcher();
+    SpmThreadCtx *spm_thread_ctx = SpmInitThreadCtx(spm_matcher);
+    if (spm_thread_ctx == NULL) {
```
inliniac: can you replace this by the new FAIL_IF(expr) logic? See for example https://github.com/inliniac/suricata/commit/7f700a137cd54497c16d99a0c4418f1296a87aaa

src/detect.h
```diff
@@ -587,6 +588,11 @@ typedef struct DetectEngineCtx_ {
     ThresholdCtx ths_ctx;
 
     uint16_t mpm_matcher; /**< mpm matcher this ctx uses */
+    uint16_t spm_matcher; /**< spm matcher this ctx uses */
+
+    /* spm thread context prototype, built as spm matchers are constructed and
+     * later cloned for each thread. */
+    SpmThreadCtx *spm_thread_ctx;
```
inliniac: I think it would help me to rename this to spm_master_thread_ctx or similar. An alternative would be to use a diffirent 'SpmGlobalCtx' like construct, that would be used to create the thread context's. That is more how the Mpm variant works. Especially in the single threaded rule loading stage, it confuses a bit to already pass a 'ThreadCtx' around.

inliniac:
Other than the comments I made I like this approach. My QA likes it too, so that is good.
I will add additional DrMemory suppressions:
<pre>
diff --git a/qa/drmemory.suppress b/qa/drmemory.suppress
index 7f1da7d..d9c89a1 100644
--- a/qa/drmemory.suppress
+++ b/qa/drmemory.suppress
@@ -43,6 +43,22 @@ suricata!SigGroupBuild
 ...
 libc.so.6!__clone
 
+UNINITIALIZED READ
+name=Hyperscan false positives due to DrMemory limitations 3
+...
+libhs.so.*!hs_compile
+suricata!SpmInitCtx
+...
+suricata!main
+
+UNINITIALIZED READ
+name=Hyperscan false positives due to DrMemory limitations 4
+...
+libhs.so.*!hs_compile
+suricata!SpmInitCtx
+...
+libc.so.6!__clone
+
 WARNING
 name=Prefetch warnings 1
 ...
@@ -58,3 +74,19 @@ libhs.so.*!hs_scan
 suricata!SCHSSearch
 ...
 libc.so.6!__clone
+
+WARNING
+name=Prefetch warnings 3
+...
+libhs.so.*!hs_scan
+suricata!HSScan
+...
+suricata!main
+
+WARNING
+name=Prefetch warnings 4
+...
+libhs.so.*!hs_scan
+suricata!HSScan
+...
+libc.so.6!__clone
</pre>

inliniac:
Oh another comment: lets default to hs if it is built in. I guess we can have an 'auto' value that selects hs if available and otherwise bm. We can do the same for mpm-algo.

src/util-spm-hs.c
jviiret: The `hs_scan` function should only return an error (i.e. something other than `HS_SUCCESS` or `HS_SCAN_TERMINATED`) if it's been passed an invalid database or scratch value. So actually, this should probably be fatal and cause an exit, rather than "failing open" like this. I'll make the change in the next PR.

src/util-spm-hs.c
jviiret: I can certainly rejig this so that rather than calling `exit` on failure, we return `NULL` and propagate it up to the caller, which can decide what to do as a result -- is this what you mean, or have I misunderstood?

src/detect-content.c
jviiret: Oh, I hadn't noticed those -- sure, I'll swap it in.

src/detect.h
jviiret: I like the idea of using a different type, which will make it harder to accidentally use the wrong structure at scan time. Will add an SpmGlobalCtx structure.

jviiret:
Done in v2. This PR is superseded by #2078. (I've also tried to squash down the commits a little more logically.)

PR #2077, opened 2016-05-16
configure.ac: escape $srcdir when used in a variable - v1

jasonish:
$srcdir needs to be escaped for proper expansion when used
as part of a Makefile variable.

Fixes the issue reported on the mailing list when building Suricata within the OpenBSD ports system, however, this is not OpenBSD specific.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/225
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/230


inliniac:
Merged through #2081, thanks Jason.

PR #2078, opened 2016-05-17
Hyperscan SPM integration v2

jviiret:
> Version 2. This addresses comments in v1 as follows:
> - treat error return values from `hs_scan` as fatal: they can only result from corrupted database or scratch args
> - react more gracefully to alloc failures, rather than exiting
> - use `FAIL_IF()` in unit tests
> - add an `SpmGlobalThreadCtx` structure that is used as the prototype for constructing the per-thread `SpmThreadCtx` structures
> - add support for config "spm-algo: auto", which defaults to Hyperscan when compiled in and Boyer-Moore otherwise.
>
> The previous PR was #2075.

The algorithm is selectable with the "spm-algo" yaml option, and defaults to "hs" when compiled with Hyperscan support, and "bm" otherwise.

### DIFF
--- a/src/util-spm-hs.c
+++ b/src/util-spm-hs.c
@@ -1,0 +1,254 @@
+/* Copyright (C) 2016 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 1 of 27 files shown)


src/util-spm-hs.c
```diff
@@ -0,0 +1,254 @@
...
+static void HSDestroyGlobalThreadCtx(SpmGlobalThreadCtx *global_thread_ctx)
+{
+    if (global_thread_ctx == NULL) {
+        return;
+    }
+    hs_free_scratch(global_thread_ctx->ctx);
+    SCFree(global_thread_ctx);
+}
+
+static SpmThreadCtx *HSInitThreadCtx(void)
```
inliniac: Dead code or is it an error that it isn't used?
<pre>
util-spm-hs.c:192:22: error: unused function 'HSInitThreadCtx' [-Werror,-Wunused-function]
static SpmThreadCtx *HSInitThreadCtx(void)
                     ^
1 error generated.
</pre>

jviiret: This is dead code -- sorry, I must have missed it when I added the SpmGlobalThreadCtx structure. I'll fix it and submit a v3 PR.

jviiret:
Superseded by v3, in PR 2083.

PR #2083, opened 2016-05-17
Hyperscan SPM integration v3

jviiret:
> Version 3. Removed some dead code that was erroneously left in v2. The previous PR was #2078.

The algorithm is selectable with the "spm-algo" yaml option, and defaults to "hs" when compiled with Hyperscan support, and "bm" otherwise.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -350,6 +350,7 @@ util-hashlist.c util-hashlist.h \
 util-hash-lookup3.c util-hash-lookup3.h \
 util-host-os-info.c util-host-os-info.h \
 util-host-info.c util-host-info.h \
+util-hyperscan.c util-hyperscan.h \
 util-ioctl.h util-ioctl.c \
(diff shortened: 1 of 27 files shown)


pevma:
@jviiret  - Would any SPM stats/ref be available/added when "--enable-profiling" is used to do test runs?

jviiret:
@pevma - I hadn't considered that, actually. I guess time attributed to SPM is already attributed to the individual keywords ('content', etc) that make use of it... what sort of additional profiling information are you interested in?

inliniac:
Merged through #2086, thanks a lot @jviiret! I added https://github.com/inliniac/suricata/pull/2086/commits/8dc63f625f9a8a890d231e1d7dc7c38a30a95168 to suppress DrMemory warnings.

pevma:
@jviiret  - nothing in particular - was just making sure we would not miss out any extra info (if available).

PR #2084, opened 2016-05-18
[preview] flow worker v38

inliniac:
This branch simplifies the flow locking in detection a lot. It introduces the concept of a "FlowWorker": a single thread module that runs Flow, Stream, AppLayer and Detect under a single flow lock.

For workers, not much changes (assuming flow based load balancing works correctly). For autofp, a lot changes:
- capture thread(s) no longer handle the flow lookup, creation and updates.
- capture threads no longer handle UDP app layer
- flow based load balancing for autofp can no longer be smart and stateful, as we have no flow to keep state. Load balancing is hash based again.

In my testing I'm seeing better performance, also for autofp.

Some advantages:
- simpler code
- detection profile stats won't get influenced by flow locks anymore (cc: @wmetcalf)
- detection can keep pointers to flow data w/o risk of data going away

Please help test this.

Todo:
- I would like to move the outputs into the FlowWorker as well. This will require a big output API overhaul though. cc @jasonish 

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/426
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/431

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -220,6 +220,7 @@ flow-storage.c flow-storage.h \
 flow-timeout.c flow-timeout.h \
 flow-util.c flow-util.h \
 flow-var.c flow-var.h \
+flow-worker.c flow-worker.h \
 host.c host.h \
 host-bit.c host-bit.h \
 host-queue.c host-queue.h \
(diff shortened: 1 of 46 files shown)


inliniac:
Replaced by #2089

PR #2087, opened 2016-05-19
Fix things in prelude connector and add ipv6 support

ToToL:


inliniac:
cc @regit

src/alert-prelude.c
```diff
@@ -207,12 +207,18 @@ static int EventToImpact(const PacketAlert *pa, const Packet *p, idmef_alert_t *
     SCEnter();
 
     ret = idmef_alert_new_assessment(alert, &assessment);
-    if (unlikely(ret < 0))
+    if (unlikely(ret < 0)) {
+        SCLogDebug("%s: error creating assessment: %s.",
```
regit: If it is really an error then we should use a SCLogError but we don't want to flood logs if it happens for every alerts.

ToToL: It is an error, creating an assessment should always be possible. But as you said, if it happened one time, it could happen many time, so I use debug instead of Error.

inliniac:
Ran QA on this, but it fails:
<pre>
alert-prelude.c: In function ‘PacketToData’:
alert-prelude.c:515:27: error: ‘TCPVars’ has no member named ‘ts’
alert-prelude.c:522:31: error: invalid operands to binary != (have ‘TCPOpt’ and ‘void *’)
alert-prelude.c:525:33: error: invalid operands to binary != (have ‘TCPOpt’ and ‘void *’)
make[3]: *** [alert-prelude.o] Error 1
</pre>

Please submit a new PR. See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Incorporating-feedback

PR #2088, opened 2016-05-20
SPM API: handle NULL pointers gracefully (Bug #1786)

jviiret:
Fix for redmine bug 1786: https://redmine.openinfosecfoundation.org/issues/1786

This will handle minimal DetectEngineCtx structures (used in delayed detect mode) safely, since they don't get SPM global contexts allocated.

Also added BUG_ON checks for valid spm_table entries.


inliniac:
Merged through #2091, thanks Justin!

PR #2017, opened 2016-04-18
app-layer-tls: validity dates from TLS certificate - v10

thus:
Get validity dates (notBefore and notAfter) from TLS certificate.

New detection keywords (tls_notbefore and tls_notafter), added validity dates to extended JSON log and extended TLS log output, and new lua functions TlsGetCertNotBefore() and TlsGetCertNotAfter().

https://redmine.openinfosecfoundation.org/issues/1583

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/17
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/17

PR #2090, opened 2016-05-20
Parsing and eve log output of tls validity timestamps.

maxtors:
Added extraction of UTCTIME validity fields from TLS certs along with output of this to the TLS EVE log.

Example of Eve log output:
  "tls": {
    "subject": "unknown=Private Organization, unknown=US, unknown=Delaware, serialNumber=5157550, unknown=88 Colin P Kelly, Jr Street, unknown=94107, C=US, ST=California, L=San Francisco, O=GitHub, Inc., CN=github.com",
    "issuerdn": "C=US, O=DigiCert Inc, OU=www.digicert.com, CN=DigiCert SHA2 Extended Validation Server CA",
    "fingerprint": "d7:9f:07:61:10:b3:92:93:e3:49:ac:89:84:5b:03:80:c1:9e:2f:8b",
    "sni": "github.com",
    **"not-valid-before": "2010-03-15T23:00:00.000000+0100",**
    **"not-valid-after": "2017-05-18T12:00:00.000000+0200",**
    "version": "TLS 1.2"
  }

### DIFF
--- a/src/util-decode-der-get.c
+++ b/src/util-decode-der-get.c
@@ -39,6 +39,7 @@
 
 static const uint8_t SEQ_IDX_ISSUER[] = { 0, 2 };
 static const uint8_t SEQ_IDX_SUBJECT[] = { 0, 4 };
+static const uint8_t SEQ_IDX_VALIDITY[] = { 0, 3 };
 
 static const char *Oid2ShortStr(const char *oid)
 {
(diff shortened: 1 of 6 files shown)


maxtors:
I'm thinking that maybe the keywords "not-valid-before" / "not-valid-after" should be using underscores, seeing that all other "multi word keywords" in the JSON format uses underscores.

inliniac:
cc @thus

thus:
There is already (a lot of) work in progress to add this functionality to Suricata, see #2017.

The implementation in this PR does not take UTC into consideration, as both 'mktime' and 'strptime' operate on local time, not UTC. A lot of the headaches that is solved in #2017, is to use UTC instead of local time.

Is the year parsed as it should? I only see parsing of a two digit year here, so it looks like it might break if it is for a date before year 2000. According to the RFC (RFC 5280, 4.1.2.5.1) the year should be 19YY if (the two digit year) is greater than or equal to 50, and 20YY if it is below 50.

Another thing that is not considered here, is the use of GeneralizedTime (see RFC 5280, 4.1.2.5.2), which must be used for dates after the year of 2049.

src/util-decode-der-get.c
```diff
@@ -205,6 +206,75 @@ int Asn1DerGetIssuerDN(const Asn1Generic *cert, char *buffer, uint32_t length,
...
 
+int Ans1DerGetValidityParseTimestamp(char *dst, uint32_t dst_len,
+        char *src, uint32_t src_len)
+{
+    int rc = -1;
+    struct timeval ts;
+    struct tm tm;
+    time_t epoch;
+
+    if (strptime(src, "%d%m%y%H%M%SZ", &tm) == NULL)
```
thus: This does not take into consideration if the year is 19YY or 20YY. It also uses local time instead of UTC, which would work poorly, since the validity dates are using UTC.

src/util-decode-der-get.c
```diff
@@ -205,6 +206,75 @@ int Asn1DerGetIssuerDN(const Asn1Generic *cert, char *buffer, uint32_t length,
...
+        goto parsing_error;
+
+    epoch = mktime(&tm);
+    memset(&ts, 0, sizeof(struct timeval));
+    ts.tv_sec = (uint32_t) epoch;
+    CreateIsoTimeString(&ts, dst, dst_len);
```
thus: This function uses 'localtime_r' which also use local time, instead of UTC.

src/util-decode-der-get.c
```diff
@@ -205,6 +206,75 @@ int Asn1DerGetIssuerDN(const Asn1Generic *cert, char *buffer, uint32_t length,
...
+{
+    int rc = -1;
+    struct timeval ts;
+    struct tm tm;
+    time_t epoch;
+
+    if (strptime(src, "%d%m%y%H%M%SZ", &tm) == NULL)
+        goto parsing_error;
+
+    epoch = mktime(&tm);
```
thus: 'mktime' use local time, and not UTC.

maxtors:
I think that, with all the work allready done in #2017, and the poor timestamp handeling in this PR is enough to close it. And rather wait for #2017.

PR #2093, opened 2016-05-21
Suricata file mgic, captuer file not open

androuxx:
i use suricata 3 on my ubuntu 14.0.4
i try to capture the files on my network with file-store feature, i use file-store
```
- file-store:
      enabled: yes       # set to yes to enable
      log-dir: files    # directory to store the files
      force-magic: yes   # force logging magic on all stored files
      force-md5: yes     # force logging of md5 checksums
      force-filestore: yes # force storing of all files
      waldo: file.waldo # waldo file to store the file_id across runs
```
but the capture file no open and this size  smaller the the original file
suricata_file2
why that is happen ?
and the can i rename the output captures files as the original name with the extension like "art.pdf" and "art.meta" ?


inliniac:
Please don't abuse pull requests for reporting issues. To report issues, go here: https://redmine.openinfosecfoundation.org/projects/suricata/issues

PR #2097, opened 2016-05-23
Changed StatsInitCtx to use ConfGetChildValueInt to get interval.

maxtors:
Encountered an error were the value "yes" was set under stats.interval. This lead to the value being set to 0 (zero). And well, alot of stats were generated =). So changed this to get an int instead.

### DIFF
--- a/src/counters.c
+++ b/src/counters.c
@@ -234,9 +234,15 @@ static void StatsInitCtx(void)
             SCLogDebug("Stats module has been disabled");
             SCReturn;
         }
-        const char *interval = ConfNodeLookupChildValue(stats, "interval");
-        if (interval != NULL)
-            stats_tts = (uint32_t) atoi(interval);
+        
+        intmax_t interval;
+        if (ConfGetChildValueInt(stats, "interval", &interval) == 0) {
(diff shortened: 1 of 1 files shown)


src/counters.c
```diff
@@ -234,9 +234,15 @@ static void StatsInitCtx(void)
...
             SCReturn;
         }
-        const char *interval = ConfNodeLookupChildValue(stats, "interval");
-        if (interval != NULL)
-            stats_tts = (uint32_t) atoi(interval);
+        
+        intmax_t interval;
+        if (ConfGetChildValueInt(stats, "interval", &interval) == 0) {
+            stats_enabled = FALSE;
+            SCLogDebug("Stats module has been disabled, invalid interval");
```
inliniac: This should probably be an error instead? Will be hard understand for normal users otherwise.

maxtors: Agree on that, should it be a fatal error message (SC_ERR_FATAL) and not some kind of fallback to the default value, since maybe other systems / scripts / etc, could be dependant on the correct stats output interval.

inliniac: Think a warning with fallback to default values would be fine here.

PR #2098, opened 2016-05-23
Reduced scope of variables based on cppcheck.

maxtors:


inliniac:
Needs rebase, but also a split up. I don't mind taking some of these changes, but they will have to be in smaller commits.

PR #2099, opened 2016-05-24
Dev dhcp v0.1

decanio:
DHCP parser and JSON logging support.

- PR decanio: https://buildbot.openinfosecfoundation.org/builders/decanio/builds/21
- PR decanio-pcap: https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/21


decanio:
Sorry.  Too early in the morning.  Wrong comment on the PR.  This one is DHCP parser and logging support.

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,668 @@
...
+    DHCP_DECODER_EVENT_EMPTY_MESSAGE,
+};
+
+SCEnumCharMap dhcp_decoder_event_table[] = {
+    {"EMPTY_MESSAGE", DHCP_DECODER_EVENT_EMPTY_MESSAGE},
+};
+
+static dhcpState dhcpGlobalState;
+
+static dhcpTransaction *dhcpTxAlloc(dhcpState *dhcp)
```
jasonish: My first thought is that this function name needs to be capitalized, as well the dhcpTransaction type.

scripts/setup-app-layer-logger.sh
```diff
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
```
inliniac: this belongs in a separate commit

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,668 @@
...
+
+    SCFree(tx);
+}
+
+static void *dhcpStateAlloc(void)
+{
+    /* TBD: possibly make this per vlan */
+    dhcpState *state = &dhcpGlobalState;
+
+    if (SC_ATOMIC_CAS(&state->initialized, 0, 1) == 1) {
```
inliniac: what is this CAS's purpose?

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,668 @@
...
+     * either. */
+    if (input == NULL || input_len == 0) {
+        return 0;
+    }
+
+    /* Normally you would parse out data here and store it in the
+     * transaction object, but as this is echo, we'll just record the
+     * request data. */
+
+    if (input_len >= DHCP_MIN_FRAME_LEN) {
```
inliniac: if (input_len < DHCP_MIN_FRAME_LEN)\n goto end;
would have the same effect, except the whole block after can move one step left. Much preferred for readability.

src/output-json-dhcp.c
```diff
@@ -0,0 +1,439 @@
...
+        goto error;
+    }
+
+    uint32_t offset;
+    /* TBD: harden this loop against bad len values */
+    DHCPOpt *dhcp_opt = (DHCPOpt *)dhcptx->request_buffer;
+    for (offset = 0; offset < dhcptx->request_buffer_len; offset += (2 + dhcp_opt->len)) {
+        dhcp_opt = (DHCPOpt *)&dhcptx->request_buffer[offset];
+        switch (dhcp_opt->code) {
+            case 53: {
```
inliniac: can we used names instead of integers here?

inliniac:
I'm still a bit fuzzy on how the state and tx' work. The state seems to have a lock implying concurrent access. How does that affect transactions? The tx is looked up under lock, but then other operations on it seem to be w/o lock.

PR #2103, opened 2016-05-24
Removed repeated string comparison

serializingme:
In the libhtp configuration parsing code, there is an unnecessary string comparison.

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -2293,8 +2293,7 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


src/app-layer-htp.c
```diff
@@ -2293,8 +2293,7 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
                 continue;
             }
 
-        } else if (strcasecmp("request-body-limit", p->name) == 0 ||
-                   strcasecmp("request_body_limit", p->name) == 0) {
```
inliniac: These 2 strings are different.

serializingme: Oh, wow... that was embarrassing :D

PR #2094, opened 2016-05-23
MPM: Add "mpm-algo: auto", use Hyperscan by default when supported

jviiret:
Ticket: https://redmine.openinfosecfoundation.org/issues/1789

As per the ticket, this change adds the "mpm-algo: auto" option, which will select Hyperscan by default if Suricata has been built with Hyperscan support, and use the current behaviour (ac, or ac-tile on tilera) if not.

This has been done by defining `DEFAULT_MPM` appropriately, which required a few small fixes to other paths that use the MPM matchers.

Finally, there are a couple of small fixes to the Hyperscan integration:
- making an error from `hs_scan()` fatal in the MPM path, like for SPM
-  allowing zero-byte scans, which can occur for some paths. Hyperscan is OK with these, but a passing it a NULL buffer will cause it to return an error, so I've added a guard.

PR #2105, opened 2016-05-25
MPM: Add "mpm-algo: auto", use Hyperscan by default when supported (v2)

jviiret:
> Version 2. Squashes a memory leak in app-layer-smtp (see 75d1518) by ensuring that its MPM contexts are cleaned up. The previous PR was #2094.

Ticket: https://redmine.openinfosecfoundation.org/issues/1789

As per the ticket, this change adds the "mpm-algo: auto" option, which will select Hyperscan by default if Suricata has been built with Hyperscan support, and use the current behaviour (ac, or ac-tile on tilera) if not.

This has been done by defining `DEFAULT_MPM` appropriately, which required a few small fixes to other paths that use the MPM matchers.

Finally, there are a couple of small fixes to the Hyperscan integration:
- making an error from `hs_scan()` fatal in the MPM path, like for SPM
-  allowing zero-byte scans, which can occur for some paths. Hyperscan is OK with these, but a passing it a NULL buffer will cause it to return an error, so I've added a guard.

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -1528,6 +1528,8 @@ int AppLayerProtoDetectSetup(void)
     memset(&alpd_ctx, 0, sizeof(alpd_ctx));
 
     uint16_t spm_matcher = SinglePatternMatchDefaultMatcher();
+    uint16_t mpm_matcher = PatternMatchDefaultMatcher();
+
     alpd_ctx.spm_global_thread_ctx = SpmInitGlobalThreadCtx(spm_matcher);
     if (alpd_ctx.spm_global_thread_ctx == NULL) {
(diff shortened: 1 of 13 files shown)


inliniac:
Merged through #2115, thanks Justin!

I updated the DrMemory suppressions: https://github.com/inliniac/suricata/pull/2115/commits/333f2cb3107fc05859cafff7a3d1127795fbc541
And also fixed the broken use of a global mpm thread ctx in the SMTP code: https://github.com/inliniac/suricata/pull/2115/commits/eec66c7b4f7d8d3272d8229c9164f0b4c4955426 I guess that never hurt before as no mpm algo implementation did anything (critical) with the thread ctx.

PR #2106, opened 2016-05-25
netmap: close sw ring before hw rings

gureedo:
Fix issue https://redmine.openinfosecfoundation.org/issues/1714


inliniac:
Merged, thanks Aleksey!

PR #2102, opened 2016-05-24
Dev feat509 hashing v0.7

serializingme:
New version of the pull request #2059 adding common configurations parsing code. Was rebased to the latest master.

PR #2109, opened 2016-05-26
Dev feat509 hashing v0.8

serializingme:
New version of the pull request #2102 fixing a major bug where the detection function wasn't comparing the file hash (MD5, SHA1 or SHA256) with the correct in memory hashes (e.g., comparing the file MD5 hash with the SHA256 hashes). This was clear because it was matching files not having any of the specified hashes (in the case of a negated rule), when they did.

### DIFF
--- a/src/util-detect-file-hash.c
+++ b/src/util-detect-file-hash.c
@@ -1,0 +1,377 @@
+/* Copyright (C) 2007-2016 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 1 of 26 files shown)


src/util-detect-file-hash.c
```diff
@@ -0,0 +1,377 @@
...
+    if (s->file_flags & FILE_SIG_NEED_MD5 && file->flags & FILE_MD5) {
+        match = HashMatchHashTable(filehash->hash, file->md5, sizeof(file->md5));
+    }
+    else if (s->file_flags & FILE_SIG_NEED_SHA1 && file->flags & FILE_SHA1) {
+        match = HashMatchHashTable(filehash->hash, file->sha1, sizeof(file->sha1));
+    }
+    else if (s->file_flags & FILE_SIG_NEED_SHA256 && file->flags & FILE_SHA256) {
+        match = HashMatchHashTable(filehash->hash, file->sha256, sizeof(file->sha256));
+    }
+
```
serializingme: For those curious, the problem with #2102 was in the previous three conditions, where it wasn't checking to what hashing algorithm the signature refereed to.

inliniac:
Can you rename the commit 'Common code added' to be more descriptive on the subject line? E.g. 'file hashing: add common code'. That way while browsing commits/git blame it's more obvious what the commit is about at first glance.

serializingme:
Done, I have changed the messages to better align with the Suricata team commit format.

pevma:
I rerun the my previous tests - it looks all matching is good now on sha1/sha256/md5 when using white/black listings of the checksums. Thanks!

inliniac:
Can you do a rebase?

PR #2110, opened 2016-05-27
Dev app layer stats v9

regit:
This patchset adds flow and transaction counters for each application layers. The result is the following:
```JSON
    "app-layer": {
      "flow": {
        "http": 9310,
        "ftp": 0,
        "smtp": 0,
        "tls": 71,
        "ssh": 0,
        "imap": 0,
        "msn": 0,
        "smb": 170,
        "dcerpc_udp": 0,
        "dns_udp": 870,
        "dcerpc_tcp": 2,
        "dns_tcp": 0
      },
      "tx": {
        "dns_udp": 21433,
        "http": 12766,
        "smtp": 0,
        "dns_tcp": 0
      }
    },
```

This is a rework of #2046 with a new approach on transaction counters.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/172
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/168

PR #2111, opened 2016-05-27
Dev app layer stats v10

regit:
This patchset adds flow and transaction counters for each application layers. The result is the following:
```JSON
    "app-layer": {
      "flow": {
        "http": 9310,
        "ftp": 0,
        "smtp": 0,
        "tls": 71,
        "ssh": 0,
        "imap": 0,
        "msn": 0,
        "smb": 170,
        "dcerpc_udp": 0,
        "dns_udp": 870,
        "dcerpc_tcp": 2,
        "dns_tcp": 0
      },
      "tx": {
        "dns_udp": 21433,
        "http": 12766,
        "smtp": 0,
        "dns_tcp": 0
      }
    },
```

This is a rework of #2110 with a patch on tx counter to fix memory allocation error spotted by coccinelle tests.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/173
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/169

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -622,6 +622,11 @@ void AppLayerParserSetTransactionInspectId(ThreadVars *tv, AppLayerParserState *
         else
             break;
     }
+
+    /* account tx as soon as we have a request to server */
+    if (direction == 0) {
+        AppLayerIncTxCounter(ipproto, alproto, tv, idx - pstate->inspect_id[0]);
+    }
     pstate->inspect_id[direction] = idx;
 
     SCReturn;
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -599,7 +599,7 @@ uint64_t AppLayerParserGetTransactionInspectId(AppLayerParserState *pstate, uint
     SCReturnCT(pstate->inspect_id[direction & STREAM_TOSERVER ? 0 : 1], "uint64_t");
 }
 
-void AppLayerParserSetTransactionInspectId(AppLayerParserState *pstate,
+void AppLayerParserSetTransactionInspectId(ThreadVars *tv, AppLayerParserState *pstate,
                                            const uint8_t ipproto, const AppProto alproto,
                                            void *alstate, const uint8_t flags)
 {
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -149,6 +149,13 @@ struct AppLayerParserState_ {
  * Post 2.0 let's look at changing this to move it out to app-layer.c. */
 static AppLayerParserCtx alp_ctx;
 
+int AppLayerParserProtoIsRegistered(uint8_t ipproto, AppProto alproto)
+{
+    int ipproto_map = FlowGetProtoMapping(ipproto);
+
+    return (alp_ctx.ctxs[ipproto_map][alproto].StateAlloc != NULL) ? 1 : 0;
+}
+
 AppLayerParserState *AppLayerParserStateAlloc(void)
 {
     SCEnter();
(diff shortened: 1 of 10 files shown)


inliniac:
As discussed in IRC: In JQ it's hard to use this app-layer:
<pre>
$ cat tmp/eve.json |jq -c 'select(.stats)|.stats.app-layer'
jq: error: layer/0 is not defined at <top-level>, line 1:
select(.stats)|.stats.app-layer                          
jq: 1 compile error
</pre>
It can be done using:
<pre>
$ cat tmp/eve.json |jq -c 'select(.stats)|.stats."app-layer"'
{"flow":{"http":1624,"ftp":0,"smtp":0,"tls":51,"ssh":0,"imap":0,"msn":0,"smb":143,"dcerpc_udp":0,"dns_udp":301,"dcerpc_tcp":2,"dns_tcp":0},"tx":{"dns_udp":0,"http":0,"smtp":0,"tls":0,"dns_tcp":0}}
</pre>
Which is not pretty. So probably using "app_layer" would be better.

inliniac:
It seems that with --disable-detection, the tx counting is broken:
<pre>
{"flow":{"http":9311,"ftp":0,"smtp":0,"tls":71,"ssh":0,"imap":0,"msn":0,"smb":170,"dcerpc_udp":0,"dns_udp":878,"dcerpc_tcp":2,"dns_tcp":0},"tx":{"dns_udp":0,"http":0,"smtp":0,"tls":0,"dns_tcp":0}}
</pre>
When I do specify rules:
<pre>
{"flow":{"http":9311,"ftp":0,"smtp":0,"tls":71,"ssh":0,"imap":0,"msn":0,"smb":170,"dcerpc_udp":0,"dns_udp":885,"dcerpc_tcp":2,"dns_tcp":0},"tx":{"dns_udp":21455,"http":12767,"smtp":0,"tls":71,"dns_tcp":0}}
</pre>

regit:
Already updated branch to use app_layer.

regit:
On disable detection, I'm trying another place to add counters.

regit:
On disable detection, alternative is to update https://github.com/regit/suricata/blob/master/src/output-tx.c#L179 so we can increase the counter in the function. @inliniac would it be clean to do so ?

inliniac:
On 27-05-16 16:25, Eric Leblond wrote:
> On disable detection, alternative is to update
> https://github.com/regit/suricata/blob/master/src/output-tx.c#L179 so we
> can increase the counter in the function. @inliniac
> <https://github.com/inliniac> would it be clean to do so ?
> 

Seems to me that we need to increment counters when the actual event
happens, so in this case when a tx is created.

regit:
I see TX logging as a kind of logging if a TX is incomplete or malformed, should we account/logged it ?

inliniac:
On 27-05-16 16:31, Eric Leblond wrote:
> I see TX logging as a kind of logging if a TX is incomplete or
> malformed, should we account/logged it ?

I don't see how it makes sense to make tx counting depend on tx logging
which may or may not be enabled.

In (almost?) all of the code we count things where they happen. Not sure
why that should be different here.

regit:
One of the reason of this choice of implementation was that it was leading to a small patch compare to previous approach. I will try to see if I can cook something independent from output and detect.

inliniac:
On 27-05-16 17:21, Eric Leblond wrote:
> One of the reason of this choice of implementation was that it was
> leading to a small patch compare to previous approach. I will try to see
> if I can cook something independent from output and detect.

Perhaps it makes sense to introduce a generic 'new tx created'-hook and
then implement tx counting on top of that.

regit:
Yes, I was going to look after something similar.

PR #2113, opened 2016-05-27
Dev offloading v10

regit:
This is an attempt to implement offloading/shunting of flows in suricata. Currently it only features an NFQ implementation of offloading mechanism. I will work later on an EBPF one for AF_PACKET.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/175
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/171

### DIFF
--- a/src/util-device.h
+++ b/src/util-device.h
@@ -27,6 +27,7 @@ typedef struct LiveDevice_ {
     char *dev;  /**< the device (e.g. "eth0") */
     enum RunModes runmode; /**< the runmode (e.g. "RUNMODE_NFLOG") */
     int ignore_checksum;
+    void (*offload_callback)(Packet *p);
     SC_ATOMIC_DECLARE(uint64_t, pkts);
     SC_ATOMIC_DECLARE(uint64_t, drop);
     SC_ATOMIC_DECLARE(uint64_t, invalid_checksums);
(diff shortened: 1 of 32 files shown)


inliniac:
I think the term 'offload' is a bit unfortunate. It's way to broad and non-descriptive. I would prefer 'shunt' although I think it's not considered a nice word. I guess 'bypass' could be something, or perhaps 'flow bypass'. Iirc that is what we really mean right? Ask the capture method to stop sending us packets for a flow.

regit:
I agree on 'bypass'. That will be better.

inliniac:
I think we should hide this pattern in a function:
<pre>
if (p->livedev && p->livedev->offload_callback) {
    p->livedev->offload_callback(p);
}
</pre>

src/util-device.h
```diff
@@ -20,27 +20,31 @@
...
 #include "queue.h"
 #include "unix-manager.h"
+#include "runmodes.h"
 
 /** storage for live device names */
 typedef struct LiveDevice_ {
     char *dev;  /**< the device (e.g. "eth0") */
+    enum RunModes runmode; /**< the runmode (e.g. "RUNMODE_NFLOG") */
     int ignore_checksum;
+    void (*offload_callback)(Packet *p);
```
inliniac: Our naming for functions and function pointers means this should be OffloadCallback

inliniac:
This PR adds a lot of very verbose unittests with lots of code duplication. Can you guys clean that up as well? See 2c7cd1c22a6618dbc5207e927f3f99636c7fa204 as an example.

inliniac:
Commit message " offloading: add crypted part " doesn't make much sense to me. It speaks about 'crypted files' ?

PR #2114, opened 2016-05-27
Dev app layer stats v11

regit:
This is an update of #2111. The place where the transaction counter increase has been changed. It is now in AppLayerParserParse. And due to unittests, this implies there is far more changes involved as it was necessary to bring a ThreadVars to the function.

Other change is "app_layer" instead of "app-layer' in JSON to avoid any problem in the shell.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/176
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/172


regit:
Found an issue, closing.

PR #2116, opened 2016-05-29
App layer stats v13

regit:
Rework of #2114 fixing a memory handling issue.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/178
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/174

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -140,6 +140,8 @@ struct AppLayerParserState_ {
      * we don't need a var per direction since we don't log a transaction
      * unless we have the entire transaction. */
     uint64_t log_id;
+    /* Count of number of transactions */
+    uint64_t tx_cnt;
 
     /* Used to store decoder events. */
     AppLayerDecoderEvents *decoder_events;
@@ -624,6 +626,7 @@ void AppLayerParserSetTransactionInspectId(AppLayerParserState *pstate,
         else
             break;
     }
+
     pstate->inspect_id[direction] = idx;
 
     SCReturn;
@@ -1004,6 +1007,14 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow 
         }
     }
 
+    if (AppLayerParserProtocolIsTxEventAware(f->proto, alproto)) {
+        uint64_t p_tx_cnt = pstate->tx_cnt;
+        pstate->tx_cnt = AppLayerParserGetTxCnt(f->proto, alproto, f->alstate);
+        if (tv) {
+            AppLayerIncTxCounter(f->proto, alproto, tv,
+                                 pstate->tx_cnt - p_tx_cnt);
+        }
+    }
     /* next, see if we can get rid of transactions now */
     AppLayerParserTransactionsCleanup(f);
 
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -149,6 +149,13 @@ struct AppLayerParserState_ {
  * Post 2.0 let's look at changing this to move it out to app-layer.c. */
 static AppLayerParserCtx alp_ctx;
 
+int AppLayerParserProtoIsRegistered(uint8_t ipproto, AppProto alproto)
+{
+    int ipproto_map = FlowGetProtoMapping(ipproto);
+
+    return (alp_ctx.ctxs[ipproto_map][alproto].StateAlloc != NULL) ? 1 : 0;
+}
+
 AppLayerParserState *AppLayerParserStateAlloc(void)
 {
     SCEnter();
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -906,7 +906,7 @@ int AppLayerParserSetTxDetectState(uint8_t ipproto, AppProto alproto,
 
 /***** General *****/
 
-int AppLayerParserParse(AppLayerParserThreadCtx *alp_tctx, Flow *f, AppProto alproto,
+int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow *f, AppProto alproto,
                         uint8_t flags, uint8_t *input, uint32_t input_len)
 {
     SCEnter();
@@ -1298,7 +1298,7 @@ int AppLayerParserRequestFromFile(AppProto alproto, char *filename)
             }
             //PrintRawDataFp(stdout, buffer, result);
 
-            (void)AppLayerParserParse(alp_tctx, f, alproto, flags, buffer, result);
+            (void)AppLayerParserParse(NULL, alp_tctx, f, alproto, flags, buffer, result);
             if (done)
                 break;
         }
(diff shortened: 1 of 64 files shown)


src/app-layer-parser.c
```diff
@@ -149,6 +149,13 @@ struct AppLayerParserState_ {
  * Post 2.0 let's look at changing this to move it out to app-layer.c. */
 static AppLayerParserCtx alp_ctx;
 
+int AppLayerParserProtoIsRegistered(uint8_t ipproto, AppProto alproto)
+{
+    int ipproto_map = FlowGetProtoMapping(ipproto);
```
inliniac: type should be uint8_t like elsewhere

src/app-layer.c
```diff
@@ -66,6 +67,16 @@ struct AppLayerThreadCtx_ {
 #endif
 };
 
+typedef struct AppLayerCounters_ {
+    char *name;
```
inliniac: const?

src/app-layer.c
```diff
@@ -80,6 +91,14 @@ static inline int ProtoDetectDone(const Flow *f, const TcpSession *ssn, uint8_t
             (FLOW_IS_PM_DONE(f, direction) && FLOW_IS_PP_DONE(f, direction)));
 }
 
+static void AppLayerIncFlowCounter(uint8_t ipproto, AppProto alproto, ThreadVars *tv)
+{
+    if (tv) {
```
inliniac: I think tv will only be NULL in unittests, so it may be worth only doing this check then. Otherwise, add 'likely'

src/app-layer.c
```diff
@@ -648,6 +678,80 @@ void AppLayerRegisterGlobalCounters(void)
...
+    AppLayerProtoDetectSupportedAppProtocols(alprotos);
+
+    for (alproto = 0; alproto < ALPROTO_MAX; alproto++) {
+        if (alprotos[alproto] == 1) {
+            char *str = "app_layer.flow.";
+            char *alproto_str = AppLayerGetProtoName(alproto);
+            int alproto_len = strlen(alproto_str) + 1;
+            uint8_t ipproto_map = FlowGetProtoMapping(ipproto);
+
+            if (AppLayerParserProtoIsRegistered(ipproto, alproto) &&
```
inliniac: What is this about 'other proto'?

src/app-layer-parser.c
```diff
@@ -140,6 +140,8 @@ struct AppLayerParserState_ {
      * we don't need a var per direction since we don't log a transaction
      * unless we have the entire transaction. */
     uint64_t log_id;
+    /* Count of number of transactions */
+    uint64_t tx_cnt;
```
inliniac: Why does this need to be part of the state? It's only used in AppLayerParserParse so it can be a local variable.

src/app-layer-parser.c
```diff
@@ -1004,6 +1007,14 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow
         }
     }
 
+    if (AppLayerParserProtocolIsTxEventAware(f->proto, alproto)) {
```
inliniac: did you mean AppLayerParserProtocolIsTxAware ?

src/app-layer.c
```diff
@@ -99,6 +101,18 @@ static void AppLayerIncFlowCounter(uint8_t ipproto, AppProto alproto, ThreadVars
     }
 }
 
+void AppLayerIncTxCounter(uint8_t ipproto, AppProto alproto, ThreadVars *tv,
```
inliniac: I think every function that has Threadvars as an argument puts it first in the arg list. Might be good to do that here as well.

src/app-layer.c
```diff
@@ -99,6 +101,18 @@ static void AppLayerIncFlowCounter(uint8_t ipproto, AppProto alproto, ThreadVars
...
+                          uint64_t step)
+{
+    if (!AppLayerParserProtocolIsTxAware(ipproto, alproto))
+        return;
+
+    if (tv) {
```
inliniac: tv should never be NULL unless for unittests, so I think this can be optimized

src/app-layer.c
```diff
@@ -99,6 +101,18 @@ static void AppLayerIncFlowCounter(uint8_t ipproto, AppProto alproto, ThreadVars
     }
 }
 
+void AppLayerIncTxCounter(uint8_t ipproto, AppProto alproto, ThreadVars *tv,
+                          uint64_t step)
+{
+    if (!AppLayerParserProtocolIsTxAware(ipproto, alproto))
```
inliniac: if we already check this prior to calling this function, then we can probably leave this check out here

src/app-layer.c
```diff
@@ -706,6 +721,16 @@ void AppLayerSetupCounters(uint8_t ipproto)
                 snprintf(applayer_counters[ipproto_map][alproto].name,
                          strlen(str) + alproto_len + strlen(ipproto_suffix),
                          "%s%s%s", str, alproto_str, ipproto_suffix);
+                if (AppLayerParserProtocolIsTxAware(ipproto, alproto)) {
+                    applayer_counters[ipproto_map][alproto].tx_name =
+                        SCMalloc(strlen(tx_str) + alproto_len + strlen(ipproto_suffix));
```
inliniac: please use a 'size' variable here instead of "strlen(tx_str) + alproto_len + strlen(ipproto_suffix)" multiple times

src/app-layer.c
glongo: 'other proto' is the opposite of ipproto. It's used to check if handle the case that a proto is registered for tcp and udp.

src/app-layer-parser.c
glongo: ack

src/app-layer.c
glongo: ack

src/app-layer.c
glongo: ack

src/app-layer.c
glongo: ack

src/app-layer-parser.c
```diff
@@ -1004,6 +1007,14 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow
...
+        uint64_t p_tx_cnt = pstate->tx_cnt;
+        pstate->tx_cnt = AppLayerParserGetTxCnt(f->proto, alproto, f->alstate);
+        if (tv) {
```
glongo: would be good to add 'likely' here also?

src/app-layer.c
glongo: ack

src/app-layer.c
glongo: ack

src/app-layer-parser.c
regit: yes

src/app-layer-parser.c
regit: In fact, what the code is adding to the tv counter at each pass in the parser the difference of tx on a flow between now and the previous pass. So we take into account the new TX and we have to keep the tx_cnt between the pass.

src/app-layer.c
regit: We construct the name later and we use it in StatsRegisterCounter which is not currently used a const so this would require some patching in these.

src/app-layer-parser.c
inliniac: I don't follow this. We can take the current tx_cnt at the start of AppLayerParse, and again at the end of it. Then consider the diff to be the new tx number?

PR #2117, opened 2016-05-30
logging - do not use thread modules for logging - v1

jasonish:

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/240
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/245


inliniac:
Looks good overall, but breaks my lua logging tests. Will send you a test case.

inliniac:
I guess the next step would be to get rid of the higher level logging thread modules, like TMM_TXLOGGER so that we have a single (non-thread module) function that we can call to trigger all the logging.

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2126

PR #2123, opened 2016-05-31
Unix socket auto

regit:
Implement 'auto' value for unix socket command configuration to have it activated when running in live mode.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/180
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/176

### DIFF
--- a/src/counters.c
+++ b/src/counters.c
@@ -244,10 +244,7 @@ static void StatsInitCtx(void)
 
         /* if the unix command socket is enabled we do the background
          * stats sync just in case someone runs 'dump-counters' */
-        int unix_socket = 0;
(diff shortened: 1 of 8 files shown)


inliniac:
I've been thinking about whether this use of the Time API is a proper way to query the engine's run mode. I tend to think it may be better to handle this more globally, in the instance perhaps.

inliniac:
whoops didn't mean to close just yet

inliniac:
Merged in #2250, so this time I can close it :)

PR #2126, opened 2016-05-31
logging - do not use thread modules for logging - v3

jasonish:
Previous PR: #2117 

Changes from last PR:
- Fix lua output.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/241
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/246


jasonish:
Closing. Accidentally pushed commits against this branch.

PR #2130, opened 2016-06-01
[preview] output cleanup v4

inliniac:
Add 2 new log levels: PERF and CONFIG. SCLogConfig is used to log configuration settings, like in the stream engine. SCLogPerf would be used to things related to performance, like profiling, afpacket info, etc.

Goal is to shrink the default output.

By default:
-v info
-vv perf
-vvv config


inliniac:
Merged through #2136

PR #2138, opened 2016-06-04
detect-filemagic: fix heap-use-after-free

norg:
This fixes the heap-use-after-free issue with sm being freed without
being removed from the signature (s) list. Move the protocol check for
rules with filemagic before the alloc and make the error log more
precise.

This fixes https://redmine.openinfosecfoundation.org/issues/1584

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/24
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/24


inliniac:
Thanks Andreas!

PR #2141, opened 2016-06-08
(preview) logging - do not use thread modules for logging - v4

jasonish:
Previous PR: #2126, #2117 

Converts most loggers, those that didn't have a Func on the thread module to non-thread modules.

As profiling was based on the thread module ID, new profiling support had to be added which is currently the only reason for the enum giving each logger its own ID.

As loggers are no longer thread modules, rename and move registration to output.c.

Provide a single entry point for all logging. Currently a thread module for logging, but could be changed so it can be called by the flow worker.

Could still probably use some naming cleanup.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/254
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/259

PR #2142, opened 2016-06-09
logging - do not use thread modules for logging - v5

jasonish:
Previous PRs: #2141, #2126, #2117

Could still probably use some naming cleanup

Changes from last PR:
- Free some memory leaks.
- Fix file and file data loggers.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/255
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/260


jasonish:
Closing. Accidentally pushed against this branch. v6 is almost ready for preview.

PR #2144, opened 2016-06-09
TEST check if zero copy works correcty

inliniac:
This is TEST/DEBUG code to validate that the buffer we get from packet
capture (p->ext_pkt in 0copy) remains valid.

The idea is that at the end of the lifetime of the packet we check if
the Ethernet pointer into ext_pkt still contains the same data as our
local copy of it.

With PF_RING it has been observed that this is not the case, which is
a violation of our expectations. We expect the pointer to remain valid
until the next call to pfring_recv.


inliniac:
I'll revive this if we ever suspect 0copy issues again.

PR #2146, opened 2016-06-11
configure: enable nfqueue by default and set flags

norg:
This change enables nfqueue by default if the needed lib is installed
and also sets the correct CPPFLAGS which is necessary for some systems
like OpenSuse.

This should solve: https://redmine.openinfosecfoundation.org/issues/1525

But fails on our buildbot due to old libnetfilter_queue version:

Build failure for norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/25
Build failure for norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/25


inliniac:
Back to the drawing board.

PR #2153, opened 2016-06-14
util-threshold-config: parse suppress rules with spaces in ip list

norg:
This modified regex allows spaces witihn the ip list for supress rules
like [10.0.0.1, 10.0.0.2].

This would also "fix" the issue from https://redmine.openinfosecfoundation.org/issues/1515

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/27
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/27


inliniac:
Merged through #2160, thanks Andreas!

PR #2157, opened 2016-06-18
Next 30x v1

inliniac:
@regit @norg @jasonish please have a look if anything is missing. This is to be 3.0.2

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1237,7 +1237,7 @@
             HTP_LDADD="../libhtp/htp/libhtp.la"
             AC_SUBST(HTP_LDADD)
             # make sure libhtp is added to the includes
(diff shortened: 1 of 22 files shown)


norg:
Looks fine to me, all my other commits should be just 3.1

pevma:
@norg  - ok , cause i was about to ask if we didn't miss  - https://github.com/inliniac/suricata/pull/2154

regit:
I think we could add 4defc5acc21e92b53d6993e6133871e3636a9193 which is fixing a problem for some setup and should not have any negative impact.

jasonish:
Looks fine to me. Anything I have is pending for 3.1 as well.

inliniac:
Great points, added those in #2158

PR #2161, opened 2016-06-20
[preview] disable offloading v1

inliniac:
First stab at disabling offloading through ioctl commands instead of just warning on the settings. Implemented for FreeBSD (ifcaps) and Linux (ethtool).

TODO:
- testing
- make configurable

cc: @regit @gureedo 

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/459
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/464

### DIFF
--- a/src/util-device.c
+++ b/src/util-device.c
@@ -39,6 +39,23 @@ static int live_devices_stats = 1;
 static int LiveSafeDeviceName(const char *devname,
                               char *newdevname, size_t destlen);
 
+static int g_live_devices_disable_offloading = 1;
+
+void LiveSetOffloadDisable(void)
+{
+    g_live_devices_disable_offloading = 1;
+}
+
+void LiveSetOffloadWarn(void)
+{
+    g_live_devices_disable_offloading = 0;
+}
+
+int LiveGetOffload(void)
+{
+    return g_live_devices_disable_offloading;
+}
+
 /**
  *  \brief Add a pcap device for monitoring
  *
--- a/src/util-ioctl.c
+++ b/src/util-ioctl.c
@@ -236,6 +236,33 @@ int GetIfaceCaps(const char *ifname)
     return ifr.ifr_curcap;
 }
 #endif
+#ifdef SIOCSIFCAP
+int SetIfaceCaps(const char *ifname, int caps)
+{
+    struct ifreq ifr;
+
+    int fd = socket(AF_INET, SOCK_DGRAM, 0);
+    if (fd < 0) {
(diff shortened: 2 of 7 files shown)


src/util-ioctl.c
```diff
@@ -345,6 +399,62 @@ static int GetIfaceOffloadingLinux(const char *dev, int csum, int other)
...
+        if (GetEthtoolValue(dev, ETHTOOL_GSG, &value) == 0 && value != 0) {
+            SCLogInfo("%s: disabling sg offloading", dev);
+            SetEthtoolValue(dev, ETHTOOL_SSG, 0);
+        }
+#endif
+#ifdef ETHTOOL_GFLAGS
+        if (GetEthtoolValue(dev, ETHTOOL_GFLAGS, &value) == 0) {
+            if (value & ETH_FLAG_LRO) {
+                SCLogInfo("%s: disabling lro offloading", dev);
+                SetEthtoolValue(dev, ETHTOOL_SSG, value & ~ETH_FLAG_LRO);
```
inliniac: whoops wrong command

pevma:
These could also be considered - udp-fragmentation-offload, rx-vlan-offload, tx-vlan-offload

src/util-device.c
```diff
@@ -39,6 +39,23 @@ static int live_devices_stats = 1;
 static int LiveSafeDeviceName(const char *devname,
                               char *newdevname, size_t destlen);
 
+static int g_live_devices_disable_offloading = 1;
```
regit: I'm really ok with the global idea as running ethtool is painful. But if we disable offloading by default we have an impact on the system and returning to initial state is currently not done. So the question is, do we have user that could run from time to time suricata and suffer from performance loss due to the removal of offloading. I don't see much cases but we may have an important one which is firewall with optional IDS capability.

inliniac: On 21-06-16 08:28, Eric Leblond wrote:
> I'm really ok with the global idea as running ethtool is painful. But if
> we disable offloading by default we have an impact on the system and
> returning to initial state is currently not done. So the question is, do
> we have user that could run from time to time suricata and suffer from
> performance loss due to the removal of offloading. I don't see much
> cases but we may have an important one which is firewall with optional
> IDS capability.

Not 100% sure how important this really is, but I'll cook up some logic
to restore the original settings at exit.

inliniac:
Replaced by #2162

PR #2165, opened 2016-07-05
app-layer-parser: fix segfault

decanio:
app-layer-parser: fix segfault that had occured when a protocol's eve logger
is enabled, but its app layer parser is disabled.

- PR decanio: https://buildbot.openinfosecfoundation.org/builders/decanio/builds/23
- PR decanio-pcap: https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/23

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -819,8 +819,9 @@ int AppLayerParserGetStateProgressCompletionStatus(AppProto alproto,
 {
     SCEnter();
     int r = 0;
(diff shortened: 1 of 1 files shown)


inliniac:
I don't think this is the proper fix. The issue is not that there is a crash in this scenario, but rather that we register a logger for a disabled protocol. When a protocol is disabled non of the callbacks are registered.

I'm currently testing a fix where we don't register the logger in this case.

PR #2167, opened 2016-07-07
(preview) logging - do not use thread modules for logging - v6

jasonish:
Previous PRs: Previous PRs: #2142, #2141, #2126, #2117

Could still probably use some naming cleanup

Changes from last PR:
- Hook into flow worker

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/266
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/271


src/flow-worker.c
```diff
@@ -195,9 +195,10 @@ TmEcode FlowWorker(ThreadVars *tv, Packet *p, void *data, PacketQueue *preq, Pac
                 Detect(tv, x, detect_thread, NULL, NULL);
                 FLOWWORKER_PROFILING_END(x, PROFILE_FLOWWORKER_DETECT);
             }
-#if 0
+
             //  Outputs
-#endif
+            OutputLoggerLog(tv, x, fw->output_thread, preq, unused);
```
inliniac: it seems a bit odd to pass on 'unused' here, as that makes it 'used' :)

If we need it, then we should rename it. Otherwise, lets replace it's use here by NULL or remove the argument from the OutputLoggerLog func completely.

src/flow-worker.c
```diff
@@ -219,11 +220,15 @@ TmEcode FlowWorker(ThreadVars *tv, Packet *p, void *data, PacketQueue *preq, Pac
...
-#if 0
-    // Outputs
 
-    // StreamTcpPruneSession (from TmqhOutputPacketpool)
-#endif
+    // Outputs.
+    OutputLoggerLog(tv, p, fw->output_thread, preq, unused);
+
+    /*  Release tcp segments. Done here after alerting can use them. */
+    if (p->flow != NULL && p->proto == IPPROTO_TCP) {
```
inliniac: With this in place can you update the comment block on the top of this file?

src/log-pcap.c
```diff
@@ -171,6 +168,11 @@ void TmModulePcapLogRegister(void)
     (prof).total += (UtilCpuGetTicks() - pcaplog_profile_ticks); \
     (prof).cnt++
 
+static int PcapLogCondition(ThreadVars *tv, const Packet *p)
+{
+    return TRUE;
```
inliniac: I'm fine with this for now, but in the longer term we need to update this. At the start of the actual log function we have:
<pre>
    PcapLogThreadData *td = (PcapLogThreadData *)thread_data;
    PcapLogData *pl = td->pcap_log;

    if ((p->flags & PKT_PSEUDO_STREAM_END) ||
        ((p->flags & PKT_STREAM_NOPCAPLOG) &&
         (pl->use_stream_depth == USE_STREAM_DEPTH_ENABLED)) ||
        (IS_TUNNEL_PKT(p) && !IS_TUNNEL_ROOT_PKT(p)) ||
        (pl->honor_pass_rules && (p->flags & PKT_NOPACKET_INSPECTION)))
    {
        return TM_ECODE_OK;
    }
</pre>

src/output-stats.c
```diff
@@ -154,6 +150,9 @@ static TmEcode OutputStatsLogThreadInit(ThreadVars *tv, void *initdata, void **d
 
                 SCLogDebug("%s is now set up", logger->name);
             }
+        } else {
+            /* XXX Temporary. */
```
inliniac: is this still relevant?

src/suricata.c
```diff
@@ -597,6 +559,7 @@ void usage(const char *progname)
     printf("\t--erf-in <path>                      : process an ERF file\n");
 #ifdef HAVE_DAG
     printf("\t--dag <dagX:Y>                       : process ERF records from DAG interface X, stream Y\n");
+4ckphn43z
```
inliniac: heh?

src/output.c
```diff
@@ -18,9 +18,18 @@
 /**
  * \file
  *
+ * \author OISF, Jason Ish <jason.ish@oisf.net>
  * \author Endace Technology Limited, Jason Ish <jason.ish@endace.com>
  *
- * Output registration functions
+ * The root logging out for all non-application logging.
+ *
+ * The loggers are made up of a hierarchy of loggers. At the top we
```
inliniac: I think this info should be part of the commit as well.

src/output.c
```diff
@@ -769,3 +915,176 @@ void OutputNotifyFileRotation(void) {
...
+    OutputRegisterRootLoggers();
+    OutputRegisterLoggers();
+}
+
+/**
+ * \brief Register all root loggers.
+ */
+void OutputRegisterRootLoggers(void)
+{
+    OutputPacketLoggerRegister();
```
inliniac: I guess these can be made conditional? Only if a logger of a certain type is registered, register the root logger as well?

src/output.c
```diff
@@ -769,3 +915,176 @@ void OutputNotifyFileRotation(void) {
         *(flag->flag) = 1;
     }
 }
+
+TmEcode OutputLoggerLog(ThreadVars *tv, Packet *p, void *thread_data,
+    PacketQueue *pq, PacketQueue *postpq)
```
inliniac: I think PacketQueue *pq, PacketQueue *postpq can both be removed. No output should be creating pseudo packets.

inliniac:
Can you submit a rebased PR? Then I will do a QA run.

src/flow-worker.c
jasonish: Its not used anywhere. I have removed it from all the logger functions.

src/output.c
jasonish: Good idea.  I put it on the commit where the root logger was introduced.

src/output.c
jasonish: Yes, thats the idea.  Wanted to keep some of that extra logic out for now. But will need to add it. Do you want it before the QA run?

src/output.c
jasonish: Done.

src/output.c
inliniac: On 13-07-16 19:11, Jason Ish wrote:
> Yes, thats the idea. Wanted to keep some of that extra logic out for
> now. But will need to add it. Do you want it before the QA run?
> 

No need, it's more like an optimization anyway.

jasonish:
Closed. Rebased and comments addressed here: https://github.com/inliniac/suricata/pull/2176

PR #2172, opened 2016-07-11
preview: directional logging - v2

jasonish:
Replaces PR #2170 with a PR that applies directly to master.

Take a different approach to per directional logging than the previous PRs did (#1992). Instead of touching the core tx logging logic, allow a logger name to be registered multiple times, where the TS and TC logging functions can be registered separately.

For DNS the result is that requests are logged before the response, and requests will be timestamped with the request timestamp, not the response timestamp.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/271
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/276

### DIFF
--- a/src/app-layer-dns-common.c
+++ b/src/app-layer-dns-common.c
@@ -226,7 +226,7 @@ int DNSGetAlstateProgress(void *tx, uint8_t direction)
     if (direction & STREAM_TOCLIENT) {
         /* response side of the tx is done if we parsed a reply
          * or if we tagged this tx as 'reply lost'. */
-        return (dns_tx->replied|dns_tx->reply_lost) ? 1 : 0;
+        return (dns_tx->replied|dns_tx->reply_lost) ? 2 : 1;
     }
     else {
         /* tx is only created if we have a complete request,
@@ -253,11 +253,11 @@ int DNSGetTxLogged(void *alstate, void *tx, uint32_t logger)
 
 /** \brief get value for 'complete' status in DNS
  *
- *  For DNS we use a simple bool. 1 means done.
+ *  For DNS 1 means request done, 2 means reply done (or lost).
  */
 int DNSGetAlstateProgressCompletionStatus(uint8_t direction)
 {
-    return 1;
+    return 2;
 }
 
 void DNSSetEvent(DNSState *s, uint8_t e)
--- a/src/output-json-dns.c
+++ b/src/output-json-dns.c
@@ -274,7 +274,8 @@ static void LogAnswers(LogDnsLogThread *aft, json_t *js, DNSTransaction *tx, uin
 
 }
 
-static int JsonDnsLogger(ThreadVars *tv, void *thread_data, const Packet *p, Flow *f, void *alstate, void *txptr, uint64_t tx_id)
+static int JsonDnsLoggerToServer(ThreadVars *tv, void *thread_data,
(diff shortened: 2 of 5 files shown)


src/app-layer-dns-common.c
```diff
@@ -253,11 +253,11 @@ int DNSGetTxLogged(void *alstate, void *tx, uint32_t logger)
 
 /** \brief get value for 'complete' status in DNS
  *
- *  For DNS we use a simple bool. 1 means done.
+ *  For DNS 1 means request done, 2 means reply done (or lost).
  */
 int DNSGetAlstateProgressCompletionStatus(uint8_t direction)
 {
-    return 1;
+    return 2;
```
inliniac: why does this need to change? We can just use 0 not ready and 1 ready for both sides, right?

jasonish: No, because with WithProgress we want it to fire in specific states, one where that state is 0.  If we use 0 as a progress state when registering, it does not get accepted as 0 due to this: 

https://github.com/inliniac/suricata/blob/master/src/output-tx.c#L88

So while want to get called when tc is 0, and ts is 1, the above will force it to be 1 and 1. So implementing tri-state for the specific protocol seemed easier then changing the tx handling logic.

inliniac: On 11-07-16 20:29, Jason Ish wrote:
> In src/app-layer-dns-common.c
> <https://github.com/inliniac/suricata/pull/2172#discussion_r70310964>:
> 
>>   */
>>  int DNSGetAlstateProgressCompletionStatus(uint8_t direction)
>>  {
>> -    return 1;
>> +    return 2;
> 
> No, because with WithProgress we want it to fire in specific states, one
> where that state is 0. If we use 0 as a progress state when registering,
> it does not get accepted as 0 due to this:
> 
> https://github.com/inliniac/suricata/blob/master/src/output-tx.c#L88
> 
> So while want to get called when tc is 0, and ts is 1, the above will
> force it to be 1 and 1. So implementing tri-state for the specific
> protocol seemed easier then changing the tx handling logic.

I see. So the bug/issue is actually hidden. I think we should try to fix
that registation thing. The progress stuff is hard enough w/o having to
account for unrelated issues as well :)

jasonish: As all registration goes through functions, I think it would be easy to make -1 be the value that means assign it in OutputRegisterTxLogger, thus making 0 a valid value when registered WIthProgress().

inliniac: On 11-07-16 20:36, Jason Ish wrote:
> In src/app-layer-dns-common.c
> <https://github.com/inliniac/suricata/pull/2172#discussion_r70312347>:
> 
>>   */
>>  int DNSGetAlstateProgressCompletionStatus(uint8_t direction)
>>  {
>> -    return 1;
>> +    return 2;
> 
> As all registration goes through functions, I think it would be easy to
> make -1 be the value that means assign it in OutputRegisterTxLogger,
> thus making 0 a valid value when registered WIthProgress().

Simple enough, lets try that.

jasonish: Ok, works. Seems low risk. In prscript right now.

jasonish:
Closed. See https://github.com/inliniac/suricata/pull/2173

PR #2175, opened 2016-07-13
Config option for disabling libhtp decompression

WGH-:
Suricata has a config option to configure libhtp decompression layer limit, but no option to disable decompression altogether. 0 is interpreted as no limit, and negative values are not accepted, either.

This pull request adds such option.


inliniac:
Why would you want to disable this?

WGH-:
When you pass reconstructed HTTP sessions to other tools, `Content-Encoding` header is preserved, which may lead to some confusion. It can be difficult to determine whether content has been successfully decompressed or it has failed for some reason (it wasn't compressed to begin with, decompression layer limit has been hit, data was corrupted, and so on).

Additionally, although it may sound as a rather uncommon use case, you may want to examine additional metadata in gzip, like original file name.

inliniac:
Documentation update is missing. Can you resubmit the PR?

PR #2177, opened 2016-07-13
Issue 1231 - SSL Negation (Snort compatibility) - v5

jasonish:
Previous PRs: #1146, #1147, #1474, #2043, #2044 

Redmine issue: https://redmine.openinfosecfoundation.org/issues/1231

First this required the current state being passed to the ssl_state match function instead of the cumulative state, otherwise ssl_state:client_hello would also match while in the state server_hello. Note that this patch is a bit naive as I don't know the SSL protocol that well, so its more of a preview or RFC.

This patch accepts "," as a state separator to be compatible with Snort, but keeps "|" as a separator for compatibility with existing Suricata rules.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/275
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/280

CC: @thus

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -244,7 +244,7 @@ static int SSLv3ParseHandshakeType(SSLState *ssl_state, uint8_t *input,
 
     switch (ssl_state->curr_connp->handshake_type) {
         case SSLV3_HS_CLIENT_HELLO:
-            ssl_state->flags |= SSL_AL_FLAG_STATE_CLIENT_HELLO;
+            ssl_state->current_flags = SSL_AL_FLAG_STATE_CLIENT_HELLO;
 
             /* skip version */
             input += SSLV3_CLIENT_HELLO_VERSION_LEN;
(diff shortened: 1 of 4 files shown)


inliniac:
@thus, any feedback on this one? Thanks!

thus:
I've taken a quick look, and there seems to be some issues.

**Direction issues**
(matching the following packet instead of the current)

There seems to be some kind of direction issue. The following rule should match the client_hello packet sent from the client:
```
alert tls any any -> any any (msg:"TLS client_hello test"; ssl_state:client_hello; sid:123;)
```
Instead, it matches the TCP ACK that the server sends in reply to the client_hello. When running Suricata with --simulate-ips it matches the client_hello packet as it should, but it also matches the ACK from the server.

This also applies to the other "SSL states". When trying to match a server_hello, we instead match the ACK from the client, and with --simulate-ips we get both the server_hello from the server (as we should) and the ACK from the client.

**Not resetting the state**

Another issue is that the "SSL state" is not reset when parsing a new TLS record. This is a issue because not all TLS records sets the "SSL state". Because of this, the old state is still set when parsing new records, leading to matches on the wrong packets. One example of this is when using ssl_state:client_keyx, we also get a match on "change cipher spec", since it doesn't set the state.

jasonish:
> I've taken a quick look, and there seems to be some issues.
> 
> Direction issues
> (matching the following packet instead of the current)
> 
> There seems to be some kind of direction issue. The following rule should match the client_hello packet sent from the client:
> 
> alert tls any any -> any any (msg:"TLS client_hello test"; ssl_state:client_hello; sid:123;)
> Instead, it matches the TCP ACK that the server sends in reply to the client_hello. When running Suricata with --simulate-ips it matches the client_hello packet as it should, but it also matches the ACK from the server.
> 
> This also applies to the other "SSL states". When trying to match a server_hello, we instead match the ACK from the client, and with --simulate-ips we get both the server_hello from the server (as we should) and the ACK from the client.
> 

I don't think this is anything specific with this ssl support.  That is how Suricata IDS mode works - the alert won't trigger until the alerting packets are acked.  As you saw, the alert is more immediate in IPS mode.

> Not resetting the state
> 
> Another issue is that the "SSL state" is not reset when parsing a new TLS record. This is a issue because not all TLS records sets the "SSL state". Because of this, the old state is still set when parsing new records, leading to matches on the wrong packets. One example of this is when using ssl_state:client_keyx, we also get a match on "change cipher spec", since it doesn't set the state.

Will look into this one. Thanks.

inliniac:
Jason did you look at the state issue?

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2261 with respect to the state issue.

PR #2178, opened 2016-07-13
logging - do not use thread modules for logging - v8

jasonish:
Previous PRs: #2176, #2167, #2142, #2141, #2126, #2117

Changes from last PR:
- Remove some flow locks in loggers (oops). Merged into existing commit where they should have been done already.
- Remove flow locks from the tcp stream segment foreach function as its only called from the logging modules.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/276
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/281


jasonish:
See https://github.com/inliniac/suricata/pull/2179

PR #2179, opened 2016-07-14
logging - do not use thread modules for logging - v9

jasonish:
Previous PRs: #2178, #2176, #2167, #2142, #2141, #2126, #2117

Changes from last PR:
- Remove flow locking in JSON drop log.
- Remove flow locking from Lua layer.

As profiling was based on the thread module ID, new profiling support had to be added which is currently the only reason for the enum giving each logger its own ID.

As loggers are no longer thread modules, rename and move registration to output.c.

Provide a single entry point for all logging. Currently a thread module for logging, but could be changed so it can be called by the flow worker.

And lastly call logging from the flow worker where the flow is already locked.

Could still probably use some naming cleanup.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/277
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/282

PR #2181, opened 2016-07-15
logging - do not use thread modules for logging - v10

jasonish:
Previous PRs: #2179, #2178, #2176, #2167, #2142, #2141, #2126, #2117

Changes from last PR:
- More lock cleanup in flowbit handling.

And lastly call logging from the flow worker where the flow is already locked.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/279
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/284


src/log-pcap.c
```diff
@@ -171,6 +168,11 @@ void TmModulePcapLogRegister(void)
     (prof).total += (UtilCpuGetTicks() - pcaplog_profile_ticks); \
     (prof).cnt++
 
+static int PcapLogCondition(ThreadVars *tv, const Packet *p)
+{
+    return TRUE;
```
inliniac: this seems too broad. E.g. pseudo packets should be excluded. Not every condition here https://github.com/inliniac/suricata/blob/master/src/log-pcap.c#L358 can be handled in this function though.

src/output-stats.c
```diff
@@ -170,15 +166,12 @@ static TmEcode OutputStatsLogThreadDeinit(ThreadVars *tv, void *thread_data)
...
-                    "TmModuleGetByName for %s failed", logger->name);
-            exit(EXIT_FAILURE);
-        }
-
-        if (tm_module->ThreadDeinit) {
-            tm_module->ThreadDeinit(tv, store->thread_data);
+        if (logger->ThreadDeinit) {
+            logger->ThreadDeinit(tv, store->thread_data);
+        } else {
+            /* XXX Temporary. */
```
inliniac: Debug?

src/output-tx.c
```diff
@@ -249,27 +263,22 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
  *  loggers */
 static TmEcode OutputTxLogThreadInit(ThreadVars *tv, void *initdata, void **data)
 {
+    SCLogNotice("OutputTxLogThreadInit");
```
inliniac: Debug -> SCEnter();

src/output-tx.c
```diff
@@ -249,27 +263,22 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data, PacketQ
...
     OutputLoggerThreadData *td = SCMalloc(sizeof(*td));
     if (td == NULL)
         return TM_ECODE_FAILED;
     memset(td, 0x00, sizeof(*td));
 
     *data = (void *)td;
+    SCLogNotice("Thread data at %p", td);
```
inliniac: Debug

src/output.c
```diff
@@ -768,3 +915,175 @@ void OutputNotifyFileRotation(void) {
...
     }
 }
+
+TmEcode OutputLoggerLog(ThreadVars *tv, Packet *p, void *thread_data)
+{
+    LoggerThreadStore *thread_store = (LoggerThreadStore *)thread_data;
+    RootLogger *logger = TAILQ_FIRST(&RootLoggers);
+    LoggerThreadStoreNode *thread_store_node = TAILQ_FIRST(thread_store);
+    while (logger && thread_store_node) {
+        if (logger->LogFunc != NULL) {
```
inliniac: in what case can this condition be false?

src/output.c
```diff
@@ -768,3 +915,175 @@ void OutputNotifyFileRotation(void) {
...
+
+    RootLogger *logger;
+    TAILQ_FOREACH(logger, &RootLoggers, entries) {
+
+        void *child_thread_data = NULL;
+        if (logger->ThreadInit != NULL) {
+            if (logger->ThreadInit(tv, initdata, &child_thread_data) == TM_ECODE_OK) {
+                LoggerThreadStoreNode *thread_store_node =
+                    SCCalloc(1, sizeof(*thread_store_node));
+                BUG_ON(thread_store_node == NULL);
```
inliniac: todo

src/suricata.c
```diff
@@ -597,6 +559,7 @@ void usage(const char *progname)
     printf("\t--erf-in <path>                      : process an ERF file\n");
 #ifdef HAVE_DAG
     printf("\t--dag <dagX:Y>                       : process ERF records from DAG interface X, stream Y\n");
+4ckphn43z
```
inliniac: don't forget to remove :)

src/util-profiling.c
```diff
@@ -87,6 +87,9 @@ SCProfilePacketData packet_profile_app_pd_data6[257];
 SCProfilePacketData packet_profile_detect_data4[PROF_DETECT_SIZE][257];
 SCProfilePacketData packet_profile_detect_data6[PROF_DETECT_SIZE][257];
 
+SCProfilePacketData packet_profile_log_data4[LOGGER_SIZE][257];
+SCProfilePacketData packet_profile_log_data6[LOGGER_SIZE][257];
```
inliniac: It looks like the update func doesn't handle the tunnel packets separately, so then the size can be 256 instead.

src/util-profiling.c
```diff
@@ -685,6 +691,58 @@ void SCProfilingDumpPacketStats(void)
...
+            if (pd->cnt == 0) {
+                continue;
+            }
+
+            FormatNumber(pd->tot, totalstr, sizeof(totalstr));
+            double percent = (long double)pd->tot /
+                (long double)total * 100;
+
+            fprintf(fp, "%-24s    IPv6     %3d  %12"PRIu64"     %12"PRIu64"   %12"PRIu64"  %12"PRIu64"  %12s  %-6.2f\n",
+                    PacketProfileLoggertIdToString(m), p, pd->cnt, pd->min, pd->max, (uint64_t)(pd->tot / pd->cnt), totalstr, percent);
```
inliniac: would it be hard to use the registration name instead?

inliniac:
Minor comments only. If you rebase and update I'll make this the start of the 3.2 dev branch soon.

src/output.c
jasonish: Probably never, just became a bit idiomatic to check on funcptrs before calling them - this one is probably the exception.

src/output.c
jasonish: Exit in this case do you think?

src/util-profiling.c
jasonish: Yes, but only because the output system is deregistered before the profiling output runs, so I lose my registration table to lookup the name by ID.

I suppose we could have a table mapping IDs to names as well, centralizing the creation of all the names that could be access by ID?

inliniac: On 24-08-16 16:04, Jason Ish wrote:
> In src/util-profiling.c
> <https://github.com/inliniac/suricata/pull/2181#discussion_r76062128>:
> 
>> +    }
>> +    for (m = 0; m < LOGGER_SIZE; m++) {
>> +        int p;
>> +        for (p = 0; p < 257; p++) {
>> +            SCProfilePacketData *pd = &packet_profile_log_data6[m][p];
>> +
>> +            if (pd->cnt == 0) {
>> +                continue;
>> +            }
>> +
>> +            FormatNumber(pd->tot, totalstr, sizeof(totalstr));
>> +            double percent = (long double)pd->tot /
>> +                (long double)total * 100;
>> +
>> +            fprintf(fp, "%-24s    IPv6     %3d  %12"PRIu64"     %12"PRIu64"   %12"PRIu64"  %12"PRIu64"  %12s  %-6.2f\n",
>> +                    PacketProfileLoggertIdToString(m), p, pd->cnt, pd->min, pd->max, (uint64_t)(pd->tot / pd->cnt), totalstr, percent);
> 
> Yes, but only because the output system is deregistered before the
> profiling output runs, so I lose my registration table to lookup the
> name by ID.
> 
> I suppose we could have a table mapping IDs to names as well,
> centralizing the creation of all the names that could be access by ID?

Let's safe it for later then.

PR #2182, opened 2016-07-23
rule-reload: remember pending USR2 signals

norg:
We did ignore additional USR2 signals while a rule-reload was running.
This changes the counter to be incremented with every additional USR2
signal so we don't ignore them anymore but it's still limited to prevent
huge overload or even overflow.

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/29
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/29

This solves https://redmine.openinfosecfoundation.org/issues/1277

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -303,7 +303,11 @@ static void SignalHandlerSigterm(/*@unused@*/ int sig)
  */
 static void SignalHandlerSigusr2(int sig)
 {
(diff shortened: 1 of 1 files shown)


inliniac:
Merged through #2250, thanks Andreas!

PR #2184, opened 2016-08-01
Issue 1854 - fix unified2 tag logging - v1

jasonish:
Redmine issue: https://redmine.openinfosecfoundation.org/issues/1854

tag: fix unified tagging and make unified2 compliant

Logging of tagged packets in unified2 was not working as the tag alert structure was never being initialized. Instead of fixing that, making logging of tagged packets the same as in Snort, where there is no event type for tagged packets, but instead the event_id, and event_second field are used to reference the event structure that initiated the tag.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/283
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/288

### DIFF
--- a/src/alert-unified2-alert.c
+++ b/src/alert-unified2-alert.c
@@ -693,6 +693,7 @@ error:
 static int Unified2PacketTypeAlert(Unified2AlertThread *aun, const Packet *p, uint32_t event_id, int stream)
 {
     int ret = 0;
+    DetectTagDataEntry *tag = p->tag;
 
     if (!(aun->unified2alert_ctx->flags & UNIFIED2_ALERT_FLAGS_EMIT_PACKET))
         return 1;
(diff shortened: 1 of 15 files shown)


inliniac:
I worry about synchronization issues. IIRC DetectTagDataEntry are owned by Host or Flow. A packet reference to it would need to be careful about simultaneous access/updates and also about the data going away because of garbage collection. How are these things handled?

jasonish:
Hmm, that did cross my mind.  Simple fix I suppose would be to copy the data that is required in order to attribute the packet back to the event ID.

Might be worth some extra discussion though.  Say you have 2 rules that end up tagging the same packets, the packets only get attributed to one of the alerts, the first one (event id, event_second).

For eve I was thinking about NOT attributing the packet records with a back reference.  External tooling could use the flow id, or address/port parameters to match up packet records of interest which means the current tag logic would work for eve.  This extra is just to be compatible with Snort's current unified2 output.

jasonish:
Simpler fix here: https://github.com/inliniac/suricata/pull/2190

PR #2187, opened 2016-08-09
app-layer-ssl: Validity dates from TLS certificates (v11)

thus:
Get validity dates (notBefore and notAfter) from TLS certificates.

New detection keywords (tls_cert_notbefore and tls_cert_notafter), added validity dates to extended JSON log and extended TLS log output, and new lua functions TlsGetCertNotBefore() and TlsGetCertNotAfter().

Updates:
- Renamed tls_notbefore to tls_cert_notbefore
- Renamed tls_notafter to tls_cert_notafter
- Added additional unit test

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/37
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/37


pevma:
This will very useful. Pending PR acceptance into master - please update the wiki with a few examples - rules and lua functions alike.

thus:
@pevma any ETA on when the migration to Sphinx for the user guide is done? Would it make sense to add a own repo for documentation to simplify the process of adding documentation (e.g. suricata-documentation) where more people could have write access?

pevma:
@thus  - Sorry i didn't explain it better before. It is going to be soon. At the moment we need to update both redmine and do a PR as described here - https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Converting_Wiki_Documentation_to_Sphinx

thus:
@pevma I know the drill, I just think the migration process is dragging out in time. The migration warning on the top of the pages in the user guide on the wiki are confusing users. The edit access on most of the pages are removed, so the content is just getting more and more outdated.

I think it would be better to just remove the entire user guide from the wiki and link to the new sphinx user guide (or use a redmine plugin to include it onto the site), but that is just my opinion :)

pevma:
@thus - understood. There is work in progress and this will be addressed hopefully by next week's end.

src/util-time.c
```diff
@@ -349,3 +370,135 @@ void CreateTimeString (const struct timeval *ts, char *str, size_t size)
...
+    oldlen = inlen = strlen(string);
+
+    /* Skip trailing whitespace */
+    while (inlen > 0 && isspace(string[inlen - 1]))
+        inlen--;
+
+    char *instr = NULL;
+
+    if (inlen < oldlen) {
+        instr = SCMalloc(inlen + 1);
```
inliniac: can we use a stack alloc'd string instead?

src/detect-tls-cert-validity.c
```diff
@@ -0,0 +1,1331 @@
...
+    else {
+        goto error;
+    }
+
+    sm->ctx = (void *)dd;
+
+    s->flags |= SIG_FLAG_APPLAYER;
+    s->alproto = ALPROTO_TLS;
+
+    SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_AMATCH);
```
inliniac: I'm trying to phase out AMATCH. Now that TLS uses TX (even if just one TX) I'd like to see it use the same logic as the DNS response (DE_STATE_FLAG_DNSRESPONSE_INSPECT). Pretty much a separate inspect engine for TLS responses. See DetectEngineInspectGenericList in detect-engine-dns.c

thus:
Replaced by #2259

PR #2194, opened 2016-08-20
decode: support Cisco Fabric Path / DCE - v1

jasonish:
[dce-dns-request.zip](https://github.com/inliniac/suricata/files/428377/dce-dns-request.zip)

Cisco Fabric Path is ethernet wrapped in an ethernet like header
with 2 extra bytes.  The ethernet type is in the same location
so the ethernet decoder can be used with some validation
for the extra length.

This is just simple validation and decode. No extra counters or invalid events specific to fabric path.

I am not able to share the original pcap I received, but was able to create this one.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/286
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/291

### DIFF
--- a/src/decode-ethernet.c
+++ b/src/decode-ethernet.c
@@ -85,6 +85,12 @@ int DecodeEthernet(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
             DecodeMPLS(tv, dtv, p, pkt + ETHERNET_HEADER_LEN,
                        len - ETHERNET_HEADER_LEN, pq);
             break;
+        case ETHERNET_TYPE_DCE:
+            if (len > ETHERNET_DCE_HEADER_LEN) {
+                DecodeEthernet(tv, dtv, p, pkt + ETHERNET_DCE_HEADER_LEN,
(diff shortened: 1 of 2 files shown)


src/decode-ethernet.c
```diff
@@ -85,6 +85,12 @@ int DecodeEthernet(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
             DecodeMPLS(tv, dtv, p, pkt + ETHERNET_HEADER_LEN,
                        len - ETHERNET_HEADER_LEN, pq);
             break;
+        case ETHERNET_TYPE_DCE:
+            if (len > ETHERNET_DCE_HEADER_LEN) {
```
inliniac: I think I would like to see an event if this condition is false. Otherwise we'll completely silently ignore packets w/o any logging or possibility to match on the packet.

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2203

PR #2196, opened 2016-08-23
eve: log tag packets as packet events - v2

jasonish:
Create a new eve event type, "packet" for logging packets that
are tagged as part of an event. The packet is still at the top
level to keep it consistent with alert event types.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/1830

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/287
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/292

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -175,6 +175,15 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
     json_object_set_new(js, "alert", ajs);
 }
 
+static void AlertJsonPacket(const Packet *p, json_t *js)
+{
+    unsigned long len = GET_PKT_LEN(p) * 2;
+    uint8_t encoded_packet[len];
+    Base64Encode((unsigned char*) GET_PKT_DATA(p), GET_PKT_LEN(p),
+        encoded_packet, &len);
+    json_object_set_new(js, "packet", json_string((char *)encoded_packet));
+}
+
 static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
 {
     MemBuffer *payload = aft->payload_buffer;
@@ -183,7 +192,7 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
 
     int i;
 
-    if (p->alerts.cnt == 0)
+    if (p->alerts.cnt == 0 && !(p->flags & PKT_HAS_TAG))
         return TM_ECODE_OK;
 
     json_t *js = CreateJSONHeader((Packet *)p, 0, "alert");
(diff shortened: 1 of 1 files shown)


src/output-json-alert.c
```diff
@@ -368,6 +377,14 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
     json_object_clear(js);
     json_decref(js);
 
+    if (p->flags & PKT_HAS_TAG) {
+        MemBufferReset(aft->json_buffer);
+        json_t *packetjs = CreateJSONHeader((Packet *)p, 0, "packet");
```
inliniac: should packetjs be checked for NULL here?

inliniac:
Wonder if this should be optional?

jasonish:
> Wonder if this should be optional?

I thought about it and would be happy to make it optional. But I think it should be enabled by default, to carry out any intentions of rules tag keywords unless told not to.

In Snort AFAIK, its not optional and always on.

inliniac:
On 23-08-16 17:28, Jason Ish wrote:
>     Wonder if this should be optional?
> 
> I thought about it and would be happy to make it optional. But I think
> it should be enabled by default, to carry out any intentions of rules
> tag keywords unless told not to.
> 
> In Snort AFAIK, its not optional and always on.

I don't object to enabled by default in new installs, just wonder if
this may break ppls existing setups in upgrades. I guess this would be
setups where ppl run eve and u2 and use tags.

jasonish:
> I don't object to enabled by default in new installs, just wonder if
> this may break ppls existing setups in upgrades. I guess this would be
> setups where ppl run eve and u2 and use tags.

Do you expect anything to break?  Tagging doesn't work for unified2 in the current release either.

But I can see a case for wanting to disable it - when you already have FPC.

inliniac:
On 23-08-16 17:49, Jason Ish wrote:
>     I don't object to enabled by default in new installs, just wonder if
>     this may break ppls existing setups in upgrades. I guess this would be
>     setups where ppl run eve and u2 and use tags.
> 
> Do you expect anything to break? Tagging doesn't work for unified2 in
> the current release either.

The only thing I can think of is that in an existing setup the log
volume to something like ELK would explode if suddenly the tag/packet
records are in eve as well.

> But I can see a case for wanting to disable it - when you already have FPC.

In such cases not adding 'tag' to rules should do it as well.

src/output-json-alert.c
jasonish: On Tue, Aug 23, 2016 at 9:15 AM, Victor Julien <notifications@github.com>
wrote:

> In src/output-json-alert.c
> <https://github.com/inliniac/suricata/pull/2196#discussion_r75887672>:
>
> > @@ -368,6 +377,14 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
> >      json_object_clear(js);
> >      json_decref(js);
> >
> > +    if (p->flags & PKT_HAS_TAG) {
> > +        MemBufferReset(aft->json_buffer);
> > +        json_t *packetjs = CreateJSONHeader((Packet *)p, 0, "packet");
>
> should packetjs be checked for NULL here?
>
Yep, it should.

jasonish:
> In such cases not adding 'tag' to rules should do it as well.

I guess I'm thinking more of the case where Talos/ET uses the tag keyword.

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2199

PR #2201, opened 2016-08-24
logging - do not use thread modules for logging - v12

jasonish:
Previous PRs: #2181, #2179, #2178, #2176, #2167, #2142, #2141, #2126, #2117

Changes from last PR:
- stricter condition on pcap-log
- debug output removed
- if 0 (dead) code removed
- better exit on output module init failure
- reduce size of profile array

And lastly call logging from the flow worker where the flow is already locked.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/294
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/299


inliniac:
Needs rebase.

PR #2203, opened 2016-08-26
decode: support Cisco Fabric Path / DCE - v2

jasonish:
Cisco Fabric Path is ethernet wrapped in an ethernet like header
with 2 extra bytes.  The ethernet type is in the same location
so the ethernet decoder can be used with some validation
for the extra length.

This PR adds an event for DCE packets that are too small.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/296
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/301


inliniac:
https://redmine.openinfosecfoundation.org/issues/1874

PR #2204, opened 2016-08-28
mail-store: add mail-store feature for logging email content to the f…

decanio:
mail-store: add mail-store feature for logging email content to the filesystem
in a manner similar to the existing file-store feature.

Enable with something like this:

  # output module to store extracted mail to disk
  #
  # The mails are stored to the log-dir in a format "mail.<id>" where <id> is
  # an incrementing number starting at 1. For each file "mail.<id>" a meta
  # file "mail.<id>.meta" is created.
- mail-store:
    enabled: yes      # set to yes to enable
    log-dir: mail    # directory to store the files
    force-mailstore: yes # force storing of all files
    #waldo: mail.waldo # waldo file to store the file_id across runs

prscript errors out.  Here's the build info:

You can watch build progress at https://buildbot.openinfosecfoundation.org/builders/decanio/builds/25
You can watch build progress at https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/25


inliniac:
I wonder if it would make more sense to add it as an option to the existing file store. Seems like this is just a lot of code duplication here.

PR #2197, opened 2016-08-23
[preview] prefilter engines v73

inliniac:
As #2192, #2193 and #2195, adding:
- profiling
- icmp seq and icmp id support
- fixes and cleanup

Profiling writes to packet_stats.log:
<pre>
Prefilter                        IP ver   Proto   cnt            min            max            avg            tot          %% 
--------------------             ------   -----   ----------     ------------   ------------   -----------    ---------    ---
payload                           IPv4       1           271             3541          78312         10678         2.9m  0.00  
payload                           IPv4       6       7889689              337       34196300         29919       236.1b  28.16 
payload                           IPv4      17         11389             2377         649994         21475       244.6m  0.03  
stream                            IPv4       6       7889689              359       21837652         23182       182.9b  21.82 
http_uri                          IPv4       6         79975              519        4225543         25672         2.1b  0.24  
http_raw_uri                      IPv4       6         79975              379        2749632          6334       506.6m  0.06  
http_header (request)             IPv4       6        446006             2124       66473748        104662        46.7b  5.57  
http_user_agent                   IPv4       6         79752              855        5356546         16164         1.3b  0.15  
http_raw_header (request)         IPv4       6         79339             1015       15886229         53390         4.2b  0.51  
http_method                       IPv4       6         79975              414         760744          5130       410.3m  0.05  
file_data (smtp)                  IPv4       6          9513              490           5260           910         8.7m  0.00  
file_data (http response)         IPv4       6        444447             1275       44328241         79206        35.2b  4.20  
http_client_body                  IPv4       6         79339             1511        3104952          4823       382.7m  0.05  
http_cookie (request)             IPv4       6        446006              671        7805244         13251         5.9b  0.71  
http_cookie (response)            IPv4       6        446006              560        6973052          3983         1.8b  0.21  
flags                             IPv4       6       7889689              306       19892565          2107        16.6b  1.98  
flow                              IPv4       6       3810940              292       15720078          1517         5.8b  0.69  
dsize                             IPv4       6       1392034              324        5260418          1139         1.6b  0.19  
dsize                             IPv4      17         11383              350        1034349          1174        13.4m  0.00  
http_raw_header (response)        IPv4       6        365108              899        9422963         16901         6.2b  0.74  
http_stat_code                    IPv4       6        366254              502        1383656          2300       842.4m  0.10  
payload                           IPv6       0           132              799          12464          1784       235.5k  0.00  
payload                           IPv6       6       2963833              338       66755456         39337       116.6b  13.91 
[...]

PR #2207, opened 2016-08-30
[preview] prefilter engines v76

inliniac:
Changes since #2197:
- rebase to master
- make 'packet header engines' optional: --set detect.prefilter.default=auto to enable, or explicitly use 'prefilter' keyword in rules
- add additional logic for prefilter engines for rules that match on special src/dst ports or alproto
- revert the removal of the syn specific non-mpm/non-prefilter list
- various performance optimizations and profiling options

Note: the profiling code itself is quite expensive

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/476
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/481


pevma:
Maybe a comment with the available options in yaml - "mpm/auto" and what they do or how they differ here - 
https://github.com/inliniac/suricata/pull/2207/files#diff-90

inliniac:
Yeah adding that. In short the setting 'mpm' is what has been the default all along. So Suricata prefilters on MPM. The setting 'auto' looks at the rules to see if it can add prefilter engines for keywords like dsize, flags, flow, icode, itype, etc.

src/detect-engine-prefilter.c
```diff
@@ -58,6 +58,31 @@
 static int PrefilterStoreGetId(const char *name);
 #endif
 
+/* Return true is the list is sorted smallest to largest */
```
gozzy: Hmm... It doesn't seem to return anything ;)

inliniac:
Replaced by #2212

src/detect-engine-prefilter.c
inliniac: Good point, removing.

PR #2215, opened 2016-09-02
stream: fix depth reached detection

regit:
When a segment only partially fit in streaming depth, the stream
depth reached flag was not set resulting in a continuous
inspection of the rest of the session.

I've seen this issue when working on shunting where I had continuous stream instead of having them shunt.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/184
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/180

### DIFF
--- a/src/stream-tcp-private.h
+++ b/src/stream-tcp-private.h
@@ -157,7 +157,9 @@ enum
 #define STREAMTCP_STREAM_FLAG_KEEPALIVE         0x0004
 /** Stream has reached it's reassembly depth, all further packets are ignored */
 #define STREAMTCP_STREAM_FLAG_DEPTH_REACHED     0x0008
-// vacancy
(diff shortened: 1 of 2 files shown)


inliniac:
Why do we need a new flag if all it does is set another flag? Maybe I'm just missing the point here.

regit:
If I remember correctly if we just set the depth reached flag then we won't inspect the current segment and it has to be done to honor the stream depth.
Digging more if it does not make it clear for you.

inliniac:
I think this needs  a couple of things:
1. ticket
2. better explanation of the issue
3. better explanation of the fix. Especially why we need an extra flag instead directly setting STREAMTCP_STREAM_FLAG_DEPTH_REACHED

PR #2216, opened 2016-09-03
About suricata pcap autofp mode

jarwin123:
hello,
    i use pcap mode capture packet only one device , i cann't  Start mulit thread in one NIC. i read suricata some code ,i found ` (void)ConfGet("netmap.live-interface", &live_dev)`; the live_dev alway return 0x0, i modify suricata.yaml ,but the result is the same , how can i modify suricata.yaml? thank you!


inliniac:
Pull requests are not support tickets. See https://redmine.openinfosecfoundation.org/projects/suricata/issues

PR #2217, opened 2016-09-05
Ruleset stats v17

glongo:

This addresses comments from the last PR, in particular:
- Remove DONE state (https://github.com/inliniac/suricata/commit/9f540448b3a4c1d165e9aafd3aa8c5c1282a2ecf)
- Block live reload when -s/-S is specified (https://github.com/inliniac/suricata/commit/ee6d592661feaa4e2854a0c5f414a5f25244221d)

Last PR: https://github.com/inliniac/suricata/pull/1998
Ticket: https://redmine.openinfosecfoundation.org/issues/1585

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/119
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/118

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -342,6 +342,7 @@ util-decode-asn1.c util-decode-asn1.h \
 util-decode-der.c util-decode-der.h \
 util-decode-der-get.c util-decode-der-get.h \
 util-decode-mime.c util-decode-mime.h \
+util-detect.c util-detect.h \
 util-device.c util-device.h \
 util-enum.c util-enum.h \
 util-error.c util-error.h \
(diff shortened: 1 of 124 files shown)


inliniac:
I'm not too excited about the mass change of SigInit and DetectEngineAppendSig. If such a mass changes are done I would like to propose another: replace SigInit with DetectEngineAppendSig everywhere. Then don't change the DetectEngineAppendSig arguments, but instead use a var in the DetectEngineCtx to store and access an error string. Much like the already present:
<pre>
    /** Store rule file and line so that parsers can use them in errors. */
    char *rule_file;
    int rule_line;
</pre>

inliniac:
Commit message "detect-engine: remove DONE state" should explain the reason of the change.

inliniac:
Does this showing of failed rules work if there are many? E.g. if someone has a broken HOME_NET thousands of rules will fail to load. This would easily lead to many megabytes of transfer over the unix socket.

inliniac:
In any case this needs a rebase as well. Closing.

regit:
Good catch on lot of failed rules issue.

PR #2223, opened 2016-09-05
netmap: Ignore interface 'up' status

michalpurzynski:
Let's ignore interface status as reported in kernel, when using kernel bypass packet capture mechanisms.

I have tested it and it presents no issues whatsoever (on FreeBSD).


inliniac:
Can you give this a commit message more in line with our style. Something like: "netmap: ignore interface 'up' status". You can leave out that it works, that is implied by the PR :)

michalpurzynski:
Your wish... etc ;)

inliniac:
You changed the pull request title, not the commit message. The commit message is what gets recorded in the git history, so that is what I care about most.

PR #2232, opened 2016-09-09
Mixed mode v10

glongo:
This addresses comments in the previous patchset: https://github.com/inliniac/suricata/pull/1857

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/124
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/123

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -358,6 +358,14 @@ typedef struct PktProfiling_ {
 /* forward declartion since Packet struct definition requires this */
 struct PacketQueue_;
 
+/*
+   since a pkt could come from different runmodes, we need to set
+   a mode to be able to do an action later.
+   (e.g. if a pkt come from NFQ, pkt_mode will be set to IPS and
+         it means that we'll be able to drop it.)
+*/
+enum PktMode {PKT_MODE_IDS, PKT_MODE_IPS, PKT_MODE_AUTO};
+
 /* sizes of the members:
  * src: 17 bytes
  * dst: 17 bytes
@@ -411,6 +419,8 @@ typedef struct Packet_
     uint32_t flow_hash;
 
     struct timeval ts;
+
+    enum PktMode pkt_mode;
 
     union {
         /* nfq stuff */
@@ -894,6 +904,8 @@ int PacketCallocExtPkt(Packet *p, int datalen);
 int PacketCopyData(Packet *p, uint8_t *pktdata, int pktlen);
 int PacketSetData(Packet *p, uint8_t *pktdata, int pktlen);
 int PacketCopyDataOffset(Packet *p, int offset, uint8_t *data, int datalen);
+int PacketModeIsIPS(const Packet *p);
+int PacketModeIsIDS(const Packet *p);
 const char *PktSrcToString(enum PktSrcEnum pkt_src);
 
 DecodeThreadVars *DecodeThreadVarsAlloc(ThreadVars *);
--- a/src/stream-tcp-inline.c
+++ b/src/stream-tcp-inline.c
@@ -41,9 +41,14 @@ extern int stream_inline;
  *  \retval 0 no
  *  \retval 1 yes
  */
-int StreamTcpInlineMode(void)
+int StreamTcpInlineMode(const Packet *p)
 {
-    return stream_inline;
+    if (stream_inline == PKT_MODE_IDS)
+        return 0;
+    else if (stream_inline == PKT_MODE_IPS)
+        return 1;
+    else /* implied AUTO */
+        return (stream_inline && PacketModeIsIPS(p));
 }
 
 /**
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -229,7 +229,7 @@ SC_ATOMIC_DECLARE(unsigned int, engine_stage);
 volatile uint8_t suricata_ctl_flags = 0;
 
 /** Run mode selected */
-int run_mode = RUNMODE_UNKNOWN;
+RunModesList runmodeslist = { {RUNMODE_UNKNOWN, RUNMODE_UNKNOWN}, 0};
 
 /** Engine mode: inline (ENGINE_MODE_IPS) or just
   * detection mode (ENGINE_MODE_IDS by default) */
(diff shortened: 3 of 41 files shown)


inliniac:
Breaks build for netmap:
<pre>
runmode-netmap.c:283:36: error: too few arguments to function call, single argument 'runmode' was not specified
    int nlive = LiveGetDeviceCount();
                ~~~~~~~~~~~~~~~~~~ ^
./util-device.h:40:1: note: 'LiveGetDeviceCount' declared here
int LiveGetDeviceCount(enum RunModes runmode);
^
runmode-netmap.c:300:54: error: too few arguments to function call, expected 2, have 1
        const char *live_dev = LiveGetDeviceName(ldev);
                               ~~~~~~~~~~~~~~~~~     ^
./util-device.h:41:1: note: 'LiveGetDeviceName' declared here
const char *LiveGetDeviceName(int number, enum RunModes runmode);
^
runmode-netmap.c:330:58: error: too few arguments to function call, expected 2, have 1
            const char *live_dev = LiveGetDeviceName(ldev);
                                   ~~~~~~~~~~~~~~~~~     ^
./util-device.h:41:1: note: 'LiveGetDeviceName' declared here
const char *LiveGetDeviceName(int number, enum RunModes runmode);
^
3 errors generated.
</pre>
Please review all the runmodes, including tilera, ipfw, etc.

src/decode.h
```diff
@@ -358,6 +358,14 @@ typedef struct PktProfiling_ {
 /* forward declartion since Packet struct definition requires this */
 struct PacketQueue_;
 
+/*
+   since a pkt could come from different runmodes, we need to set
+   a mode to be able to do an action later.
+   (e.g. if a pkt come from NFQ, pkt_mode will be set to IPS and
+         it means that we'll be able to drop it.)
+*/
+enum PktMode {PKT_MODE_IDS, PKT_MODE_IPS, PKT_MODE_AUTO};
```
inliniac: The AUTO value here is not being used except for a single 'set' to stream_inline somewhere. Whats the point of it?

src/stream-tcp-inline.c
```diff
@@ -41,9 +41,14 @@ extern int stream_inline;
  *  \retval 0 no
  *  \retval 1 yes
  */
-int StreamTcpInlineMode(void)
+int StreamTcpInlineMode(const Packet *p)
 {
-    return stream_inline;
+    if (stream_inline == PKT_MODE_IDS)
```
inliniac: can you turn this into a switch statement? Then the AUTO will pop up in greps.

src/stream-tcp-inline.c
```diff
@@ -41,9 +41,14 @@ extern int stream_inline;
...
+        return 0;
+    else if (stream_inline == PKT_MODE_IPS)
+        return 1;
+    else /* implied AUTO */
+        return (stream_inline && PacketModeIsIPS(p));
```
inliniac: why is stream_inline part of this comparison? It's value is PKT_MODE_AUTO, so != 0

src/suricata.c
```diff
@@ -1111,20 +1148,30 @@ static void SCPrintElapsedTime(SCInstance *suri)
 static int ParseCommandLineAfpacket(SCInstance *suri, const char *optarg)
 {
 #ifdef HAVE_AF_PACKET
-    if (suri->run_mode == RUNMODE_UNKNOWN) {
-        suri->run_mode = RUNMODE_AFP_DEV;
+    RunModesList *runmodes = &suri->runmodeslist;
+    if (runmodes->run_mode[runmodes->runmodes_cnt] == RUNMODE_UNKNOWN) {
+        if (!RunmodeIsAlreadyRan(runmodes, RUNMODE_AFP_DEV)) {
```
inliniac: IsAlreadyRan is a bit of a confusing name, what does 'Ran' mean here? What about ..AlreadyAdded? of ..IsSet or something similar.

src/decode.h
inliniac: Ok I found out later.

inliniac:
Needs a bit of work.

src/stream-tcp-inline.c
glongo: stream_inline == PKT_MODE_AUTO (stream.inline: auto) means that the value returned depends on the PKT_MODE set in the capture method.
For instance: with nfqueue, this will return 1 (so we're in inline mode), instead 0 with nflog.

inliniac: "else" implies stream_inline == PKT_MODE_AUTO (which is 2), so whats the point of having "(stream_inline && PacketModeIsIPS(p));" then? stream_inline's value can only be one thing. 2. So the condition is always true and thus not needed.

glongo: if PacketModeIsIPS(p) returns 0, the condition is not true, I'm not getting the point maybe.

inliniac: On 09-09-16 16:49, Giuseppe Longo wrote:
> In src/stream-tcp-inline.c
> <https://github.com/inliniac/suricata/pull/2232#discussion_r78192965>:
> 
>>  {
>> -    return stream_inline;
>> +    if (stream_inline == PKT_MODE_IDS)
>> +        return 0;
>> +    else if (stream_inline == PKT_MODE_IPS)
>> +        return 1;
>> +    else /* implied AUTO */
>> +        return (stream_inline && PacketModeIsIPS(p));
> 
> if PacketModeIsIPS(p) returns 0, the condition is not true, I'm not
> getting the point maybe.

My point is only about stream_inline part of return (stream_inline &&
PacketModeIsIPS(p));. That part is not adding anything.

glongo: I see the point, but let me to ask a question: is there a case that nfq is running with stream.inline set to no?

inliniac: I don't see how that is relevant to what I'm trying to say here.

glongo: Just to be sure that I can remove "stream_inline".

src/suricata.c
glongo: ack

src/stream-tcp-inline.c
glongo: ack

src/stream-tcp-inline.c
glongo: ack

PR #2213, opened 2016-09-01
Eve alert dump dns fields

fooinha:
The EVE JSON output for a Suricata alert is missing the DNS fields.

This missing information requires that if we've a DNS based Snort rule, we have to correlate the flow_id fields from alert and DNS logged events.

The idea is to dump DNS fields to EVE's alert JSON output.

The alerts for DNS answer packets also logs the queries if question section is available.

Please review this, and consider this idea as an improvement for eve alert logging.

----
Example output:

```javascript
{
  "timestamp": "2016-09-01T20:57:08.943822+0000",
  "flow_id": 8782437618590984000,
  "in_iface": "eth0",
  "event_type": "alert",
  "src_ip": "192.168.65.1",
  "src_port": 53,
  "dest_ip": "172.17.0.4",
  "dest_port": 44566,
  "proto": "UDP",
  "alert": {
    "action": "allowed",
    "gid": 1,
    "signature_id": 8888888,
    "rev": 1,
    "signature": "ET DNS Answer",
    "category": "Potentially Bad Traffic",
    "severity": 2
  },
  "dns": {
    "tx": {
      "answers": [
        {
          "type": "answer",
          "id": 35844,
          "rcode": "NOERROR",
          "rrname": "www.google.com",
          "rrtype": "AAAA",
          "ttl": 280,
          "rdata": "2a00:1450:4003:0801:0000:0000:0000:2004"
        }
      ],
      "queries": [
        {
          "type": "query",
          "id": 35844,
          "rrname": "www.google.com",
          "rrtype": "AAAA",
          "tx_id": 35844
        }
      ],
      "num_tx": 1
    }
  }
}
```
----

No unit tests for this part of the code, nevertheless , I ran it and got this.

....
```text
pass
==== TEST RESULTS ====
PASSED: 3618
FAILED: 0
======================
```
----
I ran valgrind and did some basic tests and did not see any memory leak...

```txt
==22009== 48 bytes in 1 blocks are definitely lost in loss record 145 of 208
==22009==    at 0x4C2FB55: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==22009==    by 0x4E43D3D: ??? (in /usr/lib/x86_64-linux-gnu/libmagic.so.1.0.0)
==22009==    by 0x4E447ED: ??? (in /usr/lib/x86_64-linux-gnu/libmagic.so.1.0.0)
==22009==    by 0x7278F2: MagicInit (util-magic.c:83)
[...]

PR #2233, opened 2016-09-11
eve: adds dns fields for eve alert's json output

fooinha:
eve: adds dns fields for eve alert's json output

dns: uniformize json output following the same format as in alert event type

Check #2213 for more details.


fooinha:
Hi, I may have another PR that code will conflict with this one. It's about payload printable size option for EVE alert json.

Advice please.

Thanx

jasonish:
Proper review might take a little longer than usual.  I believe there is some work-in-progress going on with the DNS logging that may involve a more compressed output format.  And there will need to be some review and consensus on what that format should look like, which should be nailed down before adding DNS info to alerts.

Some of the development team is busy with training this week, then there will be some catch up to do.

src/output-json-alert.c
```diff
@@ -281,6 +299,19 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
             }
         }
 
+        if (json_output_ctx->flags & LOG_JSON_DNS) {
+            if (p->flow != NULL) {
+                FLOWLOCK_RDLOCK(p->flow);
```
inliniac: This lock is no longer necessary now that https://github.com/inliniac/suricata/pull/2235 is merged.

inliniac:
Could you have  a look at @decanio 's work in #2267 ?

src/output-json-alert.c
fooinha: I'll remove locks and try to rebase to master HEAD. Thanx.

fooinha:
Applied feature and commit message from @decanio's work in #2267 .
Alert event from a rule with DNS protocol is not filtered and style of output  remains with unified output only ( so we don't generate multiple alert events for the same DNS request/response ) .

fooinha:
Added some unit tests.

Test OutputJsonDnsQueryDiscreteTest01 -- Tests discrete query     : pass
Test OutputJsonDnsResponseUnifiedTest02 -- Tests unified response : pass

fooinha:
Output:


- https://gist.github.com/fooinha/657b4ab5f6e5cffb4bacb2a45250f346#file-slit-json

- https://gist.github.com/fooinha/657b4ab5f6e5cffb4bacb2a45250f346#file-unified-json

inliniac:
I like these outputs. @jasonish @decanio thoughts?

decanio:
I haven't been following this thread closely.  Is the objective of this PR to add DNS metadata to alerts that trigger on DNS flows, and do it in a format that is compatible with my compact DNS PR?

If so the examples look ok to me.

jasonish:
I like them to.  But I think when logging DNS details into an event it should only log the "unified" format, as we have to fit it in a single record anyways. It also means I think this PR has a dependence on #2275.

fooinha:
Hi, the goal of this PR was only to add dns info into the alert event. 
The alert event contained the query/response in an unified style.

@inliniac , pointed to @decanio work in #2267 and #2275, and I included those ideas implementation in this PR.
 In the process I changed a bit the internal functions for dns ouput, and added 3 unit tests.

Yes, when alerting we force the "unified" format, so that we don't issue multiple events per rule.
Yes, the pasted gist locations, were files created by a running suricata with PR branch code.

Thanx.

fooinha:
The closed PR #2213 has the initial part of the story.

fooinha:
@inliniac , I saw your last comments at #2275.
This PR also implements the same, but with a significant rewrite for internal functions for dns json output.

Should we also tag this one with needsdocupdate ?

inliniac:
I'm likely merging an updated version of #2275 for 3.2. This branch can then be rebased for the next major version. See https://lists.openinfosecfoundation.org/pipermail/oisf-devel/2016-October/003803.html

Btw, please have look at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow
Our workflow of github pull requests is to not modify them (and thus the branch they point to) after doing the PR. This PR is a nice example of why that is. Many unrelated commits make it messy. So closing.

decanio:
I should have an update to my PR before the day is out.

fooinha:
@decanio , do you need set of extra eyes to test it?

PR #2234, opened 2016-09-12
invalid event fixes v0

inliniac:
Discovered during the dev training in Paris: invalid event enum and name<>event array were not in sync.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/486
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/491

### DIFF
--- a/rules/decoder-events.rules
+++ b/rules/decoder-events.rules
@@ -121,6 +121,7 @@ alert pkthdr any any -> any any (msg:"SURICATA IPv6-in-IPv6 packet too short"; d
 alert pkthdr any any -> any any (msg:"SURICATA IPv6-in-IPv6 invalid protocol"; decode-event:ipv6.ipv6_in_ipv6_wrong_version; sid:2200085; rev:1;)
 
 # MPLS rules
+alert pkthdr any any -> any any (msg:"SURICATA MPLS header too small"; decode-event:mpls.header_too_small; sid:2200111; rev:1;)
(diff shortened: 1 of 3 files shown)


inliniac:
@jasonish I'm a bit confused about the difference between frag_too_large and frag_pkt_too_large. Do you have an idea to make the distinction more clear?

jasonish:
FRAG_PKT_TOO_LARGE occurs when a fragment would extend past the maximum packet size.

I can't remember the purpose behing IPV{4,6}_FRAG_TOO_LARGE, and their usage is wrong anyways.  These events are only set when a fragment is trimmed completely due to the overlap policy.  I'm thinking it would be much better to set the IPV4_FRAG_IGNORED event here since it is dropped - OR set nothing as the overlap even will get set anyways.

So I think we can remove IPV4_FRAG_TOO_LARGE and IPV6_FRAG_TOO_LARGE.

inliniac:
@jasonish any chance you can take this branch and rework it to clean that up?

jasonish:
> @jasonish any chance you can take this branch and rework it to clean that up?
>  Show all checks

Yes. Will do.

inliniac:
I merged the mpls patch. Can you do a new PR with the frag fixes and the BUG_ON addition?

PR #2236, opened 2016-09-13
flowbits: validate that there are no spaces in the name - v1

jasonish:
Fixes issue: https://redmine.openinfosecfoundation.org/issues/1889

To catch the issue where the ';' is missing we have to expand the
regex to capture the whole name string, not just the leading
valid stuff. Then verify that there are no spaces in the name
(Snort has the same restriction) and fail if there is.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/310
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/315

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -45,7 +45,7 @@
 #include "util-unittest.h"
 #include "util-debug.h"
 
-#define PARSE_REGEX         "([a-z]+)(?:,\\s*([^\\s]*))?"
+#define PARSE_REGEX         "([a-z]+)(?:,\\s*(.*))?"
 static pcre *parse_regex;
 static pcre_extra *parse_regex_study;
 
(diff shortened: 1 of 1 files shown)


src/detect-flowbits.c
```diff
@@ -182,6 +182,20 @@ static int DetectFlowbitParse(char *str, char *cmd, int cmd_len, char *name,
             SCLogError(SC_ERR_PCRE_GET_SUBSTRING, "pcre_copy_substring failed");
             return 0;
         }
+
+        /* Trim trailing whitespace. */
+        while (isblank(name[strlen(name) - 1])) {
```
inliniac: the regex should prevent it, but it may be good to add a strlen(name) > 0 condition in the while loop. If strlen returns 0 we may have a negative index. Might prevent future issues if the regex changes.

inliniac:
Should this be applied to hostbits & xbits as well? And I guess to flowint and vars and such?

src/detect-flowbits.c
jasonish: Done in new PR.

> Should this be applied to hostbits & xbits as well? And I guess to flowint and vars and such?

New PR has it done for hostbots.  The others weren't as straight forward as hostbits and flowbits.

jasonish:
See #2239.

PR #2237, opened 2016-09-14
Mixed mode v11

glongo:
Mixed mode patchset.

This addresses comment of the last PR: #2232 

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/125
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/124

### DIFF
--- a/src/alert-fastlog.c
+++ b/src/alert-fastlog.c
@@ -146,7 +146,7 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
         }
 
         char *action = "";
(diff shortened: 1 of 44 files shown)


inliniac:
Mostly looks good, but I think the runmode setting process can be done cleaner. There is a RunmodeGetNumber, but the patch still bypasses that a few times. Also I think it would be good to replace the explicit runmode[runmode_cnt] = somerunmode by a 'RunmodeSetPrimary'. All the special runmodes (e.g. unittests) would use this.

PR #2239, opened 2016-09-15
flowbits/hostbits: validate that there are no spaces in the name - v2

jasonish:
Previous PR: #2236

Fixes issue: https://redmine.openinfosecfoundation.org/issues/1889

To catch the issue where the ';' is missing we have to expand the
regex to capture the whole name string, not just the leading
valid stuff. Then verify that there are no spaces in the name
(Snort has the same restriction) and fail if there is.

Also some unit test cleanup.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/315
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/320

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -354,70 +354,19 @@ static int FlowBitsTestParse01(void)
 
 static int FlowBitsTestSig01(void)
 {
-    uint8_t *buf = (uint8_t *)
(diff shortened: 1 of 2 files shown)


inliniac:
Merged through #2242, thanks Jason!

PR #2240, opened 2016-09-15
preview: don't use pcre for high level option parsing - v1

jasonish:
This is a possible replacement for the regular expression parsing we currently do for rule options (could be applied to other parts of the rules as well).  I think it makes it easier to validate, and easier to understand when issues arise vs deciphering regular expressions.

On my slower machine it also parses 35,000 rules in 34 seconds instead of 53.

Thoughts on this approach instead of regular expressions?

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/317
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/322

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -72,9 +72,7 @@
 extern int sc_set_caps;
 
 static pcre *config_pcre = NULL;
-static pcre *option_pcre = NULL;
 static pcre_extra *config_pcre_extra = NULL;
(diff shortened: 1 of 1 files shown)


inliniac:
What can I say, the perf numbers look good. Otherwise C parsing code is ugly whether it's using pcre or not ;) No objections to going a bit deeper here.

PR #2244, opened 2016-09-19
defrag - cleanup too large events - v1

jasonish:
Based on PR #2234 

The frag_too_large event was not being used properly and was only being sent when a fragment is dropped due to complete overlap, in which case an overlap event was already set.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/319
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/324


inliniac:
Merged as part of #2250, thanks Jason.

PR #2246, opened 2016-09-19
util-decode-mime: remove quote from boundary= string.

decanio:
util-decode-mime: remove quote from boundary= string.

remove quote from the end of the boundary= string.  This was throwing off
the mime parser so that it wouldn't always catch mime boundaries causing
things like missed attachments.

https://buildbot.openinfosecfoundation.org/builders/decanio/builds/26
https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/26


inliniac:
Merged through #2250, thanks Tom!

PR #2247, opened 2016-09-20
App layer stats v17

regit:
Update of https://github.com/inliniac/suricata/pull/2116 addressing comments and rebasing on latest master.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/188
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/184

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -140,6 +140,8 @@ struct AppLayerParserState_ {
      * we don't need a var per direction since we don't log a transaction
      * unless we have the entire transaction. */
     uint64_t log_id;
+    /* Count of number of transactions */
+    uint64_t tx_cnt;
 
     /* Used to store decoder events. */
     AppLayerDecoderEvents *decoder_events;
@@ -624,6 +626,7 @@ void AppLayerParserSetTransactionInspectId(AppLayerParserState *pstate,
         else
             break;
     }
+
     pstate->inspect_id[direction] = idx;
 
     SCReturn;
@@ -1011,6 +1014,20 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow 
         }
     }
 
+    if (AppLayerParserProtocolIsTxAware(f->proto, alproto)) {
+        if (likely(tv)) {
+            /* At each pass in the parser we add to the tv counter
+             * the difference of tx on a flow between now and the previous pass
+             */
+            uint64_t p_tx_cnt = pstate->tx_cnt;
+            pstate->tx_cnt = AppLayerParserGetTxCnt(f->proto, alproto, f->alstate);
+            if (!AppLayerParserProtocolIsTxAware(f->proto, alproto)) {
+                goto end;
+            }
+            AppLayerIncTxCounter(tv, f->proto, alproto,
+                                 pstate->tx_cnt - p_tx_cnt);
+        }
+    }
     /* next, see if we can get rid of transactions now */
     AppLayerParserTransactionsCleanup(f);
 
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -149,6 +149,13 @@ struct AppLayerParserState_ {
  * Post 2.0 let's look at changing this to move it out to app-layer.c. */
 static AppLayerParserCtx alp_ctx;
 
+int AppLayerParserProtoIsRegistered(uint8_t ipproto, AppProto alproto)
+{
+    uint8_t ipproto_map = FlowGetProtoMapping(ipproto);
+
+    return (alp_ctx.ctxs[ipproto_map][alproto].StateAlloc != NULL) ? 1 : 0;
+}
+
 AppLayerParserState *AppLayerParserStateAlloc(void)
 {
     SCEnter();
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -913,7 +913,7 @@ int AppLayerParserSetTxDetectState(uint8_t ipproto, AppProto alproto,
 
 /***** General *****/
 
-int AppLayerParserParse(AppLayerParserThreadCtx *alp_tctx, Flow *f, AppProto alproto,
+int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow *f, AppProto alproto,
                         uint8_t flags, uint8_t *input, uint32_t input_len)
 {
     SCEnter();
@@ -1314,7 +1314,7 @@ int AppLayerParserRequestFromFile(AppProto alproto, char *filename)
             }
             //PrintRawDataFp(stdout, buffer, result);
 
-            (void)AppLayerParserParse(alp_tctx, f, alproto, flags, buffer, result);
+            (void)AppLayerParserParse(NULL, alp_tctx, f, alproto, flags, buffer, result);
             if (done)
                 break;
         }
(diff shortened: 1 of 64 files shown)


src/app-layer.c
```diff
@@ -648,6 +678,80 @@ void AppLayerRegisterGlobalCounters(void)
...
+                }
+                snprintf(applayer_counters[ipproto_map][alproto].name,
+                         strlen(str) + alproto_len,
+                         "%s%s", str, alproto_str);
+            }
+        }
+    }
+}
+
+void AppLayerRegisterCounters(ThreadVars *tv, uint8_t ipproto)
```
inliniac: This is per thread so that should be reflected in the name.

src/app-layer.c
```diff
@@ -648,6 +678,80 @@ void AppLayerRegisterGlobalCounters(void)
...
+    AppLayerProtoDetectSupportedAppProtocols(alprotos);
+
+    for (alproto = 0; alproto < ALPROTO_MAX; alproto++) {
+        if (alprotos[alproto] == 1) {
+            char *str = "app_layer.flow.";
+            char *alproto_str = AppLayerGetProtoName(alproto);
+            int alproto_len = strlen(alproto_str) + 1;
+            uint8_t ipproto_map = FlowGetProtoMapping(ipproto);
+
+            if (AppLayerParserProtoIsRegistered(ipproto, alproto) &&
```
inliniac: Can you explain the logic here? Don't see what this check does and why there are 2 conditions.

src/app-layer.c
```diff
@@ -648,6 +678,80 @@ void AppLayerRegisterGlobalCounters(void)
...
+    for (alproto = 0; alproto < ALPROTO_MAX; alproto++) {
+        if (alprotos[alproto] == 1) {
+            uint8_t ipproto_map = FlowGetProtoMapping(ipproto);
+            applayer_counters[ipproto_map][alproto].counter_id =
+                StatsRegisterCounter(applayer_counters[ipproto_map][alproto].name, tv);
+        }
+    }
+}
+
+void AppLayerDeRegisterCounters(uint8_t ipproto)
```
inliniac: This is not per thread, but it does have a name 'opposite' of a thread func: AppLayerRegisterCounters (per thread) vs AppLayerDeRegisterCounters (global)

src/flow-worker.c
```diff
@@ -98,6 +98,9 @@ static TmEcode FlowWorkerThreadInit(ThreadVars *tv, void *initdata, void **data)
         return TM_ECODE_FAILED;
     }
 
+    AppLayerRegisterCounters(tv, IPPROTO_UDP);
```
inliniac: Not sure why we should have 2 calls here for TCP and UDP. Might just as well move the ipproto inside the Register func.

src/output-json-stats.c
```diff
@@ -47,6 +47,8 @@
 #include "output-json.h"
 #include "output-json-stats.h"
 
+#include "app-layer.h"
```
inliniac: Seems unnecessary in this commit?

src/app-layer.c
```diff
@@ -576,6 +600,9 @@ int AppLayerSetup(void)
     AppLayerParserRegisterProtocolParsers();
     AppLayerProtoDetectPrepareState();
 
+    AppLayerSetupCounters(IPPROTO_UDP);
```
inliniac: Same here, we can move the ipproto logic inside the Setup func

src/app-layer.c
```diff
@@ -586,6 +613,9 @@ int AppLayerDeSetup(void)
     AppLayerProtoDetectDeSetup();
     AppLayerParserDeSetup();
 
+    AppLayerDeRegisterCounters(IPPROTO_UDP);
```
inliniac: Same here, we can move the ipproto logic inside the Setup func

src/app-layer.c
```diff
@@ -80,6 +91,14 @@ static inline int ProtoDetectDone(const Flow *f, const TcpSession *ssn, uint8_t
             (FLOW_IS_PM_DONE(f, direction) && FLOW_IS_PP_DONE(f, direction)));
 }
 
+static void AppLayerIncFlowCounter(ThreadVars *tv, uint8_t ipproto, AppProto alproto)
```
inliniac: If you just pass 'Flow *f' to this function you can directly use f->protomap and f->alproto. Safes one FlowGetProtoMapping call.

src/app-layer-parser.c
```diff
@@ -1011,6 +1014,20 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow
...
     }
 
+    if (AppLayerParserProtocolIsTxAware(f->proto, alproto)) {
+        if (likely(tv)) {
+            /* At each pass in the parser we add to the tv counter
+             * the difference of tx on a flow between now and the previous pass
+             */
+            uint64_t p_tx_cnt = pstate->tx_cnt;
+            pstate->tx_cnt = AppLayerParserGetTxCnt(f->proto, alproto, f->alstate);
+            if (!AppLayerParserProtocolIsTxAware(f->proto, alproto)) {
```
inliniac: Seems already checked on line 1017?

src/app-layer-parser.c
```diff
@@ -140,6 +140,8 @@ struct AppLayerParserState_ {
      * we don't need a var per direction since we don't log a transaction
      * unless we have the entire transaction. */
     uint64_t log_id;
+    /* Count of number of transactions */
+    uint64_t tx_cnt;
```
inliniac: I'm not convinced this is needed if it's only used to get a diff between tx_cnt before and after a AppLayerParse call.

src/app-layer.c
```diff
@@ -99,6 +101,15 @@ static void AppLayerIncFlowCounter(ThreadVars *tv, uint8_t ipproto, AppProto alp
     }
 }
 
+void AppLayerIncTxCounter(ThreadVars *tv, uint8_t ipproto, AppProto alproto,
+                          uint64_t step)
+{
+    if (likely(tv)) {
+        uint8_t ipproto_map = FlowGetProtoMapping(ipproto);
```
inliniac: We can avoid this if the Flow is passed to this function. Then we can use f->protomap

src/app-layer-parser.c
```diff
@@ -1011,6 +1014,20 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow
         }
     }
 
+    if (AppLayerParserProtocolIsTxAware(f->proto, alproto)) {
+        if (likely(tv)) {
+            /* At each pass in the parser we add to the tv counter
```
regit: @inliniac on the tx_cnt being part of the Flow structure, we have tried to explain it in this comment.

inliniac: I don't see why you need it to be in the state.

Why would this not work?
<pre>
p_tx_cnt = ...GetTxCnt();
...
Parser Runs
...
cur_tx_cnt = ...GetTxCnt()
if (cur_tx_cnt > p_tx_cnt) {
    new_tx_cnt = cur_tx_cnt - p_tx_cnt;
}
</pre>

src/app-layer.c
glongo: Here we check if a protocol, like dns, is registered for both tcp and udp. if yes, we add a suffix (ipproto_suffix) to the counter name, for instance: "dns_tcp", or "dns_udp".

PR #2248, opened 2016-09-20
rule load opt v9

inliniac:
Optimize rule address parsing. In a ruleset many of the rule address vars are indentical. This patch caches the parsing result. Rules now reference that result instead of each having their own copy of it. Should speed up parsing and slightly reduce memory usage.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/490
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/495


inliniac:
Merged through #2250

PR #2252, opened 2016-09-20
Dev feat509 hashing v0.9

serializingme:
New version of #2109 rebased to the latest master.


inliniac:
Merged through #2266, thanks Duarte!

I did a few cleanups and restored the 'force-md5' options for backwards compatibility.

serializingme:
Duh! That is a good catch, thanks!

PR #2249, opened 2016-09-20
detect: tls.subject and tls.issuerdn reimplementation (v2)

thus:
Reimplemented tls.subject and tls.issuerdn as tls_cert_subject and tls_cert_issuer (both as mpm keywords). Kept the old keywords to be compatible with existing rules.

Updates:
- Use unittest macros (FAIL_IF/PASS/etc).

Usage example:
```
alert tls $EXTERNAL_NET any -> $HOME_NET any (msg:"tls_cert_subject test";
        tls_cert_subject; content:"google"; nocase; sid:1;)
```

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/38
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/38

PR #2262, opened 2016-09-22
tls cert subject and issuer reimplementation and cert validity dates

thus:
Updates:
- Rebased
- #2249 has already been merged

One PR containing both #2249 and #2187.

Have also added a new detect engine for TLS validity.

This PR fixes the following issues:
- https://redmine.openinfosecfoundation.org/issues/1583
- https://redmine.openinfosecfoundation.org/issues/1896
- https://redmine.openinfosecfoundation.org/issues/1044
- https://redmine.openinfosecfoundation.org/issues/712

prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/41
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/41


src/detect-engine-tls.c
```diff
@@ -341,3 +341,45 @@ int DetectEngineInspectTlsSubject(ThreadVars *tv, DetectEngineCtx *de_ctx,
 
     return cnt;
 }
+
+int DetectEngineInspectTlsGeneric(ThreadVars *tv, DetectEngineCtx *de_ctx,
```
inliniac: This is a copy of DetectEngineInspectGenericList from detect-engine-dns.c right? Might be better to move that one in to detect-engine.c and call that.

thus: Jupp, that's correct. I'll do that instead then.

inliniac: In a separate commit please :)

thus:
Replaced by #2263

PR #2268, opened 2016-09-23
coverty: fix CID 1361873 - Insecure data handling (TAINTED_SCALAR) - v3

thus:
Update:
Took a new crack at this. Changed so that we set a decoder event if the length is wrong, but continue decoding the packet.

Set HANDSHAKE_INVALID_LENGTH event when client_hello is of wrong length.

Had to fix several unit tests because of this.

Also moved decoding of client_hello to a function, to be able to reuse this code for decoding server_hello in the future.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/44
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/44


inliniac:
Merged through #2284, thanks Mats!

PR #2270, opened 2016-09-23
tls: use new unit test macros

thus:
Use the new unit test macros in tls unit tests.

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/45
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/45


inliniac:
Merged through #2284, thanks Mats!

PR #2276, opened 2016-09-24
Bypass v7

regit:
This is an update of #2228 changing a lot of things.

Main change is that it introduces a bypassed state in flow which will allow suricata to get rid fast of bypassed flows instead of waiting for "established" timeout. This will lower the pressure on the flow table. Also as the bypassed information is added in the EVE entry of flow the user will know that bypass has been done on the flow.
  
Another change is that definition of bypass is not done anymore at registration but it is done inside the capture. For instance, in NFQ capture (which is the only one implemented in this PR), the bypass is activated if the bypass_mask is set in the configuration file.
  
Also remarks on PR #2228 have been addressed.
  
PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/193
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/189

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -385,6 +385,9 @@ void PacketDefragPktSetupParent(Packet *parent)
 void PacketBypassCallback(Packet *p)
 {
     if (p->BypassPacketsFlow) {
+        /* only set bypassed state if succesful */
+        SC_ATOMIC_SET(p->flow->flow_state, FLOW_STATE_BYPASSED);
+
         p->BypassPacketsFlow(p);
     }
 }
--- a/src/decode.c
+++ b/src/decode.c
@@ -384,6 +384,12 @@ void PacketDefragPktSetupParent(Packet *parent)
 
 void PacketBypassCallback(Packet *p)
 {
+    /* Don't try to bypass if flow is already out */
+    int state = SC_ATOMIC_GET(p->flow->flow_state);
+    if (state == FLOW_STATE_BYPASSED) {
+        return;
+    }
+
     if (p->BypassPacketsFlow) {
         /* only set bypassed state if succesful */
         SC_ATOMIC_SET(p->flow->flow_state, FLOW_STATE_BYPASSED);
--- a/src/decode.c
+++ b/src/decode.c
@@ -382,6 +382,13 @@ void PacketDefragPktSetupParent(Packet *parent)
     DecodeSetNoPayloadInspectionFlag(parent);
 }
 
+void PacketBypassCallback(Packet *p)
+{
+    if (p->BypassPacketsFlow) {
+        p->BypassPacketsFlow(p);
+    }
+}
+
 void DecodeRegisterPerfCounters(DecodeThreadVars *dtv, ThreadVars *tv)
 {
     /* register counters */
--- a/src/decode.c
+++ b/src/decode.c
@@ -384,9 +384,11 @@ void PacketDefragPktSetupParent(Packet *parent)
 
 void PacketBypassCallback(Packet *p)
 {
-    /* Don't try to bypass if flow is already out */
+    /* Don't try to bypass if flow is already out or
+     * if we have failed to do it once */
     int state = SC_ATOMIC_GET(p->flow->flow_state);
-    if (state == FLOW_STATE_BYPASSED) {
+    if ((state == FLOW_STATE_BYPASSED) ||
+           (p->flow->flags & FLOW_BYPASS_FAILED)) {
         return;
     }
 
@@ -395,6 +397,8 @@ void PacketBypassCallback(Packet *p)
         SC_ATOMIC_SET(p->flow->flow_state, FLOW_STATE_BYPASSED);
 
         p->BypassPacketsFlow(p);
+    } else {
+        p->flow->flags |= FLOW_BYPASS_FAILED;
     }
 }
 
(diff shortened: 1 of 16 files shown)


src/decode.c
```diff
@@ -382,6 +382,26 @@ void PacketDefragPktSetupParent(Packet *parent)
...
+     * if we have failed to do it once */
+    int state = SC_ATOMIC_GET(p->flow->flow_state);
+    if ((state == FLOW_STATE_BYPASSED) ||
+           (p->flow->flags & FLOW_BYPASS_FAILED)) {
+        return;
+    }
+
+    if (p->BypassPacketsFlow) {
+        /* only set bypassed state if succesful */
+        SC_ATOMIC_SET(p->flow->flow_state, FLOW_STATE_BYPASSED);
```
inliniac: Use FlowUpdateState

src/decode.h
```diff
@@ -448,6 +448,8 @@ typedef struct Packet_
     /** The release function for packet structure and data */
     void (*ReleasePacket)(struct Packet_ *);
 
+    void (*BypassPacketsFlow)(struct Packet_ *);
```
inliniac: missing comment explaining what this is for

src/detect-bypass.c
```diff
@@ -0,0 +1,244 @@
...
+    SCSigRegisterSignatureOrderingFuncs(de_ctx);
+    SCSigOrderSignatures(de_ctx);
+    SCSigSignatureOrderingModuleCleanup(de_ctx);
+
+    SigGroupBuild(de_ctx);
+    DetectEngineThreadCtxInit(&th_v, (void *)de_ctx, (void *)&det_ctx);
+
+    SCMutexLock(&f.m);
+    int r = AppLayerParserParse(alp_tctx, &f, ALPROTO_HTTP, STREAM_TOSERVER, http_buf1, http_len1);
+    if (r != 0) {
```
inliniac: use FAIL_IF

src/detect-bypass.c
```diff
@@ -0,0 +1,244 @@
...
+    FAIL_IF(http_state == NULL);
+
+    /* do detect */
+    SigMatchSignatures(&th_v, de_ctx, det_ctx, p1);
+
+    FAIL_IF(PacketAlertCheck(p1, 1));
+
+    SCMutexLock(&f.m);
+    r = AppLayerParserParse(alp_tctx, &f, ALPROTO_HTTP, STREAM_TOCLIENT, http_buf2, http_len2);
+    if (r != 0) {
```
inliniac: FAIL_IF

src/detect-bypass.c
```diff
@@ -0,0 +1,244 @@
...
+
+    http_state = f.alstate;
+    FAIL_IF(http_state == NULL);
+
+    /* do detect */
+    SigMatchSignatures(&th_v, de_ctx, det_ctx, p1);
+
+    FAIL_IF(PacketAlertCheck(p1, 1));
+
+    SCMutexLock(&f.m);
```
inliniac: wrong lock func, use FLOWLOCK_WRLOCK

(I have a patch that mass changes this for the entire tree)

src/detect-filestore.c
```diff
@@ -408,3 +416,113 @@ static void DetectFilestoreFree(void *ptr)
...
+    p->flow = &f;
+    p->flowflags |= FLOW_PKT_TOSERVER;
+    p->flowflags |= FLOW_PKT_ESTABLISHED;
+    p->flags |= PKT_HAS_FLOW | PKT_STREAM_EST;
+    f.alproto = ALPROTO_HTTP;
+
+    StreamTcpInitConfig(TRUE);
+
+    de_ctx = DetectEngineCtxInit();
+    if (de_ctx == NULL)
```
inliniac: FAIL_IF

Please update the rest of the unittests as well

src/source-nfq.c
```diff
@@ -507,6 +524,10 @@ static int NFQCallBack(struct nfq_q_handle *qh, struct nfgenmsg *nfmsg,
     PKT_SET_SRC(p, PKT_SRC_WIRE);
 
     p->nfq_v.nfq_index = ntv->nfq_index;
+    /* if bypass mask is set then we may want to bypass so set pointer */
+    if (nfq_config.bypass_mask) {
```
inliniac: so this mask is used as a global switch to see if bypass is enabled at all for nfq?

src/detect-filestore.c
```diff
@@ -408,3 +416,113 @@ static void DetectFilestoreFree(void *ptr)
...
+                               "(bypass; filestore; "
+                               "content:\"message\"; http_host; "
+                               "sid:1;)");
+    if (de_ctx->sig_list == NULL)
+        goto end;
+
+    SigGroupBuild(de_ctx);
+    DetectEngineThreadCtxInit(&th_v, (void *)de_ctx, (void *)&det_ctx);
+
+    SCMutexLock(&f.m);
```
inliniac: FLOWLOCK_WRLOCK

src/source-nfq.c
```diff
@@ -492,6 +501,14 @@ static void NFQReleasePacket(Packet *p)
     PacketFreeOrRelease(p);
 }
 
+static void NFQBypassCallback(Packet *p)
```
inliniac: this will need to act on the root packet in case of tunnel packets

src/source-nfq.c
regit: yes, it may be a bit implicit but I was seeing a new variable has bit too much.

regit:
ok, updating the branch.

PR #2280, opened 2016-09-25
app-layer-ssl: Validity dates from TLS certificates (v17)

thus:
Get validity dates (notBefore and notAfter) from TLS certificates.

New detection keywords (tls_cert_notbefore and tls_cert_notafter), added validity dates to extended JSON log and extended TLS log output, and new lua functions TlsGetCertNotBefore() and TlsGetCertNotAfter().

Updates:
- Try to omit OpenBSD 5.4 bug that is causing trouble
- Don't use strptime when timestamp is epoch, since OpenBSD and Cygwin strptime don't support %s.

This PR fixes the following issue:
- https://redmine.openinfosecfoundation.org/issues/1583

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/51
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/51


inliniac:
Merged through #2284, thanks Mats! Thanks for your patience with the portable time handling :)

PR #2281, opened 2016-09-26
pcap-log: seed ring buffer with existing files - v1

jasonish:
First cut at seending the pcap-log ring buffer with files already on disk with the goal of making it self maintaining over restarts.

I'm thinking it might be a good idea to force the timestamp at the end of the file.  This is the case with non-multi mode, but in multi mode you have to explicitly provide %t.  This would make %n the only variable that would make sense, as %i isn't consistent across restarts.  Thoughts?

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/337
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/342

### DIFF
--- a/src/log-pcap.c
+++ b/src/log-pcap.c
@@ -547,6 +547,27 @@ static void StatsMerge(PcapLogData *dst, PcapLogData *src)
     dst->profile_data_size += src->profile_data_size;
 }
 
+static void PcapLogDataFree(PcapLogData *pl)
+{
+
+    PcapFileName *pf;
+    while ((pf = TAILQ_FIRST(&pl->pcap_file_list)) != NULL) {
+        TAILQ_REMOVE(&pl->pcap_file_list, pf, next);
+        PcapFileNameFree(pf);
+    }
+    if (pl == g_pcap_data) {
+        for (int i = 0; i < MAX_TOKS; i++) {
+            if (pl->filename_parts[i] != NULL) {
+                SCFree(pl->filename_parts[i]);
+            }
+        }
+    }
+    SCFree(pl->h);
+    SCFree(pl->filename);
+    SCFree(pl->prefix);
+    SCFree(pl);
+}
+
 /**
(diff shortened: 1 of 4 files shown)


jasonish:
Please ignore the notice logging, just noticed I left a few in.

src/log-pcap.c
```diff
@@ -25,6 +25,10 @@
  * Pcap packet logging module.
  */
 
+#include <sys/types.h>
+#include <dirent.h>
```
inliniac: these will need checks in configure.ac and be included from suricata-common.h wrapped in HAVE_...H

inliniac:
%n thread number will be hard to use reliably if config/runmode changes.

Wrt forcing %t, I worry about breaking existing setups. Perhaps we can require it for the ringbuffer.

jasonish:
> %n thread number will be hard to use reliably if config/runmode changes.

Its reliable for repeated runs of the same config. Thats the best we can do, and its probably the norm in production.

> Wrt forcing %t, I worry about breaking existing setups. Perhaps we can require it for the ringbuffer.

Can you actually think of a valid use case where one would use multi without %t? Given its explicit use in normal?  But yes, will require it in multi/ring-buffer mode (as its explicit in normal mode).

IMO, too many possible config variations for this feature :)

So I think ring buffer mode will be more strict in terms of what the config can be.

src/log-pcap.c
jasonish: Oops, sys/types.h is not needed.  

Probably needs this for fnmatch as well (readdir is more likely to exist I believe).  What do we do if we don't have readdir? Just log a warning ring buffer won't pick up on restart?

inliniac:
I think we need a special 'deprecated' warning in the tree. Something that says the option will go away / get stricter because we believe no one / hardly anyone uses it. A 'speak up now' or the option goes away type of warning :)

PR #2278, opened 2016-09-24
Bypass v8

regit:
Update of #2276 fixing remarks.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/195
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/191

PR #2282, opened 2016-09-26
Bypass v9

regit:
Update of #2278.

Callback in capture method now return a status which allows to detect the bypass failure in lower level. A rebase has been done and flow flags had to be  updated to get some free room.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/199
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/195

### DIFF
--- a/src/detect-filestore.c
+++ b/src/detect-filestore.c
@@ -63,6 +63,7 @@ static int DetectFilestoreMatch (ThreadVars *, DetectEngineThreadCtx *,
         Flow *, uint8_t, File *, Signature *, SigMatch *);
 static int DetectFilestoreSetup (DetectEngineCtx *, Signature *, char *);
 static void DetectFilestoreFree(void *);
+static void DetectFilestoreRegisterTests(void);
 
 /**
  * \brief Registration function for keyword: filestore
@@ -75,7 +76,7 @@ void DetectFilestoreRegister(void)
     sigmatch_table[DETECT_FILESTORE].FileMatch = DetectFilestoreMatch;
     sigmatch_table[DETECT_FILESTORE].Setup = DetectFilestoreSetup;
     sigmatch_table[DETECT_FILESTORE].Free  = DetectFilestoreFree;
-    sigmatch_table[DETECT_FILESTORE].RegisterTests = NULL;
+    sigmatch_table[DETECT_FILESTORE].RegisterTests = DetectFilestoreRegisterTests;
     sigmatch_table[DETECT_FILESTORE].flags = SIGMATCH_OPTIONAL_OPT;
 
     DetectSetupParseRegexes(PARSE_REGEX, &parse_regex, &parse_regex_study);
@@ -291,6 +292,13 @@ static int DetectFilestoreSetup (DetectEngineCtx *de_ctx, Signature *s, char *st
     int ret = 0, res = 0;
     int ov[MAX_SUBSTRINGS];
 
+    /* filestore and bypass keywords can't work together */
+    if (s->flags & SIG_FLAG_BYPASS) {
+        SCLogError(SC_ERR_CONFLICTING_RULE_KEYWORDS,
+                   "filestore can't work with bypass keyword");
+        return -1;
+    }
+
     sm = SigMatchAlloc();
     if (sm == NULL)
         goto error;
@@ -407,4 +415,40 @@ static void DetectFilestoreFree(void *ptr)
     if (ptr != NULL) {
         SCFree(ptr);
     }
+}
+
+#ifdef UNITTESTS
+/*
+ * The purpose of this test is to confirm that
+ * filestore and bypass keywords can't
+ * can't work together
+ */
+static int DetectFilestoreTest01(void)
+{
+    DetectEngineCtx *de_ctx = NULL;
+    int result = 1;
+
+    de_ctx = DetectEngineCtxInit();
+    FAIL_IF(de_ctx == NULL);
+
+    de_ctx->flags |= DE_QUIET;
+
+    de_ctx->sig_list = SigInit(de_ctx,"alert http any any -> any any "
+                               "(bypass; filestore; "
+                               "content:\"message\"; http_host; "
+                               "sid:1;)");
+    FAIL_IF_NOT_NULL(de_ctx->sig_list);
+
+    if (de_ctx != NULL)
+        DetectEngineCtxFree(de_ctx);
+
+    return result;
+}
+#endif /* UNITTESTS */
+
+void DetectFilestoreRegisterTests(void)
+{
+#ifdef UNITTESTS
+    UtRegisterTest("DetectFilestoreTest01", DetectFilestoreTest01);
+#endif /* UNITTESTS */
 }
--- a/src/flow-manager.c
+++ b/src/flow-manager.c
@@ -264,6 +264,15 @@ static int FlowManagerFlowTimedOut(Flow *f, struct timeval *ts)
     }
 
     int server = 0, client = 0;
+
+    int state = SC_ATOMIC_GET(f->flow_state);
+    if (state == FLOW_STATE_BYPASSED)  {
+        if (FlowForceReassemblyNeedReassembly(f, &server, &client) == 1) {
+            FlowForceReassemblyForFlow(f, server, client);
+        }
+        return 1;
+    }
+
     if (!(f->flags & FLOW_TIMEOUT_REASSEMBLY_DONE) &&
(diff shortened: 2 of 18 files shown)


src/detect-filestore.c
```diff
@@ -408,3 +416,39 @@ static void DetectFilestoreFree(void *ptr)
...
+
+    de_ctx->flags |= DE_QUIET;
+
+    de_ctx->sig_list = SigInit(de_ctx,"alert http any any -> any any "
+                               "(bypass; filestore; "
+                               "content:\"message\"; http_host; "
+                               "sid:1;)");
+    FAIL_IF_NOT_NULL(de_ctx->sig_list);
+
+    if (de_ctx != NULL)
```
inliniac: de_ctx can't be NULL here

src/flow-manager.c
```diff
@@ -264,6 +264,15 @@ static int FlowManagerFlowTimedOut(Flow *f, struct timeval *ts)
     }
 
     int server = 0, client = 0;
+
+    int state = SC_ATOMIC_GET(f->flow_state);
+    if (state == FLOW_STATE_BYPASSED)  {
+        if (FlowForceReassemblyNeedReassembly(f, &server, &client) == 1) {
```
inliniac: I don't see why this commit is needed. The flow will be handled by the flow timeout mechanism rather soon after the bypass, as it's timeout is now really low.

inliniac: Commit message is hard to follow.

src/flow.c
```diff
@@ -225,28 +225,26 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
...
-                f->flags |= FLOW_TO_SRC_SEEN;
                 p->flowflags |= FLOW_PKT_TOCLIENT_FIRST;
             }
         }
+        f->tosrcpktcnt++;
+        f->tosrcbytecnt += GET_PKT_LEN(p);
     }
 
-    if ((f->flags & (FLOW_TO_DST_SEEN|FLOW_TO_SRC_SEEN)) == (FLOW_TO_DST_SEEN|FLOW_TO_SRC_SEEN)) {
+    if (f->tosrcpktcnt && f->todstpktcnt) {
```
inliniac: so the point of the flags logic was that we don't set 'established' on icmp error packets coming in. I think we lost that here.

src/flow-manager.c
regit: I did not get log after  30 sec. flow entry were only shown when shunting down suricata.

inliniac:
So I think the FLOW_BYPASS_FAILED flag needs some more thought. If we assume that the bypass can fail for multiple reasons:
1. capture method not supporting it
2. capture method not setup properly (e.g. no NFQ rule that actually enforces it)
3. capture method with limits (e.g. max 1000 nic hw filters)

Then we have to handle failure gracefully. I think we should unconditionally flag the flow as 'bypassed' and stop processing packets as early as possible. I'm thinking we exit 'FlowWorker' right after it looked up the flow and checked that this 'bypassed' flag is set.

Such a generic flag would replace the 'bypass failed' flag I think.

inliniac:
Another thing is that if we still continue to see packets the very short flow timeout connect to the bypass state may actually be harmful. If the flow is discard from suricata but more packets come in, we'll just create a new flow for that traffic.

src/flow-manager.c
inliniac: That is strange. Looks like this fix is hiding another issue then.

regit:
Yes on timeout, even if 5 seconds is long here. But if we have a flag that discard all new packets for the flow then that should be not anymore a potential problem.

inliniac:
Not sure I follow. If we have a 5 sec timeout, the flow is gone after 5 sec. What if a packet for that flow comes in after 10 sec? A new flow will be created. A ssh session would be an example of this. Of course we have the same issue with our regular timeouts, but they are set quite a bit higher by default.

regit:
If we omit Netfilter misconfiguration, the bypass call is acknowledged by the lower level so we should not receive any new packets from the flow after the call. Only already queued packets will be seen by suricata and a 5 seconds delay between 2 packets of the same flow is really unlikely to occur.
I fully agree on the early bypass comment, it seems a really good idea to exit as soon as possible and thus provide a "software" bypass fallback.

inliniac:
Another case may be that the capture method sets a timeout on the filter it uses, and may discard the filter for a flow too early. Or maybe it discards a filter because it needs to make space for new filters. In general I think the bypass should be handled completely inside suricata, and then we offload to capture if we're lucky.

regit:
We have a timing issue to consider if we do that: suppose the HW bypass method is bypassing for a while a flow then entry for the flow is removed. In that case we could see packets from a flow after a long time (without knowing if it was alive or not). So if we want to deal with that it means we should keep a bypassed flow forever or assume we can see an already bypassed flow as new if ever HW is playing a trick to us.

inliniac:
Yeah this is a trade off. Perhaps it would make sense to keep a 'light' flow around for a longer time. Just the flow, but all applayer/flowbit/etc stuff already freed.

regit:
So to summarize, I think we should implement a suricata-only bypass that get the bypassed packets out of flow worker just after the lookup.
But in case we have a lower level method that is doing the bypass, we may have a mixed timeout/lower level bypass stop that would cause some problem and make it almost useless. So I think the suricata-only bypass should be or disabled or should have low timeout value when used with a lower level bypass.

regit:
One other important point on callback failure. It could also be due to the fact that the protocol is not supported by callback method. For instance, AF_PACKET is not supporting now IPv6 so it was convenient to be able to say for the capture method to say, well I can't bypass. So it may really be interesting to make the distinction between failed bypassed and successful bypass.

src/detect-filestore.c
regit: ack

inliniac:
On 26-09-16 11:19, Eric Leblond wrote:
> 
> 
> In src/flow-manager.c <https://github.com/inliniac/suricata/pull/2282>:
> 
>> @@ -264,6 +264,15 @@ static int FlowManagerFlowTimedOut(Flow *f, struct timeval *ts)
>      }
>  
>      int server = 0, client = 0;
> +
> +    int state = SC_ATOMIC_GET(f->flow_state);
> +    if (state == FLOW_STATE_BYPASSED)  {
> +        if (FlowForceReassemblyNeedReassembly(f, &server, &client) == 1) {
> 
> I did not see that before a rebase on latest code. Could it be related
> to recent flow manager changes ?

It shouldn't but you never know :)

regit:
I propose to introduce two state in flow_state:

1. LOCAL_BYPASSED that is set when bypass callback is failing
2. CAPTURE_BYPASSED that is is set when bypass callback is a success

LOCAL_BYPASSED will trigger the same timeout policy as a ESTABLISHED and suricata will get packets out of flow worker after the hash lookup.

For CAPTURE_BYPASSED, we keep the same timeout policy as the one which is in implemented by this PR. 

This means we can get rid of the BYPASS FAILED which is replace by LOCAL_BYPASSED.

inliniac:
If LOCAL_BYPASSED is equal to ESTABLISHED, how to do transition to 'CLOSED'?

regit:
Right. We could have a reasonnably short timeout for instance inherited from emergency ?

regit:
Note: we discard the packets for the flow and don't do any tracking so we can't relay on closing info RST, FIN because we don't track TCP window and ack.

inliniac:
Might make sense to stay close to 'closed', but otoh lots of ppl set that to 0. New setting with a default of 30/60sec or so? Not very pretty.

regit:
I was thinking about using emergency established because it is a similar case: flow is there and we don't want to spend too much time with it.

inliniac:
I think in many cases it's set too low to be useful.

inliniac:
A key difference with regular flow handling is that performance here may very well depend on *long* timeouts. The shorter the timeout the higher the change of a new mid-stream flow based on bypassed traffic.

regit:
Well, if a bypassed flow don't send anything for 100 sec, actual default then we don't really care about it in most cases.

regit:
I've seen that it is not trivial to access the emergency timeout anymore. We could consider that the feature is important enough for users for them to have new timeout to set.

inliniac:
Yeah I think that makes sense.

regit:
ok, we will add a new variable for local bypass flow timeout.

PR #2285, opened 2016-09-26
lua: add function to check if flow has alerts

thus:

Example script:

``` lua
function init (args)
    local needs = {}
    needs["type"] = "flow"
    return needs
end

function setup (args)
end

function log (args)

    if has_alerts then
        file:flush()
    end
end

function deinit (args)
    file:close(file)
end
```

prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/52
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/52

### DIFF
--- a/src/detect.c
+++ b/src/detect.c
@@ -1943,6 +1943,12 @@ end:
     }
     PACKET_PROFILING_DETECT_END(p, PROF_DETECT_CLEANUP);
 
+    if (alerts > 0 && pflow) {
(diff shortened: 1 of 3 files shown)


inliniac:
Perhaps a eve field could be nice as well.

inliniac:
Needs rebase.

thus:
Replaced by #2483

PR #2300, opened 2016-09-27
Issue 1893 - TLS direction fixes - v1

jasonish:
Issue: https://redmine.openinfosecfoundation.org/issues/1893

The first commit fixes the direction on the TLS events to set the src_ip to be the client, and the dest_ip to be the server regardless of IDS or IPS mode.

The second commit adds direction to the fingerprint match.  Before this, if you and an any -> any rule it would trigger in the to_server and to_client directions.  It should only trigger in the to_client direction.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/341
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/346

### DIFF
--- a/src/detect-tls.c
+++ b/src/detect-tls.c
@@ -653,13 +653,20 @@ static int DetectTlsFingerprintMatch (ThreadVars *t, DetectEngineThreadCtx *det_
 
     int ret = 0;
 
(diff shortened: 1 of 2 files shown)


inliniac:
Merged through #2302, thanks Jason!

PR #2303, opened 2016-09-28
wirefuzz-fix-v2

pevma:
Add a switch to wirefuzz to use asan/lsan options.

### DIFF
--- a/qa/wirefuzz.pl
+++ b/qa/wirefuzz.pl
@@ -71,6 +71,7 @@ my $logdir;
 my $configfile;
 my $editeratio;
 my $valgrindopt;
+my $asan_options;
 my $shuffle;
 my $useltsuri;
 my $ltsuribin;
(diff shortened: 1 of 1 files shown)


inliniac:
It seems all this does is set an env var? Which leads to the question: why not set it directly?

pevma:
It offers an alternative (as opposed to setting and unsetting globally multiple times) to multiple QA runs combos on different OSs. For example gitlab with docker ( https://hub.docker.com/r/pevma/sqard/tags/ ) while fuzzing and unittesting with different asan options.

inliniac:
Whats the difference between:
ASAN_OPTIONS=detect_leaks=0 wirefuzz.pl ...
and
wirefull.pl -a=ASAN_OPTIONS=detect_leaks=0 ...

pevma:
Not sure if it makes a diff on diff OSs when used in wrapper scripts. Most likely not. I guess I was concentrated on each individual run line.

inliniac:
Not convinced this adds anything, so closing.

Btw, adding a feature and having 'fix' in the PR title is rather confusing. Please use more accurate titles.

PR #2304, opened 2016-09-28
nfq: fix coverity 1373412

regit:
Code was checking for p->root to be null and dereferencing it
afterwards.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/207
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/203

### DIFF
--- a/src/source-nfq.c
+++ b/src/source-nfq.c
@@ -504,7 +504,11 @@ static void NFQReleasePacket(Packet *p)
 static int NFQBypassCallback(Packet *p)
 {
     if (IS_TUNNEL_PKT(p)) {
-        SCMutex *m = p->root ? &p->root->tunnel_mutex : &p->tunnel_mutex;
+        if (p->root == NULL) {
+            return 0;
+        }
+
+        SCMutex *m = &p->root->tunnel_mutex;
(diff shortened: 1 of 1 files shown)


src/source-nfq.c
```diff
@@ -504,7 +504,11 @@ static void NFQReleasePacket(Packet *p)
 static int NFQBypassCallback(Packet *p)
 {
     if (IS_TUNNEL_PKT(p)) {
-        SCMutex *m = p->root ? &p->root->tunnel_mutex : &p->tunnel_mutex;
+        if (p->root == NULL) {
```
inliniac: This breaks the tunnel logic. The idea is that IS_TUNNEL_PKT returns true if this packet is either the root packet or the encapsulated packet. In case it's the root p->root == NULL (see IS_TUNNEL_ROOT_PKT), if p->root != NULL it's the inner packet. The mark should always be set on the root packet of course, as that is the only one getting verdicted by nfq.

inliniac:
Very likely same issue as https://redmine.openinfosecfoundation.org/issues/1917

PR #2306, opened 2016-09-28
doc: attempt to parse version if not in environment - v1

jasonish:
When the docs are built in tree the version is grabbed from the environment, this is not the case on readthedocs.  If the environment does not contain the version, then attempt to parse it from configure.ac. This appears to work fine on readthedocs, fixing the version in the HTML and PDF.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/342
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/347


inliniac:
Merged through #2324. Thanks Jason!

PR #2289, opened 2016-09-26
[preview] prefilter engines v97

inliniac:
Rebase of #2269 to current master.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/497
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/502

PR #2310, opened 2016-09-29
prefilter engines v99g

inliniac:
Large overhaul of internals of the detection engine. Cleans up registration and calling of prefilter engines (e.g. MPM) and 'inspect engines' (per rule). New keywords need to touch much fewer files.

Introduces more generic prefilter engines than MPM. Prefiltering can support rules looking ttl, icode, seq, window and many more. This should make non-pattern based rules much more attractive. By default this prefiltering is not yet enabled. Use 'prefilter' in rules (like fast_pattern after the keyword that it should apply to) or enable globally: "detect.prefilter.default=auto".

Adds 2 new keywords: http_request_line and http_response_line. Both are 'stickybuffers' like file_data. Both are MPM enabled.

Known issue:
- modbus inspection leads to duplicate alerts. cc: @daviddiallo 

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/500
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/505


inliniac:
This one replaced #2289

PR #2317, opened 2016-09-30
Modbus midstream v5

regit:
This is a series of fixes for problems @glongo and me have found when working on modbus protocol.

Main one is a fix of midstream algorithm that was not detecting correctly the server/client inversion. Commits a085fb9 to 3099ce2 are addressing that.

The two last patches fixe a problem in modbus not continuing correctly transaction inspection in case of message loss.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/217
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/213

### DIFF
--- a/src/app-layer-modbus.c
+++ b/src/app-layer-modbus.c
@@ -43,6 +43,7 @@
 #include "util-misc.h"
 
 #include "stream.h"
+#include "stream-tcp.h"
 
 #include "app-layer-protos.h"
 #include "app-layer-parser.h"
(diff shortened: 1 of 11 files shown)


inliniac:
I think this needs a much more fundamental approach.

1. the app layer protocol detect code needs to return a direction indicator that also works for pattern based detection (e.g. GET|20| on TOCLIENT likely means we're HTTP and in the wrong dir)
2. a single place (or perhaps one for udp and one for tcp) should check this and switch over flow _and_ packet completely
3. switching should be much more comprehensive. Luckily now that we have flow workers we can do it completely under one lock so no sync issues. But think about:
- packet flow flags
- flow packet counters, byte counters, alproto_ts/tc fields, flow flags, etc, etc. 
- it should also replace the TCP only flipping that currently exists

When both current packet and flow are flipped properly there is no need for a special flag anymore I think.

src/app-layer-modbus.c
```diff
@@ -1342,8 +1343,13 @@ static int ModbusParseResponse(Flow                 *f,
             if (tx == NULL)
                 SCReturnInt(0);
 
-            SCLogDebug("MODBUS_DECODER_EVENT_UNSOLICITED_RESPONSE");
-            ModbusSetEvent(modbus, MODBUS_DECODER_EVENT_UNSOLICITED_RESPONSE);
+            /* check that midstream is enable and that we are at start of flow */
+            if (StreamTcpMidstreamIsEnabled() && (f->data_al_so_far[0] == 0)) {
+                AppLayerParserManualSetTransactionInspectId(pstate, header.transactionId + 1, 0);
```
inliniac: Can you send me a test case for this? I would really hate to need a hack like this. This should stay hidden from the protocol implementation.

regit: Done by mail.

inliniac: A protocol implementation can register the data in which it expects the first data. Normally toserver. I think that if in midstream, when seeing the first data in the wrong direction, we can automagically handle these things 'under the hood'

inliniac:
Needs a new approach.

PR #2322, opened 2016-10-01
eve: output log to Apache kafka

fooinha:
Initial work for logging output into the distributed streaming platform Apache Kafka.

Allows to send eve messages to a list of brokers of Apache Kafka.
It uses the client implementation from [librdkafka](https://github.com/edenhill/librdkafka).

It's set to be conditional at configure/compile time with the  --enable-rdkafka [Enable Kafka support] option.

Configuration settings and default values: 

```yaml
      filetype: regular #regular|syslog|unix_dgram|unix_stream|redis|kafka
       filename: eve.json
       kafka:
        broker-list: 127.0.0.1:9092
        topic: suricata
        compression: snappy # none, gzip, snappy
        partition: -1 # below < 0 means Unassigned Partition
        max-retries: 1
        backoff-ms: 10
        buffer-max-messages: 2048
        log-level: 6
```

It's an alternative for redis output. 

FIrst implementation using librdkafka1 version 0.8.1 ( 0.8.6-1.1 -> Ubuntu Xenial )


fooinha:
The build job will need to be changed in order to include the librdkafka-dev package and the --enable-rdkafka configure option

inliniac:
The build job at Travis CI is controlled by .travis.yml, so you can add your test options there.

Can you have a look at why the shutdown crash of the commented out code happens?

fooinha:
The shutdown crash happens inside librdkafka when  freeing configuration propertries.

```c
(gdb) signal SIGINT
Continuing with signal SIGINT.
[1816] 3/10/2016 -- 06:57:52 - (suricata.c:2607) <Notice> (main) -- Signal Received.  Stopping engine.
....

*** Error in `/root/suricata/src/suricata': free(): invalid pointer: 0x000000000070ade3 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x77725)[0x7ffff5f78725]
/lib/x86_64-linux-gnu/libc.so.6(+0x7ff4a)[0x7ffff5f80f4a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7ffff5f84abc]
/usr/lib/x86_64-linux-gnu/librdkafka.so.1(+0x1b0a7)[0x7ffff7bc30a7]
/usr/lib/x86_64-linux-gnu/librdkafka.so.1(rd_kafka_conf_destroy+0x11)[0x7ffff7bc30e1]
/root/suricata/src/suricata[0x69f35c]

#6  0x00007ffff7bc30a7 in rd_kafka_anyconf_clear (prop=0x7ffff7dd2ca0 <rd_kafka_properties+224>, prop=0x7ffff7dd2ca0 <rd_kafka_properties+224>, conf=0xaa72560) at rdkafka_defaultconf.c:719
719				free(*str);
(gdb) l
714		{
715		case _RK_C_STR:
716		{
717			char **str = _RK_PTR(char **, conf, prop->offset);
718			if (*str) {
719				free(*str);
720				*str = NULL;
721			}
722		}
723		break;
```
I've changed the .travis.yaml.

Thanx

src/util-logopenfile.c
```diff
@@ -516,6 +870,12 @@ int LogFileFreeCtx(LogFileCtx *lf_ctx)
             SCFree(lf_ctx->redis_setup.key);
     }
 #endif
+#ifdef HAVE_LIBRDKAFKA
+
+    if (lf_ctx->type == LOGFILE_TYPE_KAFKA) {
+        SCLogFileCloseKafka(lf_ctx);
```
inliniac: Should this be done under lock? I see redis doesn't do it, but that may be a mistake.

fooinha: Placed the lock here.

Thanx.

fooinha:
Added ppa for librdkafka ( 0.8.1 ) ... with precise builds.

src/util-logopenfile.c
fooinha: By the way the crash was due a bad termination sequence for rdkafka.

https://github.com/edenhill/librdkafka/wiki/Proper-termination-sequence

inliniac:
Can you please follow our guide and use new PRs + new branches for changes? https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow

PR #2333, opened 2016-10-05
prelude: add protocol information through JSON

ToToL:
Prelude is based on what can be found in packet but now, Suricata provide more information in JSON format for HTTP, TLS, SSH and SMTP. This PR add the support for prelude to read these JSON and integrate them into IDMEF alerts.

### DIFF
--- a/src/alert-prelude.c
+++ b/src/alert-prelude.c
@@ -53,6 +53,12 @@
 #include "util-print.h"
 
 #include "output.h"
+#include "output-json.h"
+#include "output-json-http.h"
+#include "output-json-tls.h"
+#include "output-json-ssh.h"
+#include "output-json-smtp.h"
+#include "output-json-email-common.h"
 #include "util-privs.h"
 #include "util-optimize.h"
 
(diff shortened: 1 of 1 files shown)


inliniac:
In general I'm not a fan of the approach of using the output of one (eve) to generate the output of the next (prelude). But on the other hand, I can see how it's convenient.

@jasonish @regit any thoughts?

regit:
Good with me. Json is not going out before a while so dependency is not a problem I think.

jasonish:
I didn't like it at first, but after more thought I'm not sure if one output depending on the other, in particular where the output that is depended on is first class to Suricata, that it should be too much of an issue?

I guess there is a chance that changes to eve could break this output. Who's responsibility is it to fix up in that case?

inliniac:
I plan to follow this: https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Support_Status#Outputs

So then it would be the responsibility of @ToToL, if he accepts it :)

ToToL:
No problem to support this :) I'm in !

I though I understood that JSON was the "internal" output. That's why I use it. If you prefer I do it in an other way, tell me how to.

inliniac:
There isn't really a unified way we store things internally. The JSON data is filled by manually calling lots of protocol specific code. So I guess there are 3 options
1. keep as is
2. call the various per protocol data retrievers manually
3. add a new common middle layer

Probably best to stick with what you did for now. Like Eric said the EVE format is not likely to change in major ways.

inliniac:
Closing this as it needs a rebase and some refactoring as the locking has changed since this was made.

Btw, I'll try to respond a bit quicker on the next PR. My goal is to half my response time ;)

inliniac:
On 08-06-17 15:05, Thomas ANDREJAK wrote:
> rebase is simple but can you tell me more about locking ? what has change ?

In the current code if you have p->flow != NULL it means the flow is locked.

ToToL:
so I can't read the flow ?

Sorry I don't really understand what I have to change.

src/alert-prelude.c
```diff
@@ -467,11 +586,95 @@ static int PacketToDataV6(const Packet *p, const PacketAlert *pa, idmef_alert_t
...
+    SshState *ssh_state;
+    SSLState *ssl_state;
+
     SCEnter();
 
     if (unlikely(p == NULL))
         SCReturnInt(0);
 
+    if (p->flow != NULL) {
+        FLOWLOCK_RDLOCK(p->flow);
```
inliniac: you can drop this lock. If there is a flow (meaning p->flow != NULL) you are guaranteed it's already locked before your code was invoked. So you don't need to consider flow locking and can just use the flow.

inliniac:
On 08-06-17 15:42, Thomas ANDREJAK wrote:
> so I can't read the flow ?
> 
> Sorry I don't really understand what I have to change.

See https://github.com/inliniac/suricata/pull/2333#discussion_r120895465

Hope this helps make clear what I meant :)

ToToL:
yes, thanks : https://github.com/ToToL/suricata/commit/7581bc304941e72d842be9038803172a7afbe6e3

I have to test the build then I will push this

PR #2334, opened 2016-10-05
Prelude update

ToToL:
Update many things around Prelude. More clean code. More macro.
Also, add IPV6 support for prelude alerts


inliniac:
Merged through #2353, thanks!

PR #2335, opened 2016-10-05
[preview] flow file flags v3

inliniac:
Attempt to address flow flag pressure.
Plus some other fixes.


inliniac:
Merged in #2343

PR #2330, opened 2016-10-04
doc: exclude docs in partials/ from reference errors

jasonish:
These docs are already included with the include statement,
but older versions of Sphinx still complain that they
are not in a table of contents.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/8
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/358

PR #2336, opened 2016-10-05
doc - fixups - v4

jasonish:
Previous PR: #2330 

- Explicitly exclude partials/ from as they are not intended to be used in a TOC, but only included by other files. This is to keep older versions of Sphinx from erroring out, newer versions handle this case fine.
- Documents were missing from EXTRA_DIST causing make distcheck to fail.
- The wrong pdf file name was being removed causing make distcheck to fail.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/14
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/364


inliniac:
Thanks Jason!

PR #2339, opened 2016-10-06
logging: don't log that json is disabled in each logger - v1

jasonish:
A warning log is already emitted if eve-log is enabled in the
configuration but json support is not built so the logger
registration functions can be silent.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/15
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/365


inliniac:
Merged in #2343, thanks Jason!

PR #2340, opened 2016-10-06
macOS: thread return value affects newer macOS as well - v1

jasonish:
ALl OS X/macOS versions since 10.10 return EDEADLK here instead
of EBUSY. Assume they will moving forward as well.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/16
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/366


inliniac:
Merged in #2343, thanks Jason!

PR #2344, opened 2016-10-07
Bypass encrypted fix

regit:
This patchset is a kind of RFC. It allows to trigger bypass early for TLS flows. I feel a bit unease with the choices so discussion is welcome.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/218
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/0

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -1213,6 +1213,13 @@ static int SSLv3Decode(uint8_t direction, SSLState *ssl_state,
                handshake must be done */
             ssl_state->flags |= SSL_AL_FLAG_HANDSHAKE_DONE;
 
+            /* Encrypted data, reassembly not asked, bypass asked, let's sacrifice
+             * heartbeat lke inspection to be able to be able to bypass the flow */
+            if ((ssl_config.no_reassemble == 1) && (StreamTcpBypassEnabled())) {
+                AppLayerParserStateSetFlag(pstate, APP_LAYER_PARSER_NO_REASSEMBLY);
+                AppLayerParserStateSetFlag(pstate, APP_LAYER_PARSER_NO_INSPECTION);
+            }
+
             break;
 
         case SSLV3_HANDSHAKE_PROTOCOL:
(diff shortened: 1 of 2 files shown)


inliniac:
Can you explain the issue first, including how to reproduce it?

src/app-layer-ssl.c
```diff
@@ -1213,6 +1213,13 @@ static int SSLv3Decode(uint8_t direction, SSLState *ssl_state,
                handshake must be done */
             ssl_state->flags |= SSL_AL_FLAG_HANDSHAKE_DONE;
 
+            /* Encrypted data, reassembly not asked, bypass asked, let's sacrifice
+             * heartbeat lke inspection to be able to be able to bypass the flow */
+            if ((ssl_config.no_reassemble == 1) && (StreamTcpBypassEnabled())) {
```
inliniac: Not sure I follow the logic here. Why should this depend on bypass? Isn't the purpose of the no_reassemble option to disable tracking of ssl after the handshake?

regit:
Regarding first commit, my thinking is the following, we should not completely bypass a flow before both end have reached the stream depth or have reached a certain state. Justification is that we need the hack to treat the other side so we can't really decide to cut only one side.

About second commit, I was doing some tests lately on local bypass, I've been surprised to see that TLS session were not seen in the flow entry as bypassed. This was due to the fact that the inspection is done till stream depth so Suricata get for example heartbeat message.

src/app-layer-ssl.c
regit: The option is disabling tracking but I was thinking that maybe some people want anyway to do packet matching. So I've chooze the most careful solution. My initial implementation was just setting the flags.

inliniac: I reviewed the logic and I think we can do this without the bypass check.

inliniac:
Commit message of https://github.com/inliniac/suricata/pull/2344/commits/64f1730565ada2fd026686e0d292c1f73293bc4c needs some love.

inliniac:
Needs an update.

PR #2348, opened 2016-10-09
eve dns json output - style , alerts and unit tests.

fooinha:
eve: adds dns fields for eve alert's json output

eve: more compact log for dns output json ( as defined by @decanio  ) 

```
DNS log output format is selected using new outputs.eve-log.types.dns.style configuration option

- dns:
    # The style of logging:
    #    discrete: the classic style of an event per question and answer
    #    split: one event per request, one event per response
    #    unified: one event containing request and response
    style: discrete

To emit a single DNS log per DNS transacton use style: unified

To emit a pair of DNS logs per DNS transaction use style: split
the latter preserves individual timestamps of the query and answer arrival times.

The classic DNS log format is preseved using style: discrete

DNS filtering capabilities are preserved with all log styles.
```

eve: dns json output log unit tests

### DIFF
--- a/src/app-layer-dns-udp.c
+++ b/src/app-layer-dns-udp.c
@@ -53,7 +53,7 @@
 /** \internal
  *  \brief Parse DNS request packet
  */
-static int DNSUDPRequestParse(Flow *f, void *dstate,
+int DNSUDPRequestParse(Flow *f, void *dstate,
                               AppLayerParserState *pstate,
                               uint8_t *input, uint32_t input_len,
                               void *local_data)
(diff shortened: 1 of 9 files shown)


inliniac:
Does this include @decanio 's actual code changes? If so you can't remove/squash his commit(s).

fooinha:
It includes my implementation of @decanio idea, without using his code.

inliniac:
See https://github.com/inliniac/suricata/pull/2356#issuecomment-311621438

We'd like to think some more about a new format.

PR #2349, opened 2016-10-11
Bypass encrypted fix v2

regit:
Update of #2344 updating a commit message, fixing a test and rebasing to latest master.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/219
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/1

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -1213,6 +1213,13 @@ static int SSLv3Decode(uint8_t direction, SSLState *ssl_state,
                handshake must be done */
             ssl_state->flags |= SSL_AL_FLAG_HANDSHAKE_DONE;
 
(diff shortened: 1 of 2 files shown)


inliniac:
This patch makes my heartbleed tests fail. What the fix exposes is that we have 'no-reassemble' enabled by default, but we didn't notice because it wasn't enforced for SSL3+.

Since I don't want to break Heartbleed detection rules, I guess we can switch to 'disabled' by default for 'no-reassemble'.

inliniac:
Merged through #2353, adding https://github.com/inliniac/suricata/pull/2353/commits/b789d2ae3d13cdb0d0bf6a5e40eafa96b82ced04 to disable the default 'no-reassembly' setting.

PR #2351, opened 2016-10-12
Fix port parsing in config file, added one more corresponding test.

gozzy:
Some examples from wiki caused parsing errors.
For example, "[1:80,![2,4]]" was treated as a mistake.

Also fixed loop detection in variables declaration.

### DIFF
--- a/src/detect-engine-port.c
+++ b/src/detect-engine-port.c
@@ -1121,6 +1121,8 @@ static int DetectPortParseDo(const DetectEngineCtx *de_ctx,
                     goto error;
             }
             n_set = 0;
+        } else if (depth == 1 && s[u] == ',') {
+            range = 0;
         }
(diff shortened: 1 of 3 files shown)


src/detect-engine-port.c
```diff
@@ -2391,6 +2394,50 @@ static int PortTestFunctions06(void)
 }
 
 /**
+ * \test Test general functions
+ */
+static int PortTestFunctions07(void)
```
inliniac: Can you rewrite this to use the new FAIL_IF/PASS macro's? See https://github.com/inliniac/suricata/blob/master/src/detect-tls-cert-validity.c#L643 for a simple example.

gozzy: Yes. sure.

PR #2354, opened 2016-10-12
Fix port parsing in config file, added one more corresponding test V3

gozzy:
Some examples from wiki caused parsing errors.
For example, "[1:80,![2,4]]" was treated as a mistake.

Also fixed loop detection in variables declaration. For example,
'A: "HOME_NET, !$HOME_NET"' resulted in parsing error.

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -826,12 +826,6 @@ static int DetectAddressParse2(const DetectEngineCtx *de_ctx,
     char *rule_var_address = NULL;
     char *temp_rule_var_address = NULL;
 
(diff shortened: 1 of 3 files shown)


gozzy:
Should be better now, I could not make it crash anymore. Maybe someone is luckier than me... :)

inliniac:
Merged in #2440, thanks Alex!

inliniac:
Btw is there a ticket for this issue? If not can you open it?

gozzy:
I've just opened a bug #1967 for this.

inliniac:
https://redmine.openinfosecfoundation.org/issues/1967

PR #2356, opened 2016-10-12
output-json-dns: implment more compact DNS log format

decanio:
DNS log output format is selected using new outputs.eve-log.types.dns.style
configuration option

```
- dns:
    # The style of logging:
    #    discrete: the classic style of an event per question and answer
    #    split: one event per request, one event per response
    #    unified: one event containing request and response
    style: discrete
```

To emit a single DNS log per DNS transacton use style: unified

To emit a pair of DNS logs per DNS transaction use style: split

the latter preserves individual timestamps of the query and answer arrival
times.

The classic DNS log format is preseved using style: discrete

DNS filtering capabilities are preserved with all log styles.

https://buildbot.openinfosecfoundation.org/builders/decanio/builds/34
https://buildbot.openinfosecfoundation.org/builders/decanio-pcap/builds/0

### DIFF
--- a/doc/userguide/configuration/suricata-yaml.rst
+++ b/doc/userguide/configuration/suricata-yaml.rst
@@ -314,6 +314,11 @@ integration with 3rd party tools like logstash.
             # the example below adds three additional fields when uncommented
             #custom: [Accept-Encoding, Accept-Language, Authorization]
         - dns:
+            # The style of logging:
+            #    discrete: the classic style of an event per question and answer
+            #    split: one event per request, one event per response
+            #    unified: one event containing request and response
+            style: discrete
             # control logging of queries and answers
             # default yes, no to disable
             query: yes     # enable logging of DNS queries
--- a/src/output-json-dns.c
+++ b/src/output-json-dns.c
@@ -898,6 +898,9 @@ static int JsonDnsLoggerToClient(ThreadVars *tv, void *thread_data,
         DNSQueryEntry *query = NULL;
         TAILQ_FOREACH(query, &tx->query_list, next) {
 
+            if (unlikely(dnslog_ctx->filter & LOG_ANSWERS) == 0)
+                continue;
+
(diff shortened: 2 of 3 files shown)


jasonish:
Previous PR: https://github.com/inliniac/suricata/pull/2275

jasonish:
[dns-nxdomain.pcap.zip](https://github.com/inliniac/suricata/files/529916/dns-nxdomain.pcap.zip)
Discrete support looks OK, no change from existing behaviour.

I am still seeing some lack of information in split mode. Attached is a simple DNS request returning NXDOMAIN.  In split mode I don't have the rrname - https://gist.github.com/jasonish/5d8ebe85547dff5a2f9142328187ee2f.

The same PCAP in discrete mode produces one request record and 2 response records (https://gist.github.com/jasonish/5fb2232b1f6aa626751efcadf2e2be58).

With unified I have the request context in the same event, but the answer list only contains one item (https://gist.github.com/jasonish/43597e380fc69497d62915f7fa24cf69).  There does seem to be a discrepancy from discrete, is there an obvious reason for this?

The pcap is attached.

jasonish:
Looks like all issues mentioned here, https://github.com/inliniac/suricata/pull/2275#issuecomment-251138985 are resolved with the exception of the docs.

inliniac:
Doc needs to be a bit more than just pasting the example config in :)

jasonish:
Note that eve is also documented in the output chapter:
http://suricata.readthedocs.io/en/latest/output/index.html

inliniac:
@decanio @jasonish is this something that you'd like to see in 4.0? If so can either of you rebase it and add some docs soon?

jasonish:
I'm indifferent for now.

I would be hesitant to add anything like "unified" at this time though. The Rust DNS experiment does not correlate responses to requests and I believe this is better in high volume/flood scenarios.

inliniac:
I think we need to think a bit more on how to do a new log format. For example multiple txt sub-records currently lead to new eve records.

PR #2361, opened 2016-10-15
Dev bug1919 hashing v0.1

serializingme:
Fixed the bug reported in the issue tracker (https://redmine.openinfosecfoundation.org/issues/1919) by ensuring that only a sequence of valid characters is sent to be parsed as an hash. As an example, follows the contents of a file supplied with the `filemd5` keyword:

> \# Comment example
> 000193a93dc5f3ee945704754d6
> 
> 000393a93dc5f3ee945704754d6e0a36
> 000493a93dc5f3ee945704754d6e0a36 # Comment example
> 000593a93dc5f3ee945704754d6e0a36# Comment example
> 000693a93d 5f3ee945704754d6e0a36
> 000793a93dc5f3ee945704754d6e0a;6

The line filtering will ensure the lines that are sent to be parsed are as follows:

> 000193a93dc5f3ee945704754d6
> 000393a93dc5f3ee945704754d6e0a36
> 000493a93dc5f3ee945704754d6e0a36
> 000593a93dc5f3ee945704754d6e0a36
> 000693a93d
> 000793a93dc5f3ee945704754d6e0a

The hash parsing function will then ensure the hexadecimal representation is correct for the specified hash (e.g. it will check that it is 32 characters in length and that each byte doesn't fall outside the 0 to 256 range).

### DIFF
--- a/src/util-detect-file-hash.c
+++ b/src/util-detect-file-hash.c
@@ -193,6 +193,8 @@ int DetectFileHashMatch (ThreadVars *t, DetectEngineThreadCtx *det_ctx,
     SCReturnInt(ret);
 }
 
+static const char *hexcodes = "ABCDEFabcdef0123456789";
+
 /**
  * \brief Parse the filemd5, filesha1 or filesha256 keyword
  *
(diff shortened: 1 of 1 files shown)


inliniac:
Merged in #2364, thanks Duarte!

PR #2369, opened 2016-10-20
Dev eve output redis async v2

fooinha:
Initial support for reading redis replies asynchronously.
The libevent adapter for the hiredis async API is used.

Command pipelining for the async: true mode is not implemented, and it only will be added after discussion and review.

Changes after review in #2362 

Thanx


src/util-logopenfile.c
```diff
@@ -592,26 +863,35 @@ static int  LogFileWriteRedis(LogFileCtx *file_ctx, const char *string, size_t s
...
-                SCConfLogReopenRedis(file_ctx);
-                break;
-            case REDIS_REPLY_INTEGER:
-                SCLogDebug("Redis integer %lld", reply->integer);
-                break;
-            default:
-                SCLogError(SC_ERR_INVALID_VALUE,
-                        "Redis default triggered with %d", reply->type);
+        /* We may lose the reply if disconnection happens*/
+        if (reply)  {
```
regit: Could you please put this part in a separate commit as this is a fix we need to backport to stable.

fooinha:
Replaced by #2379.

PR #2374, opened 2016-10-23
Add NULL-terminator to app-layer template (fix #1930)

chifflier:
Fix the template app-layer: event table has to be NULL-terminated


inliniac:
Merged through #2815, thanks Pierre!

PR #2376, opened 2016-10-24
eve: check redis reply. We may lose the reply if disconnection happens.

fooinha:
Checks redis reply pointer on no pipeline mode.

We may lose the reply if disconnection happens. And we need to reconnect if that happens.

Separate commit as suggested by @regit 

Thanx.


inliniac:
Can you make sure the commit message follows the guidelines? Please see https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits

Thanks!

PR #2377, opened 2016-10-24
eve: check redis reply in non pipeline mode

fooinha:
We may lose the reply if disconnection happens.
Reconnection is needed.

Different commit as @regit suggested.

Better commit message.

Thanx

PR #2379, opened 2016-10-24
Dev eve output redis async v3

fooinha:
Initial support for reading redis replies asynchronously.
The libevent adapter for the hiredis async API is used.

Command pipelining for the async: true mode is not implemented, and it only will be added after discussion and review.

Thanx @regit for review.

Replaces #2362 and  #2369.
It is merged with dependent PR #2377.

Thanx


fooinha:
Replaced by #2385

PR #2380, opened 2016-10-25
dns: support back to back requests without a response - v1

jasonish:
Address the issue where a DNS response would not be logged when
the traffic is like:
- Request 1
- Request 2
- Response 1
- Response 2
which can happen on dual stack machines where the request for A
and AAAA are sent out at the same time on the same UDP "session".

A "window" is used to set the maximum number of outstanding
responses before considering the olders lost.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/30
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/380

### DIFF
--- a/src/app-layer-dns-common.c
+++ b/src/app-layer-dns-common.c
@@ -31,6 +31,10 @@
 #include "util-memcmp.h"
 #include "util-atomic.h"
 
+/* The number of outstanding replies to allow on a session before
+ * marking the responses as lost. */
+#define UNREPLIED_WINDOW 2
+
 typedef struct DNSConfig_ {
     uint32_t request_flood;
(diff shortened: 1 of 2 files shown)


pevma:
I had exactly the pcap for the job :) to which I run some tests. It seems it does the job.
With the PR:  
> grep '"event_type":"dns"' eve.json  |wc -l
> 4529961

with current master (3.2dev (rev 397c541)) -

> grep '"event_type":"dns"' eve.json  |wc -l
> 3622168

inliniac:
Replaced by #2386

PR #2382, opened 2016-10-25
nfq fixes v2

inliniac:
https://github.com/inliniac/suricata/pull/2368 with support for defrag reassembled packets and a fix for nfq_set_mark keyword.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/17
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/525

### DIFF
--- a/src/detect-mark.c
+++ b/src/detect-mark.c
@@ -222,9 +222,18 @@ int DetectMarkPacket(ThreadVars *t, DetectEngineThreadCtx *det_ctx, Packet *p, S
 #ifdef NFQ
     const DetectMarkData *nf_data = (const DetectMarkData *)ctx;
     if (nf_data->mask) {
-        p->nfq_v.mark = (nf_data->mark & nf_data->mask)
-                        | (p->nfq_v.mark & ~(nf_data->mask));
-        p->flags |= PKT_MARK_MODIFIED;
+        if (!(IS_TUNNEL_PKT(p))) {
+            p->nfq_v.mark = (nf_data->mark & nf_data->mask)
+                | (p->nfq_v.mark & ~(nf_data->mask));
+            p->flags |= PKT_MARK_MODIFIED;
+        } else {
+            Packet *tp = p->root ? p->root : p;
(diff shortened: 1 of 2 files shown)


inliniac:
@regit can you have a look?

regit:
I'm good with all but the root pkt mark as it will break backward compatibility. But I'm not sure we have users in that case.

inliniac:
Replaced by #2383

inliniac:
On 25-10-16 17:24, Eric Leblond wrote:
> 
> 
> In src/detect-mark.c
> <https://github.com/inliniac/suricata/pull/2382#pullrequestreview-5675543>:
> 
>> @@ -222,9 +222,18 @@ int DetectMarkPacket(ThreadVars *t, DetectEngineThreadCtx *det_ctx, Packet *p, S
>  #ifdef NFQ
>      const DetectMarkData *nf_data = (const DetectMarkData *)ctx;
>      if (nf_data->mask) {
> -        p->nfq_v.mark = (nf_data->mark & nf_data->mask)
> -                        | (p->nfq_v.mark & ~(nf_data->mask));
> -        p->flags |= PKT_MARK_MODIFIED;
> +        if (!(IS_TUNNEL_PKT(p))) {
> +            p->nfq_v.mark = (nf_data->mark & nf_data->mask)
> +                | (p->nfq_v.mark & ~(nf_data->mask));
> +            p->flags |= PKT_MARK_MODIFIED;
> +        } else {
> +            Packet *tp = p->root ? p->root : p;
> 
> I'm not sure of this one. We have a case similar as bypass: If we do
> that on GRE tunnel, we could have connmark implemented and result will
> be a all sessions inside the GRE tunnel to be marked. This could be a
> real problem for users with these kind of configuration.

We'll we just mark the packet, the iptables/nftables ruleset then
determines what is done with it.

But I'll limit it to defrag packets like with bypass for now.

PR #2375, opened 2016-10-23
yara: initial support for yara rules match

fooinha:
Adds basic support for yara rules matching to suricata file signatures.
Feature redmine ticket 1828 -> https://redmine.openinfosecfoundation.org/issues/1828.

Example for rule ... For now only if any match occurs. 

`alert http any any -> any any (msg:"YARA MATCHED"; yara; sid:2; rev:1;)`

_yara_ keyword options should be discussed. 

Example for configuration and rules loading.

```yaml
#
# Yara rules
#
default-yara-rules-path: /etc/suricata/rules/yara
yara-rules-files:
 - jar.yara
 - eicarupdatetest.yara

```

Logging and debug:
```
[17252] 23/10/2016 -- 15:56:47 - (detect-yara-rule.c:291) <Info> (YaraRuleScanCallback) -- Tag rule:[java_01] tag:[compression]
[17252] 23/10/2016 -- 15:56:47 - (detect-yara-rule.c:291) <Info> (YaraRuleScanCallback) -- Tag rule:[java_01] tag:[forensic]
[17252] 23/10/2016 -- 15:56:47 - (detect-yara-rule.c:291) <Info> (YaraRuleScanCallback) -- Tag rule:[java_01] tag:[filetype]
[17252] 23/10/2016 -- 15:56:47 - (detect-yara-rule.c:294) <Info> (YaraRuleScanCallback) -- Matched rule [java_01]

10/23/2016-15:56:47.874971  [**] [1:2:1] YARA MATCHED [**] [Classification: (null)] [Priority: 3] {TCP} 151.101.16.209:80 -> 10.10.10.2:60352

```
Notes: last libyara version has some basic issues. https://github.com/VirusTotal/yara/issues/541

It adds DetectLoadCompleteSigPathWithDefaultPath that could or should be moved to another PR.

PR #2388, opened 2016-10-26
dev yara rules v2

fooinha:
Adds basic support for yara rules matching to suricata file signatures.
Feature redmine ticket 1828 -> https://redmine.openinfosecfoundation.org/issues/1828.

Replaces #2375.

_yara_ keyword optional options:

```
file:<filename> evaluates rules from a dedicated yara rules file
         if file is not specified global rules will be used.

rules:<rule-list> only alert if matching rule exists in this list

tags:<tags-list> only alert if maching rule have any tag from the list


```

Example for rules ... 

```
alert http any any -> any any (msg:"Match any global with rule name"; yara: rules:java,java_01; sid:2; rev:1;)
alert http any any -> any any (msg:"Match any global with tag"; yara: tags:compression,foo,bar,deadbeef; sid:3; rev:1;)
alert http any any -> any any (msg:"Match jar yara"; yara: file:jar.yar; sid:4; rev:1;)

```

Example for configuration and rules loading.

``` yaml
#
# Yara rules
#
default-yara-rules-path: /etc/suricata/rules/yara
yara-rules-files:
 - jar.yara
 - eicarupdatetest.yara

```

Logging and debug:

```
10/27/2016-23:42:35.167714  [**] [1:2:1] Match any global with rule name [**] [Classification: (null)] [Priority: 3] {TCP} 151.101.16.209:80 -> 10.10.10.2:51964
10/27/2016-23:42:35.167714  [**] [1:3:1] Match any global with tag [**] [Classification: (null)] [Priority: 3] {TCP} 151.101.16.209:80 -> 10.10.10.2:51964
10/27/2016-23:42:35.167714  [**] [1:4:1] Match jar yara [**] [Classification: (null)] [Priority: 3] {TCP} 151.101.16.209:80 -> 10.10.10.2:51964

```

Notes: last libyara version has some basic issues. https://github.com/VirusTotal/yara/issues/541

It adds DetectLoadCompleteSigPathWithDefaultPath that could or should be moved to another PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,12 +55,21 @@ matrix:
         apt:
           sources:
             - sourceline: ppa:npalix/coccinelle
+            - sourceline: ppa:fooinha/libyara3
           packages:
             - *packages
             - coccinelle
+            - libyara-dev
     # Linux, gcc, -DNDEBUG.
     - os: linux
       compiler: gcc
+      addons:
+        apt:
+          sources:
+            - sourceline: ppa:fooinha/libyara3
+          packages:
+            - *packages
+            - libyara-dev
       env:
         - NAME="linux,gcc,ndebug"
         - *default-cflags
@@ -68,6 +77,13 @@ matrix:
     # Linux, clang. For this build we'll also enable -Wshadow.
     - os: linux
       compiler: clang
+      addons:
+        apt:
+          sources:
+            - sourceline: ppa:fooinha/libyara3
+          packages:
+            - *packages
+            - libyara-dev
       env:
         - NAME="linux,clang"
         - *default-cflags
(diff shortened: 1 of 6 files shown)


inliniac:
I'd like to revisit this as I think it could be useful. The 'yara: file: <filename>;' syntax isn't working for me, I don't like the keyword: subkey: value; approach. Perhaps simply split into: yara_file:<filename>; yara_rule; yara_tag; ?

inliniac:
Do I understand it correctly that this PR allows for 'global' yararules as well as per rule yararules?

fooinha:
Yes it allows global file for YARA rules, so you don't need to specify it on per rule basis.

inliniac:
But it can also be per suricata rule, right? Yara can be expensive, so it could make sense to do something like:
alert http any any -> any any (content:"somehost"; http_host; yara_file:somerule.yara; sid:1;)

Then the yara rule would only be called if the content condition is true. Would this work?

fooinha:
Yes, it can be per suricata rule.
I also allowed global rules, so we don't compile many times the same YARA files.
I didn't know how to minimize the startup burden if we only accepted per rule YARA files.

Let's work in the content condition ... I dunno the answer to that, but, yes, it would be better to bet on this behaviour.

PR #2394, opened 2016-11-01
FTP: adds patterns from server to client

fooinha:
Without registering server side patterns the flow stats for
FTP can't be incremented, and the app-layer FTP stats were always 0.

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -329,6 +329,16 @@ static int FTPRegisterPatternsForProtocolDetection(void)
     {
         return -1;
     }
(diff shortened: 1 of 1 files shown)


inliniac:
The issue here is similar to what you mention in https://github.com/inliniac/suricata/pull/2396#issuecomment-258021591 "220" is also used by SMTP and perhaps other protocols as well. We currently don't have a good way to deal with this.

inliniac:
cc @jasonish

fooinha:
Maybe we need to add support for registering more complex pattern matching mechanisms for the proto app layer modules.

inliniac:
This will have to wait for @jasonish work on improved protocol detection.

PR #2395, opened 2016-11-02
luaL_newstate maybe null in mac osx

missdiog:
show info when call luaL_newstate fail. 
the error  often happen on mac osx when luajit link without "-pagezero_size 10000 -image_base 100000000"


inliniac:
Needs rebase. Please submit a new PR. Thanks!

PR #2396, opened 2016-11-02
app-layer: account flow

fooinha:
only account flow if we have both sides and have same proto.

Just to be sure.

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -373,8 +373,8 @@ static int TCPProtoDetect(ThreadVars *tv,
                 AppLayerParserGetStreamDepth(f->proto, f->alproto));
         FlagPacketFlow(p, f, flags);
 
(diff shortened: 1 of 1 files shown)


inliniac:
Not sure I agree. First, we need to log for each flow. I think this patch won't log at all if protos are not equal.

Second: the patch should not introduce all that debug code in the same commit.

fooinha:
I see what you mean. A side of flow for the app-layer proto may not register a pattern, and it will not be possible to match proto and the flow will not be incremented? 

Also please check this #2394. The stats for ftp were not being updated. I got confused by this.

I saw this  mismatch protocol while registering conflicting patterns both for ftp and irc protos for the "PASS" command sent to server.

inliniac:
@jasonish is looking at improving the protocol detection to deal with such cases.

PR #2338, opened 2016-10-06
eve ouput kafka v2

fooinha:
Initial work for logging output into the distributed streaming platform Apache Kafka.

Allows to send eve messages to a list of brokers of Apache Kafka.
It uses the client implementation from [librdkafka](https://github.com/edenhill/librdkafka).

It's set to be conditional at configure/compile time with the  --enable-rdkafka [Enable Kafka support] option.

Configuration settings and default values: 

``` yaml
      filetype: regular #regular|syslog|unix_dgram|unix_stream|redis|kafka
       filename: eve.json
       kafka:
        broker-list: 127.0.0.1:9092
        topic: suricata
        compression: snappy # none, gzip, snappy
        partition: -1 # below < 0 means Unassigned Partition
        max-retries: 1
        backoff-ms: 10
        buffer-max-messages: 100000
        log-level: 6
```

It's an alternative for redis output. 

Rebased to upstream/master HEAD.

PPA for precise build included at travis.yaml

PR #2397, opened 2016-11-05
dev ouput kafka

fooinha:
Continuation of #2338.

This PR adds the ability to output every suricata's logs to kafka .


missdiog:
nice work!

PR #2398, opened 2016-11-07
Issue 1924 - dns malformed response - v1

jasonish:
Fix issue https://redmine.openinfosecfoundation.org/issues/1924

Allow DNS messages with a data length of 0 without marking them as malformed.

Also fixes the direction on the malformed rules.

Prscript:
- https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/392
- https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/40


inliniac:
Merged through #2414, thanks Jason.

PR #2400, opened 2016-11-08
pcap-log: seed ring buffer on start up - v2

jasonish:
Changes from last PR:
- Check for dirent.h and fnmatch.h in configure, and only build ring buffer initialization if they both exist.

Request for comments on if there is a better/nicer way to do this?

On start, look for existing pcap log files and add them to
the ring buffer. This makes pcap-log self maintaining over
restarts removing the need for external tools to clear
orphaned files.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/42
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/394


inliniac:
Merged through #2414, thanks Jason.

PR #2402, opened 2016-11-09

counterthreatunit:

### DIFF
--- a/doc/userguide/rules/suri-rule-guide-for-the-snort-expert.rst
+++ b/doc/userguide/rules/suri-rule-guide-for-the-snort-expert.rst
@@ -1,0 +1,906 @@
+====================================================
+The Suricata Rule Writing Guide For The Snort Expert
+====================================================
+
+Overview
+--------
+
+Suricata is a modern and powerful IDS/IPS engine that supports the
+popular Snort rule syntax.  However, there are some important
+differences and capabilities between how the engines work and how rules
+are applied that need to be taken into consideration when attempting to
+craft the best Suricata rule(s) possible.  This guide is intended to
+highlight the major differences between Snort and Suricata that the rule
+writer needs to be aware of when converting Snort rules to Suricata
+and/or writing Suricata rules.  The target audience for this document is
+people who are very familiar with the Snort engine and rules, who want
+to write superlative rules for Suricata.
+
+Where not specified, the statements below apply to Suricata.
+
+In general, references to Snort refer to the version 2.9 branch and
+references to Suricata apply to version 3.1 and before (where noted).
+
+Contents
+--------
+
+.. contents::
+
+Other References
+----------------
+
+Suricata Documentation:
+
+-  :doc:`index`
+-  :doc:`../index`
+-  :doc:`snort-compatibility`
+
+Automatic Protocol Detection
+----------------------------
+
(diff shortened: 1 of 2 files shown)


doc/userguide/rules/suri-rule-guide-for-the-snort-expert.rst
```diff
@@ -0,0 +1,906 @@
...
+
+  -  Split up longer content matches if you are running a version of Suricata before 3.0.
+
+-  :doc:`snort-compatibility`
+-  `https://redmine.openinfosecfoundation.org/issues/1281 <https://redmine.openinfosecfoundation.org/issues/1281>`_
+
+``flow`` Keyword
+-------------------
+
+-  Suricata does not support ``not_established`` for the ``flow`` keyword.
```
inliniac: This has been fixed in 3.2. As these docs target 3.2 I think this should be updated.

doc/userguide/rules/suri-rule-guide-for-the-snort-expert.rst
```diff
@@ -0,0 +1,906 @@
...
+Relative PCRE
+-------------
+
+-  You can do relative PCRE matches in normalized/special buffers with Suricata.  Example::
+
+     content:".php?sign="; http_uri; pcre:"/^[a-zA-Z0-9]{8}$/UR";
+     
+-  With Snort you can't combine the “relative” PCRE option ('R') with other buffer options like normalized URI ('U') – you get a syntax error.
+
+``tls.*`` Keywords
```
inliniac: I think it would be good to recommend the tls_cert_* keywords here. They are new in 3.2.

inliniac:
Thanks David, this is very useful.

doc/userguide/rules/suri-rule-guide-for-the-snort-expert.rst
```diff
@@ -0,0 +1,906 @@
...
+   use as the fast pattern match.
+
+-  When in doubt about what is going to be use as the fast pattern match
+   by Suricata, set ``fast_pattern`` explicitly in the rule and/or
+   run Suricata with the ``--engine-analysis`` switch and view the
+   generated file (``rules_fast_pattern.txt``).
+
+-  Like Snort, the fast pattern match is checked before ``flowbits``
+   in Suricata.
+
```
inliniac: depth _and_ offset I must add.

inliniac:
Could you do a new PR with the requested updates?

See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Incorporating-feedback

Thanks!

doc/userguide/rules/suri-rule-guide-for-the-snort-expert.rst
counterthreatunit: Actually the docs were written for 3.1 (since this is what was/is stable) but I can update for 3.2.

doc/userguide/rules/suri-rule-guide-for-the-snort-expert.rst
counterthreatunit: I think it makes sense to update the tls-keywords document (looks like it has already been updated some) and reference that here so we don't have to maintain the same information in two places.

PR #2405, opened 2016-11-09
af-packet: add VLAN header when needed in IPS mode

regit:
When packet is coming from a real ethernet card, the kernel is
stripping the vlan header and delivering a modified packet so
we need to insert the VLAN header back before sending the packet
on the wire.

To do so, we pass an option to the raw socket to add a reserve
before the packet data. It will get Suricata some head room to
to move the ethernet addresses before there actual place and
and insert the VLAN header in the correct place.

PR scripts:
* PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/226
* PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/8

Redmine issue https://redmine.openinfosecfoundation.org/issues/1780

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -666,6 +666,8 @@ TmEcode AFPWritePacket(Packet *p)
 {
     struct sockaddr_ll socket_address;
     int socket;
+    uint8_t *pstart;
+    size_t plen;
 
     if (p->afp_v.copy_mode == AFP_COPY_MODE_IPS) {
         if (PACKET_TEST_ACTION(p, ACTION_DROP)) {
@@ -691,7 +693,21 @@ TmEcode AFPWritePacket(Packet *p)
     if (p->afp_v.peer->flags & AFP_SOCK_PROTECT)
         SCMutexLock(&p->afp_v.peer->sock_protect);
     socket = SC_ATOMIC_GET(p->afp_v.peer->socket);
-    if (sendto(socket, GET_PKT_DATA(p), GET_PKT_LEN(p), 0,
+    /* We have a vlan ID without header so first VLAN have been stripped */
+    if (p->vlan_idx && !p->vlanh[0] && p->afp_v.relptr) {
+        pstart = GET_PKT_DATA(p) - VLAN_HEADER_LEN;
+        plen = GET_PKT_LEN(p) + VLAN_HEADER_LEN;
+        /* move ethernet addresses */
+        memmove(pstart, GET_PKT_DATA(p), 2 * ETH_ALEN);
+        /* write vlan info */
+        *(uint16_t *)(pstart + 2 * ETH_ALEN) = htons(0x8100);
+        *(uint16_t *)(pstart + 2 * ETH_ALEN + 2) = htons(p->vlan_id[0]);
+    } else {
+        pstart = GET_PKT_DATA(p);
+        plen = GET_PKT_LEN(p);
+    }
+
+    if (sendto(socket, pstart, plen, 0,
                (struct sockaddr*) &socket_address,
                sizeof(struct sockaddr_ll)) < 0) {
         SCLogWarning(SC_ERR_SOCKET, "Sending packet failed on socket %d: %s",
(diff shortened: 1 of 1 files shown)


src/source-af-packet.c
```diff
@@ -691,7 +693,21 @@ TmEcode AFPWritePacket(Packet *p)
     if (p->afp_v.peer->flags & AFP_SOCK_PROTECT)
         SCMutexLock(&p->afp_v.peer->sock_protect);
     socket = SC_ATOMIC_GET(p->afp_v.peer->socket);
-    if (sendto(socket, GET_PKT_DATA(p), GET_PKT_LEN(p), 0,
+    /* We have a vlan ID without header so first VLAN have been stripped */
+    if (p->vlan_idx && !p->vlanh[0] && p->afp_v.relptr) {
+        pstart = GET_PKT_DATA(p) - VLAN_HEADER_LEN;
+        plen = GET_PKT_LEN(p) + VLAN_HEADER_LEN;
+        /* move ethernet addresses */
+        memmove(pstart, GET_PKT_DATA(p), 2 * ETH_ALEN);
```
inliniac: Does this code get called when we are not in zero copy mode? In that case the reserve call won't do much for us, will it? I worry that GET_PKT_DATA will then point to a suricata-allocated memory block and memmove will write before it.

regit: The test is ` if (p->vlan_idx && !p->vlanh[0] && p->afp_v.relptr)` and the last test in there check that we have the release function set. And we have this release pointer set only if we are in zero copy mode.

inliniac:
If no one tests this in time for 3.2.1 we'll merge it in the dev branch and backport later if no bug reports.

gozzy:
I hope that we'll try it out today.

gozzy:
The fix itself is fine, VLAN tags are correctly passed when working in inline mode.

However, we've noticed another problem: VLAN-tagged packets seem to loose 'priority' and 'CFI' attributes (https://wiki.wireshark.org/VLAN), probably because of this:

p->vlan_id[0] = h.h2->tp_vlan_tci & 0x0fff;

The mask 0x0fff cuts those attributes and reconstructed packet slightly differs from the original one. I don't know whether it's critical, but probably this should be kept in mind for a later fix.

inliniac:
On 25-01-17 14:27, Alexander Gozman wrote:
> The fix itself is fine, VLAN tags are correctly passed when working in
> inline mode.
> 
> However, we've noticed another problem: VLAN-tagged packets seem to
> loose 'priority' and 'CFI' attributes (https://wiki.wireshark.org/VLAN),
> probably because of this:
> 
> p->vlan_id[0] = h.h2->tp_vlan_tci & 0x0fff;
> 
> The mask 0x0fff cuts those attributes and reconstructed packet slightly
> differs from the original one. I don't know whether it's critical, but
> probably this should be kept in mind for a later fix.
> 

Great catch, thanks Alex. I'd like to see this fixed before merge.

gozzy:
I can try to make another PR, but I have some doubts about a proper way to handle this case. The most obvious idea is to add two bitfields into the Packet structure and store the attributes there. Like:

struct Packet {
    ...
    uint16_t vlan_id[2];
    uint8_t vlan_idx;
    ....
}

Then, in source-af-packet.c (and, maybe, source-pfring.c):

...
p->priority = (h.h2->tp_vlan_tci & 0xe000) >> 13;
p->cfi = (h.h2->tp_vlan_tci & 0x1000) >> 12;
...

Finally, when reconstructing a packet:

uint16_t tag = htons(p->vlan_id[0] | (p->priority << 13) | (p->cfi << 12));
*(uint16_t *)(pstart + 2 * ETH_ALEN + 2) = tag;

gozzy:
Or, maybe, we should just remove the mask 0x0fff and store a whole tag. Anyway, if one needs to get VLAN ID, priority or CFI, suricata provides macros like GET_VLAN_ID, GET_VLAN_PRIORITY or GET_VLAN_CFI.

inliniac:
I don't think the rest of the code is expecting the prio & cfi to be part of the vlan_id field. I think I prefer to add 2 new fields like you suggested in https://github.com/inliniac/suricata/pull/2405#issuecomment-275146178

Maybe we can store the vlan_id as vlan_id:12, although I'm not sure how that works in an array.

regit:
Good catch indeed @gozzy. Do you want me to take code update ?

gozzy:
@regit, it would be great if you could make it 'cause at the moment I don't realise all the details about VLAN handling. It's done in source-af-packet/pfring, in DecodeVLAN function...

inliniac:
Needs improvements

inliniac:
I think the priority and cfi could be saved in Packet::AFPPacketVars. This issue is afpacket specific so we can use a afpacket specific store. Think that would make a fix much easier. Care to try @gozzy ?

gozzy:
Well, I can try if @regit is not already updating the patch.

inliniac:
@regit seems rather busy and since you have a test setup I think it may be easiest if you try, but I'll let @regit speak for himself :)

Thanks for considering anyway!

regit:
OK, I'm getting a look at it right now. But @gozzy, you will be the one to test it if ok with you.

PR #2407, opened 2016-11-14
RFC: Prototype for extracting files and metadata to Redis.

zrasmor:
Please review proof of concept and provide feedback. This is something that is a priority for us, so I'm happy to continue developing this feature to completion.

NOTES:
- This has only really been tested with forced extraction on all files - I ran into some issues getting the HTTP request/response headers when using the filestore keyword with signatures - I may have missed something silly though
 - There's certainly opportunities for improvement with constructing the metadata. In the prototype, I was partially experimenting with how I could extract HTTP header fields from the libHTP APIs. Ultimately, we'd at least like to see raw request/response headers, TCP quad, direction (request or response). Other folks might appreciate adding the flow ID also.

TODO:
- Cleaner separation from FileStore module (Register separate FileDataModule??)
- Decide how to handle truncated files
- Ensure that design is compatible with TLS cert extraction and future SMTP extraction
- Create configurable parameters in suricata.yaml
    * Redis instance (IP and port)
    * Redis Queue Name
    * Redis key expiry - currently hardcoded to 5 min
    * Redis key prefix (i.e. suricata_<hostname>_) - good practice so that keys are distinguishable from other keys in Redis
    * Hashing algorithm - Redis keys currently contain md5 as part of key, doesn't need to be limited to md5
    * Caching enable/disable/expiry - enable caching to not store files we have already seen recently
 - Unit testing
 - I've probably missed a few other TODOs

### DIFF
--- a/src/log-filestore.c
+++ b/src/log-filestore.c
@@ -1,4 +1,5 @@
 /* Copyright (C) 2007-2013 Open Information Security Foundation
+ * Copyright (C) 2016 Lockheed Martin Corporation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -59,6 +60,11 @@
 #include "util-memcmp.h"
 #include "stream-tcp-reassemble.h"
 
+#ifdef HAVE_LIBHIREDIS
+#include "hiredis/hiredis.h"
+#endif
+#include <jansson.h>
+
 #define MODULE_NAME "LogFilestoreLog"
 
 static char g_logfile_base_dir[PATH_MAX] = "/tmp";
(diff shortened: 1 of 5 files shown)


zrasmor:
Note: Travis check failed because prototype requires libnss, which does not appear to be included in Travis build.

fooinha:
You can edit the .travis.yml file in your branch and add the required package (libnss3-dev ? ) for installation with apt-get

inliniac:
I would like to suggest a alternate approach. We have a streaming data logging API in https://github.com/inliniac/suricata/blob/master/src/output-streaming.c I think the file data would probably fit in there. I've been thinking about moving the existing filedata logging into that API as well.

PR #2411, opened 2016-11-19
Add Ethernet header logging

regit:
This PR is adding optional Ethernet header logging to EVE events. It updates CreateJSONHeader function so most protocols are covered.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/227
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/9

PR #2415, opened 2016-11-23
Ethernet header v2

regit:
A rework of #2411 addressing remarks done on the PR.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/228
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/10

### DIFF
--- a/src/util-logopenfile.h
+++ b/src/util-logopenfile.h
@@ -114,6 +114,9 @@ typedef struct LogFileCtx_ {
     /* flag to avoid multiple threads printing the same stats */
     uint8_t flags;
 
+    /* flag to store the options */
+    uint8_t options_flags;
+
     /* Flag if file is a regular file or not.  Only regular files
      * allow for rotataion. */
     uint8_t is_regular;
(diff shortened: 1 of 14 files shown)


src/util-logopenfile.h
```diff
@@ -114,6 +114,9 @@ typedef struct LogFileCtx_ {
     /* flag to avoid multiple threads printing the same stats */
     uint8_t flags;
 
+    /* flag to store the options */
+    uint8_t options_flags;
```
inliniac: I don't think this is the proper place to add these options. They are about the JSON output, so they should move into something JSON specific.

regit: I've just checked the code by removing this and trying to find a structure different from LogFileCtx that could host that. Result is that the LogFileCtx is the first common structure that can be used in all the CreateJSONHeader call contexts.
So I see the following possibilities:
* We add a union to contain per output type options and variable in LogFileCtx
* We rework all JSON output module to use a intermediate structure that will have a pointer in structure like LogDnsFileCtx

On last point, if we look at DNS we have:
```
typedef struct LogDnsFileCtx_ {
    LogFileCtx *file_ctx;
    uint64_t flags; /** Store mode */
} LogDnsFileCtx;

typedef struct LogDnsLogThread_ {
    LogDnsFileCtx *dnslog_ctx;
    /** LogFileCtx has the pointer to the file and a mutex to allow multithreading */
    uint32_t dns_cnt;

    MemBuffer *buffer;
} LogDnsLogThread;
```
So we could have a 'wrapper' JSONLogFileCtx in LogDnsFileCtx that contain the LogFileCtx and the JSON options.

PR #2416, opened 2016-11-24

counterthreatunit:


counterthreatunit:
Closing per VictorJ and resubmitting.  This PR threw an AppVeryor error since AppVeyor wasn't configured properly at the time the PR was submited.

PR #2423, opened 2016-11-24
TLS logging for session resumption

rtkrruvinskiy:

was never seen.


inliniac:
@thus can you have a look?

thus:
I'll take a look :)

rtkrruvinskiy:
I've created https://redmine.openinfosecfoundation.org/issues/1969 for this PR.

src/log-tlslog.c
```diff
@@ -307,10 +308,17 @@ static int LogTlsLogger(ThreadVars *tv, void *thread_data, const Packet *p,
...
     MemBufferWriteString(aft->buffer,
-                         "%s %s:%d -> %s:%d  TLS: Subject='%s' Issuerdn='%s'",
-                         timebuf, srcip, sp, dstip, dp,
-                         ssl_state->server_connp.cert0_subject,
-                         ssl_state->server_connp.cert0_issuerdn);
+                         "%s %s:%d -> %s:%d  TLS: ",
+                         timebuf, srcip, sp, dstip, dp);
+
+    if ((ssl_state->flags & SSL_AL_FLAG_SESSION_RESUMED) != 0) {
+        MemBufferWriteString(aft->buffer, "Session resumed");
```
thus: "Session resumed" should be replaced by something easier to parse, like SESSION='resumed'.

src/log-tlslog.c
```diff
@@ -307,10 +308,17 @@ static int LogTlsLogger(ThreadVars *tv, void *thread_data, const Packet *p,
...
     MemBufferReset(aft->buffer);
     MemBufferWriteString(aft->buffer,
-                         "%s %s:%d -> %s:%d  TLS: Subject='%s' Issuerdn='%s'",
-                         timebuf, srcip, sp, dstip, dp,
-                         ssl_state->server_connp.cert0_subject,
-                         ssl_state->server_connp.cert0_issuerdn);
+                         "%s %s:%d -> %s:%d  TLS: ",
+                         timebuf, srcip, sp, dstip, dp);
+
+    if ((ssl_state->flags & SSL_AL_FLAG_SESSION_RESUMED) != 0) {
```
thus: Couldn't this be written simpler? Like:
```if ((ssl_state->flags & SSL_AL_FLAG_SESSION_RESUMED) {```

File: src/log-tlslog.c [hunk 3]
```diff
@@ -307,10 +308,17 @@ static int LogTlsLogger(ThreadVars *tv, void *thread_data, const Packet *p,
...
+        MemBufferWriteString(aft->buffer, "Session resumed");
+    } else {
+        MemBufferWriteString(aft->buffer,
+                             "Subject='%s' Issuerdn='%s'",
+                             ssl_state->server_connp.cert0_subject,
+                             ssl_state->server_connp.cert0_issuerdn);
```
thus: Instead of this if-else, I think you should do individual if checks to see if subject and issuer is set and print if they are, like is done with the other fields in LogTlsLogExtended().

src/output-json-tls.c
```diff
@@ -74,13 +74,20 @@ typedef struct JsonTlsLogThread_ {
...
     /* tls.issuerdn */
-    json_object_set_new(js, "issuerdn",
-                        json_string(ssl_state->server_connp.cert0_issuerdn));
+    if (ssl_state->server_connp.cert0_subject != NULL) {
+        json_object_set_new(js, "issuerdn",
+                            json_string(ssl_state->server_connp.cert0_issuerdn));
+    }
 
+    if ((ssl_state->flags & SSL_AL_FLAG_SESSION_RESUMED) != 0) {
+        json_object_set_new(js, "session_resumed", json_true());
```
thus: I think something like "session":"resumed" would look better, even if it's a string an not a boolean.

thus:
I would guess that this is a feature that not everyone wants enabled. I therefore think it should be made optional through configuration. E.g. add something like "log-session-resumption: yes|no" to the logger in suricata.yaml. My opinion is that the default should be that this is disabled.

thus:
Also needs to rebase with master.

rtkrruvinskiy:
@thus To confirm, are you requesting that I make the change to compare the session id in the HELLO messages and check the extension, or were you just musing and are ultimately okay with the current approach?

src/log-tlslog.c
rtkrruvinskiy: It could. I like being explicit in bitmask checks as to what it is I'm checking for, but I'm fine with doing it the other way.

src/log-tlslog.c
rtkrruvinskiy: Should SESSION be in all caps? I don't see all-caps tokens elsewhere in the log line.

src/log-tlslog.c
rtkrruvinskiy: Will change.

src/output-json-tls.c
rtkrruvinskiy: Will change.

rtkrruvinskiy:
@thus Why would this be disabled by default? If this is disabled, TLS transactions with session resumption are not logged, at all. Why would that be desirable behaviour?

src/log-tlslog.c
thus: Perhaps 'Session' instead. I see that it's a mix of both all-caps tokens (SNI, VERSION, NOTBEFORE, NOTAFTER) and not (Subject, Issuerdn).

thus:
I'm okay with the current approach, so no need to change it.

thus:
After thinking it through, I agree that it should be enabled by default, but I still think it should be possible for users to disable it through configuration.

rtkrruvinskiy:
@thus Sounds good. I'll add a configuration section for this. Out of curiosity, though, could you tell me when someone wouldn't want to output these transactions?

rtkrruvinskiy:
@thus I've pushed a commit that has the changes you asked for. Thanks!

thus:
The use case I can think of (there are probably others) is when using Suricata to harvest TLS certificate meta-data. Then TLS sessions that are resumed are just noise.

thus:
Could you open a new PR with the changes and rebase? That would make it easier to do a new review. Thanks :)

inliniac:
Please see https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Create-a-new-branch-for-incorporating-feedback

Thanks!

PR #2427, opened 2016-11-26
tls: ALPN extension

fooinha:
- tls_alpn signature keyword
- log alpn values to tls.log and eve if extendend output

Rules examples:

```bash
alert tls any any -> any any (msg:"SURICATA TLS CLIENT h2"; tls_alpn:client:h2; sid:77770001; rev:1;)
alert tls any any -> any any (msg:"SURICATA TLS SERVER h2"; tls_alpn:server:h2; sid:77770001; rev:1;)

```

tls.log output:

```bash
11/26/2016-02:36:37.057656 10.10.10.2:49310 -> 23.207.78.44:443  TLS: Subject='C=US, ST=California, L=San Jose, O=eBay, Inc., OU=Site Operations san1-v5, CN=www.ebay.com' Issuerdn='C=US, O=Symantec Corporation, OU=Symantec Trust Network, CN=Symantec Class 3 Secure Server CA - G4' SHA1='90:02:74:76:73:f1:5d:e6:3e:f1:08:e6:cd:22:db:40:6f:43:73:70' SNI='www.ebay.com' VERSION='TLS 1.2' NOTBEFORE='2015-10-27T00:00:00' NOTAFTER='2017-10-27T23:59:59' C-ALPN=<h2,http/1.1> S-ALPN=<h2> 
```

eve.log -> tls
```json
{
  "timestamp": "2016-11-26T02:36:37.057656+0000",
  "flow_id": 1567877630298705,
  "in_iface": "eth0",
  "event_type": "tls",
  "src_ip": "10.10.10.2",
  "src_port": 49310,
  "dest_ip": "23.207.78.44",
  "dest_port": 443,
  "proto": "TCP",
  "tls": {
    "subject": "C=US, ST=California, L=San Jose, O=eBay, Inc., OU=Site Operations san1-v5, CN=www.ebay.com",
    "issuerdn": "C=US, O=Symantec Corporation, OU=Symantec Trust Network, CN=Symantec Class 3 Secure Server CA - G4",
    "fingerprint": "90:02:74:76:73:f1:5d:e6:3e:f1:08:e6:cd:22:db:40:6f:43:73:70",
    "sni": "www.ebay.com",
    "version": "TLS 1.2",
    "notbefore": "2015-10-27T00:00:00",
    "notafter": "2017-10-27T23:59:59",
    "alpn": {
      "client": [
        "h2",
        "http/1.1"
      ],
      "server": [
        "h2"
      ]
    }
  }
}
```

eve log -> alert
```json
{
[...]

PR #2442, opened 2016-11-29
tls: tls_cipher_suite keyword

fooinha:
- Allows tls app-layer rules to match TLS handshake cypher suites

With conjunction with  #2427, can be used, for instance, for detecting http/2 services allowing blacklisted tls cipher suites ->  https://http2.github.io/http2-spec/#BadCipherSuites.

Example rules:

alert tls any any -> any any (msg:"SURICATA TLS CIPHER SUITE CLIENT"; tls_cipher_suite:client:TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,0xcca9; sid:66666666; rev:1;)

alert tls any any -> any any (msg:"SURICATA TLS CIPHER SUITE SERVER"; tls_cipher_suite:server:0xcca8; sid:66666667; rev:1;)


inliniac:
@thus can you have a look?

inliniac:
The keyword looks old style (just looking at the PR title), @thus has been adding new keywords with a different notation.

fooinha:
Is the current style using  '_'  as word separator?

thus:
@inliniac: I'll try to find some time next week to review all the TLS PR's.

thus:
@fooinha: Yup, that's right! The current style use '_' instead of '.' as a separator.

src/detect-tls-cipher-suite.c
```diff
@@ -0,0 +1,1116 @@
...
+
+/**
+ * \file
+ *
+ * \author Paulo Pacheco <fooinha@gmail.com>
+ *
+ * Implements support for tls_cipher_suite keyword.
+ */
+
+#include "ctype.h"
```
thus: This is already included in "suricata-common.h", which you include below.

src/detect-tls-cipher-suite.c
```diff
@@ -0,0 +1,1116 @@
...
+
+/**
+ * \brief gets a cipher suite description
+ *
+ * \param value - 2 bytes value for cipher suite code
+ *
+ * \retval "-" if not found
+ * \retval const char * with description if found
+ */
+const char * TlsCipherSuiteDescription(uint16_t value)
```
thus: I think everything related to cipher suite lookup should be moved to app-layer-ssl.[c|h], so we don't have to include detect-tls-cipher-suite.h everywhere we want to use the cipher suites.

thus:
I haven't had time to take a good look at the detection code yet.

The code depends on custom tls logging code to be accepted (#2491 and #2449).

thus:
Rebase is needed, btw.

fooinha:
@thus, Thanx for review.
I changed the code in order to meet all your comments.
Also rebase was done.

inliniac:
Please do new PR for incorporating feedback and/or rebases. Thanks!

PR #2448, opened 2016-12-03
log: common custom format output

fooinha:
Custom format output common base for output loggers.
Work in progress for review and comments.

Useful for #2427 ( @thus ) and future output loggers.


fooinha:
Replaced by #2581

PR #2451, opened 2016-12-04
flow: postpone FlowForceReassembly at shutdown

fooinha:
- fixes Bug 1946.

More information at redmine. Close it if non-sense.

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2686,12 +2686,6 @@ int main(int argc, char **argv)
     TmThreadDisableReceiveThreads();
 
     if (suri.run_mode != RUNMODE_UNIX_SOCKET) {
-        /* we need a packet pool for FlowForceReassembly */
-        PacketPoolInit();
-
-        FlowForceReassembly();
-        /* kill receive threads when they have processed all
-         * flow timeout packets */
         TmThreadDisablePacketThreads();
     }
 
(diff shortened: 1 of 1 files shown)


src/suricata.c
```diff
@@ -2698,6 +2692,12 @@ int main(int argc, char **argv)
         FlowDisableFlowRecyclerThread();
     }
 
+    if (suri.run_mode != RUNMODE_UNIX_SOCKET) {
+        /* we need a packet pool for FlowForceReassembly */
+        PacketPoolInit();
+        FlowForceReassembly();
```
inliniac: Whats the logic of moving this to after the packet processing threads are killed? How are timeout pseudo packets processed?

fooinha: I don't know enough to answer that.

What I saw, was that the FlowForceReassembly() was called while processing flows, and after breaking main's loop , and as a side effect, the http output was borked.

For instance, by adding a single time(1) right after suri's main loop, the http output was complete and ok for the supplied pcap files in redmine ticket 1946.

inliniac:
@fooinha  I'm looking at this, but can't seem to reproduce it. Could you help me out by sharing the test method and script? Thanks!

fooinha:
Fixed by #2518

PR #2452, opened 2016-12-08

counterthreatunit:


inliniac:
@jasonish can you have a look?

inliniac:
I'n not to excited about the title. It was a great title for your talk at SuriCon, but in our docs it fits better under our 'differences with Snort' doc.

I also think we can remove some things:
1. introduce to suricata
2. things that do not apply to the current suricata version. Sphinx docs are versioned, so the doc applies to the current code. No need to confused the reader with obsolete information.

counterthreatunit:
closing and going to re-work and re-submit

PR #2454, opened 2016-12-13
http: allow lower/mixed case in proto detect

inliniac:
In HTTP detection registered patterns were upper case only. Since the
detection is based on both sides this would still work for sessions
where one of the talkers misbehaved. If both sides misbehave this
would fail however, so this patch introduces case insensitive matching.

PRscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/38
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/546

https://redmine.openinfosecfoundation.org/issues/1984

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -2737,7 +2737,7 @@ static int HTPRegisterPatternsForProtocolDetection(void)
              * 3 is subtracted from the length since the spacing is hex typed as |xx|
              * but the pattern matching should only be one char
             */
-            register_result = AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP,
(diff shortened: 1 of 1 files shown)


poona:
Hi Victor,
You have seen cases of lowercase patterns?

pevma:
@poona  ET folks see malware like that on a good few occasions.

poona:
okay.  Thanks

inliniac:
See https://lists.openinfosecfoundation.org/pipermail/oisf-users/2016-December/006600.html for example

poona:
Nice.  Thanks for the example.

PR #2455, opened 2016-12-14
Preview/detect list api/v21

inliniac:
Large patch set to move away from the hardcoded DETECT_SM_LIST_* arrays in the Signature structure. Makes registering new detection lists much easier.

Signature struct goes from ~1100bytes to ~288bytes.

Instead of the large per Signature array of lists, the lists are now part of the inspect engines. Only those inspect engines that apply to a sig are in the sig and so is the list.


inliniac:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/39
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/547

pevma:
Would a signature struct of about 288 bytes mean that 50k rules loaded by Suri would take about 13.8MB of memory or it is not directly related?

inliniac:
On 14-12-16 22:20, Peter Manev wrote:
> Would a signature struct of about 288 bytes mean that 50k rules loaded
> by Suri would take about 13.8MB of memory or it is not directly related?

No, not directly related. It should use a little bit less memory than
before, but probably negligible vs other detection memory use (e.g.
mpm). I'm hoping for some better cache behavior, but overall this is
about cleaning up the code and improve extensibility.

inliniac:
Replaced by #2473

PR #2457, opened 2016-12-16
app-layer-tls: don't trigger decoder event when no CLIENT_HELLO extensions are present

thus:
Extensions are optional, so don't trigger "SURICATA TLS handshake invalid length" decoder event when no extensions are present in CLIENT HELLO record.

Partially fixes the following issue:
https://redmine.openinfosecfoundation.org/issues/1982

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/59
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/59


inliniac:

PR #2458, opened 2016-12-16
app-layer-tls: increase max number of tls records per packet

thus:
Tls packets may contain several records. This increase the number of allowed records per packet from 30 to 255, and adds a new and more informative decoder event when this limit is reached.

Fixes this issue (together with #2457):
https://redmine.openinfosecfoundation.org/issues/1982

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/60
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/60


inliniac:

PR #2460, opened 2016-12-16
app-layer-tls: fix tls_cert_subject prefilter bug

thus:
If check in prefilter was checking that issuer was non-NULL, when it in fact should be checking subject. This caused Suricata to segfault when encountering certificates with an empty subject field.

https://redmine.openinfosecfoundation.org/issues/1988

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/61
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/61


inliniac:

PR #2461, opened 2016-12-16
smtp: commands and replies are not case sensitive

regit:
RFC states that "Commands and replies are not case sensitive" and
patterns were registered to be case sensitive. So this patch fixes
a trivial evasion of SMTP signatures.

Redmine issue: https://redmine.openinfosecfoundation.org/issues/1989

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/230
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/12


inliniac:
Merged through #2467, thanks Eric!

PR #2463, opened 2016-12-17
engine-analysis: output to JSON format

fooinha:
Output rule engine analysis report to JSON format.

```
engine-analysis:
  # enables printing reports for fast-pattern for every rule.
  rules-fast-pattern: yes
  # enables printing reports for each rule
  rules: yes
  # output format
  output-format: txt # json
```

### DIFF
--- a/src/detect-engine-analyzer.c
+++ b/src/detect-engine-analyzer.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2012 Open Information Security Foundation
+/* Copyright (C) 2007-2016 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -19,6 +19,7 @@
  * \file
  *
  * \author Eileen Donlon <emdonlo@gmail.com>
+ * \author Paulo Pacheco <fooinha@gmail.com> - JSON output
  *
  * Rule analyzer for the detection engine
  */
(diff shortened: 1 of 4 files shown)


inliniac:
Ah this has been on my wishlist for a long time, thanks for taking it on!

I wonder if you can compact the code a bit. Now we see a lot of:
<pre>
        if(IS_OUTPUT_FORMAT(JSON)) {
            IF_HAS_JSON(matcher = json_string("content"));
        }
</pre>

What about turning that into something like
<pre>
    JSON(matcher = json_string("content"));
</pre>

JSON would then be a macro that hides the checks.

inliniac:
In general I'm not happy with the original analyzer code. It has way too much hardcoded logic. So a complete rewrite has been on my mind as well. So if you have thoughts about that, please share them. Thanks!

fooinha:
Updated PR with your macros suggestions. Much cleaner now. ;p

inliniac:
I think it's still too messy, which is largely caused by the original code. Not sure how to address it w/o a complete overhaul, but maybe that is just what it will take.

inliniac:
Are you interested in doing a completely new implementation instead? I'm thinking about a model where each of the rule keywords can register a 'Analyze' callback that does specific checks. Per rule we'd have a list of messages (info/warn/err) that we would then print to stdout, json, etc. Make sense?

inliniac:
Additionally, I would like to make sure the analyzer can run in the normal runmodes as well, and perhaps even over unix socket?

PR #2466, opened 2016-12-20
util-magic: fix build when magic is not available

regit:
If HAVE_MAGIC is not defined then we don't have the test functions
so we can't register them.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/231
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/13

### DIFF
--- a/src/util-magic.c
+++ b/src/util-magic.c
@@ -659,6 +659,7 @@ end:
 
 void MagicRegisterTests(void)
 {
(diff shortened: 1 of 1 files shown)


inliniac:
Merged through #2467, thanks Eric.

PR #2469, opened 2016-12-20
rfc - support for ts and tc probing parsers - v2

jasonish:
This PR is to address the issue where the same probing parser is called in the TS and TC direction, yet most probes (DNS for example) are designed to only check the TS direction.  Currently this means that some DNS responses can result in a failed detection.

Now when registering a parser, the TS and TC parsers must be supplied separately, then the protocol detect will only call the one in the applicable direction.

DNS is updated as an example which now has a minimal response parser.  Other protocols will need to be checked as its likely their request parsers can also be used as response parsers if they don't check too deeply.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/64
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/416

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -75,8 +75,12 @@ typedef struct AppLayerProtoDetectProbingParserElement_ {
     uint32_t min_depth;
     /* the max length of data after which this parser won't be invoked */
     uint32_t max_depth;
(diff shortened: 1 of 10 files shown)


inliniac:
Replaced by #2484

PR #2472, opened 2016-12-22
tls-store: fix bug that causes Suricata to crash

thus:
Fix bug that causes Suricata to crash when the tls.store keyword is used.

```
*** Error in `/usr/bin/suricata': free(): invalid next size (fast): 0x00007fd4b4373180 ***
```

https://redmine.openinfosecfoundation.org/issues/1997

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/62
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/62


inliniac:
Merged through #2494, thanks Mats!

PR #2478, opened 2016-12-28
Fixed typo as described in Bug #2000

StianHB:
A small typo in output-json-http.c that is described in Bug #2000, 
the typo resulted in JSON objects with a http_refer field and not http_referer field.

### DIFF
--- a/src/output-json-http.c
+++ b/src/output-json-http.c
@@ -303,7 +303,7 @@ void JsonHttpLogJSONExtended(json_t *js, htp_tx_t *tx)
     if (h_referer != NULL) {
         c = bstr_util_strdup_to_c(h_referer->value);
         if (c != NULL) {
-            json_object_set_new(js, "http_refer", json_string(c));
+            json_object_set_new(js, "http_referer", json_string(c));
             SCFree(c);
         }
     }


inliniac:
Will this not break existing parsers/setups?

StianHB:
Probably, but I don't see any other good way of fixing this. Unless you want it to be mis-typed due to backwards compatability. We had an issue when converting from pure http logs to eve-json in our parses since we assumed http_referer and not http_refer.

So it might break current implementation, but the fix would be more correct for future parseres/setups.

inliniac:
Holding this off until we can do a breaking change, so the next major version. We will need to add a warning in a upgrade document.

inliniac:
I'm having 2nd thoughts here. Even though we're doing a major release I don't want to break things.

@jasonish any thoughts?

jasonish:
I wonder if we could queue it up with other logging breaking changes for a future major release. make sure they are all the first beta for that release and emphasize it in the beta announcement. But as this wasn't in the beta, I don't think we should pop it into an RC.

pevma:
I share the same view with @jasonish

PR #2479, opened 2017-01-10
dns - better response handling - v1

jasonish:
The first commit simply updates the transaction id when a transaction is being allocated during a response. This can happen when a request is sent, and then an unexpected response is seen which allocates a new transaction. This commit gives this new transaction a new ID instead of re-using the previous one. Not doing so caused the real response to not be found properly.

The second commit fixes tx logging in the case where a newer transaction is complete but an old one isn't. Currently this will update the last logged ID causing the incomplete transaction from being logged.  Instead this commit will only update the last logged ID if every transaction was logged. This required the function to take the tx id as an argument instead of a simple increment.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2001

An output verification test exists here:
https://github.com/jasonish/suricata-verify/tree/master/dns-udp-unsolicited-response

Prscript:
- https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/421
- https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/69

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -597,12 +597,12 @@ uint64_t AppLayerParserGetTransactionLogId(AppLayerParserState *pstate)
     SCReturnCT((pstate == NULL) ? 0 : pstate->log_id, "uint64_t");
 }
 
-void AppLayerParserSetTransactionLogId(AppLayerParserState *pstate)
+void AppLayerParserSetTransactionLogId(AppLayerParserState *pstate, uint64_t tx_id)
 {
     SCEnter();
 
     if (pstate != NULL)
-        pstate->log_id++;
+        pstate->log_id = tx_id;
 
     SCReturn;
 }
--- a/src/output-tx.c
+++ b/src/output-tx.c
@@ -163,11 +163,10 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data)
 
     uint64_t total_txs = AppLayerParserGetTxCnt(p->proto, alproto, alstate);
     uint64_t tx_id = AppLayerParserGetTransactionLogId(f->alparser);
+    int not_logged = 0;
 
     for (; tx_id < total_txs; tx_id++)
     {
-        int logger_not_logged = 0;
-
         void *tx = AppLayerParserGetTx(p->proto, alproto, alstate, tx_id);
         if (tx == NULL) {
             SCLogDebug("tx is NULL not logging");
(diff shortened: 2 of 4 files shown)


inliniac:
Is this for 3.2.1?

jasonish:
> Is this for 3.2.1?

Could be I think. It can catch previously missed DNS responses, and I found that it fixes some missed DNP3 responses, in particular unsolicited responses in that protocol, which are valid as they even have the message type "unsolicted_response".

src/app-layer-parser.c
```diff
@@ -597,12 +597,12 @@ uint64_t AppLayerParserGetTransactionLogId(AppLayerParserState *pstate)
...
 }
 
-void AppLayerParserSetTransactionLogId(AppLayerParserState *pstate)
+void AppLayerParserSetTransactionLogId(AppLayerParserState *pstate, uint64_t tx_id)
 {
     SCEnter();
 
     if (pstate != NULL)
-        pstate->log_id++;
+        pstate->log_id = tx_id;
```
inliniac: should we check that the log_id can't move backwards? Wonder if req0 req1 res1 res0 would set it to 0 when logging res0?

jasonish: That could happen if tx1 was free'd and removed fromthe list before tx0 because tx1 is done and tx0 is not.  I'm not sure if that can even happen right now.  But yeah, its a simple check and we don't want it going backwards.

inliniac: I guess if I understand the code correctly log_id will be set to 1 when res0 is logged?

jasonish: Yes, if thats the only TX.  But what it really does it update the LogId when all the transactions have been logged.  If some, but not all were logged it doesn't get updated, so the next loop starts earlier in the transaction list looking for tx's that are better to be logged.

I'm realizing this isn't very good.  So you have a constant stream of overlapping transactions, you may hit a state where there is always a tx behind the latest that is not ready to be logged, and this number will never get updated.

Instead I think we need to track the maximum tx id where is, and all transactions before it were completely logged, and update the LogId to that number. That will prevent the condition I describe above.

src/output-tx.c
```diff
@@ -242,10 +241,13 @@ static TmEcode OutputTxLog(ThreadVars *tv, Packet *p, void *thread_data)
...
-            SCLogDebug("updating log tx_id %"PRIu64, tx_id);
-            AppLayerParserSetTransactionLogId(f->alparser);
-        }
+    }
+
+    /* Update the greatest transaction logged if all loggers
+     * succeeded. */
+    if (!not_logged) {
+        SCLogDebug("updating log tx_id %"PRIu64, tx_id);
+        AppLayerParserSetTransactionLogId(f->alparser, tx_id);
```
inliniac: Since the above loop cannot be broken out of, tx_id will always be equal to total_txs AFAICT. That would mean that if any tx is logged succesfully, we update log_id to the highest?

jasonish:
New PR: #2486

PR #2482, opened 2017-01-13
eve-log: custom tls output and log certificate

thus:
Add the possibility to specify in suricata.yaml which tls fields to log in eve-log.

```
- tls:
    extended: no
    custom: [subject, issuer, fingerprint, sni, version, not_before, not_after, certificate, chain]
```

Also adds support for logging certificate and full certificate chain in eve-log.

https://redmine.openinfosecfoundation.org/issues/1998

- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/64
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/64


thus:
Just to clarify, the changes to OutputJSONBuffer are to enable the different loggers to pass the flags they want to libjansson. E.g. it enables us to not use JSON_ESCAPE_SLASH in tls eve-log, since it messes with the base64 encoded certificates.

inliniac:
Can you expand a bit on how it messes with base64 encoding? Perhaps show an example?

thus:
Base64 consists of A-Z, a-z, 0-9, '+', '/', and '=' (used for padding). Pasting an entire certificate would be too verbose, so I'll use a simple string (which would serve the same purpose) instead.

The following command generates a base64 string with a slash:
```
echo "???" | base64
Pz8/Cg==
```

With JSON_ESCAPE_SLASH as a libjansson flag, this would be changed to:
```
Pz8\/Cg==
```

Which means that we would have to post process all the certificates to remove all the backslashes, before we could decode them. This seems wasteful, when we could just avoid printing the (IMO useless) backslashes in the first place.

inliniac:
@decanio do you remember why you added the JSON_ESCAPE_SLASH in the first place? It's not explained in comment or commit. 0c067646a86973ee2426defc9110337a372add00 introduced it.

thus:
Replaced by #2491

decanio:
Victor.  If I remember correctly JSON is supposed to escape the / character.  However early versions of libjansson didn't do it, so they introduced JSON_ESCAPE_SLASH. in later versions to add the capability and to introduce the capability while maintaining backward compatibility.

thus:
The RFC [1] seems to state that it is allowed to escape slash, but it doesn't seem to require it. The examples in the RFC don't do it.

What if I add an option to the configuration file, allowing the user to select to escape slashes or not per logger, and keeping escaping as a default. Would that be an acceptable solution? 

[1] "RFC 7159 - The JavaScript Object Notation (JSON) Data Interchange Format"
https://tools.ietf.org/html/rfc7159

thus:
Fixing this also enables us to close https://redmine.openinfosecfoundation.org/issues/1608 :)

inliniac:
I think making optional per eve instance would be the way to go. I don't like the current logic in this PR where in a single eve instance we can have some records with it and others without it.

PR #2483, opened 2017-01-13
flow: add functionality to check if flow has alerts

thus:

Example script:

``` lua
function init (args)
    local needs = {}
    needs["type"] = "flow"
    return needs
end

function setup (args)
end

function log (args)

    if has_alerts then
        file:flush()
    end
end

function deinit (args)
    file:close(file)
end
```

Also adds a "has_alerts" field to the flow eve-log.

prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/67
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/67

### DIFF
--- a/src/output-json-flow.c
+++ b/src/output-json-flow.c
@@ -261,6 +261,14 @@ static void JsonFlowLogJSON(JsonFlowLogThread *aft, json_t *js, Flow *f)
     json_object_set_new(hjs, "reason",
             json_string(reason));
 
+    const char *has_alerts = NULL;
+    if (FlowHasAlerts(f))
+        has_alerts = "yes";
+    else
+        has_alerts = "no";
+
+    json_object_set_new(hjs, "has_alerts", json_string(has_alerts));
+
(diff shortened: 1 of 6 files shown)


src/output-json-flow.c
```diff
@@ -261,6 +261,14 @@ static void JsonFlowLogJSON(JsonFlowLogThread *aft, json_t *js, Flow *f)
     json_object_set_new(hjs, "reason",
             json_string(reason));
 
+    const char *has_alerts = NULL;
+    if (FlowHasAlerts(f))
+        has_alerts = "yes";
+    else
+        has_alerts = "no";
+
+    json_object_set_new(hjs, "has_alerts", json_string(has_alerts));
```
inliniac: Please use json_boolean

thus: I'll fix it and send you a new PR.

thus:
Replaced by #2490

PR #2488, opened 2017-01-17
unified2 - nostamp and HUP rotation - v1

jasonish:
Redmine issue:
https://redmine.openinfosecfoundation.org/issues/1636

Adds nostamp option to remove the timestamp from the unified output file. Additionally, if nostamp is one, register for log file rotation on SIGHUP.

Of note, the limit is still enforced which will truncate the file (Snort behaviour).

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/77
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/429

### DIFF
--- a/src/alert-unified2-alert.c
+++ b/src/alert-unified2-alert.c
@@ -225,7 +225,7 @@ static int Unified2IPv4TypeAlert(ThreadVars *, const Packet *, void *);
 static int Unified2IPv6TypeAlert(ThreadVars *, const Packet *, void *);
 static int Unified2PacketTypeAlert(Unified2AlertThread *, const Packet *, uint32_t, int);
 void Unified2RegisterTests(void);
-int Unified2AlertOpenFileCtx(LogFileCtx *, const char *);
+static int Unified2AlertOpenFileCtx(LogFileCtx *, const char *, bool);
 static void Unified2AlertDeInitCtx(OutputCtx *);
 
 int Unified2Condition(ThreadVars *tv, const Packet *p);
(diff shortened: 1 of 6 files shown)


src/alert-unified2-alert.c
```diff
@@ -1283,6 +1287,13 @@ OutputCtx *Unified2AlertInitCtx(ConfNode *conf)
                 exit(EXIT_FAILURE);
             }
         }
+
+        if (ConfGetChildValueBool(conf, "nostamp", &nostamp)) {
+            if (nostamp) {
+                SCLogConfig("Disabling unified2 timestamp.");
+                file_ctx->nostamp = 1;
```
inliniac: minor: nostamp type is bool, so this should be set to 'true'?

inliniac:
I've introduced some _Bool in the past, perhaps it would be good unify on bool/stdbool.h instead.

src/alert-unified2-alert.c
jasonish: Oops, yes. That should be true.

jasonish:
> I've introduced some _Bool in the past, perhaps it would be good unify on bool/stdbool.h instead.

I think so. Its my understanding that it was introduced as _Bool to prevent conflicts with code that introduced its own bool typedef, and that you have to opt-in to bool by including stdbool, which is the preferred way unless it breaks your code.

jasonish:
Closing, see: https://github.com/inliniac/suricata/pull/2560

PR #2490, opened 2017-01-17
flow: add functionality to check if flow has alerts (v4)

thus:

Changes:
- Use json_boolean instead of json_string

Example script:

``` lua
function init (args)
    local needs = {}
    needs["type"] = "flow"
    return needs
end

function setup (args)
end

function log (args)

    if has_alerts then
        file:flush()
    end
end

function deinit (args)
    file:close(file)
end
```

Also adds a "has_alerts" field to the flow eve-log.

prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/69
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/69

### DIFF
--- a/src/output-json-flow.c
+++ b/src/output-json-flow.c
@@ -261,6 +261,12 @@ static void JsonFlowLogJSON(JsonFlowLogThread *aft, json_t *js, Flow *f)
     json_object_set_new(hjs, "reason",
             json_string(reason));
 
+    uint8_t has_alerts = 0;
+    if (FlowHasAlerts(f))
+        has_alerts = 1;
+
+    json_object_set_new(hjs, "has_alerts", json_boolean(has_alerts));
+
     json_object_set_new(js, "flow", hjs);
 
 
--- a/src/util-lua-common.c
+++ b/src/util-lua-common.c
@@ -248,6 +248,41 @@ static int LuaCallbackFlowTimeString(lua_State *luastate)
 }
 
 /** \internal
+ *  \brief fill lua stack with flow has alerts
+ *  \param luastate the lua state
+ *  \param flow flow
+ *  \retval cnt number of data items placed on the stack
+ *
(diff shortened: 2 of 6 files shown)


src/output-json-flow.c
```diff
@@ -261,6 +261,12 @@ static void JsonFlowLogJSON(JsonFlowLogThread *aft, json_t *js, Flow *f)
     json_object_set_new(hjs, "reason",
             json_string(reason));
 
+    uint8_t has_alerts = 0;
```
inliniac: How about make this a bit more compact?
e.g.
uint8_t has_alerts = FlowHasAlerts(f);

Or even:
json_object_set_new(hjs, "has_alerts", json_boolean(FlowHasAlerts(f)));

src/util-lua-common.c
```diff
@@ -248,6 +248,41 @@ static int LuaCallbackFlowTimeString(lua_State *luastate)
...
+ *  \brief fill lua stack with flow has alerts
+ *  \param luastate the lua state
+ *  \param flow flow
+ *  \retval cnt number of data items placed on the stack
+ *
+ *  Places alerts (bool)
+ */
+static int LuaCallbackHasAlertsPushToStackFromFlow(lua_State *luastate, const Flow *flow)
+{
+    if (FlowHasAlerts(flow)) {
```
inliniac: How about:
lua_pushboolean(luastate, FlowHasAlerts(flow));

src/output-json-flow.c
thus: Good idea. I'll change it.

src/util-lua-common.c
thus: That's much better. I'll change it.

thus:
Replaced by #2513

PR #2486, opened 2017-01-16
dns - better response handling - v2

jasonish:
Previous PR: #2479 

Changes from last PR:

Update the LogId to the maximum ID that has been logged where all earlier transactions have also been logged. For example, where "*" means that the request and response have been logged:

1*, 2*, 3*, 4*, 5* -> 5.

1*, 2, 3*, 4*, 5 -> 1.

1*, 2*, 3*, 4, 5* -> 4.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2001

An output verification test exists here:
https://github.com/jasonish/suricata-verify/tree/master/dns-udp-unsolicited-response

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/72
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/424

PR #2493, opened 2017-01-17
tx logging: only update logged tx id if all loggers logged - v3

jasonish:
Previous PR: #2486 

Change from last PR:
- As there can be multiple loggers registered we have to make sure that all loggers logged a transaction instead of just one of them, we also have to take into account a gap in logging, that is when a transaction is not logged, but a more recent one is - we now have a gap in the logged transactions.

Prevents the case where the logged id is incremented if a newer
transaction is complete and an older one is still outstanding.

For example, dns request0, unsolicited dns response, dns response0

would result in the valid response0 never being logged.

Similarily this could happen for:
  request0, request1, response1, response0

which would end up having request0, request1 and response1 logged,
but response0 would not be logged.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/78
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/430

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -597,12 +597,12 @@ uint64_t AppLayerParserGetTransactionLogId(AppLayerParserState *pstate)
     SCReturnCT((pstate == NULL) ? 0 : pstate->log_id, "uint64_t");
 }
 
-void AppLayerParserSetTransactionLogId(AppLayerParserState *pstate)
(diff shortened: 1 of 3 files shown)


inliniac:
Merged in #2499, thanks Jason!

PR #2496, opened 2017-01-17
util-file: fix error logic in hash computation

regit:
This patch fixes an issue with hash computation resulting in the
invalidity of at least one hash when at least two different hashes
functions were used.

Impact was setting as `force-hash: [md5, sha256]` not to be valid.
Also it could lead to false negative if too different hash functions
had to be used on a single file due to signatures.

Redmine issue: https://redmine.openinfosecfoundation.org/issues/2004

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/234
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/16


inliniac:
I took the hash patch in #2499, thanks Eric!

PR #2497, opened 2017-01-18
Stream depth fix

regit:
Stream config was initialized after the application layer causing the per application layer stream depth to be set to zero.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/235
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/17


inliniac:
I think it would be cleaner to completely split the config loading and stream engine init into 2 separate functions.

regit:
pfff, indeed ;) closing and resubmitting.

PR #2504, opened 2017-01-18
Bug 2005

regit:
This patchset fixes file size computation by using a custom field instead of using the FileSize function which is dependant of storage.

Redmine issue: https://redmine.openinfosecfoundation.org/issues/2005

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/236
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/18


inliniac:
This needs some more thought. There are more callers that use FileSize, including other file related logging that are not addressed.

Perhaps it makes sense to have something like a FileTrackedSize and FileDataSize so that the caller can be explicit about what it wants.

PR #2505, opened 2017-01-18
stream-tcp: separate stream config parsing and init

regit:
Update of #2497 addressing the remark.

This patch separates the stream config parsing and init so we can
read the config before init the application layer as the stream
reassembly depth value is needed when doing it.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/238
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/20

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2498,6 +2498,8 @@ int main(int argc, char **argv)
         exit(EXIT_FAILURE);
     }
 
+    StreamTcpParseConfig(STREAM_VERBOSE);
+
     if (PostConfLoadedSetup(&suri) != TM_ECODE_OK) {
         exit(EXIT_FAILURE);
     }
--- a/src/stream-tcp.c
+++ b/src/stream-tcp.c
@@ -303,18 +303,9 @@ static void StreamTcpSessionPoolCleanup(void *s)
     }
 }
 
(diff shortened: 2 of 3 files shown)


inliniac:
Is it intentional that the unix socket runmode is not updated? I guess the stream config would stay valid, although I haven't looked at what the clean up code does to the config, if anything. Needs clarification in commit message and perhaps also in the code itself.

regit:
The stream tcp parsing code is called inconditionaly so we have a complete structure. Regarding the init, it is done via the already existing function so we can do not have to change the code.

src/suricata.c
```diff
@@ -2498,6 +2498,8 @@ int main(int argc, char **argv)
         exit(EXIT_FAILURE);
     }
 
+    StreamTcpParseConfig(STREAM_VERBOSE);
```
inliniac: Why not in PostConfLoadedSetup()?

inliniac:
This breaks -k none. If I specify it on the commandline I still get: "[ERRCODE: SC_ERR_INVALID_CHECKSUM(11)] - 1/1th of packets have an invalid checksum, consider setting pcap-file.checksum-checks variable to no or use '-k none' option on command line."

PR #2506, opened 2017-01-19
pcap-log: fix pcre_study error check - v1

jasonish:
Code was failing on a NULL return value which can be returned
when there was nothing todo instead of an error. Instead
check the errbuf for a non-NULL value to determine error.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/81
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/433


inliniac:
Merged, thanks Jason!

PR #2507, opened 2017-01-19
Experiments/rust/quic/v2

inliniac:
Testing stuff with rust/suricata/rusticata.

cc: @chifflier @jasonish 

See also: https://github.com/rusticata https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Rust_Notes


jasonish:
I wonder if each parser should be its on project, or if rust/src should be laid out like Suricata src, with only one resulting library..

rust/src/dns.rs
rust/src/quick.rs

inliniac:
On 19-01-17 15:53, Jason Ish wrote:
> I wonder if each parser should be its on project, or if rust/src should
> be laid out like Suricata src, with only one resulting library..
> 
> rust/src/dns.rs
> rust/src/quick.rs

Either would work I think.

Depending on how we use the parsers, they may even be split in 2: 1)
pure rust/nom parser, 2) suricata glue.

Many choices to be made :)

rust/quic/src/quic.rs
```diff
@@ -0,0 +1,263 @@
...
+struct QuicParser;
+
+impl<'a> RParser<QuicParserState<'a>> for QuicParser {
+    fn new_state() -> QuicParserState<'a> {
+        QuicParserState::new(b"blah")
+    }
+
+    fn probe(i: &[u8]) -> bool {
+        match parse_quic(i) {
+            IResult::Done(_, quic_packet) => {
```
chifflier: If you don't use `quic_packet` you can use `IResult::Done(_,_)` to avoid an unused variable warning

rust/quic/src/quic.rs
```diff
@@ -0,0 +1,263 @@
...
+        connection_id:Some(&bytes[1..9]),
+        seq_id:&session_id,
+        payload:&bytes[14..],
+    });
+    let res = parse_quic(&bytes);
+    println!("{:?}",res);
+    assert_eq!(res, expected);
+}
+
+}
```
chifflier: Globally, except for the debugging code, the parser and the unit tests look good to me

rust/quic/src/rparser.rs
```diff
@@ -0,0 +1,93 @@
+// Written by Pierre Chifflier
```
chifflier: This file should also be moved to a common crate

rust/quic/src/quic.rs
```diff
@@ -0,0 +1,263 @@
...
+               payload: pl,
+           }
+   ))
+);
+
+struct QuicParser;
+
+impl<'a> RParser<QuicParserState<'a>> for QuicParser {
+    fn new_state() -> QuicParserState<'a> {
+        QuicParserState::new(b"blah")
```
chifflier: I am a bit ashamed, I know I left some "blah" in the example code :)

rust/quic/src/quic.rs
```diff
@@ -0,0 +1,263 @@
...
+                println!("rust: incomplete");
+            }
+            IResult::Error(_) => {
+                println!("rust: error");
+            }
+        }
+
+        //let d = parse_quic(i);
+        //println!("d {:?}", d);
+        let status = R_STATUS_OK;
```
chifflier: I think status should be set to OK at the beginning of the function, and or'ed with parser events, or set to error.
The case for incomplete is more difficult: if the protocols supports fragmentation, data must be copied in the parser state and tested in the next function call.

PR #2009, opened 2016-04-14
1.  The enip-feature-1495-v6 branch contains support for the ENIP/CIP

kwong-solana:
1.  The enip-feature-1495-v6 branch contains support for the ENIP/CIP
Industrial protocol
2. This is an app layer implementation which uses the "enip" protocol
and "cip_service" and "enip_command" keywords
3.  Rebased master to clean out commits done on master and then
merged to latest code
4.  Addresses failed PR's https://github.com/inliniac/suricata/pull/1810
and https://github.com/inliniac/suricata/pull/1819 and
https://github.com/inliniac/suricata/pull/1827
5.  As requested, added AFL entry points

PR #2508, opened 2017-01-19
Bug #2009: added CAP_NET_ADMIN for PCAP and af-packet modes.

gozzy:
Without this capability suricata is unable to get network interface's settings.


inliniac:
Merged, thanks Alexander!

PR #2513, opened 2017-01-20
flow: add functionality to check if flow has alerts (v6)

thus:

Example script:

``` lua
function init (args)
    local needs = {}
    needs["type"] = "flow"
    return needs
end

function setup (args)
end

function log (args)

    if has_alerts then
        file:flush()
    end
end

function deinit (args)
    file:close(file)
end
```

Also adds a "alerted" field to the flow eve-log.

prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/83
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/83

### DIFF
--- a/doc/userguide/output/lua-output.rst
+++ b/doc/userguide/output/lua-output.rst
@@ -178,6 +178,22 @@ Example:
       end
   end
 
+SCFlowHasAlerts
+~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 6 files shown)


inliniac:
Merged through #2516, thanks Mats!

PR #2010, opened 2016-04-14
Adding SCADA EtherNet/IP and CIP protocol support

kwong-solana:
1.  The enip-feature-1495-v6 branch contains support for the ENIP/CIP
Industrial protocol

2. This is an app layer implementation which uses the "enip" protocol
and "cip_service" and "enip_command" keywords

3.  Rebased master to clean out commits done on master and then
merged to latest code

4.  Addresses failed PR's
https://github.com/inliniac/suricata/pull/1810
and https://github.com/inliniac/suricata/pull/1819 and
https://github.com/inliniac/suricata/pull/1827 and
https://github.com/inliniac/suricata/pull/2009

5.  As requested, added AFL entry points

PR #2517, opened 2017-01-24
mpm/spm: check for SSSE3 and enable/disable Hyperscan support accordingly

satta:
The new Hyperscan 4.4 API (https://github.com/01org/hyperscan/releases/tag/v4.4.0) provides a function to check for SSSE3 presence at runtime. This allows us to fall back to non-Hyperscan matchers on systems without SSSE3 even when the suricata executable is built with Hyperscan support. Addresses Redmine issue #2010 (https://redmine.openinfosecfoundation.org/issues/2010).

 - Compile time check for presence of `hs_valid_platform()` available since Hyperscan 4.4.
 - Runtime check whether this call is successful (=SSSE3 available), resetting default SPM and MPM matchers to non-Hyperscan variants if not.

Many thanks for your careful review and comments.

### DIFF
--- a/src/util-mpm.c
+++ b/src/util-mpm.c
@@ -46,6 +46,9 @@
 #include "detect-engine-mpm.h"
 #endif
 #include "util-memcpy.h"
+#ifdef BUILD_HYPERSCAN
+#include "hs.h"
+#endif
 
 /**
  * \brief Register a new Mpm Context.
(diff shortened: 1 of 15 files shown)


jviiret:
(Hyperscan dev team member here)

This looks good to me, and is exactly what we intended `hs_valid_platform()` to be used for. Thanks Sascha!

satta:
Thanks Justin! If there's anything else just let me know.

src/util-spm.c
```diff
@@ -89,7 +92,20 @@ uint16_t SinglePatternMatchDefaultMatcher(void)
...
+     * check is not successful. */
+        if (hs_valid_platform() != HS_SUCCESS) {
+            SCLogInfo("SSSE3 support not detected, disabling Hyperscan for "
+                      "SPM");
+            /* Use Boyer-Moore as fallback. */
+            return SPM_BM;
+        } else {
+            return SPM_HS;
+        }
+    #elif
```
inliniac: missing condition, probably should be #else?

src/util-spm.c
```diff
@@ -102,7 +118,13 @@ void SpmTableSetup(void)
 
     SpmBMRegister();
 #ifdef BUILD_HYPERSCAN
-    SpmHSRegister();
+    #ifdef HAVE_HS_VALID_PLATFORM
+        if (hs_valid_platform() == HS_SUCCESS) {
+            SpmHSRegister();
+        }
+    #elif
```
inliniac: missing condition, probably should be #else?

src/util-spm.c
satta: Totally. Will fix this in `-v2`later today.

src/util-spm.c
satta: Right, thanks for noticing.

inliniac:
Breaks almost all of my tests because of the #elif issue.

Please see https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Create-a-new-branch-for-incorporating-feedback Thanks!

src/util-mpm.c
```diff
@@ -402,15 +405,46 @@ void MpmInitCtx (MpmCtx *mpm_ctx, uint16_t matcher)
...
+     * check is not successful. */
+        if (hs_valid_platform() != HS_SUCCESS) {
+            SCLogInfo("SSSE3 support not detected, disabling Hyperscan for "
+                      "MPM");
+            /* Fall back to best Aho-Corasick variant. */
+            mpm_default_matcher = DEFAULT_MPM_AC;
+        } else {
+            MpmHSRegister();
+        }
+    #elif
```
inliniac: Ah one more.

PR #2518, opened 2017-01-26
threads: fix missed logging at shutdown

inliniac:
At shutdown, all flows that still need work are handled by the flow
force reassembly logic. This means one or more flow end pseudo packets
are generated and pushed through the engine for final detection and
logging.

In some cases this would not work correctly. This was caused by the
flow timeout logic kicking in before all the 'live' packets were
processed. Before the flow timeout handling runs the receive threads
are disabled, however the engine did not wait for the in-flight
packets to be fully processed. In autofp mode, packets could still
be in the queue between receive thread(s) and flow worker(s).

This patch adds a new function that 'drains' all the packet threads
of any in-progress packets before moving on the flow timeout logic.

Bug #1946. https://redmine.openinfosecfoundation.org/issues/1946

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/58
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/566

### DIFF
--- a/src/tm-threads.c
+++ b/src/tm-threads.c
@@ -1521,6 +1521,64 @@ void TmThreadKillThread(ThreadVars *tv)
     return;
 }
 
+/** \internal
+ *
+ *  \brief make sure that all packet threads are done processing their
+ *         in-flight packets
+ */
+static void TmThreadDrainPacketThreads(void)
+{
+    /* value in seconds */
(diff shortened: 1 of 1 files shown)


inliniac:
@fooinha can you test this? Fixes the issue for me. Thanks!

fooinha:
This seems fixed.

PR #2520, opened 2017-01-26
Cleanup/unifyinit/v6

inliniac:
Unify setup and cleanup between the main runmode and the per pcap file unix run mode.

General cleanups to shrink main().

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/59
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/567

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -312,7 +312,7 @@ void CreateLowercaseTable()
     }
 }
 
-void GlobalInits()
+void GlobalsInitPreConfig()
 {
 #ifdef __SC_CUDA_SUPPORT__
     /* Init the CUDA environment */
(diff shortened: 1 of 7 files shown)


src/suricata.c
```diff
@@ -2032,6 +2043,89 @@ static int InitSignalHandler(SCInstance *suri)
...
+ * unix socket mode.
+ *
+ * Will be run once per pcap in unix-socket mode */
+void PostRunDeinit(const int runmode, struct timeval *start_time)
+{
+    if (runmode == RUNMODE_UNIX_SOCKET)
+        return;
+
+    /* needed by FlowForceReassembly */
+    PacketPoolInit();
```
regit: More coffee for me but looks strange. Maybe better comment stating that it can be called twice or something alike ?

src/unix-manager.c
```diff
@@ -1011,6 +1011,19 @@ void UnixManagerThreadSpawn(int mode)
     return;
 }
 
+// TODO can't think of a good name
+void UnixManagerThreadSpawnNonRunmode(void)
```
regit: UnixManagerThreadSpawnLiveMode ?

src/suricata.c
```diff
@@ -345,6 +345,77 @@ void GlobalInits()
...
+    HTPAtExitPrintStats();
+
+#ifdef DBG_MEM_ALLOC
+    SCLogInfo("Total memory used (without SCFree()): %"PRIdMAX, (intmax_t)global_mem);
+#ifdef DBG_MEM_ALLOC_SKIP_STARTUP
+    print_mem_flag = 0;
+#endif
+#endif
+
+    SCPidfileRemove(suri->pid_filename);
```
regit: We could start a other suri if PID is removed. This could result in two suricata logging to file at the same time in same file. Could we move that later ?

src/suricata.c
```diff
@@ -312,7 +312,7 @@ void CreateLowercaseTable()
     }
 }
 
-void GlobalInits()
+void GlobalsInitPreConfig()
```
regit: change not described in commit message ;)

src/suricata.c
```diff
@@ -2567,12 +2634,58 @@ static int PostConfLoadedSetup(SCInstance *suri)
...
 
+static void SuricataMainLoop(SCInstance *suri)
+{
+    while(1) {
+        if (sigterm_count || sigint_count) {
+            suricata_ctl_flags |= SURICATA_STOP;
+        }
+
+        if (suricata_ctl_flags & SURICATA_STOP) {
+            SCLogNotice("Signal Received.  Stopping engine.");
```
regit: Two spaces.

src/suricata.c
inliniac: I don't understand your question here

src/suricata.c
inliniac: Not changed by this PR, but I'll move it back anyway.

src/suricata.c
inliniac: Is the change not self explaining?

src/suricata.c
regit: We are in a Deinit function and we call a Init function. At first sight it looks like a bug so a comment explaining the case in simple term could be nice.

src/unix-manager.c
inliniac: Live mode suggests live traffic capture. We also spawn for offline pcap reading.

src/suricata.c
inliniac: It says FlowForceReassembly needs it.

src/suricata.c
regit: yes, no technical issue here.

inliniac:
Merged through #2524

PR #2521, opened 2017-01-27
output-json-alert: print 'parent' JSON object if tunnel

thus:
Print "src_ip", "dst_ip" and "proto" for root packet (p->root) as JSON object 'parent' if packet is a tunnel. 

Without this, it is difficult to find the traffic that triggered if the match was done inside a tunnel (e.g. GRE tunnel), since far from all network solutions decode tunnels.

Example alert with 'parent' object:
```json
{
  "timestamp": "2008-06-21T14:06:06.578905+0200",
  "event_type": "alert",
  "src_ip": "1.1.1.1",
  "dest_ip": "2.2.2.2",
  "proto": "ICMP",
  "icmp_type": 8,
  "icmp_code": 0,
  "alert": {
    "severity": 3,
    "category": "",
    "signature": "ICMP something something",
    "rev": 0,
    "signature_id": 1230910,
    "gid": 1,
    "action": "allowed"
  },
  "parent": {
    "proto": "GRE",
    "dest_ip": "10.0.0.2",
    "src_ip": "10.0.0.1"
  }
}

```

https://redmine.openinfosecfoundation.org/issues/2011

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/86
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/86

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -217,6 +217,25 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
     json_object_set_new(js, "alert", ajs);
 }
 
+static void AlertJsonParent(const Packet *p, json_t *js)
+{
+    json_t *parent = json_object();
+    if (parent == NULL)
+        return;
+
+    if (p->root == NULL)
+        return;
+
+    /* get a lock to access root packet fields */
+    SCMutex *m = &p->root->tunnel_mutex;
+
+    SCMutexLock(m);
+    JsonFiveTuple((const Packet *)p->root, 0, parent);
+    SCMutexUnlock(m);
+
+    json_object_set_new(js, "parent", parent);
+}
+
 static void AlertJsonPacket(const Packet *p, json_t *js)
 {
     unsigned long len = GET_PKT_LEN(p) * 2;
@@ -259,6 +278,10 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
 
         /* alert */
         AlertJsonHeader(p, pa, js);
+
+        if (IS_TUNNEL_PKT(p)) {
+            AlertJsonParent(p, js);
+        }
 
         if (json_output_ctx->flags & LOG_JSON_HTTP) {
             if (p->flow != NULL) {
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -277,7 +277,7 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
             if (p->flow != NULL) {
                 uint16_t proto = FlowGetAppProtocol(p->flow);
 
-                /* http alert */
+                /* tls alert */
                 if (proto == ALPROTO_TLS)
                     AlertJsonTls(p->flow, js);
             }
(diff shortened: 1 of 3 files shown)


src/output-json-alert.c
```diff
@@ -217,6 +217,25 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
...
 }
 
+static void AlertJsonParent(const Packet *p, json_t *js)
+{
+    json_t *parent = json_object();
+    if (parent == NULL)
+        return;
+
+    if (p->root == NULL)
+        return;
```
inliniac: parent is lost

src/output-json-alert.c
```diff
@@ -217,6 +217,25 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
...
+
+    /* get a lock to access root packet fields */
+    SCMutex *m = &p->root->tunnel_mutex;
+
+    SCMutexLock(m);
+    JsonFiveTuple((const Packet *)p->root, 0, parent);
```
inliniac: could be interesting to add in Packet::recursion_level. If we have multiple tunnel layers it will show how deep this packet is.

inliniac:
Great suggestion Mats!

Feature ticket pretty please? :)

src/output-json-alert.c
thus: It might be due to the fact that it is friday, but I don't understand what you mean. Could you explain?

src/output-json-alert.c
thus: That is indeed something that is interesting to add.

inliniac:
On 27-01-17 16:07, Mats Klepsland wrote:
> *@thus* commented on this pull request.
> 
> 
> In src/output-json-alert.c <https://github.com/inliniac/suricata/pull/2521>:
> 
>> @@ -217,6 +217,25 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
>      json_object_set_new(js, "alert", ajs);
>  }
>  
> +static void AlertJsonParent(const Packet *p, json_t *js)
> +{
> +    json_t *parent = json_object();
> +    if (parent == NULL)
> +        return;
> +
> +    if (p->root == NULL)
> +        return;
> 
> It might be due to the fact that it is friday, but I don't understand
> what you mean. Could you explain?

return after parent was allocated, but not freed.

thus:
Thanks! I'll add a feature ticket! I actually anticipated that you would ask for it, so I should have added it in the first place :P

thus:
I totally missed that. Thanks!

jasonish:
Would "tunnel" be a better name than parent?  Its describing the type and address points of the tunnel.

thus:
I actually called it "tunnel" initially, but dropped that name since I thought people might find it confusing, and think it contains the IP addreses inside the tunnel instead. 

Looking at it now, I see that the "proto" field removes that confusion, since it will always contain a tunnel protocol. 

I can change it to "tunnel" if you guys agree that it is the best name. @inliniac, what do you think?

thus:
Replaced by #2525

PR #2522, opened 2017-01-27
mpm/spm: check for SSSE3 and enable/disable HS

satta:
Update with fix for #2517. Now with correct `#else`s. 

I also ran the Dockerized `prscript` (thanks to @regit for his quick action), all builds completed successfully:
 - https://steinbiss.name/suri/hyperscan-runtime-check-2010-v2/features.html

Addresses Redmine issue #2010 (https://redmine.openinfosecfoundation.org/issues/2010).

 - Compile time check for presence of hs_valid_platform() available since Hyperscan 4.4.
- Runtime check whether this call is successful (=SSSE3 available), resetting default SPM and MPM matchers to non-Hyperscan variants if not.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -618,6 +618,7 @@
         fi
 
         AC_CHECK_LIB(hs,hs_compile,,HYPERSCAN="no")
+        AC_CHECK_FUNCS(hs_valid_platform)
         enable_hyperscan="yes"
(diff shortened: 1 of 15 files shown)


inliniac:
Merged through #2524, thanks Sascha!

PR #2525, opened 2017-01-30
output-json-alert: log 'tunnel' JSON object when alerting on traffic in tunnel

thus:
Log "src_ip", "dst_ip" and "proto" for root packet (p->root) as JSON object 'tunnel' when alerting on traffic inside a tunnel. Also log recursion level to indicate tunnel depth.

Without this, it is difficult to find the traffic that triggered if the match was done inside a tunnel (e.g. GRE tunnel), since far from all network solutions decode tunnels.

Example alert with 'tunnel' object:
```json
{
  "timestamp": "2008-06-21T14:06:06.578905+0200",
  "event_type": "alert",
  "src_ip": "1.1.1.1",
  "dest_ip": "2.2.2.2",
  "proto": "ICMP",
  "icmp_type": 8,
  "icmp_code": 0,
  "alert": {
    "severity": 3,
    "category": "",
    "signature": "ICMP something something",
    "rev": 0,
    "signature_id": 1230910,
    "gid": 1,
    "action": "allowed"
  },
  "tunnel": {
    "proto": "GRE",
    "dest_ip": "10.0.0.2",
    "src_ip": "10.0.0.1",
    "depth": 1
  }
}

```

https://redmine.openinfosecfoundation.org/issues/2011

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/87
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/87

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -217,6 +217,29 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
     json_object_set_new(js, "alert", ajs);
 }
 
+static void AlertJsonTunnel(const Packet *p, json_t *js)
+{
+    json_t *tunnel = json_object();
+    if (tunnel == NULL)
+        return;
+
+    if (p->root == NULL) {
+        json_decref(tunnel);
+        return;
+    }
+
+    /* get a lock to access root packet fields */
+    SCMutex *m = &p->root->tunnel_mutex;
+
+    SCMutexLock(m);
+    JsonFiveTuple((const Packet *)p->root, 0, tunnel);
+    SCMutexUnlock(m);
+
+    json_object_set_new(tunnel, "depth", json_integer(p->recursion_level));
+
+    json_object_set_new(js, "tunnel", tunnel);
+}
+
 static void AlertJsonPacket(const Packet *p, json_t *js)
 {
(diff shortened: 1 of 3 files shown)


inliniac:
Is the example above correct? In the code I see it as 'tunnel' I think.

thus:
You are right, the example is wrong. I've updated it now.

src/output-json-alert.c
```diff
@@ -217,6 +217,29 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
...
+    if (tunnel == NULL)
+        return;
+
+    if (p->root == NULL) {
+        json_decref(tunnel);
+        return;
+    }
+
+    /* get a lock to access root packet fields */
+    SCMutex *m = &p->root->tunnel_mutex;
```
thus: Do we really need to lock here? According to flow.h the values we are accessing are static and can be accessed without a lock.

inliniac: It's not about the flow but the packet. I think the lock *might* be unnecessary, but I'm only interested in considering that if it actually shows up in profiling. Otherwise I prefer to keep things consistent, not in the least to assist code checkers like coverity.

thus:
Replaced by #2566

PR #2526, opened 2017-01-30
Preview: DNS app-layer in Rust - v7

jasonish:
This is an example of a more complete app-layer written in Rust. There is still a minimal DNS app-layer in C for bridging into Rust, but this can be reduced even more.

The Rust code attempts to create a "dns" module as well as a "root" module where core/common stuff can go, but all using a single crate. This does limit the usability of the parsers as a library to other projects, but my focus is only Suricata right now.

As far as DNS goes there are still some things missing:
- Flood handling.
- Lost response handling.
- Log rtype filtering.
- Not all types logged with proper data.
- Lua (in fact, just noticed that this branch will fail to build if Lua is enabled)

This branch (runmode=single):
```
Runtime: 2m14.156s
dns                     IPv4       6            63             3297          16848          8470        533.6k  0.00  
dns                     IPv4      17       1436207             2262       74195067         12740         18.3b  100.00
DNS event count: 3591370
```

Master (runmode=single):
```
Runtime: 1m49.968s
dns                     IPv4       6            63             3039          24000          7561        476.4k  0.00  
dns                     IPv4      17       1436207             1122       42113061          8881         12.8b  99.99 
DNS event count: 3580285
```

The big difference in the time appears to be the name parsing as it requires indexing into an array which I understand has a higher overhead in Rust due to internal bounds checking (for safety).  There probably is still room for optimization here, which may just come from more time with Rust learning the idiomatic way of doing things the best way.

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -1,0 +1,907 @@
+/* Copyright (C) 2017 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+// TODO: Log reply handling (windowed).
+// TODO: Flood handling.
+
+extern crate std;
+extern crate libc;
+extern crate nom;
+
+use std::slice;
+use std::mem::transmute;
+use std::panic;
+use std::ptr;
+
+use core;
+
+use dns::*;
+
+/// DNS error codes.
+pub const DNS_RCODE_NOERROR:  u16 = 0;
+pub const DNS_RCODE_FORMERR:  u16 = 1;
+pub const DNS_RCODE_NXDOMAIN: u16 = 3;
+
+/// DNS record types.
+pub const DNS_RTYPE_A:     u16 = 1;
+pub const DNS_RTYPE_CNAME: u16 = 5;
+pub const DNS_RTYPE_SOA:   u16 = 6;
+pub const DNS_RTYPE_PTR:   u16 = 12;
+pub const DNS_RTYPE_MX:    u16 = 15;
+pub const DNS_RTYPE_SSHFP: u16 = 44;
+pub const DNS_RTYPE_RRSIG: u16 = 46;
+
+#[repr(u32)]
+pub enum DNSEvents {
+    UnsolicitedResponse = 1,
+    MalformedData,
+    NotRequest,
+    NotResponse,
+    ZFlagSet,
+    Flooded,
+    StateMemCapReached,
+}
+
+/// Expose a parsers state ::new to C.
+/// TODO: Move out into some common module.
+macro_rules!export_state_new {
+    ($name: ident, $parser: ident) => {
+        #[no_mangle]
+        pub extern "C" fn $name() -> *mut $parser {
+            let state = $parser::new();
+            let boxed = Box::new(state);
+            return unsafe{transmute(boxed)};
+        }
+    }
+}
+
+/// Expose a function to free a parser state to C.
+/// TODO: Move out into some common module.
+macro_rules!export_state_free {
+    ($name: ident, $parser: ident) => {
+        #[no_mangle]
+        pub extern "C" fn $name(state: *mut $parser) {
+            let _drop: Box<$parser> = unsafe{transmute(state)};
+        }
+    }
+}
+
+#[derive(Debug,PartialEq)]
+pub struct DNSHeader {
+    pub tx_id: u16,
+    pub flags: u16,
+    pub questions: u16,
+    pub answer_rr: u16,
+    pub authority_rr: u16,
+    pub additional_rr: u16,
+}
+
+#[derive(Debug)]
+pub struct DNSTransaction {
+    pub id: u64,
+    pub request: Option<DNSRequest>,
+    pub response: Option<DNSResponse>,
+    pub replied: bool,
+}
+
+#[derive(Debug)]
+pub struct DNSRequest {
+    pub header: DNSHeader,
+    pub queries: Vec<DNSQueryEntry>,
+}
+
+#[derive(Debug)]
+pub struct DNSResponse {
+    pub header: DNSHeader,
+    pub queries: Vec<DNSQueryEntry>,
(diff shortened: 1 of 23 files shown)


rust/src/dns/dns.rs
```diff
@@ -0,0 +1,907 @@
...
+                }
+                return 1;
+            }
+        }
+    }
+    return 0;
+}
+
+#[no_mangle]
+pub extern fn rs_dns_tx_inspect_query_name(
```
inliniac: I think I would prefer changing this logic. Instead of exposing this suri internal, I would probably like it better to just register a data/buffer getter that the C code uses to pull the relevant data out of the rust structures. We could call it in a  loop for multiple buffers like here.

The same logic could be used for mpm.

rust/src/dns/dns.rs
```diff
@@ -0,0 +1,907 @@
...
+        }
+    }
+
+}
+
+#[no_mangle]
+pub extern fn rs_dns_tx_get_alstate_progress(tx: &mut DNSTransaction,
+                                             direction: libc::uint8_t)
+                                             -> libc::uint8_t {
+    if direction & core::STREAM_TOCLIENT > 0 {
```
inliniac: I would like to make this direction simpler. Like just passing 0 or 1 from the C code.

rust/src/dns/dns.rs
jasonish: So more like how PrefilterTxDnsQuery is done (https://github.com/inliniac/suricata/blob/563be87e0330a2c7ab3fee04a0de0c39735fa10f/src/detect-engine-dns.c#L89) instead of DetectEngineInspectDnsQueryName (https://github.com/inliniac/suricata/blob/563be87e0330a2c7ab3fee04a0de0c39735fa10f/src/detect-engine-dns.c#L68). 

Thought I'd try both styles.

Note: Needs another function to return the length rather than blindly looping.

rust/src/json.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    fn json_string(value: *const c_char) -> *mut JsonT;
+    fn json_integer(val: u64) -> *mut JsonT;
+}
+
+pub struct Json {
+    pub js: *mut JsonT,
+}
+
+impl Json {
```
inliniac: We've seen jansson fall over under memory pressure, I wonder how we'd handle that from rust. I guess in general I also would like to know how rust deals with memory pressure.

src/app-layer-dns-common.h
```diff
@@ -134,77 +138,25 @@ typedef struct DNSHeader_ {
...
-    TAILQ_HEAD(, DNSQueryEntry_) query_list;        /**< list for query/queries */
-    TAILQ_HEAD(, DNSAnswerEntry_) answer_list;      /**< list for answers */
-    TAILQ_HEAD(, DNSAnswerEntry_) authority_list;   /**< list for authority records */
 
     AppLayerDecoderEvents *decoder_events;          /**< per tx events */
 
     TAILQ_ENTRY(DNSTransaction_) next;
     DetectEngineState *de_state;
+
+    RSDNSTransaction *rs_tx;
```
inliniac: For me conceptually it makes more sense to try to keep the tx in rust completely, and same for the state. That would avoid double allocations and such, but of course will lead to more calls into rust.

src/app-layer-dns-tcp.c
```diff
@@ -47,234 +47,99 @@
...
-        SCLogDebug("trailer type %04x class %04x", ntohs(trailer->type), ntohs(trailer->class));
-        data += sizeof(DNSQueryTrailer);
-
-        /* store our data */
-        if (dns_state != NULL) {
-            DNSStoreQueryInState(dns_state, fqdn, fqdn_offset,
-                    ntohs(trailer->type), ntohs(trailer->class),
-                    ntohs(dns_header->tx_id));
-        }
+    if (tx_id > 0) {
```
inliniac: so I guess the logic here is that after rust ran we need to try to find out if it created a tx and then create our own C owned tx, look up the rust tx and reference it in the C-tx?

rust/src/json.rs
jasonish: I did look into this a bit, and this is where Rust differs a lot from C - you get no NULL from a malloc to act on.  Instead, I think now a panic is raised, which can be handled, you just don't know exactly where it happened (without looking at a backtrace), but at least you can return a failure back into C, eg:

https://github.com/inliniac/suricata/blob/563be87e0330a2c7ab3fee04a0de0c39735fa10f/rust/src/dns/dns.rs#L470

src/detect-engine-dns.c
```diff
@@ -112,19 +95,19 @@ static void PrefilterTxDnsQuery(DetectEngineThreadCtx *det_ctx,
 
     const MpmCtx *mpm_ctx = (MpmCtx *)pectx;
     DNSTransaction *tx = (DNSTransaction *)txv;
-    DNSQueryEntry *query = NULL;
 
-    TAILQ_FOREACH(query, &tx->query_list, next) {
-        SCLogDebug("tx %p query %p", tx, query);
+    for (uint16_t i = 0; i < 0xffff; i++) {
```
inliniac: I like this approach better as for me it keeps a cleaner boundary between rust and C. Don't really like this loop 'i<64k' but I guess that isn't even needed as we'll never get that many records. We can probably just make it an endless loop as we'll break out because of rs_dns_tx_get_query_buffer retval.

src/app-layer-dns-tcp.c
jasonish: Yup, not ideal.  A pure rust TX would be the next step.  So far only hanging on to the C one for the logged flag, de_state, and decoder_events - at least for decoder events it means calling back into C from Rust.

src/log-dnslog.c
```diff
@@ -214,24 +127,86 @@ static int LogDnsLogger(ThreadVars *tv, void *data, const Packet *p, Flow *f,
...
-        LogQuery(aft, timebuf, dstip, srcip, dp, sp, dns_tx, query);
-    }
+    char *buf;
+    LogDnsFileCtx *hlog = aft->dnslog_ctx;
 
-    if (dns_tx->rcode)
-        LogAnswer(aft, timebuf, srcip, dstip, sp, dp, dns_tx, NULL);
-    if (dns_tx->recursion_desired)
-        LogAnswer(aft, timebuf, srcip, dstip, sp, dp, dns_tx, NULL);
+    for (uint16_t i = 0; i < 0xffff; i++) {
```
inliniac: have you done any profiling of this code? It seems a bit cleaner than the json object sharing, but it would be interesting to get some performance metrics.

Wonder if we could pass around some normalized data in larger blobs too. Not sure yet how that could/should look like. Perhaps we can look at how for example the kernel/userspace xfer for conntrack is done.

rust/src/dns/dns.rs
inliniac: Return the length of what?

jasonish: The number of answers so we know how many times to loop. I suppose it's pointless tho, it can just look at the return code which can tell it stop like it does now.

inliniac: Yeah in this case I don't think we'd need it.

jasonish:
Closing, see: https://github.com/inliniac/suricata/pull/2542

PR #2528, opened 2017-01-30
stream-tcp: separate stream config parsing and init

regit:
This patch separate the stream config parsing and init so we can
read the config before init the application layer as the stream
reassembly depth value is needed when doing it.

Update of #2505 fixing problem on -k option.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/244
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/26

### DIFF
--- a/src/stream-tcp.c
+++ b/src/stream-tcp.c
@@ -303,18 +303,9 @@ static void StreamTcpSessionPoolCleanup(void *s)
     }
 }
 
-/** \brief          To initialize the stream global configuration data
- *
- *  \param  quiet   It tells the mode of operation, if it is TRUE nothing will
- *                  be get printed.
- */
-
-void StreamTcpInitConfig(char quiet)
-{
+void StreamTcpParseConfig(char quiet) {
     intmax_t value = 0;
     uint16_t rdrange = 10;
-
-    SCLogDebug("Initializing Stream");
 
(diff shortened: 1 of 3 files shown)


inliniac:
This breaks --disable-detection.

master shows:
[64687] 31/1/2017 -- 11:37:13 - (stream-tcp.c:578) <Config> (StreamTcpInitConfig) -- stream.reassembly.raw: disabled

this PR:
[64041] 31/1/2017 -- 11:33:24 - (stream-tcp.c:568) <Config> (StreamTcpParseConfig) -- stream.reassembly.raw: enabled

Beginning to think we need a different approach altogether, risk of more (subtle) breakage seems high.

src/stream-tcp.c
```diff
@@ -303,19 +303,10 @@ static void StreamTcpSessionPoolCleanup(void *s)
...
 
-/** \brief          To initialize the stream global configuration data
- *
- *  \param  quiet   It tells the mode of operation, if it is TRUE nothing will
- *                  be get printed.
- */
-
-void StreamTcpInitConfig(char quiet)
-{
+void StreamTcpParseConfig(char quiet) {
```
inliniac: style

PR #2530, opened 2017-01-31
Experiments/rust/v4.5

inliniac:
More experimental code, work in progress, messy, etc, etc.

Mostly been exploring how to integrate file extraction between rust parsers and Suricata/c. The file API is Suricata is not very clean and quite complex.


inliniac:
Replaced by #2533

PR #2535, opened 2017-02-01
Support HTTP CONNECT and STARTTLS for SMTP and FTP

thus:
Add support for HTTP CONNECT and STARTTLS for SMTP and FTP.

Set flow flag 'FLOW_CHANGE_PROTO' on STARTTLS / HTTP CONNECT. This flag triggers a reset of the flows application detection state, and trigger the protocol to be detected again.

Add "from_proto" field to TLS eve-log to indicate the original protocol.

HTTP CONNECT eve-log example:
```json
{
  "http": {
    "length": 0,
    "status": 200,
    "protocol": "HTTP/1.1",
    "http_method": "CONNECT",
    "http_user_agent": "curl/7.47.0",
    "url": "www.facebook.com:443",
    "hostname": "www.facebook.com"
  },
  "tx_id": 0,
  "proto": "TCP",
  "timestamp": "2017-01-31T08:48:17.102130+0100",
  "flow_id": 1956895568964224,
  "pcap_cnt": 11,
  "event_type": "http",
  "src_ip": "95.85.1.25",
  "src_port": 34440,
  "dest_ip": "146.185.132.124",
  "dest_port": 8888
}
{
  "tls": {
    "from_proto": "http",
    "notafter": "2018-01-25T12:00:00",
    "notbefore": "2016-12-09T00:00:00",
    "version": "TLS 1.2",
    "sni": "www.facebook.com",
    "fingerprint": "93:6f:91:2b:af:ad:21:6f:a5:15:25:6e:57:2c:dc:35:a1:45:1a:a5",
    "issuerdn": "C=US, O=DigiCert Inc, OU=www.digicert.com, CN=DigiCert SHA2 High Assurance Server CA",
    "subject": "C=US, ST=California, L=Menlo Park, O=Facebook, Inc., CN=*.facebook.com"
  },
  "proto": "TCP",
  "timestamp": "2017-01-31T08:48:17.373532+0100",
  "flow_id": 1956895568964224,
  "pcap_cnt": 17,
  "event_type": "tls",
  "src_ip": "95.85.1.25",
  "src_port": 34440,
  "dest_ip": "146.185.132.124",
  "dest_port": 8888
}

```

https://redmine.openinfosecfoundation.org/issues/806
https://redmine.openinfosecfoundation.org/issues/805

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/89
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/89


thus:
@regit

inliniac:
cc @jasonish

inliniac:
Would it make sense in the STARTTLS case to not flag the flow for a new protocol detection, but simply set the new protocol to ALPROTO_TLS? In the HTTP CONNECT case this won't work, but for SMTP/FTP/etc it really should be TLS.

src/app-layer-smtp.c
```diff
@@ -1890,10 +1889,8 @@ static int SMTPParserTest01(void)
         goto end;
     }
 
-    if (!(f.flags & FLOW_NOPAYLOAD_INSPECTION) ||
-        !(ssn.flags & STREAMTCP_FLAG_APP_LAYER_DISABLED) ||
-        !(((TcpSession *)f.protoctx)->server.flags & STREAMTCP_STREAM_FLAG_NOREASSEMBLY) ||
-        !(((TcpSession *)f.protoctx)->client.flags & STREAMTCP_STREAM_FLAG_NOREASSEMBLY)) {
+    if (!FlowChangeProto(&f)) {
```
regit: I don't see how this 2 can be "equivalent". The initial code is testing something that completely disappear and can not be replace functionally by FlowChangeProto.

src/app-layer-ftp.c
```diff
@@ -189,6 +189,8 @@ static int FTPParseRequestCommand(void *ftp_state, uint8_t *input,
     if (input_len >= 4) {
         if (SCMemcmpLowercase("port", input, 4) == 0) {
             fstate->command = FTP_COMMAND_PORT;
+        } else if (SCMemcmpLowercase("auth tls", input, 8) == 0) {
```
regit: we are just sure input_len >= 4 so we should not read till 8 bytes here.

src/app-layer-ftp.c
```diff
@@ -268,6 +270,14 @@ static int FTPParseResponse(Flow *f, void *ftp_state, AppLayerParserState *pstat
                             uint8_t *input, uint32_t input_len,
                             void *local_data)
 {
+    FtpState *state = (FtpState *)ftp_state;
+
+    if (state->command == FTP_COMMAND_AUTH_TLS) {
+        if (SCMemcmp("234 ", input, 4) == 0) {
```
regit: Boundary check here too.

regit:
@inliniac I like the idea of doing a complete discovery in the case of STARTTLS. It is not too costly and will allow to detect potential joker with strange server implementation.

src/app-layer-smtp.c
thus: The intention here is to check that the flag that triggers the protocol re-detection is set as it should. The checks it did initially is for code that is removed in this PR. Would it be better to just remove the test altogether?

inliniac:
Agreed but I still think we should set our expectation. STARTTLS should be followed by TLS, if it's not that should be an event.

src/app-layer-ftp.c
thus: What about using input_len as an argument to SCMemcmpLowercase instead? E.g.:
```c
} else if (SCMemcmpLowercase("auth tls", input, input_len) == 0) {
```
Or would that be stupid?

thus:
Then I'll add a decoder event if STARTTLS is not followed by TLS :)

src/app-layer-htp.c
```diff
@@ -2010,6 +2010,20 @@ static int HTPCallbackResponse(htp_tx_t *tx)
...
+    /* handle HTTP CONNECT */
+    if (tx->request_method_number == HTP_M_CONNECT) {
+        /* any 2XX status response implies that the connection will become
+           a tunnel immediately after this packet (RFC 7230, 3.3.3). */
+        if ((tx->response_status_number >= 200) &&
+                (tx->response_status_number < 300) &&
+                (hstate->transaction_cnt == 1)) {
+            FlowSetChangeProtoFlag(hstate->f);
+            tx->request_progress = HTP_REQUEST_COMPLETE;
+            tx->response_progress = HTP_REQUEST_COMPLETE;
```
thus: 'tx->response_progress' should be set to 'HTP_RESPONSE_COMPLETE' instead. This is the reason Travis CI failed.

inliniac:
Looking forward to the next iteration of this work, thanks Mats!

src/app-layer-smtp.c
thus: @regit, any comments to my comment or answer to my question? :)

src/app-layer-ftp.c
thus: @regit, what do you think?

regit: If input_len is bigger than sizeof("auth tls") then we will read after the limit. Not sure we want that.

inliniac:
On 09-02-17 09:13, Eric Leblond wrote:
> 
> 
> In src/app-layer-ftp.c <https://github.com/inliniac/suricata/pull/2535>:
> 
>> @@ -189,6 +189,8 @@ static int FTPParseRequestCommand(void *ftp_state, uint8_t *input,
>      if (input_len >= 4) {
>          if (SCMemcmpLowercase("port", input, 4) == 0) {
>              fstate->command = FTP_COMMAND_PORT;
> +        } else if (SCMemcmpLowercase("auth tls", input, 8) == 0) {
> 
> If input_len is bigger than sizeof("auth tls") then we will read after
> the limit. Not sure we want that.

"Not sure we want that" -> did you turn in to a Brit Eric?
Understatement of the year! :-P

src/app-layer-smtp.c
regit: OK, I have missed the point it is in unittests. So it looks correct for me considering that.

thus:
@inliniac, Eric is just trying to say my suggestion sucks in a nice way. You should try that sometime ;)

inliniac:
On 09-02-17 09:20, Mats Klepsland wrote:
> @inliniac <https://github.com/inliniac>, Eric is just trying to say my
> suggestion sucks in a nice way. You should try that sometime ;)

I need a reject button on comments :-P

thus:
Replaced by #2556

PR #2536, opened 2017-02-02
TLS logging for session resumption 1.1

rtkrruvinskiy:

was never seen.

Replaces https://github.com/inliniac/suricata/pull/2423


thus:
Thanks! I'll take a look at it as soon as I get the chance.

thus:
Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/90
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/90

src/app-layer-ssl.c
```diff
@@ -265,6 +265,9 @@ static int TLSDecodeHandshakeHello(SSLState *ssl_state, uint8_t *input,
 
     /* skip session id */
     uint8_t session_id_length = *(input++);
+    if (session_id_length != 0) {
+	ssl_state->flags |= SSL_AL_FLAG_SSL_CLIENT_SESSION_ID;
```
inliniac: style: indent should be 4 spaces

rtkrruvinskiy:
@inliniac Thanks for the review! I'll make those changes.

rtkrruvinskiy:
Replaced by https://github.com/inliniac/suricata/pull/2555

PR #2537, opened 2017-02-03
Experiments/rust/v4.9 - Rust/C/Lua data access

inliniac:
So this adds an idea for a generic way for C and Lua to access Rust managed data.

The idea is that the Rust side manages a 'store' that it writes data to at certain 'id's. The stores gives a way to retrieve this data by id.

In C:
```c
uint64_t offset;
if (r_getu64(state, 1, &offset) == 1) {
```
State here is the NFS state, and that needs a be generalized a bit more. '1' is the id.

In Lua, the mechanics are similar:
```lua
offset = NfsGetU64(1);
```

Overhead seems to be minimal.

Safe usage on Lua is more like this
```lua
    x, err = NfsGetU32(3);
    if x == nil then
        print ("NfsGetU32(3) failed with " .. err);
    else
        -- do something with x
    end
```

Relevants commits:
https://github.com/inliniac/suricata/commit/efafb93e7f6bd8c9880de3ea5152ccee4b20ee64
https://github.com/inliniac/suricata/commit/8007a33164635fedc0cf1a11f0dfa70313615c5a
https://github.com/inliniac/suricata/commit/d2122de354e9935289778bc93bce46641f0135b7

Next step would be for the Rust code to register a map between names and id's that Lua could use, e.g.
```lua
offset = NfsGetU64("nfs.chunk_offset");
```

Stores could added to where they are needed, most likely the 'state' and/or the 'tx'. This would be a choice for the Rust implementation. The C api needs to be made flexible enough to deal with this.

Perhaps simply something like:
```c
r_getdata_state(AppProto alproto, void *state, uint8_t **data, uint32_t len);
r_getdata_tx(AppProto alproto, void *state, uint8_t **data, uint32_t len);
```


jasonish:
> offset = NfsGetU64("nfs.chunk_offset");

I like the idea of identifers like this. But I'm not sure how generic we can make it. Look at DNS where each response and request has multiple buffers.

In a call like:

if (r_getu64(state, 1, &offset) == 1) {

assuming the "1" is what we are after, say dns.response.rrname, we also need another field for which rrname we are after since its a variable length list. DNP3 is another where all responses are a list where you'd need to specify the what and an index.

Any thoughts on how to handle that?

inliniac:
Good point. One possible way could be to go recursive. Add a type that is really itself another store. Would add overhead of course. It would also complicate the getters logic.

inliniac:
I guess it depends on what needs to be stored. If you need a vector<u8> per authority, perhaps a hashmap where the index is the authority record number and the data the buffer.
```c
r_getdata_from_map(tx, DNS_AUTH_BUFFER, 4, &data, &len);
```
4 here would mean the 4th record. I guess the logic could simply be that you call until you get a NOT FOUND or indeed we need to also get a size var.

inliniac:
Recurses store experiments in #2538

PR #2539, opened 2017-02-05
Bug 1780 v2

regit:
Rework of #2405 that should copy the complete VLAN tag. Logic has been changed as I've discovered that we always have the ring buffer cell available when calling the write function. Which means we can use the vlan cell header to get the info at write time.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/246
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/28

@gozzy can you test this ? I've currently no setup I can use to do it.

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -980,6 +980,14 @@ static inline int AFPParsePacketV3(AFPThreadVars *ptv, struct tpacket_block_desc
     p->livedev = ptv->livedev;
     p->datalink = ptv->datalink;
 
+    /* get vlan id from header */
+    if ((!(ptv->flags & AFP_VLAN_DISABLED)) &&
+            (ppd->tp_status & TP_STATUS_VLAN_VALID || ppd->hv1.tp_vlan_tci)) {
+        p->vlan_id[0] = ppd->hv1.tp_vlan_tci & 0x0fff;
+        p->vlan_idx = 1;
+        p->vlanh[0] = NULL;
+    }
+
     if (ptv->flags & AFP_ZERO_COPY) {
         if (PacketSetData(p, (unsigned char*)ppd + ppd->tp_mac, ppd->tp_snaplen) == -1) {
             TmqhOutputPacketpool(ptv->tv, p);
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -662,10 +662,23 @@ int AFPRead(AFPThreadVars *ptv)
     SCReturnInt(AFP_READ_OK);
 }
 
-TmEcode AFPWritePacket(Packet *p)
+/**
+ * \brief AF packet write function.
+ *
+ * This function has to be called before the memory
+ * related to Packet in ring buffer is released.
+ *
+ * \param pointer to Packet
+ * \param version of capture: TPACKET_V2 or TPACKET_V3
+ * \retval TM_ECODE_FAILED on failure and TM_ECODE_OK on success
+ *
+ */
+static TmEcode AFPWritePacket(Packet *p, int version)
 {
     struct sockaddr_ll socket_address;
     int socket;
+    uint8_t *pstart;
+    size_t plen;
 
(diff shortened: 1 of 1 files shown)


gozzy:
@regit, sure, I'll take a look at this tomorrow.

src/source-af-packet.c
```diff
@@ -691,7 +704,45 @@ TmEcode AFPWritePacket(Packet *p)
...
         SCMutexLock(&p->afp_v.peer->sock_protect);
     socket = SC_ATOMIC_GET(p->afp_v.peer->socket);
-    if (sendto(socket, GET_PKT_DATA(p), GET_PKT_LEN(p), 0,
+
+    if (version == TPACKET_V2) {
+        union thdr h;
+        h.raw = p->afp_v.relptr;
+        /* Copy VLAN header from ring memory */
+        if (h.h2->tp_status & TP_STATUS_VLAN_VALID || h.h2->tp_vlan_tci) {
+            pstart = GET_PKT_DATA(p) - VLAN_HEADER_LEN;
```
inliniac: is this valid if the 'reserve' call during setup failed?

src/source-af-packet.c
```diff
@@ -691,7 +704,45 @@ TmEcode AFPWritePacket(Packet *p)
     if (p->afp_v.peer->flags & AFP_SOCK_PROTECT)
         SCMutexLock(&p->afp_v.peer->sock_protect);
     socket = SC_ATOMIC_GET(p->afp_v.peer->socket);
-    if (sendto(socket, GET_PKT_DATA(p), GET_PKT_LEN(p), 0,
+
+    if (version == TPACKET_V2) {
+        union thdr h;
+        h.raw = p->afp_v.relptr;
+        /* Copy VLAN header from ring memory */
+        if (h.h2->tp_status & TP_STATUS_VLAN_VALID || h.h2->tp_vlan_tci) {
```
inliniac: should this be && instead of ||?

regit: No, I had the same thought when looking at it today but the logic is to take it if kernel exports the flag or it does not export it (and in that case we use the value).

src/source-af-packet.c
regit: Good catch, making reserve call an error and exiting if it is the case.

inliniac:
On 05-02-17 19:32, Eric Leblond wrote:
> 
> 
> In src/source-af-packet.c <https://github.com/inliniac/suricata/pull/2539>:
> 
>> @@ -691,7 +704,45 @@ TmEcode AFPWritePacket(Packet *p)
>      if (p->afp_v.peer->flags & AFP_SOCK_PROTECT)
>          SCMutexLock(&p->afp_v.peer->sock_protect);
>      socket = SC_ATOMIC_GET(p->afp_v.peer->socket);
> -    if (sendto(socket, GET_PKT_DATA(p), GET_PKT_LEN(p), 0,
> +
> +    if (version == TPACKET_V2) {
> +        union thdr h;
> +        h.raw = p->afp_v.relptr;
> +        /* Copy VLAN header from ring memory */
> +        if (h.h2->tp_status & TP_STATUS_VLAN_VALID || h.h2->tp_vlan_tci) {
> 
> No, I had the same thought when looking at it today but the logic is to
> take it if kernel exports the flag or it does not export it (and in that
> case we use the value).

I would appreciate a bit more explanation, I still don't follow this logic.

regit:
@inliniac The flag was not introduced in af_pacetk since the beginning. It has been added in 2011 to fix the problem of VID 0 that was a valid tag but could not be distinguished from no-vlan. So in the test, we check if the flag has been set. This is the case of recent kernel. For older kernel, we do our best by using tp_vlan_tci if not null. I agree that we may not see a lot of pre June 2011 kernel in the wild in 2017 so we could switch to a && logic.

inliniac:
Replaced by #2540

PR #2540, opened 2017-02-05
Bug 1780 v3

regit:
Update of #2539 fixing reverse call arror handling.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/247
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/29

Redmine issue: https://redmine.openinfosecfoundation.org/issues/1780

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2011-2016 Open Information Security Foundation
+/* Copyright (C) 2011-2017 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -662,10 +662,23 @@ int AFPRead(AFPThreadVars *ptv)
     SCReturnInt(AFP_READ_OK);
 }
 
-TmEcode AFPWritePacket(Packet *p)
+/**
+ * \brief AF packet write function.
+ *
+ * This function has to be called before the memory
+ * related to Packet in ring buffer is released.
+ *
+ * \param pointer to Packet
+ * \param version of capture: TPACKET_V2 or TPACKET_V3
+ * \retval TM_ECODE_FAILED on failure and TM_ECODE_OK on success
+ *
+ */
+static TmEcode AFPWritePacket(Packet *p, int version)
 {
     struct sockaddr_ll socket_address;
     int socket;
+    uint8_t *pstart;
+    size_t plen;
 
     if (p->afp_v.copy_mode == AFP_COPY_MODE_IPS) {
         if (PACKET_TEST_ACTION(p, ACTION_DROP)) {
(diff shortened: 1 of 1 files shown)


src/source-af-packet.c
```diff
@@ -1686,6 +1745,21 @@ static int AFPSetupRing(AFPThreadVars *ptv, char *devname)
     }
 #endif
 
+    /* Let's reserve head room so we can add the VLAN header in IPS
+     * or TAP mode before write the packet */
+    if (ptv->copy_mode != AFP_COPY_MODE_NONE) {
+        /* Only one vlan is extracted from AFP header so
+         * one VLAN header length is enough. */
+        int reserve = VLAN_HEADER_LEN;
+        if (setsockopt(ptv->socket, SOL_PACKET, PACKET_RESERVE, (void *) &reserve,
```
inliniac: Do you know when this feature was introduced? If this breaks existing AF_PACKET IPS setups we can't take it into 3.2.1. AFAICS this reserve is also done if no VLANS are on the network.

src/source-af-packet.c
```diff
@@ -691,7 +704,45 @@ TmEcode AFPWritePacket(Packet *p)
     if (p->afp_v.peer->flags & AFP_SOCK_PROTECT)
         SCMutexLock(&p->afp_v.peer->sock_protect);
     socket = SC_ATOMIC_GET(p->afp_v.peer->socket);
-    if (sendto(socket, GET_PKT_DATA(p), GET_PKT_LEN(p), 0,
+
+    if (version == TPACKET_V2) {
+        union thdr h;
+        h.raw = p->afp_v.relptr;
+        /* Copy VLAN header from ring memory */
+        if (h.h2->tp_status & TP_STATUS_VLAN_VALID || h.h2->tp_vlan_tci) {
```
inliniac: it would be good to have the explanation here for why this strange looking check makes sense

src/source-af-packet.c
regit: So `git log -G PACKET_RESERVE net/packet/` gives me:
```
Author: Patrick McHardy <kaber@trash.net>
Date:   Fri Jul 18 18:05:19 2008 -0700

    packet: add PACKET_RESERVE sockopt
```
So I think we are good as it is really long ago.

src/source-af-packet.c
regit: I propose:
```
        /* Copy VLAN header from ring memory. For post june 2011 kernel we test
         * the flag. It is not defined for older kernel so we do best effort and
         * test for non null value of the TCI header */
        if (h.h2->tp_status & TP_STATUS_VLAN_VALID || h.h2->tp_vlan_tci) {
```

src/source-af-packet.c
inliniac: What kernel version was this introduced in?

regit: `git tag --contains 8913336a7e8d56e984109a3137d6c0e3362596a4` says it is 2.6.27

src/source-af-packet.c
inliniac: Good by me. Maybe make it test for non zero value. In my mind 'null' refers to a pointer value NULL.

src/source-af-packet.c
inliniac: Great, thanks. No objections to a fatal error then.

PR #2541, opened 2017-02-06
Bug 1780 v4

regit:
Rework of #2540 adding comments.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/248
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/30

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2011-2016 Open Information Security Foundation
+/* Copyright (C) 2011-2017 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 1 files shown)


inliniac:
@gozzy can you give this one a test?

gozzy:
@inliniac we're testing v3 patch right now. Seems to work with and without VLANs, but we have not completed checks with VLAN priority and CFI.

gozzy:
Here are the results for v3 patch (v4 is technically the same):
- non-tagged traffic: OK
- tagged traffic without 'priority' and 'CFI' attributes: OK
- tagged traffic with 'priority' and 'CFI' via tcpreplay: OK. We could not test it with IXIA because it processes VLAN parameters in a strange manner (firmware bug?)... However, ARP packets produced by IXIA pass through suricata correctly.

inliniac:
Did you test AF_PACKET v2, v3 or both?

inliniac:
I mean tpacket v2 or v3 of course.

gozzy:
It seemed that we have checked v2.

Tried to enable v3 by specifying "tpacket-v3: yes" in yaml-file now and saw the following after running tcpreplay: packets on 'outgoing' interface contain 2 VLAN tags: the first one is arbitrary (?), the second one is correct. V2 mode doesn't behave like that. It can be easily checked with tcpreplay, no special hardware is required.

inliniac:
Thanks for testing @gozzy. So it seems tpacketv2 works well with this patch. V3 is broken. Wonder where that random data comes from. Any thoughts @regit ?

regit:
@gozzy can you remove the last patch of the branch ? (git reset --hard HEAD^) and see how it goes

gozzy:
@regit ok, I'll give it a try now.

gozzy:
This doesn't seem to affect anything, we still see double tagged packets in tcpdump after tcpreplaying tagged sample.

regit:
@gozzy can you get a try to https://github.com/regit/suricata/tree/bug-1780-v5 ? I've simply removed VLAN fetching from v3..

gozzy:
@regit, works like a charm, no double tags and all fields have expected values.

regit:
ok thanks a lot! I'm going to check the code and resubmit.

PR #2542, opened 2017-02-06
Preview: DNS app-layer in Rust - v8

jasonish:
Previous PR: #2526 

Major changes since last PR:
- The C transaction has been removed making the C app-layer much more minimal.
- Lua should now work.
- Memcap handling.

Notes:
- All C functions called by Rust are registered as function pointers so "cargo test" can continue to work (https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Rust_Notes#Allow-cargo-test-To-Work)
- I rewrote the buffering loop for TCP to make it a little more straight forward - this could probably move to Rust.
- According to internal profiling and wall time, Rust is still slower:
```
dns                     IPv4       6            63              726           9831          2085        131.4k  0.00  
dns                     IPv4      17       1436213              753         444141          3786          5.4b  100.00
real	0m8.446s
user	0m17.377s
sys	0m0.970s
```
- Git master:
```
dns                     IPv4       6            63              717           7275          1678        105.7k  0.00  
dns                     IPv4      17       1436220              441         157356          1842          2.6b  99.99 
real	0m6.355s
user	0m13.200s
sys	0m0.170s
```
- Attempts to optimize the hot spots have not proved useful, including more C like "unsafe" code. This may just be the cost of Rust for this use case.
- The Rust code, particularly dns.rs is a bit scattered. Some organizing and grouping of functions is needed.
- Will fail to build if jansson is not available at this time, otherwise Travis tests are passing: https://travis-ci.org/jasonish/suricata/builds/198110887

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,9 @@ script:
   - |
     if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
         export CFLAGS="${CFLAGS} ${EXTRA_CFLAGS}"
-        ./configure --enable-nfqueue --enable-unittests --enable-hiredis ${ARGS}
+        export LDFLAGS="-lrt"
+        ./configure --enable-nfqueue --enable-unittests \
+            --enable-hiredis ${ARGS}
     fi
   - |
     if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
@@ -122,6 +124,7 @@ before_install:
             sudo apt-get install -y libjansson-dev
         fi
 
+        curl https://sh.rustup.rs -sSf | sh -s -- -y
     fi
   - |
     if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
@@ -129,5 +132,6 @@ before_install:
         brew install pkg-config libmagic libyaml nss nspr jansson libnet lua \
             pcre hiredis
     fi
+  - export PATH=$PATH:$HOME/.cargo/bin
   - ./qa/travis-libhtp.sh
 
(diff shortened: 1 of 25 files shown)


inliniac:
To what extend do you consider this an 'apples to apples' comparison currently? The difference 2.6b vs 5.4b is quite big.

jasonish:
Very very close. I think the only thing that is missing is the log filtering, and these tests are run with logging off.  I think its a pretty straight forward port to Rust, not going overboard with Rustisms, but at the same time not being as C-like as possible (which doesn't help performance anyways).

I've also not yet considered any sort of generic stores. The use of HashMaps are a measurable performance hit, plus it starts to diverge from "apples" to "apples" at that point.

The profiling numbers above are with Rust stable release.  If I move to Rust nightly, which allows us to use jemalloc with libs (note: this is the default for a normal rust app, not for a rust lib) the performance is better:
```
dns                     IPv4       6            63              705           4209          1322         83.3k  0.00  
dns                     IPv4      17       1436214              600         603108          2365          3.4b  100.00
real	0m6.159s
user	0m12.847s
sys	0m0.390s
```

inliniac:
Ah yes, so that is a big difference. As far as you know, will this nightly/jemalloc behavior become the default any time soon?

This leads to a related point: we probably have to assume older versions of rustc will continue to be used by many ppl as distro's lock the version. So for performance and language features this may have a big impact.

jasonish:
> Ah yes, so that is a big difference. As far as you know, will this nightly/jemalloc behavior become the default any time soon?

Not that I can see. Looks like this is an issue for Firefox as well though:
https://github.com/rust-lang/rust/issues/33082

> This leads to a related point: we probably have to assume older versions of rustc will continue to be used by many ppl as distro's lock the version. So for performance and language features this may have a big impact.

Yes, a valid concern when depending on a relatively young language, which is probably why rustup is the recommended way of installing Rust.

inliniac:
Replaced by #2686

PR #2543, opened 2017-02-07
af-packet: add VLAN header when needed in IPS mode

regit:
Update of #2541 fixing tpacket_v3 consistency.

When packet is coming from a real ethernet card, the kernel is
stripping the vlan header and delivering a modified packet so
we need to insert the VLAN header back before sending the packet
on the wire.

To do so, we pass an option to the raw socket to add a reserve
before the packet data. It will get Suricata some head room to
to move the ethernet addresses before there actual place and
and insert the VLAN header in the correct place.

We get VLAN info from the ring buffer as the call of AFPWrite is
always done in the release function so we still have access to the
memory.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/249
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/31

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2011-2016 Open Information Security Foundation
+/* Copyright (C) 2011-2017 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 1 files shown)


inliniac:
CentOS6:
<pre>
source-af-packet.c: In function ‘AFPReleasePacketV3’:
source-af-packet.c:773: error: ‘TPACKET_V3’ undeclared (first use in this function)
source-af-packet.c:773: error: (Each undeclared identifier is reported only once
source-af-packet.c:773: error: for each function it appears in.)
</pre>

regit:
grmbllll

PR #2545, opened 2017-02-08
dns: fix flood handling to avoid evasions - v1

jasonish:
If a flow hit a DNS flood state with valid requests, new requests
that would match a block rule would be passed, or if not in IPS
mode, not logged/alerted.

To handle this the flood logic will not give up parsing anymore,
but instead the window is used to control how many outstanding
requests can be in memory, with a flood alert raised when number
of outstanding requests hits the limit, and the alert state cleared
when all DNS requests have been marked as replied (or reply lost).

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/85
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/437

### DIFF
--- a/src/app-layer-dns-common.c
+++ b/src/app-layer-dns-common.c
@@ -31,12 +31,7 @@
 #include "util-memcmp.h"
 #include "util-atomic.h"
 
-typedef struct DNSConfig_ {
-    uint32_t request_flood;
-    uint32_t state_memcap;  /**< memcap in bytes per state */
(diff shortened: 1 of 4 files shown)


inliniac:
I've privately shared a pcap with you that makes this PR cry.

inliniac:
Needs more thought.

PR #1827, opened 2016-02-01
1.  The enip-feature-1495-v4 branch contains support for the ENIP/CIP

kwong-solana:
1.  The enip-feature-1495-v4 branch contains support for the ENIP/CIP
Industrial protocol
2. This is an app layer implementation which uses the "enip" protocol
and "cip_service" and "enip_command" keywords
3.  Rebased master to clean out commits done on master and then 
merged to latest code
4.  Addresses failed PR's https://github.com/inliniac/suricata/pull/1810 
and https://github.com/inliniac/suricata/pull/1819

PR #2546, opened 2017-02-08
mpm-ac: fix integer overflow on allocation

satta:
The size of a memory buffer to be reallocated during state creation in the Aho-Corasick implementation was kept in a signed `int` instead of a `size_t`, leading to an overflow when large lists of long and diverse patterns cause the amount of AC states to blow up (>2GB). This case can be triggered for reproduction using a script like https://gist.github.com/satta/e47f7a6f0e44c9898804121e4d8c5f2e.

Addresses Redmine issues #1827 (https://redmine.openinfosecfoundation.org/issues/1827) and #1843 (https://redmine.openinfosecfoundation.org/issues/1843).

- A new inline function `SCACCheckSafeSizetMult()` to check whether a `size_t` multiplication overflows, `exit()` ing in this case.
- Type change of the `size` variable in `SCACReallocState()` to `size_t`.

I ran the Dockerized prscript, builds completed successfully:

- https://steinbiss.name/suri/fix-integer-overflow-1827-1843-v1/features.html

### DIFF
--- a/src/util-mpm-ac.c
+++ b/src/util-mpm-ac.c
@@ -122,6 +122,28 @@ static void SCACGetConfig()
 
 /**
  * \internal
+ * \brief Check if size_t multiplication would overflow and perform operation
+ *        if safe. In case of an overflow we exit().
+ *
+ * \param a First size_t value to multiplicate.
+ * \param b Second size_t value to multiplicate.
+ *
+ * \retval The product of a and b, guaranteed to not overflow.
+ */
+static inline size_t SCACCheckSafeSizetMult(size_t a, size_t b)
+{
+    /* check for safety of multiplication operation */
+    if (b > 0 && a > SIZE_MAX / b) {
+        SCLogError(SC_ERR_MEM_ALLOC, "%"PRIuMAX" * %"PRIuMAX" > %"
+                   PRIuMAX" would overflow size_t calculating buffer size",
+                   (uintmax_t) a, (uintmax_t) b, (uintmax_t) SIZE_MAX);
+        exit(EXIT_FAILURE);
+    }
+    return a * b;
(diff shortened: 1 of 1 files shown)


src/util-mpm-ac.c
```diff
@@ -131,10 +153,10 @@ static void SCACGetConfig()
 static inline int SCACReallocState(SCACCtx *ctx, uint32_t cnt)
 {
     void *ptmp;
-    int size = 0;
+    size_t size = 0;
 
     /* reallocate space in the goto table to include a new state */
-    size = cnt * ctx->single_state_size;
+    size = SCACCheckSafeSizetMult((size_t) cnt, (size_t) ctx->single_state_size);
     ptmp = SCRealloc(ctx->goto_table, size);
```
inliniac: Interestingly, scan-build-3.9 now warns that size could be 0:
<pre>
util-mpm-ac.c:160:12: warning: Call to 'realloc' has an allocation size of 0 bytes
    ptmp = SCRealloc(ctx->goto_table, size);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./util-mem.h:193:14: note: expanded from macro 'SCRealloc'
    ptrmem = realloc((x), (a)); \
             ^~~~~~~~~~~~~~~~~
1 warning generated.
</pre>
I guess a simple size check would fix it. It is out goal to get to 0 issues in scan-build (and other checkers), even if they are sometimes far-fetched or even wrong. So we'll need a fix.

satta: Sure, I'll look into it. I can totally understand the goal of static checker cleanness -- we have also used scan-build (as a successor of splint) in previous projects.

PR #2547, opened 2017-02-08
alert: silence compiler type warning

satta:
The `ts_ecr` and `ts_val` struct fields are integer types (being timestamps), not pointers. This leads GCC 6.3.0 to complain about comparisons to `NULL`:

```
alert-prelude.c: In function ‘PacketToData’:
alert-prelude.c:515:35: warning: comparison between pointer and integer
             if (p->tcpvars.ts_val != NULL) {
                                   ^~
alert-prelude.c:518:35: warning: comparison between pointer and integer
             if (p->tcpvars.ts_ecr != NULL) {
                                   ^~
```
This PR adresses this by comparing these values to 0.

I ran the Dockerized prscript, builds completed successfully:
- https://steinbiss.name/suri/silence-type-warning-v1/features.html


inliniac:
Merged through #2550, thanks Sascha!

PR #2551, opened 2017-02-13
dns flood handling  - v2

jasonish:
The most recent commit is the under review it. It opportunistically increments the inspect_id as it comes across NULL transactions that will never be non-null. This is not specific to DNS, but helps DNS when logging is enabled. See note below for when logging is not enabled.

Note: There is an issue where detection is enabled (rules or no rules) but there are no loggers where DNSGetAlstateProgress never gets called in the TOCLIENT direction, this leads to a build of transactions as they are never free'd by the app-layer, even though they are marked lost to make them "complete". This is probably because there is no traffic in the TOCLIENT direction.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/87
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/439

### DIFF
--- a/src/app-layer-dns-common.c
+++ b/src/app-layer-dns-common.c
@@ -197,7 +197,8 @@ void *DNSGetTx(void *alstate, uint64_t tx_id)
 
     /* no luck with the fast tracks, do the full list walk */
     TAILQ_FOREACH(tx, &dns_state->tx_list, next) {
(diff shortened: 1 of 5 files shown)


inliniac:
Is there anything in here worth salvaging? I think master fixes the inspect_id thing already.

jasonish:
No, I don't think so. Plus I think the Rust approach of not correlating responses with requests is the better approach in the long run which makes this a non-issue with DNS.

Closing.

PR #2553, opened 2017-02-14
logging: support file rotation

thus:
Add support for log file rotation, without the need for external tools like logrotate.

The pull request basically adds three new features for all the loggers. It allows the user to use date modifiers in the log filename in suricata.yaml, e.g.:

```yaml
outputs:
  - eve-log:
      filename: eve-%Y-%m-%d-%H:%M:%S.json
```

It also adds log file rotation based on either minute/hour/day or a relative value like X seconds/minutes/hours/days/weeks, e.g.:

```yaml
outputs:
  - eve-log:
      filename: eve-%Y-%m-%d-%H:%M:%S.json
      rotate-interval: minute
```

The example above rotates the eve-log every minute. This could be changed to 'rotate-interval: 120s' to rotate the file every 120 seconds instead.

The last feature is to create new directories when needed. This enables us to use date modifiers to create directory structures (like often done by tools like rsyslog), e.g.:

```yaml
outputs:
  - eve-log:
      filename: /var/log/suricata/%Y/%m/%d/eve-%H.json
      rotate-interval: hour
```

The example above would create a new folder per day and rotate the eve-log every hour.

https://redmine.openinfosecfoundation.org/issues/1323
(it rotates on time and not size, so I'm not sure it  entirely solves this issue)

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/95
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/95

### DIFF
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -181,6 +181,21 @@ YAML::
 To reduce verbosity the output can be filtered by supplying the record types
 to be logged under ``custom``.
 
+Date modifiers in filename
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to use date modifiers in the eve-log filename.
+
+::
+
+   outputs:
+     - eve-log:
+         filename: eve-%s.json
+
+The example above adds epoch time to the filename. All the date modifiers from the
+C library should be supported. See the man page for ``strftime`` for all supported
+modifiers.
+
 Rotate log file
 ~~~~~~~~~~~~~~~
 
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -181,6 +181,35 @@ YAML::
 To reduce verbosity the output can be filtered by supplying the record types
 to be logged under ``custom``.
 
+Rotate log file
+~~~~~~~~~~~~~~~
+
+Eve-log can be configured to rotate based on time.
+
+::
+
+  outputs:
+    - eve-log:
+        filename: eve-%Y-%m-%d-%H:%M.json
+        rotate-interval: minute
+
+The example above creates a new log file each minute, where the filename contains
+a timestamp. Other supported ``rotate-interval`` values are ``hour`` and ``day``.
+
+In addition to this, it is also possible to specify the ``rotate-interval`` as a
+relative value. One example is to rotate the log file each X seconds.
+
+::
+
+  outputs:
+    - eve-log:
+        filename: eve-%Y-%m-%d-%H:%M:%S.json
+        rotate-interval: 30s
+
+The example above rotates eve-log each 30 seconds. This could be replaced with
+``30m`` to rotate every 30 minutes, ``30h`` to rotate every 30 hours, ``30d``
+to rotate every 30 days, or ``30w`` to rotate every 30 weeks.
+
 Multiple Logger Instances
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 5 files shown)


inliniac:
cc: @jasonish

jasonish:
This would still require external tools to actually delete the files right?  I think its worth mentioning because we also document logrotate usage: http://suricata.readthedocs.io/en/latest/configuration/log-rotation.html. Putting dates in the filenames will likely require a custom cleanup script.

Related issue:
https://redmine.openinfosecfoundation.org/issues/1978

thus:
You are right. It will not handle cleanup.

We could add something like:
```yaml
outputs:
  - eve-log:
      filename: /var/log/suricata/%Y/%m/%d/eve-%H.json
      rotate-interval: hour
      rotate-files: 50
```

And make each logger (if 'rotate-files' is specified) delete log files when it reach the specified amount of files. Would that be desirable?

jasonish:
How easy/hard/reliable would it be to do that for all variations of filenames that could be specified by the user?

By the way, we've had requests for unified2 like filename handling for eve, and I'm happy to see that this PR addresses this, I can do something like:
```
filename: eve.%s
rotate-interval: hour
```

to get similar behaviour.  It doesn't do limit size rotation, but I think thats fine.

thus:
I haven't given this much thought yet, but my immidiate thought is to track the files per logger when Suricata is running (simple TAILQ, or something similar). On Suricata init the loggers could try to detect the already written files based on the pattern in the file path. It should be doable, but it requires some more thinking :)

jasonish:
Yeah, its the being able to resume after a restart that is key. IMO, not point in doing it at all if it can't pick up after a restart.

But then again, I wonder if its even worth it.  If you are moving from the default logging to something like this then I assume you have a reason.  Custom cleanup might just be part of the users responsibility then? After all, its not Suricata's responsibility at all right now, it just happens that with the defaults, logrotate can easily be configured to handle it for us.

jasonish:
@thus I think its OK to not worry about cleanup. But I think it deserves a mention in the docs.

thus:
I'm just throwing some thoughts up in the air.

What about only supporting a subset of the date modifiers available and using glob to get the files from the previous run based on the filename? Then we could just replace the date modifiers with stuff glob understands like wildcards, etc.

All the modifiers below (except epoch time) is fixed length and should therefore be simple to replace with something to feed to glob. I have not added modifiers that are either locale controlled and modifiers that contain whitespace.

```
%C - the century number (2-digit integer)
%d - the day of the month (01 to 31)
%F - equivalent to %Y-%m-%d
%G - ISO 8601 week-based year (4-digit integer)
%g - like %G, but without century (2-digit integer)
%H - the hour (00 to 23)
%I - the hour (01 to 12)
%j - the day of the year (001 to 366)
%m - the month (01 to 12)
%M - the minute (00 to 59)
%p - "AM" or "PM"
%P - "am" or "pm"
%R - equivalent to %H:%M
%s - epoch time
%S - the second (00 to 60)
%T - equivalent to %H:%M:%S
%u - the day of the week (monday=1)
%U - the week number from first sunday in the year (00 to 53)
%V - ISO 8601 week number (01 to 53)
%w - the day of the week (sunday=1)
%W - the week number from first monday in the year (00 to 53)
%y - the year, short form (2-digit integer)
%Y - the year with century (4-digit integer)
```

Another possibility would be to write the log files to a file when exiting Suricata, and read it when initializing the logger when starting Suricata. However, this would not work if unexpected stuff happens like if Suricata is killed with SIGKILL/SIGSTOP or if Suricata segfaults.

thus:
Or we could just leave cleaning up to the user. Then we won't accidentally remove the wrong files at least :)

inliniac:
I think in the case of a complicated multi-directory scheme it's fine to assume the user knows what he/she's doing. So we can leave the cleanup to the user.

thus:
Replaced by #2565

PR #2554, opened 2017-02-14
rfc: json builder as an alternative to jansson - v5

jasonish:
This is an implementation of a JSON string builder to build JSON strings with the idea of being more efficient than Jansson as our requirements are not too complex.

For a given pcap, with catch-all rules and logging payload, printable payload and packet, the run time is typically 1.5x faster with this than using libjansson.

Only alert, dns, http and netflow have been been migrated. Jansson has been completely disabled for the purpose of development to make sure not existing dependency has crept in, which is also the reason for some #if 0's still lingering.

Prscript:
Build successful for jasonish-pcap
Build failure for jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/440

Note: Prscript fails due to use of strncat.  I used strncat as it performed measurably better than strlcat, but I'll revisit that.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -695,68 +695,68 @@
         exit 1
     fi
 
-  # libjansson
-    enable_jansson="no"
-    AC_ARG_WITH(libjansson_includes,
-            [  --with-libjansson-includes=DIR  libjansson include directory],
-            [with_libjansson_includes="$withval"],[with_libjansson_includes=no])
-    AC_ARG_WITH(libjansson_libraries,
-            [  --with-libjansson-libraries=DIR    libjansson library directory],
-            [with_libjansson_libraries="$withval"],[with_libjansson_libraries="no"])
+  dnl # libjansson
+  dnl   enable_jansson="no"
+  dnl   AC_ARG_WITH(libjansson_includes,
+  dnl           [  --with-libjansson-includes=DIR  libjansson include directory],
+  dnl           [with_libjansson_includes="$withval"],[with_libjansson_includes=no])
+  dnl   AC_ARG_WITH(libjansson_libraries,
+  dnl           [  --with-libjansson-libraries=DIR    libjansson library directory],
+  dnl           [with_libjansson_libraries="$withval"],[with_libjansson_libraries="no"])
 
-    if test "$with_libjansson_includes" != "no"; then
-        CPPFLAGS="${CPPFLAGS} -I${with_libjansson_includes}"
-    fi
+  dnl   if test "$with_libjansson_includes" != "no"; then
+  dnl       CPPFLAGS="${CPPFLAGS} -I${with_libjansson_includes}"
+  dnl   fi
 
-    enable_jansson="no"
-    enable_unixsocket="no"
+  dnl   enable_jansson="no"
+  dnl   enable_unixsocket="no"
 
-    AC_ARG_ENABLE(unix-socket,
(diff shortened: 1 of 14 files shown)


inliniac:
wonder if it makes sense to not use string functions like strncat at all on the internal side. Might get some additional speed by not having to look for the end of the string/buffer all the time.

jasonish:
So after this I moved back to strlcat and didn't see a performance difference.  But yeah, some time is spent in strlen.  I did have a version that tracked its own length, but I did update everything to use non str functions.  My guess is it will be marginally faster, but I also want to try something like SCJsonSetBase64(key, raw_data) so base64 data can be written directly to the buffer.

inliniac:
Since this change requires updating of all json users, it would be a create opportunity to abstract things a bit too? If we ever want to support bson or similar, we'd have to add a new api. Maybe can have a simple macro wrapping layer:

Something like, instead of:
SCJsonSetString(js, "type", "query");
do something like:
OutputSetString(TYPE_JSON, js, "type", "query");

Initially it'd just be calling the json calls directly, but later it would allow us to add different outputs like bson?

Needs some thought on how to design the API of course.

jasonish:
Maybe.. My only thought is we would want to implement another one right
away as well.  Just to make sure we're not creating a pattern that is not
applicable to other types.  This is very much built around the idea of a
string builder, I don't think it would be possible to make a similar API
that did something like ProtoBuf/Capnproto/Flatbuf, without one or the
other being butchered. But it does look like this pattern could work for
bson and msgpack.

When I think of an abstraction layer, I think of output-json-* being passed
an object that is in this abstraction layer and then they output JSON in
the most efficient way.  Likewise if we had BSON, or this same abstraction
being passed into Rust or Lua. But that might not be enough of a savings
over what we already have.

inliniac:
Complication: #2333 would be heavily affected by this.

jasonish:
Yes, it would be as this doesn't really write to a format thats usable by others.

I have been putting some thought into an intermediate log format and I don't think it will be trivial, especially if we have outputs that are flat file like, as well as structured.

I think what we would end up with is an intermediate object that would look something like a json_t. It would support key/values where values can be strings, numbers or lists. And lists can contain lists, objects, numbers and strings. It starts to mirror json_t pretty close.


The benefit would be that application specific details would only be translated once from internal data structures into this intermediate format. The intermediate format would then be passed to all the log renders, json, prelude, fast, etc. So the actual log output would only need to know how to iterate over this intermediate structure. If done right it shouldn't have any more overhead than the current JSON support, since its essentially following the same model - fill in intermediate data struct, serialize to text. Just not sure if keeping libjansson around would make sense tho. 

In a message queue scenario (https://redmine.openinfosecfoundation.org/issues/352), this is the object that would be put on the queue.

Would also make creating custom outputs as dyanmically loaded modules much easier I think, as no knowledge of the internal transaction objects is required.

Do you think its worth thinking about further? But for now, I think this PR should be closed.

inliniac:
Lets table that for now, although I am interested in the perf benefit you reported. Perhaps a ugly wrapper that allows for your code or libjansson in a not too terrible way?

inliniac:
I guess the rust thing gets in the way there too...

jasonish:
> I guess the rust thing gets in the way there too...

Rust loggers would just build up this intermediate structure as well.  I don't think I'd go so far as calling it ugly. It could potentially be a nicer API for logging. It just not trivial.

PR #2555, opened 2017-02-15
TLS logging for session resumption 1.2

rtkrruvinskiy:

was never seen.

Replaces https://github.com/inliniac/suricata/pull/2536


thus:
Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/97
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/97

src/log-tlslog.c
```diff
@@ -291,8 +297,11 @@ static int LogTlsLogger(ThreadVars *tv, void *thread_data, const Packet *p,
         return 0;
     }
 
-    if (ssl_state->server_connp.cert0_issuerdn == NULL ||
-            ssl_state->server_connp.cert0_subject == NULL) {
+    if ((ssl_state->server_connp.cert0_issuerdn == NULL ||
+            ssl_state->server_connp.cert0_subject == NULL) &&
+        ((ssl_state->flags & SSL_AL_FLAG_SESSION_RESUMED) == 0 ||
+         (hlog->flags & LOG_TLS_SESSION_RESUMPTION) == 0)) {
```
thus: The indentation for the various parts of this "if" is not aligned with each other.

src/output-json-tls.c
```diff
@@ -156,8 +164,10 @@ static int JsonTlsLogger(ThreadVars *tv, void *thread_data, const Packet *p,
         return 0;
     }
 
-    if (ssl_state->server_connp.cert0_issuerdn == NULL ||
-            ssl_state->server_connp.cert0_subject == NULL)
+    if ((ssl_state->server_connp.cert0_issuerdn == NULL ||
+            ssl_state->server_connp.cert0_subject == NULL) &&
+        ((ssl_state->flags & SSL_AL_FLAG_SESSION_RESUMED) == 0 ||
+         (tls_ctx->flags & LOG_TLS_SESSION_RESUMPTION) == 0))
```
thus: Same as commented above.

thus:
Except for the nitpick comments above, I think it looks good :)

rtkrruvinskiy:
@thus In the identation, I aligned the parentheses of the two parts of the branch (because for me it makes it easy to see them), but I can easily change that. For clarity, how should I handle the indentation? In the part of the `if` that was already there, it looks like the second line is indented five spaces after the parenthesis on the first line. Should all subsequent lines also be indented at the same level as the second line?

thus:
I personally find it more visually appealing to line up the following lines in an 'if' statement. The reason for using eight spaces (counted from the start of the 'if') is to avoid that any part of the 'if' statement aligns with the code coming right after, which (IMO) would make the code harder to read.

rtkrruvinskiy:
Replaced by https://github.com/inliniac/suricata/pull/2562

PR #2558, opened 2017-02-16
Makefile: fix race condition in make install-full - v1

jasonish:
Break out the directory creation for install-conf and
install-rules into their own targets to avoid the mkdir
race conditional when these targets are executed in
parallel.

Addresses issue:
https://redmine.openinfosecfoundation.org/issues/1470
which triggered on OSX/macOS.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/89
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/441

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,8 +14,13 @@ install-data-am:
 
 install-full: install install-conf install-rules
 
-install-conf:
+"$(DESTDIR)$(e_sysconfdir)":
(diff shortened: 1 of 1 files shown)


jasonish:
Looking into the windows build issue.

PR #2561, opened 2017-02-16
Makefile: fix race condition in make install-full

jasonish:
Use recursive make for the install process so it
is executed in a predictable order.

Addresses issue:
https://redmine.openinfosecfoundation.org/issues/1470
which triggered on OSX/macOS.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/91
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/443


inliniac:
Merged through #2572, thanks Jason!

PR #2569, opened 2017-02-17
Doc: specified tls buffers that can be used for fast_pattern

counterthreatunit:
also removed lines saying which Suricata version specific buffer support was added


counterthreatunit:
going to fix newline issue and re-submit

PR #2570, opened 2017-02-17
doc: specify TLS buffers that can be used for fast_pattern

counterthreatunit:
also removed lines saying which Suricata version specific buffer support was added


inliniac:
Can you do proper commit message formatting? The requirements are detailed here https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits, esp point 5

counterthreatunit:
OK.  I will re-submit.

PR #2573, opened 2017-02-17
pidfile: fail if the pid file exists - v1

jasonish:
Changes the pidfile check function to fail just on the
existence of the file to address issue
https://redmine.openinfosecfoundation.org/issues/1335
but log a message if the pid file appears to be stale.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/92
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/444

### DIFF
--- a/src/util-pidfile.c
+++ b/src/util-pidfile.c
@@ -89,8 +89,13 @@ void SCPidfileRemove(const char *pid_filename)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


inliniac:
Merged through #2592, thanks Jason!

PR #2574, opened 2017-02-17
TPACKET_V3 fixes: packet parsing and VLAN handling

gozzy:
1. We still need to handle tp_vlan_tci to pass VLAN information properly...
   AFPWritePacket() now handles both V2 and V3 cases.
2. AFPParsePacketV3() saved tpacket_block_desc structure instead of tpacket3_hdr.
   The bug was introduced when V3 support was added, but somehow it worked for some setups.
   C makes everything possible :)

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -679,6 +679,8 @@ static TmEcode AFPWritePacket(Packet *p, int version)
     int socket;
     uint8_t *pstart;
     size_t plen;
+    union thdr h;
+    uint16_t vlan_tci = 0;
 
     if (p->afp_v.copy_mode == AFP_COPY_MODE_IPS) {
         if (PACKET_TEST_ACTION(p, ACTION_DROP)) {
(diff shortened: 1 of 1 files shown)


gozzy:
It seems that I've found some issues related to TPACKET_V3. Somehow it worked when I tested it before, but with 3.2.1 release, TPACKET_V3 started to behave weird.

Needs further tests and I hope to do this next week... By now I've ran some tests on virtual machines and everything seems to work.

@inliniac, @regit - comments or suggestions are welcome.

gozzy:
IXIA is happy too, tagged traffic passes suricata successfully.

src/source-af-packet.c
```diff
@@ -968,12 +980,19 @@ static inline int AFPParsePacketV3(AFPThreadVars *ptv, struct tpacket_block_desc
...
+        p->vlanh[0] = NULL;
+    }
+
     if (ptv->flags & AFP_ZERO_COPY) {
         if (PacketSetData(p, (unsigned char*)ppd + ppd->tp_mac, ppd->tp_snaplen) == -1) {
             TmqhOutputPacketpool(ptv->tv, p);
             SCReturnInt(AFP_FAILURE);
         }
-        p->afp_v.relptr = pbd;
+        p->afp_v.relptr = ppd;
```
gozzy: AFAIU, this line was the root of all evil. Everything else was induced by it.

inliniac:
I'm running this now with V3 and it seems to work correctly. No VLANs though.

@regit any thoughts on this PR?

src/source-af-packet.c
regit: OK, I think that was triggered by the fact I've added room at the head of the ring cell breaking the alignment. I'm really in favor of separating both fixes.

src/source-af-packet.c
```diff
@@ -705,25 +707,35 @@ static TmEcode AFPWritePacket(Packet *p, int version)
         SCMutexLock(&p->afp_v.peer->sock_protect);
     socket = SC_ATOMIC_GET(p->afp_v.peer->socket);
 
+    h.raw = p->afp_v.relptr;
+
+    /* Copy VLAN header from ring memory. For post june 2011 kernel we test
```
regit: Comment is linked to l.716 so it should not be moved.

src/source-af-packet.c
```diff
@@ -968,12 +980,19 @@ static inline int AFPParsePacketV3(AFPThreadVars *ptv, struct tpacket_block_desc
     p->livedev = ptv->livedev;
     p->datalink = ptv->datalink;
 
+    if ((!(ptv->flags & AFP_VLAN_DISABLED)) &&
```
regit: This is a parsing fix it should be in a separate commit.

regit:
@inliniac Just commented. It was really hard to read. I'm proposing to separate patches and rewording commit messages.

gozzy:
So... There should be two commits, like:
1. Parsing/handling VLAN in V3
2. Fix relptr saving

Right?

gozzy:
Or even three commits?

regit:
@gozzy I'm in favor of three commits because they functionally fix different things. (1 for write, 1 for parsing, 1 for relptr).

gozzy:
Ok.

PR #2575, opened 2017-02-17
tcp: fix checksum validation when 0xffff - v1

jasonish:
Issue:
https://redmine.openinfosecfoundation.org/issues/2041

One approach is to fixing this issue to just validate the
checksum instead of regenerating it and comparing it. This
method is used in some kernels and other network tools.

When validating, the current checksum is passed in as an
initial argument which will cause the final checksum to be 0
if OK. If generating a checksum, 0 is passed and the result
is the generated checksum.

If OK, I'll apply this same idea to the other checksum functions.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/93
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/445


inliniac:
I can't think of drawbacks, so let's do it.

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2596

PR #2563, opened 2017-02-17
eve: async mode for redis output - v5

fooinha:
doc: async: true ## if redis replies are read asynchronously
eve: detects libevent for async redis at configure
eve: moves redis output code to new file
eve: redis pipeline batch commands with async mode

Replaces #2385

PR #2576, opened 2017-02-18
eve: async mode for redis output - v6

fooinha:

Replaces #2563 with fixes

- Files renamed
- Fixes the default configuration event truncated write issue 
- Fixes reconnect issues ... I tried a very aggressive re-connection ratio, and it went ok. but requires more testing. 
- Fixes some memory leaks
- A bit of refactor to separate sync/async logic in different functions.
- ECHO response to detect if redis is available
- QUIT command to request disconnection from redis

@jasonish , fyi...  

Thanx

### DIFF
--- a/src/util-log-redis.c
+++ b/src/util-log-redis.c
@@ -1,0 +1,607 @@
+/* vi: set et ts=4: */
+/* Copyright (C) 2007-2016 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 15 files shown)


src/util-log-redis.c
```diff
@@ -0,0 +1,607 @@
...
+    if (is_async) {
+        SCLogWarning(SC_ERR_NO_LIBEVENT, "async option not available.");
+    }
+    is_async = 0;
+#endif //ifndef HAVE_LIBEVENT
+
+    log_ctx->redis_setup.is_async = is_async;
+
+    if (!log_ctx->redis_setup.key) {
+        SCLogError(SC_ERR_MEM_ALLOC, "Unable to allocate redis key name");
```
jasonish: This now fails if you don't have a redis configuration. It should default to "suricata".  Looks like it only uses the default now if a "redis" node exists, but in my default case a redis node does not exist.

jasonish:
I still have connection/reconnection issues with async mode.

In non-async if I start Suricata with Redis not available, or stop and restart Redis, it will re-connect and start logging.

However, with async mode if I start Suricata with Redis not available, it never connects.  If I stop Redis after its connected it never reconnects.

fooinha:
Thanx, I try to replicate it.

fooinha:
Replaced by #2601

PR #2577, opened 2017-02-19
doc: Snort compatibility overhaul

counterthreatunit:
Overhaul of the snort-compatibility document.  Renamed file from "snort-compatibility" to "differences-from-snort".

Based on "The Suricata Rule Writing Guide for The Snort Expert" 2016 SuriCon talk.


inliniac:
in #2570 the references to older suri versions are removed, but here I see quite a few of them still. I think it makes sense to remove them here as well.

Same comment about git message formatting applies here too.

counterthreatunit:
In the original version, there were discussions on how previous Suricata versions behaved, and notes on when certain features were added.  I understand the condemning of the former since the docs are now versioned but I do see value in the latter, especially since the docs only started being versioned recently (3.x branch?).

However, I can make the requested changes and re-format commit.

PR #2449, opened 2016-12-03
log: TLS/SSL custom log

fooinha:
Custom format output for TLS/SSL

Work in progress for review and comments.

Requires #2448.
Useful for #2427 ( @thus ) and future output loggers.

PR #2581, opened 2017-02-20
log: common custom format output

fooinha:
Re-adapt the HTTP custom format parser for a common place, so it can be available to any line based output logger. 

Needed by #2449.

Replaces #2448 

@thus, @inliniac


inliniac:
I'll consider this one as part of https://github.com/inliniac/suricata/pull/2449#issuecomment-290462805, thanks!

PR #2582, opened 2017-02-20
doc: specify TLS buffers that can be used for fast_pattern

counterthreatunit:
Updated notes on the following buffers indicating that they can be used for fast_pattern:
    tls_cert_subject
    tls_cert_issuer
    tls_sni

I also removed references to older Suricata versions since the docs are versioned now.


inliniac:
I've merged this through #2592 with a slight git commit message adjustment: I've added 'doc: <msg>' on the subject line. For future commits, please do the same. It helps us a lot when going back in time to find changes.

Thanks David!

PR #2583, opened 2017-02-20
TPACKET_V3 fixes v2

gozzy:
Splitted #2574.


gozzy:
@regit, should be better now.
@inliniac, should I create a ticket for this?

inliniac:
Yeah please do. Also, the commit message formatting could use a bit of refinement :)

regit:
On commit message: please just prefix with 'af-packet'

regit:
That's far better. Really easier to read now.

PR #2586, opened 2017-02-20
Bug #2044

gozzy:
Rework of #2583.


inliniac:
Merged through #2592, thanks Alexander!

PR #2587, opened 2017-02-20
logging - address coverity issues - v4

jasonish:
The first commit fixes a false positive, but still confusing code where the mutex was unlocked via a different name than the lock.

The second commit is to address another false positive where Coverity could not reason that the Write function was always called under lock. This is an attempt to make it more explicit, but required some refactoring.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/95
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/447

### DIFF
--- a/src/alert-debuglog.c
+++ b/src/alert-debuglog.c
@@ -304,11 +304,8 @@ static TmEcode AlertDebugLogger(ThreadVars *tv, const Packet *p, void *thread_da
         }
     }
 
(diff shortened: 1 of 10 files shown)


jasonish:
Closing, see https://github.com/inliniac/suricata/pull/2588

PR #2565, opened 2017-02-17
logging: support file rotation

thus:
Add support for log file rotation, without the need for external tools like logrotate.

The pull request basically adds three new features for all the loggers. It allows the user to use date modifiers in the log filename in suricata.yaml, e.g.:

```yaml
outputs:
  - eve-log:
      filename: eve-%Y-%m-%d-%H:%M:%S.json
```

It also adds log file rotation based on either minute/hour/day or a relative value like X seconds/minutes/hours/days/weeks, e.g.:

```yaml
outputs:
  - eve-log:
      filename: eve-%Y-%m-%d-%H:%M:%S.json
      rotate-interval: minute
```

The example above rotates the eve-log every minute. This could be changed to 'rotate-interval: 120s' to rotate the file every 120 seconds instead.

The last feature is to create new directories when needed. This enables us to use date modifiers to create directory structures (like often done by tools like rsyslog), e.g.:

```yaml
outputs:
  - eve-log:
      filename: /var/log/suricata/%Y/%m/%d/eve-%H.json
      rotate-interval: hour
```

The example above would create a new folder per day and rotate the eve-log every hour.

https://redmine.openinfosecfoundation.org/issues/1323
(it rotates on time and not size, so I'm not sure it  entirely solves this issue)

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/99
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/99

PR #2593, opened 2017-02-21
logging: support custom file permissions per logger

thus:
Support setting file permissions per logger using 'filemode', e.g.:

```yaml  
outputs:
  - eve-log:
      enabled: yes
      filetype: regular
      filename: eve.json
      filemode: 660
```

This pull request conflicts with #2565, so one of them will have to be rebased if the other is merged.

https://redmine.openinfosecfoundation.org/issues/2046

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/112
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/112

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -31,6 +31,7 @@
 #include "tm-modules.h"      /* LogFileCtx */
 #include "conf.h"            /* ConfNode, etc. */
 #include "output.h"          /* DEFAULT_LOG_* */
+#include "util-byte.h"
 #include "util-logopenfile.h"
 #include "util-logopenfile-tile.h"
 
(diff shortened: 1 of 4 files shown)


src/util-logopenfile.c
```diff
@@ -185,6 +187,15 @@ SCLogOpenFileFp(const char *path, const char *append_setting)
     if (ret == NULL)
         SCLogError(SC_ERR_FOPEN, "Error opening file: \"%s\": %s",
                    path, strerror(errno));
+
+    if (mode != 0) {
+        int r = chmod(path, mode);
+        if (r < 0) {
+            SCLogError(SC_WARN_CHMOD, "Error: could not chmod '%s' to %d: %s",
+                       path, mode, strerror(errno));
```
jasonish: Should probably call SCLogWarning as well.

jasonish:
Still requires a doc update as it no longer includes the example in the configuration.

jasonish:
@inliniac I don't mind this feature, there's been times when I would have used it.

src/util-logopenfile.c
thus: I missed that one, thanks!

thus:
Replaced by #2598

PR #2595, opened 2017-02-21
util-daemon: user-configurable umask setting

thus:
Make umask user-configurable by adding 'daemon-umask' to suricata.yaml.

```yaml
# Daemon umask.
# Suricata will use this umask if it is provided. By default it will use the
# umask passed on by the parent process.
daemon-umask: 022
```

https://redmine.openinfosecfoundation.org/issues/1678

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/113
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/113

### DIFF
--- a/src/util-daemon.c
+++ b/src/util-daemon.c
@@ -28,6 +28,7 @@
 #include "runmodes.h"
 #include "util-daemon.h"
 #include "util-debug.h"
+#include "util-byte.h"
 #include "conf.h"
 
 #ifndef OS_WIN32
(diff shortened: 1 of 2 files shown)


jasonish:
It looks like this only applies when running as a daemon? Is that still applicable these days when things like systemd run Suricata in the foreground? Would this setting still be valid in that case as well?

thus:
@jasonish, would a more general `umask: 022`, that works daemonized or not, be more useful? And use the umask passed on by the shell as the default one.

inliniac:
@jasonish @thus any conclusions reached on this?

thus:
I agree with @jasonish, that it would be better to implement it to work both when running as a daemon and not.

jasonish:
> @jasonish, would a more general umask: 022, that works daemonized or not, be more useful? And use the umask passed on by the shell as the default one.

Yes, this would be more useful I think.

thus:
Then I'll do that :)

PR #2596, opened 2017-02-21
tcp: fix checksum validation when 0xffff - v2

jasonish:
Previous PR: #2575

Issue:
https://redmine.openinfosecfoundation.org/issues/2041

One approach is to fixing this issue to just validate the
checksum instead of regenerating it and comparing it. This
method is used in some kernels and other network tools.

When validating, the current checksum is passed in as an
initial argument which will cause the final checksum to be 0
if OK. If generating a checksum, 0 is passed and the result
is the generated checksum.

Apply the same logic to the other IP and UDP checksums for consistency.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/100
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/452


inliniac:
Merged, thanks Jason!

PR #2599, opened 2017-02-23
app-layer-tls: check if validity dates are already set

thus:
Fix bug where a client certificate (if present), would overwrite the validity dates of the server certificate.

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/116
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/116


inliniac:
How would it ever be valid for a client cert to write ssl_state->server_connp.cert0_not_before? I assume that the *server* part of that var name means it's about the server?

thus:
@inliniac, stop asking reasonable questions!

Now that you mention it, it would probably be a better solution to check the direction of the traffic, and store it in either client_connp or server_connp based on that.

I'll make the changes tomorrow and resubmit.

PR #2600, opened 2017-02-23
app-layer-tls: check if validity dates are already set (3.2.x)

thus:
Fix bug where a client certificate (if present), would overwrite the validity dates of the server certificate.

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/117
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/117


thus:
See comment in #2599

PR #2601, opened 2017-02-23
eve: async mode for redis output - v7

fooinha:
eve: redis output defaults if conf is missing

Replaces #2576 

@jasonish

### DIFF
--- a/doc/userguide/configuration/suricata-yaml.rst
+++ b/doc/userguide/configuration/suricata-yaml.rst
@@ -267,6 +267,7 @@ integration with 3rd party tools like logstash.
       #redis:
       #  server: 127.0.0.1
       #  port: 6379
+      #  async: true ## if redis replies are read asynchronously
       #  mode: list ## possible values: list (default), channel
       #  key: suricata ## key or channel to use (default to suricata)
       # Redis pipelining set up. This will enable to only do a query every
--- a/src/util-log-redis.c
+++ b/src/util-log-redis.c
@@ -1,0 +1,625 @@
+/* vi: set et ts=4: */
+/* Copyright (C) 2007-2016 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Paulo Pacheco <fooinha@gmail.com>
+ *
+ * File-like output for logging:  redis
+ */
+#include "suricata-common.h" /* errno.h, string.h, etc. */
+#include "util-log-redis.h"
+#include "util-logopenfile.h"
+
+#ifdef HAVE_LIBHIREDIS
+
+#ifdef HAVE_LIBEVENT_PTHREADS
+#include <event2/thread.h>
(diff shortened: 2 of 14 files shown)


src/output.c
```diff
@@ -1002,8 +1006,16 @@ void OutputRegisterRootLogger(ThreadInitFunc ThreadInit,
     TAILQ_INSERT_TAIL(&RootLoggers, logger, entries);
 }
 
+//XXX FIXME  - What the best place and names for this
+void OutputLibEventUsePthreads() {
+#ifdef HAVE_LIBEVENT_PTHREADS
+    evthread_use_pthreads();
+#endif
+}
+
```
jasonish: Perhaps do this in util-log-redis.c, SCLogRedisInit(), then you can do the ifdef in there, and keep the libevent stuff out of src/output.c as libevent usage is only for redis, for now.

src/util-log-redis.c
```diff
@@ -0,0 +1,602 @@
...
+ *         Emits and awaits response for an async QUIT command.
+ *         It's used to disconnect with redis
+ *  \param ctx redis context
+ */
+static void SCLogAsyncRedisSendQuit(SCLogRedisContext * ctx)
+{
+    if (ctx->connected) {
+        redisAsyncCommand(ctx->async, SCRedisAsyncQuitCommandCallback, ctx, "QUIT");
+        SCLogInfo("QUIT Command sent to redis. Connection will terminate!");
+        event_base_dispatch(ctx->ev_base);
```
jasonish: I'm not sure about the use of event_base_dispatch here.  It basically makes this call blocking, as its not going to return until event_base_breakloop() is called.  Is there a reason these can't be event_base_loop(ctx->ev_base, EVLOOP_NONBLOCK) like used in the Write message which are non-blocking from the perspective of Suricata?

jasonish:
Overall this is working pretty well.  ASAN is happy.  Connect/reconnect work well.

Async has the advantage of not blocking Suricata during write if the socket becomes backed up, which could happen with the non-async mode, just like unix_stream.

I'd still like to run it for an extended period of time though.

src/util-log-redis.c
```diff
@@ -0,0 +1,602 @@
...
+static int SCConfLogReopenAsyncRedis(LogFileCtx *log_ctx)
+{
+    SCLogRedisContext * ctx = log_ctx->redis;
+    const char *redis_server = log_ctx->redis_setup.server;
+    int redis_port = log_ctx->redis_setup.port;
+
+    /* only try to reconnect once per second */
+    if (log_ctx->redis_setup.tried >= time(NULL)) {
+        return -1;
+    }
```
jasonish: This never seems to be true, so it attempts to reconnect many times per second, at least with auto-fp.  tried is reset to 0 at the bottom of this function probably causing this.

fooinha: I'll try to fix this and keep behavior ...

doc/userguide/configuration/suricata-yaml.rst
```diff
@@ -267,6 +267,7 @@ integration with 3rd party tools like logstash.
       #redis:
       #  server: 127.0.0.1
       #  port: 6379
+      #  async: true ## if redis replies are read asynchronously
```
jasonish: I thought this was on the wiki but its not, documentation commits should be in their own commit please, separate from code commits.

fooinha: two different commits now.

src/output.c
fooinha: moved this to util-log-redis.c ... and it's called only if needed by output json init register.

src/util-log-redis.c
fooinha: Yes! ... replaced this one to event_base_loop(ctx->ev_base, EVLOOP_NONBLOCK)

src/util-log-redis.c
fooinha: Fixed this, also the state variables that control this were moved from redis setup struct to context struct.

src/util-log-redis.c
fooinha: we'll keep  event_base_dispatch(ctx->ev_base) here , because we're terminating, otherwise we may leak... 
ASAN confirms.
I'll re-visit this.

fooinha:
Replaced by #2604

PR #2602, opened 2017-02-23
TLS logging for session resumption 1.4

rtkrruvinskiy:

was never seen.

Replaces https://github.com/inliniac/suricata/pull/2562


rtkrruvinskiy:
I've tried to keep the behaviour of the JSON logger the same as in the previous version of this PR. Namely, in the case where we say `session-resumption: no`, TLS transactions with resumed sessions are not logged *at all* (which is the current behaviour of master). Independently of that, I also added `session_resumed` field toggle for custom logging, to control whether the `"session": "resumed"` field itself is output, in keeping with the new code since the last time I rebased.

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -232,6 +233,14 @@ Example of regular TLS logging:
       "issuerdn": "C=US, O=Google Inc, CN=Google Internet Authority G2"
   }
 
+Example of regular TLS logging for resumed sessions:
+
+::
+
+  "tls": {
+      "session": "resumed"
```
inliniac: can we think of other values that "session" could get? If not then I think we should turn it into a bool.

src/log-tlslog.c
```diff
@@ -294,8 +300,10 @@ static int LogTlsLogger(ThreadVars *tv, void *thread_data, const Packet *p,
         return 0;
     }
 
-    if (ssl_state->server_connp.cert0_issuerdn == NULL ||
-            ssl_state->server_connp.cert0_subject == NULL) {
+    if ((ssl_state->server_connp.cert0_issuerdn == NULL ||
+            ssl_state->server_connp.cert0_subject == NULL) &&
+            ((ssl_state->flags & SSL_AL_FLAG_SESSION_RESUMED) == 0 ||
+            (hlog->flags & LOG_TLS_SESSION_RESUMPTION) == 0)) {
```
inliniac: micro optimization: I wonder if we should check for hlog->flags & LOG_TLS_SESSION_RESUMPTION first? Safes a lot of ssl_state access if disabled, and since it's a more or less global variable it probably more likely to be hot in cache.

doc/userguide/output/eve/eve-json-format.rst
rtkrruvinskiy: I initially had it as `"session_resumed": true` but changed it to this variant following PR feedback from previous review rounds. I'd be happy to change it back (or to something else). Should I change it back?

src/log-tlslog.c
rtkrruvinskiy: To confirm, do you mean the following?
```c
if (((hlog->flags & LOG_TLS_SESSION_RESUMPTION) == 0 ||
            (ssl_state->flags & SSL_AL_FLAG_SESSION_RESUMED) == 0) &&
            (ssl_state->server_connp.cert0_issuerdn == NULL ||
            ssl_state->server_connp.cert0_subject == NULL)) {
```
I'd be happy to make that change.

rtkrruvinskiy:
@inliniac @thus I was wondering if I could get a final clarification on the comments above. I'm looking forward to having this merged!

doc/userguide/output/eve/eve-json-format.rst
inliniac: @jasonish any preference on your end? Mostly thinking about whats easiest to use in Kibana and Evebox.

src/log-tlslog.c
inliniac: yeah thats what I had in mind, thanks :)

doc/userguide/output/eve/eve-json-format.rst
jasonish: I think a bool is better if you wanted to search on "resumed" sessions. Technically it doesn't matter, but its easier to think of it as a bool I think, if there are no other values that make sense.

Just to confirm, "sni" still exists so we can back track to the initial session?

rtkrruvinskiy:
Replaced by https://github.com/inliniac/suricata/pull/2668

PR #2603, opened 2017-02-23
(rfc) logging - non-blocking writes to socket - v1

jasonish:

Related issue:
https://redmine.openinfosecfoundation.org/issues/2039

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/102
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/454

### DIFF
--- a/src/log-httplog.c
+++ b/src/log-httplog.c
@@ -709,6 +709,12 @@ static void LogHttpLogDeInitCtx(OutputCtx *output_ctx)
 {
     LogHttpFileCtx *httplog_ctx = (LogHttpFileCtx *)output_ctx->data;
     uint32_t i;
+
+    if (httplog_ctx->file_ctx->dropped) {
+        SCLogNotice("%s: %"PRIu64" events dropped instead of logged.",
+                MODULE_NAME, httplog_ctx->file_ctx->dropped);
+    }
+
     for (i = 0; i < httplog_ctx->cf_n; i++) {
         SCFree(httplog_ctx->cf_nodes[i]);
     }
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -120,6 +120,47 @@ static int SCLogUnixSocketReconnect(LogFileCtx *log_ctx)
     return log_ctx->fp ? 1 : 0;
 }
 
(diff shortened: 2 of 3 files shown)


src/util-logopenfile.c
```diff
@@ -120,37 +120,72 @@ static int SCLogUnixSocketReconnect(LogFileCtx *log_ctx)
     return log_ctx->fp ? 1 : 0;
 }
 
+static int SCLogFileWriteSocket(const char *buffer, int buffer_len,
+        LogFileCtx *ctx)
+{
+    int tries = 0;
+    int ret = 0;
+    int reopen = false;
```
inliniac: bool?

src/log-httplog.c
```diff
@@ -709,6 +709,12 @@ static void LogHttpLogDeInitCtx(OutputCtx *output_ctx)
 {
     LogHttpFileCtx *httplog_ctx = (LogHttpFileCtx *)output_ctx->data;
     uint32_t i;
+
+    if (httplog_ctx->file_ctx->dropped) {
```
inliniac: can we generalize this somehow when the other outputs are supported as well?

jasonish: Yes, so the counter is generalized, its just the logging.

Might be kind of hard as the code calling ogHttpLogDeInitCtx (or other DeInitCtx) doesn't know if they are using common writer or not, the calling code just has an opage void data pointer to logger specific data.

What might be more useful is a stat counter? Either per logger, or a simpler single counter for dropped log messages?

jasonish: Could put it in LogFileFreeCtx:
```
[29123] 28/3/2017 -- 19:26:21 - (util-logopenfile.c:639) <Notice> (LogFileFreeCtx) -- 20197 events dropped to socket /tmp/socket
```

I'm thinking a counter in stats would be good as well - perhaps 2 counts. logging.events=<total number of events logged>. logging.dropped=<count> where count is the number of events that weren't delivered, either due to a socket that would block, and socket that was disconnected, or whatever.

Not sure if the complexity of breaking it down into per-logger stats is worth it or not.

jasonish:
Closing. See #2649

PR #2605, opened 2017-02-28
Extracting HTTPS URL´s from SMTP, currently only HTTP is supported

noperle:
Hello :)
We are looking for extracting HTTPS URL´s from SMTP. Currently, only HTTP is supported
Can you please add this?


inliniac:
Not sure why all the extra commits are in there, please fix that first.

PR #2607, opened 2017-03-03
doc: Snort compatibility overhaul v2

counterthreatunit:
Overhaul of the snort-compatibility document.

Renamed file from "snort-compatibility" to "differences-from-snort".

Based on "The Suricata Rule Writing Guide for The Snort Expert" 2016 SuriCon talk.


doc/userguide/rules/differences-from-snort.rst
```diff
@@ -0,0 +1,823 @@
...
+   Snort ``http_cookie`` buffer contents::
+
+        monsterelmo
+
+-  Corresponding PCRE modifier: ``C`` (same as Snort)
+
+``http_user_agent`` Buffer
+--------------------------
+
+-  Suricata has a ``http_user_agent`` buffer, Snort does not.
```
inliniac: I wonder if it makes sense to list the things we support that Snort doesn't here. I think this document shouldn't be a feature comparison, but rather a description of how things differ where Snort users might have expectations based on their Snort experience.

I think this block (and similar ones below), could be moved to the regular keyword documentation?

counterthreatunit: Hi Victor,

In my mind this is a document that highlights differences from Snort when it comes to rules. As such, some feature comparisons are appropriate and necessary although the document is certainly not designed to be a holistic comparison of the products.

Since Suricata claims to support, for the most part, the Snort rule language and a lot of people writing rules for Suricata have a background writing Snort rules, are transitioning from Snort, or write rules for both platforms, the thought is that this would be a useful reference document and a benefit to the Suricata community.  While somewhat differing from traditional Suricata product documentation, I believe it does meet a need and I and my colleagues have used this document (or previous/similar/internal revisions) over the years as a valuable reference. (Yes, I do refer to and reference my own documentation.)

If you would like to go a different direction, feel free to do so but I'd rather not just put this work in a blog post somewhere where it can't be updated, maintained, and curated like the living document I think it should be.

Thank you.

inliniac:
Merged as part of #2715, thanks a lot @counterthreatunit !

PR #2619, opened 2017-03-23
rfc - app-layer gap handling - v4

jasonish:
Requesting comments on this way of handling gaps in the app-layer.  DNP3 is a simple case as its based on frames with a start sequence.

There is one stream test case that is broken from these changes tho.

There are ifdef's of older code as I'm still using them for reference.. Hence this is just an RFC.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/107
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/459


jasonish:
See https://github.com/inliniac/suricata/pull/2660

PR #2623, opened 2017-03-27
travis: macos: unlink all deps, then relink - v2

jasonish:
Kind of ugly, but first unlink all dependencies then install.
The deps that don't get an upgrade will remain unlinked, so
relink all dependencies as relinking an already linked dep
does not error out.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/108
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/460

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -127,12 +127,29 @@ before_install:
     if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
         brew update
 
-        # Unlink pcre in case its already installed.
-        brew unlink pcre || true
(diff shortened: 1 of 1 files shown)


inliniac:
I wonder why prscript accepted this as it's not in sync with master. I'll rebase it but wonder if prscript's check is broken.

inliniac:
Merged after rebase, thanks Jason.

PR #2604, opened 2017-02-24
dev eve output redis async v8

fooinha:
**commit**
doc: async: true ## if redis replies are read asynchronously

**commit**
eve: detects libevent for async redis at configure
eve: moves redis output code to new file - util-log-redis.{c,h}
eve: redis pipeline batch commands with async mode
eve: redis ECHO and QUIT commands for async mode
eve: redis output defaults if conf is missing

Replaces #2601

@jasonish

PR #2636, opened 2017-03-30
Dev eve output redis async v9

fooinha:
commit

commit
eve: redis output defaults if conf is missing

Replaces #2604

Many thanx to @jasonish for endless support when testing this. Please review.

### DIFF
--- a/src/util-log-redis.c
+++ b/src/util-log-redis.c
@@ -1,0 +1,607 @@
+/* vi: set et ts=4: */
+/* Copyright (C) 2007-2016 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Paulo Pacheco <fooinha@gmail.com>
+ *
+ * File-like output for logging:  redis
+ */
+#include "suricata-common.h" /* errno.h, string.h, etc. */
+#include "util-log-redis.h"
+#include "util-logopenfile.h"
+
+#ifdef HAVE_LIBHIREDIS
+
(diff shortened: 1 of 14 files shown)


src/util-log-redis.c
```diff
@@ -0,0 +1,607 @@
...
+        redisAsyncCommand(ctx->async, SCRedisAsyncQuitCommandCallback, ctx, "QUIT");
+        SCLogInfo("QUIT Command sent to redis. Connection will terminate!");
+    }
+
+    redisAsyncFree(ctx->async);
+    event_base_dispatch(ctx->ev_base);
+    ctx->async = NULL;
+    event_base_free(ctx->ev_base);
+    ctx->ev_base = NULL;
+    ctx->connected = -1;
```
jasonish: I see that ctn->connected can have values -1, 0, 1 and 2. If there is any significance to values other than 0, 1 I think these should be #define'd labels instead of just numbers.

src/util-log-redis.c
```diff
@@ -0,0 +1,607 @@
...
+                file_ctx,
+                "%s %s %s",
+                file_ctx->redis_setup.command,
+                file_ctx->redis_setup.key,
+                string);
+
+        event_base_loop(ctx->ev_base, EVLOOP_NONBLOCK);
+        return 0;
+    }
+
```
jasonish: I was wondering a bit about pipeline mode with async, and I don't think its required.  The Hiredis documentation suggests the same: https://github.com/redis/hiredis#sending-commands-and-their-callbacks, and I think we should get rid of async when pipelining.

One of the reasons to pipeline when not async is to limit the amount of time blocking for replies, and that is not an issue when async. This send block above is the only one we need. If there is data ready to be read, it will be read. If there isn't, it won't block waiting for it. So pipelining doesn't buy us anything.

src/util-log-redis.c
```diff
@@ -0,0 +1,607 @@
...
+    if (ctx->sync != NULL)  {
+        redisFree(ctx->sync);
+    }
+    ctx->sync = redisConnect(redis_server, redis_port);
+    if (ctx->sync == NULL) {
+        SCLogError(SC_ERR_SOCKET, "Error connecting to redis server.");
+        ctx->tried = time(NULL);
+        return -1;
+    }
+    if (ctx->sync != NULL && ctx->sync->err) {
```
jasonish: No need to check for NULL here as its done above. In fact, Coverity might even flag this.

src/util-log-redis.c
fooinha: These values are not required, simple boolean 0 and 1 is enough . I've added when trying them create the not hammering the logs logic. I'll change this.

src/util-log-redis.c
fooinha: Makes perfect sense, thanx for seeing this. I will remove the batch count and pipeline logic when in async mode

fooinha:
Replaced by #2640

PR #2638, opened 2017-03-31
Master 3.1.x

admin2020:
Do we have have any parameter or logs to check how many flows are active (flowbits)?


inliniac:
Don't use pull requests for questions please, we have a ticket system here: https://redmine.openinfosecfoundation.org/projects/suricata/issues

PR #2639, opened 2017-03-31
TEST flow: track longest packet 'pause' / 'gap'

inliniac:
Track the max difference between packet updates in a flow. Print
the max value as 'max_pause' in the flow record. Useful for
determining what a reasonable timeout value could be for flows.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/79
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/587

### DIFF
--- a/src/flow.c
+++ b/src/flow.c
@@ -249,6 +249,9 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
 
     int state = SC_ATOMIC_GET(f->flow_state);
 
+    if (f->lastts.tv_sec > 0) {
+        f->lastts_gap = MAX((p->ts.tv_sec - f->lastts.tv_sec), f->lastts_gap);
+    }
     if (state != FLOW_STATE_CAPTURE_BYPASSED) {
         /* update the last seen timestamp of this flow */
         COPY_TIMESTAMP(&p->ts, &f->lastts);
--- a/src/flow-util.h
+++ b/src/flow-util.h
@@ -57,6 +57,7 @@
         (f)->alproto_tc = 0; \
         (f)->de_ctx_version = 0; \
         (f)->thread_id = 0; \
+        (f)->lastts_gap = 0; \
         (f)->alparser = NULL; \
         (f)->alstate = NULL; \
         (f)->sgh_toserver = NULL; \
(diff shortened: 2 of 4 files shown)


pevma:
That is a very useful counter. Seems to work as expected.
An example of jq query i used - 

> tail -F /var/data/regit/log/suricata/eve.json | jq -c 'select((.proto=="TCP") and (.flow.max_pause > 100 and .flow.max_pause < 110))' |jq '.'

src/flow.c
```diff
@@ -249,6 +249,9 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
 
     int state = SC_ATOMIC_GET(f->flow_state);
 
+    if (f->lastts.tv_sec > 0) {
+        f->lastts_gap = MAX((p->ts.tv_sec - f->lastts.tv_sec), f->lastts_gap);
```
inliniac: need to make sure p->ts > f->lastts here.

inliniac:
I don't really like the 'max_pause' label name... any thoughts?

pevma:
max_pause_interval?

jasonish:
I believe the common term is inter-packet arrival time. I'm not a fan of the name "pause" here, as its not really a "pause", its more a lull. Inter-packet arrival time also infers direction, which this patch is not taking into account.

How about simply, "packet_interval_max"? Mean and min might be interesting to track as well, however is the naming keeps us open to adding them with the same name scheme.

pevma:
I like the idea - how about - just switching it to "max_packet_interval"?

jasonish:
> I like the idea - how about - just switching it to "max_packet_interval"?
>  Show all checks

My only thought there is if we were to add "mean" and "min" at some point, what looks better:

max_packet_interval
mean_packet_interval
min_packet_interval

OR

packet_interval_min
packet_interval_max
packet_interval_mean

I like the latter as in my head I kind of picture:
```
{
  "packet_interval": {
    "min": 1,
    "max" 100,
    "mean": 50
  }
}

pevma:
Sure.
If min/max/avg are added (or to be added) in then definitely the later makes more sense i think.

PR #2646, opened 2017-04-04
unix-socket: return failure on failure - v1

jasonish:
UnixManagerThreadInit needs to return a failure code if the socket
fails to initialize to avoid entering the UnixManager loop which
will continuously fail on the call to bind, as no listening
socket was setup.

This can occur when the socket fails to initialize due to a
permissions error and fatal init errors is not on.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/117
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/469

### DIFF
--- a/src/unix-manager.c
+++ b/src/unix-manager.c
@@ -914,7 +914,7 @@ static TmEcode UnixManagerThreadInit(ThreadVars *t, void *initdata, void **data)
         } else {
             SCLogWarning(SC_ERR_INITIALIZATION,
                     "Unable to create unix command socket");
-            return TM_ECODE_OK;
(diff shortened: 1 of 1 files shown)


inliniac:
Is there still a difference in behaviour between with and without init errors fatal now?

jasonish:
Yes. With init failures on it will exit. Otherwise the module will just fail, but Suricata will continue on - like it did before but without trying to accept on a non existent socket.

inliniac:
Merged through #2653, thanks Jason!

PR #2649, opened 2017-04-05
logging - non-blocking writes to socket - v2

jasonish:
Previous PR: #2603 

Related issue:
https://redmine.openinfosecfoundation.org/issues/2039

Changes from last PR:
- Log drops for eve in one location on exit (only if there are drops)
- Retry loop cleanup.
- If not running live, don't do a non-blocking write.

I looked at a stats counter, but it was non-trivial to one via the stats API to do the various entry points into the logger (ie: flow). Will revisit though.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/120
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/472

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -121,6 +121,57 @@ static int SCLogUnixSocketReconnect(LogFileCtx *log_ctx)
     return log_ctx->fp ? 1 : 0;
 }
 
+static int SCLogFileWriteSocket(const char *buffer, int buffer_len,
+        LogFileCtx *ctx)
+{
+    int tries = 0;
+    int ret = 0;
+    bool reopen = false;
+
+    if (ctx->fp == NULL && ctx->is_sock) {
+        SCLogUnixSocketReconnect(ctx);
+    }
+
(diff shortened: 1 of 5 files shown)


src/util-logopenfile.c
```diff
@@ -121,6 +121,63 @@ static int SCLogUnixSocketReconnect(LogFileCtx *log_ctx)
...
+    int ret = 0;
+    bool reopen = false;
+    int flags = 0;
+
+    if (ctx->fp == NULL && ctx->is_sock) {
+        SCLogUnixSocketReconnect(ctx);
+    }
+
+    /* Set flags for a non-blocking write if live. */
+    if (TimeModeIsLive()) {
```
inliniac: I would like to see this handled globally somehow. Seems wasteful to check this on every write. It won't ever change at runtime.

jasonish:
Closing, see https://github.com/inliniac/suricata/pull/2651

PR #2651, opened 2017-04-05
logging - non-blocking writes to socket - v3

jasonish:
Previous PR: #2649 

### Changes from last PR:
- Check for offline/live runmode during log file setup. As the time value for this is set after the outputs are initialized, one more global seemed to be the least intrusive way to get at this value in logging setup. Another option was to passive the SCInstance around some, but that would have been more intrusive.

### Previous Description

Changes from last PR:
- Log drops for eve in one location on exit (only if there are drops)
- Retry loop cleanup.
- If not running live, don't do a non-blocking write.

I looked at a stats counter, but it was non-trivial to one via the stats API to do the various entry points into the logger (ie: flow). Will revisit though.

### Prscript
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/121
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/473

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -720,6 +720,11 @@ static void OutputJsonDeInitCtx(OutputCtx *output_ctx)
 {
     OutputJsonCtx *json_ctx = (OutputJsonCtx *)output_ctx->data;
     LogFileCtx *logfile_ctx = json_ctx->file_ctx;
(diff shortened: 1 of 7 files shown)


inliniac:
Merged in #2655, thanks Jason!

PR #2645, opened 2017-04-04
TCP reassembly buffer v184.4

inliniac:
Switch over TCP engine to use streaming buffer API. Major rewrite and cleanup.

- Gets rid of the StreamMsg's and instead inspects the streaming buffer.
- Improve overlap detection accuracy
- simplify config: no more per size segment pools

PR #2656, opened 2017-04-07
TCP reassembly buffer v184.8

inliniac:
Rebase of #2645, plus some small adjustments.

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/89
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/597


pevma:
After some live runs I noticed (running on the same machine/traffic/config):  
- that with this PR (as opposed to master) Suri enters emergency mode now and then. ()where as usually never happens with master)
- that i have more drops and reassembly gaps 
than the usual/master stats.

Will continue to investigate...

gozzy:
@pevma, it looks strange - on our [inline mode] tests with IXIA this version behaved better that the one from master branch (AFAIR, there were no memcap drops or reassembly gaps in stats.log). Probably it's related to reassembly.segment-prealloc parameter, but I am not sure about that. Anyway, your tests differ from ours and obviously may give different results.

inliniac:
In IPS mode a bigger improvement is expected due to some changes in how it works.

inliniac:
Merged through #2673

PR #2658, opened 2017-04-10
template script: typo in app-layer setup script

jasonish:
Check for ed was failing, as it was actually looking for edx.

No prscript as its failing with a 502 Bad Gateway, but I think its clear that prscript wouldn't test this one.


inliniac:
Merged through #2661, thanks Jason.

PR #2660, opened 2017-04-11
stream gap handling - v5

jasonish:
Previous PR: #2619 

Changes from last PR:
- DNS TCP gap handling.

Requesting comments on this way of handling gaps in the app-layer. There are ifdef's of older code as I'm still using them for reference.. Hence this is just an RFC.

Protocols support gap handling:
- DNP3 - this is a simple record based protocol with a fixed start sequence.
- DNS (TCP) - Its very likely, though not always true that a TCP segment will be the start of a request/response. So probe it like it was new.

There is one stream test case that is broken from these changes tho.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/124
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/476

### DIFF
--- a/src/app-layer-dnp3.c
+++ b/src/app-layer-dnp3.c
@@ -242,10 +242,13 @@ static int DNP3CheckUserDataCRCs(const uint8_t *data, uint32_t len)
  *
  * \retval 1 if valid, 0 if not.
  */
-static int DNP3CheckStartBytes(const DNP3LinkHeader *header)
+static int DNP3CheckStartBytes(const uint8_t *input, uint32_t input_len)
 {
-    return header->start_byte0 == DNP3_START_BYTE0 &&
-        header->start_byte1 == DNP3_START_BYTE1;
+    if (input_len > 1 && input[0] == DNP3_START_BYTE0 &&
+            input[1] == DNP3_START_BYTE1) {
+        return 1;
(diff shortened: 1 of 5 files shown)


inliniac:
This will need a rebase after the stream updates.

On a gap, I would like to invoke the app-layer with a value of the number of skipped bytes. So protocols might be able to use to catch up. E.g. HTTP with a content-len field might see that the gap is in a file, but otherwise the HTTP tracking is unaffected.

inliniac:
The gap invocation could perhaps just be a extra call to AppLayerHandleTCPData with data NULL, data_len set to the gap size and flags set to STREAM_GAP.

jasonish:
> The gap invocation could perhaps just be a extra call to AppLayerHandleTCPData with data NULL, data_len set to the gap size and flags set to STREAM_GAP.

Yeah, I was wanting to get all that info into the app-layer, with the new data to skip the extra call, and state tracking required, but that would change the API across all app layer. So perhaps the extra call, and then the app-layer having to set a gap flag for the next call is better.

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2696

PR #2664, opened 2017-04-14
Added HPFeed Support

Philelis:
Added HPFeed Support adapted from [threatstream/suricata]( https://github.com/threatstream/suricata)
It is nice to use hpfeeds, so I tried to get the old setup from [threatstream/suricata]( https://github.com/threatstream/suricata) working on the new versions of suricata.
Hopefully you like it :)


inliniac:
Can you explain a bit what hpfeed is? I'm not familiar with it. A quick glance at the code suggests it sends alerts over the network. I didn't see references to encryption, but maybe I didn't look close enough.

Philelis:
Yeah Hpfeeds sends alerts over network to a server.
The description from Github [rep/hpfeeds](https://github.com/rep/hpfeeds) is: 
> hpfeeds is a lightweight authenticated publish-subscribe protocol that supports arbitrary binary payloads.

> Access to channels is given to so-called Authkeys which essentially are pairs of an identifier and a secret. The secret is sent to the server by hashing it together with a per-connection nonce. This way no eavesdroppers can obtain valid credentials. Optionally the protocol can be run on top of SSL/TLS, of course.

Authentication: 
>   -  Server sends a nonce per connection
>   -  Client sends id and sha1(nonce+authkey) hash
>   -  Server looks up authkey by id and checks hash
>   -  Server looks up subscribe/publish ACL for this client

Many people uses it for [threatstream/mhn](https://github.com/threatstream/mhn) which is a management platform for honeypots. They only have a really old version of suricata and so it would be nice that they can use every time the newest version without doing some extra lessons in programming.

inliniac:
Thank you. What reason do you see for this to be part of the Suricata process, instead of using EVE and post-processing that? In general my idea is to limit the number of output methods, and especially to be cautious with outputs that talk over the network.

cc: @jasonish

Philelis:
My intention was for using with the other project, they use it. Other Honeypots like Cowrie, Kippo or Dioanaea have support for it and I thought it might be a good idea for suricata. I know your sceptics and you don't have to implement it. 
Using EVE is a good solution, but it is only logging into files, isn't it? I have to check to use hpfeeds as extra part. I think its possible (for snort there is a working solution) and might be an addon for surricata without fix implementation in suricata.

jasonish:
I have to agree with Victor. I think "agents" like this are much better suited to be done as external processes.

It looks like it was done this way for Snort, https://github.com/threatstream/snort_hpfeeds - should be even easier to do for Suricata with the eve output already being in JSON.

inliniac:
Yeah I think an external solution is better also because currently the outputs actually block the packet path, so if the network is slow it would slow down the whole IDS.

PR #2657, opened 2017-04-08
Dev tls cipher suite v5 - DO NOT MERGE

fooinha:
- Style and code cleanup and review fixes.

- Load list from configuration file is not implemented. It will be after above questions are solved. 

----

- Allows tls app-layer rules to match TLS handshake cypher suites
Example rules:

```
alert tls any any -> any any
(msg:"SURICATA TLS CIPHER SUITE CLIENT"; 
tls_cipher_suite:client:TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,0xcca9;
sid:66666666; 
rev:1;
)
```
```
alert tls any any -> any any 
(msg:"SURICATA TLS CIPHER SUITE SERVER";
tls_cipher_suite:server:0xcca8;
sid:66666667;
rev:1;
)
```

- Custom tls logging for server cypher suite %C 

- Documentation

- First review round.

Replaces #2654.

fyi, @thus

PR #2665, opened 2017-04-14
Dev tls cipher suite v6

fooinha:
- Style and code cleanup and review fixes.

- Load list from configuration file (tls.cipher-suites.config) configured at suricata.yaml
   - Fallback to default values if configuration file load is not done.

----

- Allows tls app-layer rules to match TLS handshake cypher suites
Example rules:

```
alert tls any any -> any any
(msg:"SURICATA TLS CIPHER SUITE CLIENT"; 
tls_cipher_suite:client:TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,0xcca9;
sid:66666666; 
rev:1;
)
```
```
alert tls any any -> any any 
(msg:"SURICATA TLS CIPHER SUITE SERVER";
tls_cipher_suite:server:0xcca8;
sid:66666667;
rev:1;
)
```

- Custom tls logging for server cypher suite %C 

- Documentation

Replaces #2657 .

fyi, @thus


src/app-layer-ssl.c
```diff
@@ -265,17 +653,45 @@ static int TLSDecodeHandshakeHello(SSLState *ssl_state, uint8_t *input,
...
+        if (cipher_suites > 1) {
+            ssl_state->curr_connp->num_cipher_suites = cipher_suites / 2;
+            ssl_state->curr_connp->cipher_suites = SCMalloc(cipher_suites);
+            if (ssl_state->curr_connp->cipher_suites) {
+                memcpy(ssl_state->curr_connp->cipher_suites, input, cipher_suites);
+            }
+
+            input += cipher_suites;
+        }
+    }
```
inliniac: else if?

tls.cipher-suites.config
```diff
@@ -0,0 +1,326 @@
+0x0000,TLS_NULL_WITH_NULL_NULL
```
inliniac: it's simple, so that part I like. I wonder if using yaml would be more consistent and generally nicer? Might lead to easier parsing code too.

@thus  @jasonish what do you think?

src/app-layer-ssl.c
```diff
@@ -80,11 +80,348 @@ SCEnumCharMap tls_decoder_event_table[ ] = {
     { NULL,                          -1 },
 };
 
+// http://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4
+static SSLCipherSuite g_suites_default[] = {
```
inliniac: Can this live in it's own c or h file?

inliniac: With the file handling code I think would be best.

inliniac:
Replaced by #2671

PR #2667, opened 2017-04-18
rfc - rust integration - v3

jasonish:
RFC for incorporating Rust into the build.

Unlike other approaches, uses a single crate and a single static library.

On dist, the Rust dependencies are vendored (copied) and provided in the resulting tarball allowing the build to be done offline (cargo won't go out to the network to get deps).

Prscript failed due to a bad path in make distcheck, will be fixing up.

Prscript:
- You can watch build progress at https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/477
- You can watch build progress at https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/125
- Waiting for build completion
- Build successful for jasonish-pcap
- Build failure for jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/477

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1929,6 +1929,24 @@
     fi
     AM_CONDITIONAL([HAVE_PDFLATEX], [test "x$enable_pdflatex" != "xno"])
 
+# Cargo/Rust.
+  AC_PATH_PROG(HAVE_CARGO, cargo, "no")
+  AC_PATH_PROG(HAVE_RUSTC, rustc, "no")
+  if test "x$HAVE_CARGO" != "xno"; then
+     if test "x$HAVE_RUSTC" != "xno"; then
+        enable_rust=yes
+        AC_DEFINE([HAVE_RUST],[1],[Enable Rust language])
+	if test "x$enable_debug" = "xyes"; then
+	   RUST_LDADD="../rust/target/debug/libsuricata.a -lrt -ldl"
+	else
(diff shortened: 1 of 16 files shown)


rust/Cargo.lock
```diff
@@ -0,0 +1,21 @@
+[root]
+name = "suricata"
+version = "4.0.0-dev"
```
inliniac: Would there be a way to make this dynamic? E.g. be set based on the configure version? Perhaps it would mean this file would be checked in as a .in file?

configure.ac
```diff
@@ -1929,6 +1929,30 @@
     fi
     AM_CONDITIONAL([HAVE_PDFLATEX], [test "x$enable_pdflatex" != "xno"])
 
+# Cargo/Rust.
```
inliniac: indent looks off

rust/Cargo.lock
jasonish: Doing it as a .in file works, the main problem being that our current version, "4.0dev" will cause Cargo to error out:
```
Caused by:
  cannot parse '4.0' as a semver for key `package.version`
```
The closest thing to our version that will parse is 4.0.0-dev.

inliniac: I think we could use that in configure then, right? Haven't tried, but not expecting side effects. Unless it rejects the dash for some reason.

inliniac:
I like the simplicity.

jasonish:
Comments addressed in https://github.com/inliniac/suricata/pull/2670

PR #2668, opened 2017-04-19
TLS logging for session resumption 1.5

rtkrruvinskiy:

was never seen.

Replaces https://github.com/inliniac/suricata/pull/2602


rtkrruvinskiy:
@inliniac I've made the changes you suggested in the previous PR. Thanks!

inliniac:
Merged through #2672, thanks Ray. I did rewrite the commit message. https://github.com/inliniac/suricata/pull/2672/commits/753997310919cc041ac2b4058a7a46037bc2d613

See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits

rtkrruvinskiy:
Thanks, Victor. I'll do better on the commit message next time.

PR #2669, opened 2017-04-19
Call LiveGetShortName on correct variable

rtkrruvinskiy:
https://github.com/inliniac/suricata/commit/bb2d8a713382984b90b20d425230f32ce4106b3e introduced
the `dev` variable and replaced referenced to `live_dev` with `dev`, except in one place.
As a result, suricata can now log '[ERRCODE: SC_ERR_INVALID_VALUE(130)] - Name of device should not be null'
on start-up with pfring.

https://redmine.openinfosecfoundation.org/issues/2097


inliniac:
Merged through #2672, thanks Ray. I did rewrite the commit message. https://github.com/inliniac/suricata/pull/2672/commits/0dc57c937e3a01dfe3022bd64e4db64caea1a019

PR #2670, opened 2017-04-19
rfc - rust integration - v5

jasonish:
Previous PR: #2667 

Changes from last PR:
- Indent fixup in configure.ac.
- Cargo will now get version from autoconf. Required using a version of 4.0.0-dev.
- Configure check for cargo-vendor. If Rust is enabled, fail make dist of cargo-vendor is not installed.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/129
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/481

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1,0 +1,21 @@
+[root]
+name = "suricata"
+version = "4.0.0-dev"
+dependencies = [
+ "libc 0.2.21 (registry+https://github.com/rust-lang/crates.io-index)",
+ "nom 2.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "libc"
+version = "0.2.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "nom"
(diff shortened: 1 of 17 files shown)


rust/Cargo.lock
```diff
@@ -0,0 +1,21 @@
+[root]
+name = "suricata"
+version = "4.0.0-dev"
```
inliniac: How about this version?

jasonish:
Closing, see: https://github.com/inliniac/suricata/pull/2676

rust/Cargo.lock
jasonish: Cargo.lock is actually a generated file. Its generated from Cargo.toml. I'm not sure exactly when it gets updated tho. Will have to look into this more closely.

The lock file allows us to pin the dependency versions for reproducible builds, so I think its worthwhile using. There is some argument on whether or not it should even go into version control.  But for an end-user app, as opposed to a library it seems to make sense to check it in. That way everyone is building against the same dependencies.

PR #2674, opened 2017-04-23
Metadata and source/target identification

regit:
Source and destination IP are choozen from the flow and this is not corresponding to the nature of the attack. There is in fact no way to know in an alert which from source and destination is the target of the attack. This patchset is adressing this by updating the way metadata are used. The special keyword `target`is introduced and it is used to indicate which side of the flow (`client`or `server`) is the target of the attack.

IDMEF has a definition of Source and Target which is the one defined here. So the Prelude code has been updated to invert src and dest IP if necessary (as Source was src and Target dest by default).
 
On EVE side, if the target metdata is defined, the output is now the following:
```
  "alert": {
    "action": "allowed",
    "gid": 1,
    "signature_id": 1,
    "rev": 1,
    "signature": "connection to home",
    "severity": 3,
    "source": {
      "ip": "2001:31d0:000a:f68a:0000:0000:0000:0001",
      "port": 80
    },
    "target": {
      "ip": "2a01:0e34:ee97:b130:c685:08ff:dab3:c9c8",
      "port": 48390
    }
```
As a work on metadata was done, this patchset also adds an option to log metadata key -> value in the alert event. With this information a user will be able to better categorize the events. This is even more true if we consider the enhanced version of Emerging threats ruleset where a lot of metadata have been added.
For instance we have, rules in this ruleset looking like that
```
alert http $EXTERNAL_NET $HTTP_PORTS -> $HOME_NET any (msg:"ET WEB_CLIENT Obfuscated Javascript // ptth"; flow:from_server,established; content:"200"; http_stat_code; file_data; content:"//|3a|ptth"; classtype:bad-unknown; sid:2012325; rev:4; metadata:affected_product Web_Browsers, affected_product Web_Browser_Plugins, attack_target Client_Endpoint, deployment Perimeter, tag Web_Client_Attacks, target client, signature_severity Major, created_at 2011_02_21, updated_at 2016_07_01;)
```
Here `target client` has been added. With that metadata, we have the following alert:
```
  "alert": {
    "action": "allowed",
    "gid": 1,
    "signature_id": 2012325,
    "rev": 4,
    "signature": "ET WEB_CLIENT Obfuscated Javascript // ptth",
    "category": "Potentially Bad Traffic",
    "severity": 2,
    "source": {
      "ip": "64.27.26.79",
      "port": 80
    },
    "target": {
      "ip": "192.168.2.5",
      "port": 1148
    },
    "metadata": {
      "target": "client",
      "tag": "Web_Client_Attacks",
      "deployment": "Perimeter",
      "attack_target": "Client_Endpoint",
      "affected_product": "Web_Browsers"
    }
  },
```

As Prelude build was not issuing a warning, this patchset is also fixing this.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/253
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/35

### DIFF
--- a/doc/userguide/rules/meta.rst
+++ b/doc/userguide/rules/meta.rst
@@ -173,9 +173,10 @@ format of priority is:
 Metadata
 --------
 
-Suricata ignores the words behind meta data.  Suricata supports this
-keyword because it is part of the signature language.  The format is:
+Suricata partially ignores the words behind meta data. Suricata only looks for
+the target keyword that is used to determine which side of the IP connection is
+the target of the attack.  The format is:
 
 ::
 
-  metadata:......;
+  metadata:...., target [client|server], ....;
--- a/src/alert-prelude.c
+++ b/src/alert-prelude.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2014 Open Information Security Foundation
+/* Copyright (C) 2007-2017 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -253,7 +253,8 @@ static int EventToImpact(const PacketAlert *pa, const Packet *p, idmef_alert_t *
  *
  * \return 0 if ok
  */
-static int EventToSourceTarget(const Packet *p, idmef_alert_t *alert)
+static int EventToSourceTarget(const PacketAlert *pa, const Packet *p,
+                               idmef_alert_t *alert)
 {
     int ret;
     idmef_node_t *node;
@@ -265,6 +266,8 @@ static int EventToSourceTarget(const Packet *p, idmef_alert_t *alert)
     static char saddr[128], daddr[128];
     uint8_t ip_vers;
     uint8_t ip_proto;
+    uint16_t sp, dp;
+    uint8_t invert = 0;
 
     SCEnter();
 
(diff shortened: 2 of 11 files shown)


inliniac:
As discussed in IRC recently, I think it would make sense for metadata to be used a bit less freely. E.g. instead of
<pre>
metadata:affected_product Web_Browsers
</pre>
where is affected_product is the key, use something like:
<pre>
metadata:tag affected_product Web_Browsers
</pre>
So that we don't give up the entire metadata field to this free-form key-value pair feature.

There is existing metadata use in Snort's rulesets and there may be in other rulesets now and it the future.

regit:
Discussion on IRC was not clear. This code only parse and use the `target` keyword. The rest of metadata key value is what is already used in Emerging Threats enhanced ruleset. But I'm ok to use `tag` concept.

inliniac:
I'm a bit confused by "This code only parse and use the target keyword". Whats this then?
https://github.com/inliniac/suricata/pull/2674/files#diff-e2a291d82d2cff36c51007fafaa79b86R291

regit:
@inliniac this is just for optional metadata logging. The only one that has a functional meaning is `target` the rest is free form.

inliniac:
That was my point. I don't think thats a good idea in it's current form.

regit:
so, would you accept if I'm using `tag target client` instead of `target client` or do you want another change ?

inliniac:
On 25-04-17 10:56, Eric Leblond wrote:
> so, would you accept if I'm using |tag target client| instead of |target
> client| or do you want another change ?

there are 2 different things:

1. target keyword. I'm ok with that, as it's limited and clear.

2. free form key/value pair stuff. I'm not ok with that in it's current
form. By allowing (or actually encouraging) users to use *any*
key/value, we will get into trouble with this later if we want to add
more special keywords to metadata. This is why I'm proposing to use a
special key, for example 'tag', as a prefix to the key/value pair.

Example: metadata:target client, tag affected_product Web_Browsers, tag
otherkey othervalue, etc.

This way we don't claim the whole metadata 'space'.

Not sure if 'tag' is the proper key, consider it more like an example.
Maybe a better name can be found.

regit:
Ok there is a misunderstanding here. The metadata I expose in the example in this PR is taken from something that is published by Emerging Threats. It is their enhanced ruleset that is available at https://rules.emergingthreats.net/open/suricata-1.3-enhanced/emerging.rules.tar.gz. So, we as Suricata do not claim anything in this PR but the `target` keyword. The rest of the metadata is just an example of how this target work can integrate with other people usage of metadata.

inliniac:
On 25-04-17 11:19, Eric Leblond wrote:
> Ok there is a misunderstanding here. The metadata I expose in the
> example in this PR is taken from something that is published by Emerging
> Threats. It is their enhanced ruleset that is available at
> https://rules.emergingthreats.net/open/suricata-1.3-enhanced/emerging.rules.tar.gz.
> So, we as Suricata do not claim anything in this PR but the |target|
> keyword. The rest of the metadata is just an example of how this target
> work can integrate with other people usage of metadata.

I disagree. If we start supporting metadata as a free-form key value
pair keyword that we log to the output, we clearly define what the
keyword is. Thats a much stronger position that just saying 'it is just
and example'. This is why I'm pushing back on this. I think we need to
very careful about declaring metadata to have this specific meaning and
format. I'm aware that ET uses this, but perhaps they can change their
ways. I'm also aware that TALOS uses it quite differently, and we can't
ignore that either.

regit:
On metadata, I was following the documentation of snort: http://manual-snort-org.s3-website-us-east-1.amazonaws.com/node31.html#SECTION00448000000000000000
Do we want to do something else ?
For me this is a good definition and we should maybe put that in the documentation so it is clear for everyone.

jasonish:
From Snort's documentation I think there is a safe way to do it. I don't think we should ask the rule writers to change their ways to get the metadata into a rule output, but we can suggest best practice.  Like the Snort documentation says, its
 metadata:key1 value1, key2 value2
so we parse up to the first white space as key, then up to the comma as value.

One point, we cannot error out on loading a rule if metadata parsing fails though.

Or to be safer, just split the metadata at each , and store as a list and forget about parsing as key/value.

jasonish:
But I'm not sure if I'm in favour of the other changes to eve output at this time.  Like the "source" and "target" objects, especially if src_ip and friends are missing. At this point, I think we've set those in stone. Not to say they aren't nice. It might just be too much.

inliniac:
On 25-04-17 17:03, Jason Ish wrote:
> From Snort's documentation I think there is a safe way to do it. I don't
> think we should ask the rule writers to change their ways to get the
> metadata into a rule output, but we can suggest best practice. Like the
> Snort documentation says, its
> metadata:key1 value1, key2 value2
> so we parse up to the first white space as key, then up to the comma as
> value.
> 
> One point, we cannot error out on loading a rule if metadata parsing
> fails though.
> 
> Or to be safer, just split the metadata at each , and store as a list
> and forget about parsing as key/value.

I think all of this is leading to a really poor user experience. I wish
we could just introduce a new keyword for this purpose, where we can set
clear rules and validate input.

Btw, Snort/TALOS usage is not consistent. Their use of 'policy' looks
like 'policy balanced-ips alert'.

jasonish:
> Btw, Snort/TALOS usage is not consistent. Their use of 'policy' looks like 'policy balanced-ips alert'.

Thats consistent to me. The key is "policy" and the value is "balanced-ips alert". Its not the metadata parsers job to interpret the value. That is left up to users of the specific metadata, as in the value should be opaque to the metadata parser.

inliniac:
My preference would be to implement this in proper rule keywords. This means that the rule writers have to add support for it, which they can do in a 4.0+ version of the ruleset.

jasonish:
> My preference would be to implement this in proper rule keywords. This means that the rule writers have to add support for it, which they can do in a 4.0+ version of the ruleset.

Yes, I'd agree with that. My only real objection was adding a new convention to the content of metadata.

regit:
OK, looks like I will use MetaData as new keywords :P

regit:
I propose to use  `idmef` as new keyword with the following syntax:
```
idmef: target [client|server][, source [client|server]]
```
@inliniac good with you ?

inliniac:
Can you explain the logic of picking idmef?

regit:
@inliniac the IDMEF RFC is the one defining source and target. And the alert enhancement is in fact a proper implementation of this RFC. Hence the naming. Also IDMEF has some nice extensions in the definition of source and target that we could use later (like potentially doing target: rcpt_to).

inliniac:
Where is this rfc?

regit:
@inliniac https://tools.ietf.org/html/rfc4765#section-4.1

inliniac:
Are you planning to add more of the fields they define to this new idmef keyword? It seems in general there is overlap between our regular keywords like msg, classtype, etc and what the idmef spec has.

Wonder if it makes more sense to just make these standalone keywords like 'target: [client|server]; source: [client|server];'

regit:
I did not discuss adding more with @ToToL but I think it could be interesting. There is some good idea in the RFC that we could pick.

inliniac:
What about the rest of my comment? As much of what idmef is already part of rules, would it make sense to add source, target and more later as standalone keywords?

cc @jasonish

jasonish:
I'm not really a fan of adding "idmef", it feels to me like it's creating precedent for adding a keyword for the spec we borrow ideas from. 

If it allowed for an XML encoded string representing the idmef spec then it would make more sense I think.

regit:
@inliniac sorry bad wording on my side. I think we could benefit of adding other keywords linked to IDMEF rfc. That's the main reason why I did choose idmef as keyword.

regit:
@jasonish not fan of XML in the signature on my side. My point is that it is one of our single RFC so why not taking the name.

inliniac:
Not a fan of the idmef approach either. I think we can drop the prefix as it confuses and breaks away from our normal syntax. I think it would be interesting to add a page on in the documentation about how our rule keywords map to the idmef rfc.

regit:
I don't get the point on breaking normal syntax but if you want I can move target to the top of signature. This would be enough for my actual purpose.

jasonish:
> @jasonish not fan of XML in the signature on my side. My point is that it is one of our single RFC so why not taking the name.

That was a bit tongue in cheek, but I think thats the only case where idmef makes sense. Otherwise it seems to be there to give credit to idmef, but provide no functional value.  idmef.target, idmef.source would probably make better keywords. But if they don't conflict why not just target, and source. Or perhaps some other naming that fits us better. After all we're not implementing IDMEF, just taking the concept right?

jasonish:
Are both target and source needed? If there was only target, is there any case where source could not be inferred?

regit:
@jasonish Not now but I hope I will have the time to do something like:
```
source: src, target: dest, smtp.rcpt_to
```
So we will build a more complex source or target based on the event.

regit:
OK so to sum up the discussion, following your proposal and my last point, let's do:
```
source: [src_ip|dest_ip]
target: [src_ip|dest_ip]
```
That will possibly evolve to something like
```source: src_ip, smtp.from```

Good with you ?

ToToL:
After reading all of this I agree with the conclusion : Even I'm a huge defenser of IDMEF, I don't think it's a good idea to have idmef in a keyword if it's just because you do it this way to do as the philosophy of IDMEF (maybe some other definitions do this in the same way)

But it may be interesting to select the preset of output : 
* idmef : source is the attacker, target is the goal of the attacker
* network : source is the source of the stream, target is the target of the stream
* etc.
=> Maybe this will be interesting also for other fields.

jasonish:
In what case would you not want the "idmef" case? If convention can rule out ambiguity, we should just stick with convention.

regit:
Updated in #2761

PR #2675, opened 2017-04-28
rust - dns - v5

jasonish:
A new Rust DNS implementation that is stateless. Builds on rust-core PR, https://github.com/inliniac/suricata/pull/2676.

Rust must be enabled for this DNS parser to be used. If not enabled, the current C implementation is used.

Current issues:
- TCP not done, so some tests fail.
- Lua not done.

On the other hand it handles the dns stress test with ease, logging all 340000+ DNS requests within 9 seconds.

TODOs:
- TCP (should be pretty trivial)
- Finish Lua. I'm thinking of taking a different approach and exposing some Lua to Rust. Will likely be less work than relaying various stuff from Rust -> C -> Lua.


jasonish:
See https://github.com/inliniac/suricata/pull/2678

PR #2538, opened 2017-02-05
Experiments/rust/v5.0

inliniac:
Playing some more with data store api. Now with recursive stores:

Add a store for each 'Xid' and push the xid to a queue:
```rust
        self.store.pushback_u32(NfsStorageId::XidQueue as u32, r.hdr.xid);
        match self.store.get_store_ref_create(r.hdr.xid) {
            Some(store) => {
                store.set_u32(NfsStorageId::Procedure as u32, r.procedure);
            },
            _ => {},
        }

```

Then in C for each Xid in the queue, get it's store, do something with it and drop it:
```c
    Store *data_store = r_nfstcp_getstore(state);
    if (data_store != NULL) {
        uint32_t xid;
        while (r_popfront_u32(data_store, 7, &xid) == 1) {
            Store *nested_store;
            if (r_getstore(data_store, xid, &nested_store) == 1) {
                uint32_t procedure;
                if (r_getu32(nested_store, 6, &procedure) == 1) {
                    if (procedure == 3) {
                        uint8_t *data;
                        uint32_t len;
                        if (r_getdata(nested_store, 8, &data, &len) == 1) {
                            char *c = BytesToString(data, len);
                            if (c != NULL) {
                                SCLogNotice("NFSv3 LOOKUP %s", c);
                                SCFree(c);
                            }
                        }
                    } else if (procedure == 8) {
                        uint8_t *data;
                        uint32_t len;
                        if (r_getdata(nested_store, 9, &data, &len) == 1) {
                            char *c = BytesToString(data, len);
                            if (c != NULL) {
                                SCLogNotice("NFSv3 CREATE %s", c);
                                SCFree(c);
                            }
                        }
                    }
                }
                r_dropstore(data_store, xid);
            }
        }
    }
```

Still have to figure out how to expose the recursion to Lua.

Also, currently this mixes 'fixed' id's with dynamic ones (xid), so that needs some more thought as well. In profiling this still hardly shows up.

PR #2676, opened 2017-04-28
rust: core integration - v6

jasonish:
Previous PR: #2670 

Changes from last PR:
- libjansson wrapper
- Rust must be explicitly enabled.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/134
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/486

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,8 +47,10 @@ matrix:
     - os: linux
       compiler: gcc
       env:
-        - NAME="linux,gcc,cocci"
+        - NAME="linux,gcc,cocci,rust"
         - *default-cflags
+        - ENABLE_RUST="yes"
+        - ARGS="--enable-rust"
       addons:
         apt:
           sources:
@@ -137,6 +139,11 @@ script:
     fi
 
 before_install:
+  - |
+    if [[ "$ENABLE_RUST" == "yes" ]]; then
+        curl https://sh.rustup.rs -sSf | sh -s -- -y
+    fi
+  - export PATH=$HOME/.cargo/bin:$PATH
   - |
     if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
         brew update
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,7 +5,7 @@ ACLOCAL_AMFLAGS = -I m4
 EXTRA_DIST = ChangeLog COPYING LICENSE suricata.yaml.in \
              classification.config threshold.config \
              reference.config
-SUBDIRS = $(HTP_DIR) src qa rules doc contrib scripts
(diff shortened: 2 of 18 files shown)


inliniac:
So the proposed file layout is:


rust/
rust/src/*.rs <- common files
rust/src/dns/*.rs <- dns app layer
rust/src/_applayer_/*.rs other app layer

I thought before you wanted to do a flat file layout, but maybe I misunderstood that.

inliniac:
I'm not yet convinced the exposing of the libjansson api like this is the best idea. I'd like to minimize the use of C api's form the Rust side of things.

I wonder if an approach like the data store in https://github.com/inliniac/suricata/pull/2538 that can be used from C makes for a cleaner boundary. The logger would then live in C however.

Another way I guess would be to fully construct a JSON/log message in rust, and only call LogFileCtx::Write from Rust on that message.

jasonish:
> I'm not yet convinced the exposing of the libjansson api like this is the best idea. I'd like to minimize the use of C api's form the Rust side of things.

To be honest, this is where Rust's FFI excels, calling into C.

> 
> I wonder if an approach like the data store in #2538 that can be used from C makes for a cleaner boundary. The logger would then live in C however.

It hasn't clicked with me how that would work with DNS without making it more complex. Also, if the logging was going to be done from C, to be honest, its almost easier to do the decoding in C.  The more bridge code there, the less valuable using Rust becomes I think.

> 
> Another way I guess would be to fully construct a JSON/log message in rust, and only call LogFileCtx::Write from Rust on that message.

This would be an OK option, some code duplication in logging but lets the Rust log stuff that is decoded in Rust, etc.

I've taken the same approach with Lua, by passing the Lua state into Rust.  IMO its so much cleaner, simpler, than defining data stores for passing data back (again, opinions may vary here). Probably easier to teach as well when it comes to training as its using familiar APIs from the C side already.

jasonish:
> So the proposed file layout is:
> 
> rust/
> rust/src/.rs <- common files
> rust/src/dns/.rs <- dns app layer
> rust/src/applayer/*.rs other app layer
> 
> I thought before you wanted to do a flat file layout, but maybe I misunderstood that.

No, not a flat file layout, but a single crate with modules. I wasn't keen on the multi-create layout. 

So given the above, the "common" files is the top module, or core module. Might make sense to move that one level down? Common, core, or whatever? Then you have the DNS module, and so on.

jasonish:
Out of date. The DNS branch contains this branch in its earlier commits. See:
https://github.com/inliniac/suricata/pull/2708

PR #2678, opened 2017-04-28
rust - dns - v7

jasonish:
Previous PR: #2675

Changes from last PR:
- TCP support.
- Passes all unit tests.

TODOs:
- Lua.

One prscript test fails in a pcap. But I'm not sure what. Could be Lua.

Prscript:
- Build failure for jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/135
- Build successful for jasonish


rust/gen-c-headers.py
```diff
@@ -0,0 +1,130 @@
+#! /usr/bin/env python2
```
inliniac: Can you add a description to this? What it does, how to use it, etc?

rust/src/dns/mod.rs
```diff
@@ -18,6 +18,12 @@
 use log::*;
 use conf;
 
+pub mod parser;
+pub use self::parser::*;
+
+pub mod dns;
+pub use self::dns::*;
```
inliniac: whats this 'self' use here?

rust/src/dns/dns.rs
```diff
@@ -0,0 +1,66 @@
+/* Copyright (C) 2017 Open Information Security Foundation
```
inliniac: commit message is a bit strange, with ref to dhcp

src/app-layer-dns-tcp-rust.c
```diff
@@ -0,0 +1,203 @@
...
+#include "rust-dns-dns.h"
+
+static void RustDNSTCPParserRegisterTests(void);
+
+static int RustDNSTCPParseRequest(Flow *f, void *state,
+        AppLayerParserState *pstate, uint8_t *input, uint32_t input_len,
+        void *local_data)
+{
+    SCEnter();
+    SCReturnInt(rs_dns_parse_request_tcp(f, state, pstate, input, input_len,
```
inliniac: iirc there were some issues with calling a function in a SCReturn macro like this. Don't remember exactly, but I think there were cases in which is wasn't called.

rust/src/dns/mod.rs
jasonish: Just left overs from getting modules to work I think. I can remove them by doing "use dns::dns::..." in the modules that need rust/src/dns/dns.rs...

jasonish:
See https://github.com/inliniac/suricata/pull/2686

PR #2686, opened 2017-05-02
rust - dns - v8

jasonish:
Previous PR: #2678 

Changes from last PR:
- Complete Lua support
- Added description to header generator.
- Generated headers now carry copyright and warning about editing.
- Remove "self" usage.
- Fix DHCP in commit message.
- Don't call function in macro.

TODOs:
- Some less common types are not logged.
- Log filter is not done.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/141
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/493


rust/src/dns/dns.rs
```diff
@@ -0,0 +1,801 @@
...
+    pub answers: Vec<DNSAnswerEntry>,
+    pub authorities: Vec<DNSAnswerEntry>,
+}
+
+#[derive(Debug)]
+pub struct DNSTransaction {
+    pub id: u64,
+    pub request: Option<DNSRequest>,
+    pub response: Option<DNSResponse>,
+    pub logged: u32,
```
inliniac: would it make sense to define our own type in rust so that if we change things on the C side we can easily update it for all Rust parsers? Same would apply to C itself btw.

jasonish: I'm not sure what type you are referring to?

inliniac: In this case just the u32 of logged. But in general I wonder about this: would it make sense to define types for these kinds of fields in a common file.

rust/src/dns/log.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+    else {
+        return "".to_string();
+    }
+}
+
+///  Log the SSHPF in an DNSAnswerEntry.
+fn dns_log_sshfp(js: &Json, answer: &DNSAnswerEntry)
+{
+    // Need at least 3 bytes - TODO: log something if we don't?
```
inliniac: maybe something like "malformed" ?

src/rust-core.h
```diff
@@ -0,0 +1,27 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/*
+ * DO NOT EDIT. This file is automatically generated.
```
inliniac: can you add the command that create a file to this message?

inliniac:
I first compiled this succesfully, but later I got:
<pre>
  CCLD     suricata
/usr/bin/ld: ../rust/target/release/libsuricata.a(std-6a154fe0.0.o): undefined reference to symbol 'roundf@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libm.so.6: error adding symbols: DSO missing from command line
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Makefile:1194: recipe for target 'suricata' failed
</pre>
Using clang-3.8 on Ubuntu 16.04 with Ubuntu's Rust (rustc 1.7.0).

Perhaps we need -lm, at least thats what man roundf suggests.

With ASAN enabled it works and I see this in my ldd:
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f67e4a8e000)
So I guess asan adds -lm.

src/rust-core.h
inliniac: (and to other similar messages)

inliniac:
I've not yet checked myself, but I wonder if we can use doxygen for the rust side as well and what that means to function comment blocks.

jasonish:
> So I guess asan adds -lm.

Will add -lm to the extra flags added on Linux. Something that I'll need to do is test build on FreeBSD, Windows, and others to see what we need to do there as well.

inliniac:
On 09-05-17 17:36, Jason Ish wrote:
>     So I guess asan adds -lm.
> 
> Will add -lm to the extra flags added on Linux. Something that I'll need
> to do is test build on FreeBSD, Windows, and others to see what we need
> to do there as well.

Perhaps better to do that from configure so that we can just test for
the lib?

jasonish:
Currently doing this:
https://github.com/jasonish/suricata/blob/rust-core-8/configure.ac#L217

But that should probably be extended to test for those libs as well.

jasonish:
> I've not yet checked myself, but I wonder if we can use doxygen for the rust side as well and what that means to function comment blocks.

I haven't seen any evidence that Doxygen supports Rust, unfortunately.  There is "cargo doc" that uses the Rust documentation generate tool.  But that means no cross linking with the C docs.

rust/src/dns/log.rs
jasonish: I'm not sure if events should be logged from the logger - if logging is turned off it will never be logged. That would mean moving validation into post-parsing. Since it parse fully valid, but still not be a valid sshfp.

Here's a thought.  If we can't log it, maybe do something like:
```
{
    ...
    "sshfp": {
        "invalid": true
    }
}
```

Or even just log it as empty...?

inliniac:
On 09-05-17 18:07, Jason Ish wrote:
> *@jasonish* commented on this pull request.
> 
> 
> In rust/src/dns/log.rs
> <https://github.com/inliniac/suricata/pull/2686#discussion_r115531189>:
> 
>> +                       addr[10],
> +                       addr[11],
> +                       addr[12],
> +                       addr[13],
> +                       addr[14],
> +                       addr[15]);
> +    }
> +    else {
> +        return "".to_string();
> +    }
> +}
> +
> +///  Log the SSHPF in an DNSAnswerEntry.
> +fn dns_log_sshfp(js: &Json, answer: &DNSAnswerEntry)
> +{
> +    // Need at least 3 bytes - TODO: log something if we don't?
> 
> I'm not sure if events should be logged from the logger - if logging is
> turned off it will never be logged. That would mean moving validation
> into post-parsing. Since it parse fully valid, but still not be a valid
> sshfp.

Right. So the validation should ideally be done in the parser instead.
Then we could set events to match on.

> Here's a thought. If we can't log it, maybe do something like:
> 
> |{ ... "sshfp": { "invalid": true } } |

I like this in addition to doing an event.

> Or even just log it as empty...?

jasonish:
> Log filter is not done.

Actually, as the C code is still here, filtering is already handled.  Its something that would need to move into Rust if/when the C DNS logger gets removed completely.

inliniac:
On 09-05-17 18:11, Jason Ish wrote:
>     Log filter is not done.
> 
> Actually, as the C code is still here, filtering is already handled. Its
> something that would need to move into Rust if/when the C DNS logger
> gets removed completely.

Maybe good to create a QA test that will fail if no filtering is in
place? Then we won't forget :)

rust/src/dns/dns.rs
jasonish: For the rust side, could go as simple as:
```
type Logger = u32;
```
At least then if we ever change its real type, the compiler will fail until we fix up all uses of it.

Or we could go one step further:
```
pub struct Logged {
    logged: u32,
}

impl Logged {

    pub fn set_logged(&mut self, logger: u32) {
        self.logged |= logger;
    }
    
    pub fn is_logged(&self, logger: u32) -> bool {
        if self.logged & logger != 0 {
            true
        } else {
            false
        }
    }
}
```
Then its self contained.

Also want to look at making this a trait of a transaction, but being more idiomatic Rust may be a work in progress..

inliniac:
On 09-05-17 18:33, Jason Ish wrote:
> *@jasonish* commented on this pull request.
> 
> 
> In rust/src/dns/dns.rs
> <https://github.com/inliniac/suricata/pull/2686#discussion_r115537183>:
> 
>> +}
> +
> +#[derive(Debug)]
> +pub struct DNSResponse {
> +    pub header: DNSHeader,
> +    pub queries: Vec<DNSQueryEntry>,
> +    pub answers: Vec<DNSAnswerEntry>,
> +    pub authorities: Vec<DNSAnswerEntry>,
> +}
> +
> +#[derive(Debug)]
> +pub struct DNSTransaction {
> +    pub id: u64,
> +    pub request: Option<DNSRequest>,
> +    pub response: Option<DNSResponse>,
> +    pub logged: u32,
> 
> For the rust side, could go as simple as:
> 
> |type Logger = u32; |
> 
> At least then if we ever change its real type, the compiler will fail
> until we fix up all uses of it.
> 
> Or we could go one step further:
> 
> |pub struct Logged { logged: u32, } impl Logged { pub fn set_logged(&mut
> self, logger: u32) { self.logged |= logger; } pub fn is_logged(&self,
> logger: u32) -> bool { if self.logged & logger != 0 { true } else {
> false } } } |
> 
> Then its self contained.

I like this. Wonder if it will lead to the same binary result.

> Also want to look at making this a trait of a transaction, but being
> more idiomatic Rust may be a work in progress..

Still need to read up on the trait concept.

jasonish:
> I like this. Wonder if it will lead to the same binary result.

In release mode it looks like it inlines, creating an identical assembly output. In debug mode it calls out to the function.

jasonish:
> I like this. Wonder if it will lead to the same binary result.

Oh, the Rust looks something like this:

https://github.com/jasonish/suricata/commit/008c382b202bb9474b740e9fe9bca6fe3b16b0cc

Appears the 0 abstraction overhead holds up here.

inliniac:
On 09-05-17 19:28, Jason Ish wrote:
>     I like this. Wonder if it will lead to the same binary result.
> 
> Oh, the Rust looks something like this:
> 
> jasonish/suricata@008c382
> <https://github.com/jasonish/suricata/commit/008c382b202bb9474b740e9fe9bca6fe3b16b0cc>
> 
> Appears the 0 abstraction overhead holds up here.
> 

Nice, I like it.

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2703

PR #2689, opened 2017-05-03
dhcp - dhcp app-layer - v1.ish

jasonish:
This PR takes Tom's latest DHCP and add some updates to it:

1. Make DNS and router IP addresses lists. Per https://www.iana.org/assignments/bootp-dhcp-parameters/bootp-dhcp-parameters.xhtml, these are referred to in the plural.  I'd like to check others.

2. I was getting duplicate log messages. This was due to the request being on the flow 0.0.0.0 -> 255.255.255.255, and the response being 10.16.1.1 -> 10.16.1.109 and the global state. From the app-layer perspective we still have 2 states, but with a single transaction owned by both states. So it ends up getting logged 2x. I think its a bit of a race condition, but for me it was almost always being logged 2x.

My fix is to actually allocate a state for each state that just wraps the global state and give each TX a pointer to this state. This allows each tx to be owned by a specific state. One thing to point out is on response we change the state that owns the TX, this is to help with the logging.

3. Rework how logging is done so we following the scheme of other request/response protocols where the source address is always the client.

Outstanding issues:

==

Given a capture like:
```
Client --> DHCP Discover
Server --> DHCP Offer
Client --> DHCP Request
Server --> DHCP ACK
```
The discover/offer and request/ack get the same transaction ID, so the request overwrites the data already recorded in the discover...  2 thoughts here. Ignore discovers for now. Its really the request/ack we are interested in. Or extend the logic handling so the request makes a new transaction, so the transaction will need a type in addition to the tx id.  For the 99% use case I don't see much value here.

==

I think I really want a "brief" logging method:

https://gist.github.com/jasonish/38ce5ec4979e83f950eea95a824aa1be

Where the current format would be the extended format.  Again, I think its about meeting 80% of the uses cases with a simpler log format.

==

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/145
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/497

### DIFF
--- a/src/app-layer-dhcp.c
+++ b/src/app-layer-dhcp.c
@@ -91,17 +91,17 @@ static void DHCPTxFree(DHCPGlobalState *dhcp, void *tx, uint32_t locked)
     }
 }
 
-static DHCPTransaction *DHCPTxAlloc(DHCPGlobalState *dhcp, uint32_t xid)
+static DHCPTransaction *DHCPTxAlloc(DHCPState *dhcp, uint32_t xid)
 {
     DHCPTransaction *tx;
 
     /* limit outstanding transactions */
-    if (unlikely(dhcp->transaction_count > dhcp_config_max_transactions)) {
+    if (unlikely(dhcp->global->transaction_count > dhcp_config_max_transactions)) {
         /* toss out the oldest */
-        tx = TAILQ_FIRST(&dhcp->tx_list);
+        tx = TAILQ_FIRST(&dhcp->global->tx_list);
         if (likely(tx != NULL)) {
-            TAILQ_REMOVE(&dhcp->tx_list, tx, next);
-            DHCPTxFree(dhcp, tx, 1);
+            TAILQ_REMOVE(&dhcp->global->tx_list, tx, next);
+            DHCPTxFree(dhcp->global, tx, 1);
         }
     }
 
@@ -111,40 +111,42 @@ static DHCPTransaction *DHCPTxAlloc(DHCPGlobalState *dhcp, uint32_t xid)
     }
 
     tx->xid = xid;
+    tx->state = dhcp;
 
     /* Increment the transaction ID on the state each time one is
      * allocated. */
-    tx->tx_id = dhcp->transaction_max++;
-    dhcp->transaction_count++;
+    tx->tx_id = dhcp->global->transaction_max++;
+    dhcp->global->transaction_count++;
 
-    TAILQ_INSERT_TAIL(&dhcp->tx_list, tx, next);
+    TAILQ_INSERT_TAIL(&dhcp->global->tx_list, tx, next);
 
     return tx;
 }
 
-static SC_ATOMIC_DECLARE(uint64_t, DHCPGlobalStateAllocCount);
-
 static void *DHCPGlobalStateAlloc(void)
 {
-    /* TBD: possibly make this per vlan */
-    DHCPGlobalState *state = &dhcpGlobalState;
+    DHCPState *state = SCCalloc(1, sizeof(*state));
+    if (unlikely(state == NULL)) {
+        return NULL;
+    }
+    state->global = &dhcpGlobalState;
     return state;
 }
 
 static void DHCPGlobalStateFree(void *state)
 {
-    DHCPGlobalState *dhcp_state = state;
-    DHCPTransaction *tx;
-    uint64_t count = SC_ATOMIC_SUB(DHCPGlobalStateAllocCount, 1);
-    /* free in-flight transactions with last DHCPGlobalStateFree */
-    if (count == 0) {
-        SCMutexLock(&dhcp_state->lock);
-        while ((tx = TAILQ_FIRST(&dhcp_state->tx_list)) != NULL) {
-            TAILQ_REMOVE(&dhcp_state->tx_list, tx, next);
-            DHCPTxFree(dhcp_state, tx, 1);
+    DHCPState *dhcp_state = state;
+    DHCPGlobalState *global = dhcp_state->global;
+    DHCPTransaction *tx, *tmp;
+    SCMutexLock(&global->lock);
+    TAILQ_FOREACH_SAFE(tx, &global->tx_list, next, tmp) {
+        if (tx->state == state) {
+            TAILQ_REMOVE(&global->tx_list, tx, next);
+            DHCPTxFree(global, tx, 1);
         }
-        SCMutexUnlock(&dhcp_state->lock);
     }
+    SCMutexUnlock(&global->lock);
+    SCFree(dhcp_state);
 }
 
 /**
@@ -155,26 +157,19 @@ static void DHCPGlobalStateFree(void *state)
  */
 static void DHCPGlobalStateTxFree(void *state, uint64_t tx_id)
 {
-    DHCPGlobalState *dhcp = state;
+    DHCPGlobalState *dhcp = ((DHCPState *)state)->global;
     DHCPTransaction *tx = NULL, *ttx;
 
     SCLogDebug("Freeing transaction %"PRIu64, tx_id);
 
     SCMutexLock(&dhcp->lock);
     TAILQ_FOREACH_SAFE(tx, &dhcp->tx_list, next, ttx) {
-
-        /* Continue if this is not the transaction we are looking
-         * for. */
-        if (tx->tx_id != tx_id) {
-            continue;
+        if (tx->state == state && tx->tx_id == tx_id) {
+            TAILQ_REMOVE(&dhcp->tx_list, tx, next);
+            DHCPTxFree(dhcp, tx, 1);
+            SCMutexUnlock(&dhcp->lock);
+            return;
         }
-
-        /* Remove and free the transaction. */
-        TAILQ_REMOVE(&dhcp->tx_list, tx, next);
-        DHCPTxFree(dhcp, tx, 1);
-
-        SCMutexUnlock(&dhcp->lock);
-        return;
     }
     SCMutexUnlock(&dhcp->lock);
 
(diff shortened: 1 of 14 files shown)


jasonish:
cc @decanio

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,699 @@
...
+
+    if (dhcptx->de_state != NULL)
+        DetectEngineStateFree(dhcptx->de_state);
+
+    SCFree(tx);
+
+    if (unlikely(locked == 0)) {
+        SCMutexLock(&dhcp->lock);
+    }
+    dhcp->transaction_count++;
```
inliniac: whats the logic of incrementing a transaction count on freeing a tx?

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,699 @@
...
+     * either. */
+    if (input == NULL || input_len == 0) {
+        return 0;
+    }
+
+    /* Normally you would parse out data here and store it in the
+     * transaction object, but as this is echo, we'll just record the
+     * request data. */
+
+    if (input_len >= DHCP_MIN_FRAME_LEN) {
```
inliniac: if we reverse this check and return from the function (goto end), the entire (long) blocks indent can be reduced a level

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,699 @@
...
+        BOOTPHdr *bootp = (BOOTPHdr *)input;
+
+        if ((bootp->op == BOOTP_REQUEST) &&
```
inliniac: I think it makes sense to split out the request and response parsers. The function is very long.

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,699 @@
...
+                            DHCPTxFree(global, tx, 0);
+                            goto end;
+                        }
+                        memcpy(tx->request_buffer, dhcp, tx->request_buffer_len);
+                        /* response to release not required */
+                        tx->response_unneeded = 1;
+                        tx->request_seen = 1;
+                        break;
+                    default:
+                        SCLogDebug("DHCP unknown %d", dhcp->args[0]);
```
inliniac: should this be an event?

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,699 @@
...
+    if ((dhcptx->request_seen && dhcptx->response_seen) ||
+        (dhcptx->request_seen && dhcptx->response_unneeded)) {
+        return 1;
+    }
+
+    return 0;
+}
+
+/**
+ * \brief ???
```
inliniac: todo

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,699 @@
...
+ */
+static DetectEngineState *DHCPGetTxDetectState(void *vtx)
+{
+    DHCPTransaction *tx = vtx;
+    return tx->de_state;
+}
+
+/**
+ * \brief ???
```
inliniac: todo

src/app-layer-dhcp.c
```diff
@@ -0,0 +1,699 @@
...
+
+SCEnumCharMap dhcp_decoder_event_table[] = {
+    {"EMPTY_MESSAGE", DHCP_DECODER_EVENT_EMPTY_MESSAGE},
+};
+
+static uint32_t dhcp_config_max_transactions = 32;
+
+static DHCPGlobalState dhcpGlobalState;
+
+static void DHCPTxFree(DHCPGlobalState *dhcp, void *tx, uint32_t locked)
```
inliniac: not a big fan of the conditional locking, esp since it's called with 0/1 so at the callers it's unclear what it means. Perhaps a unlocked func and lock wrapper would make sense, although it would have to be done such that all the actual freeing isn't blocking other threads.

src/app-layer-dhcp.h
```diff
@@ -0,0 +1,180 @@
...
+    DetectEngineState *de_state;
+
+    uint8_t reverse_flow; /*<< Set when the flow is the reverse of
+                           * what Suricata detected. This is because
+                           * the response can be on a new flow, which
+                           * Suricata will flag as to server. */
+
+    TAILQ_ENTRY(DHCPTransaction_) next;
+
+} DHCPTransaction;
```
inliniac: Think the struct can be a bit more compact by reshuffling fields

src/app-layer-dhcp.h
```diff
@@ -0,0 +1,180 @@
...
+                               * value. */
+
+    /* Indicates the current transaction being logged. 
+     */
+    uint64_t log_id;
+
+    uint32_t transaction_count; /**< A count of the number of
+                                 * in-progress transactions. */
+
+    uint16_t events; /**< Number of application layer events created
```
inliniac: would it make sense to move this to DHCPState? When detection is done that should be the correct state right?

src/output-json-dhcp.c
```diff
@@ -0,0 +1,610 @@
...
+    uint32_t        count;
+    MemBuffer       *buffer;
+} LogDHCPLogThread;
+
+static inline int JsonDHCPOptEnd(DHCPOpt *dhcp_opt)
+{
+    return (dhcp_opt->code == DHCP_OPT_END) ? 1 : 0;
+}
+
+static int JsonDHCPLogger(ThreadVars *tv, void *thread_data,
```
inliniac: can this function be split into more readable chunks?

src/output-json-dhcp.c
```diff
@@ -0,0 +1,610 @@
...
+                                     dhcptx->request_client_ip_bytes[0],
+                                     dhcptx->request_client_ip_bytes[1],
+                                     dhcptx->request_client_ip_bytes[2],
+                                     dhcptx->request_client_ip_bytes[3]);
+                    json_object_set_new(reqjs, "client_ip", json_string(ipaddr));
+                }
+                }
+                break;
+            case DHCP_OPT_HOSTNAME: {
+                char *s = BytesToString(dhcp_opt->args, dhcp_opt->len);
```
inliniac: BytesToString can return NULL, does json_string handle that correctly?

inliniac: I prefer checking even if it does

src/output-json-dhcp.c
```diff
@@ -0,0 +1,610 @@
...
+                if (offset + option_size > dhcptx->response_buffer_len) {
+                    goto error;
+                }
+            } else {
+                option_size = offsetof(DHCPOpt, len);
+            }
+
+            switch (dhcp_opt->code) {
+                case DHCP_OPT_TYPE: {
+                    char *s = (char *)"";
```
inliniac: why not make 's' const char * instead of all the casting?

src/output-json-dhcp.c
```diff
@@ -0,0 +1,610 @@
...
+
+    OutputCtx *output_ctx = SCCalloc(1, sizeof(*output_ctx));
+    if (unlikely(output_ctx == NULL)) {
+        SCFree(dhcplog_ctx);
+        return NULL;
+    }
+    output_ctx->data = dhcplog_ctx;
+    output_ctx->DeInit = OutputDHCPLogDeInitCtxSub;
+
+    SCLogNotice("dhcp log sub-module initialized.");
```
inliniac: don't forget to remove in final pr

inliniac:
I agree a default 'brief' format like that would be great. The current more detailed format should stay as well of course.

src/app-layer-dhcp.c
jasonish: I can't see why this is required? Will remove.

src/output-json-dhcp.c
jasonish: Fixed, and squashed into compiler fixups commit.

src/app-layer-dhcp.c
jasonish: Template bits not removed. Fixed.

src/app-layer-dhcp.c
jasonish: Template bits not removed. Fixed.

src/app-layer-dhcp.c
jasonish: I'm not sure if it should, yet. This more of a we don't handle this message yet, and not an uknown. When/if we exhaustively handle all known message types then maybe it should be alert? But the scope of this is more about logging messages to track allocated addresses.

src/output-json-dhcp.c
jasonish: It does, but I agree, something like this looks better I think:
```
            case DHCP_OPT_HOSTNAME:
                if (dhcp_opt->len) {
                    char *val = BytesToString(dhcp_opt->args, dhcp_opt->len);
                    if (val != NULL) {
                        json_object_set_new(reqjs, "hostname", json_string(val));
                        SCFree(val);
                    }
                }
                break;
```

src/output-json-dhcp.c
jasonish: more template cleanup

src/app-layer-dhcp.h
jasonish: There is currently no DHCP specific detection. But moving it shouldn't be much of a problem. Probably better to make the global state as small as possible anyways.

src/app-layer-dhcp.c
jasonish: With the increment of the transaction removed, there is no longer a need for the lock.

jasonish: Nevermind, thats supposed to be a decrement to track the number of outstanding transactions. But this needs to be refactored anyways, as DHCPTxFree is called in a few places without the Tx being removed from the list.

src/app-layer-dhcp.h
jasonish: I managed to remove all the holes. Would it make sense to move the buffer fields lower down, into the second cacheline as they are access less frequently?

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2721

PR #2690, opened 2017-05-04
Suripcap v3

regit:
Here's a work by @glongo and me adding a script to check the result of processing by Suricata of pcap files. First it test if the application layer counters statistics are correct. But most interesting part is
 that it can test for specific events by using filter.

Configuration is a YAML file looking like:
```
%YAML 1.1
---

- test: test TLS
  filename: docker/pcaps/tls.pcap
  ruleset: none
  config: ../suricata.yaml
  options: -k none
  app-layer-events:
    - name: tls
      flow: 5
      tx: 0
  metadata:
    - name: SELKS certificate
      event_type: tls
      count: 4
      filter: "tls.serial='00:97:E6:47:09:8E:EA:C9:B4' tls.issuerdn='C=FR, ST=IDF, L=Paris, O=Stamus, CN=SELKS'"
    - name: Session resumed
      event_type: tls
      count: 1
      filter: "tls.session_resumed=true"
```

Here, suripcap is checking that Suricata is seeing 5 TLS flow in the pcap (info taken from stats entry). On metadata, suripcap check that there is 4 TLS event matching the filter on serial and issuerdn and 1 for the session resumed.

It is also possible to implement test on any type of events like for alert:
```
    - name: juve
      event_type: alert
      count: 10
      filter: "alert.signature_id=1 alert.severity=2 http.hostname='www.google.it'"
```

Different Suricata configuration can be provided and it is also possible to specify options for the command line.

Currently the default configuration is only using a single pcap available in the source code. Also, the test has not been integrated into `make check` as the pcap is not part of the distributed files so distcheck is failling.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/260
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/42

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2039,7 +2039,7 @@ AC_SUBST(CONFIGURE_PREFIX)
 AC_SUBST(CONFIGURE_SYSCONDIR)
 AC_SUBST(CONFIGURE_LOCALSTATEDIR)
 
-AC_OUTPUT(Makefile src/Makefile qa/Makefile qa/coccinelle/Makefile rules/Makefile doc/Makefile doc/userguide/Makefile contrib/Makefile contrib/file_processor/Makefile contrib/file_processor/Action/Makefile contrib/file_processor/Processor/Makefile contrib/tile_pcie_logd/Makefile suricata.yaml scripts/Makefile scripts/suricatasc/Makefile scripts/suricatasc/suricatasc)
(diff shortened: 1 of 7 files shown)


jasonish:
Two thoughts here:

* Its somewhat similar to what I've been working on over here: https://github.com/jasonish/suricata-verify

* While I think its nice to have directly in the source tree, I wonder if growth over time, and storing pcaps may cause an excessively large git checkout? Unless the scope is to keep it pretty tight, whereas I'd like to see suricata-verify become something more like Victor's private tests over time, but all open.

regit:
Main idea behind "suripcap" was to have something really easy to extend. Someone without any knowledge in any language is able to edit the yaml file and add some tests.

Regarding the pcap files and the growth in size, I agree with you and that's why I did not add the script to `make check` as this meat we would need to embed the pcap in the source. Regarding that my plan was more to be able to run on an external pcap files source containing for instance private pcap. This way developers would be able to add suripcap tests to their own QA.

inliniac:
Think it would be useful to add a few real tests to the PR.

regit:
OK, preparing that.

PR #2696, opened 2017-05-08
app-layer gap handling - v6.1

jasonish:
Previous PR: #2660 

For parsers configured to handle gaps, pass stream gaps onto them by sending NULL data with a positive input length where the length if the number of bytes lost in the gap.

Add gap handling to DNS for an example. In the DNS case a flag is set on the state that a gap occurred. On the next input data the data is re-probed to see if it can continue.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/146
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/498

### DIFF
--- a/src/stream-tcp-reassemble.c
+++ b/src/stream-tcp-reassemble.c
@@ -893,7 +893,12 @@ static void GetAppBuffer(TcpStream *stream, const uint8_t **data, uint32_t *data
     }
 }
 
-static inline bool CheckGap(TcpStream *stream, Packet *p)
+/**
+ * \brief Check if there is a gap in the application data.
+ *
+ * \retval the size of the gap in bytes, 0 if no gap.
+ */
+static inline uint32_t CheckGap(TcpStream *stream, Packet *p)
 {
     const uint64_t app_progress = STREAM_APP_PROGRESS(stream);
     uint64_t last_ack_abs = STREAM_BASE_OFFSET(stream);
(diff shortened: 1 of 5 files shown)


src/stream-tcp-reassemble.c
```diff
@@ -1081,12 +1088,15 @@ int StreamTcpReassembleAppLayer (ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
                         "(base %u, isn %u, last_ack %u => diff %u) p %"PRIu64,
                         stream->base_seq, stream->isn, stream->last_ack,
                         stream->last_ack - (stream->isn + ackadd), p->pcap_cnt);
+                gap = stream->last_ack - (stream->isn + ackadd);
```
jasonish: There is an issue when we hit this case. The seg_list continues to be NULL, so its never passed to the app-layer. Not sure yet where I should be looking here.

src/stream-tcp-reassemble.c
```diff
@@ -934,9 +939,9 @@ static inline bool CheckGap(TcpStream *stream, Packet *p)
 static int ReassembleUpdateAppLayer (ThreadVars *tv,
         TcpReassemblyThreadCtx *ra_ctx,
         TcpSession *ssn, TcpStream *stream,
-        Packet *p, enum StreamUpdateDir dir)
+        Packet *p, enum StreamUpdateDir dir, uint32_t gap)
 {
-    const uint64_t app_progress = STREAM_APP_PROGRESS(stream);
+    const uint64_t app_progress = STREAM_APP_PROGRESS(stream) + gap;
```
inliniac: instead of adding the gap here I think it makes more sense to update app_progress right after the GAP call to the applayer

src/stream-tcp-reassemble.c
inliniac: In a preview/test PR #2702 I have removed this seg_list based gap check. I think that makes your PR easier to implement as well.

src/stream-tcp-reassemble.c
jasonish: Ok, I'll have to rebase against your changes then make this change.

Currently I am never able to get the gap check below this to trigger. It was also the gap check before this which triggered, so the gap had to be used here to get at the next data (as the gap call had already been made to the app-layer before calling ReassembleUpdateAppLayer).

jasonish:
Closing, see https://github.com/inliniac/suricata/pull/2704

PR #2699, opened 2017-05-09
Misc 20170510

regit:
Address packet loss issue in IPS by adding a new option and warn user when IPS is used with tpacket_v3.

Redmine issues:
- https://redmine.openinfosecfoundation.org/issues/2100
- https://redmine.openinfosecfoundation.org/issues/2099

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/261
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/43

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -264,11 +264,17 @@ static void *ParseAFPConfig(const char *iface)
                     iface,
                     aconf->out_iface);
             aconf->copy_mode = AFP_COPY_MODE_IPS;
+            if (aconf->flags & AFP_TPACKET_V3) {
+                SCLogError(SC_ERR_RUNMODE, "Using tpacket_v3 in IPS mode will result in high latency");
+            }
         } else if (strcmp(copymodestr, "tap") == 0) {
             SCLogInfo("AF_PACKET TAP mode activated %s->%s",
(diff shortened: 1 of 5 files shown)


src/runmode-af-packet.c
```diff
@@ -264,11 +264,17 @@ static void *ParseAFPConfig(const char *iface)
                     iface,
                     aconf->out_iface);
             aconf->copy_mode = AFP_COPY_MODE_IPS;
+            if (aconf->flags & AFP_TPACKET_V3) {
+                SCLogError(SC_ERR_RUNMODE, "Using tpacket_v3 in IPS mode will result in high latency");
```
inliniac: should be warning if all it does it warn

src/stream-tcp-inline.c
```diff
@@ -35,6 +35,7 @@
 
 /** defined in stream-tcp-reassemble.c */
 extern int stream_inline;
+extern int stream_drop_invalid;
```
inliniac: Why is this global var and not a field/flag of TcpStreamCnf ?

inliniac:
Can you send me the TCP case you've been looking at?

PR #2701, opened 2017-05-10
dns: fix log filtering

jasonish:
Previously only a subset of the records could be selected
in custom. Now allow any to be selected.

For example, using:
```
custom: [aaaa]
````
Would not log AAAA records, but instead no DNS records would be logged.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/147
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/499


jasonish:
cc @decanio

inliniac:
Merged through #2716, thanks Jason!

PR #2704, opened 2017-05-10
Dev gap 7

jasonish:
Previous PR: #2696 

Changes since last PR:
- Adapt for commit https://github.com/jasonish/suricata/commit/6a19f6a5e2c882d6568c2091c78d0fa57854d7e9

Outstanding issue:
- A gap in TS, or a gap in TC direction is handled as expected. However, if a TS packet is lost and the following TC packet is lost, we only get to the point of CheckGap in a single direction, not both.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/151
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/503

### DIFF
--- a/src/stream-tcp-reassemble.c
+++ b/src/stream-tcp-reassemble.c
@@ -877,8 +877,9 @@ static void GetAppBuffer(TcpStream *stream, const uint8_t **data, uint32_t *data
         StreamingBufferBlock *blk = stream->sb.block_list;
 
         if (blk->offset > offset) {
-            SCLogDebug("gap, want data at offset %"PRIu64", got data at %"PRIu64,
-                    offset, blk->offset);
+            SCLogDebug("gap, want data at offset %"PRIu64", "
+                    "got data at %"PRIu64". GAP of size %"PRIu64,
+                    offset, blk->offset, blk->offset - offset);
             *data = NULL;
             *data_len = 0;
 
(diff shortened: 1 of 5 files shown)


src/stream-tcp-reassemble.c
```diff
@@ -936,31 +967,41 @@ static int ReassembleUpdateAppLayer (ThreadVars *tv,
...
 
     const uint8_t *mydata;
     uint32_t mydata_len;
     GetAppBuffer(stream, &mydata, &mydata_len, app_progress);
+    SCLogDebug("%"PRIu64" got %p/%u", p->pcap_cnt, mydata, mydata_len);
+
     if (mydata == NULL || mydata_len == 0) {
-        if (CheckGap(stream, p)) {
+        /* Only check for gap and signal if not already done. */
+        if (gap == 0 && ((gap = CheckGap(ssn, stream, p)) > 0)) {
```
inliniac: this gap ==0  check is now always true

src/stream-tcp-reassemble.c
```diff
@@ -893,10 +894,16 @@ static void GetAppBuffer(TcpStream *stream, const uint8_t **data, uint32_t *data
...
+/** \internal
+ *  \brief check to see if we should declare a GAP
+ *  Call this when the app layer didn't get data at the requested
+ *  offset.
+ */
+static inline uint32_t CheckGap(TcpSession *ssn, TcpStream *stream, Packet *p)
 {
     const uint64_t app_progress = STREAM_APP_PROGRESS(stream);
     uint64_t last_ack_abs = STREAM_BASE_OFFSET(stream);
+    uint32_t gap = app_progress - last_ack_abs;
```
inliniac: I'm not convinced this is correct. last_ack_abs indicates the right edge of what is ack'd. This may include both gaps and segment data.

When this function is called the only thing we know for sure is that there was no data at the exact offset of 'app_progress'.  One byte later we may have data.

We can have something like:
<pre>
[gap][data][gap][data]
</pre>
all in the window between app_progress and last_ack_abs

jasonish:
Understood. Perhaps inspecting the streaming buffer would be better, relative to what's being requested?

jasonish:
I'm actually finding that having GetAppBuffer will in mydata_len with the size of the gap, and using mydata == NULL with len > 0 as the gap check is handling more cases than my gap check above. What do you think of that approach?

There is a difference tho. The gap notification is done where there is actually more data. Before it would be done on a following ack, even if no new data was present.

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2710

PR #2703, opened 2017-05-10
rust - dns - v9

jasonish:
Previous PR: #2686 

Changes from last PR:
- Add -lm on Linux
- Create a LoggerFlags type to encapsulate tracking loggers
- Type filtering for DNS logs (I thought it was working before, but only request/reply filtering was).

TODOs:
- Do a lib check in configure for extra libraries added.
- Not all DNS types are logged (all comon ones are).
- Add command that generated files to generated files (forgot).
- Post parse validation so we can log and alert on things like an invalid SSHFP (this would be a new feature, as this isn't done in the C version either).
- Look further into Rust documentation generation. While I had it somewhat work, no index page was created.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/148
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/500

PR #2706, opened 2017-05-11
rust - dns - v10

jasonish:
Previous PR: #2703 

Changes from last PR:
- Do not included generated headers in version controlled sources. They will be generated on the fly, and as part of "make dist" so Python is not a requirement to build from a release tarball.
- Generate headers have "-gen" to suggest they are generated.
- Add [profile.release] debug=true

TODOs (from last PR):
- Do a lib check in configure for extra libraries added.
- Not all DNS types are logged (all common ones are).
- Add command that generated files to generated files (forgot).
- Post parse validation so we can log and alert on things like an invalid SSHFP (this would be a new feature, as this isn't done in the C version either).
- Look further into Rust documentation generation. While I had it somewhat work, no index page was created.

PRscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/152
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/504


jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2708

PR #2710, opened 2017-05-12
dev - application layer gap handling - v8

jasonish:
Previous PR: #2704 

Changes from last PR:
- GetAppBuffer can calculate the size of the gap, that is the amount of missing data before actual data. Use this for gap notification.
- This means that the gap is sent to the app-layer immediately upon the ack, instead it won't be notified until there is data after the gap. I'm am not yet sure if this matters though, but it makes for a simpler implementation.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/156
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/508

### DIFF
--- a/src/app-layer-dns-common.h
+++ b/src/app-layer-dns-common.h
@@ -232,6 +232,8 @@ typedef struct DNSState_ {
     uint16_t offset;
     uint16_t record_len;
     uint8_t *buffer;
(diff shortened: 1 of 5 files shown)


inliniac:
Merged through #2737, thanks Jason!

PR #2713, opened 2017-05-16
Experiments/rust 0.4

chifflier:
This PR adds the feature to write application layer parsers in Rust.

Note: it does not contain the detection code, for now


inliniac:
So some things that we need to consider:
- dynamic registration of alproto's inside Suricata. This shouldn't be very hard, but is still a bit of work.
- how to expose things to detection. In our initial integration we're simply going to expose dedicated Rust functions to C. But it would be nice if we can come up with a generic way at least for the common cases like content buffers and simple numeric fields
- similarly, how to expose things to logging (including alert record metadata), Lua, etc.

rust/interface/src/rparser.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Helper macros
+// We can't define them (can we ?) in the trait, partly because of the 'no_mangle' and 'extern'
+// stuff.
+// This forces to use macros in addition to the trait, but at least provides a proper way of
+// encapsulating the translation of C variables to rust.
+
+// Double-box, because we are wrapping the RParser trait, and a Box<Trait> has the size of two
+// pointers (128 bits) so cannot be caster to a single pointer.
```
inliniac: So adding a trait adds to the size? I must admit that the whole trait concept hasn't really landed yet for me. Whats the advantage of using it here?

rust/src/ipsec.rs
```diff
@@ -0,0 +1,165 @@
+extern crate libc;
+
+use std;
+use std::mem;
+use libc::c_char;
+
+use rparser::*;
+
+use ipsec_parser::*;
```
inliniac: is this where parse_ikev2_header is defined? Is that a separate crate?

rust/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+                              new_state:    new,
+                              free_state:   free,
+                          }
+                          ),
+            .. r
+        }
+    }
+}
+
+lazy_static! {
```
inliniac: what does lazy_static do?

rust/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Global parsers registry.
+    /// The registry is built at compile time and is immutable.
+    static ref HASHMAP: HashMap<String, RustParser> = {
+        let mut m = HashMap::new();
+        m.insert("ntp".to_string(),RustParser::new("rust-ntp", 17, "123", 0,
```
inliniac: I assume this would be done dynamically later, right?

rust/src/logger.rs
```diff
@@ -0,0 +1,39 @@
+//! Logging module
```
inliniac: whats the purpose of this when there are also SCLog* macro's?

src/Makefile.am
```diff
@@ -470,6 +471,12 @@ AM_CPPFLAGS = $(all_includes)
 # the library search path.
 suricata_LDFLAGS = $(all_libraries) ${SECLDFLAGS}
 suricata_LDADD = $(HTP_LDADD)
+EXTRA_suricata_DEPENDENCIES =
+
+if BUILD_RUST
+suricata_LDADD += $(RUST_LDADD)
+EXTRA_suricata_DEPENDENCIES += $(RUST_LDADD)
```
inliniac: this is a nice trick :)

src/util-rust.c
```diff
@@ -0,0 +1,218 @@
...
+        return ALPROTO_TLS;
+    else {
+        SCLogError(SC_ERR_INITIALIZATION, "Unknown Rust protocol");
+        return -1;
+    }
+
+}
+
+
+static int RustParseToServer(Flow *f, void *state,
```
inliniac: I like how this would make the C-side of things much leaner. I like to see what we learn from the first couple of 'hard coded' protocols to figure out how to make this flexible enough.

rust/interface/src/rparser.rs
chifflier: Hum, not really. A trait is a common interface for all parsers (which is a good thing).
The only 'trick' here is that the memory representation of a trait is 2 pointers (the object and vtable), so we just need to box (use a single pointer to point on this structure).
There is no real overhead (the compiler abstracts the Box), it's just the code is a bit weird so I documented it.

rust/src/ipsec.rs
chifflier: Oh, I forgot to comment that. The ipsec-parser crate is of https://github.com/rusticata/ipsec-parser
It was supposed to be merged in a later commit

rust/src/lib.rs
chifflier: It allows to define a static (immutable) structure (the hash table), using initialization code. If you don't use lazy_static, you can't use code, only declaration of values

rust/src/lib.rs
chifflier: Actually, this is where the rust parsers are all defined (you need to do that once) so the C code can fetch that to register them.

rust/src/logger.rs
chifflier: Compatibility: this way, you can use crates that use the debug! etc. macros without changing a single line of code
Testing: you can test your parsers outside Suricata (debug! will send to stderr) or inside (to Suricata's logging facility)

rust/src/lib.rs
inliniac: Interesting, thanks!

rust/src/lib.rs
inliniac: yeah, I thought maybe these per protocol calls could move into the files where they are implemented.

rust/src/logger.rs
inliniac: Right, gotcha. So would this be transparent in the sense that it would still show the correct file and line number in Suricata's output?

Function name seems to be tricky in Rust.

rust/interface/src/rparser.rs
inliniac: But does this lead to a state that is one pointer size bigger than needed? Not a big deal just trying to understand.

PR #2721, opened 2017-05-24
dhcp - v3.ish

jasonish:
Previous PR: #2689 

Changes from last PR:
- Refactor parser and logger into request and response handling functions to break up larger functions.
- Remove unused event. I'm not sure what events, if any we should trigger at this time. I would say malformed, but much of the parsing is done during logging, not during parsing.
- Refactor some transaction handling. In error cases, the Tx would be free'd but not removed from the tx list.
- Reorganized some data structures to be more space efficient.
- By default only log in a "short" mode and an extended mode to get a full break out of request and reply logging.

Example of default DHCP log:
https://gist.github.com/jasonish/62419300b83f84a4ec6a81eacc9c8e41

Example of extended log:
https://gist.github.com/jasonish/6fcda7ab03cf08c703f34e9bef7a3322

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/159
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/511

CC: @decanio

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -16,6 +16,7 @@ app-layer.c app-layer.h \
 app-layer-dcerpc.c app-layer-dcerpc.h \
 app-layer-dcerpc-udp.c app-layer-dcerpc-udp.h \
 app-layer-detect-proto.c app-layer-detect-proto.h \
+app-layer-dhcp.c app-layer-dhcp.h \
 app-layer-dnp3.c app-layer-dnp3.h \
 app-layer-dnp3-objects.c app-layer-dnp3-objects.h \
 app-layer-dns-common.c app-layer-dns-common.h \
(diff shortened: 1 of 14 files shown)


jasonish:
One thing that I am still not sure how should be best handled is an issue I brought up in the last PR:

Given a capture like:
```
Client --> DHCP Discover
Server --> DHCP Offer
Client --> DHCP Request
Server --> DHCP ACK
```
The discover/offer and request/ack get the same transaction ID, so the request overwrites the data already recorded in the discover...  2 thoughts here. Ignore discovers for now. Its really the request/ack we are interested in. Or extend the logic handling so the request makes a new transaction, so the transaction will need a type in addition to the tx id.  For the 99% use case I don't see much value here.

For most use cases I think only handling the request and ack is enough, and that gives us the short form log. Perhaps discretely logging the discover and offer, and the request and the ack in extended mode could be useful, but will require additional transaction logic. Currently in the above scenario the request and ack data overwrites the discover/offer as they are on the same transaction, which makes me think that just ignoring discover/offer for now might be the best solution?

jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2722

PR #2724, opened 2017-05-28
Feature 2114 redis rpushoutput

securitym0nkey:
This commit implements my own [Feature Request](https://redmine.openinfosecfoundation.org/issues/2114).
No heavy load test done, but in my lab it seems to work.

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -35,6 +35,7 @@
 #include "util-logopenfile-tile.h"
 
 const char * redis_push_cmd = "LPUSH";
+const char * redis_rpush_cmd = "RPUSH";
 const char * redis_publish_cmd = "PUBLISH";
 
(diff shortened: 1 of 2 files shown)


jasonish:
Yes, the idea is sound, and is a must do for Logstash I think.  I haven't tested it, but it looks simple enough.  I had planned on testing soon-ish though.

fooinha:
Yes, this idea looks good to me, and it will better for Logstash.

Maybe we should improve the config parsing and error checking for redis mode config key.
It's misleading in several ways.

jasonish:
@securitym0nkey  Just noticed this was made against 3.2.x. Any chance you can rebase it against git master please?

securitym0nkey:
@jasonish I did a 2nd PR #2736 for this change based on master. Any change to get this feature also into the 3.2.x?

inliniac:
This will need a doc update as well.

PR #2725, opened 2017-05-30
log-tlsstore: fix error handling

regit:
In case of realloc error, the length of the encoding buffer was not
resetted and this could result in trying to write to NULL pointer.

PR builds:
* https://buildbot.openinfosecfoundation.org/builders/regit/builds/272
* https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/55


inliniac:
Merged through #2727, thanks Eric! Applying to master as well.

inliniac:
Merged into master through #2732, thanks Eric!

PR #2726, opened 2017-05-30
doc: rephrase nocase placement explanation

norg:
just a small doc change, no need for prscript run


inliniac:
Cherry-picked into 3.2.x through #2727. Leaving open as it has to go into master still.

inliniac:
Merged through #2732, thanks Andreas!

PR #2735, opened 2017-06-02
Filestore optim

regit:
Two filestore related commits. The first one allows not to write meta file. And the second one try to limit the number of open/write/close sequence when storing files.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/274
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/57

### DIFF
--- a/src/log-filestore.c
+++ b/src/log-filestore.c
@@ -64,6 +64,8 @@
 
 static char g_logfile_base_dir[PATH_MAX] = "/tmp";
 
+SC_ATOMIC_DECLARE(int, filestore_open_file_cnt);  /**< Atomic counter, to link relative event */
+
 typedef struct LogFilestoreLogThread_ {
     LogFileCtx *file_ctx;
     /** LogFilestoreCtx has the pointer to the file and a mutex to allow multithreading */
@@ -301,7 +303,7 @@ static void LogFilestoreLogCloseMetaFile(const File *ff)
 }
 
 static int LogFilestoreLogger(ThreadVars *tv, void *thread_data, const Packet *p,
-        const File *ff, const uint8_t *data, uint32_t data_len, uint8_t flags)
+        File *ff, const uint8_t *data, uint32_t data_len, uint8_t flags)
 {
     SCEnter();
     LogFilestoreLogThread *aft = (LogFilestoreLogThread *)thread_data;
@@ -340,12 +342,20 @@ static int LogFilestoreLogger(ThreadVars *tv, void *thread_data, const Packet *p
             SCLogDebug("failed to create file");
             return -1;
         }
+        if (SC_ATOMIC_GET(filestore_open_file_cnt) < FileGetMaxOpenFiles()) {
+            ff->fd = file_fd;
+            SC_ATOMIC_ADD(filestore_open_file_cnt, 1);
+        }
     /* we can get called with a NULL ffd when we need to close */
     } else if (data != NULL) {
-        file_fd = open(filename, O_APPEND | O_NOFOLLOW | O_WRONLY);
-        if (file_fd == -1) {
-            SCLogDebug("failed to open file %s: %s", filename, strerror(errno));
-            return -1;
+        if (ff->fd == -1) {
+            file_fd = open(filename, O_APPEND | O_NOFOLLOW | O_WRONLY);
+            if (file_fd == -1) {
+                SCLogDebug("failed to open file %s: %s", filename, strerror(errno));
+                return -1;
+            }
+        } else {
+            file_fd = ff->fd;
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/log-filestore.c
```diff
@@ -330,33 +332,60 @@ static int LogFilestoreLogger(ThreadVars *tv, void *thread_data, const Packet *p
...
+        } else {
+            file_fd = ff->fd;
         }
     }
 
     if (file_fd != -1) {
         ssize_t r = write(file_fd, (const void *)data, (size_t)data_len);
         if (r == -1) {
             SCLogDebug("write failed: %s", strerror(errno));
+            close(file_fd);
```
inliniac: seems file_fd is closed twice here. First here.

src/log-filestore.c
```diff
@@ -330,33 +332,60 @@ static int LogFilestoreLogger(ThreadVars *tv, void *thread_data, const Packet *p
...
+            if (ff->fd != -1) {
+                SC_ATOMIC_SUB(filestore_open_file_cnt, 1);
+            }
+            ff->fd = -1;
+        }
+        if (ff->fd == -1) {
+            close(file_fd);
```
inliniac: Second here

src/util-file.c
```diff
@@ -149,6 +155,25 @@ void FileForceTrackingEnable(void)
     g_file_force_tracking = 1;
 }
 
+void FileWriteMetaDisable(void)
```
inliniac: why are these meta things globals in util-file.c? They are specific to the file output logger right?

src/util-file.h
```diff
@@ -69,6 +71,7 @@ typedef struct File_ {
     uint32_t file_track_id;         /**< id used by protocol parser. Optional
                                      *   only used if FILE_USE_TRACKID flag set */
     uint32_t file_store_id;         /**< id used in store file name file.<id> */
+    int fd;
```
inliniac: comment missing, esp since -1 has a magic meaning

src/util-file.c
regit: I put there aside with the stream-depth which is there too.

inliniac: probably should not be a global at all to make it easier to support multiple instances of file-store (if that doesn't already work)

src/log-filestore.c
regit: Right, updating.

src/log-filestore.c
```diff
@@ -330,33 +332,60 @@ static int LogFilestoreLogger(ThreadVars *tv, void *thread_data, const Packet *p
     if (flags & OUTPUT_FILEDATA_FLAG_OPEN) {
         aft->file_cnt++;
 
-        /* create a .meta file that contains time, src/dst/sp/dp/proto */
-        LogFilestoreLogCreateMetaFile(p, ff, filename, ipver);
+        if (FileWriteMeta()) {
```
inliniac: since a function call is added here it might make sense to actually move that into the LogFilestoreLogCreateMetaFile, etc functions. That keeps the main loop cleaner.

src/log-filestore.c
```diff
@@ -502,6 +537,21 @@ static OutputCtx *LogFilestoreLogInitCtx(ConfNode *conf)
...
+        uint32_t file_count = 0;
+        if (ParseSizeStringU32(file_count_str,
+                               &file_count) < 0) {
+            SCLogError(SC_ERR_SIZE_PARSE, "Error parsing "
+                       "file-store.max-open-files "
+                       "from conf file - %s.  Killing engine",
+                       stream_depth_str);
+            exit(EXIT_FAILURE);
+        } else {
+            FileSetMaxOpenFiles(file_count);
```
inliniac: file_count is u32, but the actual counter plain int. Please sync those.

regit: yep.

src/log-filestore.c
regit: OK

src/util-file.h
regit: indeed

src/util-file.c
regit: Easy way to do it would be to add the info to OutputCtx. Want me to do that ?

regit: I don't see how we could use multiple instances of file-store. Doing twice the writing seems a terrible idea.
But do you want me to add a conf to the data field of OutputCtx ? Currently I only see this used by the meta.

inliniac: it seems all the log-filestore.c settings are globals. Might be nice to fix it but we can keep things as globals for now. I would like to see it move into log-filestore.c thought as it only has an effect there.

regit: OK. I agree this is not super nice and that can be improved.

regit:
OK, I'm given a bit of test on updated branch and resubmitting.

PR #2749, opened 2017-06-06
setting-up-ipsinline-for-linux: alternative setup with mangle PREROUTING

selivan:
Added alternative iptables option with mangle PREROUTING table. And small fix to easier check for NFQ support in Suricata.

### DIFF
--- a/doc/userguide/setting-up-ipsinline-for-linux.rst
+++ b/doc/userguide/setting-up-ipsinline-for-linux.rst
@@ -117,7 +117,7 @@ The options -i (input) -o (output) can be combined with all previous mentioned o
 
 If you would stop Suricata and use internet, the traffic will not come through. To make internet work correctly, you have to erase all iptable rules.
 
-To simply inspect all traffic on the host, before routing desicion and input/output separation, you can use mangle PREROUTING table:
+To simply inspect all incoming traffic on the host, before routing decision that separates `INPUT` and `FORWARD`, you can use `PREROUTING` chain from `mangle` table:
 
 ::
 
--- a/doc/userguide/setting-up-ipsinline-for-linux.rst
+++ b/doc/userguide/setting-up-ipsinline-for-linux.rst
@@ -15,7 +15,7 @@ To check if you have NFQ enabled in your Suricata, enter the following command:
 ::
 
 
-  suricata --build-info
+  suricata --build-info | grep -i NFQ
 
 and examine if you have NFQ between the features.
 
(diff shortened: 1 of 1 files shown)


inliniac:
Thanks!

@regit does this make sense to you?

doc/userguide/setting-up-ipsinline-for-linux.rst
```diff
@@ -117,6 +117,14 @@ The options -i (input) -o (output) can be combined with all previous mentioned o
 
 If you would stop Suricata and use internet, the traffic will not come through. To make internet work correctly, you have to erase all iptable rules.
 
+To simply inspect all traffic on the host, before routing desicion and input/output separation, you can use mangle PREROUTING table:
```
regit: Typo on decision

doc/userguide/setting-up-ipsinline-for-linux.rst
```diff
@@ -117,6 +117,14 @@ The options -i (input) -o (output) can be combined with all previous mentioned o
...
+
+::
+
+
+  sudo iptables -t mangle -I PREROUTING -j NFQUEUE
```
regit: This is wrong: this setup only works for FORWARDED traffic. In the case of INPUT/OUTPUT we will only get the INPUT traffic. Look at the output of iptables after doing a SSH from the local box to an external host.
```
iptables -L -nv -t mangle
Chain PREROUTING (policy ACCEPT 273 packets, 287K bytes)
 pkts bytes target     prot opt in     out     source               destination         
   18  4627            tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            tcp spt:22
    0     0            tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            tcp dpt:22
``` 
We don't see the packet in the output direction.

selivan:
@redit, oops, you are right. Fixed that: 0dd3db193246b8d2ffc1a36a06540d5ed2599360.

Anyway, I think sometimes using mangle PREROUTING chain makes sense: if Suricata is running in IPS mode, dropping packet on earlier processing stage is better.

inliniac:
Would it make sense to make this part of a more advanced nfq guide instead? The guide here is really focused at beginners.

Regardless, please don't update existing PRs https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Create-a-new-branch-for-incorporating-feedback

Also, having a PR with commits that fix commits in the PR is bad practice. See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits

PR #2744, opened 2017-06-05
Filestore optim v4

regit:
Update of #2741 with changes lost during after a `git stash` added. Two patches were merged as one as it was the same feature.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/285
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/69

PR #2750, opened 2017-06-06
Filestore optim v5

regit:
Update of #2744 with an attempt to fix coverity warning.


inliniac:
Looks like this doesn't solve the Coverity issue.

inliniac:
Merged through #2752, thanks Eric.

The coverity issue is still open. It looks like an FP, but the code is a bit hard to read. So a cleanup would be nice regardless.

PR #2751, opened 2017-06-06
Create CONTRIBUTING.md

selivan:
Add github guidelines for contributing that mention signing the Contribution Agreement


selivan:
@inliniac : this is better then manually sending letters :) Also you may consider adding PR template that states:

[x] I've sighed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/

inliniac:
I think this is a good idea, but we'll certainly craft some different language. So you can leave this to the team. Thanks!

PR #2755, opened 2017-06-06
Metadata v6

regit:
Update of #2674 introducing a new keyword and adding documentation. It also contains the metadata output of initial PR.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/287
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/71


doc/userguide/rules/meta.rst
```diff
@@ -182,3 +182,20 @@ only as optional fields in alert.  The format is:
...
+
+The format is:
+
+::
+
+ target: [src_ip|dest_ip]
+
+If the value is src_ip then the source IP in the generated event (src_ip
+field in JSON) is the target of the attack. If target is set to dest_ip
+then the tarfet is the destination IP in the generated event.
```
inliniac: typo 'tarfet'

src/detect.h
```diff
@@ -909,6 +909,8 @@ typedef struct DetectEngineThreadCtx_ {
     uint8_t *base64_decoded;
     int base64_decoded_len;
     int base64_decoded_len_max;
+    uint8_t alert_flags;
```
inliniac: I don't follow why this has to be in det_ctx. It seems it's a per-sig flag and you have the sig available at every place you need it. So it makes much more sense to me to have it in the sig.

inliniac: This would also make the target keyword much simpler as it wouldn't need it's own SigMatch and all.

src/decode.h
```diff
@@ -268,15 +268,21 @@ typedef struct PacketAlert_ {
...
+#define PACKET_ALERT_FLAG_TX            BIT_U8(3)
 /** action was changed by rate_filter */
-#define PACKET_ALERT_RATE_FILTER_MODIFIED   0x10
+#define PACKET_ALERT_RATE_FILTER_MODIFIED   BIT_U8(4)
+/** alert has source which is target of attack */
+#define PACKET_ALERT_SRC_IS_TARGET   BIT_U8(5)
+/** alert has destination which is target of attack */
+#define PACKET_ALERT_DEST_IS_TARGET   BIT_U8(6)
+
+#define PACKET_ALERT_HAS_IDMEF   (PACKET_ALERT_DEST_IS_TARGET|PACKET_ALERT_SRC_IS_TARGET)
```
inliniac: PACKET_ALERT_HAS_TARGET ? The IDMEF name doesn't really make sense to me here.

inliniac:
Can you split this PR into the target/source part and the metadata keyword logic. Not convinced yet about the latter and I want the former to already get in. Thanks!

regit:
Working on target-keyword branch.

src/detect.h
regit: It simplify thing indeed :+1:

PR #2756, opened 2017-06-07
rust: build fixups and nom update - 1

jasonish:
Update nom to ~3.0.
    
Prefix dependencies with ~, which will allow for newer patch
versions only. Minor version updates should get a test before
using.
    
Remove Cargo.lock from the repo, but still generate as part
of the vendoring process for release builds. This will ensure
that all users of a particular distribution tarball will be
linking against the same Rust dependencies.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/170
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/522


inliniac:
Thanks Jason. I feared updating to nom 3 would be an issue. Glad it wasn't :)

PR #2759, opened 2017-06-07
rule-parser: ignore duplicated msg keyword

norg:
This fixes the leak resulted by duplicated msg keyword from https://redmine.openinfosecfoundation.org/issues/2074

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/48
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/50


inliniac:
Does this fail if suricata is started with -T ? If not I think it should.

PR #2761, opened 2017-06-09
Target keyword

regit:
Update of #2755 removing metadata logging and merging some patches. It also feature a complete rework of target keyword following comment.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/289
- PR regit-pcap: https://buildbot.

### DIFF
--- a/src/detect-target.c
+++ b/src/detect-target.c
@@ -1,0 +1,162 @@
+/* Copyright (C) 2017 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Eric Leblond <eric@regit.org>
+ *
+ * target keyword allow rules writer to specify information about target of the attack
+ */
+
+#include "suricata-common.h"
+#include "util-unittest.h"
+
+#include "detect-parse.h"
+#include "detect-engine.h"
+
+#include "detect-target.h"
+
+/**
+ * \brief Regex for parsing our keyword options
+ */
+#define PARSE_REGEX  "^\\s*(src_ip|dest_ip)\\s*$"
+
+static pcre *parse_regex;
+static pcre_extra *parse_regex_study;
+
+/* Prototypes of functions registered in DetectTargetRegister below */
+static int DetectTargetSetup (DetectEngineCtx *, Signature *, const char *);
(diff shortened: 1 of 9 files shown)


src/detect-target.c
```diff
@@ -0,0 +1,162 @@
...
+    const char *value;
+
+    ret = pcre_exec(parse_regex, parse_regex_study,
+                    targetstr, strlen(targetstr), 0, 0, ov, MAX_SUBSTRINGS);
+    if (ret < 1) {
+        SCLogError(SC_ERR_PCRE_MATCH, "pcre_exec parse error, ret %" PRId32 ", string %s", ret, targetstr);
+        return 0;
+    }
+
+    res = pcre_get_substring(targetstr, ov, MAX_SUBSTRINGS, 1, &value);
```
inliniac: memory is allocated by pcre_get_substring but not freed anywhere. Please use pcre_copy_substring instead. See also https://redmine.openinfosecfoundation.org/issues/1046

src/detect-target.c
```diff
@@ -0,0 +1,162 @@
...
+    if (res < 0) {
+        SCLogError(SC_ERR_PCRE_GET_SUBSTRING, "pcre_get_substring failed");
+        return 0;
+    }
+
+    /* now check key value */
+    if (!strncmp(value, "src_ip", 7)) {
```
inliniac: this will match on src_ip_regit too. Please just use strcmp

src/detect-target.c
```diff
@@ -0,0 +1,162 @@
...
+        s->flags |= SIG_FLAG_SRC_IS_TARGET;
+    } else if (!strncmp(value, "dest_ip", 8)) {
+        s->flags |= SIG_FLAG_DEST_IS_TARGET;
+    } else {
+        SCLogError(SC_ERR_INVALID_VALUE, "only 'src_ip' and 'dest_ip' are supported for target target option");
+        goto error;
```
inliniac: seems this goto is unnecessary

src/detect-target.c
```diff
@@ -0,0 +1,162 @@
...
+
+    res = pcre_get_substring(targetstr, ov, MAX_SUBSTRINGS, 1, &value);
+    if (res < 0) {
+        SCLogError(SC_ERR_PCRE_GET_SUBSTRING, "pcre_get_substring failed");
+        return 0;
+    }
+
+    /* now check key value */
+    if (!strncmp(value, "src_ip", 7)) {
+        s->flags |= SIG_FLAG_SRC_IS_TARGET;
```
inliniac: I think there needs to a check that makes sure that multiple (conflicting) instances of target won't be accepted.

src/detect-target.c
```diff
@@ -0,0 +1,162 @@
...
+ * \param targetstr pointer to the user provided target options
+ *
+ * \retval 0 on Success
+ * \retval -1 on Failure
+ */
+static int DetectTargetSetup(DetectEngineCtx *de_ctx, Signature *s, const char *targetstr)
+{
+    int ret = DetectTargetParse(s, targetstr);
+    if (ret < 0)
+        goto error;
```
inliniac: useless goto

src/output-json-alert.c
```diff
@@ -215,6 +215,76 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
     if (p->tenant_id > 0)
         json_object_set_new(ajs, "tenant_id", json_integer(p->tenant_id));
 
+    if (pa->s->flags & SIG_FLAG_HAS_TARGET) {
```
inliniac: I think it would be cleaner if this block would move into it's own function.

PR #2763, opened 2017-06-09
RFC Rust/nfs/gap/v4

inliniac:
Gap handling for NFS3, not fully complete yet.

@jasonish can you have a look at the approach and let me know what you think?


jasonish:
The checking for gaps looks fine. I'm curious about what you do on a gap? Looks like the file is truncated? What about leaving holes in files?

pevma:
@jasonish good point.Should it not be the same way we do that for files in http/smtp - so it is homogeneous?

inliniac:
@jasonish yeah I'm truncating as a first step. My initial goal is to allow the parser to continue at all.

@pevma for smtp/http etc we don't do any of this. A gap is fatal and nothing is done beyond that.

inliniac:
Replaced by #2777

PR #2769, opened 2017-06-12
Pcap Folder Mode

dbcfd:
Run mode to allow pcap files to be placed into a folder and run by suricata. Files will continually be picked up and processed till engine is told to stop.

This satisfies the use case of an external system capturing packets and feeding them into multiple files over time.

@dkhelwig


src/runmode-pcap-folder.c
```diff
@@ -0,0 +1,265 @@
...
+
+    if (ConfGet("pcap-folder.folder", &folder) == 0) {
+        SCLogError(SC_ERR_RUNMODE, "Failed retrieving pcap-folder from Conf");
+        exit(EXIT_FAILURE);
+    }
+
+    RunModeInitialize();
+    TimeModeSetOffline();
+
+    PcapFolderGlobalInit();
```
inliniac: With the exception of this line this file seems a complete copy of the regular pcap-file runmode. I'm not a fan of code duplication like this as it's likely to get out of sync at some point.

src/runmode-pcap-folder.c
```diff
@@ -0,0 +1,265 @@
...
+    tm_module = TmModuleGetByName("FlowWorker");
+    if (tm_module == NULL) {
+        SCLogError(SC_ERR_RUNMODE, "TmModuleGetByName for FlowWorker failed");
+        exit(EXIT_FAILURE);
+    }
+    TmSlotSetFuncAppend(tv, tm_module, NULL);
+
+    TmThreadSetCPU(tv, WORKER_CPU_SET);
+
+#ifndef AFLFUZZ_PCAP_RUNMODE
```
inliniac: Is this AFL logic functional for folder?

src/runmode-unix-socket.c
```diff
@@ -355,6 +379,111 @@ static TmEcode UnixSocketPcapFilesCheck(void *data)
...
+        if (this->currentfolder) {
+            SCFree(this->currentfolder);
+        }
+        this->currentfolder = NULL;
+
+        PostRunDeinit(RUNMODE_PCAP_FOLDER, NULL /* no ts */);
+        return TM_ECODE_OK;
+    }
+
+    if (this->currentfolder == NULL)
```
inliniac: style: { on same line for code blocks, on new line for function decl. See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Coding_Style

src/runmode-unix-socket.c
```diff
@@ -389,6 +518,29 @@ void UnixSocketPcapFile(TmEcode tm)
...
+            unix_manager_folder_task_failed = 1;
+            break;
+        case TM_ECODE_OK:
+            unix_manager_folder_task_running = 1;
+            break;
+    }
+#endif
+}
+
+void UnixSocketPcapFolderSetLastProcessed(time_t time)
```
inliniac: minor: 'time' is the name of a common function, I prefer to avoid naming clashes/confusion

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+#include <limits.h> // PATH_MAX
+#include <sys/types.h>  // stat
+#include <sys/stat.h>   // stat
+#include <unistd.h>     // stat
+
+#ifdef __linux__
+#  include <linux/version.h>
+#  if LINUX_VERSION_CODE >= KERNEL_VERSION(2,6,21)
+#    define PCAP_FOLDER_ENABLE_MONITOR
+#    include <sys/inotify.h> // inotify
```
inliniac: before using it check if it exists in configure.ac and then use #ifdef HAVE_...

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+#include "util-cuda-buffer.h"
+#include "util-mpm-ac.h"
+#include "util-cuda-handlers.h"
+#include "detect-engine.h"
+#include "detect-engine-mpm.h"
+#include "util-cuda-vars.h"
+
+#endif /* __SC_CUDA_SUPPORT__ */
+
+#include <dirent.h> // dirent
```
inliniac: these should be moved to suricata-common.h and checked before including

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+
+    return ret;
+}
+
+int PcapFolderSortByStatTime(const void * vleft, const void * vright)
+{
+    if (vleft == vright)
+        return 0;
+
+    time_t leftTime = 0, rightTime = 0;
```
inliniac: style: we use snake_case for variables

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+void PcapFolderFreeMemBuffer(MemBuffer * buffer)
+{
+    if (buffer == NULL)
+        return;
+    
+    size_t offset = 0;
+
+    while (offset < MEMBUFFER_OFFSET(buffer))
+    {
+        SCFree(*((char**)(MEMBUFFER_BUFFER(buffer) + offset)));
```
inliniac: whats the goal here? Freeing in a membuffer is not how the API is meant to be used

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+        time_t lastProcessed = 0;
+
+        DIR * directory = NULL;
+
+        directory = opendir(pcap_g.dirpath);
+        if (directory == NULL)
+        {
+            switch (errno)
+            {
+                case EACCES:
```
inliniac: bad indent in this block. Also it might be good to move in a util func as this function is overly long

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+                SCLogError(SC_ERR_FOPEN, "Permission denied\n");
```
inliniac: text passed to the SCLog* API should not have a trailing newline

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+                        if (!RunModeUnixSocketIsActive()) {
+                            EngineStop();
+                        } else {
+                            UnixSocketPcapFolder(TM_ECODE_DONE);
+                            SCReturnInt(TM_ECODE_DONE);
+                        }
+                    }
+                }
+
+                char * mem = SCMalloc(written + 1);
```
inliniac: use SCStrdup instead?

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+        SCFree(*((char**)(MEMBUFFER_BUFFER(buffer) + offset)));
+        offset += sizeof(char*);
+    }
+    MemBufferFree(buffer);
+}
+
+/**
+ *  \brief Main PCAP file reading Loop function
+ */
+TmEcode ReceivePcapFolderLoop(ThreadVars *tv, void *data, void *slot)
```
inliniac: this function needs to be split up into smaller chunks

src/source-pcap-folder.c
```diff
@@ -0,0 +1,943 @@
...
+#  if LINUX_VERSION_CODE >= KERNEL_VERSION(2,6,21)
+#    define PCAP_FOLDER_ENABLE_MONITOR
+#    include <sys/inotify.h> // inotify
+#  endif
+#endif
+#include "util-buffer.h"
+
+extern int max_pending_packets;
+
+typedef struct PcapFolderGlobalVars_ {
```
inliniac: there is a lot of duplication of code here. I think this could be sharing a lot with the exiting pcap-file support.

jasonish:
Would it make more sense to use the term "directory" instead of "folder"? Suricata already uses the term directory instead of folder, so I think its best to be consistent.

PR #2770, opened 2017-06-12
stream: don't do protocol detection on gap - v1

jasonish:
A gap notification has no data.

Also, break out the gap handling into its own code block to
simplify the conditional statements.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2141

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/180
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/532

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -548,15 +548,25 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
         alproto = f->alproto_tc;
     }
 
+    /* If a gap notification, relay the notification on to the
+     * app-layer if known. */
+    if (flags & STREAM_GAP) {
+        if (alproto == ALPROTO_UNKNOWN) {
+            StreamTcpSetStreamFlagAppProtoDetectionCompleted(stream);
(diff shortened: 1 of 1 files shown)


src/app-layer.c
```diff
@@ -548,15 +548,25 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
         alproto = f->alproto_tc;
     }
 
+    /* If a gap notification, relay the notification on to the
+     * app-layer if known. */
+    if (flags & STREAM_GAP) {
+        if (alproto == ALPROTO_UNKNOWN) {
+            StreamTcpSetStreamFlagAppProtoDetectionCompleted(stream);
+            SCLogDebug("ALPROTO_UNKNOWN flow %p, due to GAP in stream start", f);        } else {
```
inliniac: this line looks a bit funny with the } else { there

jasonish: Just so happens that the text wrapping and my size of text editor window made it look correct! Will fix.

jasonish:
See https://github.com/inliniac/suricata/pull/2773

PR #2771, opened 2017-06-12
Prevent panic when malformed rrnames encountered

DominoTree:
Might be a better way to handle this, but this will prevent panics for now

### DIFF
--- a/rust/src/dns/log.rs
+++ b/rust/src/dns/log.rs
@@ -418,7 +418,7 @@ fn dns_log_json_failure(r: &DNSResponse, index: usize, flags: u64)
     js.set_string("type", "answer");
     js.set_integer("id", r.header.tx_id as u64);
     js.set_string("rcode", &dns_rcode_string(r.header.flags));
-    js.set_string("rrname", std::str::from_utf8(&query.name[..]).unwrap());
+    js.set_string("rrname", std::str::from_utf8(&query.name[..]).unwrap_or(""));
 
     return js.unwrap();
 }


DominoTree:
We occasionally have malformed DNS traffic cross the wire on our network which cause the Rust implementation to crash.  This patch will drop the malformed `rrname` and allow the software to continue running.

jasonish:
Thanks. I'll have to look at the other unwraps as well.

But I wonder if this should be validated during parsing so an event can be raised? We'd probably still want to log what we can of the DNS message, but perhaps a flag stating malformed name - or log a byte representation of the name?

Thoughts?

DominoTree:
That's not a bad idea, maybe we could just spit out a string with a bunch of hex and prefix it with "MALFORMED" or something - I feel like eve-log would still deal with it properly

jasonish:
@DominoTree Please see this commit https://github.com/inliniac/suricata/pull/2780/commits/19689f5651c4763d38576831b67ba55fa14e56bf which is part of PR https://github.com/inliniac/suricata/pull/2780.

The idea is to print what we can, hex for what we can't...

Not sure if I want to raise a malformed alert for this or not at this time, as the structure of the DNS message was valid if we get to logging.

DominoTree:
@jasonish That looks good to me.  I haven't looked at the RFCs in a while but I'm not sure that it's technically even malformed just because the DNS request can't be converted to valid UTF8.

inliniac:
Fixed in #2781, so closing. Thanks!

PR #2778, opened 2017-06-13
bypass: add explicit flag in stream engine

regit:
TCP reassembly is now deactivated more frequently and triggering a
bypass on it is resulting in missing some alerts due forgetting
about packet based signature.

So this patch is introducing a dedicated flag that can be set in
the app layer and transmitted in the streaming to trigger bypass.

It is currently used by the SSL app layer to trigger bypass when
the stream becomes encrypted.

- [*] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [*] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [*] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Redmin issue: https://redmine.openinfosecfoundation.org/issues/2143

PRscript:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/293
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/77

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -141,7 +141,7 @@ typedef struct AppLayerParserCtx_ {
 } AppLayerParserCtx;
 
 struct AppLayerParserState_ {
-    uint8_t flags;
+    uint16_t flags;
 
     /* Indicates the current transaction that is being inspected.
      * We have a var per direction. */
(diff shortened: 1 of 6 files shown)


src/app-layer-parser.h
```diff
@@ -35,6 +35,7 @@
 #define APP_LAYER_PARSER_NO_INSPECTION          0x02
 #define APP_LAYER_PARSER_NO_REASSEMBLY          0x04
 #define APP_LAYER_PARSER_NO_INSPECTION_PAYLOAD  0x08
+#define APP_LAYER_PARSER_BYPASS_READY           0x10
```
inliniac: can you switch this to BIT_U16(x) macro?

src/app-layer-parser.c
```diff
@@ -141,7 +141,7 @@ typedef struct AppLayerParserCtx_ {
 } AppLayerParserCtx;
 
 struct AppLayerParserState_ {
-    uint8_t flags;
+    uint16_t flags;
```
inliniac: could be good to add cocci flags check support here too

src/stream-tcp-private.h
```diff
@@ -183,6 +183,8 @@ enum
 #define STREAMTCP_STREAM_FLAG_NEW_RAW_DISABLED 0x0200
 /** Raw reassembly disabled completely */
 #define STREAMTCP_STREAM_FLAG_DISABLE_RAW 0x400
+/** Stream can be bypass */
+#define STREAMTCP_STREAM_FLAG_BYPASS 0x0800
 // vacancy 1x
```
inliniac: no more vacancy if you add a flag, so please remove this comment

src/stream-tcp-private.h
```diff
@@ -183,6 +183,8 @@ enum
 #define STREAMTCP_STREAM_FLAG_NEW_RAW_DISABLED 0x0200
 /** Raw reassembly disabled completely */
 #define STREAMTCP_STREAM_FLAG_DISABLE_RAW 0x400
+/** Stream can be bypass */
+#define STREAMTCP_STREAM_FLAG_BYPASS 0x0800
```
inliniac: why did you do this per stream? The only use is per session, so a session flag seems more appropriate

src/app-layer-parser.c
regit: uint8_t is enough should I switch to that ?

PR #2780, opened 2017-06-14
rust/dns - fix panics - v1

jasonish:
First commit fixes a panic due to bad validation of the input length of TCP DNS frames. Unit tests were added to check for this condition.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2144

The second commit addresses a panic in converting the rrname bytes to a printable string when the rrname contains bytes that don't convert to unicode. First we try to convert to unicode, if that fails the printable characters will be printed, with the unprintable ones printed like \xff.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2148

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/184
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/537

### DIFF
--- a/rust/src/dns/log.rs
+++ b/rust/src/dns/log.rs
@@ -19,6 +19,7 @@ extern crate libc;
 
 use std;
 use std::string::String;
+use std::ascii::AsciiExt;
 
 use json::*;
 use dns::dns::*;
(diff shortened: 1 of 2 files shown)


rust/src/dns/log.rs
```diff
@@ -286,6 +287,28 @@ pub fn dns_rrtype_string(rrtype: u16) -> String {
     }.to_string()
 }
 
+fn safe_bytes_to_string(input: &[u8]) -> String {
```
inliniac: have you considered using String::from_utf8_lossy ?

jasonish: Honestly I didn't, but looking at it I still think I prefer this approach so I can see the actual "lost" bytes instead of a replacement char used for lost bytes.

jasonish:
Oops, Rust 1.7.0 didn't like something there.

jasonish:
Rust 1.7.0 build issue, see https://github.com/inliniac/suricata/pull/2781

rust/src/dns/log.rs
inliniac: Yeah I think I agree. Just wanted to make sure it was considered :)

DominoTree: `String::from_utf8_lossy` reminds me of `payload_printable` in eve-log 😃

PR #2781, opened 2017-06-14
rust/dns - fix panics - v2

jasonish:
This PR fixes a compilation issue with Rust 1.7..0

First commit fixes a panic due to bad validation of the input length of TCP DNS frames. Unit tests were added to check for this condition.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2144

The second commit addresses a panic in converting the rrname bytes to a printable string when the rrname contains bytes that don't convert to unicode. First we try to convert to unicode, if that fails the printable characters will be printed, with the unprintable ones printed like \xff.

Another option would be to use from_utf8_lossy().

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2148

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/185
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/538

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -468,6 +468,8 @@ impl DNSState {
     ///
     /// Always buffer and read from the buffer. Should optimize to skip
     /// the buffer if not needed.
+    ///
+    /// Returns the number of messages parsed.
     pub fn parse_request_tcp(&mut self, input: &[u8]) -> i8 {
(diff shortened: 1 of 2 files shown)


inliniac:
Merged in #2782, thanks Jason!

PR #2783, opened 2017-06-15
Rust: fix tx ID assignment (0 was never assigned)

chifflier:

Describe changes:
- This is a trivial change in the Rust code to increment tx id **after** assigning ID to the transaction. Otherwise, first transaction get ID 1, while suricata will request ID 0 as first valid ID.
- not sure if other parsers (C) are affected (I checked SMTP, it is not)

### DIFF
--- a/rust/src/nfs/nfs3.rs
+++ b/rust/src/nfs/nfs3.rs
@@ -320,9 +320,9 @@ impl NFS3State {
 
     pub fn new_tx(&mut self) -> NFS3Transaction {
         let mut tx = NFS3Transaction::new();
-        self.tx_id += 1;
         tx.id = self.tx_id;
-        return tx;
+        self.tx_id += 1;
+        tx
     }
 
     pub fn free_tx(&mut self, tx_id: u64) {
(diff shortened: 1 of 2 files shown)


rust/src/nfs/nfs3.rs
```diff
@@ -320,9 +320,9 @@ impl NFS3State {
 
     pub fn new_tx(&mut self) -> NFS3Transaction {
         let mut tx = NFS3Transaction::new();
-        self.tx_id += 1;
         tx.id = self.tx_id;
-        return tx;
+        self.tx_id += 1;
```
inliniac: I don't think this is correct, if you look at the lookup code in get_tx_by_id you can see it accounts for the off by one: if tx.id == tx_id + 1 {

inliniac: same on the dns side, as I copied the logic from dns

jasonish:
I don't know why the C DNS does this, but when I did DNP3, which I used DNS as a template for I tried to use 0 for the first tx and I believe I rant into issues in transaction logging.

Anyways, if you fix up the other bits in dns/dns.rs and dns/log.rs to deal with a 0 based index, it appears that it works correctly. At least all my DNS tests at https://github.com/jasonish/suricata-verify which verify the output all pass. Otherwise I'm in favour of a 0 based index, its simpler.

chifflier:
Please do not consider merging yet, still investing if the +1 is required or not

chifflier:
Canceling, this is an incorrect solution. Sorry for the noise

PR #2784, opened 2017-06-15
python: use python path found during configure - v1

jasonish:
Also look for Python under more names. For example, on OpenBSD
if you just install Python 2, you will only get a python2.7
executable.

Also renames HAVE_PYTHON_CONFIG to simply HAVE_PYTHON for cosmetic reasons.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/186
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/539

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -86,7 +86,7 @@
 
     AC_ARG_ENABLE(python,
            AS_HELP_STRING([--enable-python], [Enable python]),,[enable_python=yes])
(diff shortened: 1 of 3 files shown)


inliniac:
Merged through #2786, thanks Jason!

PR #2790, opened 2017-06-17
Rust dns string fixes 3

jasonish:
More fixing of Rust to DNS string handling.

Move the safe string handling into the json module for now. It will replace non-printable characters with a hex code. Ubuntu 14.04 still doesn't have json_stringn, so we can't pass full unicode strings to it yet. So this might be a good middle ground for now without decoding unicode.  For example, we could pass through the unicode bytes, but a but like \xff will cause jansson not to log that field.

Likewise for Lua, use pushlstring so we don't have to worry about NULLs there.

Also updates Travis to use Ubuntu 14.04.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/192
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/545

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 sudo: false
 language: c
 
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -143,18 +143,6 @@ pub struct DNSQueryEntry {
     pub rrclass: u16,
 }
 
-impl DNSQueryEntry {
-
-    pub fn name(&self) -> &str {
-        let r = std::str::from_utf8(&self.name);
-        if r.is_err() {
-            return "";
-        }
-        return r.unwrap();
-    }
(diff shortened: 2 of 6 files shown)


jasonish:
Something that might be useful is a configure check for json_stringn.. If it exists, we could pass valid unicode strings to it, then fall back to this safe string printing..  So that if you at json_stringn, you'll get a larger printable string set.

inliniac:
We have a similar issue in the SCLog* layer.

jasonish:
> We have a similar issue in the SCLog* layer.

This is a little safer as we mostly control what we log with SCLog functions. However, I'll probably change it to CString::from_vec_unchecked just in case someones really wants to log a string with a null byte. This will result in the same as if logged in C, the string will only be printed up to the null byte.

PR #2794, opened 2017-06-20
Rust add external parsers 2

chifflier:

Describe changes:
- This PR adds the '--enable-rust-external' configure flag, for (embedded) rust app layers with external dependencies
- It adds support for the NTP protocol, implemented in Rust. The decoding logic is embedded, only the raw parser is an external rust crate (https://github.com/rusticata/ntp-parser)
- The parser is mostly used as an example, and does protocol validation for NTP version 3 and 4.
- Some event rules have been added, to detect malformed NTP data
- This is meant to be an example parser, before working on similar integration for IKEv2 (and later, SNMP, SSH alternative parser, etc.)
- All comments from #2792 have been merged: directories, debug statements, build warnings, rules sids

Thanks,
Pierre

### DIFF
--- a/rules/ntp-events.rules
+++ b/rules/ntp-events.rules
@@ -1,0 +1,3 @@
+alert ntp any any -> any any (msg:"SURICATA NTP malformed request data"; flow:to_server; app-layer-event:ntp.malformed_data; classtype:protocol-command-decode; sid:2222001; rev:1;)
+alert ntp any any -> any any (msg:"SURICATA NTP malformed response data"; flow:to_client; app-layer-event:ntp.malformed_data; classtype:protocol-command-decode; sid:2222002; rev:1;)
+
--- a/rust/src/ntp/ntp.rs
+++ b/rust/src/ntp/ntp.rs
@@ -84,6 +84,9 @@ impl NTPState {
 }
 
 impl NTPState {
+    /// Parse an NTP request message
+    ///
+    /// Returns The number of messages parsed, or -1 on error
     fn parse(&mut self, i: &[u8], _direction: u8) -> i8 {
         match parse_ntp(i) {
             IResult::Done(_,ref msg) => {
@@ -94,7 +97,7 @@ impl NTPState {
                     tx.xid = msg.ref_id;
                     self.transactions.push(tx);
                 }
-                0
+                1
             },
             IResult::Incomplete(_) => {
                 SCLogDebug!("Insufficient data while parsing NTP data");
@@ -366,4 +369,25 @@ pub extern "C" fn rs_ntp_state_get_event_info(event_name: *const libc::c_char,
         *event_id = event as libc::c_int;
     };
     0
+}
+
+#[cfg(test)]
+mod tests {
+    use super::NTPState;
+
+    #[test]
+    fn test_ntp_parse_request_valid() {
+        // A UDP NTP v4 request, in client mode
+        const REQ : &[u8] = &[
+            0x23, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x18, 0x57, 0xab, 0xc3, 0x4a, 0x5f, 0x2c, 0xfe
+        ];
+
+        let mut state = NTPState::new();
+        assert_eq!(1, state.parse(REQ, 0));
+    }
 }
--- a/rust/src/ntp/ntp.rs
+++ b/rust/src/ntp/ntp.rs
@@ -1,0 +1,369 @@
+/* Copyright (C) 2017 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+// written by Pierre Chifflier  <chifflier@wzdftpd.net>
+
+extern crate ntp_parser;
+use self::ntp_parser::*;
+use core;
+use applayer;
+use libc;
+use std;
+use std::ffi::CStr;
+
+use log::*;
+
+use nom::IResult;
+
+#[repr(u32)]
+pub enum NTPEvent {
+    UnsolicitedResponse = 0,
+    MalformedData,
+    NotRequest,
+    NotResponse,
+}
+
+
+
+pub struct NTPState {
+    /// List of transactions for this session
+    transactions: Vec<NTPTransaction>,
+
+    /// Detection engine states counter
+    de_state_count: u64,
+
+    /// Events counter
+    events: u16,
+
+    /// tx counter for assigning incrementing id's to tx's
+    tx_id: u64,
+}
+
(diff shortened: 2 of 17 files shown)


jasonish:
Should we really have --enable-rust-external? I'm unclear what it gets us. What happens if we use an external nom parser for something more critical like TLS or HTTP2? Would these be wrapped behind --enable-rust-external as well?

inliniac:
The --enable-rust-external is meant to be a clear separation between the parser that we develop/bundle, review and support on one side, and the parsers that depend on (potentially lots of) external crates that are developed by third parties, have different licenses and ownership, different goals and release schedules on the other side. I consider it an experiment within the Rust experiment. Perhaps in the future we can do away with this, but for now I want to keep things manageable and cleanly separated.

This will also mean that 'core' code cannot depend on these modules for now.

jasonish:
Just to be a bit pedantic, NFS and DNS rely on external crates that are developed by third parties, as well. The name to me means external Rust code will not be pulled in (a technical thing). Where it is actually a support level type of thing. If its short term, its probably OK. If its going to exist long term, perhaps the name needs a rethink? --enable-rust-experimental to enable experimental rust modules?

inliniac:
I'm ok with finding another name if that makes things more clear.

inliniac:
In general, with Rust and C all the same, we need to be careful about accepting external dependencies. Every one of them is a liability. I've spent countless hours on working around issues in libraries we use. We also see problems with original developers going away, libhtp is the biggest example for that, but also lua plugins that are buggy but unmaintained, etc.

In my mind something like nom and crc is more comparable to libpcre, libnss, etc. General purpose libraries/frameworks that are (hopefully) well maintained. While the protocol dissectors, at this stage, are more comparable to libhtp. Small separate projects that are much more likely to need updates by us if we would depend on them. That's why we maintain libhtp, we need to change things in there too often.

Maybe things will be better because of cargo, but it remains to be seen how distro's handle packaging. Already some crates are broken on Ubuntu LTS' version of rust.

chifflier:
Hi,
I used the external keywords for the reasons Victor described. But it's also true that it also represents the experimental side of the parser, in some way. External crates often need to have exact version requirements to avoid breaking when the the external crate is updated and breaks the API.
If you think another name is better please tell me, it's quite trivial to change.

Some random thoughts:
I think Rust support is already marked as experimental, so adding one layer of experimental will just make documentation a bit weird :) Maybe something like -staging ? Or, if the Rust support becomes mainstream, rename it at that moment ?

jasonish:
Sounds like in the long run, given that external crates are going to have different licenses, and different levels of quality, that a more fine grained approach might be needed, such as:

--enable-ntp
--enable-ike

And so on, rather than having one flag pull in a grab bag of licenses. This can probably wait though.

Note: I've left "rust" out of the examples above. An end user will need to know that Rust is required to build Suricata (when it becomes required). They do not need to know, and should not need to care that a feature is written in Rust or C.

But I think my main issue with the name "external" now has nothing to do with licenses, support, etc. It just sounds like it disables pulling in any external Rust code.

chifflier:
Ok, so what do you prefer (for this pull request, not on the long term):
- something like --enable-rust-experimental ?
- something like --enable-rust-third-party, --enable-rust-external or --enable-rust-unsupported to emphasize on the fact there is a part not supported by OISF
- any other idea

I have no clear preference, I would just like to avoid one word to block the PR :)

src/app-layer-ntp.c
```diff
@@ -0,0 +1,307 @@
...
+    return 0;
+}
+
+void RegisterNTPParsers(void)
+{
+    const char *proto_name = "ntp";
+
+    /* Check if NTP UDP detection is enabled. If it does not exist in
+     * the configuration file then it will be enabled by default. */
+    if (AppLayerProtoDetectConfProtoDetectionEnabled("tcp", proto_name)) {
```
inliniac: tcp used instead of udp

src/app-layer-ntp.c
```diff
@@ -0,0 +1,307 @@
...
+
+            SCLogDebug("Unittest mode, registeringd default configuration.");
+            AppLayerProtoDetectPPRegister(IPPROTO_UDP, NTP_DEFAULT_PORT,
+                ALPROTO_NTP, 0, NTP_MIN_FRAME_LEN, STREAM_TOSERVER,
+                NTPProbingParser, NULL);
+
+        }
+        else {
+
+            if (!AppLayerProtoDetectPPParseConfPorts("tcp", IPPROTO_UDP,
```
inliniac: tcp

src/app-layer-ntp.c
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    else {
+        SCLogNotice("Protocol detecter and parser disabled for NTP.");
+        return;
+    }
+
+    if (AppLayerParserConfParserEnabled("udp", proto_name)) {
+
+        SCLogNotice("Registering NTP protocol parser.");
```
inliniac: notice

src/app-layer-ntp.c
```diff
@@ -0,0 +1,307 @@
...
+        AppLayerParserRegisterDetectStateFuncs(IPPROTO_UDP, ALPROTO_NTP,
+            NULL, NTPGetTxDetectState, NTPSetTxDetectState);
+
+        AppLayerParserRegisterGetEventInfo(IPPROTO_UDP, ALPROTO_NTP,
+            NTPStateGetEventInfo);
+        AppLayerParserRegisterGetEventsFunc(IPPROTO_UDP, ALPROTO_NTP,
+            NTPGetEvents);
+    }
+    else {
+        SCLogNotice("NTP protocol parsing disabled.");
```
inliniac: notice

rules/ntp-events.rules
```diff
@@ -0,0 +1,3 @@
+alert ntp any any -> any any (msg:"SURICATA NTP malformed request data"; flow:to_server; app-layer-event:ntp.malformed_data; classtype:protocol-command-decode; sid:2222001; rev:1;)
```
inliniac: can you add a reference to the sid assignment page?

rust/src/ntp/ntp.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    fn free(&mut self) {
+        if self.events != std::ptr::null_mut() {
+            core::sc_app_layer_decoder_events_free_events(&mut self.events);
+        }
+    }
+}
+
+impl Drop for NTPTransaction {
```
inliniac: do you see this being used? I think all tx frees are explicit, or at least they should be as the C-side is responsible.

inliniac:
On 21-06-17 08:45, Pierre Chifflier wrote:
> Ok, so what do you prefer (for this pull request, not on the long term):
> 
>   * something like --enable-rust-experimental ?
>   * something like --enable-rust-third-party, --enable-rust-external or
>     --enable-rust-unsupported to emphasize on the fact there is a part
>     not supported by OISF
>   * any other idea
> 
> I have no clear preference, I would just like to avoid one word to block
> the PR :)

No strong preference on my end. So Jason if you have a strong preference
let us know, otherwise I'm just going to accept as-is for now.

rust/src/ntp/ntp.rs
chifflier: The Drop Trait is called when the object is destroyed, this is a destructor.

Transactions are freed here:
https://github.com/inliniac/suricata/pull/2794/files#diff-1652b119a6a57d40eb26e9b50c527dcdR115
I made the `self.transactions.clear()` explicit, but it is not required (vector is cleared when the object becomes out of reach).

When C calls `rs_ntp_state_free`, the state is freed, automatically freeing the transactions.
The other code path to free a transaction is `rs_ntp_state_tx_free`, which gets the object in a temporary variable dropped immediately (calling the `Drop` Trait).
See `free_tx`:
https://github.com/inliniac/suricata/pull/2794/files#diff-1652b119a6a57d40eb26e9b50c527dcdR134

jasonish:
> On 21-06-17 08:45, Pierre Chifflier wrote:
>  Ok, so what do you prefer (for this pull request, not on the long term):
> 
>    * something like --enable-rust-experimental ?
>    * something like --enable-rust-third-party, --enable-rust-external or
>      --enable-rust-unsupported to emphasize on the fact there is a part
>      not supported by OISF
>    * any other idea
> 
>  I have no clear preference, I would just like to avoid one word to block
>  the PR :)
> No strong preference on my end. So Jason if you have a strong preference
> let us know, otherwise I'm just going to accept as-is for now.
> Merge state
>  Show all checks

I think I'd lean to towards --enable-rust-experimental? Reason being, in the past 2 days I've referred to external crates 2 times now, in reference to nom, and a time handling one I thought about pulling but might not be needed now since Ubuntu updated its Rust.  Though I suppose saying "crate" and "external" together is redundant.

chifflier:
Replaced by #2798

PR #2795, opened 2017-06-20
rust - cargo home, logging fixups - 1

jasonish:
When following the procedure of "make && sudo make install", Cargo will re-download and rebuild the Rust components as root $HOME/.cargo is used during the build. To fix this, save the path to cargo (for rustup users), and determine the path for CARGO_HOME at configure time and expose them as variables to the build so the "make" and "sudo make install" will see the same values preventing a rebuild.

Also do safe string handling in the SCLog routines and include a pure Rust logger for when not running as part of Suricata (unit tests).

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/196
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/549

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2007,9 +2007,15 @@
           CFLAGS="${CFLAGS} -I../rust/gen/c-headers"
           AC_SUBST(RUST_SURICATA_LIB)
           AC_SUBST(RUST_LDADD)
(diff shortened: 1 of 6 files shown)


jasonish:
Closing. See https://github.com/inliniac/suricata/pull/2801

PR #2799, opened 2017-06-22
Alert app proto

regit:

Describe changes:
- Add app_proto to alert events
- Some doc on EVE

PR builds: 
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/297
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/81

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -243,6 +243,10 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
         }
     }
 
+    if (p->flow) {
+        json_object_set_new(js, "app_proto",
+                json_string(AppProtoToString(p->flow->alproto)));
+    }
     /* Add tx_id to root element for correlation with other events. */
     json_object_del(js, "tx_id");
     if (pa->flags & PACKET_ALERT_FLAG_TX)
--- a/doc/userguide/output/eve/eve-json-format.rst
+++ b/doc/userguide/output/eve/eve-json-format.rst
@@ -74,6 +74,7 @@ the signature.
      "gid": 1,
      "signature_id": 1,
      "rev": 1,
+     "app_proto": "http",
      "signature": "HTTP body talking about corruption",
      "severity": 3,
      "source": {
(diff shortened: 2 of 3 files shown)


src/output-json-alert.c
```diff
@@ -243,6 +243,10 @@ void AlertJsonHeader(const Packet *p, const PacketAlert *pa, json_t *js)
         }
     }
 
+    if (p->flow) {
+        json_object_set_new(js, "app_proto",
```
inliniac: 2 thoughts:

1. perhaps it makes sense to turn the code here into a util func: https://github.com/inliniac/suricata/blob/master/src/output-json-flow.c#L191 and call that. That gives a bit more info in some sigs, like for the app-layer protocol detect warnings.

2. alternatively, what about adding the partial flow record to the alert record? Just like how we add http etc we can add the flow record. This would then automatically include the app_proto

regit: IMO, 2. may be a better idea, getting metric information could be interesting. Maybe we could do: print partial flow record or output app_proto ?

inliniac: in "print partial flow record or output app_proto" what would control which way to go?

regit: If `app-layer` or `flow` is set to yes then we print partial flow record (that include `app_proto`). If not we just print `app_proto`.

inliniac: i guess a flow switch could work, defaulting to 'enabled'. I think the eve block is getting a bit cluttered though, which is why I added 'applayer' to replace the individual protocol switches.

Perhaps we can enable all these things by default and only mention the options in the yaml documentation. That leaves a cleaner yaml.

regit: On doc that's what I've done in this PR.

I'm gonna try to implement the flow logging in the coming days.

inliniac: What I meant on the doc side is to have certain options appear *only* in the docs, so that the default yaml we ship isn't as verbose. So in this example, if we add a 'flow' switch to trigger flow logging, then don't add it to the default yaml, but do add docs for it. So ppl can disable it, but by default it's enabled and hidden from yaml. Make sense?

regit: Yes I did understand it that way.

PR #2807, opened 2017-06-26
log: wrap rotation and write in lock

jasonish:
The application log is subject to rotation, so the check for
rotation, the actual rotation and write needs to be done under
lock to ensure the file pointer is in a consisten state
at the time of write().

Fixes issue:
https://redmine.openinfosecfoundation.org/issues/2155

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/199
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/552

### DIFF
--- a/src/util-debug.c
+++ b/src/util-debug.c
@@ -577,11 +577,13 @@ SCError SCLogMessage(const SCLogLevel log_level, const char *file,
                                           log_level, file, line, function,
                                           error_code, message) == 0)
                 {
+                    SCMutexLock(&op_iface_ctx->fp_mutex);
                     if (op_iface_ctx->rotation_flag) {
                         SCLogReopen(op_iface_ctx);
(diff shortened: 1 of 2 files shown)


src/util-debug.c
```diff
@@ -714,6 +716,7 @@ static inline SCLogOPIfaceCtx *SCLogInitFileOPIface(const char *file,
         goto error;
     }
 
+    SCMutexInit(&iface_ctx->fp_mutex, NULL);
```
inliniac: no destroy anywhere?

inliniac: added it myself

inliniac:
Merged through #2814, thanks Jason! I added a destroy call as well.

PR #2809, opened 2017-06-27
Suripcap v6

regit:
Update of #2806 with a rebase and a fix on python module installation.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/306
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/90

PR #2810, opened 2017-06-27
Suripcap v7

regit:
Update of #2809 fixing travis-ci.

PR builds:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/307
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/91


inliniac:
I'm not yet convinced we need to make this a part of the suricata repo. It can be used in travis from an external repo as well, just like how we pull in libhtp during travis builds.

cc @jasonish

regit:
Positive points on having it inside:
- tests can be dependent of the version (as output evolve)
- it is here so developers can use it on their own tests directly

Negative points:
- We can not easily get big pcap in the repo

regit:
Maybe we can make this evolve to have some yaml tests used by travis that depends on a series of pcap files stored in a diff repository.

jasonish:
> I'm not yet convinced we need to make this a part of the suricata repo. It can be used in travis from an external repo as well, just like how we pull in libhtp during travis builds.
> 
> cc @jasonish

Yeah, I agree. I'm not sure if it should exist in the repo. But of course, I'm biased by having done https://github.com/jasonish/suricata-verify.

What I may add next is the ability to specify pcap by URL, then they will be downloaded and cached. Still a risk of them disappearing which is why they are currently all checked in, but it could be an option for large ones.

inliniac:
I'll let you guys fight this out @jasonish and @regit ;)

regit:
As discussed in Cork, we could push the suripcap code into a separate repo (OISF maintained) and integrate it via travis-ci.

PR #2813, opened 2017-06-27
rust/dns: handle multiple txt strings - 1

jasonish:
Fix handling of TXT records when there are multiple strings
in a single TXT record. For now, conform to the C implementation
where an answer record is created for each string in a single
txt record.

Also removes the data_len field from the answer entry. In Rust,
the length is available from actual data, which after decoding
may actually be different than the encoded data length, so just
use the length from the actual data.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/200
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/553

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -149,7 +149,6 @@ pub struct DNSAnswerEntry {
     pub rrtype: u16,
     pub rrclass: u16,
     pub ttl: u32,
-    pub data_len: u16,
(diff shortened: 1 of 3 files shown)


inliniac:
Merged through #2814, thanks Jason. I added a destroy call as well.

PR #2736, opened 2017-06-02
Feature 2114 redis rpushoutput - based on master

securitym0nkey:
New PR for feature 2114 as it was ask to rebase #2724 on master. 'Rewrote' the code to fit with the master branch.

PR #2821, opened 2017-07-02
Feature 2114 redis rpushoutput master

securitym0nkey:

Describe changes:

This is a resubmit for PR #2736 now with a doc update.
It implements my own Feature Request [2114](https://redmine.openinfosecfoundation.org/issues/2114)

### DIFF
--- a/doc/userguide/configuration/suricata-yaml.rst
+++ b/doc/userguide/configuration/suricata-yaml.rst
@@ -268,7 +268,9 @@ integration with 3rd party tools like logstash.
       #  server: 127.0.0.1
       #  port: 6379
       #  async: true ## if redis replies are read asynchronously
-      #  mode: list ## possible values: list (default), rlist (uses RPUSH instead of LPUSH), channel
+      #  mode: list ## possible values: list|lpush (default), rpush, channel|publish
+      #             ## lpush and rpush are using a Redis list. "list" is an alias for lpush
+      #             ## publish is using a Redis channel. "channel" is an alias for pubish
       #  key: suricata ## key or channel to use (default to suricata)
       # Redis pipelining set up. This will enable to only do a query every
       # 'batch-size' events. This should lower the latency induced by network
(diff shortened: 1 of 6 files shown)


doc/userguide/configuration/suricata-yaml.rst
```diff
@@ -268,7 +268,7 @@ integration with 3rd party tools like logstash.
       #  server: 127.0.0.1
       #  port: 6379
       #  async: true ## if redis replies are read asynchronously
-      #  mode: list ## possible values: list (default), channel
+      #  mode: list ## possible values: list (default), rlist (uses RPUSH instead of LPUSH), channel
```
inliniac: I'm not sure why 'list' was named that way originally, but to me it would make sense to actually use 'lpush' and 'rpush' as names instead. Perhaps we can add that right away with backwards compatibility. Thoughts?

inliniac:
In this case a single commit will be sufficient. It's subject should be prefixed with 'redis:'

doc/userguide/configuration/suricata-yaml.rst
securitym0nkey: I agree that "list" was not the perfect name in the first place. Like the idea to use "lpush" and "rpush" while keeping "list" as an "alias" for "lpush" for backwards compatibility.

securitym0nkey: @inliniac i did an additional look at this. I would suggest not just using 'lpush' and 'rpush' with backwards compatibility of 'list' i would also propose to use 'publish' for the 'channel' mode. That would be more consistent as 'publish' the redis command for channels like [lr]push are the commands for lists. 

I also noticed that currently we use channel whenever mode is configured to any value unequal to "list". I think we either should exit here with an error or log a warning and fallback to the default which would be "list".

What do you think?

securitym0nkey:
hi @inliniac,
what do you think about my latest changes to this branch?

inliniac:
This looks sane to me. @fooinha @jasonish any thoughts?

jasonish:
Yes, I'm in favour of this.

inliniac:
@securitym0nkey can you resubmit this in a clean PR with a single commit? Thanks!

PR #2823, opened 2017-07-03
New option for logging hash of truncated files

aoshiken:
Hi,

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2147

Describe changes:
- Added a new option for eve-log files type named 'force-hash-truncated', this way the user can log the hash for truncated files if needed, the option is disabled by default.

Regarding the documentation I've just changed one file because I don't see clearly what files to change

PR #2824, opened 2017-07-06
New option for logging hash of truncated files

aoshiken:
The prior iteration of this PR was #2823 

https://redmine.openinfosecfoundation.org/issues/2147

Changes since PR #2823 :
- Fixed coding style according to documentation
- Squashed some commits for comments
- Reverting part of commit that moved code

### DIFF
--- a/src/output-filedata.c
+++ b/src/output-filedata.c
@@ -124,6 +124,32 @@ static int CallLoggers(ThreadVars *tv, OutputLoggerThreadStore *store_list,
     return file_logged;
 }
 
+
+/* for hashing truncated files... */
+#ifdef HAVE_NSS
+static void EndHashTruncated( File *ff )
+{
+    if ( ff != NULL ) {
+        if (ff->md5_ctx) {
+            unsigned int len = 0;
+            HASH_End(ff->md5_ctx, ff->md5, &len, sizeof(ff->md5));
+            ff->flags |= FILE_MD5;
+        }
+        if (ff->sha1_ctx) {
+            unsigned int len = 0;
+            HASH_End(ff->sha1_ctx, ff->sha1, &len, sizeof(ff->sha1));
+            ff->flags |= FILE_SHA1;
+        }
+        if (ff->sha256_ctx) {
+            unsigned int len = 0;
+            HASH_End(ff->sha256_ctx, ff->sha256, &len, sizeof(ff->sha256));
+            ff->flags |= FILE_SHA256;
+        }
+    }
+}
+#endif
+
+
 static TmEcode OutputFiledataLog(ThreadVars *tv, Packet *p, void *thread_data)
 {
     BUG_ON(thread_data == NULL);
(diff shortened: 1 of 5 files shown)


src/output-json-file.c
```diff
@@ -78,6 +78,42 @@ typedef struct JsonFileLogThread_ {
     MemBuffer *buffer;
 } JsonFileLogThread;
 
+/* for hashing truncated files... */
+#ifdef HAVE_NSS
+static void FileWriteJsonHash( const File *ff, json_t *fjs )
```
inliniac: I think it makes sense to also use this in the default FILE_STATE_CLOSED case

src/util-file.c
```diff
@@ -875,7 +915,24 @@ static int FileCloseFilePtr(File *ff, const uint8_t *data,
     if ((flags & FILE_TRUNCATED) || (ff->flags & FILE_HAS_GAPS)) {
         ff->state = FILE_STATE_TRUNCATED;
         SCLogDebug("flowfile state transitioned to FILE_STATE_TRUNCATED");
-
+        /* for hashing truncated files... */
+#ifdef HAVE_NSS
```
inliniac: perhaps this can be turned into a small util func and be called from both truncated and closed cases

src/output-filedata.c
```diff
@@ -124,6 +124,32 @@ static int CallLoggers(ThreadVars *tv, OutputLoggerThreadStore *store_list,
     return file_logged;
 }
 
+
+/* for hashing truncated files... */
+#ifdef HAVE_NSS
+static void EndHashTruncated( File *ff )
```
inliniac: I don't understand in what case we actually need this. When a file is marked as 'truncated' the same logic is added in this PR, so in what case is this needed?

src/output-json-file.c
aoshiken: But the default state seems to be UNKNOWN, not FILE_STATE_CLOSED, that's another case different from the default one and the hash its already calculated within so I suppose you're referring to the UNKNOWN default case.

Sounds right to you or am I wrong?

And what about FILE_STATE_ERROR? Should I add the "FileWriteJsonHash()" there?

src/output-filedata.c
aoshiken: I'm sorry but I don't remember exactly, I recall some flow path leading me to here but probably you're right and I'm wrong so I'm going to remove this in the next PR iteration

inliniac: I think the file state should always be either closed, error or truncated when it is logged, otherwise we have a bug in the file tracking.

src/util-file.c
aoshiken: Agreed, in fact in the first PR version was coded as you say but I removed it for the sake of changing as less as possible

src/output-json-file.c
inliniac: I meant call this util func for case FILE_STATE_CLOSED: as well. Now the code is duplicated between FileWriteJsonRecord and FileWriteJsonHash

aoshiken: Ah! Ok... :)

aoshiken:
New PR #2898  with the requested changes.

Regards

PR #2837, opened 2017-07-13
rust/dns/tcp - probe even if payload is short - 1

jasonish:
As the DNS probe just uses the query portion of a response, don't
require there to be as many bytes as specified in the TCP DNS
header. This can occur in large responses where probe is called
without all the data.

Fixes the cases where the app proto is recorded as failed.

Fixes issue:
https://redmine.openinfosecfoundation.org/issues/2169

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/201
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/554

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -546,9 +546,7 @@ fn probe(input: &[u8]) -> bool {
 pub fn probe_tcp(input: &[u8]) -> bool {
     match nom::be_u16(input) {
         nom::IResult::Done(rem, len) => {
-            if rem.len() >= len as usize {
-                return probe(rem);
-            }
+            return probe(rem);
         },
         _ => {}
     }


inliniac:
Would it make sense to have an approach like this? https://github.com/inliniac/suricata/blob/master/rust/src/nfs/nfs.rs#L2075

jasonish:
I don't know. My intention in probing is simply to see if its parseable.. Validation of the response flag on responses, etc comes after parsing. Could do a minimum message size check beforehand I suppose to avoid attempting to parse messages that have no chance of succeeding.

inliniac:
Merged in #2838 with a small compile warning fix.

PR #2841, opened 2017-07-14
dnp3: use BasicSearch instead of memmem

jasonish:
Mingw doesn't support memmem.

Note, there is a unit test, DNP3ProbingParserTest that tests a probe with a banner, which uses this function.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/203
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/556


inliniac:
Merged as part of #2850, thanks Jason!

PR #2257, opened 2016-09-21
flash decompression v4

glongo:
This adds new decompression API for flash files compressed with zlib/lzma.

This version addresses comment in the last PR: https://github.com/inliniac/suricata/pull/1793

Ticket: https://redmine.openinfosecfoundation.org/issues/1600

Prscript:
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/126
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/125

PR #2843, opened 2017-07-17
Flash decompression v5

glongo:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/1600

Describe changes:
- Address comment in #2257 
- Optimize buffer decompression

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/139
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/3

### DIFF
--- a/src/detect-engine-hsbd.c
+++ b/src/detect-engine-hsbd.c
@@ -56,6 +56,7 @@
 
 #include "util-unittest.h"
 #include "util-unittest-helper.h"
+#include "util-file-decompression.h"
 #include "app-layer.h"
 #include "app-layer-htp.h"
 #include "app-layer-htp-mem.h"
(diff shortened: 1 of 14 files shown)


src/detect-engine-hsbd.c
```diff
@@ -270,12 +299,20 @@ int DetectEngineInspectHttpServerBody(ThreadVars *tv,
...
-    const uint8_t *buffer = DetectEngineHSBDGetBufferForTX(tx, tx_id,
-                                                     de_ctx, det_ctx,
-                                                     f, htp_state,
-                                                     flags,
-                                                     &buffer_len,
-                                                     &stream_start_offset);
+    const uint8_t *buffer = NULL;
+
+    if (det_ctx->decompressed_hsbd_buffer) {
+        buffer = det_ctx->decompressed_hsbd_buffer;
```
inliniac: Can you look at how DetectEngineHSBDGetBufferForTX manages it's buffers. It uses an array in det_ctx that contains the actual buffers so that we can have one per tx (in case we inspect multiple txs at the same time). It automatically makes sure that 2+ calls result in only one buffer creation, the next calls just use the cached results. I would like the flash functionality to use this same logic.

inliniac:
The OSX builder fails due to missing lzma header. That should be handled better.

src/detect-engine-hsbd.c
inliniac: Also, ideally there is NO memory allocation for the decompression most of the time. In the regular server body case memory allocations are only done to grow the buffer. In other cases the memory is simply reused.

PR #2848, opened 2017-07-19
detect: reject dsize rules that can't match

inliniac:
Rules can contain conflicting statements and lead to a unmatchable rule.

2 examples are rejected by this patch:

1. dsize < content
2. dsize < content@offset

Bug #2187


inliniac:
Merged as part of #2850

PR #2858, opened 2017-07-25
Make util-hash.c faster

paragonie-scott:
Restriction: Hash table must be an even power of 2. This appears to be the case everywhere it's used, however.

Describe changes:
- Optimize hashing, as per http://mziccard.me/2015/05/08/modulo-and-division-vs-bitwise-operations/
- Requires hash table to be a power of 2, as per https://blog.mozilla.org/nnethercote/2014/11/04/please-grow-your-buffers-exponentially/

### DIFF
--- a/src/util-hash.c
+++ b/src/util-hash.c
@@ -38,7 +38,7 @@ HashTable* HashTableInit(uint32_t size, uint32_t (*Hash)(struct HashTable_ *, vo
     if (size == 0) {
         goto error;
     }
-    if (size & (size - 1) == 0) {
+    if ((size & (size - 1)) != 0) {
         //printf("ERROR: Size must be a power of 2\n"); 
         goto error;
     }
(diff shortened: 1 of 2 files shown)


inliniac:
The problem here is that we have config options that directly translate to the size arg passed to HashTableInit. As we haven't enforced a power of 2 limit before we may end up breaking a lot of setups.

This doesn't mean it's a bad idea, but it does mean it's not going to make 4.0. But even for the next major version I'd like to see some actual perf analysis before introducing breaking changes.

paragonie-scott:
Everywhere this code is called in the code, it currently uses a power of 2. https://github.com/inliniac/suricata/search?utf8=%E2%9C%93&q=HashTableInit&type=

inliniac:
Ah good point, I mixed things up with the flow hash and other similar hashes.

paragonie-scott:
Yeah, I hadn't looked at those. I also missed a constant that was `1000` because I misread it to be `65536`.

paragonie-scott:
This ought to be safe to merge now. Please do feel free to benchmark it first, of course.

PR #2864, opened 2017-08-03
Fix bug with TLS client certificates (Redmine Bug #2050)

thus:
Decoding client certificate overwrites the validity dates from the server certificate, so we therefore don't decode it, since we don't do anything with it (right now) anyway.

Fixes:

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/127
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/127


inliniac:
Merged in #2868, thanks Mats!

PR #2869, opened 2017-08-03
prelude: add protocol information through JSON

ToToL:
Prelude is based on what can be found in packet but now, Suricata provide more information in JSON format for HTTP, TLS, SSH and SMTP. This PR add the support for prelude to read these JSON and integrate them into IDMEF alerts.

### DIFF
--- a/src/alert-prelude.c
+++ b/src/alert-prelude.c
@@ -53,6 +53,16 @@
 #include "util-print.h"
 
 #include "output.h"
+
+#ifdef HAVE_LIBJANSSON
+#include "output-json.h"
+#include "output-json-http.h"
+#include "output-json-tls.h"
+#include "output-json-ssh.h"
+#include "output-json-smtp.h"
+#include "output-json-email-common.h"
+#endif
+
 #include "util-privs.h"
 #include "util-optimize.h"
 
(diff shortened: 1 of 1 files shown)


ToToL:
up

inliniac:
Sorry for the delays. So originally this was meant to go into 4.0, but it didn't make it. I thought I could merge it into 4.0.x as we consider prelude 'community supported' (https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Support_Status). This would make it OK to do bigger updates on the stable branch.

However, we realized that prelude is enabled by default in a few places (most importantly the debian packaging), so updates on the stable branch will have the same restrictions as other parts of the engine.

Therefore, I will consider this branch when 4.1.x-dev opens.

Thanks for your patience.

ToToL:
Too bad we can't make it for 4.0, the first version of this PR was in 2016.

OK, wait for 4.1, I hope this version will be released soon :)

For this :
> This would make it OK to do bigger updates on the stable branch.

Do you have some updates in mind that I would have to do ?

Thanks
Regards

victorjulien:
If you're interested in revisiting this, now would be a good time. 4.1dev is open. It needs a rebase, and probably some testing again. Thanks!

PR #2873, opened 2017-08-07
tls: reimplement TLS certificate fingerprint keyword

thus:
Add `tls_cert_fingerprint` (mpm) keyword, meant to replace `tls.fingerprint` keyword.

Example:
```
alert tls any any -> any any (msg:"TLS fingerprint test"; tls_cert_fingerprint;
       content:"58:95:2f:ca:1a:7d:81:9d:9b:08:3f:1c:d7:94:34:f5:7a:26:f5:54"; sid:1;)
```

Rule profiling below, where the new keyword (tls_cert_fingerprint) is sid 99999 and the old keyword (tls.fingerprint) is sid 11111, shows that there is obviously a performance gain.

```
   Num      Rule         Gid      Rev      Ticks        %      Checks   Matches  Max Ticks   Avg Ticks   Avg Match   Avg No Match
  -------- ------------ -------- -------- ------------ ------ -------- -------- ----------- ----------- ----------- -------------- 
  1        11111        1        1        2431715994   98.93  275594   222      1279778     8823.54     21667.77    8813.19    
  2        99999        1        1        26243937     1.07   9388     222      195288      2795.48     18704.57    2410.16    

```

https://redmine.openinfosecfoundation.org/issues/2062

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/129
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/129


src/detect-engine-tls.c
```diff
@@ -348,6 +348,82 @@ int DetectEngineInspectTlsSerial(ThreadVars *tv, DetectEngineCtx *de_ctx,
...
+ */
+static void PrefilterTxTlsFingerprint(DetectEngineThreadCtx *det_ctx,
+                                      const void *pectx, Packet *p, Flow *f,
+                                      void *txv, const uint64_t idx,
+                                      const uint8_t flags)
+{
+    SCEnter();
+
+    const MpmCtx *mpm_ctx = (MpmCtx *)pectx;
+    SSLState *ssl_state = f->alstate;
```
inliniac: can this be const as well?

src/detect-tls-cert-fingerprint.c
```diff
@@ -0,0 +1,469 @@
...
+                            certificate, sizeof(certificate));
+    FLOWLOCK_UNLOCK(&f);
+
+    FAIL_IF(r != 0);
+
+    SigMatchSignatures(&tv, de_ctx, det_ctx, p3);
+
+    FAIL_IF_NOT(PacketAlertCheck(p3, 1));
+
+    if (alp_tctx != NULL)
```
inliniac: these conditional checks should be unnecessary. We shouldn't get here if any is NULL.

thus:
Fixed comments in #2949

PR #2874, opened 2017-08-07
Add user-configurable umask setting

thus:
Make umask user-configurable by setting 'umask' in suricata.yaml.

This works both when daemonizing and when running in the foreground.

https://redmine.openinfosecfoundation.org/issues/1678

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/130
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/130

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -154,6 +154,7 @@
 #include "util-debug.h"
 #include "util-error.h"
 #include "util-daemon.h"
+#include "util-byte.h"
 #include "reputation.h"
 
 #include "output.h"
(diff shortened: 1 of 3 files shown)


inliniac:
In master we set it to 027 currently. In this PR it's only set if the new config option is present. Would it make sense to fall back to the current behavior if the config setting is missing?

thus:
I think it makes most sense that the umask is inherited from the shell that Suricata is started from, instead of setting it to 027 if it's not configured. That's my opinion, but I will of course change it if you disagree.

jasonish:
I think it makes sense to inherit as well, but it might also be desirable to not change default behaviour. 

When running interactively, it makes sense to inherit, and when running as a service, systemd can set the umask.

inliniac:
So @jasonish you're in favor of taking this in for 4.1?

jasonish:
Yes, makes sense to me to do it this way.

thus:
Replaced by #3294

PR #2875, opened 2017-08-08
tls: add more subject/issuer attributes (Bug #1782)

thus:
Moved the subject/issuer attributes to a lookup table sorted by occurrence, and added all the x509 attributes from openssl and a few other commonly used attributes.

Also, OID is now printed instead of "unknown", if it's not in the lookup table.

https://redmine.openinfosecfoundation.org/issues/1782

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/131
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/131


inliniac:
Would it make sense to turn this into a hash table instead? Looks like this code does linear lookups.

thus:
That would make sense. I thought of doing that, but due to laziness, I just ended up extending the existing implementation. I'll change it to use a hash table instead.

PR #2876, opened 2017-08-09
app-layer-ssl: fix bug with >255 records in one stream

thus:
Fix bug that occurs when one stream contains more than 255 TLS records.

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/132
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/132


inliniac:
Merged in #2882, thanks Mats!

PR #2881, opened 2017-08-23
eve/alert: include rule text in alert output

rs-natano:

https://redmine.openinfosecfoundation.org/issues/2020

Describe changes:

For SIEM analysis it is often useful to refer to the actual rules to find out why a specific alert has been triggered when the signature message does not convey enough information.

Turn on the new rule flag to include the rule text in eve alert output. The feature is turned off by default.

With a rule like this:

    alert dns $HOME_NET any -> 8.8.8.8 any (msg:"Google DNS server contacted"; sid:42;)

The eve alert output might look something like this (pretty-printed for readability):

    {
      "timestamp": "2017-08-14T12:35:05.830812+0200",
      "flow_id": 1919856770919772,
      "in_iface": "eth0",
      "event_type": "alert",
      "src_ip": "10.20.30.40",
      "src_port": 50968,
      "dest_ip": "8.8.8.8",
      "dest_port": 53,
      "proto": "UDP",
      "alert": {
        "action": "allowed",
        "gid": 1,
        "signature_id": 42,
        "rev": 0,
        "signature": "Google DNS server contacted",
        "category": "",
        "severity": 3,
        "rule": "alert dns $HOME_NET any -> 8.8.8.8 any (msg:\"Google DNS server contacted\"; sid:43;)"
      },
      "app_proto": "dns",
      "flow": {
        "pkts_toserver": 1,
        "pkts_toclient": 0,
        "bytes_toserver": 81,
        "bytes_toclient": 0,
        "start": "2017-08-14T12:35:05.830812+0200"
      }
    }

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -1071,7 +1071,10 @@ int SigParse(DetectEngineCtx *de_ctx, Signature *s, const char *sigstr, uint8_t 
     SignatureParser parser;
     memset(&parser, 0x00, sizeof(parser));
 
-    s->sig_str = sigstr;
+    s->sig_str = SCStrdup(sigstr);
+    if (unlikely(s->sig_str == NULL)) {
+        SCReturnInt(-1);
+    }
 
     int ret = SigParseBasics(de_ctx, s, sigstr, &parser, addrs_direction);
     if (ret < 0) {
@@ -1099,8 +1102,6 @@ int SigParse(DetectEngineCtx *de_ctx, Signature *s, const char *sigstr, uint8_t 
 
         } while (ret == 1);
     }
-
-    s->sig_str = NULL;
 
     DetectIPProtoRemoveAllSMs(s);
 
@@ -1252,6 +1253,9 @@ void SigFree(Signature *s)
     }
     if (s->addr_dst_match6 != NULL) {
         SCFree(s->addr_dst_match6);
+    }
+    if (s->sig_str != NULL) {
+        SCFree(s->sig_str);
     }
 
     SigRefFree(s);
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -90,6 +90,7 @@
 #define LOG_JSON_FLOW              BIT_U16(11)
 #define LOG_JSON_HTTP_BODY         BIT_U16(12)
 #define LOG_JSON_HTTP_BODY_BASE64  BIT_U16(13)
+#define LOG_JSON_RULE              BIT_U16(14)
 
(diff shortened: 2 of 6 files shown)


src/detect-parse.c
```diff
@@ -1071,7 +1071,10 @@ int SigParse(DetectEngineCtx *de_ctx, Signature *s, const char *sigstr, uint8_t
     SignatureParser parser;
     memset(&parser, 0x00, sizeof(parser));
 
-    s->sig_str = sigstr;
+    s->sig_str = SCStrdup(sigstr);
```
jasonish: This will leak if SigParseBasics fails to parse the rule.

src/detect-parse.c
```diff
@@ -1100,8 +1103,6 @@ int SigParse(DetectEngineCtx *de_ctx, Signature *s, const char *sigstr, uint8_t
         } while (ret == 1);
     }
 
-    s->sig_str = NULL;
-
```
jasonish: So maybe dup the string based on the value of r.

src/detect-parse.c
rs-natano: No. The only caller of SigParse() calls SigFree() in case the function returned an error. See https://github.com/rs-natano/suricata/blob/9af2e9a50cfdb9d75d206614a8e3b03cb8486590/src/detect-parse.c#L1644 . SigFree() also frees sig_str when not NULL. Note that all other dynamically allocated members of the signature struct are handled in the same way (class_msg, sp, dp, ...); otherwise the cleanup code would be much more complex.

src/detect-parse.c
rs-natano: See previous comment.

src/detect-parse.c
rs-natano: Makes sense?

jasonish: Yes, it does. Took too much of a micro-look at the patch.

src/output-json-alert.c
```diff
@@ -523,6 +524,13 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
             AlertJsonPacket(p, js);
         }
 
+        /* signature text */
+        if (json_output_ctx->flags & LOG_JSON_RULE) {
+            hjs = json_object_get(js, "alert");
+            BUG_ON(!json_is_object(hjs));
```
inliniac: this will need a proper check

rs-natano: What's wrong about the check? When the "alert" key is not there it's a bug in the code before, isn't it?

rs-natano: Ah I see now, the "alert" key might not be there in out of memory conditions. I will create an updated PR.

inliniac: yeah we generally try to allow for alloc issues. Thanks!

rs-natano: Updated PR: https://github.com/inliniac/suricata/pull/2897

PR #2883, opened 2017-08-30
Add --queue-bypass to setting up iptable rules

jeffdyke:
I found this critically important b/c if nothing is listening to the queue (you have started suricata yet, or stopped it), packets are simply stuck and the server becomes unusable on existing and new connections, and requires a reboot to flush iptables.  If there is a better way of handling this, please replace with that.  
Thanks.

Describe changes:
- Documentation Only: Add flag to iptables so the server does not become unresponsive if nothing is listening to NFQUEUE

### DIFF
--- a/doc/userguide/setting-up-ipsinline-for-linux.rst
+++ b/doc/userguide/setting-up-ipsinline-for-linux.rst
@@ -59,10 +59,12 @@ In case of the host situation, these are the two most simple iptable rules;
 ::
 
 
-  sudo iptables -I INPUT -j NFQUEUE
-  sudo iptables -I OUTPUT -j NFQUEUE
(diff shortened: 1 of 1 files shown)


inliniac:
I think the --queue-bypass option shouldn't be set in the default 'getting started guide'. However I think we have a clear need for a more 'advanced iptables/nfqueue guide'. Any interest in getting that going?

Queue bypass could be part of that, traffic marking, etc.

jeffdyke:
I think that is fair, as there is also a question if putting `NFQUEUE` in `MANGLE` would be be a better solution.  I'm not an iptables expert, perhaps i can add a simpler noteworthy PR to link off this page b/c i feel that if you're going to add these rules you need to understand your options and their implications. Thanks for the consideration.  I'll try some different configurations and see what i can come up with.

PR #2884, opened 2017-08-31
Dns compact logging v1

glongo:

https://redmine.openinfosecfoundation.org/issues/2199

Describe changes:
This patchset implements a new DNS output format to be able to have DNS records
in one event rather than an event for each record.

It also introduces the possibility to choose
a version for eve logging. Using the version 2 is needed
to have dns summarization.

There are two different formats of dns answer: 'detailed' or 'aggs'.
* `detailed` show "rrname", "rrtype", "rdata" and "ttl" for each record.
* `aggs` show aggregate records by their type.

Both output formats can be written in one event via the 'both' format.
In the end, it's possible also to add dns information when an alert is logged.

Here there are some examples about the formats described above:
https://gist.github.com/glongo/e04167b15f0785513d234c80e758996a

The main reason of the new format is to summarize dns answers such that only one is stored into a database because in some cases there can be a lot of answers for a single dns query.

The patchset is currently incomplete as logging for the rust DNS parser is not implemented,
but me and @regit did decide to do a preview of the PR so everybody agree on the format.

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/143
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/7

### DIFF
--- a/doc/userguide/configuration/suricata-yaml.rst
+++ b/doc/userguide/configuration/suricata-yaml.rst
@@ -342,6 +342,9 @@ integration with 3rd party tools like logstash.
             # default yes, no to disable
             query: yes     # enable logging of DNS queries
             answer: yes    # enable logging of DNS answers
+            # log answer event in one line
+            # set to 'detailed', 'aggs' or 'both'
+            #answer-type: both
             # control which RR types are logged
             # all enabled if custom not specified
             #custom: [a, aaaa, cname, mx, ns, ptr, txt]
--- a/doc/userguide/configuration/suricata-yaml.rst
+++ b/doc/userguide/configuration/suricata-yaml.rst
@@ -278,6 +278,7 @@ integration with 3rd party tools like logstash.
       enabled: yes
       filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
       filename: eve.json
+      version: 2 #1|2, 2 will enable new logging format (actually used only for dns logging)
       #prefix: "@cee: " # prefix to prepend to each log entry
       # the following are valid when type: syslog above
       #identity: "suricata"
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -45,6 +45,7 @@
 #include "detect-reference.h"
 #include "app-layer-parser.h"
 #include "app-layer-dnp3.h"
+#include "app-layer-dns-common.h"
 #include "app-layer-htp.h"
 #include "app-layer-htp-xff.h"
 #include "util-classification-config.h"
@@ -55,6 +56,7 @@
 #include "output-json.h"
 #include "output-json-alert.h"
 #include "output-json-dnp3.h"
+#include "output-json-dns.h"
 #include "output-json-http.h"
 #include "output-json-tls.h"
 #include "output-json-ssh.h"
@@ -90,8 +92,9 @@
 #define LOG_JSON_FLOW              BIT_U16(11)
 #define LOG_JSON_HTTP_BODY         BIT_U16(12)
 #define LOG_JSON_HTTP_BODY_BASE64  BIT_U16(13)
+#define LOG_JSON_DNS               BIT_U16(14)
 
-#define LOG_JSON_METADATA_ALL  (LOG_JSON_APP_LAYER|LOG_JSON_HTTP|LOG_JSON_TLS|LOG_JSON_SSH|LOG_JSON_SMTP|LOG_JSON_DNP3|LOG_JSON_VARS|LOG_JSON_FLOW)
+#define LOG_JSON_METADATA_ALL  (LOG_JSON_APP_LAYER|LOG_JSON_HTTP|LOG_JSON_TLS|LOG_JSON_SSH|LOG_JSON_SMTP|LOG_JSON_DNP3|LOG_JSON_VARS|LOG_JSON_FLOW|LOG_JSON_DNS)
 
 #define JSON_STREAM_BUFFER_SIZE 4096
 
@@ -100,6 +103,7 @@ typedef struct AlertJsonOutputCtx_ {
     uint16_t flags;
     uint32_t payload_buffer_size;
     HttpXFFCfg *xff_cfg;
+    EveVersion version;
 } AlertJsonOutputCtx;
 
 typedef struct JsonAlertLogThread_ {
@@ -180,6 +184,35 @@ static void AlertJsonDnp3(const Flow *f, json_t *js)
         }
     }
 
+    return;
+}
+
+static void AlertJsonDns(const Flow *f, json_t *js)
+{
+#ifndef HAVE_RUST
+    DNSState *dns_state = (DNSState *)FlowGetAppState(f);
+    if (dns_state) {
+        uint64_t tx_id = AppLayerParserGetTransactionLogId(f->alparser);
+        DNSTransaction *tx = AppLayerParserGetTx(f->proto, ALPROTO_DNS,
+                                                 dns_state, tx_id);
+        if (tx) {
+            json_t *dnsjs = json_object();
+            if (unlikely(dnsjs == NULL)) {
+                return;
+            }
+
+            json_t *qjs = JsonDNSLogQuery(tx, tx_id);
+            if (qjs != NULL) {
+                json_object_set_new(dnsjs, "query", qjs);
+            }
+            json_t *ajs = JsonDNSLogAnswer(tx, tx_id);
+            if (ajs != NULL) {
+                json_object_set_new(dnsjs, "answer", ajs);
+            }
+            json_object_set_new(js, "dns", dnsjs);
+        }
+    }
+#endif
     return;
 }
 
@@ -465,8 +498,16 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
                 json_object_set_new(js, "app_proto",
                         json_string(AppProtoToString(p->flow->alproto)));
             }
-        }
 
+            if (json_output_ctx->version == EVE_VERSION_2) {
+                if (json_output_ctx->flags & LOG_JSON_DNS) {
+                    uint16_t proto = FlowGetAppProtocol(p->flow);
+                    if (proto == ALPROTO_DNS) {
+                        AlertJsonDns(p->flow, js);
+                    } 
+                }
+            }
+        }
 
         /* payload */
         if (json_output_ctx->flags & (LOG_JSON_PAYLOAD | LOG_JSON_PAYLOAD_BASE64)) {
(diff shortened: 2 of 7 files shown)


inliniac:
@jasonish any thoughts on the proposed 'v2' format?

jasonish:
I was told the why this new format is needed, but I can't remember it. Can the PR description be updated with a why this output is needed (use case)?

glongo:
The main reason of the new format is to summarize dns answers such that only one is stored into a database.
In some cases there can be a lot of answers for a single dns query.
For instance, `host k9s.ru` command shown 240 asnwers for one query.

jasonish:
Sorry, I should have been more explicit, its the "aggs" output I'm more interested in the use case for.

I think we also have to think about how the new DNS parser in Rust is stateless. It doesn't match up requests with responses as that is a known performance issues on heavy DNS traffic. The idea there was to compact the responses, but still have one log entry per request, and one per response, unlike now, where a response can generate many log entries. This was only not done in the Rust parse to keep compatibility for now.

One way around this is that responses contain all the information in the request, which is good for simply logging metadata, but I also like the idea of specifically logging requests discretely from responses as more of an audit/debug trail.

But I could see many cases where simply logging the response is enough. And would keep it compatible with the new parsers stateless model as well.

I think it needs a little more though, in terms of the full set of variations that can be logged. I'll try to further outline my thoughts on that in the next few days.

jasonish:
Sorry, for some reason I thought this merged the request and response logging into a single log entry. But it looks like you are keeping the request and response discrete, but compacting the multiple responses into one log entry where they are currently multiple. Still going to define my thoughts tho. If we're changing the format we need to get it right :)

doc/userguide/configuration/suricata-yaml.rst
```diff
@@ -278,6 +278,7 @@ integration with 3rd party tools like logstash.
       enabled: yes
       filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
       filename: eve.json
+      version: 2 #1|2, 2 will enable new logging format (actually used only for dns logging)
```
jasonish: I've added some comments here, https://redmine.openinfosecfoundation.org/issues/2167 with respect to versioning. My worry about doing it here, is we'd have to up it to 3 when we break TLS, up it to 4 if we break http, and so on. So I think we need 2 levels of versioning. In this case "eve" would stay at version 1, and "dns" would move to version 2.

jasonish:
I'm really keen to see this get in, but it brings up 2 problems that are not specifically related to DNS, but more general eve issues.

First, I don't think this changes the version of eve. This changes the version of the "dns" object in "eve". Can we set do a version: 2 in the "dns" object? We could add version: 1 as well, but that should be assumed if no version field exists.

Second, lets take the chance to fix up "dns" section in the suricata.yaml. I would like to see something like the following:
```
        - dns:
            # Use version 2 logging. Without setting a version the
            # version will fallback to 1 for backwards compatibility.
            version: 2

            # Enable/disable this logger. Default: enabled.
            #enabled: no

            # By default both requests and responses are logged.
            #requests: no
            #responses: no

            # Format of answer logging:
            # - detailed: array item per answer
            # - aggs: answers aggregated by type
            # Default: both.
            #answer-format: [detailed, aggs]

            # Answer types to log.
            # Default: all
            #answer-types: [a, aaaa, cname, mx, ns, ptr, txt]

        - dns:
            # Version 1 (deprecated) DNS logger.
            version: 1

            enabled: no

            # control logging of queries and answers
            # default yes, no to disable
            query: yes     # enable logging of DNS queries
            answer: yes    # enable logging of DNS answers
            # control which RR types are logged
            # all enabled if custom not specified
            #custom: [a, aaaa, cname, mx, ns, ptr, txt]
```

This would make fresh installs use the new DNS logging. But very easily changed to the old style. Existing configurations would keep the old logging, yet still fairly easy to upgrade.

Its logs verbosely by default, that is, details and aggs by default.

Refers to aggs and detailed as format, not types to avoid ambiguity with answer types.

Renames custom to answer-types, as custom is too ambiguous.

Should allow for version 1 and version 2 DNS logging to be done at the same time. Both will be logged in the "dns" object, so the version field will need to be referenced. As its opt-in for people upgrading, lets not worry about breaking parsers.

I'm happy to these changes if a more concrete example is desired.

jasonish:
I wonder if we should use "extended" instead of "detailed". That is what is used in http and tls for example.

glongo:
I'm fine to follow this way.
I would also say that this implementation uses `eve.version` to add dns metadata in an alert, so moving` version` in `dns` section means that dns metadata will be added regardless of version used.

If there are no objections/thoughts, I close this PR and proceed with a new one.

victorjulien:
Would it make sense to have naming like 'compact' and 'verbose' or something? Esp  the 'aggs' doesn't really work for me. Perhaps writing it as 'aggregates' would be better already.

victorjulien:
In a new PR please add examples for all log record types in the PR message itself, thanks.

jasonish:
> Would it make sense to have naming like 'compact' and 'verbose' or something? Esp the 'aggs' doesn't really work for me. Perhaps writing it as 'aggregates' would be better already.

The problem with this naming is there is no compact variant, yet anyways.  We have the per answer, which is "detailed" here, or the aggs which has some post-processing done to it and doesn't fit anywhere in the compact->verbose spectrum - its just an alternate data form.

"aggs" doesn't work that well for me either but had grown indifferent to it.  Perhaps "grouped" is a better name?

This leaves us with default (detailed) and grouped and they are not mutually exclusive.  I'd see something like compact and verbose being mutually exclusive of each other. I suppose they don't need to be, but we don't know what compact would look like, or if its even needed.

glongo:
"extended" and "aggregated" ?

victorjulien:
On 28-11-17 13:40, Jason Ish wrote:
>     Would it make sense to have naming like 'compact' and 'verbose' or
>     something? Esp the 'aggs' doesn't really work for me. Perhaps
>     writing it as 'aggregates' would be better already.
> 
> The problem with this naming is there is no compact variant, yet
> anyways. We have the per answer, which is "detailed" here, or the aggs
> which has some post-processing done to it and doesn't fit anywhere in
> the compact->verbose spectrum - its just an alternate data form.
> 
> "aggs" doesn't work that well for me either but had grown indifferent to
> it. Perhaps "grouped" is a better name?

Yeah I think 'grouped' works better for me.

> This leaves us with default (detailed) and grouped and they are not
> mutually exclusive. I'd see something like compact and verbose being
> mutually exclusive of each other. I suppose they don't need to be, but
> we don't know what compact would look like, or if its even needed.

For new setups 'grouped' should probably be the default? Current format
is way too verbose.

jasonish:
> For new setups 'grouped' should probably be the default? Current format is way too verbose.

Not sure I agree. Grouped is lossy, it looks like an ElasticSearch'ism to me :)

The new detailed is still more compact than current master. It results in only one response message for the response, but contains the answers as a list. Instead of an event per answer like we have in master now.

I tried to see if "group" or "aggregate" is more the technical term for this, but neither wins.  MongoDB, ES, use aggregate, SQL and other databases use group.

victorjulien:
Ah ok so I am a bit confused with the various record types now. So can someone post an example of all the record types for the same DNS transaction here? So current, plus the various new ones proposed. Thanks.

jasonish:
Leaving aggs aside, the most value in the PR is logging all answers in a single event..  Sorry, not running this branch yet so don't have real examples. Pulling from @glongo's examples above.  So with this PR we have a DNS response logged like:
```
{
  "timestamp": "2017-08-29T15:27:11.621917+0200",
  "flow_id": 1531685928692682,
  "in_iface": "eth0",
  "event_type": "dns",
  "src_ip": "192.168.1.1",
  "src_port": 53,
  "dest_ip": "192.168.1.21",
  "dest_port": 60387,
  "proto": "UDP",
  "dns": {
    "type": "answer",
    "id": 43691,
    "rcode": "NOERROR",
    "answers": [
      {
        "rrname": "www.suricata-ids.org",
        "rrtype": "CNAME",
        "ttl": 3600,
        "rdata": "suricata-ids.org"
      },
      {
        "rrname": "suricata-ids.org",
        "rrtype": "A",
        "ttl": 300,
        "rdata": "192.0.78.24"
      },
      ....
},
```

Notice that answers is a list. In current master each answer gets blown out into its own event.

So in git master, a DNS request for k9s.ru results in 241 events, 1 request and 240 responses. With this PR we'd more correctly get 2 events. 1 request and 1 response. IMO, we *need* this.

The aggs can be added and gives you groupings:
```
    "aggs": {
      "A": [
        "192.0.78.24",
        "192.0.78.25"
      ],
      "CNAME": [
        "suricata-ids.org"
      ]
}
```

This is handy for ES I think, as you can then search for DNS responses where "192.168.78.24" was returned for an A request. But as its lossy, and a post-processing of the actual response, I don't think it should be the default.

glongo:
This is an example of query/response with the current format:
```
{
  "timestamp": "2017-07-31T15:01:20.802000+0200",
  "flow_id": 2095032066653392,
  "pcap_cnt": 155,
  "event_type": "dns",
  "src_ip": "192.168.1.17",
  "src_port": 35586,
  "dest_ip": "8.8.8.8",
  "dest_port": 53,
  "proto": "UDP",
  "dns": {
    "type": "query",
    "id": 9061,
    "rrname": "tools.ietf.org",
    "rrtype": "A",
    "tx_id": 0
  }
}
{
  "timestamp": "2017-07-31T15:01:20.868355+0200",
  "flow_id": 2095032066653392,
  "pcap_cnt": 156,
  "event_type": "dns",
  "src_ip": "8.8.8.8",
  "src_port": 53,
  "dest_ip": "192.168.1.17",
  "dest_port": 35586,
  "proto": "UDP",
  "dns": {
    "type": "answer",
    "id": 9061,
    "rcode": "NOERROR",
    "rrname": "tools.ietf.org",
    "rrtype": "A",
    "ttl": 237,
    "rdata": "64.170.98.42"
  }
}
```

There are 3 answers logged for this query, but I paste just one to be short.

This below is the answer for the same query with `detailed` and `aggs` formats:
```
{
  "timestamp": "2017-07-31T15:01:20.868355+0200",
  "flow_id": 429956997856464,
  "pcap_cnt": 156,
  "event_type": "dns",
  "src_ip": "8.8.8.8",
  "src_port": 53,
  "dest_ip": "192.168.1.17",
  "dest_port": 35586,
  "proto": "UDP",
  "dns": {
    "type": "answer",
    "id": 9061,
    "rcode": "NOERROR",
    "answers": [
      {
        "rrname": "tools.ietf.org",
        "rrtype": "A",
        "ttl": 237,
        "rdata": "64.170.98.42"
      },
      {
        "rrname": "tools.ietf.org",
        "rrtype": "A",
        "ttl": 237,
        "rdata": "4.31.198.61"
      },
      {
        "rrname": "tools.ietf.org",
        "rrtype": "A",
        "ttl": 237,
        "rdata": "4.31.198.62"
      }
    ],
    "aggs": {
      "A": [
        "64.170.98.42",
        "4.31.198.61",
        "4.31.198.62"
      ]
    }
  }
}
```

victorjulien:
Ok I see the distinction between 'detailed' and 'aggs' better now. If it's all the same to you Jason I would like 'aggs' to become 'grouped'. Otherwise the output looks great.

jasonish:
> If it's all the same to you Jason I would like 'aggs' to become 'grouped'.

Works for me!

jasonish:
So in summary, the changes should be:
- Rename aggs to grouped.
- Don't increment the eve version, but instead increment the dns version.
- Implement my suggested configuration? (aggs->grouped) of course.
- I think its fine to log detail and grouped by default? Can change that before release I suppose.

victorjulien:
Thanks @jasonish, agreed.

glongo:
Updated by #3031

PR #2208, opened 2016-08-31
Fixes v0

inliniac:
Fixes for:
- https://redmine.openinfosecfoundation.org/issues/1853
- https://redmine.openinfosecfoundation.org/issues/1880

Prscript:
- PR inliniac-pcap: https://buildbot.openinfosecfoundation.org/builders/inliniac-pcap/builds/477
- PR inliniac: https://buildbot.openinfosecfoundation.org/builders/inliniac/builds/482

PR #2886, opened 2017-09-01
skip over double periods in JSON output tokenizing

satta:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#2208](https://redmine.openinfosecfoundation.org/issues/2208)

Describe changes:
- Interface name shortening introduces double periods (`..`) as spacers, which cause issues during JSON stats serialization as there `.` characters are also used as separators to define nesting of the JSON output. This commit makes sure that `..` are skipped during tokenizing.

- http://dropbox.tetrinetsucht.de/wf/gcc/
- http://dropbox.tetrinetsucht.de/wf/pcaps/

(from local dockerized run)

### DIFF
--- a/src/counters.c
+++ b/src/counters.c
@@ -905,7 +905,7 @@ void StatsSpawnThreads(void)
 uint16_t StatsRegisterCounter(const char *name, struct ThreadVars_ *tv)
 {
     uint16_t id = StatsRegisterQualifiedCounter(name,
-                                                 (tv->thread_group_name != NULL) ? tv->thread_group_name : tv->name,
+                                                 (tv->thread_group_name != NULL) ? tv->thread_group_name : tv->printable_name,
                                                  &tv->perf_public_ctx,
                                                  STATS_TYPE_NORMAL, NULL);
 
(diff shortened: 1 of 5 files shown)


inliniac:
wonder if it makes sense to replace the double dots instead, e.g. by 2 underscores. That way it would still be clear that it's shortened? abc__ed

satta:
Yes, that would work as well of course (and I also think it would be cleaner). Just wanted to come up with  minimal impact change with my PR.

Would you be OK with replacing the spacer with underscores? I would be happy to update this PR.

inliniac:
Was thinking to only replace it in the stats output, but maybe it makes sense to change this globally? Not sure if/how that would impact others...

satta:
IMHO interface names, including the shortened ones, should be consistent across all outputs. So I would suggest to change it globally or keep the dots and only fix the JSON output...

Not sure about the impact either. Any ideas how to find out?

jasonish:
What causes the interface naming shortening? Ideally the JSON output wouldn't use the shortened name, but the full one.

satta:
The shortening  is apparently done intentionally and was introduced with https://github.com/inliniac/suricata/commit/10d1450e4995630834736bcb5ed192a463d87659 with Suricata 3.1. I'm not sure about the reasoning behind it, I guess @inliniac or @maxtors can say more.

I guess it shouldn't be too difficult to keep the an unshortened thread name as well and use that in the JSON  output.

inliniac:
On 04-09-17 10:49, Sascha Steinbiss wrote:
> The shortening is apparently done intentionally and was introduced with
> 10d1450
> <https://github.com/inliniac/suricata/commit/10d1450e4995630834736bcb5ed192a463d87659>
> with Suricata 3.1. I'm not sure about the reasoning behind it, I guess
> @inliniac <https://github.com/inliniac> or @maxtors
> <https://github.com/maxtors> can say more.
> 
> I guess it shouldn't be too difficult to keep the an unshortened thread
> name as well and use that in the JSON output.

It's done because the name is used in thread names. Those can be max 16
chars. We're doing something like W#01-eth0, so we effectively have max
11 for the iface.

satta:
I see. Would you expect the thread names to be verbatim in the JSON output (where no such limitation exists) or would you be fine with having unshortened ones instead?

PR #2901, opened 2017-09-20
New eve log option for wait before writing to a blocked UNIX stream socket 

aoshiken:

Link to [Redmine ticket](https://redmine.openinfosecfoundation.org/issues/2215)

Changes:
When Suricata is configured with pcap offline analysis and, at the same time, for writing its eve log events to a UNIX socket some events could be lost if Suricata writes to the socket queue quicker than the other socket peer.

That behaviour has been observed analyzing pcap files generated by malware that uses DGA (Domains Generated Algorithm), that implies a lot of DNS flows.

This PR adds a new eve-log option named 'unix-retry-wait' for specifying an amount of microseconds to wait before retrying a write to a UNIX stream socket when its write queue is full.

I know it's a workaround but I don't see any other method suitable.

Regards

### DIFF
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -15,6 +15,9 @@ The most common way to use this is through 'EVE', which is a firehose approach w
       enabled: yes
       filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
       filename: eve.json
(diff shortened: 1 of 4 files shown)


inliniac:
Wouldn't it make more sense to simply using blocking in offline/pcap mode?

aoshiken:
Makes sense

PR #2902, opened 2017-09-22
Expectation ftp v0.14

regit:

 [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/550

Describe changes:
- Expectation API
- Add FTP data support
- Add filestore for FTP Data
- Add ftpcommand keyword

Known issue:

In some cases, the end of FTP-data connection is not detected correctly and the application layer is not receiving an EOF event. When file extraction is used and when this happen, this result in the file being in state UNKNOWN although extraction is complete.

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/327
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/111


src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+ *
+ * Some protocols like FTP create dynamic parallel flow (called expectation). In
+ * order to assign a application layer protocol to these expectation, Suricata
+ * needs to parse message of the initial protocol and create and maintain a list
+ * of expected flow.
+ *
+ * Application layers must use the here described API to implement this mechanism.
+ *
+ * When parsing a application layer message describing a parallel flow, the
+ * application layer can call AppLayerExpectationCreate() to declared an
```
inliniac: s/declared/declare/g

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+ * expectation. By doing that the next flow coming with corresponding IP parameters
+ * will be assigned the specified application layer. The resulting Flow will
+ * also have a Flow storage set that can be retrieved at index
```
inliniac: can you explain here what it does/stores?

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+#include "debug.h"
+
+#include "ippair-storage.h"
+#include "flow-storage.h"
+
+#include "app-layer-expectation.h"
+
+#include "util-print.h"
+
+static int expectation_id = -1;
```
inliniac: by convention prefix with 'g_'

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+static int expectation_data_id = -1;
+
+SC_ATOMIC_DECLARE(uint32_t, expectation_count);
+
+#define EXPECTATION_TIMEOUT 30
+
+typedef struct _Expectation {
```
inliniac: Expectation_

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+ * \param alproto the protocol that need to be set on the expected flow
+ * \param data pointer to data that will be attached to the expected flow
+ *
+ * \return -1 if error
+ * \return 0 if success
+ */
+
+int AppLayerExpectationCreate(Flow *f, int direction, Port src, Port dst, AppProto alproto, void *data)
+{
+    Expectation *exp = SCCalloc(1, sizeof(*exp));
```
inliniac: can you move this down so the null check is directly following it?

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+    exp->sp = src;
+    exp->dp = dst;
+    exp->alproto = alproto;
+    exp->ts = f->lastts;
+    exp->data = data;
+    exp->direction = direction;
+
+    if (GetFlowAddresses(f, &ip_src, &ip_dst) == -1)
+        goto error;
+    if (direction & STREAM_TOSERVER) {
```
inliniac: the ippairs should be directionless

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+        lexp = exp->next;
+        if ( (exp->direction & direction) &&
+             ((exp->sp == 0) || (exp->sp == f->sp)) &&
+             ((exp->dp == 0) || (exp->dp == f->dp))) {
+            alproto = exp->alproto;
+            if (FlowSetStorageById(f, expectation_data_id, exp->data) != 0) {
+                SCLogError(SC_ERR_UNKNOWN_VALUE, "Unable to set flow storage");
+            }
+            exp->data = NULL;
+            (void) IPPairDecrUsecnt(ipp);
```
inliniac: where is this increased?
Also, should this move into RemoveExpectationAndGetNext ?

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+    return lexp;
+}
+
+/**
+ * Function doing a lookup in expectation list
+ *
+ * \return an AppProto value if found
+ * \return ALPROTO_UNKNOWN if not found
+ */
+AppProto AppLayerExpectationLookup(Flow *f, int direction)
```
inliniac: when I see 'Lookup' in the func name, I'm not expecting things like FlowSetStorageById to happen. Should this be called differently? E.g. ...Handle()?

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+
+    pexp = NULL;
+    while (exp) {
+        lexp = exp->next;
+        if ( (exp->direction & direction) &&
+             ((exp->sp == 0) || (exp->sp == f->sp)) &&
+             ((exp->dp == 0) || (exp->dp == f->dp))) {
+            alproto = exp->alproto;
+            if (FlowSetStorageById(f, expectation_data_id, exp->data) != 0) {
+                SCLogError(SC_ERR_UNKNOWN_VALUE, "Unable to set flow storage");
```
inliniac: error print in packet path

src/app-layer-ftp.c
```diff
@@ -231,28 +255,110 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
-                   state->current_line_len);
-            state->port_line_len = state->current_line_len;
+                memcpy(state->port_line, state->current_line,
+                        state->current_line_len);
+                state->port_line_len = state->current_line_len;
+                break;
+            case FTP_COMMAND_RETR:
+            case FTP_COMMAND_STOR:
+                {
+                    char *data = SCCalloc(state->current_line_len + sizeof(int64_t) + 1, sizeof(*data));
```
inliniac: this seems a strange way to pass data. Can we use a struct instead?

src/app-layer-ftp.c
```diff
@@ -231,28 +255,110 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
+                    if (data == NULL)
+                        SCReturnInt(-1);
+                    memcpy(data + sizeof(int64_t), state->current_line, state->current_line_len);
+                    *(int64_t *) data = FlowGetId(f);
+                    int ret  = AppLayerExpectationCreate(f, STREAM_TOSERVER, 0,
+                            state->dyn_port, ALPROTO_FTPDATA, data);
+                    if (ret == -1) {
+                        SCLogError(SC_ERR_INVALID_VALUE, "No expectation created.");
```
inliniac: error in packet path

src/app-layer-ftp.c
```diff
@@ -231,28 +255,110 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
+                break;
+            default:
+                break;
         }
     }
 
     return 1;
 }
 
+static int FTPParsePassiveResponse(Flow *f, void *ftp_state, uint8_t *input, uint32_t input_len)
```
inliniac: passing ftp_state as void * seems unnecessary, please update to FtpState ptr

src/app-layer-ftp.c
```diff
@@ -231,28 +255,110 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
+
+    dyn_port = 256 * part1 + part2;
+#endif
+    state->dyn_port = dyn_port;
+
+
+    return 0;
+}
+
+static int FTPParsePassiveResponseV6(Flow *f, void *ftp_state, uint8_t *input, uint32_t input_len)
```
inliniac: void *ftp_state

src/app-layer-ftp.c
```diff
@@ -414,15 +532,217 @@ static int FTPRegisterPatternsForProtocolDetection(void)
...
+/**
+ * \brief This function is called to retrieve a ftp request
+ * \param ftp_state the ftp state structure for the parser
+ * \param input input line of the command
+ * \param input_len length of the request
+ * \param output the resulting output
+ *
+ * \retval 1 when the command is parsed, 0 otherwise
+ */
+static int FTPDataParse(Flow *f, void *ftp_state,
```
inliniac: void *ftp_data

src/app-layer-ftp.c
```diff
@@ -414,15 +532,217 @@ static int FTPRegisterPatternsForProtocolDetection(void)
...
+        ftpdata_state->files = FileContainerAlloc();
+        if (ftpdata_state->files == NULL) {
+            SCLogError(SC_ERR_MEM_ALLOC, "Could not create file container");
+            SCReturnInt(-1);
+        }
+
+        filename = data + sizeof(int64_t) + 5;
+        ftpdata_state->filename = SCStrdup(filename);
+        if (ftpdata_state->filename == NULL) {
+            SCLogInfo("Unable to allocate filename");
```
inliniac: info msg in packet path

src/app-layer-ftp.c
```diff
@@ -414,15 +532,217 @@ static int FTPRegisterPatternsForProtocolDetection(void)
...
+        if (input_len >= 4 && SCMemcmpLowercase("retr", data + sizeof(int64_t), 4) == 0) {
+            ftpdata_state->command = FTP_COMMAND_RETR;
+        }
+        if (input_len >= 4 && SCMemcmpLowercase("stor", data + sizeof(int64_t), 4) == 0) {
+            ftpdata_state->command = FTP_COMMAND_STOR;
+        }
+
+        if (FileOpenFile(ftpdata_state->files, &sbcfg, (uint8_t *) filename, strlen(filename),
+                         input, input_len, flags) == NULL) {
+            SCLogError(SC_ERR_FOPEN, "Can't open file");
```
inliniac: err in pkt path

src/app-layer-ftp.c
```diff
@@ -414,15 +532,217 @@ static int FTPRegisterPatternsForProtocolDetection(void)
...
+        const char *data = (char *)FlowGetStorageById(f, AppLayerExpectationGetDataId());
+        const char *filename;
+        if (data == NULL) {
+            SCReturnInt(-1);
+        }
+
+        ftpdata_state->files = FileContainerAlloc();
+        if (ftpdata_state->files == NULL) {
+            SCLogError(SC_ERR_MEM_ALLOC, "Could not create file container");
+            SCReturnInt(-1);
```
inliniac: should FlowFreeStorageById(f, AppLayerExpectationGetDataId()); be called in this case as well?

src/app-layer-ftp.c
```diff
@@ -414,15 +532,217 @@ static int FTPRegisterPatternsForProtocolDetection(void)
...
+            ret = FileAppendData(ftpdata_state->files, input, input_len);
+            if (ret == -2) {
+                ret = 0;
+                SCLogDebug("FileAppendData() - file no longer being extracted");
+            } else if (ret < 0) {
+                SCLogDebug("FileAppendData() failed: %d", ret);
+                ret = -2;
+            }
+        } else {
+            ret = FileCloseFile(ftpdata_state->files, NULL, 0, flags);
```
inliniac: this 'ret' is never checked/used

src/app-layer-ftp.c
```diff
@@ -414,15 +532,217 @@ static int FTPRegisterPatternsForProtocolDetection(void)
...
+static int FTPDataGetAlstateProgressCompletionStatus(uint8_t direction)
+{
+    return FTPDATA_STATE_FINISHED;
+}
+
+static int FTPDataGetAlstateProgress(void *tx, uint8_t direction)
+{
+    //FtpDataState *ftpdata_state = (FtpDataState *)tx;
+
+    /* FIXME */
```
inliniac: agreed

src/detect-ftpcommand.c
```diff
@@ -0,0 +1,278 @@
...
+static int g_ftpdata_buffer_id = 0;
+
+/**
+ * \brief Registration function for ftpcommand: keyword
+ *
+ * This function is called once in the 'lifetime' of the engine.
+ */
+void DetectFtpcommandRegister(void) {
+    /* keyword name: this is how the keyword is used in a rule */
+    sigmatch_table[DETECT_FTPCOMMAND].name = "ftpcommand";
```
inliniac: I think this is a bit of a misnomer: probably 'ftpdata_command' or similar would make more sense. A generic ftpcommand would be expected to match against all ftp commands.

src/detect-ftpcommand.c
```diff
@@ -0,0 +1,278 @@
...
+    /* setup function is called during signature parsing, when the ftpcommand
+     * keyword is encountered in the rule */
+    sigmatch_table[DETECT_FTPCOMMAND].Setup = DetectFtpcommandSetup;
+    /* free function is called when the detect engine is freed. Normally at
+     * shutdown, but also during rule reloads. */
+    sigmatch_table[DETECT_FTPCOMMAND].Free = DetectFtpcommandFree;
+    /* registers unittests into the system */
+    sigmatch_table[DETECT_FTPCOMMAND].RegisterTests = DetectFtpcommandRegisterTests;
+
+    DetectAppLayerInspectEngineRegister("ftpdata",
```
inliniac: ftpdata_command ? ftpdata is too much like 'file_data' in my mind, and makes me think of data inspection

regit: Well ftp-data is used in protocol definition so I did choose that name:
```
grep data /etc/services 
ftp-data	20/tcp
```
I agree we are not handling the 20 port but that was the idea behind the naming of the application layer.

src/detect-ftpcommand.c
regit: What about `ftpdata_transfer` ?

src/app-layer-expectation.c
regit: yes make sense

src/app-layer-expectation.c
regit: Moved to RemoveExpectationAndGetNext

src/app-layer-ftp.c
regit: It is used as return value.

src/app-layer-ftp.c
regit: Don't really know what to put here as the transaction is over when the TCP session is dead. Any idea ?

src/app-layer-ftp.c
victorjulien: it's overwritten at line 605

src/detect-ftpcommand.c
regit: Humm looks like I misunderstood at first. Good with me with ftpdata_command.

src/app-layer-ftp.c
regit: OK found something.

src/app-layer-expectation.c
regit: Increased when we fetch IPPair via IPPairLookupIPPairFromHash that does increase the ref cnt if I'm not mistaken.

src/app-layer-expectation.c
```diff
@@ -0,0 +1,287 @@
...
+            exp = RemoveExpectationAndGetNext(ipp, pexp, exp, lexp);
+            continue;
+        }
+        pexp = exp;
+        exp = lexp;
+    }
+
+out:
+    if (ipp)
+        IPPairUnlock(ipp);
```
victorjulien: think this should be IPPairRelease. It decrements the use_cnt that was increased during lookup.

src/app-layer-expectation.c
victorjulien: also, is this a memleak? if data is not stored in flow (error) exp->data is still just set to NULL

src/app-layer-expectation.c
victorjulien: in general I think this func needs an explanation of what it does. It *looks* like it transfers and expectation from ippair to flow and then does some house keeping. It should explain that in comment

regit: Adding comment about the implemented process in the doxygen comment.

src/app-layer-expectation.c
regit: Yes potentially but the set storage funciton should only fails in case the storage is not correctly setup so this is a programmation error if it occurs.
Should we keep the SCLogError in this case ?

A fix would be to probide during the creation of the expectation a cleaning function that could be used to free the structure. I found this a bit overkill for something that should only be reached in case of implementation error.

src/app-layer-expectation.c
regit: Good catch.

PR #2905, opened 2017-09-27
Adds PID to file name, writes to a temp file.

gauravjsingh:
Adds the pid to the file name to reduce the likelihood that a file is
overwritten when suricata is unexpectedly killed.  The number in the
waldo file is only written out during a clean shutdown.  In the event
of an improper shutdown, extracted files will be written using the old
number and existing files with the same name will be overwritten.

Writes extracted files and their metadata to a temporary file prefixed
with a '.'.  Renames the files when they are completely done being
written and the meta file is closed.  As-is there is no way to know that
a file on disk is still being written to by suricata.

Describe changes:
-
-
-

### DIFF
--- a/src/log-filestore.c
+++ b/src/log-filestore.c
@@ -22,6 +22,9 @@
  *
  */
 
+#include <sys/types.h>
+#include <unistd.h>
+
 #include "suricata-common.h"
 #include "debug.h"
 #include "detect.h"
@@ -63,6 +66,7 @@
 #define MODULE_NAME "LogFilestoreLog"
 
 static char g_logfile_base_dir[PATH_MAX] = "/tmp";
+static char working_file_prefix[PATH_MAX] = ".";
 
 SC_ATOMIC_DECLARE(uint32_t, filestore_open_file_cnt);  /**< Atomic counter of simultaneously open files */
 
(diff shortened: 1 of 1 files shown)


src/log-filestore.c
```diff
@@ -22,6 +22,9 @@
  *
  */
 
+#include <sys/types.h>
```
inliniac: these should be unnecessary as they are included in suricata-common.h, so please remove them here

src/log-filestore.c
```diff
@@ -336,8 +348,19 @@ static void LogFilestoreLogCloseMetaFile(const File *ff)
         fprintf(fp, "SIZE:              %"PRIu64"\n", FileTrackedSize(ff));
 
         fclose(fp);
+        /* Move working files to their final location now that we are done
+         * writing them.*/
+        if (rename(working_filename, final_filename) != 0) {
+            SCLogInfo("renaming file %s to %s failed", working_filename,
```
inliniac: these should probably be warnings? OTOH we generally don't want to create output at pkt runtime for the risk of flooding. That risk should be quite low here though.

inliniac:
Any thoughts on making this optional?

I wonder if there might be use cases that depend on the current behavior. Can't really think of any, most ppl probably wait with their processing until the file is complete.

gauravjsingh:
Sorry for the delay. I'll open a new PR after addressing the comments.

For the last comment, should the option be to add PID and prefix, or do you think they should be separate options?

PR #2908, opened 2017-09-30
tls: don't set event on small input data

victorjulien:
Test PR

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -1418,7 +1418,7 @@ static int SSLDecode(Flow *f, uint8_t direction, void *alstate, AppLayerParserSt
     }
 
     /* if we have more than one record */
-    uint32_t max_records = input_len / SSL_RECORD_MINIMUM_LENGTH;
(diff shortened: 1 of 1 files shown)


src/app-layer-ssl.c
```diff
@@ -1418,7 +1418,7 @@ static int SSLDecode(Flow *f, uint8_t direction, void *alstate, AppLayerParserSt
     }
 
     /* if we have more than one record */
-    uint32_t max_records = input_len / SSL_RECORD_MINIMUM_LENGTH;
+    uint32_t max_records = MAX((input_len / SSL_RECORD_MINIMUM_LENGTH),1);
```
regit: Where is the check that is updated due to this change ?

src/app-layer-ssl.c
```diff
@@ -1418,7 +1418,7 @@ static int SSLDecode(Flow *f, uint8_t direction, void *alstate, AppLayerParserSt
...
     while (input_len > 0) {
         if (counter > max_records) {
             SCLogDebug("Looks like we have looped quite a bit. Reset state "
```
victorjulien: @regit here

regit: Ok got it.

PR #2920, opened 2017-10-01
dns: log flags field

macpas:
This adds dns header's flags in eve
log.

Signed-off-by: Eric Leblond <eric@regit.org>

Describe changes:
-ADD DNS header's flags in eve log
-
-

### DIFF
--- a/src/app-layer-dns-common.h
+++ b/src/app-layer-dns-common.h
@@ -194,6 +194,7 @@ typedef struct DNSAnswerEntry_ {
 typedef struct DNSTransaction_ {
     uint16_t tx_num;                                /**< internal: id */
     uint16_t tx_id;                                 /**< transaction id */
+    uint16_t flags;                                 /**< dns flags */
     uint32_t logged;                                /**< flags for loggers done logging */
     uint8_t replied;                                /**< bool indicating request is
                                                          replied to. */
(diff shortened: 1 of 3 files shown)


inliniac:
@jasonish if we take this in there will have to be something similar on the Rust side I think

jasonish:
I am all for adding the flags, I'm just not sure if adding them as a string containing the hex value is the best idea. JSON, being a serialization format it would make more sense just to log it as a number. Or perhaps break it out like the TCP flags in the flow event.

I know a number isn't as easily human read, but it is much easier read by machines and scripts.

jasonish:
> We can still also leave this field in, so that it's easy to aggregate on and spot outliers.

I still think its more true to JSON to store it as a number rather than a string tho. But I won't fuss about it.

macpas:
example, like tcp flags in netflow log: "tcp":{"tcp_flags":"1b","syn":true,"fin":true,"psh":true,"ack":true},

instead of 
"dns":{"type":"answer","id":44910,"flags":"0x8180","rcode":"NOERROR",.....}
you would expect this:
"dns":{"type":"answer","id":44910,"flags":"0x8180","qr":true,"rd":true,"ra":true, "rcode":"NOERROR",.....}..
so we would do it for those 5 flags: QR/AA/TC/RD/AA.
Am I right ?
or we keep the basic hex dump, maybe removing also the 0x in order to have the same display done in tcp_flags ?

PR #2921, opened 2017-10-02
template scripts: cleanup plus test - v3

jasonish:
The PR contains the following changes to the template setup scripts:
- Consistently use "-" instead of "_" in script names.
- Allow all scripts to be run from the top level source directory, or ./src.
- Perform the same name validation before doing anything.
- Add a script that runs a script, does a build, runs another script, build, etc.
- Finally, call this script on Travis, but not on the same build doing Cocinelle so it doesn't add any extra time to the Travis run.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/210
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/563


scripts/check-setup.sh
```diff
@@ -0,0 +1,39 @@
...
+make
+
+./scripts/setup-app-layer-logger.sh Echo
+make
+
+./scripts/setup-decoder.sh Udplite
+make
+
+./scripts/setup-simple-detect.sh simpledetect
+make
```
victorjulien: I think for automated QA this should be the only 'make'. Also, I would suggest using 'make distcheck' instead, so that the addition of files to the buildsystem is checked as well

jasonish:
Closing. See https://github.com/OISF/suricata/pull/2923

PR #2922, opened 2017-10-02
Json shortening fix v2

satta:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#2208](https://redmine.openinfosecfoundation.org/issues/2208)

First version of this PR was https://github.com/OISF/suricata/pull/2886.

Describe changes:
- Interface name shortening introduces double periods (`..`) as spacers, which cause issues during JSON stats serialization as there `.` characters are also used as separators to define nesting of the JSON output. This commit makes sure that `..` are skipped during tokenizing.
- To approach this problem more generically, we also make sure that in addition to the internal thread names (which are restricted to a limited number of characters, which prompted the shortening in the first place) we also keep a 'printable' version of the thread names, to be used in stats output.

- http://dropbox.tetrinetsucht.de/wf/gcc/
- http://dropbox.tetrinetsucht.de/wf/pcaps/

(from local dockerized run)

### DIFF
--- a/src/output-json-stats.c
+++ b/src/output-json-stats.c
@@ -68,6 +68,8 @@ static json_t *OutputStats2Json(json_t *js, const char *key)
     const char *dot = index(key, '.');
     if (dot == NULL)
         return NULL;
+    if (*(dot + 1) == '.' && *(dot + 2) != '\0')
+        dot = index(dot + 2, '.');
 
     size_t predot_len = (dot - key) + 1;
     char s[predot_len];
--- a/src/util-runmodes.c
+++ b/src/util-runmodes.c
@@ -192,8 +192,16 @@ int RunModeSetLiveCaptureAutoFp(ConfigIfaceParserFunc ConfigParser,
 
             threads_count = ModThreadsCount(aconf);
             for (thread = 0; thread < threads_count; thread++) {
+                char *printable_threadname = SCMalloc(sizeof(char) * (16+strlen(live_dev)));
+                if (unlikely(printable_threadname == NULL)) {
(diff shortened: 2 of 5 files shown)


src/output-json-stats.c
```diff
@@ -68,6 +68,8 @@ static json_t *OutputStats2Json(json_t *js, const char *key)
     const char *dot = index(key, '.');
     if (dot == NULL)
         return NULL;
+    if (*(dot + 1) == '.' && *(dot + 2) != '\0')
```
victorjulien: is this a safe access? It looks a bit fishy to me. OTOH if we have dot, *(dot+1) would exist (could be \0) and if it is not \0 *(dot+2) should also be valid. Still I'd be happier if this would include strlen(dot) > 2

src/util-runmodes.c
```diff
@@ -192,8 +192,16 @@ int RunModeSetLiveCaptureAutoFp(ConfigIfaceParserFunc ConfigParser,
 
             threads_count = ModThreadsCount(aconf);
             for (thread = 0; thread < threads_count; thread++) {
+                char *printable_threadname = SCMalloc(sizeof(char) * (16+strlen(live_dev)));
```
victorjulien: whats the 16 here?

src/output-json-stats.c
satta: That was my thinking as well (indeed I spent quite some time pondering about these cases) but I wouldn't mind including additional checks.

src/util-runmodes.c
satta: This was supposed to be an upper bound of the non-device-name part of the printable thread name, e.g the `W#01-`of `W#01-abcdefghijklmno`. In my next PR, I will also make this more dynamic.

PR #2929, opened 2017-10-08
Expectation ftp v16

regit:
Update of #2902 addressing all comments. Only https://github.com/OISF/suricata/pull/2902#discussion_r143326079 is still to discuss and may require an update.

- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made

Change since previous version:
- Add transaction handling
- Fix counters in app-layer stats

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/338
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/122


src/app-layer-expectation.c
```diff
@@ -0,0 +1,285 @@
...
+    ipp = IPPairGetIPPairFromHash(&ip_src, &ip_dst);
+    if (ipp == NULL)
+        goto error;
+
+    iexp = IPPairGetStorageById(ipp, g_expectation_id);
+    exp->next = iexp;
+    IPPairSetStorageById(ipp, g_expectation_id, exp);
+
+    SC_ATOMIC_ADD(expectation_count, 1);
+    IPPairUnlock(ipp);
```
victorjulien: IPPairRelease?  IPPairGetIPPairFromHash increments use_cnt, IPPairUnlock doesn't decrement.

src/app-layer-expectation.c
```diff
@@ -0,0 +1,285 @@
...
+        return -1;
+
+    exp->sp = src;
+    exp->dp = dst;
+    exp->alproto = alproto;
+    exp->ts = f->lastts;
+    exp->data = data;
+    exp->direction = direction;
+
+    if (GetFlowAddresses(f, &ip_src, &ip_dst) == -1)
```
victorjulien: can AppLayerExpectationGet be used here? Looks like code is the same

src/app-layer-expectation.c
```diff
@@ -0,0 +1,285 @@
...
+int AppLayerExpectationGetDataId(void)
+{
+    return g_expectation_data_id;
+}
+
+static Expectation * RemoveExpectationAndGetNext(IPPair *ipp,
+                                Expectation *pexp, Expectation *exp,
+                                Expectation *lexp)
+{
+    (void) IPPairDecrUsecnt(ipp);
```
victorjulien: why is this here?

src/app-layer-expectation.c
```diff
@@ -0,0 +1,285 @@
...
+    Expectation *lexp = NULL;
+    Expectation *pexp = NULL;
+
+    int x = SC_ATOMIC_GET(expectation_count);
+    if (x == 0) {
+        return ALPROTO_UNKNOWN;
+    }
+
+    Expectation *exp = AppLayerExpectationGet(f, direction, &ipp);
+    time_t ctime = time(NULL);
```
victorjulien: this is real time, in other places it seems packet time (so pcap time too) is used

src/detect-ftpcommand.c
```diff
@@ -0,0 +1,278 @@
...
+ * \retval 0 no match
+ * \retval 1 match
+ */
+static int DetectFtpcommandMatch(ThreadVars *t, DetectEngineThreadCtx *det_ctx,
+        Flow *f, uint8_t flags,
+        void *state, void *txv,
+        const Signature *s, const SigMatchCtx *m)
+{
+    const DetectFtpcommandData *ftpcommandd = (const DetectFtpcommandData *) m;
+    FtpDataState *ftp_state = (FtpDataState *)state;
```
victorjulien: ftp_state can be a const ptr too

src/detect-ftpcommand.c
```diff
@@ -0,0 +1,278 @@
...
+static int g_ftpdata_buffer_id = 0;
+
+/**
+ * \brief Registration function for ftpcommand: keyword
+ *
+ * This function is called once in the 'lifetime' of the engine.
+ */
+void DetectFtpcommandRegister(void) {
+    /* keyword name: this is how the keyword is used in a rule */
+    sigmatch_table[DETECT_FTPCOMMAND].name = "ftpdata_command";
```
victorjulien: as long as this does not allow matching on ftp commands in general, but only on the ftpdata_command, I would like to see the files, functions and keyword ID also reflect that in their names

src/app-layer-ftp.c
```diff
@@ -194,9 +202,31 @@ static int FTPParseRequestCommand(void *ftp_state, uint8_t *input,
...
+    }
+
+    if (input_len >= 4 && SCMemcmpLowercase("stor", input, 4) == 0) {
+        fstate->command = FTP_COMMAND_STOR;
+    }
+
     return 1;
 }
 
+struct ftp_transfer_cmd {
```
victorjulien: naming style mismatch

src/app-layer-ftp.c
```diff
@@ -228,31 +258,120 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
-            }
-            memcpy(state->port_line, state->current_line,
-                   state->current_line_len);
-            state->port_line_len = state->current_line_len;
+                memcpy(state->port_line, state->current_line,
+                        state->current_line_len);
+                state->port_line_len = state->current_line_len;
+                break;
+            case FTP_COMMAND_RETR:
+                direction = STREAM_TOCLIENT;
```
victorjulien: need explicit fall through
also comment explaining logic is welcome here

src/app-layer-ftp.c
```diff
@@ -228,31 +258,120 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
+    part2 = atoi((char *)ptr + 1);
+    ptr = memrchr(input, ',', (ptr - input) - 1);
+    if (ptr == NULL)
+        return -1;
+    part1 = atoi((char *)ptr + 1);
+
+    dyn_port = 256 * part1 + part2;
+#endif
+    state->dyn_port = dyn_port;
+
```
victorjulien: minor style

src/app-layer-ftp.c
```diff
@@ -228,31 +258,120 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
+    uint16_t dyn_port = rs_ftp_epsv_response(input, input_len);
+    if (dyn_port == 0) {
+        return -1;
+    }
+
+    state->dyn_port = dyn_port;
+#else
+    uint8_t *ptr;
+
+    ptr = memrchr(input, '|', input_len);
```
victorjulien: declare the var here too?

src/app-layer-ftp.c
```diff
@@ -414,15 +545,219 @@ static int FTPRegisterPatternsForProtocolDetection(void)
...
+    flags |= FILE_USE_DETECT;
+    if (ftpdata_state->files == NULL) {
+        struct ftp_transfer_cmd *data = (struct ftp_transfer_cmd *)FlowGetStorageById(f, AppLayerExpectationGetDataId());
+        if (data == NULL) {
+            SCReturnInt(-1);
+        }
+
+        ftpdata_state->files = FileContainerAlloc();
+        if (ftpdata_state->files == NULL) {
+            SCLogError(SC_ERR_MEM_ALLOC, "Could not create file container");
```
victorjulien: error msg in packet path, might be useful to use an event

src/app-layer-expectation.c
regit: I want to have ref cnt to 1 so it will not be suppressed. The expectation search will decref twice.

src/app-layer-expectation.c
regit: good catch.

src/app-layer-expectation.c
victorjulien: I think it's best to keep the lookup logic and storing of expectation logic separate for this. With a lookup use_cnt is incremented, and it should be decremented if you are done searching. The storing of an expectation should bump the use_cnt and when it's removed it should be decremented. To avoid the possibility of int overflows it's probably best to increment only when the first is set and decrement only when the last is removed.

doc/userguide/rules/ftp-keywords.rst
```diff
@@ -0,0 +1,31 @@
+FTP/FTP-DATA Keywords
+=====================
+
+ftpdata_command
+---------------
+
+Filter ftp-data channel based on command used on the FTP command channel.
+Currently supported commands are RETR (get on a file) and STOR (put on a
+file)
```
norg: Might want do add an ending **.**

doc/userguide/rules/ftp-keywords.rst
```diff
@@ -0,0 +1,31 @@
...
+  ftpdata_command:stor
+
+Signature example::
+
+ alert ftp-data any any -> any any (msg:"FTP store password"; filestore; filename:"password"; ftpdata_command:stor; sid:3; rev:1;)
+
+ftpbounce
+---------
+
+Detect FTP bounce attacks
```
norg: Might want do add an ending **.**

doc/userguide/file-extraction/file-extraction.rst
```diff
@@ -4,9 +4,9 @@ File Extraction
 Architecture
 ~~~~~~~~~~~~
 
-The file extraction code works on top of the HTTP and SMTP parsers. The HTTP parser takes care of dechunking and unzipping the request and/or response data if necessary. The HTTP/SMTP parsers runs on top of the stream reassembly engine.
+The file extraction code works on top of the HTTP, SMTP, NFS and NTP parsers. The HTTP parser takes care of dechunking and unzipping the request and/or response data if necessary. The application layer parsers runs on top of the stream reassembly engine.
```
norg: If we want to add more protocols we might want to write **protocol parsers** and have a list of supported protocols below?

src/app-layer-ftp.c
regit: Just removed the log message for now. I did not find any similar event. Should we use a counter instead ?

src/app-layer-expectation.c
regit: It is the removal function. So I need to make ref count go down by 1 to remove the reference hold by the expectation.

src/app-layer-expectation.c
regit: I'm doing that. Code is not decrementing the counter in the only case of the creation function.

src/app-layer-expectation.c
regit: In fact no it was not exactly the same as one was getting an entry (creating it if needed) and the other one was just a lookup.

victorjulien:
Can you make sure there is a ticket attached to the next iteration of this?

regit:
Yep will do, this is https://redmine.openinfosecfoundation.org/issues/550

PR #2932, opened 2017-10-10
Ruleset stats v23

glongo:
A patchset updated that aims at improving information regarding ruleset returned to the user by suricata.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/1585

Previous PR: https://github.com/OISF/suricata/pull/2847

Describe changes:
- Remove SCStrdup for static strings 
- Rename "tenancy" with "tenants" 
- Don't remove `reload-rules` command for backward compatibility

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/146
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/10

PR #2933, opened 2017-10-11
Ruleset stats v24

glongo:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/1585

Previous PR: https://github.com/OISF/suricata/pull/2932

Describe changes:
- Fixes for QA

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/147
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/11

### DIFF
--- a/doc/userguide/unix-socket.rst
+++ b/doc/userguide/unix-socket.rst
@@ -63,6 +63,11 @@ The set of existing commands is the following:
 * capture-mode: display capture system used
 * conf-get: get configuration item (see example below)
 * dump-counters: dump Suricata's performance counters
+* ruleset-reload-rules: reload ruleset and wait for completion
+* ruleset-reload-nonblocking: ask a ruleset reload and return
+* ruleset-last-reload: return time of last reload
+* ruleset-stats: display the number of rules loaded and failed
+* ruleset-failed-rules: display the list of failed rules
 
 You can access to these commands with the provided example script which
 is named ``suricatasc``. A typical session with ``suricatasc`` will looks like:
--- a/src/detect-http-method.c
+++ b/src/detect-http-method.c
@@ -65,7 +65,7 @@ static int DetectHttpMethodSetup(DetectEngineCtx *, Signature *, const char *);
 void DetectHttpMethodRegisterTests(void);
 void DetectHttpMethodFree(void *);
 static void DetectHttpMethodSetupCallback(Signature *s);
-static _Bool DetectHttpMethodValidateCallback(const Signature *s);
+static _Bool DetectHttpMethodValidateCallback(const Signature *s, const char **sigerror);
 
 /**
  * \brief Registration function for keyword: http_method
@@ -144,7 +144,7 @@ static void DetectHttpMethodSetupCallback(Signature *s)
  *  \retval 1 valid
  *  \retval 0 invalid
  */
-static _Bool DetectHttpMethodValidateCallback(const Signature *s)
+static _Bool DetectHttpMethodValidateCallback(const Signature *s, const char **sigerror)
 {
     const SigMatch *sm = s->init_data->smlists[g_http_method_buffer_id];
     for ( ; sm != NULL; sm = sm->next) {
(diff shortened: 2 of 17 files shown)


doc/userguide/unix-socket.rst
```diff
@@ -63,6 +63,11 @@ The set of existing commands is the following:
 * capture-mode: display capture system used
 * conf-get: get configuration item (see example below)
 * dump-counters: dump Suricata's performance counters
+* ruleset-reload-rules: reload ruleset and wait for completion
+* ruleset-reload-nonblocking: ask a ruleset reload and return
```
norg: I would write **trigger** or **request** a ruleset reload. Or copy the part above and write **reload rulesat and proceed without waiting**.

doc/userguide/unix-socket.rst
```diff
@@ -63,6 +63,11 @@ The set of existing commands is the following:
...
+* ruleset-last-reload: return time of last reload
```
norg: Maybe **ruleset-reload-time** to keep it similiar to the two above?

glongo:
Sounds good to me, thanks.

src/detect-http-method.c
```diff
@@ -153,16 +153,32 @@ static _Bool DetectHttpMethodValidateCallback(const Signature *s)
...
+                *sigerror = SCStrdup("http_method pattern with trailing tab");
+                if (*sigerror == NULL) {
+                    SCLogError(SC_ERR_MEM_ALLOC, "Error allocating memory");
+                    return FALSE;
+                }
+                SCLogError(SC_ERR_INVALID_SIGNATURE, "%s", *sigerror);
                 return FALSE;
             } else if (cd->content[0] == 0x09) {
-                SCLogError(SC_ERR_INVALID_SIGNATURE, "http_method pattern with leading tab");
+                *sigerror = SCStrdup("http_method pattern with leading tab");
```
victorjulien: in this callback there is a SCStrdup of the error

src/detect-http-raw-header.c
```diff
@@ -143,13 +143,15 @@ int DetectHttpRawHeaderSetup(DetectEngineCtx *de_ctx, Signature *s, const char *
                                                   ALPROTO_HTTP);
 }
 
-static _Bool DetectHttpRawHeaderValidateCallback(const Signature *s)
+static _Bool DetectHttpRawHeaderValidateCallback(const Signature *s, const char **sigerror)
 {
     if ((s->flags & (SIG_FLAG_TOCLIENT|SIG_FLAG_TOSERVER)) == (SIG_FLAG_TOCLIENT|SIG_FLAG_TOSERVER)) {
-        SCLogError(SC_ERR_INVALID_SIGNATURE,"http_raw_header signature "
+        *sigerror = "http_raw_header signature "
```
victorjulien: but in this callback we pass a static str. Seems one of them is wrong

src/detect-http-method.c
```diff
@@ -153,16 +153,32 @@ static _Bool DetectHttpMethodValidateCallback(const Signature *s)
         const DetectContentData *cd = (const DetectContentData *)sm->ctx;
         if (cd->content && cd->content_len) {
             if (cd->content[cd->content_len-1] == 0x20) {
-                SCLogError(SC_ERR_INVALID_SIGNATURE, "http_method pattern with trailing space");
+                *sigerror = "http_method pattern with trailing space";
```
victorjulien: we even see these 2 patterns within this one callback

src/output-json-stats.c
```diff
@@ -79,6 +185,20 @@ static json_t *OutputStats2Json(json_t *js, const char *key)
     json_t *value = json_object_iter_value(iter);
     if (value == NULL) {
         value = json_object();
+        /* If multi-tenancy is disabled, it means we have only one engine.
+           In this case, we add engine stats in detect obj.
+           The output will be:
+           "detect":{"engine":{"last_reload":"...","rules_loaded":...,"rules_failed":...}, ...}
```
victorjulien: Instead of 2 formats like this:
``` json
"detect":{"engine":{"last_reload":"...","rules_loaded":...,"rules_failed":...}, ...}

"tenancy":[{"id":1,"last_reload":"...","rules_loaded":...,"rules_failed":...},
           {"id":2,"last_reload":"...","rules_loaded":...,"rules_failed":...}]
```

why not a single one like this?
``` json
// single
"detect":[{"id":1,"last_reload":"...","rules_loaded":...,"rules_failed":...}]

// multi
"detect":[{"id":1,"last_reload":"...","rules_loaded":...,"rules_failed":...},
          {"id":2,"last_reload":"...","rules_loaded":...,"rules_failed":...}]
```

src/unix-manager.c
```diff
@@ -652,20 +656,130 @@ static TmEcode UnixManagerCaptureModeCommand(json_t *cmd,
...
+        json_object_set_new(server_msg, "message",
+                            json_string("Live rule reload not possible if -s "
+                                        "or -S option used at runtime."));
+        SCReturnInt(TM_ECODE_FAILED);
+    }
+
+    int r = DetectEngineReloadStart();
+
+    if (r == 0 && do_wait) {
+        while (DetectEngineReloadIsIdle() == 0)
```
victorjulien: did you test loading a large ruleset with config settings that make loading very slow. Can you break out of this loop then? Or is suricata stuck until loading is complete.

src/output-json-stats.c
glongo: there is also `alert` field in `detect`, this is the reason why I thought to use
```
"detect":{"engine":[...], "alert":...}
```

src/unix-manager.c
glongo: I tested it loading 19082 rules, if you use the command `ruleset-reload-rules`, Suricata (unix client) is blocked until loading is completed.
If you use instead the command `ruleset-reload-nonblocking` it doesn't stuck.

victorjulien: My point is that if someone presses 'Ctrl-C' while the blocking load is going it might be good to break out of the loop? Did you test this?

src/output-json-stats.c
victorjulien: I'm fine with making it 'detect.engines'. The main point is to have a single logic for both with and without tenants

src/unix-manager.c
glongo: Actually when 'ctrl-c' is pressed, Suricata is killed after that the reload is completed.
I've added the condition below to try to break the loop, but it doesn't work as expected:
```
while (DetectEngineReloadIsIdle() == 0 && suricata_ctl_flags == 0) {
    usleep(100)
}
```

victorjulien: Why not?

glongo: What I expect with the check above is that when I press 'ctrl-c' the loop should be broken without waiting.
So in unix client I should see "done" just after the signal is sent:
```
>>> ruleset-reload-rules
Success:
"done"
```
but this occurs when all the rules are reloaded.

victorjulien: Have you tried figuring out why this happens?

PR #2946, opened 2017-10-22
NSM: add DNS header flag for dns event record

macpas:

Describe changes:
- add dns flag code in eve log
- add dns flag in hexa code and various data (like tcp flags) (suggested in code review of PR 2920)
- eg: "flags":"8180","qr":true,"rd":true,"ra":true
- done with C  code and rust code as suggested in previous code review (PR 2920)

### DIFF
--- a/src/app-layer-dns-udp.c
+++ b/src/app-layer-dns-udp.c
@@ -316,6 +316,7 @@ static int DNSUDPResponseParse(Flow *f, void *dstate,
             tx->recursion_desired = 1;
         }
 
+        tx->flags = ntohs(dns_header->flags);
         tx->replied = 1;
     }
     if (f != NULL) {
--- a/src/output-json-dns.c
+++ b/src/output-json-dns.c
@@ -467,8 +467,19 @@ static void OutputAnswer(LogDnsLogThread *aft, json_t *djs,
 
     /* dns */
     char flags[7] = "";
-    snprintf(flags, sizeof(flags), "0x%4x", tx->flags);
+    snprintf(flags, sizeof(flags), "%4x", tx->flags);
     json_object_set_new(js, "flags", json_string(flags));
+    if (tx->flags & 0x8000)
+        json_object_set_new(js, "qr", json_true());
+    if (tx->flags & 0x0400)
+        json_object_set_new(js, "aa", json_true());
+    if (tx->flags & 0x0200)
+        json_object_set_new(js, "tc", json_true());
+    if (tx->flags & 0x0100)
+        json_object_set_new(js, "rd", json_true());
+    if (tx->flags & 0x0080)
+        json_object_set_new(js, "ra", json_true());
+
 
     /* rcode */
     char rcode[16] = "";
--- a/src/output-json-dns.c
+++ b/src/output-json-dns.c
@@ -465,6 +465,11 @@ static void OutputAnswer(LogDnsLogThread *aft, json_t *djs,
     /* id */
     json_object_set_new(js, "id", json_integer(tx->tx_id));
 
+    /* dns */
+    char flags[7] = "";
+    snprintf(flags, sizeof(flags), "0x%4x", tx->flags);
+    json_object_set_new(js, "flags", json_string(flags));
+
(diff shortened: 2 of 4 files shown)


victorjulien:
Can you fix the formatting of this pull request? You should be able to edit it without close/resubmit.

victorjulien:
Formatting is still off. The boxes should actually look like proper checkboxes. There are extra spaces in there.

src/app-layer-dns-udp.c
```diff
@@ -316,6 +316,7 @@ static int DNSUDPResponseParse(Flow *f, void *dstate,
             tx->recursion_desired = 1;
         }
 
+        tx->flags = ntohs(dns_header->flags);
```
victorjulien: should this also be set for DNS/TCP?

cc: @jasonish

macpas: Right, I believe so. I am doing the update within app-layer-dns-tcp.c file (line 513, adding this code tx->flags = ntohs(dns_header->flags); ). 
So far I do not have any dns over tcp pcap sample file. I will try to get some in my lab to verify this.

victorjulien: I've emailed you a pcap.

victorjulien:
I think we'll need to add support for Rust as well, right? cc: @jasonish

jasonish:
> I think we'll need to add support for Rust as well, right? cc: @jasonish

Looks like Rust support is included here.

src/output-json-dns.c
```diff
@@ -465,6 +465,22 @@ static void OutputAnswer(LogDnsLogThread *aft, json_t *djs,
     /* id */
     json_object_set_new(js, "id", json_integer(tx->tx_id));
 
+    /* dns */
+    char flags[7] = "";
+    snprintf(flags, sizeof(flags), "%4x", tx->flags);
+    json_object_set_new(js, "flags", json_string(flags));
```
jasonish: If its a string, of a hex number, I wonder if it should include the leading 0x? I'm trying to find some existing JSON precedent. But as it is a serialization format I think it should just be logged as an integer, the viewer can convert it back to hex?

Does our JSON have existing precedent of logging hex as a string without the leading 0x?

victorjulien: We do this in flow/netflow records, for tcp flags set. The only feedback I've ever had on that field was negative though. But perhaps it's a documentation issue.

victorjulien:
Right, didn't look before talking. Disregard the Rust comment.

src/output-json-dns.c
macpas: In previous commit (PR 2920), I was adding the 0x. However, since for TCP netflow, the 0x is not print in the json, I have proposed for this current PR to do the same: for both, no 0x in keep coherency and same philosophy for both (DNS or Netflow TCP flag): no 0x for the hexa output.

jasonish:
> In previous commit (PR 2920), I was adding the 0x. However, since for TCP netflow, the 0x is not print in the json, I have proposed for this current PR to do the same: for both, no 0x in keep coherency and same philosophy for both (DNS or Netflow TCP flag): no 0x for the hexa output.

Ok, then yes, this might be the best way to go for now.

victorjulien:
I think we do need a documentation update as well, explaining what the new fields mean. Esp for the hex flags field this seems important.

src/app-layer-dns-udp.c
macpas: ok received and used to check the update for TCP. It works fine.

src/output-json-dns.c
macpas: Yes agreed, I need to update documentation (eve-json-format.rst) file with this feature: explain new fields, and provide the 0x notice in this section.

macpas:
Yes agreed, I need to update documentation (eve-json-format.rst) file with this feature: explain new fields, and provide the 0x notice in this section.

victorjulien:
Please close a PR if you replace it by a newer version, thanks!

victorjulien:
Replaced by #3028

PR #2951, opened 2017-10-26
Dyn register rust 01

chifflier:

None

Describe changes:
- This PR simplifies the registration to Rust parsers, based on the existing C method (it calls the same functions in the same way). If it works and is satisfying, it could be extended to all parsers.


victorjulien:
How hard would it be to move a bit more back into C? I was thinking maybe both C and Rust parsers can setup the registration struct and then the C function would take it in and run the registration. That way C also can have the benefits of the struct approach.

chifflier:
Moving back the registration should not be too complicated. Function pointers are the same, option types on pointers are optimized by the compiler and thus passed as {NULL or function pointer} to C. The only type requiring adaptation are strings, but there is CString type for that.
I'll do that for the next PR.
That should not change anything but moving the `register_parser` to C code. If you have other comments on the PR, I'll wait for them to merge all of them.

rust/src/parser.rs
```diff
@@ -0,0 +1,308 @@
...
+
+use libc;
+use libc::{c_void,c_char,c_int};
+use std::ffi::CString;
+
+
+// Application layer protocol identifiers (app-layer-protos.h)
+
+pub const ALPROTO_UNKNOWN : AppProto = 0;
+pub const ALPROTO_FAILED : AppProto = 20;
```
victorjulien: this seems like asking for trouble, as adding new protos in the C side will possibly change it. Can we somehow set it like you did in ALPROTO_NTP?

rust/src/parser.rs
```diff
@@ -0,0 +1,308 @@
...
+    pub has_events:      Option<HasEventsFn>,
+    /// Function to get events
+    pub get_events:      Option<GetEventsFn>,
+    /// Function to get an event description
+    pub get_eventinfo:   Option<GetEventInfoFn>,
+
+    // Missing:
+    // - LocalStorageAlloc / LocalStorageFree
+    // - SetTxMpmIDs / GetTxMpmIDs
+    // - GetFiles
```
victorjulien: the files are a bit trick between rust and C right now. Do you see any potential issues?

rust/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub mod json;
 pub mod applayer;
 pub mod filecontainer;
 pub mod filetracker;
+#[macro_use] pub mod parser;
```
victorjulien: minor: I think it's nicer to have the 'pub mod parser;' on a new line

jasonish:
I to would like to see this moved back to C, then Rust calling into the C logic with a struct. It would let us validate the use case from both sides, with only one copy of the logic.

Also, there are some protos, DNP3 and Modbus that do not get enabled by default if their configuration is missing. So I wonder if that part should still be handled by each protocol.

rust/src/lib.rs
chifflier: done

rust/src/parser.rs
chifflier: oh, I missed that one.
updated as well

rust/src/parser.rs
chifflier: Not sure I understand what you mean.
C and Rust files are quite similar and must be kept in sync (and moving the registration to C adds the declaration of the structure in the C code as well). Hopefully, the structures will not change often.
If we need to do the declaration automatically, it's possible to use rust-bindgen.

For the functions not wrapped, it's not a problem: we can add the declaration of the C functions (for ex `AppLayerParserRegisterGetFilesFunc`) and call the after registering the parser. Or we can add a field to the structure.

chifflier:
Ok, I moved the registration code to the C side, so any parser will be able to use it. There is no downside, the structures and functions are ABI-compatible (this is guaranteed by LLVM).
I'm only facing a cosmetic issue: at the moment, the C code does not have `typedef` for the functions. To avoid using `void *` fields, or duplicating the declaration, I think I'll add `typedef` for every function used.

rust/src/parser.rs
victorjulien: Maybe it's not a problem at all, just wanted to mention that it's a bit of a hack now :)

victorjulien:
typedefs sound good. Can you do this in a separate commit?

chifflier:
Hmm, maybe I'll reconsider the typedefs, or at least minimize the changes. That would create a huge change in `app-layer-parser.c`. It will go in a separate commit

chifflier:
Superseded by #2953

PR #2953, opened 2017-10-27
Dyn register rust 02

chifflier:

None

Describe changes:

Changes since #2951


src/app-layer-register.h
```diff
@@ -0,0 +1,73 @@
...
+ * \file
+ *
+ * \author Pierre Chifflier <chifflier@wzdftpd.net>
+ */
+
+#ifndef __APP_LAYER_REGISTER_H__
+#define __APP_LAYER_REGISTER_H__
+
+typedef struct AppLayerParser {
+    const char *Name;
```
victorjulien: in C our style is to have Funcs and FuncPtrs like this, but variables_like_this. This includes struct members.

src/app-layer-parser.h
```diff
@@ -103,6 +103,12 @@ void AppLayerParserThreadCtxFree(AppLayerParserThreadCtx *tctx);
 int AppLayerParserConfParserEnabled(const char *ipproto,
                                     const char *alproto_name);
 
+/** \brief Prototype for parsing functions */
+typedef int (*ParserFPtr)(Flow *f, void *protocol_state,
```
victorjulien: As this is now public I like it to be prefixed with AppLayer.

chifflier: Should I use something like `AppLayerParserFPtr` ?
note: there is `ProbingParserFPtr` in `src/app-layer-detect-proto.h`

chifflier: Changed to `AppLayerParserFPtr`

src/app-layer-register.h
chifflier: Ok, updated

chifflier:
Replaced by #2955

PR #2955, opened 2017-10-28
Dyn register rust 03

chifflier:

None

Describe changes:

Changes since #2951

Changes since #2953
- Rebased on master


src/app-layer-register.c
```diff
@@ -0,0 +1,176 @@
...
+                    p->ProbeTS, p->ProbeTC);
+            }
+
+        }
+    }
+    else {
+        SCLogDebug("Protocol detecter and parser disabled for %s.", p->name);
+        return alproto;
+    }
+
```
jasonish: Would it be reasonable to refactor this into 2 functions. So parsers that don't follow default logic for being enabled could still use the registration bit?

chifflier: Sure, done for the next PR
This puts back some code when the parser registers itself (to check the config), but allows more flexibility: each protocol can choose the default behavior, etc.

chifflier:
Replaced by #2959

PR #2957, opened 2017-10-30
Pcap Directory Mode (https://redmine.openinfosecfoundation.org/issues/2222)

dbcfd:
Pcap file mode that is passed a folder will process all files in that folder until the folder is moved/deleted or suricata is interrupted.

Engine state will not reset between files.

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2222

Describe changes:
- If source-pcap-file is passed a directory instead of a file, all files in that directory will be processed in order of modified time, without state resetting between files.
- source-pcap-file will continue to process files until suricata is interrupted or the directory is moved/deleted
- runmode-unix-socket is updated to provide a current time to determine what state of processing of the directory is

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #2958, opened 2017-10-30
Pcap Folder Mode

dbcfd:
Version 2 of https://github.com/OISF/suricata/pull/2957

Pcap file mode that is passed a folder will process all files in that folder until the folder is moved/deleted or suricata is interrupted.

Engine state will not reset between files.

https://redmine.openinfosecfoundation.org/issues/2222

Describe changes:
- If source-pcap-file is passed a directory instead of a file, all files in that directory will be processed in order of modified time, without state resetting between files.
- source-pcap-file will continue to process files until suricata is interrupted or the directory is moved/deleted
- runmode-unix-socket is updated to provide a current time to determine what state of processing of the directory is

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -470,7 +470,7 @@
         LDFLAGS="$LDFLAGS -lgxpci -lgxio -ltmc"
         ],
         [AC_MSG_RESULT([no])])
-
+  
   #libpcre
     AC_ARG_WITH(libpcre_includes,
             [  --with-libpcre-includes=DIR  libpcre include directory],
--- a/configure.ac
+++ b/configure.ac
@@ -470,7 +470,7 @@
         LDFLAGS="$LDFLAGS -lgxpci -lgxio -ltmc"
         ],
         [AC_MSG_RESULT([no])])
-  
+ 
   #libpcre
     AC_ARG_WITH(libpcre_includes,
             [  --with-libpcre-includes=DIR  libpcre include directory],
--- a/doc/userguide/unix-socket.rst
+++ b/doc/userguide/unix-socket.rst
@@ -152,7 +152,9 @@ You can add multiple files without waiting the result: they will be
 sequentially processed and the generated log/alert files will be put
 into the directory specified as second arguments of the pcap-file
 command. You need to provide absolute path to the files and directory
-as suricata don’t know from where the script has been run.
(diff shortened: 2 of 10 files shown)


configure.ac
```diff
@@ -470,7 +470,7 @@
         LDFLAGS="$LDFLAGS -lgxpci -lgxio -ltmc"
         ],
         [AC_MSG_RESULT([no])])
-
+
```
norg: is this necessary?

doc/userguide/unix-socket.rst
```diff
@@ -152,7 +152,9 @@ You can add multiple files without waiting the result: they will be
 sequentially processed and the generated log/alert files will be put
 into the directory specified as second arguments of the pcap-file
 command. You need to provide absolute path to the files and directory
-as suricata don’t know from where the script has been run.
+as suricata don’t know from where the script has been run. If you pass
+in a directory instead of a file, all files in the directory will be processed
+until you use the interrupt command or delete the directory.
```
norg: We could fix the "don't" into "doesn't" right away. I would also write "If you pass a directory" without the "in". And maybe add what the interrupt command is.

configure.ac
dbcfd: Nope, unintended whitespace from when we were attempting to add inotify support. Missed when that was removed.

dbcfd:
Closing for https://github.com/OISF/suricata/pull/2960

src/runmode-unix-socket.c
```diff
@@ -222,21 +242,21 @@ static TmEcode UnixSocketAddPcapFile(json_t *cmd, json_t* answer, void *data)
 #else
     if(stat(filename, &st) != 0) {
 #endif /* OS_WIN32 */
-        json_object_set_new(answer, "message", json_string("File does not exist"));
+        json_object_set_new(answer, "message", json_string("filename does not exist"));
```
Gladskih: What does it mean? I've come here googling meaning of the error message.

dbcfd: Can you describe what the correct message would be?

Gladskih: File does not exist would be right in my case. I think there is a confusion in execution brunches due to conditional compilation but not shure where exactly (my main experience is in C# so I have eyebleading reading C).

PR #2959, opened 2017-10-31
Dyn register rust 04

chifflier:

None

Describe changes:

Changes since #2951

Changes since #2953
- Rebased on master

Changes since #2955 
- Registration function split in two (to register protocol detection, and protocol parsing functions), allow more flexibility in default parameters. The parser registration does not check for configuration (like protocol being enabled), this is left to the caller.


src/app-layer-register.c
```diff
@@ -0,0 +1,191 @@
...
+#include "app-layer-register.h"
+
+static const char * IpProtoToString(int ip_proto);
+
+AppProto AppLayerRegisterProtocolDetection(const struct AppLayerParser *p, int enable_default)
+{
+    AppProto alproto;
+    const char *ip_proto_str = NULL;
+
+    if (p == NULL)
```
victorjulien: this should probably print an error or simply abort()/BUG_ON as this is a coding error, not a user config error

src/app-layer-register.c
```diff
@@ -0,0 +1,191 @@
...
+        return ALPROTO_FAILED;
+
+    alproto = StringToAppProto(p->name);
+    if (alproto == ALPROTO_UNKNOWN || alproto == ALPROTO_FAILED)
+        return alproto;
```
victorjulien: same here

src/app-layer-register.c
```diff
@@ -0,0 +1,191 @@
...
+
+    ip_proto_str = IpProtoToString(p->ip_proto);
+    if (ip_proto_str == NULL) {
+            SCLogDebug("Unknown or unsupported ip_proto field in parser");
```
victorjulien: indent look off

victorjulien: also this is probably another coding error if this triggers

src/app-layer-register.c
```diff
@@ -0,0 +1,191 @@
...
+                        p->name, ip_proto_str);
+            }
+        }
+
+    }
+
+    return alproto;
+}
+
+AppProto AppLayerRegisterParser(const struct AppLayerParser *p)
```
victorjulien: first 3 checks look like coding errors, so probably bail hard

victorjulien:
Minor comment/question: is there any difference between:
<pre>
CString::new("ntp").unwrap();
"failed\0".as_ptr()
</pre>
I see both patterns in the PR and I wonder if we should settle on one. Stack vs heap alloc perhaps?

chifflier:
I'd say the main difference is that a CString is always NULL-terminated, while you need to add the '\0' when using the `str` type. This is error-prone, especially for the C side, if you forget the '\0'. The drawback of using a CString is that you need to keep the object alive to use the pointer, and so cannot use `CString::new("ntp").unwrap().as_ptr()` as a oneliner (`CString would be dropped before the pointer).

That said, the produced code is clearly different: a `CString` requires the allocation of a vector, checking the result, etc. Using a `str` is clearly better (it's just a constant), if adding the '\0' is not a problem for you.

src/app-layer-register.c
chifflier: I copied the behavior from `app-layer-parser.c`: these errors are now fatal (they can only result from a programming mistake and cannot be recovered), they print an error and exit using
`SCLogError(SC_ERR_FATAL, ...) + exit(EXIT_FAILURE)`.

Since this function can now only return a valid `AppProto`, I update the inline documentation and simplified the rust calls.

src/app-layer-register.c
chifflier: Done

src/app-layer-register.c
chifflier: Done

src/app-layer-register.c
victorjulien: You can use FatalError, it takes the same args as SCLogError but exits

chifflier: I see, the macro does both things. Done!

chifflier:
Replaced by #2977

PR #2972, opened 2017-11-02
Configurable Log Event Drop

dbcfd:
Make retries configurable for logs, so that block does not automatically result in a dropped event.

https://redmine.openinfosecfoundation.org/issues/2215

Describe changes:
- Provide new config option retries.blocked, defaulted to 0 to determine number of retries if log write is blocked
- Provide new config option retries.interrupted, defaulted to 1 to determine number of retries if log write is interrupted
- Provide  new config option retries.reconnecting, defaulted to 1 to determine number of retries when reconnecting to log

### DIFF
--- a/doc/userguide/configuration/suricata-yaml.rst
+++ b/doc/userguide/configuration/suricata-yaml.rst
@@ -241,6 +241,10 @@ There are several types of output. The general structure is:
       enabled: yes
       filename: fast.log
       append: yes/no
+      retries:
+          reconnecting: 1    #number of retries when reconnecting
+          interrupted: 1     #number of retries if write is interrupted
+          blocked: 0         #number of retries if write is blocked
 
 Enabling all of the logs, will result in a much lower performance and
 the use of more disc space, so enable only the outputs you need.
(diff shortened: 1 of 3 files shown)


src/util-logopenfile.h
```diff
@@ -128,6 +128,15 @@ typedef struct LogFileCtx_ {
     /* Socket types may need to drop events to keep from blocking
      * Suricata. */
     uint64_t dropped;
+
+    /* Number of retries when reconnected */
+    intmax_t retries_when_reconnecting;
```
victorjulien: these can be 'int's instead

doc/userguide/configuration/suricata-yaml.rst
```diff
@@ -241,6 +241,10 @@ There are several types of output. The general structure is:
       enabled: yes
       filename: fast.log
       append: yes/no
+      retries:
+          reconnecting: 1    #number of retries when reconnecting
```
victorjulien: should this be 2 space indent instead for yaml?

jasonish:
While I don't see anything particular issues with the code, I have to wonder what size of retry account, in particular when blocked, you need to use to actually see the retry succeed.

In the usual use case, the socket is going to block due to a slow resource. Typically a TCP socket waiting for its buffers to drain.  Here we immediately do the retry, on a much faster resource (the local cpu). You'd probably need very large values to actually see the retry succeed as there are no sleeps (and we don't want them in live mode), perhaps waiting a millisecond or more (round trip TCP, say waiting for a socket). 

Do you have test values that were used to see the block retry case succeed?

When I look more at the problem being solved, I wonder if blocking in offline mode (and socket mode, when being fed pcaps) solves the issue for the 99%.

dbcfd:
Closing for https://github.com/OISF/suricata/pull/2975

PR #2974, opened 2017-11-06
pfring: hw bypass support

cardigliano:
This patch adds support for hw bypass by enabling flow offload in the network
card (when supported) and implementing the BypassPacketsFlow callback.
Hw bypass support is disabled by default, and can be enabled by setting
"bypass: yes" in the pfring interface configuration section in suricata.yaml.

### DIFF
--- a/src/runmode-pfring.c
+++ b/src/runmode-pfring.c
@@ -195,6 +195,7 @@ static void *ParsePfringConfig(const char *iface)
     cluster_type default_ctype = CLUSTER_ROUND_ROBIN;
     int getctype = 0;
     const char *bpf_filter = NULL;
+    int bool_val;
 
     if (unlikely(pfconf == NULL)) {
         return NULL;
@@ -351,6 +352,13 @@ static void *ParsePfringConfig(const char *iface)
             pfconf->checksum_mode = CHECKSUM_VALIDATION_RXONLY;
         } else {
             SCLogError(SC_ERR_INVALID_ARGUMENT, "Invalid value for checksum-checks for %s", pfconf->iface);
+        }
+    }
+
+    if (ConfGetChildValueBoolWithDefault(if_root, if_default, "bypass", &bool_val) == 1) {
+        if (bool_val) {
+            SCLogConfig("Using bypass support in PF_RING for iface %s (if supported by underlying hw)", pfconf->iface);
+            pfconf->flags |= PFRING_CONF_FLAGS_BYPASS; 
         }
     }
 
--- a/src/source-pfring.c
+++ b/src/source-pfring.c
@@ -127,9 +127,9 @@ static SCMutex pfring_bpf_set_filter_lock = SCMUTEX_INITIALIZER;
 #define LIBPFRING_REENTRANT   0
 #define LIBPFRING_WAIT_FOR_INCOMING 1
 
-typedef enum {
-    PFRING_FLAGS_ZERO_COPY = 0x1
-} PfringThreadVarsFlags;
+/* PfringThreadVars flags */
+#define PFRING_FLAGS_ZERO_COPY (1 << 0)
+#define PFRING_FLAGS_BYPASS    (1 << 1)
 
 /**
  * \brief Structure to hold thread specific variables.
@@ -142,6 +142,7 @@ typedef struct PfringThreadVars_
     /* counters */
     uint64_t bytes;
     uint64_t pkts;
+    uint64_t shunt_pkts;
 
     uint16_t capture_kernel_packets;
     uint16_t capture_kernel_drops;
@@ -213,6 +214,9 @@ static inline void PfringDumpCounters(PfringThreadVars *ptv)
         SC_ATOMIC_ADD(ptv->livedev->drop, pfring_s.drop - th_drops);
         StatsSetUI64(ptv->tv, ptv->capture_kernel_packets, pfring_s.recv);
         StatsSetUI64(ptv->tv, ptv->capture_kernel_drops, pfring_s.drop);
+#ifdef PF_RING_FLOW_OFFLOAD
+        ptv->shunt_pkts = pfring_s.shunt;
+#endif
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/source-pfring.c
```diff
@@ -613,6 +671,11 @@ void ReceivePfringThreadExitStats(ThreadVars *tv, void *data)
             StatsGetLocalCounterValue(tv, ptv->capture_kernel_packets),
             StatsGetLocalCounterValue(tv, ptv->capture_kernel_drops));
     SCLogPerf("(%s) Packets %" PRIu64 ", bytes %" PRIu64 "", tv->name, ptv->pkts, ptv->bytes);
```
regit: Indent seems off.

src/runmode-pfring.c
```diff
@@ -354,6 +355,13 @@ static void *ParsePfringConfig(const char *iface)
         }
     }
 
+    if (ConfGetChildValueBoolWithDefault(if_root, if_default, "bypass", &bool_val) == 1) {
+        if (bool_val) {
+            SCLogConfig("Using bypass support in PF_RING for iface %s (if supported by underlying hw)", pfconf->iface);
```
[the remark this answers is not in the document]
cardigliano: There is no API for this atm

src/source-pfring.c
```diff
@@ -213,6 +214,9 @@ static inline void PfringDumpCounters(PfringThreadVars *ptv)
         SC_ATOMIC_ADD(ptv->livedev->drop, pfring_s.drop - th_drops);
         StatsSetUI64(ptv->tv, ptv->capture_kernel_packets, pfring_s.recv);
         StatsSetUI64(ptv->tv, ptv->capture_kernel_drops, pfring_s.drop);
+#ifdef PF_RING_FLOW_OFFLOAD
+        ptv->shunt_pkts = pfring_s.shunt;
```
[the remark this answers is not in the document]
cardigliano: Same as capture.kernel_packets/drops?

src/source-pfring.c
```diff
@@ -493,6 +541,16 @@ TmEcode ReceivePfringThreadInit(ThreadVars *tv, const void *initdata, void **dat
         }
     }
 
+    if (pfconf->flags & PFRING_CONF_FLAGS_BYPASS) {
+#ifdef PF_RING_FLOW_OFFLOAD
+        SCLogInfo("Bypass is supported by this Pfring version");
+        opflag |= PF_RING_FLOW_OFFLOAD | PF_RING_FLOW_OFFLOAD_NOUPDATES;
+        ptv->flags |= PFRING_FLAGS_BYPASS;
+#else
+        SCLogInfo("Bypass is not supported by this Pfring version, please upgrade");
```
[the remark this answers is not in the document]
cardigliano: Ok

src/source-pfring.c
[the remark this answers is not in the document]
regit: Yes something like bypassed or kernel_bypassed

src/source-pfring.c
```diff
@@ -283,6 +287,42 @@ static inline void PfringProcessPacket(void *user, struct pfring_pkthdr *h, Pack
...
+    }
+
+    r.rule_family_type = generic_flow_id_rule;
+    r.rule_family.flow_id_rule.action = flow_drop_rule;
+    r.rule_family.flow_id_rule.thread = 0;
+    r.rule_family.flow_id_rule.flow_id = p->pfring_v.flow_id;
+
+    SCLogDebug("Bypass set for flow ID = %u", p->pfring_v.flow_id);
+
+    if (pfring_add_hw_rule(p->pfring_v.ptv->pd, &r) < 0) {
```
victorjulien: what does it mean if this failed? Should we report failure somehow? Perhaps in a counter?

src/source-pfring.c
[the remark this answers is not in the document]
victorjulien: can we move this check to a global pfring init? Now it's per thread which would spit out the same message many times if many threads are in use.

src/source-pfring.h
```diff
@@ -55,6 +57,16 @@ typedef struct PfringIfaceConfig_
...
 
+/**
+ * \brief per packet Pfring vars
+ *
+ * This structure is used to pass packet metadata in callbacks.
+ */
+typedef struct PfringPacketVars_
+{
+    PfringThreadVars *ptv;
+    u_int32_t flow_id;
```
victorjulien: please use uint32_t

src/source-pfring.c
cardigliano: If for some reason it fails (e.g. hw resources are exhausted) it should fall back to local bypass. Not sure if we need to report it.

src/source-pfring.c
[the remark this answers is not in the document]
cardigliano: What about moving this to ParsePfringConfig()?

src/source-pfring.h
cardigliano: Ok

src/source-pfring.c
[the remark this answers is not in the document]
victorjulien: Sounds great.

src/source-pfring.c
victorjulien: I guess that depends on whether the user can somehow act on this. E.g. assign more resources. If it's purely about hardware limitations then there is not much to be done probably.

cardigliano: With the cards we have today with offload support, it's a pure hw limitation, no user action is needed nor helpful

victorjulien: ok

Another point: shouldn't this func be checked for in configure.ac?

cardigliano: It's under #ifdef PF_RING_FLOW_OFFLOAD, do you think it's not enough?

victorjulien: I think it would be better to add it to configure.ac so that configure output will show if it's available or not. This will help us support users that try to use it.

PR #2977, opened 2017-11-07
Dyn register rust 05

chifflier:

None

Describe changes:

Changes since #2951

Changes since #2953
- Rebased on master

Changes since #2955 
- Registration function split in two (to register protocol detection, and protocol parsing functions), allow more flexibility in default parameters. The parser registration does not check for configuration (like protocol being enabled), this is left to the caller.

Changes since #2959 
- Rebased on master
- Errors in registration function are now fatal. Use this to simplify the rust registration code (we now assume the function can never fail)


src/app-layer-ntp.c
```diff
@@ -38,250 +38,9 @@
...
-
-        AppLayerParserRegisterGetEventInfo(IPPROTO_UDP, ALPROTO_NTP,
-            NTPStateGetEventInfo);
-        AppLayerParserRegisterGetEventsFunc(IPPROTO_UDP, ALPROTO_NTP,
-            NTPGetEvents);
-    }
-    else {
-        SCLogDebug("NTP protocol parsing disabled.");
-    }
+    register_ntp_parser();
```
victorjulien: should this use the rs_ prefix like other Rust code we call from C?

chifflier: Sure, that's a trivial change.
Do you prefer me to add one patch to this PR, or re-send another one?

victorjulien: Lets wait to see if @jasonish comes up with other change requests. Normally I'd prefer a new PR,but it seems a bit excessive here :)

src/app-layer-register.h
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+#ifndef __APP_LAYER_REGISTER_H__
+#define __APP_LAYER_REGISTER_H__
+
+typedef struct AppLayerParser {
+    const char *name;
+    const char *default_port;
+    int ip_proto;
+    const char *proto_name;
```
jasonish: What is the purpose or proto_name? It doesn't appear to be used in the register functions.

Otherwise good. Was able to migrate the DNP3 parser to these new functions with success.

chifflier: The intended use is to provide a way to differentiate the name (used as an identifier) from the display name. But it's true that it's currently unused. We may want to keep it (for later use) or remove, as you prefer.

jasonish: IMO, remove until needed.

victorjulien: I've taken this into my '4.1 next' branch I'm currently QAing. Removed the unused field.

victorjulien:
Merged in #3009. Did remove the unused field and slightly reformat git commit msgs.

Thanks Pierre!

PR #2986, opened 2017-11-08
Metadata v9

regit:
Update of the #2973 addressing comments

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2253

Describe changes:
- Fix error handling (hash got the value)
- Improve unittest

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/344
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/128

PR #2990, opened 2017-11-14
Metadata v10

regit:
Update of #2986 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2253

Describe changes:
- Optional output values as array for all keys or selected
- Documentation update

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/347
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/131

### DIFF
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -170,6 +170,7 @@ Metadata::
             # http-body: yes           # enable dumping of http body in Base64
             # http-body-printable: yes # enable dumping of http body in printable format
             metadata: yes              # add L7/applayer fields, flowbit and other vars to the alert
+            rule-metadata: yes         # add rule metadata as key value in the alert event
 
 Alternatively to the `metadata` key it is also possible to select the application
 layer metadata to output on a per application layer basis ::
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -42,10 +42,18 @@ The most common way to use this is through 'EVE', which is a firehose approach w
             # payload-buffer-size: 4kb # max size of payload buffer to output in eve-log
             # payload-printable: yes   # enable dumping payload in printable (lossy) format
             # packet: yes              # enable dumping of packet (without stream segments)
-            http: yes                # enable dumping of http fields
-            tls: yes                 # enable dumping of tls fields
-            ssh: yes                 # enable dumping of ssh fields
-            smtp: yes                # enable dumping of smtp fields
+            # http-body: yes           # enable dumping of http body in Base64
+            # http-body-printable: yes # enable dumping of http body in printable format
+            metadata: yes              # add L7/applayer fields, flowbit and other vars to the alert
+            # http: yes                # enable dumping of http fields
+            # tls: yes                 # enable dumping of tls fields
+            # ssh: yes                 # enable dumping of ssh fields
+            # smtp: yes                # enable dumping of smtp fields
+
+            rule-metadata:             # dumping of key/value pairs defined by metadata keyword of rule
+              enabled: no              # set to yes to enable
+              output-array: no         # output value of key as an array
+              array-keys: [tag]        # comma separated array of keys to output as array
 
             # Enable the logging of tagged packets for rules using the
             # "tag" keyword.
@@ -170,7 +178,6 @@ Metadata::
             # http-body: yes           # enable dumping of http body in Base64
             # http-body-printable: yes # enable dumping of http body in printable format
             metadata: yes              # add L7/applayer fields, flowbit and other vars to the alert
-            rule-metadata: yes         # add rule metadata as key value in the alert event
 
 Alternatively to the `metadata` key it is also possible to select the application
 layer metadata to output on a per application layer basis ::
@@ -186,6 +193,18 @@ layer metadata to output on a per application layer basis ::
 
 The `vars` will enable dumping of a set of key/value based on flowbits and other vars
 such as named groups in regular expression.
+
+It is also possible to log key/value pairs defined by the metadata keyword on rule. To
+do so you need to set `enabled` to yes under `rule-metadata` ::
+
+        - alert:
+            metadata: yes              # add L7/applayer fields, flowbit and other vars to the alert
+            rule-metadata:             # dumping of key/value pairs defined by metadata keyword of rule
+              enabled: yes              # set to yes to enable
+              output-array: no         # output value of key as an array
+              array-keys: [tag]        # comma separated array of keys to output as array
+
+This will add a `alert.metadata` object in the alert event.
 
 DNS
 ~~~
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -43,6 +43,7 @@
 #include "detect-engine.h"
 #include "detect-engine-mpm.h"
 #include "detect-reference.h"
+#include "detect-metadata.h"
 #include "app-layer-parser.h"
 #include "app-layer-dnp3.h"
 #include "app-layer-htp.h"
@@ -90,6 +91,7 @@
 #define LOG_JSON_FLOW              BIT_U16(11)
 #define LOG_JSON_HTTP_BODY         BIT_U16(12)
 #define LOG_JSON_HTTP_BODY_BASE64  BIT_U16(13)
+#define LOG_JSON_RULE_METADATA     BIT_U16(14)
 
 #define LOG_JSON_METADATA_ALL  (LOG_JSON_APP_LAYER|LOG_JSON_HTTP|LOG_JSON_TLS|LOG_JSON_SSH|LOG_JSON_SMTP|LOG_JSON_DNP3|LOG_JSON_VARS|LOG_JSON_FLOW)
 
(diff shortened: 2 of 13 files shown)


src/output-json-alert.c
```diff
@@ -230,9 +239,49 @@ static void AlertJsonSourceTarget(const Packet *p, const PacketAlert *pa,
...
+                    json_object_set_new(mjs, kv->key, jkey);
+                } else {
+                    json_array_append_new(jkey, json_string(kv->value));
+                }
+            } else {
+                json_object_set_new(mjs, kv->key, json_string(kv->value));
+            }
+            kv = kv->next;
+        }
+        json_object_set_new(ajs, "metadata", mjs);
```
victorjulien: I don't think it's a good idea to have a object 'metadata' that can be either a map or an array. How would parsers deal with that?

victorjulien: Seems a pure array approach makes much more sense to me still. It would work w/o config for all metadata fields, not just for the one the user had to configure.

regit: Metadata is always a map but some keys in the object can be depending on the key value a array or a single value.

```

Comment (victorjulien) [reply src/output-json-alert.c]: Ok it's not metadata but per field, the issue remains the same. Parsers can't assume they will know the type.

Whats wrong with a straight array? It's simpler and requires no config.

Comment (jasonish) [reply src/output-json-alert.c]: I can see what @regit is trying to do here... Make it a scalar where it makes sense, and a list where that makes sense.. Based on the current status of metadata found in rules. Unfortunately I don't feel we can go ahead with such a thing without a formal specification/schema of what metadata should look like, and assign data types to the metadata types.

I also don't want to see it get hardcoded into the config, its hard to update and do the best thing over time.

Which unfortunately leaves arrays as the lowest common denominator.

I'm late to reviewing this PR, so still digesting it.

File: doc/userguide/output/eve/eve-json-output.rst [hunk 2]
```diff
@@ -42,10 +42,18 @@ The most common way to use this is through 'EVE', which is a firehose approach w
...
+            metadata: yes              # add L7/applayer fields, flowbit and other vars to the alert
+            # http: yes                # enable dumping of http fields
+            # tls: yes                 # enable dumping of tls fields
+            # ssh: yes                 # enable dumping of ssh fields
+            # smtp: yes                # enable dumping of smtp fields
+
+            rule-metadata:             # dumping of key/value pairs defined by metadata keyword of rule
+              enabled: no              # set to yes to enable
+              output-array: no         # output value of key as an array
+              array-keys: [tag]        # comma separated array of keys to output as array
```
jasonish: Setting output-array to yes makes them all arrays. Leaving it at no, but adding tags to array-keys makes only those listed arrays. I think that needs to be made clear here.

doc/userguide/output/eve/eve-json-output.rst
```diff
@@ -186,6 +194,18 @@ layer metadata to output on a per application layer basis ::
...
+do so you need to set `enabled` to yes under `rule-metadata` ::
+
+        - alert:
+            metadata: yes              # add L7/applayer fields, flowbit and other vars to the alert
+            rule-metadata:             # dumping of key/value pairs defined by metadata keyword of rule
+              enabled: yes              # set to yes to enable
+              output-array: no         # output value of key as an array
+              array-keys: [tag]        # comma separated array of keys to output as array
+
+This will add a `alert.metadata` object in the alert event.
```
jasonish: Looking at an alert with metadata I found it confusing as to what the metadata is.  Its rule metadata, I wonder if that should be more clear in the key name, or if the fields should be prefixed by by rule_ or something? Some are clear... But for example..

alert.metadata.updated_at: ....

What was updated at that time? I know its the rule, but its ambiguous.

doc/userguide/output/eve/eve-json-output.rst
regit: OK, I can update the comment here.

src/output-json-alert.c
regit: With array you need to have array access in all your searches so this is quite an overload in term of processing and coding.

doc/userguide/output/eve/eve-json-output.rst
regit: alert.rule-metadata ?

doc/userguide/output/eve/eve-json-output.rst
regit: @victorjulien I can also set the default to array if you want to.

jasonish:
On Tue, Nov 14, 2017 at 4:22 PM, Eric Leblond <notifications@github.com>
wrote:

> ------------------------------
>
> In src/output-json-alert.c
> <https://github.com/OISF/suricata/pull/2990#discussion_r150863922>:
>
> > +                json_t *jkey = json_object_get(mjs, kv->key);
> +                if (jkey == NULL) {
> +                    jkey = json_array();
> +                    if (jkey == NULL)
> +                        return;
> +                    json_array_append_new(jkey, json_string(kv->value));
> +                    json_object_set_new(mjs, kv->key, jkey);
> +                } else {
> +                    json_array_append_new(jkey, json_string(kv->value));
> +                }
> +            } else {
> +                json_object_set_new(mjs, kv->key, json_string(kv->value));
> +            }
> +            kv = kv->next;
> +        }
> +        json_object_set_new(ajs, "metadata", mjs);
>
> With array you need to have array access in all your searches so this is
> quite an overload in term of processing and coding.
>
Is this still true with ES?

For example:
{"metadata": {"updated_at": ["2010_09_23"], "created_at": ["2010_09_23"]}}
OR
{"metadata": {"updated_at": "2010_09_23", "created_at": "2010_09_23"}}

Would both hit an ES search like: metadata.updated_at:"2010_09_23"

Didn't test just now, but I thought I tested this before. So will need to
retest soon.

jasonish:
> Is this still true with ES?
> 
> For example:
> {"metadata": {"updated_at": ["2010_09_23"], "created_at": ["2010_09_23"]}}
> OR
> {"metadata": {"updated_at": "2010_09_23", "created_at": "2010_09_23"}}
> 
> Would both hit an ES search like: metadata.updated_at:"2010_09_23"
> 
> Didn't test just now, but I thought I tested this before. So will need to
> retest soon.
> Merge state

At least with a query string type search, this seems to hold true with Elastic Search 5.

victorjulien:
What I dislike is that this is going to assign magic properties to a field that was supposed to be nothing more than a collection of k/v pairs. Now we need confusing config to make it work. I think we either need to formalize the metadata field completely (meaning define fields, their meaning, reject everything else), or we keep it simple and make it an array.

I think formalizing the metadata field has things going for it, but it would probably need to happen in a new keyword instead.

jasonish:
> What I dislike is that this is going to assign magic properties to a field that was supposed to be nothing more than a collection of k/v pairs. Now we need confusing config to make it work. I think we either need to formalize the metadata field completely (meaning define fields, their meaning, reject everything else), or we keep it simple and make it an array.
> 
> I think formalizing the metadata field has things going for it, but it would probably need to happen in a new keyword instead.

Yes, I agree.  Arrays is the only generic approach at this time. And for many uses cases, ELK for instance, I don't think it makes the query any more difficult.

regit:
Regarding parsing it will just be a square bracket nightmare if we have array. All code accessing the metadata will need to access element 0 in 90% of the cases.

jasonish:
Just analyzed the current ET Pro ruleset for Suricata and found all these metatags listed more than once in 1 or more rules:
```
   2222 affected_product
      1 attack_target
      2 created_at
    542 deployment
      4 former_category
    719 malware_family
      1 signature_severity
   1397 tag
      2 updated_at
```

Other observation was that they all seem to honour the "key val" convention.

IMO, Suricata needs to use lists. I don't think this provides any downsides to ELK. Custom integrations can post-process log as needed.

regit:
Good analysis, could we default to array and let user update the settings if he wants to ?++Le 20 nov. 2017 8:58 PM, Jason Ish <notifications@github.com> a écrit :Just analyzed the current ET Pro ruleset for Suricata and found all these metatags listed more than once in 1 or more rules:
   2222 affected_product
      1 attack_target
      2 created_at
    542 deployment
      4 former_category
    719 malware_family
      1 signature_severity
   1397 tag
      2 updated_at

Other observation was that they all seem to honour the "key val" convention.
IMO, Suricata needs to use lists. I don't think this provides any downsides to ELK. Custom integrations can post-process log as needed.

jasonish:
> Good analysis, could we default to array and let user update the settings if he wants to ?

I think I'd be OK with that.  I'd like to see it arrays by default. And very intentional that a user turns that off for specific fields, to limit accidental changing of those. Maybe even take that option out of the default config and just add it to the documentation (I don't think I really care if its in the default suricata.yaml or not actually).

Should probably also document what happens if its not a list, and multiple ones are seen. First one win, last one? Are they concatenated, perhaps separated by spaces?

victorjulien:
Shall we split this in two? Step one: log rule metadata as list. Step two: the optional logging as map. That leads to a simple first PR.

jasonish:
Merged here: https://github.com/OISF/suricata/pull/3209

However, was modified to log all metadata fields as lists.

PR #2991, opened 2017-11-14
[discussion] flowbit logging: add to flows; log as array - 1

jasonish:
Putting this out here for discussion:
- First, add the vars/flowbit logging to flow records.
- Second, log the flowbits as an array instead of a map where the flowbit is a key with a true value. This is the point of discussion, and I think it will also make searching/correlation with Elastic Search easier.
Example
```
  "vars": {
    "flowbits": [
      "traffic/id/facebook",
      "traffic/label/social-network",
      "traffic/id/facebook-messenger",
      "traffic/label/im"
    ]
  },
```

As this changes Eve this also involves the discussion of versioning. As "vars" is also an event type, this complicates things. Do we update the version of "eve", as vars is also added to non-"var" events. Or do we just update the version of "vars".

Thought: Introduce a new object, "attributes", "parameters" or even "metadata' which then contains the list of flowbits, etc. More or less what the "vars" is I guess.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/215
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/568

### DIFF
--- a/src/output-json-flow.c
+++ b/src/output-json-flow.c
@@ -277,6 +277,14 @@ static void JsonFlowLogJSON(JsonFlowLogThread *aft, json_t *js, Flow *f)
 
     json_object_set_new(hjs, "alerted", json_boolean(FlowHasAlerts(f)));
 
+    if (f->flowvar) {
+        json_t *js_vars = json_object();
+        if (likely(js_vars != NULL)) {
+            JsonAddFlowvars(f, js_vars);
+            json_object_set_new(js, "vars", js_vars);
+        }
+    }
+
     json_object_set_new(js, "flow", hjs);
 
 
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -226,11 +226,11 @@ void JsonAddFlowvars(const Flow *f, json_t *js_vars)
             const char *varname = VarNameStoreLookupById(fb->idx, VAR_TYPE_FLOW_BIT);
             if (varname) {
                 if (js_flowbits == NULL) {
-                    js_flowbits = json_object();
+                    js_flowbits = json_array();
                     if (js_flowbits == NULL)
                         break;
                 }
-                json_object_set_new(js_flowbits, varname, json_boolean(1));
+                json_array_append(js_flowbits, json_string(varname));
             }
         }
         gv = gv->next;
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -154,7 +154,7 @@ static void JsonAddPacketvars(const Packet *p, json_t *js_vars)
     }
 }
 
-static void JsonAddFlowvars(const Flow *f, json_t *js_vars)
+void JsonAddFlowvars(const Flow *f, json_t *js_vars)
 {
     if (f == NULL || f->flowvar == NULL) {
         return;
--- a/src/output-json.h
+++ b/src/output-json.h
@@ -39,6 +39,7 @@ typedef struct OutputJSONMemBufferWrapper_ {
 
 int OutputJSONMemBufferCallback(const char *str, size_t size, void *data);
 
+void JsonAddFlowvars(const Flow *f, json_t *js_vars);
 void JsonAddVars(const Packet *p, const Flow *f, json_t *js);
 void CreateJSONFlowId(json_t *js, const Flow *f);
 void JsonTcpFlags(uint8_t flags, json_t *js);


victorjulien:
I think it would make sense if we list our entire naming tree including the types.

I agree flowbits should be logged as an array instead. I think we can leave 'vars' and deprecate it. Then add a new object like you suggest. Not sure on the name. Some of the vars are about labeling, some are about augmenting & some are about extraction. Not sure about a good common name.

jasonish:
> I think it would make sense if we list our entire naming tree including the types.

I'm not sure what you mean here?

victorjulien:
On 14-11-17 16:21, Jason Ish wrote:
>     I think it would make sense if we list our entire naming tree
>     including the types.
> 
> I'm not sure what you mean here?

Our object naming tree is getting so large that I'm getting fuzzy about
what we have and what would make sense to add. So thinking it might be
good idea to create an overview/list to get some clarity.

jasonish:
On Tue, Nov 14, 2017 at 4:23 PM, Victor Julien <notifications@github.com>
wrote:

> On 14-11-17 16:21, Jason Ish wrote:
> > I think it would make sense if we list our entire naming tree
> > including the types.
> >
> > I'm not sure what you mean here?
>
> Our object naming tree is getting so large that I'm getting fuzzy about
> what we have and what would make sense to add. So thinking it might be
> good idea to create an overview/list to get some clarity.


Of everything that can be logged to eve? Yeah, I agree. That is something I
can take a look at doing.

victorjulien:
On 14-11-17 16:26, Jason Ish wrote:
> On Tue, Nov 14, 2017 at 4:23 PM, Victor Julien <notifications@github.com>
> wrote:
> 
>> On 14-11-17 16:21, Jason Ish wrote:
>> > I think it would make sense if we list our entire naming tree
>> > including the types.
>> >
>> > I'm not sure what you mean here?
>>
>> Our object naming tree is getting so large that I'm getting fuzzy about
>> what we have and what would make sense to add. So thinking it might be
>> good idea to create an overview/list to get some clarity.
> 
> 
> Of everything that can be logged to eve? Yeah, I agree. That is something I
> can take a look at doing.

It would probably also be helpful in the rule metadata conversation.

victorjulien:
Would it make sense to have a separate logging logic for flowbits with a 'special' name? E.g. if a bit starts with 'traffic/' have it go into it's own json array.

jasonish:
> Would it make sense to have a separate logging logic for flowbits with a 'special' name? E.g. if a bit starts with 'traffic/' have it go into it's own json array.

While I think it would make sense in the output, I hate to add magic to some names, just like the "tags" in @regit's metadata patch. I guess its less magic if the convention is documented from the start so it might be a fine idea for the long term.

The nice thing about flowbits is they exist now, but would a "label" keyword be better in the long term? This is explicitly for labelling flows on the output? It might be so close the flowbits that it just adds confusion. 

Also, even in the vars I think "flowbit":true is still a flawed way of logging such data. But as you said above, maybe the whole vars is set for deprecation in which case its better leave as is, and make the replacement "correct".

jasonish:
I took one of my eve files and attempted to document all items and types. Its not exhaustive as its only based on samples I have.

I wonder if in the long run we should try to keep something like this up to date. Including version added, version removed, etc. The initial version could just be "4.0" and somewhat auto-generated. And moving forward any change to eve requires updates to this document before merging (which in the case maybe its best to go into sphinx).

https://docs.google.com/spreadsheets/d/1zGOjKLoeVxsnr8TNlQVvcYey1wvZ_AfLCnA8_qSoNNw/edit?usp=sharing

inliniac:
On 21-11-17 14:37, Jason Ish wrote:
> I took one of my eve files and attempted to document all items and
> types. Its not exhaustive as its only based on samples I have.
> 
> I wonder if in the long run we should try to keep something like this up
> to date. Including version added, version removed, etc. The initial
> version could just be "4.0" and somewhat auto-generated. And moving
> forward any change to eve requires updates to this document before
> merging (which in the case maybe its best to go into sphinx).
> 
> https://docs.google.com/spreadsheets/d/1zGOjKLoeVxsnr8TNlQVvcYey1wvZ_AfLCnA8_qSoNNw/edit?usp=sharing
> 


Good stuff. Would it make sense to have a look at json schema's?

https://redmine.openinfosecfoundation.org/issues/1369

jasonish:
> Good stuff. Would it make sense to have a look at json schema's?

In the long run, yes. But it looks like a lot of work. I also find it harder to read, so wonder if more people would get more value out of a simple table.

Back to the long run, perhaps as we overhaul sections, or add new items, we make sure they can be expressed as a schema.

victorjulien:
Like with 'vars', the new object would contain flowbits (tags? labels? flags?), flowvars (extracted data), flowints, plus the same for hosts, ippairs. We could consider splitting as well.

victorjulien:
I think we have agreement that we need a new design for the various things we can add to records: vars, metadata, rule, etc.

PR #2993, opened 2017-11-16
New feature : set dynamic host policy with p0f

mephesto1337:

None

Describe changes:
- Add new Unix Socket command to set host reassembly policy dynamically
- An additionnal patch is required for p0f, but it written

### DIFF
--- a/src/unix-manager.c
+++ b/src/unix-manager.c
@@ -34,6 +34,7 @@
 #include "util-device.h"
 #include "util-signal.h"
 #include "util-buffer.h"
+#include "util-host-os-info.h"
 
 #if (defined BUILD_UNIX_SOCKET) && (defined HAVE_SYS_UN_H) && (defined HAVE_SYS_STAT_H) && (defined HAVE_SYS_TYPES_H)
 #include <sys/un.h>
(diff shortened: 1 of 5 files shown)


src/unix-manager.c
```diff
@@ -728,6 +729,65 @@ static TmEcode UnixManagerListCommand(json_t *cmd,
...
+        if (!json_is_string(jarg)) {
+            errmsg= "ipv6 is not a string";
+            goto error;
+        }
+        ipv6 = json_string_value(jarg);
+    }
+
+    if ( ipv4 ) {
+        SCLogInfo("Set host %s to OS %s", ipv4, ostype);
+        if (SCHInfoAddHostOSInfo(ostype, ipv4, 1) < 0) {
```
victorjulien: sadly, this function is not thread safe. We can have multiple readers of the data structure storing the host os info. Adding a single mutex wouldn't work either as that is too much of a contention point. So this will need some more thought.

src/unix-manager.c
```diff
@@ -728,6 +729,65 @@ static TmEcode UnixManagerListCommand(json_t *cmd,
...
+        }
+    } else {
+        errmsg = "no valid ipv4 or ipv6";
+        goto error;
+    }
+
+    json_object_set_new(answer, "message", json_string("OK"));
+    return TM_ECODE_OK;
+
+    error:
```
victorjulien: style: label at beginning of line

src/unix-manager.c
```diff
@@ -728,6 +729,65 @@ static TmEcode UnixManagerListCommand(json_t *cmd,
...
+
+    SCLogInfo("error: %s", errmsg);
```
victorjulien: SCLogError

src/unix-manager.c
mephesto1337: What would be the better, spinlocks or build an RCU structure around the radix tree ?

victorjulien: As discussed during SuriCon brainstorm, I think the easiest way would be to use the host table. It would only support single IP-addresses though.

PR #2996, opened 2017-11-17
ConfGetInt in conf.c: NULL-pointer dereference

whotwagner:

Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/2275

Describe changes:

 - This changes will validate strval in ConfGetInt right before strtoimax is called. If strval is NULL, the function will exit with an proper error.

PRScript output (if applicable):

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -386,6 +386,12 @@ int ConfGetInt(const char *name, intmax_t *val)
     if (ConfGet(name, &strval) == 0)
         return 0;
 
+    if (strval == NULL) {
+        SCLogError(SC_ERR_INVALID_YAML_CONF_ENTRY, "malformed integer value "
+                "for %s: NULL", name);
+        return 0;
+    }
+
     errno = 0;
     tmpint = strtoimax(strval, &endptr, 0);
     if (strval[0] == '\0' || *endptr != '\0') {


src/conf.c
```diff
@@ -386,9 +386,15 @@ int ConfGetInt(const char *name, intmax_t *val)
...
+    if (strval == NULL) {
+        SCLogError(SC_ERR_INVALID_YAML_CONF_ENTRY, "malformed integer value "
+                "for %s: NULL", name);
+        return 0;
+    }
+
     errno = 0;
     tmpint = strtoimax(strval, &endptr, 0);
-    if (strval[0] == '\0' || *endptr != '\0') {
+    if (*endptr != '\0') {
```
victorjulien: why is the first condition removed here?

victorjulien:
Could you fixup the formatting of the PR description? It should look like https://github.com/OISF/suricata/pull/2990

victorjulien:
Could also update the commit subject line to be something like
"conf: fix NULL-pointer dereference in ConfGetInt"

We add the subsystem first, then a short description. No need to mention the filename, that is already visible in the diff itself. Thanks!

src/conf.c
whotwagner: Because this condition is checked before strtoimax() now so that strval can't be NULL when it's used by strtoimax. Since strtoimax does not change strval, another check of strval would be useless.

victorjulien: The checks are not the same though. One checks that the ptr != NULL, the other checks that the first char is not \0

dbcfd: Shouldn't the strval[0] check be moved up with strval == NULL so you don't even bother calling strtoimax?

whotwagner: How embarrassing. Of course you are right with the first char. I put the strval[0]-check back to it's place, but if you want I would also move it up to the NULL-check before strtoimax.

victorjulien:
Merged in #3001, thanks!

PR #3002, opened 2017-11-20
Expectation ftp v19

regit:
Update of #2998

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/550

Describe changes:
- rebase on latest master
- merge 2 first commits on app-layer-expectation

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/351
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/135

PR #3003, opened 2017-11-21
Expectation ftp v20

regit:
Update of #3002 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/550

Describe changes:
- Fix SCLogDebug call

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/352
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/136


src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+#define EXPECTATION_TIMEOUT 30
+
+typedef struct Expectation_ {
+    struct timeval ts;
+    Port sp;
+    Port dp;
+    AppProto alproto;
+    int direction;
+    void *data;
+    void (*DFree)(void *);
```
victorjulien: is this to free the data field? Comment!

src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+
+    SC_ATOMIC_ADD(expectation_count, 1);
+    /* As we are creating the expectation, we release lock on IPPair without
+     * setting the ref count to 0. This way the IPPair will be kept till
+     * cleanup */
+    IPPairUnlock(ipp);
+    return 0;
+
+error:
+    if (exp != NULL)
```
victorjulien: exp can't be NULL AFAICS

src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+     */
+    (void) IPPairDecrUsecnt(ipp);
+    SC_ATOMIC_SUB(expectation_count, 1);
+    if (pexp == NULL) {
+        IPPairSetStorageById(ipp, g_expectation_id, lexp);
+    } else {
+        pexp->next = lexp;
+    }
+    if (exp->data)
+        SCFree(exp->data);
```
victorjulien: Is this the field that should be freed by the DFree func ptr?

src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+
+    int x = SC_ATOMIC_GET(expectation_count);
+    if (x == 0) {
+        return ALPROTO_UNKNOWN;
+    }
+
+    Expectation *exp = AppLayerExpectationGet(f, direction, &ipp);
+    time_t ctime = f->lastts.tv_sec;
+
+    if (exp == NULL)
```
victorjulien: seems to make sense to check this right after the 'Get'?

src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+    IPPair *ipp = NULL;
+    Expectation *lexp = NULL;
+    Expectation *pexp = NULL;
+
+    int x = SC_ATOMIC_GET(expectation_count);
+    if (x == 0) {
+        return ALPROTO_UNKNOWN;
+    }
+
+    Expectation *exp = AppLayerExpectationGet(f, direction, &ipp);
```
victorjulien: add comment like: "'takes reference of the ip pair in 'ipp'"

src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+    pexp = NULL;
+    while (exp) {
+        lexp = exp->next;
+        if ( (exp->direction & direction) &&
+             ((exp->sp == 0) || (exp->sp == f->sp)) &&
+             ((exp->dp == 0) || (exp->dp == f->dp))) {
+            alproto = exp->alproto;
+            f->alproto_ts = alproto;
+            f->alproto_tc = alproto;
+            if (FlowSetStorageById(f, g_expectation_data_id, exp->data) != 0) {
```
victorjulien: comment like: transfer ownership of exp->data to flow

src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    Expectation *exp = AppLayerExpectationGet(f, direction, &ipp);
+    time_t ctime = f->lastts.tv_sec;
+
+    if (exp == NULL)
+        goto out;
+
+    pexp = NULL;
+    while (exp) {
```
victorjulien: it seems that if we have multiple expectations that apply to a flow each of them would be used to update the flow. Would FlowSetStorageById handle that case w/o loosing memory?

src/app-layer-ftp.c
```diff
@@ -228,31 +272,121 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
+                state->port_line_len = state->current_line_len;
+                break;
+            case FTP_COMMAND_RETR:
+                /* change direction (default to server) so expectation will handle
+                 * the correct message when expectation will match.
+                 */ 
+                direction = STREAM_TOCLIENT;
+                // fallthrough
+            case FTP_COMMAND_STOR:
+                {
```
victorjulien: can you add a comment that if we get here 'current_line_len' is at least >=5? This is the type of code that looks suspicious as it doesn't check things itself.

I would also not be opposed to actually checking the side here, purely as a defensive measure so future changes to FTPGetLine/FTPParseRequestCommand don't cause unintended issues

src/app-layer-ftp.c
```diff
@@ -228,31 +272,121 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
+                    struct FtpTransferCmd *data = SCCalloc(1, sizeof(struct FtpTransferCmd));
+                    if (data == NULL)
+                        SCReturnInt(-1);
+                    data->DFree = FtpTransferCmdFree;
+                    data->filename = SCCalloc(state->current_line_len - 4, sizeof(char));
+                    if (data->filename == NULL) {
+                        SCFree(data);
+                        SCReturnInt(-1);
+                    }
+                    memcpy(data->filename, state->current_line + 5, state->current_line_len - 5);
```
victorjulien: it seems the goal here is to create a nul-terminated string. Two things:

1. please make this explicit
2. is this guaranteed to be a valid string? As it comes from the wire my suspicion is 'no'. So a byte array instead?

src/app-layer-ftp.c
```diff
@@ -446,6 +795,32 @@ void RegisterFTPParsers(void)
...
+        AppLayerParserRegisterGetTx(IPPROTO_TCP, ALPROTO_FTPDATA, FTPDataGetTx);
+
+        AppLayerParserRegisterGetTxCnt(IPPROTO_TCP, ALPROTO_FTPDATA, FTPDataGetTxCnt);
+
+        AppLayerParserRegisterGetStateProgressFunc(IPPROTO_TCP, ALPROTO_FTPDATA, FTPDataGetAlstateProgress);
+
+        AppLayerParserRegisterGetStateProgressCompletionStatus(ALPROTO_FTPDATA,
+                FTPDataGetAlstateProgressCompletionStatus);
+
+        sbcfg.buf_size = 4096;
```
victorjulien: seems the file data itself is excluded from the memcap. To include it register alloc/free funcs here for the streaming api

src/app-layer-ftp.c
```diff
@@ -465,6 +840,34 @@ void FTPAtExitPrintStats(void)
 #endif
 }
 
+
+#ifdef HAVE_LIBJANSSON
+json_t *JsonFTPDataAddMetadata(const Flow *f)
+{
+    FtpDataState *ftp_state = NULL;
```
victorjulien: const

src/app-layer-ftp.h
```diff
@@ -131,15 +132,39 @@ typedef struct FtpState_ {
...
     DetectEngineState *de_state;
 } FtpState;
 
+enum {
+    FTPDATA_STATE_IN_PROGRESS,
+    FTPDATA_STATE_FINISHED,
+};
+
+/** FTP Data State for app layer parser */
+typedef struct FtpDataState_ {
```
victorjulien: I think the layout can be optimized

src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+    } else if (FLOW_IS_IPV6(f)) {
+        FLOW_COPY_IPV6_ADDR_TO_PACKET(&f->src, ip_src);
+        FLOW_COPY_IPV6_ADDR_TO_PACKET(&f->dst, ip_dst);
+    } else {
+        return -1;
+    }
+    return 0;
+}
+
+static Expectation *AppLayerExpectationGet(Flow *f, int direction, IPPair **ipp)
```
victorjulien: AppLayerExpectationLookup? It wraps around IPPairLookupIPPairFromHash, not IPPairGetIPPairFromHash

src/app-layer-expectation.c
```diff
@@ -0,0 +1,311 @@
...
+ *
+ * \return expectation data identifier
+ */
+int AppLayerExpectationGetDataId(void)
+{
+    return g_expectation_data_id;
+}
+
+static Expectation * RemoveExpectationAndGetNext(IPPair *ipp,
+                                Expectation *pexp, Expectation *exp,
```
victorjulien: what are the names pexp and lexp referring to?

src/app-layer-ftp.c
victorjulien: *checking the size

src/app-layer-expectation.c
regit: Just updated that part of the code to only get the DFree in ExpectationData.

src/app-layer-expectation.c
regit: Adding documentation to the function.

src/app-layer-ftp.c
regit: From https://tools.ietf.org/html/rfc959 at section 5.3.2 we are suppose to have ASCII only here. I'm adding a check here. Mister @chifflier do you know if I should be more laxist here ?

victorjulien: Making it a byte array makes it match behavior of our other protocol parsers and gives consistent matching logic in our keywords. So this is what I think we should do.

regit: OK, will do that.

src/app-layer-ftp.c
regit: I got the idea but don't manage to understand what to do here. Can you tell me which functions to use ?

victorjulien: see https://github.com/OISF/suricata/blob/master/src/app-layer-htp.c#L2274

The idea is to register your own alloc/free funcs that track memuse/memcap

regit: Do you want me to introduce a new memcap or should I use the reassembly one ?

victorjulien: Not sure what reassembly has to do with it.

Whats needed is a FTP specific memcap.

regit: Ok :+1:

regit:
Gonna resubmit.

PR #3004, opened 2017-11-21
Dns response 2198 v1

StianHB:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues/2198) ticket:2198

Describe changes:
- Created the dns_response keyword that allows matching in rules in A record responses in DNS
- Updated the user-docs to reflect the new keyword

PR #3006, opened 2017-11-22
detect/dns: added functionality for dns_response keyword

StianHB:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues/) ticket:[2198](https://redmine.openinfosecfoundation.org/projects/suricata/issues/2198)

Describe changes:
- Created the dns_response keyword that allows matching in rules in A record responses in DNS
- Updated the user-docs to reflect the new keyword

Incorporated changes from #3004 

Could use some feedback on the segfault from the unit test. Tried it again on my end with no issue.


jasonish:
> Could use some feedback on the segfault from the unit test. Tried it again on my end with no issue.

The segfaul is happening on the Rust enabled builds. Take a look at the #ifdef RUST in the query name handling. Similar will need to be done for the response handling. I can help out a bit here.

Cherry pick this commit from my repo:
https://github.com/jasonish/suricata/commit/ef0805628438cb582800fb36b1099fd250bdede4

This adds the function to Rust to access the answer rdata. However, I just realized that it doesn't give you the rdata type like A, and AAAA and so on.

@victorjulien What are your thoughts on allowing humanized matching of IP addresses in the response data?

src/app-layer-dns-common.c
```diff
@@ -851,11 +851,6 @@ const uint8_t *DNSReponseParse(DNSState *dns_state, const DNSHeader * const dns_
         case DNS_RECORD_TYPE_A:
         {
             if (datalen == 0 || datalen == 4) {
-                //PrintRawDataFp(stdout, data, ntohs(head->len));
```
victorjulien: code cleanups like this should go into their own commits

src/detect-dns-response.c
```diff
@@ -0,0 +1,212 @@
...
+    s->alproto = ALPROTO_DNS;
+    return 0;
+}
+
+
+
+#ifdef UNITTESTS
+
+/** \test simple dns response match on A record */
+static int DetectDnsResponseTest01(void)
```
victorjulien: can you convert this to fully using the FAIL_IF style macro's only? No conditions, just a linear path of checks until PASS.

src/detect-engine-dns.c
```diff
@@ -119,6 +120,48 @@ int DetectEngineInspectDnsQueryName(ThreadVars *tv,
...
+        if(answer->type == DNS_RECORD_TYPE_A) {
+            det_ctx->discontinue_matching = 0;
+            det_ctx->buffer_offset = 0;
+            det_ctx->inspection_recursion_counter = 0;
+
+            const uint8_t *temp = (const uint8_t *) ((uint8_t *) answer +
+                                    sizeof(DNSAnswerEntry) + answer->fqdn_len);
+
+            char a[16] = "";
+            PrintInet(AF_INET, (const void *) temp, a, sizeof(a));
```
victorjulien: is this safe? PrintInet unconditionally reads 4 bytes from 'temp'. I would like to see a safety check here.

src/detect-engine-dns.c
```diff
@@ -119,6 +120,48 @@ int DetectEngineInspectDnsQueryName(ThreadVars *tv,
     return r;
 }
 
+int DetectEngineInspectDnsResponse(ThreadVars *tv,
```
victorjulien: we had a DetectEngineInspectDnsResponse, it wrapped around a 'generic list' inspect function. Now it's a buffer inspect function. This means that the old logic is gone which means that Lua scripts will be broken.

detect-lua.c:1030 sets up lua scripts in the 'generic list' for the list with name 'dns_response'. Luckily this name is only used internally, so you can change it. But you will have to make sure not to break this.

doc/userguide/rules/dns-keywords.rst
```diff
@@ -22,6 +22,17 @@ used in a rule all contents following it are affected by it.  Example:
 The dns_query keyword affects all following contents, until pkt_data
 is used or it reaches the end of the rule.
 
+dns_response
+------------
+
+The **dns_response** keyword allows you to match IPs in A record responses.
+
+ Example:
+ alert dns any any -> any any (msg:"Test dns_reputation option";
```
victorjulien: Lets not make any reference to 'reputation' here as it's a simple match. Lets reserve reputation for when we can make large number of ip's.

victorjulien:
"What are your thoughts on allowing humanized matching of IP addresses in the response data?"

Good question. I guess it's convenient to have it human readable. I guess we could also make it binary and then add a inet_ntoa transform when that API is ready?

jasonish:
> "What are your thoughts on allowing humanized matching of IP addresses in the response data?"
> 
> Good question. I guess it's convenient to have it human readable. I guess we could also make it binary and then add a inet_ntoa transform when that API is ready?

Right now, as "dns_query" refers to the request name, I think it makes perfect sense for the "dns_response" keyword to refer to the answer data. This could be an IP address, or a hostname. I personally think it should be kept as binary data for now.  Perhaps an xform, or a more specific keyword down the road? But keeping it binary lets you match on A, or CNAME, or TXT, etc.

src/detect-engine-dns.c
victorjulien: A possible source of confusion is that internally we name 'buffers'. These names may be the same as a keyword inspecting them, but this doesn't have to be the case. It would probably be good to first rename the existing 'dns_response' buffer to 'dns_response_generic' or something. Then in a 2nd commit, add your 'dns_response' keyword. That should solve the confusion around what 'dns_response' means.

PR #3005, opened 2017-11-21
preview: faster rule parsing by not using pcre - v4

jasonish:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [ x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Don't use pcre to parse the rule options at the top level of parsing. Performs much better, especially on lower end CPUs. Could also allow for better context when parsing fails.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/217
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/570

PR #3007, opened 2017-11-22
preview: faster rule parsing by not using pcre - v5

jasonish:

Previous PRs:
- https://github.com/OISF/suricata/pull/3005

Don't use pcre to parse the rule options at the top level of parsing. Performs much better, especially on lower end CPUs. Could also allow for better context when parsing fails.

Changes from last PR:
- Handle rules with carriage returns at the end.
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/218
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/571

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -70,11 +70,6 @@
 
 extern int sc_set_caps;
 
-static pcre *config_pcre = NULL;
-static pcre *option_pcre = NULL;
-static pcre_extra *config_pcre_extra = NULL;
-static pcre_extra *option_pcre_extra = NULL;
-
 static void SigMatchTransferSigMatchAcrossLists(SigMatch *sm,
         SigMatch **src_sm_list, SigMatch **src_sm_list_tail,
         SigMatch **dst_sm_list, SigMatch **dst_sm_list_tail);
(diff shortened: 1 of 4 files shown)


victorjulien:
Nice checkboxes @jasonish ;-)

This passes my QA. What do you consider 'preview' about this?

I guess it would be nice to take it further, and replace (most of) the individual rule keyword parser pcre usage as well. One reason for doing this would be that we're on pcre1 and upstream works on pcre2 mostly.

jasonish:
I guess I don't consider this final one a preview. I'd like to look into some other keywords, but when it comes to performance, doing it here was the big gain, doing each keyword would be much less. But it could get us off pcre1 like you say.

But I'd like to see this in before that is complete for the performance gains I see on lower end CPUs.

src/detect-parse.c
```diff
@@ -581,70 +553,59 @@ int SigMatchListSMBelongsTo(const Signature *s, const SigMatch *key_sm)
...
-        exit(1);
+    /* Look for the end of this option, handling escaped semicolons. */
+    char *optend = optstr;
+    for (;;) {
+        optend = strchr(optend, ';');
+        if (optend == NULL) {
+            SCLogError(SC_ERR_INVALID_SIGNATURE, "no terminating \";\" found");
+            goto error;
+        }
+        if (*(optend -1) == '\\') {
```
victorjulien: if there is only a ';', then this would be an out of bounds read?

src/detect-parse.c
```diff
@@ -973,41 +916,137 @@ static int SigParseAction(Signature *s, const char *action)
...
         Signature *s, const char *sigstr, SignatureParser *parser, uint8_t addrs_direction)
 {
-#define MAX_SUBSTRINGS 30
-    int ov[MAX_SUBSTRINGS];
-    int ret = 0;
-
-    ret = pcre_exec(config_pcre, config_pcre_extra, sigstr, strlen(sigstr), 0, 0, ov, MAX_SUBSTRINGS);
-    if (ret != 8 && ret != 9) {
-        SCLogDebug("pcre_exec failed: ret %" PRId32 ", sigstr \"%s\"", ret, sigstr);
+    char *index, *dup = SCStrdup(sigstr);
```
victorjulien: I wonder if we can get rid of this strdup. It seems needed to insert the \0's so that strlcpy knows what to copy. Perhaps we can copy w/o the need for the \0?

victorjulien:
"But I'd like to see this in before that is complete for the performance gains I see on lower end CPUs."

Agreed

src/detect-parse.c
jasonish: Yes, it can read the byte before the buffer. Strangely couldn't get ASAN to trip on it. I've also added optend > optstr to prevent this.

src/detect-parse.c
jasonish: Yes. SigParseToken and SigParseList now also track the length of data to be copied. The strdup has been removed.

jasonish:
Closing. See https://github.com/OISF/suricata/pull/3015

PR #3008, opened 2017-11-22
Adds options to mark when a file is final.

gauravjsingh:
This takes the form of an option to add the pid of the process to file
names. Additionally, it adds a suffix to the file name to indicate it is
not finalized.

Adding the pid to the file name reduces the likelihood that a file is
overwritten when suricata is unexpectedly killed. The number in the
waldo file is only written out during a clean shutdown. In the event
of an improper shutdown, extracted files will be written using the old
number and existing files with the same name will be overwritten.

Writes extracted files and their metadata to a temporary file suffixed
with '.tmp' and '.tmp.meta' respectivecly. Renames the files when they
are completely done being written. As-is there is no way to know that
a file on disk is still being written to by suricata.

Describe changes:
- Followup to PR #2984.
- Changes are using a suffix instead of a prefix (".tmp", and ".tmp.meta"), making it the standard behavior, and not requiring write-meta for this option.

### DIFF
--- a/doc/userguide/file-extraction/file-extraction.rst
+++ b/doc/userguide/file-extraction/file-extraction.rst
@@ -44,8 +44,10 @@ drop dir must be configured.
       waldo: file.waldo # waldo file to store the file_id across runs
       max-open-files: 0 # how many files to keep open (O means none)
       write-meta: yes   # write a .meta file if set to yes
+      working-file-prefix: . # if nonempty, and write-meta is yes, use the given prefix to indicate that the file is not finalized.
+      include-pid: yes  # include the pid in filenames if set to yes.
 
-Each file that is stored with have a name "file.<id>". The id will be reset and files will be overwritten unless the waldo option is used. A "file.<id>.meta" file is generated containing file metadata if write-meta is set to yes (default).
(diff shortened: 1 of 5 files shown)


jasonish:
> Changes are using a suffix instead of a prefix (".tmp", and ".tmp.meta"), making it the standard behavior, and not requiring write-meta for this option.

I think this should be .meta.tmp. Its easier for unix tools to work with... ls *.tmp, find -name \*.tmp, etc.

doc/userguide/file-extraction/file-extraction.rst
```diff
@@ -44,8 +44,10 @@ drop dir must be configured.
       waldo: file.waldo # waldo file to store the file_id across runs
       max-open-files: 0 # how many files to keep open (O means none)
       write-meta: yes   # write a .meta file if set to yes
+      working-file-prefix: . # if nonempty, and write-meta is yes, use the given prefix to indicate that the file is not finalized.
```
victorjulien: this seems out of sync with the code

gauravjsingh:
Addressed comments in #3019

PR #3052, opened 2017-12-05
hardening: unix suricata-command.socket permission update

macpas:

Describe changes:
So far, the suricata socket suricata-command.socket has the rights rw-r----- suricata:user. 
When suricata is used with restricted access, an other application (suricatasc like) that needs to access to the command socket also with restricted access can not write to the socket since it is not the owner (e.g suricata within container, with an hardened value for umask and hardened rights for users). 
The socket should be set as rw-rw----. Use chmod instead of fchmod and set it after the socket creation.

### DIFF
--- a/src/unix-manager.c
+++ b/src/unix-manager.c
@@ -163,19 +163,6 @@ static int UnixNew(UnixCommand * this)
     }
     this->select_max = this->socket + 1;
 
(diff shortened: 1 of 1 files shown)


regit:
The part of the PR text explaining the change is really useful and should be part of the commit message. I've done the change in PR to come.

victorjulien:
Merged through #3184, thanks!

PR #3069, opened 2017-12-09
Rotate log files based on file size

thus:
Rotate log files based on size. Similar to the time-based rotation.

Enable by adding 'rotate-size' to the logger in suricata.yaml:
```yaml
- eve-log:
    enabled: yes
    filetype: regular
    filename: eve-%s.json
    rotate-size: 50mb
```    

Also works together with time-based rotation, which makes it possible to rotate either when reaching a specific file size, or after a specified time.

https://redmine.openinfosecfoundation.org/issues/2107

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/134
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/134

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -29,6 +29,7 @@
 #include "conf.h"            /* ConfNode, etc. */
 #include "output.h"          /* DEFAULT_LOG_* */
 #include "util-byte.h"
+#include "util-misc.h"
 #include "util-logopenfile.h"
 #include "util-logopenfile-tile.h"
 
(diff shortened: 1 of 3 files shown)


thus:
Changes: rebased to be in sync with master

src/util-logopenfile.c
```diff
@@ -424,6 +438,18 @@ SCConfLogOpenGeneric(ConfNode *conf,
         }
     }
 
+    /* Rotate log file based on size */
+    const char *rotate_size = ConfNodeLookupChildValue(conf, "rotate-size");
+    if (rotate_size != NULL) {
+        log_ctx->flags |= LOGFILE_ROTATE_SIZE;
```
regit: Better to have it after the parsing of the option in case we later consider we do not need to exit in case of failure.

src/util-logopenfile.h
```diff
@@ -137,6 +137,7 @@ typedef struct LogFileCtx_ {
 #define LOGFILE_HEADER_WRITTEN  0x01
 #define LOGFILE_ALERTS_PRINTED  0x02
 #define LOGFILE_ROTATE_INTERVAL 0x04
+#define LOGFILE_ROTATE_SIZE     0x08
```
regit: Could be nice to convert the flags to  BIT_U8(x)

src/util-logopenfile.c
```diff
@@ -195,14 +196,26 @@ static int SCLogFileWrite(const char *buffer, int buffer_len, LogFileCtx *log_ct
     } else
 #endif
     {
-
```
regit: I don't like white space fix in patch changing code but who cares ;)

src/util-logopenfile.c
```diff
@@ -195,14 +196,26 @@ static int SCLogFileWrite(const char *buffer, int buffer_len, LogFileCtx *log_ct
...
+                (log_ctx->size_current + buffer_len > log_ctx->size_limit)) {
+            SCConfLogReopen(log_ctx);
+
+            /* Reset time-based rotation, if used */
+            if (log_ctx->flags & LOGFILE_ROTATE_INTERVAL) {
+                time_t now = time(NULL);
+                log_ctx->rotate_time = now + log_ctx->rotate_interval;
+            }
+        }
+
```
regit: Don't like this line break.

src/util-logopenfile.c
thus: Would it make sense to just report the error and remove the exit in case of failure in this patch instead of waiting with this for later?

After you mentioned it, I don't think this is something that should make exit Suricata.

thus:
Replaced by #3123

src/util-logopenfile.c
thus: In #3123 I tried changing it so that it doesn't set the flag if it fails instead of exiting. I also changed 'rotate-interval' so it behaves in the same way.

PR #3081, opened 2017-12-12
Unix reopen command

regit:
Add a unix socket command to trigger log files reopening.

Describe changes:
- Add the command
- Documentation

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/358
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/142

### DIFF
--- a/doc/userguide/unix-socket.rst
+++ b/doc/userguide/unix-socket.rst
@@ -67,6 +67,7 @@ The set of existing commands is the following:
 * capture-mode: display capture system used
 * conf-get: get configuration item (see example below)
 * dump-counters: dump Suricata's performance counters
+* reopen-log-files: reopen log files (to be run after external log rotation)
 * ruleset-reload-rules: reload ruleset and wait for completion
 * ruleset-reload-nonblocking: reload rulesat and proceed without waiting
 * ruleset-reload-time: return time of last reload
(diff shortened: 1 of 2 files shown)


victorjulien:
@jasonish can you have a look at this? Looks ok to me but I do wonder if there is any risk around synchronization/threading. E.g. if both unix-socket and signals code would act in parallel.

pevma:
for info - this PR solves a problem of logrotating that i have observed on live traffic boxes that i have discussed with @jasonish a while back ago.

jasonish:
What about making the "flag" value volatile? Looking at this, I think it should have been made volatile from the start. It should not require synchronization then over what volatile already provides.

inliniac:
On 13-12-17 14:10, Jason Ish wrote:
> What about making the "flag" value volatile? Looking at this, I think it
> should have been made volatile from the start. It should not require
> synchronization then over what volatile already provides.

In that case I prefer making it an atomic. Those are free except for
'sets', so should be a good fit.

But the list we're walking is static after init, right?

jasonish:
> In that case I prefer making it an atomic. Those are free except for 'sets', so should be a good fit. 

Yup, even better.

> But the list we're walking is static after init, right?

Correct.

victorjulien:
Added the atomic flag thing as a ticket: https://redmine.openinfosecfoundation.org/issues/2378

I don't think it should block this PR.

victorjulien:
Merged through #3104, thanks Eric!

PR #3121, opened 2017-12-28
Don't panic on malformed NFS traffic

DominoTree:

https://redmine.openinfosecfoundation.org/issues/2397

Describe changes:
Removed a call to the panic!() macro if an unsupported/invalid NFS version is encountered. Not sure this is the correct way to do this, but it seems to have solved the issue in my environment. Perhaps this might make more sense with a Result<> wrapper or something?

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -95,9 +95,8 @@ pub static mut SURICATA_NFS3_FILE_CONFIG: Option<&'static SuricataFileContext> =
 #[repr(u32)]
 pub enum NFSEvent {
     MalformedData = 0,
-    /* remove 'Padding' when more events are added. Rustc 1.7 won't
-     *   accept a single field enum with repr(u32) */
-    Padding,
+    NonExistingVersion = 1,
+    UnsupportedVersion = 2,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


rust/src/nfs/nfs.rs
```diff
@@ -1069,7 +1069,10 @@ impl NFSState {
                 SCLogDebug!("NFSv2 reply record");
                 return self.process_reply_record_v2(r, &xidmap);
             },
-            _ => { panic!("unsupported NFS version"); },
+            _ => {
+                SCLogNotice!("unsupported NFS version");
```
regit: Looks like having a NFSv4 message would be legit so we should ignore this value.

For weird value, using a log notice message could introduce a message storm if ever a lot of weird packets are sent. It is better to use a Debug level message and set an event instead with 
```
self.set_event(NFSEvent::MalformedData);
```
Ideally, a dedicated event like NonExistingVersion would be better.

DominoTree: Changes seem simple enough, I'll get them sorted :)

DominoTree:
Went ahead and added an NFS event for `UnsupportedVersion` as well because it seemed to make sense. If not, let me know and I'll go ahead and remove it.

victorjulien:
Merged in #3133, thanks! I did a small commit msg cleanup.

PR #3122, opened 2017-12-30
Initial JA3 integration

thus:

Features implemented:
* JSON logger for JA3.
* JA3 metadata in alerts.

The authors of JA3 is presenting on Shmoocon in late January, so it would be really cool if we could have JA3 support in Suricata by then ;)

Redmine:

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/137
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/137


src/app-layer-ssl.c
```diff
@@ -258,55 +263,214 @@ static int SSLSetTxMpmIDs(void *vtx, uint64_t mpm_ids)
...
+
+        JA3Buffer *ja3_cipher_suites = Ja3BufferInit();
+        if (ja3_cipher_suites == NULL)
+            return -1;
+
+        uint16_t processed_len = 0;
+        while (processed_len < cipher_suites_length)
+        {
+            if (!(HAS_SPACE(2)))
+                goto invalid_length;
```
victorjulien: ja3_cipher_suites memleak

src/app-layer-ssl.c
```diff
@@ -258,55 +263,214 @@ static int SSLSetTxMpmIDs(void *vtx, uint64_t mpm_ids)
...
+        SSLSetEvent(ssl_state,
+                    TLS_DECODER_EVENT_HANDSHAKE_INVALID_LENGTH);
+        return -1;
+    }
+
+    /* Skip compression methods */
+    uint8_t compression_methods_length = **input;
+    *input += 1;
+
+    *input += compression_methods_length;
```
victorjulien: should we first have a check for HAS_SPACE(compression_methods_length)?

src/app-layer-ssl.c
```diff
@@ -378,30 +543,186 @@ static int TLSDecodeHandshakeHello(SSLState *ssl_state, uint8_t *input,
...
-                memcpy(ssl_state->curr_connp->sni, input, sni_strlen - 1);
+                memcpy(ssl_state->curr_connp->sni, *input, sni_strlen - 1);
                 ssl_state->curr_connp->sni[sni_strlen-1] = 0;
 
-                input += sni_len;
+                *input += sni_len;
+                break;
+            }
+
+            case SSL_EXTENSION_ELLIPTIC_CURVES:
```
victorjulien: I thnk it would make sense to break up this function/loop as it's quite long

src/util-ja3.c
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * \internal
+ * \brief Resize buffer if it is full.
+ *
+ * \param buffer The buffer.
+ *
+ * \retval 0 on success.
+ * \retval -1 on failure.
+ */
+static int Ja3BufferResizeIfFull(JA3Buffer *buffer, unsigned len)
```
victorjulien: please use explicit type instead of just 'unsigned', so something like uint32_t

victorjulien:
Can you post examples of the tls eve addition and the ja3 eve records to this PR?

src/app-layer-ssl.c
thus: Do we need it? We are not dereferencing input as far as I can see.

src/app-layer-ssl.c
thus: Will fix.

src/app-layer-ssl.c
thus: I'll create some new functions for you :)

src/util-ja3.c
thus: Will do.

thus:
Example of JA3 fields added to TLS log:

```json
{
  "timestamp": "2017-11-28T20:52:15.413711+0100",
  "flow_id": 248041851312109,
  "pcap_cnt": 10,
  "event_type": "tls",
  "src_ip": "192.168.86.32",
  "src_port": 60596,
  "dest_ip": "192.168.31.104",
  "dest_port": 443,
  "proto": "TCP",
  "tls": {
    "session_resumed": true,
    "sni": "something.com",
    "ja3_hash": "2efb07037a97b06201ab4fe7ec0c326e",
    "ja3_str": "771,49195-49199-52393-52392-49196-49200-49162-49161-49171-49172-51-57-47-53-10,0-23-65281-10-11-35-16-5-13-21,29-23-24-25,0"
  }
}
```

Example of the same entry, but in JA3 log:

```json
{
  "timestamp": "2017-11-28T20:52:15.413711+0100",
  "flow_id": 248041851312109,
  "pcap_cnt": 10,
  "event_type": "ja3",
  "src_ip": "192.168.86.32",
  "src_port": 60596,
  "dest_ip": "192.168.31.104",
  "dest_port": 443,
  "proto": "TCP",
  "ja3": {
    "hash": "2efb07037a97b06201ab4fe7ec0c326e",
    "str": "771,49195-49199-52393-52392-49196-49200-49162-49161-49171-49172-51-57-47-53-10,0-23-65281-10-11-35-16-5-13-21,29-23-24-25,0"
  }
}
```

The last example (the 'ja3' tag) is how it shows up as meta data to an alert as well.

src/app-layer-ssl.c
victorjulien: it's moved forward for the next caller isn't it? Seems unsafe to do this w/o check?

thus: Yes, it is. I'll add a check if that makes you sleep better at night ;)

thus:
Replaced by #3141

PR #3123, opened 2017-12-31
Rotate log files based on file size

thus:
Rotate log files based on size. Similar to the time-based rotation.

Enable by adding 'rotate-size' to the logger in suricata.yaml:
```yaml
- eve-log:
    enabled: yes
    filetype: regular
    filename: eve-%s.json
    rotate-size: 50mb
```    

Also works together with time-based rotation, which makes it possible to rotate either when reaching a specific file size, or after a specified time.

https://redmine.openinfosecfoundation.org/issues/2107

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/138
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/138

### DIFF
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -243,7 +243,7 @@ modifiers.
 Rotate log file
 ~~~~~~~~~~~~~~~
 
-Eve-log can be configured to rotate based on time.
+Eve-log can be configured to rotate based on time, file size, or both combined.
 
 ::
 
@@ -251,9 +251,16 @@ Eve-log can be configured to rotate based on time.
     - eve-log:
         filename: eve-%Y-%m-%d-%H:%M.json
         rotate-interval: minute
+        rotate-size: 100mb
 
-The example above creates a new log file each minute, where the filename contains
-a timestamp. Other supported ``rotate-interval`` values are ``hour`` and ``day``.
+The example above creates a new log file either each minute or when the file
+reaches 100 mb in size. The filename contains a timestamp to avoid overwriting
+itself when rotating.
+
+If both time-based and size-based rotation is used at the same time, then the
+rotation time is reset upon rotation, even if the rotation was based on size.
+
+Other supported ``rotate-interval`` values are ``hour`` and ``day``.
 
 In addition to this, it is also possible to specify the ``rotate-interval`` as a
 relative value. One example is to rotate the log file each X seconds.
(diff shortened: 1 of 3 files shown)


thus:
Changes:
* Added the changes suggested by @regit in #3069.
* Removed the exit on failure on 'rotate-size', instead it just disables the functionality and issues an error message.
* Did the same thing to 'rotate-interval' as mentioned above.

victorjulien:
Setting this to a low limit (5mb) seems to reopen an existing file a few times:
<pre>
-rw-rw-r-- 1 victor victor 5,0M jan 23 07:29 tmp/eve-1516688980.json
-rw-rw-r-- 1 victor victor 5,0M jan 23 07:29 tmp/eve-1516688981.json
-rw-rw-r-- 1 victor victor  15M jan 23 07:29 tmp/eve-1516688982.json
-rw-rw-r-- 1 victor victor  10M jan 23 07:29 tmp/eve-1516688983.json
-rw-rw-r-- 1 victor victor  15M jan 23 07:29 tmp/eve-1516688984.json
-rw-rw-r-- 1 victor victor  10M jan 23 07:29 tmp/eve-1516688985.json
-rw-rw-r-- 1 victor victor  15M jan 23 07:29 tmp/eve-1516688986.json
-rw-rw-r-- 1 victor victor  10M jan 23 07:29 tmp/eve-1516688987.json
-rw-rw-r-- 1 victor victor 2,6M jan 23 07:29 tmp/eve-1516688988.json
</pre>
Of course 5mb is a contrived example, but I'm sure ppl in the field will hit this condition unintentionally on high speed sensors.

2 questions
1. can this lead to data loss? (mostly thinking about non-append mode here)
2. should we explicitly deal with this somehow?

thus:
1. Yes, it could lead to data loss in non-append mode. In append mode, the same file would just be opened (and appended to) several times.
2. Could we add a big fat warning in the documentation about this? People won't know of this feature without reading the docs anyway, since it's not in the default suricata.yaml.

victorjulien:
I don't like the warning suggestion, I prefer a solution that simply doesn't include potential data loss.

A higher resolution will make the issue less likely, but even then we might get into the same issue. How about something like: if a file X for this name already exists, create X-01 ? Guess this would require some state keeping to avoid trying each suffix to be tried until we find X-999. Other solutions?

thus:
I think that would be a good solution.

What about logging a warning if the user doesn't use any date modifiers in the filename when using log rotation? If this is the case, then the log files will be overwritten every time Suricata is restarted. If we should "allow" filenames without date modifiers, then we would have to have a "waldo" file for keeping track of things, which sounds like added complexity. What do you think?

Should we log a warning and tell the user to increase the value if the file rotates like this?

PR #3138, opened 2018-01-09
flowbits: analyze and dump to json

victorjulien:
Analyze flowbits to find which bits are only checked.

Track whether they are set and checked on the same level of 'statefulness'
for later used.

Dump flowbits to json including the sids that set/check etc the bit.

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/71
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/73


victorjulien:
Merged as part of #3182

PR #3141, opened 2018-01-11
Initial JA3 integration

thus:

Features implemented:
* JSON logger for JA3.
* JA3 metadata in alerts.

The authors of JA3 is presenting on Shmoocon in late January, so it would be really cool if we could have JA3 support in Suricata by then ;)

Redmine:

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/139
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/139


thus:
Updates: made changes suggested in #3122

thus:
Replaced by #3145

PR #3163, opened 2018-01-16
conf: avoid included empty files

maurizioabba:
Skip invalid empty files included in configuration file.

https://redmine.openinfosecfoundation.org/issues/2418

Describe changes:
- skip nodes from "include" if the file parsed is empty
-
-

### DIFF
--- a/src/conf-yaml-loader.c
+++ b/src/conf-yaml-loader.c
@@ -250,6 +250,9 @@ ConfYamlParse(yaml_parser_t *parser, ConfNode *parent, int inseq)
                         state = CONF_INCLUDE;
                         goto next;
                     }
+                    if (value[0] == '\0') {
+                        goto next;
+                    }
 
                     if (parent->is_seq) {
                         if (parent->val == NULL) {


jasonish:
While this fixes the issue, I'd rather fix the state transitions that allows this happen. With some debugging enabled, I found it interested that libyaml gives us a scalar event for an empty file, and we treat the first scalar event as a key. Instead I'd rather check if the key is non-empty, before moving into the next state. I've come up with:
```
diff --git a/src/conf-yaml-loader.c b/src/conf-yaml-loader.c
index c04ce1b1e..5379c7b70 100644
--- a/src/conf-yaml-loader.c
+++ b/src/conf-yaml-loader.c
@@ -206,6 +206,14 @@ ConfYamlParse(yaml_parser_t *parser, ConfNode *parent, int inseq)
             char *tag = (char *)event.data.scalar.tag;
             SCLogDebug("event.type=YAML_SCALAR_EVENT; state=%d; value=%s; "
                 "tag=%s; inseq=%d", state, value, tag, inseq);
+
+            /* Skip over empty scalar values while in KEY state. This
+             * tends to only happen on an empty file, where a scalar
+             * event probably shouldn't fire anyways. */
+            if (state == CONF_KEY && strlen(value) == 0) {
+                goto next;
+            }
+
             if (inseq) {
                 char sequence_node_name[DEFAULT_NAME_LEN];
                 snprintf(sequence_node_name, DEFAULT_NAME_LEN, "%d", seq_idx++);
```

which I will recreate a PR for. But thank you for bringing this to our attention. You are right, we don't want this to happen.

PR #3165, opened 2018-01-16

maurizioabba:

Describe changes:
- Increase Decoder function signature to accept pkt len of size uint32_t
-
-

### DIFF
--- a/src/decode-ipv4.c
+++ b/src/decode-ipv4.c
@@ -485,7 +485,7 @@ static int DecodeIPV4Options(Packet *p, uint8_t *pkt, uint16_t len, IPV4Options 
     return 0;
 }
 
-static int DecodeIPV4Packet(Packet *p, uint8_t *pkt, uint16_t len)
+static int DecodeIPV4Packet(Packet *p, uint8_t *pkt, uint32_t len)
 {
     if (unlikely(len < IPV4_HEADER_LEN)) {
         ENGINE_SET_INVALID_EVENT(p, IPV4_PKT_TOO_SMALL);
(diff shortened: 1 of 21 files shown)


src/decode-ipv4.c
```diff
@@ -485,7 +485,7 @@ static int DecodeIPV4Options(Packet *p, uint8_t *pkt, uint16_t len, IPV4Options
     return 0;
 }
 
-static int DecodeIPV4Packet(Packet *p, uint8_t *pkt, uint16_t len)
+static int DecodeIPV4Packet(Packet *p, uint8_t *pkt, uint32_t len)
```
victorjulien: so in what condition could an IP packet be bigger than 64k?

src/decode-tcp.c
```diff
@@ -47,12 +47,12 @@
     (dst).len  = (src).len; \
     (dst).data = (src).data
 
-static int DecodeTCPOptions(Packet *p, uint8_t *pkt, uint16_t len)
+static int DecodeTCPOptions(Packet *p, uint8_t *pkt, uint32_t len)
```
victorjulien: not sure I understand why TCP options could be >64k?

src/decode-ipv4.c
maurizioabba: You are perfectly right, there is no possibility for an IP packet to be bigger than 64k. However, the variable passed is a uint32 (from DecodeEthernet). I modified these two decoders (both IP and TCP) to get to a standard among decoders. Of course, if you prefer to restrict the change only where is needed, I will just remove the layer 3 and layer 4 - related modification.

src/decode-tcp.c
maurizioabba: Again, just for standard-enforcing. I'm ok with reversing the change to uint16

maurizioabba:
Replaced by https://github.com/OISF/suricata/pull/3171

PR #3166, opened 2018-01-16
signal: use centralized pthread_sigmask for signals

maurizioabba:
according to its man page, sigprocmask has undefined behavior in
multithreaded environments. Instead of explictly blocking the handling
of SIGUSR2 in every thread, direct block handling SIGUSR2 before
creating the threads and enable again the handling of this signal
afterwards. In this way, only the main thread will be able to manage
this signal properly.

https://redmine.openinfosecfoundation.org/issues/2420

Describe changes:
- use pthread_sigmask instead of sigprocmask
- remove blocking signal handling in the thread spawn
- add signal handling block before spawning threads
- regain signal handling from main thread after spawning the various threads

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2146,7 +2146,6 @@ static int InitSignalHandler(SCInstance *suri)
     UtilSignalHandlerSetup(SIGINT, SignalHandlerSigint);
     UtilSignalHandlerSetup(SIGTERM, SignalHandlerSigterm);
 #ifndef OS_WIN32
-    UtilSignalHandlerSetup(SIGUSR2, SignalHandlerSigusr2);
     UtilSignalHandlerSetup(SIGHUP, SignalHandlerSigHup);
     UtilSignalHandlerSetup(SIGPIPE, SIG_IGN);
     UtilSignalHandlerSetup(SIGSYS, SIG_IGN);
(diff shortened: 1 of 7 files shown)


src/suricata.c
```diff
@@ -2778,6 +2779,16 @@ int main(int argc, char **argv)
 
     (void)SCSetThreadName("Suricata-Main");
 
+    /* Ignore SIGUSR2 as early as possble. We redeclare interest
+     * once we're done launching threads. The goal is to either die
+     * completely or handle any and all SIGUSR2s correctly.
+     */
+    UtilSignalHandlerSetup(SIGUSR2, SIG_IGN);
```
victorjulien: this fails on mingw (windows / #ifdef OS_WIN32). It might be nicer to create a little wrapper 'UtilSignalSetupUSR2' or something, that handles the #ifdef'ing internally.

victorjulien: btw it fails because SIGUSR2 is undefined

maurizioabba: Out of curiosity, is there any idea for performing rule reload under windows? I don't have a Win machine I can currently test on, but I think there is no rule reload there, am I correct?

victorjulien: there are 2 levels of windows support currently. Cygwin has USR2 support and reloads too (iirc). The newer MinGW based code (OS_WIN32 define) does not have USR2 (and unix socket) and has the rule reload code ifdef'd out

maurizioabba: In the end, I'd prefer not to create a wrapper, we are already dealing with a wrapper here and it may be a little bit overwhelming. However, I correct the patch. Will create a PR shortly

maurizioabba:
Substitute by https://github.com/OISF/suricata/pull/3167

PR #3167, opened 2018-01-17
signal: use centralized pthread_sigmask for signals

maurizioabba:
according to its man page, sigprocmask has undefined behavior in
multithreaded environments. Instead of explictly blocking the handling
of SIGUSR2 in every thread, direct block handling SIGUSR2 before
creating the threads and enable again the handling of this signal
afterwards. In this way, only the main thread will be able to manage
this signal properly.

https://redmine.openinfosecfoundation.org/issues/2420

Describe changes:
- use pthread_sigmask instead of sigprocmask
- remove blocking signal handling in the thread spawn
- add signal handling block before spawning threads
- regain signal handling from main thread after spawning the various threads

### DIFF
--- a/src/counters.c
+++ b/src/counters.c
@@ -323,10 +323,6 @@ static void StatsReleaseCtx(void)
  */
 static void *StatsMgmtThread(void *arg)
 {
-#ifndef OS_WIN32
-    /* block usr2.  usr2 to be handled by the main thread only */
-    UtilSignalBlock(SIGUSR2);
-#endif
     ThreadVars *tv_local = (ThreadVars *)arg;
     uint8_t run = 1;
     struct timespec cond_time;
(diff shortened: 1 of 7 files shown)


victorjulien:
Merged in #3183, thanks!

victorjulien:
So I'm just noticing now that this actually breaks USR2 for me. Any idea why my reloads don't work anymore @maurizioabba ?

maurizioabba:
That's really weird. Looks fine from my side. what behavior are you experiencing? Are we speaking about unix system?

victorjulien:
Looks like the patch is not unblocking usr2 in the 'delayed detect' case. It's interesting that it worked before, as it seems we didn't actually set up usr2 correctly for that case.

maurizioabba:
Sure, I see the issue in the delayed_detect mode. At line 2513, we explicitly set up the SignalHandler setup only if delayed_detect == 0.
A possible solution would be to add
UtilSignalHandlerSetup(SIGUSR2, SignalHandlerSigusr2);
After line 2518.
In this way, SIGUSR2 will be managed only after the first DetectEngineReload.
On my tests, this will work.

Shall I open a new PR? A new ticket on redmine? What should be the appropriate procedure here?

PR #3172, opened 2018-01-17

maurizioabba:

Note that after v1 L3 and L4 decoder length increases have been removed.

Describe changes:
- Increase Decoder function signature to accept pkt len of size uint32_t
-
-

### DIFF
--- a/src/decode-ethernet.c
+++ b/src/decode-ethernet.c
@@ -39,7 +39,7 @@
 #include "util-debug.h"
 
 int DecodeEthernet(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
-                   uint8_t *pkt, uint16_t len, PacketQueue *pq)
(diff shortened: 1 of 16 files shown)


src/decode-ethernet.c
```diff
@@ -39,7 +39,7 @@
 #include "util-debug.h"
 
 int DecodeEthernet(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
-                   uint8_t *pkt, uint16_t len, PacketQueue *pq)
+                   uint8_t *pkt, uint32_t len, PacketQueue *pq)
```
victorjulien: since we're passing length as "len - ETHERNET_HEADER_LEN" to the DecodeIPV4 func, it seems we can now get into trouble by sending > USHRT_MAX. I think this will lead to a int overflow / wrap around. Think we need explicit checks to avoid this case?

PR #3175, opened 2018-01-18
filestore v2 - v4

jasonish:

Previous PR:
https://github.com/OISF/suricata/pull/3139

https://redmine.openinfosecfoundation.org/issues/2303

This PR introduces a new version of the filestore output to address the issues with the current one.

New directory layout:
```
.../filestore/00/
.../filestore/01/
...
.../filestore/ff/
.../filestore/tmp/
```

And files are named by their SHA256. Files in progress are stored in the tmp/ directory, then when completed are renamed into the directory that shares the first 2 characters of its SHA256 sum. If the file already exists its timestamps are updated. This layout provides depuplication of saved files and eliminates the need for "waldo" file.

As SHA256 is required, this module require libnss, and it will force SHA256 checksums. An additional change was made to even force the SHA256 if the file is not in a complete state, as we need this for the final filename.

Metadata is disabled by default as we already have the `fileinfo` records in the main eve-log. If `write-fileinfo` is enabled, then a `fileinfo` file will be written for each occurrence of the file. These files are named <SHA256>.<TV.TV_SEC>.<FILE_ID>.json. This is unique enough that each occurrence of the file will get its own fileinfo file, but subsequent runs over the same pcap should collide causing the file to be overwritten.

A new tool is also installed (provided Python is installed) name `suricatactl` provides a command for pruning the filestore. The idea is that more miscellaneous commands will be added over time.  Example usage:
```
suricatactl filestore prune --age 7d
```
which will remove all files from the filestore over 7 days old.

Known issues:
- The new filestore output is named output-filestore.c, which started as an exact copy of log-filestore.c then modified as required. Both filestores have the same configuration name and use the "version" field to determine if they should be enabled or not. Currently the output code handles a NULL returned from an output init function as an error, which is what the filestore init functions return if for a different version. I need a better way to handle this. They could simply have different names I suppose, like "file-store" and "file-store2" instead of adding a version field. Changing the configuration name would obviously change the filenames. 
- The idea without output-filestore.c was to start using a consistent naming scheme for loggers, as we currently use log-, output- and alert-. But then we also use output- for output support files that are not individual loggers.
- Nothing is done about files in tmp/ that may be left hanging due to an abnormal exit. I'm thinking of simply removing all existing files in tmp/ on startup.

Changes from last PR:
- Remove exit stats.
- Replace find -delete with find -print0 | xargs -0 rm -f
- Move #include utime.h into suricata-common.h.
- Allow the caller to the file info build to decide if the stored flag should be true or not.
- Removed exit stats.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/253
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/606

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,7 +5,7 @@ ACLOCAL_AMFLAGS = -I m4
 EXTRA_DIST = ChangeLog COPYING LICENSE suricata.yaml.in \
              classification.config threshold.config \
              reference.config
-SUBDIRS = $(HTP_DIR) rust src qa rules doc contrib scripts etc
+SUBDIRS = $(HTP_DIR) rust src qa rules doc contrib scripts etc python
 
 CLEANFILES = stamp-h[0-9]*
 
--- a/configure.ac
+++ b/configure.ac
@@ -2125,7 +2125,7 @@ AC_SUBST(CONFIGURE_SYSCONDIR)
 AC_SUBST(CONFIGURE_LOCALSTATEDIR)
 AC_SUBST(PACKAGE_VERSION)
 
-AC_OUTPUT(Makefile src/Makefile rust/Makefile rust/Cargo.toml rust/.cargo/config qa/Makefile qa/coccinelle/Makefile rules/Makefile doc/Makefile doc/userguide/Makefile contrib/Makefile contrib/file_processor/Makefile contrib/file_processor/Action/Makefile contrib/file_processor/Processor/Makefile contrib/tile_pcie_logd/Makefile suricata.yaml scripts/Makefile scripts/suricatasc/Makefile scripts/suricatasc/suricatasc etc/Makefile etc/suricata.logrotate etc/suricata.service)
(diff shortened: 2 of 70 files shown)


victorjulien:
Seeing test failures on docker based builders:
===> filestore-v2.3-fserror: FAIL: stats.file_store.fs_errors: expected 1; got 0

Maybe caused by how docker runs? Maybe in the container the user is root?

jasonish:
> Maybe caused by how docker runs? Maybe in the container the user is root?

Thats probably it. It sets up directory permissions to fail, but probably won't fail for root. Maybe suricata-verify needs some more skip condition checks

victorjulien:
my fedora25 docker builder has another error I didn't see before:
===> dns-udp-dns-log-unanswered: expected 4 responses, found 2
FAILED: verification failed

victorjulien:
Hmm I restarted the job and now the same test passes.

KingJiJi:
hello ,I have a problem：
i find suricata file-store work well at  first few minutes . it can find and store 100% files.
But after a few minutes,it will only find and store 1% or less files.

My version  Suricata version 5.0.0
Thank you.

victorjulien:
> hello ,I have a problem：
> i find suricata file-store work well at first few minutes . it can find and store 100% files.
> But after a few minutes,it will only find and store 1% or less files.
> 
> My version Suricata version 5.0.0
> Thank you.

Support requests and bug reports should go here https://redmine.openinfosecfoundation.org/projects/suricata

XCM-jj:
> > hello ,I have a problem：
> > i find suricata file-store work well at first few minutes . it can find and store 100% files.
> > But after a few minutes,it will only find and store 1% or less files.
> > My version Suricata version 5.0.0
> > Thank you.
> 
> Support requests and bug reports should go here https://redmine.openinfosecfoundation.org/projects/suricata

ok.think you

PR #3178, opened 2018-01-21
Optimization: reset bidirectional flag for rules with same src and dst

gozzy:

Describe changes:
If one created bidirectional rule like 'alert tcp any any <> any any ...', the rule was checked twice (for each packet in every direction(. This is suboptimal and may give duplicated alerts. To avoid this, bidirectional rules are now checked for the same src and dst (addresses and ports) and if it's the case, the rule is treated as unidirectional and a corresponding message is logged.

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -1797,6 +1797,51 @@ error:
 }
 
 /**
+ * \brief Checks if a signature has the same source and destination
+ * \param s parsed signature
+ *
+ *  \retval 1 if source and destination are the sane, 0 otherwise
+ */
+static int SigHasSameSourceAndDestination(const Signature *s)
+{
+    int ret = 0;
+
+    if (!(s->flags & SIG_FLAG_SP_ANY) || !(s->flags & SIG_FLAG_DP_ANY)) {
+        if (DetectPortCmp(s->sp, s->dp) != 1) {
+            goto end;
+        }
+    }
+
+    if (!(s->flags & SIG_FLAG_SRC_ANY) || !(s->flags & SIG_FLAG_DST_ANY)) {
+        DetectAddress *src_v4 = s->init_data->src->ipv4_head;
+        DetectAddress *dst_v4 = s->init_data->dst->ipv4_head;
+        DetectAddress *src_v6 = s->init_data->src->ipv6_head;
+        DetectAddress *dst_v6 = s->init_data->dst->ipv6_head;
+
+        if (src_v4 != NULL && dst_v4 != NULL) {
+            if (DetectAddressCmp(src_v4, dst_v4) != ADDRESS_EQ) {
+                goto end;
+            }
+        } else if (!(src_v4 == NULL && dst_v4 == NULL)) {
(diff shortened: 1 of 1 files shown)


gozzy:
I don't think that initial behavior was a bug, therefore I've not created a ticked. However I suppose that it slows down packet processing and could be optimized a little bit.

gozzy:
Any thoughts on this?

src/detect-parse.c
```diff
@@ -1796,6 +1796,51 @@ static Signature *SigInitHelper(DetectEngineCtx *de_ctx, const char *sigstr,
...
+    }
+
+    if (!(s->flags & SIG_FLAG_SRC_ANY) || !(s->flags & SIG_FLAG_DST_ANY)) {
+        DetectAddress *src_v4 = s->init_data->src->ipv4_head;
+        DetectAddress *dst_v4 = s->init_data->dst->ipv4_head;
+        DetectAddress *src_v6 = s->init_data->src->ipv6_head;
+        DetectAddress *dst_v6 = s->init_data->dst->ipv6_head;
+
+        if (src_v4 != NULL && dst_v4 != NULL) {
+            if (DetectAddressCmp(src_v4, dst_v4) != ADDRESS_EQ) {
```
victorjulien: both src and dst can be a list of DetectAddress objects, but DetectAddressCmp doesn't compare the list but just a single object.

src/detect-parse.c
```diff
@@ -1796,6 +1796,51 @@ static Signature *SigInitHelper(DetectEngineCtx *de_ctx, const char *sigstr,
     return NULL;
 }
 
+/**
+ * \brief Checks if a signature has the same source and destination
+ * \param s parsed signature
+ *
+ *  \retval 1 if source and destination are the sane, 0 otherwise
+ */
+static int SigHasSameSourceAndDestination(const Signature *s)
```
victorjulien: 'static bool'

src/detect-parse.c
```diff
@@ -1796,6 +1796,51 @@ static Signature *SigInitHelper(DetectEngineCtx *de_ctx, const char *sigstr,
...
+{
+    int ret = 0;
+
+    if (!(s->flags & SIG_FLAG_SP_ANY) || !(s->flags & SIG_FLAG_DP_ANY)) {
+        if (DetectPortCmp(s->sp, s->dp) != 1) {
```
victorjulien: same list issue here

src/detect-parse.c
```diff
@@ -1796,6 +1796,51 @@ static Signature *SigInitHelper(DetectEngineCtx *de_ctx, const char *sigstr,
...
+                goto end;
+            }
+        } else if (!(src_v6 == NULL && dst_v6 == NULL)) {
+            goto end;
+        }
+    }
+
+    ret = 1;
+
+end:
```
victorjulien: no need to use a label and goto's, just return false from each check

gozzy:
Replaced with #3231

PR #3181, opened 2018-01-22
Filestore stream-depth fixes

glongo:

https://redmine.openinfosecfoundation.org/issues/2264

Describe changes:
Currently if file-store.stream-depth is set, it doesn't behave correctly.
If its value is zero, it works as expected, but if the value is fixed it won't be used.
Even if stream-depth is enabled in an application layer, it doesn't work because the value is overwritten.
This patchset fixes the usage of stream-depth when:
- global file-store.stream-depth is enabled
- stream-depth is enabled for an application layer
- file-store keyword is used with http protocol

- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/169
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/33

### DIFF
--- a/doc/userguide/file-extraction/file-extraction.rst
+++ b/doc/userguide/file-extraction/file-extraction.rst
@@ -63,6 +63,15 @@ of the filename. For example, if the SHA256 hex string of an extracted
 file starts with "f9bc6d..." the file we be placed in the directory
 `filestore/f9`.
 
+The size of a file that can be stored depends on ``file-store.stream-depth``,
+if this value is reached a file can be truncated and not stored completely.
+If not enabled, ``strem.reassembly.depth`` will be considered.
+
+Setting ``file-store.stream-depth`` to 0 permits to store any files.
+
+A protocol parser, like modbus, could permit to set a different
+store-depth value and use it rather than ``file-store.stream-depth``.
+
 Using the SHA256 for file names allows for automatic de-duplication of
 extracted files. However, the timestamp of a pre-existing file will be
 updated if the same files is extracted again, similar to the `touch`
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -473,6 +473,14 @@ void AppLayerHtpNeedFileInspection(void)
     SCReturn;
 }
 
+void AppLayerHtpUseStreamDepth(htp_tx_t *tx)
+{
+    HtpTxUserData *tx_ud = (HtpTxUserData *) htp_tx_get_user_data(tx);
+    if (tx_ud) {
+        tx_ud->use_stream_depth = TRUE;
+    }
+}
+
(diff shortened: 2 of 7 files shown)


doc/userguide/file-extraction/file-extraction.rst
```diff
@@ -63,6 +63,15 @@ of the filename. For example, if the SHA256 hex string of an extracted
 file starts with "f9bc6d..." the file we be placed in the directory
 `filestore/f9`.
 
+The size of a file that can be stored depends on ``file-store.stream-depth``,
+if this value is reached a file can be truncated and not stored completely.
+If not enabled, ``strem.reassembly.depth`` will be considered.
```
norg: stre*a*m.reassembly

glongo:
I'll send a new PR soon.

victorjulien:
I don't really understand the http part here, and I wonder if this could be generalized to not be http specific?

glongo:
Although file-store.stream-depth is enabled, without the http part a request/response body is not reassembled when it should be because the stream-depth value is set.
This happens because while reassembly only the body limit specified in libhtp settings is considered, but
 the file-store.stream-depth is excluded.
As a result, it could happen that a file is truncated but it should be stored instead.
This below in an example:

Config:
> outputs.14.file-store.enabled = yes
outputs.14.file-store.stream-depth = 0
stream.reassembly.depth = 1mb

Rule:
> alert http any any -> any any (msg:"FILE store all"; filestore; sid:1; rev:1;)

In this rule the filestore keyword is used, so it means that file-store.stream-depth value should be unlimited

> $ wget http://d.7-zip.org/a/7z1701-x64.exe
> $ cat var/log/suricata/files/file.1.meta

> TIME:              02/08/2018-17:19:49.681144
SRC IP:            178.62.126.102
DST IP:            192.168.1.3
PROTO:             6
SRC PORT:          80
DST PORT:          60292
APP PROTO:         http
HTTP URI:          /a/7z1701-x64.exe
HTTP HOST:         d.7-zip.org
HTTP REFERER:      <unknown>
HTTP USER AGENT:   Wget/1.19.2 (linux-gnu)
FILENAME:          /a/7z1701-x64.exe
MAGIC:             <unknown>
STATE:             TRUNCATED
SIZE:              103988

As you can see the file is truncated.

The http part doesn't need to be generalized because this issue happen only in this case (http proto), in other case like smtp, it works correctly.

victorjulien:
Isn't that because the http body limits are hit instead of the stream depth?

glongo:
Yes, the body limits are hit and the stream depth not

victorjulien:
So it seems what we need is a clean way to update those body limits. Right now they are either 'unlimited' or use a fixed value.

regit:
@victorjulien Not sure we got what you are saying. The proposed code is using a dynamic approach: if filestore is used on a TX, then we set a flag on the TX that is used to set body limit to the value of file reassembly. This is the same approach as the one on stream but applied to the protocol.

PR #3208, opened 2018-02-01
suricata-update: bundle suricata update

jasonish:
Add autoconf/automake support for installing suricata-update
if found in the top level suricata-update.

Requires this PR to suricata-update:
https://github.com/OISF/suricata-update/pull/28

There are still some things we can do better, but are outside the simple scope of bundling. Especially when a --prefix is used, it would be nice for suricata-update to default to the same locations as Suricata. I think this could be done by looking at the --build-info of Suricata itself. First check if a suricata programs exists in the same directory as suricata-update (the case when bundling), then fall back to suricata on the path. Of course --suricata would always take precedence. These changes would be done in suricata-update.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/266
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/619

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ src/suricata
 stamp-h1
 src/build-info.h
 qa/log/
+
+/suricata-update/*
+!/suricata-update/Makefile.am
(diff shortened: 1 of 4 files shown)


victorjulien:
Merged in #3297, thanks Jason!

PR #3210, opened 2018-02-01
Initial JA3 integration

thus:

Features implemented:
* JSON logger for JA3.
* JA3 metadata in alerts.

Redmine:

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/143
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/143


thus:
Changes:
* Fixed another compiler warning..
* Rebased with master

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -254,6 +254,8 @@ In addition to this, custom logging also allows the following fields:
 
 * "certificate": The TLS certificate base64 encoded
 * "chain": The entire TLS certificate chain base64 encoded
+* "ja3_hash": The JA3 hash, which is an md5sum of the JA3 string
```
norg: a m5sum instead of an md5sum?

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -319,3 +321,21 @@ Example of TFTP logging:
       "file": "rfc1350.txt",
       "mode": "octet"
    }
+
+Event type: JA3
+---------------
+
+* "hash": The JA3 hash, which is an md5sum of the JA3 string
```
norg: a m5sum instead of an md5sum?

doc/userguide/output/eve/eve-json-output.rst
```diff
@@ -107,13 +107,27 @@ YAML::
...
 
 The default is to log certificate subject and issuer. If ``extended`` is
 enabled, then the log gets more verbose.
 
 By using ``custom`` it is possible to select which TLS fields to log.
 
+JA3
+~~~
+
+JA3 fingerprints are logged one record per TLS session.
```
norg: on record? or did I misread this sentence?

norg:
is there a specific reason why md5sum is used instead of newer sha2?

doc/userguide/output/eve/eve-json-format.rst
thus: What is a 'm5sum'? Isn't 'md5sum' correct here?

doc/userguide/output/eve/eve-json-format.rst
thus: Same as above :)

doc/userguide/output/eve/eve-json-output.rst
thus: I was thinking it as one JSON record per TLS session. Should it be changed to something else?

thus:
The reason why md5 was chosen, is because that's what the authors of JA3 has used for their implementation. I have no idea why they chose md5 over other (newer) hash algorithms, I just thought it was best to use the same since this PR implements support for JA3 :)

victorjulien:
As discussed on chat, this has some issues that need to be addressed.

thus:
Replaced by #3215

PR #3213, opened 2018-02-02
app-layer: add tx iterator API

victorjulien:

    [..........TTTT]

    [.T..T...TTTT.T]

caller must keep.

behaviour will be used.

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/84
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/86


victorjulien:
Replaced by #3214

PR #3214, opened 2018-02-05
Optimization/tx iterator/v2

victorjulien:

    [..........TTTT]

    [.T..T...TTTT.T]

caller must keep.

behaviour will be used.

Changes since #3213:
- add NFS iterator implementation

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/85
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/87

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -423,6 +423,29 @@ impl NFSState {
         return None;
     }
 
+    // for use with the C API call StateGetTxIterator
+    pub fn get_tx_iterator(&mut self, min_tx_id: u64,
+            ret_tx_id: &mut u64, state: &mut u64) ->
+        Option<&NFSTransaction>
+    {
+        let mut index = *state as usize;
+        let len = self.transactions.len();
+
+        // find tx that is >= min_tx_id
+        while index < len {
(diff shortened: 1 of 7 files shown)


jasonish:
Quickly adapting UDP/DNS looks like: https://gist.github.com/jasonish/c84653483ebeea419f7562e0389f95c0

Look about right to you?

jasonish:
When the actual transaction object is used as an iterator we run into a problem in AppLayerParserTransactionCleanup, as we may free the transaction currently used to maintain the state. This will be a common use case in the C app-layer parsers where the transaction object holds the pointer to the next one.

If the state object holds the next transaction, we lack the context of this being the first call, or a subsequent call to the iterator. Which is probably why in languages that have iterators we see the creation, then a call to next. So its almost like we need one more context variable to know if its the first call or not, so we can choose how to define the initial state.

One attempt at this is: https://gist.github.com/jasonish/80ceedc2260cc05cbddf1c64a05a1074, but its not generic. Each app-layer has to it itself, I'd prefer a more generic approach.

victorjulien:
Replaced by #3220

PR #3215, opened 2018-02-05
Initial JA3 integration

thus:

Features implemented:
* JSON logger for JA3.
* JA3 metadata in alerts.

Redmine:

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/144
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/144


thus:
Changes:
* Fixed double free

src/util-error.h
```diff
@@ -337,8 +337,10 @@ typedef enum {
     SC_ERR_PF_RING_VLAN,
     SC_ERR_CREATE_DIRECTORY,
     SC_WARN_FLOWBIT,
-
     SC_ERR_MAX,
+    SC_WARN_NO_JA3_SUPPORT,
```
victorjulien: place before max

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+ * \brief Allocate new buffer.
+ *
+ * \return pointer to buffer on success.
+ * \return NULL on failure.
+ */
+JA3Buffer *Ja3BufferInit(void)
+{
+    JA3Buffer *buffer = SCCalloc(1, sizeof(JA3Buffer));
+    if (buffer == NULL) {
+        SCLogError(SC_ERR_MEM_ALLOC, "Error allocating memory for JA3 buffer");
```
victorjulien: as this is called at runtime I prefer no error message

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * \brief Free allocated buffer.
+ *
+ * \param buffer The buffer to free.
+ */
+void Ja3BufferFree(JA3Buffer *buffer)
+{
+    if (buffer == NULL) {
+        SCLogError(SC_ERR_INVALID_ARGUMENT, "Buffer should not be NULL");
```
victorjulien: this can only be a code error, right? Then you can use DEBUG_VALIDATE_BUG_ON (include util-validate.h)

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+ * \param buffer The buffer.
+ * \param len    The length of the data that should fit into the buffer.
+ *
+ * \retval 0 on success.
+ * \retval -1 on failure.
+ */
+static int Ja3BufferResizeIfFull(JA3Buffer *buffer, uint32_t len)
+{
+    if (buffer == NULL) {
+        SCLogError(SC_ERR_INVALID_ARGUMENT, "Buffer should not be empty");
```
victorjulien: same here

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+{
+    if (buffer1 == NULL || buffer2 == NULL) {
+        SCLogError(SC_ERR_INVALID_ARGUMENT, "Buffers should not be NULL");
+        return -1;
+    }
+
+    /* If buffer1 contains no data, then we just copy the second buffer
+       instead of appending its data. */
+    if (buffer1->data == NULL) {
+        Ja3BufferFree(buffer1);
```
victorjulien: buffer1 is freed here

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+        *buffer1 = *buffer2;
```
victorjulien: then here it's written to, that looks like use after free

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * \internal
+ * \brief Return number of digits in number.
+ *
+ * \param num The number.
+ *
+ * \return digits Number of digits.
+ */
+static unsigned NumberOfDigits(unsigned num)
```
victorjulien: please use explicit types

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+/**
+ * \brief Add value to buffer.
+ *
+ * \param buffer The buffer.
+ * \param value  The value.
+ *
+ * \retval 0 on success.
+ * \retval -1 on failure.
+ */
+int Ja3BufferAddValue(JA3Buffer *buffer, int value)
```
victorjulien: this function takes the buffer in, then on various error conditions frees it. This leaves it to the called to make sure there is no dangling ptr. I'm not seeing a clean handling of this in the callers, other than returning error codes there.

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+        if (buffer->data == NULL) {
+            SCLogError(SC_ERR_MEM_ALLOC,
+                       "Error allocating memory for JA3 data");
+            Ja3BufferFree(buffer);
+            return -1;
+        }
+        buffer->size = JA3_BUFFER_INITIAL_SIZE;
+    }
+
+    unsigned value_len = NumberOfDigits(value);
```
victorjulien: explicit type

src/app-layer-ssl.c
```diff
@@ -265,55 +270,390 @@ static void SSLSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags)
...
+                                            const uint8_t *initial_input)
+{
+    if (!(HAS_SPACE(1))) {
+        SCLogDebug("TLS handshake invalid length");
+        SSLSetEvent(ssl_state,
+                    TLS_DECODER_EVENT_HANDSHAKE_INVALID_LENGTH);
+        return -1;
+    }
+
+    uint8_t session_id_length = **input;
```
victorjulien: I'm not too excited about this ptr-to-ptr use. This makes already fragile code even harder to track. It might be better to create a new data type that we pass around containing the ptr and size. Plus some helper funcs to safely read mem.

src/app-layer-ssl.c
```diff
@@ -265,55 +270,390 @@ static void SSLSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags)
...
+
+    return 0;
+}
+
+static inline int TLSDecodeHSHelloRandom(SSLState *ssl_state, uint8_t **input,
+                                         const uint32_t input_len,
+                                         const uint8_t *initial_input)
+{
+    /* Skip random */
+    *input += SSLV3_CLIENT_HELLO_RANDOM_LEN;
```
victorjulien: it's unclear why some of these funcs do check with 'HAS_SPACE' and others don't

src/util-ja3.c
```diff
@@ -0,0 +1,290 @@
...
+ *
+ * \retval 1 if disabled.
+ * \retval 0 otherwise.
+ */
+int Ja3IsDisabled(const char *type)
+{
+    int is_enabled = 0;
+
+    /* Check if JA3 is enabled */
+    ConfGetBool("app-layer.protocols.tls.ja3-fingerprints", &is_enabled);
```
pevma: As per our discussion with  @thus  - we ought to make that configure option more visible - aka add it in yaml/doc

src/output-json-ja3.c
```diff
@@ -0,0 +1,339 @@
...
+        return 0;
+    }
+
+    if (ja3_ctx->include_metadata) {
+        JsonAddMetadata(p, f, js);
+    }
+
+    json_t *tjs = json_object();
+    if (tjs == NULL) {
+        free(js);
```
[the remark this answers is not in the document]
thus: Will fix.

src/util-error.h
thus: Bah, must have screwed things up when rebasing. Will fix.

src/util-ja3.c
thus: Okay, I'll remove the error message here :)

src/util-ja3.c
thus: That's right. This shouldn't happen unless a bug is introduced. I'll change it to what you suggested.

src/util-ja3.c
thus: I'll fix.

src/util-ja3.c
thus: I'll change it :)

src/util-ja3.c
thus: Your wish is my command!

src/app-layer-ssl.c
thus: I have a plan for fixing this. Don't know if you will approve, but at least you get to close another PR if you don't ;)

src/app-layer-ssl.c
thus: I'm using HAS_SPACE before dereferencing variables, but it would probably be better to use it every time a pointer is moved. I'll fix.

src/util-ja3.c
thus: I'll add a notice in the docs and add it to the yaml.

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -254,6 +254,8 @@ In addition to this, custom logging also allows the following fields:
 
 * "certificate": The TLS certificate base64 encoded
 * "chain": The entire TLS certificate chain base64 encoded
+* "ja3_hash": The JA3 hash, which is an md5sum of the JA3 string
```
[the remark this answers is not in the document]
thus: Haha, now your previous comment makes sense! I was wondering what a m5sum was :)

Google looks like it thinks you are right:
- 83700 hits for "an md5sum"
- 235000 hits for "a md5sum"

I'll correct it :)

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -319,3 +321,21 @@ Example of TFTP logging:
       "file": "rfc1350.txt",
       "mode": "octet"
    }
+
+Event type: JA3
+---------------
+
+* "hash": The JA3 hash, which is an md5sum of the JA3 string
```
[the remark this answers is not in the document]
thus: I'll correct it :)

doc/userguide/output/eve/eve-json-output.rst
```diff
@@ -107,13 +107,27 @@ YAML::
...
 
 The default is to log certificate subject and issuer. If ``extended`` is
 enabled, then the log gets more verbose.
 
 By using ``custom`` it is possible to select which TLS fields to log.
 
+JA3
+~~~
+
+JA3 fingerprints are logged one record per TLS session.
```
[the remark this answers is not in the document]
thus: Your suggestion is better. I'll correct it.

src/util-ja3.c
thus: This is by design, buffer1 is set to the value of buffer2 if buffer1 is empty, instead of copying buffer2. Should this be changed?

src/util-ja3.c
thus: Would it be more clear what's happening if the 'Ja3Buffer' is freed in the callers instead of in 'Ja3BufferAddValue'? This should really only happen if we are out of memory (I think).

thus:
I'll try my best to send you a new PR tomorrow! :)

src/util-ja3.c
victorjulien: Yes. The memory at 'buffer1' is freed. The line *buffer1 = *buffer2 copies the content of buffer2 (so the data at this memory location, not the pointer address itself) to the memory buffer1 points to. But this was just freed. So this is use after free.

thus: Shit, that's true. You are right, as usual!

thus:
Replaced by #3252

PR #3197, opened 2018-01-29
Startup: Check if logrotation will still work after priv drop (Issue #2386)

richi235:
## Checkboxes

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

## Redmine Ticket
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2386

## Description of Changes:
Write permission on the logdir is necesarry to create new files in there, which is necesary after logrotation to continue logging. (use cases: exe.json and others)

If suricata could open the log files but not create new ones it silently failed, not writing any logs after the first logrotation. This patch adds a check before priv drop and warns+exit()s if we will not be able to create new files after priv drop.

If we won't drop privs we check if the current (and general) user will be able to create logfiles after log rotation.

For details and discussion see:
https://redmine.openinfosecfoundation.org/issues/2386

## PRScript
[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR richi235-pcap: https://buildbot.openinfosecfoundation.org/builders/richi235-pcap/builds/1
- PR richi235: https://buildbot.openinfosecfoundation.org/builders/richi235/builds/1

PR #2386, opened 2016-10-26
dns - support multiple requests before response - v2

jasonish:
Previous PR: #2380 

Changes since last PR:
- Address issue with private pcap.
- Window is now dynamic based on the number of requests seen without a response.  Worst case is the flood/givenup condition.
- Fix issue with multiple requests in a TCP buffer.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/33
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/383

PR #3218, opened 2018-02-05
 Startup: Check if logrotation will still work after priv drop (Issue #2386) v2

richi235:
## Checkboxes

## Redmine Ticket
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2386

## Description of Patch:
Write permission on the logdir is necesarry to create new files in there, which is necesary after logrotation to continue logging. (use cases: exe.json and others)

If suricata could open the log files but not create new ones it silently failed, not writing any logs after the first logrotation. This patch adds a check before priv drop and warns+exit()s if we will not be able to create new files after priv drop.

If we won't drop privs we check if the current (and general) user will be able to create logfiles after log rotation.

For details and discussion see:
https://redmine.openinfosecfoundation.org/issues/2386

## Previous PR (v1)
 #3197

## Changes from v1

  * remove all superfluous `#include`s of standard unix headers
  * don't use malloc and heap for supplementary `gids` array but a variable length array on stack available since C99
  * write `const` before the type of the var it constifies because it's suricata convention that way
  * make suricata struct instance a global variable in `util-conf.c`, to avoid passing it as param, to avoid circular header inclusion problems
  * directly pass `strerror()` output string to `FatalError()`
  * directly call `FatalError()` in suricata.c
  * rephrase error messages
  * implement semantics of root user (root can always write to files, even if he doesn't own them)
  

## PRScript
- PR richi235-pcap: https://buildbot.openinfosecfoundation.org/builders/richi235-pcap/builds/3
- PR richi235: https://buildbot.openinfosecfoundation.org/builders/richi235/builds/3

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2039,6 +2039,46 @@ static TmEcode ParseCommandLine(int argc, char** argv, SCInstance *suri)
         }
     }
 
+#ifndef OS_WIN32
+    /* Get the suricata user ID to given user ID */
+    if (suri->do_setuid == TRUE) {
+        if (SCGetUserID(suri->user_name, suri->group_name,
+                        &suri->userid, &suri->groupid) != 0) {
+            SCLogError(SC_ERR_UID_FAILED, "failed in getting user ID");
+            return TM_ECODE_FAILED;
+        }
+    /* Get the suricata group ID to given group ID */
+    } else if (suri->do_setgid == TRUE) {
+        if (SCGetGroupID(suri->group_name, &suri->groupid) != 0) {
+            SCLogError(SC_ERR_GID_FAILED, "failed in getting group ID");
+            return TM_ECODE_FAILED;
+        }
+    }
+
+    if (set_log_directory
+        && ConfigCheckFileWritablePrivDrop(ConfigGetLogDirectory()) != TM_ECODE_DONE ) {
+
+        if (suri->do_setuid || suri->do_setgid) { /* We will drop privileges */
+            FatalError(SC_ERR_LOGDIR_CMDLINE,
+                      "After dropping privileges to another user the logfile directory \"%s\" "
+                      "defined on the commandline (-l parameter) will no longer be writeable. "
+                      "This means after logrotation it will not be possible to create a new "
+                      "logfile and all following logmessages will get lost. "
+                      "To avoid this, please make the default logdir writeable "
+                      "for the privilege drop user or configure another logdir. Exiting ..." ,
+                       ConfigGetLogDirectory());
+        } else { /* We will not drop privileges */
+            FatalError(SC_ERR_LOGDIR_CMDLINE,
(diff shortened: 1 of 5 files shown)


victorjulien:
Should this have worked?

<pre>
sudo ./src/suricata -c suricata.yaml -i eth0 -l /root/tmp --user victor --group root
[136405] 9/2/2018 -- 15:51:09 - (suricata.c:2069) <Error> (ParseCommandLine) -- [ERRCODE: SC_ERR_LOGDIR_CMDLINE(117)] - After dropping privileges to another user the logfile directory "/root/tmp" defined on the commandline (-l parameter) will no longer be writeable. This means after logrotation it will not be possible to create a new logfile and all following logmessages will get lost. To avoid this, please make the default logdir writeable for the privilege drop user or configure another logdir. Exiting ...
</pre>

src/suricata.c
```diff
@@ -2039,6 +2039,46 @@ static TmEcode ParseCommandLine(int argc, char** argv, SCInstance *suri)
...
+
+    if (set_log_directory
+        && ConfigCheckFileWritablePrivDrop(ConfigGetLogDirectory()) != TM_ECODE_DONE ) {
+
+        if (suri->do_setuid || suri->do_setgid) { /* We will drop privileges */
+            FatalError(SC_ERR_LOGDIR_CMDLINE,
+                      "After dropping privileges to another user the logfile directory \"%s\" "
+                      "defined on the commandline (-l parameter) will no longer be writeable. "
+                      "This means after logrotation it will not be possible to create a new "
+                      "logfile and all following logmessages will get lost. "
```
victorjulien: log messages

src/util-conf.c
```diff
@@ -51,21 +59,91 @@ const char *ConfigGetLogDirectory()
...
+        /* Step 1: getting and preparing data */
+        int supplementary_gid_count;
+
+        /* With 0 as first arg getgroups returns the number of existing supplementary gids */
+        if ( -1 == (supplementary_gid_count = getgroups(0 , NULL)) ) {
+            FatalError(SC_ERR_FATAL, "Getting group IDs of suricata process failed with following error message: %s",
+                       strerror(errno));
+        }
+
+        /* Using a variable length array here, available since C99, recommended in
```
victorjulien: please remove this comment, vla's are considered 'known'

victorjulien:
Seems to work well. Just wondering if that group==root thing should be handled?

richi235:
> Just wondering if that group==root thing should be handled?
It depends on what were the permissions of `/root/tmp` in your test.

In general the root group is not special in any way and has no special privileges, I researched this beforehand, see: 
https://unix.stackexchange.com/questions/44077/what-does-it-mean-to-be-in-group-0

richi235:
no longer relevant. updated version in #3244

PR #3220, opened 2018-02-06
Optimization/tx iterator/v8

victorjulien:

    [..........TTTT]

    [.T..T...TTTT.T]

caller must keep.

behaviour will be used.

Changes since #3214:
- make iterator return a struct { ptr, id, has_next } (tuple)
- remove unused HasTxDetectState call
- simplify GetTxDetectState

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/86
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/88

### DIFF
--- a/rust/gen-c-headers.py
+++ b/rust/gen-c-headers.py
@@ -88,6 +88,8 @@ type_map = {
     "core::AppLayerDecoderEvents": "AppLayerDecoderEvents",
     "AppLayerDecoderEvents": "AppLayerDecoderEvents",
     "core::AppLayerEventType": "AppLayerEventType",
+    "applayer::AppLayerGetTxIterTuple": "AppLayerGetTxIterTuple",
+    "AppLayerGetTxIterTuple": "AppLayerGetTxIterTuple",
     "AppLayerEventType": "AppLayerEventType",
     "CLuaState": "lua_State",
     "Store": "Store",
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -15,6 +15,29 @@
  * 02110-1301, USA.
  */
 
+extern crate libc;
+use std;
+
+#[repr(C)]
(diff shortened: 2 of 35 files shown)


jasonish:
My DNS support now looks like this, and works well with no terminator tricks:
```
diff --git a/src/app-layer-dns-udp.c b/src/app-layer-dns-udp.c
index 49ebf6d49..32a657f14 100644
--- a/src/app-layer-dns-udp.c
+++ b/src/app-layer-dns-udp.c
@@ -387,6 +387,32 @@ static void DNSUDPConfigure(void)
     DNSConfigSetGlobalMemcap(global_memcap);
 }
 
+static AppLayerGetTxIterTuple DNSUDPGetTxIterator(const uint8_t ipproto,
+        const AppProto alproto, void *alstate, uint64_t min_tx_id,
+        uint64_t max_tx_id, AppLayerGetTxIterState *istate)
+{
+    AppLayerGetTxIterTuple result = { NULL, 0, false };
+    DNSState *dns_state = alstate;
+    DNSTransaction *tx = (DNSTransaction *)istate->un.ptr;
+    if (tx == NULL) {
+        tx = TAILQ_FIRST(&dns_state->tx_list);
+    }
+
+    while (tx != NULL) {
+        if (tx->tx_num >= min_tx_id + 1) {
+            DNSTransaction *next = TAILQ_NEXT(tx, next);
+            istate->un.ptr = next;
+            result.tx_ptr = tx;
+            result.tx_id = tx->tx_num - 1;
+            result.has_next = next != NULL;
+            return result;
+        }
+        tx = TAILQ_NEXT(tx, next);
+    }
+
+    return result;
+}
```

victorjulien:
Merged after rebase in #3223

PR #3222, opened 2018-02-06
metadata: fix parsing when not k/v - v1

jasonish:
Handle metadata in the format of:
```
metadata: autosign, tag value, created_at YYYY-MM-DD;
```

Only values that match the key/val format will actually be added for logging. Single values like "autosign" are skipped for now.

I felt this was easier done without regular expressions.

PRScript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/270
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/623

### DIFF
--- a/src/detect-metadata.c
+++ b/src/detect-metadata.c
@@ -34,14 +34,6 @@
 
 #include "util-unittest.h"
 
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged in #3230, thanks!

PR #3225, opened 2018-02-07
conf/yaml: don't allow empty key values - v1

jasonish:
Another approach to https://github.com/OISF/suricata/pull/3163.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2418

Prevent the include of an empty file from creating an empty configuration key. This happens because libyaml givers on a scalar event on an empty file, and the first scalar event is received while in the state of looking for a key, so we set an empty key. Instead, don't transition the state if the value is an empty string.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/271
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/624


maurizioabba:
Sounds good! It passes my scenarios

victorjulien:
Merged in #3230, thanks!

PR #3226, opened 2018-02-07
Fixes/40x/v9

victorjulien:
Backports from the master branch.

Passes


jasonish:
Looks like the .travis.yml needs to be updated to use a newer version of Rust. But then I wonder, if what required that really needed to be back ported?

PR #3231, opened 2018-02-10
Optimization: reset bidirectional flag for rules with same src and dst v2

gozzy:
Reworked #3178

### DIFF
--- a/src/detect-engine-port.c
+++ b/src/detect-engine-port.c
@@ -753,6 +753,33 @@ DetectPortLookupGroup(DetectPort *dp, uint16_t port)
 }
 
 /**
+ * \brief Used to check if a DetectPort list contains an instance with
+ *        a similar DetectPort.  The comparison done is not the one that
+ *        checks the memory for the same instance, but one that checks that the
+ *        two instances hold the same content.
+ *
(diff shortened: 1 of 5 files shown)


src/detect-engine-port.c
```diff
@@ -779,6 +806,38 @@ int DetectPortJoin(DetectEngineCtx *de_ctx, DetectPort *target,
...
+ * \retval false On failure.
+ */
+bool DetectPortListsAreEqual(DetectPort *list1, DetectPort *list2)
+{
+    DetectPort *item = list1;
+    DetectPort *it = list2;
+
+    // First, Check if every item of list1 presents in list2.
+    while (item != NULL && it != NULL) {
+        if (!DetectPortLookupInList(list2, item)) {
```
victorjulien: this seems inefficient: instead of comparing each list member, this does a list lookup per member. Why not compare if port at item == it?

gozzy: I assumed that two lists with similar members do not preserve their order in common case, and 'it'  is only used to check if lists sizes differ. Or am I getting something wrong?

gozzy:
Replaced with #3248

PR #3248, opened 2018-02-14
Optimization: reset bidirectional flag for rules with same src and dst v3

gozzy:
Reworked #3231


gozzy:
@victorjulien are there any issues with this PR? I don't know whether it's better to abandon it or not :)

victorjulien:
Merged in #3304, thanks Alex!

PR #3252, opened 2018-02-16
Initial JA3 integration

thus:

Features implemented:
* JSON logger for JA3.
* JA3 metadata in alerts.

Redmine:

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/145
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/145


thus:
Updates:
- Fixed changes requested by @victorjulien, @norg and @pevma in #3215

thus:
Fixed all changes requested except (https://github.com/OISF/suricata/pull/3215#discussion-diff-166747461R175) which I need some more clarifications on from @victorjulien.

victorjulien:
<pre>
  CC       util-ja3.o
util-ja3.c:118:9: warning: Value stored to 'buffer1' is never read
        buffer1 = buffer2;
        ^         ~~~~~~~
1 warning generated.
</pre>

src/app-layer-ssl.c
```diff
@@ -250,54 +255,416 @@ static void SSLSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags)
...
     input += SSLV3_CLIENT_HELLO_VERSION_LEN;
 
-    /* skip random */
+    return (input - initial_input);
+}
+
+static inline int TLSDecodeHSHelloRandom(SSLState *ssl_state, uint8_t *input,
+                                         const uint32_t input_len)
+{
+    uint8_t *initial_input = input;
```
victorjulien: this can probably be a const to make clear it shouldn't ever change: actually a 'const uint8 * const', as neither the ptr itself as the mem being pointed to should change. (probably applies to many funcs below too)

src/app-layer-ssl.c
```diff
@@ -250,54 +255,416 @@ static void SSLSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags)
...
+                    return -1;
+                }
+            }
+
+            processed_len += 2;
+        }
+
+        rc = Ja3BufferAppendBuffer(ssl_state->ja3_str, ja3_cipher_suites);
+        if (rc != 0) {
+            return -1;
```
victorjulien: not all error paths seem to lead to freeing of ja3_cipher_suites?

src/app-layer-ssl.c
```diff
@@ -250,54 +255,416 @@ static void SSLSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags)
...
+{
+    uint8_t *initial_input = input;
+
+    /* There must not be more than one extension of the same
+       type (RFC5246 section 7.4.1.4). */
+    if (ssl_state->curr_connp->sni) {
+        SCLogDebug("Multiple SNI extensions");
+        SSLSetEvent(ssl_state,
+                TLS_DECODER_EVENT_MULTIPLE_SNI_EXTENSIONS);
+        return -1;
```
victorjulien: should this be fatal or can we set warning and ignore the extra sni?

src/detect-tls-ja3-hash.c
```diff
@@ -0,0 +1,229 @@
...
+    sigmatch_table[DETECT_AL_TLS_JA3_HASH].desc = "content modifier to match the JA3 hash buffer";
+    sigmatch_table[DETECT_AL_TLS_JA3_HASH].url = DOC_URL DOC_VERSION "/rules/ja3-keywords.html#ja3_hash";
+    sigmatch_table[DETECT_AL_TLS_JA3_HASH].Match = NULL;
+    sigmatch_table[DETECT_AL_TLS_JA3_HASH].Setup = DetectTlsJa3HashSetup;
+    sigmatch_table[DETECT_AL_TLS_JA3_HASH].Free = NULL;
+    sigmatch_table[DETECT_AL_TLS_JA3_HASH].RegisterTests = DetectTlsJa3HashRegisterTests;
+
+    sigmatch_table[DETECT_AL_TLS_JA3_HASH].flags |= SIGMATCH_NOOPT;
+
+    DetectAppLayerMpmRegister("ja3_hash", SIG_FLAG_TOSERVER, 2,
```
victorjulien: can you use the *Register2 API calls here? See detect-http-request-line.c for an example. Simpler code, more features.

src/util-ja3.c
```diff
@@ -0,0 +1,283 @@
...
+    if (buffer1 == NULL || buffer2 == NULL) {
+        SCLogError(SC_ERR_INVALID_ARGUMENT, "Buffers should not be NULL");
+        return -1;
+    }
+
+    /* If buffer1 contains no data, then we just copy the second buffer
+       instead of appending its data. */
+    if (buffer1->data == NULL) {
+        Ja3BufferFree(buffer1);
+        buffer1 = buffer2;
```
victorjulien: this is no buffer copy, just a (unused) pointer assignment

src/app-layer-ssl.c
thus: On error it's freed in Ja3BufferAppendBuffer, so it's cleared on all error paths.

src/app-layer-ssl.c
thus: We could just warn and continue decoding the rest of the rest of the packet. Changing the code so it does that instead.

src/util-ja3.c
thus: Changed it. Please check if the new solution is better when the new PR is opened :)

jasonish:
Just some comments on the logging.

- Do we need a "ja3" event type? Can't we just use the "ja3" fields that are already added to the "tls" event type?

- If tls event types already have the ja3 fields, then the alert's don't need a special flag to log ja3 metadata, they will already get it by enabling tls metadata on alerts.

- I was also trying really hard to find a better name for ja3.str but couldn't. But perhaps using the full name "string" would be better than str.

thus:
@victorjulien, what do you think? Do you want me to remove the JA3 logger?

I think it makes sense to have a separate alert meta for JA3. It makes it simple to turn it off if the user don’t need it, without turning off tls.

@victorjulien, what’s your take on the ja3_str name? Do you want me to change it?

jasonish:
> I think it makes sense to have a separate alert meta for JA3. It makes it simple to turn it off if the user don’t need it, without turning off tls.

Is there a use case for logging "ja3" and not "tls"? Otherwise, could enabling ja3 be a flag on the tls type instead of its own type? Just trying to make sure every new toggle has its use case.

The most user friendly way I think is to have a single toggle for ja3. That could be app-layer.protocols.tls.ja3-fingerprints. If its on, tls records automatically include ja3, as well as alerts if tls metadata is on. Even with a separate ja3 record, I think it would make sense to include the ja3 in the tls record by default.

thus:
Replaced by #3280

PR #3261, opened 2018-03-01
travis: redirect unittest output to file in all builds

jasonish:
On error, print the last 500 lines of output then exit 1.

Shoud allow us to see why a build fail on the debug tests,
when the error was burried in an output file we weren't
making visible.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/275
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/628

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -168,17 +168,21 @@ script:
             --with-libnspr-includes=/usr/local/opt/nspr/include/nspr \
             --with-libnspr-libraries=/usr/local/opt/nspr/lib
     fi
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged in #3263, thanks Jason!

victorjulien:
As @chifflier pointed out in chat, it seems we no longer run cocci. I suspect it's related to this patch as it removes 'make check' which would run cocci.

PR #3274, opened 2018-03-08
Feature: WinDivert IPS functionality on Windows

smithjacobj:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2455

Describe changes:
- Enables IPS functionality on Windows using the open-source (LGPLv3)
WinDivert driver and API.

> From https://www.reqrypt.org/windivert-doc.html: "WinDivert is a
user-mode capture/sniffing/modification/blocking/re-injection package
for Windows Vista, Windows Server 2008, Windows 7, and Windows 8.
WinDivert can be used to implement user-mode packet filters, packet
sniffers, firewalls, NAT, VPNs, tunneling applications, etc., without
the need to write kernel-mode code."

- adds `--windivert [filter string]` and `--windivert-forward [filter
string]` command-line options to enable WinDivert IPS mode.
`--windivert[-forward] true` will open a filter for all traffic. See
https://www.reqrypt.org/windivert-doc.html#filter_language for more
information.

Limitation: currently limited to `autofp` runmode.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable): I do not currently have access to buildbot.

PR #3275, opened 2018-03-08
Feature: WinDivert IPS functionality on Windows

smithjacobj:

Describe changes:
- Enables IPS functionality on Windows using the open-source (LGPLv3)
WinDivert driver and API.

> From https://www.reqrypt.org/windivert-doc.html: "WinDivert is a

information.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable): I do not currently have access to buildbot.

-----

This is a replacement PR for #3274 

Should I consider updating `appveyor.yml` to include `--enable-windivert=yes`?


victorjulien:
Thanks Jacob, this is very cool. I will try to play with it soon.

I do need think & discuss internally how to deal with the license. GNU itself says GPLv2 and LGPLv3 are not compatible, so this would put some limitations on how it can be used/distributed.

smithjacobj:
I did not previously understand the incompatibility with older GPL versions but hopefully we can find a solution. Easiest might be to talk to basil@reqrypt.org and see about also (L)GPLv2 licensing WinDivert, but I'm not a licensing expert.

smithjacobj:
@victorjulien I made contact with basil@reqrypt.org and he is receptive to dual-licensing under GPLv2 as well. His only concern is the labor of updating the license/copyright notices.

configure.ac
```diff
@@ -1051,6 +1051,45 @@
         fi
     fi
 
+  # WinDivert support
+    AC_ARG_ENABLE(windivert,
+        AS_HELP_STRING([--enable-windivert],[Enable WinDivert support [default=no]]),,
+        [enable_windivert="no"])
+    
+  # WinDivert can only be enabled on Windows builds
+    AC_CHECK_DECL([OS_WIN32],,[enable_windivert="no"])
```
victorjulien: OS_WIN32 not set for CYGWIN. Probably not a problem as MinGW/native builds will be the new default from 4.1+.

doc/userguide/setting-up-ipsinline-for-windows.rst
```diff
@@ -0,0 +1,47 @@
+Setting up IPS/inline for Windows
+=================================
+
+This guide explains how to work with Suricata in layer 4 inline mode using
+WinDivert on Windows.
+
+First start by compiling Suricata with WinDivert support. For instructions, see
+`Windows Installation
+<https://redmine.openinfosecfoundation.org/attachments/download/1175/SuricataWinInstallationGuide_v1.4.3.pdf>`_.
```
victorjulien: @pevma good reminder that we need the windows install docs in this guide, not in a pdf on redmine. Any progress on that end?

doc/userguide/setting-up-ipsinline-for-windows.rst
```diff
@@ -0,0 +1,47 @@
...
+
+A quick start is to examine all traffic, in which case you can use the following
+command:
+
+::
+  
+  suricata -c suricata.yaml --windivert[-forward] true
+
+The filter is automatically stopped and normal traffic resumes when Suricata is
+stopped.
```
victorjulien: maybe good to also add a few examples like 'only tcp/80' or 'only icmp' or 'all tcp'?

src/Makefile.am
```diff
@@ -2,6 +2,7 @@ noinst_HEADERS = action-globals.h \
     app-layer-nbss.h app-layer-dcerpc-common.h \
     debug.h \
 	flow-private.h queue.h source-nfq-prototypes.h \
+    source-windivert-prototypes.h \
```
victorjulien: indent looks off

src/runmode-windivert.c
```diff
@@ -0,0 +1,72 @@
+/* Copyright (C) 2007-2011 Open Information Security Foundation
```
victorjulien: 2018

src/runmode-windivert.c
```diff
@@ -0,0 +1,72 @@
...
+int RunModeIpsWinDivertAutoFp(void)
+{
+    SCEnter();
+    int ret = 0;
+#ifdef WINDIVERT
+    RunModeInitialize();
+
+    TimeModeSetLive();
+
+    LiveDeviceHasNoStats(); // TODO: this is copied from runmode-nfq.c, is this
```
victorjulien: I think this is correct. Stats will have to use the CaptureStats* calls for IPS specific counters

src/runmode-windivert.h
```diff
@@ -0,0 +1,32 @@
+/* Copyright (C) 2007-2010 Open Information Security Foundation
```
victorjulien: 2018

src/runmodes.c
```diff
@@ -147,6 +147,12 @@ static const char *RunModeTranslateModeToName(int runmode)
 #endif
         case RUNMODE_UNIX_SOCKET:
             return "UNIX_SOCKET";
+        case RUNMODE_WINDIVERT:
+#ifdef WINDIVERT
+            return "WINDIVERT";
```
victorjulien: indent looks off

src/source-windivert-prototypes.h
```diff
@@ -0,0 +1,31 @@
+/* Copyright (C) 2007-2010 Open Information Security Foundation
```
victorjulien: 2018

victorjulien: please check all files for the copyright year

src/source-windivert.c
```diff
@@ -0,0 +1,682 @@
...
+    SCMutexInit(&wd_qv->counters_mutex, NULL);
+
+    g_wd_num++;
+
+unlock:
+    SCMutexUnlock(&g_wd_init_lock);
+
+    if (ret == 0) {
+        /* not really a "device" identical to nfq/ipfw, but best descriptor */
+        LiveRegisterDevice(filter_str);
```
victorjulien: have you seen this value being used anywhere? In other runmodes it may be used in thread names and unix socket. The latter is obviously irrelevant here. Not sure about the former.

src/source-windivert.c
```diff
@@ -0,0 +1,682 @@
...
+    TmSlot *slot;
+
+    CaptureStats stats;
+} WinDivertThreadVars;
+
+#define WINDIVERT_MAX_QUEUE 16
+static WinDivertThreadVars g_wd_tv[WINDIVERT_MAX_QUEUE];
+static WinDivertQueueVars g_wd_qv[WINDIVERT_MAX_QUEUE];
+static uint16_t g_wd_num = 0;
+static SCMutex g_wd_init_lock;
```
victorjulien: can be initialized at declaration: add: = SCMUTEX_INITIALIZER;

src/source-windivert.c
```diff
@@ -0,0 +1,682 @@
...
+    Packet *p = PacketGetFromQueueOrAlloc();
+    if (unlikely(p == NULL)) {
+        SCLogDebug(
+                "PacketGetFromQueueOrAlloc() - failed to obtain Packet buffer");
+        SCReturnInt(TM_ECODE_FAILED);
+    }
+    PKT_SET_SRC(p, PKT_SRC_WIRE);
+
+    /* WinDivert needs to be fed a buffer, so we must make one available. It is
+     * highly likely we'll encounter segmentation offload so we'll just give our
```
victorjulien: segmentation offload will lead to detection issues with certain rules using dsize, and these are meant to look at the original packets.

src/source-windivert.c
```diff
@@ -0,0 +1,682 @@
...
+     * pool packets external buffers.
+     */
+    PacketCallocExtPkt(p, MAX_PAYLOAD_SIZE);
```
victorjulien: I don't think this is a good fit. It's not resizing ext pkt if needed. The PacketCopyData function will expand packets as needed.

src/source-windivert.c
```diff
@@ -0,0 +1,682 @@
...
+ * \brief WinDivert verdict module packet entry function
+ */
+TmEcode VerdictWinDivert(ThreadVars *tv, Packet *p, void *data, PacketQueue *pq,
+        PacketQueue *postpq)
+{
+    SCEnter();
+
+    TmEcode ret = TM_ECODE_OK;
+
+    /* \todo do we need to specifically handle tunnel packets like NFQ? */
```
victorjulien: on receiving tunnel packets Suricata will split a packet into multiple. The encapsulated packets remain connected to the original packet (p->root). If we need to drop based on something detected in the encapsulated packet, we need to effectively drop the 'root' packet. This is what the tunnel handling does, so I'm pretty sure you'll need to handle it as well somehow.

src/suricata.c
```diff
@@ -1809,6 +1825,45 @@ static TmEcode ParseCommandLine(int argc, char** argv, SCInstance *suri)
                 }
             }
 #endif
+            else if(strcmp((long_opts[option_index]).name, "windivert-forward") == 0) {
```
victorjulien: I don't think the difference between windivert and windivert-forward is documented?

victorjulien:
When this is merge, you'll definitely have to come present it at SuriCon :)

src/source-windivert.c
smithjacobj: I'm definitely _way open_ to suggestions here; the way it's currently written is because `WinDivertRecv` takes a buffer and writes to it, rather than providing a buffer to use/copy.

I could probably alloc an intermediate buffer per thread that then gets `PacketCopyData`'d

configure.ac
smithjacobj: Would it be preferable to check for `$host in mingw32/cygwin" then?

src/runmodes.c
smithjacobj: I'm not seeing it?
image

src/source-windivert.c
smithjacobj: Thread names. I'll do what NFQ does, which is just use the queue number (in this case I'll use the array index).

src/suricata.c
smithjacobj: It's explained in the userdoc like using the forward layer of NFQ, but maybe I can make it clearer in the usage printout.

src/source-windivert.c
smithjacobj: I'm seeing this in `source-pcap.c:344-348` - namely the global option to disable offloading, and the `DisableIfaceOffloading function` (of which are only supported on POSIX OS's with `ethtool` from what I can tell):

```
if (LiveGetOffload() == 0) {
    (void)GetIfaceOffloading((char *)pcapconfig->iface, 1, 1);
} else {
    DisableIfaceOffloading(ptv->livedev, 1, 1);
}
```

Perhaps I should look into a solution for disabling offload on Windows? Or at least make the suggestion in the docs?

configure.ac
victorjulien: It's up to you. If you want to support CYGWIN as well it will have to change. But I don't mind limiting this new feature to MinGW/native win.

src/runmodes.c
victorjulien: Yeah you are right, please disregard.

src/source-windivert.c
victorjulien: Probably fine to add this to the docs. Perhaps a bit more than just saying 'disable offloading'. I wouldn't know how to do, although maybe it's an easy google.

src/source-windivert.c
victorjulien: I think the cleanest way to address this would be to add support for your packet size for ConfigGetCaptureValue() in suricata.c. It uses some logic like getting the MTU to figure out what default allocation size we should use in the PacketPool.

victorjulien:
"Should I consider updating appveyor.yml to include --enable-windivert=yes?" Maybe not right away. A first iteration would be classified as 'community supported' as is defined here https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Support_Status#Community

We don't want that to affect other things yet.

If you are interested in helping improve the AppVeyor setup, for example by adding support for MinGW, I'd be happy to take your help!

src/source-windivert.c
smithjacobj: Thank you for the explanation. This situation should be handled properly now.

src/source-windivert.c
smithjacobj: [This document](https://docs.microsoft.com/en-us/windows-hardware/drivers/network/using-registry-values-to-enable-and-disable-task-offloading) may come in handy but I need to understand if restarting devices is enough or if it requires a full Windows restart.

victorjulien:
Can you submit a new version when you've implemented the feedback? Thanks!

PR #3283, opened 2018-03-13
LZ4 compression for pcap files

mfil:
- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2456

Describe changes:
- Add option to compress pcap log files with lz4 before writing to disk

Update to https://github.com/OISF/suricata/pull/3188, https://github.com/OISF/suricata/pull/3200, https://github.com/OISF/suricata/pull/3206, and https://github.com/OISF/suricata/pull/3250
- fix bugs and formatting issues pointed out by victorjulien and jasonish
- squash fixup commits
- exit with error when asked to use compression in sguil mode
- add lz4 file extension

PR #3206, opened 2018-01-31
LZ4 compression for pcap files

mfil:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Add option to compress pcap log files with lz4 before writing to disk

Update to https://github.com/OISF/suricata/pull/3188 and https://github.com/OISF/suricata/pull/3200
- Fixed a bug where pcap logs in multi mode were always uncompressed
- Added ifdefs to build without liblz4 installed
- Fix Appveyor and Travis CI failures

PR #3200, opened 2018-01-30
LZ4 compression for pcap files

mfil:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Add option to compress pcap log files with lz4 before writing to disk

Update to https://github.com/OISF/suricata/pull/3188
- Fixed a bug where pcap logs in multi mode were always uncompressed
- Added ifdefs to (hopefully) build without liblz4 installed

PR #3285, opened 2018-03-14
LZ4 compression for pcap files

mfil:

https://redmine.openinfosecfoundation.org/issues/2456

Describe changes:
- Add option to compress pcap log files with lz4 before writing to disk

Update to https://github.com/OISF/suricata/pull/3188, https://github.com/OISF/suricata/pull/3200, https://github.com/OISF/suricata/pull/3206, https://github.com/OISF/suricata/pull/3250, and https://github.com/OISF/suricata/pull/3283
- fix bugs and formatting issues pointed out by victorjulien and jasonish, including the memory leaks
- add lz4 file extension
- rename lz4-compression-level option to lz4-level
- squash fixup commits
- exit with error when asked to use compression in sguil mode

### DIFF
--- a/src/log-pcap.c
+++ b/src/log-pcap.c
@@ -26,6 +26,11 @@
  */
 
 #include "suricata-common.h"
+#include "util-fmemopen.h"
+
+#ifdef HAVE_LIBLZ4
+#include <lz4frame.h>
+#endif /* HAVE_LIBLZ4 */
 
 #if defined(HAVE_DIRENT_H) && defined(HAVE_FNMATCH_H)
 #define INIT_RING_BUFFER
@@ -62,7 +67,7 @@
 
 #define DEFAULT_LOG_FILENAME            "pcaplog"
 #define MODULE_NAME                     "PcapLog"
-#define MIN_LIMIT                       1 * 1024 * 1024
+#define MIN_LIMIT                       4 * 1024 * 1024
 #define DEFAULT_LIMIT                   100 * 1024 * 1024
 #define DEFAULT_FILE_LIMIT              0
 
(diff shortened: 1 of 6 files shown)


src/log-pcap.c
```diff
@@ -481,6 +632,57 @@ static PcapLogData *PcapLogDataCopy(const PcapLogData *pl)
...
+
+        /* Initialize a new compression context. */
+
+        LZ4F_errorCode_t errcode =
+               LZ4F_createCompressionContext(&copy_comp->lz4f_context, 1);
+        if (LZ4F_isError(errcode)) {
+            SCLogError(SC_ERR_PCAP_LOG_COMPRESS,
+                    "LZ4F_createCompressionContext failed: %s",
+                    LZ4F_getErrorName(errcode));
+            return NULL;
```
jasonish: In all these failure cases, the previous allocation (buffer, pcap_buf, pcap_buf_wrapper) are left hanging, they should be cleaned up before returning.

src/log-pcap.c
```diff
@@ -1077,6 +1295,111 @@ static OutputInitResult PcapLogInitCtx(ConfNode *conf)
...
+
+            /* Use SCFmemopen so we can make pcap_dump write to a buffer. */
+
+            comp->pcap_buf_size = sizeof(struct pcap_file_header) +
+                    sizeof(struct pcap_pkthdr) + PCAP_SNAPLEN;
+            comp->pcap_buf = SCMalloc(comp->pcap_buf_size);
+            if (comp->pcap_buf == NULL) {
+                SCLogError(SC_ERR_MEM_ALLOC, "SCMalloc failed: %s",
+                        strerror(errno));
+                exit(EXIT_FAILURE);
```
jasonish: We probably shouldn't exit here, but instead return an OutputInitResult of failure. And again, cleanup any allocations.

mfil: I tried staying in line with the rest of the function, e.g. at line 1290,
```C
if (stat(pl->dir, &stat_buf) != 0) {
    SCLogError(SC_ERR_LOGDIR_CONFIG, "The sguil-base-dir directory \"%s\" " 
           "supplied doesn't exist. Shutting down the engine",
            pl->dir);
    exit(EXIT_FAILURE);
}
```
Or is there a special reason why it needs to exit in those cases? I don't mind making the change you requested, I just want to understand the code better.

jasonish: Sorry, I missed that. I do see 2 cases here. One is an error in configuration where we exit out, and can be fixed by the user. The other is a malloc error, which the user potentially can't fix, and perhaps we should fail more gracefully there? @victorjulien Any thoughts here on exiting?

victorjulien: Generally we try to continue on malloc errors at runtime. At startup time they are allowed to be fatal.

PR #3308, opened 2018-03-22
tls: code cleanup

thus:
Cleanup in TLS code.

Changes:
- Total rewrite of certificate handling function. Only decode the first certificate in the chain.
- Use TAILQ_EMPTY instead of declaring two variables in the SSL state.
- Use BIT_U32 flags in the SSL state.

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/152
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/152


src/app-layer-ssl.c
```diff
@@ -255,6 +259,249 @@ static void SSLSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags)
...
+}
+
+static inline int TlsDecodeHSCertificateFingerprint(SSLState *ssl_state,
+                                                    uint8_t *input,
+                                                    uint32_t cert_len)
+{
+    if (unlikely(ssl_state->server_connp.cert0_fingerprint != NULL))
+        return 0;
+
+    ssl_state->server_connp.cert0_fingerprint = SCMalloc(SHA1_STRING_LENGTH *
```
victorjulien: prefer SCCalloc here so the newly allocated memory is zeroed

src/app-layer-ssl.c
```diff
@@ -255,6 +259,249 @@ static void SSLSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags)
...
+    TAILQ_INSERT_TAIL(&ssl_state->server_connp.certs, cert, next);
+
+    return 0;
+}
+
+static int TlsDecodeHSCertificate(SSLState *ssl_state,
+                                  const uint8_t * const initial_input,
+                                  const uint32_t input_len)
+{
+    uint8_t *input = (uint8_t *)initial_input;
```
victorjulien: this can be const too right? The ptr itself can change, but the content can't?

src/app-layer-ssl.c
```diff
@@ -255,6 +259,249 @@ static void SSLSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags)
...
+            rc = TlsDecodeHSCertificateFingerprint(ssl_state, input, cert_len);
+            if (rc != 0)
+                goto error;
+
+            DerFree(cert);
+        }
+
+        rc = TlsDecodeHSCertificateAddCertToChain(ssl_state, input, cert_len);
+        if (rc != 0)
+            return -1;
```
victorjulien: why doesn't this jump to error?

src/app-layer-ssl.c
thus: Will do :)

src/app-layer-ssl.c
thus: Sure, just have to typecast a couple of places, but yeah, the content can't be changed.

src/app-layer-ssl.c
thus: Because it has nothing to clean up. I'm changing it to jump to error, since that makes it easier to understand what's happening.

PR #3309, opened 2018-03-22
configure: allow to disable libnss and libnspr

rbgarga:
Let user chose to disable libnss and libnspr support even if these
libraries are installed in the system. Default remains to enable when
libraries are found and disable parameter were not used

https://redmine.openinfosecfoundation.org/issues/2461

Describe changes:
Since libnss and libnspr are not mandatory, there is an option on FreeBSD ports tree to let user to chose to disable it during build phase. It works as expected when building binary packages on a clean environment. When user build the port locally and has one of these libraries installed in the system, if it choses to disable them, configure script doesn't provide a explicit option to have it disabled and then binary package ends up broken because it has binaries linked against these libraries but doesn't have the correct dependency registered to make sure nss/nspr are installed on the system.

This change adds 2 new configure parameters: `--disable-nss` and `--disable-nspr`. When they are used it will not link against respective libraries. The default behavior is not changed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1586,15 +1586,8 @@
     fi
 
   # libnspr
-    enable_nspr="no"
-
-    # Try pkg-config first:
-    PKG_CHECK_MODULES([libnspr], nspr,, [with_pkgconfig_nspr=no])
-    if test "$with_pkgconfig_nspr" != "no"; then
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged as part of #3315, thanks!

PR #3316, opened 2018-03-26
suricatasc - move into python directory - v1

jasonish:
Moves suricatasc into the new top level python directory containing suricatactl.

- Moves the suricatasc packet to suricata.sc.
- Keeps the suricatasc binary.
- Keeps import suricatasc compatibilty, as that will import suricata.sc.
- Fix suricatasc when installed to a non-default directory, as happens with --prefix /somewhere/else is used.

I made the assumption that there are users of the Python suricatasc library which is why I kept compatibilty. I can revert that if needed. It would be a one line change to any users of the library to import from the new location (suricata.sc instead of suricatasc).

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/280
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/633


victorjulien:
Merged in #3317, thanks Jason!

PR #3339, opened 2018-04-11
Feature: WinDivert IPS functionality on Windows

smithjacobj:

Describe changes:
- Enables IPS functionality on Windows using the open-source (LGPLv3/GPLv2)
WinDivert 1.4.1 driver and API.

> From https://www.reqrypt.org/windivert-doc.html: "WinDivert is a

information.

- Added interface offload query/disable/restore for Win32 via WMI (see `win32-syscall.h/c`)

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable): I do not currently have access to buildbot.

-----

This is a replacement PR for #3337


victorjulien:
screenshot from 2018-04-14 10-55-10

This PR adds a new compile warning. This is w/o enabling windivert.

victorjulien:
Could you clean up the commit message to look like a single commit instead of a squash of many?

configure.ac
```diff
@@ -232,8 +232,8 @@
             RUST_LDADD="-ldl -lrt -lm"
             ;;
         *-*-mingw32*)
-            CFLAGS="${CFLAGS} -DOS_WIN32"
-            LDFLAGS="${LDFLAGS} -lws2_32"
+            CFLAGS="${CFLAGS} -DOS_WIN32 -DNTDDI_VERSION=NTDDI_VISTA -D_WIN32_WINNT=_WIN32_WINNT_VISTA"
```
victorjulien: should this be conditional based on whether windivert is enabled?

configure.ac
```diff
@@ -1394,7 +1433,7 @@
 
         AC_ARG_WITH(libhtp_includes,
                 [  --with-libhtp-includes=DIR  libhtp include directory],
-                [with_libhtp_includes="$withval"],[with_libhtp_includes=no])
+                [with_libhtp_includes="$withval"],[with_libhtp_includes="no"])
```
victorjulien: Please make this a separate commit as it's unrelated to windivert

src/source-windivert.c
```diff
@@ -0,0 +1,844 @@
...
+#endif
+
+#ifndef WINDIVERT
+/* Gracefully handle the case where no WinDivert support is compiled in */
+
+TmEcode NoWinDivertSupportExit(ThreadVars *, const void *, void **);
+
+void TmModuleReceiveWinDivertRegister(void)
+{
+    memset(&tmm_modules[TMM_RECEIVEWINDIVERT], 0, sizeof(TmModule));
```
victorjulien: instead of these individual memsets, I think we can memset the whole thing at startup? RegisterAllModules looks like a good candidate.

configure.ac
smithjacobj: Good question. It _can_ be conditional, but the WMI offload query/disable code requires Vista or newer. I suspect you still want XP compatibility if possible? I can probably flag the functionality in an `#ifdef` on `NTDDI_VERSION`/`_WIN32_WINNT`, and have `--enable-windivert` switch those on by default.

src/source-windivert.c
smithjacobj: Is this a request that I change this for the PR or a note for your reference? Right now, this seems to be the default behavior for other thread modules to zero themselves.

configure.ac
smithjacobj: OK.

smithjacobj:
> This PR adds a new compile warning. This is w/o enabling windivert.

Are you hoping to eliminate this warning in this PR? I can work on it, the changes to defining API versions mean that Windows provides its own definitions for POSIX host/network byte order conversions.

configure.ac
victorjulien: I don't specifically care about XP. Whats the lowest OS version still supported by Microsoft for the general public?

src/source-windivert.c
victorjulien: Hmm maybe I need more coffee, but I don't see the other modules doing this. I think it would be a cleaner solution to do the whole thing at once, but it's not a blocker. So if you're up for it, go ahead :)

victorjulien:
In general PRs should not introduce new warnings. Perhaps it would make sense to do a separate commit / PR for the changing Windows API, then build windivert commit(s) on top of that?

TechnikEmpire:
Just FYI, this warning coming from ws2tcpip.h is because mingw is garbage and most of their windows headers are wrong. The actual official header has the proper function definitions here with the const qualifier.

configure.ac
smithjacobj: Windows 7 is the oldest supported now. Windows Vista support technically ended a year ago.

I made the change to make this conditional testing another issue, so it'll be optional without WinDivert.

victorjulien:
On 24-04-18 16:00, Jacob Masen-Smith wrote:
> *@smithjacobj* commented on this pull request.
> 
> 
> In configure.ac
> <https://github.com/OISF/suricata/pull/3339#discussion_r183740898>:
> 
>> @@ -232,8 +232,8 @@
>              RUST_LDADD="-ldl -lrt -lm"
>              ;;
>          *-*-mingw32*)
> -            CFLAGS="${CFLAGS} -DOS_WIN32"
> -            LDFLAGS="${LDFLAGS} -lws2_32"
> +            CFLAGS="${CFLAGS} -DOS_WIN32 -DNTDDI_VERSION=NTDDI_VISTA -D_WIN32_WINNT=_WIN32_WINNT_VISTA"
> 
> Windows 7 is the oldest supported now. Windows Vista support technically
> ended a year ago.
> 
> I made the change to make this conditional testing another issue, so
> it'll be optional without WinDivert.

Cool. Btw, I'm not opposed to changes breaking things for Vista, XP or
even lower.

src/source-windivert.c
smithjacobj: Nope, I need more coffee, I was wrong. I based it on `source-nfq.c` but that uses a bunch of manual `NULL` assignments. I extrapolated the `memset` zeroing from that.

There currently doesn't seem to be any full zeroing of the `tmm_modules` allocation.

smithjacobj:
It does look like I need to consider Win Server 2008 R2, which is end-of-life in 2020, but the major architectural changes happen around Vista and most of the API changes after that are mostly iterative, so it's not so hard maintaining Vista compatibility.

PR #3344, opened 2018-04-18
fileinfo: fix IP parameters order in events

regit:
All fileinfo events should have a source IP that is the one sending
the data and a destination IP that is the receiving part. This was
not the case as the side information was not contained in the File
structure.

This patch adds a side information to the File allowing the
application layer to define which side should be considered at
logging time.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2492

Describe changes:
- Add FileSetSide function to set if file is to_client or to_server
- Make SMTP, HTTP and FTP use that API
- Protocols with Rust parser are NOT covered

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/396
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/179

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -740,6 +740,17 @@ static int FTPDataParse(Flow *f, FtpDataState *ftpdata_state,
                          input, input_len, flags) == NULL) {
             SCLogDebug("Can't open file");
             ret = -1;
+        } else {
+            switch (data->cmd) {
+                case FTP_COMMAND_STOR:
+                    FileSetSide(ftpdata_state->files->tail, FILE_TO_SERVER);
+                    break;
+                case FTP_COMMAND_RETR:
+                    FileSetSide(ftpdata_state->files->tail, FILE_TO_CLIENT);
+                    break;
+                default:
+                    break;
+            }
         }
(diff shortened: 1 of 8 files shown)


victorjulien:
I wonder why we would do a per proto update instead of just using the direction implied by the file container. The logging code knows the direction, so why not make sure this is used? https://github.com/OISF/suricata/blob/master/src/output-file.c#L174

regit:
Do you want to update the API used by JsonFileLogger or do you want to set the flag introduced in the File in a generic way ?

inliniac:
On 19-04-18 11:12, Eric Leblond wrote:
> Do you want to update the API used by JsonFileLogger or do you want to
> set the flag introduced in the File in a generic way ?

I'm unconvinced we need the per file flag at all.

regit:
Ok then I update the Logger API to get the direction at the time of the call.

With actual code, I don't see how to get the information:
```
static int JsonFileLogger(ThreadVars *tv, void *thread_data, const Packet *p, const File *ff)
```

PR #3353, opened 2018-04-24
output: fix logging wrong direction in tls upgrade

victorjulien:
When upgrading to TLS from HTTP logging of the final HTTP tx could
have the wrong direction. This was due to the original packet triggering/
finalizing the upgrade would be used as the base for both the toserver
and toclient pseudo packet meaning it was wrong in one direction.

This patch creates a pseudo packet in the same way as the flow timeout
code does, so it no longer takes the raw original packet in.

Bug #2430

passed


victorjulien:
https://redmine.openinfosecfoundation.org/issues/2480

PR #3355, opened 2018-04-24
Adds WinDivert support to Windows builds

smithjacobj:

Enables IPS functionality on Windows using the open-source (LGPLv3/GPLv2)
WinDivert driver and API.

From https://www.reqrypt.org/windivert-doc.html : "WinDivert is a

    information.

Additionally:
- `tmm_modules` now zeroed during `RegisterAllModules`
- fixed Windows Vista+ `inet_ntop` call in `PrintInet`
- fixed `GetRandom` bug (nonexistent keys) on fresh Windows installs

---

This is a replacement for #3337


doc/userguide/setting-up-ipsinline-for-windows.rst
```diff
@@ -0,0 +1,64 @@
...
+command in an elevated command prompt or terminal:
+
+::
+  
+  suricata -c suricata.yaml --windivert [filter string]
+
+For information on the WinDivert filter language, see
+https://www.reqrypt.org/windivert-doc.html#filter_language
+
+If Suricata is running on a gateway and is meant to protect the computers behind
```
norg: just minor suggestion, "protect the network" instead of "computers" :)

src/source-windivert.c
```diff
@@ -0,0 +1,891 @@
...
+     * for direct data alloc size, and this is meaningless if large segments are
+     * coalesced before they reach WinDivert */
+    bool success = false;
+    if (wd_tv->offload_enabled) {
+        /* allocate external, if not already */
+        PacketCallocExtPkt(p, MAX_PAYLOAD_SIZE);
+
+        success = WinDivertRecv(wd_tv->filter_handle, p->ext_pkt,
+                                MAX_PAYLOAD_SIZE, &p->windivert_v.addr,
+                                &p->pktlen);
```
victorjulien: our coccinelle check for pkt access triggers on this:
```
Using concurrency level 28
Testing cocci file: access-pkt-packet.cocci
Invalid Packet->pktlen usage, GET_PKT_LEN macro must be used at /builds/inliniac/suricata-ci/suricata/qa/coccinelle/../../src/source-windivert.c:342
ERROR: Job failed: exit code 1
```

victorjulien: I guess the way to work around this would be to use a local var to get the pktlen from WinDivertRecv and then use the SET_PKT_LEN macro.

src/source-windivert.c
```diff
@@ -0,0 +1,891 @@
...
+
+    g_wd_num++;
+
+unlock:
+    SCMutexUnlock(&g_wd_init_lock);
+
+    if (ret == 0) {
+        // stringify queue index to use as thread name descriptor
+        char wd_num_str[6];
+        wd_num_str[sizeof(wd_num_str) - 1] = 0;
```
victorjulien: instead of this line you can simply init the var at declaration:
```c
char wd_num_str[6] = "";
```

src/source-windivert.c
```diff
@@ -0,0 +1,891 @@
...
+    }
+
+    for (IP_ADAPTER_ADDRESSES *if_info = if_info_list; if_info != NULL;
+         if_info = if_info->Next) {
+
+        if (WinDivertIfaceMatchFilter(wd_qv->filter_str, if_info->IfIndex)) {
+            SCLogInfo("Found adapter %s matching WinDivert filter %s",
+                      if_info->AdapterName, wd_qv->filter_str);
+
+            LiveDevice *new_ldev = malloc(sizeof(LiveDevice));
```
victorjulien: please use SCMalloc and check the result

victorjulien: actually I'd prefer SCCalloc so the struct is zeroed

src/source-windivert.c
```diff
@@ -0,0 +1,891 @@
...
+            new_ldev->dev = SCStrdup(if_info->AdapterName);
```
victorjulien: check result of SCStrdup

src/source-windivert.c
```diff
@@ -0,0 +1,891 @@
...
+ * \brief WinDivert verdict module packet entry function
+ */
+TmEcode VerdictWinDivert(ThreadVars *tv, Packet *p, void *data, PacketQueue *pq,
+                         PacketQueue *postpq)
+{
+    SCEnter();
+
+    TmEcode ret = TM_ECODE_OK;
+
+    /* \todo do we need to specifically handle tunnel packets like NFQ? */
```
victorjulien: comment obsolete?

src/source-windivert.c
```diff
@@ -0,0 +1,891 @@
...
+    size_t count = (sizeof(tests) / sizeof(tests[0]));
+
+    for (size_t i = 0; i < count; i++) {
+        struct testdata test = tests[i];
+
+        bool actual = WinDivertIfaceMatchFilter(test.filter, test.if_index);
+        if (actual != test.expected) {
+            printf("WinDivertIfaceMatchFilter(\"%s\", %d) == %d, expected %d\n",
+                   test.filter, test.if_index, actual, test.expected);
+            // FAIL;
```
victorjulien: todo?

src/suricata.c
```diff
@@ -2405,12 +2463,30 @@ static int ConfigGetCaptureValue(SCInstance *suri)
...
+                    break;
+                }
+
+                g_default_mtu = DEFAULT_MTU;
+                default_packet_size = DEFAULT_PACKET_SIZE;
+                break;
+#endif /* WINDIVERT */
             case RUNMODE_PCAP_DEV:
             case RUNMODE_AFP_DEV:
+            /* \bug: are PCAP/AFP supposed to fall through to strip_trailing_plus? */
```
victorjulien: can you remove this comment and add it as a ticket or a PR?

src/util-ioctl.c
```diff
@@ -53,6 +57,8 @@
  */
 static int GetIfaceMaxHWHeaderLength(const char *pcap_dev)
 {
+    /* \bug: this isn't necessarily universal (my Ubuntu WiFi device is wlp1s0),
```
victorjulien: same here: please record things like this elsewhere

src/util-print.c
```diff
@@ -268,7 +268,16 @@ const char *PrintInet(int af, const void *src, char *dst, socklen_t size)
 {
     switch (af) {
         case AF_INET:
+#if defined(OS_WIN32) && NTDDI_VERSION >= NTDDI_VISTA
+            // because Windows has to provide a non-conformant inet_ntop, of
+            // course!
+            ; // case has to start with a statement
```
victorjulien: I prefer adding braces instead, something like:
```c
case AF_INET:
#if defined(OS_WIN32) && NTDDI_VERSION >= NTDDI_VISTA
{
  struct in_addr _src;
  memcpy(&_src, src, sizeof(struct in_addr));
  return inet_ntop(af, &_src, dst, size);
}
```

src/util-random.c
```diff
@@ -24,8 +24,8 @@
  * SEI CERT C Coding Standard MSC30-C
  */
 
-#include "suricata-common.h"
 #include "util-random.h"
+#include "suricata-common.h"
```
victorjulien: why is this needed?

src/win32-misc.h
```diff
@@ -24,6 +24,8 @@
 #ifndef __WIN32_MISC_H__
 #define __WIN32_MISC_H__
 
+#include <sdkddkver.h>
```
victorjulien: I'm not familiar with this header, is it available on all windows platforms?

victorjulien: it should probably be checked for existence in configure.ac

src/win32-syscall.c
```diff
@@ -0,0 +1,1651 @@
...
+{
+    DWORD err = NO_ERROR;
+    IP_ADAPTER_ADDRESSES *if_info_list;
+
+    ULONG size = 0;
+    err = GetAdaptersAddresses(AF_UNSPEC, 0, NULL, NULL, &size);
+    if (err != ERROR_BUFFER_OVERFLOW) {
+        return err;
+    }
+    if_info_list = malloc((size_t)size);
```
victorjulien: SCMalloc

src/win32-syscall.c
```diff
@@ -0,0 +1,1651 @@
...
+    if (if_info_list == NULL) {
+        return ERROR_NOT_ENOUGH_MEMORY;
+    }
+    err = GetAdaptersAddresses(AF_UNSPEC, 0, NULL, if_info_list, &size);
+    if (err != NO_ERROR) {
+        free(if_info_list);
```
victorjulien: SCFree

src/win32-syscall.c
```diff
@@ -0,0 +1,1651 @@
...
+    }
+
+    if (*pif_info == NULL) {
+        ret = ERROR_NOT_FOUND;
+    }
+
+    return ret;
+}
+
+#if NTDDI_VERSION < NTDDI_VISTA
```
victorjulien: does this still make sense? Is there something older than Vista we'd want to support? I'm thinking there is not. As this is new functionality we can easily set a minimum supported version to be something recent. (e.g. just Win10 + latest Server?)

src/win32-syscall.c
```diff
@@ -0,0 +1,1651 @@
...
+    /* now search for the right adapter in the list */
+    IP_ADAPTER_ADDRESSES *if_info = NULL;
+    for (if_info = if_info_list; if_info != NULL; if_info = if_info->Next) {
+        /* -1 (uint) is an invalid value */
+        if (if_info->Mtu == (uint32_t)-1) {
+            continue;
+        }
+
+        /* we want to return the largest MTU value so we allocate enough */
+        if (if_info->Mtu > mtu) {
```
victorjulien:
```c
mtu = MAX(mtu, if_info->Mtu);
```

victorjulien:
Btw, I'm in the process of setting up a vm for testing.

src/win32-misc.h
sgeto: > I'm not familiar with this header, is it available on all windows platforms?

Yes, It's a very common one.

src/win32-syscall.c
smithjacobj: That was a matter of assumption I made based on [this discussion](https://github.com/OISF/suricata/pull/3339#discussion_r181546625) that XP compatibility was still desirable.

src/source-windivert.c
smithjacobj: That would zero the first index though, not the last one. Zeroing the last one means that, as long as I keep the length values correct when setting it, no c-string functions should ever overrun the buffer as the last index is always NULL.

src/util-random.c
smithjacobj: It's not, not sure why I did it so I'm reverting it.

src/win32-misc.h
smithjacobj: I'm pretty sure it's been available since Windows 2000. If we have any expectation of somebody trying to build it on Windows ME :laughing: then I guess it might need to be added to `configure.ac`

src/win32-syscall.c
smithjacobj: To be specific, the minimum Windows OS this functionality requires is Vista, for both the WMI APIs and WinDivert (which relies on Windows Filtering Platform introduced in Vista.)

PR #3358, opened 2018-04-30
suricata-update integration - v2

jasonish:
More suricata-update integration:

- Install the engine provided rules in $datadir/suricata/rules (eg. /usr/share/suricata/rules).
- Create a python module, defaults.py that has the compile time directories such as sysconfdir, e_datarulesdir so Python tools can pick the locations that Suricata was compiled with. This seemed like the best idea at the time, but executing Suricata to extract the information is also possible. Even if not the best option for suricata-update, its still useful for other Python tools, ie: to prevent suricatasc.py being a .in file (will do another PR for that)
- If suricata-update is bundled, use it during "install-rules".
- If suricata-update is bundled, set the default-rule-path to the expected directory and enable only one rule file, suricata.rules.

Requires update suricata-update with this PR:
- https://github.com/OISF/suricata-update/pull/36

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/286
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/639

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,6 +30,12 @@ install-conf:
 	install -m 770 -d "$(DESTDIR)$(e_localstatedir)"
 
 install-rules:
+if HAVE_SURICATA_UPDATE
+	$(DESTDIR)$(bindir)/suricata-update \
+		--suricata $(DESTDIR)$(bindir)/suricata \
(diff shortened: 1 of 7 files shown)


victorjulien:
Merged in #3368, thanks!

PR #3362, opened 2018-05-02
Restore core dump privileges after uid/gid change (1271)

B0fH:

https://redmine.openinfosecfoundation.org/issues/1271

Describe changes:
Per man(2) prctl, this flag is set to the value of '/proc/sys/fs/suid_dumpable' - which defaults to zero - when the effective user and/or group for the process is changed, effectively preventing suricata from creating core dumps after the main thread drops capabilities and changes the process uid/gid. This PR resets the flag (in order to allow suricata to create core dumps) after the process's effective UID/GID have been changed.

### DIFF
--- a/src/util-privs.c
+++ b/src/util-privs.c
@@ -93,7 +93,7 @@ void SCDropMainThreadCaps(uint32_t userid, uint32_t groupid)
     #ifdef HAVE_SYS_PRCTL_H
     if(prctl(PR_SET_DUMPABLE, 1) < 0) {
         SCLogError(SC_ERR_CHANGING_CAPS_FAILED, "restoring core dump privileges"
-                " failed");
+                " failed, error (%s)", strerror(errno));
     }
     #endif
     
--- a/src/util-privs.c
+++ b/src/util-privs.c
@@ -90,6 +90,13 @@ void SCDropMainThreadCaps(uint32_t userid, uint32_t groupid)
         exit(EXIT_FAILURE);
     }
 
+    #ifdef HAVE_SYS_PRCTL_H
+    if(prctl(PR_SET_DUMPABLE, 1) < 0) {
+        SCLogError(SC_ERR_CHANGING_CAPS_FAILED, "restoring core dump privileges"
+                " failed");
+    }
+    #endif
+    
     SCLogInfo("dropped the caps for main thread");
 }
 


src/util-privs.c
```diff
@@ -90,6 +90,13 @@ void SCDropMainThreadCaps(uint32_t userid, uint32_t groupid)
         exit(EXIT_FAILURE);
     }
 
+    #ifdef HAVE_SYS_PRCTL_H
+    if(prctl(PR_SET_DUMPABLE, 1) < 0) {
+        SCLogError(SC_ERR_CHANGING_CAPS_FAILED, "restoring core dump privileges"
+                " failed");
```
jasonish: Looks like prctl() sets errno on failure, would it make sense to log that as well?

B0fH: Indeed, I've updated the PR to include errno(via strerror)  in the error message.

victorjulien:
@bluhm @eroullit any thoughts on this, esp in relation to your privs work?

bluhm:
The proposed change does not affect OpenBSD.  There a changed UID program does not write a core dump by default.  The program cannot circumvent this by itself.  Only root can globally activate in the kernel, that such core dumps are written to /var/crash/ .  Only root may access this directory.  There is no prctl(), this is Linux specific.

catenacyber:
Core dumps may leak sensitive information (ie network traffic).
Can this behavior be optional ?

jasonish:
> Core dumps may leak sensitive information (ie network traffic).

Is this really a problem? I think the primary use case here is when Suricata is started as root then configured to switch to a user account after the network device is open for reading. So whomever is starting Suricata already has the privileges to read network data.

jasonish:
Tests OK. My test scenarios are setuid binary switching to user, and starting as root then switching to user.  In both cases this allows the coredump to be created, which I believe is better default behaviour.

jasonish:
However, maybe re-enabling should follow the same logic as enabling is before dropping privileges:

https://github.com/OISF/suricata/blob/master/src/util-coredump-config.c#L97

jasonish:
Also, I seem to be able to fix this issue just by doing the coredump configuration after the privileges are dropped which seems like a simpler fix.  Am I missing anything as to why this might not be a good idea?
```
diff --git a/src/suricata.c b/src/suricata.c
index cf1742098..0d7476af5 100644
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2851,8 +2851,6 @@ static int PostConfLoadedSetup(SCInstance *suri)
 #endif
     SCAsn1LoadConfig();
 
-    CoredumpLoadConfig();
-
     DecodeGlobalConfig();
 
     LiveDeviceFinalize();
@@ -3014,6 +3012,7 @@ int main(int argc, char **argv)
     }
 
     SCDropMainThreadCaps(suricata.userid, suricata.groupid);
+    CoredumpLoadConfig();
     PreRunPostPrivsDropInit(suricata.run_mode);
 
     PostConfLoadedDetectSetup(&suricata);
```

B0fH:
Hi Jason -
 This seems to make more sense than my patch which just restores the PR_SET_DUMPABLE flag, my only concern would be the Suricata crashing before the call to SCDropMainThreadCaps(), maybe it makes sense to call CoredumpLoadConfig() twice?

Thanks, 
 Elazar

----------------------------------------

jasonish:
Closing. Fixed with
https://github.com/victorjulien/suricata/commit/178d420f36d37702e5909d2e55eef5b143d77467

PR #3366, opened 2018-05-03
output-file: Ensure that we evaluate whether a file has been stored (2500) v2

B0fH:

https://redmine.openinfosecfoundation.org/issues/2500

Describe changes:
The previous logic did not take into account whether a file has been stored or not before executing the logger functions. This would result in stored almost always (except where force file store for all files is enabled) evaluating to false as the store/no store evaluation happens after the logger calls. This change ensures that either FILE_STORE or FILE_NOSTORE is set before executing the logger functions.

**Test Case**
With filestore configured to only log selected files: `force-filestore: no`

Rule:
`
alert http any any -> any any (msg:"FILESTORE executable file"; flow:established,to_client; filesize:5KB<>5MB; filemagic:"executable"; filestore; noalert; sid:95500001; rev:1;)
`

Lua test script:
[test.lua.zip](https://github.com/OISF/suricata/files/1971074/test.lua.zip)

Expected result:
Without the change, stored will always evaluate to false even if the file is ultimately stored due to the filestore keyword in the rule.

### DIFF
--- a/src/output-file.c
+++ b/src/output-file.c
@@ -114,7 +114,7 @@ static void OutputFileLogFfc(ThreadVars *tv,
 
             SCLogDebug("ff %p state %u", ff, ff->state);
 
-            if (ff->state > FILE_STATE_OPENED (ff->flags & FILE_STORE || ff->flags & FILE_NOSTORE)) {
+            if (ff->state > FILE_STATE_OPENED && (ff->flags & FILE_STORE || ff->flags & FILE_NOSTORE)) {
                 bool file_logged = false;
 #ifdef HAVE_MAGIC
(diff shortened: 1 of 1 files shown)


B0fH:
Closing this request as it breaks logging for application protocols other than HTTP in a case where the file isn't stored. This due to the fact that FILE_NOSTORE is only set on File Flow object(ff) in app-layer-htp-file.c/HTPFileOpen() but not in the other application protocol handlers. 

On a related note, files that aren't stored may not get pruned, as the logic in util-file.c/FilePruneFile() looks for the FILE_STORED or FILE_NOSTORE flag.

PR #3369, opened 2018-05-06
output-file: Ensure that we evaluate whether a file has been stored (2500) v3

B0fH:

https://redmine.openinfosecfoundation.org/issues/2500

Describe changes:
The previous logic did not take into account whether a file has been stored or not before executing the logger functions. This would result in stored almost always (except where force file store for all files is enabled) evaluating to false as the store/no store evaluation happens after the logger calls. This change ensures that either FILE_STORED or FILE_NOSTORE is set before executing the logger functions. It also ensures that files which haven't been marked FILE_STORE after the stream is closed are marked FILE_NOSTORE.

**Test Case**
With filestore configured to only log selected files: `force-filestore: no`

Rule:
`alert http any any -> any any (msg:"FILESTORE executable file"; flow:established,to_client; filesize:5KB<>5MB; filemagic:"executable"; filestore; noalert; sid:95500001; rev:1;)`

Lua test script:
[test.lua.zip](https://github.com/OISF/suricata/files/1977368/test.lua.zip)

### DIFF
--- a/src/output-filedata.c
+++ b/src/output-filedata.c
@@ -145,6 +145,12 @@ static void OutputFiledataLogFfc(ThreadVars *tv, OutputLoggerThreadStore *store,
 
             if (!(ff->flags & FILE_STORE)) {
                 SCLogDebug("ff FILE_STORE not set");
+                /* if the stream has been closed and we still
+                 * haven't stored the file then mark it no-store */
+                if (ff->state >= FILE_STATE_CLOSED &&
+                    (p->flowflags & STREAM_EOF)) {
+                    ff->flags |= FILE_NOSTORE;
+                }
                 continue;
             }
 
--- a/src/output-filedata.c
+++ b/src/output-filedata.c
@@ -148,7 +148,7 @@ static void OutputFiledataLogFfc(ThreadVars *tv, OutputLoggerThreadStore *store,
                 /* if the stream has been closed and we still
                  * haven't stored the file then mark it no-store */
                 if (ff->state >= FILE_STATE_CLOSED &&
-                    (p->flowflags & STREAM_EOF)) {
+                    (p->flags & PKT_STREAM_EOF)) {
                     ff->flags |= FILE_NOSTORE;
                 }
                 continue;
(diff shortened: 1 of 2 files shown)


src/output-filedata.c
```diff
@@ -145,6 +145,12 @@ static void OutputFiledataLogFfc(ThreadVars *tv, OutputLoggerThreadStore *store,
 
             if (!(ff->flags & FILE_STORE)) {
                 SCLogDebug("ff FILE_STORE not set");
+                /* if the stream has been closed and we still
+                 * haven't stored the file then mark it no-store */
+                if (ff->state >= FILE_STATE_CLOSED &&
+                    (p->flowflags & STREAM_EOF)) {
```
victorjulien: This makes cocci unhappy:
```
Testing cocci file: struct-flags.cocci
Invalid usage of flags field at /builds/inliniac/suricata-ci/suricata/qa/coccinelle/../../src/output-filedata.c:151, flags value is incorrect (wrong family).
```

B0fH:
Now that I'm looking at the flags, this probably isn't the right solution either. I had originally looked at something like the following:

               if (ff->state >= FILE_STATE_CLOSED &&
                   (file_close || file_trunc)) {
                    ff->flags |= FILE_NOSTORE;
                }

However, it would take upwards of 5-10 minutes at time for the PKT_PSEUDO_STREAM_END flag to be set, delaying file logging. I'm not sure why this is, could be something specific to my setup. Maybe this makes more sense:

                if (ff->state >= FILE_STATE_CLOSED &&
                   (p->flags & PKT_STREAM_EOF)) {
                    ff->flags |= FILE_NOSTORE;
                }

victorjulien:
A suricata-verify test would make it a lot easier to test and analyze

B0fH:
Hi Victor -
 I saw the note on the original bug - I'll try and see what test can be built for the test framework. Just need to find the time.

Thanks,
 Elazar

----------------------------------------

victorjulien:
If you find the time, please do a SV test or a bug ticket. Thanks!

PR #3345, opened 2018-04-19
Add support for kerberos 01

chifflier:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: none

Describe changes:
- This PR adds support for a new protocol, Kerberos (RFC 4120)
- It follows exchanges on port 88, both TCP and UDP
- It is written in Rust, and the parser itself is shared with nfsv4 and samba (`kerberos-parser`)
- It adds a logger for some parameters (client and server realm+principal, cryptographic parameters), in JSON
- It adds new rules for events. The rule for weak crypto algorithms is configurable, it can generate tons of alerts

Example of JSON output for an `AS-REP` message (only "krb5" key):
```
  "krb5": {
    "msg_type": "KRB_AS_REP",
    "cname": "testuser",
    "realm": "INTERNET.LOCAL",
    "sname": "krbtgt/INTERNET.LOCAL",
    "encryption": "aes256-cts-hmac-sha1-96"
  }
```

Output for a failed `AS-REP`:
```
  "krb5": {
    "msg_type": "KRB_ERROR",
    "failed_request": "KRB_AS_REQ",
    "error_code": "KDC_ERR_PREAUTH_REQUIRED",
    "cname": "testuser",
    "realm": "INTERNET.LOCAL",
    "sname": "krbtgt/INTERNET.LOCAL",
    "encryption": "<none>"
  }
```

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #3371, opened 2018-05-09
Rust add kerberos 02

chifflier:

Describe changes:
- Update of #3345
- New detection keywords: `krb5.msg_type`, `krb5.cname`, `krb5.sname` and `krb5.err_code`


src/detect-krb5-name.c
```diff
@@ -0,0 +1,411 @@
...
+    SCLogDebug("g_krb5_name_list_id %d", g_krb5_name_list_id);
+}
+
+static int DetectEngineInspectKRB5Generic(ThreadVars *tv,
+        DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx,
+        const Signature *s, const SigMatchData *smd,
+        Flow *f, uint8_t flags, void *alstate,
+        void *txv, uint64_t tx_id)
+{
+    return DetectEngineInspectGenericList(tv, de_ctx, det_ctx, s, smd,
```
victorjulien: This is a pattern matching one, right? I think this should be implemented differently then. Needs to be a 'sticky buffer' keyword, see smb_share for example.

chifflier: Hi,
after running some tests I'm not sure a sticky buffer is the best way to go: the client and server names are not simple strings, but PrincipalNames, which are encoded using a name-type and a list of strings (see https://tools.ietf.org/html/rfc4120#section-6.2). Using sticky buffers would mean that we can only match the entire buffer ?
Multiple strings are common, for ex in Active Directory, where you can find PrincipalNames like ( "ldap", "DOMAIN.COM" ), or ("ldap") only, quite often: using a keyword means we can match "ldap" easily. A substring match would not be great either, in some types the name can be encoded.
The current (proposed) approach leaves the matching on the rust side, to deal with these cases (it checks all substrings for a match).
Any idea ? Do you think a sticky buffer is better ?

victorjulien: I'm not sure I understand the issue. If you need to match multiple things it can be done like with DNS (dns_query) where we can have multiple query buffers.

Can you mock up some example rules that would show the issue?

chifflier: The server name is a variable list of items (which are usually strings, but could be something else). For ex, it can be just ["krbtgt"], or ["krbtgt","DOMAIN.COM"], or it could contain more strings. I would like a rule `krb5_sname:krbtgt` to match any of these server names.

Another case is a Kerberos login using a certificate. I think the type used for the client name will be a X.509 DN, which is not a string but a DER object. In that case, `krb5_cname:toto` should match "CN=toto,DC=DOMAIN,DC=COM".

- Can I allocate multiple sticky buffers for one transaction, the number being known only after the transaction is created ?
- Where do I free these buffers ?

Also, `dns_query` seems to handle multiple buffers, but the code looks quite complicated. As I do not know what method is better (sticky buffers vs AppLayerTxMatch), I'm just asking for some guidance here on which one to choose :)

victorjulien: I think there can be multiple ways: one would be with how the buffer is presented to the detection engine. Maybe it's possible to simply present it as a whole.

The dns_query approach seems best still, even if it's a bit more complex. It will perform much much better than any AppLayerTxMatch based approach.

Another thing to keep in mind is what we now have a transforms api. So perhaps a new transform could be used to take an input buffer and turn it into what you want.

The inspection buffer memory is managed for you, so no need to free things.

chifflier:
Updated by #3380

PR #3346, opened 2018-04-19
detect-engine: make output configurable (bug #2493)

dbcfd:
Detect engine has a new rules analysis method that is outputting analysis as json to a file. This
output is not currently configurable, so checking existing config settings to see if rules analysis
should be output.

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

https://redmine.openinfosecfoundation.org/issues/2493https://redmine.openinfosecfoundation.org/issues/2493

Describe changes:
- Use config to determine if rules analysis should be output

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #3375, opened 2018-05-11
Moved EngineAnalysisRules2 w/EngineAnalysisRules

smithjacobj:

- [**N/A**] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2493

Describe changes:
-EngineAnalysisRules2 was in a strange location where it did not respect
the --engine-analysis flag. It has been moved to the same call location
as EngineAnalysisRules.

### DIFF
--- a/src/detect-engine-build.c
+++ b/src/detect-engine-build.c
@@ -1820,9 +1820,6 @@ static int SigMatchPrepare(DetectEngineCtx *de_ctx)
             SigMatch *sm = s->init_data->smlists[type];
             s->sm_arrays[type] = SigMatchList2DataArray(sm);
         }
-#ifdef HAVE_LIBJANSSON
-        EngineAnalysisRules2(de_ctx, s);
-#endif
         /* free lists. Ctx' are xferred to sm_arrays so won't get freed */
         uint32_t i;
         for (i = 0; i < s->init_data->smlists_array_size; i++) {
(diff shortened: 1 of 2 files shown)


dbcfd:
Here was the original PR: https://github.com/OISF/suricata/pull/3346. Probably unneeded at this point, unless we also want the new method to respect the config option.

I think this is a better approach than the config only approach, but would need @victorjulien to weigh in on why EngineAnalysisRules2 is where it is.

victorjulien:
I'm ok with this, the place it was now is more of a dev artifact. I would like to add support for having it be part of a normal run though, not only as a --engine-analysis run.

Right now the engine analysis is a special runmode on it's own. That has it's use, but I'd like an additional non-terminating analysis option too. I was sort of working towards that.

smithjacobj:
The additional stuff seems fine - though it would be preferable that it can also be disabled. Our use case was having issues triggering threat detection in antivirus because of the rule signatures being written out to the `rules.json` file.

dbcfd:
@victorjulien Is it possible to merge this one, and then make it available via an option? Or do we want to hold up merge to add the support for making it optional.

victorjulien:
I've merged this with a small commit message update. Thanks @smithjacobj. @dbcfd feel free to do a PR against master.

PR #3383, opened 2018-05-23
util-random: workaround getrandom unavailability

regit:
getrandom syscall availability is detected at runtime. So it is
possible that the build is done on a box that supports it but
the run is done on a system with no availability. So a workaround
solution is needed to fix this case.

Also we have seen some issue in docker environment where the build
is detecting getrandom but where it does not work at runtime. Result was
suricata being stuck in init phase.

For both reasons, the code is updated to have a call to a fallback
function if ever the getrandom call returns that the syscall is
not available.

https://redmine.openinfosecfoundation.org/issues/2505

Describe changes:
- workaround getrandom failure

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/404
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/186

### DIFF
--- a/src/util-random.c
+++ b/src/util-random.c
@@ -27,6 +27,33 @@
 #include "suricata-common.h"
 #include "util-random.h"
 
+#if defined(HAVE_CLOCK_GETTIME)
+
+static long int RandomGetClock(void)
+{
+    struct timespec ts;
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged in #3384, thanks!

PR #3390, opened 2018-06-04

maurizioabba:

It's necessary to change several primitive for Packet manipulation, to
unify the parameter "packet length" whenever we are before IP decoding.

Add tests before calling DecodeIPVX function to avoid a possible
integer overflow over the len parameter.

Describe changes:
- Change packet length to uint32_t
- Unify packet lengths for defrag utils and PacketCopyData
-

### DIFF
--- a/src/decode-erspan.c
+++ b/src/decode-erspan.c
@@ -43,7 +43,7 @@
  * \brief Function to decode ERSPAN packets
  */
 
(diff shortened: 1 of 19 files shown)


victorjulien:
I wonder why we have all the checks before calling DecodeIPV4 and DecodeIPV6 instead of moving the check inside those? Seems more robust against future changes and a smaller patch overall.

maurizioabba:
The fact is that IP packet length must be a UINT16 (as per IP header), therefore it would look inconsistent to have a parameter length of uint32. That's why all checks are performed before IP level

victorjulien:
Merged through #3446, thanks!

PR #3433, opened 2018-07-24
Added Dockerfile

monokal:

- [ Yes] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ Yes] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ Yes] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Describe changes:
- Added working Dockerfile for Suricata 4.0.5, along with entrypoint script and simple doc.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable): na


docker/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+    mv "/tmp/boost_$(echo $BOOST_VERSION | tr . _)/boost" /tmp/hyperscan/include/
+
+# Build Hyperscan.
+RUN cd /tmp/hyperscan/build && \
+    cmake -DBUILD_STATIC_AND_SHARED=1 ../ && \
+    make && \
+    make install
+
+# Download and extract Suricata source.
+RUN wget -O /tmp/suricata.tar.gz "https://www.openinfosecfoundation.org/download/suricata-${SURICATA_VERSION}.tar.gz" && \
```
regit: Looks to me it could be interesting to build from the source directory as the Dockerfile is hosted in the source tree.

docker/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+    tar xzvf /tmp/suricata.tar.gz -C /tmp
+
+# TODO: Verify tarball against signature:
+# "https://www.openinfosecfoundation.org/download/suricata-${SURICATA_VERSION}.tar.gz.sig"
+
+# Compile and install Suricata from source.
+RUN cd "/tmp/suricata-${SURICATA_VERSION}" && \
+    ./configure --enable-geopip --enable-lua && \
+    make && \
+    make install-full
```
regit: install-full will download the rules which is useless given the fact you are also installing and using suricata-update.

You could do `make install && make install-conf` instead.

docker/entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+suricata-update && suricata $@
```
regit: I did not find any edit of the suricata.yaml in suricata 4.0.x this means the classical structure of the ruleset is used (import of list of files present in the ET ruleset). So the `suricata.rules` file created by `suricata-update` is unused.

Things should change in 4.1.x as suricata.rules should be activated if you have suricata-update.

docker/README.md
```diff
@@ -0,0 +1,22 @@
...
+```bash
+docker build suricata:latest .
+```
+
+```bash
+docker run -d \
+    --name suricata \
+    --privileged \
+    --network host \
+    --cap-add NET_ADMIN \
```
regit: I'm a bit rusted on docker side but if you run in privileged mode do you really need to add the capnet ?

jmtaylor90:
I am not sure how much it matters but the official github for hyperscan moved to
https://github.com/intel/hyperscan

docker/Dockerfile
```diff
@@ -0,0 +1,110 @@
+FROM centos:latest
```
regit: Some other distributions are shipping hyperscan so this could result in a smaller docker image to use that instead of centos as the build prerequisites will be smaller.

docker/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+
+# The Suricata version to install. This should match the version in the download
+# URL. E.g: https://www.openinfosecfoundation.org/download/suricata-<VERSION>.tar.gz
+ENV SURICATA_VERSION '4.0.5'
```
victorjulien: you can use 'current' here and it will always pull the latest stable (suricata-current.tar.gz is kept in sync with last stable)

docker/Dockerfile
victorjulien: yes, like debian with backports which would bring both latest suricata and hyperscan versions.

In general I'm not quite sure why to pick one or the other as a base for docker. Guess image size is the main factor?

monokal:
Thanks all. I'll amend the PR shortly taking all of your feedback on-board. Initially this came from my external repo hence it not building from the source tree along with other bits.

docker/Dockerfile
monokal: Perhaps Ubuntu's new minimal image could be a good call. https://blog.ubuntu.com/2018/07/09/minimal-ubuntu-released

monokal:
Here we are guys, with all feedback taken in to account.

docker/Dockerfile
monokal: Change made to do so and pushed!

docker/Dockerfile
monokal: The issue with install-conf was that it doesn't add suricata to PATH, whereas install-full does. Not the end of the world obviously, could be added to PATH as an additional step in the Dockerfile.

docker/entrypoint.sh
monokal: Amended to add `-S /var/lib/suricata/rules/suricata.rules` in the entrypoint.

docker/README.md
monokal: Correct! I've amended the doc.

docker/Dockerfile
monokal: Thanks, I've amended the Dockerfile to build from the source tree now that it's part of the repo.

docker/Dockerfile
monokal: Changed to the latest minimal Ubuntu Image and pushed. All tested good.

monokal:
Any chance of reviewing this please @regit / @victorjulien? I'm having to keep my fork up to date currently.

dbcfd:
Does it make sense to add a .dockerignore (docs, benches, .github)?

Also since the dockerfile is being added to a docker directory, add the entrypoint script there rather than echo'ing it as part of the docker run?

Finally, any reason for ubuntu rather than a lighter distro?

monokal:
Requested changes were pushed 6 months ago...

monokal:
@victorjulien @regit ^

praseodym:
One disadvantage of the approach taken in this PR is that the resulting container image will still contain all build dependencies. I’ve recently created [a Dockerfile with a multi-stage build approach](
https://github.com/praseodym/docker-suricata/blob/master/Dockerfile) so that the final image is a lot smaller.

That Dockerfile does use a dist tarball, but it can be trivially adopted to use raw (git) sources instead. Or, it could just install a pre-built Debian package, but that would remove any compile-time flexibility (e.g. using other capture libraries) from the Docker build.

victorjulien:
@praseodym interesting, what command is used to create the image?

victorjulien:
I do agree that the image shouldn't have the build artifacts

praseodym:
> @praseodym interesting, what command is used to create the image?

@victorjulien: a regular `docker build -t suricata .` will suffice. The Dockerfile uses [a multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/): a stage with build dependencies builds Suricata and the resulting build artifacts are then copied onto a clean Debian image to create the final Docker image.

victorjulien:
I think for now we'll leave Dockerfiles outside of our tree.

PR #3442, opened 2018-08-03
eve/http: add tx->request_port_number as http_port

maurizioabba:
Add the port specified in the hostname (if any) to the http object in
eve. The port may be different from the dest_port used by the TCP flow.

https://redmine.openinfosecfoundation.org/issues/2562

Describe changes:
- add http_port to the http object with the port specified in the hostname (if any)
-
-

### DIFF
--- a/src/output-json-http.c
+++ b/src/output-json-http.c
@@ -202,6 +202,13 @@ static void JsonHttpLogJSONBasic(json_t *js, htp_tx_t *tx)
         }
     }
 
+    /* port */
+    if (tx->request_port_number >= 0)
+    {
+        json_object_set_new(js, "http_port",
+                json_integer(tx->request_port_number));
+    }
+
     /* uri */
     if (tx->request_uri != NULL)
     {
--- a/doc/userguide/output/eve/eve-json-format.rst
+++ b/doc/userguide/output/eve/eve-json-format.rst
@@ -180,6 +180,21 @@ Event with non-extended logging:
       "http_content_type": "application\/x-gzip"
   }
 
+In case the hostname shows a port number (such as in case the header "Hostname: www.digip.org:1337") is present:
+
+::
+
+
+  "http": {
+      "http_port": 1337,
+      "hostname": "www.digip.org",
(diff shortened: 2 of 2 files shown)


src/output-json-http.c
```diff
@@ -202,6 +202,13 @@ static void JsonHttpLogJSONBasic(json_t *js, htp_tx_t *tx)
         }
     }
 
+    /* port */
+    if (tx->request_port_number >= 0)
```
victorjulien: is this only set if different? If not, does it make sense to compare it against the flow dest port and only log it if different?

maurizioabba: not necessarily, this is set only if the hostname header has a port in it, or if the url has it. The idea behind the patch is to log an information that otherwise will be lost. I can foresee several cases where tcp dest  port != http port but request is "normal", ex http proxy. I'm not sure whether we are ok to lose the information "hostname has a port" because the tcp port coincide with this value.

victorjulien:
On 03-08-18 16:50, Maurizio Abba wrote:
> *@maurizioabba* commented on this pull request.
> 
> 
> In src/output-json-http.c
> <https://github.com/OISF/suricata/pull/3442#discussion_r207568197>:
> 
>> @@ -202,6 +202,13 @@ static void JsonHttpLogJSONBasic(json_t *js, htp_tx_t *tx)
>          }
>      }
>  
> +    /* port */
> +    if (tx->request_port_number >= 0)
> 
> not necessarily, this is set only if the hostname header has a port in
> it, or if the url has it. The idea behind the patch is to log an
> information that otherwise will be lost. I can foresee several cases
> where tcp dest port != http port but request is "normal", ex http proxy.
> I'm not sure whether we are ok to lose the information "hostname has a
> port" because the tcp port coincide with this value.

Right, if it's only added when the port is in the uri/hostname I'm all
for logging it. Can you add a comment stating this?

maurizioabba:
of course. Also, I'll wait to merge this before working on fix-2563 (http headers dump) so that I won't have to adjust the offsets too many time (they patch the same file).

Replaced by https://github.com/OISF/suricata/pull/3444

PR #3459, opened 2018-08-22
problem: 2nd key in bpf_map_get_next_key usage is the current one.

JustinAzoff:
The 2nd key passed to bpf_map_get_next_key is the one that contains
the current key in the iteration.

If the map only contains a single key, after one call to
bpf_map_get_next_key key is undefined and next_key is the first item.
doing key = next_key and calling bpf_map_get_next_key ends the loop
because there are no more keys and bpf_map_lookup_elem is never called
with the first key.  I believe in maps containing more than one key this
will just result in the first key being skipped.

linux/samples/bpf/sampleip_user.c is one of the only uses of
bpf_map_get_next_key I can find, and they do:

    key = 0, i = 0;
    while (bpf_map_get_next_key(fd, &key, &next_key) == 0) {
            bpf_map_lookup_elem(fd, &next_key, &value);
            counts[i].ip = next_key;
            counts[i++].count = value;
            key = next_key;
    }

Note how bpf_map_lookup_elem is called with next_key, not key.

To avoid having to change a ton of references to key, I just renamed key
to prev_key and nextkey to key.

- [ ] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/

Describe changes:
-
-
-

### DIFF
--- a/src/util-ebpf.c
+++ b/src/util-ebpf.c
@@ -391,7 +391,7 @@ static int EBPFForEachFlowV4Table(LiveDevice *dev, const char *name,
                                   struct timespec *ctime)
 {
     int mapfd = EBPFGetMapFDByName(dev->dev, name);
-    struct flowv4_keys key = {}, next_key;
+    struct flowv4_keys prev_key = {}, key;
     int found = 0;
     unsigned int i;
     unsigned int nr_cpus = UtilCpuGetNumProcessorsConfigured();
@@ -401,7 +401,7 @@ static int EBPFForEachFlowV4Table(LiveDevice *dev, const char *name,
     }
 
     uint64_t hash_cnt = 0;
-    while (bpf_map_get_next_key(mapfd, &key, &next_key) == 0) {
+    while (bpf_map_get_next_key(mapfd, &prev_key, &key) == 0) {
         bool purge = true;
         uint64_t pkts_cnt = 0;
         uint64_t bytes_cnt = 0;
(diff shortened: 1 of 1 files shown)


regit:
Good catch and thanks for this PR. Code looks ok but there is some issues.

First of all can you sign the contribution agreement ? https://suricata-ids.org/about/contribution-agreement/

Second the commit message is not conform to what we expect. See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits for more information. In the case of your commit, you should have a first line with something like `util-ebpf: fix iteration on map` or something alike.

JustinAzoff:
Gotcha.. I'll work on those things.

I believe I have found another bug related to how that code works.. It's really subtle and you'd only notice it if you are outputting some counters tracking how many total loop iterations you did.  I was noticing I was seeing 38 loop iterations for only 20 flows.

When iterating using `bpf_map_get_next_key`, if you delete the current key using `bpf_map_delete_elem`, then that key is no longer valid for the next call to `next_key`.  The result is that the iteration starts over again at the beginning.  The end result is eventually the same, but it will turn expiring N flows out of M total into an `O(N*M)` operation.

In my library I fixed this by doing

```
int bpf_map_get_next_key_and_delete(int fd, const void *key, void *next_key, int *delete)
{
    int res = bpf_map_get_next_key(fd, key, next_key);
    if (*delete) {
        bpf_map_delete_elem(fd, key);
        *delete = 0;
    }
    return res;
}
```
and doing

```
    int delete_prev_key = 0;
    while (bpf_map_get_next_key_and_delete(v4_fd, &prev_key, &key, &delete_prev_key) == 0) {
        ...
        if (expire) {
            ....
            delete_prev_key=1;
        }
    }

PR #3462, opened 2018-08-27
Add support for TLSv1.3

thus:
This PR contains various stuff that was fixed on the path to get TLSv1.3 support:
- Add support for TLSv1.3 (including all the drafts)
- Add support for session resumption by both session ID and session ticket
- Add 'raw' matching mode to "tls.version" keyword (e.g: "tls.version:0x7f10")

https://redmine.openinfosecfoundation.org/issues/2279

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/163
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/163

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -635,50 +635,54 @@ static inline int TLSDecodeHSHelloCipherSuites(SSLState *ssl_state,
     if (!(HAS_SPACE(2)))
         goto invalid_length;
 
-    uint16_t cipher_suites_length = *input << 8 | *(input + 1);
-    input += 2;
-
-    if (!(HAS_SPACE(cipher_suites_length)))
-        goto invalid_length;
-
-    if ((ssl_state->current_flags & SSL_AL_FLAG_STATE_CLIENT_HELLO) &&
-            ssl_config.enable_ja3) {
-        int rc;
-
-        JA3Buffer *ja3_cipher_suites = Ja3BufferInit();
-        if (ja3_cipher_suites == NULL)
-            return -1;
-
-        uint16_t processed_len = 0;
-        /* coverity[tainted_data] */
-        while (processed_len < cipher_suites_length)
-        {
-            if (!(HAS_SPACE(2))) {
-                Ja3BufferFree(&ja3_cipher_suites);
-                goto invalid_length;
-            }
-
-            uint16_t cipher_suite = *input << 8 | *(input + 1);
-            input += 2;
-
-            if (TLSDecodeValueIsGREASE(cipher_suite) != 1) {
-                rc = Ja3BufferAddValue(&ja3_cipher_suites, cipher_suite);
-                if (rc != 0) {
-                    return -1;
-                }
-            }
-
-            processed_len += 2;
-        }
-
-        rc = Ja3BufferAppendBuffer(&ssl_state->ja3_str, &ja3_cipher_suites);
-        if (rc == -1) {
-            return -1;
-        }
-
+    if (ssl_state->current_flags & SSL_AL_FLAG_STATE_SERVER_HELLO) {
+        /* Skip cipher suite */
+        input += 2;
     } else {
-        /* Skip cipher suites */
-        input += cipher_suites_length;
+        uint16_t cipher_suites_length = *input << 8 | *(input + 1);
+        input += 2;
+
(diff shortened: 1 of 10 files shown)


src/app-layer-ssl.c
```diff
@@ -818,6 +819,56 @@ static inline int TLSDecodeHSHelloExtensionSni(SSLState *ssl_state,
...
+            goto invalid_length;
+
+        uint8_t supported_ver_len = *input;
+        input += 1;
+
+        if (!(HAS_SPACE(supported_ver_len)))
+            goto invalid_length;
+
+        /* Use the first (and prefered) version as client version */
+        ssl_state->curr_connp->version = *input << 8 | *(input + 1);
```
victorjulien: should we record all advertised versions here for logging and also perhaps validate that the top one is in fact the best version?

src/util-lua-tls.c
```diff
@@ -179,6 +180,51 @@ static int GetCertInfo(lua_State *luastate, const Flow *f, int direction)
         case TLS_VERSION_12:
             snprintf(ssl_version, sizeof(ssl_version), "TLS 1.2");
             break;
+        case TLS_VERSION_13:
```
victorjulien: we seem to have this number to string conversion 3 times, can you see if this can be consolidated into a single util func?

src/detect-ssl-version.c
```diff
@@ -239,6 +261,10 @@ static DetectSslVersionData *DetectSslVersionParse(const char *str)
                 ssl->data[TLS12].ver = TLS_VERSION_12;
                 if (neg == 1)
                     ssl->data[TLS12].flags |= DETECT_SSL_VERSION_NEGATED;
+            } else if (strncasecmp("tls1.3", tmp_str, 6) == 0) {
```
victorjulien: strNcasecmp seems wrong here, it would match "tls1.333333" as well. Could you update them all?

src/detect-tls-version.c
```diff
@@ -185,14 +187,17 @@ static DetectTlsVersionData *DetectTlsVersionParse (const char *str)
             tmp_str += 1;
         }
 
-        if (strcmp("1.0", tmp_str) == 0) {
+        if (strncmp("1.0", tmp_str, 3) == 0) {
```
victorjulien: strNcmp seems wrong here. This would match 1.0.1 as well.

src/app-layer-ssl.h
```diff
@@ -166,6 +166,9 @@ typedef struct SSLStateConnp_ {
     /* ssl server name indication extension */
     char *sni;
 
+    char *session_id;
+    uint8_t ssl3_session_id_length;
```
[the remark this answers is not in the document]
thus: No, but it was meant as a "ssl3" style client hello record, but I'll change it to something more appropriate.

src/app-layer-ssl.c
thus: The top one might not be the best TLS version, but it is the one that the client wants most to use.

Would it make sense to do the work of storing all advertised versions for logging purposes in another PR? I mean, since we don't store the client version now.

src/util-lua-tls.c
thus: I actually thought of this myself, but I was hoping that you wouldn't mention it, haha :P

src/detect-ssl-version.c
thus: Would checking the string length be a sane thing to do instead?

src/detect-tls-version.c
thus: Same question here as above :)

src/app-layer-ssl.c
victorjulien: Yeah that sounds good.

src/detect-ssl-version.c
victorjulien: These are all nul terminated strings, so it shouldn't be necessary to explicitly check the length.

thus: My mind was thinking of strstr not on strcmp (which does exact match), so yeah, using strncasecmp does not make sense at all! I'll change to strcasecmp.

thus:
Replaced by #3464

PR #3467, opened 2018-08-30
Bug fix: Added livedev on flow and set vlan and iface in stream-tcp and flow-timeout

bolzzy:
Proposed bug fix for bug [#2057](https://redmine.openinfosecfoundation.org/issues/2057): 
Adding livedev and flow on psuedopackets in stream-tcp and flow-timeout.

Describe changes:
- Added and setting livedev on flow
- Setting VLAN and interface from flow to psuedopackets in stream-tcp and flow-timeout

This should fix the problem dealing with missing vlan and in_iface fields in logs caused by psuedo packets missing vlan and iface.

I have tested on some live traffic and it seems to work out fine, but I'm not 100% sure it does not affect corner cases with wierd flows etc.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable): N/A

### DIFF
--- a/src/flow-timeout.c
+++ b/src/flow-timeout.c
@@ -90,6 +90,18 @@ static inline Packet *FlowForceReassemblyPseudoPacketSetup(Packet *p,
     p->flags |= PKT_HAS_FLOW;
     p->flags |= PKT_PSEUDO_STREAM_END;
 
+    if (f->vlan_id[0] > 0) {
+        p->vlan_id[0] = f->vlan_id[0];
+        p->vlan_idx=1;
+
+        if (f->vlan_id[1] > 0) {
+            p->vlan_id[1] = f->vlan_id[1];
+            p->vlan_idx=2;
+        }
+    }
+
+    p->livedev = f->livedev;
+
     if (f->flags & FLOW_NOPACKET_INSPECTION) {
         DecodeSetNoPacketInspectionFlag(p);
     }
--- a/src/flow-util.c
+++ b/src/flow-util.c
@@ -153,6 +153,7 @@ void FlowInit(Flow *f, const Packet *p)
     f->recursion_level = p->recursion_level;
     f->vlan_id[0] = p->vlan_id[0];
     f->vlan_id[1] = p->vlan_id[1];
+    f->livedev = p->livedev;
 
     if (PKT_IS_IPV4(p)) {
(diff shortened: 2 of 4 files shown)


src/flow-timeout.c
```diff
@@ -90,6 +90,18 @@ static inline Packet *FlowForceReassemblyPseudoPacketSetup(Packet *p,
     p->flags |= PKT_HAS_FLOW;
     p->flags |= PKT_PSEUDO_STREAM_END;
 
+    if (f->vlan_id[0] > 0) {
```
victorjulien: while vlan 0 is officially reserved, it seems it is actually used. So I don't think we should assume this fields value being 0 to mean there is no vlan.

victorjulien: the value uses just 11 bits, so we could use one of the other bits to indicate the field has been set.

src/flow-util.c
```diff
@@ -153,6 +153,7 @@ void FlowInit(Flow *f, const Packet *p)
     f->recursion_level = p->recursion_level;
     f->vlan_id[0] = p->vlan_id[0];
     f->vlan_id[1] = p->vlan_id[1];
+    f->livedev = p->livedev;
```
victorjulien: @regit can a livedev ever disappear at runtime?

src/flow.h
```diff
@@ -345,6 +345,10 @@ typedef struct Flow_
     uint8_t recursion_level;
     uint16_t vlan_id[2];
 
+    /** Incoming interface */
+    struct LiveDevice_ *livedev;
```
victorjulien: this ptr should be reset in FLOW_RECYCLE and FLOW_INITIALIZE

victorjulien:
Commit message formatting should follow our guidelines: https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria#Commits

Please also explain the why of the change, and what effect the change has.

src/flow-timeout.c
bolzzy: That does not work as flow does not currently have vlan_idx.
But checking if the value is set ( "if vlan_id[0] " ) should still be sufficient as this also supports vlan 0, or?

regit: I think you need to add it or we could missed to set and log vlan 0 on packet. Or at reverse log a vlan 0 that is not real.

PR #3469, opened 2018-08-31
RFC - template: add gap handling flow - v1

jasonish:
The idea here is to better describe what might be received as input to the parse function. As far as I can tell, we should either be EOF, GAP or we should have data.

Thoughts?

### DIFF
--- a/src/app-layer-template.c
+++ b/src/app-layer-template.c
@@ -215,16 +215,20 @@ static int TemplateParseRequest(Flow *f, void *state,
 
     SCLogNotice("Parsing echo request: len=%"PRIu32, input_len);
 
-    /* Likely connection closed, we can just return here. */
-    if ((input == NULL || input_len == 0) &&
-        AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF)) {
-        return 0;
-    }
-
-    /* Probably don't want to create a transaction in this case
-     * either. */
-    if (input == NULL || input_len == 0) {
-        return 0;
+    if (input == NULL) {
+        if (AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF)) {
+            /* This is a signal that the stream is done. Do any
+             * cleanup if needed. Usually nothing is required here. */
+            return 0;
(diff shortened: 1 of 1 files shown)


src/app-layer-template.c
```diff
@@ -215,16 +215,20 @@ static int TemplateParseRequest(Flow *f, void *state,
...
+    if (input == NULL) {
+        if (AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF)) {
+            /* This is a signal that the stream is done. Do any
+             * cleanup if needed. Usually nothing is required here. */
+            return 0;
+        } else if (flags & STREAM_GAP) {
+            /* This is a signal that there has been a gap in the
+             * stream. This only needs to be handled if gaps were
+             * enabled during protocol registration. The input_len
+             * contain the size of the gap. */
```
victorjulien: typo: s/contain/contains/g

victorjulien:
We can be passed the STREAM_MIDSTREAM flag too I think. This would indicate we're not sure we are starting at the start of the stream. But input would have data.

victorjulien:
Think it would be good to refresh and finish this.

src/app-layer-template.c
```diff
@@ -215,16 +215,20 @@ static int TemplateParseRequest(Flow *f, void *state,
...
+            return 0;
+        }
+        /* This should not happen. */
+        return -1;
```
catenacyber: Use `BUG_ON`here ?

catenacyber:
> we should either be EOF, GAP or we should have data.

Seems good to me.
We can have also a flag `STREAM_START` but I am not sure it can make sense

src/app-layer-template.c
victorjulien: `DEBUG_VALIDATION_BUG_ON` before `return -1`, so that it is handled as a protocol error in the normal case, but with `--enable-debug-validation` it `abort()`s.

victorjulien:
@jasonish can you rebase this?

jasonish:
Continued at https://github.com/OISF/suricata/pull/5184.

PR #3473, opened 2018-09-10
Optimization/defrag rbtree/v1

jasonish:
Use an RB tree to store fragments. Improves performance of IP fragment re-assembly.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/314
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/667

### DIFF
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -100,6 +100,8 @@ static int default_policy = DEFRAG_POLICY_BSD;
  * context. */
 static DefragContext *defrag_context;
 
+RB_GENERATE(IP_FRAGMENTS, Frag_, rb, DefragRbFragCompare);
+
 /**
  * Utility/debugging function to dump the frags associated with a
  * tracker.  Only enable when unit tests are enabled.
@@ -149,15 +151,15 @@ DefragFragInit(void *data, void *initdata)
 void
 DefragTrackerFreeFrags(DefragTracker *tracker)
 {
-    Frag *frag;
+    Frag *frag, *next;
 
     /* Lock the frag pool as we'll be return items to it. */
     SCMutexLock(&defrag_context->frag_pool_lock);
 
-    while ((frag = TAILQ_FIRST(&tracker->frags)) != NULL) {
-        TAILQ_REMOVE(&tracker->frags, frag, next);
-
-        /* Don't SCFree the frag, just give it back to its pool. */
+    for (frag = RB_MIN(IP_FRAGMENTS, &tracker->fragment_tree);
+         frag != NULL; frag = next) {
+        next = RB_NEXT(IP_FRAGMENTS, &tracker->fragment_tree, frag);
+        RB_REMOVE(IP_FRAGMENTS, &tracker->fragment_tree, frag);
         DefragFragReset(frag);
         PoolReturn(defrag_context->frag_pool, frag);
     }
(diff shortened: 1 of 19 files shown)


src/defrag.c
```diff
@@ -604,15 +612,29 @@ DefragInsertFrag(ThreadVars *tv, DecodeThreadVars *dtv, DefragTracker *tracker,
...
+            prev = RB_MIN(IP_FRAGMENTS, &tracker->fragment_tree);
+            next = IP_FRAGMENTS_RB_NEXT(prev);
+        } else {
+            prev = IP_FRAGMENTS_RB_PREV(next);
+            if (prev == NULL) {
+                prev = next;
+                next = IP_FRAGMENTS_RB_NEXT(prev);
+            }
+        }
+        while (prev != NULL) {
```
victorjulien: I think this could be much simplified. In the TCP case I've made sure that the compare func primarily looks at the seq (offset here) and secondary at the length. This means that I can simply RB_INSERT. Only if exactly equal it rejects the insert. In other cases, you can take the inserted frag, check RB_PREV to see if that overlaps us and check RB_NEXT to see if we overlap that. This way no looping is required.

victorjulien: Whats the case for having exact duplicates in the tree?

jasonish: In the "LAST" policy, the data received last takes precedence. So it may mean removing a fragment that matches it in offset and length. We don't actually remove it from the tree, it just gets flagged as skipped and the new fragment inserted.

I think the insert could be simplified by moving some of this logic into the compare function, but that would require relaying the personality through, and moving some of the logic. I didn't want to go that far yet.

victorjulien: But in this case it would be enough to remove the "old" duplicate and replace it by the new frag for that exact offset+len right?

jasonish: It could. Not saying it couldn't be done, but this would change the lookup based on what personality is being used? Something I was trying hard to avoid for this run - not touching the personality code.

With "last", if only a portion overlaps, we remove that data from the prior packet. So this lookup catches both of those cases with having the personality code in one section.

victorjulien: I'm fine with keeping it this way for now.

Wondering though, can we currently insert large amounts of duplicate frags?

victorjulien: (currently as in this PR)

src/defrag.c
```diff
@@ -388,11 +380,13 @@ Defrag6Reassemble(ThreadVars *tv, DefragTracker *tracker, Packet *p)
      * fragments are inserted if frag_offset order. */
     Frag *frag;
     int len = 0;
-    TAILQ_FOREACH(frag, &tracker->frags, next) {
+    Frag *first = RB_MIN(IP_FRAGMENTS, &tracker->fragment_tree);
```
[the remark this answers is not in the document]
jasonish: Fixed.

src/defrag.c
```diff
@@ -149,15 +151,15 @@ DefragFragInit(void *data, void *initdata)
...
+    Frag *frag, *next;
 
     /* Lock the frag pool as we'll be return items to it. */
     SCMutexLock(&defrag_context->frag_pool_lock);
 
-    while ((frag = TAILQ_FIRST(&tracker->frags)) != NULL) {
-        TAILQ_REMOVE(&tracker->frags, frag, next);
-
-        /* Don't SCFree the frag, just give it back to its pool. */
+    for (frag = RB_MIN(IP_FRAGMENTS, &tracker->fragment_tree);
```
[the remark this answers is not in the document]
jasonish: Changed to RB_FOREACH_SAFE.

src/defrag.c
jasonish: Yes, with the last policy this would be possible. I have a fix for this now as well. When inserting a new fragment, if the previous has been marked as skip, or has been trimmed to the point of being remove, remove it from the tree then instead of hanging onto it.

jasonish:
Closing. See https://github.com/OISF/suricata/pull/3474

PR #3481, opened 2018-09-18
wip - rust - templates and scripts - v1

jasonish:
Changes:
- C stub app-layer-template-rust for Rust based parsers.
- C stub output-json-template-rust for Rust based parsers.
- Replace setup-app-layer.sh and setup-app-layer-logger.sh with setup-app-layer.py.

And the start of an app-layer written in Rust. I was thinking of a protocol that had length prefixed messages like "\x0cHello World!". But given we want to turn this into Gopher, maybe something like "12:Hello World!" would be better?

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/321
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/674

PR #3484, opened 2018-09-18
rust - app-layer templates and scripts - v2

jasonish:
Previous PR:
https://github.com/OISF/suricata/pull/3481

Changes:
- Complete Rust parser and logger for fictitious request/reply protocol.
- Sample pcap (may want to remove)
- Rewrite of setup-app-layer.sh and setup-app-layer-logger.sh in Python as a combined script.
- Reference Rust source when --rust is used.

I changed the dummy protocol to be a little closer to what gopher might be.. Its ascii prefixed messages, where the prefix is the length of the message.

Request  -> 12:Hello World!
Response -> 3:Bye

Its easier to use netcat to experiment with.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/323
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/677


rust/src/applayertemplate/mod.rs
```diff
@@ -0,0 +1,518 @@
...
+    _flow: *const Flow,
+    state: *mut libc::c_void,
+    _pstate: *mut libc::c_void,
+    input: *const libc::uint8_t,
+    input_len: u32,
+    _data: *const libc::c_void,
+    _flags: u8,
+) -> i8 {
+    let _eof = unsafe {
+        if AppLayerParserStateIssetFlag(_pstate, APP_LAYER_PARSER_EOF) > 0 {
```
victorjulien: _pstate is a bit ugly here, the leading _ should mean it's not used?

rust/src/applayertemplate/mod.rs
```diff
@@ -0,0 +1,518 @@
...
+
+        while current.len() > 0 {
+            match parser::parse_message(current) {
+                nom::IResult::Done(rem, response) => {
+                    current = rem;
+
+                    match self.find_request() {
+                        Some(tx) => {
+                            tx.response = Some(response);
+                            println!("Found response for request:");
```
victorjulien: should we use SCLog* here?

rust/src/applayertemplate/mod.rs
```diff
@@ -0,0 +1,518 @@
...
+    _flow: *const Flow,
+    state: *mut libc::c_void,
+    _pstate: *mut libc::c_void,
+    input: *const libc::uint8_t,
+    input_len: u32,
+    _data: *const libc::c_void,
+    _flags: u8,
+) -> i8 {
+    let _eof = unsafe {
+        if AppLayerParserStateIssetFlag(_pstate, APP_LAYER_PARSER_EOF) > 0 {
```
victorjulien: _pstate

src/app-layer-template-rust.c
```diff
@@ -0,0 +1,74 @@
+/* Copyright (C) 2015 Open Information Security Foundation
```
victorjulien: year

src/app-layer-template-rust.h
```diff
@@ -0,0 +1,30 @@
+/* Copyright (C) 2015 Open Information Security Foundation
```
victorjulien: year

rust/src/applayertemplate/logger.rs
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use libc;
+use std;
+use json::*;
+use super::TemplateTransaction;
```
victorjulien: what does this do?

rust/src/applayertemplate/mod.rs
jasonish: Oops. Renamed to pstate. Also using eof, but only like
```
diff --git a/rust/src/applayertemplate/mod.rs b/rust/src/applayertemplate/mod.rs
index 9daa161b5..5bb6705a5 100644
--- a/rust/src/applayertemplate/mod.rs
+++ b/rust/src/applayertemplate/mod.rs
@@ -300,19 +300,24 @@ pub extern "C" fn rs_template_state_tx_free(
 pub extern "C" fn rs_template_parse_request(
     _flow: *const Flow,
     state: *mut libc::c_void,
-    _pstate: *mut libc::c_void,
+    pstate: *mut libc::c_void,
     input: *const libc::uint8_t,
     input_len: u32,
     _data: *const libc::c_void,
     _flags: u8,
 ) -> i8 {
-    let _eof = unsafe {
-        if AppLayerParserStateIssetFlag(_pstate, APP_LAYER_PARSER_EOF) > 0 {
+    let eof = unsafe {
+        if AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF) > 0 {
             true
         } else {
             false
         }
     };
+
+    if eof {
+    }
+
     let state = cast_pointer!(state, TemplateState);
     if state.parse_request(buf) {
```

The idea with the template is to just show how to test for EOF. Do you think we should do anything else with it here?

rust/src/applayertemplate/mod.rs
jasonish: Thought about it? But I had every intention that the user would remove that stuff there and not convert to SCLogDebug.  I'm indifferent about it though.

rust/src/applayertemplate/logger.rs
jasonish: This pulls TemplateTransaction into scope from mod.rs.
 
Something more general...  We have been doing stuff like "dhcp/mod.rs", "dhcp/dhcp.rs". It seems idiomatic in Rust to put the stuff that you might put into a file named the same as the module, into mod.rs itself. So I went with that approach there. Even if I had used templater.rs, we'd have something like 'super::template::TemplateTransaction' or '::applayertemplate::template::TemplateTransaction' where "super" is the preferred way.

rust/src/applayertemplate/mod.rs
victorjulien: Not at this time I think. It's more used to close things like files.

rust/src/applayertemplate/mod.rs
victorjulien: I prefer SCLog, probably SCLogNotice in the template that the dev can change to Debug

rust/src/applayertemplate/logger.rs
victorjulien: Hmm not sure I'm a fan on this. I actually like the separation of mod.rs and actual content in other parsers. Anyway we can defer this to later.

jasonish:
Closing. See https://github.com/OISF/suricata/pull/3485.
Addresses issues mentioned here.

PR #3491, opened 2018-09-27
training - move rust detect to v2 and add mpm - v2

jasonish:
Based on ams_2018/v2.


victorjulien:
This seems still relevant, can you do a new PR against master?

PR #3494, opened 2018-10-03
Flow/stream-tcp: bug 2057 fix missing vlan and iface on psuedopacket v2

bolzzy:
Ref old PR: https://github.com/OISF/suricata/pull/3467

Describe changes:
This should fix the bug related to missing vlan and in_iface on logged/alerted psuedopackets. Without vlan and iface on output / alert of pseudopackets we are unable to split the output to the correct SIEM.
I have not noticed any issues related to this patch on sensors in IDS mode with live traffic.

### DIFF
--- a/src/flow-timeout.c
+++ b/src/flow-timeout.c
@@ -89,6 +89,16 @@ static inline Packet *FlowForceReassemblyPseudoPacketSetup(Packet *p,
     p->flags |= PKT_HAS_FLOW;
     p->flags |= PKT_PSEUDO_STREAM_END;
 
+    if (f->vlan_id[0]) {
+        p->vlan_id[0] = f->vlan_id[0];
+        p->vlan_idx=1;
+
+        if (f->vlan_id[1]) {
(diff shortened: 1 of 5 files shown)


src/flow-timeout.c
```diff
@@ -88,6 +88,17 @@ static inline Packet *FlowForceReassemblyPseudoPacketSetup(Packet *p,
     p->flags |= PKT_STREAM_EOF;
     p->flags |= PKT_HAS_FLOW;
     p->flags |= PKT_PSEUDO_STREAM_END;
+    p->livedev = f->livedev;
+
+    if (f->vlan_id[0]) {
```
praseodym: This won’t work for vlan 0, which is why an explicit vlan index is used in other places. #3624 adds that index to the flow struct and I think it would be better to copy it back to the pseudo packet as well.

bolzzy: You are right, fixing this in new PR. Thanks!

bolzzy:
updated fix in #3636 , closing.

PR #3518, opened 2018-10-21
Fix two JA3 bugs

thus:
Fix JA3 bug that occurs when decoding empty extensions in ClientHello packet. Also fix JA3 bug that occurs when encountering earlier drafts of TLSv1.3.

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/171
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/171


victorjulien:
Merged in #3520, thanks Mats!

PR #3530, opened 2018-10-28
app-layer-ssl: add support for early data in TLSv1.3

thus:
This pull request does the following three things:
- It stops decoding from breaking when encountering TLSv1.3 0-RTT.
- Fixes flow and inspection bypass for TLSv1.3.
- Handles all version numbers above TLSv1.2 as TLSv1.3.

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/172
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/172

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -2197,9 +2197,13 @@ static int SSLv3Decode(uint8_t direction, SSLState *ssl_state,
                 }
             }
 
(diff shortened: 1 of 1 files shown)


thus:
Replaced by #3532

PR #3533, opened 2018-10-29
userguide: add documentation for ssl_version keyword

thus:
Add some documentation for the ssl_version keyword to the TLS-keywords section in the Userguide.

https://redmine.openinfosecfoundation.org/issues/2592


victorjulien:
Merged in #3536, thanks Mats!

PR #3535, opened 2018-10-30
app-layer-ssl: log facebook custom TLSv1.3 draft versions

thus:
Since we know what they are, we should log Facebook's custom TLSv1.3 draft implementations (https://github.com/facebookincubator/fizz) as strings and not just log the version number as hex (e.g 0xfb1a).

This PR also change the logging format of the existing TLSv1.3 drafts. This means that they are logged as "TLS 1.3 draft-26" instead of "TLS 1.3 (draft 26)". 

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/174
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/174


victorjulien:
Merged in #3536, thanks Mats!

PR #3540, opened 2018-11-06
source-nfq: support queue range and increase maximum number of queues to 128

gozzy:

Describe changes:
- increase maximum number of queues from 16 to 128


norg:
why 128? I guess this will be enough for most cases but when we change it we might even go to the max value from NFQUEUE Itself? (65535)

gozzy:
@norg yes, I have thought about 65535 but supposed that it will be too much. Perhaps we should really allow to use as much qeues as possible.

PR #3550, opened 2018-11-19
Nfq queue range v2

gozzy:

Describe changes:

This replaces #3540.

### DIFF
--- a/src/source-nfq.h
+++ b/src/source-nfq.h
@@ -89,7 +89,8 @@ typedef struct NFQGlobalVars_
 } NFQGlobalVars;
 
 void NFQInitConfig(char quiet);
-int NFQRegisterQueue(char *queue);
+int NFQRegisterQueue(const char *queue);
+int NFQParseAndRegisterQueues(const char *queues);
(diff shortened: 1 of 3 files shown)


src/source-nfq.h
```diff
@@ -30,7 +30,8 @@
 #include <linux/netfilter.h>		/* for NF_ACCEPT */
 #include <libnetfilter_queue/libnetfilter_queue.h>
 
-#define NFQ_MAX_QUEUE 16
+// Netfilter's limit
+#define NFQ_MAX_QUEUE 65535
```
victorjulien: The reason why this number is so low is that it's used in the definition of a few data structures, like g_nfq_t and g_nfq_q. Esp the last one is quite big: 120 bytes (on 64bit) * NFQ_MAX_QUEUE. I think this is excessive in the case of 65535.

Ideally this whole logic would be done differently, where the data structures are allocated on demand and not in these big arrays.

gozzy: Ok, I will try to find a way to handle this.

PR #3556, opened 2018-11-25
suricata-update: configures fixups - v1

jasonish:
1) Do not install suricata-update if python-yaml is not found. A warning message, similar to how we handle other dependencies will be displayed telling the user how to install it.
2) Allow --disable-suricata-update to keep suricata-update from being installed even if all dependencies are met.
3) Print datarootdir in build-info. This is useful as its where the enging provided rules are now installed.
4) Include a script for bundling suricata-update, I think its useful to have something like this in-tree for others who want to do their own builds and bundle it correctly.

Related redmine issues:
- https://redmine.openinfosecfoundation.org/issues/2705

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/330
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/684

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ qa/log/
 
 /suricata-update/*
 !/suricata-update/Makefile.am
+!/suricata-update/bundle.sh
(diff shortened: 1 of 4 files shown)


victorjulien:
Merged in #3563, thanks Jason. I did drop the bundle.sh patch for now as I didn't really see the point of adding it a point release and in general would like to discuss how to handle utils like that.

PR #3569, opened 2018-12-05
rust/dns: add v1 dns logging

jasonish:
Enabled v1 json dns logging for Rust enabled builds.

Previous PR:
https://github.com/OISF/suricata/pull/3566

Changes from last PR:
- Remove debug log line (logged as notice).

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2704

Suricata-Verify has been updated to test v1 json dns logging with Rust enabled builds:
- https://github.com/OISF/suricata-verify/pull/3

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/334
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/688

PR #3572, opened 2018-12-06
rust/dns: add v1 dns logging - v3

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/3569

Changes from last PR:
- Fix Lua support functions to match that of C, allowing the dns-lua suricata-verify test to pass.

Redmine issues:
- https://redmine.openinfosecfoundation.org/issues/2730
- https://redmine.openinfosecfoundation.org/issues/2704

Suricata-Verify has been updated as well:
- https://github.com/OISF/suricata-verify/pull/4

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/335
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/689

### DIFF
--- a/rust/src/dns/log.rs
+++ b/rust/src/dns/log.rs
@@ -598,3 +598,122 @@ pub extern "C" fn rs_dns_log_json_answer(tx: &mut DNSTransaction,
 
     return std::ptr::null_mut();
 }
+
+// Version 1 logging support.
+
+fn dns_log_json_answer_v1(header: &DNSHeader, answer: &DNSAnswerEntry)
+                       -> Json
+{
+    let js = Json::object();
+
+    js.set_string("type", "answer");
+    js.set_integer("id", header.tx_id as u64);
+    js.set_string("flags", format!("{:x}", header.flags).as_str());
+    if header.flags & 0x8000 != 0 {
+        js.set_boolean("qr", true);
+    }
+    if header.flags & 0x0400 != 0 {
+        js.set_boolean("aa", true);
+    }
+    if header.flags & 0x0200 != 0 {
+        js.set_boolean("tc", true);
+    }
+    if header.flags & 0x0100 != 0 {
+        js.set_boolean("rd", true);
+    }
+    if header.flags & 0x0080 != 0 {
+        js.set_boolean("ra", true);
+    }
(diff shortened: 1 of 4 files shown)


rust/src/dns/lua.rs
```diff
@@ -93,11 +88,26 @@ pub extern "C" fn rs_dns_lua_get_query_table(clua: &mut CLuaState,
 
             lua.settable(-3);
         }
+    } else if let Some(response) = &tx.response {
```
victorjulien: is it possible to have both no request and no response? In this case returning 1 at the end looks suspicious. Though a table is created unconditionally.

rust/src/dns/log.rs
```diff
@@ -598,3 +598,122 @@ pub extern "C" fn rs_dns_log_json_answer(tx: &mut DNSTransaction,
...
+}
+
+#[no_mangle]
+pub extern "C" fn rs_dns_log_json_answer_v1(tx: &mut DNSTransaction,
+                                         i: libc::uint16_t,
+                                         flags: libc::uint64_t)
+                                         -> *mut JsonT
+{
+    let index = i as usize;
+    for response in &tx.response {
```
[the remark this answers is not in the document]
jasonish: Its on my todo list to remove using Options as iterators, but I didn't think an idiomatic change belonged in a 0.0.1 release in an effort to minimize the amount of code changed.  The v1 stuff was mostly a revert of a commit, so carried that back in with it. But its consistent with other stuff in dns/log.rs.  Happy to make the change though.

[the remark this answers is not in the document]
victorjulien: I prefer to have it fixed now before myself or anyone else spends time being confused again :)

rust/src/dns/lua.rs
jasonish: Yes, but this is for consistency with C, which always returns 1 and an empty table as well.

victorjulien: Ok. There should probably be a comment about this behavior in the code then.

jasonish:
Closing. See https://github.com/OISF/suricata/pull/3574 for update.

PR #3582, opened 2018-12-13
NFQ queue range v4

gozzy:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Describe changes:
- added ability to specify Netfilter queue range with '-q' option (similar to iptables '--queue-balance' option)
- increase maximum number of queues from 16 to 65535
- NFQ context are created dynamically instead of on-stack arrays

This replaces #3552.

PR #3594, opened 2018-12-20
Nfq queue range v5

gozzy:

Describe changes:

- NFQ contexts are created dynamically instead of on-stack arrays

This replaces #3582.

### DIFF
--- a/src/source-nfq.c
+++ b/src/source-nfq.c
@@ -48,6 +48,7 @@
 #include "util-debug.h"
 #include "util-error.h"
 #include "util-byte.h"
+#include "util-cpu.h"
 #include "util-privs.h"
 #include "util-device.h"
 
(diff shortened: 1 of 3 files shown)


gozzy:
@regit should be a bit better now

src/source-nfq.c
```diff
@@ -887,17 +899,30 @@ int NFQParseAndRegisterQueues(const char *queues)
...
-            }
-        }
-    } else if (NFQRegisterQueue(queue_start) != 0) {
-        SCLogError(SC_ERR_INVALID_ARGUMENT, "queue(s) argument '%s' is not "
-                                        "valid", queues);
-        return -1;
+        num_queues = queue_end - queue_start + 1; // +1 due to inclusive range
     }
 
+    g_nfq_t = (NFQThreadVars *)SCCalloc(num_queues, sizeof(NFQThreadVars));
```
regit: We need to free allocated array at Suricata exit.

gozzy: Do you have an idea about a better place for that? ReceiveNFQThreadExitStats() seems to be the only option by now.

BTW, I've found another small leak - nfq_open() is called without nfq_close().

PR #3596, opened 2018-12-21
netmap: switch to nm_* API

victorjulien:
Process multiple packets at nm_dispatch. Use zero copy for workers
recv mode.

Add configure check netmap check for API 11+ and find netmap api version.

Add netmap guide to the userguide.

https://redmine.openinfosecfoundation.org/issues/2741

Describe changes:
- reimplementation of the NETMAP capture method on top of the nm_* API

N/A

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1448,6 +1448,67 @@
         fi
 
         AC_CHECK_HEADER(net/netmap_user.h,,[AC_ERROR(net/netmap_user.h not found ...)],)
+
+        have_recent_netmap="no"
+        AC_COMPILE_IFELSE([AC_LANG_PROGRAM([
+        #include <net/netmap_user.h>
+        ],[
+        #ifndef NETMAP_API
+        #error "outdated netmap, need one with NETMAP_API"
+        #endif
+        #if NETMAP_API < 11
+        #error "outdated netmap, need at least api version 11"
+        #endif
+        ])], [have_recent_netmap="yes"])
+        if test "x$have_recent_netmap" != "xyes"; then
+            echo "ERROR: outdated netmap"
+            exit 1
+        fi
+        have_netmap_version="unknown"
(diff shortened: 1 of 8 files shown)


vmaffione:
Hi!
  Just some clarification that may be useful.
The `nm_*` API (nm_open and friends)  is available in both 11, 12 and 13 (and future versions) and there are no differences.
This API 11 uses the `NIOCREGIF` and `NIOCGETINFO` commands to interact with netmap.
However, since netmap got many new features since 11 (pipes, monitors, zerocopy monitors, null port, netmap passthrough, partial openings, VALE persistent ports, sync-kloop, virtual offloadings, etc.), we ran out of space in the `struct nmreq`, which is the one used for `NIOCREGIF` and `NIOCGETINFO`.

Therefore we decided to introduce a completely new control API, extensible in such a way that we can keep adding new commands and options without breaking backward compatibility.
The new control API is based on the `NIOCCTRL` command, and it is provided in version 12 and 13.
We provide backward binary compatibility to users of API 11 (e.g. `NIOCREGIF` + `NIOCGETINFO`), and that's why suricata keeps working (together with all the in-tree netmap applications such as `pkt-gen`, `bridge`, `lb`, etc.).
Internally, netmap converts requests done using the API 11 to requests using the current API (13).
The old API (11) is now contained in the "netmap_legacy.h" header, while "netmap.h" contains the new API (and includes "netmap_legacy.h").
In conclusion, since you are using the old API (11), you don't need to check if the local netmap API is 11, 12, 13 (or more), as long as it is at least 11.

victorjulien:
Thanks for the clarifications @vmaffione. I've added the version checks mostly for informational purposes for support requests. I did notice that some nm_open flags aren't supported on the 11 version of the API, so thats why it's checked there.

victorjulien:
Merged in #3673

PR #3609, opened 2019-01-07
NFQ queue range v6

gozzy:

Describe changes:
- NFQ contexts are created dynamically instead of on-stack arrays

This replaces #3594.

### DIFF
--- a/src/source-nfq.c
+++ b/src/source-nfq.c
@@ -48,6 +48,7 @@
 #include "util-debug.h"
 #include "util-error.h"
 #include "util-byte.h"
+#include "util-cpu.h"
 #include "util-privs.h"
 #include "util-device.h"
 
@@ -136,8 +137,8 @@ typedef struct NFQThreadVars_
 /* shared vars for all for nfq queues and threads */
 static NFQGlobalVars nfq_g;
 
-static NFQThreadVars g_nfq_t[NFQ_MAX_QUEUE];
-static NFQQueueVars g_nfq_q[NFQ_MAX_QUEUE];
+static NFQThreadVars *g_nfq_t;
+static NFQQueueVars *g_nfq_q;
 static uint16_t receive_queue_num = 0;
 static SCMutex nfq_init_lock;
 
@@ -765,6 +766,22 @@ TmEcode ReceiveNFQThreadInit(ThreadVars *tv, const void *initdata, void **data)
     return TM_ECODE_OK;
 }
 
+static void NFQDestroyQueue(NFQQueueVars *nq)
+{
+    if (unlikely(nq == NULL)) {
+        return;
+    }
+
+    SCLogDebug("starting... will close queuenum %" PRIu32 "", nq->queue_num);
+    NFQMutexLock(nq);
+    if (nq->qh != NULL) {
+        nfq_destroy_queue(nq->qh);
+        nq->qh = NULL;
+        nfq_close(nq->h);
+        nq->h = NULL;
+    }
+    NFQMutexUnlock(nq);
+}
 
 TmEcode ReceiveNFQThreadDeinit(ThreadVars *t, void *data)
 {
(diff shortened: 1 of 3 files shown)


src/source-nfq.c
```diff
@@ -804,65 +822,120 @@ TmEcode VerdictNFQThreadDeinit(ThreadVars *tv, void *data)
...
+ *  \retval -1 on failure.
+ */
+int NFQParseAndRegisterQueues(const char *queues)
+{
+    uint16_t queue_start = 0;
+    uint16_t queue_end = 0;
+    uint16_t num_queues = 1;    // if argument is correct, at least one queue will be created
+
+    // Either "id" or "start:end" format (e.g., "12" or "0:5")
+    int count = sscanf(queues, "%hu:%hu", &queue_start, &queue_end);
```
victorjulien: I'm not familiar with this 'h' modifier use in sccanf, what does it do?

victorjulien:
I have no objection to the array approach, but I do think it would be nicer if we could have a much more dynamic solution. Instead of a large array that is indexed by queue num use a list or hash or something. I don't really see use cases for more than a couple of hundred queues max, so this array will be very lightly populated if there is only a single high queue number.

Any ideas?

src/source-nfq.c
gozzy: In this case 'h' modifies 'u' and means 'unsigned short int'. Without 'h' it would have been 'unsigned int'.

victorjulien:
On 29-01-19 13:13, Alexander Gozman wrote:
> *@gozzy* commented on this pull request.
> 
> 
> In src/source-nfq.c
> <https://github.com/OISF/suricata/pull/3609#discussion_r251806907>:
> 
>> +/**
> + *  \brief Parses and adds Netfilter queue(s).
> + *
> + *  \param string with the queue number or range
> + *
> + *  \retval 0 on success.
> + *  \retval -1 on failure.
> + */
> +int NFQParseAndRegisterQueues(const char *queues)
> +{
> +    uint16_t queue_start = 0;
> +    uint16_t queue_end = 0;
> +    uint16_t num_queues = 1;    // if argument is correct, at least one queue will be created
> +
> +    // Either "id" or "start:end" format (e.g., "12" or "0:5")
> +    int count = sscanf(queues, "%hu:%hu", &queue_start, &queue_end);
> 
> In this case 'h' modifies 'u' and means 'unsigned short int'. Without
> 'h' it would have been 'unsigned int'.

Oh cool, so is it equivalent to doing %5u or is it even smarter and
takes the max value USHRT_MAX into account?

gozzy:
> Oh cool, so is it equivalent to doing %5u or is it even smarter and takes the max value USHRT_MAX into account?

Some sscanf implementations that I saw used different internal flags for 'u' and 'hu' and, as a result, do different type casts. I suppose that if we want unsigned short we should let sscanf know about it :)

gozzy:
> I have no objection to the array approach, but I do think it would be nicer if we could have a much more dynamic solution. Instead of a large array that is indexed by queue num use a list or hash or something. I don't really see use cases for more than a couple of hundred queues max, so this array will be very lightly populated if there is only a single high queue number.
> 
> Any ideas?

Do I understand correctly that it would be better to turn  *g_nfq_t and *g_nfq_q into lists or anything other than array?

gozzy:
@inliniac, @regit 
Are there any more things to do or current version is acceptable? :)

BTW, I have one more fix on the way: NFQ does not increase packets counter and "iface-stat" command returns zero.

victorjulien:
I think the list would be nice as a future improvement, but is not a requirement for this PR. I will test this PR and merge if I don't find issues.

Wrt the stats, happy to see a separate PR for that. Thanks!

gozzy:
Sure, the stats will be fixed in a separate PR a bit later (there's a dependency on this one).

victorjulien:
Merged in #3641, thanks Alexander!

PR #2765, opened 2017-06-09
rust fixups - v2

jasonish:
Replaces PR #2760 
- Fixes dist/distcheck if Python is not available. Ideally it should be to generate headers, but as Rust is experimental lets not force it yet.
- Add DNS GAP handling to Rust DNS.

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/177
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/529

PR #3610, opened 2019-01-09
Fix Issue 2765 - rule option: migrate to GeoIP2 database and libmaxminddb API.

bmeeks8:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#2765](https://redmine.openinfosecfoundation.org/issues/2765)

Describe changes:
- Update "geoip" rule option support code to use the libmaxminddb library for reading "iso_country" code for an IP address from the GeoIP2/GeoLite2 database format. The legacy GeoIP database format and supporting library (libgeoip) has been discontinued by MaxMind and replaced by GeoIP2 and the libmaxminddb library.

### DIFF
--- a/src/detect-geoip.c
+++ b/src/detect-geoip.c
@@ -94,9 +94,9 @@ static int InitGeolocationEngine(DetectGeoipData *geoipdata)
     (void)ConfGet("geoip-database", &filename);
 
     if (filename == NULL) {
-        SCLogWarning(SC_ERR_INVALID_ARGUMENT, "Unable to locate a GeoIP2 
-                     database filename in YAML conf.  GeoIP rule matching 
-                     is disabled.");
+        SCLogWarning(SC_ERR_INVALID_ARGUMENT, "Unable to locate a GeoIP2"  
+                     "database filename in YAML conf.  GeoIP rule matching "
+                     "is disabled.");
         geoipdata->mmdb_status = MMDB_FILE_OPEN_ERROR;
         return FALSE;
     }
@@ -109,8 +109,8 @@ static int InitGeolocationEngine(DetectGeoipData *geoipdata)
         return TRUE;
     }
 
-    SCLogWarning(SC_ERR_INVALID_ARGUMENT, "Failed to open GeoIP2 database: %s.  
-                 Error was: %s.  GeoIP rule matching is disabled.", filename, 
+    SCLogWarning(SC_ERR_INVALID_ARGUMENT, "Failed to open GeoIP2 database: %s. " 
+                 "Error was: %s.  GeoIP rule matching is disabled.", filename, 
                  MMDB_strerror(status));
     geoipdata->mmdb_status = status;
     return FALSE;
--- a/src/detect-geoip.c
+++ b/src/detect-geoip.c
@@ -205,7 +205,7 @@ static int CheckGeoMatchIPv4(const DetectGeoipData *geoipdata, uint32_t ip)
     {
         for (i = 0; i < geoipdata->nlocations; i++) {
             if (country != NULL && strcmp(country, (char *)geoipdata->location[i])==0) {
-                free(country);
+                SCFree(country);
                 return 1;
             }
         }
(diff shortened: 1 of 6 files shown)


bmeeks8:
Apparently there is no _libmaxminddb_ library package available for Cygwin, thus the test build on appveyor is failing because it says it cannot find the library.  I've searched the web and did not find a binary package for Cygwin I could reference.  I did find source code for maybe creating a binary package, though, using _cygport_.

So does this mean "geoip" support in Suricata is going to die with the end-of-life of the legacy GeoIP database, or is there another alternative for including the GeoIP2 format supported by _libmaxminddb_ in Suricata?

src/detect-geoip.c
```diff
@@ -84,27 +84,85 @@ void DetectGeoipRegister(void)
...
-static GeoIP *InitGeolocationEngine(void)
+static int InitGeolocationEngine(DetectGeoipData *geoipdata)
 {
-    return GeoIP_new(GEOIP_MEMORY_CACHE);
+    const char *filename = NULL;
+
+    /* Get location and name of GeoIP2 database from YAML conf */
+    (void)ConfGet("geoip-database", &filename);
+
+    if (filename == NULL) {
```
victorjulien: would it make sense to use a default location in this case?

src/detect-geoip.c
```diff
@@ -84,27 +84,85 @@ void DetectGeoipRegister(void)
...
+    if (result.found_entry) {
+        mmdb_error = MMDB_get_value(&result.entry, &entry_data, "country", 
+                                    "iso_code", NULL);
+        if (mmdb_error != MMDB_SUCCESS)
+            return NULL;
+
+        /* If ISO country code was found, then return it */
+        if (entry_data.has_data) {
+            if (entry_data.type == MMDB_DATA_TYPE_UTF8_STRING) {
+                    char *country_code = strndup((char *)entry_data.utf8_string,
```
victorjulien: Why is the string copied here? It seems unnecessary. If it is needed, please use SCStrdup instead.

src/detect-geoip.c
```diff
@@ -125,29 +183,44 @@ static const char *GeolocateIPv4(GeoIP *geoengine, uint32_t ip)
...
+        return 0;
+
     /* Check if NOT NEGATED match-on condition */
     if ((geoipdata->flags & GEOIP_MATCH_NEGATED) == 0)
     {
-        for (i = 0; i < geoipdata->nlocations; i++)
-            if (country != NULL && strcmp(country, (char *)geoipdata->location[i])==0)
+        for (i = 0; i < geoipdata->nlocations; i++) {
+            if (country != NULL && strcmp(country, (char *)geoipdata->location[i])==0) {
+                free(country);
```
victorjulien: SCFree

victorjulien:
Our Windows support efforts are now focused on MinGW, which does have a libmaxminddb package. In any case I would like to get this work merged, even if not all platforms support it.

src/detect-geoip.c
bmeeks8: Will change this.

src/detect-geoip.c
bmeeks8: The new MaxMind library, when passed a NULL filename, will return a file open error and terminate.  This behavior is different from the legacy libGeoIP library which would search a default path for a DB file before giving up.

Do you want to include all the code baggage of locating a default path and DB file across Linux/Windows platforms, or just log the error message if a valid DB file and path is not given in suricata.yaml?

bmeeks8: After further reflection, I propose leaving the code "as-is" whereby it logs an appropriate warning message and disables the geoip rule option when the GeoIP2 database is not specified in _suricata.yaml_.  The _libmaxminddb_ library will not use a default path when NULL is provided for the DB name.  There are at least two different DB names possible for the Country database depending on whether the user has the free GeoLite2 edition or the paid subscription GeoIP2 edition.  The user is responsible for configuring the DB download anyway, so the user may as well specify the location of the DB to Suricata via the _suricata.yaml_ entry.

bmeeks8:
I believe all review comments have been resolved and this request is ready for acceptance and merging.

src/detect-geoip.c
victorjulien: Ok fair enough, but then I would prefer a proper SCStrndup implementation to be introduced. It should be quite simple to add that.

PR #3612, opened 2019-01-14
Mpls improvements

wesleyvanderree:

- https://redmine.openinfosecfoundation.org/issues/2771
- https://redmine.openinfosecfoundation.org/issues/2772

Describe changes:
- Improved decoding vlan by adding mpls support.
- Added MPLS labels to the alert output.


src/output-json.c
```diff
@@ -739,6 +739,28 @@ json_t *CreateJSONHeader(const Packet *p, enum OutputJsonLogDirection dir,
         }
     }
 
+    /* mpls */
+    if (p->mpls_idx > 0) {
+        json_t *js_mpls;
+        switch (p->mpls_idx) {
```
victorjulien: @jasonish this is the same as the vlan logic, but I'm not quite happy with that. I know we discussed this a while ago. Do you think we should go for an array in all cases here?

jasonish: Yes. I think if it can be an array, it should always be an array. I know JSON schema allows for both, but its just simpler if its always the same datatype.

src/decode.h
```diff
@@ -434,6 +434,9 @@ typedef struct Packet_
     uint16_t vlan_id[2];
     uint8_t vlan_idx;
 
+    uint16_t mpls_id[2];
```
victorjulien: mpls label space seems to limited only by packet size (or mtu), so 2 might be a bit too low. I see limits being used in practice between 4 and 6 when I google it.

A second issue is that the label appears to be 20 bits in size, so a uint16_t is too small.

src/output-json.c
```diff
@@ -739,6 +739,19 @@ json_t *CreateJSONHeader(const Packet *p, enum OutputJsonLogDirection dir,
         }
     }
 
+    /* mpls */
+    if (p->mpls_idx > 0) {
+        json_t *js_mpls;
+        js_mpls = json_array();
+
+        for (int i = 0; i < p->mpls_idx; i++) {
+            if (unlikely(js != NULL)) {
```
victorjulien: checking for js here doesn't really make sense, for 2 reasons:
1. it should check js_mpls instead
2. is should be check outside of the loop

src/output-json.c
```diff
@@ -739,6 +739,19 @@ json_t *CreateJSONHeader(const Packet *p, enum OutputJsonLogDirection dir,
         }
     }
 
+    /* mpls */
+    if (p->mpls_idx > 0) {
+        json_t *js_mpls;
+        js_mpls = json_array();
+
+        for (int i = 0; i < p->mpls_idx; i++) {
```
victorjulien: this compares a signed int (i) with a unsigned int. Some compilers warn on this

wesleyvanderree:
Closed due to new PR: https://github.com/OISF/suricata/pull/3668

PR #3616, opened 2019-01-18
Netmap 11 to 12 fix

muratbalaban43:
Although, there are no visible API changes, moving from API version
11 to 12, for every rx/tx ring that is being opened, one needs to
explicitly call NETMAP_IF to get the fresh address of the netmap_if
struct.

When the netmap-rework branch is done, it will also handle the
situation, but anyhow, this is a quick fix for 4.1.3 so that Suricata
can still happily run on FreeBSD 12 & 11 newcoming releases, since
they all come with the new netmap code.

Contributed by: Sunny Valley Networks

Attn: @inliniac

Describe changes:
-
-
-

### DIFF
--- a/src/source-netmap.c
+++ b/src/source-netmap.c
@@ -268,6 +268,7 @@ static int NetmapOpen(char *ifname, int promisc, NetmapDevice **pdevice, int ver
 {
     NetmapDevice *pdev = NULL;
     struct nmreq nm_req;
+    struct netmap_if *nifp = NULL;
 
     *pdevice = NULL;
 
@@ -367,7 +368,7 @@ static int NetmapOpen(char *ifname, int promisc, NetmapDevice **pdevice, int ver
             nm_req.nr_ringid = i | NETMAP_NO_TX_POLL;
         } else {
             nm_req.nr_flags = NR_REG_SW;
-            nm_req.nr_ringid = NETMAP_NO_TX_POLL;
+            nm_req.nr_ringid = i | NETMAP_NO_TX_POLL;
         }
         if (ioctl(pring->fd, NIOCREGIF, &nm_req) != 0) {
             SCLogError(SC_ERR_NETMAP_CREATE,
@@ -385,14 +386,15 @@ static int NetmapOpen(char *ifname, int promisc, NetmapDevice **pdevice, int ver
                            strerror(errno));
                 break;
             }
-            pdev->nif = NETMAP_IF(pdev->mem, nm_req.nr_offset);
-        }
+            pdev->nif = nifp = NETMAP_IF(pdev->mem, nm_req.nr_offset);
+        } else
+		nifp = NETMAP_IF(pdev->mem, nm_req.nr_offset);
 
         if ((i < pdev->rx_rings_cnt) || (i == pdev->rings_cnt)) {
-            pring->rx = NETMAP_RXRING(pdev->nif, i);
+            pring->rx = NETMAP_RXRING(nifp, i);
         }
         if ((i < pdev->tx_rings_cnt) || (i == pdev->rings_cnt)) {
-            pring->tx = NETMAP_TXRING(pdev->nif, i);
+            pring->tx = NETMAP_TXRING(nifp, i);
         }
         SCSpinInit(&pring->tx_lock, 0);
         success_cnt++;


vmaffione:
Hi,
  No fix should be necessary, just a recompilation.
As you say, there is no visible change for the legacy API (NIOCREGIF).
If this is not true, it may be that support for the legacy API has some bugs.
Could you please share a minimal configuration of suricata, so that I can try myself and see where the problem is?

muratbalaban43:
@vmaffione , here is the relavent netmap section. Suricata-netmap bridges igb0 hw <-> sw rings:

netmap:
  - interface: default
    threads: auto
    copy-mode: ips
    disable-promisc: yes #  promiscuous mode
    checksum-checks: auto


  - interface: igb0
    copy-iface: igb0+
  
  - interface: igb0+
    copy-iface: igb0

vmaffione:
Is this a complete configuration that I can just copy-paste and it should work?
I'm sorry, I've never used suricata before.

vmaffione:
Thanks a lot, I will try asap.

vmaffione:
Hi,
  I tried to install suricata from these github sources, on FreeBSD 12.0-RELEASE, and without this patch.
I use the following commands to run suricata over an em interface:

````
sudo ifconfig em1 up -arp promisc -rxcsum -txcsum -rxcsum6 -txcsum6 -tso -tso4 -tso6 -lro -vlanmtu -vlanhwtag -vlanhwfilter -vlanhwtso
sudo suricata -c /etc/suricata/suricata.yaml --netmap=em1 -v
````

This is the netmap section of my suricata.yaml:
````
netmap:
  - interface: default
    threads: auto
    copy-mode: ips
    disable-promisc: yes #  promiscuous mode
    checksum-checks: auto

  - interface: em1
    copy-iface: em1+

  - interface: em1+
    copy-iface: em1

````

and I see packets being captured
````
[100078] 3/2/2019 -- 14:01:59 - (util-device.c:329) <Notice> (LiveDeviceListClean) -- Stats for 'em1':  pkts: 4892, drop: 0 (0.00%), invalid chksum: 0
````

So why is this patch needed?

**Comment by vmaffione** [comment]
(more information here https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=230465#c14)

**Comment by fichtner** [comment]
You're not using em1+ likely because you stuff --netmap=em1 from the command line

**Comment by vmaffione** [comment]
My bad, sorry. I'm not familiar with suricata.
The question is, though, what should I do to reproduce the issue?

**Comment by fichtner** [comment]
Give the yaml to suricata, do not mention interfaces on the command line. Run traffic through em1. It should block instead of flowing through when suricata is running.

**Comment by vmaffione** [comment]
It does not seem to be a valid commmand line:
````
[vmaffione@fbsd11rel ~]$ sudo suricata -c /etc/suricata/suricata.yaml -v
Suricata 5.0.0-dev (rev e710b066)
USAGE: suricata [OPTIONS] [BPF FILTER]

	-c <path>                            : path to configuration file
	-T                                   : test configuration file (use with -c)
	-i <dev or ip>                       : run in pcap live mode
	-F <bpf filter file>                 : bpf filter file
	-r <path>                            : run in pcap file/offline mode
	-s <path>                            : path to signature file loaded in addition to suricata.yaml settings (optional)
	-S <path>                            : path to signature file loaded exclusively (optional)
	-l <dir>                             : default log directory
	-D                                   : run as daemon
	-k [all|none]                        : force checksum check (all) or disabled it (none)
	-V                                   : display Suricata version
	-v[v]                                : increase default Suricata verbosity
	--list-app-layer-protos              : list supported app layer protocols
	--list-keywords[=all|csv|<kword>]    : list keywords implemented by the engine
	--list-runmodes                      : list supported runmodes
	--runmode <runmode_id>               : specific runmode modification the engine should run.  The argument
	                                       supplied should be the id for the runmode obtained by running
	                                       --list-runmodes
	--engine-analysis                    : print reports on analysis of different sections in the engine and exit.
	                                       Please have a look at the conf parameter engine-analysis on what reports
	                                       can be printed
	--pidfile <file>                     : write pid to this file
	--init-errors-fatal                  : enable fatal failure on signature init error
	--disable-detection                  : disable detection engine
	--dump-config                        : show the running configuration
	--build-info                         : display build information
	--pcap[=<dev>]                       : run in pcap mode, no value select interfaces from suricata.yaml
	--pcap-file-continuous               : when running in pcap mode with a directory, continue checking directory for pcaps until interrupted
	--pcap-file-delete                   : when running in replay mode (-r with directory or file), will delete pcap files that have been processed when done
	--pcap-buffer-size                   : size of the pcap buffer value from 0 - 2147483647
	--netmap[=<dev>]                     : run in netmap mode, no value select interfaces from suricata.yaml
	--simulate-ips                       : force engine into IPS mode. Useful for QA
	--erf-in <path>                      : process an ERF file
	--set name=value                     : set a configuration value


To run the engine with default configuration on interface eth0 with signature file "signatures.rules", run the command as:

suricata -c suricata.yaml -s signatures.rules -i eth0 
````

**Comment by muratbalaban43** [comment]
@vmaffione , try adding --netmap (without interface name)

suricata -c suricata.yaml --netmap -s signatures.rules

**Comment by vmaffione** [comment]
Yes, indeed, thanks.
I'm able to reproduce the issue :)

**Comment by muratbalaban43** [comment]
All welcome 👍

**Comment by vmaffione** [comment]
I see the problem now, and your patch is the right fix. In the original netmap implementation, you could only open all the RX/TX/host rings, so all or none.
This means that the `struct netmap_if` always had all the `ring_ofs` entries valid, e.g. pointing to all the TX, RX and host rings.
And suricata is written making this assumption.

However, with the introduction of the "partial opening" feature, that was not the case anymore.
If you open `netmap:ix-3`, for example, only the fourth TX ring and the fourth RX ring are reachable through the `netmap_if`. The others just point to a special invalid ring with all the slots pointing to a special invalid buffer, which means that you can't use those rings.
In other words, each `NIOCREGIF` registration allocates an independent instance of `struct netmap_if`, and thus if you separately open hw rings and sw rings you'll get two `struct netmap_if`, one with the valid hw rings, and the other with valid sw rings. That's why you need to use NETMAP_IF after each NIOCREGIF.

File: src/source-netmap.c [hunk 1]
```diff
@@ -367,7 +368,7 @@ static int NetmapOpen(char *ifname, int promisc, NetmapDevice **pdevice, int ver
             nm_req.nr_ringid = i | NETMAP_NO_TX_POLL;
         } else {
             nm_req.nr_flags = NR_REG_SW;
-            nm_req.nr_ringid = NETMAP_NO_TX_POLL;
+            nm_req.nr_ringid = i | NETMAP_NO_TX_POLL;
```
vmaffione: The `i` here is not necessary, and may be rejected by netmap in the future.

src/source-netmap.c
```diff
@@ -385,14 +386,15 @@ static int NetmapOpen(char *ifname, int promisc, NetmapDevice **pdevice, int ver
                            strerror(errno));
                 break;
             }
-            pdev->nif = NETMAP_IF(pdev->mem, nm_req.nr_offset);
-        }
+            pdev->nif = nifp = NETMAP_IF(pdev->mem, nm_req.nr_offset);
```
vmaffione: I would remove the `nif` field from the `pdev`, that's very dangerous.

src/source-netmap.c
```diff
@@ -385,14 +386,15 @@ static int NetmapOpen(char *ifname, int promisc, NetmapDevice **pdevice, int ver
...
+        } else
+		nifp = NETMAP_IF(pdev->mem, nm_req.nr_offset);
```
vmaffione: Remove `else` and always compute the nifp before using `NETMAP_RXRING` and `NETMAP_TXRING`.

vmaffione:
Btw this change of behavior happened way before switching from 11 to 12, so the latter control API change is not responsible for this issue (as the title suggests).

victorjulien:
@vmaffione thanks a lot for your feedback, appreciate it a lot that you've taken the time for this!

@muratbalaban43 can you submit a new PR with the suggested updates? Also the coding style was a bit off and the commit message needs to be cleaned up a bit.

Please see https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Github_work_flow#Create-a-new-branch-for-incorporating-feedback
and
https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Code_Submission_Quality_Criteria

Thanks!

vmaffione:
Of course you do that because you want to process each ring in a different thread.. please ignore my last message, and let's go for Murat's patch :)

muratbalaban43:
@vmaffione Thanks for the comments. @victorjulien , I'll be creating another request. Thank you for the pointers.

PR #3622, opened 2019-01-24
detect-geoip: migrate to GeoIP2 MMDB database format (fix Issue 2765) - v6

bmeeks8:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [Issue 2765](https://redmine.openinfosecfoundation.org/issues/2765)

Describe changes:
- Update the geoip rule option code to use the MaxMind MMDB GeoIP2 database format and supporting API from the libmaxminddb library.
- Remove legacy libgeoip library and "--enable-geoip" build option from AppVeyor CygWin build because there is no libmaxminddb library for CygWin.
- Add a SCStrndup() wrapper function in util-mem.h for strndup(). The MaxMind GeoIP2 library returns string data without a null terminator and with a data length parameter instead.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,7 +31,7 @@ init:
 install:
     - ps: 'Start-FileDownload "http://cygwin.com/$env:CYG_SETUP" -FileName "$env:CYG_SETUP"'
     - '%CYG_SETUP% -gqnNdO --quiet-mode --no-shortcuts --only-site --root "%CYG_ROOT%" --site "%CYG_MIRROR%" --local-package-dir "%CYG_CACHE%" > NUL 2>&1'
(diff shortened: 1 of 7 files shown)


victorjulien:
While adding some QA support for this, I noticed that the unittests fail if the feature is compiled in, but no database is found. I also didn't find a package to install to get the db directly, only indirectly: install downloader, run downloader. Any idea on how to address this? I think I would have rather it skip the tests in this case, than cause failure.

victorjulien:
Or maybe the DB is available, but not found. Unittests run without config, so we can't rely on the user setting the right location.

bmeeks8:
> Or maybe the DB is available, but not found. Unittests run without config, so we can't rely on the user setting the right location.

Hmm... the only two solutions I know of are to bypass the unit tests for this or have the new code try the default install location that is part of the downloader utility package you found.  However, that default check is not guaranteed to find a DB either if the user has not downloaded and installed it.  So maybe the best solution of the two is to skip that unit test.

I'm not a big fan of the way MaxMind set up this new library.  They made the API much more difficult to use in my opinion as compared to the older library.

Which solution sounds best to you?

victorjulien:
I guess a third option would be to create a small little db of our own just for the tests. Though the location of that would be another issue. Could it even live in the code and be opened as fmemopen?

Bypass is starting to sound pretty good :)

PR #3634, opened 2019-01-31
configure.ac: fix --{disable,enable}-xxx options

ffontaine:
Currently, if the user provides --enable-libmagic or
--disable-libmagic, libmagic will be disabled because $enableval is not
used to know if the user provided --enable or --disable

Most of the options have this issue so fix them all by using $enableval

Fixes:
 - https://redmine.openinfosecfoundation.org/issues/2797

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -86,7 +86,7 @@
     fi
 
     AC_ARG_ENABLE(python,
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged after rebase to master. Thanks @ffontaine !

PR #3624, opened 2019-01-24
eve/flow: add vlan and in_iface fields - v1

praseodym:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable) => not applicable

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2057

Describe changes:
- eve/flow: add vlan field 
- eve/flow: add in_iface field

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
not applicable

PR #3636, opened 2019-02-01
add vlan and in_iface fields to eve/flow and pseudopackets - v2

praseodym:

- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable) => not applicable

Describe changes:
- add vlan and in_iface to logged/alerted psuedopackets
- eve/flow: add vlan field 
- eve/flow: add in_iface field

not applicable

Supersedes #3494 and #3624.

### DIFF
--- a/src/flow.h
+++ b/src/flow.h
@@ -27,6 +27,7 @@
 #include "decode.h"
 #include "util-var.h"
 #include "util-atomic.h"
+#include "util-device.h"
 #include "detect-tag.h"
 #include "util-optimize.h"
 
--- a/src/flow.h
+++ b/src/flow.h
@@ -629,4 +629,3 @@ uint8_t FlowGetDisruptionFlags(const Flow *f, uint8_t flags);
 void FlowHandlePacketUpdate(Flow *f, Packet *p);
 
 #endif /* __FLOW_H__ */
-
--- a/src/flow.h
+++ b/src/flow.h
@@ -346,6 +346,10 @@ typedef struct Flow_
     uint8_t proto;
     uint8_t recursion_level;
     uint16_t vlan_id[2];
+    uint8_t vlan_idx;
+
+    /** Incoming interface */
+    struct LiveDevice_ *livedev;
 
     /** flow hash - the flow hash before hash table size mod. */
     uint32_t flow_hash;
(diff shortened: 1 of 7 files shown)


src/flow.h
```diff
@@ -346,6 +347,10 @@ typedef struct Flow_
     uint8_t proto;
     uint8_t recursion_level;
     uint16_t vlan_id[2];
+    uint8_t vlan_idx;
+
+    /** Incoming interface */
+    struct LiveDevice_ *livedev;
```
victorjulien: I think this should be a const ptr as the livedev should never be changed from the flow

src/stream-tcp.c
```diff
@@ -6034,6 +6034,17 @@ static void StreamTcpPseudoPacketCreateDetectLogFlush(ThreadVars *tv,
     np->flags |= PKT_HAS_FLOW;
     np->flags |= PKT_IGNORE_CHECKSUM;
     np->flags |= PKT_PSEUDO_DETECTLOG_FLUSH;
+    np->livedev = f->livedev;
+
+    if (f->vlan_idx > 0) {
+        np->vlan_id[0] = f->vlan_id[0];
+        np->vlan_idx=1;
```
victorjulien: style nitpick: we need spaces before and after the =

src/flow.h
praseodym: Done, this does require casting in the pseudopacket assignment to avoid compiler warnings (`warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]`)

src/stream-tcp.c
praseodym: Changed.

praseodym:
> I've requested 2 small changes.
> 
> More fundamentally, I would like to think about the vlan addition to the flow records. As discussed here [#3612 (comment)](https://github.com/OISF/suricata/pull/3612#discussion_r250167172), I regret the per record vlan field type ambiguity. I wonder if we should already fix it here, even if the other record type still have this issue. @jasonish opinions?

#3645 addresses this. I've changed the json flow vlan field to always be an array, and created another commit to do the same for other records for consistency. I don't think this is a breaking change (unless one never expects double-tagged vlans in the output) but it should be added to the release notes.

jasonish:
> More fundamentally, I would like to think about the vlan addition to the flow records. As discussed here [#3612 (comment)](https://github.com/OISF/suricata/pull/3612#discussion_r250167172), I regret the per record vlan field type ambiguity. I wonder if we should already fix it here, even if the other record type still have this issue. @jasonish opinions?

I agree.  In theory, since existing code has to support the value being a scalar or a list, we should be able to always use a list elsewhere as well.  I guess the only issue would be any parsing code that assumes a minimum list length of 2.

PR #3645, opened 2019-02-10
add vlan and in_iface fields to eve/flow and pseudopackets - v3

praseodym:

- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable) => not applicable

Describe changes:
- add vlan and in_iface to logged/alerted psuedopackets
- eve/flow: add vlan field 
- eve/flow: add in_iface field
- eve/json: always output vlan field as array

not applicable

Supersedes #3494, #3624 and #3636.

### DIFF
--- a/src/flow-timeout.c
+++ b/src/flow-timeout.c
@@ -88,6 +88,10 @@ static inline Packet *FlowForceReassemblyPseudoPacketSetup(Packet *p,
     p->flags |= PKT_STREAM_EOF;
     p->flags |= PKT_HAS_FLOW;
     p->flags |= PKT_PSEUDO_STREAM_END;
+    p->vlan_id[0] = f->vlan_id[0];
+    p->vlan_id[1] = f->vlan_id[1];
+    p->vlan_idx = f->vlan_idx;
+    p->livedev = (struct LiveDevice_ *)f->livedev;
 
     if (f->flags & FLOW_NOPACKET_INSPECTION) {
         DecodeSetNoPacketInspectionFlag(p);
@@ -466,4 +470,3 @@ void FlowForceReassembly(void)
     FlowForceReassemblyForHash();
     return;
 }
-
--- a/src/flow-util.c
+++ b/src/flow-util.c
@@ -203,4 +203,3 @@ void FlowInit(Flow *f, const Packet *p)
 
     SCReturn;
 }
-
--- a/src/flow-util.c
+++ b/src/flow-util.c
@@ -153,6 +153,8 @@ void FlowInit(Flow *f, const Packet *p)
     f->recursion_level = p->recursion_level;
     f->vlan_id[0] = p->vlan_id[0];
     f->vlan_id[1] = p->vlan_id[1];
+    f->vlan_idx = p->vlan_idx;
+    f->livedev = p->livedev;
 
     if (PKT_IS_IPV4(p)) {
         FLOW_SET_IPV4_SRC_ADDR_FROM_PACKET(p, &f->src);
--- a/src/flow-util.h
+++ b/src/flow-util.h
@@ -40,6 +40,8 @@
         (f)->sp = 0; \
         (f)->dp = 0; \
         (f)->proto = 0; \
+        (f)->livedev = NULL; \
+        (f)->vlan_idx = 0; \
         SC_ATOMIC_INIT((f)->flow_state); \
         SC_ATOMIC_INIT((f)->use_cnt); \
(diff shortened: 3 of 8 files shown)


victorjulien:
I'm inclined to accept the vlan change for existing record types since we're early in the 5.0 cycle. Opinions @jasonish @regit ?

jasonish:
> I'm inclined to accept the vlan change for existing record types since we're early in the 5.0 cycle. Opinions @jasonish @regit ?

I agree.

victorjulien:
@jasonish I'm already running into the first compatibility issue: suricata-verify. I have a test that looks for:
```yaml
checks:
  - filter:
      count: 1
      match:
        event_type: alert
        alert.signature_id: 1
        vlan: 1111
```
which doesn't match
```json
"event_type":"alert","vlan":[1111]
```
It does match when the vlan field is
```json
vlan:1111
```

jasonish:
It probably shouldn't match [1111] as its looking for changes in output.  Here the test is looking for a scalar, not a list so the test will need to be updated.

This is quite different than a full parser that will already have to do something like:
```
if isList(object):
    ....
else:
    # Handle as scalar.
```
so its more likely the change will go unnoticed there.

victorjulien:
Any suggestions to how to deal with this in tests? Splitting the test in 2 is possible but seems exactly the kind of thing I want to avoid.

jasonish:
I assume this is so we can validate 4.x and 5.x with the same test? Otherwise I'd say just make it a list for 5.0 and move on.  We could add a version specifier to each check?

Otherwise I think we've had to create some modifier functions before the match?
```
     match:
        event_type: alert
        alert.signature_id: 1
        vlan | tolist: 1111
```
I'm not sure what exactly it would look like, but the above would attempt to convert whatever type it is to a list.  So a list would be a list, a scalar would become a single item list.

The version check inside the match being simpler of course.

victorjulien:
I wonder if we could have vlan be a special magic thing instead. I'm still planning to have some kind of eve2test tool at some point, and it would make sense if that would just work w/o special knowledge of this vlan ambiguity.

jasonish:
I'm not sure I agree.  One of the ideas of suricata-verify is to catch minute details like that.  Is the plan in 5.0 to make it always a list? If so, why can't we just add in the additional version check.  Then it will detect any change in that behaviour, vs some magic that may mask real issues.

victorjulien:
Yeah I guess that makes sense. So what would the version check look like?

victorjulien:
@praseodym @bolzzy while we resolve the suricata-verify issue, could you submit some suricata-verify tests that test this feature?

praseodym:
> @praseodym @bolzzy while we resolve the suricata-verify issue, could you submit some suricata-verify tests that test this feature?

I've created a test for single and double-tagged VLAN flows: https://github.com/OISF/suricata-verify/pull/14 

I'm not sure how to go about testing existance of `in_iface` with a pcap or how to test pseudopackets.

bolzzy:
> > @praseodym @bolzzy while we resolve the suricata-verify issue, could you submit some suricata-verify tests that test this feature?
> 
> I've created a test for single and double-tagged VLAN flows: [OISF/suricata-verify#14](https://github.com/OISF/suricata-verify/pull/14)
> 
> I'm not sure how to go about testing existence of `in_iface` with a pcap or how to test pseudopackets.

Unless suricata-verify is able to replay packets to a (virtual) interface and make suricata listen to that interface, I don't think it would be possible to verify existence of in_iface by reading a pcap directly with suricata. However, we could probbably script it with tcpreplay. Any input on this @victorjulien ?

On the other issue, I can try to produce a PCAP that triggers psuedopackets at the end of the week (and suricata-verify test if I got time).

praseodym:
> On the other issue, I can try to produce a PCAP that triggers psuedopackets at the end of the week (and suricata-verify test if I got time).

There's a Scapy script included in my suricata-verify PR which you could maybe use as a starting point :)

victorjulien:
@bolzzy no we can't test the in_iface this way. But the vlan in pseudo packets should be easy: just cut all the tcp shutdown packets from a pcap and it should generate the pseudo packets at engine shutdown.

victorjulien:
Created https://redmine.openinfosecfoundation.org/issues/2924. Once suricata-verify has been patched, I will merge this PR.

victorjulien:
Merged as part of #3815, thanks @praseodym and @bolzzy!

PR #3649, opened 2019-02-13
Test/stream events/v2

victorjulien:
https://redmine.openinfosecfoundation.org/issues/2484

Describe changes:
- when stream gaps or flow 'wrong thread' is encountered, stop setting stream events to avoid causing a flood of alerts
- add gap status to flow records

RFC: this stops both matching and stats increases of stream events after gap/wrong-thread. It may lead to reduced detection as a result, but I think it's a fair balance as the events are unreliable anyway in this scenario.

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/220
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/222

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -1131,6 +1131,7 @@ void DecodeUnregisterCounters(void);
 
 #define PKT_PSEUDO_DETECTLOG_FLUSH      (1<<27)     /**< Detect/log flush for protocol upgrade */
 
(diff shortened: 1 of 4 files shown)


victorjulien:
Replaced by #3661

PR #3651, opened 2019-02-13
app-layer-ssl: fix AddressSanitizer heap-buffer-overflow

thus:
When decoding CipherSuites there is no check to verify that the CipherSuites length is divisible by 2, causing overflows whenever the length is in fact an odd number (which should never happen with normal traffic).

https://redmine.openinfosecfoundation.org/issues/2762

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/179
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/179

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -771,6 +771,11 @@ static inline int TLSDecodeHSHelloCipherSuites(SSLState *ssl_state,
         if (!(HAS_SPACE(cipher_suites_length)))
             goto invalid_length;
 
+        /* Cipher suites length should always be divisible by 2 */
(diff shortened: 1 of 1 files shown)


pevma:
Works as expected with the offending pcap.

jmtaylor90:
@thus would you like a suricata-verify test for this? I would be happy to come up with one if so.

pevma:
Please feel free - seems like a corner case.

victorjulien:
@jmtaylor90 yes please :)

thus:
Replaced by #3664

PR #3670, opened 2019-02-20
Doc suricatasc/v1

shivan1b:
- Add missing commands and their corresponding details in unix-socket
userguide.
- Fix the message for unregister-tenant-handler

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2800

PR #3671, opened 2019-02-20
suricatactl cleanup/v1

shivan1b:
- Make code compatible with Python 3
-  Fix PyLint issues
- Clean up parser, improve help

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2793

### DIFF
--- a/python/suricata/ctl/filestore.py
+++ b/python/suricata/ctl/filestore.py
@@ -21,23 +21,23 @@ import os
 import os.path
 import time
 import re
-import glob
 import logging
 
 logger = logging.getLogger("filestore")
 
+
 class InvalidAgeFormatError(Exception):
     pass
 
+
 def register_args(parser):
     parsers = parser.add_subparsers()
-
     prune_parser = parsers.add_parser("prune")
     prune_parser.add_argument("-d", "--directory", help="filestore directory")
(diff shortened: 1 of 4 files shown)


python/suricata/ctl/filestore.py
```diff
@@ -47,39 +50,39 @@ def register_args(parser):
...
-        return val * 60 * 60 * 24
-    else:
+    val = int(matched_age.group(1))
+    unit = matched_age.group(2)
+    ts_units = ["s", "m", "h", "d"]
+    try:
+        idx = ts_units.index(unit)
+    except ValueError:
         raise InvalidAgeFormatError("bad unit: %s" % (unit))
+    multiplier = 60 ** idx if idx != 3 else 24 * 60 ** 2
```
jasonish: My only concern is this way of calculating the age. It feels more fragile to change than an if statement. but that being said I don't see a reason why we would need to extend this in the near future.

victorjulien:
Minor: commit prefix should be just 'suricatactl: ' (no colon)

Can you make these changes and do a single PR with these commits + the ones from #3670? Thanks!

PR #2853, opened 2017-07-21
samples: systemd unit file and logrotate configuration file - 2

jasonish:
Add a sample logrotate configuration file, and a systemd unit file.

These are in contrib, not sure if there was a better place. Perhaps creating an etc directory?

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/207
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/560

PR #3683, opened 2019-02-27
filestore: fix dropping of unwanted files (Issue #2853)

magenbluten:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [Issue #2853](https://redmine.openinfosecfoundation.org/issues/2853)

Describe changes:
- use `file_track_id` instead of `file_store_id` to uniquely identify files to drop

`file_store_id` is is/was used by filestore v1. file_store_id only gets updated if the file gets actually dropped to disk. in this case the file is still in memory and just selected / search for dropping. esp in filestore v2 this value is always 0. this leads to dropping of all files in a `FileContainer` when at least one file was marked for dropping.

### DIFF
--- a/src/detect-filestore.c
+++ b/src/detect-filestore.c
@@ -263,7 +263,7 @@ static int DetectFilestoreMatch (ThreadVars *t, DetectEngineThreadCtx *det_ctx, 
     /* file can be NULL when a rule with filestore scope > file
      * matches. */
     if (file != NULL) {
-        file_id = file->file_store_id;
+        file_id = file->file_track_id;
         if (file->sid != NULL && s->id > 0) {
             if (file->sid_cnt >= file->sid_max) {
                 void *p = SCRealloc(file->sid, sizeof(uint32_t) * (file->sid_max + 8));
--- a/src/util-file.c
+++ b/src/util-file.c
@@ -1256,7 +1256,7 @@ void FileStoreFileById(FileContainer *fc, uint32_t file_id)
 
     if (fc != NULL) {
         for (ptr = fc->head; ptr != NULL; ptr = ptr->next) {
-            if (ptr->file_store_id == file_id) {
+            if (ptr->file_track_id == file_id) {
                 FileStore(ptr);
             }
         }


victorjulien:
Could you provide a test case at https://github.com/OISF/suricata-verify ?

victorjulien:
I think this patch isn't correct, as file_track_id isn't used by all protocol parsers. Its fairly new addition to support the (Rust) parsers with parallel file support. HTTP, FTP and SMTP are not setting the file_track_id field.

victorjulien:
Right now the file_track_id is only set when FileOpenFileWithId is used to open the file. It's only used from Rust. Perhaps if the older FileOpenFile is used it should automatically set the track_id field or maybe it makes sense to update the C file api users to also use FileOpenFileWithId. The latter would allow for cleaning up the API, so thats probably nicer.

magenbluten:
> I think this patch isn't correct, as file_track_id isn't used by all protocol parsers. Its fairly new addition to support the (Rust) parsers with parallel file support. HTTP, FTP and SMTP are not setting the file_track_id field.

it works in my testcase for smb and http. i need to make the testcase smaller. if the `file_track_id` is not set then the default is like before: every file of the container gets dropped. you could only drop files of a container having FLAG_STORE set checking while iterating. this would be an alternative solution, but the `file_track_id` seemed the right way to do because using `file_store_id` is definitely wrong.

victorjulien:
The Rust code simply sets the transaction id as the track id, but that only works because the transaction is defined as a single transaction. In SMTP for example, multiple file attachments would all be part of a single tx, so we can't use that approach there. HTTP also has multi-file support with multi-part bodies. Not sure about how our FTP parser would have to be updated.

I do think updating HTTP/SMTP/FTP to use the track_id is probably simple, even if it means simply adding a simple tracker/counter to the per tx state.

Also I do agree that the current file_store_id use is wrong. I think your patch would be correct if HTTP/SMTP/FTP are fixed to use the track_id.

magenbluten:
so how to proceed from here? fix/update the other parsers? we have many many "unwanted" files, mostly via SMB. maybe open tickets for each parser. as soon as they are "fixed" behaviour is correct/restored. up to then all container files get dropped...

victorjulien:
Yeah lets go this route, I think it can be done in a fairly un-intrusive way and can lead to a nice code cleanup as well.

magenbluten:
sorry for the late catch-up.

how do we proceed from here? would it be possible to merge this change? because it doesn't break things. and as soon as the other parsers got fixed "everything" should be correct. even this minor change would improve our situation immediately, which is very important to the quality of our service we are offering our customers.

who is creating the tickets for the parsers? who will update them? and when. i might be able to contribute but have to check with my teammates beforehand.

victorjulien:
You can create tickets in our redmine for this, sure. Wrt merging, as I think this change is incorrect I won't merge it. A good test case is still very welcome.

norg:
We (Alex and I) started to narrow it down (see redmine issue updates) to small pcaps to reproduce the behaviour. With good test case you mean suricata-verify test cases combined with those pcaps?

With the http pcap we could show that the behaviour is wrong and won't be changed by this small fix. Thus the parsers need to be updated as well (in a seperate task) but the patch won't break the current behaviour of those but would already fix it for the ones supporting it.

You said that the patch would be correct if http,smtp,ftp are also fixed, on which we agree but we also think that the patch is already fixing the other ones without breaking the behaviour of those "old" parsers. Thus we think this would already be an improvement to the current state and help a lot.
We would also start to work on the fix for the "old" parsers to improve it for them as well, but this takes a bit more time.

Or is there anything else that you see as incorrect that needs to be addressed as well which we didn't see?
So any suggestion to make it more likely to be merged would be very appreciated :)

Thanks!

victorjulien:
Thanks @norg. Yeah if you could turn those pcaps into suricata-verify tests it would speed up things a lot. I'll then have a look to see if we're missing something obvious or we would indeed need the updates to the various parsers.

norg:
We added the requested suricata-verify tests: https://github.com/OISF/suricata-verify/pull/52

Both will fail, as expected (since  behaviour is wrong), but the smb one will pass with this PR without interfering with the http part for now.

victorjulien:
I've merged this through #3826. I've made the necessary changes to HTTP, FTP and SMTP to make sure all support the track id now.

Thanks @magenbluten and @norg!

PR #3707, opened 2019-03-06
[#2497] Error messages usability improvements

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/issues/2497 ticket:

Describe changes:

Module identifiers are assigned using SCSetModule (macro defined in util-debug.h).


jlucovsky:
Continued in #3965

PR #3708, opened 2019-03-06
New app layer event for invalid http request line

catenacyber:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2655

Describe changes:

- Follows #3705
- New app layer event for invalid http request line (using flags rather than log string comparison)
- New app layer event for unexpected request body


src/app-layer-htp.c
```diff
@@ -138,6 +138,10 @@ SCEnumCharMap http_decoder_event_table[ ] = {
         HTTP_DECODER_EVENT_REQUEST_FIELD_TOO_LONG},
     { "RESPONSE_FIELD_TOO_LONG",
         HTTP_DECODER_EVENT_RESPONSE_FIELD_TOO_LONG},
+    { "REQUEST_LINE_INVALID",
+        HTTP_DECODER_EVENT_REQUEST_LINE_INVALID},
```
jlucovsky: Suggest adding an entry to `htp_errors` for this condition.

catenacyber: Hi @jlucovsky 
I am not sure what you mean...
As I understand, `htp_errors` maps logs messages from libhtp to some identifier in `http_decoder_event_table`

In this case, we do not use libhtp logs.
Do you mean that we should ? as in the first version of this PR #3704, instead of using the numerical values of `request_method_number` and such ?

jlucovsky: I withdraw my comment; you're correct about the purpose of `htp_errors`.

catenacyber: Ok thanks @jlucovsky for taking the time to review

PR #3709, opened 2019-03-06
http: new event for auth unrecognized

catenacyber:

Describe changes:
- new app layer event for auth unrecognized
- activates libhtp auth parsing
- Fixes #984


victorjulien:
Does this mean we're enabling previously unused code paths in libhtp? If so, perhaps we should review them and/or make sure we fuzz test them.

catenacyber:
> Does this mean we're enabling previously unused code paths in libhtp?

Yes it does.
That is why I point this out.

> If so, perhaps we should review them and/or make sure we fuzz test them.

It is already in the coverage for fuzzing
(as this option is now explicitly disabled by suricata and enabled by default)
See https://storage.googleapis.com/oss-fuzz-coverage/libhtp/reports/20190305/linux/src/libhtp/htp/htp_parsers.c.html#L102

Do you want me to review the code ?

victorjulien:
Merged in #3818, thanks Philippe!

PR #3690, opened 2019-03-02
Fix memory leak with TOS handling

jlucovsky:
Memory leak detected by LEAK_SANTIZER fixed by returning memory
allocated from pcre_get_substring().

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/2833) ticket:

Describe changes:
- Ensure string allocated on successful pcre_get_substring is freed on error and normal return paths.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #3711, opened 2019-03-06
Fix memory leak with TOS handling

jlucovsky:
Memory leak detected by LEAK_SANTIZER fixed by returning memory
allocated from pcre_get_substring().

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/2833) ticket:

Changes:
- Ensure string allocated on successful pcre_get_substring is released  on error and normal return paths.

This PR is a continuation of #3690 with a fix for "use after free" per review feedback.

### DIFF
--- a/src/detect-tos.c
+++ b/src/detect-tos.c
@@ -121,7 +121,7 @@ static DetectTosData *DetectTosParse(const char *arg, bool negate)
         goto error;
     }
 
-    const char *str_ptr;
+    const char *str_ptr = NULL;
     res = pcre_get_substring((char *)arg, ov, MAX_SUBSTRINGS, 1,
                              &str_ptr);
     if (res < 0) {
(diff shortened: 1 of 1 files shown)


src/detect-tos.c
```diff
@@ -157,6 +161,9 @@ static DetectTosData *DetectTosParse(const char *arg, bool negate)
 
     return tosd;
 
+error_free:
+    if (str_ptr)
```
victorjulien:
```
** CID 1443023:  Null pointer dereferences  (REVERSE_INULL)
/src/detect-tos.c: 165 in DetectTosParse()


________________________________________________________________________________________________________
*** CID 1443023:  Null pointer dereferences  (REVERSE_INULL)
/src/detect-tos.c: 165 in DetectTosParse()
159         tosd->tos = (uint8_t)tos;
160         tosd->negated = negate;
161     
162         return tosd;
163     
164     error_free:
>>>     CID 1443023:  Null pointer dereferences  (REVERSE_INULL)
>>>     Null-checking "str_ptr" suggests that it may be null, but it has already been dereferenced on all paths leading to the check.
165         if (str_ptr)
166             pcre_free_substring(str_ptr);
167     error:
168         return NULL;
169     }
170     
```

jlucovsky:
Work continued in #3713

PR #3713, opened 2019-03-09
Fix memory leak during TOS parsing

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/2833) ticket:

Use `pcre_copy_substring` instead of `pcre_get_substring` -- the latter allocates memory while the former uses a supplied buffer.

### DIFF
--- a/src/detect-tos.c
+++ b/src/detect-tos.c
@@ -121,31 +121,33 @@ static DetectTosData *DetectTosParse(const char *arg, bool negate)
         goto error;
     }
 
-    const char *str_ptr;
-    res = pcre_get_substring((char *)arg, ov, MAX_SUBSTRINGS, 1,
-                             &str_ptr);
+    /* For TOS value */
+    char tosbytes_str[64] = "";
+    res = pcre_copy_substring((char *)arg, ov, MAX_SUBSTRINGS, 1,
+                             tosbytes_str, sizeof(tosbytes_str));
     if (res < 0) {
-        SCLogError(SC_ERR_PCRE_GET_SUBSTRING, "pcre_get_substring failed");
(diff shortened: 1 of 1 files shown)


src/detect-tos.c
```diff
@@ -121,31 +121,33 @@ static DetectTosData *DetectTosParse(const char *arg, bool negate)
         goto error;
     }
 
-    const char *str_ptr;
-    res = pcre_get_substring((char *)arg, ov, MAX_SUBSTRINGS, 1,
-                             &str_ptr);
+    /* For TOS value */
+    char tosbytes_str[64] = "";
+    res = pcre_copy_substring((char *)arg, ov, MAX_SUBSTRINGS, 1,
```
gozzy: Is `tosbytes_str` guaranteed to fit 64 bytes? AFAIK, `pcre_copy_substring()` uses `memcpy()` or `strncpy()`, and both are insecure. Maybe it'll be better to initialize the variable as `{ 0 }` and use `sizeof(tosbytes_str) - 1`?
Also, I'm not sure that we still need to cast `arg` because `pcre_copy_substring()` accepts `const char *` as its first argument

victorjulien: the manpage suggests that the function will return PCRE_ERROR_NOMEMORY if the size is too small, so I think we're good.

victorjulien:
Just to be sure, this PR fixes the leak fixed in #3711 in a different way, right?

jlucovsky:
Correct.  Memory is no longer allocated while parsing the TOS value; instead, a local scopes buffer is used instead

victorjulien:
Merged as part of #3735, thanks Jeff!

PR #3682, opened 2019-02-26
suricata-verify for travis; prefer python3 - v1

jasonish:
Bundling to separate issues in this PR if it helps review/qa.

Prefer Python 3:
- autoconf will now specifically look for python3, then python2, then python. The idea here is to prefer Python 3.
- Redmine issue: https://redmine.openinfosecfoundation.org/issues/2808
- Specially set the Mac environment to xcode8.3 to get rid of the warning that 8.1 is deprecrated and the tests will be done on 8.3 anyways.

Add suricata-verify to Travis-CI tests:
- Requires jq and pyyaml to be installed on Mac.
- As rustup can sometimes take a very long time on Travis (but not always), attempt to cache it.

You can verify that the correct Rust version is still being used here:
- 1.24.1 -- https://travis-ci.org/jasonish/suricata/jobs/498790305#L635
- 1.31.0 -- https://travis-ci.org/jasonish/suricata/jobs/498790303#L667

Prscript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/345
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/699

PR #3714, opened 2019-03-10
autoconf/python - check for distutils - v2

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/3682

Changes from last PR:
- Add a check for the Python distutils module.
- Only install suricatactl/suricatasc if we have Python and distutils.
- Only install suricata-update if we have distutils and pyyaml (and its bundled).

What I'm not sure about is if we should error out here at all, or what autoconf parameters should cause it to error out instead of continuuing without installing the Python tools.

PRScript:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/346
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/700

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -85,22 +85,26 @@
         exit 1
     fi
 
+    python_version="not set"
+    python_path="not set"
+
     AC_ARG_ENABLE(python,
-           AS_HELP_STRING([--enable-python], [Enable python]),[enable_python=$enableval],[enable_python=yes])
-    AC_PATH_PROGS(HAVE_PYTHON, python python2 python2.7, "no")
-    if test "x$enable_python" = "xno" ; then
-        echo
-        echo "   Warning! python disabled, you will not be      "
-        echo "   able to install suricatasc unix socket client   "
-        echo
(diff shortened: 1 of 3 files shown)


victorjulien:
Just adding a note: This is a bit of a breaking change as many systems will have python3, but not pyyaml for py3. This will now lead to a configure warning and not processing suricata-update even if that is available. In this case it will not fall back to python2 even if the dependencies for that are available.

victorjulien:
This is https://redmine.openinfosecfoundation.org/issues/2808 right?

jasonish:
> Just adding a note: This is a bit of a breaking change as many systems will have python3, but not pyyaml for py3. This will now lead to a configure warning and not processing suricata-update even if that is available. In this case it will not fall back to python2 even if the dependencies for that are available.

We could do a fallback check for Python 2, see if it meets all requirements. I'm hesitant to do that though.

We could also provide, and maybe should a `--with-python` argument that explicitly sets the path to the Python interpreter used.

PR #3716, opened 2019-03-10
Ebpf xdp update 8.3

regit:
Here is a big update to eBPF and XDP code. The number of commits is higher than what I would like but I was not feeling like squashing everything as the number of functionally separated updates is huge. Code has been running on a box getting some heavy eight live traffic. This box have a Netronome card so testing hardware offload was possible.

- https://redmine.openinfosecfoundation.org/issues/2654

Describe changes:
- Handle XDP hardware bypass for Netronome card
- Pinned maps for Suricata restart persistent bypass
- Global bypass switch for IPS
- Update eBPF filter example with IP address pass list
- Updated flow timeout algorithm for bypassed flows

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/437
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/218


ebpf/xdp_filter.c
```diff
@@ -45,6 +45,10 @@
 #define USE_PERCPU_HASH    1
 #define GOT_TX_PEER	1
 
+/* set to non 0 to load balance in hardware mode on RSS_QUEUE_NUMBERS queues
+ * and unset BUILD_CPUMAP */
+#define RSS_QUEUE_NUMBERS   40
```
victorjulien: would there be a way to make this not hardcoded?

victorjulien:
I think for review purposes it would be great if you could split this into multiple PRs:
1. the bpf/af-packet changes
2. the bypass triggers in suri
3. the new tooling (scbpf)

ebpf/xdp_filter.c
regit: In theory yes, we can use a map that stores the number of threads and eBPF code could then get the number of threads from it. Problem is that this setup is specific to hardware bypass. So we could have something working by:
- making the build of this map conditional to harware offload
- get Suricata set the map if ever it is defined in the code

I'm wondering if I should introduce a xdp_hw_filter.c that would be usable without modification for hardware offload. I'm a bit reluctant on that as from eBPF point of view, the C code is just configuration. But it is clearly not the mentality and use case for something like Suricata.

regit:
OK on splitting, it makes sense.

Could you give a try to this PR to your test setup ? The documentation should have been updated enough so it is not painful to test (and you can contact me directly). Idea is to check that hardware offload is working on a second device.

regit:
Replaced by:
- #3719 
- #3720 
- #3721

PR #3719, opened 2019-03-11
Suri bypass update

regit:
Extracted from #3716, this is the change on Suricata bypass strategy.

Describe changes:
- More aggressive bypass strategy
- Thread renaming
- New counter for bypassed flow

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/439
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/220

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -464,6 +464,7 @@ void DecodeRegisterPerfCounters(DecodeThreadVars *dtv, ThreadVars *tv)
     dtv->counter_flow_udp = StatsRegisterCounter("flow.udp", tv);
     dtv->counter_flow_icmp4 = StatsRegisterCounter("flow.icmpv4", tv);
     dtv->counter_flow_icmp6 = StatsRegisterCounter("flow.icmpv6", tv);
+    dtv->counter_flow_pkts_bypassed = StatsRegisterCounter("flow.pkts_bypassed", tv);
 
     dtv->counter_defrag_ipv4_fragments =
         StatsRegisterCounter("defrag.ipv4.fragments", tv);
--- a/src/flow-manager.c
+++ b/src/flow-manager.c
@@ -275,6 +275,8 @@ static int FlowManagerFlowTimedOut(Flow *f, struct timeval *ts)
     int server = 0, client = 0;
 
     if (!(f->flags & FLOW_TIMEOUT_REASSEMBLY_DONE) &&
+            SC_ATOMIC_GET(f->flow_state) != FLOW_STATE_CAPTURE_BYPASSED &&
+            SC_ATOMIC_GET(f->flow_state) != FLOW_STATE_LOCAL_BYPASSED &&
             FlowForceReassemblyNeedReassembly(f, &server, &client) == 1) {
         FlowForceReassemblyForFlow(f, server, client);
         return 0;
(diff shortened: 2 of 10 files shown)


src/flow-manager.c
```diff
@@ -275,6 +275,8 @@ static int FlowManagerFlowTimedOut(Flow *f, struct timeval *ts)
     int server = 0, client = 0;
 
     if (!(f->flags & FLOW_TIMEOUT_REASSEMBLY_DONE) &&
+            SC_ATOMIC_GET(f->flow_state) != FLOW_STATE_CAPTURE_BYPASSED &&
```
victorjulien: If we force reassembly on bypass, can't we simply set the FLOW_TIMEOUT_REASSEMBLY_DONE flag then?

src/decode.c
```diff
@@ -464,6 +464,7 @@ void DecodeRegisterPerfCounters(DecodeThreadVars *dtv, ThreadVars *tv)
     dtv->counter_flow_udp = StatsRegisterCounter("flow.udp", tv);
     dtv->counter_flow_icmp4 = StatsRegisterCounter("flow.icmpv4", tv);
     dtv->counter_flow_icmp6 = StatsRegisterCounter("flow.icmpv6", tv);
+    dtv->counter_flow_pkts_bypassed = StatsRegisterCounter("flow.pkts_bypassed", tv);
```
victorjulien: this will only work for internal bypass, right? Not for driver/hw level bypass, as then suri won't see the packets. I think this may be confusing for ppl. Perhaps the counter should depend on the actual setting we use in af-packet.

src/stream-tcp.c
```diff
@@ -4841,18 +4841,13 @@ int StreamTcpPacket (ThreadVars *tv, Packet *p, StreamTcpThread *stt,
         /* check for conditions that may make us not want to log this packet */
 
         /* streams that hit depth */
-        if ((ssn->client.flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) &&
+        if ((ssn->client.flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) ||
```
victorjulien: I believe I've seen this before. This would lead to potential detection and logging bypass. E.g. depth 1MiB, http post > 1MiB, response is not logged even though it's below the depth for its direction.

src/stream-tcp-reassemble.c
```diff
@@ -1767,6 +1767,13 @@ int StreamTcpReassembleHandleSegment(ThreadVars *tv, TcpReassemblyThreadCtx *ra_
                 ssn, stream, p->payload_len,
                 (stream->flags & STREAMTCP_STREAM_FLAG_NOREASSEMBLY) ? "true" : "false");
 
+        /* In the case of bypass Suricata has to honor the inspection for all packets
+         * before stream depth so we need to check the depth.By adding the stream depth
+         * reached flag, it will trigger the bypass. */
+        if (StreamTcpBypassEnabled()) {
+            StreamTcpReassembleCheckDepth(ssn, stream, TCP_GET_SEQ(p), p->payload_len);
```
victorjulien: I'm a bit confused about this. The last data packet (p->payload_len > 0) should have been checked, so depth flag should have been set then. The exception would be if NOREASSEMBLY flag is set, but perhaps the bypass needs to consider that flag instead?

src/stream-tcp.c
regit: Yes, it was sent once but your explanation looks better this time :)

victorjulien:
I cherry-picked the flow bypass thread name patch and the encrypt handling fix in #3735

victorjulien:
Closing this as it should be rebased and comments should be addressed/answered.

PR #3720, opened 2019-03-11
scbpf: add basic eBPF map handling tool

regit:
SCbpf part of #3716 

Describe changes:
- basic tool to handle IPv4 maps

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/440
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/221

### DIFF
--- a/ebpf/scbpf/scbpf.c
+++ b/ebpf/scbpf/scbpf.c
@@ -1,0 +1,96 @@
+#include <linux/unistd.h>
+#include <linux/bpf.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <unistd.h>
+#include <string.h>
+#include <assert.h>
+#include <errno.h>
+
+#include <sys/types.h>
+#include <sys/socket.h>
+
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
+
+#include <bpf/bpf.h>
+
+#define BPF_F_ADD	(1 << 0)
+#define BPF_F_GET	(1 << 1)
+
+#define BPF_F_KEY	(1 << 2)
+#define BPF_F_VAL	(1 << 3)
+#define BPF_F_KEY_VAL	(BPF_F_KEY | BPF_F_VAL)
+
+static void usage(void)
+{
+	printf("Usage: scbpf [...]\n");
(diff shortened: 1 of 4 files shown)


ebpf/scbpf/scbpf.c
```diff
@@ -0,0 +1,96 @@
+#include <linux/unistd.h>
```
victorjulien: Should this have a copyright header? Also, coding style is different from Suricata's.

regit: Indeed.

regit:
@victorjulien are you ok with the concept of adding this tool inside Suricata source code? Alternatively we could have it as a separate tool to allow different release cycle.

victorjulien:
I guess my first thought was: can this be part of suricatactl? But I'm not sure if it feasible to convert this to python?

regit:
It would not be really trivial. It may be doable. I need to look at it to see if we can get something done without pulling half of the Python planet in term of dependencies.

jasonish:
Would it make sense to have it as a sub-command of Suricata? ./suricata
--scbpf or something like that?

regit:
@jasonish we need all insert/update/delete/list operations, looks like a lot of option to add to Suricata main code. Not fan of it even more because I plan some contributions from beginner on it (intern). Python could be a nice choice, I'm gonna check if we can use that.

jasonish:
> @jasonish we need all insert/update/delete/list operations, looks like a lot of option to add to Suricata main code. Not fan of it even more because I plan some contributions from beginner on it (intern). Python could be a nice choice, I'm gonna check if we can use that.

Makes sense.  So with Python stuff (suricatactl, suricata-update) I've been very sensitive to dependencies that are required. And we've started to add some checks to configure.ac.  But I don't think we have to add all dependencies to configure.ac, just the core stuff.

So if you had this in Python, "suricatactl scebpf" or whatever, in my opinion I think it would be fine to error out with an error "such and such python module required".  I'd be curious to see what Victor thinks, but I don't see this as a "core" feature that would should check all deps for during install of Suricata.

victorjulien:
Is this still relevant now that ```bpfctrl``` exists?

regit:
Can be closed as we can use bpfctrl.

PR #3726, opened 2019-03-15
Avoid use-after-free during pid file cleanup.

bluhm:
In case the pid file is given in the config file, the file name is
stored in volatile memory.  Removal of the pid file happens after
cleanup of config memory, so we need a temporary copy.

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -398,6 +398,9 @@ static void GlobalsDestroy(SCInstance *suri)
     MpmHSGlobalCleanup();
 #endif
 
+    /* The pid file name may be in config memory, but is needed later. */
+    if (suri->pid_filename != NULL)
+        suri->pid_filename = strdup(suri->pid_filename);
     ConfDeInit();
 #ifdef HAVE_LUAJIT
     LuajitFreeStatesPool();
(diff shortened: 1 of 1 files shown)


src/suricata.c
```diff
@@ -398,6 +398,8 @@ static void GlobalsDestroy(SCInstance *suri)
     MpmHSGlobalCleanup();
 #endif
 
+    /* The pid file name may be in config memory, but is needed later. */
+    suri->pid_filename = strdup(suri->pid_filename);
```
jlucovsky: Note that `pid_filename` may be `NULL` unless

- Suricata is running as a daemon
- Suricata was started `--pidfile`.

See https://github.com/OISF/suricata/blob/master/src/suricata.c#L2127

bluhm: Right, I have added a != NULL check.

src/suricata.c
```diff
@@ -407,6 +410,8 @@ static void GlobalsDestroy(SCInstance *suri)
     SCThresholdConfGlobalFree();
 
     SCPidfileRemove(suri->pid_filename);
+    free((char *)suri->pid_filename);
```
jlucovsky: Note that you can only call `free` if the memory you're returning has been allocated by `malloc` (or equivalent).

PR #3730, opened 2019-03-19
Avoid use-after-free during pid file cleanup.

bluhm:
In case the pid file is given in the config file, the file name is
stored in volatile memory.  Removal of the pid file happens after
cleanup of config memory.  Create a copy of the name which will be
freed after the pid file has been removed.


victorjulien:
Merged in #3746, thanks!

PR #3745, opened 2019-03-24
RFC: Improved FTP active mode handling

jlucovsky:
1. https://redmine.openinfosecfoundation.org/issues/2459
1. https://redmine.openinfosecfoundation.org/issues/2527

Describe changes:
- Detect FTP active mode over IPv4 and IPv6
- Create expectation based on mode

PR for suricata-verify in progress.

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -330,6 +330,10 @@ static int FTPParseRequestCommand(void *ftp_state, uint8_t *input,
         fstate->command = FTP_COMMAND_PORT;
     }
 
+    if (input_len >= 4 && SCMemcmpLowercase("eprt", input, 4) == 0) {
+        fstate->command = FTP_COMMAND_EPRT;
+    }
+
     if (input_len >= 8 && SCMemcmpLowercase("auth tls", input, 8) == 0) {
         fstate->command = FTP_COMMAND_AUTH_TLS;
     }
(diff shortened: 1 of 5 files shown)


src/app-layer-ftp.c
```diff
@@ -455,18 +511,23 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
                     memcpy(data->file_name, state->current_line + 5, state->current_line_len - 5);
                     data->cmd = state->command;
                     data->flow_id = FlowGetId(f);
-                    int ret = AppLayerExpectationCreate(f, direction, 0,
-                                                         state->dyn_port,
-                                                         ALPROTO_FTPDATA, data);
+                    int ret = AppLayerExpectationCreate(f,
+                                            state->active ? STREAM_TOSERVER : direction,
```
regit: Should state->active be enough to say if it is to_server or to_client ? Why do we need direction here ?

src/app-layer-ftp.c
```diff
@@ -455,18 +511,23 @@ static int FTPParseRequest(Flow *f, void *ftp_state,
...
+                                            0, state->dyn_port, ALPROTO_FTPDATA, data);
                     if (ret == -1) {
                         FTPFree(data, sizeof(struct FtpTransferCmd));
                         SCLogDebug("No expectation created.");
                         SCReturnInt(-1);
                     } else {
-                        SCLogDebug("Expectation created.");
+                        SCLogDebug("Expectation created [direction: %s, dynamic port %"PRIu16"].",
+                            direction == STREAM_TOCLIENT ? "to client" : "to server",
```
regit: Should we not have same logic a l.514 ? (ie test on state->active)

src/app-layer-ftp.c
```diff
@@ -552,6 +597,29 @@ static int FTPParseResponse(Flow *f, void *ftp_state, AppLayerParserState *pstat
         }
     }
 
+    if (state->command == FTP_COMMAND_EPRT) {
+        uint16_t dyn_port;
+        dyn_port = FTPGetV6PortNumber(state->port_line, state->port_line_len);
```
regit: We don't use Rust implementation here with current code.

src/app-layer-ftp.h
```diff
@@ -115,6 +116,7 @@ typedef struct FtpState_ {
     uint8_t *input;
     int32_t input_len;
     uint8_t direction;
+    uint8_t active;
```
regit: Could we get rid of direction field in favor of active one ?

src/app-layer-ftp.c
```diff
@@ -481,53 +542,37 @@ static int FTPParsePassiveResponse(Flow *f, FtpState *state, uint8_t *input, uin
 {
     uint16_t dyn_port;
 
+    dyn_port =
 #ifdef HAVE_RUST
```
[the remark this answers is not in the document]
jlucovsky: Do you mean moving the call to `rs_ftp_pasv_response` to within `FTPGetV4PortNumber` and similarly for the IPv6 counterparts? 

Note that `FTPGetV4PortNumber` is also called while parsing an FTP response from the function `FTPParseResponse`

src/app-layer-ftp.c
jlucovsky: My understanding is we're setting up an expectation for the data channel in which case the direction is flipped for active mode.

Is that correct?

src/app-layer-ftp.c
jlucovsky: Good point -- will change.

src/app-layer-ftp.c
jlucovsky: Correct -- that's because the `state->port_line` will be something like this: `"EPRT |2|2601:191:8500:2e00:7c0f:78e0:dc5b:f7c1|34325|"` (and similar when using IPv4).

The current Rust functions called from the passive handlers handle an input line like `229 Entering Extended Passive Mode (|||48758|)`

src/app-layer-ftp.h
jlucovsky: Good question ... I see that the `direction` member is only used in `FTPDataParse`. What do you think?

src/app-layer-ftp.c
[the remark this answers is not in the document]
regit: Yes exactly so we have non Rust and Rust code for all parsing.

src/app-layer-ftp.c
regit: Yes I was thinking you could do `state->active ? STREAM_TOSERVER : STREAM_TOCLIENT` or something alike.

src/app-layer-ftp.h
regit: By reading code, `state->direction` is used for the way of the message. So looks like we need to keep it.

src/app-layer-ftp.c
jlucovsky: I'm not sure if that would be more clear.

victorjulien:
@jlucovsky can you please make sure PRs are not updated after they are opened? Push changes to a new branch and then open a new PR when ready.

victorjulien:
Please submit a new PR when you have more changes.

PR #3758, opened 2019-03-30
ftp: Improved FTP active mode handling

jlucovsky:
Link to redmine ticket:

https://redmine.openinfosecfoundation.org/issues/2459
https://redmine.openinfosecfoundation.org/issues/2527
Describe changes:

- Detect FTP active mode over IPv4 and IPv6
- Create expectation based on mode
- Remove return type from TCP, ipv4 options and update tests
- Improve unknown protocol message

Suricata-verify tests: https://github.com/OISF/suricata-verify/pull/28

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -330,6 +330,10 @@ static int FTPParseRequestCommand(void *ftp_state, uint8_t *input,
         fstate->command = FTP_COMMAND_PORT;
     }
 
+    if (input_len >= 4 && SCMemcmpLowercase("eprt", input, 4) == 0) {
+        fstate->command = FTP_COMMAND_EPRT;
+    }
+
     if (input_len >= 8 && SCMemcmpLowercase("auth tls", input, 8) == 0) {
         fstate->command = FTP_COMMAND_AUTH_TLS;
     }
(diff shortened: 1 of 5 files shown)


src/app-layer-ftp.c
```diff
@@ -374,6 +378,55 @@ static void FtpTransferCmdFree(void *data)
...
+}
+
+/**
+ * \brief This function is extract a port number from the command input line for IPv4 FTP usage
+ * \param input input line of the command
+ * \param input_len length of the request
+ *
+ * \retval 0 if a port number could not be extracted; otherwise, thed ynamic port number
+ */
+static int FTPGetV4PortNumber(uint8_t *input, uint32_t input_len)
```
victorjulien: should we have validation in place on: numbers non-negative, values not leading to overflow, no missing values?

src/app-layer-ftp.c
```diff
@@ -552,6 +597,29 @@ static int FTPParseResponse(Flow *f, void *ftp_state, AppLayerParserState *pstat
         }
     }
 
+    if (state->command == FTP_COMMAND_EPRT) {
+        uint16_t dyn_port;
+        dyn_port = FTPGetV6PortNumber(state->port_line, state->port_line_len);
```
victorjulien: declare var on this line itself for more compactness

src/app-layer-ftp.c
```diff
@@ -552,6 +597,29 @@ static int FTPParseResponse(Flow *f, void *ftp_state, AppLayerParserState *pstat
...
+        }
+        state->dyn_port = dyn_port;
+        state->active = 1;
+        SCLogDebug("FTP active mode (v6): dynamic port %"PRIu16"", dyn_port);
+    }
+
+    if (state->command == FTP_COMMAND_PORT) {
+        if ((flags & STREAM_TOCLIENT)) {
+            uint16_t dyn_port;
+            dyn_port = FTPGetV4PortNumber(state->port_line, state->port_line_len);
```
victorjulien: declare var here

src/app-layer-ftp.c
```diff
@@ -552,6 +597,29 @@ static int FTPParseResponse(Flow *f, void *ftp_state, AppLayerParserState *pstat
...
+        if (dyn_port == 0) {
+            return 0;
+        }
+        state->dyn_port = dyn_port;
+        state->active = 1;
+        SCLogDebug("FTP active mode (v6): dynamic port %"PRIu16"", dyn_port);
+    }
+
+    if (state->command == FTP_COMMAND_PORT) {
+        if ((flags & STREAM_TOCLIENT)) {
```
victorjulien: what happens if direction is toserver?

src/app-layer-ftp.h
```diff
@@ -115,6 +116,7 @@ typedef struct FtpState_ {
     uint8_t *input;
     int32_t input_len;
     uint8_t direction;
+    uint8_t active;
```
victorjulien: use 'bool' instead? Plus then true/false instead of 1/0

src/app-layer-ftp.c
jlucovsky: The port number is  an unsigned 16-bite value -- all values other than 0 are valid. I'm not following the "no missing values" as a single value, not a range or other sequence, is being parsed.

src/app-layer-ftp.c
jlucovsky: Will do.

src/app-layer-ftp.c
jlucovsky: will do.

src/app-layer-ftp.c
jlucovsky: No action is necessary -- handling for `EPSV` and `PASV` will parse the parse numbers with those command lines.

src/app-layer-ftp.h
jlucovsky: Will do.

jlucovsky:
Continued in #3768

src/app-layer-ftp.c
victorjulien: My point is that it makes sense to have validation that the values are valid. If the command badly formatted we should set an event and not create the expectation.

PR #3761, opened 2019-04-01
doc: Add manpages for suricatasc and suricatactl

shivan1b:
Add the missing manpages and the corresponding Sphinx configuration
for the command line tools `suricatasc` and `suricatactl`.

These manpages have been given the structure following the manpages for `git-submodules`.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/884

### DIFF
--- a/doc/userguide/manpages/suricatactlfilestore.rst
+++ b/doc/userguide/manpages/suricatactlfilestore.rst
@@ -1,0 +1,51 @@
+Suricata Control Filestore
+==========================
+
+SYNOPSIS
+--------
+
+**suricatactl filestore** [-h] <command> [<args>]
+
+DESCRIPTION
+-----------
+
+**Filestore** allows to store files to disk if a particular signature is matched.
+
+Syntax for usage in suricata.yaml:
+
+**filestore:<direction>,<scope>;**
+
+`direction` can be either of the following.
+
+   1. request/to_server i.e. store a file in the request / to_server direction
+
+   2. response/to_client i.e. store a file in the response / to_client direction
+
+   3: both i.e. store both directions
+
+`scope` can be either:
+
+   1. `file` only store the matching file (for filename,fileext,filemagic matches)
+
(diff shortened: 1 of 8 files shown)


doc/userguide/manpages/suricatactlfilestore.rst
```diff
@@ -0,0 +1,51 @@
...
+**suricatactl filestore** [-h] <command> [<args>]
+
+DESCRIPTION
+-----------
+
+**Filestore** allows to store files to disk if a particular signature is matched.
+
+Syntax for usage in suricata.yaml:
+
+**filestore:<direction>,<scope>;**
```
victorjulien: this looks like rule syntax. Pretty sure it doesn't belong here.

shivan1b: Oops. I was supposed to add both suricata.yaml and rule configuration for filestore. Fixing this. Do you think it should be there at all? Does it make sense according to you to have a manpage dedicated to filestore?

victorjulien: I think so, but this PR seems to introduce 2 of them? suricatactlfilestore.rst and suricatactl-filestore.rst? The latter one looks better.

shivan1b: Actually, this is following the sub-command manpages as by `git-submodules`. They have one for `git-submodules` where they write about the command usage and options and then there is one `gitsubmodules` where they explain what git submodules actually are.
Since `suricatactl` is mostly a WIP and more commands shall be added in future, each command will have their separate man page. Do youthink I should still keep just the latter?

victorjulien: Ah interesting, didn't know what. I think for our purpose we can do without the explanation of what filestore is, and only do the page that explains the usage of the command.

shivan1b:
New PR #3763

PR #3764, opened 2019-04-02
http: wait for response line for filename

catenacyber:
See http evader case 481
https://redmine.openinfosecfoundation.org/issues/1656

Describe changes:
- Waits for request line (if not HTTP0.9) before looking for filename and storing file contents
- This allows to handle junk before response line, like in http evader case 481


catenacyber:
See https://github.com/OISF/libhtp/pull/194 for the big changes

victorjulien:
Could you explain the 'why' of the change in the commit message? It's too terse this way.

PR #3766, opened 2019-04-02
RFC: logging: Anomaly logging

jlucovsky:
*REQUEST FOR COMMENTS*

This changeset adds anomaly logging to suricata.

This changeset is preliminary - it exists to facilitate early and advance
feedback.


pevma:
Was wondering if there is an example of the anomaly logs?

victorjulien:
For the packets that don't have a valid ipv4 or ipv6 header, we probably should follow the same logic as with the alert logger for decoder events.

It would be good to have a few examples as part of the PR.

jlucovsky:
Continued in #3774

PR #3767, opened 2019-04-02
RFC: detect/http: Use v2 inspect and mpm engines

jlucovsky:
*REQUEST FOR COMMENTS* only. Preliminary PR.

This changeset updates the http stub detect logic to use the v2 inspect and npm engines.

### DIFF
--- a/src/detect-http-headers-stub.h
+++ b/src/detect-http-headers-stub.h
@@ -32,6 +32,7 @@
 #include "flow.h"
 #include "app-layer.h"
 #include "app-layer-parser.h"
+#include "app-layer-protos.h"
 #include "app-layer-htp.h"
 
 #include "detect.h"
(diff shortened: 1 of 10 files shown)


src/detect-http-headers-stub.h
```diff
@@ -49,259 +50,62 @@
...
-    htp_header_t *h = (htp_header_t *)htp_table_get_c(tx->request_headers,
-                                                      HEADER_NAME);
-    if (h == NULL || h->value == NULL) {
-        SCLogDebug("HTTP %s header not present in this request", HEADER_NAME);
-        return;
-    }
+        htp_header_t *h = (htp_header_t *)htp_table_get_c(tx->request_headers,
+                                                          HEADER_NAME);
+        if (h == NULL || h->value == NULL) {
+            SCLogDebug("HTTP cookie header not present in this request");
```
victorjulien: this should be changed back to something like 'SCLogDebug("HTTP %s header not present in this request", HEADER_NAME);'

src/detect-http-headers-stub.h
```diff
@@ -49,259 +50,62 @@
...
-    if (tx->response_headers == NULL)
-        return;
-    const HtpTxUserData *htud = (const HtpTxUserData *)htp_tx_get_user_data(tx);
-    /* if the request wasn't flagged as having a trailer, we skip */
-    if (htud && !htud->response_has_trailers)
-        return;
+        htp_header_t *h = (htp_header_t *)htp_table_get_c(tx->response_headers,
+                                                          HEADER_NAME);
+        if (h == NULL || h->value == NULL) {
+            SCLogDebug("HTTP header not present in this request");
```
victorjulien: would be nice to name the header that was not found here

src/detect-http-headers-stub.h
```diff
@@ -324,30 +128,55 @@ static int DetectHttpHeadersSetup(DetectEngineCtx *de_ctx, Signature *s, const c
...
+ * \brief this function setup the http.header keyword used in the rule
+ *
+ * \param de_ctx   Pointer to the Detection Engine Context
+ * \param s        Pointer to the Signature to which the current keyword belongs
+ * \param str      Should hold an empty string always
+ *
+ * \retval 0       On success
+ */
+static int DetectHttpHeadersSetupSticky(DetectEngineCtx *de_ctx, Signature *s, const char *str)
+{
```
victorjulien: this needs error handling, the proper way is https://github.com/OISF/suricata/blob/master/src/detect-http-client-body.c#L159

src/detect-http-headers-stub.h
```diff
@@ -324,30 +128,55 @@ static int DetectHttpHeadersSetup(DetectEngineCtx *de_ctx, Signature *s, const c
...
+    DetectBufferSetActiveList(s, g_buffer_id);
+    s->alproto = ALPROTO_HTTP;
+    return 0;
+}
+
 static void DetectHttpHeadersRegisterStub(void)
 {
+    sigmatch_table[KEYWORD_ID].name = KEYWORD_NAME_LEGACY;
```
victorjulien: this is overwritten by the next sets as all are setting to sigmatch_table[KEYWORD_ID]. The KEYWORD_NAME_LEGACY should be set in sigmatch_table[KEYWORD_ID].alias instead.

src/detect-http-server.c
```diff
@@ -33,6 +33,7 @@
  * used by the origin server to handle the request.
  */
 
+#define KEYWORD_NAME_LEGACY "http.server"
```
victorjulien: I guess this doesn't make sense, leave it out and add some ifdefs to the template?

src/detect-http-headers-stub.h
jlucovsky: Should there be two `sigmatch_table` entries corresponding `http.name` and `http_name` as is done here: https://github.com/OISF/suricata/blob/master/src/detect-http-cookie.c#L86:L103?

Or, should the `.alias` member be set to the legacy value?

src/detect-http-headers-stub.h
jlucovsky: Will fix.

src/detect-http-headers-stub.h
jlucovsky: Will fix.

src/detect-http-headers-stub.h
jlucovsky: Will fix.

src/detect-http-server.c
jlucovsky: Right ... and the same applies to `detect-http-location.c

src/detect-http-headers-stub.h
victorjulien: The reason why the cookie case is different is that there we went from the content modifier
```
content:"abc"; http_cookie;
```
to the sticky buffer:
```
http.cookie; content:"abc";
```
These have very different mechanics during the rule parsing, so they are registered as 2 different keywords.

In this case of the headers all keywords are already sticky buffers, so we can simply register the new name and keep the old name as an alias.

src/detect-http-server.c
victorjulien: Yes indeed.

src/detect-http-headers-stub.h
jlucovsky: Understood. Will fix.

src/detect-http-server.c
jlucovsky: Will remove from detect-http-{server,location}.c

jlucovsky:
Continued in #3775

PR #2485, opened 2017-01-14
decode-icmpv6: add missing types

norg:
There have been some ICMPv6 types missing within the DecodeICMPV6 that
are added by this commit and the code check is adjusted to always use
the DEFINE.

This fixes: https://redmine.openinfosecfoundation.org/issues/1972

- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/42
- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/42

PR #3773, opened 2019-04-04
http: logs content range

catenacyber:
Fixes #2485

https://redmine.openinfosecfoundation.org/issues/2485

Describe changes:
- Logs http content-range header value (in responses) in json output

This is a first easy step to provide the data.
Maybe it makes more sense to transfer it in the `fileinfo` section

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range for

### DIFF
--- a/src/output-json-http.c
+++ b/src/output-json-http.c
@@ -261,6 +261,13 @@ static void JsonHttpLogJSONBasic(json_t *js, htp_tx_t *tx)
                 *p = '\0';
             json_object_set_new(js, "http_content_type", SCJsonString(string));
         }
+        htp_header_t *h_content_range = htp_table_get_c(tx->response_headers, "content-range");
+        if (h_content_range != NULL) {
+            const size_t size = bstr_len(h_content_range->value) * 2 + 1;
+            char string[size];
+            BytesToStringBuffer(bstr_ptr(h_content_range->value), bstr_len(h_content_range->value), string, size);
+            json_object_set_new(js, "http_content_range", SCJsonString(string));
+        }
     }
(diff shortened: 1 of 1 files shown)


victorjulien:
I agree it should go into the fileinfo records as well.

Since the goal is to allow for automated processing, I wonder if we should log a parsed header instead of the raw value. So a json object with the range.

catenacyber:
> I agree it should go into the fileinfo records as well.

Do you mean that it should go in both `fileinfo` and `http` records ? or in `fileinfo` only ?

> Since the goal is to allow for automated processing, I wonder if we should log a parsed header instead of the raw value. So a json object with the range.

Ok, but what should we do in the case the parsing fails ? For instance if I get `Content-Range: bytes -512` ie a missing start

victorjulien:
> > I agree it should go into the fileinfo records as well.
> 
> Do you mean that it should go in both `fileinfo` and `http` records ? or in `fileinfo` only ?

Both.

> > Since the goal is to allow for automated processing, I wonder if we should log a parsed header instead of the raw value. So a json object with the range.
> 
> Ok, but what should we do in the case the parsing fails ? For instance if I get `Content-Range: bytes -512` ie a missing start

Hmm good question. I'm doubting between leaving it out (this is something that could be alerted on) and adding an extra (optional?) field for the raw value.

catenacyber:
> I'm doubting between leaving it out (this is something that could be alerted on) and adding an extra (optional?) field for the raw value.

I went for the raw value (and did not parse the units field)

PR #3777, opened 2019-04-05
Feat/user mode/v4

victorjulien:
Add a distinction between SYSTEM and USER modes when running. This is based on the runmode, where reading a pcap leads to user mode.

Currently the only consequence of the user mode is that the default-log-dir from the yaml is ignored and is set to '.' instead. If -l is used on the commandline, that still takes precedence.

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/245
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/247

### DIFF
--- a/src/runmodes.c
+++ b/src/runmodes.c
@@ -823,7 +823,7 @@ void RunModeInitializeOutputs(void)
                 OutputInitResult r = module->InitFunc(output_config);
                 if (!r.ok) {
                     FatalErrorOnInit(SC_ERR_INVALID_ARGUMENT,
-                        "output module setup failed");
+                        "output module \"%s\": setup failed", output->val);
                     continue;
                 } else if (r.ctx == NULL) {
                     continue;
--- a/src/runmodes.c
+++ b/src/runmodes.c
@@ -492,6 +492,19 @@ int RunModeOutputFiledataEnabled(void)
     return filedata_logger_count > 0;
 }
 
+bool IsRunModeSystem(enum RunModes run_mode_to_check)
+{
+    switch (run_mode_to_check) {
+        case RUNMODE_PCAP_FILE:
+        case RUNMODE_ERF_FILE:
+        case RUNMODE_ENGINE_ANALYSIS:
+            return false;
+            break;
+        default:
+            return true;
+    }
+}
+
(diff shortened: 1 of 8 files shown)


victorjulien:
https://redmine.openinfosecfoundation.org/issues/2421

abungay:
I think it might be helpful to revisit this and add a note to the docs for the suricata command line option '-r' description or similar to indicate this behaviour that the log dir will change when using this argument. When I first started trying out Suricata I thought my configuration file had issues since I was only looking at my default log dir for output using suricata -r <pcap> but I didn't know to check my CWD for pcap log output instead. I finally found out that CWD was being used instead with help from this [stackoverflow post](https://stackoverflow.com/questions/61132410/how-to-run-suricata-on-pcap-mode-and-get-results-in-fast-log) explaining what was happening.

victorjulien:
> I think it might be helpful to revisit this and add a note to the docs for the suricata command line option '-r' description or similar to indicate this behaviour that the log dir will change when using this argument. When I first started trying out Suricata I thought my configuration file had issues since I was only looking at my default log dir for output using suricata -r but I didn't know to check my CWD for pcap log output instead. I finally found out that CWD was being used instead with help from this [stackoverflow post](https://stackoverflow.com/questions/61132410/how-to-run-suricata-on-pcap-mode-and-get-results-in-fast-log) explaining what was happening.

Agreed. Could you turn this into a ticket and/or PR?

abungay:
> > I think it might be helpful to revisit this and add a note to the docs for the suricata command line option '-r' description or similar to indicate this behaviour that the log dir will change when using this argument. When I first started trying out Suricata I thought my configuration file had issues since I was only looking at my default log dir for output using suricata -r but I didn't know to check my CWD for pcap log output instead. I finally found out that CWD was being used instead with help from this [stackoverflow post](https://stackoverflow.com/questions/61132410/how-to-run-suricata-on-pcap-mode-and-get-results-in-fast-log) explaining what was happening.
> 
> Agreed. Could you turn this into a ticket and/or PR?

Ticket created at https://redmine.openinfosecfoundation.org/issues/3762.

PR #3789, opened 2019-04-10
init: pledge(2) needs "fattr" during suricata reload.

bluhm:
When killed with SIGHUP, suricata reopens the log files.  If filemode
is set in the config, it needs pledge promise "fattr" to allow the
chmod(2) on OpenBSD.

### DIFF
--- a/src/util-privs.c
+++ b/src/util-privs.c
@@ -239,7 +239,7 @@ int SCGetGroupID(const char *group_name, uint32_t *gid)
 #ifdef __OpenBSD__
 int SCPledge(void)
 {
-    int ret = pledge("stdio rpath wpath cpath unix dns bpf", NULL);
+    int ret = pledge("stdio rpath wpath cpath fattr unix dns bpf", NULL);
 
     if (ret != 0) {
         SCLogError(SC_ERR_PLEDGE_FAILED, "unable to pledge,"


victorjulien:
@eroullit can you have a look?

eroullit:
Indeed, a logrotation with a specific `filemode` option in `suricata.yaml` will trigger `chmod(2)` after `pledge(2)`.
To allow this feature to operate properly, the `fattr` promise is required.
Without the `fattr` promise:
```
[79646] 11/4/2019 -- 14:11:56 - (tm-threads.c:2157) <Notice> (TmThreadWaitOnThreadInit) -- all 3 packet processing threads, 4 management threads initialized, engine started.                         

Program received signal SIGABRT, Aborted.
[Switching to thread 512136]
chmod () at -:3
3       -: No such file or directory.
        in -
Current language:  auto; currently asm
(gdb) bt
#0  chmod () at -:3
#1  0x00000f93ff5fda8d in SCLogOpenFileFp (
    path=0xf969cbba800 "logs/eve-%Y-%m-%d-%H:%M.json", append_setting=Variable "append_setting" is not available.
)
    at util-logopenfile.c:284
#2  0x00000f93ff5fde7e in SCConfLogReopen (log_ctx=Variable "log_ctx" is not available.
) at util-logopenfile.c:508
[...]
```

With the @bluhm changeset, suricata keeps operating either after a `SIGHUP` or once the `rotate-interval` expired.
I am double checking if some other promises are missing.

bluhm:
I have commited pledge support with this fix to the OpenBSD ports tree, so it gets tested there.  Would be nice if we could get the fix also upstream.

victorjulien:
Merged in #3818, thanks a lot @bluhm

PR #3799, opened 2019-04-15
ctl/filestore: Add check for filestore directory

shivan1b:
Up until now, suricatactl would delete any directory that is provided as
an argument on command line. This patch adds a basic test for the
directories `tmp`, `00` and `ff` in order to justify that the provided
directory is actually a filestore directory.

Additionally, some code has been broken up and made more readable and
pythonic.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2843

### DIFF
--- a/python/suricata/ctl/filestore.py
+++ b/python/suricata/ctl/filestore.py
@@ -82,22 +82,32 @@ def remove_file(path, dry_run):
     return size
 
 
-def prune(args):
+def set_logger_level(args):
     if args.verbose:
         logger.setLevel(logging.DEBUG)
     if args.quiet:
         logger.setLevel(logging.WARNING)
 
-    if not args.directory:
-        print(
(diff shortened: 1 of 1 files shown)


python/suricata/ctl/filestore.py
```diff
@@ -82,22 +82,32 @@ def remove_file(path, dry_run):
...
-        print(
-            "error: the filestore directory must be provided with --directory",
-            file=sys.stderr)
-        return 1
 
-    if not args.age:
-        print("error: no age provided, nothing to do", file=sys.stderr)
-        return 1
+def perform_sanity_checks(args):
+    set_logger_level(args)
```
jasonish: This should probably be in the prune() method, but we can fix that up another time.

victorjulien:
Merged in #4188, thanks!

PR #3800, opened 2019-04-16
Privdrop without libcap ng v2

eroullit:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2931

Describe changes:
- Introduces `SCPrivsInit()` to allow suricata to drop privileges and apply capabilities (when enabled)
- Introduces `SCPrivsRaise()` to allow suricata to raise privileges to original state, based the process saved UID/GID for specific code path during processing/teardown
- Introduces `SCPrivsDrop()` to drop privileges down again once the privileged code section is over.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable): N/A

### DIFF
--- a/doc/userguide/configuration/dropping-privileges.rst
+++ b/doc/userguide/configuration/dropping-privileges.rst
@@ -1,8 +1,23 @@
 Dropping Privileges After Startup
 =================================
 
-Currently, libcap-ng is needed for dropping privileges on Suricata
-after startup. For libcap, see status of feature request number #276
+Currently, dropping privileges can be either done via libcap-ng or with POSIX
+seteuid()/setegid().
+
+When the privilege drop is done by libcap-ng, the process real, effective and saved
+user/group ID is set to the values provided by the user configuration.
+Furthermore, process capabilities are set ensure that operations requiring
+higher permissions can be performed properly.
+Once done, the privilege level stays unchanged for the lifetime of the process.
+
+When libcap-ng support is disabled, privilege drop is done via
+seteuid()/setegid() on startup. Operations requiring higher permissions are
+placed between the functions SCPrivsRaise()/SCPrivsDrop().
+
+Between SCPrivsRaise() and SCPrivsDrop(), the process privileges are raised back
+to the saved user/group level during the process runtime.
+
+For libcap, see status of feature request number #276
 -- Libcap support for dropping privileges.
 
 Most distributions have ``libcap-ng`` in their repositories.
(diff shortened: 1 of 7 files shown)


doc/userguide/configuration/dropping-privileges.rst
```diff
@@ -1,8 +1,23 @@
...
 =================================
 
-Currently, libcap-ng is needed for dropping privileges on Suricata
-after startup. For libcap, see status of feature request number #276
+Currently, dropping privileges can be either done via libcap-ng or with POSIX
+seteuid()/setegid().
+
+When the privilege drop is done by libcap-ng, the process real, effective and saved
+user/group ID is set to the values provided by the user configuration.
+Furthermore, process capabilities are set ensure that operations requiring
```
norg: are set **to**

norg:
docs look good although I'm not certain if we want to have the function calls in a userguide document, might be better within a development document :)

src/util-privs.c
```diff
@@ -136,6 +149,84 @@ void SCDropCaps(ThreadVars *tv)
...
+
+#else /* HAVE_LIBCAP_NG */
+
+static void SCSetGroupID(const uint32_t gid)
+{
+    int ret = setegid(gid);
+
+    if (ret != 0) {
+        SCLogError(SC_ERR_GID_FAILED, "unable to set the group ID,"
+                " check permissions!! gid=%u ret=%i errno=%i", gid, ret, errno);
```
victorjulien: We can use ```FatalError``` instead of SCLogError+exit.

Also I think its nicer to print the error string ```strerror(errno)``` than the numeric code.

victorjulien:
> I wonder if we should leave the linux only things out of this (afpacket, ethtool ioctl). For those we have libcap-ng. Perhaps we should simply make that lib mandatory for linux.

@satta @chifflier any thoughts on this from a  Debian perspective?

satta:
> > I wonder if we should leave the linux only things out of this (afpacket, ethtool ioctl). For those we have libcap-ng. Perhaps we should simply make that lib mandatory for linux.
> 
> @satta @chifflier any thoughts on this from a Debian perspective?

Personally I don't see much in the way of depending hard on libcap-ng for Linux. It's one additional dependency, but one that seems to move rather slowly development-wise (so we're not impacted by Debian's slow adaptation of new versions) and that builds on all platforms currently supported by Suricata on Debian (that is, only those using the Linux kernel).

norg:
Should we take the discussion to the redmine ticket again and close this PR for now until a decision is made?

PR #3804, opened 2019-04-18
ssh : code style consistency

catenacyber:
None

Describe changes:
- Adds SSH_FLAG_VERSION_PARSED to flags before each return
This way, we are sure SSHParseBanner does not get called again
And proto_version does not get leaked

Another way could be to call `SCFree(header->proto_version);` instead of `header->flags |= SSH_FLAG_VERSION_PARSED;`

Maybe we should return 0 instead of -1 if `(sw_ver_len <= 1)` so as to continue ssh parsing...

This was found by fuzzing as a potential leak (not present in suricata as returning -1 ensures we stop parsing)


victorjulien:
Merged as part of #3810, thanks Philippe!

PR #3806, opened 2019-04-18

catenacyber:
Fixes #1125

Describe changes:
- Adds a pattern to recognize SMTP traffic from server

How can we test this ?
Should we try to recognize SMTP streams without seeing them start ? (In this case, we can add others patterns such as "250 ")

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -1578,6 +1578,11 @@ static int SMTPRegisterPatternsForProtocolDetection(void)
     {
         return -1;
     }
+    if (AppLayerProtoDetectPMRegisterPatternCI(IPPROTO_TCP, ALPROTO_SMTP,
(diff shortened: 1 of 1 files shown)


victorjulien:
On thing that is tricky is that ftp, smtp, irc and maybe others are very much alike. So it's hard to be sure based on a single pattern.

catenacyber:
Thanks Victor.
What is the status of protocol detection https://redmine.openinfosecfoundation.org/issues/2757 ?

I thought Suricata kept a list of possible protocols for a flow and had different heuristics (like client and server speaking the same protocol, try whole parsing, etc...) to rule them out
How can we do otherwise ?

victorjulien:
I think this is also a valid start of FTP. On of the recent additions is that you can register a probing parser that is invoked after a pattern matched. Perhaps this is the way to go for ftp/smtp. That we have a PP that can figure out if it really is smtp or ftp.

PR #3813, opened 2019-04-23
Fuzz v1

catenacyber:

Describe changes:
- Adds two fuzz targets

This is a step for oss-fuzz initial integration
For ideal integration, there will remain more fuzz targets to be written to cover most of the code.
And we should also work on corpuses for those fuzz targets.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@
     AC_CONFIG_HEADERS([config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
     AC_CONFIG_MACRO_DIR(m4)
-    AM_INIT_AUTOMAKE
+    AM_INIT_AUTOMAKE([subdir-objects])
 
     AC_LANG_C
     AC_PROG_CC_C99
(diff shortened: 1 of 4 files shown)


victorjulien:
The test files look quite hacky, with lots of duplication between the 2. Also with their own initialization steps I fear we're going to be chasing our tails there a lot. We've already experienced this with the unix socket mode in the past.

Another thing that caught my eye is the naming of the 2 tests: I think they need to follow the code names more closely. So something like applayerparse instead of layerparse.

What I would also like to see early on is a guide on how to set this up and how to use it.

catenacyber:
> What I would also like to see early on is a guide on how to set this up and how to use it.

Here is something :
- install docker
- run git clone --branch suricata --depth 1 https://github.com/catenacyber/oss-fuzz
(we will use the orignal google repo once we merge this)
- change directory into cloned repository : cd oss-fuzz
- run python infra/helper.py build_image suricata
- run python infra/helper.py build_fuzzers --sanitizer address suricata
You can use undefined sanitizer (memory sanitizer does not work yet see https://github.com/google/oss-fuzz/issues/2145#issuecomment-485781098 )
- run python infra/helper.py run_fuzzer suricata fuzz_layerparse
(or another fuzz target such as fuzz_detectproto)

Is this what you expected as a "a guide on how to set this up and how to use it" ?

I will take your remarks into account later

catenacyber:
> The test files look quite hacky

I am not sure what you mean here, and what should be done about it.

> with lots of duplication between the 2

These duplicated code lines are also in suricata.c
Should I create a new file, ie globals.c, to put all duplicated code in it ? 

> Also with their own initialization steps I fear we're going to be chasing our tails there a lot

Do you think there should be a global `init` function that should be good for all fuzz targets and suricata itself ? (Even if for the fuzz targets, we will initialize stuff that we do not use, such as a `AppLayerProtoDetectGetCtxThread` for fuzz_applayerparserparse.c)

I was rather thinking that we should initialize the least things possible (as is done for the unit tests such as `HTPParserTest01`)

> Another thing that caught my eye is the naming of the 2 tests: I think they need to follow the code names more closely. So something like applayerparse instead of layerparse.

ok

victorjulien:
Yes, I think a global init would make a lot of sense. We've started doing this between normal modes and unix socket mode, and it has resulted in far fewer issues. It would probably be good if (most of) the unittests would start using it eventually. I can't even begin to count the lost time resulting from initialization mismatches.

catenacyber:
Ok, I will work on this global `init`

PR #3823, opened 2019-05-02

catenacyber:
Use of SMTPProbingParserTc to distinguish from FTP
after pattern recognition

Describe changes:
- Adds a pattern and a probing parser to recognize SMTP traffic from server

Modified from #3806 :
- Adds probing parser after pattern recognition to distinguish from FTP
- Adds "220-" pattern for multi-line response (see example input.pcap from suricata-verify/tests/smtp)

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -1566,6 +1566,42 @@ static int SMTPStateGetEventInfo(const char *event_name,
     return 0;
 }
 
+static AppProto SMTPProbingParserTc(Flow *f, uint8_t direction,
+        uint8_t *input, uint32_t ilen, uint8_t *rdir)
+{
+    // Server banner is 220, space or hyphen, name, protocol, server agent
+    // protocol can be either SMTP or ESTMP (extended SMTP)
+
+    // Skips response code
+    size_t pos = 4;
+    if (ilen < 4)
+        return ALPROTO_UNKNOWN;
+
+    // Skips name
+    while (pos < ilen && input[pos] != ' ')
+        pos++;
+    pos++;
+    if (pos + 4 >= ilen) {
+        return ALPROTO_UNKNOWN;
+    }
+    // Checks protocol
+    if (memcmp(input+pos, "SMTP", 4) == 0) {
+        return ALPROTO_SMTP;
+    }
+    if (pos + 5 >= ilen) {
+        // Be careful for packet splitting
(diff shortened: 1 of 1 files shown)


victorjulien:
Can you add a comment on what the expected data looks like? And a verify test.

src/app-layer-smtp.c
```diff
@@ -1566,6 +1566,42 @@ static int SMTPStateGetEventInfo(const char *event_name,
...
+    pos++;
+    if (pos + 4 >= ilen) {
+        return ALPROTO_UNKNOWN;
+    }
+    // Checks protocol
+    if (memcmp(input+pos, "SMTP", 4) == 0) {
+        return ALPROTO_SMTP;
+    }
+    if (pos + 5 >= ilen) {
+        // Be careful for packet splitting
```
victorjulien: it's not clear to me why this is done. Can't we just return UNKNOWN if we don't have enough data?

catenacyber: > Can't we just return UNKNOWN if we don't have enough data?

That is precisely what I tried to do :
Enough is 4 or 5 bytes because we can have either `SMTP` or `ESMTP` as a pattern
With TCP streaming, we are not supposed to know in how many packets this content will be.

Another way is that we can return UNKNOWN if we have less than 5 bytes.
So, we will return UNKOWN if we just GET `SMTP`, thus it is less precise.
But then, I guess that we will have more data coming afterwards from the server and then return `ALPROTO_SMTP` as expected. 
And anyways, regular servers have at least end of lines characters after `SMTP` pattern and send this in one packet.

Do you prefer it this way ?

catenacyber:
> Can you add a comment on what the expected data looks like? 

Done
Waiting for your decision about the other point to submit a new pull request.

> And a verify test.

See https://github.com/OISF/suricata-verify/pull/53

src/app-layer-smtp.c
victorjulien: If you return unknown you will be called again with the same data and more, so I think you could do something like:

```c
if (memcmp( ... "SMTP") == 0)
   return ALPROTO_SMTP;
if (len < 5)
   return ALPROTO_UNKNOWN; // we'll be called again with more data
if (memcmp(..., "ESMTP") == 0)
   return ALPROTO_SMTP;
```

catenacyber: ok

PR #3824, opened 2019-05-02
http: configures libhtp to allw spaces in uri

catenacyber:
https://redmine.openinfosecfoundation.org/issues/2881

Describe changes:
- configures libhtp to allow spaces in URIs

See https://github.com/OISF/libhtp/pull/208

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -2264,6 +2264,7 @@ static void HTPConfigSetDefaultsPhase1(HTPCfgRec *cfg_prec)
     htp_config_register_response_complete(cfg_prec->cfg, HTPCallbackResponse);
 
     htp_config_set_parse_request_cookies(cfg_prec->cfg, 0);
+    htp_config_set_allow_space_uri(cfg_prec->cfg, 1);
 
     /* don't convert + to space by default */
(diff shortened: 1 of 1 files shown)


catenacyber:
CI will fail before OISF/libhtp#208 is merged

src/app-layer-htp.c
```diff
@@ -2264,6 +2264,7 @@ static void HTPConfigSetDefaultsPhase1(HTPCfgRec *cfg_prec)
     htp_config_register_response_complete(cfg_prec->cfg, HTPCallbackResponse);
 
     htp_config_set_parse_request_cookies(cfg_prec->cfg, 0);
+    htp_config_set_allow_space_uri(cfg_prec->cfg, 1);
```
victorjulien: I think it would be good to check for this in configure and guard it with ifdefs. We did this in the past with HAVE_HTP_URI_NORMALIZE_HOOK (see git grep HAVE_HTP_URI_NORMALIZE_HOOK, although it's now unused and can be removed)

catenacyber: Ok, I did not add it to `features` in suricata.c
Should I ?

PR #3832, opened 2019-05-06

catenacyber:
Use of SMTPProbingParserTc to distinguish from FTP after pattern recognition

Link to redmine ticket:

Describe changes:
- Adds a pattern and a probing parser to recognize SMTP traffic from server

Modified from #3823 :

Added examples in comments of expected lines
Return `ALPROTO_UNKNOWN` if we do not have enough data to be sure to recognize

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -1566,6 +1566,43 @@ static int SMTPStateGetEventInfo(const char *event_name,
     return 0;
 }
 
+static AppProto SMTPProbingParserTc(Flow *f, uint8_t direction,
+        uint8_t *input, uint32_t ilen, uint8_t *rdir)
+{
+    // Server banner is 220, space or hyphen, name, protocol, server agent
+    // protocol can be either SMTP or ESTMP (extended SMTP)
+    // Examples :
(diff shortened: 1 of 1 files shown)


victorjulien:
I think the problem with this approach is that AFAICS the 220 line doesn't need to have SMTP or ESMTP in the string. So their absence does not prove the protocol is not SMTP, but their presence does make it very likely that it is SMTP. Not entirely sure how to deal with this.

catenacyber:
As I understand, there are two points :
- Regular traffic
Does this patch return the correct answer for all regular traffic ?
All pcaps I found have this pattern
But in the case of its absence (in the first line), we may want to return `ALPROTO_UNKNOWN` instead of `ALPROTO_FAILED`
Do you want me to do these changes ?

- Malicious server
Let's suppose that the attacker controls the SMTP server and wants to avoid detection by Suricata.
The easy solution I see is to replace SMTP by FTP in this banner so that Suricata thinks that the server flow is FTP instead of SMTP.
What happens if Suricata finds that a SMTP client speaks to a FTP server ?

catenacyber:
Test in https://github.com/OISF/suricata-verify/pull/53

catenacyber:
see https://redmine.openinfosecfoundation.org/issues/2757 first

catenacyber:
Closing at this needs rework after #4605, #4607 (especially) and #4681

PR #3835, opened 2019-05-07
Fuzz v2

catenacyber:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2859

Describe changes:
- Adds two fuzz targets
- Change compilation scheme so as to compile fuzz targets (option --enable-fuzztargets for configure)

Modifies #3813 
- renames fuzz targets to match the functions names
- create a new file init.c for common initialization between real suricata and fuzz targets (that means getting code out of suricata.c into init.c)
- calls `ConfInit` before `WindowsInitService` 
- new function `initGlobal` to do all the initialization before command line parsing

I think that this PR is not complete but I would like some feedback, that is why I am doing it.
I kept the history for commits so as to follow my logic.

There remains some more initialization logic in the fuzz targets.
These are inspired by afl-fuzz logic from `ParseCommandLineAFL` in suricata.c
Should they be merged together into one common initialization function ?
Do you know if these steps (like `MpmTableSetup`) depend on command line arguments ? 
(so should we merge them either into `initGlobal` or into some new `initWithArgs`)

PR #3837, opened 2019-05-07
Fuzz v3

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds three fuzz targets

Modifies #3835 :
- adds fuzz target for `SigInit` (rule parsing)
- Adds a README
- Adds onefile.c simple driver
- Code style fixes as reviewed

I think that this PR is not complete but I would like some feedback, that is why I am doing it.
I kept the history for commits so as to follow my logic.


src/tests/fuzz/fuzz_applayerparserparse.c
```diff
@@ -0,0 +1,101 @@
...
+
+
+#include "suricata-common.h"
+#include "app-layer-detect-proto.h"
+#include "flow-util.h"
+#include "app-layer-parser.h"
+
+#define HEADER_LEN 6
+
+void fuzz_openFile(const char * name) {
```
SirkoHoeer: this function is not necessary.

src/tests/fuzz/fuzz_applayerparserparse.c
```diff
@@ -0,0 +1,101 @@
...
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    Flow * f;
+    TcpSession ssn;
+
+    if (size < HEADER_LEN) {
+        return 0;
+    }
+
+    if (alp_tctx == NULL) {
```
SirkoHoeer: you could use an init function i.e. int LLVMFuzzerInitialize(int *argc, char ***argv)

catenacyber: Yes but that is not recommended by libfuzzer : http://llvm.org/docs/LibFuzzer.html#startup-initialization

src/tests/fuzz/fuzz_applayerparserparse.c
catenacyber: Indeed, it is not used yet.
In other projects, I use this function to open a file for logging.

How do you achieve this ?

src/tests/fuzz/fuzz_applayerparserparse.c
SirkoHoeer: Ok, you are right. I prefer this kind of solution

src/tests/fuzz/README
```diff
@@ -0,0 +1,11 @@
+How to run fuzzing ?
```
victorjulien: how do I control which Suricata repo+branch is used?

catenacyber: This is in `Dockerfile` used by oss-fuzz
See https://github.com/google/oss-fuzz/pull/2533/files#diff-f132db5f0e91b1506efa943dd4eb789bR30

PR #3848, opened 2019-05-11
app-layer-ftp: Potential memory leak fixed

jlucovsky:
Ensure that when handling failures during STOR command
processing, that all memory is freed on the error path.

Link to [redmine](https://redmine.openinfosecfoundation.org/attachments/1486) ticket:

Describe changes:
- Ensure memory for both command and file name is returned, as needed, on error path
- Ensure FTP memcap values are correct when returning memory via `FtpTransferCmdFree`

According to @pevma, the original problem has not recurred recently. While examining the source code, the issues addressed by this PR were noticed.


victorjulien:
Merged in #3864, thanks!

PR #3721, opened 2019-03-11
Ebpf xdp 5.0

regit:
eBPF and XDP part of #3716. It contains one patch of #3719 as this would have conflict otherwise.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Change flow timeout logic to avoid ktime in eBPF
- XDP hardware bypass compatible with netronome
- Persistant flow bypass via eBPF pinned maps and reload of bypassed flow tables at start
- Global XDP bypass switch to avoid interrupt when Suricata is down
- Updated filter.bpf providing IPv4 block list 

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/441
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/222

PR #3849, opened 2019-05-12
RFC: Ebpf xdp 5.0 v2.3

regit:
Rework of #3721 addressing most comments. Only comment missing is that VLAN support for eBPF bypass filter is still incomplete.

- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made

A few choices need to be validated here.

First, XDP filter code bypass TLS 1.2 encrypted packets seen on port 443. I did set it to on by default.

Second, a Flow storage is now used to store the bypass counters. Currently, the registration of the storage is not optional. We could win some memory by making it optional but this would require a mechanism to do so. Possibilities I see are:
- A variable at the root of the yaml
- A function call that can be done in the interfaces configuration parsing code. Problem with this approach is that in the order of call in init phase we don't parse the interfaces before finalizing the storage. So we would have to write a custom parsing function to do so.

Describe changes:
- Use a flow storage for the bypass counter to limit the impact on non bypass user
- Introduce a generic system to get counters of actually bypassed flows
- TLS encrypted packets bypass in XDP code
- Bug fixes and code cleaning

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/449
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/230


ebpf/xdp_filter.c
```diff
@@ -287,6 +309,37 @@ static int __always_inline filter_ipv4(struct xdp_md *ctx, void *data, __u64 nh_
 #endif
     }
 
+#if ENCRYPTED_TLS_BYPASS
+    if ((dport == __constant_ntohs(443)) || (sport == __constant_ntohs(443))) {
```
victorjulien: this seems to open a big hole in the ids... as long as you send packets that have port 443 as src or dst (tcp and udp?) and use these 3 first bytes, Suricata will be blind to them?

regit: Yes, full bypass it is.

regit: I can make it deactivated by default and let the user decide.

victorjulien: I don't see how this could ever be a good idea really. I mean you have no idea if the packets are really encrypted and part of a valid TLS stream.

I can see why you'd want it, but this seems to open up things way too much.

Are there other ways we could do this? Track TLS more properly in the filter or re-evaluate the in flight packets, or maybe other things can be done?

regit: OK, updated, going to propose an update

regit:
OK, I'm gonna submit a rebased version with fixes.

doc/userguide/capture-hardware/ebpf-xdp.rst
```diff
@@ -249,13 +258,25 @@ also use the ``/etc/suricata/ebpf/xdp_filter.bpf`` (in our example TCP offloadin
...
     use-mmap: yes
     ring-size: 200000
+    # Uncomment the following if you are using hardware XDP with
+    # a card like Netronome
+    # use-percpu-hash: no
 
 
 XDP bypass is compatible with AF_PACKET IPS mode. Packets from bypassed flows will be send directly 
-from one card to the second card without going by the kernel network stack.
+from one card to the second card without going by the ker nel network stack.
```
norg: did the kernel split into the **ker** and **nel** :)?

doc/userguide/capture-hardware/ebpf-xdp.rst
```diff
@@ -249,13 +258,25 @@ also use the ``/etc/suricata/ebpf/xdp_filter.bpf`` (in our example TCP offloadin
...
+
+If you are using hardware XDP offload you may have to set ``use-percpu-hash`` to false and
+build and install the XDP filter file after setting ``USE_PERCPU_HASH`` to 0.
 
-Setup symmetric hashing on the NIC
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+In the XDP filter file, you can set ``ENCRYPTED_TLS_BYPASS`` to 0 if you don't want to bypass
+the encrypted TLS 1.2 packetsin the eBPF code.
```
norg: **packets** **in**

doc/userguide/capture-hardware/ebpf-xdp.rst
```diff
@@ -249,13 +258,25 @@ also use the ``/etc/suricata/ebpf/xdp_filter.bpf`` (in our example TCP offloadin
...
+
+If you are not using vlan tracking (``vlan.use-for-tracking`` set to false in suricata.yaml) then you have also to set
```
norg: **then you also have to set**

PR #3851, opened 2019-05-13

thus:
Add new keyword `tls.cert` to match on the 'raw' TLS certificate. This keyword matches on the first certificate in the certificate chain.

Example:
```alert tls any any -> any any (msg:"test tls.cert"; tls.cert; content:"|01 02 03 04|"; sid:1;)```

Updates:
- Fixed changes requested in #3545
- Renamed keywords from "tls_" to "tls." (Redmine Feature 2283)

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/182
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/182

### CODE REVIEW

File: src/detect-tls-cert.c [hunk 1]
```diff
@@ -0,0 +1,142 @@
...
+{
+    InspectionBuffer *buffer = InspectionBufferGet(det_ctx, list_id);
+    if (buffer->inspect == NULL) {
+        const SSLState *ssl_state = (SSLState *)f->alstate;
+
+        if (TAILQ_EMPTY(&ssl_state->server_connp.certs)) {
+            return NULL;
+        }
+
+        const SSLCertsChain *cert = TAILQ_FIRST(&ssl_state->server_connp.certs);
```
victorjulien: are we inspecting only one cert?

src/detect-tls-cert.c
```diff
@@ -0,0 +1,142 @@
...
+void DetectTlsCertRegister(void)
+{
+    sigmatch_table[DETECT_AL_TLS_CERT].name = "tls.cert";
+    sigmatch_table[DETECT_AL_TLS_CERT].desc = "content modifier to match the TLS certificate buffer";
+    sigmatch_table[DETECT_AL_TLS_CERT].url = DOC_URL DOC_VERSION "/rules/tls-keywords.html#tls.cert";
+    sigmatch_table[DETECT_AL_TLS_CERT].Setup = DetectTlsCertSetup;
+#ifdef UNITTESTS
+    sigmatch_table[DETECT_AL_TLS_CERT].RegisterTests = DetectTlsCertRegisterTests;
+#endif
+    sigmatch_table[DETECT_AL_TLS_CERT].flags |= SIGMATCH_NOOPT;
```
victorjulien: add SIGMATCH_INFO_STICKY_BUFFER

src/tests/detect-tls-cert.c
```diff
@@ -0,0 +1,363 @@
...
+    SigMatch *sm = de_ctx->sig_list->sm_lists[DETECT_SM_LIST_MATCH];
+    FAIL_IF_NOT_NULL(sm);
+
+    sm = de_ctx->sig_list->sm_lists[g_tls_cert_buffer_id];
+    FAIL_IF_NULL(sm);
+
+    FAIL_IF(sm->type != DETECT_CONTENT);
+    FAIL_IF_NOT_NULL(sm->next);
+
+    SigCleanSignatures(de_ctx);
```
victorjulien: this is redundant: called by DetectEngineCtxFree

src/detect-tls-cert.c
victorjulien: Ok I see this is mentioned in the PR. Would it make sense to inspect all? Or would there be a case for a separate keyword for all certs or something?

src/detect-tls-cert.c
```diff
@@ -0,0 +1,142 @@
...
+static int g_tls_cert_buffer_id = 0;
+
+/**
+ * \brief Registration function for keyword: tls.cert
+ */
+void DetectTlsCertRegister(void)
+{
+    sigmatch_table[DETECT_AL_TLS_CERT].name = "tls.cert";
+    sigmatch_table[DETECT_AL_TLS_CERT].desc = "content modifier to match the TLS certificate buffer";
+    sigmatch_table[DETECT_AL_TLS_CERT].url = DOC_URL DOC_VERSION "/rules/tls-keywords.html#tls.cert";
```
thus: Hm, this should perhaps be changed to "tls-cert" instead of "tls.cert".

src/detect-tls-cert.c
thus: There's not really a good reason to only inspect one certificate. I'll change it to inspect the entire certificate chain :)

src/detect-tls-cert.c
thus: Will do!

src/tests/detect-tls-cert.c
thus: I'll remove it!

src/detect-tls-cert.c
```diff
@@ -0,0 +1,142 @@
...
+#include "util-unittest.h"
+#include "util-unittest-helper.h"
+
+static int DetectTlsCertSetup(DetectEngineCtx *, Signature *, const char *);
+#ifdef UNITTESTS
+static void DetectTlsCertRegisterTests(void);
+#endif
+static InspectionBuffer *GetData(DetectEngineThreadCtx *det_ctx,
+        const DetectEngineTransforms *transforms,
+        Flow *_f, const uint8_t _flow_flags,
```
thus: Should remove the underscores from the variables here as well.

src/detect-tls-cert.c
victorjulien: Ok cool. This will complicate the inspection code somewhat, because you can now have multiple buffers per tx. I would suggest checking the dns query code which has the same issue.

src/detect-tls-cert.c
victorjulien: Agreed.

src/detect-tls-cert.c
victorjulien: Yeah if they are used then you should

victorjulien:
Can you also add suricata-verify tests?

thus:
Replaced by #3855

PR #3852, opened 2019-05-14
http: adds events for each libhtp log

catenacyber:
https://redmine.openinfosecfoundation.org/issues/997

Describe changes:
- Adds http event for each warning or error log from libhtp

Here are my comments :
- Should there be a suricata-verify test for each case ?
- I am not sure that "C-T multipart/byteranges in responses not supported" should raise an event.
- Should we use the flag `HTP_STATUS_LINE_INVALID` ? The logs are more precise as they distinguish invalid protocol and invalid status...


rules/http-events.rules
```diff
@@ -57,5 +57,16 @@ alert http any any -> any any (msg:"SURICATA HTTP Request unrecognized authoriza
 alert http any any -> any any (msg:"SURICATA HTTP Request excessive header repetition"; flow:established,to_server; app-layer-event:http.request_header_repetition; flowint:http.anomaly.count,+,1; classtype:protocol-command-decode; sid:2221035; rev:1;)
 alert http any any -> any any (msg:"SURICATA HTTP Response excessive header repetition"; flow:established,to_client; app-layer-event:http.response_header_repetition; flowint:http.anomaly.count,+,1; classtype:protocol-command-decode; sid:2221036; rev:1;)
 
-# next sid 2221037
+alert http any any -> any any (msg:"SURICATA HTTP Response multipart/byteranges"; flow:established,to_client; app-layer-event:http.response_multipart_byteranges; flowint:http.anomaly.count,+,1; classtype:protocol-command-decode; sid:2221037; rev:1;)
```
victorjulien: this one should be commented out by default I think. It's more informational about suricata limitations than an actual issue.

PR #3855, opened 2019-05-15

thus:
Add new keyword `tls.cert` to match on the 'raw' TLS certificate. This keyword matches on the first certificate in the certificate chain.

Example:
```alert tls any any -> any any (msg:"test tls.cert"; tls.cert; content:"|01 02 03 04|"; sid:1;)```

Updates:
- Fixed changes requested in #3851 
- Changed the code so it inspects every certificate in the chain

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/185
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/185

### CODE REVIEW

**Comment by thus** [comment]
https://github.com/OISF/suricata-verify/pull/56

File: doc/userguide/rules/tls-keywords.rst [hunk 1]
```diff
@@ -136,6 +136,21 @@ Usage::
 
   tls_cert_valid;
 
+tls.cert
+--------
+
+Do a "raw" match on the TLS certificate. This keyword only matches on the
+first certificate in the certificate chain.
```
victorjulien: no it doesn't, it matches on all of them... right?

src/detect-tls-cert.c
```diff
@@ -0,0 +1,263 @@
...
+    const DetectEngineTransforms *transforms;
+} PrefilterMpmTlsCert;
+
+/**
+ * \brief Registration function for keyword: tls.cert
+ */
+void DetectTlsCertRegister(void)
+{
+    sigmatch_table[DETECT_AL_TLS_CERT].name = "tls.cert";
+    sigmatch_table[DETECT_AL_TLS_CERT].desc = "content modifier to match the TLS certificate buffer";
```
victorjulien: sticky buffer
Also, should this mention that we only inspect certs the server provides?

thus:
> I wonder if this should be tls.server_certs or something? Or do you not see a case for client certs being added later?

It would be cool to add client certificates too. I haven't really thought about that in a while, but it would probably not be a huge job to add (famous last words).

If I do that, would I then be allowed to keep the name "tls.cert"? ;)

doc/userguide/rules/tls-keywords.rst
thus: Yeah. I should probably fix that!

thus:
Replaced by #3857

PR #3857, opened 2019-05-15

thus:
Add new keyword `tls.certs` to match on each of the 'raw' certificates in the TLS certificate chain.

Example:
```alert tls any any -> any any (msg:"test tls.certs"; tls.certs; content:"|01 02 03 04|"; sid:1;)```

Updates:
- Fixed changes requested in #3855 
- Changed the code so it inspects every certificate in the chain
- Renamed keyword to 'tls.certs'

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/186
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/186

### CODE REVIEW

**Comment by thus** [comment]
suricata-verify testcases:
https://github.com/OISF/suricata-verify/pull/58

File: src/detect-tls-certs.c [hunk 1]
```diff
@@ -0,0 +1,263 @@
...
+        const int match = DetectEngineContentInspection(de_ctx, det_ctx, s, engine->smd,
+                                              NULL, f, (uint8_t *)buffer->inspect,
+                                              buffer->inspect_len,
+                                              buffer->inspect_offset, DETECT_CI_FLAGS_SINGLE,
+                                              DETECT_ENGINE_CONTENT_INSPECTION_MODE_STATE);
+        if (match == 1) {
+            return DETECT_ENGINE_INSPECT_SIG_MATCH;
+        }
+
+	local_id++;
```
victorjulien: on the 'local' local_id is incremented, not the one in the cbdata. So the same id is used for each iteration.

src/detect-tls-certs.c
```diff
@@ -0,0 +1,263 @@
...
+        if (buffer == NULL)
+            break;
+
+        if (buffer->inspect_len >= mpm_ctx->minlen) {
+            (void)mpm_table[mpm_ctx->mpm_type].Search(mpm_ctx,
+                    &det_ctx->mtcu, &det_ctx->pmq,
+                    buffer->inspect, buffer->inspect_len);
+        }
+
+        local_id++;
```
victorjulien: same here

src/detect-tls-certs.c
victorjulien: I don't think you need to local var, you can just use and increment the one in cbdata directly I think

thus: Oh, thanks for pointing that out! I'll fix it right away!

thus:
Replaced by #3859

PR #3862, opened 2019-05-17
Http uri double decode v1

catenacyber:
https://redmine.openinfosecfoundation.org/issues/524#change-12045

Describe changes:
- fixes double-encode configuration keywords
- adds an http event for double encoded uri

Behavior (http event generation) is only enabled when configuration keywords `double-decode-X` is true


src/app-layer-htp.c
```diff
@@ -2146,7 +2148,23 @@ static int HTPCallbackDoubleDecodeQuery(htp_tx_t *tx)
...
-    htp_urldecode_inplace(tx->cfg, HTP_DECODER_URLENCODED, tx->parsed_uri->query, &flags);
+    size_t prevlen = bstr_len(tx->parsed_uri->query);
+    htp_status_t res = htp_urldecode_inplace(tx->cfg, HTP_DECODER_URLENCODED, tx->parsed_uri->query, &flags);
+    // shorter string means that uri was encoded
+    if (res == HTP_OK && prevlen > bstr_len(tx->parsed_uri->query)) {
+        HtpTxUserData *htud = (HtpTxUserData *) htp_tx_get_user_data(tx);
+        if (likely(htud == NULL)) {
+            htud = HTPMalloc(sizeof(*htud));
+            if (unlikely(htud == NULL))
+                return HTP_ERROR;
```
victorjulien: we can probably be more forgiving here, instead of making the callback fail with an error simply don't set an event

src/app-layer-htp.c
```diff
@@ -2146,7 +2148,23 @@ static int HTPCallbackDoubleDecodeQuery(htp_tx_t *tx)
...
 
     uint64_t flags = 0;
-    htp_urldecode_inplace(tx->cfg, HTP_DECODER_URLENCODED, tx->parsed_uri->query, &flags);
+    size_t prevlen = bstr_len(tx->parsed_uri->query);
+    htp_status_t res = htp_urldecode_inplace(tx->cfg, HTP_DECODER_URLENCODED, tx->parsed_uri->query, &flags);
+    // shorter string means that uri was encoded
+    if (res == HTP_OK && prevlen > bstr_len(tx->parsed_uri->query)) {
+        HtpTxUserData *htud = (HtpTxUserData *) htp_tx_get_user_data(tx);
+        if (likely(htud == NULL)) {
+            htud = HTPMalloc(sizeof(*htud));
```
victorjulien: HTPCalloc avoids the memset call

src/app-layer-htp.c
```diff
@@ -2157,7 +2175,23 @@ static int HTPCallbackDoubleDecodePath(htp_tx_t *tx)
         return HTP_OK;
 
     uint64_t flags = 0;
-    htp_urldecode_inplace(tx->cfg, HTP_DECODER_URL_PATH, tx->parsed_uri->path, &flags);
+    size_t prevlen = bstr_len(tx->parsed_uri->path);
+    htp_status_t res = htp_urldecode_inplace(tx->cfg, HTP_DECODER_URL_PATH, tx->parsed_uri->path, &flags);
+    // shorter string means that uri was encoded
+    if (res == HTP_OK && prevlen > bstr_len(tx->parsed_uri->path)) {
+        HtpTxUserData *htud = (HtpTxUserData *) htp_tx_get_user_data(tx);
```
victorjulien: there is a lot (100%) of duplication here, can we avoid that?

src/app-layer-htp.c
catenacyber: ok, I hesitated about this one

src/app-layer-htp.c
catenacyber: Sure

src/app-layer-htp.c
catenacyber: Good idea

PR #3860, opened 2019-05-17
Request for Comments Only: App layer/Proto detection anomaly event logging

jlucovsky:
_Request for comments only_ (work in progress)

This PR includes changes for 2 issues:
- [2941](https://redmine.openinfosecfoundation.org/issues/2941)
- [2942](https://redmine.openinfosecfoundation.org/issues/2942)

Posted for early feedback ....

App layer events are logged via the transactional logging support with a "wild card" application layer protocol. This leverages the existing transactional handling in the logger and ensures that each transaction is only inspected a single time. This logging requires the capability of translating the recorded event (code) into a string. Additions to the app layer parser infrastructure provide that capability and many (but not all) of the app layer protocols have been modified to translate event ids into event strings.

Protocol and parser detection events are handled by reviewing the `app_layer_parser` field of the packet and the parser state, respectively.

Would like some feedback on the approach as well as the content/key naming of generated log records (below).

### Protocol Detection Event
```
{
  "timestamp": "2015-05-29T07:35:09.657185-0700",
  "flow_id": 1506698034184689,
  "pcap_cnt": 82775,
  "event_type": "anomaly",
  "src_ip": "10.3.162.105",
  "src_port": 62869,
  "dest_ip": "61.65.90.109",
  "dest_port": 80,
  "proto": "TCP",
  "anomaly": {
    "alproto": "http",
    "layer": "app_layer",
    "type": "packet",
    "event": "APPLAYER_DETECT_PROTOCOL_ONLY_ONE_DIRECTION"
  }
}
```

### App layer events
```
{
  "timestamp": "2015-05-29T07:35:09.657185-0700",
  "flow_id": 1506698034184689,
  "pcap_cnt": 82775,
  "event_type": "anomaly",
  "src_ip": "10.3.162.105",
  "src_port": 62869,
  "dest_ip": "61.65.90.109",
  "dest_port": 80,
  "proto": "TCP",
  "anomaly": {
    "alproto": "http",
    "layer": "applayer_parser",
    "type": "transaction",
    "event": "METHOD_DELIM_NON_COMPLIANT"
  }
}
{
  "timestamp": "2015-05-29T07:35:09.657185-0700",
  "flow_id": 1506698034184689,
  "pcap_cnt": 82775,
  "event_type": "anomaly",
  "src_ip": "10.3.162.105",
  "src_port": 62869,
  "dest_ip": "61.65.90.109",
  "dest_port": 80,
  "proto": "TCP",
  "anomaly": {
    "alproto": "http",
[...]

PR #3867, opened 2019-05-20
Request for Comments #2: App layer/proto detection anomaly event logging

jlucovsky:
_Request for comments only #2_ (work in progress); continuation of #3860

This PR includes changes for 2 issues:
- [2941](https://redmine.openinfosecfoundation.org/issues/2941)
- [2942](https://redmine.openinfosecfoundation.org/issues/2942)

Posted for early feedback ....

App layer events are logged via the transactional logging support with a "wild card" application layer protocol. This leverages the existing transactional handling in the logger and ensures that each transaction is only inspected a single time. This logging requires the capability of translating the recorded event (code) into a string. Additions to the app layer parser infrastructure provide that capability and many (but not all) of the app layer protocols have been modified to translate event ids into event strings.

Protocol and parser detection events are handled by reviewing the `app_layer_parser` field of the packet and the parser state, respectively.

### Update
This PR includes changes that apply the "wild card" logger to all protocols; this is used during the logger output loop when traversing transactions and loggers.

The second change adds logic such that a wild card logger only receives the transaction if none of the other loggers have deferred handling the transaction. This ensures that the anomaly logger receives the transaction when it's deemed complete.

### TODO
- Ensure each app layer protocol's decoder events or equivalent can be translated from and event id into a string.
- Documentation updates

#### Observations
Each transaction has an event set; for a given transaction, there are sometimes multiple, back-to-back events recorded that are the same. 

Would like some feedback on the approach as well as the content/key naming of generated log records (below).

### Protocol Detection Event
```
{
  "timestamp": "2011-01-25T10:56:06.114451-0800",
  "flow_id": 796744833386293,
  "pcap_cnt": 10968,
  "event_type": "anomaly",
  "src_ip": "10.0.2.15",
  "src_port": 2526,
  "dest_ip": "65.54.186.19",
  "dest_port": 5443,
  "proto": "TCP",
  "anomaly": {
    "alproto": "tls",
    "type": "packet",
    "event": "APPLAYER_DETECT_PROTOCOL_ONLY_ONE_DIRECTION",
    "event_no": "1 (of 1)",
    "layer": "app_layer"
  }
}
```

### App layer events
```
{
  "timestamp": "2011-01-25T10:54:08.507043-0800",
  "flow_id": 1901021719668973,
  "pcap_cnt": 4232,
  "event_type": "anomaly",
  "src_ip": "194.165.188.79",
  "src_port": 443,
  "dest_ip": "172.16.255.1",
  "dest_port": 10627,
  "proto": "TCP",
  "anomaly": {
    "alproto": "tls",
    "type": "transaction",
    "event": "INVALID_SSL_RECORD",
    "event_no": "1 (of 4)",
    "tx_id": 0,
    "layer": "applayer_parser"
  }
}
{
  "timestamp": "2011-01-25T10:54:08.507043-0800",
  "flow_id": 1901021719668973,
  "pcap_cnt": 4232,
  "event_type": "anomaly",
  "src_ip": "194.165.188.79",
  "src_port": 443,
  "dest_ip": "172.16.255.1",
  "dest_port": 10627,
  "proto": "TCP",
  "anomaly": {
    "alproto": "tls",
    "type": "transaction",
    "event": "INVALID_HANDSHAKE_MESSAGE",
    "event_no": "2 (of 4)",
    "tx_id": 0,
    "layer": "applayer_parser"
  }
}
```

### DIFF
--- a/src/output-json-anomaly.c
+++ b/src/output-json-anomaly.c
@@ -78,7 +78,8 @@ typedef struct JsonAnomalyLogThread_ {
     AnomalyJsonOutputCtx* json_output_ctx;
 } JsonAnomalyLogThread;
 
-static int AnomalyDecodeEventJson(ThreadVars *tv, JsonAnomalyLogThread *aft, const Packet *p)
+static int AnomalyDecodeEventJson(ThreadVars *tv, JsonAnomalyLogThread *aft,
+                                  const Packet *p)
 {
     bool is_ip_pkt = PKT_IS_IPV4(p) || PKT_IS_IPV6(p);
 
@@ -138,23 +139,12 @@ static int AnomalyDecodeEventJson(ThreadVars *tv, JsonAnomalyLogThread *aft, con
     return TM_ECODE_OK;
 }
 
-#if 0
-static int JsonAnomalyTxLogger(ThreadVars *tv, void *thread_data, const Packet *p, Flow *f, void *state, void *tx, uint64_t tx_id)
-{
-    SCLogInfo("Here we are");
-    return TM_ECODE_OK;
-}
-#endif
-
-static int AnomalyAppLayerDecoderEventJson(JsonAnomalyLogThread *aft, const Packet *p, AppLayerDecoderEvents *decoder_events, bool is_applayer)
+static int AnomalyAppLayerDecoderEventJson(JsonAnomalyLogThread *aft,
+                        const Packet *p, AppLayerDecoderEvents *decoder_events,
+                        bool is_applayer, const char *layer, uint64_t tx_id)
 {
     const char *alprotoname;
-
-    if (likely(p->flow)) {
-        alprotoname = AppLayerGetProtoName(p->flow->alproto);
-    } else {
-        alprotoname = NULL;
-    }
+    alprotoname = AppLayerGetProtoName(p->flow->alproto);
 
     for (int i = 0; i < decoder_events->cnt; i++) {
         MemBufferReset(aft->json_buffer);
@@ -171,8 +161,7 @@ static int AnomalyAppLayerDecoderEventJson(JsonAnomalyLogThread *aft, const Pack
         }
 
         JsonFiveTuple((const Packet *)p, LOG_DIR_PACKET, js);
-        if (p->flow)
-            JsonAddCommonOptions(&aft->json_output_ctx->cfg, p, p->flow, js);
+        JsonAddCommonOptions(&aft->json_output_ctx->cfg, p, p->flow, js);
 
         /* Use app layer proto name if available */
         if (alprotoname) {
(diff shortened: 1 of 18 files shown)


src/output-json-anomaly.c
```diff
@@ -134,6 +139,129 @@ static int AnomalyJson(ThreadVars *tv, JsonAnomalyLogThread *aft, const Packet *
...
+            return TM_ECODE_OK;
+        }
+
+        json_t *ajs = json_object();
+        if (unlikely(ajs == NULL)) {
+            json_decref(js);
+            return TM_ECODE_OK;
+        }
+
+        JsonFiveTuple((const Packet *)p, LOG_DIR_PACKET, js);
```
victorjulien: this is already called by CreateJSONHeader, is it needed here?

src/output-json-anomaly.c
```diff
@@ -134,6 +139,129 @@ static int AnomalyJson(ThreadVars *tv, JsonAnomalyLogThread *aft, const Packet *
...
+            json_object_set_new(ajs, "event", json_string(event_name));
+        } else {
+            json_object_set_new(ajs, "type", json_string("unknown"));
+            json_object_set_new(ajs, "code", json_integer(event_code));
+        }
+
+        /* log event number (and total */
+        uint32_t offset = 0;
+        char event_no_buf[32];
+        PrintBufferData(event_no_buf, &offset, 32, "%d (of %d)",
```
victorjulien: is there a reason for not using snprintf here?

src/output-json-anomaly.c
```diff
@@ -134,6 +139,129 @@ static int AnomalyJson(ThreadVars *tv, JsonAnomalyLogThread *aft, const Packet *
...
+
+    int rc = TM_ECODE_OK;
+
+    if (p->events.cnt) {
+        rc = AnomalyDecodeEventJson(tv, aft, p);
+    }
+
+    /* app layer events */
+    if (p->app_layer_events && p->app_layer_events->cnt) {
+        rc = AnomalyAppLayerDecoderEventJson(aft, p, p->app_layer_events,
```
victorjulien: rc is overwritten. AnomalyDecodeEventJson could have set it.

jlucovsky:
> Maybe I'm overlooking it, but where is made sure that we don't log the same event multiple times? I think this could happen both in p->flow->alparser and tx events.

The transaction not logged until the transaction has been deemed complete so this shouldn't happen with tx events.

I'll check the `alparser` events.

victorjulien:
So this could lead to long delays if the tx is long running and the event is generated in the beginning?

src/output-json-anomaly.c
jlucovsky: Oops ... good catch.

src/output-json-anomaly.c
jlucovsky: No; will use that here and in first anomaly logger.

src/output-json-anomaly.c
jlucovsky: Good catch.

jlucovsky:
> Maybe I'm overlooking it, but where is made sure that we don't log the same event multiple times? I think this could happen both in p->flow->alparser and tx events.

Will add a mechanism per our discussion to avoid double logging of individual events.

jlucovsky:
> So this could lead to long delays if the tx is long running and the event is generated in the beginning?

Will modify to no longer defer tx logging; instead, a mechanism to avoid double logging of individual events will be added.

jlucovsky:
> Maybe I'm overlooking it, but where is made sure that we don't log the same event multiple times? I think this could happen both in p->flow->alparser and tx events.

Will add a mechanism to ensure individual events are logged a single time.

jlucovsky:
Continued in #3880.

PR #3872, opened 2019-05-20
signature: fix memory leak in DetectBytejumpSetup

catenacyber:
None

Describe changes:
- Fixes memory leak in `DetectBytejumpSetup`

Example signature 
`alert udp any 67 -> any 68 (msg:"ET INFO Web|"; byte_jump:1,count0,relative,post_offset -9; conten16_06_24, updated_at 2016_06_24;)`

### DIFF
--- a/src/detect-bytejump.c
+++ b/src/detect-bytejump.c
@@ -581,6 +581,7 @@ static int DetectBytejumpSetup(DetectEngineCtx *de_ctx, Signature *s, const char
         if (bed_sm == NULL) {
             SCLogError(SC_ERR_INVALID_SIGNATURE, "Unknown byte_extract var "
                        "seen in byte_jump - %s\n", offset);
+            SCFree(offset);
             goto error;
         }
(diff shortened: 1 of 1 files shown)


catenacyber:
Found by fuzzing `SigInit`

src/detect-bytejump.c
```diff
@@ -581,6 +581,7 @@ static int DetectBytejumpSetup(DetectEngineCtx *de_ctx, Signature *s, const char
         if (bed_sm == NULL) {
             SCLogError(SC_ERR_INVALID_SIGNATURE, "Unknown byte_extract var "
                        "seen in byte_jump - %s\n", offset);
+            SCFree(offset);
```
jlucovsky: Should this be freed in preceding `goto error` cases?

catenacyber: Thank you @jlucovsky 
This is the only `goto` within the condition where `if (offset != NULL) {` at line 579
And it gets freed at line 589 if there is no error 
So I think we only need this one.
Does it look good to you so ?

PR #3873, opened 2019-05-20
Http evader v2

catenacyber:
See http evader case 481
https://redmine.openinfosecfoundation.org/issues/1656

Describe changes:
- Waits for request line (if not HTTP0.9) before looking for filename and storing file contents
- This allows to handle junk before response line, like in http evader case 481

Modifies #3764 :
- modifies HTTP unit tests to fix stream flags (removing STREAM_EOF when this is not the last packet)

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -1627,8 +1627,12 @@ static int HtpResponseBodyHandle(HtpState *hstate, HtpTxUserData *htud,
 
     int result = 0;
 
-    /* see if we need to open the file */
-    if (!(htud->tcflags & HTP_FILENAME_SET))
+    /* see if we need to open the file
+     * we check for tx->response_line in case of junk
+     * interpreted as body before response line
+     */
+    if (!(htud->tcflags & HTP_FILENAME_SET) &&
+        (tx->response_line != NULL || tx->is_protocol_0_9))
     {
         SCLogDebug("setting up file name");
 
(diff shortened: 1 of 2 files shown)


src/app-layer-htp.c
```diff
@@ -1627,8 +1627,12 @@ static int HtpResponseBodyHandle(HtpState *hstate, HtpTxUserData *htud,
 
     int result = 0;
 
-    /* see if we need to open the file */
-    if (!(htud->tcflags & HTP_FILENAME_SET))
+    /* see if we need to open the file
```
victorjulien: I don't understand this fix. This function is being called by HTPCallbackResponseBodyData, so the response body callback. The patch seems to suggest we can get called while there is actually no response body data. Why are we called then? And shouldn't the fix be that the callback is not called?

catenacyber: `HTPCallbackResponseBodyData` is called with data

The patch is about handling junk data before the actual response line (and dealing with HTTP 0.9 as well).

Should this junk data be thrown out ?

catenacyber:
CI is expected to fail without https://github.com/OISF/libhtp/pull/223

src/app-layer-htp.c
victorjulien: I think this junk data should then be stripped by libhtp before invoking the callback.

catenacyber: I would like to be sure that I understand "be stripped by" : 
Do you mean that Suricata should not send these bytes to libhtp callback ?

I did not choose this option because of another case : if we have more than one request and reply in the same connection.
Then, starting from the second response, it is impossible to tell if this is junk data before the response line, or extra data (more than content-length) that will be interpreted by browsers (there are http-evader cases like this).
So, is it right to do differently for the first reply than for the next ones ?

victorjulien:
Merged in #4160, thanks Philippe!

PR #3878, opened 2019-05-21
travis: move checks into script

victorjulien:
Move travis test logic into support script.

Previously some errors were not caught, e.g. https://travis-ci.org/victorjulien/suricata/jobs/535178322#L2376

$? and 'exit' are not supposed to be used in script section of the yaml file directly, but can be used in the script.


qa/travis.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+set -ev
```
jlucovsky: Suggest including `set -u` (treat unset variables as an error).

qa/travis.sh
```diff
@@ -0,0 +1,65 @@
...
+        --with-libnspr-includes=/usr/local/opt/nspr/include/nspr \
+        --with-libnspr-libraries=/usr/local/opt/nspr/lib ${ARGS}
+fi
+
+if [[ "${EXIT_CODE}" ]]; then
+   exit "${EXIT_CODE}"
+fi
+
+# Linux container builds have 2 cores, make use of them.
+if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
```
jlucovsky: Suggest using output of `nproc` instead? If there are more cores, they'll automatically be used.

qa/travis.sh
```diff
@@ -0,0 +1,65 @@
...
+    j="-j 2"
+fi
+make ${j}
+
+# Like "make check", but fail on first error. We redirect the output
+# so Travis doesn't fail the build with a too much output error.
+if [[ "${NO_UNITTESTS}" != "yes" ]]; then
+    mkdir -p ./qa/log
+    ./src/suricata -u -l ./qa/log --fatal-unittests > unittests.log 2>&1
```
jlucovsky: Nit: I have a preference for `comand &> output-file` instead of `command > output-file 2>&1`

qa/travis.sh
```diff
@@ -0,0 +1,65 @@
...
+    if [[ $? -ne 0 ]]; then
```
jlucovsky: How can the return code be non-zero since `set -e` was used?

qa/travis.sh
victorjulien: Leaving this as future todo as it causes build failures. I suspect this is because the build matrix in the yaml defines some things some of the times, but not for all builds.

qa/travis.sh
victorjulien: Leaving as a todo as well. Goal here is to fix the error checking. Leaving the logic otherwise unchanged.

qa/travis.sh
victorjulien: Good catch: trying with disabling -e just before and enabling -e just after so we can still have the diagnostic output.

victorjulien:
Merged an updated version.

PR #3888, opened 2019-05-26
logging: app-layer and proto detect layer events

jlucovsky:
This PR includes changes for 2 issues:
- [2941](https://redmine.openinfosecfoundation.org/issues/2941)
- [2942](https://redmine.openinfosecfoundation.org/issues/2942)
- [2991] (https://redmine.openinfosecfoundation.org/issues/2991)

This PR includes changes for *2991* to optimize TX logging of app-layer events. All parsers, including Rust-based parsers, have been updated.

App-layer events and protocol detection events are now logged to the anomaly log these changes.

App-layer events are logged by adding a "wild card" application layer logger. This logger is invoked while traversing the TX loggers and is invoked for all TX, regardless of the `alproto`; hence, the term "wild-card". This logger may be invoked multiple times for the same TX; a mechanism to track which events have been logged previously is employed to avoid logging an individual event twice.

Protocol detection events are logged when `app_layer_events` has an event posted. Parser events are logged when posted to the (flow) `alparser` area.

App layer parsers have been amended to include a function to translate an application layer parser specific event id into a display string.

### TODO
- Extend Rust app-layer-parsers to support the new `GetEventInfoById` function that translates an app layer parser event value into a display string. This includes SMB and NFS.
- Documentation updates.

#### Observations
Each transaction has an event set; for a given transaction, there are sometimes multiple, back-to-back events recorded that are the same. 

Would like some feedback on the approach as well as the content/key naming of generated log records (below).

### Protocol Detection Event
```
{
  "timestamp": "2014-09-09T18:29:28.179967-0700",
  "flow_id": 1831982488608705,
  "pcap_cnt": 1077,
  "event_type": "anomaly",
  "src_ip": "172.16.165.134",
  "src_port": 25,
  "dest_ip": "172.16.165.133",
  "dest_port": 49200,
  "proto": "TCP",
  "anomaly": {
    "alproto": "smtp",
    "type": "packet",
    "event": "APPLAYER_DETECT_PROTOCOL_ONLY_ONE_DIRECTION",
    "event_no": "1 (of 1)",
    "layer": "app_layer"
  }
}
```

### App layer events
```
{
  "timestamp": "2014-10-06T16:10:03.482379-0700",
  "flow_id": 1831982488608705,
  "event_type": "anomaly",
  "src_ip": "172.16.165.134",
  "src_port": 25,
  "dest_ip": "172.16.165.133",
  "dest_port": 49200,
  "proto": "TCP",
  "tx_id": 0,
  "anomaly": {
    "alproto": "smtp",
    "type": "transaction",
    "event": "INVALID_REPLY",
    "event_no": "1 (of 1)",
    "layer": "applayer_parser"
  }
}
{
  "timestamp": "2015-01-06T14:27:26.548474-0800",
  "flow_id": 833463658664645,
  "pcap_cnt": 636813,
  "event_type": "anomaly",
  "src_ip": "108.61.179.182",
  "src_port": 443,
  "dest_ip": "192.168.138.163",
  "dest_port": 49168,
  "proto": "TCP",
  "tx_id": 0,
  "anomaly": {
    "alproto": "http",
    "type": "transaction",
    "event": "REQUEST_AUTH_UNRECOGNIZED",
    "event_no": "1 (of 1)",
    "layer": "applayer_parser"
  }
}
{
  "timestamp": "2015-01-06T14:27:27.959384-0800",
  "flow_id": 833463658664645,
  "pcap_cnt": 636816,
  "event_type": "anomaly",
  "src_ip": "192.168.138.163",
  "src_port": 49168,
  "dest_ip": "108.61.179.182",
  "dest_port": 443,
  "proto": "TCP",
  "tx_id": 0,
  "anomaly": {
    "alproto": "http",
    "type": "transaction",
    "event": "GZIP_DECOMPRESSION_FAILED",
    "event_no": "2 (of 5)",
    "layer": "applayer_parser"
  }
}
{
  "timestamp": "2011-01-25T10:54:08.507043-0800",
  "flow_id": 1685566832748781,
  "pcap_cnt": 4232,
  "event_type": "anomaly",
  "src_ip": "194.165.188.79",
  "src_port": 443,
  "dest_ip": "172.16.255.1",
  "dest_port": 10627,
  "proto": "TCP",
  "tx_id": 0,
  "anomaly": {
    "alproto": "tls",
    "type": "transaction",
    "event": "INVALID_SSL_RECORD",
    "event_no": "4 (of 4)",
    "layer": "applayer_parser"
  }
}
```


src/app-layer-dnp3.c
```diff
@@ -1309,20 +1309,10 @@ static int DNP3ParseResponse(Flow *f, void *state, AppLayerParserState *pstate,
...
-
-    if (dnp3->curr && dnp3->curr->tx_num == tx_num) {
-        return dnp3->curr->decoder_events;
-    }
-
-    TAILQ_FOREACH(tx, &dnp3->tx_list, next) {
-        if (tx->tx_num == tx_num) {
-            return tx->decoder_events;
-        }
+    if (likely(tx)) {
```
victorjulien: in the rust code the tx that is passed is unconditionally dereferenced, so this assumes we never get a null ptr. I think this makes sense. So we can remove checks like this as well.

src/app-layer-enip.c
```diff
@@ -112,17 +112,10 @@ static uint64_t ENIPGetTxCnt(void *alstate)
...
-    ENIPState         *enip = (ENIPState *) state;
-    ENIPTransaction   *tx;
-
-    if (enip->curr && enip->curr->tx_num == (id + 1))
-        return enip->curr->decoder_events;
-
-    TAILQ_FOREACH(tx, &enip->tx_list, next) {
-        if (tx->tx_num == (id+1))
-            return tx->decoder_events;
+    if (likely(tx)) {
```
victorjulien: same

src/app-layer-parser.h
```diff
@@ -138,7 +138,7 @@ void AppLayerParserRegisterLocalStorageFunc(uint8_t ipproto, AppProto proto,
 void AppLayerParserRegisterGetFilesFunc(uint8_t ipproto, AppProto alproto,
                              FileContainer *(*StateGetFiles)(void *, uint8_t));
 void AppLayerParserRegisterGetEventsFunc(uint8_t ipproto, AppProto proto,
-    AppLayerDecoderEvents *(*StateGetEvents)(void *, uint64_t));
+    AppLayerDecoderEvents *(*StateGetEvents)(void *));
```
victorjulien: append ```__attribute__((nonnull))``` here to tell the compiler (and devs) we shouldn't pass a null ptr here?

src/output-json-anomaly.c
```diff
@@ -134,6 +135,137 @@ static int AnomalyJson(ThreadVars *tv, JsonAnomalyLogThread *aft, const Packet *
...
+        json_t *ajs = json_object();
+        if (unlikely(ajs == NULL)) {
+            json_decref(js);
+            return TM_ECODE_OK;
+        }
+
+        JsonAddCommonOptions(&aft->json_output_ctx->cfg, p, p->flow, js);
+
+        /* Use app layer proto name if available */
+        json_object_set_new(ajs, "alproto",
```
victorjulien: this is ```app_proto``` in eve. See JsonAddFlow for example.

Also, you can skip this numeric fallback. The compiler will yell at us if we forgot to keep AppLayerGetProtoName current. Worst case a null should be handled correctly by ```json_string()```

src/app-layer-enip.c
victorjulien: (there are a couple more places where this happens)

jasonish:
```
    "event_no": "2 (of 5)",
```
This looks a little suspicious.  Would this value ever be parsed, and is it useful in event visualization?  If its purely an informative string, then probably fine.  But if its useful to be machine parsed, it might be better to have "event_no" and some other field designated to the total events.

I also see:
```
    "layer": "app_layer"
```
and
```
    "layer": "applayer_parser"
```
Perhaps the first one should just be `applayer`.

src/output-json-anomaly.c
jlucovsky: > ```
>     "event_no": "2 (of 5)",
> ```
> 
> This looks a little suspicious. Would this value ever be parsed, and is it useful in event visualization? If its purely an informative string, then probably fine. But if its useful to be machine parsed, it might be better to have "event_no" and some other field designated to the total events.
> 

The intent was to be an informational string. Do you think that having this information would be helpful for the consumers who are machine parsing these records? If so, then perhaps an alternate presentation would be better.

> I also see:
> 
> ```
>     "layer": "app_layer"
> ```
> 
> and
> 
> ```
>     "layer": "applayer_parser"
> ```
> 
> Perhaps the first one should just be `applayer`.

Will update.

jlucovsky: > this is `app_proto` in eve. See JsonAddFlow for example.
> 
> Also, you can skip this numeric fallback. The compiler will yell at us if we forgot to keep AppLayerGetProtoName current. Worst case a null should be handled correctly by `json_string()`

Will use `app_proto` and remove the numeric fallback.

src/app-layer-parser.h
jlucovsky: Will do.

src/app-layer-dnp3.c
jlucovsky: Will do.

jlucovsky:
Continued in #3893

PR #3877, opened 2019-05-21
signature: fix memory leak in DetectBytejumpSetup

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None

Describe changes:
- Fixes memory leak in `DetectBytejumpSetup`

Example signature 
`alert udp any 67 -> any 68 (msg:"ET INFO Web|"; byte_jump:1,count0,relative,post_offset -9; conten16_06_24, updated_at 2016_06_24;)`

Modifies #3872 
- Avoids the leak in more failure cases

PR #3854, opened 2019-05-15
add fuzz-targets and environment to suricata (LibFuzzer and llvm sanitizer)

SirkoHoeer:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
(There is no link to a redmine issue)

Describe changes:

As discussed with Victor I send the Pull Request to you.
It includes:
 - 26 Fuzz Targets
 - Build scripts 
 - Docker Run-time Environment 
 - Seed File for Fuzzing 

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #3891, opened 2019-05-27
Fuzz v4

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds three fuzz targets

Modifies #3837 (with some help from #3854) :
- adds fuzz target for `ConfYamlLoadString` (config file parsing)
- adds fuzz target for `DecodePcapFile` (pcap file)
- adds fuzz target for `FlowWorker` (both rules and pcap at the same time)
- fixes use of `SC_LOG_FILE` environment variable from `SCLogSetOPIface` calling `SCLogInitFileOPIface

I think that it will be useful to merge this code as continuous fuzzing will find more bugs than I did running these targets 10 minutes or less and finding already #3877 and #3890 

What is left to be done before merging this ?
Are the following items relevant ? Is there anything else ?
- using or removing `fuzz_openFile`
- move `PostConfLoadedSetup` to init.c
- load and reload yaml config file in fuzz_sigyamlpcap.c

Should we already start oss-fuzz integration on a specific branch ? cf https://github.com/catenacyber/oss-fuzz/blob/ac9d0d47951f7c884b33ca26920717a27b84abee/projects/suricata/Dockerfile#L31

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@
     AC_CONFIG_HEADERS([config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
     AC_CONFIG_MACRO_DIR(m4)
-    AM_INIT_AUTOMAKE
+    AM_INIT_AUTOMAKE([subdir-objects])
 
     AC_LANG_C
     AC_PROG_CC_C99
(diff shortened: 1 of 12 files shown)


catenacyber:
@victorjulien do you agree with the other todos I listed ?

PR #3913, opened 2019-06-03
install: install all files for events rules

catenacyber:
https://redmine.openinfosecfoundation.org/issues/2786

Describe changes:
- Makefile target `install-rules` now install all event rules files ie `rules/*-events.rules` not not a specific list

This patch only for the case where suricata-update is not installed

Modifies #3911 with the rebase as needed

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -47,15 +47,7 @@ endif
 else
 	@echo "UNABLE to load ruleset wget or curl are not installed on system."
 endif
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)decoder-events.rules" || install -m 600 "$(top_srcdir)/rules/decoder-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)stream-events.rules" || install -m 600 "$(top_srcdir)/rules/stream-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)smtp-events.rules" || install -m 600 "$(top_srcdir)/rules/smtp-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)http-events.rules" || install -m 600 "$(top_srcdir)/rules/http-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)dns-events.rules" || install -m 600 "$(top_srcdir)/rules/dns-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -47,15 +47,7 @@ endif
...
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)smtp-events.rules" || install -m 600 "$(top_srcdir)/rules/smtp-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)http-events.rules" || install -m 600 "$(top_srcdir)/rules/http-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)dns-events.rules" || install -m 600 "$(top_srcdir)/rules/dns-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)tls-events.rules" || install -m 600 "$(top_srcdir)/rules/tls-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)modbus-events.rules" || install -m 600 "$(top_srcdir)/rules/modbus-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)app-layer-events.rules" || install -m 600 "$(top_srcdir)/rules/app-layer-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
-	@test -e "$(DESTDIR)$(e_sysconfrulesdir)dnp3-events.rules" || install -m 600 "$(top_srcdir)/rules/dnp3-events.rules" "$(DESTDIR)$(e_sysconfrulesdir)"
+	install -m 600 $(top_srcdir)/rules/*-events.rules "$(DESTDIR)$(e_sysconfrulesdir)"
```
victorjulien: this will overwrite the existing files?

catenacyber: Yes. Is it not the expected behavior ?

victorjulien: It's the old behavior, so if you change that I expect it to be mentioned/explained in the git message.

Whether it should change is something @jasonish maybe has ideas about.

catenacyber: ping @jasonish ?

jasonish: I don't think we should change this for now.

We already install these rule files into /usr/share/suricata/rules in an overwrite fashion.  I'd like to eliminate the copying to /etc/suricata/rules in time.  Maybe 5.0 is that time?

catenacyber: Thank you Jason.
But I am not sure to understand.
Are you saying that we should ignore https://redmine.openinfosecfoundation.org/issues/2786 ?
And instead always use the rules installed in `/usr/share/suricata/rules` and do not install them any longer in `/etc/suricata/rules` ?

jasonish: No.  We should not ignore it. Fix it to install the missing files, but don't change to  overwrite the files.  Just keep the same process.

We do have to decide when to drop these files I think.  5.0, sometime after?

People often edit rule files in /etc/, so we don't want to overwrite them.  But we also want to include these files in a read-only matter.  That is why we started putting them into /usr/share, where they can be picked up by suricata-update.

victorjulien:
Please submit a new PR, thanks!

PR #3914, opened 2019-06-03
Fuzz v6

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds three fuzz targets

Modifies #3908 with needed rebase

I think that it will be useful to merge this code as continuous fuzzing will find more bugs than I did running these targets 10 minutes or less and finding already #3877 and #3890 

What is left to be done before merging this ?
Are the following items relevant ? Is there anything else ?
- using or removing `fuzz_openFile`
- move `PostConfLoadedSetup` to init.c
- load and reload yaml config file in fuzz_sigyamlpcap.c

Should we already start oss-fuzz integration on a specific branch ? cf https://github.com/catenacyber/oss-fuzz/blob/ac9d0d47951f7c884b33ca26920717a27b84abee/projects/suricata/Dockerfile#L31

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -6,8 +6,11 @@ noinst_HEADERS = action-globals.h \
 	suricata-common.h threadvars.h tree.h \
     util-validate.h
 bin_PROGRAMS = suricata
+if BUILD_FUZZTARGETS
+    bin_PROGRAMS += fuzz_applayerprotodetectgetproto fuzz_applayerparserparse fuzz_siginit fuzz_confyamlloadstring fuzz_decodepcapfile fuzz_sigyamlpcap
+endif
 
-suricata_SOURCES = \
+COMMON_SOURCES = \
 alert-debuglog.c alert-debuglog.h \
 alert-fastlog.c alert-fastlog.h \
 alert-prelude.c alert-prelude.h \
@@ -282,6 +285,7 @@ host-bit.c host-bit.h \
 host-queue.c host-queue.h \
 host-storage.c host-storage.h \
 host-timeout.c host-timeout.h \
+init.c \
 ippair.c ippair.h \
 ippair-bit.c ippair-bit.h \
 ippair-queue.c ippair-queue.h \
@@ -374,7 +378,6 @@ stream-tcp-list.c stream-tcp-list.h \
 stream-tcp-reassemble.c stream-tcp-reassemble.h \
 stream-tcp-sack.c stream-tcp-sack.h \
 stream-tcp-util.c stream-tcp-util.h \
-suricata.c suricata.h \
 threads.c threads.h \
 threads-debug.h threads-profile.h \
 tm-modules.c tm-modules.h \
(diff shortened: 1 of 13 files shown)


src/Makefile.am
```diff
@@ -6,8 +6,11 @@ noinst_HEADERS = action-globals.h \
 	suricata-common.h threadvars.h tree.h \
     util-validate.h
 bin_PROGRAMS = suricata
+if BUILD_FUZZTARGETS
+    bin_PROGRAMS += fuzz_applayerprotodetectgetproto fuzz_applayerparserparse fuzz_siginit fuzz_confyamlloadstring fuzz_decodepcapfile fuzz_sigyamlpcap
```
victorjulien: can you update this to have a ~80 char line limit?

src/Makefile.am
```diff
@@ -507,13 +510,88 @@ EXTRA_DIST = tests
...
 AM_CPPFLAGS = $(all_includes)
 
+suricata_SOURCES = suricata.c suricata.h $(COMMON_SOURCES)
+
 # the library search path.
 suricata_LDFLAGS = $(all_libraries) ${SECLDFLAGS}
-suricata_LDADD = $(HTP_LDADD) $(RUST_LDADD)
+suricata_LDADD = $(RUST_SURICATA_LIB) $(HTP_LDADD) $(RUST_LDADD)
+
+fuzz_applayerprotodetectgetproto_SOURCES = tests/fuzz/fuzz_applayerprotodetectgetproto.c $(COMMON_SOURCES)
```
victorjulien: can we unify these more? Its kind of ugly to have this list of similar blocks

src/Makefile.am
```diff
@@ -507,13 +510,88 @@ EXTRA_DIST = tests
...
+fuzz_applayerprotodetectgetproto_LDFLAGS = $(all_libraries) ${SECLDFLAGS}
+fuzz_applayerprotodetectgetproto_LDADD = $(RUST_SURICATA_LIB) $(HTP_LDADD) $(RUST_LDADD)
+if HAVE_LIBFUZZINGENGINE
```
victorjulien: when would we not have this?

src/tests/fuzz/fuzz_applayerparserparse.c
```diff
@@ -0,0 +1,101 @@
...
+    Flow * f;
+    TcpSession ssn;
+
+    if (size < HEADER_LEN) {
+        return 0;
+    }
+
+    if (alp_tctx == NULL) {
+        InitGlobal();
+        run_mode = RUNMODE_UNITTEST;
```
victorjulien: why is this set?

src/tests/fuzz/fuzz_sigyamlpcap.c
```diff
@@ -0,0 +1,208 @@
...
+        //disables checksums validation for fuzzing
+        suricata.checksum_validation = 0;
+        ConfSet("stream.checksum-validation", "0");
+        suricata.sig_file = "/tmp/fuzz.rules";
+        suricata.sig_file_exclusive = 1;
+        //loads rules after init
+        suricata.delayed_detect = 1;
+
+        SupportFastPatternForSigMatchTypes();
+        //PostConfLoadedSetup(&suricata);
```
victorjulien: why isn't this used instead?

src/Makefile.am
catenacyber: ok

src/Makefile.am
catenacyber: I spent some hours looking for a solution to this. Do you have any idea/pointer ? For some macro-like function for Makefile.am files...

src/Makefile.am
catenacyber: For instance, CI does not need libFuzzingEngine to build the fuzz targets with the simple driver onefile.c 
I use it as well to debug and reproduce crashes

src/tests/fuzz/fuzz_applayerparserparse.c
catenacyber: It made sense to me that fuzzing was some kind of testing.
What does that imply ?

src/tests/fuzz/fuzz_sigyamlpcap.c
catenacyber: I will do it

PR #3918, opened 2019-06-03
boyermoore: avoid one tolower call

catenacyber:
https://redmine.openinfosecfoundation.org/issues/1218

Describe changes:
- Updates `PreBmBcNocase` so that both upper and lower character give the same jump entry
- the characters had already been lowered by previous call to `BoyerMooreCtxToNocase`
- Updates `BoyerMooreNocase` so that we do not call `tolower` for the jump entries

For performance impact, I do not know if certain compilers managed to optimize away the second call to `tolower`

### DIFF
--- a/src/suricata.h
+++ b/src/suricata.h
@@ -177,6 +177,7 @@ extern int g_disable_randomness;
 
 #include <ctype.h>
 #define u8_tolower(c) tolower((uint8_t)(c))
+#define u8_toupper(c) toupper((uint8_t)(c))
 
 void EngineStop(void);
 void EngineDone(void);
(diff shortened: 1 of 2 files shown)


victorjulien:
Replaced by #3924

catenacyber:
#3924 is another separate PR

https://redmine.openinfosecfoundation.org/issues/1220
and 
https://redmine.openinfosecfoundation.org/issues/1218

are not the same even if they are both related to BoyerMoore

victorjulien:
Ah ok, reopening then.

jlucovsky:
I'd like to see some unittests added

catenacyber:
Good idea Jeff.
Do you have some idea about how to profile this ?

catenacyber:
Actually, there are already tests for Boyer Moore like
https://github.com/OISF/suricata/blob/master/src/util-spm.c#L632

Which unit tests would you like then ?

jlucovsky:
> Actually, there are already tests for Boyer Moore like
> https://github.com/OISF/suricata/blob/master/src/util-spm.c#L632
> 
> Which unit tests would you like then @jlucovsky ?

The test case you linked seems sufficient to cover your changes. Thanks!.

catenacyber:
And there are more than just one

PR #3917, opened 2019-06-03
Ebpf xdp 5.0 v2.8

regit:
Rebase of #3904 on latest master.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- rebase on top of latest master

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/460
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/241

PR #3925, opened 2019-06-05
Ebpf xdp 5.0 v2.9

regit:
Update of #3917 addressing comments and fixing bugs in particular in accounting. Result is far better now. For example on one of our busy live test box:
```
{
  "closed": 1505552,
  "closed_delta": 68793,
  "pkts": 350285153,
  "pkts_delta": 13317202,
  "bytes": 376952042963,
  "bytes_delta": 14405785152
}
```

Describe changes:
- Fix accounting
- Silent error handling for some hw bypass
- Better log messages

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/461
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/242


victorjulien:
I wonder if I'm hitting some kind of performance issue in the flow offload. See attached picture. The number of packets processed by the card is relatively low, even though I trigger bypass very aggressively (each syn packet). top shows the 'FB' thread at ~100% all the time.

Screenshot from 2019-06-06 07-41-46

victorjulien:
Is ```flow_bypassed.closed``` authoritative for the number of flows that have been offloaded to bpf? I notice that its a lot lower than ```flow.tcp```, even though I'm trying to offload all flows (by bypassing on SYN pkts). ```flow.tcp``` is 34x higher.

regit:
It is the number of half flows removed from the eBPF map.Le 6 juin 2019 08:01, Victor Julien <notifications@github.com> a écrit :Is flow_bypassed.closed authoritative for the number of flows that have been offloaded to bpf? I notice that its a lot lower than flow.tcp, even though I'm trying to offload all flows (by bypassing on SYN pkts). flow.tcp is 34x higher.

regit:
Regarding the performance issue: the FB thread is going to be costly. The current algorithm involves one syscall per half flow. I've already pointed that in netdev conference but there was no change provided (or I did miss it). I'm gonna try to open a discussion on the subject to see if they have a less costly solution to propose in upcoming release.

Consequence of this performance issue is that it can't take some time a flow get removed but we are I think under one minute delay.

victorjulien:
What does this costly FB thread mean for the timeliness of bypass half-flow inserts? To be of most value the inserts need to be (close to) real time.

regit:
The insert shoud not be affected. This is the timeout detection loop that is taking long. You can use the .flow.bypassed counter in the flow event to see the efficiency of bypass on a per flow basis.

victorjulien:
Ok. I don't have a flow.bypassed counter and a quick grep of the code doesn't show me where it's supposed to be registered or updated.

src/source-af-packet.c
```diff
@@ -2411,10 +2425,12 @@ static int AFPBypassCallback(Packet *p)
         }
         key.port16[0] = GET_TCP_DST_PORT(p);
         key.port16[1] = GET_TCP_SRC_PORT(p);
-        if (AFPInsertHalfFlow(p->afp_v.v6_map_fd, &key, inittime) == 0) {
+        if (AFPInsertHalfFlow(p->afp_v.v6_map_fd, &key, p->flow_hash,
+                              p->afp_v.nr_cpus) == 0) {
             return 0;
         }
-        EBPFUpdateFlow(p->flow, p);
+        if (p->flow)
```
victorjulien: can't be false due to check at function start

src/source-af-packet.c
```diff
@@ -2371,19 +2378,23 @@ static int AFPBypassCallback(Packet *p)
...
         key.dst = htonl(GET_IPV4_SRC_ADDR_U32(p));
         key.port16[0] = GET_TCP_DST_PORT(p);
         key.port16[1] = GET_TCP_SRC_PORT(p);
-        if (AFPInsertHalfFlow(p->afp_v.v4_map_fd, &key, inittime) == 0) {
+        if (AFPInsertHalfFlow(p->afp_v.v4_map_fd, &key, p->flow_hash,
+                              p->afp_v.nr_cpus) == 0) {
             return 0;
         }
-        EBPFUpdateFlow(p->flow, p);
+        EBPFUpdateFlow(p->flow, p, NULL);
```
victorjulien: return code is not checked, what is the consequence of failure of EBPFUpdateFlow?

Related: EBPFUpdateFlow documentation says its about IPS. So why is this even called if we're in IDS mode? It does a lookup and potentially a mem alloc, so it seems something you'd want to avoid for IDS

regit:
> Ok. I don't have a flow.bypassed counter and a quick grep of the code doesn't show me where it's supposed to be registered or updated.

It is in the event_type==flow:

Screenshot from 2019-06-06 10-14-00

src/source-af-packet.c
```diff
@@ -2284,12 +2291,12 @@ TmEcode AFPSetBPFFilter(AFPThreadVars *ptv)
  *
  * \param mapfd file descriptor of the protocol bypass table
  * \param key data to use as key in the table
- * \param inittime time of creation of the entry (in monotonic clock)
  * \return 0 in case of error, 1 if success
  */
-static int AFPInsertHalfFlow(int mapd, void *key, uint64_t inittime)
+static int AFPInsertHalfFlow(int mapd, void *key, uint32_t hash,
```
victorjulien: Not related to this patch set, but still things I want to mention:

1. use  ```} else if (PKT_IS_IPV6(p) &&```
2. this is wrong ```((IPV6_GET_NH(p) == IPPROTO_TCP) || (IPV6_GET_NH(p) == IPPROTO_UDP))) {``` I think. At the func start you already checked if the packet is TCP or UDP. Using IPV6_GET_NH is almost always wrong.
3. you can avoid all the htonl calls by directly accessing the packet ipv4 and ipv6 headers
4. I'm curious about the per call checks for ```p->afp_v.v4_map_fd```. Can we just check those at start up? Perhaps create different wrappers for the Callback in case v4 is available and v6 isn't?

src/source-af-packet.c
```diff
@@ -2441,40 +2457,42 @@ static int AFPXDPBypassCallback(Packet *p)
...
-        inittime = curtime.tv_sec * 1000000000;
-    }
     if (PKT_IS_IPV4(p)) {
         struct flowv4_keys key = {};
         if (p->afp_v.v4_map_fd == -1) {
             return 0;
         }
-        key.src = GET_IPV4_SRC_ADDR_U32(p);
-        key.dst = GET_IPV4_DST_ADDR_U32(p);
+        key.src = p->src.addr_data32[0];
```
victorjulien: here don't have the htonl

src/source-af-packet.c
```diff
@@ -2441,40 +2457,42 @@ static int AFPXDPBypassCallback(Packet *p)
...
+        key.dst = p->dst.addr_data32[0];
         /* In the XDP filter we get port from parsing of packet and not from skb
          * (as in eBPF filter) so we need to pass from host to network order */
-        key.port16[0] = htons(GET_TCP_SRC_PORT(p));
-        key.port16[1] = htons(GET_TCP_DST_PORT(p));
+        key.port16[0] = htons(p->sp);
```
victorjulien: here we do ntohs, but in the non-XDP version we don't

src/source-af-packet.c
victorjulien: ah I see the code doc explains this

ebpf/xdp_filter.c
```diff
@@ -65,65 +84,108 @@ struct flowv6_keys {
...
+    __u16 vlan_id[2];
+};
 
 struct pair {
-    __u64 time;
-    __u64 packets;
-    __u64 bytes;
-} __attribute__((__aligned__(8)));
+    __u32 packets;
+    __u32 bytes;
```
victorjulien: I guess this has consequences for accuracy on high bw elephant flows. We sync these counters once every ~10 seconds AFAICS. This means that we'd be unable to accurately track flows faster than ~4Gbit/s? These would overflow within that window. Is this correct? Are there other side effects?

src/util-ebpf.c
```diff
@@ -313,141 +497,196 @@ int EBPFSetupXDP(const char *iface, int fd, uint8_t flags)
...
     int mapfd = EBPFGetMapFDByName(dev->dev, name);
     struct flowv4_keys key = {}, next_key;
     int found = 0;
     unsigned int i;
-    unsigned int nr_cpus = UtilCpuGetNumProcessorsConfigured();
-    if (nr_cpus == 0) {
-        SCLogWarning(SC_ERR_INVALID_VALUE, "Unable to get CPU count");
+    uint64_t hash_cnt = 0;
+
+    if (tcfg->cpus_count == 0) {
```
victorjulien: seems like something that should be checked at init

src/util-ebpf.c
```diff
@@ -313,141 +497,196 @@ int EBPFSetupXDP(const char *iface, int fd, uint8_t flags)
...
+                            uint64_t pkts_cnt, uint64_t bytes_cnt);
+
 /**
  * Bypassed flows cleaning for IPv4
  *
  * This function iterates on all the flows of the IPv4 table
  * looking for timeouted flow to delete from the flow table.
  */
-static int EBPFForEachFlowV4Table(LiveDevice *dev, const char *name,
+static int EBPFForEachFlowV4Table(ThreadVars *th_v, LiveDevice *dev, const char *name,
```
victorjulien: I added a print statement to the start and end of this func to get a sense for how long it runs and for how many entries it finds. 3 observations:
```
[2206] 6/6/2019 -- 11:15:37 - (util-ebpf.c:324) <Info> (EBPFLoadFile) -- Loaded pinned maps, will use already loaded eBPF filter
[2206] 6/6/2019 -- 11:15:37 - (runmode-af-packet.c:533) <Info> (ParseAFPConfig) -- Loaded pinned maps from sysfs
[2206] 6/6/2019 -- 11:15:37 - (util-runmodes.c:297) <Info> (RunModeSetLiveCaptureWorkersForDevice) -- Going to use 12 thread(s)
[2242] 6/6/2019 -- 11:15:37 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2206] 6/6/2019 -- 11:15:37 - (util-conf.c:115) <Info> (ConfUnixSocketIsEnable) -- Running in live mode, activating unix socket
[2206] 6/6/2019 -- 11:15:37 - (unix-manager.c:131) <Info> (UnixNew) -- Using unix socket file '/var/run/suricata/suricata-command.socket'
[2206] 6/6/2019 -- 11:15:37 - (tm-threads.c:2157) <Notice> (TmThreadWaitOnThreadInit) -- all 12 packet processing threads, 9 management threads initialized, engine started.
[2235] 6/6/2019 -- 11:15:40 - (source-af-packet.c:510) <Info> (AFPPeersListReachedInc) -- All AFP capture threads are running.
[2242] 6/6/2019 -- 11:16:00 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 260908
[2242] 6/6/2019 -- 11:16:02 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2242] 6/6/2019 -- 11:16:23 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 262144
[2242] 6/6/2019 -- 11:16:35 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2242] 6/6/2019 -- 11:16:56 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 262144
[2242] 6/6/2019 -- 11:17:07 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2242] 6/6/2019 -- 11:17:35 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 287759
[2242] 6/6/2019 -- 11:17:46 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2242] 6/6/2019 -- 11:18:22 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 359593
[2242] 6/6/2019 -- 11:18:33 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2242] 6/6/2019 -- 11:18:57 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 274873
[2242] 6/6/2019 -- 11:19:09 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2242] 6/6/2019 -- 11:19:33 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 273435
[2242] 6/6/2019 -- 11:19:44 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2242] 6/6/2019 -- 11:20:16 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 334341
[2242] 6/6/2019 -- 11:20:27 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[2242] 6/6/2019 -- 11:20:55 - (util-ebpf.c:692) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 300721
```

1. it starts before af-packet is ready. Not sure if this is intentional.
2. it seems to find more entries than my .max_entries setting of 262144
3. runtime is really long 20 seconds is no exception.

So it seems for that ~20 seconds we're doing 262k+ flow table lookups and updates. That seems to be an expensive operation.

victorjulien: Ok the more than 256k thing seems to be explained by the fact that while we loop and remove dead flows, new ones get inserted and we loop over (part of) those as well:
```
[13510] 6/6/2019 -- 11:30:05 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[13510] 6/6/2019 -- 11:30:26 - (util-ebpf.c:693) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 262144, dead 0, diff 262144
[13510] 6/6/2019 -- 11:30:38 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[13510] 6/6/2019 -- 11:31:09 - (util-ebpf.c:693) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 306348, dead 125551, diff 180797
[13510] 6/6/2019 -- 11:31:20 - (util-ebpf.c:616) <Notice> (EBPFForEachFlowV4Table) -- start
[13510] 6/6/2019 -- 11:31:55 - (util-ebpf.c:693) <Notice> (EBPFForEachFlowV4Table) -- end: hash_cnt 357770, dead 132159, diff 225611
```

victorjulien: In my test it seems about 1/4 to 1/3 of the flows are 'dead' each time. I wonder if it would make sense to skip the newly added ones? I guess this would require adding a field to 'struct pair'. I think the newly added flows can't time out yet because of FLOW_BYPASSED_TIMEOUT == 100 seconds?

victorjulien: I guess we still want to inspect them for their stats though.

victorjulien:
I guess my testing is not really addressing the intended use case. I suspect performance will suffer if we offload to many flows, and the overhead of the table maintenance likely exceeds the benefits. I will think about how I can test the intended use case of dealing with a far smaller number of elephant flows.

This is something that might have to be explained in the docs though. Some sort of guidance of that to do and not do.

src/util-ebpf.c
regit: We dump the flow table from the kernel/userspace map so we can't skip some. And also we get the stats.

src/source-af-packet.c
regit: This is linked with the parsing function that revert the u16 at read in one case and not in the other.

regit:
If there is no improvement in the flow table handling we may have to use more advanced strategy for the bypass. Maybe we can try to hardware bypass only flow with high bandwidth.

pevma:
I would personally +1 this one as this is in essence what we try to avoid most - inspecting big/unneeded to look at flows.

victorjulien:
One relatively simple way of reducing the maintenance cost by almost 50% would be to switch to using bidir flows in the bpf program. This would half the number of records needed for one Suricata flow and also half the number of flow engine updates that would have to be done by Suricata. The cost would be a bit more complexity in the bpf program, but not much.

victorjulien:
Just spoke with Simon on the counter overflow issue, and it seems that with other solutions they simply sync more often. When taking the extreme case of a single 40Gbit/s flow, the overflow would happen approximately every second. This means we'd need to sync at least at that interval too, but probably more often to allow for timing issues. This of course implies too that the actual maintenance loop we perform runs quickly enough to have a sub-second interval.

regit:
This is strictly impossible with current XDP framework. I wonder if we should just pick pckts or bytes and use one single counter. This way we can use a u64.

victorjulien:
What is 'strictly impossible'? Not clear what you're responding to.

regit:
Sorry, dumping the flow table in subsecond time is not possible.

victorjulien:
I've been playing a bit to get a feel for how the dumping of the flow table works/performs. It seems there is a significant cost to the first call to ```bpf_map_get_next_key```, maybe because it's called with a 0 key and has to do something special? The cost on this system (under unrelated load) of this call with an empty table seems to be ~1 second. If the table is full this call appears faster.

Even with a 1024 map size limit, the flow dump will never finish within a second. It's more likely to go into 2 secs. This is still true if the call to the EBPFOpFlowForKey is commented out and all flows are considered 'dead'. I think this serves to show the overhead of the flow dumping itself.

So I'm wondering if the whole u32 for counter is going to fly.

40Gbit/s flow will wrap it in ~1 sec.
10Gbit/s flow in ~4 sec.
1Gbit/s flow in ~40 sec.

Even the 40 sec doesn't seem safe. I've seen runs on over a minute.

So if the u32 counters are not usable, we need to consider what to do with them. Maybe we can get advice on how to speed up the flow table dump. Or on how to get u64 counters. But if neither is feasible it seems we should consider what it means for the implementation. I think it could be simpler as it would ignore the counters other than for handling flow timeouts.

Thoughts?

regit:
Thanks a lot for the analysis. What about we drop the bytes counter in the case of netronome hw bypass and we keep the 64 bit counters for the general case ?

victorjulien:
Dumping some more stats:
```
1.5M max_entries
[39174] 7/6/2019 -- 10:04:42 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 235.523s: total: 2352762 (9989/s) removed: 1755104 (7451/s)
[39174] 7/6/2019 -- 10:08:56 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 242.445s: total: 2400035 (9899/s) removed: 1804656 (7443/s)
[39174] 7/6/2019 -- 10:13:08 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 241.073s: total: 2404982 (9976/s) removed: 1808289 (7501/s)
[39174] 7/6/2019 -- 10:17:22 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 242.214s: total: 2397255 (9897/s) removed: 1798532 (7425/s)
[39174] 7/6/2019 -- 10:21:38 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 244.554s: total: 2403613 (9828/s) removed: 1816480 (7427/s)

262k max_entries
[43098] 7/6/2019 -- 10:35:01 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 18.458s: total: 240915 (13052/s) removed: 1 (0/s)
[43098] 7/6/2019 -- 10:35:32 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 19.719s: total: 262146 (13294/s) removed: 4 (0/s)
[43098] 7/6/2019 -- 10:36:03 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 19.662s: total: 262147 (13332/s) removed: 5 (0/s)
[43098] 7/6/2019 -- 10:36:42 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 27.258s: total: 299823 (10999/s) removed: 134595 (4937/s)
[43098] 7/6/2019 -- 10:37:25 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 32.121s: total: 362249 (11277/s) removed: 145698 (4535/s)

32k max_entries
[42123] 7/6/2019 -- 10:26:40 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 1.779s: total: 7274 (4088/s) removed: 0 (0/s)
[42123] 7/6/2019 -- 10:26:45 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.707s: total: 32768 (8839/s) removed: 0 (0/s)
[42123] 7/6/2019 -- 10:27:00 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.672s: total: 32771 (8924/s) removed: 4 (1/s)
[42123] 7/6/2019 -- 10:27:15 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.619s: total: 32768 (9054/s) removed: 4 (1/s)
[42123] 7/6/2019 -- 10:27:29 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.615s: total: 32770 (9065/s) removed: 4 (1/s)
[42123] 7/6/2019 -- 10:27:45 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.717s: total: 32769 (8815/s) removed: 4 (1/s)
[42123] 7/6/2019 -- 10:28:00 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 4.130s: total: 32770 (7934/s) removed: 4 (0/s)
[42123] 7/6/2019 -- 10:28:15 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.626s: total: 32770 (9037/s) removed: 4 (1/s)
[42123] 7/6/2019 -- 10:28:32 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 5.679s: total: 46887 (8256/s) removed: 29610 (5213/s)
[42123] 7/6/2019 -- 10:28:48 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 4.285s: total: 36307 (8473/s) removed: 6498 (1516/s)
[42123] 7/6/2019 -- 10:29:03 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.684s: total: 33414 (9070/s) removed: 1334 (362/s)
[42123] 7/6/2019 -- 10:29:17 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.629s: total: 32895 (9064/s) removed: 276 (76/s)
[42123] 7/6/2019 -- 10:29:32 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.638s: total: 32818 (9020/s) removed: 102 (28/s)
[42123] 7/6/2019 -- 10:29:47 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.678s: total: 32798 (8917/s) removed: 58 (15/s)
[42123] 7/6/2019 -- 10:30:02 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 3.658s: total: 32778 (8960/s) removed: 32 (8/s)
[42123] 7/6/2019 -- 10:30:18 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 4.197s: total: 36309 (8651/s) removed: 9092 (2166/s)
[42123] 7/6/2019 -- 10:30:34 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 5.034s: total: 43298 (8601/s) removed: 19942 (3961/s)
[42123] 7/6/2019 -- 10:30:50 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 4.109s: total: 36412 (8861/s) removed: 7054 (1716/s)

1k max_entries
[42724] 7/6/2019 -- 10:32:15 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 1.238s: total: 0 (0/s) removed: 0 (0/s)
[42724] 7/6/2019 -- 10:32:18 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 1.366s: total: 1000 (732/s) removed: 0 (0/s)
[42724] 7/6/2019 -- 10:32:30 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 1.378s: total: 1028 (746/s) removed: 6 (4/s)
[42724] 7/6/2019 -- 10:32:43 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 1.395s: total: 1028 (736/s) removed: 6 (4/s)
[42724] 7/6/2019 -- 10:32:56 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 1.382s: total: 1026 (742/s) removed: 6 (4/s)
[42724] 7/6/2019 -- 10:33:08 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 1.393s: total: 1026 (736/s) removed: 6 (4/s)
[42724] 7/6/2019 -- 10:33:21 - (util-ebpf.c:703) <Notice> (EBPFForEachFlowV4Table) -- time elapsed 1.393s: total: 1030 (739/s) removed: 6 (4/s)
```
It seems that we max out at about 8-13k bpf 'get_next' operations per second and about 8k delete operations per second.

victorjulien:
I've been told that 64bit counters on the netronome hardware is hard. I tried changing the filter to use u64 myself and the program wouldn't load. It's also not just a counter, its a atomic counter.

I'm thinking that for the netronome hw offload case we might have to accept (and document!) that we just loose all stats. Then we could simply set a bool on receiving a packet, which would indicate that since the last time the flow table was dumped we've seen at least one packet and the flow will not time out this iteration. This also means that the timeout of the flow will be quite non-deterministic, which should also be documented. I wonder if in this case we could even let the FlowManager handle this by doing direct lookups instead of this flow table dump process.

@dcbeckett do you have ideas on why we appear to see a start up cost to the flow hash dump?

regit:
@victorjulien just to be sure, you have tested with a single 64 counters or with 2 64 bit counters ?

victorjulien:
I converted both, should I try just doing one?

regit:
Yes in a previous mail I was told to use u32 as 2 u64 would be too big for the map value.

victorjulien:
It seems if I make packets u64 and comment out bytes it will actually accept the program.

regit:
OK great, my suggestion was to propose a code with only packet counter in 64 bit in the case of netronome.

victorjulien:
It seems to work, I see packets in the flow.bypassed object:
```json
{
  "timestamp": "2019-06-07T15:26:43.099385+0200",
  "flow_id": 1498799269479250,
  "in_iface": "ens5np0",
  "event_type": "flow",
  "src_ip": "16.0.0.83",
  "src_port": 39557,
  "dest_ip": "48.0.144.194",
  "dest_port": 445,
  "proto": "TCP",
  "flow": {
    "pkts_toserver": 77,
    "pkts_toclient": 0,
    "bytes_toserver": 66,
    "bytes_toclient": 0,
    "bypassed": {
      "pkts_toserver": 76,
      "pkts_toclient": 0,
      "bytes_toserver": 0,
      "bytes_toclient": 0
    },
    "start": "2019-06-07T15:26:36.197458+0200",
    "end": "2019-06-07T15:26:22.197458+0200",
    "age": -14,
    "bypass": "local",
    "state": "bypassed",
    "reason": "timeout",
    "alerted": false
  },
  "tcp": {
    "tcp_flags": "02",
    "tcp_flags_ts": "02",
    "tcp_flags_tc": "00",
    "syn": true,
    "state": "syn_sent"
  }
}
```
(btw note the negative age, that will need a fix)

victorjulien:
After discussing a bit with Simon there are a couple of extra thoughts:

1. we might be able to keep the pkts counter 32 bit. In theory a 40Gbit/s flow with 64 byte packets would be ~60M pkts/s, or overflow u32 in 71 sec. However, that is really not something that can be expected. E.g. according to Simon DPDK would be limited to something like 4M pkts/s per flow for generating packets. That would give us a 1000 sec, so plenty.

2. we may choose a different type of trade off. What if we can detect the overflow in the bpf program (e.g. new value < old value) and reset the bytes to 0. Then both the bpf program in for future packets and Suricata could see that pkt cnt > 0 and bytes == 0, so don't show the bytes in the log/stats. This way we can still have the bytes in the vast majority of cases.

victorjulien:
Another thing to consider is the number of syscalls we do. Right now we seem to process about 13k half flows per second. This means 26k/s syscalls as we do a ```bpf_map_get_next_key``` followed by a ```bpf_map_lookup_elem```. Then I saw about 8k/s of half flow removals, so a call to ```bpf_map_delete_elem```.

The total of this is about 34k/s, which I've told is getting quite close to a max number for the card.

2 takeaways from this:

1. syscalls are expensive

2. we do 4 per Suricata::Flow during hash maintenance, plus 2 per removal

So we do something expensive, then we do it many times for a single unit of data (from Suricatas view).

I wonder if can avoid this. A simple suggestion is to use bidir flows in the bpf and it's hash. This would half the number of items in the hash map, it would half the number of syscalls. It would somewhat increase the data portion 'struct pair' to allow tracking of ts and tc counters. It would slightly increase processing cost per packet (untested).

The second place where we could potentially save a lot of work, is in the ```bpf_map_get_next_key```. We should be able to know all the keys, as we inserted them ourselves. So by attaching them to the Suricata::Flow we could simply do the lookup directly, w/o the need to walk the entire bpf hash map all the time. Suricata's flow manager already walks the hash, although almost never fully due to optimizations to avoid doing just that. But I feel it should be able to do a bit of extra work for bypassed flows.

If we remove the key iteration phase, we might still want to do that at startup (to pick up or remove old flows) and maybe periodically if we want to allow other parts of the system to add flows to the hash map w/o a Suricata::Flow also present.

regit:
Agree on the fact that a method update to iterate in Suricata on bypassed flows and query when needed would lower the syscall count. I had similar thoughts. We could for instance check only flow that could potentially timeout.
Doing so we would loose some real time accounting because a single flow would be accounted only every timeout and we would also destroy flow a bit later (max destroy time at 2 * interval check).

The thing I don't like is that we have to iterate on all the flows.

ebpf/xdp_filter.c
```diff
@@ -55,7 +73,8 @@ struct flowv4_keys {
         __u16 port16[2];
     };
     __u32 ip_proto;
-} __attribute__((__aligned__(8)));
+    __u16 vlan_id[2];
```
victorjulien: does this contain just the vlan_id? The id is 12bit, so then we'd have 4 bits left and again for the 2nd vlan_id. We just support TCP or UDP, so instead of using 4 bytes for the ipproto, we could use 1 bit of one of the vlan_id fields.

I think this could be beneficial as we seem to be starved for space a bit. IIRC we can only use 64 bytes for key + value.

victorjulien:
The Suricata flow manager would not iterate on all the flows other than how it does now. The bypassed flows simply would have their own timeout and might be reevaluated a bit sooner depending on the timeout settings. This reevaluation process that is run per flow would then involve an extra step, namely the retrieval of the flow representation from bpf to see if packets were seen. If yes, the timeout is reset for that flow, if no it can time out.

The only new thing about this process would be the fetching on the flow representation from bpf.

I guess if you think this would have a big impact on the flow engine we could explore other ways too. Like for example create a separate storage with pointers to only bypassed flows (flows are ref counted so won't go away as long as you hold a ref) and scan through that at an interval.

regit:
I think it really worth a try. I'm gonna try to implement that this week end.Le 7 juin 2019 19:58, Victor Julien <notifications@github.com> a écrit :The Suricata flow manager would not iterate on all the flows other than how it does now. The bypassed flows simply would have their own timeout and might be reevaluated a bit sooner depending on the timeout settings. This reevaluation process that is run per flow would then involve an extra step, namely the retrieval of the flow representation from bpf to see if packets were seen. If yes, the timeout is reset for that flow, if no it can time out.
The only new thing about this process would be the fetching on the flow representation from bpf.
I guess if you think this would have a big impact on the flow engine we could explore other ways too. Like for example create a separate storage with pointers to only bypassed flows (flows are ref counted so won't go away as long as you hold a ref) and scan through that at an interval.

regit:
Yes, it is only used for vlanid checking. So compressing the data could be done.Le 7 juin 2019 19:50, Victor Julien <notifications@github.com> a écrit :@victorjulien commented on this pull request.



In ebpf/xdp_filter.c:
> @@ -55,7 +73,8 @@ struct flowv4_keys {
         __u16 port16[2];
     };
     __u32 ip_proto;
-} __attribute__((__aligned__(8)));
+    __u16 vlan_id[2];

does this contain just the vlan_id? The id is 12bit, so then we'd have 4 bits left and again for the 2nd vlan_id. We just support TCP or UDP, so instead of using 4 bytes for the ipproto, we could use 1 bit of one of the vlan_id fields.
I think this could be beneficial as we seem to be starved for space a bit. IIRC we can only use 64 bytes for key + value.

victorjulien:
Btw I noticed that the xdp ipv6 filter bails out if the nexthdr isn't set to 6 or 17, so it would fail is any exthdr is in place.

regit:
I'm currently working on this branch to address discussion here https://github.com/regit/suricata/tree/ebpf-xdp-5.0-v2.10

At time of writing, got 64 bits counters. Only an accounting command is missing.

src/util-ebpf.c
regit: Removing the warning and keeping the sanity check as the function is not called often.

regit:
On xdp IPv6 filter, we return XDP_PASS so it reaches Suricata as soon as it has the exthdr.

PR #3940, opened 2019-06-09
Ebpf xdp 5.0 v2.10

regit:
Update of #3925 implementing a callback strategy and addressing comments.

Describe changes:
- callback strategy for bypass to avoid loop
- 64bits counters

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/462
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/243


src/flow-manager.c
```diff
@@ -248,6 +249,26 @@ static int FlowManagerFlowTimeout(const Flow *f, enum FlowState state, struct ti
 
     /* do the timeout check */
     if (flow_times_out_at >= ts->tv_sec) {
+        /* for bypassed flow we need to check if the capture method has seen
+         * something and update the logic */
+        FlowBypassInfo *fc = FlowGetStorageById(f, GetFlowBypassInfoID());
```
victorjulien: should we wrap this in if (flow state is non-local bypass) { ?

src/flow-util.c
```diff
@@ -204,21 +204,26 @@ void FlowInit(Flow *f, const Packet *p)
...
-    return g_bypass_counter_id;
+    return g_bypass_info_id;
 }
 
-static void FlowCounterFree(void *x)
+static void FlowBypassFree(void *x)
 {
+    FlowBypassInfo *fb = (FlowBypassInfo *) x;
+
+    if (fb->bypass_data && fb->BypassFree) {
```
victorjulien: check below 'x' can be NULL, but dereference here in unconditional

regit: indeed

src/flow-manager.c
regit: Yes, good catch.

src/flow-manager.c
```diff
@@ -248,6 +249,26 @@ static int FlowManagerFlowTimeout(const Flow *f, enum FlowState state, struct ti
...
+        if (fc) {
```
victorjulien: ```if (fc && fc->BypassUpdate) {```

src/flow-manager.c
```diff
@@ -248,6 +254,51 @@ static int FlowManagerFlowTimeout(const Flow *f, enum FlowState state, struct ti
...
+                    if (counters) {
+                        counters->bypassed_pkts += pkts_tosrc + pkts_todst;
+                        counters->bypassed_bytes += bytes_tosrc + bytes_todst;
+                    }
+                    return 0;
+                } else {
+                    SCLogDebug("No new packet, dead flow %ld", FlowGetId(f));
+                    if (f->livedev) {
+                        if (FLOW_IS_IPV4(f)) {
+                            LiveDevAddBypassStats(f->livedev, -1, AF_INET);
```
victorjulien: this -1 doesn't seem to be handled? I guess the best case is that it adds it to the counter.

src/flow-manager.c
victorjulien: can this whole logic move to a (inline) util func? Fine if it stays in this file.

src/flow-manager.c
```diff
@@ -248,6 +254,51 @@ static int FlowManagerFlowTimeout(const Flow *f, enum FlowState state, struct ti
 
     /* do the timeout check */
     if (flow_times_out_at >= ts->tv_sec) {
+        /* for bypassed flow we need to check if the capture method has seen
```
victorjulien: Flow is not locked at this point. I think we should lock it inside the util func I suggested, which should only be called for bypassed flows.

victorjulien:
```
Date: 6/10/2019 -- 09:58:01 (uptime: 0d, 01h 30m 35s)
------------------------------------------------------------------------------------
Counter                                       | TM Name                   | Value
------------------------------------------------------------------------------------
capture.kernel_packets                        | Total                     | 7645799086
capture.kernel_drops                          | Total                     | 1430643515
decoder.pkts                                  | Total                     | 6216443971
decoder.bytes                                 | Total                     | 1166854336137
decoder.ipv4                                  | Total                     | 6216443970
decoder.ipv6                                  | Total                     | 1
decoder.ethernet                              | Total                     | 6216443971
decoder.tcp                                   | Total                     | 6216443970
decoder.icmpv6                                | Total                     | 1
decoder.avg_pkt_size                          | Total                     | 187
decoder.max_pkt_size                          | Total                     | 1514
flow.tcp                                      | Total                     | 134907876
flow.icmpv6                                   | Total                     | 1
tcp.sessions                                  | Total                     | 134893763
tcp.syn                                       | Total                     | 134893763
flow_mgr.new_pruned                           | Total                     | 14114
flow_mgr.bypassed_pruned                      | Total                     | 134479153
flow.spare                                    | Total                     | 231652
flow_mgr.flows_checked                        | Total                     | 136885
flow_mgr.flows_notimeout                      | Total                     | 108942
flow_mgr.flows_timeout                        | Total                     | 27943
flow_mgr.flows_removed                        | Total                     | 27943
flow_mgr.rows_checked                         | Total                     | 1000000
flow_mgr.rows_skipped                         | Total                     | 879837
flow_mgr.rows_empty                           | Total                     | 18190
flow_mgr.rows_maxlen                          | Total                     | 11
flow_bypassed.closed                          | Total                     | 139388
flow_bypassed.pkts                            | Total                     | 6698425
flow_bypassed.bytes                           | Total                     | 1240218890
tcp.memuse                                    | Total                     | 124620160
tcp.reassembly_memuse                         | Total                     | 1179648
flow.memuse                                   | Total                     | 250517296
```
It seems there is a mismatch between the flow_bypassed.* stats and the ```flow_mgr.bypassed_pruned``` stat. I think it would be really helpful to get some stats around insertions. E.g. ```flow.bypassed.success``` and ```flow.bypassed.failed```. Or even something like ```flow.bypassed.failed.egain``` and the same for other codes. Right now it seems there is no way to know if perhaps no new flows are added to the card, or if it even tries to.

src/util-ebpf.c
```diff
@@ -514,11 +514,13 @@ static bool EBPFCreateFlowForKey(struct flows_stats *flowstats, FlowKey *flow_ke
...
      * on the flow */
-    FlowCounters *fc = FlowGetStorageById(f, GetFlowBypassCounterID());
+    FlowBypassInfo *fc = FlowGetStorageById(f, GetFlowBypassInfoID());
     if (fc == NULL) {
-        fc = SCCalloc(sizeof(FlowCounters), 1);
+        fc = SCCalloc(sizeof(FlowBypassInfo), 1);
         if (fc) {
-            FlowSetStorageById(f, GetFlowBypassCounterID(), fc);
+            FlowSetStorageById(f, GetFlowBypassInfoID(), fc);
+            fc->BypassUpdate = EBPFBypassUpdate;
```
victorjulien: why is this stored in the flow? It seems like something that will be the same for all bypassed flows, so would fit better in the flow-manager ctx itself somehow

src/util-ebpf.c
```diff
@@ -533,56 +535,97 @@ static bool EBPFCreateFlowForKey(struct flows_stats *flowstats, FlowKey *flow_ke
     return false;
 }
 
+void EBPFBypassFree(void *data)
+{
+    EBPFBypassData *eb = (EBPFBypassData *)data;
+    if (eb == NULL)
+        return;
+    if (eb->key[0]) {
```
victorjulien: can the key never be 0?

src/util-ebpf.c
```diff
@@ -533,56 +535,97 @@ static bool EBPFCreateFlowForKey(struct flows_stats *flowstats, FlowKey *flow_ke
...
+ * Update lastts in the flow and do accounting
+ *
+ * */
+bool EBPFBypassUpdate(Flow *f, void *data, time_t tsec)
+{
+    EBPFBypassData *eb = (EBPFBypassData *)data;
+    if (eb == NULL) {
         return false;
+    }
+    bool activity = EBPFBypassCheckHalfFlow(f, eb, eb->key[0], 0);
```
victorjulien: How about something like:
```c
/* call both unconditionally to update counters */
bool activity = EBPFBypassCheckHalfFlow(f, eb, eb->key[0], 0);
activity |= EBPFBypassCheckHalfFlow(f, eb, eb->key[1], 1));
```

src/flow-manager.c
victorjulien: actually, it seems better to call it from ```FlowManagerFlowTimedOut``` instead, which is called after the flow is locked.

src/util-ebpf.c
regit: Are we sure an optimization will not skip the eval if activity is already true ?

victorjulien:
Could you add the following counters/stats:
1. active hflows in bpf (atomic counter between inserts/deletes)
2. packets (and bytes) tracked by local bypass and by bpf bypass. This should give a sense to how much we successfully offload
3. insert success/failure (see comment above)

All these should give indications how how full the bpf map is. If it is not full, it will also tell us something about the window between the insert into the map and the effective bpf offload time.

src/util-ebpf.c
victorjulien: I don't think it should. We use this pattern elsewhere too.

src/util-ebpf.c
regit: I was thinking about mixed capture setup. We may end up with different callback functions.

victorjulien: Ok. Just feels like spending 16 bytes per flow for that use case is a bit of a heavy price.

src/util-ebpf.c
```diff
@@ -313,141 +497,261 @@ int EBPFSetupXDP(const char *iface, int fd, uint8_t flags)
...
-    return 0;
+    for (i = 0; i < eb->cpus_count; i++) {
+        /* let's start accumulating value so we can compute the counters */
+        SCLogDebug("%d: Adding pkts %lu bytes %lu", i,
+                BPF_PERCPU(values_array, i).packets,
+                BPF_PERCPU(values_array, i).bytes);
+        pkts_cnt += BPF_PERCPU(values_array, i).packets;
+        bytes_cnt += BPF_PERCPU(values_array, i).bytes;
+    }
+    FlowBypassInfo *fc = FlowGetStorageById(f, GetFlowBypassInfoID());
```
victorjulien: it seems we can optimize things by moving this to ```EBPFBypassUpdate```, that would safe us one lookup. However, we could safe us both as in the flow manager we've already looked it up just before we call ```EBPFBypassUpdate```

regit: Indeed

regit:
On counter:
- [x] active bpf flow: via iface-bypass-stats
- [ ] packet local/capture bypass
- [ ] insert success/failure

3. to be done

src/source-af-packet.c
```diff
@@ -2487,32 +2590,61 @@ static int AFPXDPBypassCallback(Packet *p)
...
+            keys[1]->dst[i] = GET_IPV6_SRC_ADDR(p)[i];
+        }
+        keys[1]->port16[0] = htons(GET_TCP_DST_PORT(p));
+        keys[1]->port16[1] = htons(GET_TCP_SRC_PORT(p));
+        keys[1]->vlan0 = p->vlan_id[0];
+        keys[1]->vlan1 = p->vlan_id[1];
+        keys[1]->ip_proto = keys[0]->ip_proto;
+        if (AFPInsertHalfFlow(p->afp_v.v6_map_fd, keys[1],
+                              p->afp_v.nr_cpus) == 0) {
+            SCFree(keys[0]);
```
victorjulien: Unless I'm missing some kind of garbage collection, I think this leads to loss of map space. Half flow 1 is inserted, but when the 2nd fails there is not connection of the key of half flow 1 to a Suricata::Flow.

I think the only way to recover from this in the current model would be a flow table dump from the card.

It's probably better to handle this cleaner here, like remove the half flow 1 again, or store the key of a single half flow in the Suricata::Flow as well.

src/source-af-packet.c
```diff
@@ -2487,32 +2590,61 @@ static int AFPXDPBypassCallback(Packet *p)
...
+            SCFree(keys[1]);
             return 0;
         }
-        return 1;
+        LiveDevAddBypassStats(p->livedev, 1, AF_INET6);
+        return AFPSetFlowStorage(p, p->afp_v.v6_map_fd, keys[0], keys[1]);
```
victorjulien: AFPSetFlowStorage can fail, and this will lead to the same problem, but now for 2 half flows.

src/source-af-packet.c
regit: I did do that in next branch: https://github.com/regit/suricata/commit/eef83ca264e2bfa3fe82c621a7f9406d458bd287

src/source-af-packet.c
regit: Indeed, I'm gonna fix that in error handling.

src/flow-manager.c
regit: In fact, we will have the storage if and only if we have a capture bypassed flow so IMO we should keep it like that.

src/source-af-packet.c
victorjulien: Counter to what I said in chat, fixing this does help with the offloaded flow numbers. My quick and dirty fix was to call ```EBPFDeleteKey``` on the already inserted half flows in the error case for insert failure of the 2nd hflow (first updated the non-XDP callback by accident, explaining the misinfo)

src/flow-manager.c
victorjulien: Right ok. Although checking 'state' (a local variable) is probably cheaper than a memory access at the end of the structure.

src/source-af-packet.c
regit: Ok great!

regit:
I did implement that:
```
>>> iface-bypassed-stat
Success:
{
    "wlp3s0": {
        "ipv4_maps_count": 66,
        "ipv4_fail": 0,
        "ipv4_success": 82,
        "ipv6_maps_count": 0,
        "ipv6_fail": 0,
        "ipv6_success": 0
    }
}
```
It should be enough to get an idea of the insertion failure rate.

PR #3941, opened 2019-06-10
Ebpf xdp 5.0 v2.11

regit:
Updated of #3940 addressing comments.

It updates the iface counter:

```
>>> iface-bypassed-stat
Success:
{
    "wlp3s0": {
        "ipv4_maps_count": 66,
        "ipv4_fail": 0,
        "ipv4_success": 82,
        "ipv6_maps_count": 0,
        "ipv6_fail": 0,
        "ipv6_success": 0
    }
}
```

And add pkts and bytes counters in stats:

```
{
  "local_pkts": 0,
  "local_bytes": 0,
  "closed": 84,
  "pkts": 4799,
  "bytes": 2975133
}
```

Describe changes:
- Add counters
- Better error handling

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/467
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/248


src/flow-manager.c
```diff
@@ -254,6 +259,54 @@ static int FlowManagerFlowTimeout(const Flow *f, enum FlowState state, struct ti
     return 1;
 }
 
+static inline int FlowBypassedTimeout(Flow *f, struct timeval *ts,
+                                      FlowTimeoutCounters *counters)
```
victorjulien: I don't think counters here should be optional. AFAICS it will only be called with NULL from some unittests. So those should just also pass counters.

src/flow-util.c
```diff
@@ -203,3 +203,30 @@ void FlowInit(Flow *f, const Packet *p)
...
+{
+    FlowBypassInfo *fb = (FlowBypassInfo *) x;
+
+    if (fb == NULL)
+        return;
+
+    if (fb->bypass_data && fb->BypassFree) {
+        fb->BypassFree(fb->bypass_data);
+    }
+    if (x)
```
victorjulien: can't be false, as ```fb``` was already NULL checked

src/flow-worker.c
```diff
@@ -67,14 +70,17 @@ typedef struct FlowWorkerThreadData_ {
  *
  *  Handle flow creation/lookup
  */
-static inline TmEcode FlowUpdate(Packet *p)
+static inline TmEcode FlowUpdate(Packet *p, ThreadVars *tv, FlowWorkerThreadData *fw)
```
victorjulien: Normal pattern is to have tv as first arg, then fw, then p

src/flow-worker.c
```diff
@@ -67,14 +70,17 @@ typedef struct FlowWorkerThreadData_ {
...
 {
     FlowHandlePacketUpdate(p->flow, p);
 
     int state = SC_ATOMIC_GET(p->flow->flow_state);
     switch (state) {
         case FLOW_STATE_CAPTURE_BYPASSED:
+            return TM_ECODE_DONE;
```
victorjulien: if we'd count these as well separately we'd get a sense for latency for applying offload?

regit: Yes indeed.

src/source-af-packet.h
```diff
@@ -137,6 +151,7 @@ typedef struct AFPPacketVars_
     uint8_t copy_mode;
     int v4_map_fd;
     int v6_map_fd;
+    unsigned int nr_cpus;
```
victorjulien: ifdef HAVE_PACKET_EBPF for this and the map fds?

src/util-ebpf.h
```diff
@@ -40,8 +40,10 @@ struct flowv4_keys {
 		__be32 ports;
 		__be16 port16[2];
 	};
-	__u32 ip_proto;
-}  __attribute__((__aligned__(8)));
+    __u8 ip_proto:1;
```
victorjulien: I always thought you are supposed to keep the type the same for those bit notation fields.

victorjulien:
In the new strategy to let the flow manager do the bpf lookups, how are we handling entries already in the maps when Suricata starts up? Do we do a pass at Suricata start up or otherwise to pick up flows already in the map?

regit:
Yes, we are reading and creating flow entries at start to create bypassed suricata::Flow.

victorjulien:
Can you add output to show much we picked up?

src/flow-manager.c
regit: I don't understand your comment.

victorjulien:
On 11-06-19 23:14, Eric Leblond wrote:
> 
> 
> In src/flow-manager.c
> <https://github.com/OISF/suricata/pull/3941#discussion_r292662721>:
> 
>> @@ -254,6 +259,54 @@ static int FlowManagerFlowTimeout(const Flow *f, enum FlowState state, struct ti
>      return 1;
>  }
>  
> +static inline int FlowBypassedTimeout(Flow *f, struct timeval *ts,
> +                                      FlowTimeoutCounters *counters)
> 
> I don't understand your comment.

The ```counters``` pointer here is treated in the function as if it can
be NULL. I don't think that is correct. When looking at the code, the
only place I see it being NULL is in unittests. It's not reasonable to
have the runtime NULL checks just because not all unittests call the
function correctly.

src/util-ebpf.h
regit: I did not manage to find the norm but a few docs use that.

PR #3942, opened 2019-06-11
Pr/3934/jeff/ftp/v4

victorjulien:
Takes #3934 and adds a bunch of suggested changes.

I ran it on live traffic and noticed that the logs were only generated at flow end, instead of when a tx was supposed to be ready. This is fixed by https://github.com/OISF/suricata/commit/919de882b4d381bf620a1c7eae1bebd67d73a5dc
(this was easy to reproduce with a pcap, all the logs had the same 'pcap_cnt' and 'timestamp' value).

However, after this I started seeing crashes, as txs now disappear much sooner (when marked complete and they are logged, they are removed), so needed https://github.com/OISF/suricata/commit/1d1aefe8e9aee331a2b58ed6bb6ba55406f515e6 to not crash.

Then I noticed issues with directory listings, where the 150 reply 'closed' the tx, but the ```226 Directory send OK.``` was on it's own, and delayed. I tried to address this with https://github.com/OISF/suricata/commit/f6622bb972b00503dc837373664631a1caa2516e but I'm not convinced this was the correct approach.

This PR also set the correct author on the commit by @lmzke


jlucovsky:
Leaving the transaction open with status code 150 ("About to start data connection") seems ok since "150" is a marker that an event (data transfer) is beginning. I haven't measured the reporting latencies but the approach seems correct, given the meaning of "150"

victorjulien:
@jasonish any thoughts on how to deal with this case? We have a tx with command, and 2 replies: one preliminary response (150) and then some time later a final response. Logically they are part of the same transaction, but the timing will be off. The 150 won't be logged until we have the final response as well.

jasonish:
> @jasonish any thoughts on how to deal with this case? We have a tx with command, and 2 replies: one preliminary response (150) and then some time later a final response. Logically they are part of the same transaction, but the timing will be off. The 150 won't be logged until we have the final response as well.

Not really.  Can we make an array of responses? I guess the worst case is a file download, so it may take a few minutes (or hours) to complete, and only then it will be logged, correct?  Not ideal for realtime awareness, but will be much like an ftp server log.

The other way would be to log before the final replay, but then you'd need some way to associate the response transaction (which would be a new transaction) with the original transaction.

lmzke:
&lt;https://www.rfc-editor.org/rfc/rfc959.txt&gt; page 49, section `Command-Reply Sequences`, notes that some command sequences are tracked by state (as was already known/observed here).

It also shows other sequences to keep in mind.  e.g., if `150` were to be handled as its own case - unclear - then `125` can also precede `226`.

jlucovsky:
> <https://www.rfc-editor.org/rfc/rfc959.txt> page 49, section `Command-Reply Sequences`, notes that some command sequences are tracked by state (as was already known/observed here).
> 
> It also shows other sequences to keep in mind. e.g., if `150` were to be handled as its own case - unclear - then `125` can also precede `226`.

Thanks for the reference. RFC959 states:
```
1yz   Positive Preliminary reply

               The requested action is being initiated; expect another
               reply before proceeding with a new command.  (The
               user-process sending another command before the
               completion reply would be in violation of protocol; but
               server-FTP processes should queue any commands that
               arrive while a preceding command is in progress.)  This
               type of reply can be used to indicate that the command
               was accepted and the user-process may now pay attention
               to the data connections, for implementations where
               simultaneous monitoring is difficult.  The server-FTP
               process may send at most, one 1yz reply per command.
```

The implication is that any reply code beginning with 1 could be treated as a precursor to a secondary reply.

We could 
- Group 1xy completion codes with the secondary reply and increase the reporting latency of the 1xy code
- Group 1xy with the secondary completion codes thereby increasing the reporting latency for both.

When reported separately, there may be a large delay (based on file size(s)); the first option sends an indicator that there's more quickly.

My preference is for the first option -- do not report 1xy with the secondary completion codes.

jlucovsky:
Some responses may be grouped together and reported as one response, rather than separate responses, e.g. as indicated by the following response to a `USER` command:

```
  "ftp": {
    "command": "USER",
    "command_data": "cp@gilkeyphotography.com",
    "reply": [
      "User cp@gilkeyphotography.com OK. Password required\r\n230-Your bandwidth usage is restricted\r\n230-OK. Current restricted directory is /\r\n230 7168000 Kbytes used (100%) - authorized: 7168000 Kb"
    ],
    "completion_code": [
      "331"
    ],
    "reply_received": "yes"
  }
```

I think each `\r\n`-terminated line should be listed separately within the area.

Thoughts?

victorjulien:
@jlucovsky agreed, it would be best to get the codes in an array and the messages too.

Lets go with grouping them into single tx where we log when the final code is received.

jlucovsky:
Continued in #3962

victorjulien:
Replaced by #3962

PR #3960, opened 2019-06-20
configure: Fix failure with --enable-rust-debug

shivan1b:
Build was failing with --enable-rust-debug option as the check in
configure was for enable-debug and not enable-rust-debuf because of
which everytime, rust suricata lib path was set to rust/target/release.
Fix it by using the correct variable for check.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3054

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2413,7 +2413,7 @@ fi
         exit 1
     fi
 
-    if test "x$enable_debug" = "xyes"; then
+    if test "x$enable_rust_debug" = "xyes"; then
       RUST_SURICATA_LIB="../rust/target/debug/${RUST_SURICATA_LIBNAME}"
     else
       RUST_SURICATA_LIB="../rust/target/release/${RUST_SURICATA_LIBNAME}"


configure.ac
```diff
@@ -2413,7 +2413,7 @@ fi
         exit 1
     fi
 
-    if test "x$enable_debug" = "xyes"; then
+    if test "x$enable_rust_debug" = "xyes"; then
```
victorjulien: maybe this should check for either of those 2 vars, because we want the debug build for rust also if rust-debug isn't enabled. In the old scheme, --enable-debug will make sure symbols are preserved in the rust binary iirc. Do I remember correctly @jasonish ?

jasonish: We keep debug symbols in the release binary as well due to having this in the Cargo.toml:
```
[profile.release]
debug = true
```
So this just enabled/disables optimized builds.

Do we need a specific flag for Rust debug builds anymore?

victorjulien: So the only diff would be that it would lead to non-optimized rust binary?

jasonish: > So the only diff would be that it would lead to non-optimized rust binary?

Correct. That is my understanding after revisiting these options.

shivan1b: Are we OK with non-optimized rust binary then? Should I remove `enable_rust_debug` altogether?

jasonish: I'm OK with removing `enable-rust-debug` if the following behaviour is respected:
- A default ./configure will add the --release flag to cargo. This results in an optimized build, but keeps debug symbols.
- `--enable-debug` will remove the --release flag to cargo resulting in a non-optimized debug build.

This does remove some control of Rust vs C build options, but I think its much simpler to reason about.

Thoughts @victorjulien ?

shivan1b: @victorjulien how do you feel about this?

victorjulien: I'm good with Jasons suggestion.

shivan1b: @victorjulien is this already done or not required?

victorjulien: I don't understand what your asking. The idea is to implement Jasons idea.

PR #3969, opened 2019-06-24
detect: validate dsize and distance values

jmtaylor90:
I am not sure this is the right place in the code to fix this but this does work in my testing and sample bad rule. Feedback/pointers more than welcome and thanks in advance!

https://redmine.openinfosecfoundation.org/issues/2982

Describe changes:
- This change adds logic to fail rules that have a dsize value that matches a distance value.

### DIFF
--- a/src/detect-content.c
+++ b/src/detect-content.c
@@ -379,7 +379,6 @@ _Bool DetectContentPMATCHValidateCallback(const Signature *s)
     if (!(s->flags & SIG_FLAG_DSIZE)) {
         return TRUE;
     }
-
     int max_right_edge_i = SigParseGetMaxDsize(s);
     if (max_right_edge_i < 0) {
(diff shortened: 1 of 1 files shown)


src/detect-content.c
```diff
@@ -405,6 +406,12 @@ _Bool DetectContentPMATCHValidateCallback(const Signature *s)
                     cd->content_len, cd->offset, right_edge, max_right_edge);
             return FALSE;
         }
+	if (max_right_edge_i == distance_size) {
```
victorjulien: This seems to exact, can you check https://github.com/OISF/suricata-verify/pull/80/files#r298919007?

I think testing distance directly isn't correct as it's distance after a prior pattern match, of which the length should also be considered.

jmtaylor90:
reworking

PR #3981, opened 2019-06-29
log: Add module and subsystem identifiers to log

jlucovsky:
This changeset provides subsystem and module identifiers in the log when
the log format string contains `"%S"`. By convention, the log format
surrounds `"%S"` with brackets.

The subsystem name is generally the same as the thread name. The module
name is derived from the source code module name and usually consists of
the first one or 2 segments of the name using the dash character as the
segment delimiter.

Issue 2497: [redmine](https://redmine.openinfosecfoundation.org/issues/2497)

Continued from #3980 with
1. Remove extension from Rust source modules

Describe changes:

Module identifiers are derived from the source code module emitting the log message. A new CPP define `__SCFILENAME__` is assigned to `_sc_module` (`util-debug.h`) at compile time. Rust source module names are determined dynamically during the calls to the log function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -263,3 +263,9 @@ before_install:
         sudo pip install PyYAML
     fi
   - ./qa/travis-libhtp.sh
+
+after_failure:
+    - echo == Start of after failure logs ==
+    - find . -name config.log | xargs cat
+    - sleep 10
+    - echo == End of after failure logs ==
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: xenial
 sudo: false
 language: c
 
--- a/configure.ac
+++ b/configure.ac
@@ -2473,6 +2473,10 @@ fi
         fi
     fi
 
+
+# Add diagnostic filename
+CPPFLAGS="${CPPFLAGS} -D__SCFILENAME__=\\\"\$(*F)\\\""
+
 AC_SUBST(CFLAGS)
 AC_SUBST(LDFLAGS)
 AC_SUBST(CPPFLAGS)
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -67,6 +67,7 @@ pub type SCLogMessageFunc =
                   filename: *const std::os::raw::c_char,
                   line: std::os::raw::c_uint,
                   function: *const std::os::raw::c_char,
+                  subsystem: *const std::os::raw::c_char,
                   code: std::os::raw::c_int,
                   message: *const std::os::raw::c_char) -> std::os::raw::c_int;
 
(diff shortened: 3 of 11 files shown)


victorjulien:
Example output:
```
SC_LOG_FORMAT="%t [%S] - <%d> - " ./src/suricata -c suricata.yaml -l tmp/ -r ~/Desktop/tftp-rrq-shivani.pcapng --disable-detection --runmode=single -v
1/7/2019 -- 13:01:49 [Suricata-main:suricata] - <Notice> - This is Suricata version 5.0.0-dev running in USER mode
1/7/2019 -- 13:01:49 [Suricata-main:util-cpu] - <Info> - CPUs/cores online: 12
1/7/2019 -- 13:01:49 [Suricata-main:util-coredump] - <Info> - Could not set core dump size to unlimited; it's set to the hard limit.
1/7/2019 -- 13:01:49 [Suricata-main:util-logopenfile] - <Info> - fast output device (regular) initialized: fast.log
1/7/2019 -- 13:01:49 [Suricata-main:util-logopenfile] - <Info> - eve-log output device (regular) initialized: eve.json
1/7/2019 -- 13:01:49 [Suricata-main:util-logopenfile] - <Info> - stats output device (regular) initialized: stats.log
1/7/2019 -- 13:01:49 [Suricata-main:tm-threads] - <Notice> - all 1 packet processing threads, 4 management threads initialized, engine started.
1/7/2019 -- 13:01:49 [W#01:source-pcap] - <Info> - Starting file run for /home/victor/Desktop/tftp-rrq-shivani.pcapng
1/7/2019 -- 13:01:49 [W#01:source-pcap] - <Info> - pcap file /home/victor/Desktop/tftp-rrq-shivani.pcapng end of file reached (pcap err code 0)
1/7/2019 -- 13:01:49 [Suricata-main:suricata] - <Notice> - Signal Received.  Stopping engine.
1/7/2019 -- 13:01:49 [Suricata-main:suricata] - <Info> - time elapsed 0.052s
1/7/2019 -- 13:01:49 [W#01:source-pcap] - <Notice> - Pcap-file module read 1 files, 3 packets, 230 bytes
1/7/2019 -- 13:01:49 [counters] - <Info> - Alerts: 0
```
This seems to be working well from the technical side. So I think now we should think about what the best output would be. In the above example I think I'd like to see the 'util-', 'source-' and 'tm-' prefixes stripped. Would there be an easy way to do this in the transform func?

jlucovsky:
Yes ... The transform function receives a filename (sans extension) and returns the transformation. Currently, it returns *up to* the first 2 segments of a filename; a "segment" is defined as a `-` delimited string:
- `detect-engine` -> `detect-engine`
- `detect-http-content-len` -> `detect-http`
- `detect` -> `detect`
- `detect-transform-md5` -> `detect-transform`

The transform function can be redefined to
- If there are multiple "segments", skip past the first segment
- Find up to the next 2 segments

With this:
- `detect-engine` -> `engine`
- `detect-http-content-len` -> `http-content`
- `detect` -> `detect`
- `detect-transform-md5` -> `transform-md5`

Is this what you'd like?

jlucovsky:
Continued in #4006

PR #3983, opened 2019-07-01
Tftp backport/v1

shivan1b:
- https://redmine.openinfosecfoundation.org/issues/3044
- https://redmine.openinfosecfoundation.org/issues/3008

### DIFF
--- a/rust/src/dhcp/dhcp.rs
+++ b/rust/src/dhcp/dhcp.rs
@@ -220,8 +220,9 @@ impl DHCPState {
 
 #[no_mangle]
 pub extern "C" fn rs_dhcp_probing_parser(_flow: *const Flow,
-                                         input: *const libc::uint8_t,
-                                         input_len: u32) -> AppProto {
+                                         input: *const u8,
+                                         input_len: u32) -> AppProto
+{
     if input_len < DHCP_MIN_FRAME_LEN {
         return ALPROTO_UNKNOWN;
     }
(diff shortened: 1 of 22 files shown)


victorjulien:
https://ci.appveyor.com/project/OISF/suricata/builds/25658705

Rust binding included for non-Rust build. Probably a side effect of git master no longer requiring ```#ifdef RUST``` guards, but master-4.1.x still does.

rust/src/dhcp/dhcp.rs
```diff
@@ -239,21 +240,21 @@ pub extern "C" fn rs_dhcp_probing_parser(_flow: *const Flow,
 
 #[no_mangle]
 pub extern "C" fn rs_dhcp_tx_get_alstate_progress(_tx: *mut libc::c_void,
-                                                  _direction: libc::uint8_t) -> libc::c_int {
+                                                  _direction: u8) -> libc::c_int {
```
jlucovsky: nit: The `{` should be on the same line or the next line consistently (IMO). This change is inconsistent with that on line 225.

rust/src/tftp/tftp.rs
```diff
@@ -31,11 +31,28 @@ pub struct TFTPTransaction {
...
     pub mode : String,
-    pub logged : LoggerFlags
+    pub logged : LoggerFlags,
+    id: u64,
 }
 
 pub struct TFTPState {
-    pub transactions : Vec<TFTPTransaction>
+    pub transactions : Vec<TFTPTransaction>,
+    /// tx counter for assigning incrementing id's to tx's
```
jlucovsky: nit: use 2 `/` characters

src/app-layer-ikev2.c
```diff
@@ -34,33 +34,9 @@
 
 #include "app-layer-ikev2.h"
 
-#ifdef HAVE_RUST
```
jlucovsky: Is Rust mandatory in 4.1.4?

src/app-layer-tftp.c
```diff
@@ -36,14 +36,6 @@
 
 #include "app-layer-tftp.h"
 
-#ifndef HAVE_RUST
```
jlucovsky: Rust mandatory?

shivan1b:
@jlucovsky there are no compiler warnings about using `libc::c_void` but yes using `std::os::raw::c_void` seems like a better choice for our usecases. Since this is just a backport of Victor's commits from master, I can make the change of using `std::os::raw::c_void` as a separate cleanup patch if @victorjulien thinks its OK too.

shivan1b:
Please check the new PR with changes requested here: #3986

rust/src/tftp/tftp.rs
victorjulien: In Rust ```///``` has a special meaning for documenting, so it may be correct here.

PR #3984, opened 2019-07-01
detect: Add missing keyword URLs and description

shivan1b:
Add missing keyword URLs and their description. Fix the ones that
were incorrect.

Partially closes redmine ticket #2974.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2974


norg:
I just have small style ideas:
- there is a mix with "Matches" vs. "matches" using the capital sometimes (would go with the latter one)
- sometimes we end with a dot sometimes we don't :)
- "matches" vs. "match"
Nothing big but could be improved, but I bet there will be future PRs adding new keywords doing the same :)

src/detect-krb5-errcode.c
```diff
@@ -65,7 +65,7 @@ static int g_krb5_err_code_list_id = 0;
 void DetectKrb5ErrCodeRegister(void) {
     sigmatch_table[DETECT_AL_KRB5_ERRCODE].name = "krb5_err_code";
     sigmatch_table[DETECT_AL_KRB5_ERRCODE].desc = "match Kerberos 5 message type";
```
jlucovsky: s/message type/error code/

src/detect-modbus.c
```diff
@@ -517,6 +517,8 @@ void DetectModbusRegister(void)
 {
     SCEnter();
     sigmatch_table[DETECT_AL_MODBUS].name          = "modbus";
+    sigmatch_table[DETECT_AL_MODBUS].desc          = "used for matching on various properties of Modbus requests.";
+    sigmatch_table[DETECT_AL_MODBUS].url          = DOC_URL DOC_VERSION "/rules/modbus-keyword.html#modbus-keyword";
```
jlucovsky: nit: add space before `=`

shivan1b:
Replaced with #4148

PR #3988, opened 2019-07-03
ftp: removes one use of atoi

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3053

Describe changes:
- Removes one use of `atoi` and replaces it with `ByteExtractStringUint16` for better error handling

If this looks good, I can remove more uses of `atoi`


jlucovsky:
The change looks good but the last parameter should be a `const char  *` instead of `char *`.

catenacyber:
Thank you Jeff

PR #3990, opened 2019-07-03
leak: Fixes leak in AppLayerProtoDetectPMRegisterPattern

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3070

Describe changes:
- Fixes a leak in `AppLayerProtoDetectPMRegisterPattern` in case of an error

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -1478,6 +1478,7 @@ static int AppLayerProtoDetectPMRegisterPattern(uint8_t ipproto, AppProto alprot
 
     goto end;
  error:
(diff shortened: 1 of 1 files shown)


src/app-layer-detect-proto.c
```diff
@@ -1478,6 +1478,7 @@ static int AppLayerProtoDetectPMRegisterPattern(uint8_t ipproto, AppProto alprot
 
     goto end;
  error:
+    SCFree(cd);
```
victorjulien: DetectContentFree is meant to be used with this data type

PR #3992, opened 2019-07-03
Tftp vxlan backport/v2

shivan1b:
- https://redmine.openinfosecfoundation.org/issues/3044
- https://redmine.openinfosecfoundation.org/issues/3008
- https://redmine.openinfosecfoundation.org/issues/3068

### DIFF
--- a/src/decode-vxlan.c
+++ b/src/decode-vxlan.c
@@ -1,0 +1,180 @@
+/* Copyright (C) 2014 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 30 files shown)


src/decode-vxlan.c
```diff
@@ -0,0 +1,276 @@
...
+
+#include "pkt-var.h"
+#include "util-profiling.h"
+#include "host.h"
+
+#define VXLAN_HEADER_LEN        8
+#define VXLAN_DEFAULT_PORT      4789
+#define VXLAN_DEFAULT_PORT_S    "4789"
+
+static bool g_vxlan_enabled = true;
```
victorjulien: I wonder if we should disable by default, as this is a backport to stable. Thoughts @jasonish ?

jryberg:
I'm looking forward for this pull request =)

src/decode-vxlan.c
jasonish: I'm in favour of having it disabled by default in 4.1.[](url)

victorjulien: @shivan1b can you make sure it is disabled by default? That means this global should be set to false, and the suricata.yaml.in setting should also be updated. It will there also require an update to the suricata-verify tests (add --set decoder.vxlan.enabled=true to the commandline I think)

shivan1b:
Replaced with #4003

PR #3995, opened 2019-07-03
log: use SCLogError instead of fprintf

catenacyber:
None

Describe changes:
- Replaces one use of `fprintf` by `SCLogError`

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -1078,7 +1078,7 @@ static int SigParseBasics(DetectEngineCtx *de_ctx,
 
     /* Options. */
     if (index == NULL) {
-        fprintf(stderr, "no rule options.\n");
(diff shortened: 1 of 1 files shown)


src/detect-parse.c
```diff
@@ -1078,7 +1078,7 @@ static int SigParseBasics(DetectEngineCtx *de_ctx,
 
     /* Options. */
     if (index == NULL) {
-        fprintf(stderr, "no rule options.\n");
+        SCLogError(SC_ERR_INVALID_RULE_ARGUMENT, "no rule options.\n");
```
jlucovsky: nit: `\n` isn't needed when using `SCLogError`

catenacyber: Thanks Jeff

PR #4002, opened 2019-07-05
http: fixes overflow in range parsing

catenacyber:
None

Describe changes:
- Fixes off-by-one overflow in HTTP range parsing

For instance `if (len < pos + 1 || data[pos+1] != '/')`
Meant that when 
- `len == pos+1`
- first sub condition was false
- so we tried the seance one
- so we accessed `data[pos+1] == data[len]` which is one byte overflow

### DIFF
--- a/src/app-layer-htp-file.c
+++ b/src/app-layer-htp-file.c
@@ -186,7 +186,7 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
 
     if (data[pos] == '*') {
         // case with size only
-        if (len < pos + 1 || data[pos+1] != '/') {
+        if (len < pos + 2 || data[pos+1] != '/') {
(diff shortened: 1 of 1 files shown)


src/app-layer-htp-file.c
```diff
@@ -186,7 +186,7 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
 
     if (data[pos] == '*') {
         // case with size only
-        if (len < pos + 1 || data[pos+1] != '/') {
+        if (len < pos + 2 || data[pos+1] != '/') {
```
jlucovsky: Suggest `<=`  with `pos + 1` instead (here, and at other change).

catenacyber: ok

PR #4005, opened 2019-07-05
Always store the vlan id of packets

mfil:
in the suricata IRC.

request.

Describe changes:

### DIFF
--- a/src/flow-hash.c
+++ b/src/flow-hash.c
@@ -135,8 +135,9 @@ static inline uint32_t FlowGetHash(const Packet *p)
 
             fhk.proto = (uint16_t)p->proto;
             fhk.recur = (uint16_t)p->recursion_level;
-            fhk.vlan_id[0] = p->vlan_id[0];
-            fhk.vlan_id[1] = p->vlan_id[1];
+            /* g_vlan_mask zeros the vlan_id if we don't use it for tracking. */
+            fhk.vlan_id[0] = p->vlan_id[0] & g_vlan_mask;
+            fhk.vlan_id[1] = p->vlan_id[1] & g_vlan_mask;
 
             hash = hashword(fhk.u32, 5, flow_config.hash_rand);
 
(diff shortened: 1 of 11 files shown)


src/flow-hash.c
```diff
@@ -266,38 +267,92 @@ uint32_t FlowKeyGetHash(FlowKey *fk)
...
-        fhk.vlan_id[1] = fk->vlan_id[1];
+        fhk.vlan_id[0] = fk->vlan_id[0] & g_vlan_mask;
+        fhk.vlan_id[1] = fk->vlan_id[1] & g_vlan_mask;
 
         hash = hashword(fhk.u32, 11, flow_config.hash_rand);
     }
 return hash;
 }
 
+static inline bool cmp_addrs(const uint32_t addr1[4], const uint32_t addr2[4])
```
victorjulien: Our code style is that function names AreWrittenLikeThis. See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Coding_Style

src/source-pfring.c
```diff
@@ -532,9 +531,8 @@ TmEcode ReceivePfringThreadInit(ThreadVars *tv, const void *initdata, void **dat
 
     opflag = PF_RING_PROMISC;
 
-    /* if suri uses VLAN and if we have a recent kernel, we need
-     * to use parsed_pkt to get VLAN info */
-    if ((! ptv->vlan_disabled) && SCKernelVersionIsAtLeast(3, 0)) {
+    /* if we have a recent kernel, we need to use parsed_pkt to get VLAN info */
```
victorjulien: this should be in its own commit

src/flow-hash.c
mfil: I can't remove the vlanh pointers without touching pfring and af-packet because they set those pointers to NULL, but I can try to make sure that that's the only change to those files.

mfil:
> I think there are several changes that should go into different commits:
> 
>     1. flow cmp macro to inline func changes
> 
>     2. vlan handling, incl removal of ptrs from Packet
> 
>     3. pfring update
> 
> 
> This is a bit tricky to untangle, but esp for an update to the flow engine this is very important. Could you try to untangle these?

I can't completely avoid touching pfring and af-packet when removing the vlanh pointer because they set them to NULL. I can try to make sure that that's the only change.

src/source-pfring.c
mfil: Do you mean a fourth commit in addition to the list above?

victorjulien: This is what I meant when referring pfring. The removal of ```SCKernelVersionIsAtLeast(3, 0)) {``` seems unrelated to the rest of the changes, thus must go in it's own commit

victorjulien:
Yeah the vlanh ptr updates in af-packet and pfring ring are logically part of the vlan work, so they can be in the same commit. The reimplementation of the flow cmp code is logically a different thing, so that should be separated and documented in the git commit messages.

PR #4012, opened 2019-07-09
Always store vlan id

mfil:
in the suricata IRC.

request.

Describe changes:

Updates https://github.com/OISF/suricata/pull/4005
- Split up into smaller commits
- Use the correct function naming convention

### DIFF
--- a/src/source-pfring.c
+++ b/src/source-pfring.c
@@ -138,7 +138,7 @@ struct PfringThreadVars_
     ThreadVars *tv;
     TmSlot *slot;
 
-    int vlan_disabled;
+    int vlan_in_ext_header;
 
     /* threads count */
     int threads;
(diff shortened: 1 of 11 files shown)


src/source-pfring.c
```diff
@@ -256,13 +256,12 @@ static inline void PfringProcessPacket(void *user, struct pfring_pkthdr *h, Pack
      * So if it is not set, use the parsed info from PF_RING's
      * extended header.
      */
-    if ((!ptv->vlan_disabled) &&
+    if (vlan_in_ext_header &&
```
victorjulien: should this be ```ptv->vlan_in_ext_header``` instead?
```
  CC       source-pfring.o
source-pfring.c: In function 'PfringProcessPacket':
source-pfring.c:259:9: error: 'vlan_in_ext_header' undeclared (first use in this function); did you mean 'long_pkt_header'?
     if (vlan_in_ext_header &&
         ^~~~~~~~~~~~~~~~~~
         long_pkt_header
source-pfring.c:259:9: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [source-pfring.o] Error 1
Makefile:1769: recipe for target 'source-pfring.o'
```

mfil: Yes. I don't know why it compiled on my machine, or on Appveyor.

victorjulien: Yeah none of the public CI tools have it. I run a private CI that does have it.

PR #4014, opened 2019-07-09
Vxlan backport/v5

shivan1b:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3068

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -77,6 +77,7 @@ decode-tcp.c decode-tcp.h \
 decode-teredo.c decode-teredo.h \
 decode-udp.c decode-udp.h \
 decode-vlan.c decode-vlan.h \
+decode-vxlan.c decode-vxlan.h \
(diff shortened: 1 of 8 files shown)


victorjulien:
The 2nd unittest will need the same logic. It passes now as well, but as a side effect.

shivan1b:
Side effect is that the control never reaches to `DecodeVXLAN` func where the packet enqueing is handled because of `false` value returned by `DecodeVXLANEnabledForPort`.
In an ideal scenario, the `false` value should have been returned by `DecodeVXLANEnabledForPort`   because of the global setting but because port `1` does not have VXLAN enabled.

Is this correct, @victorjulien ?

shivan1b:
Replaced with #4018

victorjulien:
Yes exactly :)

PR #4016, opened 2019-07-10
Always store vlan id

mfil:
in the suricata IRC.

request.

Describe changes:

Updates https://github.com/OISF/suricata/pull/4005
- Split up into smaller commits
- Use the correct function naming convention

Updates https://github.com/OISF/suricata/pull/4012
- Fix a bug in my previous pull request
- Rebase

### DIFF
--- a/src/decode-erspan.c
+++ b/src/decode-erspan.c
@@ -64,7 +64,7 @@ int DecodeERSPAN(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p, uint8_t *pkt,
         return TM_ECODE_FAILED;
     }
 
-    if (vlan_id > 0 && dtv->vlan_disabled == 0) {
+    if (vlan_id > 0) {
         if (p->vlan_idx >= 2) {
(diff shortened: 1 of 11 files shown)


mfil:
In source-pfring.c, a comment is out of sync with the code it describes:
```
    /* If kernel is older than 3.8, VLAN is not stripped so we don't
     * get the info from packt extended header but we will use a standard
     * parsing */
    if (! SCKernelVersionIsAtLeast(3, 0)) {
        ptv->vlan_disabled = 1;
    }
```
It's kernel version 3.8 in the comment and 3.0 in the code. I don't know which one is correct, so I didn't touch either. Probably something for a separate commit?

victorjulien:
Looks like it's been this way since @regit introduced it in 1fa4233d67d068fd49155e9d153011a491125833. Guess we let the code be leading, so use 3.0.

But indeed a separate commit would be in order.

mfil:
Should I make a v5 branch, or should we do this completely separate?

victorjulien:
Lets keep it separate, unless we need a v5 for other reasons. Then I'm fine with mixing it in :)

victorjulien:
Thanks a lot Max!

PR #4020, opened 2019-07-10
ssl: register probing for port 443 if no config

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3070

Describe changes:
- Register TLS probing parser for port 443 in case of configuration missing

If `AppLayerProtoDetectPPParseConfPorts` fails, it means there is no configuration for TLS.

See SMB for another example

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -2853,10 +2853,20 @@ void RegisterSSLParsers(void)
                                           STREAM_TOSERVER,
                                           SSLProbingParser, NULL);
         } else {
-            AppLayerProtoDetectPPParseConfPorts("tcp", IPPROTO_TCP,
-                                                proto_name, ALPROTO_TLS,
-                                                0, 3,
-                                                SSLProbingParser, NULL);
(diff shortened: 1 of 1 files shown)


src/app-layer-ssl.c
```diff
@@ -2853,10 +2853,20 @@ void RegisterSSLParsers(void)
...
         } else {
-            AppLayerProtoDetectPPParseConfPorts("tcp", IPPROTO_TCP,
-                                                proto_name, ALPROTO_TLS,
-                                                0, 3,
-                                                SSLProbingParser, NULL);
+            if (AppLayerProtoDetectPPParseConfPorts("tcp", IPPROTO_TCP,
+                                                    proto_name, ALPROTO_TLS,
+                                                    0, 3,
+                                                    SSLProbingParser, NULL) == 0) {
+                SCLogWarning(SC_ERR_SMB_CONFIG, "no TLS config found, "
```
victorjulien: ```SC_ERR_SMB_CONFIG``` is wrong here. I think we can use (and introduce if needed) a more generic code.

catenacyber: Thanks, changing it

PR #4023, opened 2019-07-11
detect: validate dsize and distance values

jmtaylor90:

https://redmine.openinfosecfoundation.org/issues/2982

Describe changes:
- this update from v1 adds a content length counter and attempts to calculate what valid distance values could be
-
-

### DIFF
--- a/src/detect-content.c
+++ b/src/detect-content.c
@@ -379,13 +379,15 @@ _Bool DetectContentPMATCHValidateCallback(const Signature *s)
     if (!(s->flags & SIG_FLAG_DSIZE)) {
         return TRUE;
     }
-
     int max_right_edge_i = SigParseGetMaxDsize(s);
     if (max_right_edge_i < 0) {
         return TRUE;
     }
 
     uint32_t max_right_edge = (uint32_t)max_right_edge_i;
+    int content_total_length = 0;
+    int previous_content_length = 0;
+    int content_temp_length = 0;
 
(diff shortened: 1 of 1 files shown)


src/detect-content.c
```diff
@@ -379,20 +379,29 @@ _Bool DetectContentPMATCHValidateCallback(const Signature *s)
...
         return TRUE;
     }
-
     int max_right_edge_i = SigParseGetMaxDsize(s);
     if (max_right_edge_i < 0) {
         return TRUE;
     }
 
     uint32_t max_right_edge = (uint32_t)max_right_edge_i;
+    int content_total_length = 0;
```
jlucovsky: nit: with the way this is implemented, neither `content_total_length` nor `previous_content_length` require initialization values.

jmtaylor90: Okay, I will update in the next version. Thanks Jeff! For my own information, is it just style or functionally bad?

jlucovsky: It's not harmful but initialization in this case is not necessary either. So, it's a (minor) optimization to avoid doing unnecessary work. But, it is minor. Best to include only what's strictly necessary, IMO.

jmtaylor90: Cool, thanks!

victorjulien: I think ```content_total_length``` can be declared on first use, which is within the for loop below.

victorjulien: Looks like this will neatly apply to all three vars?

jmtaylor90:
continued in #4064

PR #4024, opened 2019-07-11
http: fixes overflow in range parsing

catenacyber:
None

Describe changes:
- Fixes off-by-one overflow in HTTP range parsing

For instance `if (len < pos + 1 || data[pos+1] != '/')`
Meant that when 
- `len == pos+1`
- first sub condition was false
- so we tried the seance one
- so we accessed `data[pos+1] == data[len]` which is one byte overflow

Continues #4002 with review taken into account


victorjulien:
Merged in #4028, thanks!

PR #4035, opened 2019-07-14
detect/transform: add tld and domain transforms

jlucovsky:
This PR adds transforms for:
1. tld - Returns the top-level-domain and passes it on
1. domain - Returns the domain and passes it on.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #4036, opened 2019-07-14
detect/transform: add tld and domain transforms

jlucovsky:
This PR adds transforms for:
1. tld - Returns the top-level-domain and passes it on

Continuation fo #4035 
This update:
1. Addresses unittest failures.

Companion PR to suricata-verify` repo; https://github.com/OISF/suricata-verify/pull/97

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:  [3074](https://redmine.openinfosecfoundation.org/projects/suricata/issues/3074)

### DIFF
--- a/src/detect-transform-domain.c
+++ b/src/detect-transform-domain.c
@@ -1,0 +1,226 @@
+/* Copyright (C) 2019 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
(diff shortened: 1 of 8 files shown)


src/detect-transform-domain.c
```diff
@@ -0,0 +1,226 @@
...
+    InspectionBufferFree(&buffer);
+    PASS;
+}
+
+static int DetectTransformDomainTest05(void)
+{
+    const uint8_t *input = (const uint8_t *)"windows.update.microsoft.com";
+    uint32_t input_len = strlen((char *)input);
+
+    const char *result = "update.microsoft.com";
```
victorjulien: I would have expected 'microsoft.com' here instead of including the subdomain.

src/detect-transform-domain.c
```diff
@@ -0,0 +1,226 @@
...
+static void TransformDomain(InspectionBuffer *buffer)
+{
+    const uint8_t *input = buffer->inspect;
+    const uint32_t input_len = buffer->inspect_len;
+    uint8_t output[input_len]; // we can only shrink
+    uint8_t *os = output;
+
+    SCLogDebug("Transforming %.*s", input_len, input);
+
+    uint8_t *dot = memchr(input, '.', input_len);
```
victorjulien: should this be memrchr to look from right to left? In my idea 'domain' would not include any subdomains.

src/detect-transform-domain.c
jlucovsky: I'll modify this transform to produce the second-level-domain instead.

jlucovsky:
Continued in #4038

PR #4042, opened 2019-07-17
RFC: feature/datasets/v40

victorjulien:
Draft of datasets work. Not as far yet as I'd hoped.

```
Datasets are sets/lists of data that can be accessed or added from
the rule language.

This patch implements 3 data types:

1. string (or buffer)
2. md5
3. sha256

The patch also implements 2 new rule keywords:

1. dataset
2. datarep

The dataset keyword allows matching against a list of values to see if
it exists or not. It can also add the value to the set. The set can
optionally be stored to disk on exit.

The datarep support matching/lookups only. With each item in the set a
reputation value is stored and this value can be matched against. The
reputation value is unsigned 16 bit, so values can be between 0 and 65535.

Datasets can be registered in 2 ways:

1. through the yaml
2. through the rules

The goal of this rules based approach is that rule writers can start using
this without the need for config changes.

A dataset is implemented using a thash hash table. Each dataset is its own
separate thash.
```

**TODOs**
- unix commands are incomplete: only dataset-add was created
- the unix dataset-add command for the string data type currently requires base64 input. It would be good if suricatasc would do this for the user
- sets are not reloaded/updated on rule reload. New sets are loaded though.
- whether a dataset is pure data or data+rep is currently being determined by the file contents. This is not great.

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/278
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/281


victorjulien:
See https://github.com/OISF/suricata/pull/4042/commits/83a8ab588490c367ea2b808f55e8d64e18165397?short_path=1002361#diff-1002361f823c745161c26ac3888c3879 for a bit more documentation and intended use.

victorjulien:
Replaced by #4047

PR #4044, opened 2019-07-17
Protocol detection 2757 v1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2757

Describe changes:
- Use both directions for protocol detection over UDP (as is done with TCP)
- Adds a probing function for protocol `DNS` for direction to client (for UDP)
- Changes `AppLayerProtoDetectPPGetProto` so that first packet does not bit mask away a protocol for both directions when config option midstream is activated 

The main goal was the first one (use both directions for UDP)
I added DNS probing function to test this functionality
And then I came across this unwanted behavior from `AppLayerProtoDetectPPGetProto` :
When the first packet is no valid DNS but on port 53 (for instance a junk request), second packet (error response from server) does not get checked for DNS as first packet bit masked away DNS for both directions (ie `probing_parser_toclient_alproto_masks` and `probing_parser_toserver_alproto_masks`)

There remains more to be done for the redmine ticket
https://redmine.openinfosecfoundation.org/issues/2757
such as "Support when multiple protocols have patterns matching (such as USER for FTP and IRC)"
But this can already be merged independently I guess...

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #4048, opened 2019-07-17
Protocol detection 2757 v2

catenacyber:

Describe changes:
- Use both directions for protocol detection over UDP (as is done with TCP)
- Adds a probing function for protocol `DNS` for direction to client (for UDP)
- Changes `AppLayerProtoDetectPPGetProto` so that first packet does not bit mask away a protocol for both directions when config option midstream is activated 

The main goal was the first one (use both directions for UDP)
I added DNS probing function to test this functionality
And then I came across this unwanted behavior from `AppLayerProtoDetectPPGetProto` :
When the first packet is no valid DNS but on port 53 (for instance a junk request), second packet (error response from server) does not get checked for DNS as first packet bit masked away DNS for both directions (ie `probing_parser_toclient_alproto_masks` and `probing_parser_toserver_alproto_masks`)

There remains more to be done for the redmine ticket
such as "Support when multiple protocols have patterns matching (such as USER for FTP and IRC)"
But this can already be merged independently I guess...

Modifies #4044 setting `f->alproto` before calling `AppLayerParserParse` thus fixing suricata-verify failures

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -608,8 +608,9 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
 
     /* if we don't know the proto yet and we have received a stream
      * initializer message, we run proto detection.
-     * We receive 2 stream init msgs (one for each direction) but we
-     * only run the proto detection once. */
+     * We receive 2 stream init msgs (one for each direction), we
+     * only run the proto detection for both and emit an event
+     * in the case protocols mismatch. */
     if (alproto == ALPROTO_UNKNOWN && (flags & STREAM_START)) {
         /* run protocol detection */
         if (TCPProtoDetect(tv, ra_ctx, app_tctx, p, f, ssn, stream,
@@ -688,6 +689,8 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
 int AppLayerHandleUdp(ThreadVars *tv, AppLayerThreadCtx *tctx, Packet *p, Flow *f)
 {
     SCEnter();
+    AppProto *alproto;
+    AppProto *alproto_otherdir;
 
     if (f->alproto == ALPROTO_FAILED) {
         SCReturnInt(0);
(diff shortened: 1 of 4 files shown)


victorjulien:
Can you move more of the explanations from the PR to the commits themselves? The PR does not become part of the git history, only the commits do.

src/app-layer.c
```diff
@@ -697,26 +700,69 @@ int AppLayerHandleUdp(ThreadVars *tv, AppLayerThreadCtx *tctx, Packet *p, Flow *
...
+        *alproto = AppLayerProtoDetectGetProto(tctx->alpd_tctx,
                                   f, p->payload, p->payload_len,
                                   IPPROTO_UDP, flags, &reverse_flow);
         PACKET_PROFILING_APP_PD_END(tctx);
 
-        if (f->alproto != ALPROTO_UNKNOWN) {
-            AppLayerIncFlowCounter(tv, f);
+        switch (*alproto) {
+            case ALPROTO_UNKNOWN:
+            if (*alproto_otherdir != ALPROTO_UNKNOWN) {
```
victorjulien: style: indent is off

src/app-layer.c
```diff
@@ -697,26 +700,69 @@ int AppLayerHandleUdp(ThreadVars *tv, AppLayerThreadCtx *tctx, Packet *p, Flow *
...
+                // Use recognized side
+                f->alproto = *alproto_otherdir;
+                AppLayerIncFlowCounter(tv, f);
+                if (*alproto_otherdir == ALPROTO_FAILED) {
+                    SCLogDebug("ALPROTO_UNKNOWN flow %p", f);
+                }
+            } else {
+                // First side of protocol is unknown
```
victorjulien: what is 'first side' referring to here?

src/app-layer.c
```diff
@@ -697,26 +700,69 @@ int AppLayerHandleUdp(ThreadVars *tv, AppLayerThreadCtx *tctx, Packet *p, Flow *
...
+                if (*alproto_otherdir == ALPROTO_FAILED) {
+                    SCLogDebug("ALPROTO_UNKNOWN flow %p", f);
+                }
+            }
+            // else wait for second side of protocol
+            break;
+            default:
+            f->alproto = *alproto;
+            if (*alproto_otherdir == ALPROTO_FAILED) {
+                *alproto_otherdir = *alproto;
```
victorjulien: I think we can leave this as failed so its logged that way in the flow log.

src/app-layer.c
```diff
@@ -697,26 +700,69 @@ int AppLayerHandleUdp(ThreadVars *tv, AppLayerThreadCtx *tctx, Packet *p, Flow *
...
+                AppLayerIncFlowCounter(tv, f);
+            } else if (*alproto_otherdir != ALPROTO_UNKNOWN) {
+                if (*alproto_otherdir != *alproto) {
+                    AppLayerDecoderEventsSetEventRaw(&p->app_layer_events,
+                                                     APPLAYER_MISMATCH_PROTOCOL_BOTH_DIRECTIONS);
+                    // reset protocol detection
+                    *alproto_otherdir = ALPROTO_UNKNOWN;
```
victorjulien: I think this can be kept set to its original value for the same reason: logging the flow record.

src/app-layer.c
catenacyber: First direction (either to_client or to_server where a packet was seen for this flow)
How would you call it nicely ?

src/app-layer.c
catenacyber: ok

src/app-layer.c
catenacyber: Ok (I meant to retry cf suricata-verify where I send junk first)

src/app-layer.c
catenacyber: ok

catenacyber:
> So this one packet may be junk logic does seem a bit arbitrary. What if there are 2? Or what if we miss the server packet (pkt loss) making it appear to be 2?

Sorry, this was a wrong wording from me.
It is not about packets, it is about directions.
So there can be any number of junk packets from the client.
We will still check the protocol spoken by the server

PR #4050, opened 2019-07-17
Fuzz v11

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds six fuzz targets

Modifies #4049 with 
- fixed rebase with latest changes to FTP
- squashing commits

Some commits can be cherry-picked independently : log, config, style

### DIFF
--- a/src/conf-yaml-loader.c
+++ b/src/conf-yaml-loader.c
@@ -195,15 +195,14 @@ ConfYamlParse(yaml_parser_t *parser, ConfNode *parent, int inseq)
             yaml_version_directive_t *ver =
                 event.data.document_start.version_directive;
             if (ver == NULL) {
-                fprintf(stderr, "ERROR: Invalid configuration file.\n\n");
-                fprintf(stderr, "The configuration file must begin with the following two lines:\n\n");
-                fprintf(stderr, "%%YAML 1.1\n---\n\n");
+                SCLogError(SC_ERR_CONF_YAML_ERROR, "ERROR: Invalid configuration file.");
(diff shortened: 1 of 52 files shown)


victorjulien:
```style: Using const when parsing input```

I like this change, but it needs a better commit message. First, it's not a *style* change. Second, the message needs to explain a bit more. For example that app-layer API is never supposed to modify its input data and therefore its input pointers are marked as const.

```move PostConfLoadedSetup to init.c```
This should be squashed into into https://github.com/OISF/suricata/pull/4050/commits/d40792f6bf8ec151f84219b21d8e89d41c15a3bf

File: src/conf-yaml-loader.c [hunk 1]
```diff
@@ -195,15 +195,14 @@ ConfYamlParse(yaml_parser_t *parser, ConfNode *parent, int inseq)
             yaml_version_directive_t *ver =
                 event.data.document_start.version_directive;
             if (ver == NULL) {
-                fprintf(stderr, "ERROR: Invalid configuration file.\n\n");
-                fprintf(stderr, "The configuration file must begin with the following two lines:\n\n");
-                fprintf(stderr, "%%YAML 1.1\n---\n\n");
+                SCLogError(SC_ERR_CONF_YAML_ERROR, "ERROR: Invalid configuration file.");
+                SCLogError(SC_ERR_CONF_YAML_ERROR, "The configuration file must begin with the following two lines: %%YAML 1.1 and ---");
```
victorjulien: Line length limit exceeded

catenacyber:
> it's not a style change

I could not figure a better category. How would you call it ?

PR #4055, opened 2019-07-18
ssh.* documentation update

travisbgreen:

- [ x ] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ x ] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ x ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

https://redmine.openinfosecfoundation.org/issues/2377

Describe changes:
- Lightly updated documentation
- Fixed example syntax error
- Corrected --list-keywords output to reflect renamed buffers

### DIFF
--- a/doc/userguide/rules/ssh-keywords.rst
+++ b/doc/userguide/rules/ssh-keywords.rst
@@ -19,9 +19,9 @@ The example above matches on SSH connections with SSH version 2.
 
 ``ssh.proto`` can be used as ``fast_pattern``.
 
-``ssh.proto`` replaces the previous keyword name: ``ssh_proto``. You may continue
-to use the previous name, but it's recommended that rules be converted to use
-the new name.
+``ssh.proto`` replaces the previous keyword names: ``ssh_proto`` & 
+``ssh.protoversion``. You may continue to use the previous name, but 
+it's recommended that rules be converted to use the new name.
 
 ssh.software
 ------------
@@ -31,7 +31,7 @@ Match on the software string from the SSH banner.
 Example::
 
   alert ssh any any -> any any (msg:"match SSH software string"; \
-      ssh.software: content:"openssh"; nocase; sid:1000020;)
+      ssh.software; content:"openssh"; nocase; sid:1000020;)
 
 The example above matches on SSH connections where the software string contains "openssh".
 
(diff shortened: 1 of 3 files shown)


src/detect-ssh-software.c
```diff
@@ -50,8 +50,8 @@
 #define KEYWORD_NAME "ssh.software"
 #define KEYWORD_NAME_LEGACY "ssh_software"
 #define KEYWORD_DOC "ssh-keywords.html#ssh-software"
-#define BUFFER_NAME "ssh_software"
-#define BUFFER_DESC "ssh software"
+#define BUFFER_NAME "ssh.software"
```
victorjulien: I'm a bit worried that this might break some config options. Each 'buffer name' is used to create a config name:
```
detect.mpm.<buffer_name>.shared
```
See https://github.com/OISF/suricata/blob/master/src/detect-engine-mpm.c#L236

Config wise, the following 2 are not equal
```
detect.mpm.ssh_software.shared
detect.mpm.ssh.software.shared
```
The latter would look like:
```yaml
detect:
  mpm:
    ssh:
      software:
        shared: true
```
The former:
```yaml
detect:
  mpm:
    ssh_software:
      shared: true
```

These config options are not part of the suricata.yaml.in, so perhaps nobody is using them. Maybe it would still be good to have a fallback option here:
```c
        /* see if we use a unique or shared mpm ctx for this type */
        int confshared = 0;
        char confstring[256] = "detect.mpm.";
        strlcat(confstring, n->name, sizeof(confstring));
        strlcat(confstring, ".shared", sizeof(confstring));
        if (ConfGetBool(confstring, &confshared) == 1)
            shared = confshared;
        else if (ConfGetBool(confstringwithunderscore, &confshared) == 1)
            shared = confshared;
```
Of course the ```confstringwithunderscore``` variable would have to be crafted somehow.

doc/userguide/rules/ssh-keywords.rst
```diff
@@ -19,9 +19,9 @@ The example above matches on SSH connections with SSH version 2.
 
 ``ssh.proto`` can be used as ``fast_pattern``.
 
-``ssh.proto`` replaces the previous keyword name: ``ssh_proto``. You may continue
-to use the previous name, but it's recommended that rules be converted to use
-the new name.
+``ssh.proto`` replaces the previous keyword names: ``ssh_proto`` & 
+``ssh.protoversion``. You may continue to use the previous name, but
```
victorjulien: It might be good to distinguish between ```ssh_proto``` and ```ssh.protoversion``` as they have quite different mechanics. ```ssh_proto``` is just the old name for ```ssh.proto```, but operates the same way as sticky buffers. ```ssh.protoversion``` is quite different.

PR #4058, opened 2019-07-19
Protocol detection 2757 v3

catenacyber:

Describe changes:
- Use both directions for protocol detection over UDP (as is done with TCP)
- Adds a probing function for protocol `DNS` for direction to client (for UDP)
- Changes `AppLayerProtoDetectPPGetProto` so that first packet does not bit mask away a protocol for both directions when config option midstream is activated 

The main goal was the first one (use both directions for UDP)
I added DNS probing function to test this functionality
And then I came across this unwanted behavior from `AppLayerProtoDetectPPGetProto` :
When the first packet is no valid DNS but on port 53 (for instance a junk request), second packet (error response from server) does not get checked for DNS as first packet bit masked away DNS for both directions (ie `probing_parser_toclient_alproto_masks` and `probing_parser_toserver_alproto_masks`)

There remains more to be done for the redmine ticket
such as "Support when multiple protocols have patterns matching (such as USER for FTP and IRC)"
But this can already be merged independently I guess...

Modifies #4048 :
- rewording commits with more details
- better identation
- leaving protocol inconsistencies for logging

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -523,6 +523,10 @@ fn probe(input: &[u8]) -> bool {
     parser::dns_parse_request(input).is_ok()
 }
 
+fn probe_srv(input: &[u8]) -> bool {
+    parser::dns_parse_response(input).is_ok()
+}
+
 /// Probe TCP input to see if it looks like DNS.
 pub fn probe_tcp(input: &[u8]) -> bool {
     match nom::be_u16(input) {
@@ -820,6 +824,19 @@ pub extern "C" fn rs_dns_probe(input: *const u8, len: u32)
         std::slice::from_raw_parts(input as *mut u8, len as usize)
     };
     if probe(slice) {
+        return 1;
+    }
+    return 0;
+}
+
+#[no_mangle]
+pub extern "C" fn rs_dns_probe_srv(input: *const u8, len: u32)
+                                   -> u8
+{
+    let slice: &[u8] = unsafe {
+        std::slice::from_raw_parts(input as *mut u8, len as usize)
+    };
+    if probe_srv(slice) {
         return 1;
     }
     return 0;
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -532,7 +532,11 @@ again_midstream:
     } else {
         /* first try the destination port */
         pp_port_dp = AppLayerProtoDetectGetProbingParsers(alpd_ctx.ctx_pp, ipproto, dp);
-        alproto_masks = &f->probing_parser_toclient_alproto_masks;
+        if (dir == idir) {
+            // do not update alproto_masks to let a chance to second packet
+            // for instance when sending a junk packet to a DNS server
+            alproto_masks = &f->probing_parser_toclient_alproto_masks;
+        }
         if (pp_port_dp != NULL) {
             SCLogDebug("toclient - Probing parser found for destination port %"PRIu16, dp);
 
--- a/src/app-layer-dns-udp-rust.c
+++ b/src/app-layer-dns-udp-rust.c
@@ -65,6 +65,21 @@ static uint16_t DNSUDPProbe(Flow *f, uint8_t direction,
     return ALPROTO_DNS;
 }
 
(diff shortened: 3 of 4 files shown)


rust/src/dns/dns.rs
```diff
@@ -825,6 +829,19 @@ pub extern "C" fn rs_dns_probe(input: *const u8, len: u32)
     return 0;
 }
 
+#[no_mangle]
+pub extern "C" fn rs_dns_probe_srv(input: *const u8, len: u32)
+                                   -> u8
+{
+    let slice: &[u8] = unsafe {
```
victorjulien: you can use the `build_slice!` macro here.

src/app-layer-dns-udp-rust.c
```diff
@@ -65,6 +65,21 @@ static uint16_t DNSUDPProbe(Flow *f, uint8_t direction,
     return ALPROTO_DNS;
 }
 
+static uint16_t DNSUDPProbeSrv(Flow *f, uint8_t direction,
```
victorjulien: This will no longer be necessary once a version of #4513 is merged.

src/app-layer.c
```diff
@@ -697,42 +700,77 @@ int AppLayerHandleUdp(ThreadVars *tv, AppLayerThreadCtx *tctx, Packet *p, Flow *
...
-        f->alproto = AppLayerProtoDetectGetProto(tctx->alpd_tctx,
+        *alproto = AppLayerProtoDetectGetProto(tctx->alpd_tctx,
                                   f, p->payload, p->payload_len,
                                   IPPROTO_UDP, flags, &reverse_flow);
         PACKET_PROFILING_APP_PD_END(tctx);
 
-        if (f->alproto != ALPROTO_UNKNOWN) {
-            AppLayerIncFlowCounter(tv, f);
+        switch (*alproto) {
+            case ALPROTO_UNKNOWN:
```
victorjulien: I think `ALPROTO_UNKNOWN` and `ALPROTO_FAILED` have so much overlap that that they should be handled together.

src/app-layer-detect-proto.c
```diff
@@ -532,7 +532,11 @@ static AppProto AppLayerProtoDetectPPGetProto(Flow *f,
     } else {
         /* first try the destination port */
         pp_port_dp = AppLayerProtoDetectGetProbingParsers(alpd_ctx.ctx_pp, ipproto, dp);
-        alproto_masks = &f->probing_parser_toclient_alproto_masks;
+        if (dir == idir) {
```
victorjulien: I think I sort of understand what you're trying to do, but I don't understand why it work work or how to follow the flow of things.

Can you update the commit message to explain how in your test case the masks are wrong (or harmful), and what this patch does to address that?

rust/src/dns/dns.rs
catenacyber: ok

src/app-layer-dns-udp-rust.c
catenacyber: This is still a necessary bug to fix, only this bug fix will need to be moved to Rust once #4604 is merged

src/app-layer.c
catenacyber: Well, this Pull Request assumes that `ALPROTO_UNKNOWN` and `ALPROTO_FAILED` are well defined for different cases :
- `ALPROTO_UNKNOWN` means I do not know, this may be protocol X to not protocol X (like I do not enough data to decide)
- `ALPROTO_FAILED` this is definitely not protocol X

If some parsers are setting them wrong, should we not fix these parsers ?

src/app-layer-detect-proto.c
catenacyber: Ok doing

catenacyber: Well, I had already done it 
```
    Do not mask protocols on both directions with only first packet
    
    For instance :
    When the first packet is no valid DNS but on port 53 (a junk request)
    second packet (error response from server) does not get checked for DNS
    as first packet bit masked away DNS for both directions
```

So 

>  how in your test case the masks are wrong (or harmful)

It is wrong to mask protocols on _both_ directions with only _first_ packet

> what this patch does to address that?

The patch applies the bit mask only for the right direction

PR #4061, opened 2019-07-19
Fuzz v15

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds six fuzz targets

Modifies #4060 with 
- adding more `const` keywords for appveyor (different ifdef)

Some commits can be cherry-picked independently : log, config, style

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -6,8 +6,14 @@ noinst_HEADERS = action-globals.h \
 	suricata-common.h threadvars.h tree.h \
     util-validate.h
 bin_PROGRAMS = suricata
+if BUILD_FUZZTARGETS
+    bin_PROGRAMS += fuzz_applayerprotodetectgetproto \
+    fuzz_applayerparserparse fuzz_siginit \
+    fuzz_confyamlloadstring fuzz_decodepcapfile \
+    fuzz_sigpcap
+endif
 
-suricata_SOURCES = \
+COMMON_SOURCES = \
 alert-debuglog.c alert-debuglog.h \
 alert-fastlog.c alert-fastlog.h \
 alert-prelude.c alert-prelude.h \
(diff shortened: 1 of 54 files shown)


src/suricata.h
```diff
@@ -135,7 +135,7 @@ typedef struct SCInstance_ {
     enum RunModes aux_run_mode;
 
     char pcap_dev[128];
-    char *sig_file;
+    const char *sig_file;
```
victorjulien: this is not related to app-layer, so should go into its own commit

src/Makefile.am
```diff
@@ -519,13 +525,88 @@ EXTRA_DIST = tests
...
 AM_CPPFLAGS = $(all_includes)
 
+suricata_SOURCES = suricata.c suricata.h $(COMMON_SOURCES)
+
 # the library search path.
 suricata_LDFLAGS = $(all_libraries) ${SECLDFLAGS}
-suricata_LDADD = $(HTP_LDADD) $(RUST_LDADD)
+suricata_LDADD = $(RUST_SURICATA_LIB) $(HTP_LDADD) $(RUST_LDADD)
+
+fuzz_applayerprotodetectgetproto_SOURCES = tests/fuzz/fuzz_applayerprotodetectgetproto.c $(COMMON_SOURCES)
```
victorjulien: would it be possible to move those into a src/fuzz/Makefile.am or fuzz/Makefile.am? or perhaps a fuzz/applayerprotodetectgetproto/Makefile.am?

src/suricata.h
catenacyber: Actually, this cannot be made `const` as it will be used in some cases in argument for `dirname`

src/Makefile.am
catenacyber: I would love to.
The only way I see to achieve that is to compile `$(COMMON_SOURCES)` ie every source file except suricata which has `main` into a library (a convenience one).
I spent some hours trying and this is the best workaround I found, though it is not pretty.
The main problem is that C and Rust files depend on one another...

Any help from an autotool expert is welcome here :-)

PR #4051, opened 2019-07-17
detect/transform: add tld and domain transforms

jlucovsky:
This PR adds transforms for:
1. tld - Returns the top-level-domain and passes it on
1. domain - Returns the domain and passes it on.

Continuation of #4041 

This update:
1. Fixes unittest failures

Companion PR to suricata-verify` repo; https://github.com/OISF/suricata-verify/pull/101

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:  [3074](https://redmine.openinfosecfoundation.org/projects/suricata/issues/3074)

PR #4063, opened 2019-07-20
detect/transform: add tld and domain transforms

jlucovsky:
This PR adds transforms for:
1. tld - Returns the top-level-domain and passes it on

Continuation of #4051 

This update:
1. Fixes unittest failures

Companion PR to suricata-verify` repo; https://github.com/OISF/suricata-verify/pull/101

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:  [3074](https://redmine.openinfosecfoundation.org/issues/3074)

### DIFF
--- a/src/detect-transform-tld.c
+++ b/src/detect-transform-tld.c
@@ -1,0 +1,207 @@
+/* Copyright (C) 2019 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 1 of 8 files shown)


src/detect-transform-tld.c
```diff
@@ -0,0 +1,207 @@
...
+
+static void TransformTLD(InspectionBuffer *buffer)
+{
+    const uint8_t *input = buffer->inspect;
+    const uint32_t input_len = buffer->inspect_len;
+    uint8_t output[input_len]; // we can only shrink
+    uint8_t *os = output;
+
+    /* from end, scan back for dot */
+    uint8_t *dot = memrchr(input, '.', input_len);
```
regit: This will not work with tld like .co.uk. See https://wiki.mozilla.org/TLD_List. (cc @stricaud)

regit: @stricaud has written a library to parse url. I don't know if it is suitable for the task but it worth at least to take a look for the code: https://github.com/stricaud/faup

jlucovsky: ICANN (and others) define TLD as the rightmost element of a name: http://archive.icann.org/en/tlds/

I think we should consider using the URL parsing library in the future.

stricaud: The relevant part is faup_tld_tree_extract() in https://github.com/stricaud/faup/blob/master/src/lib/tld-tree.c

jlucovsky:
Continued in #4104

PR #4074, opened 2019-07-26
Fuzz v16

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds six fuzz targets

Modifies #4061 with 
- Updated documentation (including libfuzzer and afl in addition to oss-fuzz)
- Use of state-of-the art `-fsanitize=fuzzer` instead of libFuzzingEngine.a
- Remove some wrong `const` keywords and use of `strdup` to create a non-const char * in fuzz_sigpcap.c
- Move more code from suricata.c to init.c

Some commits can be cherry-picked independently : log, config, check

About completeness versus afl options, two targets might be missing :
- `MimeDecParseLine`
- `DecodeDer`
Should I create fuzz targets for these 2 functions ?

Otherwise, we have a generic fuzz target fuzz_applayerparserparse which replaces all protocol-specific targets. Do you have any opinion about this design ?

Finally, there will be `#ifdef` to be reviewed such as `AFLFUZZ_PCAP_RUNMODE`

Should we include in this PR changes to .travis.yml ? (ie make a build with `--enable-fuzztargets` configure option)

### DIFF
--- a/src/conf-yaml-loader.c
+++ b/src/conf-yaml-loader.c
@@ -195,15 +195,15 @@ ConfYamlParse(yaml_parser_t *parser, ConfNode *parent, int inseq)
             yaml_version_directive_t *ver =
                 event.data.document_start.version_directive;
             if (ver == NULL) {
-                fprintf(stderr, "ERROR: Invalid configuration file.\n\n");
-                fprintf(stderr, "The configuration file must begin with the following two lines:\n\n");
-                fprintf(stderr, "%%YAML 1.1\n---\n\n");
+                SCLogError(SC_ERR_CONF_YAML_ERROR, "ERROR: Invalid configuration file.");
+                SCLogError(SC_ERR_CONF_YAML_ERROR,
+                           "The configuration file must begin with the following two lines: %%YAML 1.1 and ---");
                 goto fail;
             }
             int major = ver->major;
             int minor = ver->minor;
             if (!(major == YAML_VERSION_MAJOR && minor == YAML_VERSION_MINOR)) {
-                fprintf(stderr, "ERROR: Invalid YAML version.  Must be 1.1\n");
+                SCLogError(SC_ERR_CONF_YAML_ERROR, "ERROR: Invalid YAML version.  Must be 1.1");
                 goto fail;
             }
         }
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@
     AC_CONFIG_HEADERS([config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
     AC_CONFIG_MACRO_DIR(m4)
-    AM_INIT_AUTOMAKE
+    AM_INIT_AUTOMAKE([subdir-objects])
 
     AC_LANG([C])
     AC_PROG_CC_C99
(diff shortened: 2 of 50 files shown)


src/conf-yaml-loader.c
```diff
@@ -195,15 +195,15 @@ ConfYamlParse(yaml_parser_t *parser, ConfNode *parent, int inseq)
             yaml_version_directive_t *ver =
                 event.data.document_start.version_directive;
             if (ver == NULL) {
-                fprintf(stderr, "ERROR: Invalid configuration file.\n\n");
-                fprintf(stderr, "The configuration file must begin with the following two lines:\n\n");
-                fprintf(stderr, "%%YAML 1.1\n---\n\n");
+                SCLogError(SC_ERR_CONF_YAML_ERROR, "ERROR: Invalid configuration file.");
```
jlucovsky: I don't mind this change but I'm wondering why the change was restricted to this file? There are plenty of `fprintf` and `printf` calls elsewhere.

catenacyber: These seemed the only ones reachable from fuzzing targets.

Doing `grep fprintf src/*.c | grep stderr`, I get 
```
src/conf-yaml-loader.c:                fprintf(stderr, "ERROR: Invalid configuration file.\n\n");
src/conf-yaml-loader.c:                fprintf(stderr, "The configuration file must begin with the following two lines:\n\n");
src/conf-yaml-loader.c:                fprintf(stderr, "%%YAML 1.1\n---\n\n");
src/conf-yaml-loader.c:                fprintf(stderr, "ERROR: Invalid YAML version.  Must be 1.1\n");
src/conf-yaml-loader.c:        fprintf(stderr, "Failed to initialize yaml parser.\n");
src/conf-yaml-loader.c:        fprintf(stderr, "final value was reassigned\n");
src/suricata.c:                    fprintf(stderr, "ERROR: Failed to set pfring.cluster-id.\n");
src/suricata.c:                    fprintf(stderr, "ERROR: Failed to set pfring.cluster-type.\n");
src/suricata.c:                    fprintf(stderr, "ERROR: Failed to set engine init-failure-fatal.\n");
src/suricata.c:                            fprintf(stderr, "ERROR: Failed to set unix-command.filename.\n");
src/suricata.c:                fprintf(stderr, "ERROR: Unit tests not enabled. Make sure to pass --enable-unittests to configure when building.\n");
src/suricata.c:                fprintf(stderr, "ERROR: Unit tests not enabled. Make sure to pass --enable-unittests to configure when building.\n");
src/suricata.c:                    fprintf(stderr, "ERROR: Failed to set erf-file.file\n");
src/suricata.c:                    fprintf(stderr, "ERROR: Failed to set pcap-buffer-size.\n");
src/suricata.c:                        fprintf(stderr, "Failed to set configuration value %s.",
src/suricata.c:                fprintf(stderr, "ERROR: Failed to set engine init-failure-fatal.\n");
src/suricata.c:            fprintf(stderr, "ERROR: Unit tests not enabled. Make sure to pass --enable-unittests to configure when building.\n");
src/util-byte.c:        fprintf(stderr, "ByteExtractString: Extra characters following numeric value\n");
src/util-byte.c:        fprintf(stderr, "ByteExtractStringSigned: Extra characters following numeric value\n");
src/util-byte.c:    //fprintf(stderr, "ByteExtractStringSigned: Extracted base %d: 0x%" PRIx64 "\n", base, *res);
src/util-unittest.c:                        fprintf(stderr, "ERROR: unittest failed.\n");
```

Maybe we can change the ones in src/util-byte.c
For the other ones, I am not sure logging is initialized before we call them (especially the ones in suricata.c)

PR #4082, opened 2019-08-01
detectproto: adding missing probing parsers

catenacyber:

Describe changes:
- Adds missing probing parser in `TOCLIENT` direction

These are for symmetric protocols (at least symmetric enough for the header where the probing parsers stay)

app-layer-template.c is also modified not to have zero `TOCLIENT` probing parsers 
- Should we leave a comment that every protocol is not symmetric ? (For instance DNP3 is symmetric and HTTP is not)
- Or should we have `TemplateProbingParserTs` and `TemplateProbingParserTc` ?


jlucovsky:
> * Should we leave a comment that every protocol is not symmetric ? (For instance DNP3 is symmetric and HTTP is not)
> * Or should we have `TemplateProbingParserTs` and `TemplateProbingParserTc` ?

I like the 2nd approach  with a comment in each of these functions including guidance for new protocols to determine if a parser for both sides should be implemented.

PR #4085, opened 2019-08-02
detectproto: adding missing probing parsers

catenacyber:

Describe changes:
- Adds missing probing parser in `TOCLIENT` direction

These are for symmetric protocols (at least symmetric enough for the header where the probing parsers stay)

Modifies #4082 with
- having both `TemplateProbingParserTs` and `TemplateProbingParserTc` with a comment saying that we can use only one
- precision in comments on the returned value of these functions (which can return `ALPROTO_FAILED` with a different meaning than `AL_PROTO_UNKNOWN`)


victorjulien:
Merged in #4113, thanks Philippe!

PR #4087, opened 2019-08-02
Fuzz v17

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds six fuzz targets

Modifies #4074 with 
- use of `$LIB_FUZZING_ENGINE` to be compatible with oss-fuzz
- redirect log file of `fuzz_applayerparserparse` to /dev/null

Some commits can be cherry-picked independently : log, config, check

About completeness versus afl options, two targets might be missing :
- `MimeDecParseLine`
- `DecodeDer`
Should I create fuzz targets for these 2 functions ?

Otherwise, we have a generic fuzz target fuzz_applayerparserparse which replaces all protocol-specific targets. Do you have any opinion about this design ?

Finally, there will be `#ifdef` to be reviewed such as `AFLFUZZ_PCAP_RUNMODE`

Should we include in this PR changes to .travis.yml ? (ie make a build with `--enable-fuzztargets` configure option)

### DIFF
--- a/src/tests/fuzz/fuzz_applayerparserparse.c
+++ b/src/tests/fuzz/fuzz_applayerparserparse.c
@@ -1,0 +1,104 @@
+/**
+ * @file
+ * @author Philippe Antoine <contact@catenacyber.fr>
+ * fuzz target for AppLayerProtoDetectGetProto
+ */
+
+
+#include "suricata-common.h"
+#include "app-layer-detect-proto.h"
+#include "flow-util.h"
+#include "app-layer-parser.h"
+
+#define HEADER_LEN 6
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+AppLayerParserThreadCtx *alp_tctx = NULL;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    Flow * f;
+    TcpSession ssn;
+
+    if (size < HEADER_LEN) {
+        return 0;
+    }
+
+    if (alp_tctx == NULL) {
+        //Redirects logs to /dev/null
+        setenv("SC_LOG_OP_IFACE", "file", 0);
+        setenv("SC_LOG_FILE", "/dev/null", 0);
+
+        InitGlobal();
+        run_mode = RUNMODE_UNITTEST;
+        FlowInitConfig(FLOW_QUIET);
+        MpmTableSetup();
+        SpmTableSetup();
+        AppLayerProtoDetectSetup();
+        AppLayerParserSetup();
+        AppLayerParserRegisterProtocolParsers();
+        alp_tctx = AppLayerParserThreadCtxAlloc();
+    }
+
+    if (data[0] >= ALPROTO_MAX) {
+        return 0;
+    }
+    f = FlowAlloc();
+    if (f == NULL) {
+        return 0;
+    }
+    f->flags |= FLOW_IPV4;
+    f->src.addr_data32[0] = 0x01020304;
+    f->dst.addr_data32[0] = 0x05060708;
+    f->sp = (data[2] << 8) | data[3];
+    f->dp = (data[4] << 8) | data[5];
+    f->proto = data[1];
+    memset(&ssn, 0, sizeof(TcpSession));
+    f->protoctx = &ssn;
+    f->protomap = FlowGetProtoMapping(f->proto);
+    f->alproto = data[0];
+
+    int start = 1;
+    int flip = 0;
+    size_t offset = HEADER_LEN;
+    while (1) {
+        int done = 0;
+        uint8_t flags = 0;
+        size_t onesize = 0;
+        if (flip) {
+            flags = STREAM_TOCLIENT;
+            flip = 0;
+        } else {
+            flags = STREAM_TOSERVER;
+            flip = 1;
+        }
+        if (start > 0) {
+            flags |= STREAM_START;
+            start = 0;
+        }
+        if (size < offset + 2) {
+            onesize = 0;
+            done = 1;
+        } else {
+            onesize = ((data[offset]) << 8) | (data[offset+1]);
+            offset += 2;
+            if (size < offset + onesize) {
+                onesize = size - offset;
+                done = 1;
+            }
+        }
+        if (done) {
+            flags |= STREAM_EOF;
+        }
+
+        (void) AppLayerParserParse(NULL, alp_tctx, f, f->alproto, flags, data+offset, onesize);
+        offset += onesize;
+        if (done)
+            break;
+
+    }
+
+    FlowFree(f);
+
+    return 0;
+}
--- a/src/tests/fuzz/fuzz_applayerprotodetectgetproto.c
+++ b/src/tests/fuzz/fuzz_applayerprotodetectgetproto.c
@@ -1,0 +1,57 @@
+/**
+ * @file
+ * @author Philippe Antoine <contact@catenacyber.fr>
+ * fuzz target for AppLayerProtoDetectGetProto
+ */
+
+
+#include "suricata-common.h"
+#include "app-layer-detect-proto.h"
+#include "flow-util.h"
+#include "app-layer-parser.h"
+
+
+#define HEADER_LEN 6
+
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+AppLayerProtoDetectThreadCtx *alpd_tctx = NULL;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    Flow f;
+    TcpSession ssn;
+    bool reverse;
+
+    if (size < HEADER_LEN) {
+        return 0;
+    }
+
+    if (alpd_tctx == NULL) {
+        //global init
+        InitGlobal();
+        run_mode = RUNMODE_UNITTEST;
+        MpmTableSetup();
+        SpmTableSetup();
+        AppLayerProtoDetectSetup();
+        AppLayerParserSetup();
+        AppLayerParserRegisterProtocolParsers();
+        alpd_tctx = AppLayerProtoDetectGetCtxThread();
+    }
+
+    memset(&f, 0, sizeof(f));
+    FLOW_INITIALIZE(&f);
+    f.flags |= FLOW_IPV4;
+    f.src.addr_data32[0] = 0x01020304;
+    f.dst.addr_data32[0] = 0x05060708;
+    f.sp = (data[2] << 8) | data[3];
+    f.dp = (data[4] << 8) | data[5];
+    f.proto = data[1];
+    memset(&ssn, 0, sizeof(TcpSession));
+    f.protoctx = &ssn;
+    f.protomap = FlowGetProtoMapping(f.proto);
+
+    AppLayerProtoDetectGetProto(alpd_tctx, &f, data+HEADER_LEN, size-HEADER_LEN, f.proto, data[0], &reverse);
+
+    return 0;
+}
--- a/src/tests/fuzz/fuzz_decodepcapfile.c
+++ b/src/tests/fuzz/fuzz_decodepcapfile.c
@@ -1,0 +1,114 @@
+/**
+ * @file
+ * @author Philippe Antoine <contact@catenacyber.fr>
+ * fuzz target for AppLayerProtoDetectGetProto
+ */
+
+
+#include <pcap/pcap.h>
+
+#include "suricata-common.h"
+#include "app-layer-detect-proto.h"
+#include "defrag.h"
+#include "tm-modules.h"
+#include "source-pcap-file.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+static int bufferToFile(const char * name, const uint8_t *Data, size_t Size) {
+    FILE * fd;
+    if (remove(name) != 0) {
+        if (errno != ENOENT) {
+            printf("failed remove, errno=%d\n", errno);
+            return -1;
+        }
+    }
+    fd = fopen(name, "wb");
+    if (fd == NULL) {
+        printf("failed open, errno=%d\n", errno);
+        return -2;
+    }
+    if (fwrite (Data, 1, Size, fd) != Size) {
+        fclose(fd);
+        return -3;
+    }
+    fclose(fd);
+    return 0;
+}
+
+static int initialized = 0;
+ThreadVars tv;
+PacketQueue pq;
+DecodeThreadVars *dtv;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    pcap_t * pkts;
+    char errbuf[PCAP_ERRBUF_SIZE];
+    const u_char *pkt;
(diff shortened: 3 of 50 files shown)


catenacyber:
Travis CI failure comes from apt
Can we launch it again ?

catenacyber:
Travis CI failure comes from apt
Can we launch it again ?

src/tests/fuzz/fuzz_applayerparserparse.c
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    if (data[0] >= ALPROTO_MAX) {
+        return 0;
+    }
+    f = FlowAlloc();
+    if (f == NULL) {
+        return 0;
+    }
+    f->flags |= FLOW_IPV4;
```
victorjulien: I would like to see this mess with fewer internals, but instead use something like UTHBuildFlow or something equivalent.

src/tests/fuzz/fuzz_applayerprotodetectgetproto.c
```diff
@@ -0,0 +1,57 @@
...
+        SpmTableSetup();
+        AppLayerProtoDetectSetup();
+        AppLayerParserSetup();
+        AppLayerParserRegisterProtocolParsers();
+        alpd_tctx = AppLayerProtoDetectGetCtxThread();
+    }
+
+    memset(&f, 0, sizeof(f));
+    FLOW_INITIALIZE(&f);
+    f.flags |= FLOW_IPV4;
```
victorjulien: similar to comment above, lets use something like UTHBuildFlow

src/tests/fuzz/fuzz_decodepcapfile.c
```diff
@@ -0,0 +1,114 @@
...
+        MpmTableSetup();
+        SpmTableSetup();
+        AppLayerProtoDetectSetup();
+        DefragInit();
+        FlowInitConfig(FLOW_QUIET);
+
+        //redirect logs to /tmp
+        ConfigSetLogDirectory("/tmp/");
+        //disables checksums validation for fuzzing
+        suricata.checksum_validation = 0;
```
victorjulien: can we replace hard coding things like this by a call to ConfYamlLoadString with a minimal config?

src/tests/fuzz/fuzz_decodepcapfile.c
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    //initialize structure
+    pkts = pcap_open_offline("/tmp/fuzz.pcap", errbuf);
+    if (pkts == NULL) {
+        return 0;
+    }
+
+    //loop over packets
+    r = pcap_next_ex(pkts, &header, &pkt);
```
victorjulien: I'm not too excited about having these custom pcap loops, can we use the one for the pcap file reading code somehow?

src/tests/fuzz/fuzz_sigpcap.c
```diff
@@ -0,0 +1,181 @@
...
+    if (initialized == 0) {
+        //Redirects logs to /dev/null
+        setenv("SC_LOG_OP_IFACE", "file", 0);
+        setenv("SC_LOG_FILE", "/dev/null", 0);
+
+        InitGlobal();
+
+        run_mode = RUNMODE_PCAP_FILE;
+        //redirect logs to /tmp
+        ConfigSetLogDirectory("/tmp/");
```
victorjulien: same suggestion to load config from yaml from mem

src/tests/fuzz/fuzz_sigpcap.c
```diff
@@ -0,0 +1,181 @@
...
+        suricata.delayed_detect = 1;
+
+        SupportFastPatternForSigMatchTypes();
+        PostConfLoadedSetup(&suricata);
+
+        //dummy init before DetectEngineReload
+        DetectEngineCtx * de_ctx = DetectEngineCtxInit();
+        DetectEngineAddToMaster(de_ctx);
+
+        memset(&tv, 0, sizeof(tv));
```
victorjulien: I don't like these kinds of hard coded init steps, they are easy to mess up or forget when main logic changes.

src/tests/fuzz/fuzz_sigpcap.c
```diff
@@ -0,0 +1,181 @@
...
+        if (bufferToFile(suricata.sig_file, data, pos-1) < 0) {
+            return 0;
+        }
+    } else {
+        if (bufferToFile(suricata.sig_file, data, pos) < 0) {
+            return 0;
+        }
+    }
+
+    if (DetectEngineReload(&suricata) < 0) {
```
victorjulien: what is the goal here?

victorjulien:
In general I would like us to think about how we can reduce the amount of 'magic' glue code to the absolute minimum.

src/tests/fuzz/fuzz_applayerparserparse.c
```diff
@@ -0,0 +1,104 @@
...
+#include "flow-util.h"
+#include "app-layer-parser.h"
+
+#define HEADER_LEN 6
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+AppLayerParserThreadCtx *alp_tctx = NULL;
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
```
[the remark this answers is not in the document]
catenacyber: Do you mean document what is `alp_tctx` for ?
I am not sure I get what you want...

src/tests/fuzz/fuzz_applayerparserparse.c
catenacyber: ok, I did not know this `UTHBuildFlow`, seems handy

src/tests/fuzz/fuzz_applayerparserparse.c
```diff
@@ -0,0 +1,104 @@
...
+        }
+        if (start > 0) {
+            flags |= STREAM_START;
+            start = 0;
+        }
+        if (size < offset + 2) {
+            onesize = 0;
+            done = 1;
+        } else {
+            onesize = ((data[offset]) << 8) | (data[offset+1]);
```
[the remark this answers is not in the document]
catenacyber: We want to fuzz multiple calls to AppLayerParserParse because some parts of the code are only reached after multiple packets (in SMTP for example).
So we treat our input as a list of sizes encoded as uint16_t and buffer.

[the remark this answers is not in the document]
catenacyber: I think that the best should be this one : https://github.com/google/fuzzing/blob/master/docs/split-inputs.md#magic-separator

src/tests/fuzz/fuzz_applayerprotodetectgetproto.c
catenacyber: ok

src/tests/fuzz/fuzz_decodepcapfile.c
catenacyber: Do you mean a hard-coded minimal config in the C file ?

src/tests/fuzz/fuzz_decodepcapfile.c
catenacyber: > can we use the one for the pcap file reading code somehow?

I don't understand your sentence :-/

src/tests/fuzz/fuzz_siginit.c
```diff
@@ -0,0 +1,46 @@
...
+        run_mode = RUNMODE_UNITTEST;
+        MpmTableSetup();
+        SpmTableSetup();
+        SigTableSetup();
+        SCReferenceConfInit();
+        SCClassConfInit();
+        de_ctx = DetectEngineCtxInit();
+    }
+
+    char * buffer = malloc(size+1);
```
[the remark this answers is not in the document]
catenacyber: I propose to get a first version merged and optimize afterwards

src/tests/fuzz/fuzz_siginit.c
```diff
@@ -0,0 +1,46 @@
...
+    if (buffer) {
+        memcpy(buffer, data, size);
+        //null terminate string
+        buffer[size] = 0;
+        Signature *s = SigInit(de_ctx, buffer);
+        free(buffer);
+        if (s != NULL) {
```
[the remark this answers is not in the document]
catenacyber: ok cool

src/tests/fuzz/fuzz_sigpcap.c
```diff
@@ -0,0 +1,181 @@
...
+#include "detect-engine-threshold.h"
+#include "host-bit.h"
+#include "ippair-bit.h"
+#include "app-layer-htp.h"
+#include "util-decode-asn1.h"
+#include "detect-fast-pattern.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+static int bufferToFile(const char * name, const uint8_t *Data, size_t Size) {
```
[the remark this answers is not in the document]
catenacyber: Is there already one such function in Suricata ?
If not, should it be in util-unittest-helper.c ?

src/tests/fuzz/fuzz_sigpcap.c
catenacyber: To reload signatures.

The fuzzed input is here : a signatures file (plural for signatures) and a pcap
So we reload signatures before processing the next pcap

src/tests/fuzz/fuzz_sigpcap.c
catenacyber: So, what is the right way to do ?

catenacyber:
I added the two fuzz targets to next version
- `MimeDecParseLine`
- `DecodeDer`

Now I need to rebase on master and test with oss-fuzz

src/tests/fuzz/fuzz_applayerparserparse.c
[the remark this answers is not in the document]
victorjulien: No I mean the input data has a 'magic' layout: first byte is alproto, 2nd byte ipproto, etc. This should be documented in a comment.

src/tests/fuzz/fuzz_decodepcapfile.c
victorjulien: Yeah, something like https://github.com/OISF/suricata/blob/master/src/app-layer-htp.c#L3862

src/tests/fuzz/fuzz_sigpcap.c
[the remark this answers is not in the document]
victorjulien: Probably a 'util-fuzz-helper' or something, but yeah.

src/tests/fuzz/fuzz_decodepcapfile.c
victorjulien: We already have code for reading a pcap file and processing the packets. Can't we reuse that somehow?

src/tests/fuzz/fuzz_applayerparserparse.c
victorjulien: Maybe we can directly reuse it, or maybe we need a small separate set of Fuzz support funcs.

src/tests/fuzz/fuzz_applayerparserparse.c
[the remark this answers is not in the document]
victorjulien: The magic separator seems different than this though. Would it make sense to switch to this instead of embedding size args? The size args themselves can be invalid I suppose, so wouldn't also be testing the fuzz loop this way?

src/tests/fuzz/fuzz_sigpcap.c
victorjulien: Ah yeah, I see. Thanks.

src/tests/fuzz/fuzz_siginit.c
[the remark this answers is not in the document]
victorjulien: Ok sure.

src/tests/fuzz/fuzz_sigpcap.c
victorjulien: Not useful now: I've started work on something to register startup/cleanup routines with dependency handling. When done we should be able to say something like 'load detect engine and its dependencies (in the correct order)'.

src/tests/fuzz/fuzz_sigpcap.c
```diff
@@ -0,0 +1,181 @@
...
+        SupportFastPatternForSigMatchTypes();
+        PostConfLoadedSetup(&suricata);
+
+        //dummy init before DetectEngineReload
+        DetectEngineCtx * de_ctx = DetectEngineCtxInit();
+        DetectEngineAddToMaster(de_ctx);
+
+        memset(&tv, 0, sizeof(tv));
+        dtv = DecodeThreadVarsAlloc(&tv);
+        DecodeRegisterPerfCounters(dtv, &tv);
```
victorjulien: this one is also called from the ThreadInit in line 95, is it needed to run both?

src/tests/fuzz/fuzz_sigpcap.c
victorjulien: For now I don't really know a cleaner way to do quickly.

src/tests/fuzz/fuzz_applayerparserparse.c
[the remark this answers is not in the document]
catenacyber: ok

src/tests/fuzz/fuzz_applayerparserparse.c
catenacyber: It works, I just needed to make `enable-fuzztargets` option of configure implies `enable-unittests`
That seemed ok to me.

src/tests/fuzz/fuzz_applayerparserparse.c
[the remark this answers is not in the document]
catenacyber: Indeed it is different. I will switch to this

src/tests/fuzz/fuzz_sigpcap.c
catenacyber: This looks needed for `TMM_DECODEPCAPFILE` even if `ThreadInit` line 95 does it for `TMM_FLOWWORKER`

src/tests/fuzz/fuzz_sigpcap.c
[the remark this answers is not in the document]
catenacyber: Ok, putting it first in util-unittest-helper.c since we already use `UTHBuildFlow`

src/tests/fuzz/fuzz_decodepcapfile.c
catenacyber: ok

src/tests/fuzz/fuzz_decodepcapfile.c
catenacyber: I finally managed it. The down sides are that :
- it is less clear which functions we are fuzzing  (`DecodeEthernet` et al)
- There is more "magic glue"

You will tell me when I push next rebased version

src/tests/fuzz/fuzz_sigpcap.c
catenacyber: done

PR #4097, opened 2019-08-11
src/detect: check DetectBufferSetActiveList return code

shivan1b:
Make sure to always check the return codes of DetectBufferSetActiveList.
Also, force this warning on function prototype.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3005

### DIFF
--- a/src/detect-engine.h
+++ b/src/detect-engine.h
@@ -164,7 +164,7 @@ void DetectEngineSetParseMetadata(void);
 void DetectEngineUnsetParseMetadata(void);
 int DetectEngineMustParseMetadata(void);
 
-int DetectBufferSetActiveList(Signature *s, const int list);
+int WARN_UNUSED DetectBufferSetActiveList(Signature *s, const int list);
 int DetectBufferGetActiveList(DetectEngineCtx *de_ctx, Signature *s);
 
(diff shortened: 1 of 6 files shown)


src/detect-engine.h
```diff
@@ -164,7 +164,7 @@ void DetectEngineSetParseMetadata(void);
 void DetectEngineUnsetParseMetadata(void);
 int DetectEngineMustParseMetadata(void);
 
-int DetectBufferSetActiveList(Signature *s, const int list);
+int WARN_UNUSED DetectBufferSetActiveList(Signature *s, const int list);
```
jlucovsky: Since you've modified the offending places (where the return code was ignored), is `WARN_UNUSED` still needed?

shivan1b: Its not needed for the current codebase but is this function not going to be used in future again? If yes, I think check would prove to be helpful at all those places.

jlucovsky: I agree that this would be useful moving forward.

victorjulien:
Merged in #4112, thanks Shivani!

PR #4101, opened 2019-08-14

shivan1b:
ByteExtractString* functions.

### DIFF
--- a/src/detect-dce-iface.c
+++ b/src/detect-dce-iface.c
@@ -180,7 +180,9 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
             goto error;
         }
 
-        version = atoi(copy_str);
+        if (ByteExtractStringInt32(&version, 10, 0, (const char *)copy_str) < 0) {
+            version = 0;
+        }
         if (version > UINT16_MAX) {
             SCLogError(SC_ERR_INVALID_SIGNATURE, "DCE_IFACE interface version "
                        "invalid: %d\n", version);
--- a/src/detect-engine-port.c
+++ b/src/detect-engine-port.c
@@ -51,6 +51,7 @@
 #include "pkt-var.h"
 #include "host.h"
 #include "util-profiling.h"
+#include "util-byte.h"
 #include "util-var.h"
 
 static int DetectPortCutNot(DetectPort *, DetectPort **);
@@ -1418,14 +1419,22 @@ DetectPort *PortParse(const char *str)
         port2[0] = '\0';
         port2++;
 
-        if(DetectPortIsValidRange(port))
-            dp->port = atoi(port);
+        if (DetectPortIsValidRange(port)) {
+            if (ByteExtractStringUint16(&dp->port, 10, 0,
+                                        (const char *)port) < 0) {
+                dp->port = 0;
+            }
+        }
         else
             goto error;
 
         if (strcmp(port2, "") != 0) {
-            if (DetectPortIsValidRange(port2))
-                dp->port2 = atoi(port2);
+            if (DetectPortIsValidRange(port2)) {
+                if (ByteExtractStringUint16(&dp->port2, 10, 0,
+                                            (const char *)port2) < 0) {
+                    dp->port2 = 0;
+                }
+            }
             else
                 goto error;
         } else {
@@ -1440,7 +1449,13 @@ DetectPort *PortParse(const char *str)
             dp->port = 0;
             dp->port2 = 65535;
         } else if(DetectPortIsValidRange(port)){
-            dp->port = dp->port2 = atoi(port);
+            if ((ByteExtractStringUint16(&dp->port, 10, 0,
+                                         (const char *)port) < 0) ||
+                (ByteExtractStringUint16(&dp->port2, 10, 0,
+                                         (const char *)port) < 0)) {
+                dp->port = dp->port2 = 0;
+            }
+
         } else {
             goto error;
         }
--- a/src/detect-engine.c
+++ b/src/detect-engine.c
@@ -2342,7 +2342,10 @@ static int DetectEngineCtxLoadConf(DetectEngineCtx *de_ctx)
             }
 
             if (insp_recursion_limit != NULL) {
-                de_ctx->inspection_recursion_limit = atoi(insp_recursion_limit);
+                if (ByteExtractStringInt32(&de_ctx->inspection_recursion_limit,
+                    10, 0, (const char *)insp_recursion_limit) < 0) {
+                    de_ctx->inspection_recursion_limit = 0;
+                }
             } else {
                 de_ctx->inspection_recursion_limit =
                     DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT;
--- a/src/detect-nfs-version.c
+++ b/src/detect-nfs-version.c
@@ -42,6 +42,7 @@
 
 #include "util-unittest.h"
 #include "util-unittest-helper.h"
+#include "util-byte.h"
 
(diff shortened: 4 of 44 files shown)


src/detect-dce-iface.c
```diff
@@ -180,7 +180,9 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
             goto error;
         }
 
-        version = atoi(copy_str);
+        if (ByteExtractStringInt32(&version, 10, 0, (const char *)copy_str) < 0) {
+            version = 0;
```
shivan1b: Position 1.

src/detect-engine-port.c
```diff
@@ -1418,14 +1419,22 @@ DetectPort *PortParse(const char *str)
         port2[0] = '\0';
         port2++;
 
-        if(DetectPortIsValidRange(port))
-            dp->port = atoi(port);
+        if (DetectPortIsValidRange(port)) {
+            if (ByteExtractStringUint16(&dp->port, 10, 0,
+                                        (const char *)port) < 0) {
+                dp->port = 0;
```
shivan1b: Position 2.

src/detect-engine-port.c
```diff
@@ -1418,14 +1419,22 @@ DetectPort *PortParse(const char *str)
...
         else
             goto error;
 
         if (strcmp(port2, "") != 0) {
-            if (DetectPortIsValidRange(port2))
-                dp->port2 = atoi(port2);
+            if (DetectPortIsValidRange(port2)) {
+                if (ByteExtractStringUint16(&dp->port2, 10, 0,
+                                            (const char *)port2) < 0) {
+                    dp->port2 = 0;
```
shivan1b: Position 3.

src/detect-engine-port.c
```diff
@@ -1440,7 +1449,13 @@ DetectPort *PortParse(const char *str)
             dp->port = 0;
             dp->port2 = 65535;
         } else if(DetectPortIsValidRange(port)){
-            dp->port = dp->port2 = atoi(port);
+            if ((ByteExtractStringUint16(&dp->port, 10, 0,
+                                         (const char *)port) < 0) ||
+                (ByteExtractStringUint16(&dp->port2, 10, 0,
+                                         (const char *)port) < 0)) {
+                dp->port = dp->port2 = 0;
```
shivan1b: Position 4.

src/detect-engine.c
```diff
@@ -2342,7 +2342,10 @@ static int DetectEngineCtxLoadConf(DetectEngineCtx *de_ctx)
             }
 
             if (insp_recursion_limit != NULL) {
-                de_ctx->inspection_recursion_limit = atoi(insp_recursion_limit);
+                if (ByteExtractStringInt32(&de_ctx->inspection_recursion_limit,
+                    10, 0, (const char *)insp_recursion_limit) < 0) {
+                    de_ctx->inspection_recursion_limit = 0;
```
shivan1b: Position 5.

rezarb24:
Hi i had some questions
1-Source-Windows Version Is the link below?
3-

Suricata (Development) version is 5.0.0-beta1; released April 30, 2019.

    Linux/Mac/FreeBSD/UNIX/Windows Source: suricata-5.0.0-beta1.tar.gz

4-What is the programming language of this program?
5. What software can this program be edited?
6-How to get output after editing?
To install in windows(exe)

src/detect-nfs-version.c
```diff
@@ -282,7 +283,9 @@ static DetectNfsVersionData *DetectNfsVersionParse (const char *rawstr)
     }
 
     /* set the first value */
-    dd->lo = atoi(value1); //TODO
+    if (ByteExtractStringUint32(&dd->lo, 10, 0, (const char *)value1) < 0) { //TODO
+        return 0;
```
jlucovsky: This should be treated as an error, IMO.

src/detect-nfs-version.c
```diff
@@ -292,8 +295,9 @@ static DetectNfsVersionData *DetectNfsVersionParse (const char *rawstr)
             goto error;
         }
 
-        //
-        dd->hi = atoi(value2); // TODO
+        if (ByteExtractStringUint32(&dd->hi, 10, 0, (const char *)value2) < 0) { //TODO
+            return 0;
```
jlucovsky: This should be treated as an error, IMO.

src/detect-stream_size.c
```diff
@@ -245,7 +246,9 @@ static DetectStreamSizeData *DetectStreamSizeParse (const char *streamstr)
     }
 
     /* set the value */
-    sd->ssize = (uint32_t)atoi(value);
+    if (ByteExtractStringUint32(&sd->ssize, 10, 0, (const char *)value) < 0) {
+        return 0;
```
jlucovsky: This should be treated as an error. Also, returning from here will leave `mode, arg` and `value` dangling (leak).

src/detect-template.c
```diff
@@ -158,8 +159,12 @@ static DetectTemplateData *DetectTemplateParse (const char *templatestr)
     if (unlikely(templated == NULL))
         return NULL;
 
-    templated->arg1 = (uint8_t)atoi(arg1);
-    templated->arg2 = (uint8_t)atoi(arg2);
+    if (ByteExtractStringUint8(&templated->arg1, 10, 0, (const char *)arg1) < 0) {
+        return 0;
```
jlucovsky: Suggest returning `NULL` and also calling `SCFree(templated)` in these 2 cases.

shivan1b:
> I reviewed approx half of the changes for this PR. IMO, most (but not all) of the times `ByteExtract` returns less than 0 indicates an error condition.
> 
Thanks, Jeff.

> In a few cases, a default value should be used when this occurs. In others, an error condition exists. Sometimes, the condition can be treated as a value that's outside the range of permitted values.
> 
OK.

> Also in some cases, early returns will lead to leaks (dangling references).

I was afraid that these changes might be bringing in some bugs in the codebase. Tests only failed for 5 cases but there could be test cases for other modules which are not covered yet. Just like you pointed this one.

Do you think I should add so many of changes at once then? Seems like it is hard to review as well. I'm willing to give this more time and do the proper fixes.

jlucovsky:
> > I reviewed approx half of the changes for this PR. IMO, most (but not all) of the times `ByteExtract` returns less than 0 indicates an error condition.
> 
> Thanks, Jeff.
> 
> > In a few cases, a default value should be used when this occurs. In others, an error condition exists. Sometimes, the condition can be treated as a value that's outside the range of permitted values.
> 
> OK.
> 
> > Also in some cases, early returns will lead to leaks (dangling references).
> 
> I was afraid that these changes might be bringing in some bugs in the codebase. Tests only failed for 5 cases but there could be test cases for other modules which are not covered yet. Just like you pointed this one.
> 
> Do you think I should add so many of changes at once then? Seems like it is hard to review as well. I'm willing to give this more time and do the proper fixes.

IMO, it's ok to make this volume of changes. However, each change must be done within the context and provide appropriate error handling. Most often, that means ensuring proper error cleanup and messaging, if appropriate.

catenacyber:
Thanks @jlucovsky for the review which looks good.
@shivan1b do you still want me to review as well ?

> IMO, most (but not all) of the times ByteExtract returns less than 0 indicates an error condition.

Can you give an example where `ByteExtract` returns less than 0 but it is not an error condition.
Or did you mean `ByteExtractSomething` ?

shivan1b:
> Thanks @jlucovsky for the review which looks good.
> @shivan1b do you still want me to review as well ?
> 
@catenacyber I shall do a new PR today and tag you there. Please have a look at that. Thanks.

PR #4110, opened 2019-08-20
doc: use describe instead of option for old Sphinx - v1

jasonish:
Older versions of Sphinx will generate duplicate IDs when you have
options like:

.. option:: some-option

.. option:: some-other-option

The version of Sphinx provided on CentOS 7 has this issue, newer
versions of Sphinx do not.  As CentOS 7 is still a popular
distribution, change ".. option" to ".. describe" which has the
same visual output, but does not generate links.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/356
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/711


victorjulien:
Merged in #4180, thanks!

PR #4104, opened 2019-08-17
detect/transform: add domain transform

jlucovsky:
This PR adds a transform for:
1. domain - Returns the domain and passes it on.

Continuation of #4063 

This update:
1. Incorporates new logic for transformation (See defect)
1. Removes support for the `tld` transform due to issues with a heuristics-only approach. Proper TLD transformation requires a data store for validation and checks.
1. Adds additional test cases.

Companion PR to suricata-verify` repo; https://github.com/OISF/suricata-verify/pull/111

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:  [3074](https://redmine.openinfosecfoundation.org/issues/3074)

PR #4111, opened 2019-08-20
detect/transform: add domain transform

jlucovsky:
This PR adds a transform for:

Continuation of #4104 

This update:
1. Addresses review feedback

Companion PR to suricata-verify` repo; https://github.com/OISF/suricata-verify/pull/111

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:  [3074](https://redmine.openinfosecfoundation.org/issues/3074)

### DIFF
--- a/doc/userguide/rules/transforms.rst
+++ b/doc/userguide/rules/transforms.rst
@@ -21,6 +21,37 @@ Example::
 
 .. note:: not all sticky buffers support transformations yet
 
+domain
+------
+
+Takes the buffer, and prepends a ``.`` character providing concise domain checks. For example,
+an input string of ``hello.google.com`` would be modified and become ``.hello.google.com``. Additionally,
+adding the dot allows ``google.com`` to match against ``content:".google.com"``
+
+Example::
+
+    alert http any any -> any any (dns.query; domain; \
+        content:".microsoft.com"; sid:1;)
+
+This example will match on ``windows.update.microsoft.com`` and
+``maps.microsoft.com.au`` but not ``windows.update.fakemicrosoft.com``.
+
+This rule can be used to match on the domain only; example::
+
+    alert http any any -> any any (dns.query; domain; \
+        content:".microsoft.com"; endswith; sid:1;)
+
+This example will match on ``windows.update.microsoft.com`` but not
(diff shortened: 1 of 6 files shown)


victorjulien:
Any thoughts on the name of this transform? Its not 'domain' specific at all I think, it just puts a dot in front of whatever data it is used with. cc @jasonish

jlucovsky:
> Any thoughts on the name of this transform? Its not 'domain' specific at all I think, it just puts a dot in front of whatever data it is used with. cc @jasonish

True but it facilitates domain and tld identification.

jasonish:
> > Any thoughts on the name of this transform? Its not 'domain' specific at all I think, it just puts a dot in front of whatever data it is used with. cc @jasonish
> 
> True but it facilitates domain and tld identification.

As its not a domain I'm not sure if domain is the best name.  But it is treating it like a domain, so I don't really have a better name except for:

- `as_domain`. So in an OO language as hostname object might have some methods.  You'd expect `domain()` to return the actual domain, but maybe `.as_domain()` would return the hostname as a domain name.  With the caveat that both return a dot prefixed name.

- `dotprefix`: to be quote literal, give me the buffer with a . in front.

jlucovsky:
> > > Any thoughts on the name of this transform? Its not 'domain' specific at all I think, it just puts a dot in front of whatever data it is used with. cc @jasonish
> > 
> > 
> > True but it facilitates domain and tld identification.
> 
> As its not a domain I'm not sure if domain is the best name. But it is treating it like a domain, so I don't really have a better name except for:
> 
> * `as_domain`. So in an OO language as hostname object might have some methods.  You'd expect `domain()` to return the actual domain, but maybe `.as_domain()` would return the hostname as a domain name.  With the caveat that both return a dot prefixed name.
> * `dotprefix`: to be quote literal, give me the buffer with a . in front.

The most indicative is `dotprefix` but I like `as_domain` since that provides the purpose.

@victorjulien ?

victorjulien:
``dotprefix`` works the best for me as nothing in the code is actually specific to a domain. Only its intended use is around domains.

jlucovsky:
Closed in #4190

PR #4121, opened 2019-08-23

shivan1b:
ByteExtractString* functions.

### DIFF
--- a/src/detect-cipservice.c
+++ b/src/detect-cipservice.c
@@ -345,13 +345,21 @@ static DetectEnipCommandData *DetectEnipCommandParse(const char *rulestr)
     }
 
     unsigned long cmd = atol(rulestr);
+    if (ByteExtractStringUint64(&cmd, 10, 0, rulestr) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "invalid ENIP command"
+                ": \"%s\"", rulestr);
+        goto error;
+    }
     if (cmd > MAX_ENIP_CMD) //if command greater than 16 bit
     {
         SCLogError(SC_ERR_INVALID_SIGNATURE, "invalid ENIP command %lu", cmd);
         goto error;
     }
-
-    enipcmdd->enipcommand = (uint16_t) atoi(rulestr);
+    if (ByteExtractStringUint16(&enipcmdd->enipcommand, 10, 0, rulestr) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "invalid ENIP command"
+                ": \"%s\"", rulestr);
+        goto error;
+    }
 
     return enipcmdd;
 
--- a/src/detect-dce-iface.c
+++ b/src/detect-dce-iface.c
@@ -180,8 +180,12 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
             goto error;
         }
 
-        version = atoi(copy_str);
-        if (version > UINT16_MAX) {
+        if (ByteExtractStringInt32(&version, 10, 0, (const char *)copy_str) < 0) {
+            SCLogError(SC_ERR_INVALID_SIGNATURE, "DCE_IFACE interface version "
+                       "invalid: \"%s\n\"", copy_str);
+            goto error;
+        }
+        if (version < 0 || version > UINT16_MAX) {
             SCLogError(SC_ERR_INVALID_SIGNATURE, "DCE_IFACE interface version "
                        "invalid: %d\n", version);
             goto error;
--- a/src/detect-engine.c
+++ b/src/detect-engine.c
@@ -2342,7 +2342,15 @@ static int DetectEngineCtxLoadConf(DetectEngineCtx *de_ctx)
             }
 
             if (insp_recursion_limit != NULL) {
-                de_ctx->inspection_recursion_limit = atoi(insp_recursion_limit);
+                if (ByteExtractStringInt32(&de_ctx->inspection_recursion_limit,
+                    10, 0, (const char *)insp_recursion_limit) < 0) {
+                    SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for "
+                            "detect-engine.inspection-recursion-limit, "
+                            "resetting to %d",
+                            DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT);
+                    de_ctx->inspection_recursion_limit =
+                        DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT;
+                }
             } else {
                 de_ctx->inspection_recursion_limit =
                     DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT;
--- a/src/detect-modbus.c
+++ b/src/detect-modbus.c
@@ -51,6 +51,7 @@
 #include "detect-engine-modbus.h"
 
 #include "util-debug.h"
+#include "util-byte.h"
 
 #include "app-layer-modbus.h"
 
@@ -195,13 +196,28 @@ static DetectModbus *DetectModbusAccessParse(const char *str)
                 goto error;
 
             if (arg[0] == '>') {
-                modbus->address->min    = atoi((const char*) (arg+1));
+                if (ByteExtractStringUint16(&modbus->address->min, 10, 0,
+                                            (const char*) (arg + 1)) < 0) {
+                    SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
+                               "address: '%d'", arg[1]);
+                    goto error;
+                }
                 modbus->address->mode   = DETECT_MODBUS_GT;
             } else if (arg[0] == '<') {
-                modbus->address->min    = atoi((const char*) (arg+1));
+                if (ByteExtractStringUint16(&modbus->address->min, 10, 0,
+                                            (const char*) (arg + 1)) < 0) {
+                    SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
+                               "address: '%d'", arg[1]);
+                    goto error;
+                }
                 modbus->address->mode   = DETECT_MODBUS_LT;
             } else {
-                modbus->address->min    = atoi((const char*) arg);
+                if (ByteExtractStringUint16(&modbus->address->min, 10, 0,
+                                            (const char*) (arg)) < 0) {
+                    SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
+                               "address: '%d'", arg[0]);
+                    goto error;
+                }
             }
             SCLogDebug("and min/equal address %d", modbus->address->min);
 
@@ -213,7 +229,12 @@ static DetectModbus *DetectModbusAccessParse(const char *str)
                 }
 
                 if (*arg != '\0') {
-                    modbus->address->max    = atoi((const char*) (arg+2));
+                    if (ByteExtractStringUint16(&modbus->address->max, 10, 0,
+                                                (const char*) (arg + 2)) < 0) {
+                        SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for "
+                                "max address: '%d'", arg[2]);
+                        goto error;
+                    }
                     modbus->address->mode   = DETECT_MODBUS_RA;
                     SCLogDebug("and max address %d", modbus->address->max);
                 }
@@ -240,13 +261,28 @@ static DetectModbus *DetectModbusAccessParse(const char *str)
                         goto error;
 
                     if (arg[0] == '>') {
-                        modbus->data->min   = atoi((const char*) (arg+1));
+                        if (ByteExtractStringUint16(&modbus->data->min, 10, 0,
+                                                    (const char*) (arg + 1)) < 0) {
+                        SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
+                                "address: '%d'", arg[1]);
+                        goto error;
+                        }
                         modbus->data->mode  = DETECT_MODBUS_GT;
                     } else if (arg[0] == '<') {
-                        modbus->data->min   = atoi((const char*) (arg+1));
+                        if (ByteExtractStringUint16(&modbus->data->min, 10, 0,
+                                                    (const char*) (arg + 1)) < 0) {
+                        SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
+                                   "address: '%d'", arg[1]);
+                        goto error;
+                        }
                         modbus->data->mode  = DETECT_MODBUS_LT;
                     } else {
-                        modbus->data->min   = atoi((const char*) arg);
+                        if (ByteExtractStringUint16(&modbus->data->min, 10, 0,
+                                                    (const char*)arg) < 0) {
+                        SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
+                                   "address: '%d'", arg[0]);
+                        goto error;
+                        }
                     }
                     SCLogDebug("and min/equal value %d", modbus->data->min);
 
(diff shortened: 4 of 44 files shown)


shivan1b:
Hi, Jeff!
Could you please review this? Please pay special attention to `runmode-pcap`, `runmode-netmap`, `runmode-afpacket`, `runmode-pfring`, `util-napatech`.

src/detect-engine.c
```diff
@@ -2342,7 +2342,15 @@ static int DetectEngineCtxLoadConf(DetectEngineCtx *de_ctx)
...
             if (insp_recursion_limit != NULL) {
-                de_ctx->inspection_recursion_limit = atoi(insp_recursion_limit);
+                if (ByteExtractStringInt32(&de_ctx->inspection_recursion_limit,
+                    10, 0, (const char *)insp_recursion_limit) < 0) {
+                    SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for "
+                            "detect-engine.inspection-recursion-limit, "
+                            "resetting to %d",
+                            DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT);
+                    de_ctx->inspection_recursion_limit =
+                        DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT;
```
shivan1b: This causes test 2 of this module to fail. Here, with this change, no matter what `insp_recursion_limit` is, I am always setting it to `DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT`. However, earlier (i.e. in current master), for cases where `inspection-recursion- limit` used to be empty in `suricata.yaml`, the value was set to `0` which led to a further condition check later on (L2353 in master) and resetting the value to `-1` which (from my understanding) is invalid and causes the test to fail as it was expecting `-1`.
Question: Should the test be fixed or the value setting done above?

src/util-cpu.c
```diff
@@ -77,7 +78,16 @@ uint16_t UtilCpuGetNumProcessorsConfigured(void)
...
 	const char* envvar = getenv("NUMBER_OF_PROCESSORS");
-	nprocs = (NULL != envvar) ? atoi(envvar) : 0;
+    if (envvar != NULL) {
+        if (ByteExtractStringInt64(&nprocs, 10, 0, (const char *)envvar) < 0) {
+            SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for the number of "
+                       "cpus: '%s'", envvar);
+            return 0;
+        }
+    }
+    else {
```
jlucovsky: Suggest initializing `nprocs = 0`.

Remove the `else` corresponding to `if (envvar != NULL)`

src/util-cpu.c
```diff
@@ -77,7 +78,16 @@ uint16_t UtilCpuGetNumProcessorsConfigured(void)
 #elif OS_WIN32
 	long nprocs = -1;
 	const char* envvar = getenv("NUMBER_OF_PROCESSORS");
-	nprocs = (NULL != envvar) ? atoi(envvar) : 0;
+    if (envvar != NULL) {
+        if (ByteExtractStringInt64(&nprocs, 10, 0, (const char *)envvar) < 0) {
```
jlucovsky: Cast on `envar` not needed.

However, the first arg is expected to be `i64_t`, not `long`.

src/runmode-pfring.c
```diff
@@ -123,7 +123,11 @@ static void *OldParsePfringConfig(const char *iface)
         pfconf->threads = 1;
     } else {
         if (threadsstr != NULL) {
-            pfconf->threads = atoi(threadsstr);
+            if (ByteExtractStringUint16(&pfconf->threads, 10, 0, (const char *)threadsstr) < 0) {
+                SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for pfconf "
```
jlucovsky: Suggest changing the message ... the invalid value comes from `pfring.threads` ... `pfconf threads` is correct but `pfring.threads` would be more helpful.

src/runmode-pfring.c
```diff
@@ -123,7 +123,11 @@ static void *OldParsePfringConfig(const char *iface)
         pfconf->threads = 1;
     } else {
         if (threadsstr != NULL) {
-            pfconf->threads = atoi(threadsstr);
+            if (ByteExtractStringUint16(&pfconf->threads, 10, 0, (const char *)threadsstr) < 0) {
```
jlucovsky: Cast on `threadsstr` isn't needed.

Please check the other calls to `ByteExtract.*(` -- I suspect there are many cases where the cast isn't necessary.

src/runmode-pfring.c
```diff
@@ -141,7 +145,11 @@ static void *OldParsePfringConfig(const char *iface)
     } else if (ConfGet("pfring.cluster-id", &tmpclusterid) != 1) {
         SCLogError(SC_ERR_INVALID_ARGUMENT,"Could not get cluster-id from config");
     } else {
-        pfconf->cluster_id = (uint16_t)atoi(tmpclusterid);
+        if (ByteExtractStringUint16(&pfconf->cluster_id, 10, 0, (const char *)tmpclusterid) < 0) {
+            SCLogError(SC_ERR_INVALID_ARGUMENT, "Invalid value for cluster "
```
jlucovsky: `for cluster id` --> `for pfring.cluster-id`?

src/runmode-pfring.c
```diff
@@ -267,7 +279,11 @@ static void *ParsePfringConfig(const char *iface)
 
     /* command line value has precedence */
     if (ConfGet("pfring.cluster-id", &tmpclusterid) == 1) {
-        pfconf->cluster_id = (uint16_t)atoi(tmpclusterid);
+        if (ByteExtractStringUint16(&pfconf->cluster_id, 10, 0, (const char *)tmpclusterid) < 0) {
+            SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for "
```
jlucovsky: `for cluster id` -> `for pfring.cluster-id`?

src/runmode-pfring.c
```diff
@@ -283,7 +299,11 @@ static void *ParsePfringConfig(const char *iface)
             SCLogError(SC_ERR_INVALID_ARGUMENT,
                        "Could not get cluster-id from config");
         } else {
-            pfconf->cluster_id = (uint16_t)atoi(tmpclusterid);
+            if (ByteExtractStringUint16(&pfconf->cluster_id, 10, 0, (const char *)tmpclusterid) < 0) {
+                SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for "
+                             "cluster id: '%s'. Resetting to 1.", tmpclusterid);
```
jlucovsky: `cluster id` -> `cluster-id`?

src/runmode-pcap.c
```diff
@@ -144,7 +145,11 @@ static void *ParsePcapConfig(const char *iface)
         aconf->threads = 1;
     } else {
         if (threadsstr != NULL) {
-            aconf->threads = atoi(threadsstr);
+            if (ByteExtractStringInt32(&aconf->threads, 10, 0, (const char *)threadsstr) < 0) {
+                SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for aconf "
```
jlucovsky: `for aconf threads` -> `for pcap.threads`?

src/runmode-netmap.c
```diff
@@ -149,7 +149,9 @@ static int ParseNetmapSettings(NetmapIfaceSettings *ns, const char *iface,
             ns->threads = 0;
             ns->threads_auto = true;
         } else {
-            ns->threads = atoi(threadsstr);
+            if (ByteExtractStringUint16(&ns->threads, 10, 0, (const char *)threadsstr) < 0) {
+                goto finalize;
```
jlucovsky: 1. Suggest a message stating the configured value is invalid.
2. Not sure if jumping to `finalize` is correct.
3. `ns->threads` is indeterminate. Is a default value appropriate here?

src/runmode-napatech.c
```diff
@@ -169,7 +169,10 @@ static void *NapatechConfigParser(const char *device)
     }
 
     /* device+5 is a pointer to the beginning of the stream id after the constant nt portion */
-    conf->stream_id = atoi(device + 2);
+    if (ByteExtractStringUint16(&conf->stream_id, 10, 0, (const char *)(device + 2)) < 0) {
```
jlucovsky: I suggest changing the comment from `device+5` to `device+2` to reflect the code.

I suggest logging an error if the conversion fails.

src/detect-cipservice.c
```diff
@@ -345,13 +345,21 @@ static DetectEnipCommandData *DetectEnipCommandParse(const char *rulestr)
     }
 
     unsigned long cmd = atol(rulestr);
+    if (ByteExtractStringUint64(&cmd, 10, 0, rulestr) < 0) {
```
jlucovsky: Remove the call to `atol`

src/detect-dce-iface.c
```diff
@@ -180,8 +180,12 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
             goto error;
         }
 
-        version = atoi(copy_str);
-        if (version > UINT16_MAX) {
+        if (ByteExtractStringInt32(&version, 10, 0, (const char *)copy_str) < 0) {
+            SCLogError(SC_ERR_INVALID_SIGNATURE, "DCE_IFACE interface version "
+                       "invalid: \"%s\n\"", copy_str);
+            goto error;
```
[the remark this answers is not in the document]
jlucovsky: IMO, the code is now stricter so the test should be revisted. If the purpose of the test is to `assert version == 0` (which would've happened previously with atoi, then the test can be reworked.

src/detect-engine.c
jlucovsky: Good question. I'll defer to the others on this one.

victorjulien: It looks like the ```-1``` is used to disable it, but that doesn't make sense as a default value given that we are expected to set it to ```DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT```. Please update the tests to match the ```DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT``` value.

src/detect-modbus.c
```diff
@@ -195,13 +196,28 @@ static DetectModbus *DetectModbusAccessParse(const char *str)
                 goto error;
 
             if (arg[0] == '>') {
-                modbus->address->min    = atoi((const char*) (arg+1));
+                if (ByteExtractStringUint16(&modbus->address->min, 10, 0,
+                                            (const char*) (arg + 1)) < 0) {
+                    SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
+                               "address: '%d'", arg[1]);
```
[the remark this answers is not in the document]
victorjulien: Indeed. Print the string instead.

src/runmode-af-packet.c
```diff
@@ -195,7 +196,9 @@ static void *ParseAFPConfig(const char *iface)
             if (strcmp(threadsstr, "auto") == 0) {
                 aconf->threads = 0;
             } else {
-                aconf->threads = atoi(threadsstr);
+                if (ByteExtractStringInt32(&aconf->threads, 10, 0, (const char *)threadsstr) < 0) {
```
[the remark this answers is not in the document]
victorjulien: Threads is initialized at the start of the func.

I agree the ```goto finalized``` should not be used here. The rest of the config should still be handled.

victorjulien:
Thanks for the detailed review @jlucovsky

src/detect-modbus.c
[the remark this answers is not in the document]
shivan1b: So, it should just be `"Invalid value for min address."`, is this correct?

src/detect-nfs-version.c
```diff
@@ -282,7 +283,10 @@ static DetectNfsVersionData *DetectNfsVersionParse (const char *rawstr)
     }
 
     /* set the first value */
-    dd->lo = atoi(value1); //TODO
+    if (ByteExtractStringUint32(&dd->lo, 10, 0, (const char *)value1) < 0) { //TODO
```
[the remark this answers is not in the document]
shivan1b: @jlucovsky not sure what was this TODO for so I left it at its place.

src/util-log-redis.c
```diff
@@ -514,7 +515,11 @@ int SCConfLogOpenRedis(ConfNode *redis_node, void *lf_ctx)
         SCLogError(SC_ERR_MEM_ALLOC, "Error allocating redis server string");
         exit(EXIT_FAILURE);
     }
-    log_ctx->redis_setup.port = atoi(redis_port);
+    if (ByteExtractStringInt32(&log_ctx->redis_setup.port, 10, 0, (const char *)redis_port) < 0) {
+        SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for redis port: '%s'. "
+                   "Resetting to 6379.", redis_port);
+        log_ctx->redis_setup.port = 6379;
```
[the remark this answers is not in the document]
shivan1b: Could you please explain why exiting with failure is a better idea here?
Why I thought resetting was a good idea: Redis by default runs on `6379` and in the code as well on L465 (this branch) if the port is not given then it is set to `6379`. So, exiting if the port provided in conf was invalid probably didn't call for an exit.

src/util-cpu.c
shivan1b: Is this what you're suggesting?
 ```
   81     if (envvar != NULL) {
   82         if (ByteExtractStringInt64(&nprocs, 10, 0, (const char *)envvar) < 0) {
   83             SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for the number of "
   84                        "cpus: '%s'", envvar);
   85             return 0;
   86         }
   87     }
   88     nprocs = 0;
```

File: src/runmode-netmap.c (hunk 10 above)
Comment (shivan1b) [reply src/runmode-netmap.c]: > 1. Suggest a message stating the configured value is invalid.
OK
> 2. Not sure if jumping to `finalize` is correct.
> 3. `ns->threads` is indeterminate. Is a default value appropriate here?
all other cases use `0`, does it make sense to use that here as well. These conditional checks (L144-L158 in this branch) won't really make sense if we set the same value in every block.

File: src/detect-dce-iface.c (hunk 13 above)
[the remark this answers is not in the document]
Comment (shivan1b) [reply src/detect-dce-iface.c]: Turns out the test does a lot of things before checking if the version is 0.
I have understood the following.

**Aim of the test**
To check that uuid, version and options are parsed correctly from a given argument string which does not have a version specified.

**Procedure**
The function calls are as follows.
```
DetectDceIfaceTestParse06    ---->     DetectDceIfaceSetup   ---->    DetectDceIfaceArgParse
```
1. Since we `goto error` from the middle of the function `DetectDceIfaceArgParse`, the return value is `NULL` compared to a `DetectDceIfaceData *` under normal circumstances.
2. Since `DetectDceIfaceSetup` receives `NULL` from `DetectDceIfaceArgParse`, it returns `-1` (which is an indicator of failure) right away without doing a `SigMatch` allocation. `SigMatch` allocation and list entry happens if there was no error constructing `DetectDceIfaceData` struct.
3. The test as of now tends to fail because of (2) but it should ideally fail since we set `did` to `NULL` and there is nothing to check the `test_uuid` against. So, the entire point of test and keyword kinda fails.

**Solution (My proposal)**
Change the test argument string to something with version as that is not the sole purpose of this test.

**Caveat**
By putting this error condition we are enforcing the presence of version with dce_iface string which does not go as per docs or `PARSE_REGEX`. (As per docs and regex only `uuid` is required, `version` and `any_frag` option can be skipped)

**Question**
Do we want to enforce `version` in the `dce_iface` argument string? If yes, we should update the test, docs and regex. If no, version should be set to `0` in case of an invalid/ version string.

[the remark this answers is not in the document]
shivan1b: Yes. Just to double confirm, with the patch above, we are enforcing that we are not accepting argument string to `dce-iface` of the type `"12345678-1234-1234-1234-123456789ABC,any_frag"`, I hope its OK?
Acceptable strings can be:
`"12345678-1234-1234-1234-123456789ABC"`
`"12345678-1234-1234-1234-123456789ABC,<1,any_frag"`
`"12345678-1234-1234-1234-123456789ABC,<1"`
only.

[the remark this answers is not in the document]
shivan1b: I think I found a way to fix this. Thanks for all the assistance with this PR.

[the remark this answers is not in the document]
victorjulien: ```12345678-1234-1234-1234-123456789ABC,any_frag``` notation is used in 2 ETPRO rules. I can share them privately.

shivan1b:
Replaced with #4147

PR #4124, opened 2019-08-26
configure.ac: fix static build with pcap

ffontaine:
pcap can depends on nl-3 so use pkg-config to find these dependencies
otherwise all AC_CHECK_LIB calls will fail when building statically

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1311,6 +1311,7 @@
     AC_CHECK_HEADERS([pcap.h pcap/pcap.h pcap/bpf.h])
 
     LIBPCAP=""
+    PKG_CHECK_MODULES([PCAP],libpcap,[CPPFLAGS="${CPPFLAGS} ${PCAP_CFLAGS}" LIBS="${LIBS} ${PCAP_LIBS}"],[])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -1311,6 +1311,7 @@
     AC_CHECK_HEADERS([pcap.h pcap/pcap.h pcap/bpf.h])
 
     LIBPCAP=""
+    PKG_CHECK_MODULES([PCAP],libpcap,[CPPFLAGS="${CPPFLAGS} ${PCAP_CFLAGS}" LIBS="${LIBS} ${PCAP_LIBS}"],[])
```
victorjulien: should it look for ```pcap``` instead w/o the 'lib' prefix? Also, for our windows support we need it to look for ```${PCAP_LIB_NAME}```

ffontaine: Nope,` PKG_CHECK_MODULES` is used to retrieve the pkgconfig file which is named libpcap.pc, not pcap.pc (see https://github.com/the-tcpdump-group/libpcap/blob/master/libpcap.pc.in). Basically, this `PKG_CHECK_MODULES` call does not change anything in the way `LIBPCAP` is set, it just make sure that `AC_CHECK_LIB` calls don't fail due to missing dependencies when building statically.

PR #4125, opened 2019-08-27
install: install all files for events rules

catenacyber:
https://redmine.openinfosecfoundation.org/issues/2786

Describe changes:
- Makefile target `install-rules` now install all event rules files ie `rules/*-events.rules`

This patch only for the case where suricata-update is not installed

Modifies #3913 with not overwriting the already present files

I did not manage to have a generic way to list all the files


victorjulien:
Merged in #4139, thanks!

PR #4126, opened 2019-08-27
boyermoore: avoid one tolower call

catenacyber:
https://redmine.openinfosecfoundation.org/issues/1218

Describe changes:
- Updates `PreBmBcNocase` so that both upper and lower character give the same jump entry
- the characters had already been lowered by previous call to `BoyerMooreCtxToNocase`
- Updates `BoyerMooreNocase` so that we do not call `tolower` for the jump entries

For performance impact, I do not know if certain compilers managed to optimize away the second call to `tolower`

Modifies #3918 fixing the case where the needle has upper characters checked against lower characters

### DIFF
--- a/src/suricata.h
+++ b/src/suricata.h
@@ -178,6 +178,7 @@ extern uint16_t g_vlan_mask;
 
 #include <ctype.h>
 #define u8_tolower(c) tolower((uint8_t)(c))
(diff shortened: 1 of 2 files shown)


victorjulien:
Merged in #4139, thanks!

PR #4136, opened 2019-08-28
ftp: Add rust parsers for active mode port handling

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2916](https://redmine.openinfosecfoundation.org/issues/2916)

Describe changes:
- Remove C-language parsers for ports in active mode
- Add Rust parsers for active mode port handling.

### DIFF
--- a/rust/src/ftp/mod.rs
+++ b/rust/src/ftp/mod.rs
@@ -55,6 +55,42 @@ named!(pub ftp_pasv_response<u16>,
         )
 );
 
+// PORT 192,168,0,13,234,10
+named!(pub ftp_active_port<u16>,
+       do_parse!(
+            tag!("PORT") >>
+            ws!(digit) >> tag!(",") >> digit >> tag!(",") >>
+            digit >> tag!(",") >> digit >> tag!(",") >>
+            part1: verify!(getu16, |v| v <= std::u8::MAX as u16) >>
+            tag!(",") >>
+            part2: verify!(getu16, |v| v <= std::u8::MAX as u16) >>
+            tag!("\0") >>
+            (
+                part1 * 256 + part2
+            )
+        )
+);
+
+
+#[no_mangle]
+pub extern "C" fn rs_ftp_active_port(input: *const u8, len: u32) -> u16 {
+    let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
+    match ftp_active_port(buf) {
+        Ok((_, dport)) => {
+            return dport;
+        }
+        Err(nom::Err::Incomplete(_)) => {
+            let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
+            SCLogNotice!("port incomplete: '{:?}'", String::from_utf8_lossy(buf));
+        },
+        Err(_) => {
+            let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
+            SCLogNotice!("port error on '{:?}'", String::from_utf8_lossy(buf));
+        },
+    }
+    return 0;
+}
+
 
 #[no_mangle]
 pub extern "C" fn rs_ftp_pasv_response(input: *const u8, len: u32) -> u16 {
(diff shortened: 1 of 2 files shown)


rust/src/ftp/mod.rs
```diff
@@ -55,6 +55,42 @@ named!(pub ftp_pasv_response<u16>,
...
+        Ok((_, dport)) => {
+            return dport;
+        }
+        Err(nom::Err::Incomplete(_)) => {
+            let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
+            SCLogNotice!("port incomplete: '{:?}'", String::from_utf8_lossy(buf));
+        },
+        Err(_) => {
+            let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
+            SCLogNotice!("port error on '{:?}'", String::from_utf8_lossy(buf));
```
victorjulien: should be SCLogDebug

rust/src/ftp/mod.rs
```diff
@@ -55,6 +55,42 @@ named!(pub ftp_pasv_response<u16>,
...
+#[no_mangle]
+pub extern "C" fn rs_ftp_active_port(input: *const u8, len: u32) -> u16 {
+    let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
+    match ftp_active_port(buf) {
+        Ok((_, dport)) => {
+            return dport;
+        }
+        Err(nom::Err::Incomplete(_)) => {
+            let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
+            SCLogNotice!("port incomplete: '{:?}'", String::from_utf8_lossy(buf));
```
victorjulien: should be SCLogDebug

rust/src/ftp/mod.rs
```diff
@@ -55,6 +55,42 @@ named!(pub ftp_pasv_response<u16>,
...
+            (
+                part1 * 256 + part2
+            )
+        )
+);
+
+
+#[no_mangle]
+pub extern "C" fn rs_ftp_active_port(input: *const u8, len: u32) -> u16 {
+    let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
```
victorjulien: You can use the ```build_slice``` macro. E.g.
```rust
let slice: &[u8] = build_slice!(input, len as usize);
```

rust/src/ftp/mod.rs
```diff
@@ -88,6 +124,40 @@ named!(pub ftp_epsv_response<u16>,
...
+            tag!("|") >> tag!("\0") >>
+            (
+                port
+            )
+        )
+);
+
+#[no_mangle]
+pub extern "C" fn rs_ftp_active_eprt(input: *const u8, len: u32) -> u16 {
+    let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
```
victorjulien: build_slice

rust/src/ftp/mod.rs
```diff
@@ -88,6 +124,40 @@ named!(pub ftp_epsv_response<u16>,
...
+    match ftp_active_eprt(buf) {
+        Ok((_, dport)) => {
+            return dport;
+        },
+        Err(nom::Err::Incomplete(_)) => {
+            let buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
```
victorjulien: I wonder how Rust deals with this in the non-debug case: will this lead to warnings of unused vars?

If so, maybe it should be:
```rust
let _buf = unsafe{std::slice::from_raw_parts(input, len as usize)};
SCLogDebug!("eprt incomplete: '{:?}'", String::from_utf8_lossy(_buf));
```

src/app-layer-ftp.c
```diff
@@ -832,7 +769,7 @@ static int FTPParseResponse(Flow *f, void *ftp_state, AppLayerParserState *pstat
 
     if (state->command == FTP_COMMAND_PORT) {
         if ((flags & STREAM_TOCLIENT)) {
-            uint16_t dyn_port = FTPGetV4PortNumber(state->port_line, state->port_line_len);
+            uint16_t dyn_port = rs_ftp_active_port(state->port_line, state->port_line_len + 1);
```
victorjulien: Why is a +1 added?

rust/src/ftp/mod.rs
```diff
@@ -55,6 +55,42 @@ named!(pub ftp_pasv_response<u16>,
...
+// PORT 192,168,0,13,234,10
+named!(pub ftp_active_port<u16>,
+       do_parse!(
+            tag!("PORT") >>
+            ws!(digit) >> tag!(",") >> digit >> tag!(",") >>
+            digit >> tag!(",") >> digit >> tag!(",") >>
+            part1: verify!(getu16, |v| v <= std::u8::MAX as u16) >>
+            tag!(",") >>
+            part2: verify!(getu16, |v| v <= std::u8::MAX as u16) >>
+            tag!("\0") >>
```
victorjulien: Is this \0 on the wire? If not, it should not be parsed here.

rust/src/ftp/mod.rs
victorjulien: Can be done w/o the explicit type btw:
```rust
```

rust/src/ftp/mod.rs
victorjulien: Ah no, it should not be necessary.

rust/src/ftp/mod.rs
victorjulien: So removing the 'tag("\0")' condition and not passing +1 doesn't work?

rust/src/ftp/mod.rs
chifflier: @victorjulien , @jasonish : I think it is possible to enable only a subset of clippy. This would allow a smooth transition (enabling more warnings when code is ready) and can be checked in travis CI.

I'll have a look

chifflier: @victorjulien @jasonish : a simple check shows that there is ~ 130 functions that should be marked unsafe. I can fix that, but would you agree with a fix-everything patch fixing all these functions and adding a travis check ?

jasonish: @chifflier In general there are some clippy warnings I don't agree with.  Right now I'm more concerned with our usage of unsafe.  Would it be a good idea to mark all these functions as unsafe, instead of just around the pointer dereference?  I think it make sense to. Thanks.

jasonish: I'm indifferent to a fix everything approach.  For was initially thinking we'd do it just for new functions, then updating existing ones as we go.  But that doesn't make it harder to look through the clippy output.

rust/src/ftp/mod.rs
jlucovsky: That's correct. The passive equivalents of these functions terminate with a `)`

rust/src/ftp/mod.rs
chifflier: Clippy is quite flexible, so it can be used to check only the warnings you agree with. This is what I'm proposing to add, if you want.

Specifically on the unsafe functions, I tend to agree with clippy: functions dereferencing unsafe (extern C) pointers should be 1) minimal 2) marked as unsafe function, so the caller knows it.

jlucovsky:
Continued in #4151

PR #4137, opened 2019-08-28
ftp: Ensure non-zero command length with MPM init

jlucovsky:
This PR:
- Only initializes MPM with commands having non-zero command lengths.


victorjulien:
Merged in #4139, thanks!

PR #4147, opened 2019-09-01
Bug 3053/v3

shivan1b:

### DIFF
--- a/src/detect-byte-extract.c
+++ b/src/detect-byte-extract.c
@@ -238,8 +238,12 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                    "for arg 1 for byte_extract");
         goto error;
     }
-    bed->nbytes = atoi(nbytes_str);
-
+    if (ByteExtractStringUint8(&bed->nbytes, 10, 0,
+                               (const char *)nbytes_str) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for number of bytes"
+                   " to be extracted: \"%s\".", nbytes_str);
+        goto error;
+    }
     /* offset */
     char offset_str[64] = "";
     res = pcre_copy_substring((char *)arg, ov,
@@ -249,7 +253,11 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                    "for arg 2 for byte_extract");
         goto error;
     }
-    int offset = atoi(offset_str);
+    int offset;
+    if (ByteExtractStringInt32(&offset, 10, 0, (const char *)offset_str) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for offset: \"%s\".", offset_str);
+        goto error;
+    }
     if (offset < -65535 || offset > 65535) {
         SCLogError(SC_ERR_INVALID_SIGNATURE, "byte_extract offset invalid - %d.  "
                    "The right offset range is -65535 to 65535", offset);
@@ -305,7 +313,13 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                            "for arg %d for byte_extract", i);
                 goto error;
             }
-            int multiplier = atoi(multiplier_str);
+            int multiplier;
+            if (ByteExtractStringInt32(&multiplier, 10, 0,
+                                       (const char *)multiplier_str) < 0) {
+                SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for"
+                        "multiplier: \"%s\".", multiplier_str);
+                goto error;
+            }
             if (multiplier < DETECT_BYTE_EXTRACT_MULTIPLIER_MIN_LIMIT ||
                 multiplier > DETECT_BYTE_EXTRACT_MULTIPLIER_MAX_LIMIT) {
                 SCLogError(SC_ERR_INVALID_SIGNATURE, "multipiler_value invalid "
@@ -409,7 +423,12 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                            "for arg %d in byte_extract", i);
                 goto error;
             }
-            bed->align_value = atoi(align_str);
+            if (ByteExtractStringUint8(&bed->align_value, 10, 0,
+                                       (const char *)align_str) < 0) {
+                SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for"
+                        " align_str: \"%s\".", align_str);
+                goto error;
+            }
             if (!(bed->align_value == 2 || bed->align_value == 4)) {
                 SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid align_value for "
                            "byte_extract - \"%d\"", bed->align_value);
--- a/src/detect-dce-iface.c
+++ b/src/detect-dce-iface.c
@@ -128,7 +128,6 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
     int len = 0;
     char temp_str[3] = "";
     int version;
-
     ret = pcre_exec(parse_regex, parse_regex_study, arg, strlen(arg), 0, 0, ov,
                     MAX_SUBSTRINGS);
     if (ret < 2) {
@@ -180,8 +179,12 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
             goto error;
         }
 
-        version = atoi(copy_str);
-        if (version > UINT16_MAX) {
+        if (ByteExtractStringInt32(&version, 10, 0, (const char *)copy_str) < 0) {
+            SCLogError(SC_ERR_INVALID_SIGNATURE, "DCE_IFACE interface version "
+                       "invalid: \"%s\n\"", copy_str);
+            goto error;
+        }
+        if (version < 0 || version > UINT16_MAX) {
             SCLogError(SC_ERR_INVALID_SIGNATURE, "DCE_IFACE interface version "
                        "invalid: %d\n", version);
             goto error;
--- a/src/detect-dce-iface.c
+++ b/src/detect-dce-iface.c
@@ -169,8 +169,8 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
     if (ret == 3 || ret == 5)
         did->any_frag = 1;
 
-    /* if the regex has 4 or 5, version/operator is present in the signature */
-    if (ret == 4 || ret == 5) {
+    /* if the regex has 4 or 5, version/operator might be present in the signature */
+    if (ret == 4 || (ret == 5 && ov[4] != -1)) {
         /* first handle the version number, so that we can do some additional
          * validations of the version number, wrt. the operator */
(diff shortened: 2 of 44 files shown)


src/detect-byte-extract.c
```diff
@@ -409,7 +423,12 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                            "for arg %d in byte_extract", i);
                 goto error;
             }
-            bed->align_value = atoi(align_str);
+            if (ByteExtractStringUint8(&bed->align_value, 10, 0,
+                                       (const char *)align_str) < 0) {
+                SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for"
+                        " align_str: \"%s\".", align_str);
```
catenacyber: align_value rather than align_value ?

src/detect-dce-iface.c
```diff
@@ -128,7 +128,6 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
     int len = 0;
     char temp_str[3] = "";
     int version;
-
```
catenacyber: nit

catenacyber:
Why is commit https://github.com/OISF/suricata/pull/4147/commits/99e5f56b100cc83c7b0d4b83141bbf6e4836f901 part of this PR ?

shivan1b:
> Why is commit [99e5f56](https://github.com/OISF/suricata/commit/99e5f56b100cc83c7b0d4b83141bbf6e4836f901) part of this PR ?

Mistake. Redoing. Thanks for pointing.

shivan1b:
> Why is commit [99e5f56](https://github.com/OISF/suricata/commit/99e5f56b100cc83c7b0d4b83141bbf6e4836f901) part of this PR ?

Oops thought it was a different commit. If its not added here, the error logging in case version turns out to be empty after a ByteExtract call, the tests fail which would be incorrect.

src/detect-dce-iface.c
```diff
@@ -180,8 +179,12 @@ static DetectDceIfaceData *DetectDceIfaceArgParse(const char *arg)
             goto error;
         }
 
-        version = atoi(copy_str);
-        if (version > UINT16_MAX) {
+        if (ByteExtractStringInt32(&version, 10, 0, (const char *)copy_str) < 0) {
```
catenacyber: Cannot we do `ByteExtractStringUInt16`instead of `ByteExtractStringInt32` ? and save the letter check ?

src/detect-engine-address.c
```diff
@@ -571,7 +572,8 @@ int DetectAddressParseString(DetectAddress *dd, const char *str)
                         goto error;
                 }
 
-                int cidr = atoi(mask);
+                int cidr;
+                ByteExtractStringInt32(&cidr, 10, 0, (const char *)mask);
```
catenacyber: We should check the return value (and in the next one)

src/detect-engine-port.c
```diff
@@ -1418,14 +1419,24 @@ DetectPort *PortParse(const char *str)
         port2[0] = '\0';
         port2++;
 
-        if(DetectPortIsValidRange(port))
-            dp->port = atoi(port);
-        else
-            goto error;
+        if (strcmp(port, "") != 0) {
+            if (DetectPortIsValidRange(port)) {
+                if (ByteExtractStringUint16(&dp->port, 10, 0,
```
catenacyber: `DetectPortIsValidRange` and `ByteExtractStringUint16` are redundant
You could rather modify `DetectPortIsValidRange` to use `ByteExtractStringUint16` rather than `strtol`

src/detect-engine.c
```diff
@@ -4297,7 +4305,7 @@ static int DetectEngineTest02(void)
     if (de_ctx == NULL)
         goto end;
 
-    result = (de_ctx->inspection_recursion_limit == -1);
+    result = (de_ctx->inspection_recursion_limit == 3000);
```
catenacyber: Why this value ?

src/detect-iprep.c
```diff
@@ -321,7 +322,8 @@ int DetectIPRepSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawstr)
     }
 
     if (value != NULL && strlen(value) > 0) {
-        int ival = atoi(value);
+        int ival;
+        ByteExtractStringInt32(&ival, 10, 0, (const char *)value);
```
catenacyber: Check return value

src/detect-lua.c
```diff
@@ -278,7 +279,14 @@ int DetectLuaMatchBuffer(DetectEngineThreadCtx *det_ctx,
...
                 if (strcmp(k, "retval") == 0) {
-                    if (atoi(v) == 1)
+                    int val;
+                    if (ByteExtractStringInt32(&val, 10, 0,
+                                               (const char *)v) < 0) {
+                        SCLogError(SC_ERR_INVALID_VALUE, "Invalid value in "
+                                   "the table: '%s'", v);
+                        ret = 0;
+                    }
+                    if (val == 1)
```
catenacyber: That can be `else if`

src/reputation.c
```diff
@@ -251,7 +252,8 @@ static int SRepCatSplitLine(char *line, uint8_t *cat, char *shortname, size_t sh
 
     SCLogDebug("%s, %s", ptrs[0], ptrs[1]);
 
-    int c = atoi(ptrs[0]);
+    int c;
+    ByteExtractStringInt32(&c, 10, 0, (const char *)ptrs[0]);
```
catenacyber: Check return value

src/runmode-af-packet.c
```diff
@@ -195,7 +196,11 @@ static void *ParseAFPConfig(const char *iface)
             if (strcmp(threadsstr, "auto") == 0) {
                 aconf->threads = 0;
             } else {
-                aconf->threads = atoi(threadsstr);
+                if (ByteExtractStringInt32(&aconf->threads, 10, 0, (const char *)threadsstr) < 0) {
+                    SCLogError(SC_ERR_INVALID_VALUE, "Invalid number of "
```
catenacyber: Should this be a warning ? (as we seem to go on)

src/runmode-af-packet.c
```diff
@@ -290,7 +295,9 @@ static void *ParseAFPConfig(const char *iface)
     if (ConfGetChildValueWithDefault(if_root, if_default, "cluster-id", &tmpclusterid) != 1) {
         aconf->cluster_id = (uint16_t)(cluster_id_auto++);
     } else {
-        aconf->cluster_id = (uint16_t)atoi(tmpclusterid);
+        if (ByteExtractStringInt32(&aconf->cluster_id, 10, 0, (const char *)tmpclusterid) < 0) {
+            goto finalize;
```
catenacyber: Should we set `cluster_id = 0` ?

catenacyber:
@shivan1b In your tests, did you try to see where `ByteExtractStringSomething` failed ? (and `atoi` did not)

catenacyber:
> If its not added here, the case for dce-iface where version is not provided as a part of argument string, the tests fail which would be incorrect.

Thanks, but the link is still not obvious
That may deserve a comment in the code
Can you point me where `atoi` gets called so that it makes the test fail ?

shivan1b:
> @shivan1b In your tests, did you try to see where `ByteExtractStringSomething` failed ? (and `atoi` did not)

@catenacyber yes. It failed at 5 places in the files `src/detect-dce-iface.c`, `src/detect-engine-port.c` and `src/detect-engine.c`. I have fixed all those tests with the appropriate values as per the discussion with Jeff and Victor.
You can check the positions where it failed in #4101 . I have marked them as "Position 1", "Position 2", ...

src/detect-byte-extract.c
shivan1b: Yes, this should be align_value. Fixing in the next PR.

src/detect-dce-iface.c
shivan1b: Yes, we could. I'll declare version as `uint16_t` then else the compiler will throw warnings.

src/detect-engine-address.c
shivan1b: hmm Jeff suggested that since the very next line of code is checking the range of cidr, we could drop the check. Would you suggest otherwise?

src/detect-engine-port.c
shivan1b: Makes sense.

src/detect-dce-iface.c
catenacyber: If we could, then I guess we should :-)

src/detect-engine.c
shivan1b: One of the failing tests was this. Now that we are checking the value of `inspection_resursion_limit` with `ByteExtract*`, whenever the retval is `< 0`, we reset the `inspection_recursion_limit` to its default compared to 0 earlier.

src/detect-lua.c
shivan1b: OK.

src/runmode-af-packet.c
shivan1b: Yes, this should be. My fault.

src/detect-engine-address.c
catenacyber: > Would you suggest otherwise?

Yes, I suggest :
`if (ByteExtractStringInt32(&cidr, 10, 0, (const char *)mask) < 0 || cidr > 128)`

I think it is more readable to keep using `ByteExtractStringInt32` and such in the same way, ie control their return value.

Furthermore, we will use an uninitialized value if we do not check `ByteExtractStringInt32` return value as it will not always set `cidr`

src/runmode-af-packet.c
shivan1b: Yes, maybe. goto is making it skip every other config.

src/detect-engine.c
catenacyber: So, if I understand correctly, should we not set `result` when `ByteExtract` fails and not use `de_ctx->inspection_recursion_limit` anymore ?

Otherwise, we should at least use `DETECT_ENGINE_DEFAULT_INSPECTION_RECURSION_LIMIT` instead of plain `3000`

src/detect-engine-address.c
shivan1b: OK. Adding this in the follow up PR

src/detect-lua.c
shivan1b: hey @catenacyber now that I was making changes, I do not understand how this can be `else if`. Could you please explain?

catenacyber: `val` can only be 1 if `ByteExtractStringInt32` succeeded

So we do not need to test this condition in the case `ByteExtractStringInt32` failed, so we can put an else

shivan1b:
> > If its not added here, the case for dce-iface where version is not provided as a part of argument string, the tests fail which would be incorrect.
> 
> Thanks, but the link is still not obvious
> That may deserve a comment in the code
> Can you point me where `atoi` gets called so that it makes the test fail ?

@catenacyber it failed for this case https://github.com/OISF/suricata/blob/master/src/detect-dce-iface.c#L634
This test gets to the atoi call here: https://github.com/OISF/suricata/blob/master/src/detect-dce-iface.c#L183 and in case of an empty string for version, atoi converts it to 0.

shivan1b:
Replaced with #4168

PR #4148, opened 2019-09-01
detect: Add missing keyword URLs and description

shivan1b:
Add missing keyword URLs and their description. Fix the ones that
were incorrect.

Partially closes redmine ticket #2974.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2974


src/detect-app-layer-event.c
```diff
@@ -66,6 +66,8 @@ static int g_applayer_events_list_id = 0;
 void DetectAppLayerEventRegister(void)
 {
     sigmatch_table[DETECT_AL_APP_LAYER_EVENT].name = "app-layer-event";
+    sigmatch_table[DETECT_AL_APP_LAYER_EVENT].desc = "Match on events generated by the App Layer Parsers and the protocol detection engine.";
```
jlucovsky: nit: I'm pretty sure I had suggested the use of `Match` instead of `match` but I think that was incorrect advice as none of the existing `desc` values start with an uppercase letter. I suggest we follow convention to be consistent and thus, suggest using `match`

shivan1b: Thanks, Jeff. Shall address this in the next PR.

victorjulien: ```git grep '\.desc ='``` shows that no trailing dot is the common pattern, so lets remove it here too.

victorjulien: Ah I missed that this PR adds them. I have no objection to this and will be merging this PR. Could you do a follow up PR that changes all existing ones? There are quite few still in the old notation.

victorjulien:
Merged in #4203, thanks Shivani! Please follow up on https://github.com/OISF/suricata/pull/4148#discussion_r324588919

PR #4151, opened 2019-09-02
ftp: Add rust parsers for active mode port handling 

jlucovsky:
Continuation of #4136

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2916](https://redmine.openinfosecfoundation.org/issues/2916)

Describe changes:
- Addresses review comments
- Adds unit tests in Rust module.

### DIFF
--- a/rust/src/ftp/mod.rs
+++ b/rust/src/ftp/mod.rs
@@ -55,6 +55,40 @@ named!(pub ftp_pasv_response<u16>,
         )
 );
 
+// PORT 192,168,0,13,234,10
+named!(pub ftp_active_port<u16>,
+       do_parse!(
+            tag!("PORT") >>
+            ws!(digit) >> tag!(",") >> digit >> tag!(",") >>
+            digit >> tag!(",") >> digit >> tag!(",") >>
+            part1: verify!(getu16, |v| v <= std::u8::MAX as u16) >>
+            tag!(",") >>
+            part2: verify!(getu16, |v| v <= std::u8::MAX as u16) >>
+            tag!("\0") >>
+            (
+                part1 * 256 + part2
+            )
+        )
+);
+
+
+#[no_mangle]
+pub extern "C" fn rs_ftp_active_port(input: *const u8, len: u32) -> u16 {
+    let buf = build_slice!(input, len as usize);
+    match ftp_active_port(buf) {
+        Ok((_, dport)) => {
+            return dport;
+        }
+        Err(nom::Err::Incomplete(_)) => {
+            SCLogDebug!("port incomplete: '{:?}'", String::from_utf8_lossy(buf));
+        },
+        Err(_) => {
+            SCLogDebug!("port error on '{:?}'", String::from_utf8_lossy(buf));
+        },
+    }
+    return 0;
+}
+
 
 #[no_mangle]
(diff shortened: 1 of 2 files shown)


rust/src/ftp/mod.rs
```diff
@@ -55,6 +55,40 @@ named!(pub ftp_pasv_response<u16>,
...
+// PORT 192,168,0,13,234,10
+named!(pub ftp_active_port<u16>,
+       do_parse!(
+            tag!("PORT") >>
+            ws!(digit) >> tag!(",") >> digit >> tag!(",") >>
+            digit >> tag!(",") >> digit >> tag!(",") >>
+            part1: verify!(getu16, |v| v <= std::u8::MAX as u16) >>
+            tag!(",") >>
+            part2: verify!(getu16, |v| v <= std::u8::MAX as u16) >>
+            tag!("\0") >>
```
victorjulien: I can't get over this. It's not on the wire so it should not be part of this parser.

jlucovsky: Note that the buffer being inspected was created by the ftp parser: https://github.com/OISF/suricata/blob/master/src/app-layer-ftp.c#L665:L666

victorjulien: I don't see how that is relevant. The parser shouldn't be parsing things that were artificially added. Esp not if the only reason is to hide a problem in the parser.

victorjulien: @jasonish can you help with this? When you remove the \0 check the test fails with Incomplete(Size(1)). I think the getu16 macro emits this but not 100% sure.

jasonish: This is due to the end of the string (eof) causing it to return incomplete, as just maybe you wanted to parse more, or something like that.  I saw that @chifflier ran into this issue with nom4 and his ssh parser, but nom5 cleans it up a little.

It looks like the fix is to tell that parser that the string is considered complete, so if it hits eof while parsing the last item, it will be OK, and not fail. Unfortunately its a bit invasive to fix.  I did come up with the following new version of `ftp_active_port` that will work:

```
use nom::types::CompleteByteSlice

named!(parse_digits<CompleteByteSlice, &[u8]>,
    map!(take_while!(nom::is_digit), |b| b.0));

named!(parse_u16<CompleteByteSlice, u16>,
    map_res!(map_res!(parse_digits, str::from_utf8), u16::from_str));

named!(pub ftp_active_port1<CompleteByteSlice, u16>,
       do_parse!(
            tag!("PORT") >>
            ws!(digit) >> tag!(",") >> digit >> tag!(",") >>
            digit >> tag!(",") >> digit >> tag!(",") >>
            part1: verify!(parse_u16, |v| v <= std::u8::MAX as u16) >>
            tag!(",") >>
            part2: verify!(parse_u16, |v| v <= std::u8::MAX as u16) >>
            (
                part1 * 256 + part2
            )
        )
);
```

and the associated test:

```
    #[test]
    fn test_active_port_valid() {
        let port = ftp_active_port1(CompleteByteSlice("PORT 192,168,0,13,234,10".as_bytes()));
        assert_eq!(port, Ok((CompleteByteSlice(&b""[..]), 59914)));
    }
```

jlucovsky:
Continued in #4164

PR #4152, opened 2019-09-02
protodetect: continue when multiple patterns match

catenacyber:
This PR is for discussion rather than straight merging.

Describe changes:
- Keep searching when multiple patterns are found for protocol detection

For a proof of concept, this can happen for instance with SSH and SMB

We should not pick randomly one of the protocols.
But I am not sure about what we should do
Should we try the probing parsers of the protocols ? Try to fully parse ? Defer until more data has come ? Set an app layer event ?

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -1503,7 +1503,7 @@ AppProto AppLayerProtoDetectGetProto(AppLayerProtoDetectThreadCtx *tctx,
         AppProto pm_results[ALPROTO_MAX];
         uint16_t pm_matches = AppLayerProtoDetectPMGetProto(tctx, f,
                 buf, buflen, direction, pm_results, reverse_flow);
-        if (pm_matches > 0) {
+        if (pm_matches == 1) {
             alproto = pm_results[0];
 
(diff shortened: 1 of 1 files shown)


victorjulien:
It seems fair to take the list of candidates (so pm_results) and run the probing parsers for those if there are any. That might give a more definitive result. If not I think indeed set an event and go with the current default behaviour or using the first result

catenacyber:
> It seems fair to take the list of candidates (so pm_results) and run the probing parsers for those if there are any.

It seems fair indeed.
My problem is that the protocols with pattern matchers for detection do not all have probing parsers.
cf https://redmine.openinfosecfoundation.org/issues/2757#note-15

So, should we first add probing parsers for every protocol ?

victorjulien:
> > It seems fair to take the list of candidates (so pm_results) and run the probing parsers for those if there are any.
> 
> It seems fair indeed.
> My problem is that the protocols with pattern matchers for detection do not all have probing parsers.
> cf https://redmine.openinfosecfoundation.org/issues/2757#note-15
> 
> So, should we first add probing parsers for every protocol ?

I don't think this is needed. If a protocol has a PP, we eval that and if that matches it wins over a protocol that does not have a PP.

PR #4153, opened 2019-09-02
rules: move old way of install rules to new Makefile target

jasonish:
Stop falling back to the old method of installing rules into
/etc/suricata/rules if Suricata-Update is not available.

The goal here is to move away from the behaviour of installing
rules to /etc/suricata/rules as part of the default install
process. The engine provided rules are already installed to
/usr/share/suricata/rules, which can then be used as input
to rule management tools such as Suricata-Update.

This does not change the behaviour for Suricata release users
with the bundled Suricata-Update, but does expose the old
behaviour under a new Makefile target: install-rules-legacy.

**Note**: I'm really not sure about leaving the legacy behaviour here at all.  With 5.0 I feel like its time to move on and only put rule files in /usr/share/suricata/rules, and optionally run Suricata-Update.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3138

PRscript output:
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/361
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/716

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,6 +37,12 @@ if HAVE_SURICATA_UPDATE
 		--suricata-conf $(DESTDIR)$(sysconfdir)/suricata/suricata.yaml \
 		--no-test --no-reload
 else
+	echo "error: rules not installed as suricata-update not available"
+	exit 1
+endif
+
+# The old way of installing rules, before Suricata-Update was available.
+install-rules-legacy:
 	install -d "$(DESTDIR)$(e_sysconfrulesdir)"
 if HAVE_FETCH_COMMAND
 if HAVE_WGET_COMMAND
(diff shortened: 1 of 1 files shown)


victorjulien:
Yeah I wonder if we need the legacy feature as well. What if we just removed it and bring it back if we need it anyway for some reason?

jasonish:
> Yeah I wonder if we need the legacy feature as well. What if we just removed it and bring it back if we need it anyway for some reason?

I don't think we do.  Its kind of a broken, as when upgrading it doesn't update the engine files. Its easy enough to suggest to manually copy from the the /usr/share directory for those who need it still, which is also the way you'd have to get updated files into /etc on upgrade anyways.

Will make a new PR for this.

PR #4163, opened 2019-09-03
compile: checks libhtp lzma support

catenacyber:
none

Describe changes:
- Checks that libhtp has lzma decompression support

In the case of a non bundled version, we run a program creating a decompressor with `HTP_COMPRESSION_LZMA` (this checks if the version is recent enough) and then checks if it is not in passthrough mode (in case the recent version was compiled on a machine without liblzma)
In the case of bundled version, we just check that the version is recent enough to have HTP_COMPRESSION_LZMA in headers (liblzma check is already done).


victorjulien:
CI fails on this, can you check why?

PR #4166, opened 2019-09-04
Feature/datasets/v43

victorjulien:
Rebased version of #4047. For merge as experimental feature.

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/301
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/302


doc/userguide/rules/datasets.rst
```diff
@@ -0,0 +1,152 @@
...
+
+    dataset:<cmd>,<name>,<options>;
+
+    dataset:<set|isset|isnotset>,<name> \
+        [, type <string|md5|sha256>, save <file name>, load <file name>, state <file name>];
+
+type <type>
+  the data type: string, md5, sha256
+save <file name>
+  file name for saving the in memory data when Suricata exits
```
pevma: `the in` is probably in reverse order

doc/userguide/rules/datasets.rst
```diff
@@ -0,0 +1,152 @@
...
+load <file name>
+  file name for load the data when Suricata starts up
+state
+  sets both 'save' and 'load' to the same value
```
jasonish: When would you want to use a different combination of the save, load and state?

jasonish: Is an initial `load` file something we could see a rule distributor releasing?

doc/userguide/rules/datasets.rst
```diff
@@ -0,0 +1,152 @@
...
+  Name of an already defined dataset
+type
+  Data type: string, md5, sha256
+data
+  Data to add in serialized form (base64 for string, hex notation for md5/sha256)
+
+Example adding 'google.com' to set 'myset'::
+
+    dataset-add myset string Z29vZ2xlLmNvbQ==
+
```
jasonish: Why as base64? Can we not do that conversion on the Suricata side of the socket?

Also for hex notation, is a leading 0x required?

doc/userguide/rules/datasets.rst
```diff
@@ -0,0 +1,152 @@
...
+Example of sets for tracking unique values::
+
+    datasets:
+     - ua-seen:
+       type: string
+       state: ua-seen.lst
+     - dns-sha256-seen:
+       type: sha256
+       state: dns-sha256-seen.lst
+
```
jasonish: I don't think we should use lists here. Instead:
```
datasets:
  ua-seen:
    type: string
    state: ua-seen.lst
  dns-sha256-seen:
    type: sha256
    state: dns-sha256-seen.lst
```

If the list does make more sense, then it should look like:
```
datasets:
  - name: ua-seen
    type: string
    state: ua-seen.lst
  - name: dns-sha256-seen
    type: sha256
    state: dns-sha256-seen.lst
```

src/datasets.c
```diff
@@ -0,0 +1,990 @@
...
+
+    if (PathIsAbsolute(in_path)) {
+        strlcpy(path, in_path, sizeof(path));
+        strlcpy(out_path, path, out_size);
+        return;
+    }
+
+    const char *data_dir = ConfigGetDataDirectory();
+    if ((ret = stat(data_dir, &st)) != 0) {
+        SCLogNotice("data-dir '%s': %s", data_dir, strerror(errno));
```
jasonish: Probably an error here? Perhaps even failure, as something I explicitly configured is not going to be active.

src/datasets.c
```diff
@@ -0,0 +1,990 @@
...
+                if (dset == NULL)
+                    FatalError(SC_ERR_FATAL, "die: no dset for %s", set_name);
+                SCLogDebug("dataset %s: id %d type %s", set_name, n, set_type->val);
+                n++;
+            }
+
+            list_pos++;
+        }
+    }
+    SCLogNotice("datasets done: %p", datasets);
```
jasonish: Debug?

src/datasets.c
```diff
@@ -0,0 +1,990 @@
...
+
+    SCLogNotice("dataset: %s loading from '%s'", set->name, set->load);
+
+    FILE *fp = fopen(set->load, "r");
+    if (fp == NULL) {
+        SCLogError(SC_ERR_DATASET, "fopen '%s' failed: %s",
+                set->load, strerror(errno));
+        return -1;
+    }
+
```
jasonish: When the filename comes from state, I think its very likely the file may not exist yet, and should probably be created. I would expect an error on load though.

configure.ac
```diff
@@ -2532,6 +2532,7 @@ else
     EXPAND_VARIABLE(sysconfdir, e_sysconfrulesdir, "/suricata/rules")
     EXPAND_VARIABLE(localstatedir, e_localstatedir, "/run/suricata")
     EXPAND_VARIABLE(datadir, e_datarulesdir, "/suricata/rules")
+    EXPAND_VARIABLE(localstatedir, e_datadir, "/lib/suricata/data")
```
jasonish: This directory appears to have 2 usages.  One as a place to hold datasets that may be distributed as part of a ruleset, but the other is more like a cache, for example the state files.

I like to separate the 2, so I can `rm -rf` the cache type files to reset my state, but not lose files that may be required to startup.

It would also be nice to set some guidelines for people who wish to distribute rules that make use of datasets.

doc/userguide/rules/datasets.rst
victorjulien: Maybe I should have written 'in-memory'.

doc/userguide/rules/datasets.rst
victorjulien: We are working with binary data, so base64 seemed appropriate. It's in files too, not just the socket.

victorjulien: No leading 0x is required or supported. It's the output that ```md5sum``` and similar use.

doc/userguide/rules/datasets.rst
victorjulien: Good point: https://redmine.openinfosecfoundation.org/issues/3143

src/datasets.c
victorjulien: Fixing this and others.

doc/userguide/rules/datasets.rst
victorjulien: You can either use load, load and save, or state. Keyword parsing will block the rest.

Just a load would indeed be what I expect to be used by a rule vendor.

State would be used in case you use either rules to fill a set (dataset + set operator) or if you manage it over unix socket.

Save is something you can use to just extract data. E.g. get all unique UAs.

src/datasets.c
victorjulien: ConfigGetDataDirectory will return a default if not configured. Error will be handled later when the file is actually opened.

src/datasets.c
victorjulien: Yeah good catch. Fixing this and adding a SV test.

doc/userguide/rules/datasets.rst
jasonish: I wonder if the default directories should differ for a set vs, state. Think of the case where a ruleset tarball contains a rule, and an associated dataset, I think the best place to put that dataset is in the rule directory itself, or in a sub-directory of the rule directory named `datasets`.  In fact, that might be a good convention for rule publishers to follow as well.

The rule would then contain `dataset: load datasets/myset.lst`.  The rule parser would then look for this relative to the directory the rule is being loaded from.

Its a little less clear what to do with save, and state, and its likely those won't be used by rule publishers.

PR #4176, opened 2019-09-05
WIP -- Rules analyzer 

jlucovsky:
*Request for comments*

This is a prototype implementation that attempts to better handle the list of buffers to work with `--engine-analysis`. 

The goal was to
- Collect the buffer names in one place
- Refactor buffer handling
- Reduce issues due to missing buffers by highlighting the buffers being handled.

The purpose of this PR is to gather comments on the approach.

### DIFF
--- a/src/detect-engine-analyzer.c
+++ b/src/detect-engine-analyzer.c
@@ -859,6 +859,68 @@ void EngineAnalysisRules2(const DetectEngineCtx *de_ctx, const Signature *s)
     SCReturn;
 }
 
+typedef struct analyzer_item {
+    uint32_t    have_bufcount;
+    int32_t   item_id;
+    bool check_encoding_match;
+    bool global;
+    const char  *item_name;
+    const char  *display_name;
+} analyzer_t;
+
+analyzer_t AnalyzerItems[] = {
+    /* request keywords */
+    { 0, 0,  true, false, "http_uri", "http uri" },
+    { 0, 0,  false, false, "http_raw_uri", "http raw uri" },
+    { 0, 0,  false, true, "http_method", "http method" },
+    { 0, 0,  false, false, "http_request_line", "http request line" },
+    { 0, 0,  false, false, "http_client_body", "http client body" },
+    { 0, 0,  true, false, "http_header", "http header" },
+    { 0, 0,  false, false, "http_raw_header", "http raw header" },
+    { 0, 0,  true, false, "http_cookie", "http cookie" },
+    { 0, 0,  false, false, "http_user_agent", "http user agent" },
+    { 0, 0,  false, false, "http_host", "http host" },
+    { 0, 0,  false, false, "http_raw_host", "http raw host" },
+    { 0, 0,  false, false, "http_accept_enc", "http accept enc" },
+    { 0, 0,  false, false, "http_referer", "http referer" },
+    { 0, 0,  false, false, "http_content_type", "http content type" },
+    { 0, 0,  false, false, "http_header_names", "http header names" },
+
+    /* response keywords not listed above */
+    { 0, 0,  false, false, "http_stat_msg", "http stat msg" },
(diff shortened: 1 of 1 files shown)


src/detect-engine-analyzer.c
```diff
@@ -859,6 +859,68 @@ void EngineAnalysisRules2(const DetectEngineCtx *de_ctx, const Signature *s)
     SCReturn;
 }
 
+typedef struct analyzer_item {
```
victorjulien: style: UseThisStyleForStructNames

src/detect-engine-analyzer.c
```diff
@@ -859,6 +859,68 @@ void EngineAnalysisRules2(const DetectEngineCtx *de_ctx, const Signature *s)
...
+    { 0, 0,  false, false, "http_raw_host", "http raw host" },
+    { 0, 0,  false, false, "http_accept_enc", "http accept enc" },
+    { 0, 0,  false, false, "http_referer", "http referer" },
+    { 0, 0,  false, false, "http_content_type", "http content type" },
+    { 0, 0,  false, false, "http_header_names", "http header names" },
+
+    /* response keywords not listed above */
+    { 0, 0,  false, false, "http_stat_msg", "http stat msg" },
+    { 0, 0,  false, false, "http_stat_code", "http stat code" },
+    { 0, 0,  false, true, "file_data", "http server body"},
```
victorjulien: file_data can also refer to smb file xfer, nfs file xfer, smtp attachments

src/detect-engine-analyzer.c
```diff
@@ -859,6 +859,68 @@ void EngineAnalysisRules2(const DetectEngineCtx *de_ctx, const Signature *s)
...
+
+    /* missing request keywords */
+    { 0, 0,  false, false, "http_request_line", "http request line" },
+    { 0, 0,  false, false, "http_accept", "http accept" },
+    { 0, 0,  false, false, "http_accept_lang", "http accept lang" },
+    { 0, 0,  false, false, "http_connection", "http connection" },
+    { 0, 0,  false, false, "http_content_len", "http content len" },
+    { 0, 0,  false, false, "http_protocol", "http response line" },
```
victorjulien: wrong description

src/detect-engine-analyzer.c
```diff
@@ -859,6 +859,68 @@ void EngineAnalysisRules2(const DetectEngineCtx *de_ctx, const Signature *s)
...
+    { 0, 0,  false, false, "http_start", "http start" },
+
+    /* missing response keywords; some are listed above*/
+    { 0, 0,  false, false, "http_response_line", "http response line" },
+    { 0, 0,  false, false, "http.server", "http response line" },
+    { 0, 0,  false, false, "http.location", "http response line" },
+};
+
+/* Map items to AnalyzerItems array; the key is the item id */
+int32_t AnalyzerItemMap[300];
```
victorjulien: where does the 300 come from?

src/detect-engine-analyzer.c
jlucovsky: > I like this approach. Much cleaner.

Great ... i'll move forward with this. This was a "POC"; I'll make sure your comments are addressed in the next pr.

src/detect-engine-analyzer.c
jlucovsky: Ad-hoc value for the POC. The size of this array is should be large enough to include the id space returned by `DetectBufferTypeGetByName`

src/detect-engine-analyzer.c
jlucovsky: will fix.

src/detect-engine-analyzer.c
jlucovsky: In this module, it's only be used as an alias for the http server body buffer.

jlucovsky:
Continued in #4181

PR #4174, opened 2019-09-05
protocol parser: rdp

lmzke:
Continuation of #4131.

Initial implementation of feature [2314](https://redmine.openinfosecfoundation.org/issues/2314):
1. Add protocol parser for RDP
2. Add transactions for RDP negotiation
3. Add eve logging of transactions

This PR has incorporated all previous reviewer suggestions.  New contributor, all feedback welcome.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Sample output, based on [wireshark RDP-002.pcap](https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=view&target=RDP-002.pcap.gz):

```json
"rdp": {
  "tx_id": 0,
  "event_type": "initial_request",
  "cookie": "A70067"
}

"rdp": {
  "tx_id": 1,
  "event_type": "initial_response"
}

"rdp": {
  "tx_id": 2,
  "event_type": "connect_request",
  "client": {
    "version": "v5",
    "desktop_width": 1152,
    "desktop_height": 864,
    "color_depth": 15,
    "keyboard_layout": "en-US",
    "build": "Windows XP",
    "client_name": "ISD2-KM84178",
    "keyboard_type": "enhanced",
    "function_keys": 12,
    "product_id": 1,
    "capabilities": [
      "support_errinfo_pdf"
    ],
    "id": "55274-OEM-0011903-00107"
  },
  "channels": [
    "rdpdr",
    "cliprdr",
    "rdpsnd"
  ]
}

"rdp": {
  "tx_id": 3,
  "event_type": "connect_response"
}
```

PR #4184, opened 2019-09-09
RDP application layer - v5

jasonish:
Previous PR by @lmzke:
https://github.com/OISF/suricata/pull/4174

Changes from last PR:
- Address comments in https://github.com/OISF/suricata/pull/4174
- Disable by default.

The disabling by default isn't very clean codewise as I don't think we have a good way to tell Suricata that a parser should be off by default. Instead it assumes on by default, unless configured explicitly to off.  Did I miss something here? If not, its something I'll look into post 5.0.

Prscript was failing to even start.


victorjulien:
Merged in #4188, thanks a lot @lmzke! Thanks Jason for the integration work.

PR #4169, opened 2019-09-05
Bug 3053/v5

shivan1b:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3053

PR #4186, opened 2019-09-10
Bug 3053/v6

shivan1b:

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -45,6 +45,7 @@
 #include "util-pool.h"
 #include "util-radix-tree.h"
 #include "util-file.h"
+#include "util-byte.h"
 
 #include "stream-tcp-private.h"
 #include "stream-tcp-reassemble.h"
@@ -2640,7 +2641,15 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
                 cfg_prec->randomize = ConfValIsTrue(p->val);
             }
         } else if (strcasecmp("randomize-inspection-range", p->name) == 0) {
-            uint32_t range = atoi(p->val);
+            uint32_t range;
+            if (ByteExtractStringUint32(&range, 10, 0,
+                                        (const char *)p->val) < 0) {
+                FatalError(SC_ERR_FATAL, "Invalid value for randomize"
+                           "-inspection-range setting from conf file - \"%s\"."
+                           " It should be a valid integer less than 100."
+                           " Killing engine",
+                           p->val);
+            }
             if (range > 100) {
                 SCLogError(SC_ERR_SIZE_PARSE, "Invalid value for randomize"
                            " inspection range setting from conf file - %s."
--- a/src/detect-byte-extract.c
+++ b/src/detect-byte-extract.c
@@ -238,8 +238,12 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                    "for arg 1 for byte_extract");
         goto error;
     }
-    bed->nbytes = atoi(nbytes_str);
-
+    if (ByteExtractStringUint8(&bed->nbytes, 10, 0,
+                               (const char *)nbytes_str) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for number of bytes"
+                   " to be extracted: \"%s\".", nbytes_str);
+        goto error;
+    }
     /* offset */
     char offset_str[64] = "";
     res = pcre_copy_substring((char *)arg, ov,
@@ -249,7 +253,11 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                    "for arg 2 for byte_extract");
         goto error;
     }
-    int offset = atoi(offset_str);
+    int offset;
+    if (ByteExtractStringInt32(&offset, 10, 0, (const char *)offset_str) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for offset: \"%s\".", offset_str);
+        goto error;
+    }
     if (offset < -65535 || offset > 65535) {
         SCLogError(SC_ERR_INVALID_SIGNATURE, "byte_extract offset invalid - %d.  "
                    "The right offset range is -65535 to 65535", offset);
@@ -305,7 +313,13 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                            "for arg %d for byte_extract", i);
                 goto error;
             }
-            int multiplier = atoi(multiplier_str);
+            int multiplier;
+            if (ByteExtractStringInt32(&multiplier, 10, 0,
+                                       (const char *)multiplier_str) < 0) {
+                SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for"
+                        "multiplier: \"%s\".", multiplier_str);
+                goto error;
+            }
             if (multiplier < DETECT_BYTE_EXTRACT_MULTIPLIER_MIN_LIMIT ||
                 multiplier > DETECT_BYTE_EXTRACT_MULTIPLIER_MAX_LIMIT) {
                 SCLogError(SC_ERR_INVALID_SIGNATURE, "multipiler_value invalid "
@@ -409,7 +423,12 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                            "for arg %d in byte_extract", i);
                 goto error;
             }
-            bed->align_value = atoi(align_str);
+            if (ByteExtractStringUint8(&bed->align_value, 10, 0,
+                                       (const char *)align_str) < 0) {
+                SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid align_value: "
+                           "\"%s\".", align_str);
+                goto error;
+            }
             if (!(bed->align_value == 2 || bed->align_value == 4)) {
                 SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid align_value for "
                            "byte_extract - \"%d\"", bed->align_value);
--- a/src/detect-cipservice.c
+++ b/src/detect-cipservice.c
@@ -344,14 +344,18 @@ static DetectEnipCommandData *DetectEnipCommandParse(const char *rulestr)
         goto error;
     }
 
-    unsigned long cmd = atol(rulestr);
-    if (cmd > MAX_ENIP_CMD) //if command greater than 16 bit
-    {
-        SCLogError(SC_ERR_INVALID_SIGNATURE, "invalid ENIP command %lu", cmd);
+    uint64_t cmd;
+    if (ByteExtractStringUint64(&cmd, 10, 0, rulestr) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "invalid ENIP command"
+                ": \"%s\"", rulestr);
         goto error;
     }
-
-    enipcmdd->enipcommand = (uint16_t) atoi(rulestr);
+    if (cmd > MAX_ENIP_CMD) //if command greater than 16 bit
+    {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "invalid ENIP command %"PRIu64, cmd);
+        goto error;
+    }
+    enipcmdd->enipcommand = (uint16_t) cmd;
 
     return enipcmdd;
 
--- a/src/detect-dce-opnum.c
+++ b/src/detect-dce-opnum.c
@@ -175,16 +175,25 @@ static DetectDceOpnumData *DetectDceOpnumArgParse(const char *arg)
         if ((hyphen_token = index(dup_str_temp, '-')) != NULL) {
             hyphen_token[0] = '\0';
             hyphen_token++;
-            dor->range1 = atoi(dup_str_temp);
+            if (ByteExtractStringUint32(&dor->range1, 10, 0,
+                                        (const char *)dup_str_temp) < 0) {
+                goto error;
+            }
             if (dor->range1 > DCE_OPNUM_RANGE_MAX)
                 goto error;
-            dor->range2 = atoi(hyphen_token);
+            if (ByteExtractStringUint32(&dor->range2, 10, 0,
+                                        (const char *)hyphen_token) < 0) {
+                goto error;
+            }
             if (dor->range2 > DCE_OPNUM_RANGE_MAX)
                 goto error;
             if (dor->range1 > dor->range2)
                 goto error;
         }
-        dor->range1 = atoi(dup_str_temp);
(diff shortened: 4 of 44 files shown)


src/app-layer-htp.c
```diff
@@ -2640,7 +2641,15 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
                 cfg_prec->randomize = ConfValIsTrue(p->val);
             }
         } else if (strcasecmp("randomize-inspection-range", p->name) == 0) {
-            uint32_t range = atoi(p->val);
+            uint32_t range;
+            if (ByteExtractStringUint32(&range, 10, 0,
+                                        (const char *)p->val) < 0) {
+                FatalError(SC_ERR_FATAL, "Invalid value for randomize"
```
catenacyber: Why `FatalError` and not `SCLogError` as is done just below ?

src/detect-byte-extract.c
```diff
@@ -305,7 +313,13 @@ static inline DetectByteExtractData *DetectByteExtractParse(const char *arg)
                            "for arg %d for byte_extract", i);
                 goto error;
             }
-            int multiplier = atoi(multiplier_str);
+            int multiplier;
```
catenacyber: Should `multiplier` be declared as `int32` instead of `int` ?

src/detect-cipservice.c
```diff
@@ -344,14 +344,18 @@ static DetectEnipCommandData *DetectEnipCommandParse(const char *rulestr)
         goto error;
     }
 
-    unsigned long cmd = atol(rulestr);
+    uint64_t cmd;
+    if (ByteExtractStringUint64(&cmd, 10, 0, rulestr) < 0) {
```
catenacyber: You should directly use `ByteExtractStringUint16` to avoid further cast

src/detect-engine-port.c
```diff
@@ -1465,13 +1471,10 @@ DetectPort *PortParse(const char *str)
  */
 int DetectPortIsValidRange(char *port)
```
catenacyber: you could declare it static (so that we see that these changes do not affect other files)

src/detect-engine-port.c
```diff
@@ -1413,21 +1414,25 @@ DetectPort *PortParse(const char *str)
...
+    if ((port2 = strchr(port, ':')) != NULL) {
         /* 80:81 range format */
         port2[0] = '\0';
         port2++;
 
-        if(DetectPortIsValidRange(port))
-            dp->port = atoi(port);
-        else
-            goto error;
+        if (strcmp(port, "") != 0) {
```
catenacyber: Why do we consider empty string means port 0 and not failure ?

src/detect-fast-pattern.c
```diff
@@ -282,10 +283,11 @@ static int DetectFastPatternSetup(DetectEngineCtx *de_ctx, Signature *s, const c
                        "for fast_pattern offset");
             goto error;
         }
-        int offset = atoi(arg_substr);
-        if (offset > 65535) {
-            SCLogError(SC_ERR_INVALID_SIGNATURE, "Fast pattern offset exceeds "
-                       "limit");
+        int16_t offset;
+        if (ByteExtractStringInt16(&offset, 10, 0,
```
catenacyber: Should it rather be Uint16 ? 
Otherwise, it needs to be Int32 so that we can go up to 65535...

src/detect-ttl.c
```diff
@@ -181,7 +182,11 @@ static DetectTtlData *DetectTtlParse (const char *ttlstr)
                     return NULL;
 
                 mode = DETECT_TTL_LT;
-                ttl1 = atoi(arg3);
+                if (ByteExtractStringInt32(&ttl1, 10, 0, (const char *)arg3) < 0) {
```
catenacyber: IPv4 TTL is uint8 and not int32

But that may be part of another PR...

src/runmode-af-packet.c
```diff
@@ -290,7 +295,10 @@ static void *ParseAFPConfig(const char *iface)
     if (ConfGetChildValueWithDefault(if_root, if_default, "cluster-id", &tmpclusterid) != 1) {
         aconf->cluster_id = (uint16_t)(cluster_id_auto++);
     } else {
-        aconf->cluster_id = (uint16_t)atoi(tmpclusterid);
+        if (ByteExtractStringInt32(&aconf->cluster_id, 10, 0, (const char *)tmpclusterid) < 0) {
```
catenacyber: Uint16

src/util-proto-name.c
```diff
@@ -57,10 +58,11 @@ void SCProtoNameInit()
             if (proto_ch == NULL)
                 continue;
 
-            int proto = atoi(proto_ch);
-            if (proto >= 255)
+            int proto;
+            if (ByteExtractStringInt32(&proto, 10, 0, (const char *)proto_ch) < 0)
```
catenacyber: Uint8 ?

jlucovsky:
Be consistent with value delimiters -- either `'` or `"` but not both.

src/util-ip.c
```diff
@@ -65,7 +66,9 @@ bool IPv4AddressStringIsValid(const char *str)
 
     addr[dots][alen] = '\0';
     for (int x = 0; x < 4; x++) {
-        int a = atoi(addr[x]);
+        int a;
+        if (ByteExtractStringInt32(&a, 10, 0, (const char *)addr[x]) < 0)
+            return false;
```
jlucovsky: Suggest including at least a debug message on failure.

src/util-napatech.c
```diff
@@ -369,9 +369,12 @@ static uint32_t CountWorkerThreads(void)
 
                         char copystr[16];
                         strlcpy(copystr, lnode->val, 16);
-
-                        start = atoi(copystr);
-                        end = atoi(strchr(copystr, '-') + 1);
+                        if (ByteExtractStringUint8(&start, 10, 0, (const char *)copystr) < 0) {
+                            FatalError(SC_ERR_FATAL, "Napatech invalid worker range start");
```
jlucovsky: Suggest including the invalid value in the message.

src/util-proto-name.c
jlucovsky: I think that would be better -- `uint8`

src/app-layer-htp.c
catenacyber: Ok, but I still think it is better to have only one oh those for both cases

jlucovsky: Agreed -- consistent use is important.

shivan1b: Thanks @catenacyber and @jlucovsky . This change was as per this discussion: https://github.com/OISF/suricata/pull/4169#discussion_r321640364 , Should I replace exit calls everywhere with FatalError? Probably that's what was meant by this comment.

src/detect-byte-extract.c
shivan1b: Yes, sure.

src/detect-cipservice.c
shivan1b: Makes sense.

src/detect-engine-port.c
shivan1b: Yes OK.

src/app-layer-htp.c
victorjulien: Lets not further expand the scope of this work. Can you create a ticket for this instead?

src/detect-engine-port.c
shivan1b: hmm I had thought that since port2's value in case it is empty, it is reset to 65535 then this should be reset to 0 since it is range of ports we're working on. So if we have `:5000`, it should probably be `0:5000`.
If this is incorrect which I believe it is from your comment, I would like to learn why.

src/detect-fast-pattern.c
shivan1b: Yes OK

src/detect-dce-opnum.c
```diff
@@ -175,16 +175,25 @@ static DetectDceOpnumData *DetectDceOpnumArgParse(const char *arg)
         if ((hyphen_token = index(dup_str_temp, '-')) != NULL) {
             hyphen_token[0] = '\0';
             hyphen_token++;
-            dor->range1 = atoi(dup_str_temp);
+            if (ByteExtractStringUint32(&dor->range1, 10, 0,
+                                        (const char *)dup_str_temp) < 0) {
+                goto error;
```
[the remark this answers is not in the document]
shivan1b: Hi Jeff!
Other goto error calls also did not log any messages so I left it as-is, should I add error messages in all the cases?

src/detect-engine.c
```diff
@@ -2342,7 +2342,15 @@ static int DetectEngineCtxLoadConf(DetectEngineCtx *de_ctx)
             }
 
             if (insp_recursion_limit != NULL) {
-                de_ctx->inspection_recursion_limit = atoi(insp_recursion_limit);
+                if (ByteExtractStringInt32(&de_ctx->inspection_recursion_limit,
+                    10, 0, (const char *)insp_recursion_limit) < 0) {
+                    SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for "
+                            "detect-engine.inspection-recursion-limit, "
```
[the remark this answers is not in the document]
shivan1b: Yes OK

src/runmode-pfring.c
```diff
@@ -255,7 +263,11 @@ static void *ParsePfringConfig(const char *iface)
                 }
             }
         } else {
-            pfconf->threads = atoi(threadsstr);
+            if (ByteExtractStringUint16(&pfconf->threads, 10, 0, (const char *)threadsstr) < 0) {
+                SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for "
+                             "threads: '%s'. Resetting to 1.", threadsstr);
```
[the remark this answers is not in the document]
shivan1b: Yes it should

src/util-cpu.c
```diff
@@ -75,9 +76,12 @@ uint16_t UtilCpuGetNumProcessorsConfigured(void)
...
     return (uint16_t)nprocs;
 #elif OS_WIN32
-	long nprocs = -1;
-	const char* envvar = getenv("NUMBER_OF_PROCESSORS");
-	nprocs = (NULL != envvar) ? atoi(envvar) : 0;
+    int64_t nprocs = -1;
+    const char* envvar = getenv("NUMBER_OF_PROCESSORS");
+    if (envvar != NULL) {
+        if (ByteExtractStringInt64(&nprocs, 10, 0, envvar) < 0)
+            nprocs = 0;
```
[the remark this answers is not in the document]
shivan1b: Yes OK

src/app-layer-htp.c
shivan1b: Yes sure. Here's the ticket: https://redmine.openinfosecfoundation.org/issues/3188

src/detect-engine-port.c
catenacyber: Ok thanks, I did not know this was in a range.

catenacyber: So, should it be 0 or 0xFFFF ?

src/runmode-af-packet.c
catenacyber: If `cluster_id` is defined as an `int`, this means we have a double cast :
- first an explicit cast from `int` to `uint16` where 0x23456 becomes 0x3456 (and I am not sure what 0xFFFFFFFF becomes)
- then an implicit cast from `uint16` to `int` but the value does not change since every `uint16` value can be an `int`

If you use `ByteExtractStringUint16`, do you get a warning ?
Are there other places than this where `cluster_id` can have a value which is not an `uint16` (such as e negative value) ?
Otherwise, I would change the declaration of  `cluster_id` to be `uint16`

shivan1b: > If `cluster_id` is defined as an `int`, this means we have a double cast :
> 
> * first an explicit cast from `int` to `uint16` where 0x23456 becomes 0x3456 (and I am not sure what 0xFFFFFFFF becomes)
> * then an implicit cast from `uint16` to `int` but the value does not change since every `uint16` value can be an `int`
> 
Thanks for the explanation.

> If you use `ByteExtractStringUint16`, do you get a warning ?
Yes. Since `ByteExtractStringUint16` expects its first argument to be `uint16_t` but instead I'm giving `int`.

```
runmode-af-packet.c: In function ‘ParseAFPConfig’:
runmode-af-packet.c:298:37: warning: passing argument 1 of ‘ByteExtractStringUint16’ from incompatible pointer type [-Wincompatible-pointer-types]
  298 |         if (ByteExtractStringUint16(&aconf->cluster_id, 10, 0, (const char *)tmpclusterid) < 0) {
      |                                     ^~~~~~~~~~~~~~~~~~
      |                                     |
      |                                     int *
In file included from runmode-af-packet.c:58:
util-byte.h:176:39: note: expected ‘uint16_t *’ {aka ‘short unsigned int *’} but argument is of type ‘int *’
  176 | int ByteExtractStringUint16(uint16_t *res, int base, uint16_t len, const char *str);
      |                             ~~~~~~~~~~^~~
```
> Are there other places than this where `cluster_id` can have a value which is not an `uint16` (such as e negative value) ?
There aren't any.

> Otherwise, I would change the declaration of `cluster_id` to be `uint16`
Done. Thanks a lot!

victorjulien: Yes, this makes sense to do including for ```AFPThreadVars::cluster_id```. Please make sure to put these af-packet changes in their own commit.

src/detect-engine-port.c
shivan1b: I think it should be 0 since strcmp returns 0 in case strings are equal. In this case, if start port is empty then strcmp would return 0.

catenacyber: Looks ok to me, thanks

src/detect-lua.c
```diff
@@ -278,8 +279,16 @@ int DetectLuaMatchBuffer(DetectEngineThreadCtx *det_ctx,
                 SCLogDebug("k='%s', v='%s'", k, v);
 
                 if (strcmp(k, "retval") == 0) {
-                    if (atoi(v) == 1)
+                    int val;
+                    if (ByteExtractStringInt32(&val, 10, 0,
+                                               (const char *)v) < 0) {
+                        SCLogError(SC_ERR_INVALID_VALUE, "Invalid value in "
```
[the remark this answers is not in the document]
inashivb: Do you have a suggestion?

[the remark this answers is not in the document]
jlucovsky: Maybe something like `"SCLogError(Invalid value for \"retval\" from LUA return table: '%s'", v)`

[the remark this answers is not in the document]
inashivb: Thanks, Jeff!

inashivb:
This PR has been modified according to #4488 and broken into 4 parts which shall be coming along once 4488 is merged.

PR #4194, opened 2019-09-12
htp: set lzma memlimit from config

victorjulien:


src/app-layer-htp.c
```diff
@@ -179,6 +179,11 @@ SCEnumCharMap http_decoder_event_table[ ] = {
     { "REQUEST_LINE_INCOMPLETE",
         HTTP_DECODER_EVENT_REQUEST_LINE_INCOMPLETE},
 
+    { "LZMA_SUSPICIOUS_DICT_RATIO",
```
jlucovsky: Should this be `DIST`?

src/app-layer-htp.h
```diff
@@ -119,6 +120,9 @@ enum {
     HTTP_DECODER_EVENT_REQUEST_LINE_INVALID,
     HTTP_DECODER_EVENT_REQUEST_BODY_UNEXPECTED,
 
+    HTTP_DECODER_EVENT_LZMA_SUSPICIOUS_DICT_RATIO,
```
jlucovsky: `DIST`?

src/app-layer-htp.c
victorjulien: No it was about a dictionary size. But this is outdated code, in the newer this isn't present.

PR #4197, opened 2019-09-12
htp: set lzma memlimit from config

jasonish:


victorjulien:
Merged in #4201. Added a missing guard for the new htp function.

PR #4200, opened 2019-09-13
Register missing app-layer parser loggers for TLS and SSH

thus:
Fix so that it is possible to use a TLS or SSH Lua output script without having to have a logger enabled for the app-layer protocol.

https://redmine.openinfosecfoundation.org/issues/3162

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/190
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/190


victorjulien:
Merged in #4203, thanks Mats!

Do we need a backport to 4.1.x for this you think?

thus:
It would probably be nice to backport it to 4.1.x as well, I think :)

PR #4202, opened 2019-09-15
htp: need htp_config.h for new lzma function

jlucovsky:
`app-layer-htp.c` needs function declaration of `htp_config_set_lzma_memlimit`


victorjulien:
What problem is this solving?

jlucovsky:
> What problem is this solving?

Local compiler errors with `app-layer-htp.c` ... I'm not able to reproduce the issue so it was likely an environmental issue.

PR #4215, opened 2019-09-19
Issue 1271 - Coredump after privilege drop - v1

jasonish:
I came up with this refactor of https://github.com/OISF/suricata/pull/3362 after testing that only coredumps have to be enabled after dropping privileges, and not the whole coredump size configuration.

https://redmine.openinfosecfoundation.org/issues/1271

Somethings to note after re-enabling coredumps after dropping privileges:
- The ownership of /proc/<PID> changes to the user. Without this PR, these remain owned by root and the user the process runs as is limited in what they can do in here (if anything at all)
- The user suricata is now running as can now kill the process. Previously the user couldn't.
- A coredump will be generated. Previously it wasn't.

I think all this is acceptable. The reason for disabling coredumps after dropping privileges appears to be more for the use case where the app is setuid, where any users starts the process with elevated privileges to start. I don't see this as a common, or even recommended use case for Suricata.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/375
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/731

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -3014,6 +3014,10 @@ int main(int argc, char **argv)
     }
 
     SCDropMainThreadCaps(suricata.userid, suricata.groupid);
+
+    /* Re-enable coredumps after privileges are dropped. */
+    CoredumpEnable();
+
     PreRunPostPrivsDropInit(suricata.run_mode);
 
     PostConfLoadedDetectSetup(&suricata);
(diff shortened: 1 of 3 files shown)


victorjulien:
Merged in #4225, thanks Jason!

PR #4217, opened 2019-09-19
protodetect: retry when both sides do not agree

catenacyber:

Describe changes:
- Retry protocol detection above TCP when both sides (client and server) do not agree

Makes test https://github.com/OISF/suricata-verify/pull/117 pass

There is likely a performance impact to be assessed
We keep retrying until both sides agree. Maybe we could fix a threshold on the number of retries.
I am not using `AppLayerParserParse` results as it is designed for relaxed parsing.

This should be done for UDP as well after merging #4058 

Replaces #4158 with rebase to make tests pass

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -616,6 +616,38 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
                            data, data_len, flags) != 0) {
             goto failure;
         }
+    } else if (f->alproto_ts != f->alproto_tc &&
+               f->alproto_ts != ALPROTO_UNKNOWN && f->alproto_ts != ALPROTO_FAILED &&
+               f->alproto_tc != ALPROTO_UNKNOWN && f->alproto_tc != ALPROTO_FAILED) {
+        bool reverse_flow = false;
+        AppProto alproto_otherdir = (flags & STREAM_TOSERVER) ? f->alproto_tc : f->alproto_ts;
+
+        //partial reset for protocol detection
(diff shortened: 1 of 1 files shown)


catenacyber:
CI failure looks unrelated : sip-response-line

src/app-layer.c
```diff
@@ -616,6 +616,38 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
...
+        AppProto alproto_otherdir = (flags & STREAM_TOSERVER) ? f->alproto_tc : f->alproto_ts;
+
+        //partial reset for protocol detection
+        FLOW_RESET_PM_DONE(f, flags);
+        FLOW_RESET_PP_DONE(f, flags);
+        f->probing_parser_toserver_alproto_masks = 0;
+        f->probing_parser_toclient_alproto_masks = 0;
+
+        //test if we are now agreed with other side
+        AppProto newproto = AppLayerProtoDetectGetProto(app_tctx->alpd_tctx,
```
victorjulien: Can you update this to have a proper commit message that explains:
1. the scenario
2. why and how it doesn't work correctly now
3. how this patch fixes it and why that works.

My main question, that should be answered in the git message, is: why would resetting the protocol detection and running it again give a different result?

catenacyber: Ok, I will update the git message.

The scenario is an evil server and a legit client.
The evil server first "pretends" to speak another protocol.
Suricata classifies the flow with this faked protocol.
Then the server answers the clients requests with the right protocol...

> why would resetting the protocol detection and running it again give a different result?

Because the evil server will speak a different protocol

PR #4231, opened 2019-09-20
tls-log: fix so buffer is reset on custom logging

thus:
Move MemBufferReset() so it also works when using custom TLS logging. This avoids duplicate entries in the tls-log.

https://redmine.openinfosecfoundation.org/issues/3177

Prscript:
- PR thus-pcap: https://buildbot.openinfosecfoundation.org/builders/thus-pcap/builds/191
- PR thus: https://buildbot.openinfosecfoundation.org/builders/thus/builds/191

### DIFF
--- a/src/log-tlslog.c
+++ b/src/log-tlslog.c
@@ -451,12 +451,12 @@ static int LogTlsLogger(ThreadVars *tv, void *thread_data, const Packet *p,
         return 0;
     }
 
+    MemBufferReset(aft->buffer);
+
     /* Custom format */
     if (hlog->flags & LOG_TLS_CUSTOM) {
         LogTlsLogCustom(aft, ssl_state, &p->ts, srcip, sp, dstip, dp);
     } else {
-
-        MemBufferReset(aft->buffer);
         CreateTimeString(&p->ts, timebuf, sizeof(timebuf));
         MemBufferWriteString(aft->buffer,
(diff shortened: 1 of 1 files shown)


src/log-tlslog.c
```diff
@@ -451,12 +451,12 @@ static int LogTlsLogger(ThreadVars *tv, void *thread_data, const Packet *p,
         return 0;
     }
 
+    MemBufferReset(aft->buffer);
```
jlucovsky: May I suggest that (1) The call to `MemBufferReset` be retained in `LogTlsLogger` and (2) a call to `MemBufferReset` be added in `LogTlsLogCustom`?

thus: After reading the code once more, I understand the reasoning why you want this change. I don't like how the code in _log_tlslog.c_ is structured. I would prefer if the code was structured as in _output-json-tls.c_ where functions for logging 'basic', 'extended' and 'custom' just fills in the buffer, and all the other functionality is shared. Would you be okay with that I make this change for tls-log also instead? :)

jlucovsky: You have a valid point and yes, I'm ok with the original proposed change.

thus: I've made the changes I mentioned in #4232 :)

thus:
Replaced by #4232

PR #4240, opened 2019-09-25
filestore: Ensure configured steam-depth is gte stream.depth

jlucovsky:
_Request for comments only_  Would like comments on the solution and if the solution is complete.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2506](https://redmine.openinfosecfoundation.org/projects/suricata/issues/2506) 

Describe changes:
- Ensure configured stream depth for filestore (v1 and v2) is either 0 or greater than configured stream depth. Configured values, if non-zero, are compared against `stream.depth` ... a warning log record is written when the value is not greater than `stream.depth` and the value for `file-store.stream-depth` is set to `stream.depth`

*NOTES*
1. With a configured `file-store.stream-depth = 100600` and `stream.depth = 100000`, files larger than `100600` were created. I don't think this is proper behavior.

### DIFF
--- a/doc/userguide/file-extraction/file-extraction.rst
+++ b/doc/userguide/file-extraction/file-extraction.rst
@@ -25,7 +25,7 @@ Settings
 
 *stream.checksum_validation* controls whether or not the stream engine rejects packets with invalid checksums. A good idea normally, but the network interface performs checksum offloading a lot of packets may seem to be broken. This setting is enabled by default, and can be disabled by setting to "no". Note that the checksum handling can be controlled per interface, see "checksum_checks" in example configuration.
 
-*file-store.stream-depth* controls how far into a stream reassembly is done. Beyond this value no reassembly will be done. This means that after this value the HTTP session will no longer be tracked. By default a settings of 1 Megabyte is used. 0 sets it to unlimited. If set to no, it is disabled and stream.reassembly.depth is considered.
+*file-store.stream-depth* controls how far into a stream reassembly is done. Beyond this value no reassembly will be done. This means that after this value the HTTP session will no longer be tracked. By default a setting of 1 Megabyte is used. 0 sets it to unlimited. If set to no, it is disabled and stream.reassembly.depth is considered. Non-zero values must be greater than ``stream.stream-depth`` to be used.
 
 *libhtp.default-config.request-body-limit* / *libhtp.server-config.<config>.request-body-limit* controls how much of the HTTP request body is tracked for inspection by the http_client_body keyword, but also used to limit file inspection. A value of 0 means unlimited.
 
@@ -74,7 +74,9 @@ If not enabled, ``stream.reassembly.depth`` will be considered.
 Setting ``file-store.stream-depth`` to 0 permits to store any files.
 
 ``file-store.stream-depth`` will always override ``stream.reassembly.depth``
-when filestore keyword is used.
+when filestore keyword is used. However, it is not possible to set ``file-store.stream-depth``
+to a value less than ``stream.reassembly.depth``. Values less than this amount are ignored
+and a warning message will be displayed.
 
 A protocol parser, like modbus, could permit to set a different
 store-depth value and use it rather than ``file-store.stream-depth``.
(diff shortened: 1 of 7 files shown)


doc/userguide/file-extraction/file-extraction.rst
```diff
@@ -25,7 +25,7 @@ Settings
 
 *stream.checksum_validation* controls whether or not the stream engine rejects packets with invalid checksums. A good idea normally, but the network interface performs checksum offloading a lot of packets may seem to be broken. This setting is enabled by default, and can be disabled by setting to "no". Note that the checksum handling can be controlled per interface, see "checksum_checks" in example configuration.
 
-*file-store.stream-depth* controls how far into a stream reassembly is done. Beyond this value no reassembly will be done. This means that after this value the HTTP session will no longer be tracked. By default a settings of 1 Megabyte is used. 0 sets it to unlimited. If set to no, it is disabled and stream.reassembly.depth is considered.
+*file-store.stream-depth* controls how far into a stream reassembly is done. Beyond this value no reassembly will be done. This means that after this value the HTTP session will no longer be tracked. By default a setting of 1 Megabyte is used. 0 sets it to unlimited. If set to no, it is disabled and stream.reassembly.depth is considered. Non-zero values must be greater than ``stream.stream-depth`` to be used.
```
jasonish: Is this correct? If file-store.stream-depth is commented out it will use a default of 1 megabyte? I do find that a bit misleading. If its uncommented I would expect the global stream depth to be used and not some file-store specific default.  I think it could lead to confusion.

jasonish:
I wonder if it would make sense for `file-store.stream-depth` to not fiddle with the actual stream depth at all? Instead:
- If not set, track files til stream depth.
- If greater than stream depth, log a warning.
- If less than stream depth, do not change stream depth, but stop storing file data at that depth.
So the documentation string would be something like:
`depth into stream to extract file`.

Would that be a desirable performance enhancement?

src/log-filestore.c
```diff
@@ -61,6 +62,8 @@
 
 #define MODULE_NAME "LogFilestoreLog"
 
+#define MIN_STREAM_DEPTH_VALUE    (3 * 1024)
```
jasonish: Doesn't look like this is used.

doc/userguide/file-extraction/file-extraction.rst
jlucovsky: Perhaps my wording is unnecessarily confusing? Here's the rules
* 0 -- no limit
* non-zero -- must be greater than `stream.depth` to be used; else, `stream.depth` is used.

My wording is incorrect by referring to `stream.reassembly.depth` and `stream.stream-depth` -- these refer to `stream.depth`

src/log-filestore.c
jlucovsky: It's a remnant ... i'll remove.

jlucovsky:
> I wonder if it would make sense for `file-store.stream-depth` to not fiddle with the actual stream depth at all? Instead:
> 
> * If not set, track files til stream depth.
Agreed.
> * If greater than stream depth, log a warning.
And use `stream.depth`'s value instead?

> * If less than stream depth, do not change stream depth, but stop storing file data at that depth.
This means it's no longer an "increase only" setting. Is that desired?

>   So the documentation string would be something like:
>   `depth into stream to extract file`.
> 
> Would that be a desirable performance enhancement?

victorjulien:
> I wonder if it would make sense for `file-store.stream-depth` to not fiddle with the actual stream depth at all? Instead:
> 
> * If not set, track files til stream depth.
> * If greater than stream depth, log a warning.
> * If less than stream depth, do not change stream depth, but stop storing file data at that depth.
>   So the documentation string would be something like:
>   `depth into stream to extract file`.
> 
> Would that be a desirable performance enhancement?

No, because it undoes the primary goal of the option: override depth limitation for sessions that contain files to be stored. The whole point is to make sure we can have certain TCP sessions use a higher depth than others.

PR #4247, opened 2019-09-28
log/alert: Improve HTTP body logging documentation

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[2640](https://redmine.openinfosecfoundation.org/issues/2640)

Describe changes:
- Clarify wording of HTTP body logging for alerts
- Warn if HTTP body logging is configured but metadata is not enabled.

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -97,6 +97,8 @@
             LOG_JSON_APP_LAYER  |                                  \
             LOG_JSON_RULE_METADATA)
 
+#define _JSON_BODY_LOGGING  (LOG_JSON_HTTP_BODY | LOG_JSON_HTTP_BODY_BASE64)
+
 #define JSON_STREAM_BUFFER_SIZE 4096
 
 typedef struct AlertJsonOutputCtx_ {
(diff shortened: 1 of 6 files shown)


src/output-json-alert.c
```diff
@@ -97,6 +97,8 @@
             LOG_JSON_APP_LAYER  |                                  \
             LOG_JSON_RULE_METADATA)
 
+#define _JSON_BODY_LOGGING  (LOG_JSON_HTTP_BODY | LOG_JSON_HTTP_BODY_BASE64)
```
victorjulien: why is there a leading underscore?

src/output-json-alert.c
```diff
@@ -810,6 +812,7 @@ static void SetFlag(const ConfNode *conf, const char *name, uint16_t flag, uint1
 static void JsonAlertLogSetupMetadata(AlertJsonOutputCtx *json_output_ctx,
         ConfNode *conf)
 {
+    static bool _warn_no_meta = false;
```
victorjulien: same here, why the leading _?

src/output-json-alert.c
jlucovsky: Just to indicate it's private; however, it's private anyway since it's not in a header file.

I'll remove the leading `_`.

src/output-json-alert.c
jlucovsky: Following convention: https://github.com/OISF/suricata/blob/master/src/output-json-anomaly.c#L375:L376

Would you like the `_` removed from both files?

jlucovsky:
Continued in #4276

PR #4256, opened 2019-10-01
deprecation cleans and classification.config handling - v1

jasonish:
Deprecation notifications when enabling outputs, and in documentation.

Also install classification.config and reference.config into datadir (perhaps this should have been a separate PR).

https://redmine.openinfosecfoundation.org/issues/3209

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/386
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/742


victorjulien:
I assume the travis failure is due to the classification.config move which isn't handled by SV.

jasonish:
> The messaging is a bit inconsistent. Some places use Month, Year. Others state 6.0. Another just v6. I think I prefer the month/year everywhere.

Will fix.

jasonish:
> I assume the travis failure is due to the classification.config move which isn't handled by SV.

Yeah, not sure what to do about that.  I couldn't get the autoconf _DATA stuff to do its thing in the top directory, but works fine when in a sub-directory.  It would be an update to `suricata-verify/run.py` to set the path correctly to classification.config, not a big deal.  But I'll take another look at doing this without moving the file.

victorjulien:
> > I assume the travis failure is due to the classification.config move which isn't handled by SV.
> 
> Yeah, not sure what to do about that. I couldn't get the autoconf _DATA stuff to do its thing in the top directory, but works fine when in a sub-directory. It would be an update to `suricata-verify/run.py` to set the path correctly to classification.config, not a big deal. But I'll take another look at doing this without moving the file.

We can update SV, but it will have to be in such a way that 4.1.x will continue to work.

jasonish:
Replaced by https://github.com/OISF/suricata/pull/4258

PR #4261, opened 2019-10-02
Feat/classtype unknown/v9

victorjulien:
Makes missing classtypes in rules non-fatal. Instead creates a warning, once per unique missing classtype.

Consequence is that -T no longer errors out missing classtypes.

Perhaps another commandline option, e.g. something like ```--strict-rule-parsing``` or ```--strict-classtype-parsing``` could be added to make -T error out again in this case.

- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/316
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/317

### DIFF
--- a/src/alert-unified2-alert.c
+++ b/src/alert-unified2-alert.c
@@ -948,7 +948,7 @@ static int Unified2IPv6TypeAlert(ThreadVars *t, const Packet *p, void *data)
         phdr->generator_id = htonl(pa->s->gid);
         phdr->signature_id = htonl(pa->s->id);
         phdr->signature_revision = htonl(pa->s->rev);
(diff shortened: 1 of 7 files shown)


jasonish:
> Perhaps another commandline option, e.g. something like --strict-rule-parsing or --strict-classtype-parsing could be added to make -T error out again in this case.

Maybe a more generic `--strict` command line option. It would accept a comma separated lists of things to be strict about.

`--strict classifications,flowbits`

Could add other things that are not rule related.  Associated configuration file section could be:
```
strict:
  - classifications
  - flowbits
```

victorjulien:
Replaced by #4265

PR #4263, opened 2019-10-03
log/anomaly: Remove `event_no` from anomaly record

jlucovsky:
Describe changes:
- Remove `event_no` item from anomaly events. This field was informational only and was of little value to rules.


victorjulien:
Merged in #4271, thanks Jeff!

PR #4266, opened 2019-10-03
logging: make verbosity levels more specific - v1

jasonish:
Per issue https://redmine.openinfosecfoundation.org/issues/1851,
make the levels of verbosity fixed to specific logging levels
rather than adjusting the verbosity based on the configured
default log level.

The idea is to make it more easy to reason about.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/1851

Also fix the issue where individual loggers cannot be more
verbose than the default logging level.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/3210

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/389
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/745

### DIFF
--- a/doc/userguide/partials/options.rst
+++ b/doc/userguide/partials/options.rst
@@ -18,8 +18,18 @@
 
 .. option:: -v
 
-   The -v option enables more verbosity of Suricata's output. Supply
-   multiple times for more verbosity.
+   Increase the verbosity of the Suricata application logging by
+   increasing the log level from the default. This option can be
+   passed multiple times to further increase the verbosity.
+
+   - -v: INFO
+   - -vv: PERF
+   - -vvv: CONFIG
+   - -vvvv: DEBUG
+
+   This option will not decrease the log level set in the
+   configuration file if it is already more verbose than the level
+   requested with this option.
 
 .. Basic input options.
 
--- a/doc/userguide/partials/options.rst
+++ b/doc/userguide/partials/options.rst
@@ -16,6 +16,8 @@
 
    Test configuration.
 
+.. _cmdline-option-v:
+
 .. option:: -v
 
    Increase the verbosity of the Suricata application logging by
--- a/src/util-debug.c
+++ b/src/util-debug.c
@@ -1299,6 +1299,13 @@ void SCLogLoadConfig(int daemon, int verbose)
     SCLogInitData *sc_lid;
     int have_logging = 0;
     int max_level = 0;
+    SCLogLevel min_level = 0;
+
+    /* If verbose logging was requested, set the minimum as
+     * SC_LOG_NOTICE plus the extra verbosity. */
+    if (verbose) {
+        min_level = SC_LOG_NOTICE + verbose;
+    }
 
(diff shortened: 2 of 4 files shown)


doc/userguide/partials/options.rst
```diff
@@ -16,10 +16,22 @@
...
-   The -v option enables more verbosity of Suricata's output. Supply
-   multiple times for more verbosity.
+   Increase the verbosity of the Suricata application logging by
+   increasing the log level from the default. This option can be
+   passed multiple times to further increase the verbosity.
+
+   - -v: INFO
+   - -vv: PERF
+   - -vvv: CONFIG
+   - -vvvv: DEBUG
```
jlucovsky: This took me a few times to understand and perhaps a slight modification  would be something like:

```
Here are the logging level equivalents when specifying the verbose option:
- -v changes the log level to INFO
- -vv changes the log level to PERF
...
```

src/util-debug.c
```diff
@@ -1385,8 +1386,12 @@ void SCLogLoadConfig(int daemon, int verbose)
                     level_s);
                 exit(EXIT_FAILURE);
             }
+            max_level = MAX(max_level, level);
         }
 
+        /* Increase the level of extra verbosity was requested. */
+        level = MAX(min_level, level);
```
jlucovsky: This sets level to the max of `NOTICE` + verbosity or the configured level. IMO, the verbosity should be added to the configured level, not `NOTICE`.

jasonish: Would that still meet the requirement that -v is always info, -vv is always perf and so on?

Per the issue, we want the levels to be fixed, not relative.

src/util-debug.c
```diff
@@ -1298,6 +1298,14 @@ void SCLogLoadConfig(int daemon, int verbose)
...
     SCLogInitData *sc_lid;
     int have_logging = 0;
+    int max_level = 0;
+    SCLogLevel min_level = 0;
+
+    /* If verbose logging was requested, set the minimum as
+     * SC_LOG_NOTICE plus the extra verbosity. */
+    if (verbose) {
+        min_level = SC_LOG_NOTICE + verbose;
+    }
```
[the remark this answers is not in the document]
jasonish: We could alter the code, but the idea is that min_level is 0, but only adjusted based on SC_LOG_NOTICE if a -v is provided.

If -v is not provided, the configuration may have a default level of warn, and we wouldn't want to push that up to notice, unless a -v is provided on the command line.

doc/userguide/partials/options.rst
jasonish: Agreed on the wording, I will fix that.

As for the level used at runtime, no. Per the issue the discussion was having -v occurrences max to fixed levels based on what is the current hardcoded default of notice.

jasonish:
> @jasonish is this still current?

Yes, unless we want to revisit what should happen.  I do plan to update the doc though.

jasonish:
Merged.

PR #4281, opened 2019-10-07
doc: add upgrade page

victorjulien:


doc/userguide/upgrade.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Configuration Updates
+~~~~~~~~~~~~~~~~~~~~~
+
+New versions of Suricata will occationally include updated config files:
+``classification.config`` and ``reference.config``. Since the Suricata
+installation will not overwrite these if they exist, they should be manually
+updated. If there are no local modifications they can simply be overwritten
+by the ones Suricata supplies.
+
```
norg: It's also relevant to some parts of the suricata.yaml as well
Should we also add a diff which settings changed there or just recommend people to use diff and read the changelog/new config?

doc/userguide/upgrade.rst
```diff
@@ -0,0 +1,55 @@
...
+Major updates include new features, new default settings and often also
+remove features.
```
norg: Link to deprecation policy?

doc/userguide/upgrade.rst
```diff
@@ -0,0 +1,55 @@
...
+
+
+Upgrading 4.1 to 5.0
+--------------------
+
+Major changes
+~~~~~~~~~~~~~
+- New protocols enabled by default: snmp (new config only)
+- New protocols disabled by default: rdp, sip
+- New defaults: nfs, smb, tftp, krb5 ntp are all enabled by default
```
norg: New protocols defaults to make it more clear it's about the same topic

doc/userguide/upgrade.rst
victorjulien: The diff is quite large and a bit too technical I think. I would prefer a more textual overview of changes.

victorjulien:
Merged with some improvements through #4288

PR #4234, opened 2019-09-23
detect: hash keywords depend on filestore - v1

jasonish:
Keywords like filemd5 require the filestore to be enabled, and
will not trigger if it is not enabled, nor will it give
an indication that they will not trigger.

For these keywords, set the filestore flag as if the filestore
option had been provided.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2490

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/380
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/736

PR #4284, opened 2019-10-08
file extraction: always prune files after detect - v3

jasonish:
If a keyword like filemd5 was being used without a filestore,
or a file output enabled, it would be pruned before detection
had a chance to match.

Consolidate file pruning to the end of the flow worker so files
are available for detection even when a file output is not
enabled.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2490

Previous PR:
https://github.com/OISF/suricata/pull/4234

- Will fixing some unit tests.

### DIFF
--- a/src/flow-worker.c
+++ b/src/flow-worker.c
@@ -281,6 +281,19 @@ static TmEcode FlowWorker(ThreadVars *tv, Packet *p, void *data, PacketQueue *pr
     // Outputs.
     OutputLoggerLog(tv, p, fw->output_thread);
 
+    /* Prune any stored files. */
+    if (p->flow && p->flow->alstate) {
+        Flow *f = p->flow;
+        FileContainer *ffc_ts = AppLayerParserGetFiles(p->proto, f->alproto,
+            f->alstate, STREAM_TOSERVER);
+        FileContainer *ffc_tc = AppLayerParserGetFiles(p->proto, f->alproto,
+            f->alstate, STREAM_TOCLIENT);
+        if (ffc_ts && (p->flowflags & FLOW_PKT_TOSERVER))
+            FilePrune(ffc_ts);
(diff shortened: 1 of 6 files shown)


jasonish:
Closing. New version here:
https://github.com/OISF/suricata/pull/4234

src/flow-worker.c
```diff
@@ -281,6 +281,19 @@ static TmEcode FlowWorker(ThreadVars *tv, Packet *p, void *data, PacketQueue *pr
     // Outputs.
     OutputLoggerLog(tv, p, fw->output_thread);
 
+    /* Prune any stored files. */
+    if (p->flow && p->flow->alstate) {
+        Flow *f = p->flow;
+        FileContainer *ffc_ts = AppLayerParserGetFiles(p->proto, f->alproto,
```
victorjulien: is there a reason for first fetching both containers but then using only one? Packet is either toserver or toclient.

victorjulien: I see this comes from the ``OutputFileLog``. Wonder why we did it this way.

src/flow-worker.c
```diff
@@ -281,6 +281,19 @@ static TmEcode FlowWorker(ThreadVars *tv, Packet *p, void *data, PacketQueue *pr
     // Outputs.
     OutputLoggerLog(tv, p, fw->output_thread);
 
+    /* Prune any stored files. */
```
victorjulien: Can this move into it's own little function?

src/flow-worker.c
victorjulien: Looks like we added this in f75df19c35b6a36c00083ebaf7d06ecf4efac70e, but that is not relevant here I think.

jasonish:
Closing. Continued here: https://github.com/OISF/suricata/pull/4286

PR #4287, opened 2019-10-09
ftp: Handle malformed RETR/STOR

jlucovsky:
Ensure that RETR (STOR) have a filename -- otherwise, treat the command
string as malformed.

Added unittests for each command and verified that SEGV's occur without
parser change and no longer occur with the parser change.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[3226](https://redmine.openinfosecfoundation.org/issues/3226)

Describe changes:
- Ensure RETR/STOR also have a filename, else treat as malformed.
- Unittest cases for RETR/STOR that are malformed


victorjulien:
Looks like this fixes the issue for me.

PR #4291, opened 2019-10-10
build: cbindgen (v10)

dbcfd:
Version 10 (from https://github.com/OISF/suricata/pull/4279)

Rust headers are now generated using cbindgen. If cbindgen is present, they can
be generated during dist, otherwise they will be available for builds.

Describe changes:
- Rust headers are now created by cbindgen
- Cbindgen is run as a step under dist when cbindgen is available and a high enough version


victorjulien:
https://redmine.openinfosecfoundation.org/issues/3308

jasonish:
We can get rid of a few warnings w.r.t. lua with:
```
+[defines]
+"feature = lua" = "HAVE_LUA"
```
in cbindgen.toml.

jasonish:
I see 2 possible issues here, that we can live with:
- All that annoying WARN output.  I think cbindgen should be smarter there. If we explicitly exclude something, don't warn us about it. One idea here is to grep this output for actual errors, or just leave it as is (as I see other projects doing that have a lot of warn output as well).
- Doing anything in Rust, which will trigger a rebuild of the header file causes all C files to be rebuilt, and ccache doesn't provide much help here.

victorjulien:
> * Doing anything in Rust, which will trigger a rebuild of the header file causes all C files to be rebuilt, and ccache doesn't provide much help here.

This may be a side effect of my suggestion to include the bindings in ``suricata-common.h``. Maybe that wasn't such a good idea after all. OTOH, the more rust we'll write, the more we'll include it.

jasonish:
> > * Doing anything in Rust, which will trigger a rebuild of the header file causes all C files to be rebuilt, and ccache doesn't provide much help here.
> 
> This may be a side effect of my suggestion to include the bindings in `suricata-common.h`. Maybe that wasn't such a good idea after all. OTOH, the more rust we'll write, the more we'll include it.

I'm pretty sure thats it. I'm not opposed to having to include the rust binding in each file that needs. Sure, eventually that might be about every file, but currently its not.

dbcfd:
Closing for https://github.com/OISF/suricata/pull/4367

PR #4293, opened 2019-10-11
configure.ac: fix static build with pcap

ffontaine:
pcap can depends on nl-3 so use pkg-config to find these dependencies
otherwise all AC_CHECK_LIB calls will fail when building statically

Link to [redmine] ticket: https://redmine.openinfosecfoundation.org/issues/3232


victorjulien:
Merged in #4299, thanks!

PR #4297, opened 2019-10-12
Some last minute fixups...

jasonish:
Document EVE DNS v2 as being the default, and include its default
configuration in the manual.

Rename the current EVE DNS section to v1 format.

The idea here is to commit us to v2 being the new default for 5.0,
and documenting this as such.

coming...


victorjulien:
Merged in #4300, thanks Jason.

PR #4303, opened 2019-10-14
RFC -- Modifications to use `pcre_jit_exec`

jlucovsky:
_PREVIEW/RFC_ only

I'm posting this PR to request feedback for the approach to implement use of `pcre_jit_exec` when `PCRE_HAVE_JIT` has the value `1`.

The unittests and sv pass with the changes; I have not reconciled the use o` PCRE_HAVE_JIT` within `detect-pcre.c` (esp. see `pcre_use_jit`) and there are a 1/2 dozen or so source code modules that always use `pcre_exec` that remain to be modified.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[749](https://redmine.openinfosecfoundation.org/projects/suricata/issues/749)

Does this seem like the proper approach? Is the introduction of a macro `PCRE_EXEC` ok? Other comments?


victorjulien:
The intend was not to use the pcre jit for the rule parsing, but instead only for the matching of the pcre keyword against traffic (so in detect-pcre.c).

JIT for rule parsing is probably useless anyway, as JIT doesn't support the core aspect of our pcre based parsing: substring capture.

However, I *do* like the cleanup of having less code duplication.

src/detect-parse.h
```diff
@@ -39,6 +39,21 @@ enum {
...
+    pcre_extra *study;
+    pcre_jit_stack *jit_stack;
+    struct DetectParseRegex_ *next;
+} DetectParseRegex;
+
+#if PCRE_HAVE_JIT
+#define PCRE_EXEC(p, str, x1, x2, ss, ss_cnt) \
+        pcre_jit_exec((p)->regex, (p)->study, str, strlen(str), x1, x2, ss, ss_cnt, (p)->jit_stack)
+#else
+#define PCRE_EXEC(p, str, x1, x2, ss, ss_cnt) \
```
victorjulien: Is there a reason for making this a macro instead of a function?

jlucovsky: Given your comments about using JIT only in `detect-pcre.c`, I plan to remove the macro but will retain the "cleanup" related changes in the detect modules.

I'll ensure that only uses from `detect-pcre.c` enable/allocate/assign the JIT stack as well.

Will your OS changes (e.g., result of `PageSupportsRWX` at https://github.com/OISF/suricata/blob/master/src/detect-pcre.c#L147) need to be retained?

victorjulien: Probably, yes.

jlucovsky:
Continued in #4318

PR #4305, opened 2019-10-14
detect: return failure instead of aborting

catenacyber:
None

Describe changes:
- `DetectAddressParseString` returns failure instead of aborting

Found by fuzzing

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -533,7 +533,9 @@ int DetectAddressParseString(DetectAddress *dd, const char *str)
         str++;
 
     /* shouldn't see 'any' here */
-    BUG_ON(strcasecmp(str, "any") == 0);
+    if(strcasecmp(str, "any") == 0) {
+        return -1;
+    }
 
     strlcpy(ipstr, str, sizeof(ipstr));
     SCLogDebug("str %s", str);


src/detect-engine-address.c
```diff
@@ -533,7 +533,9 @@ int DetectAddressParseString(DetectAddress *dd, const char *str)
         str++;
 
     /* shouldn't see 'any' here */
-    BUG_ON(strcasecmp(str, "any") == 0);
+    if(strcasecmp(str, "any") == 0) {
```
victorjulien: I would rather see with that input we end up here, as the comment tells us this should be impossible

catenacyber: I understand the comment as "no one should write a rule with any here" rather than it is impossible to have any here (I need some work to regenerate an input for this case)

victorjulien: BUG_ON is not meant for user reachable errors, but for actual code bugs.

catenacyber: That is the point of my PR :-)
But I now think you mean that there may be a deeper bug...
I will try to find again the reproducer

victorjulien:
Can you open a ticket for this case?

catenacyber:
I created https://redmine.openinfosecfoundation.org/issues/3468

Why did you close the PR ?

victorjulien:
> I created https://redmine.openinfosecfoundation.org/issues/3468
> 
> Why did you close the PR ?

It fixes a symptom, not the root cause. Lets see if that can be fixed first.

catenacyber:
Ok, I will try to reproduce it to get a stack trace and see which caller of `DetectAddressParseSingle` is wrong

PR #4312, opened 2019-10-16
Experiment/git version/v1

victorjulien:
See git msgs for an explanation.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,4 @@
-    AC_INIT([suricata],[5.0.0-dev])
+    AC_INIT([suricata],[5.0.0])
     m4_ifndef([AM_SILENT_RULES], [m4_define([AM_SILENT_RULES],[])])AM_SILENT_RULES([yes])
     AC_CONFIG_HEADERS([config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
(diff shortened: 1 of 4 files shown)


victorjulien:
Guess I'm over thinking this. Goal is to make it clear that ppl are not running a release.

So I guess this scheme would work too. Example for 5.0.2

1. set AC_INIT version to 5.0.2
2. tag release 'suricata-5.0.2'
3. set AC_INIT version to 5.0.3-dev

jasonish:
Yes, its hard to avoid that version set and commit dance.

Something else we need to capture is that generally 5.0.0-dev is less than 5.0.0, or 5.0.1. Which is something we don't capture in the 4.1.x branch.

victorjulien:
What do you mean by '5.0.0-dev is less than 5.0.0, or 5.0.1'?

jasonish:
> What do you mean by '5.0.0-dev is less than 5.0.0, or 5.0.1'?

While I think semver treats "5.0.0-dev" as less than "5.0.0" (as its most likely a pre-release), visually it looks this way as well.

This is more of a think with master-4.1.x branch.  If I check that branch out I get "4.1.0-dev", which visually looks less than "4.1.0", yet master-4.1.x is "4.1.5".

victorjulien:
So then my suggestion makes sense still right? 5.0.1-dev is < 5.0.1, then after 5.0.1 is released it gets set to 5.0.2-dev which is > 5.0.1 but < 5.0.2.

victorjulien:
(btw lets ignore 4.1.x for now, once we settle on a routine we can see if/how we can apply it to 4.1.x or just stick with the current until that branch goes EOL)

jasonish:
> So then my suggestion makes sense still right? 5.0.1-dev is < 5.0.1, then after 5.0.1 is released it gets set to 5.0.2-dev which is > 5.0.1 but < 5.0.2.

Yes, your approach should capture that.

PR #4313, opened 2019-10-16
eve/dns: don't log warning if dns log version not set - v1

jasonish:
If the DNS log version is not set, we default to v2. This should
not be warning, but better logged at the config level.

A warning will still be logged if the value is set but is not
1 or 2.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/402
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/758


victorjulien:
Merged in #4319, thanks Jason.

I do think we need a ticket for this as well.

PR #4316, opened 2019-10-17
[RFC] conf: add value checks for datasets hash/prealloc

jmtaylor90:

https://redmine.openinfosecfoundation.org/issues/3240
Describe changes:
- add function to check for values of configuration values in yaml for hash
- have hash-size use validation function
- have prealloc use validation function

### DIFF
--- a/src/util-thash.c
+++ b/src/util-thash.c
@@ -204,6 +204,19 @@ static void THashDataFree(THashTableContext *ctx, THashData *h)
 #define GET_VAR(prefix,name) \
     snprintf(varname, sizeof(varname), "%s.%s", (prefix), (name))
 
+static void THashConfigValidate(const char *confvalue, char *varname)
+{
+  int varlen = strlen(confvalue);
+  for (int i=0; i<varlen; i++)
+    if (!isdigit(confvalue[i]))
+    {
+      SCLogError(SC_ERR_SIZE_PARSE, "Error parsing %s "
+                 "from key %s. Killing Engine",
+                 confvalue, varname);
+      exit(EXIT_FAILURE);
+    }
+}
+
 /** \brief initialize the configuration
  *  \warning Not thread safe */
 static void THashInitConfig(THashTableContext *ctx, const char *cnf_prefix)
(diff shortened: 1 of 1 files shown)


jmtaylor90:
Was just looking at the configuration while testing datasets and wondered if this would be a legitimate fix.

src/util-thash.c
```diff
@@ -204,6 +204,19 @@ static void THashDataFree(THashTableContext *ctx, THashData *h)
 #define GET_VAR(prefix,name) \
     snprintf(varname, sizeof(varname), "%s.%s", (prefix), (name))
 
+static void THashConfigValidate(const char *confvalue, char *varname)
+{
+  int varlen = strlen(confvalue);
```
victorjulien: I was thinking ``ByteExtractStringUint32`` would do this for us, but it quite explicitly does not:
```c
    /* This will interfere with some rules that do not know the length
     * in advance and instead are just using the max.
     */
#if 0
    else if (len && *endptr != '\0') {
        fprintf(stderr, "ByteExtractString: Extra characters following numeric value\n");
        return -1;
    }
#endif
```

I guess I would like a more generic solution though, but I fear the risk of side effects if we enable it for ``ByteExtractStringUint32``.

cc: @shivan1b

jmtaylor90: Ah, so are there other places where this sort of validation could take place where it currently doesn't? I am guessing by your comment there is :)
I was also wondering if it would be worth adding a max value check to ensure values aren't larger than their type?

victorjulien: The functions are per type, so that part is mostly covered. Though in some cases the type allows for much larger values than make sense. For example the ``hash-size`` is a ``uint32_t``, but only because ``uint16_t`` is too small, not because we need the full 32 bits.

I did check ``ByteExtractString`` for enabling the 'trailing chars' check, but it can't be done in its current form. These functions are not just used by parsing rules or config, but also to extract data from packets for ``byte_jump`` and ``byte_test`` and maybe in other places. So for parsing we might need a new set of functions to do what we want.

jmtaylor90: thanks for the additional information!

victorjulien:
I've merged this for the upcoming 5.0.3 release. In master we're still working towards a different approach. Thanks Jason!

PR #4318, opened 2019-10-17
RFC part 2 -- Modifications to use pcre_jit_exec

jlucovsky:
_PREVIEW/RFC_ only

Continuation of  #4303 

I'm posting this PR to request feedback for the approach to implement use of `pcre_jit_exec` when `PCRE_HAVE_JIT` has the value `1` for `detect-pcre.c`

The unittests and sv pass with the changes;  there are a 1/2 dozen or so source code modules that always use `pcre_exec` that remain to be modified.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[749](https://redmine.openinfosecfoundation.org/projects/suricata/issues/749)

This updates
1. Removes the `PCRE_EXEC` macro
1. Adds a shared/common function `DetectParsePcreExec`
1. Refactors usage of `MAX_SUBSTRINGS` to a single place (except for 2 places where values other than `30` are used.

Does this seem like the proper approach?


src/util-error.h
```diff
@@ -34,6 +34,7 @@ typedef enum {
     SC_ERR_PCRE_GET_SUBSTRING,
     SC_ERR_PCRE_COMPILE,
     SC_ERR_PCRE_STUDY,
+    SC_ERR_PCRE_JITSTACK,
```
victorjulien: convention is to only append new codes so codes stay static

src/detect-parse.c
```diff
@@ -2365,6 +2367,9 @@ void DetectParseFreeRegexes(void)
         if (r->study) {
             pcre_free_study(r->study);
         }
+        if (r->jit_stack) {
```
victorjulien: is this still needed in this parsing work?

src/detect-parse.c
```diff
@@ -2373,38 +2378,57 @@ void DetectParseFreeRegexes(void)
...
 
-    *parse_regex_study = pcre_study(*parse_regex, 0, &eb);
+    detect_parse->study = pcre_study(detect_parse->regex, jit ? PCRE_STUDY_JIT_COMPILE : 0 , &eb);
     if (eb != NULL) {
         FatalError(SC_ERR_PCRE_STUDY, "pcre study failed: %s", eb);
     }
 
-    DetectParseRegexAddToFreeList(*parse_regex, *parse_regex_study);
+    if (jit) {
+        detect_parse->jit_stack = pcre_jit_stack_alloc(32*1024, 512*1024);
```
victorjulien: still needed?

victorjulien:
CI failures are related to jit detection it seems.

src/util-error.h
jlucovsky: Will move to the end of the enum list.

src/detect-parse.c
jlucovsky: Yes because these are helper functions and are called from code in `detect-pcre.c`

src/detect-parse.c
jlucovsky: yes, for the same reason as before. Note that `jit` will only be enabled when invoked from `detect-pcre.c`

src/util-error.h
jasonish: > convention is to only append new codes so codes stay static

I'm curious why.  Is it possible to get the code, instead of the name output in the logging where static may actually matter?

jlucovsky:
Continued in #4324

PR #4322, opened 2019-10-19
Exp/versions/v1

victorjulien:
@jasonish see this simpler approach that makes things only depend on PROG_VER (which is actually PACKAGE_VERSION from AC_INIT).


victorjulien:
Replaced by #4323

PR #4323, opened 2019-10-19
Exp/versions/v2

victorjulien:
@jasonish see this simpler approach that makes things only depend on PROG_VER (which is actually PACKAGE_VERSION from AC_INIT).

As #4322 but with a different name to try to avoid a name clash on mingw.


jasonish:
Yes, much simpler. And will work no matter how the code is obtained.

victorjulien:
Replaced by #4326

PR #4328, opened 2019-10-26
configure.ac: remove AC_CHECK_FILE

ffontaine:
The use of AC_CHECK_FILE and AC_CHECK_FILES cause the following error
when cross-compiling:

`
  configure: error: cannot check for file existence when cross compiling
`

The solution is to check for the file directly instead of using a macro.

Link to [redmine] ticket: https://redmine.openinfosecfoundation.org/issues/3281


victorjulien:
Merged as part of #4428, thanks!

PR #4336, opened 2019-10-30
Fastlog icmp code 3266 v2

catenacyber:

Describe changes:
- Uses imp type and code instead of port from union when logging to fast.log

Modifies #3266 with fixing format string and having enough space for temporary buffer

### DIFF
--- a/src/alert-fastlog.c
+++ b/src/alert-fastlog.c
@@ -150,12 +150,21 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
             } else {
                 snprintf(proto, sizeof(proto), "PROTO:%03" PRIu32, IP_GET_IPPROTO(p));
             }
+            char srcPortOrIcmp[17] = "";
+            char dstPortOrIcmp[17] = "";
+            if (IP_GET_IPPROTO(p) == IPPROTO_ICMP) {
+                snprintf(srcPortOrIcmp, sizeof(srcPortOrIcmp), "type=%03" PRIu8 "code=%03" PRIu8, p->icmp_s.type, p->icmp_s.code);
+                snprintf(dstPortOrIcmp, sizeof(dstPortOrIcmp), "type=%03" PRIu8 "code=%03" PRIu8, p->icmp_d.type, p->icmp_d.code);
+            } else {
+                snprintf(srcPortOrIcmp, sizeof(srcPortOrIcmp), "%"PRIu32, p->sp);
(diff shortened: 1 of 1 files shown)


catenacyber:
CI failure was due to unrelated network failure

src/alert-fastlog.c
```diff
@@ -141,24 +141,30 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
...
-        }
-
-        /* Create the alert string without locking. */
-        int size = 0;
-        if (likely(decoder_event == 0)) {
-            PrintBufferData(alert_buffer, &size, MAX_FASTLOG_ALERT_SIZE, 
+            char srcPortOrIcmp[17] = "";
+            char dstPortOrIcmp[17] = "";
+            if (IP_GET_IPPROTO(p) == IPPROTO_ICMP) {
+                snprintf(srcPortOrIcmp, sizeof(srcPortOrIcmp), "type=%03" PRIu8 "code=%03" PRIu8, p->icmp_s.type, p->icmp_s.code);
```
jlucovsky: Should `"code=%03"` be `" code=%03"`?

catenacyber: If we do that scripts processing fast.log with `cut` will fail...
Not sure what the best option is...

src/alert-fastlog.c
```diff
@@ -141,24 +141,30 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
...
             } else {
                 snprintf(proto, sizeof(proto), "PROTO:%03" PRIu32, IP_GET_IPPROTO(p));
             }
-        }
-
-        /* Create the alert string without locking. */
-        int size = 0;
-        if (likely(decoder_event == 0)) {
-            PrintBufferData(alert_buffer, &size, MAX_FASTLOG_ALERT_SIZE, 
+            char srcPortOrIcmp[17] = "";
```
victorjulien: style: vars are snake_case

src/alert-fastlog.c
victorjulien: This was broken in c662383b53f059dc221d74841228adda2c9a200f. Before that sp was a union with type, dp with code. So the output simply used:
```
<ip>:<type> -> <ip>:<code>
1.2.3.4:8 -> 5.6.7.8:0
```

src/alert-fastlog.c
catenacyber: ok

src/alert-fastlog.c
catenacyber: Ok, so icmp has always same type and code for source and destination ?

PR #4348, opened 2019-11-06
transform: adds url_decode keyword

catenacyber:

Describe changes:

Modifies #4347 by fixing transformed buffer size and avoiding overflow

### DIFF
--- a/doc/userguide/rules/transforms.rst
+++ b/doc/userguide/rules/transforms.rst
@@ -106,3 +106,7 @@ Example::
 
 .. note:: depends on libnss being compiled into Suricata
 
+url_decode
+-------------------
+
+Decodes url-encoded data, ie replacing '+' with space and '%HH' with its value.
--- a/src/detect-http-client-body.c
+++ b/src/detect-http-client-body.c
@@ -258,6 +258,7 @@ static InspectionBuffer *HttpClientBodyGetDataCallback(DetectEngineThreadCtx *de
     StreamingBufferGetDataAtOffset(body->sb,
             &data, &data_len, offset);
     InspectionBufferSetup(buffer, data, data_len);
+    InspectionBufferApplyTransforms(buffer, transforms);
     buffer->inspect_offset = offset;
 
     /* move inspected tracker to end of the data. HtpBodyPrune will consider
--- a/src/detect-transform-urldecode.c
+++ b/src/detect-transform-urldecode.c
@@ -1,0 +1,168 @@
+/* Copyright (C) 2019 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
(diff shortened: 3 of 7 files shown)


src/detect-transform-urldecode.c
```diff
@@ -0,0 +1,168 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Philippe Antoine <p.antoine@catenacyber.fr>
+ *
+ * Implements the urldecode transform keyword
```
jlucovsky: nit: s/urldecode/url_decode/

src/detect-http-client-body.c
```diff
@@ -258,6 +258,7 @@ static InspectionBuffer *HttpClientBodyGetDataCallback(DetectEngineThreadCtx *de
     StreamingBufferGetDataAtOffset(body->sb,
             &data, &data_len, offset);
     InspectionBufferSetup(buffer, data, data_len);
+    InspectionBufferApplyTransforms(buffer, transforms);
```
victorjulien: this should go into a separate commit as it is not related to urldecode

src/detect-transform-urldecode.c
```diff
@@ -0,0 +1,168 @@
...
+ */
+static int DetectTransformUrlDecodeSetup (DetectEngineCtx *de_ctx, Signature *s, const char *nullstr)
+{
+    SCEnter();
+    int r = DetectSignatureAddTransform(s, DETECT_TRANSFORM_URL_DECODE);
+    SCReturnInt(r);
+}
+
+// util function so as to ease reuse sometimes
+static uint32_t utilBufferUrlDecode(const uint8_t *input, const uint32_t input_len, uint8_t *output)
```
victorjulien: style: proper name would be ``UtilBufferUrlDecode`` (starting with capital U) but I prefer just dropping the ``util`` completely.

src/detect-transform-urldecode.c
```diff
@@ -0,0 +1,168 @@
...
+}
+
+static void TransformUrlDecode(InspectionBuffer *buffer)
+{
+    const uint8_t *input = buffer->inspect;
+    const uint32_t input_len = buffer->inspect_len;
+    uint8_t output[input_len]; // we can only shrink
+
+    uint32_t output_size = utilBufferUrlDecode(input, input_len, output);
+    printf("lol %d\n", output_size);
```
victorjulien: debug leftover

doc/userguide/rules/transforms.rst
```diff
@@ -106,3 +106,7 @@ Example::
 
 .. note:: depends on libnss being compiled into Suricata
 
+url_decode
+-------------------
+
+Decodes url-encoded data, ie replacing '+' with space and '%HH' with its value.
```
victorjulien: Since the code mentions its not decoding unicode, would it make sense to mention that here as well?

victorjulien:
" Modifies #4347 by fixing transformed buffer size and avoiding overflow "

Idea: libfuzzer interface to each transform?

catenacyber:
>> " Modifies #4347 by fixing transformed buffer size and avoiding overflow "
> Idea: libfuzzer interface to each transform?

Indeed, but it would be interesting to get a sense of coverage of current fuzz targets (oss-fuzz has a nice report for coverage)

doc/userguide/rules/transforms.rst
catenacyber: Ok and I take it as a confirmation that unicode decoding should not be part of this keyword

src/detect-http-client-body.c
catenacyber: ok

src/detect-transform-urldecode.c
catenacyber: ok

src/detect-transform-urldecode.c
catenacyber: ok, sorry for the golang style

src/detect-transform-urldecode.c
catenacyber: ok

PR #4362, opened 2019-11-12
_RFC_ PCRE refactoring

jlucovsky:
Continuation of #4361

This commit is for _RFC_ to solicit feedback on
- The refactoring of PCRE related code across dozens of detect modules
- Using `pcre_jit_exec` in the PCRE detect module (https://redmine.openinfosecfoundation.org/issues/749)

This update includes:
- Use `AC_RUN` instead of `AC_COMPILE`

If the approach is solid, then a new PR will be prepared with better commit messages.


configure.ac
```diff
@@ -730,6 +730,44 @@
...
+
+       AC_MSG_CHECKING(for PCRE JIT EXEC support usability)
+       AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+                          #include <pcre.h>
+                          #include <string.h>
+                          ]],
+           [[
+           const char* regexstr = "(a|b|c|d)";
+           const char* teststr = "apple";
+           pcre *re;
```
victorjulien: can we compress this by declaring the vars on first use. Also maybe directly pass the regex and test data into the funcs? Could safe us ~6 lines?

src/detect-pcre.c
```diff
@@ -647,29 +653,38 @@ static DetectPcreData *DetectPcreParse (DetectEngineCtx *de_ctx,
                 "Falling back to regular PCRE handling (%s:%d)",
                 regexstr, de_ctx->rule_file, de_ctx->rule_line);
     }
+
+    if (jit) {
+        pd->parse_regex.jit_stack = pcre_jit_stack_alloc(32*1024, 512*124);
+        if (pd->parse_regex.jit_stack == NULL) {
+            SCLogError(SC_ERR_PCRE_JITSTACK, "pcre jit stack failed");
```
victorjulien: should this be an error or should we fall back to non-jit in this case?

src/detect-pcre.c
```diff
@@ -921,10 +931,12 @@ static void DetectPcreFree(void *ptr)
 
     DetectPcreData *pd = (DetectPcreData *)ptr;
 
-    if (pd->re != NULL)
-        pcre_free(pd->re);
-    if (pd->sd != NULL)
-        pcre_free_study(pd->sd);
+    if (pd->parse_regex.regex != NULL)
```
victorjulien: call a generic func to clean up the parse_regex here?

src/detect-pcre.c
```diff
@@ -70,15 +71,28 @@
...
 static int pcre_use_jit = 1;
 #endif
 
+/* \brief Helper function for using pcre_exec with/without JIT
+ */
+static int DetectPcreExec(DetectParseRegex *regex, const char *str, const size_t strlen,
+                            int start_offset, int options,  int *ovector, int ovector_size)
+{
+#ifdef PCRE_HAVE_JIT_EXEC
+    if (pcre_use_jit)
```
victorjulien: I suppose this could also be checked by seeing if ```regex->jit_stack``` is non-NULL?

src/detect-pcre.c
```diff
@@ -124,33 +138,27 @@ void DetectPcreRegister (void)
...
+#else
+    DetectSetupParseRegexes(PARSE_REGEX, &parse_regex);
 #endif
 
-    DetectParseRegexAddToFreeList(parse_capture_regex, parse_capture_regex_study);
+    /* setup the capture regex, as it needs PCRE_UNGREEDY we do it manually */
+    int opts = PCRE_UNGREEDY; /* pkt_http_ua should be pkt, http_ua, for this reason the UNGREEDY */
+
+#ifdef PCRE_HAVE_JIT
+    DetectSetupParseRegexesOptsWithJIT(PARSE_CAPTURE_REGEX, &parse_capture_regex, opts, pcre_use_jit == 1);
```
victorjulien: whats the purpose of this call?

victorjulien:
I would like to see a split between the changes for parsing keyword arguments, and the updates to how the ``pcre`` keyword matches traffic (using jit stack and such). The commit is currently very large, and reviewing the things that relate to detection alone becomes hard that way.

configure.ac
jlucovsky: Will do.

src/detect-pcre.c
jlucovsky: Will do.

src/detect-pcre.c
jlucovsky: Good point. In the rare case of an error, I'll emit a warning message and fall back.

src/detect-pcre.c
jlucovsky: Yes ... if fallback is supported (prev comment).

src/detect-pcre.c
jlucovsky: This creates the structure for a JIT capable exec call. Should that be used for the capture regex?

jlucovsky:
Continued in #4377

src/detect-pcre.c
victorjulien: My understanding, which may be outdated, is that the jit support and capturing substrings are mutually exclusive. Jit just doesn't support it. If this is still correct, and it should be easy to test, then having jit support in the pcre for rule parsing doesn't seem to make much sense.

It does make a lot of sense for runtime packet matching, as there we don't often do substring capture.

jlucovsky: JIT *can be* used for capturing substrings but the documentation states 

```
Just-in-time compiling is a heavyweight optimization that can greatly speed up pattern matching. However, it comes at the cost of extra processing before the match is performed. Therefore, it is of most benefit when the same pattern is going to be matched many times.
```

Thus, I'll change this in #4377 to use `DetectSetupParseRegexes` (no jit)

PR #4363, opened 2019-11-12
Fastlog icmp code 3266 v3

catenacyber:

Describe changes:
- Uses icmp type and code instead of port from union when logging to fast.log

Modifies #4336 with using the previous format <ip>:<type> -> <ip>:<code>
And using snake_case

### DIFF
--- a/src/alert-fastlog.c
+++ b/src/alert-fastlog.c
@@ -150,12 +150,18 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
             } else {
                 snprintf(proto, sizeof(proto), "PROTO:%03" PRIu32, IP_GET_IPPROTO(p));
             }
+            uint16_t src_port_or_icmp = p->sp;
+            uint16_t dst_port_or_icmp = p->dp;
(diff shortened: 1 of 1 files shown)


src/alert-fastlog.c
```diff
@@ -141,19 +141,16 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
             action = "[wDrop] ";
         }
 
-        char proto[16] = "";
+        /* Create the alert string without locking. */
+        int size = 0;
         if (likely(decoder_event == 0)) {
+            char proto[16] = "";
```
victorjulien: looking at this it seems we can do a small optimization by moving this 'proto' to the start of the func and just setting it once. We can have multiple alerts for a packet, but the ip proto will never change.

catenacyber: Ok, but should it be part of this PR ?

victorjulien: Lets do it in a separate one.

victorjulien:
Merged in #4420, thanks Philippe!

PR #4370, opened 2019-11-15
suricata-5.0.0  Lua support  Can not get cookie

t4mo:
i use suricata-5.0.0 Lua support ,but i  Can not get http_cookie。
but Available in version 4.0 ,5.0 how to get http_cookie


victorjulien:
This is not a bug tracker. See https://redmine.openinfosecfoundation.org/projects/suricata/issues

PR #4371, opened 2019-11-15
hashtable: implement different hashing functions

catenacyber:
Djb2, Murmur3 and CRC32
Use only Djb2 so far

https://redmine.openinfosecfoundation.org/issues/3322

Describe changes:
- Implements other hashing functions : Murmur3, CRC32
- Keep using Djb2 with only one implementation

This is a draft PR not to be merged.

Here are my benchmarking results.
I use the top 1 million passwords from https://github.com/danielmiessler/SecLists/blob/master/Passwords/Common-Credentials/10-million-password-list-top-1000000.txt
I computed their MD5s and SHA256s (as the hash function is used for these datasets types)

```
Hash           Time MD5       Collides MD5  Time Sha256    Collides SHA256
=============  =============  ============  ============== ==============
Murmur         274 ms         133           203 ms         30
CRC32          296 ms         106           278 ms         36
Djb2           262 ms         115           215 ms         23
```

So, there does not seem to be a clear winner.
Murmur seems faster when the data is longer (ie sha256 instead of md5) because of the init and close operations
CRC32 seems slower in every case (but can be improved with slicing by 4)
I do not know how to interpret collisions results
Any thoughts about the quality of my benchmark ?


catenacyber:
Benchamarking main is 
```
#define HASH_SIZE 64
int main() {
    FILE * f = fopen("hasha.raw", "rb");
    uint8_t * sha256 = malloc(1000000*HASH_SIZE);
    int nb = fread(sha256, HASH_SIZE, 1000000, f);
    struct timeval stop, start;
    gettimeofday(&start, NULL);
    for (int i = 0; i<nb; i++) {
        printf("hashm %x\n", HashFunctionMurmur(sha256+HASH_SIZE*i, HASH_SIZE));
    }
    gettimeofday(&stop, NULL);
    fprintf(stderr, "took m %lu us\n", (stop.tv_sec - start.tv_sec) * 1000000 + stop.tv_usec - start.tv_usec);
    gettimeofday(&start, NULL);
    for (int i = 0; i<nb; i++) {
        printf("hashc %x\n", HashFunctionCRC32(sha256+HASH_SIZE*i, HASH_SIZE));
    }
    gettimeofday(&stop, NULL);
    fprintf(stderr, "took c %lu us\n", (stop.tv_sec - start.tv_sec) * 1000000 + stop.tv_usec - start.tv_usec);
    gettimeofday(&start, NULL);
    for (int i = 0; i<nb; i++) {
        printf("hashd %x\n", HashFunctionDjb2(sha256+HASH_SIZE*i, HASH_SIZE));
    }
    gettimeofday(&stop, NULL);
    fprintf(stderr, "took d %lu us\n", (stop.tv_sec - start.tv_sec) * 1000000 + stop.tv_usec - start.tv_usec);
    free(sha256);
    return 0;
}
```

Bash one-liner to compute collisions is `grep hashm result | sort | uniq -c | awk '$1 > 1 {print $0}' | wc -l` where result is the stdout of benchmarking program (and hashm is for Murmur)

catenacyber:
compile option `-O2` seems to improve especially Murmur hash

jlucovsky:
> Benchamarking main is
> 
> ```
> #define HASH_SIZE 64
> int main() {
>     FILE * f = fopen("hasha.raw", "rb");
>     uint8_t * sha256 = malloc(1000000*HASH_SIZE);
>     int nb = fread(sha256, HASH_SIZE, 1000000, f);
>     struct timeval stop, start;
>     gettimeofday(&start, NULL);
>     for (int i = 0; i<nb; i++) {
>         printf("hashm %x\n", HashFunctionMurmur(sha256+HASH_SIZE*i, HASH_SIZE));
>     }
>     gettimeofday(&stop, NULL);
>     fprintf(stderr, "took m %lu us\n", (stop.tv_sec - start.tv_sec) * 1000000 + stop.tv_usec - start.tv_usec);
>     gettimeofday(&start, NULL);
>     for (int i = 0; i<nb; i++) {
>         printf("hashc %x\n", HashFunctionCRC32(sha256+HASH_SIZE*i, HASH_SIZE));
>     }
>     gettimeofday(&stop, NULL);
>     fprintf(stderr, "took c %lu us\n", (stop.tv_sec - start.tv_sec) * 1000000 + stop.tv_usec - start.tv_usec);
>     gettimeofday(&start, NULL);
>     for (int i = 0; i<nb; i++) {
>         printf("hashd %x\n", HashFunctionDjb2(sha256+HASH_SIZE*i, HASH_SIZE));
>     }
>     gettimeofday(&stop, NULL);
>     fprintf(stderr, "took d %lu us\n", (stop.tv_sec - start.tv_sec) * 1000000 + stop.tv_usec - start.tv_usec);
>     free(sha256);
>     return 0;
> }
> ```
> 
> Bash one-liner to compute collisions is `grep hashm result | sort | uniq -c | awk '$1 > 1 {print $0}' | wc -l` where result is the stdout of benchmarking program (and hashm is for Murmur)

I suggest you remove the `printf` calls from the timed periods. Instead, save the values within the timed regions and then display all captures values after the benchmarked time periods are complete.

catenacyber:
> I suggest you remove the printf calls from the timed periods.

Indeed it would be better, but how do you get sure that compiler does not optimize away everything ?
Anyways, since there is the same number of `printf` it should not change much the results, just reduce the uncertainty margin

jlucovsky:
> > I suggest you remove the printf calls from the timed periods.
> 
> Indeed it would be better, but how do you get sure that compiler does not optimize away everything ?
> Anyways, since there is the same number of `printf` it should not change much the results, just reduce the uncertainty margin

Moving the display statements till all of the values are computed will eliminate an element of indeterminism -- that is, the printing of the values to an output file and/or console. 

We may find that the results are not impacted with the current placement of the `printf` calls; however, if none are included in the timings, the results will be more accurate.

catenacyber:
Results without printf on sha256 :
```
took m 67890 us
took c 156471 us
took d 93517 us
```
Adding `O2` compile option
```
took m 13637 us
took c 88674 us
took d 31692 us
```
On MD5 :
```
took m 37353 us
took c 72004 us
took d 45459 us
```
On MD5 with `O2` compile option 
```
took m 7782 us
took c 34764 us
took d 14132 us
``

jlucovsky:
Murmur with `02` offers compelling results.

catenacyber:
> Murmur with 02 offers compelling results.

Did you manage to reproduce those results on your hardware ?
And there is one downside : Murmur has more collisions on my MD5s samples

victorjulien:
I can run my pcap based bench on a rebased version of this.

catenacyber:
> I can run my pcap based bench on a rebased version of this.

Please
Do you have an idea of the weight of these hash functions in the total time ?

PR #4374, opened 2019-11-18
Optimize/output/v1

victorjulien:
Various optimizations to output framework.

@jasonish can you check they changes to the 'root loggers'?


victorjulien:
Replacing this as it has an issue with unix socket mode.

PR #4377, opened 2019-11-19
detect/pcre: Use pcre_jit_exec when possible

jlucovsky:
This PR uses `pcre_jit_exec` when available with the `pcre` keyword. It also refactors like code across most of the detect modules into helper functions in `detect-parse.c`

- The refactoring of PCRE related code across dozens of detect modules
- Using `pcre_jit_exec` in the PCRE detect module 

Redmine issue:[749](https://redmine.openinfosecfoundation.org/issues/749)


jlucovsky:
Continued in #4380

PR #4381, opened 2019-11-19
Issue 3345: Add TX detect flags to DHCP - v1

jasonish:
Add TX detect flag callbacks to the DHCP parser for proper
transaction cleanup. Addresses performance issue:

Ticket 3345:
https://redmine.openinfosecfoundation.org/issues/3345

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/412
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/768

### DIFF
--- a/rust/src/applayertemplate/template.rs
+++ b/rust/src/applayertemplate/template.rs
@@ -544,6 +544,8 @@ pub unsafe extern "C" fn rs_template_register_parser() {
         set_tx_mpm_id: None,
         get_files: None,
         get_tx_iterator: Some(rs_template_state_get_tx_iterator),
+        get_tx_detect_flags: None,
+        set_tx_detect_flags: None,
     };
 
     let ip_proto_str = CString::new("tcp").unwrap();
--- a/rust/src/dhcp/dhcp.rs
+++ b/rust/src/dhcp/dhcp.rs
@@ -17,7 +17,7 @@
 
 use crate::applayer;
 use crate::core;
-use crate::core::{ALPROTO_UNKNOWN, AppProto, Flow, IPPROTO_UDP};
+use crate::core::{ALPROTO_UNKNOWN, AppProto, Flow, IPPROTO_UDP, STREAM_TOSERVER};
 use crate::core::{sc_detect_engine_state_free, sc_app_layer_decoder_events_free_events};
 use crate::dhcp::parser::*;
 use crate::log::*;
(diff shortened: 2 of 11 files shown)


victorjulien:
So if there are many protos that don't use the flags, does it mean those would all have the same issue? If so we either need to address this on a lower level, or we need to force support for the flags on the API level.

Did you create a SV test for this? It might not validate anything, but would still make analysis easier.

jasonish:
> So if there are many protos that don't use the flags, does it mean those would all have the same issue? If so we either need to address this on a lower level, or we need to force support for the flags on the API level.

I believe so as it seems to be a critical part of the transaction state machine.

When the detect flag callbacks were added was it intended that all app-layers would be updated to use them? Or is there a use case to not having these callbacks registered as well?

> 
> Did you create a SV test for this? It might not validate anything, but would still make analysis easier.

No. My current test is a pcap that takes 20 some seconds to process. But much less with this patch. Have been thinking about adding a timeout to SV tests to handle such a case.

victorjulien:
> > So if there are many protos that don't use the flags, does it mean those would all have the same issue? If so we either need to address this on a lower level, or we need to force support for the flags on the API level.
> 
> I believe so as it seems to be a critical part of the transaction state machine.
> 
> When the detect flag callbacks were added was it intended that all app-layers would be updated to use them? Or is there a use case to not having these callbacks registered as well?

One of the goals is to make sure everything that we log would also be an inspection buffer, so then every proto would need to support the flags. But as long as that hasn't been implemented yet, there can be logging-only protocols.

> > Did you create a SV test for this? It might not validate anything, but would still make analysis easier.
> 
> No. My current test is a pcap that takes 20 some seconds to process. But much less with this patch. Have been thinking about adding a timeout to SV tests to handle such a case.

Can you add it even if it is w/o any checks? It's a easy way to share test cases too :)

jasonish:
> Can you add it even if it is w/o any checks? It's a easy way to share test cases too :)

https://github.com/OISF/suricata-verify/pull/159

Will merge soon, just want a sanity check on the Python changes as I want to skip this test by default, at least until we have a fix in.

jasonish:
> One of the goals is to make sure everything that we log would also be an inspection buffer, so then every proto would need to support the flags. But as long as that hasn't been implemented yet, there can be logging-only protocols.

Got it. So any protocol that also has detect ability. I'll address the other protos this week.

PR #4382, opened 2019-11-20
RFC: github-actions: example test workflows - v1

jasonish:
This is an example of using GitHub actions to perform
the following builds:

- MacOS
- Fedora 31 (in container)
- Ubuntu 18.04 with Coccinelle
- Ubuntu 16.04, using dist tarball artifact from the
  Fedora 31 build.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1,0 +1,180 @@
+name: builds
+
+on: [push, pull_request]
+
+jobs:
+
+  fedora-31:
+    name: Fedora 31
+    runs-on: ubuntu-latest
+    container: fedora:31
+    steps:
+      - run: |
+          dnf -y install \
+                autoconf \
+                automake \
+                cargo \
+                ccache \
+                diffutils \
+                file-devel \
+                gcc \
+                gcc-c++ \
+                git \
+                jansson-devel \
+                jq \
+                lua-devel \
+                libtool \
+                libyaml-devel \
+                libnfnetlink-devel \
+                libnetfilter_queue-devel \
+                libnet-devel \
+                libcap-ng-devel \
+                libevent-devel \
+                libmaxminddb-devel \
+                libpcap-devel \
+                libtool \
+                lz4-devel \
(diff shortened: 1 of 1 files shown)


jasonish:
A PR here doesn't seem to trigger the action. It did against my own repo tho:
https://github.com/jasonish/suricata/pull/16

victorjulien:
I like it. Are there disadvantages vs using travis? Like maybe stricter resource limits?

jasonish:
> I like it. Are there disadvantages vs using travis? Like maybe stricter resource limits?

As far as I can tell, it will only do 20 concurrent jobs (5 on Mac). Each workflow has a maximum runtime of 6 hours. But otherwise no limits, extra jobs will just be queued.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#usage-limits

Other disadvantages include:
- Not able to restart just one job in the workflow.
- The UI is not quite as nice.
- No ARM builder (this is a fairly new Travis CI feature we don't use)

The pros are that it has Windows, Linux and Mac all in one place. First class Docker support so we can easily test on any distribution. Defining a build is simpler.

victorjulien:
Maybe we can expand the integration a bit further and run it parallel to travis for a few weeks. It definitely sounds interesting.

jasonish:
> Maybe we can expand the integration a bit further and run it parallel to travis for a few weeks. It definitely sounds interesting.

Any tests you'd like to see added right away? I'd like to try Windows, but that would be more time consuming.

I'm probably going to add this to Suricata-Update as well. It will really benefit from tests across multiple distros.

victorjulien:
Can we get as many from 'tier 1' in as reasonable? https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Support_Status

Windows may be a bit much work, although the appveyor config would be a good start. Since that is working well for us, I'm fine with skipping windows for this test.

jasonish:
Closing. Continued at https://github.com/OISF/suricata/pull/4392.

PR #4386, opened 2019-11-21
4.1.x: app-layer: add tx detect flags to Rust parsers - v3

jasonish:
Adds transaction detect flags to Rust parsers that didn't
already have them registered.

- https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/771
- https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/415

### DIFF
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -17,6 +17,7 @@
 
 extern crate libc;
 use std;
+use core::{STREAM_TOSERVER};
 
 #[repr(C)]
 pub struct AppLayerGetTxIterTuple {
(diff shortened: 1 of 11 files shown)


jasonish:
So it looks like detect flags are really only required for krb5 as its the only one that does detection.

rust/src/applayer.rs
```diff
@@ -97,3 +98,49 @@ macro_rules!export_tx_set_detect_state {
...
+    pub fn set(&mut self, direction: u8, flags: u64) {
+        if direction & STREAM_TOSERVER != 0 {
+            self.ts = flags;
+        } else {
+            self.tc = flags;
+        }
+    }
+
+    pub fn get(&self, direction: u8) -> u64 {
+        if (direction & STREAM_TOSERVER) != 0 {
```
jlucovsky: nit: suggest consistency with paren use ... e.g., see L110.

PR #4394, opened 2019-11-25
Check if parser supports tx detect flags.

jasonish:


src/app-layer-parser.c
```diff
@@ -1146,6 +1146,17 @@ int AppLayerParserSetTxDetectState(const Flow *f,
...
+{
+    SCEnter();
+    for (uint8_t p = 0; p < FLOW_PROTO_MAX; p++) {
+        if (alp_ctx.ctxs[p][alproto].GetTxDetectFlags != NULL) {
+            SCReturnBool(true);
+        }
+    }
+    SCReturnBool(false);
+}
+
```
jasonish: Bad hack.

src/detect-engine.c
```diff
@@ -171,6 +171,14 @@ void DetectAppLayerInspectEngineRegister(const char *name,
...
         int progress, InspectEngineFuncPtr Callback)
 {
+    if (!AppLayerParserSupportsTxDetectFlags(0, alproto)) {
+        /* FatalError(SC_ERR_INITIALIZATION, */
+        /*     "Inspect engine registered for app-layer protocol without " */
+        /*     "TX detect flag support: %s", AppProtoToString(alproto)); */
+        SCLogError(SC_ERR_INITIALIZATION,
+            "Inspect engine registered for app-layer protocol without "
+            "TX detect flag support: %s", AppProtoToString(alproto));
+    }
```
jasonish: @victorjulien Here we do not have the ipproto, but the parsers are registered on the combination of ipproto/alproto, but detect engines are only registered with alproto.

Any thoughts on how to handle this short of modifying all detect engine registrations to be ipproto/alproto?

victorjulien: Think the loop makes sense. You should probably only check the ``GetTxDetectFlags`` pointer is set if one of the mandatory fields is also set. See ``ValidateParserProto``

jasonish: All the parsers have passed ValidateParserProto by the time we do this. What would be the reason to check it again?

victorjulien: Not all the parsers register for each ipproto, so you can first check if the parser has registered for a ipproto at all, then check if it also registered the GetTxDetectFlags.

jasonish: I must be missing something.  If the parser has not registered for any ipproto, `AppLayerParserSupportsTxDetectFlags` will return false. To return true, the parser must be registered with at least one ipproto, and that one must have GetTxDetectFlags. Does this satisfy what is required?

Or should we first check that its registered for at least one ipproto, then check that at least one of the registrations has GetTxDetectFlags, but they may not be the same registered parser? I'm not sure if that makes sense to me.

victorjulien: Yeah I think you're right, it should work as is. Lets drop the `ipproto` arg to `AppLayerParserSupportsTxDetectFlags`

jasonish:
Continued at https://github.com/OISF/suricata/pull/4401

PR #4398, opened 2019-11-26
Transform urldecode 2689 v4

catenacyber:

Describe changes:

Modifies #4364 by taking comment into account and rebasing

### DIFF
--- a/src/detect-transform-urldecode.c
+++ b/src/detect-transform-urldecode.c
@@ -1,0 +1,167 @@
+/* Copyright (C) 2019 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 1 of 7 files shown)


src/detect-transform-urldecode.c
```diff
@@ -0,0 +1,167 @@
...
+#include "detect.h"
+#include "detect-engine.h"
+#include "detect-engine-prefilter.h"
+#include "detect-parse.h"
+#include "detect-transform-urldecode.h"
+
+#include "util-unittest.h"
+#include "util-print.h"
+//isxdigit
+#include <ctype.h>
```
victorjulien: already in `suricata-common.h`

victorjulien:
Not sure I understand "The use of a transform looked good to me as it can get reused and is only computed if needed, but there is the suggestion to do it like http_uri"

victorjulien:
> What should be the behavior in the case the transform fails ? (that is the buffer is not a valid url-encoded one and has patterns such as %ZZ) Right, now I keep undecoded data (like https://www.urldecoder.org for instance)

I would say keep the original content in this case.

catenacyber:
> > What should be the behavior in the case the transform fails ? (that is the buffer is not a valid url-encoded one and has patterns such as %ZZ) Right, now I keep undecoded data (like https://www.urldecoder.org for instance)
> 
> I would say keep the original content in this case.

Do you mean the whole original content ?
Like `%2A%ZZ` transforms into itself and not `*%ZZ` ?
I did "partial decode" aka "decode what you can"

> The use of a transform looked good to me as it can get reused and is only computed if needed, but there is the suggestion to do it like http_uri

I meant that a transform seems to me the right tooling for what is asked in he redmine ticket...
It seems better than a specific keyword `http_normalized_post_data`...

> Wonder if this should take an option to allow multiple levels of decoding ?

That can be added... But I am not sure it is relevant

PR #4401, opened 2019-11-27
App-Layer TX detect flags fixups - v2

jasonish:
When a detect engine is being registered, first check that
there is a parser registered for the AL proto, and if so,
make sure it has TX detect flag support.

The check to make sure the AL proto has a parser registered
is to avoid the error message when the parser is disabled
in the configuration file. I feel like there might
be a cleaner way to do this.

Note that this check is mainly for development time, and
we should not see this error in any release.

Add TX detect flags to app-layer parsers that were
missing them and register detection engines.

Related Redmine issues:
https://redmine.openinfosecfoundation.org/issues/3356

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/416
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/772

### DIFF
--- a/src/app-layer-modbus.c
+++ b/src/app-layer-modbus.c
@@ -274,6 +274,24 @@ static LoggerId ModbusGetTxLogged(void *alstate, void *vtx)
     return tx->logged;
 }
 
+static void ModbusSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags) {
+    ModbusTransaction *tx = (ModbusTransaction *)vtx;
+    if (dir & STREAM_TOSERVER) {
+        tx->detect_flags_ts = flags;
+    } else {
+        tx->detect_flags_ts = flags;
+    }
+}
+
+static uint64_t ModbusGetTxDetectFlags(void *vtx, uint8_t dir) {
+    ModbusTransaction *tx = (ModbusTransaction *)vtx;
+    if (dir & STREAM_TOSERVER) {
+        return tx->detect_flags_ts;
+    } else {
+        return tx->detect_flags_tc;
+    }
+}
+
 static uint64_t ModbusGetTxCnt(void *alstate)
 {
     return ((uint64_t) ((ModbusState *) alstate)->transaction_max);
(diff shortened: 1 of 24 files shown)


src/app-layer-parser.c
```diff
@@ -1149,6 +1149,17 @@ int AppLayerParserSetTxDetectState(const Flow *f,
     SCReturnInt(r);
 }
 
+bool AppLayerParserSupportsTxDetectFlags(AppProto alproto)
+{
+    SCEnter();
+    for (uint8_t p = 0; p < FLOW_PROTO_MAX; p++) {
```
victorjulien: minor: use `FLOW_PROTO_APPLAYER_MAX`

src/detect-engine.c
```diff
@@ -171,6 +171,13 @@ void DetectAppLayerInspectEngineRegister(const char *name,
         AppProto alproto, uint32_t dir,
         int progress, InspectEngineFuncPtr Callback)
 {
+    if (AppLayerParserForProto(alproto)) {
+        if (!AppLayerParserSupportsTxDetectFlags(alproto)) {
+            SCLogError(SC_ERR_INITIALIZATION,
```
victorjulien: I think this should be a `FatalError` or even a `BUG_ON`.

src/app-layer-parser.c
```diff
@@ -1149,6 +1149,17 @@ int AppLayerParserSetTxDetectState(const Flow *f,
     SCReturnInt(r);
 }
 
+bool AppLayerParserForProto(AppProto alproto)
```
victorjulien: Elsewhere we use
```c
    if (alproto != ALPROTO_UNKNOWN && !(AppLayerParserIsTxAware(alproto))) {
        SCLogNotice("%s logger not enabled: protocol %s is disabled",
            name, AppProtoToString(alproto));
        return -1;
    }
```
ignore the alproto check.

I have a patch for 6.0dev that renames `AppLayerParserTxAware` to `AppLayerParserIsEnabled`

src/app-layer-register.c
```diff
@@ -171,6 +171,11 @@ int AppLayerRegisterParser(const struct AppLayerParser *p, AppProto alproto)
                 p->GetTxIterator);
     }
 
+    if (p->SetTxDetectFlags && p->GetTxDetectFlags) {
```
victorjulien: do we check anywhere for the case where only one of them would be set?

jasonish: No. Adding a `BOTH_SET_OR_BOTH_UNSET` check to `ValidateParserProto` to catch the case where only one is set.

src/app-layer-parser.c
jasonish: Thanks, changed to use this. Right now trying to keep it simple to backport.

src/app-layer-modbus.c
```diff
@@ -274,6 +274,24 @@ static LoggerId ModbusGetTxLogged(void *alstate, void *vtx)
     return tx->logged;
 }
 
+static void ModbusSetTxDetectFlags(void *vtx, uint8_t dir, uint64_t flags) {
```
victorjulien: style: { on a newline

jasonish: I see others like this (like in enip). Will review everywhere for this.

Ironically I believe most of that was a complete auto-suggestion by TabNine heh.

jasonish:
Continued at https://github.com/OISF/suricata/pull/4405.

PR #4419, opened 2019-12-05
flow: expose last time as a function - v2

jasonish:
This function returns the individual components
of the timeval in output pointers making it suitable
for use over Rust FFI.

### DIFF
--- a/src/flow.c
+++ b/src/flow.c
@@ -1113,6 +1113,19 @@ void FlowUpdateState(Flow *f, enum FlowState s)
     }
 }
 
+/**
+ * \brief Get flow last time as individual values.
+ *
+ * Instead of returning a pointer to the timeval copy the timeval
+ * parts into output pointers to make it simpler to call from Rust
+ * over FFI using only basic data types.
+ */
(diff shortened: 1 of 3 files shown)


src/flow.c
```diff
@@ -1113,6 +1113,19 @@ void FlowUpdateState(Flow *f, enum FlowState s)
...
+/**
+ * \brief Get flow last time as individual values.
+ *
+ * Instead of returning a pointer to the timeval copy the timeval
+ * parts into output pointers to make it simpler to call from Rust
+ * over FFI using only basic data types.
+ */
+void FlowGetLastTimeAsParts(Flow *flow, uint64_t *secs, uint64_t *usecs)
+{
+    *secs = flow->lastts.tv_sec;
```
victorjulien: would be it better to explicitly cast tv_sec and tv_usec to uint64_t here? Think depending on arch this can be signed types

jasonish: Good idea.

jasonish:
Updated here: https://github.com/OISF/suricata/pull/4421

PR #4402, opened 2019-11-27
detect/analyzer: Suppress direction warnings

jlucovsky:
This commit suppresses direction warnings by the rules analyzer for ICMP
and ICMPV6 since it's not actionable.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:(3103)[https://redmine.openinfosecfoundation.org/issues/3103]

Describe changes:
- Suppress direction warnings for ICMP, ICMPv6

PR #4424, opened 2019-12-06
detect/analyzer: Suppress direction warnings

jlucovsky:
This commit ensures direction warnings for ICMP v4 and v6
are suppressed and corrects check so that both protocols
are checked (instead of the same protocol being checked twice).

This fixes the change from #4402 which made checked for ICMP (twice) .

Describe changes:
- Check for both v4 and v6 ICMP protocols.


victorjulien:
Merged in #4431, thanks Jeff!

PR #4425, opened 2019-12-06
perf: optimizes alert fast log

catenacyber:
None

Describe changes:
- Suggested by @victorjulien here : https://github.com/OISF/suricata/pull/4363#discussion_r354290237
- small optimization by moving this 'proto' to the start of the func and just setting it once. We can have multiple alerts for a packet, but the ip proto will never change.
- same for `src_port_or_icmp` and `dst_port_or_icmp`


jlucovsky:
See comment from #4426

PR #4426, opened 2019-12-06
Fastlog icmp code 3385 v1

catenacyber:

Describe changes:
- Uses icmp type and code instead of port from union when logging to fast.log

Backport of #4363

### DIFF
--- a/src/alert-fastlog.c
+++ b/src/alert-fastlog.c
@@ -150,12 +150,18 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
             } else {
                 snprintf(proto, sizeof(proto), "PROTO:%03" PRIu32, IP_GET_IPPROTO(p));
             }
+            uint16_t src_port_or_icmp = p->sp;
+            uint16_t dst_port_or_icmp = p->dp;
+            if (IP_GET_IPPROTO(p) == IPPROTO_ICMP) {
+                src_port_or_icmp = p->icmp_s.type;
+                dst_port_or_icmp = p->icmp_s.code;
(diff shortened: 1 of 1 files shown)


src/alert-fastlog.c
```diff
@@ -141,24 +141,27 @@ int AlertFastLogger(ThreadVars *tv, void *data, const Packet *p)
...
             }
-        }
-
-        /* Create the alert string without locking. */
-        int size = 0;
-        if (likely(decoder_event == 0)) {
-            PrintBufferData(alert_buffer, &size, MAX_FASTLOG_ALERT_SIZE, 
+            uint16_t src_port_or_icmp = p->sp;
+            uint16_t dst_port_or_icmp = p->dp;
+            if (IP_GET_IPPROTO(p) == IPPROTO_ICMP) {
```
jlucovsky: Should this handle `IPPROTO_CIMPV6`?

Suggest using if/else:
```
    if _is_icmp_
        set src and dest to type and code, respectively
   else:
        set src/dest ports
```

catenacyber: This is just a back port. Maybe we should first tackle main branch with this, don't you think ?

jlucovsky: Minimally, shouldn't both include `IPPROTO_ICMPV6`?

victorjulien: > Should this handle `IPPROTO_CIMPV6`?

Yes it should. I'm fixing this up both here and for master.

victorjulien:
Merged in #4433, thanks Philippe!

PR #4440, opened 2019-12-12
Expectation fix v1.1

regit:
Some fixes and improvement on the expectation system to avoid some problems and fix a rare memory leak (see Redmine issue).

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3378

Describe changes:
- limit expectation count to contain memory pressure
- clean expectation for flow at cleaning of the flow

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/517
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/293

### DIFF
--- a/src/app-layer-expectation.c
+++ b/src/app-layer-expectation.c
@@ -68,6 +68,7 @@ static int g_expectation_data_id = -1;
 SC_ATOMIC_DECLARE(uint32_t, expectation_count);
 
 #define EXPECTATION_TIMEOUT 30
+#define EXPECTATION_MAX_LEVEL 10
 
 typedef struct Expectation_ {
     struct timeval ts;
@@ -75,6 +76,8 @@ typedef struct Expectation_ {
     Port dp;
     AppProto alproto;
     int direction;
+    uint8_t level;
+    Flow *orig_f;
     void *data;
     struct Expectation_ *next;
 } Expectation;
@@ -188,6 +191,8 @@ int AppLayerExpectationCreate(Flow *f, int direction, Port src, Port dst,
     exp->dp = dst;
     exp->alproto = alproto;
     exp->ts = f->lastts;
+    exp->orig_f = f;
+    exp->level = 0;
     exp->data = data;
     exp->direction = direction;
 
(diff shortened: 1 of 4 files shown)


jlucovsky:
Now that the cause is known, I wonder if we could create a suricata-verify test for this.

pevma:
since this PR code update i have not been able to reproduce  the leak reported here  - https://redmine.openinfosecfoundation.org/issues/3378?next_issue_id=3377&prev_issue_id=3379

src/app-layer-expectation.c
```diff
@@ -199,9 +204,18 @@ int AppLayerExpectationCreate(Flow *f, int direction, Port src, Port dst,
 
     iexp = IPPairGetStorageById(ipp, g_expectation_id);
     exp->next = iexp;
-    IPPairSetStorageById(ipp, g_expectation_id, exp);
+    if (iexp) {
+        exp->level = iexp->level + 1;
```
victorjulien: I don't understand what is being limited here, can you explain? Does this mean we can have just 10 active ftp-data flows per ip-pair? Or is this about expectations that aren't 'consumed' yet by a flow? What happens if 10 expectations are created but only 1-9 are used. Will the 10th block any further expectations from being created? Would like to see these things explained in the commit msg and maybe also at the top of this file.

src/app-layer-expectation.c
```diff
@@ -68,13 +68,16 @@ static int g_expectation_data_id = -1;
...
+#define EXPECTATION_MAX_LEVEL 10
 
 typedef struct Expectation_ {
     struct timeval ts;
     Port sp;
     Port dp;
     AppProto alproto;
     int direction;
+    uint8_t level;
+    Flow *orig_f;
```
victorjulien: Can we do this in a way that doesn't invite ppl to actually use this flow pointer (by use I mean deref it). I understand you only use it to compare flow addresses now, but this looks like an accident waiting to happen. Flow logic is tricky enough as it is.

src/app-layer-expectation.c
regit: So we don't add the expectation if there is already 10 non converted expectation. But preparing this answer I did realize that this means, an attacker can saturate the expectation list then pass a command. This needs to be implemented with a FIFO.

victorjulien:
Replaced by #4527

PR #4454, opened 2019-12-18
Fuzz prep v5

catenacyber:
Link to redmine ticket:

- Using more `const` keywords
- new file init.c to have common initialisation

Modifies #4327 with needed rebase

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -238,6 +238,12 @@
             #include <sys/random.h> 
             ])
 
+    AC_CHECK_DECL([memmem],
+        AC_DEFINE([HAVE_MEMMEM], [1], [Use memmem]),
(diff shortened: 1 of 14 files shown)


victorjulien:
I'm not sure how I can review https://github.com/OISF/suricata/pull/4454/commits/07a6da954c3cd36e2ea7bed8548f388f2103bf1c in a sane way. The diff view is useless, so the only thing I can think of is opening init.c and suricata.c from master side by side and checking nothing is changed other than the code move.

I suppose what it shows is that commits need to be in small steps, so review is easy.

If I'm missing simply techniques for this, please let me know. Otherwise I request a new PR with this one commit broken up into bite-size chunks.

PR #4458, opened 2019-12-20
RFC: filestore: Warn if needed but not configured

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[3406](https://redmine.openinfosecfoundation.org/issues/3406)

Describe changes:
- Adds a rudimentary feature tracking system where feature providers announce features
   that can be tracked by feature consumers
- Modified filestore output to register "output::file-store" as a provided feature
- Modified filestore detect code to check if "output::file-store" is available and warn, if not.
- Modified engine analyzer to issue warning when a rule requires filestore but it's not configured

Preview -- additional work still needed:
- suricata-verify tests
- comments
- testing (1 sv-test is failing).

### DIFF
--- a/src/detect-filestore.c
+++ b/src/detect-filestore.c
@@ -35,6 +35,8 @@
 #include "detect-engine-mpm.h"
 #include "detect-engine-state.h"
 
+#include "feature.h"
+
 #include "flow.h"
 #include "flow-var.h"
 #include "flow-util.h"
@@ -322,6 +324,17 @@ continue_after_realloc_fail:
 static int DetectFilestoreSetup (DetectEngineCtx *de_ctx, Signature *s, const char *str)
 {
     SCEnter();
+
+    bool warn_not_configured = false;
+
+    if (!warn_not_configured) {
+        if (!RequiresFeature("output::file-store")) {
+            SCLogWarning(SC_WARN_ALERT_CONFIG, "One or more rule(s) depends on the "
+                         "file-store output log which is not enabled. "
+                         "Enable the output \"file-store\".");
+        }
+        warn_not_configured = true;
+    }
 
     DetectFilestoreData *fd = NULL;
     SigMatch *sm = NULL;
--- a/src/feature.c
+++ b/src/feature.c
@@ -1,0 +1,141 @@
+/* Copyright (C) 2007-2010 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Jeff Lucovsky <jeff@lucovsky.org>
+ *
+ * Implements feature tracking
+ */
+
+#include "suricata-common.h"
+#include "feature.h"
+#include "debug.h"
+
+#include "util-hash.h"
+
+typedef struct _FeatureEntryType {
+	uint16_t	id;
+	const char *feature;
+} FeatureEntryType;
+
+static SCMutex feature_table_mutex = SCMUTEX_INITIALIZER;
+static uint16_t feature_table_id = 0;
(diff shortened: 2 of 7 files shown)


src/detect-filestore.c
```diff
@@ -323,6 +325,17 @@ static int DetectFilestoreSetup (DetectEngineCtx *de_ctx, Signature *s, const ch
 {
     SCEnter();
 
+    bool warn_not_configured = false;
```
victorjulien: this will lead to a check per rule with the filestore keyword, right? Making it static will make it a one time check?

Need to consider rule reloads. The output will still be disabled, but we may want to warn at least once per ruleset.

src/feature.c
```diff
@@ -0,0 +1,141 @@
+/* Copyright (C) 2007-2010 Open Information Security Foundation
```
victorjulien: 2019

src/feature.c
```diff
@@ -0,0 +1,141 @@
...
+ */
+
+#include "suricata-common.h"
+#include "feature.h"
+#include "debug.h"
+
+#include "util-hash.h"
+
+typedef struct _FeatureEntryType {
+	uint16_t	id;
```
victorjulien: style: 'id' alignment

src/feature.c
```diff
@@ -0,0 +1,141 @@
...
+	const char *feature;
+} FeatureEntryType;
+
+static SCMutex feature_table_mutex = SCMUTEX_INITIALIZER;
+static uint16_t feature_table_id = 0;
+static HashTable *feature_hash_table;
```
victorjulien: Might want to use a `HashListTable` so we could (later) add a `--dump-features` command that walks the list.

src/feature.c
```diff
@@ -0,0 +1,141 @@
...
+        SCFree((void *)f->feature);
+    }
+    SCFree(data);
+}
+
+static void FeatureInit(void) {
+    feature_hash_table = HashTableInit(256, FeatureHashFunc,
+                                       FeatureHashCompareFunc,
+                                       FeatureHashFreeFunc);
+    BUG_ON(feature_hash_table == NULL);
```
victorjulien: `FatalError`

src/feature.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+static void FeatureAddEntry(const char *feature_name)
+{
+    FeatureEntryType *feature = SCCalloc(1, sizeof(*feature));
+    BUG_ON(feature == NULL);
```
victorjulien: `FatalError` (and the same for the rest of the `BUG_ON` statements)

src/feature.c
```diff
@@ -0,0 +1,141 @@
...
+    feature->feature = SCStrdup(feature_name);
```
victorjulien: Missing error check. Coccinelle complains about this as well.

src/feature.c
victorjulien: Actually, what is the use of this `id`?

src/feature.c
```diff
@@ -0,0 +1,141 @@
...
+ * Implements feature tracking
+ */
+
+#include "suricata-common.h"
+#include "feature.h"
+#include "debug.h"
+
+#include "util-hash.h"
+
+typedef struct _FeatureEntryType {
```
victorjulien: I don't remember why I started with the trailing underscore in this case, but I do remember why it is trailing and not leading. See for example https://doxygen.openinfosecfoundation.org/annotated.html The sort order is alphabetical and the leading underscores affect this.

But recently I started leaving the underscore out completely, so I think we can just drop it. If there is a reason to add it, please make it trailing.

src/feature.c
victorjulien: Wondering though, will we ever need heap allocated features? If not, we can just store the pointers to the constant strings.

src/detect-filestore.c
jlucovsky: I intended for this to be `static` -- will fix and also investigate the rule reload mechanism.

src/feature.c
jlucovsky: Good point ... We could add features to `feature.h` with `#define` to encourage that practice but would we ever be able to detect if it wasn't?

src/feature.c
jlucovsky: Will do.

src/feature.c
jlucovsky: Will do.

src/feature.c
jlucovsky: Will do.

I'll add `--dump-features` as part of this pr?

src/feature.c
jlucovsky: Good point. In hindsight, it's not used so I'm going to remove it.

src/feature.c
jlucovsky: Will address.

src/detect-filestore.c
jlucovsky: Will work this in.

PR #4459, opened 2019-12-20
New ByteParseString* functions

inashivb:
Add a new set of ByteExtractParse* functions and use them for strict parsing in all parser, configuration and setup functions.

### DIFF
--- a/src/defrag-hash.c
+++ b/src/defrag-hash.c
@@ -195,7 +195,7 @@ void DefragInitConfig(char quiet)
     }
     if ((ConfGet("defrag.hash-size", &conf_val)) == 1)
     {
-        if (ByteExtractStringUint32(&configval, 10, strlen(conf_val),
+        if (ByteParseStringUint32(&configval, 10, strlen(conf_val),
                                     conf_val) > 0) {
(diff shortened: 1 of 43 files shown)


victorjulien:
The prototype declarations should not be in a separate commit. Please squash https://github.com/OISF/suricata/pull/4459/commits/88c6413f66273904e9ed18a30ce72ac6d8f83e0e and https://github.com/OISF/suricata/pull/4459/commits/7f05211999ca780e132953a205bf39b9ff62911d

victorjulien:
I think it looks good overall.

I do question my own suggestion of 'ByteParse..' naming. These are all string parsers I think, so would a naming scheme like 'StringParse...' make more sense? E.g. `StringParseUint16`.

inashivb:
> I think it looks good overall.
> 
> I do question my own suggestion of 'ByteParse..' naming. These are all string parsers I think, so would a naming scheme like 'StringParse...' make more sense? E.g. `StringParseUint16`.

Yes, makes sense. Making the changes.

inashivb:
Replaced with #4462

PR #4462, opened 2019-12-21
New StringParse* functions

inashivb:

### DIFF
--- a/src/util-byte.c
+++ b/src/util-byte.c
@@ -182,7 +182,7 @@ int ByteExtractUint16(uint16_t *res, int e, uint16_t len, const uint8_t *bytes)
     return ret;
 }
 
-int ByteExtractString(uint64_t *res, int base, uint16_t len, const char *str)
+int ByteExtractString(uint64_t *res, int base, uint16_t len, const char *str, bool strict)
 {
     const char *ptr = str;
     char *endptr = NULL;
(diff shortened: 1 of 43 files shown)


src/util-byte.c
```diff
@@ -223,29 +223,24 @@ int ByteExtractString(uint64_t *res, int base, uint16_t len, const char *str)
         SCLogDebug("invalid numeric value");
         return -1;
     }
-    /* This will interfere with some rules that do not know the length
-     * in advance and instead are just using the max.
-     */
-#if 0
-    else if (len && *endptr != '\0') {
+    else if (strict && len && *endptr != '\0') {
         fprintf(stderr, "ByteExtractString: Extra characters following numeric value\n");
```
victorjulien: This should be a `SCLogError` call instead.

src/util-byte.c
```diff
@@ -337,15 +396,10 @@ int ByteExtractStringSigned(int64_t *res, int base, uint16_t len, const char *st
         SCLogError(SC_ERR_INVALID_NUMERIC_VALUE, "Invalid numeric value");
         return -1;
     }
-    /* This will interfere with some rules that do not know the length
-     * in advance and instead are just using the max.
-     */
-#if 0
-    else if (len && *endptr != '\0') {
+    else if (strict && len && *endptr != '\0') {
         fprintf(stderr, "ByteExtractStringSigned: Extra characters following numeric value\n");
```
victorjulien: `SCLogError`

inashivb:
Replaced with #4464

PR #4467, opened 2020-01-01
python:  add --with-target-python option

ffontaine:
When installing python scripts, distutils would use the python used to
run setup.py as shebang for the scripts it installs.

However, when cross-compiling, this is most often not correct.

Instead, add a --with-target-python option to allow the user to set its
own value such as '/usr/bin/env python'. To keep current behavior,
$TARGET_PYTHON is set to $HAVE_PYTHON by default

Link to ticket: https://redmine.openinfosecfoundation.org/issues/3432

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -150,6 +150,13 @@
        echo ""
     fi
 
+    # Allow the user to specify what is the target python to use in the python
+    # scripts (can be useful when cross-compiling)
+    AC_ARG_WITH(target-python,
+           AS_HELP_STRING([--with-target-python], [Target python to use in python scripts]),
+           [TARGET_PYTHON=$withval],[TARGET_PYTHON=$HAVE_PYTHON])
+    AC_SUBST(TARGET_PYTHON)
+
     # Check for python-yaml.
     have_python_yaml="no"
     if test "x$enable_python" = "xyes"; then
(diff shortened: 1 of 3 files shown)


jasonish:
This will probably produce unexpected results. For example when python, python2 and python3 all exist, ./configure will prefer python3, but this will pick up whatever python is pointing at, that is often still Python 2.

A better fix is probably a --with-python argument to point ./configure at the python program to use which would override the auto-detection.

ffontaine:
I updated the PR as requested by adding a --with-python-shebang option

configure.ac
```diff
@@ -150,6 +150,13 @@
        echo ""
     fi
 
+    # Allow the user to specify what is the python shebang to use in the python
+    # scripts (can be useful when cross-compiling)
+    AC_ARG_WITH(python-shebang,
+           AS_HELP_STRING([--with-python-shebang], [Python shebang to use in python scripts]),
```
jasonish: I think I'd rather this be `--with-target-python`. Otherwise I don't see any issues here.

configure.ac
```diff
@@ -150,6 +150,13 @@
        echo ""
     fi
 
+    # Allow the user to specify what is the python target to use in the python
+    # scripts (can be useful when cross-compiling)
+    AC_ARG_WITH(python-target,
+           AS_HELP_STRING([--with-python-target], [Python target to use in python scripts]),
```
jasonish, line 156: `--with-target-python` please.

jasonish:
I'm running into an issue where I use this option.  The scripts, `suricatasc`, `suricatasc`, `suricata-update` do get my target #! /usr/bin/python line..  In this case my default Python is Python 3, and I'm targetting Python 3..  However, the libraries files still get install into lib/python3.7 (ie: my system python), so the scripts fail to run.

Does this occur in the cross compiling scenario as well?

victorjulien:
Please submit a new PR with the requested changes if you still would like to get this merged. Thanks!

PR #4456, opened 2019-12-19
rust: cbindgen plus 2018 edition - v17

jasonish:
Builds on cbindgen PR #4448, but also migrate Rust code
to 2018 edition.

Previous PR: https://github.com/OISF/suricata/pull/4452

Changes from last PR:
- Fix make distcheck.
- Fix output-json-tftp.c

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3308

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- Fails, required cbindgen.

PR #4448, opened 2019-12-16
build: cbindgen - v14

jasonish:
Fixes to https://github.com/OISF/suricata/pull/4443:
- Fixup cargo path for builds that require cbindgen
- Don't install cbindgen for builds from distribution
  archive.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3308

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- Fail due to cbindgen not being installed.

PR #4468, opened 2020-01-02
rust: cbindgen and 2018 edition - v18

jasonish:
Builds on cbindgen PR #4448, but also migrate Rust code
to 2018 edition.

Previous PR: https://github.com/OISF/suricata/pull/4456

Changes from last PR:
- distcheck fixes. It looks like we don't need to include Cargo.lock
  at all in the distribution. As long as the sources are vendored
  we can regenerate it offline, solving an issue with out of tree
  builds.

- Fails: requires cbindgen


victorjulien:
I worry a bit about how the `rust/gen` detection and handling with `clean`/`distclean` is going to confuse developers. I see we'd start using `make maintainer-clean`, something I had to look up as I never encountered it before.

I had an older `rust/gen` and it took me quite some time to figure out how to properly updated it.

Seems like any changes to how we use `cbindgen` might result in confusion like this.

rust/Makefile.am
```diff
@@ -27,36 +28,43 @@ if RUST_CROSS_COMPILE
 RUST_TARGET = --target $(host_triplet)
 endif
 
-all-local:
-if HAVE_PYTHON
-	cd $(top_srcdir)/rust && $(HAVE_PYTHON) ./gen-c-headers.py
+all-local: $(abs_top_builddir)/rust/Cargo.lock
+if !HAVE_RUST_HEADERS
+	cbindgen --config $(abs_top_srcdir)/rust/cbindgen.toml \
```
victorjulien: I think it would be good to `echo` something if `HAVE_RUST_HEADERS` is true. E.g. "using existing Rust bindings in `rust/gen`", maybe even with a small hint of how to update/remove.

jasonish:
It is a bit of a trick to get right.  The idea is that if we do not have the headers bundled, then `distclean` will remove them, as they are generated as part of the build. However if they are bundled, then `distclean` will not remove them.

In some ways this is better than the current master is now.  For example, if you are building from a distrubtion archive with generated headers and *do not* have Python available, you can build Suricata, but after you do a `distclean` your generated headers are gone and you won't be able to rebuild Suricata.  I think we want to avoid that.

Perhaps we need to combine the 2 approaches:
- If cbindgen is available, always use it to generate the headers like we do with Python now. Always let `distclean` remove the headers.
- If cbindgen is not available, require headers to be present, but do not remove them on `distclean`.

This means as a developer, presumably you have cbindgen installed and get the behaviour you are used to. However, if you just build from distribution tarballs, you don't have to worry about it.

rust/Makefile.am
jasonish: > I think it would be good to `echo` something if `HAVE_RUST_HEADERS` is true. E.g. "using existing Rust bindings in `rust/gen`", maybe even with a small hint of how to update/remove.

Thoughts on what the hint would be? There are no real specific commands (yet) to update/remove the generated headers.

jasonish:
Another option is to revisit how Danny originally did it as part of the build.rs, where Cargo pulls down everything it needs to build the headers instead of depending on cbindgen to be externally installed.

jasonish:
> I worry a bit about how the `rust/gen` detection and handling with `clean`/`distclean` is going to confuse developers. I see we'd start using `make maintainer-clean`, something I had to look up as I never encountered it before.

So when building from git, there is no difference, as cbindgen is required, there are no bundled headers, so they always get regenerated. Did you see different?

> 
> I had an older `rust/gen` and it took me quite some time to figure out how to properly updated it.

Did this happen when building from git? Can you describe more what you were seeing?

jasonish:
I think I have some changes that make it feel more natural.. Just running into some distcheck issues, again.

jasonish:
Continued at https://github.com/OISF/suricata/pull/4477.

PR #4470, opened 2020-01-04
analysis: exit if table entries are stale

jlucovsky:
This commit causes Suricata to exit when a buffer from the analyzer
table is not recognized.

Since the table must match what's registered, exiting will bring noticed
to the condition.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[3433](https://redmine.openinfosecfoundation.org/issues/3433)

Describe changes:
- Exit with error message if a buffer name is not recognized.


victorjulien:
Merged in #4492, thanks Jeff!

PR #4471, opened 2020-01-05
Events counter

regit:
This patchset adds basic counters on JSON events. Getting the number of generated events was the main reason behind this patch as it is an important measure when dealing with log shipping and storage.

Counters added are:
- events
- bytes
- min_size
- max_size
- avg_size

Sample output:

```
tail -n1 /tmp/eve.json | jq '.stats.json'
{
  "events": 254144,
  "bytes": 125368383,
  "min_size": 231,
  "max_size": 4998,
  "avg_size": 493
}
```

Describe changes:
- Add new counters for JSON output

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/518
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/294

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -82,12 +82,62 @@ static const char *TRAFFIC_LABEL_PREFIX = "traffic/label/";
 static size_t traffic_id_prefix_len = 0;
 static size_t traffic_label_prefix_len = 0;
 
+SC_ATOMIC_DECLARE(uint64_t, json_events_counter);
+SC_ATOMIC_DECLARE(uint64_t, json_events_bytes);
+SC_ATOMIC_DECLARE(uint64_t, json_events_max_size);
+SC_ATOMIC_DECLARE(uint64_t, json_events_min_size);
+
 void OutputJsonRegister (void)
 {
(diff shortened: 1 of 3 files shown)


victorjulien:
Why is this not done using the stats API?

regit:
@inliniac I wanted to avoid to propagate the ThreadVars to the place were the counter was to be implemented so I did pick global counters. Did I miss a way to have global counter and implement max and average ?

victorjulien:
Atomics have runtime overhead, which gets worse when NUMA is a factor. I'm actually thinking about ways to reduce existing uses of it, especially 'hot' ones like for memcaps. So adding new ones will have to come with a good reason.

The ThreadVars issue is annoying indeed. Maybe its just time to make that a `__thread` variable so we can use it in much simpler ways.

victorjulien:
> Did I miss a way to have global counter and implement max and average ?

I don't think you'd even need this if the accounting is implemented using the stats api. Compare with the packet avg and max sizes. These don't use global counters, just normal ones.

PR #4451, opened 2019-12-17
detect: Implement byte_test's bitmask option

jlucovsky:
Continuation of #4449

This PR implements `bitmask` for the `byte_test` -- missing, but documented -- according to the  behavior specified in the Snort user's manual: http://manual-snort-org.s3-website-us-east-1.amazonaws.com/node32.html#SECTION004531200000000000000

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[3283](https://redmine.openinfosecfoundation.org/issues/3283)

Describe changes:
- Fixes Travis CI build error.

PR #4473, opened 2020-01-06
detect: Implement byte_test's bitmask option

jlucovsky:
Continuation of #4451

This PR implements `bitmask` for the `byte_test` -- missing, but documented -- according to the  behavior specified in the Snort user's manual: http://manual-snort-org.s3-website-us-east-1.amazonaws.com/node32.html#SECTION004531200000000000000

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[3283](https://redmine.openinfosecfoundation.org/issues/3283)

Describe changes:
- Improve bitmask application.

### DIFF
--- a/src/detect-bytetest.c
+++ b/src/detect-bytetest.c
@@ -135,7 +135,6 @@ int DetectBytetestDoMatch(DetectEngineThreadCtx *det_ctx,
         if (ptr == NULL || len <= 0) {
             SCReturnInt(0);
         }
-        //PrintRawDataFp(stdout,ptr,len);
     }
     else {
         SCLogDebug("absolute, data->offset %"PRIi32"", data->offset);
@@ -160,7 +159,7 @@ int DetectBytetestDoMatch(DetectEngineThreadCtx *det_ctx,
         extbytes = ByteExtractStringUint64(&val, data->base,
                                            data->nbytes, (const char *)ptr);
         if (extbytes <= 0) {
-            /* strtoull() return 0 if there is no numeric value in data string */
+            /* ByteExtractStringUint64() returns 0 if there is no numeric value in data string */
             if (val == 0) {
                 SCLogDebug("No Numeric value");
                 SCReturnInt(0);
@@ -227,15 +226,15 @@ int DetectBytetestDoMatch(DetectEngineThreadCtx *det_ctx,
             }
             break;
         case DETECT_BYTETEST_OP_GE:
-        if (val >= value) {
-            match = 1;
-        }
-        break;
+            if (val >= value) {
+                match = 1;
+            }
+            break;
         case DETECT_BYTETEST_OP_LE:
-        if (val <= value) {
-            match = 1;
-        }
-        break;
+            if (val <= value) {
+                match = 1;
+            }
+            break;
         default:
             /* Should never get here as we handle this in parsing. */
             SCReturnInt(-1);
--- a/src/detect-bytetest.c
+++ b/src/detect-bytetest.c
@@ -19,6 +19,7 @@
  * \file
  *
  * \author Brian Rectanus <brectanu@gmail.com>
+ * \author Jeff Lucovsky <jeff@lucovsky.org>
  *
  * Implements byte_test keyword.
  */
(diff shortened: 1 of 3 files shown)


src/detect-bytetest.c
```diff
@@ -273,7 +292,13 @@ static DetectBytetestData *DetectBytetestParse(const char *optstr, char **value,
                    "for arg %d", i + 1);
             goto error;
         }
-        args[i] = (char *)str_ptr;
+        /* args[2] is comma separated test value, offset */
+        if (i == 2) {
+            test_value = (char *) str_ptr;
+            data_offset = SCStrdup((char *) str_ptr);
```
victorjulien: Need a result check here.

src/detect-bytetest.c
```diff
@@ -295,73 +325,103 @@ static DetectBytetestData *DetectBytetestParse(const char *optstr, char **value,
...
-        if (*offset == NULL)
-            goto error;
-    } else {
-        if (ByteExtractStringInt32(&data->offset, 0, 0, args[4]) <= 0) {
-            SCLogError(SC_ERR_INVALID_VALUE, " Malformed offset: %s", str_ptr);
-            goto error;
+    /* Offset -- note that this *also* contains test_value, offset so parse accordingly */
+    if (data_offset) {
+        char *end_ptr = test_value;
+        end_ptr = data_offset;
```
victorjulien: this looks a bit odd, why assign `test_value` first?

src/detect-bytetest.c
```diff
@@ -382,6 +442,9 @@ static DetectBytetestData *DetectBytetestParse(const char *optstr, char **value,
                 data->flags |= DETECT_BYTETEST_LITTLE;
             } else if (strcasecmp("dce", args[i]) == 0) {
                 data->flags |= DETECT_BYTETEST_DCE;
+            } else if (strncasecmp("bitmask", args[i], strlen("bitmask")) == 0) {
```
victorjulien: why not a simple strcmp instead? Are we matching a partial string?

src/detect-bytetest.c
```diff
@@ -1279,6 +1367,55 @@ static int DetectBytetestTestParse22(void)
     return result;
 }
 
+/**
+ * \test Test bitmask option.
+ */
+static int DetectBytetestTestParse23(void)
```
victorjulien: please use PASS/FAIL api in new tests

src/detect-bytetest.c
```diff
@@ -1432,6 +1569,122 @@ static int DetectByteTestTestPacket05(void)
...
+    s = DetectEngineAppendSig(de_ctx, "alert dns any any -> any any "
+                              "(msg:\"Test dns_query option\"; "
+                              "byte_test:3,=,0x01,0,bitmask 0xFF;"
+                              "sid:3;)");
+    FAIL_IF_NULL(s);
+
+    SigGroupBuild(de_ctx);
+    DetectEngineThreadCtxInit(&tv, (void *)de_ctx, (void *)&det_ctx);
+
+    FLOWLOCK_WRLOCK(&f);
```
victorjulien: these locks can be removed

src/detect-bytetest.c
```diff
@@ -1432,6 +1569,122 @@ static int DetectByteTestTestPacket05(void)
...
+    int r = AppLayerParserParse(NULL, alp_tctx, &f, ALPROTO_DNS,
+                                STREAM_TOSERVER, buf, sizeof(buf));
+    if (r != 0) {
```
victorjulien: replace block by `FAIL_IF_NOT(r == 0);`

src/detect-bytetest.c
```diff
@@ -1432,6 +1569,122 @@ static int DetectByteTestTestPacket05(void)
...
+    }
+    FLOWLOCK_UNLOCK(&f);
+
+    dns_state = f.alstate;
+    FAIL_IF_NULL(dns_state);
+
+    /* do detect */
+    SigMatchSignatures(&tv, de_ctx, det_ctx, p);
+
+    if (!(PacketAlertCheck(p, 1))) {
```
victorjulien: replace with `FAIL_IF_NOT(PacketAlertCheck(p, 1))`

src/detect-bytetest.c
```diff
@@ -1432,6 +1569,122 @@ static int DetectByteTestTestPacket05(void)
...
+        FAIL;
+    }
+
+    if (!PacketAlertCheck(p, 3)) {
+        printf("sig 3 didn't alert, but it should have: ");
+        FAIL;
+    }
+
+
+    if (alp_tctx != NULL)
```
victorjulien: if the alloc/setup funcs were checked properly, these can all be conditionless

src/detect-bytetest.c
```diff
@@ -295,73 +325,103 @@ static DetectBytetestData *DetectBytetestParse(const char *optstr, char **value,
...
-            goto error;
+    /* Offset -- note that this *also* contains test_value, offset so parse accordingly */
+    if (data_offset) {
+        char *end_ptr = test_value;
+        end_ptr = data_offset;
+        while (!(isspace((unsigned char)*end_ptr) || (*end_ptr == ','))) end_ptr++;
+        str_ptr = ++end_ptr;
+        while (isspace((unsigned char)*str_ptr) || (*str_ptr == ',')) str_ptr++;
+        end_ptr = (char *)str_ptr;
+        while (!(isspace((unsigned char)*end_ptr) || (*end_ptr == ',')) && (*end_ptr != '\0'))
```
victorjulien: this parsing is done to work around the pcre substring capture limitation? Would it be cleaner to run a second pcre on this data instead?

src/detect-bytetest.c
jlucovsky: Will add.

src/detect-bytetest.c
jlucovsky: Will clean up.

src/detect-bytetest.c
jlucovsky: Will do.

src/detect-bytetest.c
jlucovsky: WIll do.

src/detect-bytetest.c
jlucovsky: Will do.

src/detect-bytetest.c
jlucovsky: Will update.

src/detect-bytetest.c
jlucovsky: Will do.

src/detect-bytetest.c
jlucovsky: Good question. This is not a substring so it could be just `strcasecmp` (like most, but not all, of the existing checks).

I'd prefer a case-insensitive match here to `strcasecmp` would work. The original code has a case-sensitive check for `relative` and a quick cross-check with the other `byte_` parsers shows inconsistent use of the three `str*cmp` functions.

What's your preference? Case-sensitive seems correct but case-insensitive would likely be ok and be detected easily with a rule parse error.

src/detect-bytetest.c
jlucovsky: I think either solution is a bit clunky.  This approach mimics that used in `detect-bytejump.c`

src/detect-bytetest.c
jlucovsky: @victorjulien preference?

victorjulien: Lets use `strcmp` for new options. I don't want to break existing rules, so existing `strcasecmp` will have to stay as is.

jlucovsky: > why not a simple strcmp instead? Are we matching a partial string?

Sigh -- my previous answer was incorrect. Due to the PCRE limitation, the string being checked is of the form `bitmask <value>` so the `strncasecmp` is necessary.

jlucovsky:
Continued in #4554

PR #4476, opened 2020-01-07
Preview: New JsonBuilder - v8

jasonish:
Migrates DHCP logging to use the new JsonBuilder. This is based on
PR #4468 to get Rust 2018 and cbindgen support, so only look at the
commits after that to get an idea of how this looks.

Related ticket:
https://redmine.openinfosecfoundation.org/issues/2726

- Fails due to missing cbindgen


jasonish:
Sorry about the diff of rust/src/dhcp/logger.js, a rustfmt got in there.

rust/src/jsonbuilder.rs
```diff
@@ -0,0 +1,539 @@
...
+        }
+        out[offset] = b'"';
+        offset += 1;
+        self.buf.push_str(std::str::from_utf8(&out[0..offset])?);
+        Ok(())
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn scjs_new_object() -> *mut JsonBuilder {
```
victorjulien: in line with what we do elsewhere, should this be `rs_js_new_object` instead?

jasonish: That depends.. I want the API to be feel "native" when calling from C, prefixing with "rs" doesn't provide that, and now that Rust is required, and we shouldn't have Rust and C implementations of the same thing, I think its less desirable.

We could also name these SCJSNewObject() or something to fit with the C style and just suppress the Rust warnings this would give us as well.

jasonish:
Continued here: https://github.com/OISF/suricata/pull/4481

PR #4484, opened 2020-01-16
Fuzz prep v7

catenacyber:
Link to redmine ticket:

- Using more `const` keywords

Modifies #4454 with :
- removing init.c file
- adds `InitGlobal` function to suricata.c (changing some order for windows)
- adds hack not to use `main` while fuzzing with `#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`

Another way is proposed in v6 : have a different file for `main` which is just a wrapper to the existing function

I am not sure about the `SCInstance suricata` variable :
It seems to be a global variable
But some functions expect it as a parameter, even if it always refer to the global variable
Should fuzz targets use their own global variable ? (to call `PostConfLoadedSetup`)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -238,6 +238,12 @@
             #include <sys/random.h> 
             ])
 
+    AC_CHECK_DECL([memmem],
+        AC_DEFINE([HAVE_MEMMEM], [1], [Use memmem]),
+        [], [
+            #include <string.h>
+            ])
+
     AC_CHECK_FUNCS([utime])
 
     OCFLAGS=$CFLAGS
--- a/src/decode.c
+++ b/src/decode.c
@@ -656,7 +656,8 @@ inline int PacketSetData(Packet *p, const uint8_t *pktdata, uint32_t pktlen)
     if (unlikely(!pktdata)) {
         return -1;
     }
-    p->ext_pkt = (uint8_t *)pktdata;
+    // ext_pkt cannot be const (because we sometimes copy)
+    p->ext_pkt = (uint8_t *) pktdata;
     p->flags |= PKT_ZERO_COPY;
 
     return 0;
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2941,11 +2941,10 @@ static void SuricataMainLoop(SCInstance *suri)
     }
 }
 
-int main(int argc, char **argv)
-{
-    SCInstanceInit(&suricata, argv[0]);
+SuricataContext context;
 
-    SuricataContext context;
+// Global initialization common to all runmodes
+int InitGlobal(void) {
     context.SCLogMessage = SCLogMessage;
     context.DetectEngineStateFree = DetectEngineStateFree;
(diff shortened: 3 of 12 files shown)


src/suricata.h
```diff
@@ -178,6 +178,8 @@ extern volatile uint8_t suricata_ctl_flags;
 extern int g_disable_randomness;
 extern uint16_t g_vlan_mask;
 
+int InitGlobal(void);
```
victorjulien: can you move this to line 196 below. Seems out of place here.

src/suricata.c
```diff
@@ -2941,11 +2941,10 @@ static void SuricataMainLoop(SCInstance *suri)
     }
 }
 
-int main(int argc, char **argv)
-{
-    SCInstanceInit(&suricata, argv[0]);
+SuricataContext context;
 
-    SuricataContext context;
+// Global initialization common to all runmodes
```
victorjulien: Please use doxygen style: `/** \brief ... */`

src/suricata.c
```diff
@@ -2242,8 +2242,10 @@ static int MayDaemonize(SCInstance *suri)
 static int InitSignalHandler(SCInstance *suri)
 {
     /* registering signals we use */
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
victorjulien: please split this commit. It does different things:
1. make PostConfLoadedSetup global
2. fix PostConfLoadedSetup using global &suricata instead of local suri ptr
3. change main
4. change signals.

configure.ac
```diff
@@ -238,6 +238,12 @@
             #include <sys/random.h> 
             ])
 
+    AC_CHECK_DECL([memmem],
```
victorjulien: why not use `AC_CHECK_FUNCS([memmem])`

src/decode.c
```diff
@@ -656,7 +656,8 @@ inline int PacketSetData(Packet *p, const uint8_t *pktdata, uint32_t pktlen)
     if (unlikely(!pktdata)) {
         return -1;
     }
-    p->ext_pkt = (uint8_t *)pktdata;
```
victorjulien: what does 'check' in the commit message refer to?

src/suricata.h
catenacyber: ok

src/suricata.c
catenacyber: ok

src/decode.c
catenacyber: I meant some keyword for "adding compiler check" (which is what `const` keyword does)
How would you name it ?

configure.ac
catenacyber: I guess because I am no auto tools expert :-p
Changing it

src/suricata.c
catenacyber: ok

PR #4486, opened 2020-01-16
Fix out-of-source offline builds with vendored sources - v1

jasonish:
Hopefully better handling of builds done from distribution
archives with vendored sources.

- Don't include Cargo.toml, its always regenerated.
- Don't include Cargo.lock, it was only included
  when sources were vendored, and can be generate
  from the vendored sources.
- Do the build from the build directory, instead
  of from the source directory. This needs to be
  done to pick up the .cargo/config that is generated
  by ./configure. We previously did the cd to pick
  up the Cargo.lock, which meant the generated
  cargo config wasn't being picked up, and there
  currently exists no command line option to tell
  it where to look.

- Failed to connect to buildbot.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -249,6 +249,7 @@ jobs:
       - run: ./configure --enable-unittests
       - run: make -j2
       - run: make check
+      - run: make distcheck
       - name: Fetching suricata-verify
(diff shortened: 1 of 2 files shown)


jasonish:
Continued at https://github.com/OISF/suricata/pull/4490.

PR #4487, opened 2020-01-17
Rust parser for DCERPC over UDP

inashivb:

Note: Everything is specific to DCERPC over UDP only even if at certain places name reads DCERPC only.


victorjulien:
Is there a SV PR that goes with this?

inashivb:
> Is there a SV PR that goes with this?

@victorjulien no, not yet. Was planning on writing S-V tests once TCP was also accommodated in. Shall make an S-V PR for UDP specific stuff soon.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,620 @@
...
+    pub fn parse_dcerpc_udp(&mut self, input: &[u8]) -> i32 {
+        // Call header parser first
+        let mut parsed = self.parse_dcerpc_udp_header(input);
+        if parsed == -1 {
+            return -1;
+        }
+
+        // Input left after successful header parsing should be 80 less
+        // than the original length
+        let mut input_left = input.len() as i32 - 80;
```
victorjulien: could this underflow is `input.len()` is smaller than 80? I see no explicit checks for this.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,620 @@
...
+pub extern "C" fn rs_dcerpc_udp_parse(
+    _flow: *mut core::Flow,
+    state: &mut DCERPCUDPState,
+    _pstate: *mut std::os::raw::c_void,
+    input: *const u8,
+    input_len: u32,
+    _data: *mut std::os::raw::c_void,
+    _flags: u8,
+) -> i32 {
+    if input_len > 0 {
```
victorjulien: merge this and the next line into a single `if` statement?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,620 @@
...
+    state: *mut std::os::raw::c_void,
+    _tx_id: u64,
+) -> *mut DCERPCUDPState {
+    let dce_state = cast_pointer!(state, DCERPCUDPState);
+    dce_state
+}
+
+#[no_mangle]
+pub extern "C" fn rs_dcerpc_udp_get_tx_cnt(_state: *mut std::os::raw::c_void) -> u8 {
+    1
```
victorjulien: can we have multiple transactions in a single state? If so, this should return the highest tx id we've used (the API is misnamed, it doesn't return a count but instead a `max_id`)

rust/src/dcerpc/dcerpc.rs
inashivb: hmm maybe. My instinct was that input len should at least be 80 if the code reaches here else it should return -1 as header parsing will fail because of incomplete header (less than 80). What do you think?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,620 @@
...
+}
+
+#[no_mangle]
+pub extern "C" fn rs_dcerpc_udp_get_alstate_progress(
+    _tx: *mut std::os::raw::c_void,
+    _direction: u8,
+) -> u8 {
+    0
```
victorjulien: this hardcodes that a tx is never considered 'complete', except when Suricata times the flow out.

rust/src/dcerpc/dcerpc.rs
victorjulien: probably best to be explicit about it. It will be more robust especially against future changes.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,620 @@
...
+        }
+    }
+
+    fn get_hdr_pkt_type(&self) -> u8 {
+        if let Some(ref hdr) = &self.dcerpchdrudp {
+            return hdr.pkt_type;
+        }
+        // Shouldn't happen
+        return 99;
+    }
```
jasonish: I think it would be more idiomatic to return Option or Result here - probably a Result if it should never really happen, or an option if it may, but unlikely to happen.

In the C binding we may want to return a value to designate error, but in Rust we should stick to the Rust idioms for new code.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,620 @@
...
+    state: &mut DCERPCUDPState,
+    _pstate: *mut std::os::raw::c_void,
+    input: *const u8,
+    input_len: u32,
+    _data: *mut std::os::raw::c_void,
+    _flags: u8,
+) -> i32 {
+    if input_len > 0 {
+        if input != std::ptr::null_mut() {
+            let buf = unsafe { std::slice::from_raw_parts(input, input_len as usize) };
```
jasonish: I guess its now idiomatic to mark the whole function unsafe when dealing with raw pointers as clippy points this out:
```
12474 error: this public function dereferences a raw pointer but is not marked `unsafe`
12475    --> src/dcerpc/dcerpc.rs:381:59
12476     |
12477 381 |             let buf = unsafe { std::slice::from_raw_parts(input, input_len as usize) };
12478     |                                                           ^^^^^
12479     |
12480     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsaf      e_ptr_arg_deref
```

I think it makes sense in our code to follow this guideline of making the function unsafe, as our unsafe functions are generally short and easy to review.

victorjulien:
Closing

PR #4488, opened 2020-01-17
New StringParse* functions

inashivb:
Related to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3053

### DIFF
--- a/src/util-byte.c
+++ b/src/util-byte.c
@@ -182,7 +182,7 @@ int ByteExtractUint16(uint16_t *res, int e, uint16_t len, const uint8_t *bytes)
     return ret;
 }
 
-int ByteExtractString(uint64_t *res, int base, uint16_t len, const char *str)
+int ByteExtractString(uint64_t *res, int base, uint16_t len, const char *str, bool strict)
 {
     const char *ptr = str;
     char *endptr = NULL;
@@ -223,29 +223,25 @@ int ByteExtractString(uint64_t *res, int base, uint16_t len, const char *str)
         SCLogDebug("invalid numeric value");
         return -1;
     }
-    /* This will interfere with some rules that do not know the length
-     * in advance and instead are just using the max.
-     */
-#if 0
-    else if (len && *endptr != '\0') {
-        fprintf(stderr, "ByteExtractString: Extra characters following numeric value\n");
+    else if (strict && len && *endptr != '\0') {
+        SCLogError(SC_ERR_INVALID_NUMERIC_VALUE, "ByteExtractString: "
+                   "Extra characters following numeric value\n");
         return -1;
     }
-#endif
 
     return (endptr - ptr);
 }
 
 int ByteExtractStringUint64(uint64_t *res, int base, uint16_t len, const char *str)
 {
-    return ByteExtractString(res, base, len, str);
+    return ByteExtractString(res, base, len, str, false);
 }
 
 int ByteExtractStringUint32(uint32_t *res, int base, uint16_t len, const char *str)
 {
     uint64_t i64;
 
-    int ret = ByteExtractString(&i64, base, len, str);
+    int ret = ByteExtractString(&i64, base, len, str, false);
     if (ret <= 0) {
         return ret;
     }
@@ -265,7 +261,7 @@ int ByteExtractStringUint16(uint16_t *res, int base, uint16_t len, const char *s
 {
     uint64_t i64;
 
-    int ret = ByteExtractString(&i64, base, len, str);
+    int ret = ByteExtractString(&i64, base, len, str, false);
     if (ret <= 0) {
         return ret;
     }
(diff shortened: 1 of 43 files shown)


victorjulien:
So the error checking by the updated callers caught my eye. This is not something you updated here, but it does look kind of funny. We see:
`if (StringParse...) <= 0)` -> error
`if ((StringParse...) == -1)` -> error
`if ((StringParse...) < 0)` -> error
`if ((StringParse...) != strlen(..))` -> error
`if ((StringParse...) > 0)` -> no error

Header says:
```
* \return n Number of bytes extracted on success
* \return -1 On error
```
So what is the correct pattern?

inashivb:
> So the error checking by the updated callers caught my eye. This is not something you updated here, but it does look kind of funny. We see:
> `if (StringParse...) <= 0)` -> error
> `if ((StringParse...) == -1)` -> error
> `if ((StringParse...) < 0)` -> error
> `if ((StringParse...) != strlen(..))` -> error
> `if ((StringParse...) > 0)` -> no error
> 
> Header says:
> 
> ```
> * \return n Number of bytes extracted on success
> * \return -1 On error
> ```
> 
> So what is the correct pattern?

Seems like Error should be whenever the retval is -1.
I do  not know about `if ((StringParse...) != strlen(..))`. At some places, the statements are like
`if (StringParseUint16(rep_value, 10, 0, key) != (int)strlen(key))`
not sure what this means. That if key = "2", then it should fail because the length of key is 1?

victorjulien:
> > So the error checking by the updated callers caught my eye. This is not something you updated here, but it does look kind of funny. We see:
> > `if (StringParse...) <= 0)` -> error
> > `if ((StringParse...) == -1)` -> error
> > `if ((StringParse...) < 0)` -> error
> > `if ((StringParse...) != strlen(..))` -> error
> > `if ((StringParse...) > 0)` -> no error
> > Header says:
> > ```
> > * \return n Number of bytes extracted on success
> > * \return -1 On error
> > ```
> > 
> > 
> > So what is the correct pattern?
> 
> Seems like Error should be whenever the retval is -1.
> I do not know about `if ((StringParse...) != strlen(..))`. At some places, the statements are like
> `if (StringParseUint16(rep_value, 10, 0, key) != (int)strlen(key))`
> not sure what this means. That if key = "2", then it should fail because the length of key is 1?

i think it means that if the function returns the number of characters it consumed in case of success, we only consider it success if it consumed **all** the characters we fed it.

src/util-byte.c
```diff
@@ -223,29 +223,25 @@ int ByteExtractString(uint64_t *res, int base, uint16_t len, const char *str)
...
     }
-    /* This will interfere with some rules that do not know the length
-     * in advance and instead are just using the max.
-     */
-#if 0
-    else if (len && *endptr != '\0') {
-        fprintf(stderr, "ByteExtractString: Extra characters following numeric value\n");
+    else if (strict && len && *endptr != '\0') {
+        SCLogError(SC_ERR_INVALID_NUMERIC_VALUE, "ByteExtractString: "
+                   "Extra characters following numeric value\n");
```
victorjulien: Please remove the function name and the trailing \n here

src/util-byte.c
```diff
@@ -337,15 +397,11 @@ int ByteExtractStringSigned(int64_t *res, int base, uint16_t len, const char *st
...
         return -1;
     }
-    /* This will interfere with some rules that do not know the length
-     * in advance and instead are just using the max.
-     */
-#if 0
-    else if (len && *endptr != '\0') {
-        fprintf(stderr, "ByteExtractStringSigned: Extra characters following numeric value\n");
+    else if (strict && len && *endptr != '\0') {
+        SCLogError(SC_ERR_INVALID_NUMERIC_VALUE, "ByteExtractStringSigned: "
```
victorjulien: remove func name and \n

src/util-byte.c
```diff
@@ -420,6 +476,73 @@ int ByteExtractStringInt8(int8_t *res, int base, uint16_t len, const char *str)
...
+
+    ret = ByteExtractStringSigned(&i64, base, len, str, true);
+    if (ret <= 0) {
+        return ret;
+    }
+
+    *res = (int32_t)i64;
+
+    if ((int64_t)(*res) != i64) {
+        SCLogError(SC_ERR_NUMERIC_VALUE_ERANGE, "Numeric value out of range "
```
victorjulien: why are these `SCLogError` here but `SCLogDebug` for the unsigned function variants?

victorjulien:
> > Seems like Error should be whenever the retval is -1.
> > I do not know about `if ((StringParse...) != strlen(..))`. At some places, the statements are like
> > `if (StringParseUint16(rep_value, 10, 0, key) != (int)strlen(key))`
> > not sure what this means. That if key = "2", then it should fail because the length of key is 1?
> 
> i think it means that if the function returns the number of characters it consumed in case of success, we only consider it success if it consumed **all** the characters we fed it.

I suppose in strict mode this is implied. But in non-strict mode it is a way for the caller to know if there is trailing data.

src/util-byte.c
inashivb: Copy paste issue. Fixed now.

inashivb:
Replaced with #4646

PR #4491, opened 2020-01-19

catenacyber:
For discussion ,not to be merged yet

Describe changes:
- Creates some file structure for the reassembled one identified with `file_range_id`

Follows #4427 with a working implementation

This is very basic and there are checks and features still missing (cf TODOs comments).
At least, it is now working against https://github.com/OISF/suricata-verify/pull/171
Is this the right way to go ?

Another solution I think of would be to give some external program the ability to inject some kinds of pseudo packets (with the reassembled file)
This way, we would be robust if the range requests come over multiple flows, and we still can run suricata rules on the reassembled file

### DIFF
--- a/src/app-layer-htp-file.c
+++ b/src/app-layer-htp-file.c
@@ -113,6 +113,11 @@ int HTPFileOpen(HtpState *s, const uint8_t *filename, uint16_t filename_len,
         }
 
         sbcfg = &s->cfg->response.sbcfg;
+        //TODO handle intersected download
+        if (s->file_range_id != 0) {
+            FileCloseFileById(files, s->file_range_id, NULL, 0, FILE_TRUNCATED);
+            s->file_range_id = 0;
+        }
 
     } else {
         if (s->files_ts == NULL) {
@@ -225,9 +230,12 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
  *  \retval -2 error parsing
  *  \retval -3 error negative end in range
  */
-int HTPFileSetRange(HtpState *s, bstr *rawvalue)
+int HTPFileOpenWithRange(HtpState *s,const uint8_t *filename, uint16_t filename_len,
+                         const uint8_t *data, uint32_t data_len,
+                         uint64_t txid, bstr *rawvalue)
 {
     SCEnter();
+    uint16_t flags = 0;
 
     if (s == NULL) {
         SCReturnInt(-1);
@@ -235,10 +243,14 @@ int HTPFileSetRange(HtpState *s, bstr *rawvalue)
 
     FileContainer * files = s->files_tc;
     if (files == NULL) {
-        SCLogDebug("no files in state");
-        SCReturnInt(-1);
+        s->files_tc = FileContainerAlloc();
+        if (s->files_tc == NULL) {
+            SCReturnInt(-1);
+        }
+        files = s->files_tc;
     }
 
+    //TODO handle file anyways
     HtpContentRange crparsed;
     if (HTPParseContentRange(rawvalue, &crparsed) != 0) {
         SCLogDebug("parsing range failed");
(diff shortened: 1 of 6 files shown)


src/app-layer-htp.h
```diff
@@ -257,6 +257,7 @@ typedef struct HtpState_ {
     uint16_t events;
     uint16_t htp_messages_offset; /**< offset into conn->messages list */
     uint32_t file_track_id;             /**< used to assign file track ids to files */
+    uint32_t file_range_id;             /**< used to assign track ids to range file */
```
jlucovsky: nit: Suggest changing "track ids" to "ids"

src/app-layer-htp-file.c
```diff
@@ -225,20 +230,27 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
  *  \retval -2 error parsing
  *  \retval -3 error negative end in range
  */
-int HTPFileSetRange(HtpState *s, bstr *rawvalue)
+int HTPFileOpenWithRange(HtpState *s,const uint8_t *filename, uint16_t filename_len,
+                         const uint8_t *data, uint32_t data_len,
+                         uint64_t txid, bstr *rawvalue)
 {
     SCEnter();
+    uint16_t flags = 0;
```
jlucovsky: nit: no need to initialize `flags`

src/app-layer-htp.c
```diff
@@ -1765,8 +1765,15 @@ static int HtpResponseBodyHandle(HtpState *hstate, HtpTxUserData *htud,
         }
 
         if (filename != NULL) {
-            result = HTPFileOpen(hstate, filename, (uint32_t)filename_len,
-                    data, data_len, HtpGetActiveResponseTxID(hstate), STREAM_TOCLIENT);
+            //set range if present
+            htp_header_t *h_content_range = htp_table_get_c(tx->response_headers, "content-range");
+            if (h_content_range != NULL) {
+                result = HTPFileOpenWithRange(hstate, filename, (uint32_t)filename_len,
+                                              data, data_len, HtpGetActiveResponseTxID(hstate), h_content_range->value);
```
jlucovsky: niit: reformat to fit w/in 80 chars

src/app-layer-htp-file.c
```diff
@@ -248,7 +260,48 @@ int HTPFileSetRange(HtpState *s, bstr *rawvalue)
         SCLogDebug("negative end in range");
         SCReturnInt(-3);
     }
-    int retval = FileSetRange(files, crparsed.start, crparsed.end);
+
+    flags = FileFlowToFlags(s->f, STREAM_TOCLIENT);
```
jlucovsky: Will this always be true? I.e., direction to client?

src/app-layer-htp-file.c
```diff
@@ -248,7 +260,48 @@ int HTPFileSetRange(HtpState *s, bstr *rawvalue)
...
+    if (crparsed.start == 0 && crparsed.end < crparsed.size) {
+        // open another file for the whole ranges
+        if (s->file_range_id != 0) {
+            // TODO handle parallel range downloads
+            if (FileCloseFileById(files, s->file_range_id, NULL, 0, FILE_TRUNCATED) != 0) {
+                SCLogDebug("close file for range failed");
+            }
+        }
+        //TODO use a flag ?
+        s->file_range_id = ~s->file_track_id;
```
jlucovsky: `file_track_ids` are integer values ... not sure why you're doing this?

catenacyber:
Thanks @jlucovsky for your review.
What is your global opinion ?
Is this the right way to go ?

src/app-layer-htp-file.c
catenacyber: I want some ID that will not be overrun by regular/usual file ids
As the comment states, it may be better to have `s->file_range_id = FLAG_FILE_RANGES | s->file_track_id;` with FLAG_FILE_RANGES = 0x80000000

src/app-layer-htp-file.c
catenacyber: As far as I know, content-range is a server header only, so yes, direction will always be to client

src/app-layer-htp-file.c
jlucovsky: That might be clearer.

jlucovsky:
> Thanks @jlucovsky for your review.
> What is your global opinion ?
> Is this the right way to go ?

This seems like the right direction.

Will you be addressing the `TODO` comments soon?

catenacyber:
> Will you be addressing the TODO comments soon?

Not next week, but afterwards I guess

src/app-layer-htp.h
catenacyber: ok

src/app-layer-htp-file.c
catenacyber: ok

src/app-layer-htp.c
catenacyber: ok

src/app-layer-htp-file.c
catenacyber: ok

PR #4495, opened 2020-01-22
Fuzz prep v10

catenacyber:
Link to redmine ticket:

- Using more `const` keywords
- adds `InitGlobal` function
- fix `PostConfLoadedSetup` to use local suri ptr
- make `PostConfLoadedSetup` global (and accessible to fuzz targets)
- disable some signal handlers while fuzzing
-  move main to a separate file (so as to reuse code from suricata.c with fuzz targets)

Modifies #4494 with code typo fix

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1,0 +1,24 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#include "suricata.h"
+
+
+int main(int argc, char **argv)
+{
+    return SC_main(argc, argv);
+}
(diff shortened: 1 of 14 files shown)


src/main.c
```diff
@@ -0,0 +1,24 @@
+/* Copyright (C) 2020 Open Information Security Foundation
```
victorjulien: I don't really understand why this patch is necessary. Can you explain?

src/suricata.c
```diff
@@ -2996,7 +2996,7 @@ int InitGlobal(void) {
     return 0;
 }
 
-int main(int argc, char **argv)
+int SC_main(int argc, char **argv)
```
victorjulien: Not a fan of this name. We actually want to move away from the SC prefix that is used on and off.

When this is in use, in what mode is Suricata running? Normal? Fuzz target? Something else?

victorjulien:
Looking good overall, just the `SC_main` related questions.

src/suricata.c
catenacyber: > When this is in use, in what mode is Suricata running? Normal? Fuzz target? Something else?

Normal mode.

Fuzz targets do not use this, but they could cf what is done for fuzzing mysql project : https://github.com/google/oss-fuzz/blob/master/projects/mysql-server/targets/fuzz_mysqld.cc#L152
This was the quickest way to get all initialization right for this project

src/main.c
catenacyber: We want to compile suricata and the fuzz targets.
We want them to share as much code as possible.
Suricata and fuzz targets do not have the same `main` function

So, Suricata's `main` function must be isolated from the rest, in a file whose compiled object file will not be used by fuzz targets.

Other solutions were :
- to move away code from suricata.c to another file init.c : downside is git history problem
- to copy/paste init code from suricata.c to every fuzz target needing it : downside is discrepancy because of lack of code sharing

Another solution would be to move away `main` and command line parsing code to another file main.c (and not have `SC_main`)

Do you see any other way ? What do you prefer ?

victorjulien:
I merged this through #4550, thanks Philippe!

PR #4499, opened 2020-01-24
Opt/smb file tx/v3

victorjulien:
After a GAP all normal transactions are closed. File transactions
are left open as they can handle GAPs in principle. However, the
GAP might have contained the closing of a file and therefore it
may remain active until the end of the flow.

This patch introduces a time based heuristic for these transactions.
After the GAP all file transactions are stamped with the current
timestamp. If 60 seconds later a file has seen no update, its marked
as closed.

This is meant to fix resource starvation issues observed in long
running SMB sessions where packet loss was causing GAPs.

https://redmine.openinfosecfoundation.org/issues/3400

Passed.

### DIFF
--- a/rust/src/smb/smb.rs
+++ b/rust/src/smb/smb.rs
@@ -1,4 +1,4 @@
-/* Copyright (C) 2017 Open Information Security Foundation
+/* Copyright (C) 2017-2020 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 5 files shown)


rust/src/smb/smb.rs
```diff
@@ -1167,8 +1197,12 @@ impl SMBState {
                     SCLogDebug!("post_gap_housekeeping: done");
                     break;
                 }
-                if let Some(SMBTransactionTypeData::FILE(_)) = tx.type_data {
+                if let Some(SMBTransactionTypeData::FILE(ref mut f)) = tx.type_data {
                     // leaving FILE txs open as they can deal with gaps.
+                    if f.post_gap_ts != 0 {
+                        f.post_gap_ts = self.ts + 60;
+                        self.check_post_gap_file_txs = true;
+                    }
```
jasonish: Something I'm not following is how does f.post_gap_ts ever not != 0, I only see it being set if not already 0.

victorjulien: Whoops ya seeing the issue. Will submit a new PR. Thanks!

victorjulien:
Replaced by #4504

PR #4513, opened 2020-01-30
DNS clean: Move more C pieces to Rust - v2

jasonish:
Moves more of the remaining C parser bits to Rust. Most of the C
was just wrappers around Rust functions anyways.

Removes DNS app-layer events that are no longer used. But on second
thought, this might cause issues with upgrades?

Also a small DHCP cleanup where it also just wrapped Rust.

- https://redmine.openinfosecfoundation.org/issues/3234

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/439
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/795


victorjulien:
Wrt to the removed events, how will the code deal with old rule files?

victorjulien:
The DNS yaml settings probably also need some review, as I think the memcap settings there are not effective?

jasonish:
> Wrt to the removed events, how will the code deal with old rule files?

Yeah, was questioning that as I was doing up the pull request description.  I think its best to remove them from the rule file, but keep them in the code just commented as deprecated?

jasonish:
> The DNS yaml settings probably also need some review, as I think the memcap settings there are not effective?

Correct. They have no meaning in the DNS parser anymore.

victorjulien:
> > Wrt to the removed events, how will the code deal with old rule files?
> 
> Yeah, was questioning that as I was doing up the pull request description. I think its best to remove them from the rule file, but keep them in the code just commented as deprecated?

Maybe the rules should soft-fail by default, and hard-fail in strict mode? Comparable to missing classtypes.

jasonish:
> Maybe the rules should soft-fail by default, and hard-fail in strict mode? Comparable to missing classtypes.

They kind of do already, its just very ugly...

```
[67210] 30/1/2020 -- 14:56:42 - (detect-app-layer-event.c:207) <Error> (DetectAppLayerEventParseAppP2) -- [ERRCODE: SC_ERR_INVALID_SIGNATURE(39)] - app-layer-event keyword's protocol "dns" doesn't have event "unsollicited_response" registered
[67210] 30/1/2020 -- 14:56:42 - (detect-engine-loader.c:184) <Error> (DetectLoadSigFile) -- [ERRCODE: SC_ERR_INVALID_SIGNATURE(39)] - error parsing signature "alert dns any any -> any any (msg:"SURICATA DNS Unsolicited response"; flow:to_client; app-layer-event:dns.unsollicited_response; classtype:protocol-command-decode; sid:2240001; rev:2;)" from file ../master/rules/dns-events.rules at line 2
```

Would be nice if that could be cleaned up to one warning mentioning the keyword.  Thought we did something like this for something else..  Will investigate.

jasonish:
> > > Wrt to the removed events, how will the code deal with old rule files?
> > 
> > 
> > Yeah, was questioning that as I was doing up the pull request description. I think its best to remove them from the rule file, but keep them in the code just commented as deprecated?
> 
> Maybe the rules should soft-fail by default, and hard-fail in strict mode? Comparable to missing classtypes.

So perhaps something like this:
```
diff --git a/src/detect-app-layer-event.c b/src/detect-app-layer-event.c
index f7b748d64..7b3f67da8 100644
--- a/src/detect-app-layer-event.c
+++ b/src/detect-app-layer-event.c
@@ -204,10 +204,16 @@ static int DetectAppLayerEventParseAppP2(DetectAppLayerEventData *data,
     }
 
     if (r < 0) {
-        SCLogError(SC_ERR_INVALID_SIGNATURE, "app-layer-event keyword's "
+        if (SigMatchStrictEnabled(DETECT_AL_APP_LAYER_EVENT)) {
+            SCLogError(SC_ERR_UNKNOWN_VALUE, "app-layer-event keyword's "
                    "protocol \"%s\" doesn't have event \"%s\" registered",
                    alproto_name, p_idx + 1);
-        return -1;
+            return -1;
+        }
+        SCLogWarning(SC_ERR_UNKNOWN_VALUE, "app-layer-event keyword's "
+            "protocol \"%s\" doesn't have event \"%s\" registered",
+            alproto_name, p_idx + 1);
+        event_id = -1;
     }
     data->event_id = event_id;
```

Now you'll just get a warning unless you do `--strict` or `--strict=app-layer-event`.  Its a bit different than classtype as we're just setting the ID to something that should never match.

victorjulien:
> Now you'll just get a warning unless you do `--strict` or `--strict=app-layer-event`. Its a bit different than classtype as we're just setting the ID to something that should never match.

I don't like this. It will continue to have runtime overhead, which should be avoided.

rust/src/dns/dns.rs
```diff
@@ -122,18 +124,94 @@ pub const DNS_RCODE_BADTRUNC: u16 = 22;
...
+    MalformedData = 0,
+    NotRequest = 1,
+    NotResponse = 2,
+    ZFlagSet = 3,
+}
+
+impl DNSEvent {
+    pub fn to_cstring(&self) -> &str {
+        match *self {
+            DNSEvent::MalformedData => "MALFORMED_DATA\0",
```
victorjulien: off topic: can we have a macro to automatically add \0 to strings like this? `my_macro!("MALFORMED_DATA")` which would lead to `"MALFORMED_DATA\0"`?

jasonish: Possible, but I'd like to put that thought on hold.  I have another branch where we derive most of this event code, so the writer of the code doesn't have to add the \0 as well.

jasonish:
> I don't like this. It will continue to have runtime overhead, which should be avoided.

Should probably consider our options.  This would be the "classtype" way.  We could leave it as is, Suricata will still startup with some error messages, only being fatal of fatal init errors were selected.

The other option would be some special handing of deprecated values.  But I'm not sure if that would be much different than now, maybe a softer warning, but still exit if init errors are fatal?

Of course we could not remove the events, but that keeps the same runtime overhead we are trying to avoid here.

rust/src/dns/dns.rs
victorjulien: Ack.

victorjulien:
(not sure why github made my last comment a 'review')

jasonish:
Continued at: https://github.com/OISF/suricata/pull/4589

PR #4519, opened 2020-02-01
[preview] Support transform options

jlucovsky:
Continuation of #4517 

This commit adds support for transform-specific options. During Setup,
 transforms have the signature string available for options detection.
 When a transform detects an option, it should convert the option into an
 internal format and supply a pointer to this format as the last argument
 to DetectSignatureAddTransform.

 Transforms that support options must provide a function in their
 Sigmatch table entry. When the transform is freed, a pointer to the
 internal format of the option is passed to this function.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:[3199](https://redmine.openinfosecfoundation.org/issues/3199)

Describe changes:
- Infrastructure to support decode options.
- Fixed CI issue (unchecked memory alloc result)

### DIFF
--- a/src/detect-engine-mpm.c
+++ b/src/detect-engine-mpm.c
@@ -173,7 +173,7 @@ void DetectAppLayerMpmRegisterByParentId(DetectEngineCtx *de_ctx,
                 for (int i = 0; i < transforms->cnt; i++) {
                     char ttstr[64];
                     (void)snprintf(ttstr,sizeof(ttstr), "%s,",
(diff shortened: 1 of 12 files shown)


victorjulien:
I think this looks good. I would like to see a real implementation of a transform that uses this. The pcre one may be the best option there as we want it anyway.

jlucovsky:
Continued in #4647

PR #4522, opened 2020-02-04

jlucovsky:
truncated name.

Describe changes:
- Restrict file name within `name` and `filename` to `NAME_MAX` characters.

### DIFF
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -1830,11 +1830,12 @@ static int FindMimeHeader(const uint8_t *buf, uint32_t blen,
  * \param search_start The start of the search (ie. boundary=\")
  * \param search_end The end of the search (ie. \")
  * \param tlen The output length of the token (if found)
+ * \param max_len The maximum offset in which to search
  *
  * \return A pointer to the token if found, otherwise NULL if not found
  */
-static uint8_t * FindMimeHeaderToken(MimeDecField *field, const char *search_start,
-        const char *search_end, uint32_t *tlen)
+static uint8_t * FindMimeHeaderTokenRestrict(MimeDecField *field, const char *search_start,
+        const char *search_end, uint32_t *tlen, int max_len)
 {
     uint8_t *fptr, *tptr = NULL, *tok = NULL;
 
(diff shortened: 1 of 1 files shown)


victorjulien:
The appveyor failures look real, can you check those?

src/util-decode-mime.c
```diff
@@ -1830,11 +1830,12 @@ static int FindMimeHeader(const uint8_t *buf, uint32_t blen,
...
  * \param search_end The end of the search (ie. \")
  * \param tlen The output length of the token (if found)
+ * \param max_len The maximum offset in which to search
  *
  * \return A pointer to the token if found, otherwise NULL if not found
  */
-static uint8_t * FindMimeHeaderToken(MimeDecField *field, const char *search_start,
-        const char *search_end, uint32_t *tlen)
+static uint8_t * FindMimeHeaderTokenRestrict(MimeDecField *field, const char *search_start,
+        const char *search_end, uint32_t *tlen, int max_len)
```
victorjulien: `max_len` can be const. Also, why is it `int` and not `uint32_t`?

src/util-decode-mime.c
```diff
@@ -1844,8 +1845,11 @@ static uint8_t * FindMimeHeaderToken(MimeDecField *field, const char *search_sta
     fptr = FindBuffer(field->value, field->value_len, (const uint8_t *)search_start, strlen(search_start));
     if (fptr != NULL) {
         fptr += strlen(search_start); /* Start at end of start string */
-        tok = GetToken(fptr, field->value_len - (fptr - field->value), search_end,
-                &tptr, tlen);
+        uint32_t search_len = field->value_len - (fptr - field->value);
+        if (max_len) {
+            search_len = MIN((int)search_len, max_len);
```
victorjulien: I suppose by casting `search_len` to `int` it would become negative if `(fptr - field->value)` is > `field->value_len`. So then we would pass this on but use it as a `uint32_t` leaving the `max_len` parameter ineffective?

src/util-decode-mime.c
```diff
@@ -1844,8 +1845,11 @@ static uint8_t * FindMimeHeaderToken(MimeDecField *field, const char *search_sta
     fptr = FindBuffer(field->value, field->value_len, (const uint8_t *)search_start, strlen(search_start));
     if (fptr != NULL) {
         fptr += strlen(search_start); /* Start at end of start string */
-        tok = GetToken(fptr, field->value_len - (fptr - field->value), search_end,
-                &tptr, tlen);
+        uint32_t search_len = field->value_len - (fptr - field->value);
+        if (max_len) {
```
victorjulien: if `max_len` is set and the field value is bigger than it, we should set an event that we can match on from the rules.

jlucovsky:
> The appveyor failures look real, can you check those?

Yes; will fix those.

src/util-decode-mime.c
jlucovsky: Will do.

src/util-decode-mime.c
jlucovsky: `FindBuffer` won't search more than `field->value_len` bytes starting at `field->value`. Thus, `fptr - field->value` should never exceed `field->value_len`.

src/util-decode-mime.c
jlucovsky: It's used to compare with another value.

jlucovsky:
Continued in #4524

src/util-decode-mime.c
victorjulien: AFAICS the value that we compare with is `uint32_t`, which we then cast to `int`.

src/util-decode-mime.c
victorjulien: My point is that this *should* logic is irrelevant if we make the check more robust. If then for some unforeseen reason our assumption turns out false (possibly due to some future change), we still have an effective check to reduce the effect.

src/util-decode-mime.c
jlucovsky: Roger; will adjust.

PR #4523, opened 2020-02-05
Dev/nom5 v1

chifflier:

not applicable

https://redmine.openinfosecfoundation.org/issues/3128

Describe changes:
- This PR upgrades all parsers to use nom 5. This is mostly a basic conversion (also to minimize code changes), so there are still macros, but at least it is now possible to use nom 5 features. Conversions from macros to functions can be done more selectively in further PRs [*]
- Some subparsers (DER, Kerberos, SNMP) have also been upgraded to latest versions, and may provide new features
- Unit tests all pass [**], and performance tests here are showing similar performances. However, more tests would be nice

[*] Note that this also showed that the macro to functions conversion is not so easy. For example, it tends to confuse type inference in the Rust compiler, and sometimes require manual type annotations.
[**] Feedback: the rust unit tests present in the code (`cargo test`) have been especially useful to diagnose regressions. We need more of them!

### DIFF
--- a/rust/Cargo.toml.in
+++ b/rust/Cargo.toml.in
@@ -17,7 +17,7 @@ strict = []
 debug = []
 
 [dependencies]
-nom = "4.2"
+nom = "5.0"
 bitflags = "1.0"
 byteorder = "1.3"
 crc = "1.8"
@@ -27,11 +27,11 @@ num-derive = "0.2"
 num-traits = "0.2"
 widestring = "0.4"
 
-der-parser = "1.1"
-kerberos-parser = "0.2"
+der-parser = "3.0"
+kerberos-parser = "0.4"
 
-ntp-parser = "0.3"
-ipsec-parser = "0.4"
-snmp-parser = "0.3.0"
-tls-parser = "0.8"
-x509-parser = "0.4"
+ntp-parser = "0.4"
+ipsec-parser = "0.5"
+snmp-parser = "0.5"
+tls-parser = "0.9"
+x509-parser = "0.6"
--- a/rust/src/common.rs
+++ b/rust/src/common.rs
@@ -1,0 +1,12 @@
+#[macro_export]
+macro_rules! take_until_and_consume (
+ ( $i:expr, $needle:expr ) => (
+    {
+      let input: &[u8] = $i;
+
+      let (rem, res) = ::nom::take_until!(input, $needle)?;
+      let (rem, _) = ::nom::take!(rem, $needle.len())?;
+      Ok((rem, res))
+    }
+  );
+);
--- a/rust/src/dhcp/parser.rs
+++ b/rust/src/dhcp/parser.rs
@@ -18,7 +18,9 @@
 use std::cmp::min;
 
 use crate::dhcp::dhcp::*;
-use nom::*;
+use nom::IResult;
+use nom::combinator::rest;
+use nom::number::complete::{be_u8, be_u16, be_u32};
 
 pub struct DHCPMessage {
     pub header: DHCPHeader,
@@ -121,7 +123,7 @@ named!(pub parse_header<DHCPHeader>,
 named!(pub parse_clientid_option<DHCPOption>,
        do_parse!(
            code:   be_u8 >>
-           len: verify!(be_u8, |v| v > 1) >>
+           len: verify!(be_u8, |&v| v > 1) >>
            _htype: be_u8 >>
            data:   take!(len - 1) >>
                (
--- a/rust/src/dhcp/parser.rs
+++ b/rust/src/dhcp/parser.rs
@@ -20,7 +20,7 @@ use std::cmp::min;
 use crate::dhcp::dhcp::*;
 use nom::IResult;
 use nom::combinator::rest;
-use nom::number::complete::{be_u8, be_u16, be_u32};
+use nom::number::streaming::{be_u8, be_u16, be_u32};
 
 pub struct DHCPMessage {
     pub header: DHCPHeader,
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -25,6 +25,8 @@ use crate::applayer::LoggerFlags;
 use crate::core;
 use crate::dns::parser;
 
+use nom::number::complete::be_u16;
+
 /// DNS record types.
 pub const DNS_RECORD_TYPE_A           : u16 = 1;
 pub const DNS_RECORD_TYPE_NS          : u16 = 2;
@@ -443,7 +445,7 @@ impl DNSState {
 
         let mut count = 0;
         while self.request_buffer.len() > 0 {
-            let size = match nom::be_u16(&self.request_buffer) {
+            let size = match be_u16(&self.request_buffer) {
                 Ok((_, len)) => len,
                 _ => 0
             } as usize;
@@ -484,7 +486,7 @@ impl DNSState {
 
         let mut count = 0;
         while self.response_buffer.len() > 0 {
-            let size = match nom::be_u16(&self.response_buffer) {
+            let size = match be_u16(&self.response_buffer) {
                 Ok((_, len)) => len,
                 _ => 0
             } as usize;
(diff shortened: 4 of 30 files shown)


victorjulien:
I get a QA error in NFS/RPC handling. For 5.0.x I created this test:
```rust
#[cfg(test)]
mod tests {
    use crate::nfs::rpc_records::*;
    use nom::Err::Incomplete;
    use nom::Needed::Size;

    #[test]
    fn test_parse_input_too_short() {
        let buf: &[u8] = &[
            0x80, 0x0, 0x0, 0x9c, 0x8e, 0x28, 0x2, 0x7e
        ];
        let r = parse_rpc_request_partial(buf);
        match r {
            Err(Incomplete(e)) => { assert_eq!(e, Size(4)); },
             _ => { panic!("failed {:?}",r); }
       }
    }
}
```
As you can see, the data is too short for the parser so it returns Err::Incomplete as expected.

However for the git master (6.0.x), the parser returns Eof. This is a passing test:
```rust
#[cfg(test)]
mod tests {
    use crate::nfs::rpc_records::*;
    use nom::Err::Error;
    use nom::error::ErrorKind::Eof;

    #[test]
    fn test_parse_input_too_short() {
        let buf: &[u8] = &[
            0x80, 0x0, 0x0, 0x9c, 0x8e, 0x28, 0x2, 0x7e
        ];

        let r = parse_rpc_request_partial(buf);
        match r {
            Err(Error((rem, ek))) => {
                    assert_eq!(rem.len(),0);
                    assert_eq!(ek, Eof);
                },
            _ => { panic!("failed {:?}",r); }
        }
    }
}
```

So it doesn NOT return `Incomplete` as I would have expected. Any idea why?

chifflier:
@victorjulien Good catch! The fix is to use the streaming API and not the complete one:
```diff
-use nom::number::complete::be_u32;
+use nom::number::streaming::be_u32;
```

This thing is, I'm not sure this can be generalized to all calls (for ex, subparsers called in `many`, `fold` and `while`-like combinators should be complete or the whole parsing will fail.

I'm committing the above fix, but also have to check if the streaming/complete error happens in other places.

chifflier:
I pushed ad4a017 to use the streaming version where possible, to get the expected behavior. Tests are OK here (as well as `suricata-verify`)

chifflier:
The changes implemented in ad4a017 fixes all known locations of possible incomplete calls.
@victorjulien , what do you think of this fix? Can you run more tests?

victorjulien:
> The changes implemented in [ad4a017](https://github.com/OISF/suricata/commit/ad4a017abf8aed23bf49f4e6d11a1cc0c8acfa13) fixes all known locations of possible incomplete calls.
> @victorjulien , what do you think of this fix? Can you run more tests?

I had to undo this for `smb1_records.rs`, then all my tests pass. It does make me wonder if we need to do this on a per case basis. I supposed one way to check would be to see if parsers have special handling for getting `err::Incomplete`?

chifflier:
> I had to undo this for `smb1_records.rs`, then all my tests pass. It does make me wonder if we need to do this on a per case basis. I supposed one way to check would be to see if parsers have special handling for getting `err::Incomplete`?

Hmm, I am curious why `smb1_records.rs` broke (I found no calls to `many`).

For the record, this last patch is, imho, the correct way to handle incomplete reads in a streaming parser. That said, nom 5 now allows *each parser* to choose its behavior (error on incomplete, or raise incomplete). If the upper layer does not differentiate `Ok` from other results (which is usually the case, for ex in `rust/src/dhcp/parser.rs`) , it all goes to error handling.
I think the streaming version should be preferred because it gives a chance of waiting for more data (for ex when handling fragmentation) and does not change parsers that do not handle incomplete reads.

The only case where the streaming versions should not be used are inside `many` combinators: unless you specify exactly the number of items to read, the next read after the last one will always return incomplete for this combinator.

So I believe the standard method of choosing combinators should be something like:
- always prefer the streaming version
- unless inside a `many` or `fold` combinator, in that case either add `complete` of use the complete combinators (the former being maybe a better choice, since it is transitive - think of parsers calling subparsers).

So my suggestion would be to use the streaming version everywhere unless explicitly required/justified.

victorjulien:
> > I had to undo this for `smb1_records.rs`, then all my tests pass. It does make me wonder if we need to do this on a per case basis. I supposed one way to check would be to see if parsers have special handling for getting `err::Incomplete`?
> 
> Hmm, I am curious why `smb1_records.rs` broke (I found no calls to `many`).

Yeah, I don't know. I didn't investigate deeply. A test that tests many many records reported to be off, and with this change it passed again.

> For the record, this last patch is, imho, the correct way to handle incomplete reads in a streaming parser. That said, nom 5 now allows _each parser_ to choose its behavior (error on incomplete, or raise incomplete). If the upper layer does not differentiate `Ok` from other results (which is usually the case, for ex in `rust/src/dhcp/parser.rs`) , it all goes to error handling.
> I think the streaming version should be preferred because it gives a chance of waiting for more data (for ex when handling fragmentation) and does not change parsers that do not handle incomplete reads.
> 
> The only case where the streaming versions should not be used are inside `many` combinators: unless you specify exactly the number of items to read, the next read after the last one will always return incomplete for this combinator.
> 
> So I believe the standard method of choosing combinators should be something like:
> 
> * always prefer the streaming version
> * unless inside a `many` or `fold` combinator, in that case either add `complete` of use the complete combinators (the former being maybe a better choice, since it is transitive - think of parsers calling subparsers).
> 
> So my suggestion would be to use the streaming version everywhere unless explicitly required/justified.

I'm not really happy with this. I wonder if it would be better to just explicitly call the parsers in their full form directly, so its clear how they behave. It's quite inconvenient that some very important distinction in the behavior of the system requires you to know it is included in the top of the file.

victorjulien:
Perhaps we can do something like this:
```diff
diff --git a/rust/src/smb/smb1_records.rs b/rust/src/smb/smb1_records.rs
index 148631fbf..a26685711 100644
--- a/rust/src/smb/smb1_records.rs
+++ b/rust/src/smb/smb1_records.rs
@@ -19,6 +19,7 @@ use crate::smb::error::SmbError;
 use crate::log::*;
 use nom::IResult;
 use nom::combinator::rest;
+use nom::number::complete::le_u8 as le_u8c;
 use nom::number::complete::{le_u8, le_u16, le_u32, le_u64};
 use crate::smb::smb::*;
 use crate::smb::smb_records::*;
@@ -51,7 +52,7 @@ pub struct Smb1WriteRequestRecord<'a> {
 
 named!(pub parse_smb1_write_request_record<Smb1WriteRequestRecord>,
     do_parse!(
-            _wct: le_u8
+            _wct: le_u8c
         >>  fid: take!(2)
         >>  _count: le_u16
         >>  offset: le_u32
```

chifflier:
> Perhaps we can do something like this:
> 
> ```diff
> diff --git a/rust/src/smb/smb1_records.rs b/rust/src/smb/smb1_records.rs
> +use nom::number::complete::le_u8 as le_u8c;
>  use nom::number::complete::{le_u8, le_u16, le_u32, le_u64};
> ```

Sure, this is one of the nice things in nom 5: each parser combinator can choose if calling complete subparsers or not.

chifflier:
[...]
> I'm not really happy with this. I wonder if it would be better to just explicitly call the parsers in their full form directly, so its clear how they behave. It's quite inconvenient that some very important distinction in the behavior of the system requires you to know it is included in the top of the file.

Well, this is one possibility. What I would find important is to have a common convention (for ex: default to complete parsers, justify/document exceptions) for code consistency and easier reviews.

I'm fine with both kind of parsers, so I'll leave the choice to the team. In both cases, tests will be very important to ensure no regression happened.

Note: according to https://github.com/Geal/nom/blob/master/doc/upgrading_to_nom_5.md all macros default to _streaming_ version. This means that in some cases we may need to transition to functions to really get the _complete_ version.

victorjulien:
I may have somehow messed up my testing, because the tests now also pass w/o that smb1 change.

victorjulien:
Merged through #4624 (#4626), thanks a lot Pierre!

PR #4524, opened 2020-02-05
smtp/mime: Restrict file name lengths

jlucovsky:
Continuation of #4522 

This commit places restrictions on the length of the file name specified
in attachments (`name=` or `filename=`) to `NAME_MAX`. Names exceeding
these limits will be truncated and processing will continue with the
truncated name.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3409](https://redmine.openinfosecfoundation.org/issues/3265)

Describe changes:
- Address CI build errors
- Create event when name is truncated.

PR #4526, opened 2020-02-06

jlucovsky:
Continuation of #4524

truncated name.

Describe changes:
- Address review comments


src/util-decode-mime.c
```diff
@@ -1,4 +1,4 @@
-/* Copyright (C) 2012 BAE Systems
+/* Copyright (C) 2012-2020 Open Information Security Foundation
```
victorjulien: Please don't do this kind of change.

src/util-decode-mime.c
```diff
@@ -1830,30 +1830,63 @@ static int FindMimeHeader(const uint8_t *buf, uint32_t blen,
...
+    size_t ss_len = strlen(search_start);
+    fptr = FindBuffer(field->value, field->value_len, (const uint8_t *)search_start, ss_len);
     if (fptr != NULL) {
-        fptr += strlen(search_start); /* Start at end of start string */
-        tok = GetToken(fptr, field->value_len - (fptr - field->value), search_end,
-                &tptr, tlen);
+        fptr += ss_len; /* Start at end of start string */
+        uint32_t offset = fptr - field->value;
+        if (offset > field->value_len) {
+            goto error_exit;
```
victorjulien: Just `return NULL` here? Not against gotos, but this one seems unnecessary.

src/util-decode-mime.c
```diff
@@ -1912,6 +1946,11 @@ static int ProcessMimeHeaders(const uint8_t *buf, uint32_t len,
                 }
                 memcpy(entity->filename, bptr, blen);
                 entity->filename_len = blen;
+
+                if (truncated_name) {
+                    state->stack->top->data->anomaly_flags |= ANOM_LONG_FILENAME;
```
victorjulien: why is the flag set in 2 places here but not at the block around line 1999 below?

victorjulien:
Can you add test cases to trigger the checks?

src/util-decode-mime.c
jlucovsky: Will restore BAE copyright.

src/util-decode-mime.c
jlucovsky: Ok.

src/util-decode-mime.c
jlucovsky: Thank; missed that.

I'm not sure why the anomaly flags are set off of `msg` *and* `top`. That seems odd.

jlucovsky:
Continued in #4552

PR #4535, opened 2020-02-11
nfs: implement post-GAP transaction cleanup

victorjulien:
Close all prior transactions in the direction of the GAP, except the
file xfers. Those use their own logic described below.

After a GAP all normal transactions are closed. File transactions
are left open as they can handle GAPs in principle. However, the
GAP might have contained the closing of a file and therefore it
may remain active until the end of the flow.

This patch introduces a time based heuristic for these transactions.
After the GAP all file transactions are stamped with the current
timestamp. If 60 seconds later a file has seen no update, its marked
as closed.

This is meant to fix resource starvation issues observed in long
running SMB sessions where packet loss was causing GAPs. Due to the
similarity of the NFS and SMB parsers, this issue is fixed for NFS
as well in this patch.

Bug #3424.
Bug #3425.

Passed.

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -119,6 +119,10 @@ pub struct NFSTransactionFile {
     /// last xid of this file transfer. Last READ or COMMIT normally.
     pub file_last_xid: u32,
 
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged through #4537

PR #4546, opened 2020-02-13
Ssh rust v1

catenacyber:

Describe changes:

This is a draft, not to be merged before discussion.
cc @chifflier

- The commits need to be squashed
But now, I kept the history of my progression just in case it might be useful
- This PR uses nom v4. #4523 means to use nom v5. Should I rebase on top of this ? (or the other way around ?) 
- This is meant to be an exact translation. We want to add features later.
- One difference though is the handling of end of line(s) at the end of the banner. This Rust version should be robust against packet splitting between CR and LF (which should be accepted by openssh), when the current version is vulnerable because end of lines skipping https://github.com/OISF/suricata/blob/master/src/app-layer-ssh.c#L235 is only done if the first packet contains the whole banner with end of lines. I will make a suricata-verify test for this
- This works locally for me with `make check` but there seem to be Rust compilation issues on other systems for the CI. Is my system wrong ?

I will add GitHub comments on my questions in the code.
And I will likely update this creation comment


rust/src/ssh/detect.rs
```diff
@@ -0,0 +1,108 @@
...
+    buffer: *mut *const u8,
+    buffer_len: *mut u32,
+    direction: u8,
+) -> u8 {
+    let tx = cast_pointer!(tx, SSHTransaction);
+    match direction {
+        STREAM_TOSERVER => {
+            let m = &tx.cli_hdr.protover;
+            if m.len() > 0 {
+                unsafe { *buffer = m.as_ptr() };
```
catenacyber: Can we do less `unsafe` in these functions ?
Like checking if the pointer is not NULL

rust/src/ssh/parser.rs
```diff
@@ -0,0 +1,187 @@
...
+    return true;
+}
+
+//may leave \r at the end to be removed
+named!(pub ssh_parse_line<&[u8], &[u8]>,
+    terminated!(
+        take_while!(is_not_lineend),
+        alt!( tag!("\n") | tag!("\r\n") |
+              do_parse!(
+                    bytes: tag!("\r") >>
```
catenacyber: I am not sure that SSH parser should accept single CR as end of banner, but that is what is being done by the C parser

rust/src/ssh/parser.rs
```diff
@@ -0,0 +1,187 @@
...
+pub struct SshRecordHeader {
+    pub pkt_len: u32,
+    padding_len: u8,
+    pub msg_code: u8,
+}
+
+named!(pub ssh_parse_record_header<SshRecordHeader>,
+    do_parse!(
+        pkt_len: verify!(be_u32, |val:u32| val > 1) >>
+        padding_len: be_u8 >>
```
catenacyber: Have tests been made against 0-length or 1-length packets without message code ?

src/app-layer-ssh.h
```diff
@@ -48,40 +48,12 @@
...
-    uint8_t buf_offset;
-    uint8_t flags;
-    uint32_t record_left;
-    uint8_t *proto_version;
-    uint8_t *software_version;
-    uint8_t *banner_buffer;
-} SshHeader;
-
 enum {
     SSH_STATE_IN_PROGRESS,
```
catenacyber: Not sure about these : can we share an enumeration between Rust and C ? (cf SSHConnectionState)

src/detect-ssh-proto-version.c
```diff
@@ -110,34 +111,35 @@ static int DetectSshVersionMatch (DetectEngineThreadCtx *det_ctx,
...
         SCLogDebug("no ssh state, no match");
         SCReturnInt(0);
     }
 
     int ret = 0;
-    if ((flags & STREAM_TOCLIENT) && (ssh_state->srv_hdr.flags & SSH_FLAG_VERSION_PARSED)) {
-        if (ssh->flags & SSH_FLAG_PROTOVERSION_2_COMPAT) {
-            SCLogDebug("looking for ssh server protoversion 2 compat");
-            if (strncmp((char *) ssh_state->srv_hdr.proto_version, "2", 1) == 0 ||
-                strncmp((char *) ssh_state->srv_hdr.proto_version, "2.", 2) == 0 ||
```
catenacyber: This test was useless

src/app-layer-ssh.c
```diff
@@ -52,557 +53,6 @@
...
-        if (ret < 0) {
-            SCLogDebug("Invalid version string");
-            SCReturnInt(-1);
-        } else if (header->flags & SSH_FLAG_VERSION_PARSED) {
-            SCLogDebug("Version string parsed, remaining length %d", ret);
-            input += input_len - ret;
-            input_len -= (input_len - ret);
-
-            uint32_t u = 0;
-            while (u < input_len && (input[u] == '\r' || input[u] == '\n')) {
```
catenacyber: This should be done consistently against packet splitting

src/app-layer-ssh.c
```diff
@@ -695,36 +159,20 @@ static int SSHParserTest01(void)
...
-    }
-
-    if (ssh_state->cli_hdr.software_version == NULL) {
-        printf("Client version string not parsed: ");
-        goto end;
-    }
-
-    if (ssh_state->cli_hdr.proto_version == NULL) {
+    void * tx = rs_ssh_state_get_tx(ssh_state, 0);
+    if ( rs_ssh_tx_get_alstate_progress(tx, STREAM_TOSERVER) != SSH_FLAG_VERSION_PARSED ) {
```
catenacyber: Not sure between using `rs_ssh_tx_get_alstate_progress` or custom `rs_ssh_tx_get_flags`

src/app-layer-ssh.c
```diff
@@ -695,36 +159,20 @@ static int SSHParserTest01(void)
...
-        goto end;
-    }
-
-    if (ssh_state->cli_hdr.software_version == NULL) {
-        printf("Client version string not parsed: ");
-        goto end;
-    }
-
-    if (ssh_state->cli_hdr.proto_version == NULL) {
+    void * tx = rs_ssh_state_get_tx(ssh_state, 0);
```
catenacyber: All the C unit tests have been translated to use Rust instead of the C structures...

src/app-layer-ssh.c
```diff
@@ -2634,11 +1719,13 @@ static int SSHParserTest25(void)
...
     FAIL_IF(r != -1);
 
-    SshState *ssh_state = f.alstate;
+    void *ssh_state = f.alstate;
     FAIL_IF_NULL(ssh_state);
-
-    FAIL_IF_NOT(!(ssh_state->cli_hdr.flags & SSH_FLAG_VERSION_PARSED));
-    FAIL_IF(ssh_state->cli_hdr.software_version);
+    void * tx = rs_ssh_state_get_tx(ssh_state, 0);
+    FAIL_IF( rs_ssh_tx_get_flags(tx, STREAM_TOSERVER) == SSH_FLAG_VERSION_PARSED );
```
catenacyber: Should we use the `FAIL_IF` macros for the older tests ?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+    }
+}
+
+pub enum SSHConnectionState {
+    SshStateInProgress,
+    SshStateBannerDone,
+    SshStateFinished,
+}
+
+bitflags! {
```
catenacyber: Maybe it should rather be an enum and not a bitflags...

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+pub enum SSHEvent {
+    InvalidBanner = 0,
+    LongBanner,
+    InvalidRecord,
+}
+
+impl SSHEvent {
+    fn from_i32(value: i32) -> Option<SSHEvent> {
+        match value {
+            0 => Some(SSHEvent::InvalidBanner),
```
catenacyber: Should something be done in C with these events ?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+const SSH_RECORD_HEADER_LEN: usize = 6;
+//TODO complete enum and parse messages contents
+const SSH_MSG_NEWKEYS: u8 = 21;
+
+pub struct SshHeader {
+    record_left: u32,
+    record_buf: Vec<u8>,
+    flags: SSHTxFlag,
+    banner: Vec<u8>,
+    //can we have these be references to banner ?
```
catenacyber: As the comment states : can we have these be slices referencing the banner ? (or is it too dangerous ?)

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+        }
+    }
+
+    fn set_event(&mut self, event: SSHEvent) {
+        let ev = event as u8;
+        core::sc_app_layer_decoder_events_set_event_raw(&mut self.transaction.events, ev);
+    }
+
+    fn parse_record(&mut self, mut input: &[u8], resp: bool) -> bool {
+        //TODO add SCLogDebug ?
```
catenacyber: Should I add `SCLogDebug` messages ?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+        let mut hdr = &mut self.transaction.cli_hdr;
+        if resp {
+            hdr = &mut self.transaction.srv_hdr;
+        }
+        //first skip record left bytes
+        if hdr.record_left > 0 {
+            //should we check for overflow ?
+            let ilen = input.len() as u32;
```
catenacyber: Will `ilen` always fit into an uint32 ?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+            if hdr.record_buf.len() + input.len() < SSH_RECORD_HEADER_LEN {
+                hdr.record_buf.extend(input);
+                return true;
+            } else {
+                let needed = SSH_RECORD_HEADER_LEN - hdr.record_buf.len();
+                hdr.record_buf.extend(&input[..needed]);
+                input = &input[needed..];
+            }
+        }
+        // Should we make the code with less duplicates ?
```
catenacyber: Complexity comes here from the fact that we can either have one slice or (one vector and one slice) :
- Treating one slice is easy
- Treating one vector and one slice needs more code, but we can translate one slice into one vector and one slice...

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+            match parser::ssh_parse_record(input) {
+                Ok((rem, head)) => {
+                    input = rem;
+                    if head.msg_code == SSH_MSG_NEWKEYS {
+                        hdr.flags = SSHTxFlag::SSH_FLAG_PARSER_DONE;
+                    }
+                    //header and complete data (not returned)
+                }
+                Err(nom::Err::Incomplete(_)) => {
+                    match parser::ssh_parse_record_header(input) {
```
catenacyber: Is there a more Rust-idiomatic way to do this ? ie get an error message "We do not have a complete record, but we still have the complete record header"

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+
+    fn parse_banner(&mut self, input: &[u8], resp: bool) -> bool {
+        let mut hdr = &mut self.transaction.cli_hdr;
+        if resp {
+            hdr = &mut self.transaction.srv_hdr;
+        }
+        let mut iline = input;
+        let mut vec = vec![13u8];
+        if hdr.banner.len() > 0 {
+            if hdr.banner[hdr.banner.len() - 1] == 13 {
```
catenacyber: This is hacky...
We still need somehow to track that the last packet ended with CR.
Any better idea on how to do this ?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+        }
+        match parser::ssh_parse_line(iline) {
+            Ok((rem, line)) => {
+                let mut setlong = false;
+                if hdr.banner.len() + line.len() <= SSH_MAX_BANNER_LEN {
+                    hdr.banner.extend(line);
+                } else if hdr.banner.len() < SSH_MAX_BANNER_LEN {
+                    hdr.banner
+                        .extend(&line[0..SSH_MAX_BANNER_LEN - hdr.banner.len()]);
+                    setlong = true;
```
catenacyber: This `setlong` was required by the borrow checker

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+        };
+        0
+    } else {
+        -1
+    }
+}
+
+/// C entry point for a probing parser.
+#[no_mangle]
+pub extern "C" fn rs_dummy_probing_parser(
```
catenacyber: Should we rather make this probing parser optional ?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,629 @@
...
+}
+
+#[no_mangle]
+pub extern "C" fn rs_ssh_state_progress_completion_status(_direction: u8) -> std::os::raw::c_int {
+    return SSHConnectionState::SshStateFinished as i32;
+}
+
+#[no_mangle]
+pub extern "C" fn rs_ssh_tx_get_flags(
+    tx: *mut std::os::raw::c_void,
```
catenacyber: Is this the right file to be for this function ?

rust/src/ssh/ssh.rs
victorjulien: It can be done. It will be tricky due to Rusts policing of holding multiple references

rust/src/ssh/ssh.rs
victorjulien: It should, the C side of things should never send us that much data. But I'm not sure if there are explicit checks for this anywhere.

rust/src/ssh/ssh.rs
victorjulien: Would suggest `set_event_long` or something to indicate we're dealing with an exception

rust/src/ssh/ssh.rs
victorjulien: Yeah I think we should do this.

rust/src/ssh/ssh.rs
victorjulien: Yeah

src/app-layer-ssh.h
victorjulien: Not sure either. Do you still need it?

rust/src/ssh/detect.rs
victorjulien: You can at least wrap both `*buffer` and `*buffer_len` assignments in a single `unsafe`. We need `unsafe` when we pass something to C though.

rust/src/ssh/parser.rs
victorjulien: I suppose this is something to test against real implementations.

rust/src/ssh/parser.rs
victorjulien: Not sure I understand what you're asking

rust/src/ssh/ssh.rs
victorjulien: Would suggest looking at how for example the `smb` parser is dealing with events.

victorjulien: The events are exposed to the main code for the app-layer-event rules and anomaly logging.

rust/src/ssh/ssh.rs
victorjulien: This is more like a state, right?

rust/src/ssh/ssh.rs
victorjulien: Where it makes sense :)

rust/src/ssh/ssh.rs
catenacyber: I tried and it seemed tricky indeed.
So, should it be done ? Is it worth the performance cost ?

rust/src/ssh/detect.rs
catenacyber: My understanding is that we need `unsafe` here because we use a pointer _from_ C
So, can we do anything about it, or just wrap the two statements together ?

rust/src/ssh/parser.rs
catenacyber: I have been looking at openSSH only, which accepts LF and CRLF as end of line/banner (but not CR only) cf https://github.com/openssh/openssh-portable/blob/master/ssh_api.c#L345

The current unit tests make sure CR only must be considered as end of line/banner

Which other real implementations should I look at ?
Should this be part of this PR ? Or should I create a new ticket for it ?

rust/src/ssh/parser.rs
catenacyber: If I send a SSH record header with length == 0 or length == 1, the C and Rust parser will fail.
I am not sure how real implementations behave...
Should this be part of this PR ? Or should I create a new ticket for it (and keep this PR as only translating and not fixing bugs) ?

src/app-layer-ssh.h
catenacyber: `SSH_STATE_BANNER_DONE` is being used in detect-ssh-*.c, output-json-ssh.c, and output-lua.c

Maybe it is being used abusively as a shortcut to `rs_ssh_state_progress_completion_status()`

rust/src/ssh/ssh.rs
catenacyber: Ok, will do

rust/src/ssh/ssh.rs
catenacyber: Ok, will do

rust/src/ssh/ssh.rs
catenacyber: Ok.
Should this be part of this PR ?
Or should I create a ticket for it ?
It will affect code of other parsers...

catenacyber:
@victorjulien I will be waiting for more answers before V2 ;-)

rust/src/ssh/detect.rs
victorjulien: The from C part is hidden by the `cast_pointer` macro. The `unsafe` here is for returning pointer and length value to the caller. This also crosses the FFI boundary, and is thus unsafe.

rust/src/ssh/parser.rs
victorjulien: Ticket sounds fine.

rust/src/ssh/parser.rs
victorjulien: Does the app-layer API ever send this up though?

src/app-layer-ssh.h
victorjulien: Ah I see. So in these places it is used to make sure logging/detection happens as soon as it can, which means before the tx is considered complete, but after we have the banner parsed. So I agree we need a way to express this in the Rust parser case as well.

src/app-layer-ssh.c
victorjulien: Yes please.

rust/src/ssh/ssh.rs
victorjulien: Ticket + PR, then rebuild this ssh on top of that?

rust/src/ssh/detect.rs
jasonish: In this case its probably best to just mark the whole function as unsafe.  Clippy  would prefer this as well:
```
12465 error: this public function dereferences a raw pointer but is not marked `unsafe`
12466   --> src/ssh/detect.rs:34:27
12467    |
12468 34 |                 unsafe { *buffer = m.as_ptr() };
12469    |                           ^^^^^^
12470    |
12471    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref
```

catenacyber: My compiler produces a warning when the function is marked unsafe :
```
warning: unnecessary `unsafe` block
   --> src/parser.rs:127:31
    |
127 |       ($ptr:ident, $ty:ty) => ( unsafe{ &mut *($ptr as *mut $ty) } );
    |                                 ^^^^^^ unnecessary `unsafe` block
    | 
   ::: src/ssh/detect.rs:23:1
    |
23  | / pub unsafe extern "C" fn rs_ssh_tx_get_protocol(
24  | |     tx: *mut std::os::raw::c_void,
25  | |     buffer: *mut *const u8,
26  | |     buffer_len: *mut u32,
...   |
29  | |     let tx = cast_pointer!(tx, SSHTransaction);
    | |              --------------------------------- in this macro invocation
...   |
52  | |     return 0;
53  | | }
    | |_- because it's nested under this `unsafe` fn
    |
    = note: #[warn(unused_unsafe)] on by default
```

rust/src/ssh/parser.rs
catenacyber: https://redmine.openinfosecfoundation.org/issues/3485

rust/src/ssh/parser.rs
catenacyber: This is used by reaching the line `header->flags |= SSH_FLAG_PARSER_DONE;` which ends up doing `AppLayerParserStateSetFlag(pstate, APP_LAYER_PARSER_NO_INSPECTION);`

src/app-layer-ssh.h
catenacyber: Ok, and what is the Rust idiomatic way ?

rust/src/ssh/ssh.rs
catenacyber: https://redmine.openinfosecfoundation.org/issues/3486
PR coming...

rust/src/ssh/parser.rs
catenacyber: Furthermore, we will need to get this right for https://redmine.openinfosecfoundation.org/issues/2698

rust/src/ssh/detect.rs
jasonish: Oh, yeah. I think what is needed is a sweeping change to cover this. For now, stick with  your code as-is.

rust/src/ssh/parser.rs
victorjulien: > This is used by reaching the line `header->flags |= SSH_FLAG_PARSER_DONE;` which ends up doing `AppLayerParserStateSetFlag(pstate, APP_LAYER_PARSER_NO_INSPECTION);`

So we're still calling the parser after this, but only with a 0 len payload?

src/app-layer-ssh.h
victorjulien: @jasonish do you know?

rust/src/ssh/ssh.rs
victorjulien: Perf impact is probably low, although in general we don't know for sure until we test/measure. But I'm fine with leaving this as is for now.

rust/src/ssh/parser.rs
catenacyber: Oh, I think I understood four misunderstanding...

By 0-length or 1-length packets
I meant 0-length or 1-length _SSH records_

This ssh record begins by field `pkt_len` encoded on u32, then has a u8 padding and u8 msg_code
The `pkt_len` includes these u8.
Both C and Rust parsers forbid the use of `pkt_len < 2`
I do not know if real implementations allow and skip these dummy SSH records... That is my question.

victorjulien: How hard would it to find out?

catenacyber: A few hours I guess... Enough to ask if someone already did this ;-)

src/app-layer-ssh.h
catenacyber: Working with cbindgen.toml, I did not manage to make cbindgen produce an enum, but I can export constants.
I did not manage to make it work with the example of https://blog.eqrion.net/announcing-cbindgen/

rust/src/ssh/ssh.rs
catenacyber: Indeed

rust/src/ssh/ssh.rs
catenacyber: We will need to update https://redmine.openinfosecfoundation.org/projects/suricata/wiki/AppLayer

catenacyber: Ok, so I added a file ssh-events.rules

src/app-layer-ssh.h
jasonish: > Working with cbindgen.toml, I did not manage to make cbindgen produce an enum, but I can export constants.
> I did not manage to make it work with the example of https://blog.eqrion.net/announcing-cbindgen/

Apparently if the enum is not used in the signature of an export function, the enum will not be exported either. To get around this, it looks like we can make this change to cbindgen.toml:
```
diff --git a/rust/cbindgen.toml b/rust/cbindgen.toml
index f6117c403..c69469443 100644
--- a/rust/cbindgen.toml
+++ b/rust/cbindgen.toml
@@ -71,7 +71,10 @@ documentation_style = "doxy"
 # found but otherwise don't appear to be used by the public API.
 #
 # default: []
-include = ["AppLayerGetTxIterTuple"]
+include = [
+    "AppLayerGetTxIterTuple",
+    "SSHEvent",
+]
 
 # A list of items to not include in the generated bindings
 # default: []
@@ -109,7 +112,7 @@ exclude = [
 # * "functions":
 #
 # default: []
-item_types = ["structs","opaque","functions"]
+item_types = ["structs","opaque","functions","enums"]
 
 # Whether applying rules in export.rename prevents export.prefix from applying.
 #
```

PR #4548, opened 2020-02-13

catenacyber:
For discussion ,not to be merged yet

Describe changes:

Follows #4491 with addressed comments and TODOs

This is basic and can be improved (cf TODOs comments).

PS : Another solution I think of would be to give some external program the ability to inject some kinds of pseudo packets (with the reassembled file)
This way, we would be robust if the range requests come over multiple flows, and we still can run suricata rules on the reassembled file

### DIFF
--- a/src/app-layer-htp-file.c
+++ b/src/app-layer-htp-file.c
@@ -113,6 +113,13 @@ int HTPFileOpen(HtpState *s, const uint8_t *filename, uint16_t filename_len,
         }
 
         sbcfg = &s->cfg->response.sbcfg;
+        if (s->file_range_ids != 0) {
+            //TODO set an event ? or event better : handle intersected downloads
+            SCLogWarning(SC_ERR_NOT_SUPPORTED,
+                         "Regular download during HTTP Range file reassembly.");
+            FileCloseFileById(files, s->file_range_ids, NULL, 0, FILE_TRUNCATED);
+            s->file_range_ids = 0;
+        }
 
     } else {
         if (s->files_ts == NULL) {
@@ -225,9 +232,12 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
  *  \retval -2 error parsing
  *  \retval -3 error negative end in range
  */
-int HTPFileSetRange(HtpState *s, bstr *rawvalue)
+int HTPFileOpenWithRange(HtpState *s,const uint8_t *filename, uint16_t filename_len,
+                         const uint8_t *data, uint32_t data_len,
+                         uint64_t txid, bstr *rawvalue)
 {
     SCEnter();
+    uint16_t flags;
 
     if (s == NULL) {
         SCReturnInt(-1);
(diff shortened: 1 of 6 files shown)


src/app-layer-htp-file.c
```diff
@@ -225,30 +232,88 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+        s->file_range_ids = FLAG_FILE_RANGES | s->file_track_id;
+
+        if (FileOpenFileWithId(files, &s->cfg->response.sbcfg, s->file_range_ids,
+                               filename, filename_len,
+                               data, data_len, flags) != 0) {
+            SCLogDebug("open file for range failed");
+            SCReturnInt(-1);
+        }
+    } else if (s->file_range_ids != 0){
+        // TODO check this is next range and same file
```
catenacyber: oops, not a TODO anymore

src/app-layer-htp-file.c
```diff
@@ -225,30 +232,88 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+        if (FileCheckNameAndOffsetById(files, s->file_range_ids,
+                                       filename, filename_len, crparsed.start) == 0) {
+            if (FileAppendDataById(files, s->file_range_ids, data, data_len) != 0) {
+                SCReturnInt(-1);
+            }
+        } else {
+            //TODO set an event ? or even better : handle unordered ranges download
```
catenacyber: This is the big question (see rust filetracker.rs so called ooo chunks).
I think that we should get a first version, and then deal with this.
However, the first version should trigger some event so that we can get alerts on this behavior.
What do you think @jlucovsky ?

src/app-layer-htp.h
```diff
@@ -257,10 +257,14 @@ typedef struct HtpState_ {
     uint16_t events;
     uint16_t htp_messages_offset; /**< offset into conn->messages list */
     uint32_t file_track_id;             /**< used to assign file track ids to files */
+    uint32_t file_range_ids;             /**< used to assign track ids to range file */
     uint64_t last_request_data_stamp;
     uint64_t last_response_data_stamp;
 } HtpState;
 
+/** Flag to track reassembled file from multiple ranges */
+#define FLAG_FILE_RANGES 0x80000000
```
catenacyber: We trust there will be no overflow...

src/util-file.c
```diff
@@ -795,9 +795,46 @@ int FileSetRange(FileContainer *ffc, uint64_t start, uint64_t end)
...
+        SCReturnInt(-1);
+    }
+
+    File *ff = ffc->head;
+    for ( ; ff != NULL; ff = ff->next) {
+        if (track_id == ff->file_track_id) {
+            if (ff->size != start) {
+                SCReturnInt(1);
+            }
+            if (ff->name_len != name_len) {
```
catenacyber: Is there a function to compare these two buffers ? (and save three lines of code)

catenacyber:
CI fails because of suricata-verify test http-evader-501 which uses `Range` but downloads the whole file in one try...
I think that we should detect this case and go back to normal file handling...

src/app-layer-htp-file.c
```diff
@@ -113,6 +113,13 @@ int HTPFileOpen(HtpState *s, const uint8_t *filename, uint16_t filename_len,
         }
 
         sbcfg = &s->cfg->response.sbcfg;
+        if (s->file_range_ids != 0) {
+            //TODO set an event ? or event better : handle intersected downloads
+            SCLogWarning(SC_ERR_NOT_SUPPORTED,
```
victorjulien: Prints to screen based on traffic are not allowed. Guess an event to start?

src/app-layer-htp-file.c
```diff
@@ -225,30 +232,88 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+            SCReturnInt(-1);
+        }
+        files = s->files_tc;
     }
 
     HtpContentRange crparsed;
     if (HTPParseContentRange(rawvalue, &crparsed) != 0) {
-        SCLogDebug("parsing range failed");
-        SCReturnInt(-2);
+        SCLogDebug("parsing range failed, going back to normal file");
```
victorjulien: is this an error/anomaly condition? If so, should we set and event?

src/app-layer-htp-file.c
```diff
@@ -225,30 +232,88 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+        return HTPFileOpen(s, filename, (uint32_t)filename_len,
+                           data, data_len,
+                           txid, STREAM_TOCLIENT);
     }
     if (crparsed.end <= 0) {
         SCLogDebug("negative end in range");
```
victorjulien: event?

src/app-layer-htp-file.c
victorjulien: In normal traffic, it should be in order, right?

src/util-file.c
victorjulien: Don't think so. You can add a util func for this.

src/util-file.c
```diff
@@ -795,9 +795,46 @@ int FileSetRange(FileContainer *ffc, uint64_t start, uint64_t end)
...
+ */
+int FileCheckNameAndOffsetById(FileContainer *ffc, uint32_t track_id,
+        const uint8_t *name, uint16_t name_len, uint64_t start) {
+    SCEnter();
+
+    if (ffc == NULL || ffc->tail == NULL) {
+        SCReturnInt(-1);
+    }
+
+    File *ff = ffc->head;
```
victorjulien: move into `for` statement?

src/app-layer-htp.h
victorjulien: I don't see how this is used. I mean I see it is being set, but thats all.

src/app-layer-htp-file.c
catenacyber: By "normal traffic", do you mean from one of the top 5 browsers ?
I do not know how to produce such a traffic...
Can it be observed somewhere ?

Either way, I am not sure it will be in order...

src/app-layer-htp.h
catenacyber: The point is to have an id different than the other files.
But maybe we can have this different id without using this flag, just incrementing the id for the next file...
What do you prefer ?

src/util-file.c
catenacyber: Should I ?

src/app-layer-htp-file.c
catenacyber: Can you point me to some similar code ? (generating an event for files)

src/util-file.c
catenacyber: ok

src/app-layer-htp-file.c
catenacyber: Not sure, I will check

src/app-layer-htp-file.c
catenacyber: Will check also

src/app-layer-htp-file.c
catenacyber: Ok, found some

src/app-layer-htp-file.c
catenacyber: It is, I am setting an event

src/app-layer-htp-file.c
catenacyber: Event also

src/app-layer-htp.h
victorjulien: If we can do without 'tricks', I prefer that.

src/util-file.c
victorjulien: Sure. Maybe just a `SCMemcmpWithLength` that takes both lengths and otherwise wraps around `SCMemcmp`. Not liking my name suggestion btw.

catenacyber: ok, naming `SCBufferCmp`

PR #4558, opened 2020-02-17

inashivb:
actions.

e.g.
to write two rules,
```
```

```
```


inashivb:
Thanks, Victor! Replaced with #4563

PR #4559, opened 2020-02-18

catenacyber:

Describe changes:

Follows #4548 with addressed comment

### DIFF
--- a/src/app-layer-htp-file.c
+++ b/src/app-layer-htp-file.c
@@ -113,6 +113,12 @@ int HTPFileOpen(HtpState *s, const uint8_t *filename, uint16_t filename_len,
         }
 
         sbcfg = &s->cfg->response.sbcfg;
+        if (s->file_range_ids != 0) {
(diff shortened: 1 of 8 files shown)


src/app-layer-htp-file.c
```diff
@@ -225,30 +231,95 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+        // open another file for the whole ranges
+        if (s->file_range_ids != 0) {
+            //TODO even better : handle intersected downloads
+            AppLayerDecoderEventsSetEventRaw(&htud->decoder_events, HTTP_DECODER_EVENT_RANGE_INTERSECTED);
+            s->events++;
+            if (FileCloseFileById(files, s->file_range_ids, NULL, 0, FILE_TRUNCATED) != 0) {
+                SCLogDebug("close file for range failed");
+            }
+        }
+        s->file_track_id++;
```
catenacyber: No side effects

PR #4562, opened 2020-02-18

catenacyber:

Describe changes:

Follows #4559 with :
- right check for whole file in one range request (end-1==size)
- do not set event for unsupported Content-Range values such as `Content-Range: bytes 200-1000/*`
cf https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range

### DIFF
--- a/src/app-layer-htp-file.c
+++ b/src/app-layer-htp-file.c
@@ -113,6 +113,12 @@ int HTPFileOpen(HtpState *s, const uint8_t *filename, uint16_t filename_len,
         }
 
         sbcfg = &s->cfg->response.sbcfg;
+        if (s->file_range_ids != 0) {
+            //TODO even better : handle intersected downloads
+            HTPSetEvent(s, NULL, direction, HTTP_DECODER_EVENT_RANGE_INTERSECTED);
+            FileCloseFileById(files, s->file_range_ids, NULL, 0, FILE_TRUNCATED);
+            s->file_range_ids = 0;
+        }
 
     } else {
         if (s->files_ts == NULL) {
@@ -225,9 +231,12 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
  *  \retval -2 error parsing
  *  \retval -3 error negative end in range
  */
-int HTPFileSetRange(HtpState *s, bstr *rawvalue)
+int HTPFileOpenWithRange(HtpState *s,const uint8_t *filename, uint16_t filename_len,
+                         const uint8_t *data, uint32_t data_len,
+                         uint64_t txid, bstr *rawvalue, HtpTxUserData *htud)
 {
     SCEnter();
+    uint16_t flags;
 
     if (s == NULL) {
         SCReturnInt(-1);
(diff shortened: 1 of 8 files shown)


victorjulien:
"Adds util function SCBufferCmp" this should be in its own commit.

src/app-layer-htp-file.c
```diff
@@ -113,6 +113,12 @@ int HTPFileOpen(HtpState *s, const uint8_t *filename, uint16_t filename_len,
         }
 
         sbcfg = &s->cfg->response.sbcfg;
+        if (s->file_range_ids != 0) {
+            //TODO even better : handle intersected downloads
```
victorjulien: what would it take to support this? Files use a unique identifier I think, was it the etag?

src/app-layer-htp-file.c
```diff
@@ -225,30 +231,92 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+        SCLogDebug("parsing range failed, going back to normal file");
+        return HTPFileOpen(s, filename, (uint32_t)filename_len,
+                           data, data_len,
+                           txid, STREAM_TOCLIENT);
     }
-    if (crparsed.end <= 0) {
-        SCLogDebug("negative end in range");
-        SCReturnInt(-3);
+    if (crparsed.end <= 0 || crparsed.size <= 0) {
+        SCLogDebug("range without all information");
```
victorjulien: is this 'event worthy' as well?

src/app-layer-htp-file.c
```diff
@@ -225,30 +231,92 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+            AppLayerDecoderEventsSetEventRaw(&htud->decoder_events, HTTP_DECODER_EVENT_RANGE_INTERSECTED);
+            s->events++;
+            if (FileCloseFileById(files, s->file_range_ids, NULL, 0, FILE_TRUNCATED) != 0) {
+                SCLogDebug("close file for range failed");
+            }
+        }
+        s->file_track_id++;
+        s->file_range_ids = s->file_track_id;
+
+        if (FileOpenFileWithId(files, &s->cfg->response.sbcfg, s->file_range_ids,
```
victorjulien: if this succeeds we continue on line 312 which calls `FileOpenFileWithId` again. Is that intentional?

src/app-layer-htp-file.c
```diff
@@ -292,6 +360,12 @@ int HTPFileStoreChunk(HtpState *s, const uint8_t *data, uint32_t data_len,
         goto end;
     }
 
+    if (s->file_range_ids != 0) {
+        result = FileAppendDataById(files, s->file_range_ids, data, data_len);
```
victorjulien: I think it would be useful if the commit message explained the overall logic of the modifications, as I'm a bit confused with why some things are apparently being done twice. Like 2 calls to `FileAppendDataById` in a row here.

src/app-layer-htp-file.c
catenacyber: Good question.
This is an unhandled case, not evil traffic...
So, is it event worthy ?

src/app-layer-htp-file.c
catenacyber: There are two types of intersected downloads :
- range download intersected with regular download
- range downloads intersected with themselves

For the first case, it only takes a boolean to know if we must call `FileAppendDataById` for the reassembled file
https://github.com/OISF/suricata/pull/4562/files#diff-d5d8efe82303f94955b7b0c979f6585aR364

For the second case, we will a dynamic array for `file_range_ids`

src/app-layer-htp-file.c
catenacyber: Indeed it is intentional : that is the core of this Pull Request.
That is what I tried to explain in my comment :

> if there is a range : open 2 files, one for the whole reassembled, and one only for the current range

src/app-layer-htp-file.c
catenacyber: I am adding the comments of the PR

catenacyber:
I pushed v6 on my repo, waiting for more answers before making a PR

src/app-layer-htp-file.c
victorjulien: So you are saying that it is valid to have ranges with zero size and/or 0 as end 'size'? Seems odd.

src/app-layer-htp-file.c
victorjulien: OK this wasn't clear to me.

src/app-layer-htp-file.c
victorjulien: Also explain the overall design in the commit message too please.

src/app-layer-htp-file.c
catenacyber: It is indeed possible to have range with an unknown size
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
But not with an unknown end...

Fixing this

src/app-layer-htp-file.c
catenacyber: And now, how does it look ?

PR #4563, opened 2020-02-18

inashivb:
actions.

e.g.
to write two rules,
```
```

```
```

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -46,12 +46,16 @@
 #include "util-debug.h"
 
 #define PARSE_REGEX         "^([a-z]+)(?:,\\s*(.*))?"
+
+#define MAX_TOKENS 100
+
 static pcre *parse_regex;
 static pcre_extra *parse_regex_study;
 
 int DetectFlowbitMatch (DetectEngineThreadCtx *, Packet *,
         const Signature *, const SigMatchCtx *);
 static int DetectFlowbitSetup (DetectEngineCtx *, Signature *, const char *);
+static char **StrSplit(char *arrptr, const char *delim);
 void DetectFlowbitFree (void *);
 void FlowBitsRegisterTests(void);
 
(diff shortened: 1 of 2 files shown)


src/detect-flowbits.c
```diff
@@ -70,6 +74,39 @@ void DetectFlowbitsRegister (void)
...
+    char *token;
+    int i = 0;
+    if (strchr(arrptr, ' ') != NULL) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Space not allowed in flowbits.");
+        goto error;
+    }
+    strarr = SCCalloc((MAX_TOKENS + 1), sizeof(char **));
+    if (strarr == NULL)
+        goto error;
+    while((token = strtok_r(arrptr, delim, &arrptr))) {
```
victorjulien: I think with `strtok_r` you're not supposed to use the same pointer in the 1st and last argument.

victorjulien: style: space between `while` and `((`

src/detect-flowbits.c
```diff
@@ -105,6 +142,15 @@ static int DetectFlowbitMatchIsset (Packet *p, const DetectFlowbitsData *fd)
 {
     if (p->flow == NULL)
         return 0;
+    if (fd->or_list_size > 0) {
+        int result = 0;
+        for (int i = 0; i < fd->or_list_size; i++) {
+            result |= FlowBitIsset(p->flow, fd->or_list[i]);
```
victorjulien: I suppose the anding of result isn't useful anymore now. I think you could simply do:
```c
for (int i = 0; i < fd->or_list_size; i++) {
    if (FlowBitIsset(p->flow, fd->or_list[i]))
        return 1;
}
return 0;
```

src/detect-flowbits.c
```diff
@@ -1078,6 +1201,208 @@ static int FlowBitsTestSig08(void)
...
+    FAIL_IF_NULL(s);
+    s = s->next  = SigInit(de_ctx,"alert ip any any -> any any (msg:\"Flowbit isset ored flowbits\"; flowbits:isset,fb3|fb4; sid:4;)");
+    FAIL_IF_NULL(s);
+
+    SigGroupBuild(de_ctx);
+    DetectEngineThreadCtxInit(&th_v, (void *)de_ctx, (void *)&det_ctx);
+
+    SigMatchSignatures(&th_v, de_ctx, det_ctx, p);
+
+    FAIL_IF(PacketAlertCheck(p, 4));
```
victorjulien: suggest adding checks for the other sids too

src/detect-flowbits.c
```diff
@@ -105,6 +142,15 @@ static int DetectFlowbitMatchIsset (Packet *p, const DetectFlowbitsData *fd)
 {
     if (p->flow == NULL)
         return 0;
+    if (fd->or_list_size > 0) {
+        int result = 0;
+        for (int i = 0; i < fd->or_list_size; i++) {
```
victorjulien: `i` is signed, `fd->or_list_size` unsigned. I suspect we may see warnings for that depending on compiler flags.

inashivb:
Thanks, Victor! Continued in #4564

PR #4564, opened 2020-02-18

inashivb:
actions.

e.g.
to write two rules,
```
```

```
```

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -46,12 +46,16 @@
 #include "util-debug.h"
 
 #define PARSE_REGEX         "^([a-z]+)(?:,\\s*(.*))?"
+
+#define MAX_TOKENS 100
+
 static pcre *parse_regex;
 static pcre_extra *parse_regex_study;
 
 int DetectFlowbitMatch (DetectEngineThreadCtx *, Packet *,
         const Signature *, const SigMatchCtx *);
 static int DetectFlowbitSetup (DetectEngineCtx *, Signature *, const char *);
+static char **StrSplit(char *arrptr, const char *delim);
 void DetectFlowbitFree (void *);
 void FlowBitsRegisterTests(void);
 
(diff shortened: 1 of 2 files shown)


src/detect-flowbits.c
```diff
@@ -105,6 +143,13 @@ static int DetectFlowbitMatchIsset (Packet *p, const DetectFlowbitsData *fd)
 {
     if (p->flow == NULL)
         return 0;
+    if (fd->or_list_size > 0) {
+        for (uint8_t i = 0; i < fd->or_list_size; i++) {
```
victorjulien: I think this will work fine if we have 2 or 3 flow bits, but it may be slow if we have many flowbits. Pretty much we walk an array here and per member we walk a list in the flow. In the common case both the list and array will be short, but what if both contain 100 items?

Since the most expensive part is most likely the list walk in the flow, I think it will be more efficient to do something like:
```c
if (fd->or_list_size > 0) {
   return FlowBitIssetFromArray(p->flow, fd->or_list, fd->or_list_size);
}
```
Then `FlowBitIssetFromArray` would walk `f->flowvar` and for each flowbit in that list check the array. I suppose a small and opportunistic optimization could be to pass the minimum and maximum of the id's in the array as well, so you can first check if the flow bit id would possibly be part of the id space. Another option could be to use something like a bloomfilter to do a first quick check.

src/detect-flowbits.c
```diff
@@ -70,6 +74,40 @@ void DetectFlowbitsRegister (void)
...
+    uint8_t i = 0;
+    if (strchr(arrptr, ' ') != NULL) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Space not allowed in flowbits.");
+        goto error;
+    }
+    strarr = SCCalloc((MAX_TOKENS + 1), sizeof(char **));
+    if (strarr == NULL)
+        goto error;
+    while ((token = strtok_r(arrptr, delim, &saveptr))) {
+        if (i >= MAX_TOKENS) {
```
jlucovsky: `i` will never exceed `MAX_TOKENS`.

Can the check be moved to just before (or after) the `i++`?

This changes things slightly -- token collection ceases when the maximum number of tokens are collected instead of generating an error when the maximum is reached and there are more tokens to collect.

src/detect-flowbits.c
```diff
@@ -245,14 +298,37 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
...
-
-    SCLogDebug("idx %" PRIu32 ", cmd %s, name %s",
-        cd->idx, fb_cmd_str, strlen(fb_name) ? fb_name : "(none)");
-
+    if (strchr(fb_name, '|') != NULL) {
+        fb_items = StrSplit(fb_name, "|");
+        if (fb_items == NULL) {
+            goto error;
+        }
+        for (i = 0; fb_items[i] != NULL ; i++) {
```
jlucovsky: Suggest surrounding this code block with `if (SCLogDebugEnabled())` to avoid traversing the array when not in debug mode (most common case).

src/detect-flowbits.c
```diff
@@ -245,14 +298,37 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
...
+        cd->or_list = SCMalloc(cd->or_list_size * sizeof(uint32_t));
+        if (unlikely(cd->or_list == NULL))
+            goto error;
+        for (uint8_t j = 0; j < cd->or_list_size ; j++) {
+            cd->or_list[j] = VarNameStoreSetupAdd(fb_items[j], VAR_TYPE_FLOW_BIT);
+            de_ctx->max_fb_id = MAX(cd->or_list[j], de_ctx->max_fb_id);
+        }
+        cd->cmd = fb_cmd;
+        // Free the allocated memory for flowbit items
+        for (int j = 0; j < cd->or_list_size ; j++) {
```
jlucovsky: Suggest adding a utility routine that frees the elements instead of doing it here (and a few other places).

```
FuncName(int howmany, char **fbitems)
```

src/detect-flowbits.c
```diff
@@ -287,6 +363,13 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
     return 0;
 
 error:
+
```
jlucovsky: Pre-existing error -- `cd->or_list` will be leaked it `SigMatchAlloc()` fails.

Suggest calling `DetectFlowbitFree(cd)` instead of `SCFree(cd)`.

src/detect-flowbits.c
inashivb: I think I am misunderstanding something here. I changed MAX_`TOKENS` to 2 and passed three flowbits in a test and it failed with the error message here.
Could you please elaborate a little more which situation would `i` not exceed `MAX_TOKENS`?

src/detect-flowbits.c
inashivb: I was also trying to use this to get the length of `fb_items` array, is there a better way to do that? Please let me know.

inashivb:
Replaced with #4737

PR #4358, opened 2019-11-10
Feature 2694 thresholds by_rule by_both v3

mordak:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/2694) ticket: 2694

Describe changes:
- Accept by_rule and by_both in rule thresholds.
- Refactor threshold application to separate threshold table storage from threshold calculation.
- Ensure the by_rule threshold table is initialized when needed.
- Add tests for threshold parsing and application.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #4569, opened 2020-02-19
Feature 2694 thresholds v4

mordak:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues/2694) ticket: https://redmine.openinfosecfoundation.org/issues/2694

Describe changes:
- Accept by_rule and by_both in rule thresholds.
- Refactor threshold application to separate threshold table storage from threshold calculation.
- Ensure the by_rule threshold table is initialized when needed.
- Add tests for threshold parsing and application.

### DIFF
--- a/src/detect-engine-threshold.c
+++ b/src/detect-engine-threshold.c
@@ -673,6 +673,32 @@ void ThresholdHashInit(DetectEngineCtx *de_ctx)
 }
 
 /**
+ * \brief Realloc threshold context hash tables
+ *
+ * \param de_ctx Detection Context
+ */
+void ThresholdHashRealloc(DetectEngineCtx *de_ctx)
+{
+    /* Return if we are already big enough */
+    uint32_t num = de_ctx->signum + 1;
+    if (num <= de_ctx->ths_ctx.th_size)
+        return;
+
+    void *ptmp = SCRealloc(de_ctx->ths_ctx.th_entry, num * sizeof(DetectThresholdEntry *));
+    if (ptmp == NULL) {
+        SCLogError(SC_ERR_MEM_ALLOC, "Error allocating memory for rule thresholds"
+                " (tried to allocate %"PRIu32" th_entrys for rule tracking)", num);
+        exit(EXIT_FAILURE);
+    } else {
+        de_ctx->ths_ctx.th_entry = ptmp;
+        for (uint32_t i = de_ctx->ths_ctx.th_size; i < num; ++i) {
+            de_ctx->ths_ctx.th_entry[i] = NULL;
+        }
+        de_ctx->ths_ctx.th_size = num;
+    }
+}
+
+/**
  * \brief Destroy threshold context hash tables
  *
  * \param de_ctx Dectection Context
--- a/src/detect-engine-threshold.c
+++ b/src/detect-engine-threshold.c
@@ -69,6 +69,11 @@
 #include "util-var-name.h"
 #include "tm-threads.h"
 
+#define TIME_DIFF_SEC(p_ts, lookup_tsh) \
+    (uint64_t)((((uint64_t)(p_ts).tv_sec * 1000000 + (p_ts).tv_usec) - \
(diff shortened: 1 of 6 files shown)


victorjulien:
Continuation of #4358

src/detect-engine-threshold.c
```diff
@@ -713,6 +672,32 @@ void ThresholdHashInit(DetectEngineCtx *de_ctx)
...
+    /* Return if we are already big enough */
+    uint32_t num = de_ctx->signum + 1;
+    if (num <= de_ctx->ths_ctx.th_size)
+        return;
+
+    void *ptmp = SCRealloc(de_ctx->ths_ctx.th_entry, num * sizeof(DetectThresholdEntry *));
+    if (ptmp == NULL) {
+        SCLogError(SC_ERR_MEM_ALLOC, "Error allocating memory for rule thresholds"
+                " (tried to allocate %"PRIu32" th_entrys for rule tracking)", num);
+        exit(EXIT_FAILURE);
```
jlucovsky: This is a change in behavior -- previously, an error message was logged with incomplete cleanup.

@victorjulien Should this be a warning instead?  When `realloc` fails, the original pointer is left unmodified.

src/detect-threshold.c
```diff
@@ -372,6 +379,43 @@ static int ThresholdTestParse05(void)
     return 0;
 }
 
+/**
+ * \test ThresholdTestParse06 is a test for thresholding by_both
+ *
+ *  \retval 1 on succces
```
jlucovsky: nit: s/cuces/success/ (here, and a few more places).

src/detect-engine-threshold.c
```diff
@@ -69,6 +69,11 @@
 #include "util-var-name.h"
 #include "tm-threads.h"
 
+#define TIME_DIFF_SEC(p_ts, lookup_tsh) \
```
jlucovsky: With renamed params, this seems like it'd be a good add to `util-time.h`

src/detect-engine-threshold.c
```diff
@@ -458,17 +435,9 @@ static int ThresholdHandlePacketHost(Host *h, Packet *p, const DetectThresholdDa
                 }
             } else {
                 DetectThresholdEntry *e = DetectThresholdEntryAlloc(td, p, sid, gid);
-                if (e == NULL) {
-                    break;
-                }
-
-                e->tv_sec1 = p->ts.tv_sec;
-                e->current_count = 1;
+                *new_tsh = e;
```
jlucovsky: The local `e` can be removed ... just assign to `*new_tsh` instead (here, and a few other places).

src/detect-engine-threshold.c
mordak: Oh, true. I must have been looking at the initialization routines when I added this. I can remove it, and have the function warn instead without changing the table.

src/detect-threshold.c
mordak: Will fix.

src/detect-engine-threshold.c
mordak: Would you like me to move it as part of this PR?

src/detect-engine-threshold.c
mordak: Okay.

src/detect-engine-threshold.c
jlucovsky: I think it'd be a good addition so yes.

src/detect-engine-threshold.c
victorjulien: Yeah lets do the warn thing, even if that also isn't ideal. It's at least not changing the behaviour.

mordak: Done in #4747. The previous behaviour was to free the existing entry on allocation failure, so that's what we do now.

src/detect-engine-threshold.c
mordak: Done in #4747. This has a knock on effect where it made sense to change the threshold structure to hold a timeval (tv1) instead of holding tv_sec1 and tv_usec1 separately. I think the consistency is better.

PR #4556, opened 2020-02-16
smtp/mime: Restrict file name lengths

jlucovsky:
Continuation of #4552

This commit places restrictions on the length of the file name specified
in attachments (`name=` or `filename=`) to `NAME_MAX`. Names exceeding
these limits will be truncated and processing will continue with the
truncated name.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3409](https://redmine.openinfosecfoundation.org/issues/3265)

Describe changes:
- Address review comments

PR #4571, opened 2020-02-19

jlucovsky:
Continuation of #4556

truncated name.

Describe changes:
- Squash merge commit


src/app-layer-smtp.c
```diff
@@ -147,6 +147,8 @@ SCEnumCharMap smtp_decoder_event_table[ ] = {
       SMTP_DECODER_EVENT_MIME_LONG_HEADER_VALUE },
     { "MIME_LONG_BOUNDARY",
       SMTP_DECODER_EVENT_MIME_BOUNDARY_TOO_LONG },
+    { "MIME_LONG_FILENAME",
```
catenacyber: Should not we add a rule for this event in rules/smtp-event.rules ?

jlucovsky: Thanks for pointing this out.

We should -- I'll make an update.

jlucovsky:
Continued in #4578

PR #4406, opened 2019-11-29
github actions: extract repo/branch names from PR message - v1

jasonish:
Create a "prep" build that parses libhtp, suricata-update,
and suricata-verify repo and versions from the pull-request
message and creates a prep archive with these versions
which can then be used by the other builds.

Also modify builds that use the distribution archive to
extract suricata into the home directory to avoid needing
working-directory on every step (this is what is done
by the github provided code checkout action).

Examples:

#libhtp-repo: https://github.com/OISF/libhtp
#libhtp-branch: master

#suricata-verify-repo: http://github.com/OISF/suricata-verify
#suricata-verify-branch: master

#suricata-update-repo: http://github.com/OISF/suricata-update
#suricata-update-branch: master

Prscript not applicable.

PR #4572, opened 2020-02-19
github: Update PR template: Add Suricata-verify

jlucovsky:
Describe changes:
- Add placeholder for suricata-verify in PR template.


victorjulien:
@jasonish should this be integrated with #4406, or perhaps #4406 be updated to make use of this?

victorjulien:
I think we can use #4900 for the same purpose, with added benefit of it being tested by CI?

jasonish:
> I think we can use #4900 for the same purpose, with added benefit of it being tested by CI?

Yeah, I should update the template in that branch.

PR #4574, opened 2020-02-19
Fix for bug #2423 "Napatech crashing"

pjyoung1:
This fixes https://redmine.openinfosecfoundation.org/issues/2423 where
a specific traffic pattern resulted in Suricata's TmThreadsSlotProcessPkt()
function to get stuck in a loop trying to read from an empty queue.  This
was the result of the packet data structure being released too soon.

    [Bug #2423: Suricata 4.0.3 and Napatech crashing](https://redmine.openinfosecfoundation.org/issues/2423)

Describe changes:
-    Moved the PacketFreeOrRelease() function from the Packet Release callback function to a point after the completion of TmThreadsSlotProcessPkt().  See, the Redmine bug report for details.

### DIFF
--- a/src/source-napatech.c
+++ b/src/source-napatech.c
@@ -655,8 +655,9 @@ static PacketQueue packets_to_release[MAX_STREAMS];
  */
 static void NapatechReleasePacket(struct Packet_ *p)
 {
-    PacketFreeOrRelease(p);
-    PacketEnqueue(&packets_to_release[p->ntpv.stream_id], p);
+    if (!PKT_IS_PSEUDOPKT(p)) {
+    	PacketEnqueue(&packets_to_release[p->ntpv.stream_id], p);
+    }
 }
 
 /**
@@ -966,6 +967,7 @@ TmEcode NapatechPacketLoop(ThreadVars *tv, void *data, void *slot)
                 ProgramFlow(rel_pkt, is_inline);
             }
 #endif
+            PacketFreeOrRelease(rel_pkt);
             NT_NetRxRelease(ntv->rx_stream, rel_pkt->ntpv.nt_packet_buf);
             rel_pkt = PacketDequeue(&packets_to_release[ntv->stream_id]);
         }


victorjulien:
Not related to this change, but something that caught my eye while reviewing:
```c
        Packet *p = PacketGetFromQueueOrAlloc();
#ifdef NAPATECH_ENABLE_BYPASS
        p->ntpv.bypass = 0;
#endif
        if (unlikely(p == NULL)) {
```
If `p == NULL` this is a NULL-ptr deref: `p->ntpv.bypass = 0;`

src/source-napatech.c
```diff
@@ -655,8 +655,9 @@ static PacketQueue packets_to_release[MAX_STREAMS];
  */
 static void NapatechReleasePacket(struct Packet_ *p)
 {
-    PacketFreeOrRelease(p);
```
victorjulien: I don't really understand this change. The `ReleasePacket` callback is meant to call `PacketFreeOrRelease` after some cleanup specific to the capture method. Why can't it be done fully here? I mean including `NT_NetRxRelease` here in the callback and getting rid of `packets_to_release` completely?

pjyoung1:
Hi Victor,

It’s been a long time since this particular piece of code was implemented.  But, as I recall the packets_to_release was set when we implemented zero copy support.  At the time it appeared that the callback might be made from a separate thread.  And it’s necessary that NT_NetRxRelease() is always called from within the same thread where the packet it received.  I can take another look.  There may be something else here that I’ve forgotten.  But, perhaps it isn’t necessary if the callback is guaranteed to always be in the same thread at NT_NetRxGet().

The PacketFreeOrRelease() is a different issue.  What I discovered with this particular bug that if:
Traffic is sent in from the pcap file I attached to the ticket.
The flow timeout occurs (flow-timeouts.tcp.closed in the .yaml)
The same traffic file is sent a second time.

Then TmThreadsSlotVarRun() gets stuck in a loop where it keeps trying to read from decode_pq and the return value is always NULL.  What I observed is that under these conditions PacketFreeOrRelease() seems to be called prematurely.  Moving it after the completion of TmThreadsSlotVarRun() eliminated the issue.  Perhaps this is actually a bug in how the decode_pq queue is created or processed during the flow timeout.  I don’t really understand what is happening in this portion of the code.

Phil

victorjulien:
> Hi Victor, It’s been a long time since this particular piece of code was implemented. But, as I recall the packets_to_release was set when we implemented zero copy support. At the time it appeared that the callback might be made from a separate thread. And it’s necessary that NT_NetRxRelease() is always called from within the same thread where the packet it received. I can take another look. There may be something else here that I’ve forgotten. But, perhaps it isn’t necessary if the callback is guaranteed to always be in the same thread at NT_NetRxGet(). The PacketFreeOrRelease() is a different issue. What I discovered with this particular bug that if: Traffic is sent in from the pcap file I attached to the ticket. The flow timeout occurs (flow-timeouts.tcp.closed in the .yaml) The same traffic file is sent a second time. Then TmThreadsSlotVarRun() gets stuck in a loop where it keeps trying to read from decode_pq and the return value is always NULL. What I observed is that under these conditions PacketFreeOrRelease() seems to be called prematurely. Moving it after the completion of TmThreadsSlotVarRun() eliminated the issue. Perhaps this is actually a bug in how the decode_pq queue is created or processed during the flow timeout. I don’t really understand what is happening in this portion of the code. Phil 

The current Napatech support in our tree only supports the `workers` runmode. This makes sense, as its the best performing mode. In this mode each thread captures packets, runs the full packet pipeline, and then returns the packet to the packet pool. So everything happens in that single thread.

(btw, the code defines `RunModeNapatechAutoFp` but does not use it)

The `ThreadVars::decode_pq` packet queue is not used by flow timeout, but only by packet decoders that decapsulate packets and by the flow-worker to make sure locally generate stream end pseudo packets are processed. Since you're in workers, in both cases packets will be taken from the thread local packet pool and returned to the same packet pool directly.

I don't see how it would be possible to get stuck in this loop:
```c
        /* handle new packets */
        while (tv->decode_pq.top != NULL) {
            Packet *extra_p = PacketDequeueNoLock(&tv->decode_pq);
            if (unlikely(extra_p == NULL))
                continue;

            /* see if we need to process the packet */
            if (s->slot_next != NULL) {
                r = TmThreadsSlotVarRun(tv, extra_p, s->slot_next);
                if (unlikely(r == TM_ECODE_FAILED)) {
                    TmThreadsSlotProcessPktFail(tv, s, extra_p);
                    return TM_ECODE_FAILED;
                }
            }
            tv->tmqh_out(tv, extra_p);
        }
```
Can you expand a bit more on what you saw?

I'm trying to understand to avoid getting in a situation where we fix a symptom, but not the cause.

pjyoung1:
What I see in this loop is:

        /* handle new packets */
        while (tv->decode_pq.top != NULL) {
            Packet *extra_p = PacketDequeueNoLock(&tv->decode_pq);
            if (unlikely(extra_p == NULL))
                continue;

that when I get here after the timeout occurs tv->decode_pq.top is non-NULL and PacketDequeueNoLock((&tv->decode_pq) repeatedly returns NULL to extra_p.  The situation is rare and was very difficult to reproduce.  Even once I got the pcap file it took a while to realize that it was dependent on the timeout.  I tried this with both AFPacket (non-zero-copy) and libPCAP and was unable to reproduce it in either or those input methods.  However, it’s hard to understand what my code could have to do with this other than corruption as a result of the packet being release pre-maturely.

The notable thing about the traffic is that the file has an http connection that is established and immediately reset.  Followed by virtually the same sequence.  Perhaps this is somehow confusing the decoder…

pjyoung1:
BTW: I just remembered that there is another reason for the packet_to_release queue.  When running in inline mode we use the PACKET_TEST_ACTION() macro to determine if a packet should be passed out of the egress port or dropped.  I need to know this before I call NT_NetRxRelease().  But, the action isn’t set in the packet during the release callback.  (See: https://redmine.openinfosecfoundation.org/issues/3122.)

From: Philip Young
Sent: Thursday, February 20, 2020 2:57 PM
To: 'OISF/suricata' <reply@reply.github.com>; OISF/suricata <suricata@noreply.github.com>
Cc: Author <author@noreply.github.com>
Subject: RE: [OISF/suricata] Fix for bug #2423 "Napatech crashing" (#4574)

What I see in this loop is:

        /* handle new packets */
        while (tv->decode_pq.top != NULL) {
            Packet *extra_p = PacketDequeueNoLock(&tv->decode_pq);
            if (unlikely(extra_p == NULL))
                continue;

that when I get here after the timeout occurs tv->decode_pq.top is non-NULL and PacketDequeueNoLock((&tv->decode_pq) repeatedly returns NULL to extra_p.  The situation is rare and was very difficult to reproduce.  Even once I got the pcap file it took a while to realize that it was dependent on the timeout.  I tried this with both AFPacket (non-zero-copy) and libPCAP and was unable to reproduce it in either or those input methods.  However, it’s hard to understand what my code could have to do with this other than corruption as a result of the packet being release pre-maturely.

The notable thing about the traffic is that the file has an http connection that is established and immediately reset.  Followed by virtually the same sequence.  Perhaps this is somehow confusing the decoder…

pjyoung1:
I'm replacing this Pull request with another that also includes a couple of other minor fixes.

PR #4575, opened 2020-02-19
fix ( theoretically but highly unlikely exploitable ) config-/runtime…

magenbluten:
let `ctx->size := 32769` with `addr_hash := 32768` you get `qid => -32768` for example. although unlikely to be exploitable, still awkward to see. i removed the bogus and wrong comment as well.

note: to get ctx size = 32769 you would need to specifiy this many queues in auto_fp mode. not a realistic scenario. creating a fake packet with "matching" hash is trivial though.

### DIFF
--- a/src/tmqh-flow.c
+++ b/src/tmqh-flow.c
@@ -220,7 +220,7 @@ void TmqhOutputFlowFreeCtx(void *ctx)
 
 void TmqhOutputFlowHash(ThreadVars *tv, Packet *p)
 {
-    int16_t qid = 0;
+    uint16_t qid = 0;
 
     TmqhFlowCtx *ctx = (TmqhFlowCtx *)tv->outctx;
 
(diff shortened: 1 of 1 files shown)


victorjulien:
Maybe I am missing the vuln, but this would only be an issue if we have more than 32k (SHRT_MAX) threads?

magenbluten:
well, you can perfectly specify 32k queues in the config file.
but then yes it is unlikely ( or impossible, dunno) that that amount 
of threads can be created. maybe somewhere else in your code you
limit the amount of queues (which also not neccessarily means threads?)

anyhow. to comment is wrong and the overflow could happen.

just make it unsigned int or uint16_t. no big deal. or leave as is if you're
not in the mood. just wanted to report that there might be an issue. i
have done that.

victorjulien:
I think the runmodes setup code limits threads to 1024 currently. Since generally 1 thread per cpu core performs best I think we can safely assume we won't hit that 1k limit anytime soon.

Anyway, I do agree the code is wrong. Just wanted to make sure I didn't overlook the issue.

victorjulien:
Addressed in #5183, thanks.

PR #4578, opened 2020-02-20

jlucovsky:
Continuation of #4571

truncated name.

Describe changes:
- Adds rule to `smtp-events.rules` for new event.


src/util-decode-mime.c
```diff
@@ -1830,22 +1831,38 @@ static int FindMimeHeader(const uint8_t *buf, uint32_t blen,
...
-        tok = GetToken(fptr, field->value_len - (fptr - field->value), search_end,
-                &tptr, tlen);
+        fptr += ss_len; /* Start at end of start string */
+        uint32_t offset = fptr - field->value;
+        if (offset > field->value_len) {
+            return tok;
+        }
+        uint32_t search_len = field->value_len - offset;
+        if (max_len && search_len > max_len) {
+            SCLogDebug("Token value length for %s exceeds limit: %d; truncating to limit: %d",
```
catenacyber: What happens if the mime message is
`Content-Disposition: attachment; filename="something"; lots of junk...`

It looks to me like we alert on a too long name, event if it is not the name which is too long.
We should compare with `tlen` instead, no

jlucovsky: Good catch.

jlucovsky:
Continued in #4585

PR #4576, opened 2020-02-20
fuzz: adds eight fuzz targets

catenacyber:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2859

Describe changes:
- Adds eight fuzz targets
- Change compilation scheme so as to compile fuzz targets (option --enable-fuzztargets for configure)

Modifies #4309 with 
- needed rebase

After this, we can 
- review and delete `#ifdef AFLFUZZ_PCAP_RUNMODE` and such
Who best can tell if we can remove these chunks of code ?
- Change .travis.yml  to make a build with `--enable-fuzztargets` configure option, then run the fuzz targets with sanitizers enabled
- Get the PR to oss-fuzz running : https://github.com/google/oss-fuzz/pull/2533 and fix the found bugs, add a status badge, try CIFuzz...

PR #4582, opened 2020-02-20
fuzz: adds eight fuzz targets

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds eight fuzz targets

Modifies #4576 with style comments taken into account

After this, we can 

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@
     AC_CONFIG_HEADERS([config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
     AC_CONFIG_MACRO_DIR(m4)
-    AM_INIT_AUTOMAKE([tar-ustar])
+    AM_INIT_AUTOMAKE([tar-ustar], [subdir-objects])
 
     AC_LANG([C])
     AC_PROG_CC_C99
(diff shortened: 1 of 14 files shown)


configure.ac
```diff
@@ -3,7 +3,7 @@
     AC_CONFIG_HEADERS([config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
     AC_CONFIG_MACRO_DIR(m4)
-    AM_INIT_AUTOMAKE([tar-ustar])
+    AM_INIT_AUTOMAKE([tar-ustar], [subdir-objects])
```
jasonish: This should be:
```
     AM_INIT_AUTOMAKE([tar-ustar subdir-objects])
```
This will eliminate the following errors during autogen:
```
src/Makefile.am:573: warning: source file 'tests/fuzz/onefile.c' is in a subdirectory,
src/Makefile.am:573: but option 'subdir-objects' is disabled
```

catenacyber: Ok, coming in fuzz-v23

src/tests/fuzz/fuzz_applayerparserparse.c
```diff
@@ -0,0 +1,137 @@
...
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+AppLayerParserThreadCtx *alp_tctx = NULL;
+
+/* input buffer is structured this way :
+ * 6 bytes header,
+ * then sequence of buffers separated by magic bytes 01 D5 CA 7A */
+
+/* The 6 bytes header is
+ * alproto
```
victorjulien: I could see value in a (set of) target(s) that allow us to specifically test a app-layer parser. For example when we introduce a new parser, or a large change (e.g. big libhtp update for http) we really want to hammer it for some time.

catenacyber: This is interesting. Here are a few elements for thinking.

This target is quite easy to get modified to test a specific app-layer.
We can copy/paste the code (n times) and enforce the value.
Or we can use an environment variable, or maybe an argument with https://llvm.org/docs/LibFuzzer.html#startup-initialization but I am not sure if we can fit this into oss-fuzz
Would that fit your use case ?

This fuzz target allows to fuzz a new parser without changes
Otherwise, you need to add a corresponding new fuzz target when you add a new parser...

I think that the fuzzing engines are smart enough, when there is one big change, to "hammer" it (because they find new code coverage)

victorjulien: Is there a way to find out in what alproto a running fuzzer spends most of its time? As an example, I would like to 'hammer' a new protocol for at least a week before doing a release. However if it turns out the fuzzer spends almost all its time in http/libhtp, the new protocol may have hardly been fuzzed.

I do think using an env var could be good enough. It would allow the manual 'hammering' while otherwise relying on the smartness of the coverage based approach.

catenacyber: > Is there a way to find out in what alproto a running fuzzer spends most of its time?

oss-fuzz generates a nice coverage report
You can do it locally with your corpus as well.

> using an env var could be good enough. It would allow the manual 'hammering'

Yes that would be manual.
I will propose something like that

victorjulien: > > Is there a way to find out in what alproto a running fuzzer spends most of its time?
> 
> oss-fuzz generates a nice coverage report
> You can do it locally with your corpus as well.

Can you update the bundled doc with instructions on how to generate this?

catenacyber: Ok, doc is here : https://google.github.io/oss-fuzz/advanced-topics/code-coverage/

src/Makefile.am
```diff
@@ -539,11 +544,116 @@ EXTRA_DIST = tests
...
 suricata_DEPENDENCIES = $(RUST_SURICATA_LIB)
 
+fuzz_applayerprotodetectgetproto_SOURCES = tests/fuzz/fuzz_applayerprotodetectgetproto.c $(COMMON_SOURCES)
+fuzz_applayerprotodetectgetproto_LDFLAGS = $(all_libraries) ${SECLDFLAGS}
+fuzz_applayerprotodetectgetproto_LDADD = $(RUST_SURICATA_LIB) $(HTP_LDADD) $(RUST_LDADD)
+if HAS_FUZZLDFLAGS
+    fuzz_applayerprotodetectgetproto_LDFLAGS += $(LIB_FUZZING_ENGINE)
+else
+    fuzz_applayerprotodetectgetproto_SOURCES += tests/fuzz/onefile.c
+endif
```
jasonish: To fix the argument to long thing during `make dist`, you can replace `fuzz_applayerprotodetectgetproto_SOURCES` with `nodist_fuzz_applayerprotodetectgetproto_SOURCES` on lines 554 and 558, and do this for all the other fuzz targets.

All these files will be included in the dist anyways when the tests directory is included.

src/tests/fuzz/README
```diff
@@ -0,0 +1,51 @@
...
+
+To compile libFuzzer, you can do the following
+```
+svn co http://llvm.org/svn/llvm-project/compiler-rt/trunk/lib/fuzzer
+cd fuzzer
+./build.sh
+```
+
+
+3) With afl
```
victorjulien: I tried this and it works. However exec/s are very low (100/s or so), even though I compiled with afl-clang-fast. The older `--enable-afl` target are much much fast. Can we support the `AFLFUZZ_PERSISTANT_MODE` somehow?

```c
#ifdef AFLFUZZ_PERSISTANT_MODE
    while (__AFL_LOOP(1000)) {
        /* reset state */
        memset(buffer, 0, sizeof(buffer));
#endif /* AFLFUZZ_PERSISTANT_MODE */
```

catenacyber: I am adding code for this. Could you test it when I make the updated PR ?

victorjulien: Absolutely.

PR #4585, opened 2020-02-21

jlucovsky:
Continuation of #4578

truncated name.

Describe changes:
- Handles this case where the field has excessive length but the token does not: `Content-Disposition: attachment; filename="something"; lots of junk...`


src/util-decode-mime.c
```diff
@@ -1830,30 +1831,77 @@ static int FindMimeHeader(const uint8_t *buf, uint32_t blen,
...
-                &tptr, tlen);
-        if (tok != NULL) {
-            SCLogDebug("Found mime token");
+        fptr += ss_len; /* Start at end of start string */
+        uint32_t offset = fptr - field->value;
+        if (offset > field->value_len) {
+            return tok;
+        }
+        uint32_t search_len = field->value_len - offset;
+        if (max_len && search_len > max_len) {
```
catenacyber: I think that we could make the code simpler.
We can call `GetToken` without prior checks...
Then directly check with `tlen` to see if we must set `toolong`

jlucovsky: Good suggestions; thanks Phillipe!

jlucovsky:
Continued in #4612

PR #4586, opened 2020-02-21
Feat/hdlc/v4

victorjulien:
Describe changes:
- basic Cisco HDLC link layer support for pcap (file) and af-packet.

Passed.

SV https://github.com/OISF/suricata-verify/pull/181


rules/decoder-events.rules
```diff
@@ -143,5 +143,8 @@ alert pkthdr any any -> any any (msg:"SURICATA ERSPAN too many vlan layers"; dec
 # Cisco Fabric Path/DCE
 alert pkthdr any any -> any any (msg:"SURICATA DCE packet too small"; decode-event:dce.pkt_too_small; classtype:protocol-command-decode; sid:2200110; rev:2;)
 
-# next sid is 2200114
+# Cisco HDLC
+alert pkthdr any any -> any any (msg:"SURICATA HDLC packet too small"; decode-event:hdlc.pkt_too_small; classtype:protocol-command-decode; sid:2200115; rev:1;)
+
+# next sid is 2200115
```
jlucovsky: nit: shouldn't the "next" sid used be `2200116`?

src/decode-hdlc.c
```diff
@@ -0,0 +1,122 @@
...
+{
+    StatsIncr(tv, dtv->counter_hdlc);
+
+    if (unlikely(len < HDLC_HEADER_LEN)) {
+        ENGINE_SET_INVALID_EVENT(p, HDLC_PKT_TOO_SMALL);
+        return TM_ECODE_FAILED;
+    }
+
+    if (unlikely(len > HDLC_HEADER_LEN + USHRT_MAX)) {
+        return TM_ECODE_FAILED;
```
jlucovsky: Should an event be set in this case too?

src/decode-hdlc.c
```diff
@@ -0,0 +1,122 @@
...
+    memset(p, 0, SIZE_OF_PACKET);
+
+    DecodeHDLC(&tv, &dtv, p, raw, sizeof(raw));
+
+    FAIL_IF_NOT(PKT_IS_IPV4(p));
+    FAIL_IF_NOT(PKT_IS_TCP(p));
+    FAIL_IF_NOT(p->dp == 80);
+
+    SCFree(p);
+    return 1;
```
victorjulien: use PASS

rules/decoder-events.rules
victorjulien: yes, good catch

src/decode-hdlc.c
victorjulien: Good question. I don't think we do this in other places either. I wonder if we should set an event, and then limit the len we pass on to what we support (for an IP packet max it to USHRT_MAX).  The event could be something like 'unexpected trailing data'. However it would require some investigation if trailing data might be normal. For example in ethernet it can be.

victorjulien: Created a ticket for this https://redmine.openinfosecfoundation.org/issues/3503

victorjulien:
Replaced by #4602.

PR #4590, opened 2020-02-22

jlucovsky:
Fix memory leaks identified during fuzzing that are caused by rule parse errors. A suricata-verify test is in progress.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3489](https://redmine.openinfosecfoundation.org/issues/3489)

Describe changes:
- Use `pcre_copy_substring` instead of `pcre_get_substring` to avoid memory leaks on parse errors
- Documentation fixup -- misplaced back-tick charracters


jlucovsky:
Continued in #4591

PR #4591, opened 2020-02-22

jlucovsky:
Continuation of #4590 

Fix memory leaks identified during fuzzing that are caused by rule parse errors. A suricata-verify test is in progress.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3489](https://redmine.openinfosecfoundation.org/issues/3489)

Describe changes:
- Increases buffer size for ssl_state values


jlucovsky:
Continued in #4593

PR #4595, opened 2020-02-24
[PREVIEW] detect/transform: Transform options and pcrexform (new transform)

jlucovsky:
This PR is a preview and includes #4519.

This PR demonstrates `pcrexform` -- a new transform supporting transform options (#4519). Transform options are supported _when the transform implementation supports them_. This means that the transform infrastructure will provide the transform-specific options container to the transform evaluation logic. When no longer needed, the infrastructure will invoke the transform's `Free` function to release the options container.

`pcrexform` applies a (mandatory) RE to the transformation buffer and emits the first capture, if any, into the transformation buffer. If there is not a match, the transformation buffer is unmodified.

- [3199](https://redmine.openinfosecfoundation.org/issues/3199)
- [3200](https://redmine.openinfosecfoundation.org/issues/3200)

Describe changes:
- Transform option support (from PR #4519) 
- Pcrexform transform supporting transform options

Companion [Suricata-verify PR](https://github.com/OISF/suricata-verify/pull/183)

### DIFF
--- a/src/detect-transform-pcrexform.c
+++ b/src/detect-transform-pcrexform.c
@@ -1,0 +1,113 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Jeff Lucovsky <jeff@lucovsky.org>
+ *
+ * Implements the pcrexform transform keyword with option support
+ */
+
+#include "suricata-common.h"
+
+#include "detect.h"
+#include "detect-engine.h"
+#include "detect-parse.h"
+#include "detect-transform-pcrexform.h"
+
+typedef struct DetectTransformPcrexformData_ {
+    pcre *parse_regex;
+    pcre_extra *parse_regex_study;
+} DetectTransformPcrexformData;
+
+static int DetectTransformPcrexformSetup (DetectEngineCtx *, Signature *, const char *);
+static void DetectTransformPcrexformFree(void *);
+static void DetectTransformPcrexform(InspectionBuffer *buffer, void *options);
+
+void DetectTransformPcrexformRegister(void)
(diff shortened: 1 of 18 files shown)


src/detect-transform-pcrexform.c
```diff
@@ -0,0 +1,113 @@
...
+    sigmatch_table[DETECT_TRANSFORM_PCREXFORM].url =
+        DOC_URL DOC_VERSION "/rules/transforms.html#pcre-xform";
+    sigmatch_table[DETECT_TRANSFORM_PCREXFORM].Transform =
+        DetectTransformPcrexform;
+    sigmatch_table[DETECT_TRANSFORM_PCREXFORM].Free =
+        DetectTransformPcrexformFree;
+    sigmatch_table[DETECT_TRANSFORM_PCREXFORM].Setup =
+        DetectTransformPcrexformSetup;
+
+    sigmatch_table[DETECT_TRANSFORM_PCREXFORM].flags |= SIGMATCH_QUOTES_OPTIONAL;
```
victorjulien: I suppose we can enforce this as its all new functionality.

src/detect-transform-pcrexform.c
```diff
@@ -0,0 +1,113 @@
...
+    // Create pxd from regexstr
+    DetectTransformPcrexformData *pxd = SCCalloc(sizeof(*pxd), 1);
+    if (pxd == NULL) {
+        SCLogDebug("pxd allocation failed");
+        SCReturnInt(-1);
+    }
+
+    DetectSetupParseRegexes(regexstr, &pxd->parse_regex, &pxd->parse_regex_study);
+
+    int r = DetectSignatureAddTransform(s, DETECT_TRANSFORM_PCREXFORM, pxd);
```
victorjulien: should we free `pxd` on failure here?

src/detect-transform-pcrexform.c
```diff
@@ -0,0 +1,113 @@
...
+    const char *input = (const char *)buffer->inspect;
+    const uint32_t input_len = buffer->inspect_len;
+    DetectTransformPcrexformData *pxd = options;
+
+    int ov[MAX_SUBSTRINGS];
+    int ret = pcre_exec(pxd->parse_regex, pxd->parse_regex_study, (char *)input,
+                        input_len, 0, 0, ov, MAX_SUBSTRINGS);
+
+    if (ret > 0) {
+        char str[128];
```
victorjulien: I think we need to consider how big this is allowed to be. 128 seems far too low. Can we find out the length before doing a stack allocation of that length?

jlucovsky: Yes ... we know that the captured string is `<=` the size of the inspection buffer. I think that `pcre_get_substring` is a good candidate to use here.

src/detect-transform-pcrexform.c
jlucovsky: Are you saying it's ok to permit optional quotes or that quotes should be required?

src/detect-transform-pcrexform.c
jlucovsky: Yes; will update.

jlucovsky:
Continued in #4611

PR #4601, opened 2020-02-25
rfb: add RFB parser - v2

satta:
Previous PR: https://github.com/OISF/suricata/pull/4600

Describe changes:
- Include RFB keyword page in documentation.
- Remove debug output.
- Enable RFB by default in detection and add default ports.
- Update copyright dates.
- Add documentation URL entries to sigmatch table entries.


jlucovsky:
There are some unaddressed comments from #4600

rust/src/rfb/rfb.rs
```diff
@@ -0,0 +1,851 @@
...
+        loop {
+            if current.len() == 0 {
+                return true;
+            }
+            match self.state {
+                parser::RFBGlobalState::TSClientProtocolVersion => {
+                    match parser::parse_protocol_version(current) {
+                        Ok((rem, request)) => {
+                            current = rem;
+                            if request.major.to_string() == "003" && request.minor.to_string() == "003" {
```
jasonish: Should be able to compare without the .to_string() here (avoids short lived copies).

rust/src/rfb/rfb.rs
```diff
@@ -0,0 +1,851 @@
...
+        if AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF) > 0 {
+            true
+        } else {
+            false
+        }
+    };
+
+    if eof {
+        // If needed, handled EOF, or pass it into the parser.
+    }
```
jasonish: The template should probably change here..  This can probably changed to just return if input_len is 0. Checking for EOF is primarily useful if some cleanup needs to be done, like protocols supporting file extraction.

satta:
> There are some unaddressed comments from #4600

I see. So far we only adressed Victor's comments -- sorry. We'll address them in v3.

> Looks really good on my first pass. Do you have any Suricata-Verify tests for RFB?

I'll look into it and add some.

rust/src/rfb/rfb.rs
fhonza: would a return -1 be correct in this case?
If understanding logic in app-layer-parser.c correctly a return false would lead to disabling the applayer-parser for this flow....

satta: Returned 1 for now.

satta:
New PR: #4610

PR #4602, opened 2020-02-25
Feat/hdlc/v5

victorjulien:
#4586, rebased and comments addressed.

Passed.


src/decode-hdlc.c
```diff
@@ -0,0 +1,122 @@
...
+                   const uint8_t *pkt, uint32_t len)
+{
+    StatsIncr(tv, dtv->counter_hdlc);
+
+    if (unlikely(len < HDLC_HEADER_LEN)) {
+        ENGINE_SET_INVALID_EVENT(p, HDLC_PKT_TOO_SMALL);
+        return TM_ECODE_FAILED;
+    }
+
+    if (unlikely(len > HDLC_HEADER_LEN + USHRT_MAX)) {
```
catenacyber: Why do we need this check ?

src/decode-hdlc.c
```diff
@@ -0,0 +1,122 @@
...
+        case ETHERNET_TYPE_IP:
+            DecodeIPV4(tv, dtv, p, pkt + HDLC_HEADER_LEN,
+                       len - HDLC_HEADER_LEN);
+            break;
+        case ETHERNET_TYPE_IPV6:
+            DecodeIPV6(tv, dtv, p, pkt + HDLC_HEADER_LEN,
+                       len - HDLC_HEADER_LEN);
+            break;
+        default:
+            SCLogDebug("p %p pkt %p ether type %04x not supported", p,
```
catenacyber: Should we set an event ? Are there no other ether types such as VLAN here ?

victorjulien:
SV OISF/suricata-verify#181

src/decode-hdlc.c
victorjulien: See https://github.com/OISF/suricata/pull/4586#discussion_r383956184 from the last PR.

jasonish:
Do we have to worry about the other non-Cisco HDLC? A peak at my 2016 PR used CHDLC just in case, but I'm not sure if its ever used in the wild.

src/decode-hdlc.c
victorjulien: I don't know, I assume so. I've done a different approach in #4613

victorjulien: I don't know, I assume so. I've done a different approach in #4613

victorjulien:
> Do we have to worry about the other non-Cisco HDLC? A peak at my 2016 PR used CHDLC just in case, but I'm not sure if its ever used in the wild.

In #4613 I've added a 'C' prefix to address this.

victorjulien:
Replaced by #4613

PR #4589, opened 2020-02-21
DNS cleanup: Move more C pieces to Rust - v3

jasonish:
Moves more of the remaining C parser bits to Rust. Most of the C
was just wrappers around Rust functions anyways.

Also a small DHCP cleanup where it also just wrapped Rust.

Previous PR: https://github.com/OISF/suricata/pull/4513

Changes from last PR:
- Remove unused DNS settings from suricata.yaml.
- Use "silent errors" for unknown app-layer decoder events.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3234

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/440
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/796

PR #4604, opened 2020-02-25
DNS cleanup: Move more C pieces to Rust - v5

jasonish:
Moves more of the remaining C parser bits to Rust. Most of the C
was just wrappers around Rust functions anyways.

Also a small DHCP cleanup where it also just wrapped Rust.

Previous PR: https://github.com/OISF/suricata/pull/4589

Changes from last PR:
- Introduce silent/ok errors for rule loading errors that
  should still allow -T to pass.
- Return silent/ok when an app-layer event code is unknown
  unless strict-rule-keywords is on, then it will return
  a normal error and -T will fail.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/443
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/800


jasonish:
Github appears to be having issue with checks being triggered right now.

rust/src/dns/dns.rs
```diff
@@ -821,9 +917,16 @@ pub extern "C" fn rs_dns_tx_get_query_rrtype(tx: &mut DNSTransaction,
...
+    _flow: *const core::Flow,
+    _dir: u8,
+    input: *const u8,
+    len: u32,
+    rdir: *mut u8,
+) -> AppProto {
+    if len == 0 || len < std::mem::size_of::<DNSHeader>() as u32 {
+        return core::ALPROTO_UNKNOWN;
+    }
     let slice: &[u8] = unsafe {
```
catenacyber: we can use the `build_slice!` macro here.

rust/src/dns/dns.rs
```diff
@@ -821,9 +917,16 @@ pub extern "C" fn rs_dns_tx_get_query_rrtype(tx: &mut DNSTransaction,
 }
 
 #[no_mangle]
-pub extern "C" fn rs_dns_probe(input: *const u8, len: u32, rdir: *mut u8)
-                               -> u8
-{
+pub extern "C" fn rs_dns_probe(
+    _flow: *const core::Flow,
+    _dir: u8,
```
catenacyber: This should be used cf https://github.com/OISF/suricata/pull/4058

rust/src/dns/dns.rs
```diff
@@ -861,15 +969,116 @@ pub extern "C" fn rs_dns_probe_tcp(direction: u8,
...
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_dns_udp_register_parser() {
+    let default_port = std::ffi::CString::new("[53]").unwrap();
+    let parser = RustParser{
+        name: b"dns\0".as_ptr() as *const std::os::raw::c_char,
+        default_port: default_port.as_ptr(),
+        ipproto: IPPROTO_UDP,
+        probe_ts: Some(rs_dns_probe),
+        probe_tc: Some(rs_dns_probe),
```
catenacyber: We should distinguish both probing parsers

rust/src/dns/dns.rs
jasonish: We could, but then I'd get an error about the function being unsafe.  Maybe could just be marked as a TODO til then.

catenacyber: ok

jasonish: Oh, but I suppose I should actually make the function unsafe to support my argument ;)

PR #4605, opened 2020-02-26
protodetect: retry when both sides do not agree

catenacyber:

Describe changes:
- Retry protocol detection above TCP when both sides (client and server) do not agree

Makes test https://github.com/OISF/suricata-verify/pull/117 pass

There is likely a performance impact to be assessed
We keep retrying until both sides agree. Maybe we could fix a threshold on the number of retries.
I am not using `AppLayerParserParse` results as it is designed for relaxed parsing.

This should be done for UDP as well after merging #4058 

Replaces #4217 with rebase and updated git commit message

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -617,6 +617,38 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
                            data, data_len, flags) != 0) {
             goto failure;
         }
+    } else if (f->alproto_ts != f->alproto_tc &&
+               f->alproto_ts != ALPROTO_UNKNOWN && f->alproto_ts != ALPROTO_FAILED &&
+               f->alproto_tc != ALPROTO_UNKNOWN && f->alproto_tc != ALPROTO_FAILED) {
+        bool reverse_flow = false;
+        AppProto alproto_otherdir = (flags & STREAM_TOSERVER) ? f->alproto_tc : f->alproto_ts;
+
+        //partial reset for protocol detection
+        FLOW_RESET_PM_DONE(f, flags);
+        FLOW_RESET_PP_DONE(f, flags);
+        f->probing_parser_toserver_alproto_masks = 0;
+        f->probing_parser_toclient_alproto_masks = 0;
+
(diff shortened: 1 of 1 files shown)


src/app-layer.c
```diff
@@ -617,6 +617,38 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
...
+        AppProto alproto_otherdir = (flags & STREAM_TOSERVER) ? f->alproto_tc : f->alproto_ts;
+
+        //partial reset for protocol detection
+        FLOW_RESET_PM_DONE(f, flags);
+        FLOW_RESET_PP_DONE(f, flags);
+        f->probing_parser_toserver_alproto_masks = 0;
+        f->probing_parser_toclient_alproto_masks = 0;
+
+        //test if we are now agreed with other side
+        AppProto newproto = AppLayerProtoDetectGetProto(app_tctx->alpd_tctx,
```
victorjulien, line 633: I see that the SV test passes with this, but I'm trying to understand why it works. The server starts talking FTP, but then on the 2nd try suri thinks it talks HTTP? Are you ignoring the FTP data somehow?

catenacyber: > The server starts talking FTP, but then on the 2nd try suri thinks it talks HTTP

Yes, that is the test case :
The _evil_ server starts talking FTP, and then serves HTTP requests

> Are you ignoring the FTP data somehow?

I think the FTP data gets ignored, because it was first parsed as FTP with a call to `AppLayerParserParse`, so we do not use it again as HTTP with `AppLayerParserParse`

We could set an event before calling `AppLayerParserStateCleanup` to have some visibility on this

The point of this PR is to have a better protocol detection against _evil_ servers, as when there is a mismatch between the protocol talked by the client, and the one talked by the server, Suricata now always choses the server's protocol.

What do you think ?

PS : #4681 is an easier first step to improve protocol detection, a more consensual case...

catenacyber:
Closing as this should come after https://github.com/OISF/suricata/pull/5759 and https://redmine.openinfosecfoundation.org/issues/4232

PR #4607, opened 2020-02-26
protodetect: handle when multiple patterns match

catenacyber:
This PR is for discussion rather than straight merging.

Describe changes:
- Keep searching when multiple patterns are found for protocol detection

Modifies #4152 with improved handling
Fixes #4606 with fixed compilation

First, we run probing parsers, restricted to the protocols found by pattern matching.
If that fails (and it should not), we run all probing parsers.

Probing parsers, ie `PPGetProto` return first successful protocol and not a list of successful protocols
Do we assume that  probing parsers do not overlap ?
In this case, we should check (with fuzzing?) that it is the case.
Otherwise, we should extend this PR

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -1498,12 +1498,13 @@ AppProto AppLayerProtoDetectGetProto(AppLayerProtoDetectThreadCtx *tctx,
 
     AppProto alproto = ALPROTO_UNKNOWN;
     AppProto pm_alproto = ALPROTO_UNKNOWN;
+    bool rflow = false;
 
     if (!FLOW_IS_PM_DONE(f, direction)) {
         AppProto pm_results[ALPROTO_MAX];
         uint16_t pm_matches = AppLayerProtoDetectPMGetProto(tctx, f,
                 buf, buflen, direction, pm_results, reverse_flow);
-        if (pm_matches > 0) {
+        if (pm_matches == 1) {
             alproto = pm_results[0];
 
             /* HACK: if detected protocol is dcerpc/udp, we run PP as well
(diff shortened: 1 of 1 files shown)


victorjulien:
Can you add a SV test for this as well?

catenacyber:
> Can you add a SV test for this as well?

Ok, I will

Why did you mark this PR as draft ?

inliniac:
On 24-07-2020 13:26, Catena cyber wrote:
>     Can you add a SV test for this as well?
> 
> Ok, I will
> 
> Why did you mark this PR as draft ?

Because its not meant for merge as is stated in the body.


-- 
---------------------------------------------
Victor Julien
http://www.inliniac.net/
PGP: http://www.inliniac.net/victorjulien.asc
---------------------------------------------

catenacyber:
Indeed, and S-V test will help, thanks :-)

catenacyber:
I tried a pcap with `GET /EHLO` and one with `SSH-\xffSMB` and I always got 1 pattern match instead of 2

My guess is that the MPM stops on the first pattern it finds.
So we cannot find more than one pattern.
Next PR is `DEBUG_VALIDATE_BUG_ON` to check this (with fuzzing)

victorjulien:
> I tried a pcap with `GET /EHLO` and one with `SSH-\xffSMB` and I always got 1 pattern match instead of 2
> 
> My guess is that the MPM stops on the first pattern it finds.

The EHLO pattern will only be found in the first 4 bytes, so it makes sense it won't detect it here. But the SSH/SMB pattern is different, I would expect that to match. Can you share the pcaps?

catenacyber:
> But the SSH/SMB pattern is different, I would expect that to match. 

I thought so as well

> Can you share the pcaps?

https://redmine.openinfosecfoundation.org/issues/2757#note-24

PR #3501, opened 2018-10-08
Global memcap v3

glongo:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2316

Describe changes:
- Patchset rebased.

This new feature add a new `global-memcap` which is used to rule all the other memcaps available.
If enabled, the sum of all the memcaps cannot be greater than the value set for the global memcap, and in case a memcap is set to zero, it won't be considered for the sum.
This brings a patch also that permits to set the memcaps value for dns and replace the static list of memcap with a dynamic one.
A user could easily see a list of memcaps running suricata with the `--display-memcaps` option:
`suricata --display-memcaps`
The idea of having a list of memcaps can be useful in case of tuning.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/190
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/54

PR #4610, opened 2020-02-27
rfb: add RFB parser - v3

satta:
Previous PR: #4601 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#3501](https://redmine.openinfosecfoundation.org/issues/3501)

Describe changes:
- Address comments from #4601 and the remaining ones from #4600.


victorjulien:
Marking as 'preview' as we'll need a cleaner git history before the final merge.

satta:
Agreed. We would be happy do merge commits before submitting an 'approved' PR.

rust/src/rfb/logger.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+    js.set("authentication", auth);
+
+    if let Some(ref reason) = tx.tc_failure_reason {
+        js.set_string("server_security_failure_reason", &reason.reason_string);
+    }
+
+    // Client/Server init
+    if let Some(s) = &tx.ts_client_init {
+        js.set_boolean("screen shared", s.shared != 0);
```
satta: I'm quite certain this is meant to be `screen_shared` with an underscore instead of a space.

rust/src/rfb/rfb.rs
```diff
@@ -0,0 +1,837 @@
...
+            *state = index as u64;
+            return Some((tx, tx.tx_id - 1, (len - index) > 1));
+        }
+
+        return None;
+    }
+}
+
+/// Probe to see if this input looks like a request or response.
+fn probe(input: &[u8]) -> bool {
```
catenacyber: Should not this be made as a fixed pattern instead of a probing function ? ie calling `AppLayerProtoDetectPMRegisterPatternCI`

src/app-layer-rfb.h
```diff
@@ -0,0 +1,30 @@
...
+ * \file
+ *
+ * \author Sascha Steinbiss <sascha.steinbiss@dcso.de>
+ */
+
+#ifndef __APP_LAYER_RFB_H__
+#define __APP_LAYER_RFB_H__
+
+void RegisterRFBParsers(void);
+void RFBParserRegisterTests(void);
```
jlucovsky: This shouldn't be needed here.

satta: Yep, looks like it's only called from within `app-layer-rfb.c`.

rust/src/rfb/rfb.rs
satta: Yes! Just changed that and will push it in the next PR. Thanks.

catenacyber: Thanks

PR #4613, opened 2020-02-27
[preview] Feat/hdlc/v8

victorjulien:
Follow up to #4602.

Renames to 'CHDLC'.

Changes behavior in a significant way: a new entry point for network layer decoders is called from all decoders that deal with calling the network layer. It does **not** validate if combinations make sense. For example, we can now have ethernet -> qinq w/o a regular vlan in between. There may be other invalid combinations that would now be decoded.

@jasonish as a side effect I needed to add 6f87507. Not sure why this change is needed now.


src/decode.h
```diff
@@ -1157,5 +1166,87 @@ static inline bool VerdictTunnelPacket(Packet *p)
...
+        case ETHERNET_TYPE_DCE:
+            if (unlikely(len < ETHERNET_DCE_HEADER_LEN)) {
+                ENGINE_SET_INVALID_EVENT(p, DCE_PKT_TOO_SMALL);
+            } else {
+                DecodeEthernet(tv, dtv, p, data, len);
+            }
+            break;
+        default:
+            SCLogDebug("unknown ether type: %" PRIx32 "", proto);
+            ENGINE_SET_INVALID_EVENT(p, VLAN_UNKNOWN_TYPE); // TODO
```
victorjulien: TODO

catenacyber: This is rather ETHER_UNKNOWN_TYPE, is it not ?

jlucovsky: `proto` is a 16-bit value ... `PRIx16`?

victorjulien:
Replaced by #4802

PR #4614, opened 2020-02-28
Fuzz v23

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds eight fuzz targets

Modifies #4582 with :
- fixing subdir-objects in configure.ac
- using `nodist_fuzz_x_SOURCES` to fix `make dist` with duplicated sources
- added coverage documentation
- target fuzz_applayerparserparse can use an environment variable to fuzz only a specific app layer protocol
- use of `AFLFUZZ_PERSISTANT_MODE`

After this, we can 

### DIFF
--- a/src/tests/fuzz/onefile.c
+++ b/src/tests/fuzz/onefile.c
@@ -1,0 +1,51 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size);
+
+int main(int argc, char** argv)
+{
+    FILE * fp;
+    uint8_t *data;
+    size_t size;
+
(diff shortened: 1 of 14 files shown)


src/tests/fuzz/onefile.c
```diff
@@ -13,6 +13,10 @@ int main(int argc, char** argv)
     if (argc != 2) {
         return 1;
     }
+#ifdef AFLFUZZ_PERSISTANT_MODE
```
victorjulien: this is set by our own `configure.ac`, and in this branch only with `--enable-afl`. Think it would be good to test for it when just using `--enable-fuzztargets` as well.

victorjulien: It does show much better speeds
Screenshot from 2020-03-03 11-46-35

PR #4616, opened 2020-02-28
Detect icmpv6hdr v1

catenacyber:

Describe changes:
- Adds `icmpv6.hdr` signature keyword
- Fixes typos in comments
- Adds missing doc for `ipv6.hdr` keyword
- Adds a generic `DetectU32Data` structure for signature matches on uint32 values
- Adds `icmpv6.mtu` signature  keyword
- Fix endianness for ICMPv6 MTU decoding

This fixes only the first evasion of the redmine ticket

Makes https://github.com/OISF/suricata-verify/pull/187 pass


src/detect-icmpv6-mtu.c
```diff
@@ -0,0 +1,190 @@
...
+    //DetectU32Data *icmpv6mtud = (DetectU32Data *)ptr;
+    SCFree(ptr);
+}
+
+/* prefilter code */
+
+static void
+PrefilterPacketIcmpv6mtuMatch(DetectEngineThreadCtx *det_ctx, Packet *p, const void *pectx)
+{
+    if (!(PKT_IS_ICMPV6(p)) || PKT_IS_PSEUDOPKT(p))
```
catenacyber: These 5 lines could be turned into one function as they are used twice... But these are only 5 lines... 🤷‍♂

src/detect-icmpv6-mtu.c
```diff
@@ -0,0 +1,190 @@
...
+    const PrefilterPacketHeaderCtx *ctx = pectx;
+    if (PrefilterPacketHeaderExtraMatch(ctx, p) == FALSE)
+        return;
+
+    /* if we match, add all the sigs that use this prefilter. This means
+     * that these will be inspected further */
+    DetectU32Data du32;
+    du32.mode = ctx->v1.u8[0];
+    du32.arg1 = ctx->v1.u32[1];
+    du32.arg2 = UINT32_MAX;
```
catenacyber: Is this the right thing to do ?

src/detect-icmpv6-mtu.c
```diff
@@ -0,0 +1,190 @@
...
+{
+    return PrefilterSetupPacketHeader(de_ctx, sgh, DETECT_ICMPV6MTU,
+            PrefilterPacketU32Set,
+            PrefilterPacketU32Compare,
+            PrefilterPacketIcmpv6mtuMatch);
+}
+
+static bool PrefilterIcmpv6mtuIsPrefilterable(const Signature *s)
+{
+    const SigMatch *sm;
```
catenacyber: This could call a generic function `PrefilteIsPrefilterableByType(DETECT_ICMPV6MTU)`

What is preferred ?

src/detect-icmpv6hdr.c
```diff
@@ -0,0 +1,128 @@
...
+ * \param s pointer to the Current Signature
+ * \param _unused unused
+ *
+ * \retval 0 on Success
+ * \retval -1 on Failure
+ */
+static int DetectICMPv6hdrSetup (DetectEngineCtx *de_ctx, Signature *s, const char *_unused)
+{
+    // ICMPv6 comes only with IPv6
+    s->proto.flags |= DETECT_PROTO_IPV6;
```
catenacyber: All this file is copied/pasted/adapted from detect-ipv6hdr.c

Except for these 3 lines where it is a mix of detect-ipv6hdr.c and detect-tcphdr.c

src/detect-util-uint.c
```diff
@@ -0,0 +1,239 @@
...
+    return NULL;
+}
+
+void
+PrefilterPacketU32Set(PrefilterPacketHeaderValue *v, void *smctx)
+{
+    const DetectU32Data *a = smctx;
+    v->u8[0] = a->mode;
+    v->u32[1] = a->arg1;
+    // not enough place for a->arg2 in PrefilterPacketHeaderValue
```
catenacyber: Is there anything better to do ?

src/detect-icmpv6-mtu.c
```diff
@@ -0,0 +1,190 @@
...
+
+    return 0;
+}
+
+/**
+ * \brief this function will free memory associated with DetectU32Data
+ *
+ * \param ptr pointer to DetectU32Data
+ */
+void DetectICMPv6mtuFree(void *ptr)
```
catenacyber: Is there a generic wrapper _function_ around `SCFree` ?

src/detect-util-uint.c
```diff
@@ -0,0 +1,239 @@
...
+                }
+
+                SCLogDebug("u32 is %"PRIu32" to %"PRIu32"",u32d->arg1, u32d->arg2);
+                if (u32d->arg1 >= u32d->arg2) {
+                    SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid u32 range. ");
+                    goto error;
+                }
+                break;
+            default:
+                u32d->mode = DETECT_UINT_EQ;
```
catenacyber: maybe we should directly error here

catenacyber:
PS : I can make V2 if I made too much noise with `git push -f`

src/detect-util-uint.c
victorjulien: 2 options I think:

1. mtu is actually u16? So then there is enough space.
2. expand space. It is currently included in:
```c
struct PrefilterPacketHeaderCtx_ {
	PrefilterPacketHeaderValue v1;                   /*     0     8 */
	uint16_t                   type;                 /*     8     2 */
	uint16_t                   value;                /*    10     2 */
	uint32_t                   sigs_cnt;             /*    12     4 */
	uint32_t *                 sigs_array;           /*    16     8 */

	/* size: 24, cachelines: 1, members: 5 */
	/* last cacheline: 24 bytes */
};
```
If we double the space this PrefilterPacketHeaderCtx_ becomes a nice half cache line

src/detect-icmpv6-mtu.c
victorjulien: this would get the actual value if there was enough space, or when it was using the u16 support

src/detect-icmpv6-mtu.c
victorjulien: Not yet, but I should submit a PR soon that turns the current alloc/free macros into functions. So maybe those could help then

victorjulien:
> PS : I can make V2 if I made too much noise with `git push -f`

Please don't push to a branch that already has a PR. Github still isn't good at keeping the previous comments sane. Just do v2 directly after submitting the PR and make changes there.

src/detect-icmpv6-mtu.c
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * \param t pointer to thread vars
+ * \param det_ctx pointer to the pattern matcher thread
+ * \param p pointer to the current packet
+ * \param m pointer to the sigmatch that we will cast into DetectU32Data
+ *
+ * \retval 0 no match
+ * \retval 1 match
+ */
+static int DetectICMPv6mtuMatch (DetectEngineThreadCtx *det_ctx, Packet *p,
```
jlucovsky: nit: parameter comment mismatch with actual.

src/detect-icmpv6hdr.c
```diff
@@ -0,0 +1,128 @@
...
+    if (DetectBufferSetActiveList(s, g_icmpv6hdr_buffer_id) < 0)
+        return -1;
+
+    return 0;
+}
+
+static InspectionBuffer *GetData(DetectEngineThreadCtx *det_ctx,
+        const DetectEngineTransforms *transforms, Packet *p, const int list_id)
+{
+    SCEnter();
```
jlucovsky: Since you're using `SCEnter` .. make sure you use the corresponding `SC` macro when returning from this function.

src/detect-util-uint.h
```diff
@@ -0,0 +1,49 @@
...
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Philippe Antoine <p.antoine@catenacyber.fr>
+ */
+
+#ifndef _DETECT_UTIL_UINT_H
```
jlucovsky: nit/style: There should be 2 `_` before and after the file name: e.g., `__DETECT_UTIL_UINT_H__`

src/detect-icmpv6-mtu.c
catenacyber: ok

src/detect-icmpv6-mtu.c
catenacyber: I do not understand your answer.
There is indeed not enough space for 1 u8 + 2 u32

src/detect-icmpv6-mtu.c
catenacyber: ok

src/detect-util-uint.c
catenacyber: > mtu is actually u16? So then there is enough space.

ICMPv6 MTU is u32 : https://github.com/OISF/suricata/blob/master/src/decode-icmpv6.h#L153

catenacyber: So, you want to add `PrefilterPacketHeaderValue v2;` ?

src/decode-icmpv6.h
```diff
@@ -117,8 +117,10 @@
 #define ICMPV6_GET_UNUSED(p)       (p)->icmpv6h->icmpv6b.icmpv6e.unused
 /** macro for icmpv6 "error_ptr" access */
 #define ICMPV6_GET_ERROR_PTR(p)    (p)->icmpv6h->icmpv6b.icmpv6e.error_ptr
+/** macro for icmpv6 "mtu" accessibility */
+#define ICMPV6_HAS_MTU(p)          ((p)->icmpv6h->type == ICMP6_PACKET_TOO_BIG)
```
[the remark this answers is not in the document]
catenacyber: > "has_mtu" is set if the type is too big? 

Yes, because the packet is too big, we explain our MTU
Adding a comment

src/detect-util-uint.h
catenacyber: ok

src/detect-icmpv6hdr.c
catenacyber: What is the `SC` macro to return a buffer ?
This is copied from detect-ipv6hdr.c

src/detect-icmpv6-mtu.c
catenacyber: Hmm... It seems to me that the comment and the code match 
we return 0 when there is no match
we return 1 when there is a match

Did I get you wrong ?

src/detect-util-uint.c
```diff
@@ -0,0 +1,239 @@
...
+
+static pcre *parse_regex;
+static pcre_extra *parse_regex_study;
+
+
+int DetectU32Match(const uint32_t parg, const DetectU32Data *du32)
+{
+    switch (du32->mode) {
+        case DETECT_UINT_EQ:
+            if (parg == du32->arg1) {
```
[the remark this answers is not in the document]
catenacyber: But I am not sure `(parg == du32->arg1) == 1`, right ?

src/detect-util-uint.c
catenacyber: The structure passed as an argument to `PrefilterPacketU32Set` is `PrefilterPacketHeaderValue` and not its container `PrefilterPacketHeaderCtx_`

src/detect-icmpv6hdr.c
jlucovsky: `SCReturnPtr()`

src/detect-util-uint.c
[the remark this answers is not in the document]
jlucovsky: `return x == y` and `if (x == y) return 1` are equivalent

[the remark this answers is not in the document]
catenacyber: But this does not seem obvious... I can switch the return type to `bool` here, but at some point I will need some readable `if (x == y) return 1 else return 0`

PR #4619, opened 2020-02-29
detect: Use pcre_copy_substring to avoid leaks 

jlucovsky:
Backport of #4590; Requires [Suricata-verify PR #189](https://github.com/OISF/suricata-verify/pull/189)

Fix memory leaks identified during fuzzing that are caused by rule parse errors. A suricata-verify test is in progress.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3507](https://redmine.openinfosecfoundation.org/issues/3507)

Describe changes:
- Use `pcre_copy_substring` instead of `pcre_get_substring` to avoid memory leaks on parse errors
- Documentation fixup -- misplaced back-tick characters

PR #4625, opened 2020-03-03

jlucovsky:
Continuation of #4619 

Backport of #4590; Requires [Suricata-verify PR #189](https://github.com/OISF/suricata-verify/pull/189)

Fix memory leaks identified during fuzzing that are caused by rule parse errors.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3507](https://redmine.openinfosecfoundation.org/issues/3507)

Describe changes:
- Use `pcre_copy_substring` instead of `pcre_get_substring` to avoid memory leaks on parse errors
- Documentation fixup -- misplaced back-tick characters


catenacyber:
Does this mean we should never use `pcre_get_substring` ? And always `pcre_copy_substring` instead ?

jlucovsky:
> Does this mean we should never use `pcre_get_substring` ? And always `pcre_copy_substring` instead ?

Generally, avoid using `pcre_get_substring` unless there's a compelling reason to do so. Most of the strings are short-lived and can result in the need for carefully crafted cleanup code.

Also, see https://redmine.openinfosecfoundation.org/issues/1046

jlucovsky:
Continued in #4718

PR #4481, opened 2020-01-13
Preview: New JsonBuilder - v10

jasonish:
Migrates DHCP logging to use the new JsonBuilder. This is based on
PR #4477 to get Rust 2018 and cbindgen support, so only look at the
commits after that to get an idea of how this looks.

Related ticket:
https://redmine.openinfosecfoundation.org/issues/2726

Previous PR: https://github.com/OISF/suricata/pull/4476

Changes from last PR:
- Drop "sc" prefix on the C functions for the JsonBuilder,
  now its exposed as jb_set_string, etc for JsonBuilder (
  maybe js_ would be better though).
- In DHCP logging, pass in a JsonBuilder instead of
  allocating a new one.
- Remove the rustfmt of logger.rs to easier see the diff.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- Fails as based on #4477 which requires cbindgen.

PR #4634, opened 2020-03-05
Replace libjannson with Rust JsonBuilder - v12

jasonish:
Introduces a new JsonBuilder module to take the place of libjannson
for creating JSON logs.

Converts DHCP and SIP for demonstration purposes.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/2726

Previous PR: https://github.com/OISF/suricata/pull/4481

Changes from last PR:
- Convert SIP logging.
- Add methods for logging metadata like pktvars, community-id.

Notes:
- For SIP we have to keep the old style logger until alert logging
  has been migrated, due to metadata logging.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/446
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/803

### DIFF
--- a/rust/src/dhcp/logger.rs
+++ b/rust/src/dhcp/logger.rs
@@ -21,8 +21,8 @@ use std::os::raw::c_void;
 use crate::dhcp::dhcp::*;
 use crate::dhcp::parser::{DHCPOptionWrapper,DHCPOptGeneric};
 use crate::dns::log::dns_print_addr;
-use crate::json::*;
 use crate::conf::ConfNode;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
 
 pub struct DHCPLogger {
     extended: bool,
@@ -57,7 +57,7 @@ impl DHCPLogger {
         return None;
     }
 
-    fn do_log(&self, tx: &DHCPTransaction) -> bool {
+    pub fn do_log(&self, tx: &DHCPTransaction) -> bool {
         if !self.extended {
             match self.get_type(tx) {
                 Some(t) => {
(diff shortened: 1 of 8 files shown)


catenacyber:
Is the far future plan to completely remove the dependency to libjansson even for C parsers ?

jasonish:
> Is the far future plan to completely remove the dependency to libjansson even for C parsers ?

Yes, and maybe not even that far...

jasonish:
Thoughts on naming...  The "jb_" prefix stands for JsonBuilder, and really only applies in the C bindings as in Rust it lives under the Suricata crate.

My first thoughts is that we don't need to stick with the lower case Rust style, by moving the ffi binding functions into their own module we could suppress that warning, and export a more Suricata  C style binding, say something like `SuriJSONNewArray`.  Assuming we might expose some of these methods in a header file to plugin writers, coming up with a consistent prefix is important, and its been expressed that `Suri` may be better than `SC`.

victorjulien:
> Thoughts on naming... The "jb_" prefix stands for JsonBuilder, and really only applies in the C bindings as in Rust it lives under the Suricata crate.
> 
> My first thoughts is that we don't need to stick with the lower case Rust style, by moving the ffi binding functions into their own module we could suppress that warning, and export a more Suricata C style binding, say something like `SuriJSONNewArray`. Assuming we might expose some of these methods in a header file to plugin writers, coming up with a consistent prefix is important, and its been expressed that `Suri` may be better than `SC`.

I'm of two minds here. On the one hand I'd like us to settle on a prefix and use it. On other hand, I wonder if we should only introduce a 'public' API once we actually support plugins.

jasonish:
> > Thoughts on naming... The "jb_" prefix stands for JsonBuilder, and really only applies in the C bindings as in Rust it lives under the Suricata crate.
> > My first thoughts is that we don't need to stick with the lower case Rust style, by moving the ffi binding functions into their own module we could suppress that warning, and export a more Suricata C style binding, say something like `SuriJSONNewArray`. Assuming we might expose some of these methods in a header file to plugin writers, coming up with a consistent prefix is important, and its been expressed that `Suri` may be better than `SC`.
> 
> I'm of two minds here. On the one hand I'd like us to settle on a prefix and use it. On other hand, I wonder if we should only introduce a 'public' API once we actually support plugins.

Its likely that a utility like JSON building will be fully exposed in that public API, so I think it makes a good case to work on this prefix.

Of note, Rust functions that are marked #[no_mangle] no longer emit a camel case warning like they did when we first started with Rust.  Had that been the case we'd probably have stuck with our C convention for Rust extern "C" functions.

victorjulien:
> Assuming we might expose some of these methods in a header file to plugin writers, coming up with a consistent prefix is important, and its been expressed that `Suri` may be better than `SC`.

The only concern I have here is that I'd hate to introduce a new prefix that lives next to the current SC/Sc... maybe we should first remove SC, or already rename it to Suri...

jasonish:
Continued here:
https://github.com/OISF/suricata/pull/4687

PR #4638, opened 2020-03-05
Fuzz v24

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds eight fuzz targets

Modifies #4614 with :
- use of `AFLFUZZ_PERSISTANT_MODE` enabled when configure has option `--enable-fuzztargets`

After this, we can 

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -465,6 +465,9 @@
     AM_CONDITIONAL([BUILD_FUZZTARGETS], [test "x$enable_fuzztargets" = "xyes"])
     AS_IF([test "x$enable_fuzztargets" = "xyes"], [
         AC_DEFINE([AFLFUZZ_NO_RANDOM], [1], [Disable all use of random functions])
+        AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[]], [[while (__AFL_LOOP(1000))]])],
+                [AC_DEFINE([AFLFUZZ_PERSISTANT_MODE], [1], [Enable AFL PERSISTANT_MODE])],
+                [])
         AC_PROG_CXX
         AC_LANG_PUSH(C++)
         tmp_saved_flags=$[]_AC_LANG_PREFIX[]FLAGS
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@
     AC_CONFIG_HEADERS([config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
     AC_CONFIG_MACRO_DIR(m4)
-    AM_INIT_AUTOMAKE([tar-ustar])
+    AM_INIT_AUTOMAKE([tar-ustar subdir-objects])
 
     AC_LANG([C])
     AC_PROG_CC_C99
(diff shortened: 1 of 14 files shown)


configure.ac
```diff
@@ -459,9 +459,49 @@
...
+
+    AC_ARG_ENABLE(fuzztargets,
+        AS_HELP_STRING([--enable-fuzztargets], [Enable fuzz targets]),[enable_fuzztargets=$enableval],[enable_fuzztargets=no])
+    AM_CONDITIONAL([BUILD_FUZZTARGETS], [test "x$enable_fuzztargets" = "xyes"])
+    AS_IF([test "x$enable_fuzztargets" = "xyes"], [
+        AC_DEFINE([AFLFUZZ_NO_RANDOM], [1], [Disable all use of random functions])
+        AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[]], [[while (__AFL_LOOP(1000))]])],
+                [AC_DEFINE([AFLFUZZ_PERSISTANT_MODE], [1], [Enable AFL PERSISTANT_MODE])],
+                [])
+        AC_PROG_CXX
```
jasonish: Defining this here results in some ugly output during autoconf:
```
../../lib/autoconf/c.m4:667: AC_LANG_COMPILER(C++) is expanded from...
../../lib/autoconf/lang.m4:329: AC_LANG_COMPILER_REQUIRE is expanded from...
../../lib/autoconf/general.m4:2678: AC_LINK_IFELSE is expanded from...
../../lib/m4sugar/m4sh.m4:639: AS_IF is expanded from...
../../lib/autoconf/general.m4:2031: AC_CACHE_VAL is expanded from...
```
It looks like moving AC_PROG_CXX outside of the AS_IF fixes this. I believe the macro is expanded either way, and won't cause a failure even if no C++ compiler can be found.

catenacyber: But do we want to require a C++ compiler when it is only needed for fuzzing ?

configure.ac
```diff
@@ -459,9 +459,49 @@
 
 # options
 
+
+    AC_ARG_ENABLE(fuzztargets,
+        AS_HELP_STRING([--enable-fuzztargets], [Enable fuzz targets]),[enable_fuzztargets=$enableval],[enable_fuzztargets=no])
+    AM_CONDITIONAL([BUILD_FUZZTARGETS], [test "x$enable_fuzztargets" = "xyes"])
+    AS_IF([test "x$enable_fuzztargets" = "xyes"], [
+        AC_DEFINE([AFLFUZZ_NO_RANDOM], [1], [Disable all use of random functions])
+        AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[]], [[while (__AFL_LOOP(1000))]])],
```
victorjulien: This has no effect as `config.h` is not included in `onefile.c`.

src/tests/fuzz/onefile.c
```diff
@@ -0,0 +1,59 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <stdio.h>
+
```
victorjulien: Add
```c
#include "config.h"
```
For AFL persistent mode to work.

configure.ac
catenacyber: You are right Jason, no failure even if no C++ compiler is found, thanks :-)

src/tests/fuzz/onefile.c
catenacyber: ok thanks

configure.ac
catenacyber: ok thanks

PR #4643, opened 2020-03-06
Detect icmpv6hdr v7

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3323 

Describe changes:
- Fixes some typos
- Adds a generic `DetectU32Data` structure for signature matches on uint32 values
- Adds `icmpv6.mtu` signature  keyword
- Fix endianness for ICMPv6 MTU decoding

This fixes only the first evasion of the redmine ticket

Makes https://github.com/OISF/suricata-verify/pull/187 pass

Modifies #4639 by taking comments into account

There is still some points to improve/discuss :
- Handle `DetectBufferTypeRegister` as fatal error... Should it be part of this PR ?
- Use `pcre_copy_substring` in detect-tcpmss.c... Should it be part of this PR ?

PR #4644, opened 2020-03-06
Detect icmpv6hdr v8

catenacyber:

Describe changes:
- Fixes some typos
- Adds a generic `DetectU32Data` structure for signature matches on uint values
- Adds `icmpv6.mtu` signature  keyword
- Fix endianness for ICMPv6 MTU decoding

This fixes only the first evasion of the redmine ticket

Makes https://github.com/OISF/suricata-verify/pull/187 pass

Modifies #4643 by taking comments into account

There is still some points to improve/discuss :
- Handle `DetectBufferTypeRegister` as fatal error... Should it be part of this PR ?
- Use `pcre_copy_substring` in detect-tcpmss.c... Should it be part of this PR ?


src/detect-util-uint.c
```diff
@@ -0,0 +1,233 @@
+/* Copyright (C) 2020 Open Information Security Foundation
```
victorjulien: I think it'd be better to make the uint in the filename explit: u32. Also, I would like to see this as 'detect-engine'. So `detect-engine-u32.[ch]`

src/detect-util-uint.c
```diff
@@ -0,0 +1,233 @@
...
+            res = pcre_copy_substring((char *) u32str, ov, MAX_SUBSTRINGS, 3, arg3, sizeof(arg3));
+            if (res < 0) {
+                SCLogError(SC_ERR_PCRE_COPY_SUBSTRING, "pcre_copy_substring failed");
+                goto error;
+            }
+            SCLogDebug("Arg3 \"%s\"", arg3);
+        }
+    }
+
+    u32d = SCMalloc(sizeof (DetectU32Data));
```
victorjulien: Suggesting SCCalloc here.

Also, I think its cleaner if this is alloc'd and then filled after all the parsing logic is done.

src/detect-util-uint.c
```diff
@@ -0,0 +1,233 @@
...
+{
+    const DetectU32Data *a = smctx;
+    if (v.u8[0] == a->mode &&
+        v.u32[1] == a->arg1 &&
+        v.u32[2] == a->arg2)
+        return true;
+    return false;
+}
+
+static int g_detect_u32_registered = 0;
```
victorjulien: minor: use type bool?

src/detect-engine-prefilter-common.h
```diff
@@ -94,4 +94,14 @@ PrefilterPacketHeaderExtraMatch(const PrefilterPacketHeaderCtx *ctx,
     return TRUE;
 }
 
+static inline bool PrefilterIsPrefilterableById(const Signature *s, enum DetectKeywordId kid) {
```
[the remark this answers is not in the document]
catenacyber: ok

src/detect-util-uint.c
catenacyber: ok

src/detect-util-uint.c
catenacyber: > I think it'd be better to make the uint in the filename explit: u32

I thought that I would put the similar structures for uint16 in the same file as is done for `ByteExtractStringUint32` in util-byte.h

Do not you think so ?

> Also, I would like to see this as 'detect-engine'

Ok for renaming with engine

src/detect-util-uint.c
catenacyber: Ok, interesting take

PR #4646, opened 2020-03-06
New StringParse* functions

inashivb:
Related to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3053

### DIFF
--- a/src/defrag-hash.c
+++ b/src/defrag-hash.c
@@ -202,7 +202,7 @@ void DefragInitConfig(char quiet)
     }
     if ((ConfGet("defrag.hash-size", &conf_val)) == 1)
     {
-        if (ByteExtractStringUint32(&configval, 10, strlen(conf_val),
+        if (StringParseUint32(&configval, 10, strlen(conf_val),
                                     conf_val) > 0) {
             defrag_config.hash_size = configval;
         } else {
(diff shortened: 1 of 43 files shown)


victorjulien:
In commit `util: Add StringParse* functions`, can you explain what the behaviour of the new function is that why it is important?

jlucovsky:
The new function sig is `StringParseUint32(&configval, 10, strlen(conf_val),  conf_val)`

Would it make sense to avoid the extra parameter (`strlen(x)`) and have `StringParseX` compute that?

victorjulien:
I think a 

> The new function sig is `StringParseUint32(&configval, 10, strlen(conf_val), conf_val)`
> 
> Would it make sense to avoid the extra parameter (`strlen(x)`) and have `StringParseX` compute that?

I supposed that depends on whether the current functionality of being able to parse only part of a string is considered useful. Is it being used anywhere in the current code?

inashivb:
> I think a
> 
> > The new function sig is `StringParseUint32(&configval, 10, strlen(conf_val), conf_val)`
> > Would it make sense to avoid the extra parameter (`strlen(x)`) and have `StringParseX` compute that?
> 
> I supposed that depends on whether the current functionality of being able to parse only part of a string is considered useful. Is it being used anywhere in the current code?

There are parts of code like `if (StringParseUint32(&data->multiplier, 10, strlen(args[i]) - 11,..` so it does not always seem to take strlen of the string being passed to it. Leaving the function signature as-is.

inashivb:
Replaced with # #4739

PR #4648, opened 2020-03-06
doc: Add chassis for dev docs

inashivb:
Setup for Suricata Developer Guide

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3344


doc/devguide/Makefile.am
```diff
@@ -0,0 +1,59 @@
+if HAVE_SURICATA_MAN
+man1_MANS = suricata.1
+endif
```
jasonish: Anything about the man page can be removed from the devguide section.  The PDF stuff can stay though.

doc/devguide/README.md
```diff
@@ -0,0 +1,14 @@
+# Suricata Developer Guide
+
+This directory contains the Suricata Developer Guide. The
+[Sphinx Document Generate](http://sphinx-doc.org) is used to build the
```
jasonish: `Generate` -> `Generator`. I realize this is a typo brought over from the user guide. Should probably fix it there as well (as a separate commit).

doc/devguide/convert.py
```diff
@@ -0,0 +1,94 @@
+#! /usr/bin/env python
```
jasonish: We can drop this file. It exists in the userguide still, but I don't think it useful anymore.

inashivb:
Thanks, Jason! Replaced with #4650

PR #3519, opened 2018-10-22
Af packet exit 4

regit:
Update of #3509 with rebase and comments being addressed

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Less error messages
- More more code in the utility function

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/421
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/203

PR #4656, opened 2020-03-09
ftp: Incorrect ftp_memuse calculation, which will finally cause FTPCh…

vanlink:
ftp_memuse calculation is inaccurate.
ftp.memuse in stats.log will increase continuously.

Bug #3519
Edit: https://redmine.openinfosecfoundation.org/issues/3519

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -262,7 +262,7 @@ static FTPString *FTPStringAlloc(void)
 static void FTPStringFree(FTPString *str)
 {
     if (str->str) {
-        FTPFree(str->str, str->len);
+        FTPFree(str->str, str->mem_len);
     }
 
     FTPFree(str, sizeof(FTPString));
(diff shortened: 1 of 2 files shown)


jlucovsky:
Thanks for your PR and highlighting this issue.

I like your solution and have a suggestion to achieve the same goal with fewer code changes.

1. Eliminate the extra parameter to `CopyCommandLine`
2. Modify `CopyCommandLine` to return the number of bytes allocated
3. Modify `output-ftp-json.c::JsonFTPLogComand()` and change `uint16_t length = response->len` to `uint16_t length = strlen(response->str)`.

This means that the allocated length is returned from `CopyCommandLength` and available in both the TX and `response` containers so the add'l element to track the length can be removed.

vanlink:
> Thanks for your PR and highlighting this issue.
> 
> I like your solution and have a suggestion to achieve the same goal with fewer code changes.
> 
> 1. Eliminate the extra parameter to `CopyCommandLine`
> 2. Modify `CopyCommandLine` to return the number of bytes allocated
> 3. Modify `output-ftp-json.c::JsonFTPLogComand()` and change `uint16_t length = response->len` to `uint16_t length = strlen(response->str)`.
> 
> This means that the allocated length is returned from `CopyCommandLength` and available in both the TX and `response` containers so the add'l element to track the length can be removed.

OK, your suggestion is better, please do it in your way, and close this PR.
Be careful of the trailing newlines/carriage returns in the request/response.

jlucovsky:
Continued in #4844

PR #4659, opened 2020-03-10
Dev/rustify x509 v1

chifflier:

not applicable

https://redmine.openinfosecfoundation.org/issues/3517

Describe changes:
- Use the pure-rust crates `x509-parser` and `der-parser` (already dependencies) for X.509 decoding
- Remove the C DER parser
- Remove the associated fuzzing code

Notes:
- The goal is (obviously) more memory safety, DER is a minefield
- I don't know if I should be happy because this removes ~1800 lines of C, sad because this was my own code, or scared because I wrote so many DER parsers :)
- I have no exact numbers, but I expect this to be a bit (but not horribly) slower than the previous decoder. This is not entirely related to Rust vs C, but more to the fact the new parser is decoding more objects. Not a problem here, but if this becomes a burden I can patch `x509-parser`
- During the patch, I found something weird in the existing `SSLStateFree`:
  - https://github.com/OISF/suricata/blob/master/src/app-layer-ssl.c#L2593-L2594 frees the **server** serial inside a block of code freeing **client** data. The client serial does not seem free()'d . I suppose this was a bug, and fixed it.
- I introduced a function `rs_cstring_free` to free any string allocated by Rust. The rule is the following: everything allocated by Rust should be freed by Rust, which explains the `Box` and `CString` allocations. See http://jakegoulding.com/rust-ffi-omnibus/objects/ for details and examples


victorjulien:
Great work Pierre. @jasonish can you have a look, especially at the cstring logic?

rust/src/common.rs
```diff
@@ -10,3 +13,14 @@ macro_rules! take_until_and_consume (
     }
   );
 );
+
+/// Free a CString allocated by Rust
+#[no_mangle]
+pub extern "C" fn rs_cstring_free(s: *mut c_char) {
```
jasonish: Should we mark the function unsafe? Clippy does complain about this.

rust/src/x509/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+#[no_mangle]
+pub extern "C" fn rs_x509_get_subject(ptr: *const X509) -> *mut c_char {
+    if ptr.is_null() {
+        return std::ptr::null_mut();
+    }
+    let x509 = cast_pointer! {ptr, X509};
+    let subject = format!("{}", x509.0.tbs_certificate.subject);
+    let c_str = CString::new(subject).unwrap();
```
jasonish: Any chance of failure here?

jasonish:
> Great work Pierre. @jasonish can you have a look, especially at the cstring logic?

Looks good.  CString logic is correct as far as I can tell. I've pulled rs_cstring_free into the jsonbuilder branch as that was doing an incorrect free.

rust/src/common.rs
chifflier: I think it is a good idea to follow clippy, so I'll mark the function unsafe

rust/src/x509/mod.rs
chifflier: I think not (the output of `format!` is guaranteed to be valid utf-8), but I also do not like `unwrap`.
I'll rewrite that to use `to_string()` and `match` instead of `unwrap`.

chifflier: Note: This check is in fact very important: `CString::new` can fail if input data contains a null byte, which can happen in a forged certificate.

chifflier:
Replaced by #4661

rust/src/x509/mod.rs
catenacyber: What should happen when we have a certificate subject such as `oisf.net\x00.evil.com` ? Match only on oisf.net or on the whole buffer ?

chifflier: @catenacyber Using #4661 the conversion will fail and a null pointer is returned, then an "invalid certificate" event is raised
The existing C code silently stops at the first \0. I want to fix that, but I propose fixing that in another PR to avoid mixing changes and bugfixes

PR #4661, opened 2020-03-10
Dev/rustify x509 v2

chifflier:
Previous PR: #4659 

Describe changes (to previous PR):
- Fix clippy warnings (mark functions dereferencing input pointers as `unsafe` instead of blocks, add doc-comments about unsafety)
- Add wrapper to create a `CString` and do not `unwrap` (return NULL if `CString::new()` fails)

PR #4663, opened 2020-03-11
Dev/rustify x509 v3

chifflier:
Previous PR: #4661

- use an explicit call to `drop` in functions called to free objects
- fix type conversion when extracting certificate validity (use `i64` on the Rust side as this is the type used in the certificate data, and convert back to `time_t` in the C code)


rust/src/common.rs
```diff
@@ -10,3 +13,29 @@ macro_rules! take_until_and_consume (
...
 );
+
+/// Convert a String to C-compatible string
+///
+/// This function will consume the provided data and use the underlying bytes to construct a new
+/// string, ensuring that there is a trailing 0 byte. This trailing 0 byte will be appended by this
+/// function; the provided data should *not* contain any 0 bytes in it.
+///
+/// Returns a valid pointer, or NULL
+pub fn rust_string_to_c(s: String) -> *mut c_char {
```
victorjulien: can this be in its own commit, with https://github.com/OISF/suricata/pull/4663/commits/9e58350b760c1bf450b7b9cfd6c2922d28953883 squashed into it?

src/app-layer-ssl.c
```diff
@@ -355,121 +353,6 @@ void SSLVersionToString(uint16_t version, char *buffer)
...
-    switch (err) {
-        case ERR_DER_UNKNOWN_ELEMENT:
-            SSLSetEvent(ssl_state,
-                        TLS_DECODER_EVENT_CERTIFICATE_UNKNOWN_ELEMENT);
-            break;
-        case ERR_DER_ELEMENT_SIZE_TOO_BIG:
-        case ERR_DER_INVALID_SIZE:
-        case ERR_DER_RECURSION_LIMIT:
-            SSLSetEvent(ssl_state,
-                        TLS_DECODER_EVENT_CERTIFICATE_INVALID_LENGTH);
```
victorjulien: in my failing test this event is missing. Are we setting events based on the new decoder?

victorjulien:
Can you squash https://github.com/OISF/suricata/pull/4663/commits/d22e0e8254cf1af1e49dd1711c2348dd2a46045e into the commit that is fixed by this?

rust/src/common.rs
chifflier: Done

src/app-layer-ssl.c
chifflier: At the moment, no. The decoder can raise errors (See list at https://docs.rs/x509-parser/0.6.5/x509_parser/error/enum.X509Error.html, and https://docs.rs/der-parser/3.0.4/der_parser/error/enum.BerError.html for the possible structure decoding errors), but at the moment they are all transformed into a `TLS_DECODER_EVENT_INVALID_CERTIFICATE` event.

It's not wrong, but maybe lacks some precision in the error. Are there signatures or detection rules that really requires this ? If yes, I can try to find a way to send more events back to C

victorjulien: Looking at the precision provided by the Rust parsers, I would actually like to request exposing all of them. I think it would be great if we can match on those and log those in the anomaly logger.

chifflier:
Continued in #4719

PR #4665, opened 2020-03-11
DCERPC parser in Rust

inashivb:


inashivb:
Some questions:
- Other parsers do not have separate State structs for UDP and TCP, I felt like they're different so I made them separate, is this incorrect? This however is making logging difficult since the state object passed from C is `void *` and could be either the TCP or UDP state.

(TO be edited, more questions coming)

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,1794 @@
...
+pub const DCERPC_TYPE_BIND: u8 = 11;
+pub const DCERPC_TYPE_BINDACK: u8 = 12;
+pub const DCERPC_TYPE_BINDNAK: u8 = 13;
+pub const DCERPC_TYPE_ALTER_CONTEXT: u8 = 14;
+pub const DCERPC_TYPE_ALTER_CONTEXT_RESP: u8 = 15;
+pub const DCERPC_TYPE_AUTH3: u8 = 16;
+pub const DCERPC_TYPE_SHUTDOWN: u8 = 17;
+pub const DCERPC_TYPE_CO_CANCEL: u8 = 18;
+pub const DCERPC_TYPE_ORPHANED: u8 = 19;
+pub const DCERPC_TYPE_RTS: u8 = 20;
```
inashivb: There's a duplicate constants error while compiling as these constants exist both in smb and dcerpc. I tried importing these into smb but it had some issues and I'm not sure if its the right way.
What would be the right way to share variables? A common file/dir or something?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,1794 @@
...
+                    None => 0,
+                };
+                if pfcflags & PFC_FIRST_FRAG > 0 {
+                    uuidentry.flags |= DCERPC_UUID_ENTRY_FLAG_FF;
+                }
+                if let Some(ref mut bind) = self.dcerpcbind {
+                    bind.uuid_list.push(uuidentry);
+                    //println!("parse_bindctxitem: bind struct: {:?}\n", bind);
+                }
+                44
```
inashivb: Note to self: Remove hardcoded 44.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,1794 @@
...
+            self.bytes_consumed += parsed as u16;
+        }
+
+        let fraglen = match self.get_hdr_fraglen() {
+            Some(x) => x,
+            None => 0,
+        };
+        if buffer.len() as u16 != fraglen {
+            println!("Possibly fragmented data, waiting for more..");
+            return 1;
```
inashivb: Fragmentation question: Is this the right way to handle fragmented data?
I figured there could be two ways to buffer the data:
1. when `Err::Incomplete` block is hit in the parsers
2. before any of the parsing begins (except header as we'd need to extract fraglen information from there)

I chose 2 and am waiting for all the data to come in and checking the buffer length against hdr.fraglen. Is this incorrect? What should be the right way to handle this?

Note to self: Add buffering logic to `Err::Incomplete` block of header.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,1794 @@
...
+                if header.rpc_vers != 4 {
+                    SCLogDebug!("DCERPC UDP Header did not validate.");
+                    return -1;
+                }
+                let mut uuidentry = DCERPCUuidEntry::new();
+                let auuid = header.activityuuid.to_vec();
+                uuidentry.uuid = auuid;
+                self.uuid_list.push(uuidentry);
+                self.dcerpchdrudp = Some(header);
+                80
```
inashivb: Note to self: Remove hardcoded retval

rust/src/dcerpc/parser.rs
```diff
@@ -0,0 +1,373 @@
...
+                   flags1: flags1,
+                   flags2: flags2,
+                   drep: drep.to_vec(),
+                   serial_hi: serial_hi,
+                   objectuuid: match parse_uuid(objectuuid) {
+                       Ok((_, vect)) => assemble_uuid(vect),
+                       Err(e) => {
+                           println!("{}", e);
+                           vec![0]
+                       },
```
inashivb: What could be a better way of doing this in Rust? @jasonish ?

rust/src/dcerpc/dcerpc.rs
victorjulien: You can use them from smb by doing something like `use crate::smb::...` I think? (or the reverse, update smb to use this here)

rust/src/dcerpc/dcerpc.rs
victorjulien: I think you should extend `self.buffer_tc` here with `buffer`.

However, I'm getting close to PoC to allow you to offload this task to the underlying API. So maybe wait a bit.

rust/src/dcerpc/parser.rs
jasonish: This is OK, but might be better, if you expect this to fail now and again to have the objectuuid an Option..  That way you can test later if it exists or not, more idiomatically and raise an alert of whatever.

inashivb: Makes sense. Thanks.

rust/src/dcerpc/dcerpc.rs
victorjulien: See #4671

victorjulien: This is now in `master`

inashivb:
Replaced with #4769

PR #4666, opened 2020-03-11
(RFC) rust: mark extern "C" (ffi) functions as unsafe

jasonish:
Currently the macros build_slice and cast_pointer wrap their code in unsafe not
requiring the containing function to be unsafe, but clippy recommends that any
public function working with raw pointers be marked unsafe. To comply with this
idiom, remove unsafe from the macros, requiring any function that uses these
macros to be unsafe itself.

Reference: 
https://rust-lang.github.io/rust-clippy/master/index.html#not_unsaf

The next step would be to find the ffi funcations that use an unsafe block
and mark them unsafe, removing the unsafe blocks.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/448
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/805

### DIFF
--- a/rust/src/parser.rs
+++ b/rust/src/parser.rs
@@ -116,7 +116,7 @@ pub struct RustParser {
 /// UNSAFE !
 #[macro_export]
 macro_rules! build_slice {
-    ($buf:ident, $len:expr) => ( unsafe{ std::slice::from_raw_parts($buf, $len) } );
+    ($buf:ident, $len:expr) => ( { std::slice::from_raw_parts($buf, $len) } );
 }
 
 /// Cast pointer to a variable, as a mutable reference to an object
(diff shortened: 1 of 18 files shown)


jasonish:
CC: @chifflier @dbcfd

victorjulien:
This seems a bit counter intuitive to me. Should it lead to the functions just be as minimal as possible to pass the objects created from the raw pointers into 'safe' functions? Ultimately we pass around data based on unsafe input, so it seems a bit strange to classify more code as unsafe while fundamentally nothing changes. All rust needs to know that we vouch for the validity of the raw pointers, right?

victorjulien:
Btw I don't object to refactoring to make code checkers happy, but I would like to see if we can avoid having *more* of the logic inside `unsafe` blocks.

chifflier:
> This seems a bit counter intuitive to me. Should it lead to the functions just be as minimal as possible to pass the objects created from the raw pointers into 'safe' functions?

Yes, that's the usual recommendation for FFI code: write `unsafe` wrappers (mostly to wrap pointers and cast types), and call safe (classic) functions as soon as possible.

> Ultimately we pass around data based on unsafe input, so it seems a bit strange to classify more code as unsafe while fundamentally nothing changes. All rust needs to know that we vouch for the validity of the raw pointers, right?

There are usually two kinds of`unsafe`:
- code that must access objects outside Rust's memory checks: syscalls, calling C, wrapping arguments
- voodoo code that casts objects or performs lifetime changes etc.

The former is quite expected at FFI borders, the latter should be avoided unless really justified.

The idiomatic Rust way is what you and @ish described (and I think in fact everyone agrees): FFI functions are usually marked `unsafe` (if they dereference pointers in particular), and must perform as little actions as possible (immediately call a Rust function, and only handle wrapping to/from C).

So, my suggestion would be to indeed mark `unsafe` these functions, but in the same time to ensure they perform only the `unsafe` part, and call safe code as soon as possible. This may require splitting functions, but should not be bad since Rust tends to try inlining functions as much as possible.

dbcfd:
I agree with Pierre's recommendation and approach.

jasonish:
So I think its a bit more than making code checkers safe, but keeping up with generally accepted practices, and this is one lint I'd like to get rid of. I've also felt that when seeing these unsafe blocks, you really have to consider the code after them as well as being unsafe. So I think moving to a practice where the ffi functions are unsafe, but do as little as possible makes sense for us.

victorjulien:
> So I think its a bit more than making code checkers safe, but keeping up with generally accepted practices, and this is one lint I'd like to get rid of. I've also felt that when seeing these unsafe blocks, you really have to consider the code after them as well as being unsafe. So I think moving to a practice where the ffi functions are unsafe, but do as little as possible makes sense for us.

But isn't the consequence of this that we should consider *all* our Rust code unsafe? It's all about dealing with data from unsafe origin.

jasonish:
> > So I think its a bit more than making code checkers safe, but keeping up with generally accepted practices, and this is one lint I'd like to get rid of. I've also felt that when seeing these unsafe blocks, you really have to consider the code after them as well as being unsafe. So I think moving to a practice where the ffi functions are unsafe, but do as little as possible makes sense for us.
> 
> But isn't the consequence of this that we should consider _all_ our Rust code unsafe? It's all about dealing with data from unsafe origin.

hats somewhat true, and is why you should consider any extern "C"
function unsafe, as the caller (our C) needs to ensure its passing valid
data as well.

dbcfd:
> > So I think its a bit more than making code checkers safe, but keeping up with generally accepted practices, and this is one lint I'd like to get rid of. I've also felt that when seeing these unsafe blocks, you really have to consider the code after them as well as being unsafe. So I think moving to a practice where the ffi functions are unsafe, but do as little as possible makes sense for us.
> 
> But isn't the consequence of this that we should consider _all_ our Rust code unsafe? It's all about dealing with data from unsafe origin.

Oh, I see now more what you're saying. `extern "C"` functions called from C code should not be marked unsafe, as they have to be marked in this fashion for interop. We should not be calling `extern "C"` functions from Rust code that have been defined in Rust code. We should be exposing a normal rust interface that is invoked, and the `extern "C"` version should only be invoked by C code.

chifflier:
> > So I think its a bit more than making code checkers safe, but keeping up with generally accepted practices, and this is one lint I'd like to get rid of. I've also felt that when seeing these unsafe blocks, you really have to consider the code after them as well as being unsafe. So I think moving to a practice where the ffi functions are unsafe, but do as little as possible makes sense for us.
> 
> But isn't the consequence of this that we should consider _all_ our Rust code unsafe? It's all about dealing with data from unsafe origin.

I would more explain that as "if the input arguments of the Rust code are trusted/verified (resp. untrusted), then the Rust code is safe (resp. unsafe)".

One cannot expect miracles from Rust: for ex, if you pass a buffer and a wrong (too short) length as arguments, then the built slice will trigger something wrong inside Rust safe code.

rust/src/parser.rs
```diff
@@ -116,45 +116,45 @@ pub struct RustParser {
...
                                        data: *const c_void,
                                        flags: u8) -> i32;
-pub type ProbeFn      = extern "C" fn (flow: *const Flow,direction: u8,input:*const u8, input_len: u32, rdir: *mut u8) -> AppProto;
+pub type ProbeFn      = unsafe extern "C" fn (flow: *const Flow,direction: u8,input:*const u8, input_len: u32, rdir: *mut u8) -> AppProto;
 pub type StateAllocFn = extern "C" fn () -> *mut c_void;
 pub type StateFreeFn  = extern "C" fn (*mut c_void);
-pub type StateTxFreeFn  = extern "C" fn (*mut c_void, u64);
-pub type StateGetTxFn            = extern "C" fn (*mut c_void, u64) -> *mut c_void;
-pub type StateGetTxCntFn         = extern "C" fn (*mut c_void) -> u64;
+pub type StateTxFreeFn  = unsafe extern "C" fn (*mut c_void, u64);
```
dbcfd: I think the linter is specifically complaining about these `extern "C"` functions.

rust/src/rdp/log.rs
```diff
@@ -25,7 +25,7 @@ use std;
 use x509_parser::parse_x509_der;
 
 #[no_mangle]
-pub extern "C" fn rs_rdp_to_json(tx: *mut std::os::raw::c_void) -> *mut JsonT {
+pub unsafe extern "C" fn rs_rdp_to_json(tx: *mut std::os::raw::c_void) -> *mut JsonT {
```
dbcfd: The linter is probably not complaining about these, since they're exposing an interface for C to call.

chifflier:
> 
> `extern "C"` functions called from C code should not be marked unsafe, as they have to be marked in this fashion for interop.

This is indeed not required, but the good practice (enforced by clippy) is to add an `unsafe` keyword if it dereferences an input pointer (because it cannot guarantee safety).

> We should not be calling `extern "C"` functions from Rust code that have been defined in Rust code. We should be exposing a normal rust interface that is invoked, and the `extern "C"` version should only be invoked by C code.

I agree with that

rust/src/parser.rs
jasonish: There is no complaint here, these types just had to be updated after the offending functions were updated.

rust/src/rdp/log.rs
jasonish: The linter was complaining about this function as it dereferences a raw pointer, and suggest the function should be unsafe, not just block.

victorjulien:
> > We should not be calling `extern "C"` functions from Rust code that have been defined in Rust code. We should be exposing a normal rust interface that is invoked, and the `extern "C"` version should only be invoked by C code.
> 
> I agree with that

Can we enforce this somehow?

chifflier:
I think I wasn't clear at some point: `extern "C"` functions are not required to be unsafe. Imagine a function adding two integers, it would be `extern "C"` but not unsafe.

The `unsafe` keyword has the usual meaning: this function is doing something that *could* be dangerous. The only difference is that, while the usual approach is to have an `unsafe` block, there is an exception for the specific case when you dereference a pointer. In that case, `clippy` suggest to mark the whole function as unsafe (not the block), so it appears in the documentation.

Marking a block or a function as `unsafe` is always possible, it's just the usual rules on when to choose which approach to use.

dbcfd:
> > `extern "C"` functions called from C code should not be marked unsafe, as they have to be marked in this fashion for interop.
> 
> This is indeed not required, but the good practice (enforced by clippy) is to add an `unsafe` keyword if it dereferences an input pointer (because it cannot guarantee safety).

The keyword is definitely nice for identifying code which has to deal with raw input, but I don't have strong opinions about whether the keyword is applied at the function level, or at the code level, since it is more of an indication to the rust developer. Both methods still require examining the code to make sure it is handling the raw input appropriately.

Definitely would like to see less logic in functions dealing with raw input, and more in a safe function that is called from the `extern "C"` function.

dbcfd:
> In that case, `clippy` suggest to mark the whole function as unsafe (not the block), so it appears in the documentation.

Ah. In that case, all of our rust functions exposed to C would likely be unsafe. Which is fine, especially if it forces us to split out safe code from unsafe code, so the exposed interfaces only deal with the untrusted input, and the remainder of the logic is in safe functions called from the exposed interface.

liurunhao:
Excuse me?How to make suricata for linux (executable program) compatible kernel version2.0+? Tell me ,please ! Thank you !

PR #4636, opened 2020-03-05
Ssh rust v5

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3445

Describe changes:
- Translates SSH C Parser to a Rust parser

This begins to look good

This is meant to be an exact translation. We want to add features later.
https://github.com/rusticata/ssh-parser might help for this

One difference though is the handling of end of line(s) at the end of the banner. This Rust version should be robust against packet splitting between CR and LF (which should be accepted by openssh), when the current version is vulnerable because end of lines skipping https://github.com/OISF/suricata/blob/master/src/app-layer-ssh.c#L235 is only done if the first packet contains the whole banner with end of lines. I made a suricata-verify test for this https://github.com/OISF/suricata-verify/pull/176

Modifies #4584 
- rebase on master with nom v5
- commits squashed together

PR #4667, opened 2020-03-12
Ssh rust v6

catenacyber:

Describe changes:

This looks good

Modifies #4636 with style issues

Do we have to do something else than code for SSH app layer event rules Signature IDs ?

### DIFF
--- a/rules/ssh-events.rules
+++ b/rules/ssh-events.rules
@@ -1,0 +1,10 @@
+# SSH app layer event rules
+#
+# SID's fall in the 2228000+ range. See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/AppLayer
+#
+# These sigs fire at most once per connection.
+#
+
+alert ssh any any -> any any (msg:"SURICATA SSH invalid banner"; flow:established; app-layer-event:ssh.invalid_banner; classtype:protocol-command-decode; sid:2228000; rev:1;)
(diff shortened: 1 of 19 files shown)


src/output-json-ssh.h
```diff
@@ -28,6 +28,6 @@ void JsonSshLogRegister(void);
 
 #include "app-layer-ssh.h"
 
-void JsonSshLogJSON(json_t *js, SshState *tx);
+void JsonSshLogJSON(json_t *js, void *tx);
```
catenacyber: Should I not remove this function ? How come CI passed when it seems still used by alert-prelude.c ?

rules/ssh-events.rules
```diff
@@ -0,0 +1,10 @@
+# SSH app layer event rules
+#
+# SID's fall in the 2228000+ range. See https://redmine.openinfosecfoundation.org/projects/suricata/wiki/AppLayer
```
victorjulien: can you update this wiki page to match this?

catenacyber: Done

src/output-json-ssh.h
jasonish: I don't think we enable prelude at compile time for any of the CI tests.  Probably should pick at least one build and turn it on.

catenacyber: And compiler/linker do not complain if we declare functions but do not define them...

PR #4670, opened 2020-03-12
Ssh rust v8

catenacyber:

Describe changes:

This looks good

Modifies #4667 with alert-prelude.c fix

### DIFF
--- a/rust/src/ssh/ssh.rs
+++ b/rust/src/ssh/ssh.rs
@@ -1,0 +1,587 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use super::parser;
+use crate::applayer::{LoggerFlags, TxDetectFlags};
+use crate::core::STREAM_TOSERVER;
+use crate::core::{self, AppProto, Flow, ALPROTO_UNKNOWN, IPPROTO_TCP};
+use crate::log::*;
+use crate::parser::*;
+use std;
+use std::ffi::{CStr, CString};
+use std::mem::transmute;
+
+static mut ALPROTO_SSH: AppProto = ALPROTO_UNKNOWN;
+
+#[repr(u32)]
+pub enum SSHEvent {
+    InvalidBanner = 0,
+    LongBanner,
+    InvalidRecord,
+}
+
+impl SSHEvent {
+    fn from_i32(value: i32) -> Option<SSHEvent> {
+        match value {
+            0 => Some(SSHEvent::InvalidBanner),
+            1 => Some(SSHEvent::LongBanner),
+            2 => Some(SSHEvent::InvalidRecord),
+            _ => None,
+        }
+    }
+}
+
+#[repr(u8)]
+#[derive(Copy, Clone, PartialOrd, PartialEq)]
+pub enum SSHConnectionState {
+    SshStateInProgress = 0,
+    SshStateBannerDone = 1,
+    SshStateFinished = 2,
+}
+
+const SSH_MAX_BANNER_LEN: usize = 256;
+const SSH_RECORD_HEADER_LEN: usize = 6;
+//TODO complete enum and parse messages contents
+const SSH_MSG_NEWKEYS: u8 = 21;
+
+pub struct SshHeader {
+    record_left: u32,
+    record_buf: Vec<u8>,
+    flags: SSHConnectionState,
+    banner: Vec<u8>,
+    //can we have these be references to banner, with start offset and size
+    pub protover: Vec<u8>,
+    pub swver: Vec<u8>,
+}
+
+impl SshHeader {
+    pub fn new() -> SshHeader {
+        SshHeader {
+            record_left: 0,
+            record_buf: Vec::with_capacity(SSH_RECORD_HEADER_LEN),
+            flags: SSHConnectionState::SshStateInProgress,
+            banner: Vec::with_capacity(SSH_MAX_BANNER_LEN),
+            protover: Vec::new(),
+            swver: Vec::new(),
+        }
+    }
+}
+
+pub struct SSHTransaction {
+    pub srv_hdr: SshHeader,
+    pub cli_hdr: SshHeader,
+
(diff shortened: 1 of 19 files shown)


rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,587 @@
...
+    }
+
+    fn set_event(&mut self, event: SSHEvent) {
+        let ev = event as u8;
+        core::sc_app_layer_decoder_events_set_event_raw(&mut self.transaction.events, ev);
+    }
+
+    fn parse_record(&mut self, mut input: &[u8], resp: bool) -> bool {
+        //TODO add SCLogDebug ?
+        let mut hdr = &mut self.transaction.cli_hdr;
```
victorjulien: Could be written as:
```rust
let mut hdr = if !resp {
    &mut self.transaction.cli_hdr
} else {
    &mut self.transaction.srv_hdr
}
```

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,587 @@
...
+        if resp {
+            hdr = &mut self.transaction.srv_hdr;
+        }
+        //first skip record left bytes
+        if hdr.record_left > 0 {
```
victorjulien: I'm finding this buffering logic quite hard to follow. Could we somehow simplify it?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,587 @@
...
+                vec.extend(input);
+                iline = &vec[..];
+            }
+        }
+        match parser::ssh_parse_line(iline) {
+            Ok((rem, line)) => {
+                let mut set_event_long = false;
+                if hdr.banner.len() + line.len() <= SSH_MAX_BANNER_LEN {
+                    hdr.banner.extend(line);
+                } else if hdr.banner.len() < SSH_MAX_BANNER_LEN {
```
victorjulien: In what case would a normal `} else {` not be sufficient here?

src/alert-prelude.c
```diff
@@ -724,25 +724,24 @@ static void PacketToDataProtoTLS(const Packet *p, const PacketAlert *pa, idmef_a
...
     json_t *js, *s_js;
-    SshState *ssh_state = (SshState *)FlowGetAppState(p->flow);
+    void *ssh_state = FlowGetAppState(p->flow);
 
     if (ssh_state == NULL)
         return;
 
-    js = json_object();
-    if (js == NULL)
+    void *tx_ptr = rs_ssh_state_get_tx(ssh_state, 0);
```
victorjulien: should we test for failure here?

src/alert-prelude.c
```diff
@@ -724,25 +724,24 @@ static void PacketToDataProtoTLS(const Packet *p, const PacketAlert *pa, idmef_a
...
-
     s_js = json_object_get(js, "server");
     if (s_js != NULL) {
         JsonToAdditionalData(NULL, s_js, alert);
     }
 
     s_js = json_object_get(js, "client");
     if (s_js != NULL) {
-        JsonToAdditionalData(NULL, s_js, alert);
+            JsonToAdditionalData(NULL, s_js, alert);
```
victorjulien: indent?

src/util-lua-ssh.c
```diff
@@ -95,13 +99,17 @@ static int GetServerSoftwareVersion(lua_State *luastate, const Flow *f)
...
 
-    SshState *ssh_state = (SshState *)state;
+    const uint8_t *software = NULL;
+    uint32_t b_len = 0;
 
-    if (ssh_state->srv_hdr.software_version == NULL)
-        return LuaCallbackError(luastate, "error: no server software version");
+    void *tx = rs_ssh_state_get_tx(state, 0);
+    if (rs_ssh_tx_get_software(tx, &software, &b_len, STREAM_TOCLIENT) != 1)
+        return LuaCallbackError(luastate, "error: no server proto version");
```
victorjulien: software version

src/util-lua-ssh.c
```diff
@@ -157,13 +169,17 @@ static int GetClientSoftwareVersion(lua_State *luastate, const Flow *f)
...
 
-    SshState *ssh_state = (SshState *)state;
+    const uint8_t *software = NULL;
+    uint32_t b_len = 0;
 
-    if (ssh_state->cli_hdr.software_version == NULL)
-        return LuaCallbackError(luastate, "error: no client software version");
+    void *tx = rs_ssh_state_get_tx(state, 0);
+    if (rs_ssh_tx_get_software(tx, &software, &b_len, STREAM_TOSERVER) != 1)
+        return LuaCallbackError(luastate, "error: no server proto version");
```
victorjulien: client software

rust/src/ssh/ssh.rs
catenacyber: I would love to but did not find a Rust way to do it (I arrived to the same code for HTTP2)

Problem comes from the fact that the buffer belongs to the `SSHState` structure
And borrow checker complains if we use the structure in another way when we use/read/reset the buffer

Do you have some pointer to a better code ?

rust/src/ssh/ssh.rs
catenacyber: ok, thanks for the Rust idiomatics

rust/src/ssh/ssh.rs
catenacyber: I just want to set the event once, and we do not need to `extend`  if the buffer is already full

src/alert-prelude.c
catenacyber: Well, for SSH, if there is a not NULL state, there is one and only one transaction (see implementation of `rs_ssh_state_get_tx`).

In the C parser, there was only one structure for both state and transaction.

So, should we `BUG_ON` to have no performance cost on production and enforce this condition during tests ?

rust/src/ssh/ssh.rs
victorjulien: I think the DNS code takes a simpler approach. But, do we still need this after the 'incomplete' logic of #4677 is merged?

src/alert-prelude.c
catenacyber: Ok, thanks

src/util-lua-ssh.c
catenacyber: ok thanks

src/util-lua-ssh.c
catenacyber: ok thanks

src/alert-prelude.c
victorjulien: `BUG_ON` sounds good to me.

catenacyber: ok for `BUG_ON` then

rust/src/ssh/ssh.rs
catenacyber: #4677 will definitely help. So I can rebase when it gets merged

catenacyber: As for DNS, the simpler code has a performance cost of slice allocation + memory copy for each call to the parser cf https://github.com/OISF/suricata/blob/master/rust/src/dns/dns.rs#L523

Can we avoid this tradeoff and have the best of both worlds ?

victorjulien: This cost seems small, but lets just wait for #4677 so we can remove the buffering both here and from DNS.

catenacyber: ok

PR #4671, opened 2020-03-12
Feat/appl records/v16

victorjulien:
Initial support for pushing TCP buffering down to app-layer API.

Implementation for SMB to show how it is used.

cc: @chifflier @dbcfd

### DIFF
--- a/rust/src/applayertemplate/template.rs
+++ b/rust/src/applayertemplate/template.rs
@@ -302,7 +302,7 @@ pub extern "C" fn rs_template_parse_request(
     input_len: u32,
     _data: *const std::os::raw::c_void,
     _flags: u8,
-) -> i32 {
+) -> AppLayerReturn {
     let eof = unsafe {
         if AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF) > 0 {
             true
@@ -317,10 +317,10 @@ pub extern "C" fn rs_template_parse_request(
 
     let state = cast_pointer!(state, TemplateState);
     let buf = build_slice!(input, input_len as usize);
-    if state.parse_request(buf) {
-        return 1;
+    if !state.parse_request(buf) {
+        return AppLayerReturn::err();
     }
-    return -1;
+    AppLayerReturn::ok()
 }
 
 #[no_mangle]
@@ -332,7 +332,7 @@ pub extern "C" fn rs_template_parse_response(
     input_len: u32,
     _data: *const std::os::raw::c_void,
     _flags: u8,
-) -> i32 {
+) -> AppLayerReturn {
     let _eof = unsafe {
         if AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF) > 0 {
             true
@@ -342,10 +342,10 @@ pub extern "C" fn rs_template_parse_response(
     };
     let state = cast_pointer!(state, TemplateState);
     let buf = build_slice!(input, input_len as usize);
-    if state.parse_response(buf) {
-        return 1;
+    if !state.parse_response(buf) {
+        return AppLayerReturn::err();
     }
-    return -1;
+    AppLayerReturn::ok()
 }
 
 #[no_mangle]
--- a/rust/src/parser.rs
+++ b/rust/src/parser.rs
@@ -48,6 +48,13 @@ impl AppLayerReturn {
             needed: 0,
         };
     }
+    pub fn incomplete(consumed: u32, needed: u32) -> AppLayerReturn {
+        return AppLayerReturn {
+            status: 1,
+            consumed: consumed,
+            needed: needed,
+        };
+    }
 }
 
 /// Rust parser declaration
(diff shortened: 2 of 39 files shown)


rust/src/parser.rs
```diff
@@ -26,6 +26,37 @@ use crate::applayer;
...
+        };
+    }
+    pub fn incomplete(consumed: u32, needed: u32) -> AppLayerReturn {
+        return AppLayerReturn {
+            status: 1,
+            consumed: consumed,
+            needed: needed,
+        };
+    }
+}
```
jasonish: So ok() means we consumed the exact number of bytes provided? It would essentially be the same as incomplete where consumed = input_len, and needed is 0?

src/app-layer-parser.c
```diff
@@ -1234,12 +1238,35 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow
...
-                flags) < 0)
+                flags);
+        if (res.status < 0)
         {
             goto error;
+        } else if (res.status > 0) {
+            TcpSession *ssn = f->protoctx;
+            if (ssn != NULL) {
+                SCLogDebug("direction %d/%s", direction, (flags & STREAM_TOSERVER) ? "toserver":"toclient");
+                assert(res.consumed <= input_len);
```
jasonish: BUG_ON?

src/stream-tcp.c
```diff
@@ -5849,6 +5849,21 @@ static inline int StreamTcpValidateAck(TcpSession *ssn, TcpStream *stream, Packe
     SCReturnInt(-1);
 }
 
+/** \brief update reassembly progress
+
+ * \param ssn TCP Session
+ * \param direction direction to set the flag in: 0 toserver, 1 toclient
+ */
+void StreamTcpUpdateAppLayerProgress (TcpSession *ssn, char direction,
+        const uint32_t progress)
```
jasonish: Style `StreamTcpUpdateAppLayerProgress (`

src/stream-tcp.h
```diff
@@ -192,6 +192,8 @@ int StreamTcpInlineDropInvalid(void);
 int StreamTcpInlineMode(void);
 
 int TcpSessionPacketSsnReuse(const Packet *p, const Flow *f, const void *tcp_ssn);
+void StreamTcpUpdateAppLayerProgress (TcpSession *ssn, char direction,
+        const uint32_t progress);
```
jasonish: And here.

src/app-layer-parser.c
victorjulien: Yeah. This should probably become a DEBUG_VALIDATE_BUG_ON later, but will leave it as a BUG_ON in the dev tree for a while

rust/src/parser.rs
victorjulien: Essentially, yes.

src/stream-tcp.c
victorjulien: Missing the issue.

jasonish: Whitespace between function name and arguments.

rust/src/parser.rs
jasonish: This might come up from others, so I'll bring it up here. I'm starting to like removing the `return` on simple methods like these.  But I am proposing to turn that lint off anyways, so it can state.

Also, this looks like a good candidate to be named AppLayerResult, but we don't quite follow the result pattern here, so I'd leave as is, maybe revisit separately later on to see if it can map to the Result pattern.

rust/src/parser.rs
```diff
@@ -26,6 +26,37 @@ use crate::applayer;
 use std::os::raw::{c_void,c_char,c_int};
 use crate::applayer::{AppLayerGetTxIterTuple};
 
+#[repr(C)]
+pub struct AppLayerReturn {
```
victorjulien: @jasonish any thoughts on naming here? Maybe `AppLayerResult` would be more consistent with `OutputInitResult` we use already in other places.

jasonish: Yes, I think Result is a better name.  It just doesn't map the well used Result types in Rust which is like Result<TheReturnValue, TheError>... But I don't think I'd worry about that too much.

victorjulien: Ok will change.

victorjulien:
Comments addressed in #4672

rust/src/applayertemplate/template.rs
```diff
@@ -342,10 +342,10 @@ pub extern "C" fn rs_template_parse_response(
     };
     let state = cast_pointer!(state, TemplateState);
     let buf = build_slice!(input, input_len as usize);
-    if state.parse_response(buf) {
-        return 1;
+    if !state.parse_response(buf) {
```
dbcfd: Seems like it might be useful to do 

```
impl From<i32> for AppLayerReturn {
    fn from(v: i32) -> Self {
        if v < 0 {
            Self::err()
        } else {
            Self::ok()
        }
    }
}```

so that most of these just become `state.parse_response(buf).into()`

victorjulien: Can you check my attempt here https://github.com/OISF/suricata/pull/4677?

PR #4672, opened 2020-03-12
Feat/appl records/v19

victorjulien:
Initial support for pushing TCP buffering down to app-layer API.

Implementation for SMB to show how it is used.

Continuation #4671 with:
- removal of `abort()`
- rename `AppLayerReturn` to `AppLayerResult`
- minor style fixes

Passed

Ticket: https://redmine.openinfosecfoundation.org/issues/3444

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -603,7 +603,7 @@ impl NFSState {
             2 => {
                 self.process_request_record_v2(r)
             },
-            _ => { 1 },
+            _ => { 0 },
         }
     }
 
@@ -1021,7 +1021,7 @@ impl NFSState {
         let consumed = self.filetracker_update(STREAM_TOSERVER, &gap, gap_size);
         if consumed > gap_size {
             SCLogDebug!("consumed more than GAP size: {} > {}", consumed, gap_size);
-            return 1;
+            return 0;
         }
         self.ts_ssn_gap = true;
         self.ts_gap = true;
@@ -1038,7 +1038,7 @@ impl NFSState {
         let consumed = self.filetracker_update(STREAM_TOCLIENT, &gap, gap_size);
         if consumed > gap_size {
             SCLogDebug!("consumed more than GAP size: {} > {}", consumed, gap_size);
-            return 1;
+            return 0;
         }
         self.tc_ssn_gap = true;
         self.tc_gap = true;
@@ -1061,7 +1061,7 @@ impl NFSState {
                 if self.tcp_buffer_ts.len() + i.len() > 1000000 {
                     SCLogDebug!("parse_tcp_data_ts: TS buffer exploded {} {}",
                             self.tcp_buffer_ts.len(), i.len());
-                    return 1;
+                    return 0;
                 };
                 v.extend_from_slice(i);
                 v.as_slice()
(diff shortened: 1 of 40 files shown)


rust/src/nfs/nfs.rs
```diff
@@ -1151,7 +1151,7 @@ impl NFSState {
                                     Err(nom::Err::Failure(_e)) => {
                                         self.set_event(NFSEvent::MalformedData);
                                         SCLogDebug!("Parsing failed: {:?}", _e);
-                                        return 1;
+                                        return 0;
```
victorjulien: Need to double check this logic.

rust/src/nfs/nfs.rs
```diff
@@ -1392,6 +1392,7 @@ impl NFSState {
                 Err(nom::Err::Failure(_e)) => { SCLogDebug!("Parsing failed: {:?}", _e); }
             }
         }
+        SCLogNotice!("status {}", status);
```
victorjulien: TODO remove

rust/src/parser.rs
```diff
@@ -26,6 +26,37 @@ use crate::applayer;
...
 
+#[repr(C)]
+pub struct AppLayerResult {
+    pub status: i32,
+    pub consumed: u32,
+    pub needed: u32,
+}
+
+impl AppLayerResult {
+    pub fn ok() -> AppLayerResult {
```
victorjulien: TODO comment to explain

rust/src/parser.rs
```diff
@@ -26,6 +26,37 @@ use crate::applayer;
 use std::os::raw::{c_void,c_char,c_int};
 use crate::applayer::{AppLayerGetTxIterTuple};
 
+#[repr(C)]
+pub struct AppLayerResult {
```
victorjulien: @jasonish thoughts about moving this to `applayer.rs` instead?

src/app-layer-parser.h
```diff
@@ -51,6 +51,10 @@
  *  completely inspected */
 #define APP_LAYER_TX_PREFILTER_MASK             ~APP_LAYER_TX_INSPECTED_FLAG
 
+#define APP_LAYER_OK (AppLayerResult) { 0, 0, 0 }
```
victorjulien: TODO comments to explain values, esp for incomplete

catenacyber:
This looks good :-)

Could we have also something like "require data until next end of line" ?
Here we use a fixed numeric value which is fine for binary protocols.
But for text protocols (like FTP), the thing we wait for is a special character...

Or could it be a further improvement ?

rust/src/nfs/nfs.rs
victorjulien: This was wrong, reverted it but also did additional nfs work.

victorjulien:
> Could we have also something like "require data until next end of line" ?
> Here we use a fixed numeric value which is fine for binary protocols.
> But for text protocols (like FTP), the thing we wait for is a special character...
> 
> Or could it be a further improvement ?

I'd like to think a bit about this, but I see the value of the idea. For now, can you create a ticket?

rust/src/parser.rs
jasonish: Yes. In fact I think parser and app-layer should probably be merged, or something.. They're all app-layer.

victorjulien: Ok will look at merging them.

catenacyber:
> I'd like to think a bit about this, but I see the value of the idea. For now, can you create a ticket?

https://redmine.openinfosecfoundation.org/issues/3531

PR #4677, opened 2020-03-14
Feat/appl records/v25

victorjulien:
#4676, including a first attempt at addressing https://github.com/OISF/suricata/pull/4671#discussion_r392262496 @dbcfd

Passed


catenacyber:
Should this work as well for probing parsers ?

victorjulien:
> Should this work as well for probing parsers ?

These shouldn't be affected right now. PP should be called again and again with new data until they give a definitive answer.

catenacyber:
> These shouldn't be affected right now. PP should be called again and again with new data until they give a definitive answer.

Ok indeed.
That gives me a new QA/fuzzing idea :
1) call the protocol detection engine with some buffer
2) If it successfully recognizes a protocol, call it with the same buffer, but with smaller lengths and check that is either returns `AL_PROTO_UNKNOWN` or the same protocol

What do you think of this ?

victorjulien:
Replaced by #4690

victorjulien:
> > These shouldn't be affected right now. PP should be called again and again with new data until they give a definitive answer.
> 
> Ok indeed.
> That gives me a new QA/fuzzing idea :
> 
> 1. call the protocol detection engine with some buffer
> 2. If it successfully recognizes a protocol, call it with the same buffer, but with smaller lengths and check that is either returns `AL_PROTO_UNKNOWN` or the same protocol
> 
> What do you think of this ?

Think this makes sense.

catenacyber:
Ok for fuzz-v26 with no bugs found in some time

PR #4660, opened 2020-03-10
output-json: add MAC addresses to EVE-JSON logs - v3

satta:
Previous PR: #4654

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#962](https://redmine.openinfosecfoundation.org/issues/962)

Describe changes (to previous PR):
- Change `dst_*` JSON keys to `dest_*`.
- Minor whitespace and formatting fixes.
- Clarify name of function to mention MAC addresses (`JSONFormatAndAddMACAddr`).
- Add suricata-verify cases (https://github.com/OISF/suricata-verify/pull/197)

PR #4678, opened 2020-03-15
output-json: add MAC addresses to EVE-JSON logs - v4

satta:
Previous PR: #4660 

- Change return value of `MacSetAdd` to `void`.

### DIFF
--- a/src/flow.c
+++ b/src/flow.c
@@ -452,6 +452,10 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
             f->flags &= ~FLOW_PROTO_DETECT_TS_DONE;
             p->flags |= PKT_PROTO_DETECT_TS_DONE;
         }
+        if (f->macset != NULL && p->ethh != NULL) {
+            MacSetAdd(f->macset, p->ethh->eth_dst, TOSERVER);
+            MacSetAdd(f->macset, p->ethh->eth_src, TOCLIENT);
+        }
     } else {
         f->tosrcpktcnt++;
         f->tosrcbytecnt += GET_PKT_LEN(p);
@@ -466,6 +470,10 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
         if (f->flags & FLOW_PROTO_DETECT_TC_DONE) {
             f->flags &= ~FLOW_PROTO_DETECT_TC_DONE;
             p->flags |= PKT_PROTO_DETECT_TC_DONE;
+        }
+        if (f->macset != NULL && p->ethh != NULL) {
+            MacSetAdd(f->macset, p->ethh->eth_dst, TOCLIENT);
+            MacSetAdd(f->macset, p->ethh->eth_src, TOSERVER);
         }
     }
 
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -393,6 +393,9 @@ void JsonAddCommonOptions(const OutputJsonCommonSettings *cfg,
     if (cfg->include_metadata) {
         JsonAddMetadata(p, f, js);
     }
+    if (cfg->include_ethernet) {
+        CreateJSONEther(js, p, f);
+    }
     if (cfg->include_community_id && f != NULL) {
         CreateJSONCommunityFlowId(js, f, cfg->community_id_seed);
     }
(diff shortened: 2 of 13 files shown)


src/output-json.c
```diff
@@ -707,6 +710,70 @@ void CreateJSONFlowId(json_t *js, const Flow *f)
...
+           append to arrays */
+        if (f != NULL && f->macset && MacSetSize(f->macset) > 0) {
+            JSONMACAddrInfo info;
+            int ret;
+            info.dst = json_array();
+            info.src = json_array();
+            ret = MacSetForEach(f->macset, JSONFlowAppendMACAddrs, &info);
+            /* should not happen, JSONFlowAppendMACAddrs is sane */
+            if (ret != 0)
+                return ret;
```
victorjulien: both `json_array`'s above are lost here (mem leak)

src/output-json.c
```diff
@@ -1023,6 +1090,15 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
             json_ctx->cfg.include_metadata = true;
         }
 
+        /* Check if ethernet information should be logged. */
+        const ConfNode *ethernet = ConfNodeLookupChild(conf, "ethernet");
+        if (ethernet && ethernet->val && ConfValIsTrue(ethernet->val)) {
+            SCLogConfig("Enabling Ethernet logging.");
```
victorjulien: Suggest: 'Enabling Ethernet MAC-address logging.'

src/util-macset.c
```diff
@@ -0,0 +1,120 @@
...
+        FatalError(SC_ERR_MEM_ALLOC, "Unable to allocate MacSet memory.");
+    }
+    ms->size = size;
+    ms->last[TOSERVER] = ms->last[TOCLIENT] = 0;
+    return ms;
+}
+
+void MacSetAdd(MacSet *ms, uint8_t *addr, int direction)
+{
+    unsigned long i = 0;
```
victorjulien: please use explicit types (e.g. `uint32_t`)

src/util-macset.c
```diff
@@ -0,0 +1,120 @@
...
+    if (ms == NULL)
+        return;
+    if (unlikely(ms->last[direction] == ms->size)) {
+        /* MacSet full */
+        return;
+    }
+    if (ms->last[direction] > 0) {
+        for (i = ms->last[direction]-1; i < ms->size; i--) {
```
victorjulien: `i` can be declared in the `for` statement.

src/util-macset.c
```diff
@@ -0,0 +1,120 @@
...
+    ms->buf[TOCLIENT] = SCCalloc(size, sizeof(MacAddr));
+    if (unlikely(ms->buf[TOCLIENT] == NULL)) {
+        FatalError(SC_ERR_MEM_ALLOC, "Unable to allocate MacSet memory.");
+    }
+    ms->size = size;
+    ms->last[TOSERVER] = ms->last[TOCLIENT] = 0;
+    return ms;
+}
+
+void MacSetAdd(MacSet *ms, uint8_t *addr, int direction)
```
victorjulien: please add a comment about the logic, esp around the purpose of the loop

src/util-macset.c
```diff
@@ -0,0 +1,120 @@
...
+            }
+        }
+    }
+    memcpy(ms->buf[direction] + ms->last[direction], addr, sizeof(MacAddr));
+    ms->last[direction]++;
+}
+
+int MacSetForEach(MacSet *ms, MacSetIteratorFunc func, void *data)
+{
+    unsigned long i = 0;
```
victorjulien: explicit type (please review the rest of the PR for this)

src/util-macset.c
```diff
@@ -0,0 +1,120 @@
...
+            if (likely(memcmp(addr2, addr, sizeof(MacAddr)) == 0)) {
+                return;
+            }
+        }
+    }
+    memcpy(ms->buf[direction] + ms->last[direction], addr, sizeof(MacAddr));
+    ms->last[direction]++;
+}
+
+int MacSetForEach(MacSet *ms, MacSetIteratorFunc func, void *data)
```
victorjulien: style: `func` -> `Func` or `FuncPtr`. Function pointers start with capital.

src/util-macset.c
```diff
@@ -0,0 +1,120 @@
...
+    for (i = 0; i < ms->last[TOCLIENT]; i++) {
+        int ret = func((uint8_t*) ms->buf[TOCLIENT][i], TOCLIENT, data);
+        if (unlikely(ret != 0)) {
+            return ret;
+        }
+    }
+    return 0;
+}
+
+unsigned long MacSetSize(MacSet *ms) {
```
victorjulien: style: { on new line for functions

victorjulien: Please review the rest of the PR for this.

src/util-macset.h
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * \file
+ *
+ * \author Sascha Steinbiss <sascha.steinbiss@dcso.de>
+ */
+
+#ifndef __MACSET_H__
+#define __MACSET_H__
+
+#include <stdint.h>
```
victorjulien: this should not be needed

src/util-macset.c
```diff
@@ -0,0 +1,120 @@
...
+
+struct MacSet_ {
+    MacAddr *buf[2];
+    unsigned long size,
+                  last[2];
+};
+
+MacSet* MacSetInit(int size)
+{
+    MacSet *ms = SCCalloc(1, sizeof(*ms));
```
victorjulien: should this memory be controlled by the flow memcap?

src/util-macset.c
```diff
@@ -0,0 +1,120 @@
...
+
+typedef uint8_t MacAddr[6];
+
+struct MacSet_ {
+    MacAddr *buf[2];
+    unsigned long size,
+                  last[2];
+};
+
+MacSet* MacSetInit(int size)
```
victorjulien: I'm not too excited for having 3 memory allocations here for the common case. Could we have just 1 that stores a single src and dest mac, and only do additional allocs in uncommon case of having multiple macs per flow side?

victorjulien: style: should be `MacSet *MacSetInit...` (position of *)

src/output-json.c
```diff
@@ -707,6 +710,70 @@ void CreateJSONFlowId(json_t *js, const Flow *f)
...
+    JSONMACAddrInfo *info = (JSONMACAddrInfo*) data;
+    if (direction == TOSERVER) {
+        JSONFormatAndAddMACAddr(info->dst, NULL, val, true);
+    } else {
+        JSONFormatAndAddMACAddr(info->src, NULL, val, true);
+    }
+    return 0;
+}
+
+int CreateJSONEther(json_t *parent, const Packet *p, const Flow *f) {
```
victorjulien: style: { on newline (multiple locations)

src/flow.c
```diff
@@ -452,6 +452,10 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
             f->flags &= ~FLOW_PROTO_DETECT_TS_DONE;
             p->flags |= PKT_PROTO_DETECT_TS_DONE;
         }
+        if (f->macset != NULL && p->ethh != NULL) {
+            MacSetAdd(f->macset, p->ethh->eth_dst, TOSERVER);
```
victorjulien: If we always have 2 mac addresses, shouldn't we handle things in a single call?

src/output-json.c
satta: Right, good catch.

src/util-macset.c
satta: I was aware of that, but noticing that flow data structures are apparently recycled (reducing the amount of new allocations) went for simplicity rather than allocation efficiency. Sure, I'll look into that, at the cost of some additional complexity.

src/util-macset.c
satta: It probably should -- will take a look how this is implemented in the rest of the flow code.

src/flow.c
satta: Definitely. Will look into it.

satta:
Next PR: #4689

PR #4645, opened 2020-03-06
kerberos: fix against packet split in record size

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None

Describe changes:
- Fixes kerberos TCP parser against packet split evasion

Problem happens if the split happens in the middle of a record size
Another problems happens if the buffer size is less than 4

Problem found while running suricata-verify tests against Suricata compiled with this patch
```
diff --git a/src/app-layer.c b/src/app-layer.c
index b614f2712..0149d6b4f 100644
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -658,8 +658,15 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
          * a start msg should have gotten us one */
         if (f->alproto != ALPROTO_UNKNOWN) {
             PACKET_PROFILING_APP_START(app_tctx, f->alproto);
+#ifndef LOLSPLIT
+            for (size_t i = 0; i < data_len; i++) {
+                r = AppLayerParserParse(tv, app_tctx->alp_tctx, f, f->alproto,
+                                        flags, data+i, 1);
+            }
+#else
             r = AppLayerParserParse(tv, app_tctx->alp_tctx, f, f->alproto,
                                     flags, data, data_len);
+#endif
             PACKET_PROFILING_APP_END(app_tctx, f->alproto);
             if (r >= 0) {
                 (*stream)->app_progress_rel += data_len;
```

PR #4680, opened 2020-03-16
parse: dns probe_tcp handles incomplete data

catenacyber:
None yet

Describe changes:
- Fixes DNS TCP parser against packet split evasion after gap

Problem happens after gap because of recovery code using `probe_tcp` which does not handle incomplete data
https://github.com/OISF/suricata/blob/master/rust/src/dns/dns.rs#L515

Problem found while running suricata-verify test `dns-tcp-ts-gap` against Suricata compiled with this patch
```
diff --git a/src/app-layer.c b/src/app-layer.c
index b614f2712..0149d6b4f 100644
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -658,8 +658,15 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
          * a start msg should have gotten us one */
         if (f->alproto != ALPROTO_UNKNOWN) {
             PACKET_PROFILING_APP_START(app_tctx, f->alproto);
+#ifndef LOLSPLIT
+            for (size_t i = 0; i < data_len; i++) {
+                r = AppLayerParserParse(tv, app_tctx->alp_tctx, f, f->alproto,
+                                        flags, data+i, 1);
+            }
+#else
             r = AppLayerParserParse(tv, app_tctx->alp_tctx, f, f->alproto,
                                     flags, data, data_len);
+#endif
             PACKET_PROFILING_APP_END(app_tctx, f->alproto);
             if (r >= 0) {
                 (*stream)->app_progress_rel += data_len;
```

This PR fixes it but I am not sure it is right.
So, this is open to discussion

Similar to #4645

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -512,8 +512,8 @@ impl DNSState {
     /// Returns the number of messages parsed.
     pub fn parse_request_tcp(&mut self, input: &[u8]) -> i8 {
         if self.gap {
-            let (is_dns, _) = probe_tcp(input);
-            if is_dns {
+            let (is_dns, _, is_incomplete) = probe_tcp(input);
+            if is_dns || is_incomplete{
                 self.gap = false;
             } else {
                 return 0
@@ -553,8 +553,8 @@ impl DNSState {
     /// Returns the number of messages parsed.
     pub fn parse_response_tcp(&mut self, input: &[u8]) -> i8 {
         if self.gap {
-            let (is_dns, _) = probe_tcp(input);
-            if is_dns {
+            let (is_dns, _, is_incomplete) = probe_tcp(input);
+            if is_dns || is_incomplete{
                 self.gap = false;
             } else {
                 return 0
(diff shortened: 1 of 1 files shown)


catenacyber:
Maybe we should wait for #4677 to get a better fix

victorjulien:
> Maybe we should wait for #4677 to get a better fix

I don't know if #4677 is really related. I think the issue here is that the DNS parser doesn't try hard enough to find the next DNS record in the stream.

catenacyber:
> I don't know if #4677 is really related. I think the issue here is that the DNS parser doesn't try hard enough to find the next DNS record in the stream.

Indeed, that is the issue.
But to find the next DNS record in the stream, the parser must have enough data.
And with #4677, it can specify how much data is needed

jasonish:
> > I don't know if #4677 is really related. I think the issue here is that the DNS parser doesn't try hard enough to find the next DNS record in the stream.
> 
> Indeed, that is the issue.
> But to find the next DNS record in the stream, the parser must have enough data.
> And with #4677, it can specify how much data is needed

So the current approach to finding the next record is rather weak, it hopes that it is called again with the start of record, rather than scanning for potential start of records as that could get very expensive.

victorjulien:
Merged in #4788, with a git commit message rewrite. Thanks Philippe.

PR #4684, opened 2020-03-16
Ssh rust v9

catenacyber:

Describe changes:

This looks good

Modifies #4670 with comments taken into account

Can be improved and rebased when #4677 gets merged

### DIFF
--- a/rust/src/ssh/ssh.rs
+++ b/rust/src/ssh/ssh.rs
@@ -1,0 +1,589 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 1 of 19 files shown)


rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,589 @@
...
+        }
+    }
+
+    fn set_event(&mut self, event: SSHEvent) {
+        let ev = event as u8;
+        core::sc_app_layer_decoder_events_set_event_raw(&mut self.transaction.events, ev);
+    }
+
+    fn parse_record(&mut self, mut input: &[u8], resp: bool) -> bool {
+        //TODO add SCLogDebug ?
```
victorjulien: please remove or add a (meaningful) message here

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,589 @@
...
+        let mut hdr = if !resp {
+            &mut self.transaction.cli_hdr
+        } else {
+            &mut self.transaction.srv_hdr
+        };
+        //first skip record left bytes
+        if hdr.record_left > 0 {
```
victorjulien: Now that #4691 is merged, I wonder how we can use it here as well. The simplest way would be to just return incomplete with an extra byte as 'needed'. Perhaps you can already do an experiment with looking for a specific byte here, just in the this parser. If successful we can then push the logic into a lower level API.

rust/src/ssh/ssh.rs
catenacyber: Ok, adding meaningful `SCLogDebug` which are not the same than the code in C

rust/src/ssh/ssh.rs
catenacyber: Yes with pleasure 👍 🥇

PR #4687, opened 2020-03-16
Replace Jansson with Rust JsonBuilder - v17

jasonish:
Introduces a new JsonBuilder module to take the place of libjannson
for creating JSON logs.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/2726

Previous PR: https://github.com/OISF/suricata/pull/4634

This PR takes a different order of attacking the conversion than
the previous PRs:
- First do some refactors to avoid getting values from the json_t
  object as this is not possible with JsonBuilder. This involves
  create a struct to cache some values like the 5 tuple.
- Add a new method to JsonBuilder that will allow it to take a
  json_t as a value. This allows the alert to be migrated without
  migrating all the metadata that is logged with it.
- Where the Jansson functions needs to remain along with a new
  function that takes a JsonBuilder the term "Eve" is used
  instead of "JSON".
- Convert DHCP, and SIP to JsonBuilder.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/452
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/809


jasonish:
With respect to the prefix, I don't think its important until we expose some sort of public API.. But just for some idea of what this would look like..
- `SuriJBOpenObject`
- `SuriJBSetString`
- `SCJBOpenObject`
- `SCJBSetString`
- `SuriJsonBuilderOpenObject`

jasonish:
My next step will be to convert the DNS parser, then run some benchmarks on Suricata, not just microbenchmarks on encoding.

jasonish:
Continued at https://github.com/OISF/suricata/pull/4831.

PR #4689, opened 2020-03-17
output-json: add MAC addresses to EVE-JSON logs - v5

satta:
Previous PR: #4678 

- Fix unlikely but potential memory leak.
- Change wording of config log message.
- Adjust code formatting and type use.
- Add informative comments where non-trivial actions are done.
- Use capitalized names for function pointers.
- Put allocations under flow memcap constraints.
- Use static memory for the likely case of only two MAC adresses per flow and allocate memory for more only when more are seen.
- Add addresses for both sides of a flow within a single call.
- Add unit tests for `MacSet` data structure.

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -393,6 +393,9 @@ void JsonAddCommonOptions(const OutputJsonCommonSettings *cfg,
     if (cfg->include_metadata) {
         JsonAddMetadata(p, f, js);
     }
+    if (cfg->include_ethernet) {
+        CreateJSONEther(js, p, f);
+    }
     if (cfg->include_community_id && f != NULL) {
         CreateJSONCommunityFlowId(js, f, cfg->community_id_seed);
     }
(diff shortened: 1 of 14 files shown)


src/output-json.c
```diff
@@ -707,6 +710,76 @@ void CreateJSONFlowId(json_t *js, const Flow *f)
...
+        if (f != NULL && f->macset && MacSetSize(f->macset) > 0) {
+            JSONMACAddrInfo info;
+            int ret;
+            info.dst = json_array();
+            info.src = json_array();
+            ret = MacSetForEach(f->macset, JSONFlowAppendMACAddrs, &info);
+            if (unlikely(ret != 0)) {
+                /* should not happen, JSONFlowAppendMACAddrs is sane */
+                json_decref(info.dst);
+                json_decref(info.src);
```
jlucovsky: Need to `json_decref(ejs)` on error return.

satta: Correct. Thanks!

satta:
New PR: #4698

PR #4690, opened 2020-03-17
Feat/appl records/v31

victorjulien:
#4677 rebased, plus fixing a corner case and doing some cleanups.

Passed.


rust/src/dns/dns.rs
```diff
@@ -701,17 +700,21 @@ pub extern "C" fn rs_dns_parse_request_tcp(_flow: *const core::Flow,
...
+            if state.parse_request_tcp(buf) >= 0 {
+                return AppLayerResult::ok();
+            } else {
+                return AppLayerResult::err();
+            }
         }
         state.request_gap(input_len);
     }
-    return 0;
+    AppLayerResult::ok()
```
jasonish: This does functionally change the behaviour of this parser on error. Currently in master, -1 can never be returned, so it will continue trying to parse this flow even after a failure. This change will ripply the failure back to the app-layer. I'm not sure whats best, but I'm not sure if we change the current behaviour in this PR.

victorjulien: I see the change, but not here. Here the `AppLayerResult::err()` condition seems to be unreachable. For UDP, I see it clearly.

victorjulien: Hmm no for UDP we already did pass down failure.

victorjulien: So do you see a patch for the TCP parser to reach the error return path?

jasonish: You're right, err is never returned as we never return < 0.

victorjulien: Fixing it up anyway.

jasonish:
Looking good. No other comments at this time.

victorjulien:
Replaced by #4691

PR #4692, opened 2020-03-18
devguide: document new app-layer retvals

victorjulien:
First attempt to document something in the devguide.

Looking for feedback both on the content as well on the structure and integration.

### DIFF
--- a/doc/devguide/app-layer/parser.rst
+++ b/doc/devguide/app-layer/parser.rst
@@ -1,0 +1,54 @@
+Parser
+======
+
+Parsers return the type `AppLayerResult`.
+
+There are 3 main results:
+ - `APP_LAYER_OK` - parser consumed the data successfully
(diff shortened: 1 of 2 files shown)


doc/devguide/app-layer/parser.rst
```diff
@@ -0,0 +1,54 @@
+Parser
+======
+
+Parsers return the type `AppLayerResult`.
+
+There are 3 main results:
```
jasonish: "3 main results" suggests there are some other non-main results. Maybe something like:
```
Parsers return one of 3 results:
```

doc/devguide/app-layer/parser.rst
```diff
@@ -0,0 +1,54 @@
+Parser
```
jasonish: Parsers?

Then maybe a new section "Return Types" which the rest lives under.

victorjulien:
Feedback implemented in #4699

PR #4674, opened 2020-03-13
rfb: add RFB parser - v8

satta:
Previous PR: #4633 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#3501](https://redmine.openinfosecfoundation.org/issues/3501)

Describe changes (from previous PR):
- Add keyword `rfb.secresult`, allowing one to alert on failed connection attempts.
- Use flow direction for logging (mostly for consistency)
- Include unknown security result number in JSON output.

Updated suricata-verify tests as well.

PR #4695, opened 2020-03-18
rfb: add RFB parser - v9

satta:
Previous PR: #4674

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#3501](https://redmine.openinfosecfoundation.org/issues/3501)

Describe changes (from previous PR):
- Rebase against current master.
- Adapt parser code to use `AppLayerResult`-based buffering control.
- Expand commit message with content summary.
- Fix comment in `src/detect-rfb-secresult.c`.


rust/src/rfb/rfb.rs
```diff
@@ -0,0 +1,799 @@
...
+                                Some(current_transaction) => {
+                                    current_transaction.ts_client_protocol_version = Some(request);
+                                }
+                                _ => {
+                                    return AppLayerResult::err();
+                                }
+                            }
+                        }
+                        Err(nom::Err::Incomplete(v)) => {
+                            if let nom::Needed::Size(n) = v {
```
victorjulien: Are you sure the size `n` here indicates the size of the full record required? Or could it also be something like "I need 2 more bytes to get some integer"? In the latter case you should probably use `n + current.len()`. You must tell the API how much data you need on top of what was already consumed. So this needs to be true: `consumed + needed > input.len()'. Currently this is enforced by the API in a hard way (`BUG_ON` statements).

satta: Okay, got it. Then it should be `current.len() + n` indeed: `current.len()` for what is not consumed yet but was tried to parse, and `n` for what nom thinks it needs to complete on top of what is already has in `current`.
I assume I the `input` I get in the next call would then be guaranteed to be larger than `current.len() + n` from the previous call?

satta:
New PR: #4697

PR #4696, opened 2020-03-18
Feat/cond logging/v5

victorjulien:
Very early preview of conditional logging. Idea is to have special rules that allow us to disable certain features. In this PR only tx logging can be disabled, per tx.

Example rule:
```
config http any any -> any any (http.host; content:"test"; \
         config:logging disable, type tx, scope tx; sid:1;)
```
If `test` is part of the `http.host` buffer, the transaction will not be logged (so no eve.http event).

The way is implemented is by adding a new per tx `AppLayerTxData` struct that can be updated from a new `config` keyword and it accessed from the `OutputTxLogger` to skip logging as needed. The idea is to extend this to add more types (e.g. disable file tracking, disable flow log for specific flows, etc) and more scopes (per packet, flow, tx, but also later host, ippair etc).

Another idea is to unify the various per-tx AppLayer variables (detect flags, logger flags, events, etc into this `AppLayerTxData` struct.

Another example does not work well:
```
config dns any any -> any any (dns.query; content:"suricata"; \
         config:logging disable, type tx, scope tx; sid:1;)
```
This is because DNS does not use proper transactions as they are meant to be used: to match req and resp. This will need to be addressed for this scheme to work.


jasonish:
> config dns any any -> any any (dns.query; content:"suricata"; \
>          config:logging disable, type tx, scope tx; sid:1;)

It would be nice to enumerate all the possibilities you think we might add under the config keyword to see if we should formalize that grammar a little.

jasonish:
> This is because DNS does not use proper transactions as they are meant to be used: to match req and resp. This will need to be addressed for this scheme to work.

With respect to DNS, isn't this just a matter of having the dns_query keyword match on response transactions as well? This is a simple enough code change, in fact there is a ticket for it, and with a quick tweak to detect-dns.c I was able to suppress response logs as well.

DNS may be unique here in that it echo's the request back in the response.  Other protocols do not do this.

Another case to consider is protocols that do have some request/reply semantics, but also have some to_server or to_client messages that are purely unidirectional, dnp3 is one example of this, but probably not a google example where you might suppress logging based on a content match.

victorjulien:
> > config dns any any -> any any (dns.query; content:"suricata"; 
> > config:logging disable, type tx, scope tx; sid:1;)
> 
> It would be nice to enumerate all the possibilities you think we might add under the config keyword to see if we should formalize that grammar a little.

Do you mean like this? https://redmine.openinfosecfoundation.org/issues/1005#note-10

victorjulien:
> > This is because DNS does not use proper transactions as they are meant to be used: to match req and resp. This will need to be addressed for this scheme to work.
> 
> With respect to DNS, isn't this just a matter of having the dns_query keyword match on response transactions as well? This is a simple enough code change, in fact there is a ticket for it, and with a quick tweak to detect-dns.c I was able to suppress response logs as well (there may be some undersirable side affects such as double logging, but that might be worth figuring out to support this).
> 
> DNS may be unique here in that it echo's the request back in the response. Other protocols do not do this.
> 
> Another case to consider is protocols that do have some request/reply semantics, but also have some to_server or to_client messages that are purely unidirectional, dnp3 is one example of this, but probably not a google example where you might suppress logging based on a content match.

Its going to require hacks in the detect engine to inspect a response packet for request side things. Not sure how intrusive that would be. But I fear this is going to be one of those things where its going to be a cost we keep paying while we build things around txs and then having to hack our way around this.

Example: we'll likely do something like 'txvars/txbits' (flowbits but per tx). This wouldn't be hard, but it will be if we have to support these cases too.

Btw if you run this code on https://github.com/OISF/suricata-verify/blob/master/tests/dns-udp-dig-a-www-suricata-ids-org/dig-a-www.suricata-ids.org.pcap, you will see something odd. `OutputTxLogger` seems to see 3, not 2 transactions.

jasonish:
> > > This is because DNS does not use proper transactions as they are meant to be used: to match req and resp. This will need to be addressed for this scheme to work.
> > 
> > 
> > With respect to DNS, isn't this just a matter of having the dns_query keyword match on response transactions as well? This is a simple enough code change, in fact there is a ticket for it, and with a quick tweak to detect-dns.c I was able to suppress response logs as well (there may be some undersirable side affects such as double logging, but that might be worth figuring out to support this).
> > DNS may be unique here in that it echo's the request back in the response. Other protocols do not do this.
> > Another case to consider is protocols that do have some request/reply semantics, but also have some to_server or to_client messages that are purely unidirectional, dnp3 is one example of this, but probably not a google example where you might suppress logging based on a content match.
> 
> Its going to require hacks in the detect engine to inspect a response packet for request side things. Not sure how intrusive that would be. But I fear this is going to be one of those things where its going to be a cost we keep paying while we build things around txs and then having to hack our way around this.
> 
> Example: we'll likely do something like 'txvars/txbits' (flowbits but per tx). This wouldn't be hard, but it will be if we have to support these cases too.
> 

A patch like this will match the query in the DNS responses:
https://gist.github.com/jasonish/70b99c53a69d5ab8ba99b0e587a90be8

Unfortunately I don't think we really want this as it will double the amount of alerts as there is no good way to associated a response to a request. But we could add a new keyword, like dns_query, but for responses, dns_response.query_name or something to specifically look at the query in the response. It would require 2 config lines to suppress these, but seems doable without reworking much.

jasonish:
> Btw if you run this code on https://github.com/OISF/suricata-verify/blob/master/tests/dns-udp-dig-a-www-suricata-ids-org/dig-a-www.suricata-ids.org.pcap, you will see something odd. `OutputTxLogger` seems to see 3, not 2 transactions.

Don't see it or looking in the wrong place.  I only see the output logger trying to access 2 transactions.

victorjulien:
Continued at #4982

PR #4699, opened 2020-03-19
devguide: document new app-layer retvals

victorjulien:
Updated version of #4692, addressing comments and adding a bit more documentation.

Idea: can we have a 'magic' key/keyword that we can use to map things to our doxygen page?

e.g.
```
.. doxygen-function:: AppLayerParserParse
```
Which would add a link to https://doxygen.openinfosecfoundation.org/app-layer-parser_8h.html#ac2758282ee3f771558135acf2bb0a466

### DIFF
--- a/doc/devguide/app-layer/index.rst
+++ b/doc/devguide/app-layer/index.rst
@@ -1,0 +1,7 @@
+App-Layer
+=========
+
+.. toctree::
+   :maxdepth: 2
+
+   parser.rst
--- a/doc/devguide/app-layer/parser.rst
+++ b/doc/devguide/app-layer/parser.rst
@@ -1,0 +1,99 @@
+*******
+Parsers
+*******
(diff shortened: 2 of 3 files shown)


jasonish:
> Idea: can we have a 'magic' key/keyword that we can use to map things to our doxygen page?
> 
> e.g.
> 
> ```
> .. doxygen-function:: AppLayerParserParse
> ```
> 
> Which would add a link to https://doxygen.openinfosecfoundation.org/app-layer-parser_8h.html#ac2758282ee3f771558135acf2bb0a466

While I'm not sure, I think we'd have to generate doxygen, parse its xml output (which I think it can output), write a sphinx plugin, or do some post-processing.  Then we'd probably want the same for Rust.

victorjulien:
Merged through #4710

PR #4701, opened 2020-03-19
Http2 v1

catenacyber:

DRAFT for discussion
Not to be merged

Describe changes:
- Adds basic HTTP2 parser

Much more work to be done :
 + Commit squashing
 + parse deeper
 + handle transactions with progress completion
 + more generic code for logging
 + tests
 + detect why 2 alerts GOAWAY ?
 + robustness against evasions


rust/src/http2/parser.rs
```diff
@@ -0,0 +1,138 @@
...
+impl fmt::Display for HTTP2FrameType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl std::str::FromStr for HTTP2FrameType {
+    type Err = String;
+
+    //TODO more automatic
```
catenacyber: @jasonish any tips here ?

src/detect-http2.c
```diff
@@ -0,0 +1,178 @@
...
+    //TODO create a new doc file for HTTP2 keywords
+    sigmatch_table[DETECT_HTTP2_FRAMETYPE].url = DOC_URL DOC_VERSION "/rules/http2-keywords.html#frametype";
+    sigmatch_table[DETECT_HTTP2_FRAMETYPE].Match = NULL;
+    sigmatch_table[DETECT_HTTP2_FRAMETYPE].AppLayerTxMatch = DetectHTTP2frametypeMatch;
+    sigmatch_table[DETECT_HTTP2_FRAMETYPE].Setup = DetectHTTP2frametypeSetup;
+    sigmatch_table[DETECT_HTTP2_FRAMETYPE].Free = DetectHTTP2frametypeFree;
+#ifdef UNITTESTS
+    sigmatch_table[DETECT_HTTP2_FRAMETYPE].RegisterTests = DetectHTTP2RegisterTests;
+#endif
+    //TODO investigate why this is needed
```
catenacyber: @victorjulien can you point me to some explanation what is going on here ?

src/detect-http2.c
```diff
@@ -0,0 +1,178 @@
...
+    if (DetectSignatureSetAppProto(s, ALPROTO_HTTP2) != 0)
+        return -1;
+
+    if (!DetectHTTP2FuncParseFrameType(str, &frame_type)) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE,
+                   "Invalid argument \"%s\" supplied to http2.frametype keyword.", str);
+        return -1;
+    }
+
+    //TODO do we need to allocate ? Could we unionize sm->ctx ?
```
catenacyber: Thoughts here ? Allocate one byte looks like an overkill

src/output-json-http2.c
```diff
@@ -0,0 +1,243 @@
...
+#include "util-logopenfile.h"
+#include "util-crypt.h"
+
+#include "output-json.h"
+#include "output-json-http2.h"
+#include "rust.h"
+
+#define MODULE_NAME "LogHttp2Log"
+
+//TODO Cannot this be generic ? and most of the file as well
```
catenacyber: @victorjulien thoughts here ?

catenacyber:
This needs rebasing on latest master and using `AppLayerResult`...

rust/src/http2/http2.rs
```diff
@@ -0,0 +1,668 @@
...
+        //second extend buffer if present
+        if self.response_buffer.len() > 0 {
+            self.response_buffer.extend(toparse);
+            //parse one header locally as we borrow self and self.response_buffer
+            match parser::http2_parse_frame_header(&self.response_buffer) {
+                Ok((rem, head)) => {
+                    let hl = head.length as usize;
+                    let rlu = rem.len();
+
+                    let mut tx = self.new_tx();
```
victorjulien: response should get the tx already created by request side

src/detect-http2.c
victorjulien: Not sure I understand what you're asking.

src/detect-http2.c
victorjulien: Right now it is necessary. I thought there was a ticket about this, can I can't find it.

src/output-json-http2.c
victorjulien: probably. cc @jasonish

rust/src/http2/http2.rs
```diff
@@ -0,0 +1,668 @@
...
+#[derive(Copy, Clone, PartialOrd, PartialEq)]
+pub enum HTTP2ConnectionState {
+    Http2StateInit = 0,
+    Http2StateMagicDone = 1,
+}
+
+pub struct HTTP2Transaction {
+    tx_id: u64,
+    pub ftype: Option<parser::HTTP2FrameType>,
+
```
victorjulien: to me it seems that it would make sense to have a tx per 'stream', so then the stream id would be the key to mapping request and responses together.

src/detect-http2.c
catenacyber: What does `DetectAppLayerInspectEngineRegister` do ?

src/detect-http2.c
catenacyber: Do you want me to work on it ?

src/detect-http2.c
victorjulien: It registers an inspection engine to the statement part of the detection engine, for a specific protocol at a specific progress level.

rust/src/http2/http2.rs
catenacyber: Yes that is what I meant by "handle transactions with progress completion"

rust/src/http2/http2.rs
catenacyber: When I said "handle transactions with progress completion", I meant I have not yet done anything good for transactions :-p

src/detect-http2.c
catenacyber: Thanks, what is "the statement part" ?

victorjulien: Maybe it's best to look at an example from ssh (detect-ssh-software-version.c)

It first creates an inspect engine to be called when the SSH parser reaches the 'banner' state:

```c
    g_ssh_banner_list_id = DetectBufferTypeRegister("ssh_banner");

    DetectAppLayerInspectEngineRegister("ssh_banner",
            ALPROTO_SSH, SIG_FLAG_TOSERVER, SSH_STATE_BANNER_DONE,
            InspectSshBanner);
```
The inspection function is a generic one that just executes a normal 'SigMatch' list:
```c
static int InspectSshBanner(ThreadVars *tv,
        DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx,
        const Signature *s, const SigMatchData *smd,
        Flow *f, uint8_t flags, void *alstate,
        void *txv, uint64_t tx_id)
{
    return DetectEngineInspectGenericList(tv, de_ctx, det_ctx, s, smd,
                                          f, flags, alstate, txv, tx_id);
}
```
(this wrapper may actually be unnecessary, could register `DetectEngineInspectGenericList` directly I think)

These things don't do anything by themselves yet, we now need one or more keywords that will need to evaluated for SSH, when the banner is available. This is what is done here for the 'ssh.software_version' keyword:

```c
static int DetectSshSoftwareVersionSetup (DetectEngineCtx *de_ctx, Signature *s, const char *str)
{
    DetectSshSoftwareVersionData *ssh = NULL;
    SigMatch *sm = NULL;

    if (DetectSignatureSetAppProto(s, ALPROTO_SSH) != 0)
        return -1;

    ssh = DetectSshSoftwareVersionParse(str);
    if (ssh == NULL)
        goto error;

    /* Okay so far so good, lets get this into a SigMatch
     * and put it in the Signature. */
    sm = SigMatchAlloc();
    if (sm == NULL)
        goto error;

    sm->type = DETECT_AL_SSH_SOFTWAREVERSION;
    sm->ctx = (void *)ssh;

    SigMatchAppendSMToList(s, sm, g_ssh_banner_list_id);
    return 0;
}
```

The last thing the setup function here does before returning, is adding the `SigMatch` to the banner inspection list.

src/detect-http2.c
victorjulien: Lets make that a separate project. I think http2 will be complex enough by itself.

rust/src/http2/http2.rs
victorjulien: Ok :)

src/detect-http2.c
catenacyber: Indeed, that is a separate project.
But do you want me to create a ticket for this and work on it ?

src/detect-http2.c
catenacyber: Many thanks for the walk through explanation :-)

src/detect-http2.c
victorjulien: Just the ticket for now. I don't think it's high priority atm.

catenacyber: https://redmine.openinfosecfoundation.org/issues/3544

rust/src/http2/parser.rs
jasonish: Not really, this type of thing is pretty common.

catenacyber: Ok, and what about the next line :

This works
```
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
```

And I would like this

```
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s.to_uppercase() {
```

which does not work like that

PR #4704, opened 2020-03-19
Fuzz v28

catenacyber:
Link to redmine ticket:

Describe changes:
- Adds eight fuzz targets

Modifies #4700 making cocci happy with cast to `uintmax_t` in `printf`

After this, we can 


victorjulien:
Merged in #4715, thanks Philippe!

PR #4707, opened 2020-03-19
Remove libhtp as a dependency, now provided by a Rust crate.

jasonish:
This also allows us to use automake convenience libraries, so show an example of that.


jasonish:
Oops, didn't mean to do this here.

PR #4698, opened 2020-03-18
output-json: add MAC addresses to EVE-JSON logs - v6

satta:
Previous PR: #4689 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#962](https://redmine.openinfosecfoundation.org/issues/962)

Describe changes (to previous PR):
- Make sure that `ejs` is freed on error.

PR #4709, opened 2020-03-19
output-json: add MAC addresses to EVE-JSON logs - v7

satta:
Previous PR: #4698 

- Use `PASS` macro to signal successful test.
- Never fail hard when encountering allocation issues.
- Properly maintain flow memuse value across allocations/deallocations.
- Minor style issues.

### DIFF
--- a/src/flow-util.c
+++ b/src/flow-util.c
@@ -201,6 +201,12 @@ void FlowInit(Flow *f, const Packet *p)
 
     f->protomap = FlowGetProtoMapping(f->proto);
 
+    if (f->macset != NULL) {
+        MacSetReset(f->macset);
+    } else {
+        f->macset = MacSetInit(10);
+    }
+
     SCReturn;
 }
 
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -393,6 +393,9 @@ void JsonAddCommonOptions(const OutputJsonCommonSettings *cfg,
     if (cfg->include_metadata) {
         JsonAddMetadata(p, f, js);
     }
+    if (cfg->include_ethernet) {
+        CreateJSONEther(js, p, f);
+    }
     if (cfg->include_community_id && f != NULL) {
         CreateJSONCommunityFlowId(js, f, cfg->community_id_seed);
     }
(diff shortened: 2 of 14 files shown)


src/flow-util.c
```diff
@@ -201,6 +201,12 @@ void FlowInit(Flow *f, const Packet *p)
 
     f->protomap = FlowGetProtoMapping(f->proto);
 
+    if (f->macset != NULL) {
```
jlucovsky: In the `init` path, when will `f->macset` be non-NULL?

satta: Please correct me if I'm wrong: I just wanted to make sure that in case `FlowInit()` is called on a recycled flow, there is no memory leak my overwriting `f->macset` with a newly allocated instance. If this is not neessary here, I'd be happy to simplify the code.

src/output-json.c
```diff
@@ -707,6 +710,76 @@ void CreateJSONFlowId(json_t *js, const Flow *f)
...
+                /* should not happen, JSONFlowAppendMACAddrs is sane */
+                json_decref(info.dst);
+                json_decref(info.src);
+                json_decref(ejs);
+                return ret;
+            }
+            json_object_set_new(ejs, "dest_macs", info.dst);
+            json_object_set_new(ejs, "src_macs", info.src);
+        }
+    } else {
```
jlucovsky: Make sure that `p->ethh` is non-NULL.

src/output-json.c
```diff
@@ -707,6 +710,76 @@ void CreateJSONFlowId(json_t *js, const Flow *f)
...
+        /* this is a packet context, so we need to add scalar fields */
+        uint8_t *src, *dst;
+        if ((PKT_IS_TOCLIENT(p))) {
+            src = p->ethh->eth_dst;
+            dst = p->ethh->eth_src;
+        } else {
+            src = p->ethh->eth_src;
+            dst = p->ethh->eth_dst;
+        }
+        json_object_set_new(ejs, "type", json_integer(ntohs(p->ethh->eth_type)));
```
jlucovsky: A hex value would be more helpful for `type`

src/output-json.c
```diff
@@ -707,6 +710,76 @@ void CreateJSONFlowId(json_t *js, const Flow *f)
...
+        JSONFormatAndAddMACAddr(ejs, "src_mac", src, false);
```
jlucovsky: Would it be possible to use `JSONFlowAppendMACAddrs` instead of calling `JSONFormatAndAddMACAddr`? Perhaps rename `JSONFlowAppend` --> `JSONMACAppend`?

This would reduce reuse code and reduce the # of calls to `JSONFormatAndAddMACAddr`.

src/output-json.c
```diff
@@ -707,6 +710,76 @@ void CreateJSONFlowId(json_t *js, const Flow *f)
...
+            info.src = json_array();
+            int ret = MacSetForEach(f->macset, JSONFlowAppendMACAddrs, &info);
+            if (unlikely(ret != 0)) {
+                /* should not happen, JSONFlowAppendMACAddrs is sane */
+                json_decref(info.dst);
+                json_decref(info.src);
+                json_decref(ejs);
+                return ret;
+            }
+            json_object_set_new(ejs, "dest_macs", info.dst);
```
jlucovsky: Is the Ethertype available? For consistency, it'd be helpful to output that -- either as a list or a single value.

src/flow-util.c
jlucovsky: Ok ... This case happens alot (that is., `f->macset != NULL`.

Can memory be leaked by resetting a macset that has been extended with an add'l `SCCalloc`?

src/util-macset.c
```diff
@@ -0,0 +1,398 @@
...
+        case MULTI_MAC:
+            if (unlikely(ms->last[side] == ms->size)) {
+                /* MacSet full, ignore item. We intentionally do not output
+                   any warning in order not to stall packet processing */
+                return;
+            }
+            /* If the set is non-empty... */
+            if (ms->last[side] > 0) {
+                /* ...we search for duplicates in the set to decide whether
+                   we need to insert the curent item. We do this backwards,
```
jlucovsky: nit: s/curent/current

src/output-json.c
satta: Hm. I could but to me that would feel unnatural, TBH. `JSONFlowAppendMACAddrs` is just meant to be a one-off utility function to match the `MacSetIteratorFunc` and IMHO only makes sense where `MacSetForEach` is called. 
`JSONFormatAndAddMACAddr` is what I feel one would want to use in general, and also it is inlined, saving us one real function call in the packet context.

I can rename `JSONFlowAppendMACAddrs` to `MacSetIterateToJSON` to make it clearer and reduce confusion, what do you think?

src/flow-util.c
satta: I can't see where. A buffer, once created, is 
- never allocated again (only when it is `NULL` initially)
- never touched during the `MacSetReset()`
- always freed on `MacSetFree()` if it is not `NULL`.

src/output-json.c
satta: Right, will look into that

src/output-json.c
satta: Not stored so far due to complexity. This also brings the question whether e want to keep an association between the ethertype and to the MAC contexts that it was seen in or not?

TBH I'd rather ask: "Do we need the ethertype at all?" Let's say the next layer is a protocol that Suricata can decode, then it would be visible in the metadata added by the next layer (e.g. IPv4/v6, VLAN, etc.). If the next layer is a protocol that Suricata can not decode, then we would not have anything to add this JSON to -- IIRC there is nothing to trigger a flow-context JSON output except for a app-layer tx or `flow` event (which would need to be based on a decodable protocol first) or an alert (where `ip` is the most generic protocol that can be defined in rules).
I've just included the ethertype so I do not deviate from Eric's first version, but personally I don't see much value there. Any comments?

src/util-macset.c
satta: Done, will be in the next iteration.

src/output-json.c
satta: Yep, better safe than sorry. Will be in the next PR.

src/output-json.c
jlucovsky: Leave things as they are. Just wanted to check. Since you're inlining the function, it should be ok.

src/output-json.c
jlucovsky: I'm not sure how much value having the ethertype adds. Perhaps others have opinions?

src/output-json.c
jlucovsky: When replaying one of my pcaps, a SEGV happened because it was NULL.

victorjulien: @jlucovsky in such based please post the bt as well.

src/output-json.c
satta: I'll look into handling the ethertype the same way as the MAC addresses in my next PR, for the sake of completeness.

satta: After taking a second look, this would mean duplicating quite a bit of logic for maintaining buffers for different types (`MacAddr`s and `uint16_t`s). Before implementing this, I would like to find out if it is even possible to have multiple different ethertypes in this context, i.e. would such packets even be appearing in the same flow?

jlucovsky: I don't think the effort is worth it.

satta:
Next PR: #4767

PR #4711, opened 2020-03-20
Ssh rust v11

catenacyber:

Describe changes:

This looks good

Modifies #4705 with
- SSHParserTest07 passing with new code

Do I translate all other tests like this one ?

### DIFF
--- a/src/app-layer-ssh.c
+++ b/src/app-layer-ssh.c
@@ -41,6 +41,7 @@
 #include "app-layer-protos.h"
 #include "app-layer-parser.h"
 #include "app-layer-ssh.h"
+#include "rust.h"
 
 #include "conf.h"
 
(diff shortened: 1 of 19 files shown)


src/app-layer-ssh.c
```diff
@@ -1050,76 +443,76 @@ static int SSHParserTest06(void)
...
-    }
+    p = PacketGetFromAlloc();
+    FAIL_IF(unlikely(p == NULL));
+    p->proto = IPPROTO_TCP;
+    p->flow = f;
 
-    if (ssh_state->cli_hdr.software_version == NULL) {
-        printf("Client version string not parsed: ");
-        goto end;
+    FLOWLOCK_WRLOCK(f);
```
victorjulien: these locks are no longer needed in tests

src/app-layer-ssh.c
```diff
@@ -1050,76 +443,76 @@ static int SSHParserTest06(void)
...
+    uint32_t seq = 2;
+    for (int i=0; i<2; i++) {
+        if (StreamTcpUTAddSegmentWithPayload(&tv, ra_ctx, &ssn.client, seq, (uint8_t *) sshbufs[i], strlen(sshbufs[i])) == -1) {
+            printf("failed to add segment 1: ");
```
victorjulien: use FAIL_IF

src/app-layer-ssh.c
```diff
@@ -1050,76 +443,76 @@ static int SSHParserTest06(void)
...
+            goto end;
+        }
+        seq += strlen(sshbufs[i]);
+        int r = StreamTcpReassembleAppLayer(&tv, ra_ctx, &ssn, &ssn.client, p, UPDATE_DIR_PACKET);
+        if (r < 0) {
+            printf("StreamTcpReassembleAppLayer failed: ");
```
victorjulien: FAIL_IF

src/app-layer-ssh.c
```diff
@@ -1050,76 +443,76 @@ static int SSHParserTest06(void)
...
+            goto end;
+        }
     }
 
-    if (ssh_state->cli_hdr.proto_version == NULL) {
-        printf("Client version string not parsed: ");
+    void *ssh_state = f->alstate;
+    if (ssh_state == NULL) {
```
victorjulien: FAIL_IF_NULL

src/app-layer-ssh.c
catenacyber: ok

src/app-layer-ssh.c
catenacyber: ok

src/app-layer-ssh.c
catenacyber: ok

src/app-layer-ssh.c
catenacyber: ok

PR #4730, opened 2020-03-26
Ssh rust v12

catenacyber:

Describe changes:

This looks good

Modifies #4711 with rebasing, all passing unit tests and comments taken into account

### DIFF
--- a/rust/src/ssh/parser.rs
+++ b/rust/src/ssh/parser.rs
@@ -1,0 +1,198 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use nom::combinator::rest;
+use nom::number::streaming::{be_u32, be_u8};
+use std::fmt;
+
+#[inline]
+fn is_not_lineend(b: u8) -> bool {
+    if b == 10 || b == 13 {
+        return false;
+    }
+    return true;
+}
+
+//may leave \r at the end to be removed
+named!(pub ssh_parse_line<&[u8], &[u8]>,
+    terminated!(
+        take_while!(is_not_lineend),
+        alt!( tag!("\n") | tag!("\r\n") |
+              do_parse!(
+                    bytes: tag!("\r") >>
+                    not!(eof!()) >> (bytes)
+                )
+            )
+    )
+);
+
+#[derive(PartialEq)]
+pub struct SshBanner<'a> {
+    pub protover: &'a [u8],
+    pub swver: &'a [u8],
+}
+
+impl<'a> SshBanner<'a> {}
+
+// Could be simplified adding dummy \n at the end
+// or use nom5 nom::bytes::complete::is_not
+named!(pub ssh_parse_banner<SshBanner>,
+    do_parse!(
+        tag!("SSH-") >>
+        protover: is_not!("-") >>
+        char!('-') >>
+        swver: alt!( complete!( is_not!(" \r\n") ) | rest ) >>
+        //remaning after space is comments
+        (SshBanner{protover, swver})
+    )
+);
+
+#[derive(PartialEq)]
+pub struct SshRecordHeader {
+    pub pkt_len: u32,
+    padding_len: u8,
+    pub msg_code: u8,
+}
+
+impl fmt::Display for SshRecordHeader {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "(pkt_len:{}, padding_len:{}, msg_code:{})",
+            self.pkt_len, self.padding_len, self.msg_code
+        )
+    }
+}
+
+named!(pub ssh_parse_record_header<SshRecordHeader>,
+    do_parse!(
+        pkt_len: verify!(be_u32, |&val| val > 1) >>
+        padding_len: be_u8 >>
+        msg_code: be_u8 >>
+        (SshRecordHeader{pkt_len, padding_len, msg_code})
+    )
+);
+
+//test for evasion against pkt_len=0or1...
+named!(pub ssh_parse_record<SshRecordHeader>,
+    do_parse!(
+        pkt_len: verify!(be_u32, |&val| val > 1) >>
+        padding_len: be_u8 >>
+        msg_code: be_u8 >>
+        take!((pkt_len-2) as usize) >>
+        (SshRecordHeader{pkt_len, padding_len, msg_code})
+    )
+);
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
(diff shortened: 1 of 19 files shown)


rust/src/ssh/parser.rs
```diff
@@ -0,0 +1,198 @@
...
+    )
+);
+
+#[derive(PartialEq)]
+pub struct SshBanner<'a> {
+    pub protover: &'a [u8],
+    pub swver: &'a [u8],
+}
+
+impl<'a> SshBanner<'a> {}
```
victorjulien, line 49: whats the purpose of this?

rust/src/ssh/parser.rs
```diff
@@ -0,0 +1,198 @@
...
+
+// Could be simplified adding dummy \n at the end
+// or use nom5 nom::bytes::complete::is_not
+named!(pub ssh_parse_banner<SshBanner>,
+    do_parse!(
+        tag!("SSH-") >>
+        protover: is_not!("-") >>
+        char!('-') >>
+        swver: alt!( complete!( is_not!(" \r\n") ) | rest ) >>
+        //remaning after space is comments
```
victorjulien, line 59: typo: remaining

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,568 @@
...
+            transaction: SSHTransaction::new(),
+        }
+    }
+
+    fn set_event(&mut self, event: SSHEvent) {
+        let ev = event as u8;
+        core::sc_app_layer_decoder_events_set_event_raw(&mut self.transaction.events, ev);
+    }
+
+    fn parse_record(
```
victorjulien, line 133: style: reduce lines for func declaration

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,568 @@
...
+                        }
+                        Err(nom::Err::Incomplete(_)) => {
+                            //we may have consumed data from previous records
+                            if input.len() < SSH_RECORD_HEADER_LEN {
+                                return AppLayerResult::incomplete(
+                                    (il - input.len()) as u32,
+                                    SSH_RECORD_HEADER_LEN as u32,
+                                );
+                            } else {
+                                //TODO BUG_ON ?
```
victorjulien, line 199: If `ssh_parse_record_header` returns Incomplete even though the input length is >= SSH_RECORD_HEADER_LEN, then thats a bug right?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,568 @@
...
+                    SCLogDebug!("SSH invalid record {}", e);
+                    self.set_event(SSHEvent::InvalidRecord);
+                    return AppLayerResult::err();
+                }
+            }
+        }
+        return AppLayerResult::ok();
+    }
+
+    fn parse_banner(
```
victorjulien, line 221: style: formatting

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,568 @@
...
+        } else {
+            &mut self.transaction.srv_hdr
+        };
+        if hdr.flags == SSHConnectionState::SshStateBannerWaitEol {
+            match parser::ssh_parse_line(input) {
+                Ok((rem, _)) => {
+                    return self.parse_record(rem, resp, pstate);
+                }
+                Err(nom::Err::Incomplete(_)) => {
+                    if input.len() > 0 {
```
victorjulien, line 238: in what case would this func be called with `input.len() == 0`?

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,568 @@
...
+                        hdr.flags = SSHConnectionState::SshStateBannerDone;
+                    }
+                    Err(e) => {
+                        SCLogDebug!("SSH invalid banner {}", e);
+                        self.set_event(SSHEvent::InvalidBanner);
+                        return AppLayerResult::err();
+                    }
+                }
+                if line.len() >= SSH_MAX_BANNER_LEN {
+                    SCLogDebug!(
```
victorjulien, line 269: style compress

victorjulien: (multiple locations)

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    let ip_proto_str = CString::new("tcp").unwrap();
+
+    if AppLayerProtoDetectConfProtoDetectionEnabled(ip_proto_str.as_ptr(), parser.name) != 0 {
+        let alproto = AppLayerRegisterProtocolDetection(&parser, 1);
+        ALPROTO_SSH = alproto;
+        if AppLayerParserConfParserEnabled(ip_proto_str.as_ptr(), parser.name) != 0 {
+            let _ = AppLayerRegisterParser(&parser, alproto);
+        }
+        SCLogNotice!("Rust ssh parser registered.");
```
victorjulien, line 564: reduce level for final PR

rust/src/ssh/parser.rs
catenacyber: ok thanks

rust/src/ssh/ssh.rs
catenacyber: What do you mean ?
For the style I do `rustfmt rust/src/ssh/*.rs`
What should I do ?

rust/src/ssh/ssh.rs
catenacyber: Same question about `rustfmt` as before

rust/src/ssh/ssh.rs
catenacyber: Same question about rustfmt as before

rust/src/ssh/ssh.rs
catenacyber: ok `SCLogDebug`

rust/src/ssh/ssh.rs
catenacyber: I do not know. This is to be safe.
I guess in case of end of connection, we get called with `input.len() == 0`.

But I guess it is better to skip that case and not consume anything...

rust/src/ssh/ssh.rs
catenacyber: Yes.

I will propose next PR with `nom::Needed::Size(x)`

rust/src/ssh/parser.rs
catenacyber: Seems to be an unused leftover from a previous implementation.

Removing it, thanks :-)

rust/src/ssh/ssh.rs
catenacyber: Upgraded to `rustfmt --config-path rust rust/src/ssh/*.rs` and no changes in the code

jasonish: This one argument per line is the rustfmt default when the line would go over 100 chars, as its the default it seems to be the most common way of formatting such functions.

Rustfmt does have an option `fn_args_layout = "compressed"` which would result in this function looking like:
```
    fn parse_record(
        &mut self, mut input: &[u8], resp: bool, pstate: *mut std::os::raw::c_void,
    ) -> AppLayerResult {
```
As I personally don't find one easier to read than the other I'd rather stick to the default. But moving to be compressed would be OK with me.

victorjulien: Can we control the rustfmt more? I'd like the function headers to look a lot more like how we do it in C. I think it's using far to much space in the default.

catenacyber: Done in #4790

Diff for this function is 
```
     fn parse_record(
-        &mut self,
-        mut input: &[u8],
-        resp: bool,
-        pstate: *mut std::os::raw::c_void,
+        &mut self, mut input: &[u8], resp: bool, pstate: *mut std::os::raw::c_void,
     ) -> AppLayerResult {
         let (mut hdr, ohdr) = if !resp {
             (&mut self.transaction.cli_hdr, &self.transaction.srv_hdr)
```

PR #4736, opened 2020-03-27
Http2 v2

catenacyber:

DRAFT for discussion
Not to be merged

Describe changes:
- Adds basic HTTP2 parser

Much more work to be done :
 + Commit squashing
 + parse deeper
 + handle transactions with progress completion
 + more generic code for logging
 + tests
 + robustness against evasions

Modifies #4701 with 
- rebasing
- buffering using `AppLayerResult::incomplete`
- parsing/logging/detecting GOAWAY error code


rust/src/http2/http2.rs
```diff
@@ -0,0 +1,627 @@
...
+                }
+                Err(nom::Err::Incomplete(_)) => {
+                    //we may have consumed data from previous records
+                    if input.len() < HTTP2_FRAME_HEADER_LEN {
+                        return AppLayerResult::incomplete(
+                            (il - input.len()) as u32,
+                            HTTP2_FRAME_HEADER_LEN as u32,
+                        );
+                    } else {
+                        //TODO BUG_ON ?
```
catenacyber: Is there something equivalent to `BUG_ON` in Rust ? @jasonish

src/detect-http2.c
```diff
@@ -0,0 +1,286 @@
...
+    sigmatch_table[DETECT_HTTP2_ERRORCODE].name = "http2.errorcode";
+    sigmatch_table[DETECT_HTTP2_ERRORCODE].desc = "match on HTTP2 error code field";
+    sigmatch_table[DETECT_HTTP2_ERRORCODE].url = DOC_URL DOC_VERSION "/rules/http2-keywords.html#errorcode";
+    sigmatch_table[DETECT_HTTP2_ERRORCODE].Match = NULL;
+    sigmatch_table[DETECT_HTTP2_ERRORCODE].AppLayerTxMatch = DetectHTTP2errorcodeMatch;
+    sigmatch_table[DETECT_HTTP2_ERRORCODE].Setup = DetectHTTP2errorcodeSetup;
+    sigmatch_table[DETECT_HTTP2_ERRORCODE].Free = DetectHTTP2errorcodeFree;
+#ifdef UNITTESTS
+    //TODO should we call multiple times DetectHTTP2RegisterTests ?
+    sigmatch_table[DETECT_HTTP2_ERRORCODE].RegisterTests = DetectHTTP2RegisterTests;
```
catenacyber: I write TODOs to ask questions, so 

Should we call multiple times `DetectHTTP2RegisterTests` ? @victorjulien

src/detect-http2.c
```diff
@@ -0,0 +1,286 @@
...
+
+{
+    uint32_t *detect = (uint32_t *)ctx;
+
+    int value = rs_http2_tx_get_errorcode(txv, flags);
+    if (value < 0) {
+        //no value, no match
+        return 0;
+    }
+    return *detect == (uint32_t) value;
```
catenacyber: I guess we should improve this...
Any thoughts ?

rust/src/http2/http2.rs
```diff
@@ -0,0 +1,627 @@
...
+                    let mut tx = self.new_tx();
+                    tx.ftype = Some(head.ftype);
+                    match head.ftype {
+                        parser::HTTP2FrameType::GOAWAY => {
+                            match parser::http2_parse_frame_goaway(rem) {
+                                Ok((_, goaway)) => {
+                                    tx.type_data = Some(HTTP2FrameTypeData::GOAWAY(goaway));
+                                }
+                                Err(nom::Err::Incomplete(_)) => {
+                                    return AppLayerResult::incomplete(
```
catenacyber: This seems the right way to do so far :
- parse frame header
- keep state of remaining frame bytes
- use tcp buffering if we need to parse deeper

Should be `HTTP2_FRAME_HEADER_LEN + 4`

src/detect-http2.c
victorjulien: Once. Otherwise it would register the same tests multiple times.

rust/src/http2/http2.rs
victorjulien: `panic!("msg");`

src/detect-http2.c
victorjulien: Seems like premature optimization to me.

rust/src/http2/http2.rs
victorjulien: `nom::Err::Incomplete(_)` should tell you how many more bytes it needs. Can you use that too?

src/detect-http2.c
catenacyber: ok thanks

src/detect-http2.c
catenacyber: So, just leave a TODO comment for now ?

victorjulien: It's likely we'll want things like negation or ranges here too

catenacyber: Ok for negation, but not sure for ranges as this field is an enumeration

rust/src/http2/http2.rs
catenacyber: ok thanks

src/detect-http2.c
catenacyber: And do not we have generic negation for whatever keyword (returning the negation of the Match function) ?

rust/src/http2/http2.rs
catenacyber: Ok.
Should we try to be smarter : when we have a composite type, u16 and u32 for instance, we know we will need 6 bytes, even if nom tells us we first need 2 bytes ?

PR #4716, opened 2020-03-21
[preview] Cleanup thread contextual info when needed

jlucovsky:
@victorjulien and @pevma This is a preview PR that corrects the issues in 3273. Please comment upon the approach.

I used these rules -- the first rule demonstrated the issue. With this PR, memory is no longer overrun.
```
alert http any any -> any any (msg:"High entropy in file name download"; file.name; content:".exe"; endswith; fast_pattern;  bsize:<21; lua:any.lua; xbits: isset, urlentropy.detected; track ip_dst, expire 3600; target: dest_ip; sid:10; rev:1;)
alert http any any -> any any (msg:"High entropy in file name download"; file.name; content:".exe"; endswith; fast_pattern;  bsize:<21; lua:any.lua;   target: dest_ip; sid:11; rev:1;)
```

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3273](https://redmine.openinfosecfoundation.org/issues/3273)

Describe changes:
- New function to free per-thread keyword items
- Free per-thread keyword items when freeing outer lua context

PR #4740, opened 2020-03-28
detect/lua: Cleanup thread contextual info when needed

jlucovsky:
Continuation of #4716 

I used these rules -- the first rule demonstrated the issue. With this PR, memory is no longer overrun.
```
alert http any any -> any any (msg:"High entropy in file name download"; file.name; content:".exe"; endswith; fast_pattern;  bsize:<21; lua:any.lua; xbits: isset, urlentropy.detected; track ip_dst, expire 3600; target: dest_ip; sid:10; rev:1;)
alert http any any -> any any (msg:"High entropy in file name download"; file.name; content:".exe"; endswith; fast_pattern;  bsize:<21; lua:any.lua;   target: dest_ip; sid:11; rev:1;)
```

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3273](https://redmine.openinfosecfoundation.org/issues/3273)

Describe changes:
- Address review comments.
- New function to free per-thread keyword items
- Free per-thread keyword items when freeing outer lua context

### DIFF
--- a/src/detect-lua.h
+++ b/src/detect-lua.h
@@ -49,6 +49,8 @@ typedef struct DetectLuaData {
     uint32_t sid;
     uint32_t rev;
     uint32_t gid;
+
+    DetectEngineCtx *de_ctx;
 } DetectLuaData;
 
 #endif /* HAVE_LUA */
--- a/src/detect-engine.c
+++ b/src/detect-engine.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2019 Open Information Security Foundation
+/* Copyright (C) 2007-2020 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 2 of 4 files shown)


src/detect-lua.h
```diff
@@ -49,6 +49,8 @@ typedef struct DetectLuaData {
     uint32_t sid;
     uint32_t rev;
     uint32_t gid;
+
+    DetectEngineCtx *de_ctx;
```
victorjulien, line 53: Would be nice if we can avoid this, at least in the git master. See also https://github.com/OISF/suricata/pull/4716#discussion_r396017441

jlucovsky: The issue's occurring because the signature is deemed invalid after the lua keyword has been processed. I need to locate the previously registered item and mark it as invalid. The only way to do that (correct me if I'm wrong), is to locate it from the `de_ctx->keyword_list`. 

I could encode the `keyword_list` into a value returned by `DetectRegisterThreadCtxFuncs` or keep track of it in a container local to `detect-lua.c` ... I don't like either of these options. Are there others?

victorjulien: I think it would make sense to give this `Free` callback another arg: `de_ctx`. Then you can do the same thing as now w/o keeping this pointer here.

jlucovsky: The `Free` function of `SigTableElmt`?

victorjulien: Yeah

victorjulien:
https://redmine.openinfosecfoundation.org/issues/2737

jlucovsky:
Continued in #4838

PR #4741, opened 2020-03-28
Byteextractstr/part1/v1

inashivb:
`StringParse` functions error out in case of trailing characters after the extracted int, `ByteExtractString` functions ignore that.

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -45,6 +45,7 @@
 #include "util-pool.h"
 #include "util-radix-tree.h"
 #include "util-file.h"
+#include "util-byte.h"
 
 #include "stream-tcp-private.h"
 #include "stream-tcp-reassemble.h"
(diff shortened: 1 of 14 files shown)


src/app-layer-htp.c
```diff
@@ -2779,7 +2780,16 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
                 cfg_prec->randomize = ConfValIsTrue(p->val);
             }
         } else if (strcasecmp("randomize-inspection-range", p->name) == 0) {
-            uint32_t range = atoi(p->val);
+            uint32_t range;
+            if (StringParseUint32(&range, 10, 0,
```
victorjulien: Could compress to be something like
```c
if (!(StringParseUint32(&range, 10, 0, (const char *)p->val) == 0 && range <= 100)) {
    // error
```

src/detect-engine-iponly.c
```diff
@@ -261,7 +264,10 @@ static int IPOnlyCIDRItemParseSingle(IPOnlyCIDRItem *dd, const char *str)
                 goto error;
 
             /* Format is cidr val */
-            dd->netmask = atoi(mask);
+            if (StringParseUint8(&dd->netmask, 10, 0,
+                                       (const char *)mask) < 0) {
+                goto error;
+            }
```
victorjulien: Could you add validation while at it? Max value is 128.

src/detect-engine.c
```diff
@@ -2299,7 +2299,15 @@ static int DetectEngineCtxLoadConf(DetectEngineCtx *de_ctx)
             }
 
             if (insp_recursion_limit != NULL) {
-                de_ctx->inspection_recursion_limit = atoi(insp_recursion_limit);
+                if (StringParseInt32(&de_ctx->inspection_recursion_limit,
+                    10, 0, (const char *)insp_recursion_limit) < 0) {
```
victorjulien: indent is a bit off here. Can you make the 2nd line of the `StringParseInt32` indent a bit more so it doesn't align exactly with the next statement?

src/detect-id.c
```diff
@@ -150,7 +151,10 @@ static DetectIdData *DetectIdParse (const char *idstr)
     }
 
     /* ok, fill the id data */
-    temp = atoi((char *)tmp_str);
+    if (StringParseUint32(&temp, 10, 0, (const char *)tmp_str) < 0) {
```
victorjulien: check `temp` value here too?

inashivb:
Replaced with #4849

PR #4745, opened 2020-03-28

inashivb:
actions.

e.g.
to write two rules,
```
```

```
```

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -48,10 +48,14 @@
 #define PARSE_REGEX         "^([a-z]+)(?:,\\s*(.*))?"
 static DetectParseRegex parse_regex;
 
+#define MAX_TOKENS 100
+
 int DetectFlowbitMatch (DetectEngineThreadCtx *, Packet *,
         const Signature *, const SigMatchCtx *);
 static int DetectFlowbitSetup (DetectEngineCtx *, Signature *, const char *);
+static char **StrSplit(char *arrptr, const char *delim);
 void DetectFlowbitFree (void *);
+void FlowBitArrayFree(int, char **);
 void FlowBitsRegisterTests(void);
 
 void DetectFlowbitsRegister (void)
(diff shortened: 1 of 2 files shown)


src/detect-flowbits.c
```diff
@@ -69,6 +73,37 @@ void DetectFlowbitsRegister (void)
     DetectSetupParseRegexes(PARSE_REGEX, &parse_regex);
 }
 
+static char **StrSplit(char *arrptr, const char *delim)
+{
+    char **strarr = NULL;
+    char *token, *saveptr;
```
victorjulien: I don't know if its strictly necessary, but I'd sleep better if `saveptr` was initialized to `NULL` here.

src/detect-flowbits.c
```diff
@@ -243,14 +293,33 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
...
-
-    SCLogDebug("idx %" PRIu32 ", cmd %s, name %s",
-        cd->idx, fb_cmd_str, strlen(fb_name) ? fb_name : "(none)");
-
+    if (strchr(fb_name, '|') != NULL) {
+        fb_items = StrSplit(fb_name, "|");
+        if (fb_items == NULL) {
+            goto error;
+        }
+        for (i = 0; fb_items[i] != NULL ; i++) {
```
victorjulien: Since the goal here is to count, I would like to see a comment about this. It's too easy to overlook the counting as this looks like a simple debug loop

src/detect-flowbits.c
```diff
@@ -243,14 +293,33 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
...
+        cd->or_list_size = i;
+        cd->or_list = SCMalloc(cd->or_list_size * sizeof(uint32_t));
+        if (unlikely(cd->or_list == NULL))
+            goto error;
+        for (uint8_t j = 0; j < cd->or_list_size ; j++) {
+            cd->or_list[j] = VarNameStoreSetupAdd(fb_items[j], VAR_TYPE_FLOW_BIT);
+            de_ctx->max_fb_id = MAX(cd->or_list[j], de_ctx->max_fb_id);
+        }
+        cd->cmd = fb_cmd;
+        FlowBitArrayFree(cd->or_list_size, fb_items);
```
victorjulien: seeing this freeing here of potentially 101 memory allocations, I wonder if any of them are necessary. From the `strtok_r` loop in `StrSplit` you could fill a static array of id's by directly calling `VarNameStoreSetupAdd` on each of them, and then ending with a allocation+copy of that static array info the `cd` struct.

inashivb:
Replaced with #4762

PR #4748, opened 2020-03-29
rfb: make sure size calculations do not overflow

satta:
Addresses #3570 by extra checking of calculated size requests.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#3570](https://redmine.openinfosecfoundation.org/issues/3570)

Describe changes:
- Add more extensive checking size request calculations for `AppLayerResult::incomplete()`. This fixes a bug caused by an unchecked integer overflow; brought to light by fuzzing.

### DIFF
--- a/rust/src/rfb/rfb.rs
+++ b/rust/src/rfb/rfb.rs
@@ -100,6 +100,18 @@ pub struct RFBState {
     state: parser::RFBGlobalState
 }
 
+#[inline]
+fn handle_incomplete(input: &[u8], current: &[u8], nom_needed: usize) -> AppLayerResult {
+    if let Some(consumed) = input.len().checked_sub(current.len()) {
+        if let Some(needed) = current.len().checked_add(nom_needed) {
+            if consumed <= (std::u32::MAX as usize) && needed <= (std::u32::MAX as usize) {
+                return AppLayerResult::incomplete(consumed as u32, needed as u32);
+            }
+        }
+    }
+    return AppLayerResult::err();
+}
+
 impl RFBState {
     pub fn new() -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


rust/src/rfb/rfb.rs
```diff
@@ -100,6 +100,18 @@ pub struct RFBState {
     state: parser::RFBGlobalState
 }
 
+#[inline]
+fn handle_incomplete(input: &[u8], current: &[u8], nom_needed: usize) -> AppLayerResult {
```
victorjulien: if this `nom_needed` would take `nom::Needed` you could reduce the code at each of the callers.

They could simply be:
```rust
Err(nom::Err::Incomplete(v)) => {
   return handle_incomplete(input, current, v);
}
```
The `if let nom::Needed::Size(n) = v {` could be done once here in `handle_incomplete'
```

satta:
> In a case like this could you briefly explain what sequence of events led to the overflow?

Sure. With the given input, the parser eventually arrived at `parser::parse_failure_reason()` which parsed from the remaining four bytes (describing the string length) that the failure string to follow would be 4294967295 bytes long. While calculating the total size of the data to request via `AppLayerResult::incomplete()`, adding the four bytes for the parsed but not consumed sting length caused the u32 length to overflow, resulting in a much smaller value triggering the bug condition.

This problem was addressed by more careful checking of values in each stet that could overflow: one subtraction, one addition (which could overflow the `usize` length values, and a final check to determine whether the result still fit into the `u32` values required by `AppLayerResult::incomplete()`. If so, we would safely convert the values and pass them to the result type. Of not, we simply return `AppLayerResult::err()` but do not erroneously and silently request the wrong amount.

rust/src/rfb/rfb.rs
satta: True. Will look into it.

victorjulien:
Thanks @satta, could you add this explanation to the git message?

satta:
> Thanks @satta, could you add this explanation to the git message?

Absolutely. Will do so in next PR.

satta:
Next PR: #4755

PR #4750, opened 2020-03-30
build: wrap fuzz targets in guard to fix 'make tags'

victorjulien:
Passed


catenacyber:
Should we add `make tags` to the CI also ?

PR #4527, opened 2020-02-10
Expectation fix v1.3

regit:
This PR adds a limitation for expectation number for IPPair and brings some cleaning methods. 

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3378

Describe changes:
- Use circle list for expectations
- Limit number of expectations per IP pair
- Clean expectation at flow end

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/520
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/296

PR #4752, opened 2020-03-30
Expectation fix v1.4

regit:
Update of #4527 

Describe changes:
- Fix coverity warning
- Fix scan-build warning

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/522
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/298

### DIFF
--- a/src/app-layer-expectation.c
+++ b/src/app-layer-expectation.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2017 Open Information Security Foundation
+/* Copyright (C) 2017-2020 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
--- a/src/app-layer-expectation.c
+++ b/src/app-layer-expectation.c
@@ -61,6 +61,7 @@
 #include "app-layer-expectation.h"
 
 #include "util-print.h"
+#include "queue.h"
 
 static int g_expectation_id = -1;
 static int g_expectation_data_id = -1;
@@ -68,6 +69,7 @@ static int g_expectation_data_id = -1;
 SC_ATOMIC_DECLARE(uint32_t, expectation_count);
 
 #define EXPECTATION_TIMEOUT 30
+#define EXPECTATION_MAX_LEVEL 10
 
 typedef struct Expectation_ {
     struct timeval ts;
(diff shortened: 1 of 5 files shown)


src/app-layer-expectation.c
```diff
@@ -85,6 +89,12 @@ typedef struct ExpectationData_ {
     void (*DFree)(void *);
 } ExpectationData;
 
+typedef struct ExpectationList_ {
+    Expectation *cqh_first; /* first element  */
```
victorjulien: shouldn't `CIRCLEQ_HEAD` be used here?

src/app-layer-expectation.c
```diff
@@ -157,11 +181,30 @@ static Expectation *AppLayerExpectationLookup(Flow *f, int direction, IPPair **i
...
+
+static ExpectationList *AppLayerExpectationRemove(IPPair *ipp,
+                                                  ExpectationList *exp_list,
+                                                  Expectation *exp)
+{
+    CIRCLEQ_REMOVE(exp_list, exp, entries);
+    AppLayerFreeExpectation(exp);
+    SC_ATOMIC_SUB(expectation_count, 1);
+    exp_list->length--;
+    if (exp_list->length <= 0) {
```
victorjulien: `length` is `uint8_t`, so `< 0` check is useless.

src/flow-manager.c
```diff
@@ -983,6 +984,8 @@ static TmEcode FlowRecycler(ThreadVars *th_v, void *thread_data)
 
                 (void)OutputFlowLog(th_v, ftd->output_thread_data, f);
 
+                if (f->flags & FLOW_HAS_EXPECTATION)
```
victorjulien: Can you move this to `FlowClearMemory`, for two reasons:

- just cleaner
- this will cover a case that is currently missed as well: `FlowGetUsedFlow`

src/app-layer-expectation.c
regit: Good catch.

src/app-layer-expectation.c
regit: We have
```
#define CIRCLEQ_HEAD(name, type)					\
struct name {								\
	struct type *cqh_first;		/* first element */		\
	struct type *cqh_last;		/* last element */		\
}
```
And I wanted to add the length variable to the structure, so to avoid to add one layer in the structure, I did decide to directly use the directly fields in the ExpectationList structure.

victorjulien: Whats the reason for avoiding this extra layer? I don't think it matters at all after compilation. Seem just cleaner to me to use the api as intended.

PR #4753, opened 2020-03-30
detect/bsize: Ensure max u64 values can be parsed

jlucovsky:
This commit increases the size of temporary containers used to hold
values. Since the values may be unsigned 64 values, ensure that the
containers are large enough to hold the max sized unsigned 64 bit value
in ascii.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3567](https://redmine.openinfosecfoundation.org/issues/3567)

Describe changes:
- Ensure temp. containers are large enough (size = 21) to hold max u64 value

### DIFF
--- a/src/detect-bsize.c
+++ b/src/detect-bsize.c
@@ -43,6 +43,8 @@ static void DetectBsizeFree (void *);
 static void DetectBsizeRegisterTests (void);
 #endif
 
+/* Number of ascii characters required to hold max u64 string */
+#define U64_MAXSTR_SIZE 21
 /**
  * \brief Registration function for bsize: keyword
  */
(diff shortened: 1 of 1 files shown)


jlucovsky:
> I don't understand the issue, or the fix. If the issue was that a buffer was overrun, would increasing the buffer size not just hide the issue? Isn't the real issue that a buffer size isn't properly validated?

The problem occurred due to an overrun of the 11-byte array during a downstream `strlen` call. Since the container is supposed to be able to contain a 64-bit value, the size was increased.

In this case, the value used is `10000000001` (11 bytes). 

It can be fixed by validating the size of the value before it's put into the (11-byte) container.

Either way.

victorjulien:
My concern is that if previously it was possible to generate input that caused strlen to freak out (presumably by passing a buffer w/o nul termination to it), wouldn't this be still possible with a larger buffer? It seems like what we were missing is proper input validation.

jlucovsky:
I'll revisit this.

PR #4756, opened 2020-03-30
Add wrappers for validating extracted int is in range

inashivb:
Wrappers for validating the extracted int values are in the provided range.

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -65,6 +65,7 @@
 #include "util-debug.h"
 #include "util-time.h"
 #include "util-misc.h"
+#include "util-byte.h"
 
 #include "util-unittest.h"
 #include "util-unittest-helper.h"
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -2779,11 +2779,11 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
                 cfg_prec->randomize = ConfValIsTrue(p->val);
             }
         } else if (strcasecmp("randomize-inspection-range", p->name) == 0) {
-            uint32_t range = atoi(p->val);
-            if (range > 100) {
-                SCLogError(SC_ERR_SIZE_PARSE, "Invalid value for randomize"
-                           " inspection range setting from conf file - %s."
-                           " It should be inferior to 100."
+            uint32_t range;
+            if (StringParseUint32wValidation(&range, 10, 0, (const char *)p->val, 0, 99) < 0) {
+                SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for randomize"
+                           "-inspection-range setting from conf file - \"%s\"."
+                           " It should be a valid integer less than 100."
                            " Killing engine",
                            p->val);
                 exit(EXIT_FAILURE);
--- a/src/util-byte.c
+++ b/src/util-byte.c
@@ -360,6 +360,107 @@ int StringParseUint8(uint8_t *res, int base, uint16_t len, const char *str)
 
     return ret;
 }
+
+int StringParseUint64wValidation(uint64_t *res, int base, uint16_t len, const char *str,
+                                 uint64_t min, uint64_t max)
+{
+    int ret;
+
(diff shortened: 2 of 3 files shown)


inashivb:
@victorjulien how does this look to you? This is a sample. Same validation functions would also have to be written for ByteExtractString* functions, I guess.

src/util-byte.c
```diff
@@ -360,6 +360,107 @@ int StringParseUint8(uint8_t *res, int base, uint16_t len, const char *str)
...
+                                 uint64_t min, uint64_t max)
+{
+    int ret;
+
+    ret = ByteExtractString(res, base, len, str, true);
+    if (ret <= 0) {
+        return ret;
+    }
+
+    if ((uint64_t)ret < min || (uint64_t)ret > max) {
```
victorjulien: `ret` is int, so it can't hold the full value range of `uint64_t`

victorjulien: Should that be `*res` instead?

src/app-layer-htp.c
```diff
@@ -2779,11 +2780,11 @@ static void HTPConfigParseParameters(HTPCfgRec *cfg_prec, ConfNode *s,
...
             }
         } else if (strcasecmp("randomize-inspection-range", p->name) == 0) {
-            uint32_t range = atoi(p->val);
-            if (range > 100) {
-                SCLogError(SC_ERR_SIZE_PARSE, "Invalid value for randomize"
-                           " inspection range setting from conf file - %s."
-                           " It should be inferior to 100."
+            uint32_t range;
+            if (StringParseUint32wValidation(&range, 10, 0, (const char *)p->val, 0, 99) < 0) {
+                SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for randomize"
```
victorjulien: Since we error here and in `StringParseUint32wValidation` meaning we get 2 error messages per single issue. Maybe leave this one and silence the one in `StringParseUint32wValidation`?

src/util-byte.c
```diff
@@ -360,6 +360,107 @@ int StringParseUint8(uint8_t *res, int base, uint16_t len, const char *str)
...
+                                 uint32_t min, uint32_t max)
+{
+    uint64_t i64;
+
+    int ret = ByteExtractString(&i64, base, len, str, true);
+    if (ret <= 0) {
+        return ret;
+    }
+
+    if ((uint32_t)ret < min || (uint32_t)ret > max) {
```
victorjulien: `*res`

victorjulien:
@jasonish @jlucovsky if you have ideas on the func naming...

src/util-byte.c
inashivb: Yes, my bad. It should be *res everywhere.

src/app-layer-htp.c
inashivb: Right. OK.

jlucovsky:
> @jasonish @jlucovsky if you have ideas on the func naming...

The general function name pattern is `StringParse<type>wValidation`

For type, I suggest using `U32, I32, ...` as a more succinct way of expressing `Uint32`, etc.

Instead of `wValidation`, I suggest `RangeCheck`.

Thus, names would be `StringParse{U64,I64,U32,I32,U16,I16,U8,I8}RangeCheck`

inashivb:
> > @jasonish @jlucovsky if you have ideas on the func naming...
> 
> The general function name pattern is `StringParse<type>wValidation`
> 
> For type, I suggest using `U32, I32, ...` as a more succinct way of expressing `Uint32`, etc.
> 
> Instead of `wValidation`, I suggest `RangeCheck`.
> 
> Thus, names would be `StringParse{U64,I64,U32,I32,U16,I16,U8,I8}RangeCheck`

Thanks, Jeff. @jasonish do you have something to add as well?

inashivb:
Hi @victorjulien and @jasonish ! Do we have consensus over the function name that Jeff suggested then?

jasonish:
I like Jeff's idea.  Inline what Rust types as well.

victorjulien:
> I like Jeff's idea. Inline what Rust types as well.

Same here. Lets go for Jeff's suggestion.

PR #4758, opened 2020-03-30
RFC: Modular file reorg - v1

jasonish:
Example of what a modular file reorg might look like.

Example modules:
- dnp3: app-layer/dnp3
- af-packet: source/af-packet

The idea is to group all parts of an app-layer together (parser, detect,
logger, etc) into a module.

Or for af-packet, bundle the runmode and the actual sourcing of packets
into a module.

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/459
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/816


jlucovsky:
So `app-layer` means `"protocol"`?

What would the other top-level directories w/in source be?

jasonish:
> So `app-layer` means `"protocol"`?

More like application layer protocol, as it won't be protocols like TCP, GRE, etc.

> What would the other top-level directories w/in source be?

Not too sure.  I think it would be a mix of what you see here, and what you see in the other re-org branch.  If something kind of feels like it could be a plugin (an app-layer parser, along with its detect and logging) then I think it makes sense to bundle it together like I did with DNP3 here.  Sources are much the same.

But not sure about all the detect-engine stuff, misc util and what not.  Perhaps that could be like the other branch and be broken into /detect-engine/, /util/.  Part of the idea with this module approach is we can tackle it module by module, and try not to handle it one sweep over the code.

jlucovsky:
I like this approach.

victorjulien:
I like the new structure, a lot. But I worry about being unable to trace back the history of changes. I tried a few things with `tig` and `git log` and both seem unable to show anything across this 'move' boundary. Does anyone know tricks around this? (this most likely applies to any reorg)

jasonish:
> But I worry about being unable to trace back the history of changes. I tried a few things with `tig` and `git log` and both seem unable to show anything across this 'move' boundary. Does anyone know tricks around this? (this most likely applies to any reorg)

No really, especially not when running log on an individual file.

A diff across the whole repo seems to be a bit smarter though as it only shows you the diff even if a move was detected.
```
diff --git a/src/runmode-af-packet.c b/src/source/af-packet/runmode.c
similarity index 99%
rename from src/runmode-af-packet.c
rename to src/source/af-packet/runmode.c
index 05761a788..a3647b27c 100644
--- a/src/runmode-af-packet.c
+++ b/src/source/af-packet/runmode.c
@@ -35,14 +35,13 @@
 #include "tm-threads.h"
 #include "conf.h"
 #include "runmodes.h"
-#include "runmode-af-packet.h"
+#include "./runmode.h"
 #include "output.h"
 #include "log-httplog.h"
 #include "detect-engine-mpm.h"
 
 #include "alert-fastlog.h"
 #include "alert-prelude.h"
-#include "alert-unified2-alert.h"
 #include "alert-debuglog.h"
 
 #include "flow-bypass.h"
@@ -56,7 +55,7 @@
 #include "util-ioctl.h"
 #include "util-ebpf.h"
 
-#include "source-af-packet.h"
+#include "./source.h"
 
 extern int max_pending_packets;
 ```

And other than the rename in the git log summary, I don't see any other sort of tracking.

victorjulien:
I've looked into this a bit more, and it seems we're good. The 3 things I use often are 'plain' git log, git log -L and tig:
* `git log` -> `git log --follow`
* `git log -L ...` already works fine
* `tig` -> `tig --follow`

victorjulien:
As discussed elsewhere, we're moving this to the early dev cycle for 7.

PR #4764, opened 2020-03-31

inashivb:
actions.

e.g.
to write two rules,
```
```

```
```

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -48,9 +48,12 @@
 #define PARSE_REGEX         "^([a-z]+)(?:,\\s*(.*))?"
 static DetectParseRegex parse_regex;
 
+#define MAX_TOKENS 100
+
 int DetectFlowbitMatch (DetectEngineThreadCtx *, Packet *,
         const Signature *, const SigMatchCtx *);
 static int DetectFlowbitSetup (DetectEngineCtx *, Signature *, const char *);
+static int FlowbitOrAddData(char *, DetectFlowbitsData *, DetectEngineCtx *);
 void DetectFlowbitFree (void *);
 void FlowBitsRegisterTests(void);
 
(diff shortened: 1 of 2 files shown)


src/detect-flowbits.c
```diff
@@ -364,37 +423,76 @@ void DetectFlowbitsAnalyze(DetectEngineCtx *de_ctx)
...
-                            uint32_t old_size = array[fb->idx].isset_sids_size;
-                            uint32_t new_size = MAX(2 * old_size, MAX_SIDS);
-
-                            void *ptr = SCRealloc(array[fb->idx].isset_sids, new_size * sizeof(uint32_t));
-                            if (ptr == NULL)
-                                goto end;
-                            array[fb->idx].isset_sids_size = new_size;
-                            array[fb->idx].isset_sids = ptr;
+                    // Handle flowbit array in case of ORed flowbits
+                    if (fb->or_list_size > 0) {
```
jlucovsky: This `if` isn't necessary ... when the value is `0`, the `for` loop test will cause the loop body to be skipped.

jlucovsky: This is pre-existing, but there's opportunity for refactoring in these code blocks.

src/detect-flowbits.c
```diff
@@ -69,6 +72,42 @@ void DetectFlowbitsRegister (void)
...
 
+static int FlowbitOrAddData(char *arrptr, DetectFlowbitsData *cd, DetectEngineCtx *de_ctx)
+{
+    char *strarr[MAX_TOKENS];
+    char *token;
+    char *saveptr = NULL;
+    uint8_t i = 0;
+
+    if (strchr(arrptr, ' ') != NULL) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Space not allowed in flowbits.");
```
victorjulien: "spaces are not allowed in flowbit-names" or something?

inashivb:
Replaced with #4768

PR #4767, opened 2020-04-01
output-json: add MAC addresses to EVE-JSON logs - v8

satta:
Previous PR: #4709  

- Make sure that `p->ethh` is not NULL.
- Remove ethertype field from packet context output for consistency with flow context output.
- Fix spelling issue.
- Add some explanatory text to the commit message.

### DIFF
--- a/src/flow.h
+++ b/src/flow.h
@@ -29,6 +29,7 @@
 #include "util-atomic.h"
 #include "util-device.h"
 #include "detect-tag.h"
+#include "util-macset.h"
 #include "util-optimize.h"
 
 /* Part of the flow structure, so we declare it here.
@@ -475,6 +476,8 @@ typedef struct Flow_
     uint32_t tosrcpktcnt;
     uint64_t todstbytecnt;
     uint64_t tosrcbytecnt;
+
+    MacSet *macset;
 } Flow;
 
 enum FlowState {
--- a/src/util-macset.c
+++ b/src/util-macset.c
@@ -1,0 +1,398 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/**
+ * \file
+ *
+ * \author Sascha Steinbiss <sascha.steinbiss@dcso.de>
+ *
+ * Set-like data store for MAC addresses. Implemented as array for memory
+ * locality reasons as the expected number of items is typically low.
+ *
+ */
+
+#include "suricata-common.h"
+#include "suricata.h"
+#include "flow-util.h"
+#include "flow-private.h"
+#include "util-macset.h"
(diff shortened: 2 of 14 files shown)


satta:
These failures do not seem to be related to the code. Can someone restart them maybe?

satta:
Uh oh. There seems to be a bigger issue with the CI checkouts.

satta:
I'm going to push to this PR again as I did before, that helped.

src/util-macset.c
```diff
@@ -0,0 +1,398 @@
...
+
+void MacSetReset(MacSet *ms)
+{
+    if (ms == NULL)
+        return;
+    ms->state[MAC_SET_SRC] = ms->state[MAC_SET_DST] = EMPTY_SET;
+    ms->last[MAC_SET_SRC] = ms->last[MAC_SET_DST] = 0;
+}
+
+void MacSetFree(MacSet *ms)
```
jlucovsky, line 221: Since all macset's are created with `MacSetInit(10)`, I suggest that one call to `SC_ATOMIC_SUB` be made with the total size being returned.

satta: We could also make this number configurable?

jlucovsky: What would be the advantage of doing that?

src/flow.h
```diff
@@ -475,6 +476,8 @@ typedef struct Flow_
     uint32_t tosrcpktcnt;
     uint64_t todstbytecnt;
     uint64_t tosrcbytecnt;
+
+    MacSet *macset;
```
victorjulien, line 480: Since this behavior is optional, we're wasting 8 bytes per flow here if the feature is not used. Maybe this would be a good case for the `FlowStorage` API?

src/util-macset.c
satta: It would allow to tune memory usage against expected variability of MAC addresses. I assume that it it unlikely to have more than 10 MAC addresses per side and flow, but if you know that there will never be more than _n_ then you would be able to configure Suricata accordingly. Just an idea.

Apart from that, I am not sure I understand what you mean. I need to consider both `ms->buf`s because they are allocated independently of each other, so I can only add/subtract their memory usage when I know they are in use. Also, they are only allocated when a second MAC address per flow is seen for the first time, so I need to actually check whether I need to subtract the allocated bytes from the global memory usage. I agree that this can be calculated locally within `MacSetFree` and only one call to `SC_ATOMIC_SUB` can be made. Is that what you meant?

victorjulien: It might be nice to be able to track (in a stats entry perhaps) what the max number of MAC addresses per side is in some real world setups. Then we can adjust the default accordingly.

satta: Sure. I can introduce a counter for that.

src/flow.h
satta: I'll take a look.

src/util-macset.c
satta: Done, will have `decoder.max_mac_addrs_src` and `decoder.max_mac_addrs_dst` in next PR.

src/flow.h
satta: Hmm. So far only _logging_ of these values has been optional (via the `outputs.eve-log.ethernet` config item). Note that this can be disabled in one EVE output and enabled in another. Bookkeeping of the addresses is always performed up to now.

I have started to implement a solution that would check it at least one of the defined (enabled) EVE-JSON outputs has their `ethernet` setting set to true, and only then register a flow storage item for the `MacSet`s. All `MacSet*` functions would no longer directly be given a `MacSet*` pointer but a `Flow*`, and if there is no registered flow storage for `MacSet`s, all functions would immediately return without any action.
Any comments on this idea?

jlucovsky: Perhaps `ProvidesFeature` and `RequiresFeature` could be used here when `ehternet` logging is configured.

This is used currently in `output-filestore.c` and `detect-filestore.c` ... The former signals the feature is available and the second indicates the dependency if a rule is seen that requires it.

satta:
New PR: #4968

PR #4768, opened 2020-04-02

inashivb:
actions.

e.g.
to write two rules,
```
```

```
```

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -48,9 +48,12 @@
 #define PARSE_REGEX         "^([a-z]+)(?:,\\s*(.*))?"
 static DetectParseRegex parse_regex;
 
+#define MAX_TOKENS 100
+
 int DetectFlowbitMatch (DetectEngineThreadCtx *, Packet *,
         const Signature *, const SigMatchCtx *);
 static int DetectFlowbitSetup (DetectEngineCtx *, Signature *, const char *);
+static int FlowbitOrAddData(char *, DetectFlowbitsData *, DetectEngineCtx *);
 void DetectFlowbitFree (void *);
 void FlowBitsRegisterTests(void);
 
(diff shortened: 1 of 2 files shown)


src/detect-flowbits.c
```diff
@@ -295,10 +351,10 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
 void DetectFlowbitFree (void *ptr)
 {
     DetectFlowbitsData *fd = (DetectFlowbitsData *)ptr;
-
     if (fd == NULL)
         return;
-
+    if (fd->or_list != NULL)
```
victorjulien, line 356: scan-build-9 doesn't like something here:
```
   CC       detect-flowbits.o
detect-flowbits.c:356:21: warning: The left operand of '!=' is a garbage value
    if (fd->or_list != NULL)
        ~~~~~~~~~~~ ^
1 warning generated.
```

src/detect-flowbits.c
```diff
@@ -243,14 +292,21 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
     cd = SCMalloc(sizeof(DetectFlowbitsData));
```
victorjulien, line 292: suggest using `SCCalloc`, this initializes the memory. I think this should fix the scan-build issue below.

src/detect-flowbits.c
```diff
@@ -69,6 +72,39 @@ void DetectFlowbitsRegister (void)
...
+            SCLogError(SC_ERR_INVALID_SIGNATURE, "Number of flowbits exceeds "
+                       "maximum allowed: %d.", MAX_TOKENS);
+            return -1;
+        }
+        strarr[i] = token;
+        arrptr = NULL;
+        i++;
+    }
+
+    cd->or_list = SCMalloc(i * sizeof(uint32_t));
```
victorjulien, line 97: suggest first setting `cd->or_list_size = i` then passing `cd->or_list_size` to SCMalloc. Result should be the same, but I prefer making it very clear how size vars map to allocations.

victorjulien: another suggestion: `SCCalloc(cd->or_list_size, sizeof(uint32_t));`

inashivb:
Replaced with #4778

PR #4770, opened 2020-04-02
4.1.x - No Rust fixups - v1

jasonish:
First enable a 4.1.x build on GitHub CI without Rust.

Second, fix the address order of the non-Rust DNS logger to match
that of Rust.

Combined with Suricata-Verify PR https://github.com/OISF/suricata-verify/pull/208,
no Rust 4.1.x builds should pass suricata-verify.

- https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/818
- https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/461

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -323,6 +323,59 @@ jobs:
       - name: Running suricata-verify
         run: python3 ./suricata-verify/run.py
 
+  ubuntu-18-04-norust:
+    name: Ubuntu 18.04 (no Rust)
+    runs-on: ubuntu-18.04
+    container: ubuntu:18.04
+    steps:
+
+      - name: Install dependencies
+        run: |
+          apt update
+          apt -y install \
+                libpcre3 \
+                libpcre3-dev \
+                build-essential \
+                autoconf \
+                automake \
+                git \
+                jq \
+                libtool \
+                libpcap-dev \
+                libnet1-dev \
+                libyaml-0-2 \
+                libyaml-dev \
+                libcap-ng-dev \
+                libcap-ng0 \
+                libmagic-dev \
+                libnetfilter-queue-dev \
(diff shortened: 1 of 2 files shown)


victorjulien:
Previously only the no-rust checked failed, but it gave no output. I restarted the build, but then got the current failures

jasonish:
> Previously only the no-rust checked failed, but it gave no output. I restarted the build, but then got the current failures

Yeah, this one should fail in suricata-verify like seen here: https://github.com/jasonish/suricata/runs/553497311?check_suite_focus=true#step:11:636

They have the same hash, wonder why this happens. Will for a change in ID and repush.

jasonish:
Ok, just the expected build fails now.

victorjulien:
Is the dns issue a real bug that we need to fix, or more a behavior change that should go into a major release? I suppose for 5.0 we fixed it in 5.0.1 so there we considered it a bug. Asking because I fear this could break things for existing users?

Btw the ticket that the commit references it a 5.0 ticket. Can you create a backport ticket and reference that instead?

jasonish:
I'm ok with *not* fixing this in 4.1.x..  4.1.x is more set in stone than 5.0 was at the time, so maybe its better left as is.

The failing SV test can be marked as requiring Rust instead.

And maybe you can cherry-pick fc33e88f8257a2017c6ea83fa36470b644366b52?

victorjulien:
I think I would like to fix this. Went back and forth a bit in my mind, but I think the original ticket for 4.1.6 (https://redmine.openinfosecfoundation.org/issues/3355) actually meant to fix this properly. I messed it up by simply cherry-picking your patch from the master, which had already removed the C version of the DNS parser.

victorjulien:
Merged in #4810, thanks Jason!

PR #4778, opened 2020-04-04

inashivb:
actions.

e.g.
to write two rules,
```
```

```
```

Note: s-v PR coming up

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -48,9 +48,12 @@
 #define PARSE_REGEX         "^([a-z]+)(?:,\\s*(.*))?"
 static DetectParseRegex parse_regex;
 
+#define MAX_TOKENS 100
+
 int DetectFlowbitMatch (DetectEngineThreadCtx *, Packet *,
         const Signature *, const SigMatchCtx *);
 static int DetectFlowbitSetup (DetectEngineCtx *, Signature *, const char *);
+static int FlowbitOrAddData(char *, DetectFlowbitsData *, DetectEngineCtx *);
 void DetectFlowbitFree (void *);
 void FlowBitsRegisterTests(void);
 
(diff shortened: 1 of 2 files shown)


src/detect-flowbits.c
```diff
@@ -69,6 +72,39 @@ void DetectFlowbitsRegister (void)
...
 }
 
+static int FlowbitOrAddData(char *arrptr, DetectFlowbitsData *cd, DetectEngineCtx *de_ctx)
+{
+    char *strarr[MAX_TOKENS];
+    char *token;
+    char *saveptr = NULL;
+    uint8_t i = 0;
+
+    if (strchr(arrptr, ' ') != NULL) {
```
victorjulien, line 82: does this mean that `flowbits: abc | def;` is also rejected? I would say leading and trailing spaces would be ok, as long as they are stripped from the actual name. So in my example the names would be `abc` and `def`.

src/detect-flowbits.c
```diff
@@ -69,6 +72,39 @@ void DetectFlowbitsRegister (void)
     DetectSetupParseRegexes(PARSE_REGEX, &parse_regex);
 }
 
+static int FlowbitOrAddData(char *arrptr, DetectFlowbitsData *cd, DetectEngineCtx *de_ctx)
```
victorjulien, line 75: nit: convention in other code is to have `de_ctx` as the first arg, the `cd` following that and the more specific args(s) last.

src/detect-flowbits.c
inashivb: I was trying to do this and turns out the way flowbits are parsed, spaces will not be allowed in flowbits (https://github.com/OISF/suricata/blob/master/src/detect-flowbits.c#L185) as from what I understood the parser is treating everything that follows `flowbits:` as one flowbit name. Just to make sure, am I supposed to change how pcre parser evaluates the flowbit names and then change the way the entire processing is carried out thereafter?
Please let me know if I have misunderstood something.

victorjulien: I wasn't trying to suggest to support spaces within the names themselves, but only to allow leading and trailing spaces before and after the names. For regular flowbits you can see they are trimmed https://github.com/OISF/suricata/blob/master/src/detect-flowbits.c#L179 (the leading spaces are handled in the parse regex).

To give an example, I think:
`flowbits:isset, a | b | c;`
Should be equivalent to:
`flowbits:isset,a|b|c;`

Does that make sense?

inashivb: Yes, thank you for clarification.

inashivb:
Replaced with #4786

PR #4781, opened 2020-04-06
detect/ssl: Fix memory leak in version parsing

jlucovsky:
This commit fixes a memory leak in the SSL version handling that
manifests when the version identifier is incomplete or incorrect.

Redmine issue: [3566](https://redmine.openinfosecfoundation.org/issues/3566)

Describe changes:
- Use `pcre_copy_substring` to avoid incomplete error exits.


victorjulien:
Merged in #4784, thanks Jeff

PR #4786, opened 2020-04-07

inashivb:
actions.

e.g.
to write two rules,
```
```

```
```

S-V PR: https://github.com/OISF/suricata-verify/pull/211

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -48,9 +48,12 @@
 #define PARSE_REGEX         "^([a-z]+)(?:,\\s*(.*))?"
 static DetectParseRegex parse_regex;
 
+#define MAX_TOKENS 100
+
 int DetectFlowbitMatch (DetectEngineThreadCtx *, Packet *,
         const Signature *, const SigMatchCtx *);
 static int DetectFlowbitSetup (DetectEngineCtx *, Signature *, const char *);
+static int FlowbitOrAddData(DetectEngineCtx *, DetectFlowbitsData *, char *);
 void DetectFlowbitFree (void *);
 void FlowBitsRegisterTests(void);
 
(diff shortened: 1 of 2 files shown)


src/detect-flowbits.c
```diff
@@ -69,6 +72,49 @@ void DetectFlowbitsRegister (void)
...
+static int FlowbitOrAddData(DetectEngineCtx *de_ctx, DetectFlowbitsData *cd, char *arrptr)
+{
+    char *strarr[MAX_TOKENS];
+    char *token;
+    char *saveptr = NULL;
+    uint8_t i = 0;
+
+    while ((token = strtok_r(arrptr, "|", &saveptr))) {
+        // Check for leading/trailing spaces in the token
+        while(isspace((unsigned char)*token)) token++;
```
victorjulien, line 84: please have the loop body on a new line
```c
while(isspace((unsigned char)*token))
    token++;
```

src/detect-flowbits.c
```diff
@@ -69,6 +72,49 @@ void DetectFlowbitsRegister (void)
...
+        if (*token == 0)
+            continue;
+        char *end = token + strlen(token) - 1;
```
victorjulien, line 87: Why the `-1`?

src/detect-flowbits.c
```diff
@@ -69,6 +72,49 @@ void DetectFlowbitsRegister (void)
...
+        while(end > token && isspace((unsigned char)*end)) end--;
+        end[1] = '\0';
```
victorjulien, line 89: Could these be written as
```c
while(end > token && isspace((unsigned char)*end))
   *(end--) = '\0';
```
?

src/detect-flowbits.c
inashivb: From what I understood, if we have the following string (`.` represents space char) with char addresses written below each char,
```
a b c . . .
1 2 3 4 5 6
^
token
```
then `token + strlen(token) = 1 + 6 = 7` but the last character is at `6` hence the `-1`.
Is there something wrong?

inashivb:
Replaced with #4847

PR #4792, opened 2020-04-08
rfb: handles incomplete asking for one more byte

catenacyber:

Describe changes:
- Handles incomplete better for RFB parser (simply asking for one more byte at a time)

`parse_supported_security_types` returns `nom::Err::Incomplete(nom::Needed::Size(2))` if we give it a one byte input (whose value is 2)
when `parse_server_security_type` returns `nom::Err::Incomplete(nom::Needed::Size(4))` if we give it a one byte input

In the first case we need two additional bytes to the first byte
In the second case, we need a total of 4 bytes, including the first one we got

> Rust nom doesn't give you a usable number especially when their is a chain of parsers

cc @satta 
I think the behavior right now is not ok, but I am not sure about the best fix.

### DIFF
--- a/rust/src/rfb/rfb.rs
+++ b/rust/src/rfb/rfb.rs
@@ -106,7 +106,7 @@ fn handle_incomplete(input: &[u8], current: &[u8], nom_needed: nom::Needed) -> A
         if let Some(consumed) = input.len().checked_sub(current.len()) {
             if let Some(needed) = current.len().checked_add(needed_size) {
                 if consumed <= (std::u32::MAX as usize) && needed <= (std::u32::MAX as usize) {
-                    return AppLayerResult::incomplete(consumed as u32, needed as u32);
+                    return AppLayerResult::incomplete(consumed as u32, (current.len() + 1) as u32);
                 }
             }
         }


satta:
I see -- the reason here is that `parse_supported_security_types` parses a variable length value: in your case, the parser can consume the first byte and then decides how many bytes more are needed (so if the first byte was 2 nom asks for 2 additional bytes). That's correct from nom's point of view. Since this approach does not match the semantics of Suricata's `AppLayerResult::incomplete`, which uses the notion of 'full records' -- including the already unsuccessfully parsed but not 'consumed' bytes --  I calculate the value to return to Suricata by taking the current length and whatever is needed to continue parsing.

Example:
In `parse_supported_security_types`, suppose we give it one byte with value 2. Then `current` is the byte `2`, and `current.len() == 2`.  So, as you say, we need 3 bytes in total and that's what we get by calculating `current.len() + needed` where `needed` is the value returned by nom. The question is, what would happen if we gave the parser 2 bytes (in `current`).
-  If nom would ask for 1 byte, then our calculation would be correct. 
- If it would ask for 2 bytes again, we would indeed have a problem and would need to rethink the incomplete reporting strategy as we can't really use the `nom::Needed::Size` value that we got from nom without also knowing the number of bytes that were _successfully_ parsed. I already wondered earlier on IRC whether we could get that value in any way.

The same applies for the second case. I would have expected to get `nom::Needed::Size(3)` when given one byte :/
Maybe I am just overlooking something? Any ideas @chifflier @victorjulien ?

catenacyber:
We reached the same analysis.
I add @jasonish for ideas...

satta:
@fhonza  and I have just traced such a case and can confirm that the `nom::Needed::Size()` always gives us the total length of what is incomplete on the nom side, with no indication of how much nom could parse without completing the record. That makes it difficult to calculate what `AppLayerResult::incomplete` needs, to say the least...

catenacyber:
You can only make it when you have fixed size structures...
Otherwise, request one more byte...
Or add dummy bytes until you do not get the incomplete from nom...

jasonish:
> Maybe I am just overlooking something? Any ideas @chifflier @victorjulien? I think that it would be worse asking for too many bytes (which would most likely never come) than just asking for one more byte and risk parsing the buffer more frequently than absolutelu needed.

Asking for one more byte would be the same scenario we had to do before the incomplete logic where the parser would buffer itself then try again when new data was received. If the protocol has no well defined header with a length I don't see away around doing this.

satta:
> > Maybe I am just overlooking something? Any ideas @chifflier @victorjulien? I think that it would be worse asking for too many bytes (which would most likely never come) than just asking for one more byte and risk parsing the buffer more frequently than absolutelu needed.
> 
> Asking for one more byte would be the same scenario we had to do before the incomplete logic where the parser would buffer itself then try again when new data was received. If the protocol has no well defined header with a length I don't see away around doing this.

I see. And since the next buffer extension will most likely be more than just the byte we asked for, this should not blow up the number of parsing tries too much, I guess.
I think we could then just keep this behaviour here, maybe just remove `needed` completely, since it's never used.

catenacyber:
At least now, each parser does not have to rewrite its buffering logic

catenacyber:
@satta here is a small patch to suricata to test against packet splitting
Not sure it is perfect with the new incomplete stuff...

```
diff --git a/src/app-layer.c b/src/app-layer.c
index e69db23cf..84df44cce 100644
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -662,6 +662,28 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
          * a start msg should have gotten us one */
         if (f->alproto != ALPROTO_UNKNOWN) {
             PACKET_PROFILING_APP_START(app_tctx, f->alproto);
+#ifndef LOLSPLIT
+            if (data_len != 0) {
+                //printf("lola %d\n", data_len);
+                size_t chunk_len = 1;
+                size_t i = 0;
+                while (i < data_len) {
+                    //printf("lolb %d+%d/%d\n", i, chunk_len, data_len);
+                    r = AppLayerParserParse(tv, app_tctx->alp_tctx, f, f->alproto,
+                                            flags, data+i, chunk_len);
+                    //printf("lolc %d %d\n", r, (*stream)->data_required);
+                    if ((*stream)->data_required > 0 && (*stream)->data_required + i <= data_len) {
+                        chunk_len = (*stream)->data_required;
+                        (*stream)->data_required = 0;
+                    } else if ((*stream)->data_required + i >= data_len) {
+                    } else {
+                        i+=chunk_len;
+                        chunk_len = 1;
+                    }
+                }
+            } else
+#endif
             r = AppLayerParserParse(tv, app_tctx->alp_tctx, f, f->alproto,
                                     flags, data, data_len);
             PACKET_PROFILING_APP_END(app_tctx, f->alproto);
```

satta:
Hmmm, okay. I must admit I am not too familiar with the code (in particular the mechanics behind `stream.data_required`etc.) to comment on that and rather leave it to @victorjulien or @jasonish. 

TBH to me it is most interesting here to have typical use cases regarding the `AppLayerResult` interface further fleshed out, in particular when interacting with nom. The better defined or designed this interface is, the more robust the code a parser author can produce will be.

PR #4798, opened 2020-04-09
detect/asn1: Fixes definite long form parsing of length field

sharksforarms:

https://redmine.openinfosecfoundation.org/issues/3628

Describe changes:
- Changed definite long form parsing of length field from additive to decoding the bytes into a uint32_t
- Updated test cases

### DIFF
--- a/src/detect-asn1.c
+++ b/src/detect-asn1.c
@@ -913,9 +913,9 @@ static int DetectAsn1Test05(void)
         buf[0] = '\x09';
         /* length, definite form, 2 octets */
         buf[1] = '\x82';
-        /* length is the sum of the following octets (257): */
-        buf[2] = '\xFE';
-        buf[3] = '\x03';
+        /* length is the representation of the following octets (257): */
+        buf[2] = '\x01';
+        buf[3] = '\x01';
 
         /* Fill the content of the number */
         uint16_t i = 4;
@@ -983,9 +983,9 @@ static int DetectAsn1Test06(void)
         buf[0] = '\x09';
         /* length, definite form, 2 octets */
         buf[1] = '\x82';
-        /* length is the sum of the following octets (256): */
-        buf[2] = '\xFE';
-        buf[3] = '\x02';
+        /* length is the representation of the following octets (256): */
+        buf[2] = '\x01';
+        buf[3] = '\x00';
 
         /* Fill the content of the number */
         uint16_t i = 4;
@@ -1201,9 +1201,9 @@ static int DetectAsn1TestReal03(void)
     buf[0] = '\x09';
     /* length, definite form, 2 octets */
     buf[1] = '\x82';
-    /* length is the sum of the following octets (257): */
-    buf[2] = '\xFE';
-    buf[3] = '\x03';
+    /* length is the representation of the following octets (257): */
+    buf[2] = '\x01';
+    buf[3] = '\x01';
 
     /* Fill the content of the number */
     uint16_t i = 4;
--- a/src/util-decode-asn1.c
+++ b/src/util-decode-asn1.c
@@ -146,15 +146,15 @@ static uint32_t SCAsn1GetLengthLongForm(Asn1Ctx *ac)
             return ASN1_PARSER_ERR;
         }
 
-        if ((uint64_t) ((uint64_t)content_len +
-            (uint64_t) ASN1_BER_GET_HIGH_TAG_NUM(raw_len)) > UINT32_MAX)
+        uint64_t tmp_len = ((uint64_t)content_len << 8) + (uint64_t) raw_len;
+        if (tmp_len > UINT32_MAX)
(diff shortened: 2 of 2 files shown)


src/detect-asn1.c
```diff
@@ -984,8 +984,8 @@ static int DetectAsn1Test06(void)
         /* length, definite form, 2 octets */
         buf[1] = '\x82';
         /* length is the sum of the following octets (256): */
-        buf[2] = '\xFE';
-        buf[3] = '\x02';
+        buf[2] = '\x01';
+        buf[3] = '\x00';
```
catenacyber, line 989: I do not know everything about ASN1, but
Why do you change this tests ?
Would you rather add a new test ?

sharksforarms: I had to fix the current test cases as they were incorrect and asserting the previous behavior

catenacyber: Ok, but why do you change the input fo the test, and not the output ?

sharksforarms: I change the input to the test as when the ASN1 data get's parsed, the length get's parsed incorrectly and the rest of the test ends up being invalid which causes the tests to fail.

For example, in `DetectAsn1Test06`, the length (bytes 3-4) is intended to be 256 (0x0100, or previously, 0xFE + 0x02), if I do not change it, the length ends up being 65,026‬ (0xFE02) which causes the double_overflow rule to trigger on `&& (node->len.len > 256 || node->data.len > 256))` which should not happen in this test case: `Error, double_overflow should not match any node`

catenacyber: Seems legit when testing the inputs on https://lapo.it/asn1js/

@chifflier any thoughts on this ASN1 stuff ?

chifflier: I just had a quick look, but this comment is wrong:
```
buf[1] = '\x82';
/* length is the sum of the following octets (257): */
buf[2] = '\xFE';
buf[3] = '\x03';
```
The length is never the sum of bytes in BER. in long form it is 
> the encoding of an unsigned binary integer equal to the number of octets in the content octets

See X.690 section 8.1.3

There are 3 encodings: definite short form, long form, and indefinite form. Which one to use depends if you are using BER or DER.

Short form is restricted to length <= 127. So, if you want to encode 256, the encoding of the length as definite long form would be indeed `82 01 00`:
- 82: bit 8 to 1 (see 8.1.3.5) + 2, for the length is encoded as 2 bytes
- 01 00" 256, the length.

So, I don't know if the expected value is 256, but `82 01 00` would be the correct encoding of 256!

chifflier: BTW, the comment says 257, so the correct encoding would be `82 01 01` following the same rules. What I do not know is if you intend to test 256 or 257.

sharksforarms: Hi @chifflier, I updated the comment previously, but it did not update on this review view, changed `sum` to `representation` for example: `/* length is the representation of the following octets (256): */`

There are 2 existing test cases: 
1. Length of 257 (82 01 01) is meant to test the double_overflow rule does trigger
2. Length of 256 (82 01 00) is meant to test the double_overflow rule does not trigger

chifflier: @sharksforarms I agree with the test case encodings and wording (representation or encoding is fine)

catenacyber: Thanks Pierre for the feedback.

> I updated the comment previously, but it did not update on this review view, changed sum to representation

@sharksforarms : can you push your commit with changed comments ?

src/util-decode-asn1.c
```diff
@@ -146,15 +146,15 @@ static uint32_t SCAsn1GetLengthLongForm(Asn1Ctx *ac)
...
+        uint64_t tmp_len = ((uint64_t)content_len << 8) + (uint64_t) raw_len;
+        if (tmp_len > UINT32_MAX)
         {
             node->flags |= ASN1_BER_EVENT_LEN_TOO_LONG;
             ac->parser_status = ASN1_STATUS_INVALID;
             return ASN1_PARSER_ERR;
         }
 
-        content_len += raw_len;
+        content_len = tmp_len;
```
catenacyber, line 157: Do we still need this to be in the loop `for (; i < oct_len; i++)` ?

catenacyber: What about ASN1 `8400010000` ?
Should not it be 4 bytes length encoded for 0x10000 so 65536 ?

sharksforarms: Hey @catenacyber, we still need the for loop in order to gather the `oct_len` bytes into `content_len`, your example is correct.

src/detect-asn1.c
sharksforarms: @catenacyber The change is pushed, would you like me to separate the commits?

PR #4799, opened 2020-04-09
conf/yaml: limit recursion depth while paring YAML - v1

jasonish:
A deeply nested YAML file can cause a stack-overflow while
reading in the configuration to do the recursive parser. Limit
the recursion level to something sane (128) to prevent this
from happening.

The default Suricata configuration has a recursion level of 128
so there is still lots of room to grow (not that we should).

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3630

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/466
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/823

### DIFF
--- a/src/conf-yaml-loader.c
+++ b/src/conf-yaml-loader.c
@@ -34,6 +34,10 @@
 #define YAML_VERSION_MAJOR 1
 #define YAML_VERSION_MINOR 1
 
+/* The maximum level of recursion allowed while parsing the YAML
+ * file. */
+#define RECURSION_LIMIT 128
+
 /* Sometimes we'll have to create a node name on the fly (integer
  * conversion, etc), so this is a default length to allocate that will
  * work most of the time. */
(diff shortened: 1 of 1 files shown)


src/conf-yaml-loader.c
```diff
@@ -170,21 +174,27 @@ ConfYamlHandleInclude(ConfNode *parent, const char *filename)
...
     ConfNode *node = parent;
     yaml_event_t event;
     memset(&event, 0, sizeof(event));
     int done = 0;
     int state = 0;
     int seq_idx = 0;
+    int retval = 0;
+
+    if (rlevel++ > RECURSION_LIMIT) {
+        FatalError(SC_ERR_FATAL, "Recusion limit hit while parsing configuration file, aborting.");
```
satta, line 188: FYI:  There's a typo here: `Recusion`.

victorjulien: Thanks, fixing this before merge.

victorjulien:
Merged with typo-fix in #4802, thanks!

PR #4800, opened 2020-04-10
Ssh rust v15

catenacyber:

Describe changes:

This looks good

Modifies #4790 with rebasing, and reverting to not trusting Nom incomplete value cf #4792

### DIFF
--- a/rust/src/ssh/ssh.rs
+++ b/rust/src/ssh/ssh.rs
@@ -1,0 +1,538 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use super::parser;
+use crate::applayer::*;
+use crate::core::STREAM_TOSERVER;
+use crate::core::{self, AppProto, Flow, ALPROTO_UNKNOWN, IPPROTO_TCP};
+use crate::log::*;
+use std;
+use std::ffi::{CStr, CString};
+use std::mem::transmute;
+
+static mut ALPROTO_SSH: AppProto = ALPROTO_UNKNOWN;
+
+#[repr(u32)]
+pub enum SSHEvent {
+    InvalidBanner = 0,
+    LongBanner,
+    InvalidRecord,
+}
+
+impl SSHEvent {
+    fn from_i32(value: i32) -> Option<SSHEvent> {
+        match value {
+            0 => Some(SSHEvent::InvalidBanner),
+            1 => Some(SSHEvent::LongBanner),
+            2 => Some(SSHEvent::InvalidRecord),
+            _ => None,
+        }
+    }
+}
+
+#[repr(u8)]
+#[derive(Copy, Clone, PartialOrd, PartialEq)]
+pub enum SSHConnectionState {
+    SshStateInProgress = 0,
+    SshStateBannerWaitEol = 1,
+    SshStateBannerDone = 2,
+    SshStateFinished = 3,
+}
+
+const SSH_MAX_BANNER_LEN: usize = 256;
+const SSH_RECORD_HEADER_LEN: usize = 6;
+//TODO complete enum and parse messages contents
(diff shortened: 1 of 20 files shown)


rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,538 @@
...
+                if input.len() < SSH_MAX_BANNER_LEN {
+                    //0 consumed, needs at least one more byte
+                    return AppLayerResult::incomplete(0 as u32, (input.len() + 1) as u32);
+                } else {
+                    SCLogDebug!(
+                        "SSH banner too long {} vs {} and waiting for eol",
+                        input.len(),
+                        SSH_MAX_BANNER_LEN
+                    );
+                    match parser::ssh_parse_banner(input) {
```
victorjulien, line 276: A match with a 'main' branch an a generic error branch can also be written in a more compact way:
```rust
if let Ok((_, banner)) = parser::ssh_parse_banner(input) {
    // main branh
} else {
    // set event and return error
}

File: src/detect-ssh-proto-version.c [hunk 2]
```diff
@@ -330,46 +332,57 @@ static int DetectSshVersionTestParse03 (void)
...
+    StreamTcpUTSetupStream(&ssn.client, 1);
+
+    f = UTHBuildFlow(AF_INET, "1.1.1.1", "2.2.2.2", 1234, 2222);
+    FAIL_IF_NULL(f);
+    f->protoctx = &ssn;
+    f->proto = IPPROTO_TCP;
+    f->alproto = ALPROTO_SSH;
+
+    p = PacketGetFromAlloc();
+    //    p = UTHBuildPacket(NULL, 0, IPPROTO_TCP);
```
victorjulien, line 369: why isn't this usable?

src/detect-ssh-proto-version.c
```diff
@@ -330,46 +332,57 @@ static int DetectSshVersionTestParse03 (void)
...
+    FAIL_IF(unlikely(p == NULL));
+    p->proto = IPPROTO_TCP;
+    p->flow = f;
+    /*p->flowflags |= FLOW_PKT_TOSERVER;
```
victorjulien, line 373: cleanup

src/detect-ssh-proto-version.c
```diff
@@ -378,276 +391,202 @@ static int DetectSshVersionTestDetect01(void)
...
+    void *ssh_state = f->alstate;
     FAIL_IF_NULL(ssh_state);
 
     /* do detect */
     SigMatchSignatures(&th_v, de_ctx, det_ctx, p);
 
-    FAIL_IF(!(PacketAlertCheck(p, 1)));
+    FAIL_IF(PacketAlertCheck(p, 1));
+
+    SigGroupCleanup(de_ctx);
```
victorjulien, line 409: these 2 can be removed actually, they are called from `DetectEngineCtxFree` (there are many non-compliant tests on this)

src/detect-ssh-proto-version.c
victorjulien: same for other tests

rust/src/ssh/ssh.rs
```diff
@@ -0,0 +1,538 @@
...
+        match value {
+            0 => Some(SSHEvent::InvalidBanner),
+            1 => Some(SSHEvent::LongBanner),
+            2 => Some(SSHEvent::InvalidRecord),
+            _ => None,
+        }
+    }
+}
+
+#[repr(u8)]
```
victorjulien, line 47: This is also used in C, but the type does not have the same size as in C. Sounds risky in a FFI interface, even if it appears to work well right now. Thoughts @jasonish ?

rust/src/ssh/ssh.rs
catenacyber: ok

src/detect-ssh-proto-version.c
catenacyber: ok

src/detect-ssh-proto-version.c
catenacyber: ok

rust/src/ssh/ssh.rs
catenacyber: Well, we have the generated header
```
enum SSHConnectionState {
    SshStateInProgress = 0,
    SshStateBannerWaitEol = 1,
    SshStateBannerDone = 2,
    SshStateFinished = 3,
};
typedef uint8_t SSHConnectionState;
```

I do not think we can specify the size of an enum in C.
So, we have the best solution : use `SSHConnectionState` and not `enum SSHConnectionState` to avoid the risk.

Is it not the case ?

src/detect-ssh-proto-version.c
catenacyber: it is :-) changing

rust/src/ssh/ssh.rs
jasonish: It looks safe to me with the typedef. I don't see anything ending up in this scenario: https://github.com/OISF/suricata/commit/ae10a92bc63565fd07cbd3d295e0f942b053b2c4. Just have to be careful to always use the typedef. Looks like its only called from tests as well.

PR #4803, opened 2020-04-10
python: prepare for publishing to PyPI - v1

jasonish:
- Renames our Python package from suricata to suricatactl.
- Drop a VERSION file when building sdist so installation
  can happen outside of the Suriata source tree where there
  is no configure.ac to parse the version from.
  Note: sdist still needs to be built in tree.

With this we can now "python setup.py sdist" then push it up
to PyPI with twine which will make our Python libraries installable
with pip: pip install suricatactl

This is mainly useful for applications that import our socket
control module.

I would also like to have this backported to 5.0.x so we can publish a release version sooner than later.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3629

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/470
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/827

### DIFF
--- a/python/.gitignore
+++ b/python/.gitignore
@@ -5,3 +5,6 @@ lib/
 scripts-*/
 suricata/config/defaults.py
 !suricata/config/defaults.py.in
+MANIFEST
+VERSION
+dist
--- a/python/MANIFEST.in
+++ b/python/MANIFEST.in
@@ -1,0 +1,1 @@
+include VERSION
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -19,7 +19,7 @@ uninstall-local:
 	rm -f $(DESTDIR)$(bindir)/suricatasc
 	rm -rf $(DESTDIR)$(prefix)/lib*/python*/site-packages/suricata
 	rm -rf $(DESTDIR)$(prefix)/lib*/python*/site-packages/suricatasc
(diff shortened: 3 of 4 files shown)


victorjulien:
Restarting CI is messy :-/

victorjulien:
I'm a bit worried about risk of introducing regressions when doing a backport. How can we build confidence that its going be fine?

jasonish:
> Restarting CI is messy :-/

I've given a kick.

jasonish:
> I'm a bit worried about risk of introducing regressions when doing a backport. How can we build confidence that its going be fine?

None of the installed code changed, just the building code, and only in trivial ways.

We can leave it in master though, and don't touch the backports.  I think there are 0 compatibility issues with the SuricataSC module in master, and Suricata socket mode in 4.1.x and 5.0.x, so no end-user convenience is lost.

jasonish:
> We can leave it in master though, and don't touch the backports. I think there are 0 compatibility issues with the SuricataSC module in master, and Suricata socket mode in 4.1.x and 5.0.x, so no end-user convenience is lost.

As in, we can push a "suricatactl-6.0.0dev0" to pip which Cuckoo can use.  "6.0.0" is greater than "6.0.0devX", so it will also make it clear that pushing this code to PIP is prerelease/experimental.

jasonish:
After some further thought I think this should only push the lib, not the scripts.

PR #4801, opened 2020-04-10
doc/install: refer to buster as Debian stable

satta:
- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Describe changes:
- Stop mentioning stretch and jessie in Debian install documentation. Buster is the current stable.

PR #4809, opened 2020-04-12
doc/install: refer to buster as Debian stable - v2

satta:
Previous PR: #4801 

Describe changes:
- Clarify the use of the term "stable" to refer to Debian, not Suricata.
- Make sure to note that backports only carry a new version when it can be built there (nowadays not always possible due to too new Rust dependency requirements).


victorjulien:
Merged in #4811, thanks Sascha!

PR #4812, opened 2020-04-14
Optimization/c11/v32

victorjulien:
Default to C11 instead of gnu99.

Use C11 atomics, with fallback to the existing GNU atomics implementation. The spinlocked version of the atomics wrappers has been removed.

Use C11 thread local storage. Fall back to GNU __thread extension. Support for thread local storage is now mandatory. Remove the packet pool fallback to posix thread local storage.

The trickiest thing was to deal with includes that use `u_int`, `u_char` and such. For that I tried to do a `m4` extension that seems to work but could use review https://github.com/OISF/suricata/commit/626075a618f91dedfe37cd96b3f9e803b06fe3e2

Passed.


victorjulien:
Replaced by #4820

PR #4813, opened 2020-04-14
fuzz: add configure wrapper for oss-fuzz

victorjulien:


src/tests/fuzz/oss-fuzz-configure.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+./configure --disable-shared --enable-fuzztargets --disable-gccmarch-native
```
catenacyber, line 2: What about adding `sh autogen.sh` and `make` ?

victorjulien: Not really seeing the point. The goal here is to place the configure args under our control. For make and autogen we don't need that.

PR #4815, opened 2020-04-14
Ssh rust v16

catenacyber:

Describe changes:

This looks good

Modifies #4800 with comments taken into account

### DIFF
--- a/rust/src/ssh/logger.rs
+++ b/rust/src/ssh/logger.rs
@@ -1,0 +1,65 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use super::ssh::SSHTransaction;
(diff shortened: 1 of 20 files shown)


rust/src/ssh/logger.rs
```diff
@@ -0,0 +1,65 @@
...
+fn log_ssh(tx: &SSHTransaction) -> Option<Json> {
+    if tx.cli_hdr.protover.len() == 0 && tx.srv_hdr.protover.len() == 0 {
+        return None;
+    }
+    let js = Json::object();
+    if tx.cli_hdr.protover.len() > 0 {
+        let cjs = Json::object();
+        cjs.set_string(
+            "proto_version",
+            std::str::from_utf8(&tx.cli_hdr.protover).unwrap(),
```
victorjulien, line 31: Is this safe? Attacker may control way we try to decode as utf8 here and if the unwrap fails we may panic?

Perhaps safer to use `.unwrap_or(...)` where the `or` case returns either an empty string or a special string like `<invalid encoding>` or something similar.

victorjulien: If it is safe and has been validated elsewhere, it needs to be stated here in a comment.

catenacyber: So, it must not be safe.
Should I use `String::from_utf8_lossy` instead ?

victorjulien: I used both approaches in smb:
```rust
        let dialect = match &state.dialect_vec {
            &Some(ref d) => str::from_utf8(&d).unwrap_or("invalid"),
            &None        => "unknown",
        };
```
and
```
                let domain = String::from_utf8_lossy(&ntlmssp.domain);
                jsd.set_string("domain", &domain);
```

But I'm not really happy with either. Both lose info. In the `lossy` case I would like to know if/that we lost info. In the `unwrap_or` case we may have been able to get interesting info out.

Wonder if we should use lossy, but add some flag if we in fact lost into. Or add an extra field in this case (raw bytes, or hex notation or base64 notation or something).

@jasonish thoughts?

catenacyber: What about replacing `set_string` with `set_string_from_bytes` ?

PR #4818, opened 2020-04-14
Http range file v6

catenacyber:

Describe changes:

Follows #4562 with rebase, rewording commits and more comments

And this first PR only addresses the Range in the same flow and not over multiple flows as happens here https://forum.suricata.io/t/suricata-5-0-1-in-ips-mode/94/22
We may want to change the whole design to adapt to this...


catenacyber:
Closing this as we want to redesign this to be way more generic

PR #4824, opened 2020-04-15
Fuzz fix v1

catenacyber:
None

- `return` instead of `exit` from `ConfYamlParse`
- fixes fuzz target applayerparse by stopping on EOF/failure as Suricata does

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -713,6 +713,7 @@ static void HTPHandleError(HtpState *s, const uint8_t dir)
             HTPSetEvent(s, htud, dir, id);
         }
     }
+    BUG_ON(msg >= 0x10000);
     s->htp_messages_offset = (uint16_t)msg;
     SCLogDebug("s->htp_messages_offset %u", s->htp_messages_offset);
 }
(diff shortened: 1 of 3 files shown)


catenacyber:
cc @jasonish for the `ConfYamlParse` change

jasonish:
Ok on the YAML part.

src/app-layer-htp.c
```diff
@@ -713,6 +713,7 @@ static void HTPHandleError(HtpState *s, const uint8_t dir)
             HTPSetEvent(s, htud, dir, id);
         }
     }
+    BUG_ON(msg >= 0x10000);
```
victorjulien, line 716: maybe this warrants a proper fix. It seems 64k events in a flow is unlikely, but in long running pipelined sessions it might not be impossible? I guess we should think about how many events a single request or response can generate, and then think of the smallest possible request/response size, and then we can calculate how many transactions within a single flow we'd need to overflow this.

I think as a 'guard' I'd prefer just stopping processing new events.

We can also bump the type to u32, but in fuzzing we might actually reach that?

catenacyber: Ok to stop processing events...

So, maybe a guard just with `--enable-debug-validation` ? So `DEBUG_VALIDATE_BUG_ON` ?

Fuzzing reached the u16 limit (for now) only with repeated `parser is in HTP_STREAM_ERROR` (for each request or response) where Suricata is not supposed to call the parser again

PR #4828, opened 2020-04-16

catenacyber:
None

Follows #4824 without already merged commits, and improving handling of libhtp too many warnings

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -689,6 +689,11 @@ static void HTPHandleError(HtpState *s, const uint8_t dir)
 
     size_t size = htp_list_size(s->conn->messages);
     size_t msg;
+    if(size >= 0x10000) {
+        DEBUG_VALIDATE_BUG_ON("Overflowing htp_messages_offset");
+        // ignore further messages
+        return;
+    }
 
     for (msg = s->htp_messages_offset; msg < size; msg++) {
         htp_log_t *log = htp_list_get(s->conn->messages, msg);


src/app-layer-htp.c
```diff
@@ -689,6 +689,11 @@ static void HTPHandleError(HtpState *s, const uint8_t dir)
 
     size_t size = htp_list_size(s->conn->messages);
     size_t msg;
+    if(size >= 0x10000) {
+        DEBUG_VALIDATE_BUG_ON("Overflowing htp_messages_offset");
```
victorjulien, line 693: is this correct use of `assert`?

Also please avoid using hex notation of numbers to improve readability. In this case the value should probably be `USHRT_MAX`, to indicate that its about the max value the type can hold.

victorjulien: It also seems likely that this limit is way to high. IIRC the events we read from htp are in an array. This would be a very big array. We should probably bail much earlier, with a special "final" event indicating that we've reached the per flow event limit, and we will produce no further events.

catenacyber: > is this correct use of assert?

I have seen it before and it works (as a pointer to a string is not 0) but I do not know if the style is a good habit

> USHRT_MAX

Ok for that

> It also seems likely that this limit is way to high.

What would be a better limit ? 1024 ?

victorjulien: Something like that. Maybe even less, like 512 or even 256.

catenacyber: Going for 512.
We'll see if fuzzing finds a way to get them

PR #4837, opened 2020-04-20

catenacyber:
None

Follows #4828 by taking comments into account

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -87,6 +87,9 @@ static SCRadixTree *cfgtree;
 /** List of HTP configurations. */
 static HTPCfgRec cfglist;
 
+/** Limit to the number of libhtp messages that can be handled */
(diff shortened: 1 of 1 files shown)


src/app-layer-htp.c
```diff
@@ -689,6 +692,11 @@ static void HTPHandleError(HtpState *s, const uint8_t dir)
 
     size_t size = htp_list_size(s->conn->messages);
     size_t msg;
+    if(size >= HTP_MAX_MESSAGES) {
+        DEBUG_VALIDATE_BUG_ON("Too many libhtp messages");
+        // ignore further messages
```
victorjulien, line 697: I think it would be good to set a 'final' event here: 'too many events' or something. Then ppl can match (or drop) on it.

catenacyber: Ok

PR #4841, opened 2020-04-21
DCERPC Parser in Rust

inashivb:
DCERPC parser in Rust.

S-V PR to go with this: https://github.com/OISF/suricata-verify/pull/218


rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+            return Endianness::Big;
+        }
+        Endianness::Little
+    }
+
+    fn get_hdr_fraglen(&self) -> Option<u16> {
+        if let Some(ref hdr) = self.header {
+            return Some(hdr.frag_length);
+        }
+        // Shouldn't happen
```
victorjulien, line 321: @catenacyber @jasonish it would be good if we can create a Rust equivalent of `DEBUG_VALIDATE_BUG_ON` for a case like this, so that the fuzzers can see if this is reachable.

victorjulien: @inashivb maybe you can create it. It should be a wrapper around `panic!` that is only in effect if the `--enable-debug-validation` configure flag is set.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+        let endianness = self.get_endianness();
+        match parser::parse_bindctx_item(input, endianness) {
+            Ok((leftover_bytes, ctxitem)) => {
+                let mut uuidentry = DCERPCUuidEntry::new();
+                uuidentry.uuid = ctxitem.uuid;
+                uuidentry.internal_id = uuid_internal_id;
+                uuidentry.ctxid = ctxitem.ctxid;
+                uuidentry.version = ctxitem.version;
+                uuidentry.versionminor = ctxitem.versionminor;
+                let pfcflags = match self.get_hdr_pfcflags() {
```
victorjulien, line 458: Not tested, could this be `let pfcflags = self.get_hdr_pfcflags().unwrap_or(0);`?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+                // Error, probably malformed data.
+                SCLogDebug!("An error occurred while parsing DCERPC BIND CTXItem");
+                -1
+            }
+        }
+    }
+
+    pub fn process_bind_pdu(&mut self, input: &[u8]) -> i32 {
+        let mut retval = 0;
+        let mut idx = 12;
```
victorjulien, line 488: need code comment why this starts at 12

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+    }
+
+    pub fn handle_common_stub(&mut self, input: &[u8], bytes_consumed: i32, dir: u8) -> i32 {
+        let fraglen = match self.get_hdr_fraglen() {
+            Some(x) => x,
+            None => {
+                return -1;
+            }
+        };
+        self.padleft = fraglen - 16 - bytes_consumed as u16; // bytes processed so far = header length
```
victorjulien, line 614: Fraglen is taken from the network directly, and it seems its value is not validated. So I fear we may underflow if it is < 16?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+        let mut input_left = input.len() as i32 - bytes_consumed;
+        let mut parsed: i32 = bytes_consumed; // parsed bytes
+        while input_left > 0 && parsed < fraglen as i32 {
+            let retval = self.handle_stub_data(&input[parsed as usize..], input_left as u16);
```
victorjulien, line 618: is this cast to u16 down from u32 of input_left safe? Is it possible that it is > 64k?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+
+        // Making it a hard requirement for a data stream to at least have the header in place
+        // else return -1 and the buffer stays empty
+        // Reason: header has the most important of data required for further proccessing
+        if self.bytes_consumed < 16 && input_len > 0 {
+            parsed = self.process_header(&buffer);
+            if parsed == -1 {
+                self.extend_buffer(buffer, direction);
+                return AppLayerResult::ok();
+                // ASK the case when header is incomplete and that's all the data, returning
```
victorjulien, line 717: Based on the retval of `process_header` this could either mean the parser told us `Incomplete` or `Error`. In case of incomplete the current logic seems correct. In case of Error we should probably return error here. Looking at the parser `parser::parse_dcerpc_header` I don't know if we can actually get an error response.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+                    retval = self.process_request_pdu(&buffer[parsed as usize..]);
+                    if retval == -1 {
+                        return AppLayerResult::err();
+                    }
+                }
+                DCERPC_TYPE_RESPONSE => {
+                    let mut response = DCERPCResponse::new();
+                    response.cmd = match self.get_hdr_type() {
+                        Some(x) => x,
+                        None => 2,
```
victorjulien, line 762: can you add a comment on where this magical `2` comes from and what it means?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+                    };
+                    self.response = Some(response);
+                    retval = self.handle_common_stub(
+                        &buffer[parsed as usize..],
+                        0,
```
victorjulien, line 767: Not sure I understand this zero here. The argument is `bytes_consumed`. Should that be set to `parsed`? I think `handle_common_stub` needs documentation of the purpose of the function args.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+        }
+        // Shouldn't happen
+        None
+    }
+
+    pub fn handle_fragment_data(&mut self, input: &[u8], input_len: u16) -> u16 {
+        let mut retval: u16 = 0;
+        let hdrflags1 = match self.get_hdr_flags1() {
+            Some(x) => x,
+            None => 0, // TODO check if this is OK
```
victorjulien, line 881: suppose this can only happen if you have no header. In what case can that happen?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+        };
+        let rpc_vers = match self.get_hdr_rpc_vers() {
+            Some(x) => x,
+            None => 0,
+        };
+        let fraglenleft = self.fraglenleft;
+
+        // Update the stub params based on the packaet type
```
victorjulien, line 889: nit: typo "packaet"

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+        match self.get_hdr_pkt_type() {
```
victorjulien, line 890: Wonder if this double `match` could be collapsed into `match self.get._hdr_ptr_type().unwrap_or(0)` so that the result would be `x` instead of `Option(x)`?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+            }
+            Err(_) => {
+                // Error, probably malformed data.
+                SCLogDebug!("An error occurred while parsing DCERPC request");
+                -1
+            }
+        }
+    }
+
+    pub fn handle_input_data(&mut self, input: &[u8]) -> AppLayerResult {
```
victorjulien, line 962: got confused a bit as I see this method twice, once in the tcp and once in the udp handling. Would it be feasible to split those 2 implementations into 2 files?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+    let state = cast_pointer!(vtx, DCERPCState);
+    if dir & core::STREAM_TOSERVER != 0 {
+        state.detect_flags_ts = flags;
+    } else {
+        state.detect_flags_tc = flags;
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_dcerpc_set_buffer(
```
victorjulien, line 1269: This function is used to *get* data. It should probably be named something like `rs_dcerpc_get_stub_data`.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2600 @@
...
+        }
+        _ => {
+            if let Some(ref resp) = state.response {
+                *len = resp.stub_data_buffer_len as u32;
+                *buf = resp.stub_data_buffer.as_ptr();
+                SCLogDebug!("DCERPC Response stub buffer: Setting buffer to: {:?}", *buf);
+            }
+        }
+    }
+    // TODO good thing to do? Returning endianness from the function whihch is not meant to do that
```
victorjulien, line 1291: maybe better to return through a function arg. Then let this function return whether it was able to provide a buffer to the caller at all.

src/detect-dce-stub-data.c
```diff
@@ -91,20 +91,9 @@ static InspectionBuffer *GetDCEData(DetectEngineThreadCtx *det_ctx,
...
 
-        if (flow_flags & STREAM_TOSERVER) {
-            data_len = dcerpc_state->dcerpc.dcerpcrequest.stub_data_buffer_len;
-            data = dcerpc_state->dcerpc.dcerpcrequest.stub_data_buffer;
-        } else if (flow_flags & STREAM_TOCLIENT) {
-            data_len = dcerpc_state->dcerpc.dcerpcresponse.stub_data_buffer_len;
-            data = dcerpc_state->dcerpc.dcerpcresponse.stub_data_buffer;
-        }
-        if (dcerpc_state->dcerpc.dcerpchdr.packed_drep[0] & 0x10) {
+        if (rs_dcerpc_set_buffer(txv, &data, &data_len, flow_flags)) {
```
victorjulien, line 96: this function appears to be able to not touch `data` and `data_len`, so we need a check for when `data == NULL || data_len == 0` before calling `InspectionBufferSetup`

rust/src/dcerpc/dcerpc.rs
catenacyber: Leaving it to you @inashivb :-)

inashivb: Hi @catenacyber and @victorjulien !
Are you looking for something like the following? Just want to confirm that I understood the requirement correctly.

```
+  17 #[macro_export]
+  18 macro_rules! validate_item (
+  19   ($item:expr) => {
+  20     if $item.is_none() {
+  21         panic!("BUG: {:?} should not be NULL!");
+  22     }
+  23   };
+  24 );
```
And then this function in `dcerpc.rs` would look like
```
~  317     fn get_hdr_fraglen(&self) -> u16 {
~  318         validate_item!(self.header);
~  319         self.header.as_ref().unwrap().frag_length
    320     }
```

inashivb: Hi @catenacyber and @victorjulien ! Is the following patch in the right direction? Just want to know if I understood the requirement correctly.
```
diff --git a/rust/Makefile.am b/rust/Makefile.am
index 158d348b7..b9d86ab75 100644
--- a/rust/Makefile.am
+++ b/rust/Makefile.am
@@ -19,6 +19,10 @@ if DEBUG
 RUST_FEATURES +=       debug
 endif
 
+if DEBUG_VALIDATION
+RUST_FEATURES +=>   debug-validate
+endif
+
 if RUST_CROSS_COMPILE 
 RUST_TARGET = --target $(host_triplet)
 endif
diff --git a/rust/src/common.rs b/rust/src/common.rs
index 668d55c86..0956e340e 100644
--- a/rust/src/common.rs
+++ b/rust/src/common.rs
   );
 );
 
+
+#[cfg(not(feature = "debug-validate"))]
+#[macro_export]
+macro_rules! validate_option_item (
+  ($item:expr) => {};
+);
+
+
+#[macro_export]
+macro_rules! validate_option_item (
+  ($item:expr) => {
+    if $item.is_none() {
+        panic!("BUG: {:?} should not be NULL!");
+    }
+  };
```
and then in `dcerpc.rs`:
```
diff --git a/rust/src/dcerpc/dcerpc.rs b/rust/src/dcerpc/dcerpc.rs
index 7d9cbaf85..850966c53 100644
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -315,6 +315,7 @@ impl DCERPCState {
     }
 
     fn get_hdr_fraglen(&self) -> Option<u16> {
+        validate_option_item!(self.header);
         if let Some(ref hdr) = self.header {
             return Some(hdr.frag_length);
         }
```

victorjulien: I think the argument should be a bool. Think it would also be good to use the same name as in C.

rust/src/dcerpc/dcerpc.rs
inashivb: Should have used the constant. Its for `DCERPC_TYPE_RESPONSE`. Fixed.

rust/src/dcerpc/dcerpc.rs
inashivb: I was deleting 16 (hardcoded header length) in `handle_common_stub` and was using this param for request's consumed data from nom parser. Suppose I could pass a 16 there and removed hardcoded header length from the function. Thanks.

inashivb: hmm running into problem doing that. Realized why I did it like this in the first place. I think documentation would be better. That bytes_consumed param is for bytes after the header.

rust/src/dcerpc/dcerpc.rs
inashivb: DCE 1.1 defines fraglen as u16 (https://pubs.opengroup.org/onlinepubs/9629399/chap12.htm), I think I was keeping this variable as i32 because in this while loop things are subtracted from it so there's a possibility of things going in negative (not in ideal scenario though). Am I thinking wrong?

inashivb:
Replaced with #4909

rust/src/dcerpc/dcerpc.rs
victorjulien: So in what case can you see it going negative?

inashivb: None. Didn't check it properly. Thanks.

PR #4844, opened 2020-04-21
detect/ftp: FTP memory accounting fixes

jlucovsky:
Continuation of #4656 

This commit continues the work started by @vanlink and corrects the
accounting of FTP memory usage against the memcap limit.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3519](https://redmine.openinfosecfoundation.org/issues/3519)

Describe changes:
- Continuation of #4656 
- Added companion `suricata-verify` test [xxx](sv-link)

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -344,7 +344,7 @@ static void FTPTransactionFree(FTPTransaction *tx)
         FTPStringFree(str);
     }
 
-    SCFree(tx);
+    FTPFree(tx, sizeof(*tx));
 }
 
 static int FTPGetLineForDirection(FtpState *state, FtpLineState *line_state)
@@ -504,6 +504,7 @@ struct FtpTransferCmd {
     uint64_t flow_id;
     uint8_t *file_name;
     uint16_t file_len;
+    uint16_t file_mem_len;
     FtpRequestCommand cmd;
 };
 
@@ -513,7 +514,7 @@ static void FtpTransferCmdFree(void *data)
     if (cmd == NULL)
         return;
     if (cmd->file_name) {
-        FTPFree(cmd->file_name, cmd->file_len);
+        FTPFree(cmd->file_name, cmd->file_mem_len);
     }
     FTPFree(cmd, sizeof(struct FtpTransferCmd));
 }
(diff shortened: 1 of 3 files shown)


src/app-layer-ftp.c
```diff
@@ -536,7 +537,7 @@ static uint32_t CopyCommandLine(uint8_t **dest, const uint8_t *src, uint32_t len
         *dest = where;
     }
     /* either 0 or actual */
-    return length;
+    return length ? length + 1 : length;
```
victorjulien, line 540: Would `return length ? length + 1 : 0;` be the same? Seems more clear that 0 will be returned then.

src/app-layer-ftp.h
```diff
@@ -206,6 +206,7 @@ typedef struct FtpDataState_ {
     DetectEngineState *de_state;
     int32_t input_len;
     int16_t file_len;
+    int16_t file_mem_len;
```
victorjulien, line 209: Is this ever something other than 0 or `file_len - 1`?

src/output-json-ftp.c
```diff
@@ -99,7 +99,7 @@ static json_t *JsonFTPLogCommand(Flow *f, FTPTransaction *tx)
         TAILQ_FOREACH(response, &tx->response_list, next) {
             /* handle multiple lines within the response, \r\n delimited */
             uint8_t *where = response->str;
-            uint16_t length = response->len;
+            uint16_t length = (uint16_t) strlen((const char *)response->str);
```
victorjulien, line 102: Why is `len` not the same as `strlen`? Or is that another off by one because of the null-byte?

src/app-layer-ftp.c
jlucovsky: Yes, exactly the same.

src/app-layer-ftp.h
jlucovsky: 0 and `file_len + 1`. I'll see if i can remove it.

src/output-json-ftp.c
jlucovsky: Here, `request->len` includes the terminating NULL.

PR #4848, opened 2020-04-22
siganture: minimizes ip CIDR for ip range

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3662

Describe changes:
- Change `IPOnlyCIDRItemParseSingle` so as not to create a single `IPOnlyCIDRItem` for every single IP address, but instead group them by netmask < 32 if possible

### DIFF
--- a/src/detect-engine-iponly.c
+++ b/src/detect-engine-iponly.c
@@ -224,16 +224,27 @@ static int IPOnlyCIDRItemParseSingle(IPOnlyCIDRItem *dd, const char *str)
             dd->ip[0] =htonl(first);
 
             if (first < last) {
-                for (first++; first <= last; first++) {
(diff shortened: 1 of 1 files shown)


victorjulien:
Typo in PR title and commit message (siganture).

@jlucovsky can you have a look since you had a different approach to the same issue.

catenacyber:
> @jlucovsky can you have a look since you had a different approach to the same issue.

I commented on yours ;-)

jlucovsky:
> @jlucovsky can you have a look since you had a different approach to the same issue.

Both implementations produce the same results and this one eliminates recursion.

Suggest adding the test cases from my change and also removing the `tmp_ip` variables.

PR #4849, opened 2020-04-22
Byteextractstr/part1/v2

inashivb:
`StringParse` functions error out in case of trailing characters after the extracted int, `ByteExtractString` functions ignore that.

### DIFF
--- a/src/reputation.c
+++ b/src/reputation.c
@@ -28,6 +28,7 @@
 #include "suricata-common.h"
 #include "util-error.h"
 #include "util-debug.h"
+#include "util-byte.h"
 #include "util-ip.h"
 #include "util-radix-tree.h"
 #include "util-unittest.h"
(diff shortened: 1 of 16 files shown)


src/reputation.c
```diff
@@ -251,10 +252,9 @@ static int SRepCatSplitLine(char *line, uint8_t *cat, char *shortname, size_t sh
 
     SCLogDebug("%s, %s", ptrs[0], ptrs[1]);
 
-    int c = atoi(ptrs[0]);
-    if (c < 0 || c >= SREP_MAX_CATS) {
+    int c;
+    if (StringParseI32RangeCheck(&c, 10, 0, (const char *)ptrs[0], 0, SREP_MAX_CATS + 1) < 0)
```
victorjulien, line 256: should this be `- 1`? Value should stay below `SREP_MAX_CATS`

src/util-byte.h
```diff
@@ -360,6 +360,111 @@ int StringParseInt16(int16_t *res, int base, uint16_t len, const char *str);
...
+ * \param res Stores result
+ * \param base Base of the number to extract
+ * \param len Number of bytes to extract (23 max or 0 for unbounded)
+ * \param len Number of bytes to extract (23 max)
+ * \param str String to extract from
+ *
+ * \return n Number of bytes extracted on success
+ * \return -1 On error
+ */
+int StringParseU64RangeCheck(uint64_t *res, int base, uint16_t len, const char *str, uint64_t min, uint64_t max);
```
victorjulien, line 375: can you add the `WARN_UNUSED` macro so ppl get a compile warning if they don't check the result of this func? (same for the others)

src/reputation.c
inashivb: Oops.

src/util-byte.h
inashivb: Yes OK.

inashivb:
Replaced with #4855

PR #4843, opened 2020-04-21
detect/lua: Cleanup thread contextual info when needed

jlucovsky:
Continuation of #4838

I used these rules -- the first rule demonstrated the issue. With this PR, memory is no longer overrun.
```
alert http any any -> any any (msg:"High entropy in file name download"; file.name; content:".exe"; endswith; fast_pattern;  bsize:<21; lua:any.lua; xbits: isset, urlentropy.detected; track ip_dst, expire 3600; target: dest_ip; sid:10; rev:1;)
alert http any any -> any any (msg:"High entropy in file name download"; file.name; content:".exe"; endswith; fast_pattern;  bsize:<21; lua:any.lua;   target: dest_ip; sid:11; rev:1;)
```

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3273](https://redmine.openinfosecfoundation.org/issues/3273)

Describe changes:
- Addressed review comments.

PR #4852, opened 2020-04-22
detect/lua: Cleanup thread contextual info when needed

jlucovsky:
Continuation of #4843 

I used these rules -- the first rule demonstrated the issue. With this PR, memory is no longer overrun.
```
alert http any any -> any any (msg:"High entropy in file name download"; file.name; content:".exe"; endswith; fast_pattern;  bsize:<21; lua:any.lua; xbits: isset, urlentropy.detected; track ip_dst, expire 3600; target: dest_ip; sid:10; rev:1;)
alert http any any -> any any (msg:"High entropy in file name download"; file.name; content:".exe"; endswith; fast_pattern;  bsize:<21; lua:any.lua;   target: dest_ip; sid:11; rev:1;)
```

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3273](https://redmine.openinfosecfoundation.org/issues/3273)

Describe changes:
- Addressed CI build failures.

### DIFF
--- a/src/detect-lua.c
+++ b/src/detect-lua.c
@@ -98,7 +98,7 @@ static int DetectLuaAppTxMatch (DetectEngineThreadCtx *det_ctx,
                                 const SigMatchCtx *ctx);
 static int DetectLuaSetup (DetectEngineCtx *, Signature *, const char *);
 static void DetectLuaRegisterTests(void);
-static void DetectLuaFree(void *);
+static void DetectLuaFree(DetectEngineCtx *, void *);
 static int g_smtp_generic_list_id = 0;
 
 static int InspectSmtpGeneric(ThreadVars *tv,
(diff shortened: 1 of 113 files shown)


src/detect-lua.c
```diff
@@ -1119,6 +1120,8 @@ static void DetectLuaFree(void *ptr)
         if (lua->filename)
             SCFree(lua->filename);
 
+        DetectUnregisterThreadCtxFuncs(de_ctx, NULL, lua, "lua");
```
victorjulien, line 1123: Why is this part of this commit? It should be in its own. A big API update commit like this should be as clean as possible as it's hard to review before merge, but also later when someone might try to analyze why something was changed.

jlucovsky: I'll move it into a separate commit. I had not done so yet because this was the sole reason for the added function and the unregistration function was in the previous commit.

jlucovsky:
Continued in #4856

PR #4859, opened 2020-04-23

catenacyber:
None

Follows #4837 by creating a new app-layer decoder event


victorjulien:
Merged in #4864, thanks!

PR #4866, opened 2020-04-24
signature: minimizes ip CIDR for ip range

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3662

Describe changes:
- Change `IPOnlyCIDRItemParseSingle` so as not to create a single `IPOnlyCIDRItem` for every single IP address, but instead group them by netmask < 32 if possible

Modifies #4862 by
- avoiding shift overflows
- optimizing the number of CIDRS in case of example of 192.168.1.0-192.168.1.255 (`-1` added)

### DIFF
--- a/src/detect-engine-iponly.c
+++ b/src/detect-engine-iponly.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2010 Open Information Security Foundation
+/* Copyright (C) 2007-2020 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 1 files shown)


src/detect-engine-iponly.c
```diff
@@ -199,41 +199,53 @@ static int IPOnlyCIDRItemParseSingle(IPOnlyCIDRItem *dd, const char *str)
...
+                           first + (((uint32_t) 1)<<(32-(new->netmask-1))) - 1 <= last) {
+                        new->netmask--;
+                    }
                     new->ip[0] = htonl(first);
                     dd = dd->next;
+                    if (new->netmask == 0) {
+                        //case 0.0.0.0-255.255.255.255 as 0.0.0.0/0
+                        break;
+                    }
+                    first+=((uint32_t) 1)<<(32-new->netmask);
```
victorjulien, line 248: `(uint32_t) 1` can also be written as `1UL`, which I think is cleaner.

victorjulien: Maybe `UL` should be added to all the constants in this patch?

catenacyber: Ok thanks for the trick :-)

PR #4873, opened 2020-04-24
(Draft) - First attempt at output plugin support - v1

jasonish:
A an initial draft of what plugins could look like, with an
example of a pluggable file types for Eve logging.

Some notes:
- This is Linux only. Plugins on Windows are quite different
  requiring building your plugin along with full Suricata source,
  or a Suricata DLL being made available. I'm not sure about
  the Mac.
- We explicitly list the symbols that plugins can call. This
  may limit some cool plugins, but allows us to restrict what
  is called to an API we would consider stable.
- Plugins are loaded from a hard codec directory. I expect to
  use a configuration section like:
```
plugins:
  enabled: true
  directories:
    - /usr/lib/suricata/plugins
    - /var/lib/suricata/plugins
```
- We currently pass the whole LogFileCtx into the plugin. May
  want to consider only passing what is needed for the plugin,
  this would also help keep the ABI stable.
- I expect multi-threaded logging to change the calling convention
  for this file type plugs a bit
  (https://redmine.openinfosecfoundation.org/issues/3293)
- Maybe internal file types should be registered in much
  the same way (fairly large refactor). I'll probably try
  this with the Redis output.

### DIFF
--- a/plugins/example-output/dummy.c
+++ b/plugins/example-output/dummy.c
@@ -1,0 +1,56 @@
+#include <stdio.h>
+#include "suricata-plugin.h"
+#include "dummy.h"
+
+#define OUTPUT_NAME "plugin:dummy"
+
+typedef struct Context_ {
+} Context;
+
+static int dummy_write(const char *buffer, int buffer_len, LogFileCtx *ctx) {
+    printf("dummy_write\n");
+    printf("dummy_write: rotate=%d\n", ctx->rotation_flag);
+    return 0;
+}
+
+static void dummy_close(LogFileCtx *ctx) {
+    printf("dummy_close\n");
+    SCFree(ctx->plugin);
+}
+
+static int dummy_open(LogFileCtx *ctx, ConfNode *conf) {
+    printf("conf: %p\n", conf);
+    Context *context = SCCalloc(1, sizeof(Context));
+    ctx->Write = dummy_write;
+    ctx->Close = dummy_close;
+    ctx->plugin = context;
+    return 0;
+}
+
+/**
+ * Called by Suricata to initialize the module. This module registers
+ * new file type to the JSON logger.
+ */
+void sc_plugin_init(void)
+{
+    PluginFileType *my_output = SCCalloc(1, sizeof(PluginFileType));
+    my_output->name = OUTPUT_NAME;
+    my_output->Open = dummy_open;
+    RegisterPluginFileType(my_output);
+}
+
+/**
+ * This function will be called by Suricata after it has loaded the
+ * module. Here we register some details of the module as well as the
+ * initialization function.
+ *
+ * This is really just a courtesy by the plugin author where they can
+ * register some details of their module and follow a convention we
+ * set out.
+ */
+void sc_plugin_register(SCPlugin *plugin) {
+    plugin->name = "dummy-output-plugin";
+    plugin->author = "Jason Ish <jason.ish@oisf.net>";
+    plugin->license = "GPLv2";
+    plugin->Init = sc_plugin_init;
+}
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -66,6 +66,8 @@
 
 #include "source-pcap-file.h"
 
+#include "suricata-plugin.h"
+
 #define DEFAULT_LOG_FILENAME "eve.json"
 #define DEFAULT_ALERT_SYSLOG_FACILITY_STR       "local0"
 #define DEFAULT_ALERT_SYSLOG_FACILITY           LOG_LOCAL0
@@ -82,10 +84,18 @@ static const char *TRAFFIC_LABEL_PREFIX = "traffic/label/";
 static size_t traffic_id_prefix_len = 0;
 static size_t traffic_label_prefix_len = 0;
 
+static TAILQ_HEAD(, PluginFileType_) output_types =
+    TAILQ_HEAD_INITIALIZER(output_types);
+
+void RegisterPluginFileType(PluginFileType *plugin)
+{
+    SCLogNotice("Registering JSON file type plugin %s", plugin->name);
+    TAILQ_INSERT_TAIL(&output_types, plugin, entries);
+}
+
 void OutputJsonRegister (void)
 {
     OutputRegisterModule(MODULE_NAME, "eve-log", OutputJsonInitCtx);
-
     traffic_id_prefix_len = strlen(TRAFFIC_ID_PREFIX);
     traffic_label_prefix_len = strlen(TRAFFIC_LABEL_PREFIX);
 }
@@ -898,7 +908,9 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
             output_s = ConfNodeLookupChildValue(conf, "type");
         }
 
-        if (output_s != NULL) {
+        if (output_s == NULL) {
+            json_ctx->json_out = LOGFILE_TYPE_FILE;
+        } else {
             if (strcmp(output_s, "file") == 0 ||
                 strcmp(output_s, "regular") == 0) {
                 json_ctx->json_out = LOGFILE_TYPE_FILE;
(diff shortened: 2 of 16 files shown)


src/suricata-plugin.h
```diff
@@ -0,0 +1,30 @@
...
+    char *name;
+    char *license;
+    char *author;
+
+    void (*Init)(void);
+} SCPlugin;
+
+typedef struct PluginFileType_ {
+    char *name;
+    int (*Open)(LogFileCtx *ctx, ConfNode *conf);
```
victorjulien, line 24: I agree this should not expose `LogFileCtx` for compatibility reasons.

src/suricata-plugin.h
```diff
@@ -0,0 +1,30 @@
...
+#include <config.h>
+
+#undef HAVE_LUA
+#undef HAVE_NSS
+
+#include "suricata-common.h"
+#include "util-logopenfile.h"
+
+typedef struct SCPlugin_ {
+    char *name;
```
victorjulien, line 15: `const char *`

plugins/example-output/dummy.c
```diff
@@ -0,0 +1,56 @@
...
+    ctx->Close = dummy_close;
+    ctx->plugin = context;
+    return 0;
+}
+
+/**
+ * Called by Suricata to initialize the module. This module registers
+ * new file type to the JSON logger.
+ */
+void sc_plugin_init(void)
```
victorjulien, line 34: so the idea is that each plugin type has this common init function type and within this the plugin type specific setup will be done? (e.g. output, app-layer, etc)

plugins/example-output/dummy.c
```diff
@@ -0,0 +1,56 @@
...
+{
+    PluginFileType *my_output = SCCalloc(1, sizeof(PluginFileType));
+    my_output->name = OUTPUT_NAME;
+    my_output->Open = dummy_open;
+    RegisterPluginFileType(my_output);
```
victorjulien, line 39: Should this also register a free func?

Also, some plugins might have more complex shutdown requirements (e.g. properly close db session).  Should that be done in a 'Free', or should there be an additional 'Close'?

plugins/example-output/dummy.c
jasonish: Correct. It would be a generic interface for any plugin. The plugin doesn't need to declare itself to be app-layer, output, or whatever. My feeling is that might create too much overhead, at least until we see patterns emerge.

plugins/example-output/dummy.c
jasonish: I could see registering a destroy callback to associate the init function.  

Specifically to registering an output file type, it might make more sense to declare the Write/Open/Close callbacks during register, rather than in the open callback - which was just a bit of hack to limit modifications at this time. This will also help get rid of LogFileCtx in the open callback.

In this case Open would open the database connection, Close would close it.  And then a generic destroy would destroy any global context a plugin may have created.

src/util-plugin.c
```diff
@@ -0,0 +1,75 @@
...
+        void (*registerf)(SCPlugin *) = dlsym(lib, "sc_plugin_register");
+        if (registerf == NULL) {
+            SCLogNotice("Plugin missing registration function: %s", path);
+        } else {
+            SCPlugin plugin;
+            memset(&plugin, 0, sizeof(plugin));
+            (*registerf)(&plugin);
+            BUG_ON(plugin.name == NULL);
+            BUG_ON(plugin.author == NULL);
+            BUG_ON(plugin.license == NULL);
```
jlucovsky, line 40: Should we check if the plugin license is compatible?

src/util-plugin.c
```diff
@@ -0,0 +1,75 @@
...
+#include "suricata-plugin.h"
+
+#ifdef HAVE_DLFCN_H
+
+#include <dlfcn.h>
+
+static void InitPlugin(char *path)
+{
+    void *lib = dlopen(path, RTLD_LAZY);
+    if (lib != NULL) {
```
jlucovsky, line 29: Should print a warning/notice stating that `path` could not be opened using `dlerror`

src/util-plugin.c
```diff
@@ -0,0 +1,75 @@
...
+        if (registerf == NULL) {
+            SCLogNotice("Plugin missing registration function: %s", path);
+        } else {
+            SCPlugin plugin;
+            memset(&plugin, 0, sizeof(plugin));
+            (*registerf)(&plugin);
+            BUG_ON(plugin.name == NULL);
+            BUG_ON(plugin.author == NULL);
+            BUG_ON(plugin.license == NULL);
+            BUG_ON(plugin.Init == NULL);
```
jlucovsky, line 41: Suggest we have the plugin type be part of `plugin` ... that would allow controls on the types and/or number of plugins introduced. There might be benefit to this in controlled environments (FIPS?) or for compliance?

src/util-plugin.c
```diff
@@ -0,0 +1,75 @@
...
+static void InitPlugin(char *path)
+{
+    void *lib = dlopen(path, RTLD_LAZY);
+    if (lib != NULL) {
+        SCLogNotice("Loading plugin %s", path);
+        void (*registerf)(SCPlugin *) = dlsym(lib, "sc_plugin_register");
+        if (registerf == NULL) {
+            SCLogNotice("Plugin missing registration function: %s", path);
+        } else {
+            SCPlugin plugin;
```
jlucovsky, line 35: suggest `SCPlugin plugin = { 0 };`

src/output-json.c
```diff
@@ -82,10 +84,18 @@ static const char *TRAFFIC_LABEL_PREFIX = "traffic/label/";
 static size_t traffic_id_prefix_len = 0;
 static size_t traffic_label_prefix_len = 0;
 
+static TAILQ_HEAD(, PluginFileType_) output_types =
+    TAILQ_HEAD_INITIALIZER(output_types);
+
+void RegisterPluginFileType(PluginFileType *plugin)
+{
+    SCLogNotice("Registering JSON file type plugin %s", plugin->name);
+    TAILQ_INSERT_TAIL(&output_types, plugin, entries);
```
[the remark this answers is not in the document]
jlucovsky: Suggest we introduce the concept of "plugin priority" ... and order the Q by it.

plugins/example-output/dummy.c
```diff
@@ -0,0 +1,56 @@
+#include <stdio.h>
+#include "suricata-plugin.h"
+#include "dummy.h"
+
+#define OUTPUT_NAME "plugin:dummy"
```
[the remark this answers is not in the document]
jlucovsky: nit: suggest we include the term "template" in the filename, functions, typedefs instead of "dummy" to match existing files with similar purposes.

plugins/example-output/dummy.c
jlucovsky: This is a good idea ... give the plugin a chance to finalize resources and any connections.

src/suricata-plugin.h
jasonish: My current working branch has removed exposing LogFileCtx. Resolving this comment.

plugins/example-output/dummy.c
jasonish: I think in the end we'll have:
- Register
- Init
- ThreadInit
- Open
- Close
- ThreadDeInit
- Destroy

The thread init/deinit would be more clearly seen if something like the pcap logger were done as a plugin.Or multi-threaded JSON output.

plugins/example-output/dummy.c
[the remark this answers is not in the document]
jasonish: When I wrote the plugin initially I thought I might need to parse "plugin" then load the name. But I don't think this is needed at all. So while we could do a convention here, I'm more likely to name this plugin "template-json-filetype-plugin".

src/output-json.c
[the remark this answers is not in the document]
jasonish: I'm not sure what a priority would do.  Yeah, we have to check for duplicate names, including the internal names.  But since you have to explicitly name something to use, I'm not sure how priorities fit?

src/util-plugin.c
jasonish: Maybe some sort of opt-in check. Or opt-out flag. We should not prevent an end-user from using a plugin with an evil license.

src/util-plugin.c
jasonish: Do you have an idea of types? What I don't want to introduce requiring multiple plugins to fill out one feature.  Take an app-layer than may need to register an app-layer parser, a logger and detect.

src/output-json.c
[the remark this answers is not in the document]
jlucovsky: In the future, there will be open source Suricata plugins for various things and folks will include them without careful analysis (I think).

Might be helpful to introduce the concept of priority at an early stage to handle unintended duplications (multiple output plugins, eg)

src/util-plugin.c
jlucovsky: Yes, this seems like a control we should provide for the user (opt-in/opt-out)

src/util-plugin.c
jlucovsky: Perhaps `types` could be an array to allow for the scenario you posted?

Types: output (e.g, redis, "live" connector,  kafka, ...), app-layer, protocol decoder, detect.

src/output-json.c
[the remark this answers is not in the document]
jasonish: Wouldn't the name prevent duplications?  If you loaded a "kafka" file type, a second one of the same name would fail to load. You could have "my-kafka-plugin", and "your-kafka-plugin", but to use them you have to explicitly name the one you want to use. Nothing would be implicit.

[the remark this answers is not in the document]
jasonish: DONE: Duplicate name check as well as not allowing a name to conflict with a built-in (regular, syslog, redis, etc).

jasonish:
Continued at https://github.com/OISF/suricata/pull/5131.

PR #4897, opened 2020-04-28

inashivb:

### DIFF
--- a/src/util-cpu.c
+++ b/src/util-cpu.c
@@ -27,6 +27,7 @@
 #include "util-error.h"
 #include "util-debug.h"
 #include "util-cpu.h"
+#include "util-byte.h"
 
 /**
  * Ok, if they should use sysconf, check that they have the macro's
@@ -75,9 +76,15 @@ uint16_t UtilCpuGetNumProcessorsConfigured(void)
 
     return (uint16_t)nprocs;
 #elif OS_WIN32
-	long nprocs = -1;
-	const char* envvar = getenv("NUMBER_OF_PROCESSORS");
-	nprocs = (NULL != envvar) ? atoi(envvar) : 0;
+    int64_t nprocs = 0;
+    const char* envvar = getenv("NUMBER_OF_PROCESSORS");
+    if (envvar != NULL) {
+        if (StringParseInt64(&nprocs, 10, 0, envvar) < 0) {
+            SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for number of "
+                         "processors: %s, resetting to 0", envvar);
+            nprocs = 0;
+        }
+    }
     if (nprocs < 1) {
         SCLogError(SC_ERR_SYSCALL, "Couldn't retrieve the number of cpus "
                    "configured from the NUMBER_OF_PROCESSORS environment variable");
--- a/src/util-log-redis.c
+++ b/src/util-log-redis.c
@@ -26,6 +26,7 @@
 #include "suricata-common.h" /* errno.h, string.h, etc. */
 #include "util-log-redis.h"
 #include "util-logopenfile.h"
+#include "util-byte.h"
 
 #ifdef HAVE_LIBHIREDIS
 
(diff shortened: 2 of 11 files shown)


src/util-cpu.c
```diff
@@ -75,9 +76,15 @@ uint16_t UtilCpuGetNumProcessorsConfigured(void)
...
     return (uint16_t)nprocs;
 #elif OS_WIN32
-	long nprocs = -1;
-	const char* envvar = getenv("NUMBER_OF_PROCESSORS");
-	nprocs = (NULL != envvar) ? atoi(envvar) : 0;
+    int64_t nprocs = 0;
+    const char* envvar = getenv("NUMBER_OF_PROCESSORS");
+    if (envvar != NULL) {
+        if (StringParseInt64(&nprocs, 10, 0, envvar) < 0) {
+            SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for number of "
```
jlucovsky, line 83: Suggest using `SC_WARN` code?

src/util-cpu.c
```diff
@@ -75,9 +76,15 @@ uint16_t UtilCpuGetNumProcessorsConfigured(void)
 
     return (uint16_t)nprocs;
 #elif OS_WIN32
-	long nprocs = -1;
-	const char* envvar = getenv("NUMBER_OF_PROCESSORS");
-	nprocs = (NULL != envvar) ? atoi(envvar) : 0;
+    int64_t nprocs = 0;
```
jlucovsky, line 79: 63 bits -- `int64_t` seems large for the cpu count

src/util-log-redis.c
```diff
@@ -514,7 +515,10 @@ int SCConfLogOpenRedis(ConfNode *redis_node, void *lf_ctx)
         SCLogError(SC_ERR_MEM_ALLOC, "Error allocating redis server string");
         exit(EXIT_FAILURE);
     }
-    log_ctx->redis_setup.port = atoi(redis_port);
+    if (StringParseInt32(&log_ctx->redis_setup.port, 10, 0, (const char *)redis_port) < 0) {
```
jlucovsky, line 518: This could be range checked - ports are 16-bit values.

src/util-napatech.c
```diff
@@ -800,10 +800,16 @@ static uint32_t CountWorkerThreads(void)
...
-
-                        ByteExtractStringUint8(&start, 10,  0, copystr);
-                        ByteExtractStringUint8(&end, 10,  0, strchr(copystr, '-') + 1);
-
+                        if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
+                            SCLogError(SC_ERR_INVALID_VALUE, "Napatech invalid"
+                                       " worker range start: %s", copystr);
+                            exit(EXIT_FAILURE);
+                        }
+                        if (StringParseUint8(&end, 10, 0, (const char *) (strchr(copystr, '-') + 1)) < 0) {
```
jlucovsky, line 808: Should we check that `end > start`?

src/util-privs.c
```diff
@@ -154,7 +155,10 @@ int SCGetUserID(const char *user_name, const char *group_name, uint32_t *uid, ui
 
     /* Get the user ID */
     if (isdigit((unsigned char)user_name[0]) != 0) {
-        userid = atoi(user_name);
+        if (ByteExtractStringUint32(&userid, 10, 0, (const char *)user_name) < 0) {
+            SCLogError(SC_ERR_UID_FAILED, "invalid user name: '%s'", user_name);
```
jlucovsky, line 159: s/invalid user name/invalid user id value/

src/util-cpu.c
victorjulien: I think some of the OS specific APIs return `long`, so thats why that is used here. Converting it w/o down casting makes sense to me.

src/util-cpu.c
inashivb: Does `SC_WARN_POOR_RULE` fit the category here? I do not see any appropriate SC_WARN codes for invalid value.

victorjulien: I don't object to the original code you picked. We want to overhaul the error codes anyway.

inashivb:
Replaced with #4937

PR #4907, opened 2020-05-01
Ssh rust v18

catenacyber:

Describe changes:

This looks good

Modifies #4906 with 
- Modifying `set_string_from_bytes` to use `json_stringn` to allow null characters
- replacing`String::from_utf8_lossy` by using `set_string_from_bytes`


catenacyber:
Also rebased on latest master to get CI working

catenacyber:
Apparently, not all libjannsson versions have `json_stringn` (comment from output-json-dnp3-objects.c)

PR #4908, opened 2020-05-01

catenacyber:

Describe changes:
- adds file flag for `file_data` keyword

with `if ((s->flags & SIG_FLAG_FILESTORE) || s->file_flags != 0) {`

There is no specific use of `FILE_SIG_NEED_FILE`

Modifies #4905 with rebase to get CI working

### DIFF
--- a/src/detect-file-data.c
+++ b/src/detect-file-data.c
@@ -189,6 +189,7 @@ static int DetectFiledataSetup (DetectEngineCtx *de_ctx, Signature *s, const cha
     if (DetectBufferSetActiveList(s, DetectBufferTypeGetByName("file_data")) < 0)
         return -1;
 
+    s->file_flags |= FILE_SIG_NEED_FILE;
     SetupDetectEngineConfig(de_ctx);
     return 0;
 }


src/detect-file-data.c
```diff
@@ -189,6 +189,7 @@ static int DetectFiledataSetup (DetectEngineCtx *de_ctx, Signature *s, const cha
     if (DetectBufferSetActiveList(s, DetectBufferTypeGetByName("file_data")) < 0)
         return -1;
 
+    s->file_flags |= FILE_SIG_NEED_FILE;
```
victorjulien, line 192: I'd like to use a different flag here. The `FILE_SIG_*` are for use with the `File` API, and `file_data` doesn't always use that API. For HTTP it uses different logic. So I want to avoid risking confusion there.

I think a simple new flag `SIG_FLAG_INIT_FILEDATA` or something in `Signature::init_data::flags` could be set during parsing and checked during validation.

catenacyber: Let me rephrase to make sure I understand.
Instead of using the condition `s->file_flags != 0` in `SigValidate`, you would like to use something like `s->flags & SIG_FLAG_INIT_FILEDATA `, correct ?

PR #4909, opened 2020-05-01
Dcerpc single tx/v2

inashivb:
DCERPC parser in Rust.

S-V PR to go with this: https://github.com/OISF/suricata-verify/pull/226


rust/Makefile.am
```diff
@@ -19,6 +19,10 @@ if DEBUG
 RUST_FEATURES +=	debug
 endif
 
+if DEBUG_VALIDATION
+RUST_FEATURES +=>   debug-validate
```
victorjulien, line 23: does the `>` belong here?

rust/src/common.rs
```diff
@@ -14,6 +14,24 @@ macro_rules! take_until_and_consume (
...
+#[cfg(not(feature = "debug-validate"))]
+#[macro_export]
+macro_rules! debug_validate_bug_on (
+  ($item:expr) => {};
+);
+
+
+#[cfg(feature = "debug-validate")]
+#[macro_export]
+macro_rules! debug_validate_bug_on (
```
victorjulien, line 27: indent looks off

rust/src/common.rs
```diff
@@ -14,6 +14,24 @@ macro_rules! take_until_and_consume (
...
+  ($item:expr) => {
+    if $item {
+        panic!("Value should not be NULL!");
```
victorjulien, line 30: curious: can we just feed $item to `panic!` so its printed in the backtrace?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -782,7 +789,7 @@ impl DCERPCState {
     }
 }
 
-fn evaluate_stub_params(
+pub fn evaluate_stub_params(
```
victorjulien, line 792: this doesn't belong in this commit

rust/src/dcerpc/dcerpc.rs
```diff
@@ -979,6 +979,7 @@ pub unsafe extern "C" fn rs_dcerpc_get_stub_data(
         }
     }
     *endianness = state.get_hdr_drep_0() & 0x10;
+    1
```
victorjulien, line 989: does this mean it always returns 1?

src/detect-dce-stub-data.c
```diff
@@ -91,20 +91,14 @@ static InspectionBuffer *GetDCEData(DetectEngineThreadCtx *det_ctx,
     InspectionBuffer *buffer = InspectionBufferGet(det_ctx, list_id);
     if (buffer->inspect == NULL) {
         uint32_t data_len = 0;
-        uint8_t *data = NULL;
+        const uint8_t *data = NULL;
+        uint8_t endianness;
 
-        DCERPCState *dcerpc_state = txv;
-        if (dcerpc_state == NULL)
+        if (rs_dcerpc_get_stub_data(txv, &data, &data_len, &endianness, flow_flags) != 1 ||
```
victorjulien, line 97: if rs_dcerpc_get_stub_data can only return 1, whats the point of this `!= 1` check?

rust/src/dcerpc/dcerpc_udp.rs
```diff
@@ -0,0 +1,511 @@
...
+ * 02110-1301, USA.
+ */
+
+use std::mem::transmute;
+
+use crate::core;
+use crate::applayer::AppLayerResult;
+use crate::dcerpc::parser;
+use crate::log::*;
+use crate::dcerpc::dcerpc::{DCERPCRequest, DCERPCResponse, DCERPCUuidEntry, PFC_FIRST_FRAG, DCERPC_TYPE_REQUEST, DCERPC_TYPE_RESPONSE};
```
victorjulien, line 24: very long line

rust/src/dcerpc/dcerpc_udp.rs
```diff
@@ -0,0 +1,511 @@
...
+    pub fn handle_input_data(&mut self, input: &[u8]) -> AppLayerResult {
+        // Call header parser first
+        let mut parsed = self.process_header(input);
+        if parsed == -1 {
+            return AppLayerResult::err();
+        }
+
+        // Input left after successful header parsing should be 80 less
+        // than the original length
+        if input.len() < 80 {
```
victorjulien, line 205: can `process_header` succeed if this check is true? If not, the check should probably be moved to before the call to `process_header`

rust/src/dcerpc/dcerpc_udp.rs
```diff
@@ -0,0 +1,511 @@
...
+        }
+        let mut input_left = input.len() as i32 - 80;
+        let pkt_type = self.get_hdr_pkt_type().unwrap_or(0);
+        let fraglen = self.get_hdr_fraglen().unwrap_or(0);
+        self.fraglenleft = fraglen;
+        self.create_new_query(pkt_type);
+        // Parse rest of the body
+        while parsed >= DCERPC_UDP_HDR_LEN && parsed < fraglen as i32 && input_left > 0 {
+            let retval = self.handle_fragment_data(&input[parsed as usize..], input_left as u16);
+            if retval > 0 || retval > input_left as u16 {
```
victorjulien, line 216: should this be `retval >= input_left`?  Seems `input_left` can never reach 0?

rust/src/dcerpc/dcerpc_udp.rs
```diff
@@ -0,0 +1,511 @@
...
+    _flow: *mut core::Flow,
+    state: &mut DCERPCUDPState,
+    _pstate: *mut std::os::raw::c_void,
+    input: *const u8,
+    input_len: u32,
+    _data: *mut std::os::raw::c_void,
+    _flags: u8,
+) -> AppLayerResult {
+    if input_len > 0 && input != std::ptr::null_mut() {
+        let buf = std::slice::from_raw_parts(input, input_len as usize);
```
victorjulien, line 268: we have a nice macro for this... `build_slice!`?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2127 @@
...
+                self.header = Some(header);
+                (input.len() - leftover_bytes.len()) as i32
+            }
+            Err(nom::Err::Incomplete(_)) => {
+                // Insufficient data.
+                SCLogDebug!("Insufficient data while parsing DCERPC header");
+                -1
+            }
+            Err(nom::Err::Error(_)) => {
+                SCLogDebug!("EoF reached with the data");
```
victorjulien, line 436: how can you tell EOF from other possible errors?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2127 @@
...
+                let numctxitems = header.numctxitems;
+                self.bind = Some(header);
+                for i in 0..numctxitems {
+                    retval = self.handle_bindctxitem(&input[idx as usize..], i as u16);
+                    if retval == -1 {
+                        return -1;
+                    }
+                    idx = retval + idx;
+                }
+                (input.len() - leftover_bytes.len()) as i32 + retval * numctxitems as i32
```
victorjulien, line 496: I have trouble parsing this: can you explain what is being returned here?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2127 @@
...
+        let fraglen = self.get_hdr_fraglen().unwrap_or(0);
+        if fraglen < bytes_consumed as u16 + DCERPC_HDR_LEN {
+            return -1;
+        }
+        self.padleft = fraglen - DCERPC_HDR_LEN - bytes_consumed as u16; // bytes processed so far = header length
+        let mut input_left = input.len() as i32 - bytes_consumed;
+        let mut parsed: i32 = bytes_consumed; // parsed bytes
+        while input_left > 0 && parsed < fraglen as i32 {
+            let retval = self.handle_stub_data(&input[parsed as usize..], input_left as u16);
+            if retval > 0 || retval > input_left as u16 {
```
victorjulien, line 620: same as with udp, should this be `retval >= input_left`?

victorjulien: Wait a minute: if retval > input_left, input_left -= retval would underflow input_left, right?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -0,0 +1,2127 @@
...
+    _flow: *mut core::Flow,
+    state: &mut DCERPCState,
+    _pstate: *mut std::os::raw::c_void,
+    input: *const u8,
+    input_len: u32,
+    _data: *mut std::os::raw::c_void,
+    flags: u8,
+) -> AppLayerResult {
+    if input_len > 0 && input != std::ptr::null_mut() {
+        let buf = std::slice::from_raw_parts(input, input_len as usize);
```
victorjulien, line 851: `build_slice!`

rust/src/dcerpc/dcerpc.rs
inashivb: input_left I think happened because of replacing a lot of occurences while renaming things but it would still be incorrect. I think neither of the cases reached this part so nothing failed. I think it should be `retval > 0 && retval < input_len`

rust/src/common.rs
inashivb: rustfmt does not seem to change anything here. What do you think should be changed?

victorjulien: I see 2 space indent instead the expected 4

rust/src/dcerpc/dcerpc.rs
inashivb: I'll add comments in the code. But,
Bytes parsed (DCERPCBind): `input.len() - leftover_bytes.len()`
Bytes parsed per bindctxitem: `retval * numctxitems` (`retval` would always be 44 if parser works fine.)
Total bytes parsed: Bytes parsed (DCERPCBind) + Bytes parsed per bindctxitem

rust/src/common.rs
inashivb: I see the same indent in other macro as well `take_until_and_consume`. Should I change for both?

rust/src/common.rs
inashivb: `$item` is just a bool value. So true or false would be printed in its place. Do we want that? Panic would probably tell us where it occurred `thread 'dcerpc::dcerpc::tests::test_handle_bindctxitem' panicked at 'Value should not be NULL! true', src/dcerpc/dcerpc.rs:304:9`
What do you think?

inashivb:
Replaced with #4915

rust/src/dcerpc/dcerpc.rs
victorjulien: This use of retval is confusing, as it only takes the `retval` from the last loop iteration. So I'd suggest either updating a `consumed` variable from within the loop, or replacing the retval by a hardcoded value (44 in this case).

rust/src/common.rs
victorjulien: I'm fine with this (+ requested change from newer PR)

rust/src/common.rs
jasonish: Rustfmt does not format inside macros.

PR #4823, opened 2020-04-15
detect: Add byte_math detector

jlucovsky:
Continuation of #4805 

This PR adds `byte_math` to the suite of `byte` functions with a strict implementation based on the Snort documents.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2311](https://redmine.openinfosecfoundation.org/issues/2311)

Describe changes:
- Address review comments

PR #4911, opened 2020-05-02
detect: Add byte_math detector

jlucovsky:
Continuation of #4823

This PR adds `byte_math` to the suite of `byte` functions with a strict implementation based on the Snort documents.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2311](https://redmine.openinfosecfoundation.org/issues/2311)

Describe changes:
- Address review comments


src/detect-bytemath.c
```diff
@@ -0,0 +1,1210 @@
...
+
+static int DetectByteMathParseTest02(void)
+{
+    /* bytes value invalid */
+    DetectByteMathData *bmd = DetectByteMathParse(NULL, "bytes 257, offset 2, oper +, "
+                                                  "rvalue 39, result bar", NULL);
+
+    FAIL_IF(bmd != NULL);
+
+    DetectByteMathFree(NULL, bmd);
```
victorjulien, line 782: If read this correctly, this can only get called if `bmd == NULL`?

src/detect-bytemath.c
```diff
@@ -0,0 +1,1210 @@
...
+    PASS;
+}
+
+static int DetectByteMathParseTest02(void)
+{
+    /* bytes value invalid */
+    DetectByteMathData *bmd = DetectByteMathParse(NULL, "bytes 257, offset 2, oper +, "
+                                                  "rvalue 39, result bar", NULL);
+
+    FAIL_IF(bmd != NULL);
```
victorjulien, line 780: `FAIL_IF_NOT_NULL(bmd)` has the same effect, but makes it even more clear that we expect the result to NULL.

src/detect-bytemath.c
victorjulien: If so, this call can be removed.

jlucovsky:
Continued in #4913

PR #4915, opened 2020-05-02
Dcerpc single tx/v3

inashivb:
DCERPC parser in Rust.

S-V PR to go with this: https://github.com/OISF/suricata-verify/pull/226


rust/src/common.rs
```diff
@@ -14,6 +14,24 @@ macro_rules! take_until_and_consume (
...
+  ($item:expr) => {};
+);
+
+
+#[cfg(feature = "debug-validate")]
+#[macro_export]
+macro_rules! debug_validate_bug_on (
+  ($item:expr) => {
+    if $item {
+        panic!("Value should not be NULL!");
```
victorjulien, line 30: `NULL` isn't right here: it triggers on `false`.

victorjulien:
The commit adding `debug_validate_bug_on` should be in its own commit w/o adding checks to the dcerpc parser. These can be part of the normal parser adding commit.

Can you also expand the git commit messages? They are really sparse right now. Especially the commit adding the Rust dceprc parser itself should explain more about how it works, what it implements and such.

rust/src/common.rs
inashivb: Do you mean the panic statement should be changed to something else? It gets triggered if header is None.

inashivb:
> The commit adding `debug_validate_bug_on` should be in its own commit w/o adding checks to the dcerpc parser. These can be part of the normal parser adding commit.
> 
> Can you also expand the git commit messages? They are really sparse right now. Especially the commit adding the Rust dceprc parser itself should explain more about how it works, what it implements and such.

Yes sure. Thanks!

rust/src/common.rs
victorjulien: I think NULL in the string is confusing. It's about a condition being false. NULL suggests a pointer value.

inashivb:
Replaced with #4926

PR #4917, opened 2020-05-04
Fuzz tcpsplit v1

catenacyber:
None

Describe changes:
- Adds a fuzz target with differential fuzzing against TCP splitting
- Adds `const` keyword for some function arguments

Open questions :
- Can we know if for some app layer protocol, it can be seen over TCP ? (how ?)
- Code style

Tested with input file
`alert ssh any any -> any any (msg:"SSH"; ssh.protoversion:2_compat; sid:1;)%0A%00%05%00%124%00%16SSH-2.0-test%0A`
- Ok with current version
- Triggers `abort` with dummy patch 
```
diff --git a/src/app-layer-ssh.c b/src/app-layer-ssh.c
index 910803f4a..662f7770e 100644
--- a/src/app-layer-ssh.c
+++ b/src/app-layer-ssh.c
@@ -343,7 +343,7 @@ static int EnoughData(const uint8_t *input, uint32_t input_len)
 {
     uint32_t u;
     for (u = 0; u < input_len; u++) {
-        if (input[u] == '\r' || input[u] == '\n')
+        if (input[u] == '\r' || input[u] == '\n' || input == '.')
             return TRUE;
     }
     return FALSE;
```

### DIFF
--- a/src/detect-engine.c
+++ b/src/detect-engine.c
@@ -2739,7 +2739,12 @@ TmEcode DetectEngineThreadCtxInit(ThreadVars *tv, void *initdata, void **data)
     memset(det_ctx, 0, sizeof(DetectEngineThreadCtx));
 
     det_ctx->tv = tv;
+#if defined(FUZZ)
+    det_ctx->de_ctx = (DetectEngineCtx *)initdata;
+    det_ctx->de_ctx->ref_cnt++;
+#else
     det_ctx->de_ctx = DetectEngineGetCurrent();
+#endif
     if (det_ctx->de_ctx == NULL) {
 #ifdef UNITTESTS
         if (RunmodeIsUnittests()) {
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2159,11 +2159,13 @@ static void SetupDelayedDetect(SCInstance *suri)
 
 static int LoadSignatures(DetectEngineCtx *de_ctx, SCInstance *suri)
 {
+#ifndef FUZZ
     if (SigLoadSignatures(de_ctx, suri->sig_file, suri->sig_file_exclusive) < 0) {
         SCLogError(SC_ERR_NO_RULES_LOADED, "Loading signatures failed.");
         if (de_ctx->failure_fatal)
             return TM_ECODE_FAILED;
     }
+#endif
 
     return TM_ECODE_OK;
 }
@@ -2282,7 +2284,7 @@ static void PostRunStartedDetectSetup(const SCInstance *suri)
     }
 }
 
-static void PostConfLoadedDetectSetup(SCInstance *suri)
+void PostConfLoadedDetectSetup(SCInstance *suri)
 {
     DetectEngineCtx *de_ctx = NULL;
     if (!suri->disabled_detect) {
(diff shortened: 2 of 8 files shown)


src/detect-engine.c
```diff
@@ -2739,7 +2739,12 @@ TmEcode DetectEngineThreadCtxInit(ThreadVars *tv, void *initdata, void **data)
     memset(det_ctx, 0, sizeof(DetectEngineThreadCtx));
 
     det_ctx->tv = tv;
+#if defined(FUZZ)
```
catenacyber, line 2742: inspired by unit tests mode a few lines below

src/suricata.c
```diff
@@ -2159,11 +2159,13 @@ static void SetupDelayedDetect(SCInstance *suri)
 
 static int LoadSignatures(DetectEngineCtx *de_ctx, SCInstance *suri)
 {
+#ifndef FUZZ
     if (SigLoadSignatures(de_ctx, suri->sig_file, suri->sig_file_exclusive) < 0) {
```
catenacyber, line 2163: Not sure this is the best way

src/tests/fuzz/fuzz_tcpsplit.c
```diff
@@ -0,0 +1,463 @@
...
+        dp: 44818\n\
+        sp: 44818\n\
+    sip:\n\
+      enabled: yes\n\
+";
+
+TcpReassemblyThreadCtx *ra_ctx = NULL;
+uint64_t forceLayer = 0;
+
+static Packet *TestHelperBuildPacketAndFlow(const char *src, const char *dst,
```
catenacyber, line 127: Any good replacement ?

src/tests/fuzz/fuzz_tcpsplit.c
```diff
@@ -0,0 +1,463 @@
...
+
+    return p;
+
+error:
+    SCFree(p);
+    return NULL;
+}
+
+static int TestHelperAddSegmentWithPayload(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx, TcpStream *stream, uint32_t seq, const uint8_t *payload, uint16_t len, Packet *p)
+{
```
catenacyber, line 200: inspired by StreamTcpUTAddSegmentWithPayload

The whole thing is inspired by tests such as `DetectSshVersionTestDetect01` from ssh rust branch

src/tests/fuzz/fuzz_tcpsplit.c
```diff
@@ -0,0 +1,463 @@
...
+    p = TestHelperBuildPacketAndFlow("192.168.1.5", "192.168.1.1", (data[2] << 8) | data[3], (data[4] << 8) | data[5]);
+    if(p == NULL) {
+        goto end;
+    }
+    psplit = TestHelperBuildPacketAndFlow("192.168.1.7", "192.168.1.2", (data[2] << 8) | data[3], (data[4] << 8) | data[5]);
+    if(psplit == NULL) {
+        goto end;
+    }
+    p->flow->protoctx = &ssn;
+    psplit->flow->protoctx = &ssnsplit;
```
catenacyber, line 330: TODO : have these in the helper function

catenacyber:
This is a follow up to generalize bug findings such as https://github.com/OISF/suricata/pull/4792 or https://github.com/OISF/suricata/pull/4792

catenacyber:
This will need corpus generation to serialized format from rule and pcap (and use S-V directories).

Are there rules who are meant to be sensitive to TCP splitting ?
Should we restrict this to app-layer rules ?

PR #4916, opened 2020-05-03
detect: Add byte_math detector

jlucovsky:
Continuation of #4913

This PR adds `byte_math` to the suite of `byte` functions with a strict implementation based on the Snort documents.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2311](https://redmine.openinfosecfoundation.org/issues/2311)

Describe changes:
- Recreates commits such that each can be built independently (in order)
- Adds documentation.

PR #4918, opened 2020-05-05
detect: Add byte_math detector 

jlucovsky:
Continuation of #4916

This PR adds `byte_math` to the suite of `byte` functions with a strict implementation based on the Snort documents.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2311](https://redmine.openinfosecfoundation.org/issues/2311)

Describe changes:
- Fixes issues found when adding `suricata-verify` test
- Changes a warning message that causes S-V to fail. See S-V pr

[Suricata-verify PR#228](https://github.com/OISF/suricata-verify/pull/228)


src/detect-bytejump.c
```diff
@@ -576,14 +577,14 @@ static int DetectBytejumpSetup(DetectEngineCtx *de_ctx, Signature *s, const char
...
-        if (bed_sm == NULL) {
+        DetectByteIndexType index;
+        if (!DetectByteRetrieveSMVar(offset, s, &index)) {
             SCLogError(SC_ERR_INVALID_SIGNATURE, "Unknown byte_extract var "
                        "seen in byte_jump - %s\n", offset);
             goto error;
         }
-        data->offset = ((DetectByteExtractData *)bed_sm->ctx)->local_id;
-        data->flags |= DETECT_BYTEJUMP_OFFSET_BE;
+        data->offset = index;
```
victorjulien, line 586: maybe it would be better to not overload `offset` like this. We could add a new field to the struct or make offset and 'var_id' a union if they can never be used together.

src/detect-isdataat.c
```diff
@@ -240,14 +239,14 @@ int DetectIsdataatSetup (DetectEngineCtx *de_ctx, Signature *s, const char *isda
...
-        if (bed_sm == NULL) {
+        DetectByteIndexType index;
+        if (!DetectByteRetrieveSMVar(offset, s, &index)) {
             SCLogError(SC_ERR_INVALID_SIGNATURE, "Unknown byte_extract var "
                        "seen in isdataat - %s\n", offset);
             goto end;
         }
-        idad->dataat = ((DetectByteExtractData *)bed_sm->ctx)->local_id;
-        idad->flags |= ISDATAAT_OFFSET_BE;
+        idad->dataat = index;
```
victorjulien, line 248: I guess the same idea here.

src/detect-bytemath.c
```diff
@@ -0,0 +1,1196 @@
...
+    sigmatch_table[DETECT_BYTEMATH].RegisterTests = DetectByteMathRegisterTests;
+
+    DetectSetupParseRegexes(PARSE_REGEX, &parse_regex);
+}
+
+int DetectByteMathDoMatch(DetectEngineThreadCtx *det_ctx, const SigMatchData *smd,
+                          const Signature *s, const uint8_t *payload,
+                          uint16_t payload_len, uint64_t rvalue, uint64_t *value, uint8_t endian)
+{
+    DetectByteMathData *data = (DetectByteMathData *)smd->ctx;
```
victorjulien, line 137: const

src/detect-bytemath.c
```diff
@@ -0,0 +1,1196 @@
...
+                          BYTE_BIG_ENDIAN : BYTE_LITTLE_ENDIAN;
+        extbytes = ByteExtractUint64(&val, endianness, data->nbytes, ptr);
+        if (extbytes != data->nbytes) {
+            SCLogDebug("error extracting %d bytes of numeric data: %d",
+                       data->nbytes, extbytes);
+            return 0;
+        }
+    }
+
+    if (extbytes > len) {
```
victorjulien, line 203: is this a condition that can be reached? It seems more like a bug to me. If it is, we should probably use BUG_ON?

src/detect-bytemath.c
```diff
@@ -0,0 +1,1196 @@
...
+
+    int sm_list;
+    if (s->init_data->list != DETECT_SM_LIST_NOTSET) {
+        if (DetectBufferGetActiveList(de_ctx, s) == -1)
+            goto error;
+
+        sm_list = s->init_data->list;
+
+        if (data->flags & DETECT_BYTEMATH_FLAG_RELATIVE) {
+            prev_pm = DetectGetLastSMFromLists(s, DETECT_CONTENT, DETECT_PCRE, -1);
```
victorjulien, line 592: should we check the result here?

jlucovsky: Good question. `prev_pm` is null-checked later.

src/detect-bytejump.c
jlucovsky: See `DetectByteRetrieveSMVar` ... although the var name is `index` ... the value is the same.

src/detect-bytemath.c
victorjulien: Are you referring to the check at line 683? That would suggest its okay to have a byte_math setting of 'relative', while there is no previous match to be relative to.

jlucovsky: Yes -- that's the check I'm referring to. 

I'll have to modify the other `byte-` methods to return an error in this case as they use this pattern. I'll update my pr.

victorjulien: In the other keywords it may be deliberate, not sure. May be for compatibility reasons with older rules. Guess we can test with the ET set.

jlucovsky:
Continued in #4981

PR #4922, opened 2020-05-06
detect/pcre: Use the keyword context for JIT stack

jlucovsky:
When PCRE `jit` is available, store the JIT stack in the keyword context
instead of on a global id. This ensures proper cleanup and
re-initialization over a rule reload.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3681)[https://redmine.openinfosecfoundation.org/issues/3681)

Describe changes:
- Eliminated global id for JIT stack 
- Use the keyword context to track the JIT stack

### DIFF
--- a/src/detect-pcre.c
+++ b/src/detect-pcre.c
@@ -82,19 +82,18 @@ static int pcre_use_jit = 1;
 #define PCRE_JIT_MIN_STACK 32*1024
 #define PCRE_JIT_MAX_STACK 512*1024
 
-static int g_jitstack_thread_ctx_id = -1;
-
 #endif
 
 /* \brief Helper function for using pcre_exec with/without JIT
  */
-static int DetectPcreExec(DetectEngineThreadCtx *det_ctx, DetectParseRegex *regex, const char *str, const size_t strlen,
+static int DetectPcreExec(DetectEngineThreadCtx *det_ctx, DetectPcreData *pd, DetectParseRegex *regex, const char *str, const size_t strlen,
                             int start_offset, int options,  int *ovector, int ovector_size)
 {
 #ifdef PCRE_HAVE_JIT_EXEC
-    if (det_ctx && g_jitstack_thread_ctx_id != -1) {
+    if (det_ctx && pd->thread_ctx_id != -1) {
         pcre_jit_stack *jit_stack = (pcre_jit_stack *)
-            DetectThreadCtxGetKeywordThreadCtx(det_ctx, g_jitstack_thread_ctx_id);
+            DetectThreadCtxGetKeywordThreadCtx(det_ctx, pd->thread_ctx_id);
+        SCLogDebug("[%s:%d] Using jit_stack %p", __FUNCTION__, __LINE__, jit_stack);
         if (jit_stack) {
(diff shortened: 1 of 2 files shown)


src/detect-pcre.c
```diff
@@ -836,12 +835,14 @@ static void *DetectPcreThreadInit(void *data /*@unused@*/)
     if (jit_stack == NULL) {
         SCLogWarning(SC_WARN_PCRE_JITSTACK, "Unable to allocate PCRE JIT stack; will continue without JIT stack");
     }
+    SCLogDebug("[%s:%d] Using jit_stack %p", __FUNCTION__, __LINE__, jit_stack);
```
victorjulien, line 838: I don't understand why this explicitly adds `__FUNCTION__, __LINE__`, something `SCLogDebug` already provides?

src/detect-pcre.c
```diff
@@ -866,14 +867,12 @@ static int DetectPcreSetup (DetectEngineCtx *de_ctx, Signature *s, const char *r
...
-                DetectPcreThreadFree, 1);
-        if (g_jitstack_thread_ctx_id == -1) {
-            SCLogWarning(SC_WARN_REGISTRATION_FAILED,
-                    "Unable to register JIT stack functions. PCRE JIT stack disabled.");
-        }
+    pd->thread_ctx_id = DetectRegisterThreadCtxFuncs(de_ctx, "pcre",
+            DetectPcreThreadInit, (void *)pd,
+            DetectPcreThreadFree, 1);
+    if (pd->thread_ctx_id == -1) {
+        SCLogWarning(SC_WARN_REGISTRATION_FAILED,
```
victorjulien, line 874: is this now risking generating a warning for each pcre occurrence in the rules? That would be very noisy.

src/detect-pcre.h
```diff
@@ -39,6 +39,8 @@
 typedef struct DetectPcreData_ {
     /* pcre options */
     DetectParseRegex parse_regex;
+
+    int thread_ctx_id;
```
victorjulien, line 43: Documentation missing. I would also suggest adding `jit` to the name to indicate it's only used by jit support. Should it also be in `PCRE_HAVE_JIT_EXEC` guards?

src/detect-pcre.c
```diff
@@ -777,7 +776,7 @@ static int DetectPcreParseCapture(const char *regexstr, DetectEngineCtx *de_ctx,
     while (1) {
         SCLogDebug("\'%s\'", regexstr);
 
-        ret = DetectPcreExec(NULL, &parse_capture_regex, regexstr, strlen(regexstr), 0, 0, ov, MAX_SUBSTRINGS);
+        ret = DetectPcreExec(NULL, NULL, &parse_capture_regex, regexstr, strlen(regexstr), 0, 0, ov, MAX_SUBSTRINGS);
```
victorjulien, line 779: Passing in these NULLs effectively disables the JIT support here, right? Is that the only reason to deal with NULLs to `DetectPcreExec`. If so, it might be cleaner and faster to have dedicated logic for parsing here. That way we can keep the routine used in the packet runtime leaner and cleaner.

jlucovsky:
Continued in #4931

PR #4928, opened 2020-05-08
Http2 v7

catenacyber:

DRAFT for discussion
Not to be merged

Describe changes:
- Adds basic HTTP2 parser

Much more work to be done :
 + more commit squashing
 + dynamic table for headers
 + parse remaning frame types
 + tests
 + robustness against evasions

Modifies #4853 with 
- headers parsing


src/app-layer-http2.c
```diff
@@ -0,0 +1,71 @@
...
+#include "app-layer-parser.h"
+
+#include "app-layer-http2.h"
+#include "rust.h"
+
+static int HTTP2RegisterPatternsForProtocolDetection(void)
+{
+    //TODO is this too restrictive and can be evaded ?
+    /* TODO investigate why this makes AppLayerTest09 fail/leak
+     * and not leak if we only give a 5-bytes pattern
```
catenacyber, line 42: Any thoughts here on the length on the pattern ?

rust/src/http2/parser.rs
```diff
@@ -0,0 +1,1483 @@
...
+
+#[derive(Clone)]
+pub struct HTTP2FrameHeaderBlock {
+    pub name: String,
+    pub value: String,
+}
+
+fn http2_parse_headers_block_indexed(input: &[u8]) -> IResult<&[u8], HTTP2FrameHeaderBlock> {
+    //TODO ask why we need this over error[E0282]: type annotations needed for `((&[u8], usize), O1)`
+    fn parser(input: &[u8]) -> IResult<&[u8], (u8, HTTP2FrameHeaderBlock)> {
```
catenacyber, line 470: Any rust/nom style advice here ?

rust/src/http2/parser.rs
```diff
@@ -0,0 +1,1483 @@
...
+            "SETTINGS_INITIAL_WINDOW_SIZE" => Ok(HTTP2SettingsId::SETTINGSINITIALWINDOWSIZE),
+            "SETTINGS_MAX_FRAME_SIZE" => Ok(HTTP2SettingsId::SETTINGSMAXFRAMESIZE),
+            "SETTINGS_MAX_HEADER_LIST_SIZE" => Ok(HTTP2SettingsId::SETTINGSMAXHEADERLISTSIZE),
+            _ => Err(format!("'{}' is not a valid value for HTTP2SettingsId", s)),
+        }
+    }
+}
+
+//TODO move elsewhere generic
+#[derive(PartialEq, Debug)]
```
catenacyber, line 1129: Thoughts about this ?

rust/src/http2/http2.rs
```diff
@@ -0,0 +1,726 @@
...
+                let start = self.response_frame_size as usize;
+                input = &input[start..];
+                self.response_frame_size = 0;
+            }
+        }
+        //then parse all we can
+        while input.len() > 0 {
+            match parser::http2_parse_frame_header(input) {
+                Ok((rem, head)) => {
+                    let mut tx = self.new_tx();
```
victorjulien, line 361: the goal of a tx should be to match a response to a request, so not create txs for responses, but look up the request side tx.

src/app-layer-http2.c
victorjulien: What is the max size you can use before other things break?

victorjulien:
The current PR implements the HTTP2 header parsing, which is a bit like a transport layer on top of which there will be the HTTP dialect. How do you see this HTTP layer wrt transactions, detection and logging?

The ultimate goal is to see if we can overload the current http keywords in suricata to also support the http/2 dialect. I'm not entirely positive yet this is possible, but it seems the dialects are similar enough.

rust/src/http2/http2.rs
catenacyber: I know :-) this comes later in my TODO list.
I think I am getting a sense of how it should work with HTTP2

catenacyber:
> The current PR implements the HTTP2 header parsing, which is a bit like a transport layer on top of which there will be the HTTP dialect. How do you see this HTTP layer wrt transactions, detection and logging?

I was rather thinking to make it first standalone (independent of HTTP1)

But I guess it makes sense to overload HTTP1 when we can, and create new keywords/log sections when the stuff is new...

One tricky part will be to make sure that later changes to the code to one HTTP version are consistent with the code in the other http version

I will take a deeper look at this after having a good transaction handling...

src/app-layer-http2.c
catenacyber: 16 bytes pattern "PRI * HTTP/2.0\r\n" works
18 bytes pattern  "PRI * HTTP/2.0\r\n\r\n" does not

victorjulien: Ok lets go for the 16 byte pattern for now. Maybe add a note to a relevant ticket, or add a new ticket, for the length issue.

rust/src/http2/parser.rs
victorjulien: cc @jasonish

rust/src/http2/parser.rs
victorjulien: About what specifically?

src/app-layer-http2.c
catenacyber: Ok

rust/src/http2/parser.rs
catenacyber: Should I move this code elsewhere ?
And if it is the case, where precisely ?

This code is basically the Rust translation of detect-engine-uint.c

catenacyber:
> How do you see this HTTP layer wrt transactions, detection and logging?

And that brings forth the question from https://redmine.openinfosecfoundation.org/issues/1275#note-18

rust/src/http2/parser.rs
jasonish: If the question is about the style of function inside a function, I think not using nom macros, and instead using nom functions (new nom style) would give you a more pleasing style.  I have yet to master converting from macros to functions though.

catenacyber: I had to use this function (based on some other parser in Suricata) because if I use instead of 

`let (i2, indexed) = parser(input)?;` 

the corresponding function

`let (i2, indexed) = bits!(...)?;`

It does not compile with error E0282

jasonish: Probably needs some type annotation like seen in this example:
https://docs.rs/nom/5.1.1/nom/bits/fn.bits.html#example

My quick attempts to convert it failed as well.

catenacyber: Leaving it this way is fine I suppose then ?

PR #4920, opened 2020-05-05
Replace Jansson with Rust JsonBuilder - v27

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/4901

Changes from last PR:
- convert http
- convert tls
- convert file/fileinfo
- convert anomaly

PR #4929, opened 2020-05-08
JsonBuilder - v28

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/4920

Changes from last PR:
- Handle NULL values (https://github.com/OISF/suricata/pull/4920#issuecomment-625717218). We copy libjannson behaviour here and don't log the field if the value is NULL.
- Cleanup string escaping to match that of other JSON encoders. This mainly means that where we used to log "\\x00" we now long "\u0000", and so on.
- Use "append" instead of "add" for all functions that append to arrays (had mixed append and add before).
- Remove invoking the formatter to render hex bytes (is very slow)
- Make commit messages more consistent.

Notes carried over from last PR:
- Optimize constant output (https://github.com/OISF/suricata/pull/4920#discussion_r422019047)


victorjulien:
I'm seeing a case where `jq` couldn't parse a dns record. It comes down to:
```json
"dns":{"version":2,"type":"answer","id":23672,"flags":"8183","qr":true,"rd":true,"ra":true,"rrname":}}
```
Note that there is no value for `rrname`.

`jq` says:
`parse error: Unmatched '}' at line 602197, column 302`

jasonish:
> `parse error: Unmatched '}' at line 602197, column 302`

Ok, I think I can see how this can happen. Is this a pcap you can share? Would be good to see the actual contents of the rrname that is tripping it up,

victorjulien:
> > `parse error: Unmatched '}' at line 602197, column 302`
> 
> Ok, I think I can see how this can happen. Is this a pcap you can share? Would be good to see the actual contents of the rrname that is tripping it up,

Not easily. It's >100GiB pcap. Happy to run a test branch to validate your ideas.

jasonish:
Continued at https://github.com/OISF/suricata/pull/4936.

PR #4925, opened 2020-05-06
Testimony feature 3675 v4

MalakhatkoVadym:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3675

Describe changes:
- Adds testimony packet source to suricata

Testimony is a single-machine, multi-process architecture for sharing AF_PACKET data across processes.
https://github.com/google/testimony

Modifies [#4924](https://github.com/OISF/suricata/pull/4924) by fixing Deref counter for TestimonySocketConfig.

### Testing
I tested with 2 testimonyd's running: 

sudo testimonyd --config=/etc/testimony.conf
```
[
  {
      "SocketName": "/tmp/testimony.sock"
    , "Interface": "eth0"
    , "BlockSize": 1048576
    , "NumBlocks": 16
    , "FanoutSize": 4
    , "BlockTimeoutMillis": 1000
    , "User": "root"
  }
]
```
interface - "eth0"
socket path - "/tmp/testimony.sock"
Fanout size - "4" (packets are balanced between 4 memory blocks, so suricata can use receive thread for each memory block).

sudo testimonyd --config=/etc/testimony2.conf
```
[
  {
      "SocketName": "/tmp/testimony2.sock"
    , "Interface": "eth1"
    , "BlockSize": 1048576
    , "NumBlocks": 16
    , "BlockTimeoutMillis": 1000
    , "FanoutSize": 1
    , "User": "root"
  }
]

```
interface - "eth1"
socket path - "/tmp/testimony2.sock"
Fanout size - "1" (packets are not balanced between memory blocks).

sudo suricata -c /etc/suricata/suricata.yaml -k none --testimony=/tmp/testimony.sock --testimony=/tmp/testimony2.sock

suricata.yaml :

```
testimony:
  - socket: /tmp/testimony.sock
    fanout-size: 4
```

PR #4930, opened 2020-05-09
Testimony feature 3675 v5

MalakhatkoVadym:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3675

Describe changes:
- Adds testimony packet source to suricata

Testimony is a single-machine, multi-process architecture for sharing AF_PACKET data across processes.
https://github.com/google/testimony

Modifies [#4925](https://github.com/OISF/suricata/pull/4925) by adding a separate page with documentation about testimony.

### Testing
I tested with 2 testimonyd's running: 

sudo testimonyd --config=/etc/testimony.conf
```
[
  {
      "SocketName": "/tmp/testimony.sock"
    , "Interface": "eth0"
    , "BlockSize": 1048576
    , "NumBlocks": 16
    , "FanoutSize": 4
    , "BlockTimeoutMillis": 1000
    , "User": "root"
  }
]
```
interface - "eth0"
socket path - "/tmp/testimony.sock"
Fanout size - "4" (packets are balanced between 4 memory blocks, so suricata can use receive thread for each memory block).

sudo testimonyd --config=/etc/testimony2.conf
```
[
  {
      "SocketName": "/tmp/testimony2.sock"
    , "Interface": "eth1"
    , "BlockSize": 1048576
    , "NumBlocks": 16
    , "BlockTimeoutMillis": 1000
    , "FanoutSize": 1
    , "User": "root"
  }
]

```
interface - "eth1"
socket path - "/tmp/testimony2.sock"
Fanout size - "1" (packets are not balanced between memory blocks).

sudo suricata -c /etc/suricata/suricata.yaml -k none --testimony=/tmp/testimony.sock --testimony=/tmp/testimony2.sock

suricata.yaml :

```
testimony:
  - socket: /tmp/testimony.sock
    fanout-size: 4
```


norg:
Thanks for the extended documentation, that will be helpful. Can you explain a bit more about the project? I saw that you discussed with us at IRC and I mentioned that I was wondering if testimony was abandoned due to last commits in 2018 but you seem to have taken care of it now?

I ask because adding such features also requires maintenance and also requires the underlying tools to be mainainted.

MalakhatkoVadym:
> Thanks for the extended documentation, that will be helpful. Can you explain a bit more about the project? I saw that you discussed with us at IRC and I mentioned that I was wondering if testimony was abandoned due to last commits in 2018 but you seem to have taken care of it now?
> 
> I ask because adding such features also requires maintenance and also requires the underlying tools to be mainainted.

@norg The main goal is to have this issue fixed https://redmine.openinfosecfoundation.org/issues/2219. There is a product called stenographer https://github.com/google/stenographer. And I'm currently working on the integration of Suricata alerts with it, so pcap files with alerted packets would be saved with additional information about alerts. And the most efficient way to have Suricata and Stenographer working together is Testimony. (there is also an option to use pf-ring dna cluster, but it's very hard to configure https://www.ntop.org/pf_ring/accelerating-suricata-with-pf_ring-dna/)

So, I will support 'suricata' 'testimony' and 'testimony' itself for a long time. And will test suricata+testimony+stenographer configuration.

norg:
So you would run this on the same machine? That's interesting, kinda curious to see this from a performance perspective. Another approach would be to have a dedicated system to just run stenographer and set up some sort of communication between those.

MalakhatkoVadym:
> So you would run this on the same machine? That's interesting, kinda curious to see this from a performance perspective. Another approach would be to have a dedicated system to just run stenographer and set up some sort of communication between those.

There'll be such an option also (to run stenographer on a dedicated machine). Stenographer has some kind of rest api (stenocurl and stenoread tools are used for interaction with it). And I will add a possibility to configure host(ip) and port of stenographer. But in terms of running everything on one host, yes, Testimony is the best option.

MalakhatkoVadym:
Added more documentation and fixed typo.

PR #4932, opened 2020-05-11
dns: Remove parser buffering code

jlucovsky:
This PR removes the buffering code from the DNS TCP parsers.

`make check` Rust tests succeed; `suricata-verify` tests also succeed.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3538](https://redmine.openinfosecfoundation.org/issues/3538)

Describe changes:
- Move DNS query splice test to Rust.


src/detect-dns-query.c
```diff
@@ -1256,13 +929,9 @@ static void DetectDnsQueryRegisterTests(void)
     UtRegisterTest("DetectDnsQueryTest01", DetectDnsQueryTest01);
     UtRegisterTest("DetectDnsQueryTest02", DetectDnsQueryTest02);
     UtRegisterTest("DetectDnsQueryTest03 -- tcp", DetectDnsQueryTest03);
-    UtRegisterTest("DetectDnsQueryTest04 -- tcp splicing",
-                   DetectDnsQueryTest04);
```
catenacyber, line 1260: Why remove these tests ?

jlucovsky: A replacement for `DetectBsizeSigTest01` was added to the Rust test set. These tests require TCP buffering which isn't possible when directly invoking the application layer parsing.

catenacyber: > These tests require TCP buffering which isn't possible when directly invoking the application layer parsing.

It is possible as I did recently when translating the SSH parser to Rust and using ApplayerResult::Incomplete
You can see in `DetectSshVersionTestDetect01` cf https://github.com/OISF/suricata/blob/master/src/detect-ssh-proto-version.c#L339

And I think it is particularly useful to keep these tests when doing such a change ;-)

jasonish: From the DNS perspective I think its probably OK to not test this as a unit test. We should have tests that that check the correct response is returned when the data is incomplete.

While there should be some larger tests that test that the buffering +  protocol parsing does work, it probably does not need to be implemented for every protocol.

What might be more intersting is suricata-verify tests with pcaps that cause the parsers to be called with incomplete data.

victorjulien: Related: https://redmine.openinfosecfoundation.org/issues/3725

catenacyber: I agree that they can be replaced by suricata-verify tests.

I wish I had known this ticket while translating the SSH tests ;-)

victorjulien:
Can you rebase and resubmit?

jlucovsky:
Continued in #5020

PR #4933, opened 2020-05-11
fuzz: improves sigpcap target

catenacyber:
None

Describe changes:
- improves sigpcap fuzz target so that it can cover alert generation
ie in function DetectRun, get past `scratch.sgh == NULL` condition

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2161,8 +2161,10 @@ static int LoadSignatures(DetectEngineCtx *de_ctx, SCInstance *suri)
 {
     if (SigLoadSignatures(de_ctx, suri->sig_file, suri->sig_file_exclusive) < 0) {
         SCLogError(SC_ERR_NO_RULES_LOADED, "Loading signatures failed.");
+#ifndef FUZZ
         if (de_ctx->failure_fatal)
             return TM_ECODE_FAILED;
+#endif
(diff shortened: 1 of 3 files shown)


src/suricata.c
```diff
@@ -2161,8 +2161,10 @@ static int LoadSignatures(DetectEngineCtx *de_ctx, SCInstance *suri)
 {
     if (SigLoadSignatures(de_ctx, suri->sig_file, suri->sig_file_exclusive) < 0) {
         SCLogError(SC_ERR_NO_RULES_LOADED, "Loading signatures failed.");
+#ifndef FUZZ
```
victorjulien, line 2164: Can't the fuzz target simply set `de_ctx->failure_fatal`. I want to avoid such ifdefs whenever possible.

victorjulien: Think all you'd need to do is `(void)ConfSetFinal("engine.init-failure-fatal", "0");`

victorjulien: Though I'm surprised this is enabled by default? Or did you add it to the target explicitly somehow?

catenacyber: Well seen, I was just afraid so I put this ifdef without testing first if it worked without it, testing now

PR #4937, opened 2020-05-13
Byteextractstr/part2/v2

inashivb:

### DIFF
--- a/src/util-classification-config.c
+++ b/src/util-classification-config.c
@@ -34,6 +34,7 @@
 #include "util-error.h"
 #include "util-debug.h"
 #include "util-fmemopen.h"
+#include "util-byte.h"
 
 /* Regex to parse the classtype argument from a Signature.  The first substring
  * holds the classtype name, the second substring holds the classtype the
@@ -281,11 +282,9 @@ int SCClassConfAddClasstype(DetectEngineCtx *de_ctx, char *rawstr, uint16_t inde
         SCLogInfo("pcre_copy_substring() failed");
         goto error;
     }
-    if (strlen(ct_priority_str) == 0) {
+    if (strlen(ct_priority_str) == 0 || StringParseInt32(&ct_priority, 10, 0, (const char *)ct_priority_str) < 0) {
         goto error;
     }
-
-    ct_priority = atoi(ct_priority_str);
 
     /* Create a new instance of the parsed Classtype string */
     ct_new = SCClassConfAllocClasstype(ct_id, ct_name, ct_desc, ct_priority);
--- a/src/util-cpu.c
+++ b/src/util-cpu.c
@@ -27,6 +27,7 @@
 #include "util-error.h"
 #include "util-debug.h"
 #include "util-cpu.h"
+#include "util-byte.h"
 
 /**
  * Ok, if they should use sysconf, check that they have the macro's
@@ -75,9 +76,15 @@ uint16_t UtilCpuGetNumProcessorsConfigured(void)
 
     return (uint16_t)nprocs;
 #elif OS_WIN32
-	long nprocs = -1;
-	const char* envvar = getenv("NUMBER_OF_PROCESSORS");
-	nprocs = (NULL != envvar) ? atoi(envvar) : 0;
+    int64_t nprocs = 0;
+    const char* envvar = getenv("NUMBER_OF_PROCESSORS");
+    if (envvar != NULL) {
+        if (StringParseInt64(&nprocs, 10, 0, envvar) < 0) {
+            SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for number of "
+                         "processors: %s, resetting to 0", envvar);
+            nprocs = 0;
+        }
+    }
     if (nprocs < 1) {
         SCLogError(SC_ERR_SYSCALL, "Couldn't retrieve the number of cpus "
                    "configured from the NUMBER_OF_PROCESSORS environment variable");
--- a/src/util-log-redis.c
+++ b/src/util-log-redis.c
@@ -26,6 +26,7 @@
 #include "suricata-common.h" /* errno.h, string.h, etc. */
 #include "util-log-redis.h"
 #include "util-logopenfile.h"
+#include "util-byte.h"
 
 #ifdef HAVE_LIBHIREDIS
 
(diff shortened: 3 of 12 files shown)


src/util-classification-config.c
```diff
@@ -281,12 +282,10 @@ int SCClassConfAddClasstype(DetectEngineCtx *de_ctx, char *rawstr, uint16_t inde
         SCLogInfo("pcre_copy_substring() failed");
         goto error;
     }
-    if (strlen(ct_priority_str) == 0) {
+    if (strlen(ct_priority_str) == 0 || StringParseInt32(&ct_priority, 10, 0, (const char *)ct_priority_str) < 0) {
```
victorjulien, line 285: Is this `strlen(ct_priority_str) == 0` check still necessary? How will `StringParseInt32` respond to a 0-length input string?

src/util-cpu.c
```diff
@@ -75,9 +76,15 @@ uint16_t UtilCpuGetNumProcessorsConfigured(void)
...
-	long nprocs = -1;
-	const char* envvar = getenv("NUMBER_OF_PROCESSORS");
-	nprocs = (NULL != envvar) ? atoi(envvar) : 0;
+    int64_t nprocs = 0;
+    const char* envvar = getenv("NUMBER_OF_PROCESSORS");
+    if (envvar != NULL) {
+        if (StringParseInt64(&nprocs, 10, 0, envvar) < 0) {
+            SCLogWarning(SC_ERR_INVALID_VALUE, "Invalid value for number of "
+                         "processors: %s, resetting to 0", envvar);
+            nprocs = 0;
```
victorjulien, line 85: resseting to 0 will trigger the next check right away? `if (nprocs < 1) {`.

src/util-log-redis.c
```diff
@@ -514,7 +515,10 @@ int SCConfLogOpenRedis(ConfNode *redis_node, void *lf_ctx)
         SCLogError(SC_ERR_MEM_ALLOC, "Error allocating redis server string");
         exit(EXIT_FAILURE);
     }
-    log_ctx->redis_setup.port = atoi(redis_port);
+    if (StringParseU16RangeCheck(&log_ctx->redis_setup.port, 10, 0, (const char *)redis_port, 0, 65534) < 0) {
```
victorjulien, line 518: Why is 65535 excluded from this check?

src/util-napatech.c
```diff
@@ -800,10 +800,21 @@ static uint32_t CountWorkerThreads(void)
 
                         char copystr[16];
                         strlcpy(copystr, lnode->val, 16);
-
-                        ByteExtractStringUint8(&start, 10,  0, copystr);
-                        ByteExtractStringUint8(&end, 10,  0, strchr(copystr, '-') + 1);
-
+                        if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
+                            SCLogError(SC_ERR_INVALID_VALUE, "Napatech invalid"
```
victorjulien, line 804: Please use `FatalError` to replace the `SCLogError(..); exit(EXIT_FAILURE);` pattern.

src/util-napatech.c
```diff
@@ -948,16 +959,29 @@ int NapatechGetStreamConfig(NapatechStreamConfig stream_config[])
...
                     char copystr[16];
                     strlcpy(copystr, stream->val, 16);
-                    ByteExtractStringUint16(&start, 10,  0, copystr);
-                    ByteExtractStringUint16(&end, 10,  0, strchr(copystr, '-') + 1);
+                    if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
+                        SCLogError(SC_ERR_INVALID_VALUE, "Napatech invalid "
+                                   "stream_id start: %s", copystr);
+                        exit(EXIT_FAILURE);
+                    }
+                    if (StringParseUint8(&end, 10, 0, (const char *) (strchr(copystr, '-') + 1)) < 0) {
```
victorjulien, line 967: This looks risky. What if `strchr` fails?

src/util-proto-name.c
```diff
@@ -60,10 +61,9 @@ void SCProtoNameInit()
             if (proto_ch == NULL)
                 continue;
 
-            int proto = atoi(proto_ch);
-            if (proto >= 255)
+            uint8_t proto;
+            if (StringParseU8RangeCheck(&proto, 10, 0, (const char *)proto_ch, 0, 254) < 0)
```
victorjulien, line 65: this looks odd. I would expect 255 to be within range.

src/util-classification-config.c
inashivb: It'll throw an error `Invalid numeric value`. Removing the check.

src/util-cpu.c
inashivb: Yes and it'll give an additional error message to make things clearer:
```
 89         SCLogError(SC_ERR_SYSCALL, "Couldn't retrieve the number of cpus "
 90                    "configured from the NUMBER_OF_PROCESSORS environment variable");
```
Should the same error message and return statement be in the `StringParse` check block?

src/util-log-redis.c
inashivb: Shouldn't have been. Fixed.

src/util-napatech.c
inashivb: Changing for this one. Rest will be fixed later by https://redmine.openinfosecfoundation.org/issues/3188

src/util-napatech.c
inashivb: It indeed is. Fixed.

src/util-cpu.c
victorjulien: I don't like the sequence of "hey you have an error we're setting something to 0" directly followed by "hey 0 isn't valid, we're going to use 1". If 0 isn't valid we shouldn't be setting it.

So I think the Warning should say we're setting to 1 and then set to 1 :)

inashivb: > I don't like the sequence of "hey you have an error we're setting something to 0" directly followed by "hey 0 isn't valid, we're going to use 1". If 0 isn't valid we shouldn't be setting it.
> 
Yes, that does look weird.

> So I think the Warning should say we're setting to 1 and then set to 1 :)
Didn't get this part. 1 is a valid value for number of processors but we failed so why should we set to 1? Or was it a figure of speech?

victorjulien: Ah I misread the original code. Forget the setting to 1 part. Still this check should just issue a single warning if the value is invalid.

inashivb: OK. Thanks!

inashivb:
Replaced with #4959

PR #3701, opened 2019-03-04
Add to docs 2081 v2

x62smith:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues/2081) ticket:

Original PR:
https://github.com/OISF/suricata/pull/3695#discussion_r261888898

Describe changes:
Corrected spelling and made punctuation in the tables uniform.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

PR #4941, opened 2020-05-15
Log tenant_id for all eve-json messages

internetionals:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#3701](https://redmine.openinfosecfoundation.org/issues/3701)

Describe changes:
- Log `tenant_id` for all eve-log types

(wouldn't work, but our CI had no problems with the patched version, for what that is worth)

Traceback (most recent call last):
  File "qa/prscript.py", line 254, in <module>
    ret = TestGithubSync(args.branch)
  File "qa/prscript.py", line 144, in TestGithubSync
    request = urllib2.Request(GITHUB_BASE_URI + username + "/" + args.repository + "/commits?sha=" + branch + "&per_page=1")
TypeError: cannot concatenate 'str' and 'NoneType' objects

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -455,6 +455,9 @@ void EveAddCommonOptions(const OutputJsonCommonSettings *cfg,
     if (cfg->include_community_id && f != NULL) {
         CreateEveCommunityFlowId(js, f, cfg->community_id_seed);
     }
+    if (f != NULL && f->tenant_id > 0) {
+        jb_set_uint(js, "tenant_id", f->tenant_id);
+    }
 }
 
 /**
--- a/doc/userguide/configuration/multi-tenant.rst
+++ b/doc/userguide/configuration/multi-tenant.rst
@@ -215,3 +215,10 @@ unregister-tenant-handler <tenant id> vlan <vlan id>
 
 The registration of tenant and tenant handlers can be done on a
 running engine.
+
+Eve JSON output
+---------------
+
+When multi-tenant support is configured and the detect engine is active then
+all EVE-types that report based on flows will also report the corresponding
+``tenant_id`` for events matching a tenant configuration.
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -501,5 +501,12 @@ YAML::
       # Seed value for the ID output. Valid values are 0-65535.
       community-id-seed: 0
 
+Multi Tenancy
+-------------
+
+Suricata can be configured to support multiple tenants with different detection
+engine configurations. When these tenants are configured and the detection
+engine is running then all EVE logging will also report the ``tenant_id`` field
+for traffic matching a specific tenant.
 
 .. _deprecation policy: https://suricata-ids.org/about/deprecation-policy/


jasonish:
@victorjulien Could we have a tenant_id of 0 when multi-tenancy is enabled? If so, I think it would make sense to log the tenant_id of 0, if there are other tenants..  However, if a tenant_id of 0 is not possible with multi-tenancy enabled, then this is OK.

internetionals:
As far as I understand it `tenant_id == 0` is the used to indicate no specific tenant (so this is true in the non-multi-tenant case) in which case we don't want to log this field.

In the multi-tenant case this field could possibly also be `0` when a tenant is configured as `default`. My understanding is that this can happen when one configures vlan-based multi-tenancy and one receives traffic via a vlan that is not included in the tenant mapping.

src/output-json.c
```diff
@@ -705,6 +705,9 @@ void CreateJSONFlowId(json_t *js, const Flow *f)
     if (f->parent_id) {
         json_object_set_new(js, "parent_id", json_integer(f->parent_id));
     }
+    if (f->tenant_id > 0) {
```
jlucovsky, line 708: nit: Suggest `if (f->tenant)`

internetionals: All other if-statements doing anything with `tenant_id` are using `!= 0`, `== 0` and `> 0` explicitly. Perhaps `!= 0` would be more idiomatic? I originally based this change on the code from `src/out-json-alert.c` which is one of `> 0` users.

victorjulien, line 708: I don't think this belongs in this function. Probably the best place would be `JsonAddCommonOptions`.

I am a bit conflicted about whether its a good idea at all. I see why you'd want it, but at the same time the tentants are detection engine construct, not a global construct (at this time). So it really only has meaning in detection engine related output, which means alerts. Trying to think if we can just use it like this, or that we need to consider edge cases like can 2 sides of a flow have different tenants?

internetionals: > I am a bit conflicted about whether its a good idea at all. I see why you'd want it, but at the same time the tentants are detection engine construct, not a global construct (at this time).

As far as I have understood from the code the tenant-id is determined during receive time and passed along to the flow logic. The detect engine effectively pulls it out of the flow data when it detects something. Isn't the flow data central to all engines?

This is also the reason why I added this field in this function as this is the place where the `flow_id` gets added.

> Trying to think if we can just use it like this, or that we need to consider edge cases like can 2 sides of a flow have different tenants?

I haven't looked into the flow bookkeeping. If the fields that the multi-detect selections are based on (or the tenant_id themselves) are part of what makes up a unique flow then a flow would be unique. If this doesn't hold true then this is equally broken for the detection engine as half of the flow could have been received using vlan A and half over vlan B and the detection engine also only flags one of the two (probably using the tenant_id of the packet that instantiated the flow).

internetionals: @victorjulien: Do you see a way forward for this change?

I can look at moving this behind the "JsonAddCommonOptions" logic.

But I don't understand how the current logic works regarding the "the tentants are detection engine construct, not a global construct". We want to use it for things like http and dns eve logging and I'm trying to see how these event types would have different `tenant_id` values then alerts from the detection engine that use the http or dns decoder results.
Or put another way: what would it entail for this to be a global construct? Perhaps we can help with that.

victorjulien: Other than in the case of processing a pcap using unix socket the tenants are only assigned in the detection engine. So if you'd run suri with `--disable-detection` it wouldn't assign them.

That said I don't have an objection to logging this when its set (meaning > 0). It would be good to add a comment to the docs to explain the dependency on the detect engine.

internetionals: Sorry for the delay, I've just returned from vacation. Thanks for the clarification, I'll also take a look at the necessary documentation changes.

internetionals:
Sorry that it took so long, but I was maintaining the main patch for the 5.0 branch and had to find the time to get a build/development environment for the master branch. Apparently the json generation had been changed to another library, so the fix turned out to be simple.

internetionals:
Is there anything more I can do to speed this along?

norg:
> Is there anything more I can do to speed this along?

Please don't force-push and create a new clean PR with the most recent commits you want to include.

victorjulien (approved):

victorjulien:
Merged in #5693, thanks!

PR #4900, opened 2020-04-30
github actions: extract repo/branch names from PR message - v3

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/4406

Create a "prep" build that parses libhtp, suricata-update and
suricata-verify repo and branch information from the pull
request message and turn these into artifacts that are
used by the builders

Examples:

libhtp-repo: https://github.com/OISF/libhtp
libhtp-branch: 0.5.x

suricata-verify-repo: http://github.com/OISF/suricata-verify
suricata-verify-branch: master

suricata-update-repo: http://github.com/OISF/suricata-update
suricata-update-branch: master

Other changes:
- Build cbindgen in debug mode, speeds up build.
- checkout@v2 (faster)
- working directory cleanup

PR #4942, opened 2020-05-15
Remove file-store v1 - v1

jasonish:
Remove file-store v1.

Attempt to document how to update the configuration from v1 to v2.

https://redmine.openinfosecfoundation.org/issues/2959

Suricata-Verify fails on test specific to file-store v1.

suricata-verify-repo: http://github.com/jasonish/suricata-verify
suricata-verify-branch: lt-version/v1


doc/userguide/upgrade.rst
```diff
@@ -30,6 +30,12 @@ by the ones Suricata supplies.
 Major updates include new features, new default settings and often also
 remove features.
 
+Upgrading 5.0 to 6.0
+--------------------
+
+Removals
+~~~~~~~~
+- File-store v1 has been removed. If using file extraction, the file-store configuration will need to be updated to version 2.
```
victorjulien, line 38: I think we should link from here to the "Update File-store v1 Configuration to V2" chapter.

victorjulien:
I don't understand the SV failure. Shouldn't the linked SV branch pass?

doc/userguide/file-extraction/config-update.rst
```diff
@@ -0,0 +1,38 @@
...
+      force-filestore: no # force storing of all files
+      stream-depth: 1mb   # reassemble 1mb into a stream, set to no to disable
+      waldo: file.waldo   # waldo file to store the file_id across runs
+      max-open-files: 0   # how many files to keep open (O means none)
+      write-meta: yes     # write a .meta file if set to yes
+      include-pid: yes    # include the pid in filenames if set to yes.
+
+the following changes will need to be made to convert to a V2 style configuration:
+
+* A ``version`` field must be set to 2.
```
norg, line 20: should it default to 2 if not set at all?

victorjulien: Right now the only way to use v2 is by setting the version. If we'd skip this I fear the only result will be that we take the v1 config as v2.

jasonish:
> I don't understand the SV failure. Shouldn't the linked SV branch pass?

That PR hasn't been merged yet: https://github.com/OISF/suricata/pull/4900

doc/userguide/file-extraction/config-update.rst
jasonish: I had been wondering about what to default to as well. Ideally things should just work with the minimal amount of configuration.  But with this being the first release without v1, I think we should enforce the version for now.

jasonish:
Continued at https://github.com/OISF/suricata/pull/4957

doc/userguide/file-extraction/config-update.rst
norg: Valid point, we could error out in later versions if old settings are still set.

jasonish: `log-dir` was renamed to `dir`.  So that is something we could check.  But in reality I'd like:
```
- file-store:
    enabled: yes
```
to just work with reasonable default.

Anyways, I'm OK with forcing the version to be set for now, then relaxing it later on..

victorjulien: We don't want to auto-convert since we don't want to reuse the `log-dir`/`dir`, right? So then I rather keep the requirement for now. We can remove it with 7 to shrink the default config. We should then still add some checks for v1 config, in case ppl go from 5->7.

So perhaps we do need some extra checks anyway, in case ppl just add a `version: 2` to their v1 config and expect it to work.

jasonish: So currently in this PR, just a warning is logged if version isn't 2.  So the filestore is disabled, but Suricata keeps running.  Should we be more harsh so its brought to the users attention sooner?

PR #4943, opened 2020-05-16
Http2 v9

catenacyber:

DRAFT for discussion
Not to be merged

Describe changes:
- Adds basic HTTP2 parser

Much more work to be done :
 + more commit squashing
 + parse remaining frame types
 + merge with HTTP1 for logging and detection ?
 + tests
 + robustness against evasions

Modifies #4940 with 
- detection on headers


rust/src/http2/detect.rs
```diff
@@ -0,0 +1,440 @@
...
+                return 0;
+            }
+        },
+        STREAM_TOCLIENT => match &tx.type_data {
+            Some(HTTP2FrameTypeData::HEADERS(hd)) => {
+                if (i as usize) < hd.blocks.len() {
+                    let mut vec = Vec::new();
+                    for j in 0..hd.blocks[i as usize].name.len() {
+                        vec.push(hd.blocks[i as usize].name[j]);
+                        /* this escaping is needed
```
catenacyber, line 410: Discussion welcomed here

src/detect-http2.c
```diff
@@ -0,0 +1,874 @@
...
+    pectx->mpm_ctx = mpm_ctx;
+    pectx->transforms = &mpm_reg->transforms;
+
+    return PrefilterAppendTxEngine(de_ctx, sgh, PrefilterTxHttp2Header,
+            mpm_reg->app_v2.alproto, mpm_reg->app_v2.tx_min_progress,
+            pectx, PrefilterMpmHttp2HeaderFree, mpm_reg->name);
+}
+
+//TODOask generic versions ?
+static int DetectEngineInspectHttp2Header(
```
catenacyber, line 831: Should we make this code generic instead of copying/pasting it ?

jasonish: My opinion is not yet. Its also some to explore on the Rust side as well. I think we have enough patterns now to find the generic stuff and abstract it out, but I'm thinking of doing that after this release cycle. I know this refers to the C side of things, but my feelings are similar.

rust/src/http2/detect.rs
jasonish: The escaping looks fine.  I found it a little faster, if you can determine a maximum size for the output vector to do something like `let mut out = vec![0; size]` then set the indexes directly. You have to make sure you don't overrun the vector though.  You can see where I do this in the JsonBuilder branch for the escaping done there.  In the JsonBuilder I think the performance difference was worth the extra code, I'm not sure about this case.

2 other things about this block of code.  I think `if let Some(HTTP2FrameTypeData::HEADERS(hd)) = &tx.type_data` would be a little cleaner than the match. And second, I'd like to see the escaping moved out of the extern "C" function to lessen the code in unsafe (whether or not an extern "C" is marked unsafe, I consider it unsafe).

rust/src/http2/parser.rs
```diff
@@ -0,0 +1,1718 @@
...
+            error: HTTP2HeaderDecodeStatus::HTTP2HeaderDecodeSizeUpdate,
+        },
+    ));
+}
+
+fn http2_parse_headers_block<'a>(
+    input: &'a [u8],
+    dyn_headers: &mut Vec<HTTP2FrameHeaderBlock>,
+) -> IResult<&'a [u8], HTTP2FrameHeaderBlock> {
+    //caller garantees o have at least one byte
```
jasonish, line 1216: We now have a debug_validate in master which might be useful here.  However, it does conflict with unit tests.

rust/src/http2/detect.rs
```diff
@@ -0,0 +1,440 @@
...
+                Err(_) => {
+                    return -1;
+                }
+            }
+        }
+        Err(_) => {
+            return -1;
+        }
+    }
+}
```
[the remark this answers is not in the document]
catenacyber: Nice :-)

rust/src/http2/logger.rs
```diff
@@ -0,0 +1,80 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use super::http2::{HTTP2FrameTypeData, HTTP2Transaction};
+use super::parser;
+use crate::json::*;
+use std;
+
+fn log_http2(tx: &HTTP2Transaction) -> Option<Json> {
```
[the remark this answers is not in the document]
catenacyber: indeed

rust/src/http2/parser.rs
```diff
@@ -0,0 +1,1718 @@
...
+    //TODO5 detection on (GOAWAY) additional data = length
+    pub length: u32,
+    pub ftype: HTTP2FrameType,
+    pub flags: u8,
+    pub reserved: u8,
+    stream_id: u32,
+}
+
+named!(pub http2_parse_frame_header<HTTP2FrameHeader>,
+    do_parse!(
```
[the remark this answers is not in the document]
catenacyber: I saw it.
Do you want me to translate all macros to functions ?

rust/src/http2/parser.rs
catenacyber: It is not yet in master, is it ?
`grep -r debug_validate rust/src/` yields no result

catenacyber: cf https://github.com/OISF/suricata/pull/4969

rust/src/http2/parser.rs
```diff
@@ -0,0 +1,1718 @@
...
+        )
+    }
+    let (i2, indexed) = parser(input)?;
+    match http2_frame_header_static(indexed.1, dyn_headers) {
+        Some(h) => Ok((i2, h)),
+        _ => Err(Err::Error((i2, ErrorKind::MapOpt))),
+    }
+}
+
+fn http2_huffman_table_len5(n: u32) -> Option<u8> {
```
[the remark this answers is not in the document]
catenacyber: Good idea

rust/src/http2/parser.rs
```diff
@@ -0,0 +1,1718 @@
...
+        1 => Some(HTTP2FrameHeaderBlock {
+            name: ":authority".as_bytes().to_vec(),
+            value: Vec::new(),
+            error: HTTP2HeaderDecodeStatus::HTTP2HeaderDecodeSuccess,
+        }),
+        2 => Some(HTTP2FrameHeaderBlock {
+            name: ":method".as_bytes().to_vec(),
+            value: "GET".as_bytes().to_vec(),
+            error: HTTP2HeaderDecodeStatus::HTTP2HeaderDecodeSuccess,
+        }),
```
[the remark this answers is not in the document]
catenacyber: > 89 insertions(+), 327 deletions(-)

Thanks :-)

rust/src/http2/detect.rs
catenacyber: Thanks

I am going rather with `Vec::with_capacity` to avoid reallocs and keep `push`

PR #4944, opened 2020-05-16
fuzz: do not reuse global variable named suricata

catenacyber:
No redonne but https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22259

Describe changes:
- Changes global variable name in fuzz targets so as not to reuse `suricata`

Fixes oss-fuzz builds (using a new feature of clang apparently)


victorjulien:
Merged through #4945 after rebasing on top of #4939.

Thanks Philippe.

PR #4951, opened 2020-05-19
Http gap 3359 v1

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3559

Describe changes:
- Handles gaps during HTTP parsing (when possible, mostly body responses with content-length)
- To do so, modifies util-file to handle gaps (and `PrintRawDataFp`) 
- Code cleaning in app-layer-htp.c so as to make it easier to reason on this

Needs libhtp patch https://github.com/OISF/libhtp/pull/295

What testing should we do ?
I think of a simple S-V test with one removed packet to show that parsing keeps going on ok
Any ideas ?

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -1155,9 +1155,9 @@ static int HTTPParseContentTypeHeader(uint8_t *name, size_t name_len,
  *  If the request contains a multipart message, this function will
  *  set the HTP_BOUNDARY_SET in the transaction.
  */
-static int HtpRequestBodySetupMultipart(htp_tx_data_t *d, HtpTxUserData *htud)
+static int HtpRequestBodySetupMultipart(htp_tx_t *tx, HtpTxUserData *htud)
 {
-    htp_header_t *h = (htp_header_t *)htp_table_get_c(d->tx->request_headers,
+    htp_header_t *h = (htp_header_t *)htp_table_get_c(tx->request_headers,
             "Content-Type");
     if (h != NULL && bstr_len(h->value) > 0) {
         uint8_t *boundary = NULL;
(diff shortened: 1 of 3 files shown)


victorjulien:
> What testing should we do ?
I think of a simple S-V test with one removed packet to show that parsing keeps going on ok
Any ideas ?

Yeah a set of tests with all the loss scenarios we can think of:
* gap within the body
* multiple different gaps in the same body
* gap that starts within body but extends beyond it
* gap that is exactly the body size
* gap after CL header but before end of headers?
* methods: GET/HEAD, POST/PUT, etc
* etc.
I know (or suspect) the current PR(+libhtp) doesn't cover all these cases yet, but ultimately we'd want to support all of them.

Perhaps some scripting can help to generate the pcaps.

catenacyber:
> gap after CL header but before end of headers?

How could we support this while we do not know the length or remaining headers ?
I do not see how this brings enough information to handle this differently as the general case.

By the way, the general case (when we the gap is in headers for instance) is still to fail (as is the case now I think) when we could probe for the next request/response. But I do not see how to do this without being vulnerable to HTTP smuggling... But then again, it may be still better than nothing...

> I know (or suspect) the current PR(+libhtp) doesn't cover all these cases yet, but ultimately we'd want to support all of them.

Every case should be supported but it is likely there is a bug :-p

catenacyber:
I think that this PR shows a bug in the S-V test http-sticky-server

There should be 29 alerts and not only 28 (as you can count in the pcap with `grep -a Omniture http-sticky-server.pcap | wc -l`

You can see more precisely the one missed with Wireshark filter `tcp.port == 56401 `

There should be indeed 2 alerts here and not only one
You can check this with
`jq 'select(.event_type=="alert" and .dest_port == 56401)' output/eve.json`

catenacyber:
> > What testing should we do ?
> > I think of a simple S-V test with one removed packet to show that parsing keeps going on ok
> > Any ideas ?
> 
> Yeah a set of tests with all the loss scenarios we can think of:
> 
> * gap within the body
> * multiple different gaps in the same body
> * gap that starts within body but extends beyond it
> * gap that is exactly the body size
> * gap after CL header but before end of headers?
> * methods: GET/HEAD, POST/PUT, etc
> * etc.
>   I know (or suspect) the current PR(+libhtp) doesn't cover all these cases yet, but ultimately we'd want to support all of them.
> 
> Perhaps some scripting can help to generate the pcaps.

PR #4956, opened 2020-05-19
stream/tcp: fix stream side after direction change

ilya-bakhtin:
TCP stream objects may be swapped during protocol detection while opposing stream packet is being processed.
So the stream level functions try to insert this packet segment into the wrong stream after the detection phase.
It results either in StreamTcpReassembleHandleSegmentHandleData error or in packet data insertion to a wrong direction stream.

This fix detects if the stream objects content was swapped and uses a proper direction stream in such a case.

Describe changes:
As mentioned in the commit description, the protocol detection procedure may change the flow direction and swap TCP source/destination stream contents. I propose to check if the streams were swapped and use the correct side if so.
The check condition used looks unnecessarily strong; however, i would prefer to avoid false positives.


ilya-bakhtin:
Well, i forgot to note that another option is to propagate explicit information about direction change to all functions on the stack. However, it looks much more invasive and does not make a big sense.

catenacyber:
Thanks for this PR.
Could you provide a suricata-verify test for this ?

ilya-bakhtin:
I've provided the reproducer pcap.
stream.midstream must be enabled in suricata.yaml to see the difference.
Originally packet #3 is handled by the improper stream side. My fix allows the detection of such a condition.
Please see examples of correct and wrong http.log
[http-start-from-response.zip](https://github.com/OISF/suricata/files/4973033/http-start-from-response.zip)

src/stream-tcp-reassemble.c
```diff
@@ -1767,14 +1767,22 @@ int StreamTcpReassembleHandleSegment(ThreadVars *tv, TcpReassemblyThreadCtx *ra_
...
     if ((dir == UPDATE_DIR_OPPOSING || dir == UPDATE_DIR_BOTH) &&
         StreamTcpReassembleHandleSegmentUpdateACK(tv, ra_ctx, ssn, opposing_stream, p) != 0)
     {
         SCLogDebug("StreamTcpReassembleHandleSegmentUpdateACK error");
         SCReturnInt(-1);
     }
-
+    /* The procedure above may swap content of ssn->server and ssn->client structures */
+    /* We have to continue with initial content of the stream in such case */
+    if (base_seq != stream->base_seq && isn != stream->isn && base_seq == opposing_stream->base_seq && isn == opposing_stream->isn)
```
catenacyber, line 1781: Why do we need the condition for `base_seq` ?
Is the `isn` test not enough ?

victorjulien: What if the `isn` is the same on both sides? Small chance I suppose, except if the server would do it deliberately.

catenacyber: Can we then copy the pointer to `stream` then and check it afterwards ?

ilya-bakhtin:
>Can we then copy the pointer to stream then and check it afterwards?
The pointers are not changed. The content is swapped.

>What if the isn is the same on both sides?
Well, it will work the same as previously i.e. the direction will not be updated.

After some analysis, I may propose a method based on flow flag. I hope its much more accurate. Please see the latest commit.

catenacyber:
Do you know where `FlowSwap` gets called ?

ilya-bakhtin:
Sure i do. The place that affects TCP streams processing is at app-layer.c:374 from TCPProtoDetect.

Other places are at app-layer.c:736 from AppLayerHandleUdp and stream-tcp.c:939
UDP stuff does not relate to my fix at all.
stream-tcp swaps direction in a case of (SYN+ACK) received as a very first packet. Usually, such a packet does not contain any payload. However, if it will do then the fix will also handle this condition properly.

src/stream-tcp-reassemble.c
```diff
@@ -1838,14 +1838,22 @@ int StreamTcpReassembleHandleSegment(ThreadVars *tv, TcpReassemblyThreadCtx *ra_
         dir = UPDATE_DIR_BOTH;
     }
 
+    const bool reversed_before_ack_handling = p->flow != NULL && (p->flow->flags & FLOW_DIR_REVERSED) != 0;
```
catenacyber, line 1841: Thanks again for this good bug report and investigation.

As this part of the code is treacherous, I wonder about the best fix.
Do we need the condition `p->flow != NULL` ?
Could we rather test `p->flowflags` change only ? as `PacketSwap` gets called also in `TCPProtoDetect`
Or have `StreamTcpReassembleAppLayer` somehow return if `AppLayerHandleTCPData` reversed the stream ?

ilya-bakhtin:
Well, it seems p->flow must be always available in a case of TCP.
Anyhow i've prepared a change with new packet flowflag that explicitly indicates the flow direction change.

catenacyber:
If p->flow is never NULL, I think it is ok to use it...

ilya-bakhtin:
Ok, please see the latest one with all unnecessary stuff removed

src/stream-tcp-reassemble.c
```diff
@@ -1838,6 +1838,7 @@ int StreamTcpReassembleHandleSegment(ThreadVars *tv, TcpReassemblyThreadCtx *ra_
         dir = UPDATE_DIR_BOTH;
     }
 
+    const bool reversed_before_ack_handling = (p->flow->flags & FLOW_DIR_REVERSED) != 0;
     /* handle ack received */
     if ((dir == UPDATE_DIR_OPPOSING || dir == UPDATE_DIR_BOTH) &&
         StreamTcpReassembleHandleSegmentUpdateACK(tv, ra_ctx, ssn, opposing_stream, p) != 0)
```
catenacyber, line 1844: All this code can be put in the if block `if ((dir == UPDATE_DIR_OPPOSING || dir == UPDATE_DIR_BOTH)`, can it not ?

ilya-bakhtin:
Do you mean something like
if (dir == UPDATE_DIR_OPPOSING || dir == UPDATE_DIR_BOTH)
{
    store flag
    if (StreamTcpReassembleHandleSegmentUpdateACK(tv, ra_ctx, ssn, opposing_stream, p) != 0)
    {...}
    if flag changed then fix stream
}
I agree it may look more clear.
Although nobody can change a flag if the function is not called, it looks reasonable to place all this game with flags closer to it as much as possible.

catenacyber:
Yes, that is what I mean, this way, we do not consume resources if we do not fulfill the condition

ilya-bakhtin:
I've made some variables local to 'if (dir == UPDATE_DIR_OPPOSING || dir == UPDATE_DIR_BOTH)' block

catenacyber:
This commit looks good to me.
Could you create the Suricata-verify test with the supplied pcap ? cf https://github.com/OISF/suricata-verify
and reference it in this Pull Request ?
Let me know if you have questions about it.

ilya-bakhtin:
I've added the test to my fork of suricata-verify:
https://github.com/ilya-bakhtin/suricata-verify/tree/tcp_stream_after_swap

catenacyber:
> I've added the test to my fork of suricata-verify:

Thanks, can you make a Pull Request to suricata-verify with it ?

ilya-bakhtin:
suricata-verify pull request: https://github.com/OISF/suricata-verify/pull/305

catenacyber:
Comment edited to have https://github.com/OISF/suricata-verify/pull/311

catenacyber (approved):
Looks good to me

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4202 so as not to lose track of this for 6.0.2

PR #4960, opened 2020-05-20
MQTT parser - v3

satta:
Previous PR: #4934 

- Add support for transactional logging. This involves a major change in how messages are handled and output, as well as moving more detection code into Rust.
- Add protocol parser events for unexpected transaction situations (seen responses without requests, invalid tx identifiers, CONNECT not seen, ...).
- Add `eve-log.mqtt.passwords` option to disable the logging of passwords.
- Remove unsafe use of `unwrap()`.
- Rename `UNKNOWN` to `UNASSIGNED` msg type to clarify its meaning.
- Remove `NULL` tests in `SCFree()`.
- Remove unnecessary typecasts in some cases.
- Add comments for more complex option parsing.
- Add tests for repeated, contradictory flag rule parsing.
- Remove PCRE usage where not necessary.
- Do not enable MQTT parser by default.
- Update documentation to reflect new logging format.

Suricata-verify targets have also been updated in the `mqtt` branch there to match the new format introduced here.


rust/src/mqtt/detect.rs
```diff
@@ -0,0 +1,481 @@
...
+    buffer_len: *mut u32,
+) -> u8 {
+    for msg in tx.msg.iter() {
+        if let MQTTOperation::CONNECT(ref cv) = msg.op {
+            let p = &cv.client_id;
+            if p.len() > 0 {
+                *buffer = p.as_ptr();
+                *buffer_len = p.len() as u32;
+                return 1;
+            }
```
catenacyber, line 223: Are we sure we will have only one `CONNECT` in one transaction ? (and thus not multiple buffers to match)

rust/src/mqtt/detect.rs
```diff
@@ -0,0 +1,481 @@
...
+    topic: *const std::os::raw::c_char,
+) -> u8 {
+    *result = false;
+    for msg in tx.msg.iter() {
+        match msg.op {
+            MQTTOperation::SUBSCRIBE(ref subv) => {
+                if let Ok(c_topic) = CStr::from_ptr(topic).to_str() {
+                    for topic in subv.topics.iter() {
+                        if topic.topic_name == c_topic {
+                            *result = true;
```
catenacyber, line 402: Why not a sticky buffer here as well ?
cf https://github.com/OISF/suricata/pull/4943/files#diff-7ba3ac4dbe7f1d3aaba36f3d70a6ad6eR831

rust/src/mqtt/detect.rs
```diff
@@ -0,0 +1,481 @@
...
+                return 0;
+            }
+            _ => return 0
+        }
+    }
+    return 0;
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_mqtt_tx_get_reason_code(
```
catenacyber, line 426: Do we really need `unsafe` for this function ?

rust/src/mqtt/logger.rs
```diff
@@ -0,0 +1,250 @@
...
+    return Some(js);
+}
+
+#[no_mangle]
+pub extern "C" fn rs_mqtt_logger_log(_state: &mut MQTTState, tx: *mut std::os::raw::c_void, flags: u32) -> *mut JsonT {
+    let tx = cast_pointer!(tx, MQTTTransaction);
+    match log_mqtt(tx, flags) {
+        Some(js) => js.unwrap(),
+        None => std::ptr::null_mut(),
+    }
```
catenacyber, line 249: None is never returned, is it ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+    MissingUnsubscribe,
+    DoubleConnect,
+    UnintroducedMessage,
+    InvalidQosLevel,
+    MissingMsgId,
+    UnassignedMsgtype,
+}
+
+impl MQTTEvent {
+    fn from_i32(value: i32) -> Option<MQTTEvent> {
```
catenacyber, line 50: Trait `FromPrimitive` also here

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+        }
+        return tx;
+    }
+
+    fn handle_msg(&mut self, msg: MQTTMessage, toclient: bool) {
+        match msg.op {
+            MQTTOperation::CONNECT(ref conn) => {
+                self.protocol_version = conn.protocol_version;
+                if let Some(tx) = self.get_tx_by_pkt_id(MQTT_CONNECT_PKT_ID) {
+                    &mut MQTTState::set_event(tx, MQTTEvent::DoubleConnect);
```
catenacyber, line 190: `self.get_tx_by_pkt_id` seems costly in time
Could we have a flag in the state ?
Or with the progress ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+                    tx.pkt_id = Some(MQTT_CONNECT_PKT_ID);
+                    self.transactions.push(tx);
+                }
+            },
+            MQTTOperation::PUBLISH(ref publish) => {
+                if !self.connected {
+                    let mut tx = self.new_tx(msg, toclient);
+                    &mut MQTTState::set_event(&mut tx, MQTTEvent::UnintroducedMessage);
+                    self.transactions.push(tx);
+                    return;
```
catenacyber, line 202: Why return ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+                            self.transactions.push(tx);
+                        } else {
+                            let mut tx = self.new_tx(msg, toclient);
+                            &mut MQTTState::set_event(&mut tx, MQTTEvent::MissingMsgId);
+                            self.transactions.push(tx);
+                        }
+
+                    },
+                    _ => {
+                        let mut tx = self.new_tx(msg, toclient);
```
catenacyber, line 226: This line is done for every case

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+                }
+            },
+            MQTTOperation::SUBSCRIBE(ref subscribe) => {
+                if !self.connected {
+                    let mut tx = self.new_tx(msg, toclient);
+                    &mut MQTTState::set_event(&mut tx, MQTTEvent::UnintroducedMessage);
+                    self.transactions.push(tx);
+                    return;
+                }
+                let pkt_id = subscribe.message_id as u32;
```
catenacyber, line 239: Scope can be restricted

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+                    },
+                    _ => {
+                        let mut tx = self.new_tx(msg, toclient);
+                        &mut MQTTState::set_event(&mut tx, MQTTEvent::InvalidQosLevel);
+                        self.transactions.push(tx);
+                    }
+                }
+            },
+            MQTTOperation::CONNACK(ref _connack) => {
+                if let Some(tx) = self.get_tx_by_pkt_id(MQTT_CONNECT_PKT_ID) {
```
catenacyber, line 289: Is there no risk of collision against `MQTT_CONNECT_PKT_ID` ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+                }
+            },
+            MQTTOperation::UNASSIGNED => {
+                let mut tx = self.new_tx(msg, toclient);
+                tx.complete = true;
+                &mut MQTTState::set_event(&mut tx, MQTTEvent::UnassignedMsgtype);
+                self.transactions.push(tx);
+            },
+            _ => {
+                if !self.connected {
```
catenacyber, line 375: Can we reach here ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+            }
+        }
+    }
+
+    fn parse_request(&mut self, input: &[u8]) -> bool {
+        if input.len() == 0 {
+            return true;
+        }
+
+        self.request_buffer.extend(input);
```
catenacyber, line 393: Why do we need to bufferize ?
Can we not use the `AppLayerResult::Incomplete` ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+        let len = self.transactions.len();
+
+        while index < len {
+            let tx = &self.transactions[index];
+            if tx.tx_id < min_tx_id + 1 {
+                index += 1;
+                continue;
+            }
+            *state = index as u64;
+            return Some((tx, tx.tx_id - 1, (len - index) > 1));
```
catenacyber, line 480: Why minus one ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,777 @@
...
+pub extern "C" fn rs_mqtt_probing_parser(
+    _flow: *const Flow,
+    _direction: u8,
+    input: *const u8,
+    input_len: u32,
+    _rdir: *mut u8,
+) -> AppProto {
+    let buf = build_slice!(input, input_len as usize);
+    if parse_message(buf, MQTT_DEFAULT_PROTOCOL_VERSION).is_ok() {
+        return unsafe { ALPROTO_MQTT };
```
catenacyber, line 502: Should we not return `ALPROTO_UNKNOWN` if we get `Incomplete` from nom ?

rust/src/mqtt/mqtt_message.rs
```diff
@@ -0,0 +1,181 @@
...
+            crate::mqtt::mqtt_message::MQTTOperation::PINGRESP => "PINGRESP",
+            crate::mqtt::mqtt_message::MQTTOperation::DISCONNECT(_) => "DISCONNECT",
+            crate::mqtt::mqtt_message::MQTTOperation::AUTH(_) => "AUTH",
+            _ => "UNASSIGNED",
+        }
+        .to_string()
+    }
+
+    pub fn message_type_id(&self) -> u32 {
+        match self.op {
```
catenacyber, line 162: Why do we need this function ?
Do not we have this number in the header ?

rust/src/mqtt/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+}
+
+// DATA TYPES
+
+named!(#[inline], pub parse_mqtt_string<String>,
+       do_parse!(
+           length: be_u16
+           >> content: take!(length)
+           >>  (
+                 str::from_utf8(&content).unwrap_or("NON-UTF8").to_string()
```
catenacyber, line 73: What does `unwrap_or("NON-UTF8")` ?

rust/src/mqtt/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+    )
+}
+
+named!(#[inline], pub parse_connect<MQTTConnectData>,
+       do_parse!(
+           protocol_string: parse_mqtt_string
+           >> protocol_version: be_u8
+           >> flags: parse_connect_variable_flags
+           >> keepalive: be_u16
+           >> properties: call!(parse_properties, protocol_version == 5)
```
catenacyber, line 198: Why not use `cond!` ?

rust/src/mqtt/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+                            header: header,
+                            op: MQTTOperation::PUBLISH(publish),
+                        };
+                        Ok((&input[skiplen+len..], msg))
+                    }
+                    Err(e) => Err(e),
+                },
+                4..=7 => {
+                    match parse_msgidonly(rem, len, protocol_version) {
+                        Ok((_rem, msgidonly)) => {
```
catenacyber, line 522: `rem` and `len` are redundant, can we just use `rem` ?

rust/src/mqtt/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+) -> IResult<&[u8], MQTTMessageIdOnly> {
+    if protocol_version < 5 {
+        // before v5 we don't even have to care about reason codes
+        // and properties, lucky us
+        return parse_msgidonly_v3(input);
+    }
+    match be_u16(input) {
+        Ok((rem, message_id)) => {
+            remaining_len -= 2;
+            if remaining_len == 0 {
```
catenacyber, line 271: This feels dangerous and of integer underflow

Can we not have this with nom ?
like `alt! (complete! (be_u8) | value!(0))` for reason_code

rust/src/mqtt/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+            let len = header.remaining_length as usize;
+            // This is the length of the fixed header that we need to skip
+            // before returning the remainder. It is the sum of the length
+            // of the flag byte (1) and the length of the message length
+            // varint.
+            let skiplen = 1 + varint_length(len);
+            let message_type = header.message_type;
+            if fullrem.len() < len {
+                return Err(Err::Incomplete(Needed::Size(len - fullrem.len())));
+            }
```
catenacyber, line 483: Is there a danger of DOS/memory exhaustion with specifying a very big `header.remaining_length` ?

rust/src/mqtt/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+    // be between 2 and 5 bytes long.
+    match parse_fixed_header(input) {
+        Ok((fullrem, header)) => {
+            let len = header.remaining_length as usize;
+            // This is the length of the fixed header that we need to skip
+            // before returning the remainder. It is the sum of the length
+            // of the flag byte (1) and the length of the message length
+            // varint.
+            let skiplen = 1 + varint_length(len);
+            let message_type = header.message_type;
```
catenacyber, line 480: Can we add an assert that `skiplen == input.len() - fuller.len()` ?

rust/src/mqtt/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+                        header: header,
+                        op: match message_type {
+                            12 => MQTTOperation::PINGREQ,
+                            13 => MQTTOperation::PINGRESP,
+                            _ => MQTTOperation::UNASSIGNED,
+                        },
+                    };
+                    if len > rem.len() {
+                        return Err(Err::Incomplete(Needed::Size(len - rem.len())));
+                    }
```
catenacyber, line 589: Do we need this check for Incomplete ?

rust/src/mqtt/detect.rs
```diff
@@ -0,0 +1,481 @@
...
+#[derive(Debug)]
+#[allow(non_camel_case_types)]
+enum MQTTFlagState {
+    DONT_CARE = 0,
+    MUST_BE_SET = 1,
+    CANT_BE_SET = 2,
+    INVALID,
+}
+
+impl From<u8> for MQTTFlagState {
```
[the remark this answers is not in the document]
satta: I guess I can look into it.

rust/src/mqtt/detect.rs
```diff
@@ -0,0 +1,481 @@
...
+//use crate::log::*;
+use crate::mqtt::mqtt::{MQTTTransaction, MQTTState};
+use crate::mqtt::mqtt_message::MQTTOperation;
+use std::ffi::CStr;
+use std::ptr;
+
+#[derive(Debug)]
+#[allow(non_camel_case_types)]
+enum MQTTFlagState {
+    DONT_CARE = 0,
```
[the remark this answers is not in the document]
satta: The C side has its own copy of it (https://github.com/OISF/suricata/pull/4960/files/b7dad68512a57af5eaa62abfc6068c86cd31a5dc#diff-885d36f7625b49de91f02db7c4960bbcR54). I can try making the Rust version the only one, and have the C version just use this enum.

rust/src/mqtt/detect.rs
satta: The spec only specifies one CONNECT (http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/csprd02/mqtt-v3.1.1-csprd02.html#_Toc385349764) in a compliant communication: 

> A Client can only send the CONNECT Packet once over a Network Connection. The    
> Server MUST process a second CONNECT Packet sent from a Client as a protocol
> violation and disconnect the Client

rust/src/mqtt/detect.rs
satta: Because there might be multiple topics in the subscription. You are right, a possible different approach would be to create a sticky buffer with all topics, e.g. separated by newlines, as it is done with HTTP headers. I didn't go with that because it didn't really feel intuitive to me, but I can see the advantage (e.g. being able to match on topics in a particular order, etc.)

Could you please probably specify a file/line for the corresponding place in the HTTP2 code? Thanks!

rust/src/mqtt/mqtt.rs
satta: Ok

rust/src/mqtt/mqtt.rs
satta: True, this seems to be redundant now.

rust/src/mqtt/detect.rs
satta: Well, it dereferences a raw pointer from C (`result`) to write the extracted value. This wouldn't work without `unsafe`, right?

```
error[E0133]: dereference of raw pointer is unsafe and requires unsafe function or block
   --> src/mqtt/detect.rs:437:21
    |
437 |                     *result = rcode;
    |                     ^^^^^^^^^^^^^^^ dereference of raw pointer
    |
    = note: raw pointers may be NULL, dangling or unaligned; they can violate aliasing rules and cause data races: all of these are undefined behavior
```

rust/src/mqtt/detect.rs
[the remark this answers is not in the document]
catenacyber: It should make the code simpler :-)

rust/src/mqtt/detect.rs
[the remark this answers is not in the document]
catenacyber: For example, this enum is shared between Rust and C : https://github.com/OISF/suricata/blob/master/rust/src/ssh/ssh.rs#L48

rust/src/mqtt/detect.rs
catenacyber: Ok, thanks

rust/src/mqtt/detect.rs
catenacyber: Indeed, but you may have `unsafe` only for the dereference, and not the whole function.

rust/src/mqtt/detect.rs
catenacyber: Line 831 of detect-http2.c (GitHub seems to have hidden it even if the link ends with R831)

No need to create a big buffer (even if it is one way to go)
You can use `InspectionBufferMultipleForListGet` and such (other keywords do, in dns for instance)

satta: Ahh. I didn't know that. Thanks for the pointer.

rust/src/mqtt/detect.rs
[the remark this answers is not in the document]
catenacyber: Well, that is the magic of if.
`HTTP2FrameType` implements `from_u8` just by deriving `FromPrimitive`(not sure it is the right rust wording)

It is used on line 84 of parser.rs
https://github.com/OISF/suricata/pull/4943/files#diff-1731300ec3665501b1d0927c8ae1dc40R84

rust/src/mqtt/detect.rs
catenacyber: You are welcome to the knowledge :-)

rust/src/mqtt/detect.rs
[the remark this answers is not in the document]
satta: > Well, that is the magic of if.
> `HTTP2FrameType` implements `from_u8` just by deriving `FromPrimitive`(not sure it is the right rust wording)

Ahhh got it. Adding `use num_traits::FromPrimitive;` and a `#[derive(FromPrimitive)]` was enough for me to drop my own implementation. Nice!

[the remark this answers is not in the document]
catenacyber: Glad to share my rust ~~infinite knowledge~~ learning curve

rust/src/mqtt/detect.rs
satta: True. Will move the `unsafe` into the inner parts of the  code. Some of the functions there don't  need `unsafe` at all.

rust/src/mqtt/logger.rs
satta: This is actually obsolete after switch to JsonBuilder.

rust/src/mqtt/mqtt.rs
satta: I was trying to only get _one_ mutable reference late in the code flow, otherwise I would need to pass around and store references with lifetimes. Trying to keep things simple...
Moving this line before the `match msg...` would require this... right?

rust/src/mqtt/mqtt.rs
satta: Done.

rust/src/mqtt/mqtt.rs
satta: Well,  if no `CONNECT` was seen before any other operation, we can simply set the corresponding event and skip this message. However, now that I think about it, you are right -- if we arrived here, then something must have been parsed which is syntactically correct, so might as well store it... _and_ set the event as well. It might still be interesting to see.

satta: Removing the `return`.

rust/src/mqtt/mqtt.rs
satta: Yes, with the `PING*` message types. I will add them explicitly so this is visible in the code itself.

rust/src/mqtt/mqtt.rs
catenacyber: Can you add this as a comment ? thanks for the justification

rust/src/mqtt/mqtt.rs
catenacyber: thanks

rust/src/mqtt/mqtt.rs
satta: We actually already have a `connected` field in the state. Makes sense to use it here.

rust/src/mqtt/mqtt.rs
satta: Done. Maybe there is a more concise way to do it, but this works for now. I'd rather have a clear view of what is happening 
 each case instead of just reducing the line count.

rust/src/mqtt/mqtt.rs
satta: There is no risk. Message IDs (also called 'packet identifiers' in some spec versions) in MQTT messages are only 16 bit, so no ID parsed from traffic can collide with that special value `MQTT_CONNECT_PKT_ID` (which is defined to be `std::u32::MAX`).

rust/src/mqtt/mqtt.rs
satta: This is 1:1 from the applayer template code, I didn;t change it. I guess this has to do with the fact that transaction ID 0 is never given out.

rust/src/mqtt/mqtt.rs
satta: Yep. Made this more explicit to only return `ALPROTO_MQTT` on successful parsing (returned `Ok` result).

rust/src/mqtt/mqtt_message.rs
satta: Correct. Changed the code in `detect.rs` and removed this function.

rust/src/mqtt/parser.rs
satta: It tries to get the string from the `from_utf8` conversion function, and in case this fails, just returns the fixed string `NON-UTF8`. Since in MQTT valid strings are _always_ UTF8, this case would trigger only when parsing a non-compliant message. This code is just to avoid runtime issues with `unwrap`.

However, I can see that it might be better to just use `String::from_utf8_lossy` in this case, which would preserve the original string but just replace non-UTF8 characters with a replacement. See https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy for more info. I will use that in this case.

rust/src/mqtt/mqtt.rs
catenacyber: Thanks for the explanation. 
Can you add this as a comment ?

rust/src/mqtt/mqtt.rs
catenacyber: @jasonish thoughts here ?

rust/src/mqtt/parser.rs
catenacyber: Thanks for the explanation for myself :-)

rust/src/mqtt/mqtt.rs
satta: Done.

rust/src/mqtt/detect.rs
satta: Done. This is really cool!

rust/src/mqtt/parser.rs
satta: I think it's easier to not calculate any specific size any more and just request 1 more byte, as discussed earlier in another thread. It's just not worth risking to miss something when in the majority of cases the missing data is there with the next packet  (and hence parser call) anyway.

rust/src/mqtt/parser.rs
satta: I would need to dive deeper and try to get my head around this again, but I think it was because of the variety here and the fact that `cond!` would already consume something that would need to be part of another parsed field in case the protocol version is not 5 (as there is no space for properties in that case). I definitely remember I tried various combinations involving `cond!` and it was more complicated that doing it like this. If you have a better solution, please feel free to propose it.

rust/src/mqtt/mqtt.rs
jasonish: Yeah, @satta is correct. The ID in the parser is always off by one from the ID outside the parser for reasons that predate me.

rust/src/mqtt/parser.rs
satta: Also, the remaining length is at most 256MB due to protocol range constraints (https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/csprd02/mqtt-v3.1.1-csprd02.html#_Toc385349759). There's not much one can do about that.

rust/src/mqtt/parser.rs
satta: I guess. That would check whether `varint_length()` did the right thing...?

rust/src/mqtt/parser.rs
satta: I don't think so. I think it was left from earlier code.

rust/src/mqtt/parser.rs
catenacyber: I guess that I would rather compute `skiplen = input.len() - fullrem.len()` than using `varint_length`
So, not sure what the best is

rust/src/mqtt/mqtt.rs
catenacyber: Thanks Jason. Let me know if you understand the reason someday

rust/src/mqtt/parser.rs
catenacyber: Keep the working version, thanks

rust/src/mqtt/parser.rs
catenacyber: > I think it's easier to not calculate any specific size any more and just request 1 more byte, as discussed earlier in another thread

Indeed, but you seem to misunderstand my point.

256 MB looks quite big with respect to DOS by memory exhaustion. The TCP streaming engine will bufferize these 256 Mb. So, if I start XX connections and get into this state for these connections and send data slower as it gets more bufferized, Suricata will run out of memory...

> There's not much one can do about that.

Well, we can either 
- at least set a limit, and set an event when the limit is crossed
- have a more "streaming parser" keeping a more complex state, so as not to bufferize all the data (yes this is hard :-/ maybe this 256 Mb is not reachable by every message type to start)

Is this clearer ?

satta: Indeed it is.

I get your point about the potential for DOS. Unfortunately MQTT is optimized for space efficiency and does not provide a lot of information upfront about where the space is used, be it for topics, message payloads, etc.  -- you have to parse it to find out! 
So I would propose to introduce a maximum message size to inspect that can be configured in the `mqtt` YAML config. Messages with an indicated length exceeding that would be logged as events, but not parsed until they are complete. What do you think? I think such an approach would at least make the problem manageable. Another idea would be to introduce a memcap value for the parser, and to stop parsing new messages once that memcap has been hit.
Any comments?

victorjulien: In this worst case scenario of a 256mb record, what would we do with that data? If nothing much, we can just stream it through. In the smb and nfs parsers I've implemented some logic to track how many bytes to 'skip'. So the data still flows though the parser, but it doesn't do much except the 'skip' bookkeeping. Maybe that could be used here as well?

catenacyber: Indeed, SMB has an hardcoded limit of 1Mbyte if I remember correctly.
Configurable limit is even better :-)

satta: > In this worst case scenario of a 256mb record, what would we do with that data? If nothing much, we can just stream it through.

Well, the payload is logged in the EVE output -- so if someone sends a big binary payload (like a HD camera image) that would end up in the payload field, escaped for JSON.

> In the smb and nfs parsers I've implemented some logic to track how many bytes to 'skip'. So the data still flows though the parser, but it doesn't do much except the 'skip' bookkeeping. Maybe that could be used here as well?

Well, skipping means not storing data, so there is nothing to output. One could decide to leave out the parsed data in a  large message (i.e. a message that exceeds the limit), set a flag (similar to the filestore TRUNCATED flag) and simply report the message, message type, etc. without the type-specific metadata. That would be a good compromise IMHO - you know that there has been a large message, there are communication and flow metadata, but there is no requirement to parse (and hence buffer) the whole message.

victorjulien: > > In this worst case scenario of a 256mb record, what would we do with that data? If nothing much, we can just stream it through.
> 
> Well, the payload is logged in the EVE output -- so if someone submits or receives a big binary payload (like a HD camera image, etc.) that would end up in the payload field, escaped for JSON.

@jasonish we should probably define an upper limit for both the total eve record size as well as individual fields sizes. Can't imagine the mayhem 256mb records would cause :)

satta: +1 :)

jasonish: > > > In this worst case scenario of a 256mb record, what would we do with that data? If nothing much, we can just stream it through.
> > 
> > 
> > Well, the payload is logged in the EVE output -- so if someone submits or receives a big binary payload (like a HD camera image, etc.) that would end up in the payload field, escaped for JSON.
> 
> @jasonish we should probably define an upper limit for both the total eve record size as well as individual fields sizes. Can't imagine the mayhem 256mb records would cause :)

2 ideas here. We could enforce an upper bound on the size the builder will build. But I think the parser should be aware, and do truncation if necessary, so you don't lose all context on the record.

rust/src/mqtt/parser.rs
satta: Well, we need to check here whether the message has omitted reason code and properties, and need to break early and not try parsing the properties then. I don't know how to do that with nom macros, it's a similar problem to the one you pointed out earlier. TBH I'd rather have explicit, documented parsing steps here just to make it clear what is happening, but I agree that there should be extra checks to make sure there is no integer underflow in these unchecked arithmetics. Will add them.

catenacyber: Does not my nom `alt! (complete! (be_u8) | value!(0))` do the trick for `reason_code` ?
Then for `properties`, you can wrap with `alt! (complete! ( STUFF ))` to get `None ` in the relevant case, or the properties if you have data for them.

About the general philosophy of explicit documented parsing steps versus nom magic, this is a good question I do not have the answer to.
It seemed to me that there was more nom magic than explicit parsing steps, so I tried this approach in my parsers even if it felt not natural to me.

rust/src/mqtt/parser.rs
satta: Currently looking at this -- we can simply decide, after parsing the 2-5 byte fixed header, whether we want to parse the rest of the record or not depending on a configurable maximum parse depth. However, AFAICS we would not be able to partially parse the first portion of the data up to the limit, at least I don't know how to do this in nom without ditching the macros and manually checking the remaining capacity after each parsed item. Parsing a complete record with `do_parse!` would require to read (i.e. buffer!) everything before the final parsed struct is returned with an `IResult` of `Ok`, and that's what we would like to avoid.

My idea was to simply check whether the specified length exceeds the limit and then return and log a simplified special message type that only contains the type and skipped length of the large record. After that I would simply accept (without applayer buffering) and eat upcoming data until we can continue parsing the next message in the connection.
That means we would get something like:
```json
{
  "timestamp": "2020-06-19T16:59:24.964522+0200",
  "flow_id": 1025328194301771,
  "pcap_cnt": 27,
  "event_type": "mqtt",
  "src_ip": "0000:0000:0000:0000:0000:0000:0000:0001",
  "src_port": 50509,
  "dest_ip": "0000:0000:0000:0000:0000:0000:0000:0001",
  "dest_port": 1883,
  "proto": "TCP",
  "mqtt": {
    "publish": {
      "qos": 1,
      "retain": false,
      "dup": false,
      "truncated": true,
      "skipped_length": 1000007
    }
  }
}
```
Such an entry has the message type, the message size and context (src/dest, timestamp, ...) but no additional metadata since we decided not to parse it. In particular, we don't have a message ID to connect it to a transaction (i.e. in the case above we could not connect it to the `puback` that could follow). This is because most of the 256MB could already be used before reaching the position in the record where the message ID occurs.

Would you be OK with these restrictions? I would agree that this is a rare case because I would expect most of the MQTT traffic out there to be smaller than any reasonable limit (i.e. 1MB).

catenacyber: Yes, that is what I had in mind :-)

rust/src/mqtt/parser.rs
satta: Done.

rust/src/mqtt/parser.rs
satta: Got it :)

satta:
New PR: #5098

PR #4961, opened 2020-05-20
travis: add test for enable-debug-validation

inashivb:
A test for `--enable-debug-validation` has been added to travis configuration file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -156,6 +156,14 @@ matrix:
         - *default-cflags
         - ARGS="--enable-debug"
         - ENABLE_DEBUG="yes"
+    # Linux, gcc, debug-validate.
+    - os: linux
+      compiler: gcc
+      env:
+        - NAME="linux,gcc,debug-validate"
+        - *default-cflags
+        - ARGS="--enable-debug-validation"
+        - NO_UNITTESTS="yes"
     # Linux, gcc, no jansson.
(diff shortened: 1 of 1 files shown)


victorjulien:
exit code -6 means `abort()`, so it looks like something in the dcerpc code triggered this?
```
===> dcerpc-dce-iface-01: FAILED: got exit code -6, expected 0
```

victorjulien:
@jasonish how do you feel about this being in travis? Can't remember if we're trying to unify on github-ci.

inashivb:
> exit code -6 means `abort()`, so it looks like something in the dcerpc code triggered this?
> 
> ```
> ===> dcerpc-dce-iface-01: FAILED: got exit code -6, expected 0
> ```

It seems like the match block is hit after bindack is parsed when the request object is obviously not there and `request.is_none()` check triggers the panic. Not sure if I should just remove that check. Although, given the pcap, it seems like the right thing to do.
Thoughts?

jasonish:
> @jasonish how do you feel about this being in travis? Can't remember if we're trying to unify on github-ci.

I think its fine to keep Travis alive, but I'd rather see GitHub CI as the primary, so this should probably be added to one of the GitHub CI builds.  Probably don't matter which one.

inashivb:
Replaced with #4969

PR #4953, opened 2020-05-19
detect/bsize: Validate against `content` buffer when available

jlucovsky:
Continuation of #4950

This PR adds additional validation when using the `bsize` keyword. If a `content` keyword immediately precedes `bsize`, then the `bsize` value is checked to see if a match is possible using the operation (`=, <, >, <>`) and the value.

An error is raised if `bsize` value prevents a match, e.g., the content length exceeds the `bsize` value.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3682](https://redmine.openinfosecfoundation.org/issues/3682)

Describe changes:
- Rewords commits to remove markdown.

Companion [Suricata PR #233](https://github.com/OISF/suricata-verify/pull/233)

PR #4963, opened 2020-05-20

jlucovsky:
Continuation of #4953

Describe changes:
- Addressed review comments. All preceding `content` values are inspected.
- Documentation update.

### DIFF
--- a/src/detect-bsize.c
+++ b/src/detect-bsize.c
@@ -65,6 +65,8 @@ void DetectBsizeRegister(void)
 #define DETECT_BSIZE_RA 2
 #define DETECT_BSIZE_EQ 3
 
+const char *bsize_mode_strings[] = { "<", ">", "<>", "="};
+
 typedef struct DetectBsizeData {
     uint8_t mode;
     uint64_t lo;
@@ -266,6 +268,43 @@ static DetectBsizeData *DetectBsizeParse (const char *str)
     return bsz;
 }
 
+static bool DetectBsizeCheckContent(Signature *s, int list, DetectBsizeData *bsz)
+{
+    SigMatch *sm = s->init_data->smlists[list];
+    for (; sm != NULL; sm = sm->next) {
+        if (sm->type != DETECT_CONTENT)
+            continue;
+
+        DetectContentData *cd = (DetectContentData *) sm->ctx;
+        SCLogNotice("Content %.*s, content length %"PRIu16, cd->content_len, cd->content, cd->content_len);
+        bool possible;
+
+        switch (bsz->mode) {
+            case DETECT_BSIZE_EQ:
+                possible = bsz->lo == cd->content_len;
+                break;
+            case DETECT_BSIZE_GT:
+                possible = bsz->lo < cd->content_len;
+                break;
+            case DETECT_BSIZE_LT:
+                possible = cd->content_len < bsz->lo;
+                break;
+            case DETECT_BSIZE_RA:
+                possible = bsz->lo < cd->content_len && cd->content_len < bsz->hi;
+                break;
+        }
+        if (!possible) {
+            SCLogError(SC_ERR_INVALID_SIGNATURE,
+                    "bsize match impossible: content len %d and bsize op '%s' "
+                    "values lo=%"PRIu64"; hi=%"PRIu64,
+                    cd->content_len,
+                    bsize_mode_strings[bsz->mode], bsz->lo,bsz->hi);
+            return false;
+        }
+    }
+
+    return true;
+}
 /**
  * \brief this function is used to parse bsize data into the current signature
  *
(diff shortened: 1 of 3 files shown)


src/detect-bsize.c
```diff
@@ -266,6 +268,43 @@ static DetectBsizeData *DetectBsizeParse (const char *str)
...
 
+static bool DetectBsizeCheckContent(Signature *s, int list, DetectBsizeData *bsz)
+{
+    SigMatch *sm = s->init_data->smlists[list];
+    for (; sm != NULL; sm = sm->next) {
+        if (sm->type != DETECT_CONTENT)
+            continue;
+
+        DetectContentData *cd = (DetectContentData *) sm->ctx;
+        SCLogNotice("Content %.*s, content length %"PRIu16, cd->content_len, cd->content, cd->content_len);
```
victorjulien, line 279: debug

src/detect-bsize.c
```diff
@@ -266,6 +268,43 @@ static DetectBsizeData *DetectBsizeParse (const char *str)
     return bsz;
 }
 
+static bool DetectBsizeCheckContent(Signature *s, int list, DetectBsizeData *bsz)
```
victorjulien, line 271: const args

src/detect-bsize.c
```diff
@@ -266,6 +268,43 @@ static DetectBsizeData *DetectBsizeParse (const char *str)
...
+{
+    SigMatch *sm = s->init_data->smlists[list];
```
victorjulien, line 273: const

src/detect-bsize.c
```diff
@@ -291,6 +330,13 @@ static int DetectBsizeSetup (DetectEngineCtx *de_ctx, Signature *s, const char *
     DetectBsizeData *bsz = DetectBsizeParse(sizestr);
     if (bsz == NULL)
         goto error;
+
+    if (s->init_data->smlists[list] != NULL) {
```
victorjulien, line 334: by putting this here, the validation happens while new conditions can still be added. What happens with:
`http.uri; content:"abc"; bsize:3; content:"abcdef";`?

src/detect-bsize.c
```diff
@@ -266,6 +268,43 @@ static DetectBsizeData *DetectBsizeParse (const char *str)
     return bsz;
 }
 
+static bool DetectBsizeCheckContent(Signature *s, int list, DetectBsizeData *bsz)
+{
+    SigMatch *sm = s->init_data->smlists[list];
+    for (; sm != NULL; sm = sm->next) {
+        if (sm->type != DETECT_CONTENT)
```
victorjulien, line 275: while at it, it might be useful to see if we can find other conflicts too? E.g. on the uri buffer we can have `urilen`, which does pretty much the same as `bsize`. Would be good they don't have conflicting, mutually exclusive settings.

src/detect-bsize.c
```diff
@@ -266,6 +268,43 @@ static DetectBsizeData *DetectBsizeParse (const char *str)
...
+            continue;
+
+        DetectContentData *cd = (DetectContentData *) sm->ctx;
+        SCLogNotice("Content %.*s, content length %"PRIu16, cd->content_len, cd->content, cd->content_len);
+        bool possible;
+
+        switch (bsz->mode) {
+            case DETECT_BSIZE_EQ:
+                possible = bsz->lo == cd->content_len;
```
victorjulien, line 284: just checking for content_len is too limiting. What happens if we have:
`http.uri; content:"abc"; offset:3; depth:3; bsize:3;`? The sig can't match because we need 6 bytes of data.

victorjulien:
The `DetectContentPropagateLimits` function is probably interesting to look at as well. It shows the effects of offset/depth/within/distance. These can form chains that we can use to calculate a min and max depth. This can be used to say if `bsize` is sane.

There is an optimization opportunity as well: A `bsize` setting in case of `content:"abc"; bsize:8;` can also be used to effectively use `content:"abc"; depth:8; bsize:8;`. This matters as the fast_pattern/mpm algorithms understand 'depth' and use it to limit the effective search of the pattern.

victorjulien:
See also `DetectUrilenApplyToContent`

jlucovsky:
> A bsize setting in case of content:"abc"; bsize:8;

The changes in this PR would trigger an error since the content length is 3 and 8 bytes are requested.

src/detect-bsize.c
jlucovsky: Should we address  `urilen` potential conflicts in another redmine issue?

victorjulien:
> > A bsize setting in case of content:"abc"; bsize:8;
> 
> The changes in this PR would trigger an error since the content length is 3 and 8 bytes are requested.

That would be wrong. `bsize` is a check for the buffer size, like a universal `urilen` or `dsize`. It is perfectly valid to have a rule that says `http.uri; content:"/index.php"; bsize:>1024;` to check for a very long uri that starts with `/index.php`. The goal of the ticket is to make sure the `bsize` value isn't too limiting that the patterns can't possibly fit the `bsize` 'budget'.

jlucovsky:
Continued in #4991

PR #4964, opened 2020-05-20
Byteextractstr/part2/v4

inashivb:

### DIFF
--- a/src/util-ip.c
+++ b/src/util-ip.c
@@ -26,6 +26,7 @@
 
 #include "suricata-common.h"
 #include "util-ip.h"
+#include "util-byte.h"
 
 /** \brief determine if a string is a valid ipv4 address
  *  \retval bool is addr valid?
@@ -65,9 +66,9 @@ bool IPv4AddressStringIsValid(const char *str)
 
     addr[dots][alen] = '\0';
     for (int x = 0; x < 4; x++) {
-        int a = atoi(addr[x]);
-        if (a < 0 || a >= 256) {
-            SCLogDebug("out of range");
+        int a;
+        if (StringParseI32RangeCheck(&a, 10, 0, (const char *)addr[x], 0, 255) < 0) {
+            SCLogDebug("invalid value for address byte: %s", addr[x]);
             return false;
         }
     }
--- a/src/util-napatech.c
+++ b/src/util-napatech.c
@@ -800,10 +800,19 @@ static uint32_t CountWorkerThreads(void)
 
                         char copystr[16];
                         strlcpy(copystr, lnode->val, 16);
-
-                        ByteExtractStringUint8(&start, 10,  0, copystr);
-                        ByteExtractStringUint8(&end, 10,  0, strchr(copystr, '-') + 1);
-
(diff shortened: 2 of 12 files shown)


src/util-ip.c
```diff
@@ -65,9 +66,9 @@ bool IPv4AddressStringIsValid(const char *str)
 
     addr[dots][alen] = '\0';
     for (int x = 0; x < 4; x++) {
-        int a = atoi(addr[x]);
-        if (a < 0 || a >= 256) {
-            SCLogDebug("out of range");
+        int a;
+        if (StringParseI32RangeCheck(&a, 10, 0, (const char *)addr[x], 0, 255) < 0) {
```
catenacyber, line 70: Should we make the variable `a` an `int32_t` ?
It feels risky somehow...

src/util-napatech.c
```diff
@@ -800,10 +800,19 @@ static uint32_t CountWorkerThreads(void)
...
-                        ByteExtractStringUint8(&end, 10,  0, strchr(copystr, '-') + 1);
-
+                        if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
+                            FatalError(SC_ERR_INVALID_VALUE, "Napatech invalid"
+                                       " worker range start: '%s'", copystr);
+                        }
+                        char *end_str = strchr(copystr, '-');
+                        if (end_str != NULL && StringParseUint8(&end, 10, 0, (const char *) (end_str + 1)) < 0) {
+                            FatalError(SC_ERR_INVALID_VALUE, "Napatech invalid"
+                                       " worker range end: '%s'", strchr(copystr, '-') + 1);
```
catenacyber, line 810: It looks like we should `FatalError` if `end_str == NULL`

src/util-proto-name.c
```diff
@@ -60,10 +61,9 @@ void SCProtoNameInit()
             if (proto_ch == NULL)
                 continue;
 
-            int proto = atoi(proto_ch);
-            if (proto >= 255)
+            uint8_t proto;
+            if (StringParseU8RangeCheck(&proto, 10, 0, (const char *)proto_ch, 0, 255) < 0)
                 continue;
```
catenacyber, line 66: should be 254 to be the same as before.
Or before the check was off by one ?

src/util-ip.c
inashivb: Just for my understanding, could you please explain why is this risky?

src/util-napatech.c
inashivb: That's correct. Adding it in the next revision.

src/util-proto-name.c
inashivb: Before, the check was off by one. Victor suggested to include 255 in the range.

inashivb:
Replaced with #4970

src/util-ip.c
catenacyber: Because `int` size may be 32 or 64 bits.
So, it may not be the same thing as the `int32_t` used by `StringParseI32RangeCheck` with its first pointer argument.
And it is better to avoid int size conversions when we can because there can be subtle bugs

PR #4954, opened 2020-05-19
detect/asn1: Migrate to Rust - v1

sharksforarms:
Draft PR: Request for review/comments

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2977

Describe changes:
- Migration of ASN1 decoder and detect module to Rust

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

TODO:
- 4.0.0 release of https://github.com/rusticata/der-parser/
- https://github.com/rusticata/kerberos-parser/pull/3
- https://github.com/rusticata/snmp-parser/pull/7
- Once the above PRs are merged, update `rust/Cargo.toml.in`

PR #4966, opened 2020-05-20
detect/asn1: Fix relative_offset keyword option - v1

sharksforarms:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3720

Describe changes:
- Fix relative_offset keyword option to be relative in regards to the
last content match
- Change relative_offset to int32_t with bounds check to allow the full
range of the packet buffer size (uint16_t)
- Added checks for over/underflows
- Changed the offset type to uint16_t because the offset is applied to
the payload length, which is a uint16_t
- Adjusted test cases to work relative to the content match
- Added test case to verify bounds

Note: I fixed this in https://github.com/OISF/suricata/pull/4954 as well

### DIFF
--- a/src/detect-asn1.c
+++ b/src/detect-asn1.c
@@ -148,15 +148,37 @@ static int DetectAsn1Match(DetectEngineThreadCtx *det_ctx, Packet *p,
     }
 
     const DetectAsn1Data *ad = (const DetectAsn1Data *)ctx;
-    int32_t offset;
+    uint16_t offset;
     if (ad->flags & ASN1_ABSOLUTE_OFFSET) {
         offset = ad->absolute_offset;
     } else if (ad->flags & ASN1_RELATIVE_OFFSET) {
-        offset = ad->relative_offset;
+        // relative offset in regards to the last content match
+
+        // Range check the relative_offset before downcasting to uint16_t
+        if (ad->relative_offset > UINT16_MAX
+            || ad->relative_offset < -UINT16_MAX) {
+                return 0;
+        }
+
+        // Check for over/underflows
+        if (ad->relative_offset > 0) {
+            if (det_ctx->buffer_offset > (UINT16_MAX - (uint16_t)ad->relative_offset)) {
+                // overflow: offset exceeds max payload length
+                return 0;
+            }
+        } else {
+            if (det_ctx->buffer_offset < (uint16_t)(-ad->relative_offset)) {
+                // underflow: offset is invalid in regards to the buffer
+                return 0;
+            }
+        }
+
+        offset = det_ctx->buffer_offset + ad->relative_offset;
     } else {
         offset = 0;
     }
-    if (offset >= (int32_t)p->payload_len) {
+
+    if (offset >= p->payload_len) {
         return 0;
     }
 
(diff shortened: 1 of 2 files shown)


src/detect-asn1.h
```diff
@@ -36,12 +36,11 @@
 typedef struct DetectAsn1Data_ {
     uint8_t flags;     /* flags indicating the checks loaded */
     uint32_t oversize_length;   /* Length argument if needed */
-    int32_t absolute_offset;   /* Length argument if needed */
+    uint16_t absolute_offset;   /* Length argument if needed */
```
jlucovsky, line 39: What the rationale for changing the type?
Why leave the type of `relative_offset` unchanged?

src/detect-asn1.c
```diff
@@ -148,15 +148,37 @@ static int DetectAsn1Match(DetectEngineThreadCtx *det_ctx, Packet *p,
...
-    int32_t offset;
+    uint16_t offset;
     if (ad->flags & ASN1_ABSOLUTE_OFFSET) {
         offset = ad->absolute_offset;
     } else if (ad->flags & ASN1_RELATIVE_OFFSET) {
-        offset = ad->relative_offset;
+        // relative offset in regards to the last content match
+
+        // Range check the relative_offset before downcasting to uint16_t
+        if (ad->relative_offset > UINT16_MAX
```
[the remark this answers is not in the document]
sharksforarms: Range checking again as a defensive measure before downcasting, I can remove this check however as relative_offset is never touched otherwise

src/detect-asn1.h
sharksforarms: absolute_offset: Represents an offset from [0, payload_len], `Packet::payload_len` is represented as a `uint16_t`, so [0, uint16::max]
relative_offset: Represents an offset relative to the last content match which is between [0, uint16::max] so relative to this, we need a type to hold the range [-uint16::max, uint16::max] which is why it's int32, with a bounds check on parsing and also contains overflow checks

That was my train of thought, open to suggestions :)

src/detect-asn1.c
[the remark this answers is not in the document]
victorjulien: If we're positive it's been checked at parsing we would use `BUG_ON()` instead.

src/detect-asn1.c
```diff
@@ -148,15 +148,37 @@ static int DetectAsn1Match(DetectEngineThreadCtx *det_ctx, Packet *p,
...
+            || ad->relative_offset < -UINT16_MAX) {
+                return 0;
+        }
+
+        // Check for over/underflows
```
victorjulien, line 163: I find these checks hard to follow. I think I would feel better seeing the checks done as u32 (or i32) and then only down casting after the values have been validated. Easier to follow for me anyway.

sharksforarms: Agreed, re-reading this I get confused. I'll see if I can come up with a better solution

src/detect-asn1.h
victorjulien: Makes sense to me.

sharksforarms:
https://github.com/OISF/suricata/pull/4978

PR #4968, opened 2020-05-21
output-json: add MAC addresses to EVE-JSON logs - v9

satta:
Previous PR: #4767

- Only do one call to `SC_ATOMIC_SUB`.
- Use flow storage to store MacSets.
- Add counters for maximum number of MAC addresses ( `decoder.max_mac_addrs_src` and `decoder.max_mac_addrs_dst`)

Suricata-verify tests (https://github.com/OISF/suricata-verify/pull/197) have also been adapted.

### DIFF
--- a/src/flow.c
+++ b/src/flow.c
@@ -410,7 +410,7 @@ static inline void FlowUpdateTTL(Flow *f, Packet *p, uint8_t ttl)
  *
  *  \note overwrites p::flowflags
  */
-void FlowHandlePacketUpdate(Flow *f, Packet *p)
+void FlowHandlePacketUpdate(Flow *f, Packet *p, ThreadVars *tv, DecodeThreadVars *dtv)
 {
     SCLogDebug("packet %"PRIu64" -- flow %p", p->pcap_cnt, f);
 
@@ -453,6 +453,14 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
             f->flags &= ~FLOW_PROTO_DETECT_TS_DONE;
             p->flags |= PKT_PROTO_DETECT_TS_DONE;
         }
+        if (MacSetFlowStorageEnabled()) {
+            MacSet *ms = FlowGetStorageById(f, MacSetGetFlowStorageID());
+            if (ms != NULL && p->ethh != NULL) {
+                MacSetAddWithCtr(ms, p->ethh->eth_src, p->ethh->eth_dst, tv,
+                                dtv->counter_max_mac_addrs_src,
+                                dtv->counter_max_mac_addrs_dst);
+            }
+        }
     } else {
         f->tosrcpktcnt++;
         f->tosrcbytecnt += GET_PKT_LEN(p);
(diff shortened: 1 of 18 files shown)


src/flow.c
```diff
@@ -453,6 +453,14 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
             f->flags &= ~FLOW_PROTO_DETECT_TS_DONE;
             p->flags |= PKT_PROTO_DETECT_TS_DONE;
         }
+        if (MacSetFlowStorageEnabled()) {
```
victorjulien, line 456: `if (p->ethh && MacSetFlowStorageEnabled())` to avoid a small check and the following `FlowGetStorageById` check for non-ethernet.

victorjulien: Would also like to see this logic move into a small `inline` util func to keep this code as clean as possible

src/flow.c
```diff
@@ -468,6 +476,15 @@ void FlowHandlePacketUpdate(Flow *f, Packet *p)
             f->flags &= ~FLOW_PROTO_DETECT_TC_DONE;
             p->flags |= PKT_PROTO_DETECT_TC_DONE;
         }
+        if (MacSetFlowStorageEnabled()) {
+            MacSet *ms = FlowGetStorageById(f, MacSetGetFlowStorageID());
+            SCLogNotice("macset:::(%d) %p", MacSetGetFlowStorageID(), ms);
```
victorjulien, line 481: left over debug comment

satta: ah, missed this case, thanks!

src/flow.c
satta: > `if (p->ethh && MacSetFlowStorageEnabled())` to avoid a small check and the following `FlowGetStorageById` check for non-ethernet.

Sure, I can pull up the `p->ethh` NULL check to the outer `if` confdition but will need to keep the check for the result of `FlowGetStorageById` being NULL, right?
My idea:
```C
static inline void FlowUpdateEthernet(Flow *f, EthernetHdr *ethh, ThreadVars *tv,
                                      DecodeThreadVars *dtv, bool toserver)
{
    if (ethh && MacSetFlowStorageEnabled()) {
        MacSet *ms = FlowGetStorageById(f, MacSetGetFlowStorageID());
        if (ms != NULL) {
            if (toserver) {
                MacSetAddWithCtr(ms, ethh->eth_src, ethh->eth_dst, tv,
                                 dtv->counter_max_mac_addrs_src,
                                 dtv->counter_max_mac_addrs_dst);
            } else {
                MacSetAddWithCtr(ms, ethh->eth_dst, ethh->eth_src, tv,
                                 dtv->counter_max_mac_addrs_dst,
                                 dtv->counter_max_mac_addrs_src);
            }
        }
    }
}
```

victorjulien: Yes that looks sane. The `FlowGetStorageById` result needs to be checked indeed.

Minor style comment: our convention is to have `tv` and `dtv` as the first to args of the func.

satta: OK, got it.

satta:
New PR: #4975

PR #4970, opened 2020-05-21
Byteextractstr/part2/v5

inashivb:

### DIFF
--- a/src/util-log-redis.c
+++ b/src/util-log-redis.c
@@ -26,6 +26,7 @@
 #include "suricata-common.h" /* errno.h, string.h, etc. */
 #include "util-log-redis.h"
 #include "util-logopenfile.h"
+#include "util-byte.h"
 
 #ifdef HAVE_LIBHIREDIS
 
@@ -514,7 +515,9 @@ int SCConfLogOpenRedis(ConfNode *redis_node, void *lf_ctx)
         SCLogError(SC_ERR_MEM_ALLOC, "Error allocating redis server string");
         exit(EXIT_FAILURE);
     }
-    log_ctx->redis_setup.port = atoi(redis_port);
+    if (StringParseU16RangeCheck(&log_ctx->redis_setup.port, 10, 0, (const char *)redis_port, 0, 65535) < 0) {
+        FatalError(SC_ERR_INVALID_VALUE, "Invalid value for redis port: %s", redis_port);
+    }
     log_ctx->Close = SCLogFileCloseRedis;
 
 #ifdef HAVE_LIBEVENT
--- a/src/util-napatech.c
+++ b/src/util-napatech.c
@@ -800,10 +800,20 @@ static uint32_t CountWorkerThreads(void)
 
                         char copystr[16];
                         strlcpy(copystr, lnode->val, 16);
-
-                        ByteExtractStringUint8(&start, 10,  0, copystr);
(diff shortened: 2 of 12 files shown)


src/util-napatech.c
```diff
@@ -948,16 +958,28 @@ int NapatechGetStreamConfig(NapatechStreamConfig stream_config[])
...
-                    ByteExtractStringUint16(&end, 10,  0, strchr(copystr, '-') + 1);
+                    if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
+                        FatalError(SC_ERR_INVALID_VALUE, "Napatech invalid "
+                                   "stream_id start: '%s'", copystr);
+                    }
+                    char *end_str = strchr(copystr, '-');
+                    if ((end_str == NULL) || (end_str != NULL && StringParseUint8(&end,
+                                    10, 0, (const char *) (end_str + 1)) < 0)) {
+                        FatalError(SC_ERR_INVALID_VALUE, "Napatech invalid "
+                                   "stream_id end: '%s'", strchr(copystr, '-') + 1);
```
victorjulien, line 969: if `end_str` is `NULL` it means that this will try to print as a string `NULL + 1`?

src/util-proto-name.c
```diff
@@ -60,10 +61,9 @@ void SCProtoNameInit()
             if (proto_ch == NULL)
                 continue;
 
-            int proto = atoi(proto_ch);
-            if (proto >= 255)
+            uint8_t proto;
+            if (StringParseU8RangeCheck(&proto, 10, 0, (const char *)proto_ch, 0, 255) < 0)
```
victorjulien, line 65: I suppose the range check is useless here, as every value a u8 can contain is valid

src/util-log-redis.c
```diff
@@ -514,7 +515,9 @@ int SCConfLogOpenRedis(ConfNode *redis_node, void *lf_ctx)
         SCLogError(SC_ERR_MEM_ALLOC, "Error allocating redis server string");
         exit(EXIT_FAILURE);
     }
-    log_ctx->redis_setup.port = atoi(redis_port);
+    if (StringParseU16RangeCheck(&log_ctx->redis_setup.port, 10, 0, (const char *)redis_port, 0, 65535) < 0) {
```
victorjulien, line 518: range check seems to offer no extra value over checking that the value fits a u16, which the non range check version should already do, right?

src/util-proto-name.c
catenacyber: The range should be useful here.

But `StringParseU8RangeCheck` should do it before doing the cast ie
First
```
    if (*res < min || *res > max) {
        return -1;
    }
```
Then 
`    *res = (uint8_t)u64;`

victorjulien: Should the regular `StringParseU8` validate that the value fits the type as well?

catenacyber: I would expect so

victorjulien: @inashivb can you double check this?

catenacyber:
Is the status "waiting for a replacement PR" ?

inashivb:
> Is the status "waiting for a replacement PR" ?

Yes. I shall send one by tomorrow. Sorry picked up other stuff.

catenacyber:
Thanks, nothing to be sorry about :-)
I just wondered if I had missed something

PR #4973, opened 2020-05-22
Hassh feature 2698 v1

MalakhatkoVadym:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2698

Describe changes:
- Adds hassh and hasshServer for ssh fingerprinting to suricata.
- Adds lua logging to get hassh fingerprints.

### Testing

1. Enable hassh in suricata.yaml in app-layer.protocols.ssh.hassh:

```
app-layer:
  protocols:
    ssh:
      enabled: yes
      hassh: yes
```

2. Check your hassh parameters (hassh, server hassh, hassh string, hassh server string) (connect with ssh to tested host) using tool https://github.com/salesforce/hassh/tree/master/python 
or using suricata's lua scripting:
change suricata.yaml:
```
outputs:
  - lua:
      enabled: yes
      scripts-dir: /etc/suricata/lua-output/
      scripts:
         - hassh.lua
```
hassh.lua:

```
function init (args)
    local needs = {}
    needs["protocol"] = "ssh"
    return needs
end

function setup (args)
    filename = "/tmp/hassh.log"
    file = assert(io.open(filename, "a"))
end

function log(args)
    hassh = HasshGet()
    file:write (hassh .. " -hassh \n")

    hasshServer = HasshServerGet()
    file:write (hasshServer .. " -hasshServer \n")

    hasshString = HasshGetString()
    file:write (hasshString .. " -hasshString \n")

    hasshServerString = HasshServerGetString()
    file:write (hasshServerString .. " -hasshServerString \n")

    file:flush()
end

function deinit (args)
    file:close(file)
end

return 0
```

3. Add rules for hassh detection:
```
alert ssh any any -> any any (msg:"match SSH hash"; hassh; content:"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"; sid:1000010;)
alert ssh any any -> any any (msg:"match SSH hash-server"; hasshServer; content:"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"; sid:1000020;)
alert ssh any any -> any any (msg:"match SSH hash-string"; hassh.string; content:"none,zlib@openssh.com,zlib"; sid:1000030;)
alert ssh any any -> any any (msg:"match SSH hash-server-string"; hasshServer.string; content:"umac-64-etm@openssh.com,umac-128-etm@openssh.com,"; sid:1000040;)
```
4. Check alerts.

PR #4974, opened 2020-05-23
Hassh feature 2698 v2

MalakhatkoVadym:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2698

Describe changes:
- Adds hassh and hasshServer for ssh fingerprinting to suricata.
- Adds lua logging to get hassh fingerprints.

Modifies #4973 by fixing commits.

### Testing

1. Enable hassh in suricata.yaml in app-layer.protocols.ssh.hassh:

```
app-layer:
  protocols:
    ssh:
      enabled: yes
      hassh: yes
```

2. Check your hassh parameters (hassh, server hassh, hassh string, hassh server string) (connect with ssh to tested host) using tool https://github.com/salesforce/hassh/tree/master/python 
or using suricata's lua scripting:
change suricata.yaml:
```
outputs:
  - lua:
      enabled: yes
      scripts-dir: /etc/suricata/lua-output/
      scripts:
         - hassh.lua
```
hassh.lua:

```
function init (args)
    local needs = {}
    needs["protocol"] = "ssh"
    return needs
end

function setup (args)
    filename = "/tmp/hassh.log"
end

function log(args)
    hassh = HasshGet()
    file:write (hassh .. " -hassh \n")

    hasshServer = HasshServerGet()
    file:write (hasshServer .. " -hasshServer \n")

    hasshString = HasshGetString()
    file:write (hasshString .. " -hasshString \n")

    hasshServerString = HasshServerGetString()
    file:write (hasshServerString .. " -hasshServerString \n")

    file:flush()
end

function deinit (args)
    file:close(file)
end

return 0
```

3. Add rules for hassh detection:
```
alert ssh any any -> any any (msg:"match SSH hash"; hassh; content:"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"; sid:1000010;)
alert ssh any any -> any any (msg:"match SSH hash-server"; hasshServer; content:"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"; sid:1000020;)
alert ssh any any -> any any (msg:"match SSH hash-string"; hassh.string; content:"none,zlib@openssh.com,zlib"; sid:1000030;)
alert ssh any any -> any any (msg:"match SSH hash-server-string"; hasshServer.string; content:"umac-64-etm@openssh.com,umac-128-etm@openssh.com,"; sid:1000040;)
```
4. Check alerts.


rust/src/ssh/detect.rs
```diff
@@ -90,3 +90,135 @@ pub extern "C" fn rs_ssh_tx_get_software(
...
+) -> u8 {
+    let tx = cast_pointer!(tx, SSHTransaction);
+    match direction {
+        STREAM_TOSERVER => {
+            let m = &tx.cli_hdr.hassh;
+            if m.len() > 0 {
+                unsafe {
+                    *buffer = m.as_ptr();
+                }
+                unsafe {
```
catenacyber, line 109: Can you group the `unsafe` ?

rust/src/ssh/detect.rs
```diff
@@ -90,3 +90,135 @@ pub extern "C" fn rs_ssh_tx_get_software(
...
+    unsafe {
+        *buffer_len = 0;
+    }
+
+    return 0;
+}
+
+#[no_mangle]
+pub extern "C" fn rs_ssh_tx_get_hassh_server(
+    tx: *mut std::os::raw::c_void,
```
catenacyber, line 129: Why do we need two functions ? (When one parameter is `direction`)

rust/src/ssh/ssh.rs
```diff
@@ -156,20 +160,31 @@ impl SSHState {
...
-                    if head.msg_code == SSH_MSG_NEWKEYS {
-                        hdr.flags = SSHConnectionState::SshStateFinished;
-                        if ohdr.flags >= SSHConnectionState::SshStateFinished {
-                            unsafe {
-                                AppLayerParserStateSetFlag(
-                                    pstate,
-                                    APP_LAYER_PARSER_NO_INSPECTION
+                    match head.msg_code {
+                        parser::MessageCode::SshMsgKexinit if unsafe {SSHHasshIsEnabled()} && !input.is_empty() => {
+                        	match parser::parse_packet_key_exchange(&input[SSH_RECORD_HEADER_LEN..]) {
```
catenacyber, line 165: Why do we need `!input.is_empty()` ? (and not use `rem` ?)

rust/src/ssh/ssh.rs
```diff
@@ -179,8 +194,19 @@ impl SSHState {
...
                             hdr.record_left = head.pkt_len - 2 - remlen;
                             //header with rem as incomplete data
-                            if head.msg_code == SSH_MSG_NEWKEYS {
-                                hdr.flags = SSHConnectionState::SshStateFinished;
+                            match head.msg_code {
+                                parser::MessageCode::SshMsgNewKeys => {
+                                    hdr.flags = SSHConnectionState::SshStateFinished;
+                                }
+                                parser::MessageCode::SshMsgKexinit if unsafe {SSHHasshIsEnabled()} && !input.is_empty() => {
+                                    match parser::parse_packet_key_exchange(&input[SSH_RECORD_HEADER_LEN..]) {
```
catenacyber, line 202: Does it make sense to parse the incomplete `SshMsgKexinit` record ?

rust/src/ssh/parser.rs
```diff
@@ -81,7 +98,10 @@ named!(pub ssh_parse_record_header<SshRecordHeader>,
         pkt_len: verify!(be_u32, |&val| val > 1) >>
         padding_len: be_u8 >>
         msg_code: be_u8 >>
-        (SshRecordHeader{pkt_len, padding_len, msg_code})
+        (SshRecordHeader{pkt_len: pkt_len,
+            padding_len: padding_len,
+            msg_code: num::FromPrimitive::from_u8(msg_code).unwrap_or(MessageCode::SshMsgUndefined)})
```
catenacyber, line 103: I understand this pattern, but we are losing information, that we might want in future versions (ie detect on `ssh.msh_code = 8`)...
Not sure what the best is...

rust/src/ssh/parser.rs
```diff
@@ -92,10 +112,91 @@ named!(pub ssh_parse_record<SshRecordHeader>,
...
+    pub mac_algs_client_to_server: &'a [u8],
+    pub mac_algs_server_to_client: &'a [u8],
+    pub comp_algs_client_to_server: &'a [u8],
+    pub comp_algs_server_to_client: &'a [u8],
+    pub langs_client_to_server: &'a [u8],
+    pub langs_server_to_client: &'a [u8],
+    pub first_kex_packet_follows: u8,
+}
+
+use md5::compute;
```
catenacyber, line 138: Why is it md5 ? Is it a standard ?

src/detect-ssh-hassh-server.c
```diff
@@ -0,0 +1,205 @@
...
+        if (cd->flags & DETECT_CONTENT_NOCASE) {
+            *sigerror = "hasshServer should not be used together with "
+                        "nocase, since the rule is automatically "
+                        "lowercased anyway which makes nocase redundant.";
+            SCLogWarning(SC_WARN_POOR_RULE, "rule %u: %s", s->id, *sigerror);
+        }
+
+        if (cd->content_len == 32)
+            return TRUE;
+
```
catenacyber, line 142: Should we check that we have hexadecimal characters as well ?

rust/src/ssh/logger.rs
```diff
@@ -35,6 +35,18 @@ fn log_ssh(tx: &SSHTransaction) -> Option<Json> {
                 &tx.cli_hdr.swver,
             );
         }
+        if tx.cli_hdr.hassh.len() > 0 {
+            cjs.set_string_from_bytes(
+                "hassh",
```
catenacyber, line 40: Is not it a big field to log unconditionally ?

doc/userguide/rules/hassh-keywords.rst
```diff
@@ -0,0 +1,63 @@
...
+-----
+
+Match on hassh (md5 of of hassh algorithms of client).
+
+Example::
+  alert ssh any any -> any any (msg:"match hassh"; \
+      hassh; content:"ec7378c1a92f5a8dde7e8b7a1ddf33d1";\
+      sid:1000010;)
+      
+``hassh`` is a 'sticky buffer'.
```
catenacyber, line 18: I wonder if a sticky buffer makes sense for some md5...

doc/userguide/rules/index.rst
```diff
@@ -16,6 +16,7 @@ Suricata Rules
    dns-keywords
    tls-keywords
    ssh-keywords
+   hassh-keywords
```
catenacyber, line 19: I think hassh should be port of ssh keywords

MalakhatkoVadym: Better to have keywords like: ssh.hassh, ssh.hassh.server, ssh.hassh.string, ssh.hassh.server.string?

rust/src/ssh/detect.rs
MalakhatkoVadym: Now, there are two functions ```rs_ssh_tx_get_hassh``` (for hassh) and  ```rs_ssh_tx_get_hassh_string``` (for hassh string)

rust/src/ssh/parser.rs
MalakhatkoVadym: No, but it's the most lightweight and best-fitting library.
I checked a few other options (https://lib.rs/keywords/md5)
Good options are: 
https://lib.rs/crates/md-5
https://lib.rs/crates/rust-crypto
I will make some performance tests.

catenacyber: This comes in answer to other remarks as well
But did you know about md5 transform ?
https://suricata.readthedocs.io/en/suricata-5.0.3/rules/transforms.html#to-md5

doc/userguide/rules/index.rst
catenacyber: Yes like ssh.something
Why not `ssh.algorithms` ?

catenacyber: Oh hassh seems known already, so ok for `ssh.hassh`

rust/src/ssh/detect.rs
catenacyber: I meant two functions like `rs_ssh_tx_get_hassh` and `rs_ssh_tx_get_hassh_server` when we could have one and use the direction

But then, with the md5 transform, we could have only one new keyword instead of 4 and still all the functionalities : https://suricata.readthedocs.io/en/suricata-5.0.3/rules/transforms.html#to-md5

rust/src/ssh/parser.rs
MalakhatkoVadym: yes, I looked at it, but I was trying to make hassh functionality very similar to ja3/ja3s, because it's very similar 
https://suricata.readthedocs.io/en/suricata-5.0.0/rules/ja3-keywords.html
https://suricata.readthedocs.io/en/suricata-5.0.0/lua/lua-functions.html?highlight=ja3#ja3

I can use function like ```char *Ja3GenerateHash(JA3Buffer *buffer)``` to get hash.

victorjulien: > Why is it md5 ? Is it a standard ?

It seems to be "HASSH is a network fingerprinting standard which can be used to identify specific Client and Server SSH implementations. The fingerprints can be easily stored, searched and shared in the form of a small MD5 fingerprint." (https://github.com/salesforce/hassh)

src/detect-ssh-hassh-server.c
MalakhatkoVadym:
```suggestion
if (cd->content_len != 32)
        {
            *sigerror = "Invalid length of the specified hassh (should "
                        "be 32 characters long). This rule will therefore "
                        "never match.";
            SCLogWarning(SC_WARN_POOR_RULE,  "rule %u: %s", s->id, *sigerror);
            return FALSE;
        }
        for (size_t i = 0; i < cd->content_len; ++i)
        {
            if(!isxdigit(cd->content[i])) 
            {
                *sigerror = "Invalid hassh string (should be string of hexademical characters)."
                            "This rule will therefore never match.";
                SCLogWarning(SC_WARN_POOR_RULE,  "rule %u: %s", s->id, *sigerror);
                return FALSE;
            }
        }
```

rust/src/ssh/parser.rs
catenacyber: Thanks @victorjulien 
I gather it is good to use `ssh.hassh` as a keyword as it is readable.
Then, should we somehow internally translate `ssh.hassh` to some `ssh.algorithms; to_md5` ?

doc/userguide/rules/hassh-keywords.rst
MalakhatkoVadym: I was checking ja3/ja3s (it's sticky buffer) use cases. 
https://suricata.readthedocs.io/en/suricata-4.1.0/rules/ja3-keywords.html
So, I decided to make hassh 'sticky buffer' also.

rust/src/ssh/parser.rs
MalakhatkoVadym: I added all possible message code options. There are few ways to store undefined code:
1. Add logging for undefined message code:
```suggestion
            msg_code: num::FromPrimitive::from_u8(msg_code).unwrap_or(MessageCode::SshMsgUndefined)})msg_code: num::FromPrimitive::from_u8(msg_code).unwrap_or_else(|| { 
                SCLogDebug!("SSH undefined message code {}", msg_code);
                MessageCode::SshMsgUndefined
            })})
```

2. Change SshMsgUndefined enum value to be able to store ```u8``` and add separate function for ```to_u8``` conversion:
```
impl MessageCode {
    fn from_u8(value: u8) -> MessageCode {
        match value {
            1 => SshMsgDisconnect,
            2 => SshMsgIgnore,
            3 => SshMsgUnimplemented,
            4 => SshMsgDebug,
            5 => SshMsgServiceRequest,
            6 => SshMsgServiceAccept,
            20 => SshMsgKexinit,
            21 => SshMsgNewKeys,
            30 => SshMsgKexdhInit,
            31 => SshMsgKexdhReply,
            _ => SshMsgUndefined(value),
    }
}
```

3. Remove enum from ```struct SshRecordHeader``` and use const values.

Which option can you suggest?

rust/src/ssh/parser.rs
victorjulien: I think we can't really avoid having the md5 done directly as we should also consider the output case. In logging we'd also want to log the fingerprint (just like with ja3(s)).

Side thought: it would be interesting to see some benchmarks of different md5/sha1/sha256 implementations. How do the rust ones compare to libnss?

catenacyber: So, thanks @victorjuilen and thanks @MalakhatkoVadym

Remaining questions is wether to use rust md5 crate or libnss md5 if I understand, right ?

MalakhatkoVadym: @catenacyber @victorjulien 
I created a function for benchmarking:
```
char* GenerateHash(const unsigned char* buffer, int len) {
    char *ja3_hash = SCMalloc(MD5_STRING_LENGTH * sizeof(char));
    if (ja3_hash == NULL) {
        SCLogError(SC_ERR_MEM_ALLOC,
                   "Error allocating memory for JA3 hash");
        return NULL;
    }

    unsigned char md5[MD5_LENGTH];
    HASH_HashBuf(HASH_AlgMD5, md5, (unsigned char *)buffer, len);

    int i, x;
    for (i = 0, x = 0; x < MD5_LENGTH; x++) {
        i += snprintf(ja3_hash + i, MD5_STRING_LENGTH - i, "%02x", md5[x]);
    }
    return ja3_hash;
}
```
rust md5 is : 9606 nanosec
libnsstime is : 18699 nanosec
rust md5 is : 10682 nanosec
libnsstime is : 16428 nanosec
rust md5 is : 4378 nanosec
libnsstime is : 7393 nanosec
rust md5 is : 4627 nanosec
libnsstime is : 7426 nanosec
rust md5 is : 10845 nanosec
libnsstime is : 15630 nanosec
rust md5 is : 10076 nanosec
libnsstime is : 16080 nanosec

Average time for rust md5 library is better than calling libnss function, so I'll leave md5 `compute` function.

rust/src/ssh/ssh.rs
MalakhatkoVadym: yes, my bad, `!input.is_empty()` is unnecessary, and `rem` is not used, because it's empty for msg_code == 20 (I was trying to use it firstly). Will try to find a problem with `rem`.

rust/src/ssh/ssh.rs
MalakhatkoVadym: yes, you're right, incorrect hassh can be created.

rust/src/ssh/logger.rs
MalakhatkoVadym: I think, it's ok to log hassh parameters if hassh is enabled in `app-layer.protocols.ssh.hassh`.

doc/userguide/rules/hassh-keywords.rst
catenacyber: ok

rust/src/ssh/logger.rs
catenacyber: ok

rust/src/ssh/parser.rs
catenacyber: The second option looks nice (but I am not a Rust expert)

MalakhatkoVadym: Used second option in https://github.com/OISF/suricata/pull/5048

PR #4977, opened 2020-05-25
RFC Feature/60x/decoder data store/v8

victorjulien:
Initial work on 'decoder store', where the packet fields stored in the `Packet` are more dynamic.

From https://github.com/OISF/suricata/commit/efaf3bc9b902f968102d9c2e44edf82924108af0
A fixed size static store is part of the Packet structure. If that space
isn't sufficient the store can dynamically expand. It uses a list of
storage blocks for this.

The primary goals are:
- don't use space in the packet when a feature/proto isn't in use
- reduce (or avoid) the need for new protocols to manually add to the
  Packet structure

In general the storage is a TLV (type length value) store, but some types
get special support to reduce the space overhead of the TLV header.

Implemented so far:
- VLAN (2 bytes)
- MPLS (4 bytes)
- TLV (3 bytes + the data to be stored)

Right now the only 'consumer' is EVE, where the current PR adds
```json
  "decoder_store": [
    {
      "mac": "52:54:00:12:35:02"
    },
    {
      "mac": "08:00:27:bf:4f:8a"
    },
    {
      "vlan": 666
    },
    {
      "vlan": 4094
    }
  ],
```
and
```json
  "decoder_store": [
    {
      "mac": "00:00:00:06:fd:01"
    },
    {
      "mac": "17:73:74:61:74:65"
    },
    {
      "mpls": 291956
    }
  ],
```

Looking for general feedback. Also wondering if this should be extended to the `Flow`, and perhaps somehow even be taken into account for flow hash lookups.

### DIFF
--- a/src/decode-store.c
+++ b/src/decode-store.c
@@ -255,6 +255,64 @@ void DecodeStoreCleanup(Packet *p)
     s->next = NULL;
 }
 
+void DecodeStoreIterate(const Packet *p,
+        void (*Callback)(void *cb_data, const uint8_t type, const uint8_t etype, const uint8_t size, const uint8_t *data),
+        void *cb_data)
+{
+    for (const struct DecodeStore *s = &p->decode_store; s != NULL; s = s->next) {
+        uint8_t offset = 0;
+        while (offset < s->data_offset) {
+            uint8_t type, size;
+            if (GetTypeAndSize(&s->data[offset], s->data_offset - offset, &type, &size) != 1)
+                break;
+
+            switch (type) {
+                case DECODE_STORE_TYPE_MPLS: {
+                    SCLogDebug("MPLS type %d size %u", type, size);
+                    DecodeStoreBuiltinMPLS v;
+                    memcpy(&v, &s->data[offset], sizeof(v));
+                    union {
+                        uint32_t label;
+                        uint8_t data[4];
+                    } u;
+                    u.label = v.value;
+                    Callback(cb_data, type, 0, 4, u.data);
+                    break;
+                }
+                case DECODE_STORE_TYPE_VLAN: {
+                    SCLogDebug("VLAN type %d size %u", type, size);
+                    DecodeStoreBuiltinVLAN v;
+                    memcpy(&v, &s->data[offset], sizeof(v));
+                    union {
+                        uint16_t vlan_id;
+                        uint8_t data[2];
+                    } u;
+                    u.vlan_id = v.value;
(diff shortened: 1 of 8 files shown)


jasonish:
> Looking for general feedback. Also wondering if this should be extended to the `Flow`, and perhaps somehow even be taken into account for flow hash lookups.

I think its important to have this same information in flow records as well. Not sure I like how generic this is to get at this fields under the decoder_store. I also feel like that "decoder_store" name should be an internal implementation detail and not be exposed in the JSON.  Is there a better name we can use here?

src/decode-store.c
```diff
@@ -0,0 +1,320 @@
...
+        return;
+    }
+    s = s->next;
+    Copy(s, data, data_size);
+    SCLogDebug("got additional store. Done");
+    PrintInfo(s);
+}
+
+
+typedef union DecodeStoreBuiltinVLAN {
```
jasonish, line 88: Not liking the "Builtin" in the name too much.  "DecoderStoreTypeVlan", "DecodeStoreTypeGeneric", etc.

src/decode-store.c
```diff
@@ -0,0 +1,320 @@
...
+    Append(p, buf, buf_size);
+    SCLogDebug("TLV added %u datalen %u total %u", etype, data_len, hdr.len);
+    return 0;
+}
+
+static inline int GetTypeAndSize(const uint8_t *data, uint8_t size, uint8_t *type, uint8_t *out_size)
+{
+    assert(size >= 1);
+
+    *type = *data & 0x0f;
```
catenacyber, line 151: Will 16 enough for types ?

catenacyber:
What is the status on this ?
To be closed ?

src/decode-store.c
victorjulien: the idea was that there are some built-in types and then a flexible general purpose TLV type that allows 256 subtypes. The builtins should be extra compact/efficient.

satta:
Bump -- this would be interesting/useful to me as well as I'm looking at MPLS logging atm and I already felt guilty for adding more fields to the packet and flow structs in my branch ;)

satta:
I am also still curious what the status is here.

victorjulien:
@satta, not sure. I think it needs a rebase and I also really dislike the macro hack (TOP/BOT) think I did... feel free to play around with it.

satta:
Yeah, quite a big rebase as quite a bit seems to have changed in the packet department. I'll see when i can find the time to do the rebase (and hopefully get a better idea of the change in the process).

PR #4936, opened 2020-05-12
JsonBuilder - v30

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/4929

Changes from last PR:
- Fix escaping of non-ascii, but valid UTF-8 strings
- Bring back our own encoding for invalid UTF-8 strings where we display unprintable bytes as \x<HEX>.
- Minor optimizations in JsonBuilder (consolidate 2 static string pushes into one), remove a format!.

PR #4979, opened 2020-05-25
JsonBuilder - v31

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/4936

Changes from last PR:
- Convert SMTP
- Convert RFB
- Add "set_mark" and "reset_to_mark" to JsonBuilder

SMTP makes use of optionally returning a new JSON object
to extend the current one and I'm not sure on the best
way to support this in JsonBuilder. One option is to
create a new object, and then extend the root object,
however this eliminates much of the benefit of JsonBuilder.
The other option is to save state (mark), which I'm still
working on. It can't be used for multiple levels of saved
state as that requires more allocations, so may as just
create new objects. Still investigating.

suricata-verify-repo: https://github.com/jasonish/suricata-verify
suricata-verify-branch: dhcp-jsonbuilder/v1


victorjulien:
Side issue:
```
suricata-verify-repo: https://github.com/jasonish/suricata-verify
suricata-verify-branch: dhcp-jsonbuilder/v1
```
I would really like this to switch to a PR based approach, at least by default. Now I'll have to pull things from whatever branch/repo to test validate. Opening a PR is also a statement of intent to merge / get feedback. Just a branch is... just a branch :)

jasonish:
Continued at https://github.com/OISF/suricata/pull/5007.

jasonish:
> I would really like this to switch to a PR based approach, at least by default. Now I'll have to pull things from whatever branch/repo to test validate. Opening a PR is also a statement of intent to merge / get feedback. Just a branch is... just a branch :)

See https://github.com/OISF/suricata/pull/5007.

PR #4982, opened 2020-05-26
Feat/cond logging/v8

victorjulien:
Continuing #4696.

Consolidate `detect_flags` and `logged` flags in the new `AppLayerTxData`. I'm thinking about also moving the `de_state` and `events` into it, to unify everything into a single object.

Next is a way to signal protocols that use single direction 'transactions'. I was thinking about having a registration time flag similar to `APP_LAYER_PARSER_OPT_ACCEPT_GAPS`: `APP_LAYER_PARSER_UNIDIR_TXS` or something.

Then for a case where we need to inform the parser we need to store something for the opposing side, we'd have something like (pseudo code):
```c
static void ConfigApplyTx(Flow *f,
        const uint64_t tx_id, const DetectConfigData *config)
{
    void *tx = AppLayerParserGetTx(f->proto, f->alproto, f->alstate, tx_id);
    if (tx) {
        AppLayerTxData *txd = AppLayerParserGetTxData(f->proto, f->alproto, tx);
        if (txd) {
            txd->log_flags |= BIT_U8(config->type);
            if (APP_LAYER_PARSER_UNIDIR_TXS) {
                AppLayerTxData request = AppLayerTxData::InitWith(txd->log_flags)
                AppLayerParserPropagateTxData(f->proto, f->alproto, tx, &request);
```
It would then be up to the per parser `AppLayerParserPropagateTxData` callback implementation to make sure the settings are properly propagated to the 'opposing side' tx.

Passed.


jasonish:
> It would then be up to the per parser `AppLayerParserPropagateTxData` callback implementation to make sure the settings are properly propagated to the 'opposing side' tx.

As discussed outside of this ticket, something like DNS could use a FIFO of fixed to store this information.  Where this could fail is when the number of in-flight requests/responses is greater than the FIFO.  For DNS, this could be an array of u16's.  Ultimately this would be simpler than moving DNS back to a request/response like transaction. I suppose it would only see the overhead if configurable logging was enabled?

rust/src/applayer.rs
```diff
@@ -24,6 +24,40 @@ use crate::filecontainer::FileContainer;
 use crate::applayer;
 use std::os::raw::{c_void,c_char,c_int};
 
+#[repr(C)]
+#[derive(Debug)]
+pub struct AppLayerTxData {
+    /// config: log flags
+    log_flags: u8,
+
+    logged: LoggerFlags,
```
jasonish, line 33: naming: this is a little confusing.. log_flags and LoggerFlags?

victorjulien:
> > It would then be up to the per parser `AppLayerParserPropagateTxData` callback implementation to make sure the settings are properly propagated to the 'opposing side' tx.
> 
> As discussed outside of this ticket, something like DNS could use a FIFO of fixed to store this information. Where this could fail is when the number of in-flight requests/responses is greater than the FIFO. For DNS, this could be an array of u16's. Ultimately this would be simpler than moving DNS back to a request/response like transaction. I suppose it would only see the overhead if configurable logging was enabled?

I think for DNS you'd want a lookup method that can take the DNS tx id (from the traffic, not the internal suri tx) and do a quick lookup to see if there is a config to 'propagate'. If there'd be a data structure where you can do a lookup in a Hashmap style store, but also in a cheap way iterate it by age, it might solve the issue of what happens if we get lots of requests that we don't see a reply for.

jasonish:
> I think for DNS you'd want a lookup method that can take the DNS tx id (from the traffic, not the internal suri tx) and do a quick lookup to see if there is a config to 'propagate'. If there'd be a data structure where you can do a lookup in a Hashmap style store, but also in a cheap way iterate it by age, it might solve the issue of what happens if we get lots of requests that we don't see a reply for.

I've seen some LinkedHashMaps, but they make heavy use of unsafe and are not in the standard library.  To start, we could probably wrap a VecDeque and a HashSet.

https://gist.github.com/jasonish/0d62ff12392605fd39bb9177eff3ec5f

For DNS its simple, as the data we need to store is just a u16.  But if the data were an object, we could wrap it in Rc.

Also, in this example we don't purge after an item is found as removal from the middle would be expensive. So we only purge to clear up room. I suppose it might make sense to use a HashMap with a bool value that could be set when after an entry has been acted on. My thinking here is that a tx_id is reused before a previous one is purged. This might require a bit more thought, as when we age out the older duplicate tx_id, we don't want to purge it from the map either.  But I think this example serves to get a basic idea across.

jasonish:
> This might require a bit more thought, as when we age out the older duplicate tx_id, we don't want to purge it from the map either. But I think this example serves to get a basic idea across.

Could solve this if the content of the map and list was a tuple of (dns_tx_id, internal_tx_id). Then we have enough context except when there are duplicate dns_tx_id's on a state, but that would cause an issue with any request/response correlation.

src/util-config.h
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+/**
+ * \file
+ *
+ * \author Victor Julien <victor@inliniac.net>
+ */
+
+#ifndef __UTIL_CONFIG_H__
+#define __UTIL_CONFIG_H__
```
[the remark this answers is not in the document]
victorjulien: Yeah agreed. Not sure about a name though. This feature will touch many things eventually, so putting it in 'app-layer' wouldn't make sense. Maybe just in detect-config-common.h or something.

rust/src/applayer.rs
victorjulien: Fixed

victorjulien:
Ultimately the way the protocol applies the config request is up to the protocol. In #5037 i've implemented the callback.

Replaced by #5037

PR #4983, opened 2020-05-26
fix Bug #3726 suricata crash on rule reloading

amirabell:
This changseset fixes a bug causing a segmentation fault.
When rules are reloaded and a rule using libmagic matches, suricata
crashes due to an improper reinitialization of the thread contexts.

https://redmine.openinfosecfoundation.org/issues/3726

Describe changes:
- When invoking DetectFilemagicSetup we always call DetectRegisterThreadCtxFuncs 
- otherwise when rules are reloaded g_magic_thread_ctx_id will point to the wrong struct
- and it will lead to a segmentation fault.

### DIFF
--- a/src/detect-filemagic.c
+++ b/src/detect-filemagic.c
@@ -416,13 +416,11 @@ static int DetectFilemagicSetup (DetectEngineCtx *de_ctx, Signature *s, const ch
     if (filemagic == NULL)
         return -1;
 
-    if (g_magic_thread_ctx_id == -1) {
-        g_magic_thread_ctx_id = DetectRegisterThreadCtxFuncs(de_ctx, "filemagic",
-                DetectFilemagicThreadInit, (void *)filemagic,
-                DetectFilemagicThreadFree, 1);
-        if (g_magic_thread_ctx_id == -1)
-            goto error;
-    }
+    g_magic_thread_ctx_id = DetectRegisterThreadCtxFuncs(de_ctx, "filemagic",
+            DetectFilemagicThreadInit, (void *)filemagic,
+            DetectFilemagicThreadFree, 1);
+    if (g_magic_thread_ctx_id == -1)
+        goto error;
 
     /* Okay so far so good, lets get this into a SigMatch
      * and put it in the Signature. */


src/detect-filemagic.c
```diff
@@ -416,13 +416,11 @@ static int DetectFilemagicSetup (DetectEngineCtx *de_ctx, Signature *s, const ch
...
         return -1;
 
-    if (g_magic_thread_ctx_id == -1) {
-        g_magic_thread_ctx_id = DetectRegisterThreadCtxFuncs(de_ctx, "filemagic",
-                DetectFilemagicThreadInit, (void *)filemagic,
-                DetectFilemagicThreadFree, 1);
-        if (g_magic_thread_ctx_id == -1)
-            goto error;
-    }
+    g_magic_thread_ctx_id = DetectRegisterThreadCtxFuncs(de_ctx, "filemagic",
```
victorjulien, line 419: So we are overwriting this global id during reload? Seems rather dangerous?

amirabell: Actually if you invoke DetectRegisterThreadCtxFuncs with mode=1 (last parameter) it should return the same id (if existent) and it should be safe.

catenacyber: The fix seems consistent with detect-pcre.c and detect-lua.c

@amirabell
How can we reproduce the bug ? Is it some race condition as I understand ?

amirabell: It doesn't look to me a race condition. 
I have updated the redmine ticket and uploaded the build-info output and a .zip with the pcap and the rule file to reproduce the bug.

Steps to reproduce the bug:
- Start suricata in PCAP mode
- Send signal SIGUSR2 to reload the signatures
- Replay the pcap on the sniffing interface

victorjulien: > Actually if you invoke DetectRegisterThreadCtxFuncs with mode=1 (last parameter) it should return the same id (if existent) and it should be safe.

But if it does, then what is the effect of running it again?

victorjulien:
Btw can you fixup the checkboxes by removing the spaces around the x's? `[x]` vs `[ x ]`

src/detect-filemagic.c
amirabell: > > Actually if you invoke DetectRegisterThreadCtxFuncs with mode=1 (last parameter) it should return the same id (if existent) and it should be safe.
> 
> But if it does, then what is the effect of running it again?

I meant in general but it is not what happens in this case. I try to explain it as clear as possible.
Suppose we have signatures for both libmagic and lua.
When we start suricata at some point we invoke DetectEngineThreadCtxInitKeywords in detect-engine.c and we populate det_ctx->keyword_ctxs_array for every thread based on de_ctx->keyword_list. Every element has an id and in my setup  g_magic_thread_ctx_id has id 0. 

When we reload the rules we invoke again DetectFilemagicSetup but we keep the same g_magic_thread_ctx_id  and we do not insert in de_ctx->keyword_list the item corresponding to the libmagic engine because we do not invoke DetectRegisterThreadCtxFuncs. This means that an id equal to 0 will match the item registered for another engine (lua for instance) and we then invoke the function associated to libmagic with context data completely unrelated. 
This causes a segfault later on inside libmagic

victorjulien:
Merged in #5533, thanks!

PR #4985, opened 2020-05-28
Http2 v10

catenacyber:

DRAFT for discussion
Not to be merged

Describe changes:
- Adds basic HTTP2 parser

Much more work to be done :
 + more commit squashing
 + parse remaining frame types
 + merge with HTTP1 for logging and detection ?
 + tests
 + robustness against evasions

Modifies #4943 with Rust comments taken into account (one commit for each comment)


rust/src/http2/detect.rs
```diff
@@ -0,0 +1,399 @@
...
+                return -1;
+            }
+            Some(x) => {
+                return *x as i32;
+            }
+        },
+        _ => {}
+    }
+
+    return -1;
```
jasonish, line 51: A the 2 match arms are idential, this could be compressed down to:
```
    match direction {
        STREAM_TOSERVER | STREAM_TOCLIENT => match &tx.ftype {
            None => -1,
            Some(x) => *x as i32
        }
        _ => -1
    }
```

rust/src/http2/detect.rs
```diff
@@ -0,0 +1,399 @@
...
+}
+
+#[no_mangle]
+pub extern "C" fn rs_http2_parse_frametype(
+    str: *const std::os::raw::c_char,
+) -> std::os::raw::c_int {
+    let ft_name: &CStr = unsafe { CStr::from_ptr(str) };
```
jasonish, line 58: Unless there is a specific reason not to (a define function type or something), I'd just go ahead and mark this function unsafe to avoid the clippy warning.

rust/src/http2/detect.rs
```diff
@@ -0,0 +1,399 @@
...
+                        vec.push(hd.blocks[i as usize].name[j]);
+                        /* this escaping is needed
+                         * to avoid amibguity with separator between name and value
+                         */
+                        if hd.blocks[i as usize].name[j] == ':' as u8 {
+                            vec.push(':' as u8);
+                        }
+                    }
+                    vec.push(':' as u8);
+                    vec.push(' ' as u8);
```
jasonish, line 376: very minor, but I wonder if this would save a few cycles:
```
    vec.extend(&[':' as u8, ' ' as u8]);
```

rust/src/http2/detect.rs
catenacyber: Thanks for the Rust tip, I need to get http2 transactions defined right before having a good code here

rust/src/http2/detect.rs
catenacyber: So, leaving this way for now

rust/src/http2/detect.rs
catenacyber: What do you mean ?
What is the clippy warning ?

In general, I try to restrict `unsafe` to the smallest part of the code

jasonish: > In general, I try to restrict `unsafe` to the smallest part of the code

Some discussion on this here: https://github.com/OISF/suricata/pull/4666

Consensus on the Rust side is that if the function uses the raw pointers passed in, the whole thing should be unsafe, and clippy (the Rust linter) will pick this up.

catenacyber: Thanks for the interesting pointer.

I will make the change, and make better reviews next time :-)

Is there an easy command to run clippy on my code ?

catenacyber: `unsafe` added in current v12

PR #4989, opened 2020-05-29
Opt/afp csum/v3

victorjulien:
Use `TP_STATUS_CSUM_VALID` to bypass calculating the csums in AF_PACKET.

Optimize checksum handling by not dereferencing the livedev for something we can keep local.

@regit


victorjulien:
Non-draft proposal #4990

PR #4991, opened 2020-05-29

jlucovsky:
Continuation of #4963

Describe changes:
- Addressed review comments.

### DIFF
--- a/src/detect-bsize.c
+++ b/src/detect-bsize.c
@@ -65,6 +65,8 @@ void DetectBsizeRegister(void)
 #define DETECT_BSIZE_RA 2
 #define DETECT_BSIZE_EQ 3
 
+const char *bsize_mode_strings[] = { "<", ">", "<>", "="};
+
 typedef struct DetectBsizeData {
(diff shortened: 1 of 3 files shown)


src/detect-bsize.c
```diff
@@ -266,6 +268,57 @@ static DetectBsizeData *DetectBsizeParse (const char *str)
...
+            continue;
+
+        DetectContentData *cd = (DetectContentData *) sm->ctx;
+        if (cd == NULL)
+            continue;
+
+        SCLogDebug("Content %.*s, content-len %"PRIu16" depth: %"PRIu16,
+                cd->content_len, cd->content, cd->content_len, cd->depth);
+
+        if (cd->flags & DETECT_CONTENT_DEPTH) {
```
victorjulien, line 289: Just offset is another case. Does this test pass?
```
TEST_FAIL("alert http any any -> any any (http.uri; content:\"abc\"; offset:3; bsize:3; sid:15;)");
```

jlucovsky:
Continued in #4997

PR #4995, opened 2020-05-29
threads/runmode: Changes to thread config behaviour

JoshuaLumb:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3108

Describe changes:
- Previously defining receive thread (RX) counts and worker thread (W) counts displayed counterintuitive behaviour where config values could be ignored or overridden. 
- Factored out a new function in threads.c to return a desired worker thread count compliant with existing queue creation/assignment.
- Runmode autofp RX thread count is now independent of W thread count and is set under interface settings in the config file. Runmode autofp W thread count can be changed from default by enabling CPU affinity settings and specifying number of desired threads under worker-cpu-set.
- Runmode workers W thread count is determined by interface settings in the config file or thread affinity settings (lowest value).
- This is one possible interpretation of a desirable thread count configuration methodology. Others are possible (e.g. allowing "auto" settings to determine W/RX threads as a ratio or e.g. W = MaxThreads - RX, etc.) but desire discussion.
- This does not address any potential NUMA considerations, as mentioned in https://redmine.openinfosecfoundation.org/issues/3695

### DIFF
--- a/src/tm-threads.c
+++ b/src/tm-threads.c
@@ -2178,6 +2178,22 @@ void TmThreadsGetMinimalTimestamp(struct timeval *ts)
     SCLogDebug("ts->tv_sec %"PRIuMAX, (uintmax_t)ts->tv_sec);
 }
 
+uint16_t TmThreadsGetWorkerThreadMax()
+{
+    uint16_t ncpus = UtilCpuGetNumProcessorsOnline();
+    int thread_max = TmThreadGetNbThreads(WORKER_CPU_SET);
+    /* always create at least one thread */
+    if (thread_max == 0)
+            thread_max = ncpus * threading_detect_ratio;
+    if (thread_max < 1)
+        thread_max = 1;
+    if (thread_max > 1024) {
+        SCLogWarning(SC_ERR_RUNMODE, "limited number of 'worker' threads to 1024. Wanted %d", thread_max);
+        thread_max = 1024;
+    }
+    return thread_max;
+}
+
 /**
  *  \retval r 1 if packet was accepted, 0 otherwise
  *  \note if packet was not accepted, it's still the responsibility
--- a/src/tm-threads.h
+++ b/src/tm-threads.h
@@ -243,6 +243,7 @@ int TmThreadsInjectPacketsById(Packet **, int id);
 void TmThreadsInitThreadsTimestamp(const struct timeval *ts);
 void TmThreadsSetThreadTimestamp(const int id, const struct timeval *ts);
 void TmThreadsGetMinimalTimestamp(struct timeval *ts);
+uint16_t TmThreadsGetWorkerThreadMax(void);
 bool TmThreadsTimeSubsysIsReady(void);
 
 #endif /* __TM_THREADS_H__ */
--- a/src/util-runmodes.c
+++ b/src/util-runmodes.c
@@ -97,18 +97,8 @@ int RunModeSetLiveCaptureAutoFp(ConfigIfaceParserFunc ConfigParser,
     char qname[TM_QUEUE_NAME_MAX];
 
     /* Available cpus */
-    uint16_t ncpus = UtilCpuGetNumProcessorsOnline();
     int nlive = LiveGetDeviceCount();
-    int thread_max = TmThreadGetNbThreads(WORKER_CPU_SET);
-    /* always create at least one thread */
-    if (thread_max == 0)
-        thread_max = ncpus * threading_detect_ratio;
-    if (thread_max < 1)
-        thread_max = 1;
-    if (thread_max > 1024) {
-        SCLogWarning(SC_ERR_RUNMODE, "limited number of 'worker' threads to 1024. Wanted %d", thread_max);
-        thread_max = 1024;
-    }
+    uint16_t thread_max = TmThreadsGetWorkerThreadMax();
 
     char *queues = RunmodeAutoFpCreatePickupQueuesString(thread_max);
     if (queues == NULL) {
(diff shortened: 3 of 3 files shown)


victorjulien:
CI fail is a github-ci issue we can ignore.

@norg since you've been looking at these thread counts as well can you review/test the logic?

norg:
I see the described changes which make sense, although I struggle to see them all in the code, only some if those.

I will play around with it.

Since there is also some open discussion on the interpretation part I would bring this up again. So we should decide with https://suricata.readthedocs.io/en/latest/performance/runmodes.html in mind how we would like to have all those scenarios covered in the default.
So for example, we use the amount of cores as the default, this might make sense if we have worker runmode where each thread has the whole pipeline but in autofp I would argue that when we have several packet capture threads and processing threads in addition a 50/50 split might make sense as default.

But I agree it's counterintuitive and we should align the runmodes as best as possible.

norg:
So I took some time to look into that in detail. I see a diff in runmode af-packet between the patch and master. It does not affect pcap runmode (at least from my tests).

I think it's the first step in the right direction. So for example on autofp with 8 core cpu, thread-ratio set to 1 and threads set to 16 in the af-packet section I end up with 8 runner, 8 receive on master, 8 runner, 16 receive with the patch. So it's more clear to have the independence between the capture part and the processing part.

What didn't work for me was setting another value in the affinity settings in the worker-cpu-set, but changing the ratio value changed the amount of threads and the new MIN check came into play. Same between autofp and workers runmode. And yes affinity setting itself was enabled.

I would say it's already an improvement, the move of the function makes sense even without the change on the thread counts. But I would argue that we should define what we want to achieve in general with those different runmode types (like pcap, afpacket etc) and runmodes (workers, autofp). This also depends on the options those run mode types provide.

JoshuaLumb:
Thanks for the review @norg - much appreciated! Here's a quick breakdown of what my system gets under specified settings in runmode autofp with this patch:

Interface Thread Settings | worker-cpu-set | detect-thread-ratio | RX | W
-- | -- | -- | -- | --
auto | - | 1.0 | 12 | 12
auto | - | 2.0 | 12 | 24
auto | 3 | 1.0 | 12 | 3
auto | 3 | 2.0 | 12 | 3
3 | - | 1.0 | 3 | 12
3 | 3 | 1.0 | 3 | 3
3 | - | 2.0 | 3 | 24

I'm curious about what is causing the difference in worker-cpu-set behaviour between your tests and mine. Perhaps we can compare config files?

In regards to documentation and standards - what do you think would be best in terms of expectations? Is the table above appropriate or should the settings behave another way?

norg:
> I'm curious about what is causing the difference in worker-cpu-set behaviour between your tests and mine. Perhaps we can compare config files?

```
threading:
  set-cpu-affinity: yes
  cpu-affinity:
    - management-cpu-set:
        cpu: [ 0 ]  # include only these CPUs in affinity settings
    - receive-cpu-set:
        cpu: [ 0 ]  # include only these CPUs in affinity settings
    - worker-cpu-set:
        cpu: [ "1-5" ]
        mode: "exclusive"
        prio:
          default: "high"
  detect-thread-ratio: 1.0
```

So I would expect 5 W but I still get 8, changing the cpu value doesn't have any effect. So feel free to share your section.

> In regards to documentation and standards - what do you think would be best in terms of expectations? Is the table above appropriate or should the settings behave another way?

I still think the table reflects a better approach but not 100% sure if it's the best approach. I struggle a bit with what would be the best and most reasonable expectation. I would also suggest that we change the output from:

```
<Notice> - all 32 packet processing threads, 4 management threads initialized, engine started
```

To something like:

```
<Notice> - all 32 packet processing (24 Capture, 8 Processing) threads, 4 management threads initialized, engine started
```

JoshuaLumb:
> So I would expect 5 W but I still get 8, changing the cpu value doesn't have any effect. So feel free to share your section.

From what I can tell the *-cpu-set cpu value doesn't dictate number of threads to spawn but rather where those threads should bind. Try this to get 5 worker threads:

cpu-affinity:
    . . .
    - worker-cpu-set:
        cpu: [ "all" ]
        mode: "exclusive"
        threads: 5

> I still think the table reflects a better approach but not 100% sure if it's the best approach. I struggle a bit with what would be the best and most reasonable expectation.

Is there an individual or group we can consult for input on this subject?

> I would also suggest that we change the output from:
> 
> ```
> <Notice> - all 32 packet processing threads, 4 management threads initialized, engine started
> ```
> 
> To something like:
> 
> ```
> <Notice> - all 32 packet processing (24 Capture, 8 Processing) threads, 4 management threads initialized, engine started
> ```

Agreed! I'll make sure to include that change when I create PR v2.

norg:
> > So I would expect 5 W but I still get 8, changing the cpu value doesn't have any effect. So feel free to share your section.
> 
> From what I can tell the *-cpu-set cpu value doesn't dictate number of threads to spawn but rather where those threads should bind. Try this to get 5 worker threads:
> 

My bad, I forgot to set the `threads:` correctly. So I can ack that it works as in your table.

> > I still think the table reflects a better approach but not 100% sure if it's the best approach. I struggle a bit with what would be the best and most reasonable expectation.
> 
> Is there an individual or group we can consult for input on this subject?
> 

Not really, it's us as the OISF Dev team to decide in the end but the whole community for feedback :)

src/tm-threads.c
```diff
@@ -2178,6 +2178,22 @@ void TmThreadsGetMinimalTimestamp(struct timeval *ts)
     SCLogDebug("ts->tv_sec %"PRIuMAX, (uintmax_t)ts->tv_sec);
 }
 
+uint16_t TmThreadsGetWorkerThreadMax()
+{
+    uint16_t ncpus = UtilCpuGetNumProcessorsOnline();
+    int thread_max = TmThreadGetNbThreads(WORKER_CPU_SET);
+    /* always create at least one thread */
+    if (thread_max == 0)
+            thread_max = ncpus * threading_detect_ratio;
```
victorjulien, line 2187: nit: indent looks off

JoshuaLumb:
Reopened [here](https://github.com.us3.cas.ms/OISF/suricata/pull/5192)

PR #4997, opened 2020-05-30

jlucovsky:
Continuation of #4991

Describe changes:
- Incorporate `offset` value when validating bsize value.

### DIFF
--- a/src/detect-bsize.c
+++ b/src/detect-bsize.c
@@ -65,6 +65,8 @@ void DetectBsizeRegister(void)
 #define DETECT_BSIZE_RA 2
 #define DETECT_BSIZE_EQ 3
 
+const char *bsize_mode_strings[] = { "<", ">", "<>", "="};
+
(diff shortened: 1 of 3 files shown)


src/detect-bsize.c
```diff
@@ -266,6 +268,68 @@ static DetectBsizeData *DetectBsizeParse (const char *str)
...
+        case DETECT_BSIZE_LT:
+            possible = value< bsz->lo;
+            break;
+    }
+    return possible;
+}
+
+static bool DetectBsizeCheckContent(const SigMatch *sm, const DetectBsizeData *bsz)
+{
+    uint16_t bytes_required;
```
victorjulien, line 293: if none of the members of the list are `content` the error message will use this uninitialized variable.

victorjulien: also: the error won't make sense at all in this case

jlucovsky:
Continued in #5014

PR #4999, opened 2020-05-31
Draft: add fileinfo to alert

jlucovsky:
Draft PR for [issue 2015](https://redmine.openinfosecfoundation.org/issues/2015)

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -415,6 +415,7 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
 
         MemBufferReset(aft->json_buffer);
 
+
         /* alert */
         AlertJsonHeader(json_output_ctx, p, pa, js, json_output_ctx->flags);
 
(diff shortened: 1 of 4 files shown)


src/output-json-alert.c
```diff
@@ -390,6 +390,95 @@ static void AlertAddPayload(AlertJsonOutputCtx *json_output_ctx, json_t *js, con
     }
 }
 
+static json_t *JsonBuildFileInfoRecord(const Packet *p, const File *ff)
```
victorjulien, line 393: Would like to avoid this duplication with `JsonBuildFileInfoRecord` from output-json-file.c

src/output-json-alert.c
```diff
@@ -493,6 +583,26 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
             }
         }
 
+        if (pa->s->file_flags & FILE_SIG_NEED_FILE) {
```
victorjulien, line 586: Wonder if this should do this unconditionally. If there are files, add them to the record.

src/output-json-alert.c
jlucovsky: I plan to refactor as much as possible. This was just a POC implementation.

src/output-json-alert.c
jlucovsky: Should we make fileinfo output in alert records configurable?

jasonish: I hesitate to say yes as it adds another toggle to the configuration.

Probably bundle it under metadata. If alert metadata is false, then don't include it. If true, then include it?

jlucovsky: Will use metadata config setting and then include all files, regardless of whether the sig needs them.

victorjulien: Sounds good.

PR #5001, opened 2020-06-01
Proto/2

jlucovsky:
Continuation of #4998

This PR improves the output of unknown protocol values. Values that were zero-padded will no longer have leading 0's, e.g., `006` becomes `6`.

For compatibility no output changes are made to the fast nor syslog outputs.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Remove zero-padding changes from alert, drop and log

PR #5006, opened 2020-06-02
output: Improve protocol output handling.

jlucovsky:
Continuation of #5001

This PR improves the output of unknown protocol values. Values that were zero-padded will no longer have leading 0's, e.g., `006` becomes `6`.

For compatibility no output changes are made to the fast nor syslog outputs.

Describe changes:
- Update commit message format.

### DIFF
--- a/src/alert-syslog.c
+++ b/src/alert-syslog.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2014 Open Information Security Foundation
+/* Copyright (C) 2007-2020 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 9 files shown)


src/alert-syslog.c
```diff
@@ -209,6 +209,13 @@ static TmEcode AlertSyslogIPv4(ThreadVars *tv, const Packet *p, void *data)
     if (p->alerts.cnt == 0)
         return TM_ECODE_OK;
 
+    char proto[16] = "";
+    if (SCProtoNameValid(IPV4_GET_IPPROTO(p))) {
+        strlcpy(proto, known_proto[IPV4_GET_IPPROTO(p)], sizeof(proto));
```
victorjulien, line 214: this seems like an unnecessary copy. We can use a ptr instead. For the non-SCProtoNameValid we can use a local string and point the same pointer to that

victorjulien:
Replaced by #5019

PR #5008, opened 2020-06-02
Enable rdp/sip by default; deprecate drop.log -- v1

jasonish:
Enable SIP by default: 
https://redmine.openinfosecfoundation.org/issues/3256

Enable RDP by default: 
https://redmine.openinfosecfoundation.org/issues/3255

Remove old style drop.log: 
https://redmine.openinfosecfoundation.org/issues/2381


victorjulien:
Merged as part of #5012, thanks!

PR #5014, opened 2020-06-03
detect/bsize: Validate against content buffer when available

jlucovsky:
Continuation of #4997

This PR adds additional validation when using the `bsize` keyword. If a one or more `content` keywords immediately precedes `bsize`, then the `bsize` value is checked against each to see if a match is possible using the operation (`=, <, >, <>`) and the value.

An error is raised if `bsize` value prevents a match, e.g., the content length exceeds the `bsize` value. The `bsize` operation and values are used to do the evaluation.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3682](https://redmine.openinfosecfoundation.org/issues/3682)

Describe changes:
- Address review comments

Companion [Suricata PR #233](https://github.com/OISF/suricata-verify/pull/233)

PR #5028, opened 2020-06-06
detect/bsize: Validate against content buffer when available

jlucovsky:
Continuation of #5014

Describe changes:
- Address review comments

### DIFF
--- a/doc/userguide/rules/payload-keywords.rst
+++ b/doc/userguide/rules/payload-keywords.rst
@@ -267,17 +267,42 @@ You can also use the negation (!) before isdataat.
 bsize
 -----
 
-With the bsize keyword, you can match on the length of the buffer. This adds precision to the content match, previously this could have been done with isdataat.
+With the ``bsize`` keyword, you can match on the length of the buffer. This adds
+precision to the content match, previously this could have been done with ``isdataat``.
+
+An optional operator can be specified; if no operator is present, the operator will
+default to '='. When a relational operator is used, e.g., '<', '>' or '<>' (range),
+the bsize value will be compared using the relational operator. Ranges are inclusive.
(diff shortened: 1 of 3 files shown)


catenacyber:
Could we use detect-engine-uint.c helper functions ?

jlucovsky:
> Could we use detect-engine-uint.c helper functions ?

I don't think that would be a benefit in this case. `DetectCheckValue` has different semantics than the match function in `detect-engine-uint.c`.

doc/userguide/rules/payload-keywords.rst
```diff
@@ -267,18 +267,43 @@ You can also use the negation (!) before isdataat.
...
+
+   alert dns any any -> any any (msg:"test bsize rule"; dns.query; content:"short"; bsize:<10; sid:124; rev:1;)
+
+.. container:: example-rule
+
+   alert dns any any -> any any (msg:"test bsize rule"; dns.query; content:"longer string"; bsize:>10; sid:125; rev:1;)
+
+.. container:: example-rule
+
+   alert dns any any -> any any (msg:"test bsize rule"; dns.query; content:"middle"; bsize:5<>15; sid:126; rev:1;)
```
inashivb, line 305: This should probably be `bsize:`6<>15` as length of "middle" is 6.

jlucovsky:
Continued in #5576

doc/userguide/rules/payload-keywords.rst
catenacyber: What does `6<>15` mean ? A range ?

jlucovsky: Yes -- a range.

PR #5030, opened 2020-06-06
ftp/eve: Convert to JsonBuilder

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3714](https://redmine.openinfosecfoundation.org/issues/3714)

Describe changes:
- Convert to JsonBuilder
- Remove unused helper function (no longer needed)

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2017 Open Information Security Foundation
+/* Copyright (C) 2007-2020 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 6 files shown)


src/output-json-ftp.c
```diff
@@ -60,41 +60,35 @@ typedef struct LogFTPLogThread_ {
...
     }
-
-    json_object_set_new(cjs, "command", json_string(tx->command_descriptor->command_name));
+    jb_set_string(jb, "command", tx->command_descriptor->command_name);
     uint32_t min_length = tx->command_descriptor->command_length + 1; /* command + space */
     if (tx->request_length > min_length) {
-        json_object_set_new(cjs, "command_data",
-                            JsonAddStringN((const char *)tx->request + min_length,
-                                           tx->request_length - min_length));
+        char *s = BytesToString(tx->request + min_length,
```
victorjulien, line 79: @jasonish iirc we talked about a direct `jb_set_bytes` or something to avoid an expensive operation like this?

src/output-json-ftp.c
```diff
@@ -107,40 +101,61 @@ static json_t *JsonFTPLogCommand(Flow *f, FTPTransaction *tx)
                 if (pos >= 3)  {
                     /* Gather the completion code if present */
                     if (isdigit(where[0]) && isdigit(where[1]) && isdigit(where[2])) {
-                        json_array_append_new(js_respcode_list,
-                                              JsonAddStringN((const char *)where, 3));
+                        char *s = BytesToString(where, 3);
+                        if (s != NULL) {
+                            jb_append_string(js_respcode_list, s);
```
victorjulien, line 106: the goal of the jsonbuilder transition is to avoid creating temporary objects like this, but instead "stream" the whole object into a single `jb`.

src/app-layer-ftp.c
```diff
@@ -1439,36 +1439,37 @@ uint16_t JsonGetNextLineFromBuffer(const char *buffer, const uint16_t len)
...
+        jb_set_string(jb, "filename", fname);
     }
     switch (ftp_state->command) {
         case FTP_COMMAND_STOR:
-            json_object_set_new(ftpd, "command", json_string("STOR"));
+            jb_set_string(jb, "command", "STOR");
             break;
         case FTP_COMMAND_RETR:
-            json_object_set_new(ftpd, "command", json_string("RETR"));
+            jb_set_string(jb, "command", "RETR");
```
victorjulien, line 1462: @jasonish I still think we can optimize these constant cases with some preproc magic to become a single string that we append

src/output-json-ftp.c
jlucovsky: `set_string_from_bytes` ... I'll update

src/output-json-ftp.c
jlucovsky: I modeled my implementation after that in `output-json-email-common.c`.

Note that I'm building 2 arrays as the response is processed and 2 arrays will be created during that. I'm not sure how the new interfaces support this?

src/output-json-ftp.c
jasonish: `jb_set_string_from_bytes` should work, just looks like it doesn't have an extern "C" wrapper yet (haven't needed it from C yet would be the reason).

src/output-json-ftp.c
jasonish: There isn't in some cases.  In DNS, to avoid looping throught the responses 2x, I had to use intermediate objects as well in `dns_log_json_answer`.

src/app-layer-ftp.c
jasonish: Yeah, need to look at this again.

src/output-json-ftp.c
jasonish: See this PR: https://github.com/OISF/suricata/pull/5040

Should be able to:
```
jb_append_string_from_bytes(js_respcode_list, where, 3);
```

jlucovsky:
Continued in #5049

PR #5031, opened 2020-06-06
ssh/eve: convert to jsonbuilder

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3708

Describe changes:
- convert ssh to jsonbuilder

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

I am not sure there are any S-V tests with SSH by the way.
Tested with https://github.com/OISF/suricata-verify/pull/176

### DIFF
--- a/rust/src/ssh/logger.rs
+++ b/rust/src/ssh/logger.rs
@@ -16,49 +16,40 @@
  */
 
 use super::ssh::SSHTransaction;
-use crate::json::*;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
 
-fn log_ssh(tx: &SSHTransaction) -> Option<Json> {
+fn log_ssh(tx: &SSHTransaction) -> Result<Option<JsonBuilder>, JsonError> {
     if tx.cli_hdr.protover.len() == 0 && tx.srv_hdr.protover.len() == 0 {
-        return None;
+        return Ok(None);
     }
-    let js = Json::object();
+    let mut js = JsonBuilder::new_object();
     if tx.cli_hdr.protover.len() > 0 {
-        let cjs = Json::object();
-        cjs.set_string_from_bytes(
-            "proto_version",
-            &tx.cli_hdr.protover,
-        );
+        let mut cjs = JsonBuilder::new_object();
+        cjs.set_string_from_bytes("proto_version", &tx.cli_hdr.protover)?;
         if tx.cli_hdr.swver.len() > 0 {
-            cjs.set_string_from_bytes(
-                "software_version",
-                &tx.cli_hdr.swver,
-            );
+            cjs.set_string_from_bytes("software_version", &tx.cli_hdr.swver)?;
         }
-        js.set("client", cjs);
+        cjs.close()?;
+        js.set_object("client", &cjs)?;
     }
     if tx.srv_hdr.protover.len() > 0 {
-        let sjs = Json::object();
-        sjs.set_string_from_bytes(
-            "proto_version",
-            &tx.srv_hdr.protover,
-        );
+        let mut sjs = JsonBuilder::new_object();
+        sjs.set_string_from_bytes("proto_version", &tx.srv_hdr.protover)?;
         if tx.srv_hdr.swver.len() > 0 {
-            sjs.set_string_from_bytes(
-                "software_version",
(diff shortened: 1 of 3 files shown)


src/output-json-alert.c
```diff
@@ -147,12 +147,10 @@ static void AlertJsonSsh(const Flow *f, JsonBuilder *js)
...
     if (ssh_state) {
         void *tx_ptr = rs_ssh_state_get_tx(ssh_state, 0);
-        json_t *tjs = rs_ssh_log_json(tx_ptr);
+        JsonBuilder *tjs = rs_ssh_log_json(tx_ptr);
         if (unlikely(tjs == NULL))
             return;
-
-        jb_set_jsont(js, "ssh", tjs);
-        json_decref(tjs);
+        jb_set_object(js, "ssh", tjs);
```
catenacyber, line 153: Should I close it afterwards ?

rust/src/ssh/logger.rs
```diff
@@ -16,49 +16,40 @@
...
     }
-    let js = Json::object();
+    let mut js = JsonBuilder::new_object();
     if tx.cli_hdr.protover.len() > 0 {
-        let cjs = Json::object();
-        cjs.set_string_from_bytes(
-            "proto_version",
-            &tx.cli_hdr.protover,
-        );
+        let mut cjs = JsonBuilder::new_object();
```
victorjulien, line 27: there is no need to create another object here, you can just `js.open_object("client")?;` and then add the client fields. Then use `js.close()?;` when the client section is done.

The goal is to stream it all into one `JsonBuilder` object so we avoid allocs and copies.

src/output-json-ssh.c
```diff
@@ -75,25 +75,26 @@ static int JsonSshLogger(ThreadVars *tv, void *thread_data, const Packet *p,
         return 0;
     }
 
-    json_t *js = CreateJSONHeader(p, LOG_DIR_FLOW, "ssh", NULL);
+    JsonBuilder *js = CreateEveHeaderWithTxId(p, LOG_DIR_FLOW, "ssh", NULL, tx_id);
```
victorjulien, line 78: we're creating a JsonBuilder here, so we should use only this one. The rust code should not create its own, but instead take this one as an argument and continue to add to it.

rust/src/ssh/logger.rs
```diff
@@ -16,49 +16,40 @@
  */
 
 use super::ssh::SSHTransaction;
-use crate::json::*;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
 
-fn log_ssh(tx: &SSHTransaction) -> Option<Json> {
+fn log_ssh(tx: &SSHTransaction) -> Result<Option<JsonBuilder>, JsonError> {
```
victorjulien, line 21: this should not return a JsonBuilder, but take a JsonBuilder as input that it will add its output to.

victorjulien:
Have a look at the flow of the nfs logger wrt how the jsonbuilder object is set up and passed around.

rust/src/ssh/logger.rs
catenacyber: ok

src/output-json-ssh.c
catenacyber: ok

rust/src/ssh/logger.rs
catenacyber: ok, but this induces adding empty json for some alerts (see next PR)

catenacyber:
nfs logger is still with jansson (file rust/src/nfs/log.rs)

I took a look at every rust parser using `jsonbuilder` and I saw none providing the possibility of logging nothing...

victorjulien:
> nfs logger is still with jansson (file rust/src/nfs/log.rs)
> 
> I took a look at every rust parser using `jsonbuilder` and I saw none providing the possibility of logging nothing...

Did you look at the correct git revision? This is in master:
https://github.com/OISF/suricata/commit/b3b5802c85c8e893fdf6ae144880836bff0a16ee

PR #5032, opened 2020-06-07
ssh/eve: convert to jsonbuilder

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3708

Describe changes:
- convert ssh to jsonbuilder

Modifies #5031 with comments taken into account

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

I am not sure there are any S-V tests with SSH by the way.
Tested with https://github.com/OISF/suricata-verify/pull/176

### DIFF
--- a/rust/src/ssh/logger.rs
+++ b/rust/src/ssh/logger.rs
@@ -16,49 +16,33 @@
  */
 
 use super::ssh::SSHTransaction;
-use crate::json::*;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
 
-fn log_ssh(tx: &SSHTransaction) -> Option<Json> {
+fn log_ssh(tx: &SSHTransaction, js: &mut JsonBuilder) -> Result<(), JsonError> {
     if tx.cli_hdr.protover.len() == 0 && tx.srv_hdr.protover.len() == 0 {
-        return None;
+        return Err(JsonError::InvalidState);
     }
-    let js = Json::object();
     if tx.cli_hdr.protover.len() > 0 {
-        let cjs = Json::object();
-        cjs.set_string_from_bytes(
-            "proto_version",
-            &tx.cli_hdr.protover,
-        );
+        js.open_object("client")?;
+        js.set_string_from_bytes("proto_version", &tx.cli_hdr.protover)?;
         if tx.cli_hdr.swver.len() > 0 {
-            cjs.set_string_from_bytes(
-                "software_version",
-                &tx.cli_hdr.swver,
-            );
+            js.set_string_from_bytes("software_version", &tx.cli_hdr.swver)?;
         }
-        js.set("client", cjs);
+        js.close()?;
     }
     if tx.srv_hdr.protover.len() > 0 {
-        let sjs = Json::object();
-        sjs.set_string_from_bytes(
-            "proto_version",
-            &tx.srv_hdr.protover,
-        );
+        js.open_object("server")?;
+        js.set_string_from_bytes("proto_version", &tx.srv_hdr.protover)?;
         if tx.srv_hdr.swver.len() > 0 {
-            sjs.set_string_from_bytes(
-                "software_version",
-                &tx.srv_hdr.swver,
-            );
+            js.set_string_from_bytes("software_version", &tx.srv_hdr.swver)?;
         }
-        js.set("server", sjs);
+        js.close()?;
     }
-    return Some(js);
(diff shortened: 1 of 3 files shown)


rust/src/ssh/logger.rs
```diff
@@ -16,49 +16,33 @@
...
 
 use super::ssh::SSHTransaction;
-use crate::json::*;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
 
-fn log_ssh(tx: &SSHTransaction) -> Option<Json> {
+fn log_ssh(tx: &SSHTransaction, js: &mut JsonBuilder) -> Result<(), JsonError> {
     if tx.cli_hdr.protover.len() == 0 && tx.srv_hdr.protover.len() == 0 {
-        return None;
+        return Err(JsonError::InvalidState);
```
catenacyber, line 23: Is this the way to go ?

src/output-json-alert.c
```diff
@@ -147,12 +147,10 @@ static void AlertJsonSsh(const Flow *f, JsonBuilder *js)
...
         void *tx_ptr = rs_ssh_state_get_tx(ssh_state, 0);
-        json_t *tjs = rs_ssh_log_json(tx_ptr);
-        if (unlikely(tjs == NULL))
-            return;
-
-        jb_set_jsont(js, "ssh", tjs);
-        json_decref(tjs);
+        jb_open_object(js, "ssh");
+        // log even if no data
+        rs_ssh_log_json(tx_ptr, js);
```
catenacyber, line 152: Return value is not tested
We may have `"ssh" : {}` in the json

src/output-json-ssh.c
```diff
@@ -75,25 +75,25 @@ static int JsonSshLogger(ThreadVars *tv, void *thread_data, const Packet *p,
...
-    json_t *tjs = rs_ssh_log_json(txptr);
-    if (unlikely(tjs == NULL)) {
-        free(js);
+    jb_open_object(js, "ssh");
+    if (!rs_ssh_log_json(txptr, js)) {
+        jb_free(js);
         return 0;
     }
-    json_object_set_new(js, "ssh", tjs);
+    if (jb_close(js)) {
```
catenacyber, line 92: When should we test the return value of `jb_close` ?
Sometimes, it is checked, and sometimes not...

src/output-json-alert.c
victorjulien: Look at the mark support https://github.com/OISF/suricata/commit/b3b5802c85c8e893fdf6ae144880836bff0a16ee#diff-e2a291d82d2cff36c51007fafaa79b86R515

catenacyber: Thanks

rust/src/ssh/logger.rs
jasonish: No, the JsonError should probably only be emitted from JsonBuilder functions.  InvalidState means you tried to "append_string" or something while not in array, etc.

src/output-json-ssh.c
jasonish: In Rust I'd say always, as its easy to do everywhere with '?'.  In C we're not checking it really.  Not for a good reason, but more for the reason that we didn't check libjansson responses either, all of which can fail in more or less the same place as a JsonBuilder.

The bigger questions I suppose is what to do on error? It will most likely be a memory allocation panic as we go out of our way to handle UTF8 encoding errors.  I suppose JsonError::Invalid state could happen as well, but thats likely a programming error (should add a debug validate on that - just inside jsonbuilder)

rust/src/ssh/logger.rs
catenacyber: ok, so changing the function prototype to return `Result<bool`

src/output-json-ssh.c
catenacyber: Ok, so removing the check for now

rust/src/ssh/logger.rs
jasonish: Only use bool if that will be actually used.. Otherwise `Result<(), JsonError>` is enough to signal success/error.

jasonish: Ah, yes.. It is used, I see.

PR #5036, opened 2020-06-08
Dcerpc multi tx/v2

inashivb:
- Multiple transaction support for DCERPC
- Logging support for DCERPC
- travis test for enable-debug-validation
- gh-checks test for enable-debug-validation
- fixes in tests as required by enable-debug-validation

S-V PR to go with this: https://github.com/OISF/suricata-verify/pull/245

Redmine issues:
https://redmine.openinfosecfoundation.org/issues/3693
https://redmine.openinfosecfoundation.org/issues/3694


rust/src/dcerpc/dcerpc_udp.rs
```diff
@@ -67,37 +69,48 @@ pub struct DCERPCUDPState {
...
-        self.response = Some(response);
+
+    fn evaluate_serial_no(&mut self) -> u16 {
+        let mut serial_hi: u8 = 0;
+        let mut serial_lo: u8 = 0;
+        if let Some(ref hdr) = &self.header {
+            serial_hi = hdr.serial_hi;
+            serial_lo = hdr.serial_lo;
+        }
+        let mut serial_no = Cursor::new(vec![serial_hi, serial_lo]);
```
victorjulien, line 100: What is a Cursor and what does it do here?

rust/src/dcerpc/log.rs
```diff
@@ -0,0 +1,45 @@
...
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use crate::dcerpc::dcerpc::*;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
+
+fn log_dcerpc_header(js: &mut JsonBuilder, tx: &DCERPCTransaction) -> Result<(), JsonError> {
```
victorjulien, line 21: think it would be good to add the DCERPC version in use.

rust/src/dcerpc/log.rs
```diff
@@ -0,0 +1,45 @@
...
+    if tx.req_done == true {
+        js.set_string("request", &dcerpc_type_string(tx.req_cmd))?;
+        js.open_object("req")?;
+        js.set_uint("opnum", tx.opnum as u64)?;
```
victorjulien, line 25: is `opnum` valid for all request types?

rust/src/dcerpc/dcerpc_udp.rs
inashivb: Cursor is used to allow reads or writes on in-memory buffers from what I understand. I picked it up from the example usage of [byteorder](https://docs.rs/byteorder/1.1.0/byteorder/trait.ReadBytesExt.html#method.read_u16) because I wanted to ensure that the data is read according to endianness.
But, there is a mistake. I have hard coded `LittleEndian` here so doing this all was useless. Changing it with the endianness received from header might be the way to go about it.

Do you think the same? Am I doing something wrong?

rust/src/dcerpc/log.rs
inashivb: No, just the standard DCERPC requests. For the rest, it'll be 0. Do you think it should be structured differently? Or maybe a check about the request type before adding opnum to the array?

victorjulien: In the SMB/DCE logic the opnum field is only added for REQUEST

rust/src/dcerpc/log.rs
victorjulien: I notice a few more things that SMB/DCE outputs, interfaces and more detailed bindack results

rust/src/dcerpc/log.rs
inashivb: OK. I'll match again with the SMB?DCE logic and add the missing parts.

inashivb:
Replaced with #5055

PR #5037, opened 2020-06-08
Feat/cond logging/v15

victorjulien:
Continuation of #4982:
- adding support for AppLayerTxData to all protocols
- add new AppLayerTxConfig for storing and requesting config (https://github.com/OISF/suricata/commit/26e044eec4b08d980e9818a54b51b62204962ffb, https://github.com/OISF/suricata/commit/d5f0f19661c122f8f0a769fa51782c0280807801)
- add ApplyTxConfig callback that can be used for 'special' protocols wrt how txs are handled https://github.com/OISF/suricata/commit/64fd37629324309a5f750f0b7cbced7b53b867d6


jasonish:
Was able to suppress DNS responses pretty easily using a rule like:
```
config dns any any -> any any (dns.query; content:"google.com"; endswith; config:logging disable, type tx, scope tx; sid:1; rev:1;)
```
What I used in the DNS parser was a HashSet on the state. When parsing a response I use the response DNS tx_id in a call to remove(), this will return true if an item was really removed, otherwise false.  Its not immediately clear how I should suppress the logging for the response though, but for test purposes I'm just not creating the transaction.

victorjulien:
> Was able to suppress DNS responses pretty easily using a rule like:
> 
> ```
> config dns any any -> any any (dns.query; content:"google.com"; endswith; config:logging disable, type tx, scope tx; sid:1; rev:1;)
> ```
> 
> What I used in the DNS parser was a HashSet on the state. When parsing a response I use the response DNS tx_id in a call to remove(), this will return true if an item was really removed, otherwise false. Its not immediately clear how I should suppress the logging for the response though, but for test purposes I'm just not creating the transaction.

The DNS tx has a `AppLayerTxData tx_data` field. Inside this you have a `AppLayerTxConfig` member. The `ApplyTxConfig` callback passes you a config request, so you the idea is that you apply this `AppLayerTxConfig` to your (new) tx. That way when the tx logging code runs it sees that the flag to not log is set.

Of course not creating a tx has the same effect, but from the top of my head I don't know if that will have other side effects, and also we may have other config options later that wouldn't work with that model. In general for this test I would like to see it done 'correctly' so we can validate the whole logic flow.

So what I think should be stored in the `HashSet` is (a copy of) the `AppLayerTxConfig` that was supplied by the `ApplyTxConfig` callback. Then when you fetch that from the response parser, you can create the tx normally but store the `AppLayerTxConfig` in it. Then it *should* work.

victorjulien:
Replaced by #5155

PR #5038, opened 2020-06-08

catenacyber:

Describe changes:
- adds new signature flag `SIG_FLAG_FILEDATA` for `file_data` keyword

with `if ((s->flags & SIG_FLAG_FILESTORE | SIG_FLAG_FILEDATA) || ) {`
Modifies #4908 with comments taken into account using `flags` and not `file_flags`

### DIFF
--- a/src/detect.h
+++ b/src/detect.h
@@ -250,6 +250,8 @@ typedef struct DetectPort_ {
 /** Info for Source and Target identification */
 #define SIG_FLAG_DEST_IS_TARGET         BIT_U32(26)
 
+#define SIG_FLAG_FILEDATA               BIT_U32(27) /**< signature has filedata keyword */
+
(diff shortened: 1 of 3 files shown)


src/detect.h
```diff
@@ -250,6 +250,8 @@ typedef struct DetectPort_ {
 /** Info for Source and Target identification */
 #define SIG_FLAG_DEST_IS_TARGET         BIT_U32(26)
 
+#define SIG_FLAG_FILEDATA               BIT_U32(27) /**< signature has filedata keyword */
```
victorjulien, line 253: Like I said in the last PR, I would like this to be a INIT flag (for Signature::init_data::flags) as we only use it during init. I want to preserve the runtime signature flag space.

catenacyber: Ah. I was astonished indeed that signature flag space seemed to cheap.
I did not know about this `init_flags`

PR #5043, opened 2020-06-08

catenacyber:

Describe changes:
- adds new signature flag `SIG_FLAG_FILEDATA` for `file_data` keyword

with `if ((s->flags & SIG_FLAG_FILESTORE | SIG_FLAG_FILEDATA) || ) {`

Modifies #5038 with comments taken into account using `init_flags` and not `flags`


victorjulien:
Merged in #5059, thanks!

PR #5056, opened 2020-06-09
smb/eve: convert to jsonbuilder

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3712

### DIFF
--- a/rust/src/smb/log.rs
+++ b/rust/src/smb/log.rs
@@ -17,7 +17,7 @@
 
 use std::str;
 use std::string::String;
-use crate::json::*;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
 use crate::smb::smb::*;
 use crate::smb::smb1::*;
 use crate::smb::smb2::*;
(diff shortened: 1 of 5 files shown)


rust/src/smb/log.rs
```diff
@@ -194,162 +195,164 @@ fn smb_common_header(state: &SMBState, tx: &SMBTransaction) -> Json
...
+            jsb.set_uint("size", x.size)?;
 
             let gs = fuid_to_string(&x.guid);
-            js.set_string("fuid", &gs);
+            jsb.set_string("fuid", &gs)?;
         },
         Some(SMBTransactionTypeData::NEGOTIATE(ref x)) => {
             if x.smb_ver == 1 {
-                let jsa = Json::array();
+                let mut jsa = JsonBuilder::new_array();
```
victorjulien, line 233: @jasonish does this lead to a mem alloc? If so can we avoid it? Really all we need is to add a `"name":[ ...` and then later a `]`?

jasonish: Yes it would.. Can use open_array here followed by a close.

inashivb: Thanks, Jason!

inashivb:
Replaced with #5068

PR #5058, opened 2020-06-09
detect-dsize: Add ! operator for dsize matching v2

JoshuaLumb:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2538

Describe changes:
- Addressing the first request from the above ticket: added ! operator to dsize keyword allowing a user to match != values (e.g. dsize:!100;)
- Unit tests added for parsing and matching with ! operator.
- Requires some discussion: There is another approach to this that allows the combination of multiple dsize values and operators that satisfies the second request in the ticket i.e. "dsize:![100,150];". The approach involves changing to flags instead of a dsize mode and adding checks that could possibly reduce performance in detect.c. Is this something the Suricata devs would be interested in seeing?
- Updated to address typos and unit test macros.

### DIFF
--- a/src/detect-dsize.c
+++ b/src/detect-dsize.c
@@ -44,9 +44,9 @@
 #include "util-profiling.h"
 
 /**
- *  dsize:[<>]<0-65535>[<><0-65535>];
+ *  dsize:[<>!]<0-65535>[<><0-65535>];
  */
-#define PARSE_REGEX "^\\s*(<|>)?\\s*([0-9]{1,5})\\s*(?:(<>)\\s*([0-9]{1,5}))?\\s*$"
+#define PARSE_REGEX "^\\s*(<|>|!)?\\s*([0-9]{1,5})\\s*(?:(<>)\\s*([0-9]{1,5}))?\\s*$"
 static DetectParseRegex parse_regex;
 
 static int DetectDsizeMatch (DetectEngineThreadCtx *, Packet *,
@@ -82,6 +82,8 @@ DsizeMatch(const uint16_t psize, const uint8_t mode,
             const uint16_t dsize, const uint16_t dsize2)
 {
     if (mode == DETECTDSIZE_EQ && dsize == psize)
+        return 1;
+    else if (mode == DETECTDSIZE_NE && dsize != psize)
         return 1;
     else if (mode == DETECTDSIZE_LT && psize < dsize)
         return 1;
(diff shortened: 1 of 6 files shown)


src/detect.c
```diff
@@ -766,9 +767,11 @@ static inline void DetectRulePacketRules(
 
         if (unlikely(sflags & SIG_FLAG_DSIZE)) {
             if (likely(p->payload_len < s->dsize_low || p->payload_len > s->dsize_high)) {
-                SCLogDebug("kicked out as p->payload_len %u, dsize low %u, hi %u",
-                        p->payload_len, s->dsize_low, s->dsize_high);
-                goto next;
+                if (!(s->dsize_mode == DETECTDSIZE_NE)) {
```
victorjulien, line 770: could we support the NE operator here?

Think it could be cleaner to move the check into a small inline util func.

src/detect-dsize.c
```diff
@@ -83,6 +83,8 @@ DsizeMatch(const uint16_t psize, const uint8_t mode,
 {
     if (mode == DETECTDSIZE_EQ && dsize == psize)
         return 1;
+    else if (mode == DETECTDSIZE_NE && dsize != psize)
```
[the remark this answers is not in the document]
JoshuaLumb: Good call, thanks!

src/detect.c
JoshuaLumb: Something along the lines of
`static inline bool SigDsizeModeNotEqual(const Signature *s) {
    return s->dsize_mode == DETECTDSIZE_NE;
}`
in detect-dsize.h perhaps?

victorjulien: I was thinking about a bit wider scope of the inline func:
```c
static inline bool SigDsizePrefilter(...)
{
        if (unlikely(sflags & SIG_FLAG_DSIZE)) {
            if (likely(p->payload_len < s->dsize_low || p->payload_len > s->dsize_high)) {
                SCLogDebug("kicked out as p->payload_len %u, dsize low %u, hi %u",
                        p->payload_len, s->dsize_low, s->dsize_high);
                return false
            }
        }
        return true;
}
```
But then with the `NE` support added in as well.

JoshuaLumb: That makes way more sense. :)

JoshuaLumb:
[Reopened as v3](https://github.com/OISF/suricata/pull/5154)

PR #5061, opened 2020-06-09
Fix 32bit pcap counters wrap-around

roligugus:
Fixes issue 2845.
pcap_stats is based on 32bit counters and given a big enough throughput
will overflow them. This was reported by people using Myricom cards which
should only be a happenstance. The problem exists for all pcap-based
interfaces.
Let's use internal 64bit counters that drag along the pcap_stats and
handle pcap_stats wrap-around as we update the 64bit stats "often enough"
before the pcap_stats can wrap around twice.

https://redmine.openinfosecfoundation.org/issues/2845

Describe changes:
- Keep internal pcap stats as 64 bits around and take pcap_stats() one-time roll-over into account when adding current stats. Use the 64 bit stats to "essentially store the last value" to detect wrap-around and, once updated, use them also for printing the actual stats.
- The current pcap stats are polled every second which assures that we don't roll over twice. At 10G with min packets (60/84 bytes with overhead), i.e. ~15million packets/s, we've got around 290s before the 32bit stats are filled up.
- Added unit testing for the roll-over calculation in UpdatePcapStatsValue64(), plus one test for the 64 bit stats.
- Also fixed a potential div/0 in "drop percent" calculation in case there is no traffic
- No documentation changes unless I missed something

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/source-pcap.c
+++ b/src/source-pcap.c
@@ -40,12 +40,30 @@
 #include "util-optimize.h"
 #include "util-checksum.h"
 #include "util-ioctl.h"
+#include "util-unittest.h"
 #include "tmqh-packetpool.h"
 
 #define PCAP_STATE_DOWN 0
 #define PCAP_STATE_UP 1
 
 #define PCAP_RECONNECT_TIMEOUT 500000
+
+/**
+ * \brief 64bit pcap stats counters.
+ *
+ * libpcap only supports 32bit counters. They will eventually wrap around.
+ *
+ * Keep track of libpcap counters as 64bit counters to keep on counting even
+ * if libpcap's 32bit counters wrap around.
+ * Requires pcap_stats() to be called before 32bit stats wrap around twice,
+ * which we do.
+ */
+typedef struct PcapStats64_
+{
+    uint64_t ps_recv;
+    uint64_t ps_drop;
+    uint64_t ps_ifdrop;
+} PcapStats64;
 
 /**
  * \brief Structure to hold thread specific variables.
@@ -62,6 +80,7 @@ typedef struct PcapThreadVars_
     const char *bpf_filter;
 
     time_t last_stats_dump;
+    PcapStats64 last_stats64;
 
     /* data link type for the thread */
     int datalink;
@@ -99,12 +118,14 @@ static TmEcode DecodePcapThreadInit(ThreadVars *, const void *, void **);
 static TmEcode DecodePcapThreadDeinit(ThreadVars *tv, void *data);
 static TmEcode DecodePcap(ThreadVars *, Packet *, void *);
 
+void SourcePcapRegisterTests(void);
+
 /** protect pcap_compile and pcap_setfilter, as they are not thread safe:
  *  http://seclists.org/tcpdump/2009/q1/62 */
 static SCMutex pcap_bpf_compile_lock = SCMUTEX_INITIALIZER;
 
 /**
- * \brief Registration Function for RecievePcap.
+ * \brief Registration Function for ReceivePcap.
  */
 void TmModuleReceivePcapRegister (void)
 {
@@ -115,6 +136,7 @@ void TmModuleReceivePcapRegister (void)
     tmm_modules[TMM_RECEIVEPCAP].ThreadExitPrintStats = ReceivePcapThreadExitStats;
     tmm_modules[TMM_RECEIVEPCAP].cap_flags = SC_CAP_NET_RAW;
     tmm_modules[TMM_RECEIVEPCAP].flags = TM_FLAG_RECEIVE_TM;
+    tmm_modules[TMM_RECEIVEPCAP].RegisterTests = SourcePcapRegisterTests;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/source-pcap.c
```diff
@@ -522,11 +580,17 @@ static void ReceivePcapThreadExitStats(ThreadVars *tv, void *data)
          * Note: ps_recv includes dropped packets and should be considered total.
          * Unless we start to look at ps_ifdrop which isn't supported everywhere.
          */
-        SCLogInfo("(%s) Pcap Total:%" PRIu64 " Recv:%" PRIu64 " Drop:%" PRIu64 " (%02.1f%%).",
-                tv->name, (uint64_t)pcap_s.ps_recv,
-                (uint64_t)pcap_s.ps_recv - (uint64_t)pcap_s.ps_drop,
-                (uint64_t)pcap_s.ps_drop,
-                (((float)(uint64_t)pcap_s.ps_drop)/(float)(uint64_t)pcap_s.ps_recv)*100);
+        UpdatePcapStats64(&ptv->last_stats64, &pcap_s);
+        float drop_percent = likely(ptv->last_stats64.ps_recv > 0)
```
roligugus, line 584: Handle div/0 if ptv->last_stats64.ps_recv == 0.

src/source-pcap.c
```diff
@@ -641,3 +705,223 @@ void PcapTranslateIPToDevice(char *pcap_dev, size_t len)
...
+}
+
+static int UpdatePcapStatsValue64NoChange01(void)
+{
+    /*
+     * No change in counter values.
+     * Last count is within first 32bit range, i.e. same as pcap_stat range.
+     */
+    uint64_t last[] = { 0, 12345, (uint64_t)UINT32_MAX };
+    u_int current[] = { 0, 12345, UINT_MAX };
```
roligugus, line 727: Am using u_int type just to be in line with the actual pcap_stats definition. uint32_t would be the same.

src/source-pcap.c
```diff
@@ -129,14 +151,50 @@ void TmModuleDecodePcapRegister (void)
...
+
 static inline void PcapDumpCounters(PcapThreadVars *ptv)
 {
     struct pcap_stat pcap_s;
     if (likely((pcap_stats(ptv->pcap_handle, &pcap_s) >= 0))) {
-        StatsSetUI64(ptv->tv, ptv->capture_kernel_packets, pcap_s.ps_recv);
-        StatsSetUI64(ptv->tv, ptv->capture_kernel_drops, pcap_s.ps_drop);
-        (void) SC_ATOMIC_SET(ptv->livedev->drop, pcap_s.ps_drop);
-        StatsSetUI64(ptv->tv, ptv->capture_kernel_ifdrops, pcap_s.ps_ifdrop);
+        UpdatePcapStats64(&ptv->last_stats64, &pcap_s);
```
roligugus, line 190: Changes look bigger than they are. Getting 64bit stats and using them to StatsSet().

New code wraps at 80 chars, which my clang-format settings enforce, making for "bigger diff".

src/source-pcap.c
```diff
@@ -62,6 +80,7 @@ typedef struct PcapThreadVars_
     const char *bpf_filter;
 
     time_t last_stats_dump;
+    PcapStats64 last_stats64;
```
roligugus, line 83: There could probably be a better location cache-line wise. Maybe closer to "ThreadVars *tv" as they are used in the same code?

src/source-pcap.c
```diff
@@ -641,3 +705,223 @@ void PcapTranslateIPToDevice(char *pcap_dev, size_t len)
...
+{
+    /*
+     * No change in counter values.
+     * Last count is outside 32bits range.
+     */
+    uint64_t last[] = { (2ull << 32) + 0, (3ull << 32) + 12345,
+        (3ull << 32) + (uint64_t)UINT32_MAX,
+        ((uint64_t)UINT32_MAX << 32) + (uint64_t)UINT32_MAX };
+    u_int current[] = { 0, 12345, UINT_MAX, UINT_MAX };
+    FAIL_IF_NOT(ARRAY_SIZE(last) == ARRAY_SIZE(current));
```
roligugus, line 750: Doing separate last/current arrays is kind of messy to read.

An alternative would be:
```
/* Structured test data to make it easier on my eyes */
typedef struct TestData_
{
    uint64_t last; /* internal 64bit counter to drag along */
    u_int current; /* 32bit pcap stat */
} TestData;

static int UpdatePcapStatsValue64NoChange02(void)
{
    /*
     * No change in counter values.
     * Last count is outside 32bits range.
     */
    TestData data[] = {{.last = (2ull << 32) + 0, .current = 0},
            {.last = (3ull << 32) + 12345, .current = 12345},
            {.last = (3ull << 32) + (uint64_t)UINT32_MAX, .current = UINT_MAX},
            {.last = UINT64_MAX, .current = UINT_MAX}};

    for (size_t i = 0; i < ARRAY_SIZE(data); ++i) {
        uint32_t upper = Upper32(data[i].last);
        FAIL_IF_NOT(Lower32(data[i].last) == data[i].current);

        UpdatePcapStatsValue64(&data[i].last, data[i].current);
        FAIL_IF_NOT(Lower32(data[i].last) == data[i].current);
        FAIL_IF_NOT(Upper32(data[i].last) == upper);
    }

    PASS;
}

```

src/source-pcap.c
```diff
@@ -129,14 +151,50 @@ void TmModuleDecodePcapRegister (void)
     tmm_modules[TMM_DECODEPCAP].flags = TM_FLAG_DECODE_TM;
 }
 
+/**
+ * \brief Update 64 bit |last| value from |current32| value taking one
+ * wrap-around into account.
+ */
+static inline void UpdatePcapStatsValue64(uint64_t *last, uint32_t current32)
+{
+    uint32_t last32 = *last; /* slice lower 32bits */
```
victorjulien, line 160: I don't know enough about defined/undefined behavior to judge this. Would it make sense to first do a `uint64_t`with a mask that just leaves the lower 32 bits? Then downcast it?

src/source-pcap.c
```diff
@@ -641,3 +705,223 @@ void PcapTranslateIPToDevice(char *pcap_dev, size_t len)
 
     pcap_freealldevs(alldevsp);
 }
+
+#ifdef UNITTESTS
```
victorjulien, line 709: for new tests we put them in src/tests/source-pcap.c and include the file directly. See `stream-tcp-list.c tests/stream-tcp-list.c` as an example.

src/source-pcap.c
roligugus: unsigned to unsigned conversion (across different sizes) is well defined in the C standard.

As ```last``` is an uint64_t, it simply slices, i.e. it essentially "masks and downcasts" by default.

src/source-pcap.c
roligugus: Thanks for that! I copied the wrong example. ;) Will fix and open a new version.

src/source-pcap.c
roligugus: FYI, as for what integer conversion is defined, or undefined, or implementation dependent... One of the better practical explanations is: https://unspecified.wordpress.com/2011/08/08/integer-conversions-in-c/

You can always get into more detail through the CERT pages:  https://wiki.sei.cmu.edu/confluence/display/c/INT02-C.+Understand+integer+conversion+rules or the C standard. This allows you to go nuts about promotion, ranks, etc if you fancy that. ;)

roligugus:
Opened v2 to address review comments, see #5137

PR #5064, opened 2020-06-11
ikev1: add ikev1 parser

fhonza:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3767](https://redmine.openinfosecfoundation.org/issues/3767)

Describe changes:
- Add an IKEv1 parser to suricata with keywords for signature writing. Please see the included documentation (EVE-JSON fields, keywords) for more information.

Additional notes:
Alerts were created too often by the parser. After some debugging it turned out that the function tx_iterator() in ikev1.rs (from the template), in the end called from detect.c returned "previous" transactions. After changing this to return only the current transaction the number of alerts seemed correct. I was not able to understand the reason behind that, just in case this can lead to unexpected behaviour :)
It's marked with 'todo: ask oisf'. Thanks.


victorjulien:
@chifflier since you did v2, any interesting in reviewing this v1 implementation?

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+
+pub fn parse_security_association(
+    i: &[u8],
+    length: u16,
+) -> IResult<&[u8], SecurityAssociationPayload> {
+    do_parse!(
+        i,
+        domain_of_interpretation: be_u32 >>
+        situation: cond!(domain_of_interpretation == 1, take!(4)) >>
+        data: cond!(domain_of_interpretation == 1, take!(length - 8)) >>
```
chifflier, line 298: You must test `length >= 8` before doing that, or you may cause an integer underflow here.

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+        (
+            SecurityAssociationPayload {
+                domain_of_interpretation: domain_of_interpretation,
```
chifflier, line 301: (minor)
just use `domain_of_interpretation,` here

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+pub fn parse_key_exchange(
+    i: &[u8],
+    length: u16,
+) -> IResult<&[u8], KeyExchangePayload> {
+    do_parse!(
+        i,
+        key_exchange_data: take!(length) >>
+        (
+            KeyExchangePayload {
+                key_exchange_data: key_exchange_data
```
chifflier, line 318: (minor)
`KeyExchangePayload { key_exchange_data }`

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+    length: u16,
+) -> IResult<&[u8], ProposalPayload> {
+    do_parse!(
+        i,
+        proposal_number: be_u8 >>
+        proposal_type: be_u8 >>
+        spi_size: be_u8 >>
+        number_transforms: be_u8 >>
+        spi: take!(spi_size) >>
+        data: take!((length - 4) - spi_size as u16) >>
```
chifflier, line 335: Integer underflow if `length < 4`

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+
+pub fn parse_transform(
+    i: &[u8],
+    length: u16,
+) -> IResult<&[u8], TransformPayload> {
+    do_parse!(
+        i,
+        transform_number: be_u8 >>
+        transform_type: be_u8 >> be_u16 >>
+        sa_attributes: take!(length - 4) >>
```
chifflier, line 357: Integer underflow if `length < 4`

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+}
+
+pub fn parse_transform(
+    i: &[u8],
+    length: u16,
+) -> IResult<&[u8], TransformPayload> {
+    do_parse!(
+        i,
+        transform_number: be_u8 >>
+        transform_type: be_u8 >> be_u16 >>
```
chifflier, line 356: That `be_u16 >>` is badly indented, and easily mistaken (is the field really ignored?)

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+            }
+        )
+    )
+}
+
+pub fn parse_vendor_id(
+    i: &[u8],
+    length: u16,
+) -> IResult<&[u8], VendorPayload> {
+    do_parse!(
```
chifflier, line 372: better written as `map!(i, take!(length), VendorPayload)` imho

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+        vendor_id: take!(length) >>
+        (
+            VendorPayload {
+                vendor_id
+            }
+        )
+    )
+}
+
+named! { pub parse_sa_attribute<&[u8], Vec<SaAttribute>>,
```
chifflier, line 383: This function is very long and hard to review.
Hint: create subfunctions, like `AttributeType::from(format.1)`  and others to shorten it

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+            )
+        )
+    )
+}
+
+pub fn parse_nonce(
+    i: &[u8],
+    length: u16,
+) -> IResult<&[u8], NoncePayload> {
+    do_parse!(
```
chifflier, line 521: `map!(i, take!(length), NoncePayload)`

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+
+
+named! { pub parse_ikev1_payload_list<&[u8], Vec<IsakmpPayload>>,
+    many0!(
+        complete!(
+            do_parse!(
+                next_payload: be_u8 >>
+                reserved: be_u8 >>
+                payload_length: be_u16 >>
+                data: take!(payload_length - 4) >>
```
chifflier, line 540: Integer underflow if `payload_length < 4`

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+                (
+                    IsakmpPayload {
+                        payload_header: IsakmpPayloadHeader {
+                            next_payload: next_payload,
```
chifflier, line 544: Please use Field Init Shorthand (https://doc.rust-lang.org/book/ch05-01-defining-structs.html#using-the-field-init-shorthand-when-variables-and-fields-have-the-same-name)

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+                                            vendor_ids,
+                                            payload_types
+                                        ) {
+                                            return Err(())
+                                        }
+                                        cur_payload_type = isakmp_payload.payload_header.next_payload;
+                                    }
+                                },
+                                Err(nom::Err::Incomplete(_)) => {
+                                    return Err(())
```
chifflier, line 636: (this line and following)
add `ScLogDebug` calls to trace errors?

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+    transforms: &mut Vec<Vec<SaAttribute>>,
+    vendor_ids: &mut HashSet<String>,
+    payload_types: &mut HashSet<u8>) -> Result<(), ()>
+{
+    payload_types.insert(payload_type);
+
+    let element = num::FromPrimitive::from_u8(payload_type);
+    match element {
+        Some(IsakmpPayloadType::SecurityAssociation) => {
+            match parse_security_association(data, data_length) {
```
chifflier, line 604: I also suggest splitting some cases into subfunctions (`SecurityAssociation` and `Proposal`)

Do not mark subfunctions as `pub`, and the compiler will choose to inline them or not (rust is quite aggressive and usually inlines many functions)

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,732 @@
...
+                    transforms.push(attribute_list);
+                }
+            }
+            Ok(())
+        }
+        Some(IsakmpPayloadType::KeyExchange) => {
+            let res = parse_key_exchange(data, data_length);
+            if let Ok((_rem, payload)) = res {
+                if key_exchange.is_empty() {
+                    // key_exchange data should be empty, there should be only one...
```
chifflier, line 706: emit a log/warning?

rust/src/ikev1/detect.rs
```diff
@@ -0,0 +1,284 @@
...
+#[no_mangle]
+pub extern "C" fn rs_ikev1_state_get_exch_type(
+    tx: &mut IKEV1Transaction,
+    exch_type: *mut u32,
+) -> u8 {
+    if let Some(r) = tx.exchange_type {
+        unsafe{
+            *exch_type = r as u32;
+        }
+        return 1;
```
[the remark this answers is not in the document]
fhonza: first thanks for your review!
While implementing I've checked rfb/detect.rs and smb/detect.rs to get examples how to do it and both had this "return 1 on success" behaviour.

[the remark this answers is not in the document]
chifflier: you're right, it seems to be used at many places. I was just checking for consistency
Please ignore this comment :)

rust/src/ikev1/detect.rs
```diff
@@ -0,0 +1,284 @@
...
+pub extern "C" fn rs_ikev1_state_get_client_nonce(
+    state: &mut IKEV1State,
+    buffer: *mut *const u8,
+    buffer_len: *mut u32,
+) -> u8 {
+    if !state.client_nonce.is_empty() {
+        let p = &state.client_nonce;
+        if p.len() > 0 {
+            unsafe {
+                *buffer = p.as_ptr();
```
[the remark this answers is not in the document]
fhonza: Is your fear here that the content is changed from rust or from c or both?

And stupid question for implementation: wouldn't it be necessary to create a second variable on the heap and to copy the content over to survive the return out of the function? In this case it must be ensured that the allocated memory is freed, perhaps best in rust which would require a second rust function called when the variable is not used any longer from C?
Which would require a bit debugging there finding the correct location.
It's perhaps a bit error-prone to always remember that behaviour?
I've checked the other */detect.rs files but found no example.

In C an InspectionBuffer-struct is filled with a pointer to that variable which is returned to the more "inner" code of suricata, not sure where then best call the delete/rust_func.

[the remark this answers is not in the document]
chifflier: > Is your fear here that the content is changed from rust or from c or both?
> 

Unless you use the `Pin` trait in rust, a variable is not guaranteed to be at the same location, for ex if you update it. So, such affectation is correct if you check it is done once only (regardless of protocol correctness, you can have multiple
One possible scenario would be:
- first payload arrive, `client_nonce` is set
- C acquires a pointer to this nonce
- another packet with payload arrives, `client_nonce` is updated => C pointer now points to freed memory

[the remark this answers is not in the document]
chifflier: > And stupid question for implementation: wouldn't it be necessary to create a second variable on the heap and to copy the content over to survive the return out of the function? In this case it must be ensured that the allocated memory is freed, perhaps best in rust which would require a second rust function called when the variable is not used any longer from C?

That's what I was suggesting. If you look at `rust/src/common.rs`, you'll see the same for `CString` type.
Since here you may have a buffer, you cannot use the same types, but the logic is similar.
I think something like `return Box::into_raw(h) as *mut c_void;` is required to send a Box to C, and `{ Box::from_raw(ptr); }` to build the box (and free it immediatly).

> Which would require a bit debugging there finding the correct location.

What do you mean by "location"? Where to free the buffer?
It would be when you free the container C structure.
You can grep for `rs_cstring_free` to see an example.

> It's perhaps a bit error-prone to always remember that behaviour?

Indeed, you have to be very careful to do this correctly, and never forger to call functions.
That said, if I'm correct the pointer that was returned was not safe and could segfault.
`valgrind` can help a lot (also checks for memory leaks)

> I've checked the other */detect.rs files but found no example.
> 
> In C an InspectionBuffer-struct is filled with a pointer to that variable which is returned to the more "inner" code of suricata, not sure where then best call the delete/rust_func.

If you grep for `rs_krb5_tx_get_cname`, I did that for Kerberos cname matching. However, the difference in Kerberos is that the returned pointer is part of the _transaction_, and so never changes (the transaction is assigned only once). This is different for the state, which can be updated if I'm correct.
The transaction memory will be freed, so there is no leak.

So, the TL;DR is:
- if there is no way to reassign/change/free the value from Rust, you're fine (for ex. a transaction)
- or, you can create a pair of functions to allocate+copy / free memory in rust and call them from C
- or, you can allocate from C and call rust to copy data in the allocated buffer
- maybe other solutions ...

I hope it is clear!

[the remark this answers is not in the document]
fhonza: Thanks for your fast answer! i'll check your proposals!
Perhaps it's indeed the best to store more stuff in transactions. I had some other problem with alerts occuring more often than i expected, perhaps the problem was from the beginning that they are persisted in the "longer living" state.

rust/src/ikev1/ikev1.rs
```diff
@@ -0,0 +1,606 @@
...
+
+    logged: LoggerFlags,
+    de_state: Option<*mut core::DetectEngineState>,
+    events: *mut core::AppLayerDecoderEvents,
+    detect_flags: applayer::TxDetectFlags,
+}
+
+impl IKEV1Transaction {
+    pub fn new() -> IKEV1Transaction {
+        IKEV1Transaction {
```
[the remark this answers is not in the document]
fhonza: nice, didn't know the Default impl, for the state this works fine but the transaction requires applayer::LoggerFlags and core::AppLayerDecoderEvents. For the LoggerFlags this could be derived/implemented but with the AppLayerDecoderEvents i'm not sure. (defined as an empty enum)

rust/src/ikev1/ikev1.rs
```diff
@@ -0,0 +1,606 @@
...
+            let tx = &self.transactions[index];
+            if tx.tx_id < min_tx_id + 1 {
+                index += 1;
+                continue;
+            }
+            *state = index as u64;
+
+            // return Some((tx, tx.tx_id - 1, (len - index) > 1)); <- original
+
+            // todo: ask oisf next 4 lines! only return last transaction in iterator? seems to return correct number of alerts
```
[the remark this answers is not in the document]
fhonza: forget what i've written here, with changing logic to use the "immutable" transaction holding the values for alerting instead of the state the counts look much better now without this hack.

rust/src/ikev1/parser.rs
fhonza: yes, according to the rfc 2408: 
RESERVED2 (2 octets) - Unused, set to 0.

PR #5070, opened 2020-06-13
ftp/eve: Convert to JsonBuilder

jlucovsky:
Continuation of #5049

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3714](https://redmine.openinfosecfoundation.org/issues/3714)

Describe changes:
- Address review comments

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5071, opened 2020-06-14
ftp/eve: Convert to JsonBuilder

jlucovsky:
Continuation of #5070

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3714](https://redmine.openinfosecfoundation.org/issues/3714)

Describe changes:
- Address review comments -- file name length fix
- Change prefix of FTP helper functions to "Eve"

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/output-json-ftp.c
+++ b/src/output-json-ftp.c
@@ -59,7 +59,7 @@ typedef struct LogFTPLogThread_ {
     MemBuffer          *buffer;
 } LogFTPLogThread;
 
-static void JsonFTPLogCommand(Flow *f, FTPTransaction *tx, JsonBuilder *jb)
+static void EveFTPLogCommand(Flow *f, FTPTransaction *tx, JsonBuilder *jb)
 {
     /* Preallocate array objects to simplify failure case */
     JsonBuilder *js_resplist = NULL;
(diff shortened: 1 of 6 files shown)


src/output-json-ftp.c
```diff
@@ -107,40 +96,62 @@ static json_t *JsonFTPLogCommand(Flow *f, FTPTransaction *tx)
                 if (pos >= 3)  {
                     /* Gather the completion code if present */
                     if (isdigit(where[0]) && isdigit(where[1]) && isdigit(where[2])) {
-                        json_array_append_new(js_respcode_list,
-                                              JsonAddStringN((const char *)where, 3));
+                        jb_append_string_from_bytes(js_respcode_list, (const uint8_t *)where, 3);
```
victorjulien, line 99: So if I understand correctly this loop adds items to 2 arrays, which is why we have 2 separate array objects, right? If so, I think this could be optimized to use just one separate object. The 'first' array could just be streamed into the main `jb`. The 2nd array would then still need to be a separate array that is append when the loop finishes. Make sense?

jlucovsky: Good suggestion. Perhaps we should do this in other places?

jlucovsky:
Continued in #5079

PR #5074, opened 2020-06-15
Example DNS conditional logging - v1

jasonish:
Example of DNS conditional logging.

A ConfigTracker is is used with a map keyed by DNS request ID. 
The queue is only there to enforce a max size and evict entries
in order of creation. I think this needs more thought,
but serves the purpose of proving the framework around
conditional logging works.

Example rule:
```
config dns any any -> any any (dns.query; content:"google.com"; endswith; config: logging disable, type tx, scope tx; sid:1; rev:1;)
```


rust/src/dns/dns.rs
```diff
@@ -330,6 +362,8 @@ pub struct DNSState {
     pub request_buffer: Vec<u8>,
     pub response_buffer: Vec<u8>,
 
+    config: ConfigTracker,
```
victorjulien, line 365: I think this looks good. Just wonder if it would make sense to make this an `Option<ConfigTracker>` as we won't be using it in the common case most likely.

jasonish: Make sense.  This change is done.

victorjulien:
Replaced by #5155

PR #5075, opened 2020-06-15
[Draft] Prototype approach for issue 3616

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3616](https://redmine.openinfosecfoundation.org/issues/3616)

Describe changes:
- Defer content body update until all transactions have been handled.

suricata-verify-pr: 348
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-file-data.c
+++ b/src/detect-file-data.c
@@ -59,6 +59,8 @@ static void DetectFiledataSetupCallback(const DetectEngineCtx *de_ctx,
                                         Signature *s);
 static int g_file_data_buffer_id = 0;
 
+static inline HtpBody *GetResponseBody(htp_tx_t *tx);
+
 /* HTTP */
 static InspectionBuffer *HttpServerBodyGetDataCallback(DetectEngineThreadCtx *det_ctx,
         const DetectEngineTransforms *transforms,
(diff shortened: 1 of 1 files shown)


src/detect.c
```diff
@@ -1486,6 +1507,8 @@ static void DetectRunTx(ThreadVars *tv,
                     tx.tx_ptr, tx.tx_id,
                     flow_flags & STREAM_TOSERVER ? "toserver" : "toclient",
                     new_detect_flags);
+        if (alproto == ALPROTO_HTTP)
```
victorjulien, line 1510: I would prefer a solution where we can avoid adding checks here.

One way would be to make sure `HttpServerBodyGetDataCallback` is still called only once, but we cache the output for the duration of the current run. `DetectEngineInspectBufferHttpHeader` does something similar for HTTP headers.

victorjulien: Though I have to say I don't really like the solution `http_header` uses either. But it does put the cost where it belongs, in the keyword implementation. This way its not a cost that all the other keywords and protocols also have.

jlucovsky: @victorjulien This implements your second suggestion -- please review when you have time.

catenacyber:
Is there a S-V test for this ?
Does it also fix https://redmine.openinfosecfoundation.org/issues/3691 ?

victorjulien:
@jlucovsky do we have an SV test?

jlucovsky:
> @jlucovsky do we have an SV test?

https://github.com/OISF/suricata-verify/pull/340

jlucovsky:
> > @jlucovsky do we have an SV test?
> 
> [OISF/suricata-verify#340](https://github.com/OISF/suricata-verify/pull/340)

[OISF/suricata-verify#348](https://github.com/OISF/suricata-verify/pull/348)

jlucovsky:
Continued in #5549

PR #5081, opened 2020-06-17
Byteextractstr/part2/v6

inashivb:

Rework of #4970

### DIFF
--- a/src/util-affinity.c
+++ b/src/util-affinity.c
@@ -26,6 +26,7 @@
 #define _THREAD_AFFINITY
 #include "util-affinity.h"
 #include "util-cpu.h"
+#include "util-byte.h"
 #include "conf.h"
 #include "threads.h"
 #include "queue.h"
@@ -280,7 +281,10 @@ void AffinitySetupLoadFromConfig()
 
         node = ConfNodeLookupChild(affinity->head.tqh_first, "threads");
         if (node != NULL) {
-            taf->nb_threads = atoi(node->val);
+            if (StringParseInt32(&taf->nb_threads, 10, 0, (const char *)node->val) < 0) {
+                FatalError(SC_ERR_INVALID_ARGUMENT, "invalid value for threads "
+                           "count: '%s'", node->val);
+            }
             if (! taf->nb_threads) {
                 SCLogError(SC_ERR_INVALID_ARGUMENT, "bad value for threads count");
                 exit(EXIT_FAILURE);
--- a/src/util-byte.c
+++ b/src/util-byte.c
@@ -152,6 +152,9 @@ int ByteExtractUint32(uint32_t *res, int e, uint16_t len, const uint8_t *bytes)
     }
 
     ret = ByteExtract(&i64, e, len, bytes);
+    if (i64 > UINT32_MAX) {
+        return -1;
+    }
     if (ret <= 0) {
         return ret;
     }
@@ -173,6 +176,9 @@ int ByteExtractUint16(uint16_t *res, int e, uint16_t len, const uint8_t *bytes)
     }
 
     ret = ByteExtract(&i64, e, len, bytes);
+    if (i64 > UINT16_MAX) {
+        return -1;
+    }
     if (ret <= 0) {
         return ret;
     }
@@ -244,6 +250,9 @@ int ByteExtractStringUint32(uint32_t *res, int base, uint16_t len, const char *s
     if (ret <= 0) {
         return ret;
     }
+    if (i64 > UINT32_MAX) {
+        return -1;
+    }
 
     *res = (uint32_t)i64;
 
@@ -264,6 +273,9 @@ int ByteExtractStringUint16(uint16_t *res, int base, uint16_t len, const char *s
     if (ret <= 0) {
         return ret;
     }
+    if (i64 > UINT16_MAX) {
+        return -1;
+    }
 
     *res = (uint16_t)i64;
 
@@ -283,6 +295,9 @@ int ByteExtractStringUint8(uint8_t *res, int base, uint16_t len, const char *str
     int ret = ByteExtractString(&i64, base, len, str, false);
     if (ret <= 0) {
         return ret;
+    }
+    if (i64 > UINT8_MAX) {
+        return -1;
     }
 
     *res = (uint8_t)i64;
@@ -309,6 +324,9 @@ int StringParseUint32(uint32_t *res, int base, uint16_t len, const char *str)
     if (ret <= 0) {
         return ret;
     }
+    if (i64 > UINT32_MAX) {
+        return -1;
+    }
 
     *res = (uint32_t)i64;
 
@@ -329,6 +347,9 @@ int StringParseUint16(uint16_t *res, int base, uint16_t len, const char *str)
     if (ret <= 0) {
         return ret;
     }
+    if (i64 > UINT16_MAX) {
+        return -1;
+    }
 
     *res = (uint16_t)i64;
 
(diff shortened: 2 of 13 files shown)


src/util-affinity.c
```diff
@@ -280,7 +281,10 @@ void AffinitySetupLoadFromConfig()
 
         node = ConfNodeLookupChild(affinity->head.tqh_first, "threads");
         if (node != NULL) {
-            taf->nb_threads = atoi(node->val);
+            if (StringParseInt32(&taf->nb_threads, 10, 0, (const char *)node->val) < 0) {
```
catenacyber, line 284: Can `nb_threads` be negative ?
(if not, I am not sure that this should be part of this PR)

src/util-ip.c
```diff
@@ -65,9 +66,9 @@ bool IPv4AddressStringIsValid(const char *str)
 
     addr[dots][alen] = '\0';
     for (int x = 0; x < 4; x++) {
-        int a = atoi(addr[x]);
-        if (a < 0 || a >= 256) {
-            SCLogDebug("out of range");
+        int32_t a;
+        if (StringParseI32RangeCheck(&a, 10, 0, (const char *)addr[x], 0, 255) < 0) {
```
catenacyber, line 70: Would `StringParseU8` do the job ?

src/util-napatech.c
```diff
@@ -916,16 +926,28 @@ int NapatechGetStreamConfig(NapatechStreamConfig stream_config[])
 
                     char copystr[16];
                     strlcpy(copystr, stream->val, 16);
-                    ByteExtractStringUint16(&start, 10,  0, copystr);
-                    ByteExtractStringUint16(&end, 10,  0, strchr(copystr, '-') + 1);
+                    if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
```
catenacyber, line 929: Why from `ByteExtractStringUint16` to `StringParseUint8` ?

src/util-proto-name.c
```diff
@@ -60,10 +61,9 @@ void SCProtoNameInit()
             if (proto_ch == NULL)
                 continue;
 
-            int proto = atoi(proto_ch);
-            if (proto >= 255)
+            uint8_t proto;
+            if (StringParseU8RangeCheck(&proto, 10, 0, (const char *)proto_ch, 0, 255) < 0)
```
catenacyber, line 65: We do not need the range check here, do we ?

src/util-proto-name.c
```diff
@@ -60,10 +61,9 @@ void SCProtoNameInit()
...
                 continue;
-
```
catenacyber, line 66: nit ?

src/util-napatech.c
```diff
@@ -768,10 +768,20 @@ static uint32_t CountWorkerThreads(void)
 
                         char copystr[16];
                         strlcpy(copystr, lnode->val, 16);
-
-                        ByteExtractStringUint8(&start, 10,  0, copystr);
-                        ByteExtractStringUint8(&end, 10,  0, strchr(copystr, '-') + 1);
-
+                        if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
```
catenacyber, line 771: This feels like we could use Rust here ;-p

src/util-affinity.c
inashivb: Seems like ideally it should not be. But it is read from the conf so maybe its to keep a better check?

src/util-ip.c
inashivb: Yes. Changed.

src/util-napatech.c
inashivb: Because for all configuration vars, we want to make sure that the value is checked strictly i.e. not followed by any extra chars than required

src/util-proto-name.c
inashivb: Yes, we don't. Removed.

src/util-napatech.c
catenacyber: I meant why `8` instead of `16` ?

src/util-affinity.c
catenacyber: I would make it unsigned

catenacyber: I would make it unsigned

src/util-napatech.c
inashivb: This seems to be a mistake which should error out during compilation. Maybe I don't get it because of missing napatech support. Fixing. Thanks.

src/util-affinity.c
jlucovsky: Suggest `unsigned` handling too.

src/util-byte.c
```diff
@@ -152,6 +152,9 @@ int ByteExtractUint32(uint32_t *res, int e, uint16_t len, const uint8_t *bytes)
     }
 
     ret = ByteExtract(&i64, e, len, bytes);
+    if (i64 > UINT32_MAX) {
```
jlucovsky, line 155: Not sure why we're doing this. We've ensured that no more than 4 bytes are being extracted ... hence, 32 bits. How can the value exceed `0xffffffff`?

src/util-classification-config.c
```diff
@@ -281,12 +282,10 @@ int SCClassConfAddClasstype(DetectEngineCtx *de_ctx, char *rawstr, uint16_t inde
         SCLogInfo("pcre_copy_substring() failed");
         goto error;
     }
-    if (strlen(ct_priority_str) == 0) {
+    if (StringParseInt32(&ct_priority, 10, 0, (const char *)ct_priority_str) < 0) {
```
jlucovsky, line 285: Suggest unsigned here.

src/util-napatech.c
```diff
@@ -916,16 +926,28 @@ int NapatechGetStreamConfig(NapatechStreamConfig stream_config[])
 
                     char copystr[16];
                     strlcpy(copystr, stream->val, 16);
-                    ByteExtractStringUint16(&start, 10,  0, copystr);
-                    ByteExtractStringUint16(&end, 10,  0, strchr(copystr, '-') + 1);
+                    if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
+                        FatalError(SC_ERR_INVALID_VALUE, "Napatech invalid "
+                                   "stream_id start: '%s'", copystr);
```
jlucovsky, line 931: Nit ... is it "stream_id" or "stream id" (I suggest the latter)

src/util-privs.c
```diff
@@ -154,7 +155,9 @@ int SCGetUserID(const char *user_name, const char *group_name, uint32_t *uid, ui
 
     /* Get the user ID */
     if (isdigit((unsigned char)user_name[0]) != 0) {
-        userid = atoi(user_name);
+        if (ByteExtractStringUint32(&userid, 10, 0, (const char *)user_name) < 0) {
+            FatalError(SC_ERR_UID_FAILED, "invalid user id value: '%s'", user_name);
```
jlucovsky, line 159: Suggest making other places where `SCLogError` is followed by `exit` into `FatalError` (as a sep commit, of course).

src/util-byte.c
catenacyber: Ah yes, this is needed for `ByteExtractStringUint32` but redundant for `ByteExtractUint32`

src/util-privs.c
inashivb: That is going to come as a part of https://redmine.openinfosecfoundation.org/issues/3188

inashivb:
Replaced with #5089

PR #5089, opened 2020-06-20
Byteextractstr/part2/v7

inashivb:

### DIFF
--- a/src/util-napatech.c
+++ b/src/util-napatech.c
@@ -768,10 +768,20 @@ static uint32_t CountWorkerThreads(void)
 
                         char copystr[16];
                         strlcpy(copystr, lnode->val, 16);
-
-                        ByteExtractStringUint8(&start, 10,  0, copystr);
-                        ByteExtractStringUint8(&end, 10,  0, strchr(copystr, '-') + 1);
-
+                        if (StringParseUint8(&start, 10, 0, (const char *)copystr) < 0) {
+                            FatalError(SC_ERR_INVALID_VALUE, "Napatech invalid"
+                                       " worker range start: '%s'", copystr);
+                        }
+                        char *end_str = strchr(copystr, '-');
+                        if ((end_str == NULL) || (end_str != NULL && StringParseUint8(&end,
+                                        10, 0, (const char *) (end_str + 1)) < 0)) {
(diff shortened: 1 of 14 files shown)


src/util-napatech.c
```diff
@@ -916,16 +926,28 @@ int NapatechGetStreamConfig(NapatechStreamConfig stream_config[])
 
                     char copystr[16];
                     strlcpy(copystr, stream->val, 16);
-                    ByteExtractStringUint16(&start, 10,  0, copystr);
-                    ByteExtractStringUint16(&end, 10,  0, strchr(copystr, '-') + 1);
+                    if (StringParseUint16(&start, 10, 0, (const char *)copystr) < 0) {
```
jlucovsky, line 929: This may be ok --- but, Napatech stream ids are 8 bit values (max 255) so perhaps we should use `uint8`? Perhaps this is ok as `uint16_t` is used in the `NapatechStreamConfig` struct.

@pjyoung1 Any comments?

jasonish: I often like to know when a value is out of range, vs failed to parse. Can lead to much better error messages, but does separate the parseing and the range checking adding more code.

pjyoung1: I agree that this would be cleaner as uint8 instead of uint16.  (It could probably be changed to uint8_t in the NapatechStreamConfig as well.)

catenacyber: > I often like to know when a value is out of range, vs failed to parse

That could be implemented by functions such as `StringParseU64RangeCheck` returning different error codes

catenacyber:
By the way, should `atoi` be added in qa/coccinelle//banned-functions ?

inashivb:
> By the way, should `atoi` be added in qa/coccinelle//banned-functions ?

Probably yes. Maybe @victorjulien or @jasonish would be able to tell.

jasonish:
> Probably yes. Maybe @victorjulien or @jasonish would be able to tell.

I'd say yes.

catenacyber:
Shivani, will you add `atoi` as a banned function as a part of your pull requests to get rid of it ?

inashivb:
> Shivani, will you add `atoi` as a banned function as a part of your pull requests to get rid of it ?

Yes, sure. I shall do that will the last batch of changes.

PR #5094, opened 2020-06-23

inashivb:
ByteExtractString* functions.

### DIFF
--- a/src/detect-byte-extract.c
+++ b/src/detect-byte-extract.c
@@ -240,7 +240,12 @@ static inline DetectByteExtractData *DetectByteExtractParse(DetectEngineCtx *de_
                    "for arg 1 for byte_extract");
         goto error;
     }
-    bed->nbytes = atoi(nbytes_str);
+    if (StringParseUint8(&bed->nbytes, 10, 0,
+                               (const char *)nbytes_str) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for number of bytes"
+                   " to be extracted: \"%s\".", nbytes_str);
+        goto error;
+    }
 
     /* offset */
     char offset_str[64] = "";
@@ -251,7 +256,11 @@ static inline DetectByteExtractData *DetectByteExtractParse(DetectEngineCtx *de_
                    "for arg 2 for byte_extract");
         goto error;
     }
-    int offset = atoi(offset_str);
+    int offset;
+    if (StringParseInt32(&offset, 10, 0, (const char *)offset_str) < 0) {
+        SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for offset: \"%s\".", offset_str);
+        goto error;
+    }
     if (offset < -65535 || offset > 65535) {
         SCLogError(SC_ERR_INVALID_SIGNATURE, "byte_extract offset invalid - %d.  "
                    "The right offset range is -65535 to 65535", offset);
@@ -307,7 +316,13 @@ static inline DetectByteExtractData *DetectByteExtractParse(DetectEngineCtx *de_
                            "for arg %d for byte_extract", i);
                 goto error;
             }
-            int multiplier = atoi(multiplier_str);
+            int32_t multiplier;
+            if (StringParseInt32(&multiplier, 10, 0,
+                                 (const char *)multiplier_str) < 0) {
+                SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for"
+                        "multiplier: \"%s\".", multiplier_str);
+                goto error;
+            }
             if (multiplier < DETECT_BYTE_EXTRACT_MULTIPLIER_MIN_LIMIT ||
                 multiplier > DETECT_BYTE_EXTRACT_MULTIPLIER_MAX_LIMIT) {
                 SCLogError(SC_ERR_INVALID_SIGNATURE, "multipiler_value invalid "
@@ -411,7 +426,12 @@ static inline DetectByteExtractData *DetectByteExtractParse(DetectEngineCtx *de_
                            "for arg %d in byte_extract", i);
                 goto error;
             }
-            bed->align_value = atoi(align_str);
+            if (StringParseUint8(&bed->align_value, 10, 0,
+                                       (const char *)align_str) < 0) {
+                SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid align_value: "
+                           "\"%s\".", align_str);
+                goto error;
+            }
             if (!(bed->align_value == 2 || bed->align_value == 4)) {
                 SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid align_value for "
                            "byte_extract - \"%d\"", bed->align_value);
--- a/src/detect-engine-port.c
+++ b/src/detect-engine-port.c
@@ -51,12 +51,13 @@
 #include "host.h"
 #include "util-profiling.h"
 #include "util-var.h"
+#include "util-byte.h"
 
 static int DetectPortCutNot(DetectPort *, DetectPort **);
 static int DetectPortCut(DetectEngineCtx *, DetectPort *, DetectPort *,
                          DetectPort **);
 DetectPort *PortParse(const char *str);
-int DetectPortIsValidRange(char *);
+static int DetectPortIsValidRange(char *);
 
 /**
  * \brief Alloc a DetectPort structure and update counters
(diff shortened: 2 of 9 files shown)


src/detect-iprep.c
```diff
@@ -321,8 +322,8 @@ int DetectIPRepSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawstr)
     }
 
     if (value != NULL && strlen(value) > 0) {
-        int ival = atoi(value);
-        if (ival < 0 || ival > 127)
+        int8_t ival;
+        if (StringParseInt8(&ival, 10, 0, (const char *)value) < 0)
```
catenacyber, line 326: Range check with 127 ?

src/detect-modbus.c
```diff
@@ -191,13 +192,28 @@ static DetectModbus *DetectModbusAccessParse(DetectEngineCtx *de_ctx, const char
                 goto error;
 
             if (arg[0] == '>') {
-                modbus->address->min    = atoi((const char*) (arg+1));
+                if (StringParseUint16(&modbus->address->min, 10, 0,
```
catenacyber, line 195: Is this three times the same code that can be factorized ?

src/detect-byte-extract.c
```diff
@@ -251,7 +256,11 @@ static inline DetectByteExtractData *DetectByteExtractParse(DetectEngineCtx *de_
                    "for arg 2 for byte_extract");
         goto error;
     }
-    int offset = atoi(offset_str);
+    int offset;
+    if (StringParseInt32(&offset, 10, 0, (const char *)offset_str) < 0) {
```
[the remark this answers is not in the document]
jlucovsky: Or, use `Int16`?

src/detect-byte-extract.c
```diff
@@ -307,7 +316,13 @@ static inline DetectByteExtractData *DetectByteExtractParse(DetectEngineCtx *de_
                            "for arg %d for byte_extract", i);
                 goto error;
             }
-            int multiplier = atoi(multiplier_str);
+            int32_t multiplier;
+            if (StringParseInt32(&multiplier, 10, 0,
```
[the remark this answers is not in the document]
jlucovsky: With `uint16` (multiplier range is [1, 65535]

src/detect-engine-port.c
```diff
@@ -1345,15 +1351,12 @@ DetectPort *PortParse(const char *str)
...
-    long r = strtol(port, &end, 10);
-
-    if(*end == 0 && r >= 0 && r <= 65535)
-        return 1;
-    else
-        return 0;
+    uint16_t val;
+    if (StringParseUint16(&val, 10, 0, (const char *)port) < 0)
+        return -1;
+    return val;
```
jlucovsky, line 1359: The return values should be either `1` or `0` .. perhaps this static function could return a bool?

src/detect-fast-pattern.c
```diff
@@ -297,14 +299,15 @@ static int DetectFastPatternSetup(DetectEngineCtx *de_ctx, Signature *s, const c
...
+        uint16_t length;
+        if (StringParseUint16(&length, 10, 0,
+                              (const char *)arg_substr) < 0) {
+            SCLogError(SC_ERR_INVALID_SIGNATURE, "Invalid value for fast "
+                       "pattern: \"%s\"", arg_substr);
             goto error;
         }
 
-        if (offset + length > 65535) {
+        if (offset > 65535 - length) {
```
jlucovsky, line 310: Suggest using parens for clarity here.

src/detect-krb5-errcode.c
```diff
@@ -156,8 +157,10 @@ static DetectKrb5ErrCodeData *DetectKrb5ErrCodeParse (const char *krb5str)
     krb5d = SCMalloc(sizeof (DetectKrb5ErrCodeData));
     if (unlikely(krb5d == NULL))
         goto error;
-    krb5d->err_code = (int32_t)atoi(arg1);
-
+    if (StringParseInt32(&krb5d->err_code, 10, 0,
```
jlucovsky, line 160: Error codes are [0, 76]. Suggest uint8 range check or parsing as a uint8.

src/detect-byte-extract.c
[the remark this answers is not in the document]
catenacyber: `int16` is too short for `-65535` as it starts at `-32768` ;-)

src/detect-engine-port.c
catenacyber: I think the comments of the function are wrong and need to be fixed

src/detect-krb5-errcode.c
catenacyber: Do you know why `err_code` is defined as `int32` in the `krb5d` structure ?

src/detect-engine-port.c
jlucovsky: Ok ... suggest that
1. The function return a `bool` indicating success/failure
2. The function signature change to `char *port, uint16_t *port_val)`

src/detect-iprep.c
inashivb: Should I use `uint8_t` instead and do the range check for 127 since it seems ival should always be positive?

src/detect-modbus.c
inashivb: Except the last block. It uses `arg` instead of `arg + 1`.
How do you like the changes below? Please note that with these changes we assign `modbus->address->mode` before checking the value of `modbus->address->min`, I'm not sure if that would be a problem?
```
diff --git a/src/detect-modbus.c b/src/detect-modbus.c
index 9e412e759..f08d42d93 100644
--- a/src/detect-modbus.c
+++ b/src/detect-modbus.c
@@ -191,29 +191,21 @@ static DetectModbus *DetectModbusAccessParse(DetectEngineCtx *de_ctx, const char
             if (unlikely(modbus->address == NULL))
                 goto error;
 
+            int8_t idx;
             if (arg[0] == '>') {
-                if (StringParseUint16(&modbus->address->min, 10, 0,
-                                      (const char*) (arg + 1)) < 0) {
-                    SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
-                               "address: %s", (const char *)(arg + 1));
-                    goto error;
-                }
+                idx = 1;
                 modbus->address->mode   = DETECT_MODBUS_GT;
             } else if (arg[0] == '<') {
-                if (StringParseUint16(&modbus->address->min, 10, 0,
-                                      (const char*) (arg + 1)) < 0) {
-                    SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
-                               "address: %s", (const char*)(arg + 1));
-                    goto error;
-                }
+                idx = 1;
                 modbus->address->mode   = DETECT_MODBUS_LT;
             } else {
-                if (StringParseUint16(&modbus->address->min, 10, 0,
-                                      (const char*) arg) < 0) {
-                    SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
-                               "address: %s", arg);
-                    goto error;
-                }
+                idx = 0;
+            }
+            if (StringParseUint16(&modbus->address->min, 10, 0,
+                                  (const char*) (arg + idx)) < 0) {
+                SCLogError(SC_ERR_INVALID_VALUE, "Invalid value for min "
+                           "address: %s", (const char *)(arg + idx));
+                goto error;
             }
             SCLogDebug("and min/equal address %d", modbus->address->min);
```

src/detect-engine-port.c
inashivb: Point 1 done.
Point 2: Could you please tell why is it better to pass port value in this function than where it is now?

src/detect-krb5-errcode.c
inashivb: Could you please tell where did you check the error codes to be [0, 76]? I could find https://web.mit.edu/kerberos/krb5-1.5/krb5-1.5.4/doc/krb5-admin/Kerberos-V5-Library-Error-Codes.html, is this the wrong place to look at?

src/detect-engine-port.c
catenacyber: I think Jeff's suggestion is about differentiating the return code : error or failure, and the port value
Using an `int` for both at the same time may not be a good coding pattern

src/detect-modbus.c
catenacyber: The proposed change looks better indeed :-)

src/detect-iprep.c
catenacyber: Yes, even better :-) and remove the later cast in `val = (uint8_t)ival;`

src/detect-engine-port.c
inashivb: I see. Thanks @catenacyber !

inashivb:
Replaced with #5108

PR #5090, opened 2020-06-22
ikev1: add ikev1 parser - v2

fhonza:
Previous PR: [#5064](https://github.com/OISF/suricata/pull/5064)

Describe changes:
- testing of input args for null pointer for external c functions
- fixed two places returning pointers to variables which are free'd directly after returning
- store values for detecting keywords in transactions instead of the state to prevent reallocations on rust-side
- more logging in error cases
- using more rust language features
- check for integer underflows
- splitting long functions into smaller subfunctions
- shifting a function converting u8-arrays into hex (into common.rs, hope that's a good place)

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

#suricata-verify-pr: [#256](https://github.com/OISF/suricata-verify/pull/256)

PR #5097, opened 2020-06-24
ikev1: add ikev1 parser - v3

fhonza:
Previous PR: [#5090](https://github.com/OISF/suricata/pull/5090)

Describe changes:
- more specific and corrected logging output
- remove unnecessary defines
- rename JsonIKEV1AddMetadata -> EveIKEV1AddMetadata
- different handling of 2 error cases

#suricata-verify-pr: [#256](https://github.com/OISF/suricata-verify/pull/256)


fhonza:
is it possible to run the Travis CI again?

It's failing in http-evader-495 from suricata-verify, cannot reproduce it locally and it's very unlikely for such a side effect...

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,758 @@
...
+    pub transform_type: u8,
+    pub sa_attributes: &'a [u8]
+}
+
+// 4 -> Key Exchange
+pub struct KeyExchangePayload<'a> {
+    pub key_exchange_data: &'a [u8],
+}
+
+// 5 -> Identification
```
catenacyber, line 110: Are these TODOs ?

rust/src/ikev1/parser.rs
```diff
@@ -0,0 +1,758 @@
...
+
+pub fn parse_security_association(
+    i: &[u8],
+    length: u16,
+) -> IResult<&[u8], SecurityAssociationPayload> {
+    do_parse!(
+        i,
+        domain_of_interpretation: be_u32 >>
+        situation: cond!(domain_of_interpretation == 1, take!(4)) >>
+        data: cond!(domain_of_interpretation == 1 && length >= 8, take!(length - 8)) >>
```
catenacyber, line 290: This feels a bit odd.
Should not the caller limit the input slice to `length` ?

src/detect-ikev1-chosen-sa.c
```diff
@@ -0,0 +1,286 @@
...
+     * ikev1.chosen_sa_attribute:"encryption_algorithm=4"
+     * ikev1.chosen_sa_attribute:"hash_algorithm=8"
+     */
+    DetectIkev1ChosenSaData *dd = NULL;
+    int ret = 0, res = 0;
+    int ov[MAX_SUBSTRINGS];
+    char attribute[100];
+    char value[100];
+
+    ret = DetectParsePcreExec(&parse_regex, rawstr, 0, 0, ov, MAX_SUBSTRINGS);
```
catenacyber, line 154: Could we do this in Rust ?

src/detect-ikev1-chosen-sa.c
```diff
@@ -0,0 +1,286 @@
...
+ *  \retval 1 on success
+ *  \retval 0 on failure
+ */
+static int IKEV1ChosenSaParserTest (void)
+{
+    DetectIkev1ChosenSaData *de = NULL;
+    de = DetectIkev1ChosenSaParse("hash_algorithm=2");
+    if (de) {
+        DetectIkev1ChosenSaFree(NULL, de);
+        return 1;
```
catenacyber, line 273: Can we use `PASS` and `FAIL` macros ? (and `FAIL_IF_NULL`...)

src/detect-ikev1-exch-type.c
```diff
@@ -0,0 +1,287 @@
...
+            DetectEngineInspectIkev1ExchTypeGeneric);
+
+    DetectAppLayerInspectEngineRegister("ikev1.exchtype",
+            ALPROTO_IKEV1, SIG_FLAG_TOCLIENT, 1,
+            DetectEngineInspectIkev1ExchTypeGeneric);
+
+    g_ikev1_exch_type_buffer_id = DetectBufferTypeGetByName("ikev1.exchtype");
+}
+
+static int DetectEngineInspectIkev1ExchTypeGeneric(ThreadVars *tv,
```
catenacyber, line 94: Is this needed ? (it is not a tick buffer keyword)

src/detect-ikev1-key-exchange-payload-length.c
```diff
@@ -0,0 +1,306 @@
...
+        SCLogError(SC_ERR_PCRE_COPY_SUBSTRING, "pcre_copy_substring failed");
+        goto error;
+    }
+
+    dd = SCCalloc(1, sizeof(DetectIkev1KeyExchangePayloadLengthData));
+    if (unlikely(dd == NULL))
+        goto error;
+
+    if (strcmp(host_type, "client") == 0) {
+        dd->host_type = HOST_CLIENT;
```
catenacyber, line 219: Is this used ?

src/detect-ikev1-key-exchange-payload.c
```diff
@@ -0,0 +1,172 @@
...
+}
+
+void DetectIkev1KeyExchangeRegister(void)
+{
+    // register client_key_exchange
+    sigmatch_table[DETECT_AL_IKEV1_CLIENT_KEY_EXCHANGE].name = KEYWORD_NAME_CLIENT_KEY_EXCHANGE;
+    sigmatch_table[DETECT_AL_IKEV1_CLIENT_KEY_EXCHANGE].url = "/rules/" KEYWORD_DOC_CLIENT_KEY_EXCHANGE
+    sigmatch_table[DETECT_AL_IKEV1_CLIENT_KEY_EXCHANGE].desc = "sticky buffer to match on the IKEv1 client_key_exchange_payload";
+    sigmatch_table[DETECT_AL_IKEV1_CLIENT_KEY_EXCHANGE].Setup = DetectClientKeyExchangeSetup;
+    sigmatch_table[DETECT_AL_IKEV1_CLIENT_KEY_EXCHANGE].flags |= SIGMATCH_NOOPT|SIGMATCH_INFO_STICKY_BUFFER;
```
catenacyber, line 138: Can we validate the content is hexadecimal ?
like https://github.com/OISF/suricata/blob/a98a20973230bb662ddf204fb4be7b7fb657024c/src/detect-ssh-hassh.c#L213

src/detect-ikev1-key-exchange-payload.c
```diff
@@ -0,0 +1,172 @@
...
+            PrefilterGenericMpmRegister, GetClientKeyExchangeData, ALPROTO_IKEV1,
+            3);
+
+    DetectBufferTypeSetDescriptionByName(BUFFER_NAME_CLIENT_KEY_EXCHANGE, BUFFER_DESC_CLIENT_KEY_EXCHANGE);
+
+    g_buffer_client_key_exchange_id = DetectBufferTypeGetByName(BUFFER_NAME_CLIENT_KEY_EXCHANGE);
+    SCLogDebug("registering " BUFFER_NAME_CLIENT_KEY_EXCHANGE " rule option");
+
+    // register server_key_exchange
+    sigmatch_table[DETECT_AL_IKEV1_SERVER_KEY_EXCHANGE].name = KEYWORD_NAME_SERVER_KEY_EXCHANGE;
```
catenacyber, line 154: Should not we use the rule direction instead of two keywords ?

fhonza:
@catenacyber thanks & thanks for your review!

Yes i think it would make sense to have the same keywords since both IKE protocols share the same mechanics.
I would propose doing it in another feature since it would be more changes in the IKEv2 parser.

Found the C-part a bit "code intensive" for registering the keywords in suricata, perhaps it would be easier to register the same keywords for both protocols and deciding when calling the rust function which parser to use. Depending on the transaction/state type. But not sure if there would be type/pointer clashes :)

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -1287,3 +1287,109 @@ Example of RFB logging, with full VNC style authentication parameters:
...
+* "min_ver": Minor version of the ISAKMP header.
+* "contained_payload_types": List of payload types in the current packet, as numeric values.
+* "contained_payload_types_verbose": List of payload types in the current packet, in human-readable form. Needs ``extended: yes`` set in the ``ikev1`` EVE output option.
+* "encrypted_payloads": Set to ``true`` if the payloads in the packet are encrypted.
+* "exchange_type": Type of the exchange, as numeric values.
+* "exchange_type_verbose": Type of the exchange, in human-readable form. Needs ``extended: yes`` set in the ``ikev1`` EVE output option.
+* "doi": Value of the domain of interpretation (DOI).
+* "server.key_exchange_payload", "client.key_exchange_payload": Public key exchange payloads of the server and client.
+* "server.key_exchange_payload_length", "client.key_exchange_payload_length": Length of the public key exchange payload.
+* "server.nonce_payload", "client.nonce_payload": Nonce payload of the server and client.
```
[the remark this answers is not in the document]
fhonza: yes it's a cryptographic nonce, added it for checking weird values for potential exploiting.

rust/src/ikev1/ikev1.rs
```diff
@@ -0,0 +1,583 @@
...
+                tx.payload_types = Some(payload_types);
+                tx.encrypted_payloads = encrypted_payloads;
+                self.transactions.push(tx);
+                self.set_progress();
+
+                return AppLayerResult::ok(); // todo either remove outer loop or check header length-field if we have completely read everything
+            }
+            Err(nom::Err::Incomplete(_)) => {
+                SCLogDebug!("Insufficient data while parsing IKEV1");
+                return AppLayerResult::err();
```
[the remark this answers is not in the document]
fhonza: yes indeed, there is an error created in app-layer-parser.c if it's != TCP. learned it the hard way :)

rust/src/ikev1/ikev1.rs
```diff
@@ -0,0 +1,583 @@
...
+
+                if isakmp_header.maj_ver != 1 {
+                    SCLogDebug!("Unsupported ISAKMP major_version");
+                    return AppLayerResult::err();
+                }
+
+                let mut tx = self.new_tx();
+                tx.spi_initiator = Some(format!("{:016x}", isakmp_header.init_spi));
+                tx.spi_responder = Some(format!("{:016x}", isakmp_header.resp_spi));
+                tx.maj_ver = Some(isakmp_header.maj_ver);
```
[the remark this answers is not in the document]
fhonza: felt better to have it, in case someone wonders why there is a minor version, but not a major one :thinking:

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -1287,3 +1287,109 @@ Example of RFB logging, with full VNC style authentication parameters:
...
+    "spi_responder": "a00b8ef0902bb8ec",
+    "maj_ver": 1,
+    "min_ver": 0,
+    "contained_payload_types": [
+      "10",
+      "20",
+      "4",
+      "13"
+    ],
+    "contained_payload_types_verbose": [
```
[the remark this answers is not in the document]
fhonza: I think it's better to have the raw data from the packets as default and readable strings for known values. For example there is a microsoft extension of ikev1:
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-aips/eee3de64-3847-4451-978e-9513ff187d30
This is using additional payload_types for example. It's completely compatible with the parser but has undocumented types.
Could imagine that there are other implementations now or in the future as well....

src/detect-ikev1-exch-type.c
```diff
@@ -0,0 +1,287 @@
...
+
+#include "rust-bindings.h"
+
+/**
+ *   [ikev1.exchtype]:[<|>|<=|>=]<type>;
+ */
+#define PARSE_REGEX "^\\s*(<=|>=|<|>)?\\s*([0-9]+)\\s*$"
+static DetectParseRegex parse_regex;
+
+enum DetectIkev1ExchTypeCompareMode {
```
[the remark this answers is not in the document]
fhonza: That's nice! Will switch to that and add LE/GE, saves a lot code there....

doc/userguide/output/eve/eve-json-format.rst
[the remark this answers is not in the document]
catenacyber: Ok thanks

doc/userguide/output/eve/eve-json-format.rst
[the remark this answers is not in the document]
catenacyber: Ok thanks

src/detect-ikev1-exch-type.c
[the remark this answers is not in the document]
catenacyber: Great, waiting for the next PR

catenacyber:
> Yes i think it would make sense to have the same keywords since both IKE protocols share the same mechanics.
I would propose doing it in another feature since it would be more changes in the IKEv2 parser.

Thanks for this reply.

If it makes sense, there are some architecture questions which may need a thought first :
- is IKEv1 a new parser ? (as I am doing with HTTP1 and HTTP2) or a new case in the IKE(v2) parser ? (as SMB1 and AMB2)
- Do we use the same keywords ? How do we specify if we want a rule specifically apply to one version ? I think there are DCERPC keywords which work for both SMB and DCERPC.

The idea is to avoid doing some work and then undo it.

rust/src/ikev1/ikev1.rs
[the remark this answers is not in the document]
catenacyber: ok thanks

rust/src/ikev1/ikev1.rs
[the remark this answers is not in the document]
catenacyber: ok fair enough

fhonza:
Afaik the "outer shell" of ikev1/v2 is the same (ISAKMP-Header) and what the protocol does in the end, by sending key-exchange proposals to the responder and the responder selects one of them with a number of properties for the DH-Key Exchange.

The difference is mainly the structure in the payloads and the sequence of the messages. Have to admit i'm not sure if it would be simpler to unify both parsers. Could be that it ends in some ugly construct, unfortunately I know only the ikev1 rfc's.

Perhaps @chifflier has some educated opinion?

catenacyber:
If IKE shares a header between v1 and v2, I would suggest doing as SMB : one parser to Suricata (ie one value for `AppProtoEnum` in app-layer-protos.h) and early branching in the parsing function to parse v1 or v2
This way, it would also be more transparent to the end-users and signature writers.

Is is a big rework indeed. Could you give it a try and come back if you have questions ?

fhonza:
hmmm yes for seeing how it feels i can try to decide on the major_version field to call the existing ikev2 functions and see if i can store the results in the transaction/state. 
Perhaps the v4 branch will be more a basis for further discussion than a real attempt for merging :)

src/detect-ikev1-chosen-sa.c
fhonza: just wanted to be consistent with the existing keyword implementations.

src/detect-ikev1-key-exchange-payload.c
fhonza: Can you be more precise on this? Not sure I understand...

src/detect-ikev1-exch-type.c
fhonza: I've tried to remove this function but got a segfault in this line: ```sigmatch_table[smd->type].Match(det_ctx, p, s, smd->ctx)``` in detect-engine.c

Have tried to do it like you did in detect-icmpv6-mtu.c by specifying the Match-function directly, but the function signature there has no transaction (which i need to obtain the value from rust).

Do you know an example somewhere?

rust/src/ikev1/parser.rs
fhonza: not sure :)

didn't need the payload-types now but wanted to have them at least noted. Perhaps there is in the future the need to add them.

src/detect-ikev1-chosen-sa.c
catenacyber: We are now trying to move more code to Rust cf https://redmine.openinfosecfoundation.org/issues/3195

One example of signature code parsing in Rust is given :
https://github.com/OISF/suricata/blob/master/rust/src/dns/detect.rs#L34

src/detect-ikev1-key-exchange-payload.c
catenacyber: Instead of having 2 keywords `ikev1.client_key_exchange_payload` and `ikev1.server_key_exchange_payload`, could we have just one keyword `ikev1.key_exchange_payload` ? And then rely on additional classic keyword`flow:to_client` to distinguish if needed ?

fhonza: You are right, this should work. Have to admit that i didn't know it.
...perhaps i should read a bit more in the manual ;)

PR #5093, opened 2020-06-23
Fuzz dnp3 fixes v1

catenacyber:
Does not proceed final chunk if we got an error previously
Flips the direction for last chunk as usual

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) tickets:
https://redmine.openinfosecfoundation.org/issues/3772
https://redmine.openinfosecfoundation.org/issues/3773
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22832

Describe changes:
- Fixes fuzz target `fuzz_applayerparserparse` by _not_ processing a last chunk if we had an error previously
- Improves DNP3 fuzzing by not checking CRC while fuzzing
- Improves DNP3 probing parser by fixing direction based on header field
- Fixes DNP3 response parsing, as reported in #5063 

So, this is a mix of DNP3 and fuzzing, and can be split between different Pull Requests

Should I make a S-V test for the probing parser ?

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5098, opened 2020-06-24
MQTT parser - v4

satta:
Previous PR: #4960

- Allow to set maximum size limit for parsing messages.
- Remove unchecked integer arithmetics in some places.
- Convert `mqtt.(un)subscribe.topic` to sticky buffer.
- Add comment explaining that there is no risk of collision for the pseudo pkt id.
- Use `String::from_utf8_lossy()` for string conversion.
- Remove unnecessary `MQTTMessage::message_type_id()`.
- Simplify double `CONNECT` detection.
- Return `ALPROTO_UNKNOWN` on incomplete probing.
- Add comment about code repetition
- Remove explicit buffering code.
- Move `unsafe` into inner code.
- Unify use of`MQTTFlagState` between C and Rust.
- Switch to jsonbuilder.
- Remove `array_append_integer()` from JSON code.
- Update documentation.
- Convert parse functions to return `AppLayerResult`.

Suricata-verify targets have also been updated in the `mqtt` branch there to introduce checks for `max-msg-length` limits.


rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,846 @@
...
+export_tx_get_detect_state!(rs_mqtt_tx_get_detect_state, MQTTTransaction);
+export_tx_set_detect_state!(rs_mqtt_tx_set_detect_state, MQTTTransaction);
+
+#[no_mangle]
+pub extern "C" fn rs_mqtt_probing_parser(
+    _flow: *const Flow,
+    _direction: u8,
+    input: *const u8,
+    input_len: u32,
+    _rdir: *mut u8,
```
catenacyber, line 566: Can we know the direction of a mqtt message ? and so set `rdir` accordingly 
cf https://github.com/OISF/suricata/pull/5093 and more precisely https://github.com/OISF/suricata/pull/5093/commits/182ae51f483e5e7dca4f0fb83b97b96a3a00ccb0

rust/src/mqtt/parser.rs
```diff
@@ -0,0 +1,647 @@
...
+            // type.
+            if len > max_msg_size {
+                let msg = MQTTMessage {
+                    header: header,
+                    op: MQTTOperation::TRUNCATED(MQTTTruncatedData {
+                        original_message_type: message_type,
+                        skipped_length: len + skiplen,
+                    }),
+                };
+                return Ok((&input, msg));
```
catenacyber, line 495: Should not it be `fullrem` instead of `input` ?

rust/src/mqtt/logger.rs
```diff
@@ -0,0 +1,326 @@
...
+                    for prop in properties {
+                        prop.set_json(js)?;
+                    }
+                    js.close()?; // properties
+                }
+                js.close()?; // disconnect
+            },
+            MQTTOperation::TRUNCATED(ref trunc) => {
+                js.open_object(&message_code_string(trunc.original_message_type))?;
+                js.set_uint("qos", msg.header.qos_level as u64)?;
```
catenacyber, line 307: There is duplicated code for all 3 `header` fields, is not there ?

rust/src/mqtt/mqtt_message.rs
```diff
@@ -0,0 +1,190 @@
...
+    PUBCOMP(MQTTMessageIdOnly),
+    SUBSCRIBE(MQTTSubscribeData),
+    SUBACK(MQTTSubackData),
+    UNSUBSCRIBE(MQTTUnsubscribeData),
+    UNSUBACK(MQTTUnsubackData),
+    AUTH(MQTTAuthData),
+    PINGREQ,
+    PINGRESP,
+    DISCONNECT(MQTTDisconnectData),
+    TRUNCATED(MQTTTruncatedData),
```
catenacyber, line 48: Could you add a comment to specify that `TRUNCATED` is special : not in the MQTT specification, but Suricate-Specific ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,846 @@
...
+                    self.transactions.push(tx);
+                }
+            },
+            MQTTOperation::UNASSIGNED => {
+                let mut tx = self.new_tx(msg, toclient);
+                tx.complete = true;
+                &mut MQTTState::set_event(&mut tx, MQTTEvent::UnassignedMsgtype);
+                self.transactions.push(tx);
+            },
+            MQTTOperation::TRUNCATED(_) => {
```
catenacyber, line 375: I am wondering if there is more to do for certain messages type...

I think we can expect certain messages type to be truncated like `PUBLISH`, but we do not expect a truncated `CONNECT` 

What are your thoughts on this ?

rust/src/mqtt/mqtt.rs
```diff
@@ -0,0 +1,846 @@
...
+        }
+
+
+        while current.len() > 0 {
+            let mut skipped = false;
+            SCLogDebug!("request: handling {}", current.len());
+            match parse_message(current, self.protocol_version, self.max_msg_len) {
+                Ok((rem, msg)) => {
+                    SCLogDebug!("request msg {:?}", msg);
+                    match msg.op {
```
catenacyber, line 435: style : `match` with one case can be an `if`

rust/src/mqtt/logger.rs
satta: Do you mean the `qos`, `retain` and `dup` ones? True, but I didn't find a way to handle those before branching out into the different message types because with the new stateful JSONBuilder the calls to `js.set_*()` need to come after the opening of the type-specific sub-object (e.g. `js.open_object("connect")?;`) which obviously needs to be in the `match` branches because before we dont know what type we have. I don't see that as much of a problem here than just a minor style issue.
I can put them in a separate inlined function and call that in each branch to reduce code duplication if that is enough to address this concern.

rust/src/mqtt/mqtt_message.rs
satta: Of course.

rust/src/mqtt/mqtt.rs
satta: Yep, changed -- thanks.

rust/src/mqtt/mqtt.rs
satta: I think we do. It is certainly possible to have a `CONNECT` with lots of big user properties in MQTT5, which could also easily hit the limit. I think we should always only consider the length specified in the header and not make assumptions based on the message type.
All types have the same problem though: to get the packet identifier (or message ID) that allows us to tie transactions together, we need to parse the message. The ID comes after the properties which by themselves could be huge already.

rust/src/mqtt/mqtt.rs
satta: I didn't know about `rdir` before and will take a look. In general, the direction is defined by the initial communication (`CONNECT`) which identifies the broker and client. Afterwards, both sides can exchange identical messages between themselves (e.g. both can `PUBLISH`).

rust/src/mqtt/parser.rs
satta: My intention here was to treat the complete input as if it remained unparsed. This can also be seen by the fact that I am adding the `skiplen` (i.e. the header length) to the parsed value for _remaining length_ which excludes the header.

rust/src/mqtt/mqtt.rs
catenacyber: Thanks for the insights :-)

rust/src/mqtt/logger.rs
catenacyber: This is indeed a minor style issue

Inline function seems a good solution to me.

Another solution may be in a generic `js.open_object("header")?;` before the `match`
What do you think about it ?
Does it make the use of json output simpler ? to find every log with a certain sos for instance, whatever the message type ? (does that make sense ?)

rust/src/mqtt/mqtt.rs
catenacyber: Ok, so I am not sure it makes sense here...

rust/src/mqtt/parser.rs
catenacyber: > My intention here was to treat the complete input as if it remained unparsed.

Interesting. Why this choice ?

On my work with HTTP2, I try to parse as much as I can(that is the fixed header)

satta: Note we're just looking at the case of `TRUNCATED` and in this case it makes it easier to skip the rest of the input without always having to account for the variable number of 2-5 bytes already parsed for the fixed header. 

Note that we _have_ parsed the header, it's just that we have decided we want to skip it and return the whole size of the input to skip on the request/response handler side. Whether we include the header length in the returned rest or not does not make a difference here IMHO.

catenacyber: Maybe I misunderstand something.

Do we still have the `qos` and other header field for this `TRUNCATED` messages ?

victorjulien:
> @victorjulien do you want me to take the time to make a full review ?

Yes, please do! Thanks.

rust/src/mqtt/parser.rs
satta: Yes, we do. These are part of the fixed header (2-5 bytes). So we can always read these, we even have to read them to get the total length of the message -- which is the amount of bytes to be compared against the limit.  In the header there is a field that gives us the _remaining length_ in addition to the header. If the remaining length is less than the limit, we parse the message and return the remainder as always. If it is larger than the limit, we return TRUNCATED with the header values we have and pass the total length of the input to be skipped to the caller (i.e. the `handle_response`/`handle_request`). Hence we from that side we act as if that message was not parsed at all, but still carry the metadata from the fixed header. That is just what I would expect.

catenacyber: Ok, thanks for your explanations

Could you please add a simple comment that this is indeed intentional having here `input` instead of `fullrem` ?

satta: Absolutely!

satta:
Next PR: #5121

PR #5100, opened 2020-06-25
DCERPC GAP handler

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3564

Commit to look at: https://github.com/OISF/suricata/commit/8161984f8b805b6e208ca8cbda5bc76092d04fc6


rust/src/dcerpc/dcerpc.rs
```diff
@@ -301,6 +300,10 @@ pub struct DCERPCState {
     pub prev_dir: u8,
     pub prev_tx_call_id: u32,
     pub clear_bind_cache: bool,
+    pub skip_ts: u32,
```
victorjulien, line 303: unused?

victorjulien:
Unless I'm missing something, I think this will just partly work. If we have TCP request spread over 2 segments, and only the first segment is lost, how do we catch up? I think in this case the 2nd segments data will simply be evaluated as if it is the start of the next record, right?

rust/src/dcerpc/dcerpc.rs
inashivb: Yes. Had introduced it earlier trying to track the skip data, forgot removing.

inashivb:
> Unless I'm missing something, I think this will just partly work. If we have TCP request spread over 2 segments, and only the first segment is lost, how do we catch up? I think in this case the 2nd segments data will simply be evaluated as if it is the start of the next record, right?

That's right. Unless it so happens that the header checks also fall in place for the 2nd segment (RPC version 5 for TCP), it will return an error in that case. But, it still leaves the possibility of that happening. I'm not sure how can we identify a second segment.
Question:  Would  `core::STREAM_MIDSTREAM` be set when the second segment comes in?

victorjulien:
> > Unless I'm missing something, I think this will just partly work. If we have TCP request spread over 2 segments, and only the first segment is lost, how do we catch up? I think in this case the 2nd segments data will simply be evaluated as if it is the start of the next record, right?
> 
> That's right. Unless it so happens that the header checks also fall in place for the 2nd segment (RPC version 5 for TCP), it will return an error in that case. But, it still leaves the possibility of that happening. I'm not sure how can we identify a second segment.
> Question: Would `core::STREAM_MIDSTREAM` be set when the second segment comes in?

No MIDSTREAM will only be set if the TCP session was picked up midstream (so we haven't seen the 3whs). It's a similar case to gaps though, as the pick up might have happened in the middle of a record.

Wrt finding a record, in the SMB case we simply do a search for the record identifier. In SMB that is quite easy, see https://github.com/OISF/suricata/blob/master/rust/src/smb/smb.rs#L1382 https://github.com/OISF/suricata/blob/master/rust/src/smb/smb2_records.rs#L538

The NFS code (https://github.com/OISF/suricata/blob/master/rust/src/nfs/nfs.rs#L1053) takes a more brute force approach. It just runs the probe function on the input data until it succeeds.

inashivb:
Replaced with #5193

PR #5088, opened 2020-06-20
filestore: Ensure configured file-store.steam-depth is greater than or equal to stream.reassembly.depth

jlucovsky:
Continuation of #5073 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2506](https://redmine.openinfosecfoundation.org/projects/suricata/issues/2506) 

Describe changes:
- Ensure configured stream depth for filestore is either 0 or greater than configured stream reassembly. depth. 
- Issue a warning message and reset the file-store.depth value if it's less.

#suricata-verify-pr: https://github.com/OISF/suricata-verify/pull/255
#suricata-verify-repo: https://github.com/jlucovsky/suricata-verify/
#suricata-verify-branch: jlucovsky:2506/2
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5104, opened 2020-06-26
filestore: Ensure configured file-store.steam-depth is greater than or equal to stream.reassembly.depth

jlucovsky:
Continuation of #5088

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2506](https://redmine.openinfosecfoundation.org/projects/suricata/issues/2506) 

Describe changes:
- Issue a warning message when `filestore.stream-depth` is less than `stream.reassembly.depth`

#suricata-verify-pr: https://github.com/OISF/suricata-verify/pull/258
#suricata-verify-repo: https://github.com/jlucovsky/suricata-verify/
#suricata-verify-branch: jlucovsky:2506/3
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:


src/output-filestore.c
```diff
@@ -492,9 +493,14 @@ static OutputInitResult OutputFilestoreLogInitCtx(ConfNode *conf)
...
-        } else {
-            FileReassemblyDepthEnable(stream_depth);
         }
+        if (stream_depth && stream_depth <= stream_config.reassembly_depth) {
+            SCLogWarning(SC_WARN_FILESTORE_CONFIG,
+                       "file-store.stream-depth value %" PRIu32 " has "
+                       "no effect since it's less than stream.reassembly.depth "
+                       "value.", stream_depth);
+        } 
+        FileReassemblyDepthEnable(stream_depth);
```
victorjulien, line 503: the message above says that `stream_depth` has no effect, yet it is used here regardless. Is that an error?

jlucovsky: This is intentional but we can change the behavior. Since the configured value is less than `stream.reassembly.depth` its effect is NIL.

Test case `filestore-v2.8-stream-depth` covers this.

jasonish: I think its clearer to not call this function if the value is determined to have no affect. I have to jump through a few functions to verify this in code.

victorjulien: Agree with Jason.

jlucovsky:
Continued in #5162

PR #5107, opened 2020-06-26
file-hash-common: fix rule_file truncation - v1

jasonish:
Loading file hash lists uses dirname(3) on the
de_ctx->rule_file which modifies the contents,
removing the last part of the path. So on subsequent
calls the rule_file no longer contains the rule_file,
but instead just the directory name.

Mostly noticed when using "-S" with rule files outside
of the default-rule-path which requires more hunting for
the rule file.

Prscript:
- https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/840
- https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/482

### DIFF
--- a/src/detect-file-hash-common.c
+++ b/src/detect-file-hash-common.c
@@ -202,6 +202,7 @@ static DetectFileHashData *DetectFileHashParse (const DetectEngineCtx *de_ctx,
     DetectFileHashData *filehash = NULL;
     FILE *fp = NULL;
     char *filename = NULL;
+    char *rule_filename = NULL;
 
     /* We have a correct hash algorithm option */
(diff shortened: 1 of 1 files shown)


jasonish:
Something else to consider here.  When loading a file hash list, it first looks for the file specific relative to the default-rule-directory, then relative to the rule file.  When using -S, or even specifying full paths to filenames in suricata.yaml, this could result in the wrong file being loaded. I'd like to consider looking relative to the rule file first.

Found in the context of testing the dataset support in Suricata-Update.

victorjulien:
Merged in #5110, thanks!

jasonish:
Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3802

PR #5108, opened 2020-06-26
Byteextractstr/part3/v2

inashivb:
ByteExtractString* functions.

### DIFF
--- a/src/detect-engine-port.c
+++ b/src/detect-engine-port.c
@@ -51,12 +51,13 @@
 #include "host.h"
 #include "util-profiling.h"
 #include "util-var.h"
+#include "util-byte.h"
 
 static int DetectPortCutNot(DetectPort *, DetectPort **);
 static int DetectPortCut(DetectEngineCtx *, DetectPort *, DetectPort *,
                          DetectPort **);
 DetectPort *PortParse(const char *str);
-int DetectPortIsValidRange(char *);
+static bool DetectPortIsValidRange(char *, uint16_t *);
 
 /**
  * \brief Alloc a DetectPort structure and update counters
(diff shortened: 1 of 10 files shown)


src/detect-modbus.c
```diff
@@ -190,14 +191,21 @@ static DetectModbus *DetectModbusAccessParse(DetectEngineCtx *de_ctx, const char
...
             } else if (arg[0] == '<') {
-                modbus->address->min    = atoi((const char*) (arg+1));
+                idx = 1;
                 modbus->address->mode   = DETECT_MODBUS_LT;
             } else {
-                modbus->address->min    = atoi((const char*) arg);
+                idx = 0;
+            }
+            if (StringParseUint16(&modbus->address->min, 10, 0,
+                                  (const char*) (arg + idx)) < 0) {
```
jlucovsky, line 205: nit: `const char *`

src/detect-modbus.c
```diff
@@ -329,7 +360,12 @@ static DetectModbus *DetectModbusFunctionParse(DetectEngineCtx *de_ctx, const ch
             if (modbus->subfunction == NULL)
                 goto error;
 
-            *(modbus->subfunction) = atoi((const char*) arg);
+            if (StringParseUint16(&(*modbus->subfunction), 10, 0, (const char *)arg) < 0) {
```
jlucovsky, line 363: Suggest making `subfunction` a `uint16_t` instead of a pointer. This would result in no-net increase of space and would allow the checks and free of this to be removed and the call here would simply pass the address of `subfunction` rather than the address of the pointer.

src/detect-engine-port.c
```diff
@@ -1295,20 +1296,22 @@ DetectPort *PortParse(const char *str)
...
+    if ((port2 = strchr(port, ':')) != NULL) {
         /* 80:81 range format */
         port2[0] = '\0';
         port2++;
 
-        if (DetectPortIsValidRange(port))
-            dp->port = atoi(port);
-        else
-            goto error;
+        if (strcmp(port, "") != 0) {
```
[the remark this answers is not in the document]
catenacyber: Yes `:1024` means any port below 1024
We do not have to specify it `0:1024`

src/detect-modbus.c
catenacyber: I think `subfunction` is a pointer because every modbus message does not have a subfunction, cf file detect-engine-modbus.c
So, you need one more bit in the global structure to say if there is a sub function or not.

I would say this is not the point of this PR

jlucovsky: It would simplify the code and result in no-net increase in space.

catenacyber: I agree that it is good for space.
But how do you @jlucovsky make a simpler code in detect-engine-modbus.c ?

The 2 solutions I see are :
- using a int32_t for `subfunction` and having `-1` value by default to say when there is no subfunction
- add a boolean `has_subfunction ` to the structure to say if the value is a real one

jlucovsky: I suggest the second option. Avoiding the allocation and memory free  will be possible.

Thus, add a boolean, initialize to `false` and only set when the sub-function is present.

catenacyber: Should it be part of this PR ?

victorjulien: > Should it be part of this PR ?

No, this should be a separate PR. This PR is about simply converting the string to int conversions.

inashivb:
Replaced with #5144

PR #5114, opened 2020-06-29
Debug validate for JsonBuilder - v1

jasonish:
With --enable-debug-validation on, JsonBuilder will now panic if something is done in an invalid state, such as closing an already closed object which can lead to items not being logged.


victorjulien:
Merged in #5115, thanks Jason!

PR #5117, opened 2020-06-30
hyperscan: better error message if hs not compiled

inashivb:


src/detect-engine-mpm.c
```diff
@@ -643,6 +643,17 @@ uint16_t PatternMatchDefaultMatcher(void)
             }
         }
 
+        if (strcmp(mpm_algo, "hs") == 0) {
```
jlucovsky, line 646: Let me suggest an alternative
```

#ifndef BUILD_HYPERSCAN
     if (strcmp(mpm_algo, "hs") == 0) {
         FatalError(<message>);
    }
#endif

inashivb: This is indeed much better. Thank you!

inashivb:
Replaced with #5119

PR #5119, opened 2020-07-01
hyperscan: better error message if not compiled

inashivb:


src/detect-engine-mpm.c
```diff
@@ -643,6 +643,12 @@ uint16_t PatternMatchDefaultMatcher(void)
             }
         }
 
+#ifndef BUILD_HYPERSCAN
+    if (strcmp(mpm_algo, "hs") == 0) {
+        FatalError(SC_ERR_INVALID_VALUE, "Hyperscan (hs) support is "
```
victorjulien, line 648: this error provides no guidance or even enough context for users with little experience. Would suggest adding that it is about the mpm-algo option value.

src/util-spm.c
```diff
@@ -83,6 +83,12 @@ uint16_t SinglePatternMatchDefaultMatcher(void)
             }
         }
 
+#ifndef BUILD_HYPERSCAN
+        if (strcmp(spm_algo, "hs") == 0) {
+            FatalError(SC_ERR_INVALID_VALUE, "Hyperscan (hs) support is "
```
victorjulien, line 88: same but for spm-algo

src/detect-engine-mpm.c
inashivb: > this error provides no guidance or even enough context for users with little experience. Would suggest adding that it is about the mpm-algo option value.

That's correct. I ran it in dev mode and could see the file names so missed it.

How is `"Hyperscan (hs) support for mpm-algo is not compiled into Suricata"`?

jlucovsky: Add "Add`-with-libhs-includes and -with-libhs-libraries to configure."?

inashivb:
Replaced with #5149

PR #5131, opened 2020-07-06
Plugin support - v7

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/4873

Changes from last PR:
- Drop the function that is called on registration, instead look for a variable named `PluginSpec` which should be a `SCPlugin` instance which defines the plugin and what to do next. Reason: This seems to be more common after looking at other plugins.
- Add support for "source" (input) plugins.
- Include example of PF_RING converted to a plugin.

Redmine ticket:


jasonish:
Continued at https://github.com/OISF/suricata/pull/5150

PR #5134, opened 2020-07-06
app-layer-protocol: match traffic with ALPROTO_UNKNOWN to failed

JoshuaLumb:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2858

Describe changes:
- Allows use of app-layer-protocol:failed to match traffic with unknown app-layer protocols.
- Packets with unknown app layer protocols will now also match under negation conditions of known protocols - e.g. app-layer-protocol:!tls;

### DIFF
--- a/doc/userguide/rules/app-layer.rst
+++ b/doc/userguide/rules/app-layer.rst
@@ -18,7 +18,9 @@ Examples::
 
 A special value 'failed' can be used for matching on flows in which
 protocol detection failed. This can happen if Suricata doesn't know
-the protocol or when certain 'bail out' conditions happen.
+the protocol or when certain 'bail out' conditions happen. '!failed'
+will match on any flow in which detection of a known protocol was
+successful.
 
 .. _proto-detect-bail-out:
 
--- a/src/detect-app-layer-protocol.c
+++ b/src/detect-app-layer-protocol.c
@@ -75,6 +75,12 @@ static int DetectAppLayerProtocolPacketMatch(
 
         r = (data->negated) ? (f->alproto_tc != data->alproto) :
             (f->alproto_tc == data->alproto);
+    } else if (f->alproto == ALPROTO_UNKNOWN)
+    {
+        /* If rule specifies app-layer-protocol:failed match on unknown as well. */
+        SCLogDebug("packet %"PRIu64" unknown alproto, trying to match alproto_failed", p->pcap_cnt);
+        r = (data->alproto == ALPROTO_FAILED) ? !data->negated : 
+            data->negated;
     }
     else {
         SCLogDebug("packet %"PRIu64": default case: direction %02x, approtos %u/%u/%u",


catenacyber:
Thanks for this contribution.

Have you been looking into why the flow is classified as `ALPROTO_UNKNOWN` instead of `ALPROTO_FAILED ?
ie why it does not reach 
https://github.com/OISF/suricata/blob/master/src/app-layer.c#L548
`*alproto = ALPROTO_FAILED` after condition ` if (FLOW_IS_PM_DONE(f, flags) && FLOW_IS_PP_DONE(f, flags)) {` in `TCPProtoDetect` function ?

JoshuaLumb:
> Thanks for this contribution.
> 
> Have you been looking into why the flow is classified as `ALPROTO_UNKNOWN` instead of `ALPROTO_FAILED ? ie why it does not reach https://github.com/OISF/suricata/blob/master/src/app-layer.c#L548 `*alproto = ALPROTO_FAILED`after condition` if (FLOW_IS_PM_DONE(f, flags) && FLOW_IS_PP_DONE(f, flags)) {`in`TCPProtoDetect` function ?

The assignment `*alproto = ALPROTO_FAILED` will only be reached when `*alproto_otherdir != ALPROTO_UNKNOWN`. Since the user submitting [2858](https://redmine.openinfosecfoundation.org/issues/2858) describes filtering out malware that uses a "custom protocol" it's possible we would fail to match it in either direction, leading to the ultimate return of `ALPROTO_UNKNOWN`.

catenacyber:
Thanks @JoshuaLumb 

> The assignment *alproto = ALPROTO_FAILED will only be reached when *alproto_otherdir != ALPROTO_UNKNOWN

Indeed, so does `else` part with `TCPProtoDetectCheckBailConditions` get called at line https://github.com/OISF/suricata/blob/master/src/app-layer.c#L556 and ends up in calling `DisableAppLayer` at 
https://github.com/OISF/suricata/blob/master/src/app-layer.c#L272 ?

That is what I would expect (but something wrong must prevent it from happening)

JoshuaLumb:
> Indeed, so does `else` part with `TCPProtoDetectCheckBailConditions` get called at line https://github.com/OISF/suricata/blob/master/src/app-layer.c#L556 and ends up in calling `DisableAppLayer` at
> https://github.com/OISF/suricata/blob/master/src/app-layer.c#L272 ?
> 
> That is what I would expect (but something wrong must prevent it from happening)

I'm thrilled you pointed me down this path @catenacyber - thanks!

It turns out that by using certain keywords in their signature the user hamstrung the functionality they wanted from TCP stream reconstruction. Removing `flow:no_stream` resolves this and I think it's safe to close this PR. 

One strange thing arose though: using the "depth" keyword doesn't behave as I expected when retrieving payload data from a reconstructed stream. Instead of searching to a given depth in a packet's payload the search is conducted to a given depth in the entire stream. So in this case to find "ls" "pwd" and "whoami" as the user is trying to do the specified depth must be at least 3, 7, and 20 respectively. What do you think - is this strange behaviour?

catenacyber:
Glad I could help

PR #5136, opened 2020-07-06
Generate Rust documentation - v1

jasonish:
A trivial example of generating rustdoc.  In the `rust` directory a `make doc` will output HTML documentation in `target/doc`.

But in the more general documentation topic, I'm thinking of a script (or a makefile target) that would generate the rustdoc, doxygen doc, userguide and devguide all in html, and dump them in a common directory.  Could even go so far as starting a small http server to serve up the docs for local usage.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3363


victorjulien:
Is this meant for merge? Confused by "a trivial example".

Wrt a script, I think that'd make sense.

victorjulien:
Merged in #5188, thanks Jason.

PR #5137, opened 2020-07-07
32bit pcap counters can wrap-around - v2

roligugus:
Previous PR: #5061

Changes from last PR:
- Moved tests into tests/source-pcap.c as requested. Rejigged test data.
- Made PcapThreadVars' pkts also 64bits. It gets logged in ReceivePcapThreadExitStats()
- Removed PcapThreadVars' errs as it was not used at all.
- Moved last_stats64 to end of PcapThreadVars. Should there be any cache line discussion?

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/2845

### DIFF
--- a/src/source-pcap.c
+++ b/src/source-pcap.c
@@ -48,6 +48,22 @@
 #define PCAP_RECONNECT_TIMEOUT 500000
 
 /**
+ * \brief 64bit pcap stats counters.
+ *
+ * libpcap only supports 32bit counters. They will eventually wrap around.
+ *
+ * Keep track of libpcap counters as 64bit counters to keep on counting even
+ * if libpcap's 32bit counters wrap around.
+ * Requires pcap_stats() to be called before 32bit stats wrap around twice,
+ * which we do.
+ */
+typedef struct PcapStats64_ {
+    uint64_t ps_recv;
+    uint64_t ps_drop;
+    uint64_t ps_ifdrop;
+} PcapStats64;
+
+/**
  * \brief Structure to hold thread specific variables.
  */
 typedef struct PcapThreadVars_
@@ -67,9 +83,8 @@ typedef struct PcapThreadVars_
     int datalink;
 
     /* counters */
-    uint32_t pkts;
+    uint64_t pkts;
     uint64_t bytes;
-    uint32_t errs;
 
     uint16_t capture_kernel_packets;
     uint16_t capture_kernel_drops;
@@ -88,6 +103,8 @@ typedef struct PcapThreadVars_
     ChecksumValidationMode checksum_mode;
 
     LiveDevice *livedev;
+
+    PcapStats64 last_stats64;
 } PcapThreadVars;
 
 static TmEcode ReceivePcapThreadInit(ThreadVars *, const void *, void **);
@@ -99,12 +116,14 @@ static TmEcode DecodePcapThreadInit(ThreadVars *, const void *, void **);
 static TmEcode DecodePcapThreadDeinit(ThreadVars *tv, void *data);
 static TmEcode DecodePcap(ThreadVars *, Packet *, void *);
 
+void SourcePcapRegisterTests(void);
+
 /** protect pcap_compile and pcap_setfilter, as they are not thread safe:
  *  http://seclists.org/tcpdump/2009/q1/62 */
 static SCMutex pcap_bpf_compile_lock = SCMUTEX_INITIALIZER;
 
 /**
- * \brief Registration Function for RecievePcap.
+ * \brief Registration Function for ReceivePcap.
  */
 void TmModuleReceivePcapRegister (void)
 {
@@ -115,6 +134,7 @@ void TmModuleReceivePcapRegister (void)
     tmm_modules[TMM_RECEIVEPCAP].ThreadExitPrintStats = ReceivePcapThreadExitStats;
     tmm_modules[TMM_RECEIVEPCAP].cap_flags = SC_CAP_NET_RAW;
     tmm_modules[TMM_RECEIVEPCAP].flags = TM_FLAG_RECEIVE_TM;
+    tmm_modules[TMM_RECEIVEPCAP].RegisterTests = SourcePcapRegisterTests;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/tests/source-pcap.c
```diff
@@ -0,0 +1,232 @@
...
+    u_int current; /* 32bit pcap stat */
+} TestData;
+
+static int UpdatePcapStatsValue64NoChange01(void)
+{
+    /*
+     * No change in counter values.
+     * Last count is within first 32bit range, i.e. same as pcap_stat range.
+     */
+    TestData data[] = {{.last = 0, .current = 0},
```
roligugus, line 44: Same tests as before, but I rejigged test data for last and current to be paired up. Find that easier than the previous two separate arrays.

src/source-pcap.c
```diff
@@ -508,11 +566,11 @@ static void ReceivePcapThreadExitStats(ThreadVars *tv, void *data)
     if (pcap_stats(ptv->pcap_handle, &pcap_s) < 0) {
         SCLogError(SC_ERR_STAT,"(%s) Failed to get pcap_stats: %s",
                 tv->name, pcap_geterr(ptv->pcap_handle));
-        SCLogInfo("(%s) Packets %" PRIu32 ", bytes %" PRIu64 "",
-                tv->name, ptv->pkts, ptv->bytes);
+        SCLogInfo("(%s) Packets %" PRIu64 ", bytes %" PRIu64 "", tv->name,
```
roligugus, line 569: Adjusted ScLogInfo as pkts is now 64bits.

Sorry for the other formatting change. Was using git clang-format to format the patch which adjusts to 80 chars length.

src/source-pcap.c
```diff
@@ -129,14 +149,52 @@ void TmModuleDecodePcapRegister (void)
...
 }
 
+/**
+ * \brief Update 64 bit |last| value from |current32| value taking one
+ * wrap-around into account.
+ */
+static inline void UpdatePcapStatsValue64(uint64_t *last, uint32_t current32)
+{
+    /* uint64_t -> uint32_t is defined behaviour. It slices lower 32bits. */
+    uint32_t last32 = *last;
```
roligugus, line 159: @victorjulien I added a comment about "being defined behaviour". As mentioned in the previous review comment, this is defined behaviour as it's unsigned -> unsigned (smaller).

Does that help? Or would you like me to add masking and casting?

```
    // this is the same
    uint32_t last32 = (uint32_t)(*last & (uint64_t)UINT32_MAX); // or 0xFFFFFFFFull instead of UINT32_MAX
```

src/source-pcap.c
```diff
@@ -67,9 +83,8 @@ typedef struct PcapThreadVars_
     int datalink;
 
     /* counters */
-    uint32_t pkts;
+    uint64_t pkts;
     uint64_t bytes;
-    uint32_t errs;
```
roligugus, line 72: Never used. Removed it.

src/source-pcap.c
```diff
@@ -67,9 +83,8 @@ typedef struct PcapThreadVars_
     int datalink;
 
     /* counters */
-    uint32_t pkts;
```
roligugus, line 70: Also track the thread-local pkts count as 64bits.

src/source-pcap.c
```diff
@@ -88,6 +103,8 @@ typedef struct PcapThreadVars_
     ChecksumValidationMode checksum_mode;
 
     LiveDevice *livedev;
+
+    PcapStats64 last_stats64;
```
roligugus, line 107: Simply moved the stats to the end instead of having them in the middle.

Food for thought:
A better set-up might be something like that where we'd try to stuff all the fields used in PcapCallbackLoop() into the same cache line? Assumption typical cache line of 64 bytes for Intel/AMD. Having said that, this will most likely  not make a huge difference.
```
typedef struct PcapThreadVars_
{
    /* 1ST CACHE LINE - data used in PcapCallbackLoop() */
    /* thread specific handle */
    pcap_t *pcap_handle;

    time_t last_stats_dump;

    LiveDevice *livedev;
    ThreadVars *tv;
    TmSlot *slot;

    /* counters */
    uint64_t bytes;
    uint64_t pkts;
 
    /* data link type for the thread */
    int datalink;
    
    ChecksumValidationMode checksum_mode;

    /* 2ND CACHE LINE */
    PcapStats64 last_stats64;
    uint16_t capture_kernel_packets;
    uint16_t capture_kernel_drops;
    uint16_t capture_kernel_ifdrops;

    /* ptr to string from config */
    const char *bpf_filter;

    /* thread specific bpf - pcap_setfilter() makes a copy into pcap_t */
    struct bpf_program filter;

    /** callback result -- set if one of the thread module failed. */
    int cb_result;

    /* handle state */
    unsigned char pcap_state;

    /* 3RD CACHE LINE */
    /* pcap buffer size */
    int pcap_buffer_size;
    int pcap_snaplen;
} PcapThreadVars;

```

src/source-pcap.c
victorjulien: No, please keep it as is. The comment is helpful for ppl like me who depend on ubsan to know what undefined behavior is :)

src/source-pcap.c
victorjulien: If you have a setup to properly test, I'm open to reshuffling. Also, we should consider the places where we count packets. I think I count 3 here: in `livedev`, in `pkts`/`bytes` and then in the stats api through `capture_kernel_*`. And then we have of course the intermediate `last_stats64` that you're introducing (for good reason). Thats a lot of stats keeping. W/o looking into why we do all this, my gut feel is that we can reduce some of this.

roligugus: Yeah, at this time I don't have a great setup at all as I am working from home. Hence, it's included as "food for thought" comment only to redo this. It should help in theory, but that needs verification. Throughput testing might not even show any improvement as all the other stuff being done with the packet might make such a change disappear in the measurement noise. I might cachegrind this later this week which should prove that it's an improvement. Until this is verified, it's probably not worth touching, imho.

While ``last_stats64`` is used together with some of the data in other cache lines, it's only updated once a second so the impact should not be too bad not getting this perfect.

However, I realized that by putting ``last_stats64`` at the end it currently straddles a cache line which we should avoid - sizeof(PcapThreadVars) is 120 before this change. I would recommend to move some config data only used in ReceivePcapThreadInit() to the end to avoid that as a minimum. This will make ``last_stats64`` completely fit into the 2nd cache line which will be better. I.e.
```
    ...
    ChecksumValidationMode checksum_mode;

    LiveDevice *livedev;

    PcapStats64 last_stats64;

    /* 3RD CACHE LINE - data only used once during ReceivePcapThreadInit() */
    /* ptr to string from config */
    const char *bpf_filter;

    /* pcap buffer size */
    int pcap_buffer_size;
    int pcap_snaplen;
} PcapThreadVars;
```

Thoughts?

roligugus: Yeah, I saw that there's a multitude of stats counters. Was wondering why there are local counters and ``livedev`` counters as well. Did not want to touch that with a ten-foot pole to not change current behaviour. I agree with your gut feel. ;)

For what it's worth, I have a version of libpcap directly providing a 64bit stats API that I will try to get upstream. That would clean the ``last_stats64`` up again if that API was detected. However, that would be a longer term change and it needs to be accepted first as well.

roligugus: @victorjulien Sorry, running short on time to do all the things on my plate and coordinate guys at work. I plan to open another version of this pull request to move the ```PcapStats64 last_stats64``` into the 2nd cache line at least. I think then that'll be ready for merge or more changes if you agree.

I can open another pull request with the reorg of the structure to be more cache friendly  for tracking purposes and to be able to pick that up another time.

roligugus: @victorjulien I can also open another ticket to revisit all the different stats collected.

roligugus: Get rid of one ticket, gain two more? At least it's progress, eh.

victorjulien:
Merged as part of #5188, thanks!

src/source-pcap.c
roligugus: Just for tracking purposes. Opened:
- Counter usages: https://redmine.openinfosecfoundation.org/issues/3829
- Cache lines: https://redmine.openinfosecfoundation.org/issues/3830

PR #5150, opened 2020-07-08
Plugin support - v9

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5131

Changes from last PR:
- Fix source plugins by giving them a change to register
  a live device.

Redmine ticket:


examples/plugins/eve-file-type/Makefile
```diff
@@ -0,0 +1,10 @@
+SRCS :=		template.c
+
+CPPFLAGS +=	-DSURICATA_PLUGIN -I.
+CPPFLAGS +=	-I../../../src
+
+all:
+	gcc -fPIC $(CPPFLAGS) -o eve-filetype.so -shared $(SRCS)
```
victorjulien, line 7: should this be `$(CC)`?

configure.ac
```diff
@@ -2775,6 +2798,8 @@ SURICATA_BUILD_CONF="Suricata Configuration:
   Profiling enabled:                       ${enable_profiling}
   Profiling locks enabled:                 ${enable_profiling_locks}
 
+  Plugin support:                          ${plugin_support}
```
victorjulien, line 2801: Suggest: `Plugin support (experimental):`

examples/plugins/pf-ring/Makefile
```diff
@@ -0,0 +1,17 @@
...
+		runmode-pfring.c
+
+CPPFLAGS +=	-DHAVE_PF_RING_FLOW_OFFLOAD
+CPPFLAGS +=	-DSURICATA_PLUGIN -I. 
+
+# Specify the path to the Suricata source code.
+CPPFLAGS +=	-I../../../src
+
+all:
+	gcc -fPIC $(CPPFLAGS) -o $(NAME).so -shared $(SRCS)
```
victorjulien, line 14: CC

examples/plugins/pf-ring/plugin.c
```diff
@@ -0,0 +1,35 @@
...
+    LiveBuildDeviceList("pfring");
+    RunModeIdsPfringPluginRegister(plugin_slot);
+    TmModuleReceivePfringPluginRegister(receive_slot);
+    TmModuleDecodePfringPluginRegister(decode_slot);
+}
+
+void SCPluginInit(void)
+{
+    SCLogNotice("SCPluginInit");
+    SourcePlugin *plugin = SCCalloc(1, sizeof(SourcePlugin));
```
victorjulien, line 20: we use 'source' in suri, but I wonder if 'capture' might make more sense to ppl

src/decode.h
```diff
@@ -479,6 +479,9 @@ typedef struct Packet_
         WinDivertPacketVars windivert_v;
 #endif /* WINDIVERT */
 
+        /* A chunk of memory that a plugin can use for its packet vars. */
+        uint8_t plugin_v[64];
```
victorjulien, line 483: use a define for the size instead, so the plugin can use that?

src/output-json.c
```diff
@@ -1536,9 +1539,15 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
                 exit(EXIT_FAILURE);
 #endif
             } else {
-                SCLogError(SC_ERR_INVALID_ARGUMENT,
-                           "Invalid JSON output option: %s", output_s);
-                exit(EXIT_FAILURE);
+                SCPluginFileType *plugin = SCPluginFindFileType(output_s);
+                if (plugin == NULL) {
+                    SCLogError(SC_ERR_INVALID_ARGUMENT,
```
victorjulien, line 1544: `FatalError`.

configure.ac
```diff
@@ -410,6 +410,29 @@
         AC_SUBST(SECLDFLAGS)
     ])
 
+    #check for plugin support
+    AC_CHECK_HEADERS([dlfcn.h])
+    AC_MSG_CHECKING([for plugin support])
+    TMPLDFLAGS="${LDFLAGS}"
+    LDFLAGS="${LDFLAGS} -rdynamic"
```
victorjulien, line 417: what is -rdynamic used for?

src/util-plugin.c
```diff
@@ -0,0 +1,178 @@
...
+static void InitPlugin(char *path)
+{
+    void *lib = dlopen(path, RTLD_NOW);
+    if (lib == NULL) {
+        SCLogNotice("Failed to open %s as a plugin: %s", path, dlerror());
+    } else {
+        SCLogNotice("Loading plugin %s", path);
+        SCPlugin *plugin = dlsym(lib, "PluginSpec");
+        if (plugin == NULL) {
+            SCLogError(SC_ERR_PLUGIN, "Plugin does not export a plugin specification: %s", path);
```
victorjulien, line 40: should this be fatal?

src/suricata.h
```diff
@@ -158,6 +158,9 @@ typedef struct SCInstance_ {
     const char *progname; /**< pointer to argv[0] */
     const char *conf_filename;
     char *strict_rule_parsing_string;
+
+    char *source_plugin_name;
```
victorjulien, line 162: can be `const char *`?

src/suricata.c
```diff
@@ -1291,6 +1296,13 @@ static TmEcode ParseCommandLine(int argc, char** argv, SCInstance *suri)
                 return TM_ECODE_FAILED;
 #endif /* HAVE_PFRING */
             }
+            else if(strcmp((long_opts[option_index]).name , "source-plugin") == 0){
```
victorjulien, line 1299: nit style: `) {`

src/suricata-plugin.h
```diff
@@ -0,0 +1,67 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#ifndef __SURICATA_PLUGIN_H__
+#define __SURICATA_PLUGIN_H__
+
+#include "config.h"
+
+#ifdef SURICATA_PLUGIN
```
victorjulien, line 23: why is this needed?

src/suricata-common.h
```diff
@@ -489,13 +489,17 @@ typedef enum {
...
 #include "util-optimize.h"
+#ifndef SURICATA_PLUGIN
 #include <htp/htp.h>
+#endif
 #include "threads.h"
 #include "tm-threads-common.h"
 #include "util-debug.h"
 #include "util-error.h"
 #include "util-mem.h"
+#ifndef SURICATA_PLUGIN
```
victorjulien, line 500: this looks hacky. Why is this needed?

src/suricata-plugin.h
```diff
@@ -0,0 +1,67 @@
...
+} SCPlugin;
+
+/**
+ * Structure used to define a file type plugin.
+ * 
+ * Currently only used by the Eve output type.
+ */
+typedef struct SCPluginFileType_ {
+    char *name;
+    int (*Open)(ConfNode *conf, void **data);
```
victorjulien, line 50: should we have shared function pointer definitions to make sure they can't diverge?

src/suricata-plugin.h
```diff
@@ -0,0 +1,67 @@
...
+    int (*Write)(const char *buffer, int buffer_len, void *ctx);
+    void (*Close)(void *ctx);
+    TAILQ_ENTRY(SCPluginFileType_) entries;
+} SCPluginFileType;
+
+bool SCPluginRegisterFileType(SCPluginFileType *);
+
+typedef struct SourcePlugin_ {
+    char *name;
+    void (*Init)(const char *args, int plugin_slot, int receive_slot, int decode_slot);
```
victorjulien, line 60: if we only register as 'PLUGIN' types, why do we need these slots?

configure.ac
```diff
@@ -1,6 +1,6 @@
     AC_INIT([suricata],[6.0.0-dev])
     m4_ifndef([AM_SILENT_RULES], [m4_define([AM_SILENT_RULES],[])])AM_SILENT_RULES([yes])
-    AC_CONFIG_HEADERS([config.h])
+    AC_CONFIG_HEADERS([src/config.h])
```
victorjulien, line 3: should we rename to `src/autocnf.c` or similar?

src/suricata-plugin.h
jasonish: I think this should be temporary until the includes that are required by the plugins can be cleanly separated from the includes required by Suricata proper. This specific instance was to save plugins from needing to specify where to have the Lua and NSS headers if they were not picked up by default -- something a plugin really shouldn't have to worry about. However, I think I resolved this in this version of the PR so this instance of this can be removed.

src/suricata-common.h
jasonish: Likewise, this is causing an include problem when used from the plugins.  Just some header cleanup required still.  Right now I am sharing "suricata-plugin.h" between plugins and internal code.  But we're probably better off exposing what needs to be exposed to plugins in "suricata-plugin.h", and then an internal one for Suricata proper, with some shared stuff in its own header.  Should clean this up.

src/suricata-plugin.h
jasonish: This is to handle the case where multiple capture plugins may be provided (in the load path), but we only want to activate one.  The "slot" will always be the 'PLUGIN' one, but this saves the plugin from hardcoding that array ID.  So only when a capture plugin is going to be used, will it be called to complete its registration.

configure.ac
jasonish: I'd go the all the way to `src/autoconf.h`.

configure.ac
jasonish: This is what allows the plugin to call functions inside Suricata proper.  There are other options which include:
- Break out functionality used by Suricata and plugins into a shared library. Plugins and Suricata would link with that. Snort's libdaq is like this.
- Register every symbol to export in a text file (example: https://github.com/jasonish/suricata/blob/plugins/v1/src/pluginsyms.txt)
- Provide every function in a struct of function pointers passed to the plugin on load.

src/util-plugin.c
jasonish: I don't see it as  fatal.  The plugin will fail to load tho. It would most likely be a result of a non-plugin DSO being loaded.

configure.ac
victorjulien: I've seen libs use a code annotation (using some macro like `PUBLIC` or `API`). Maybe that is used to generate a `pluginsyms.txt`?

jasonish: I think that is used in combination with -rdynamic to hide some symbols.  You set a default visibility, then either hide or expose.  In one example I looked at the macro became a no-op on Linux, but expanded to something that is required on Windows - something I have it looked at yet as plugins on Windows will require Suricata-core to be a shared dll from what I can tell so far.  So it has a dependency on libsuricata.

I do need to explore this area a little more.  I think we can reasonably determine what to expose for useful capture plugins, as well as file outputs.

configure.ac
victorjulien: Should we move generated headers into a `gen` style directory? Think we have a build-info.h as well.

jasonish: > Should we move generated headers into a `gen` style directory? Think we have a build-info.h as well.

I don't think so..  Ultimately, to build plugins without the Suricata source installed we need to install the headers to /usr/include/suricata or something like that.  This holds true for libsuricata as well. I'm afraid of this complicating the process until we know more.

configure.ac
victorjulien: I think I'm edging towards limiting what we expose as opposed to defaulting to all. Even if the plugin support will be experimental for some time I think its good to already get experience with defining and maintaining a stable API, even if we'll allow for breakage during the test phase.

jasonish: Ok, I'll look at doing that.  I was playing around with -fvisibility=hidden, where we'd then use a macro to help define the function as public. This seems to affect the actual plugin, where it hides the function from Suricata - not really what we want. But doesn't have an affect on Suricata itself.

Looking at Pidgin, you do see this macro type stuff used on the API that plugins call.  Same with Snort's DAQ.  But in both these cases, the plugin is linking against a shared library, and the main program also links against that library.  So we could probably get there with libsuricata, but not now.

So we have 2 ways to control this now:
- Use -rdynamic, but control the exposed/stable API via headers and documentation.
- Or enforce what symbols are exposed - this really still requires the above, but denies a plugin from calling other functions.

src/suricata-plugin.h
jasonish: I'm not clear where they would diverge.  The implementor will get a warning if they don't set the pointer to a matching type, and we only declare them in a single place.

examples/plugins/pf-ring/plugin.c
jasonish: Capture is more clear.  Will move to that.

src/suricata-common.h
jasonish: Took a stab at fixing this without much luck.  For example, the pf-ring capture plugin needs to pull in decode.h, which pulls in a ton of stuff, re-pulling in suricata-common.h and what not resulting in "sechash.h" being pulled in which in many cases requires some paths setup. So hopefully this is temporary, until we untangle some of the includes, and/or provide some pkg-config like tool that will output compiler flags to successfully compile the plugin.

jasonish:
Continued at https://github.com/OISF/suricata/pull/5267.

PR #5151, opened 2020-07-09
hyperscan: better error message if not compiled

inashivb:


victorjulien:
Merged in #5164, thanks!

PR #5158, opened 2020-07-09
Pcap dcerpc/v1

inashivb:


rust/src/dcerpc/dcerpc.rs
```diff
@@ -867,9 +867,9 @@ impl DCERPCState {
         let fraglen = self.get_hdr_fraglen().unwrap_or(0);
 
         if (buffer.len() as u16) < fraglen {
-            SCLogDebug!("Possibly fragmented data, waiting for more..");
-            self.extend_buffer(buffer, direction);
-            return AppLayerResult::ok();
+            panic!("Possibly fragmented data, waiting for more..");
```
victorjulien, line 870: I don't think this is what I mean. Data can flow in in chunks based on how data is ack'd and all, but it may still be a single DCERPC fragment. What I think we are looking for is DCERPC stub data that is transfered in multiple DCERPC fragments (not TCP segments).

inashivb: Oh OK. Thank you for clarifying. Sending the updated PR.

inashivb:
Replaced with #5187

PR #5159, opened 2020-07-09
Output Capture: Capture session on alert v2

scottfgjordan:

Describe changes:
- This basic functionality should come at no expense when not enabled. 
- In this initial version, the tcp-segment enrichment is always occurring, regardless of whether packet capture is enabled or disabled. 

Suggested Review Order/Guide:
- log-pcap-stream.c/.h
- detect-tag-pcap.c/h
- stream-tcp .c/.h files
- util-path.c
   - The only change here is an added function that undots a temporarily hidden file. This was originally being used detect-tag-pcap.c as a way to prevent incomplete captures from being handled, but was moved as it could have some general use. 

Previous Pull Request: https://github.com/OISF/suricata/pull/5128#partial-pull-merging

**Changes from capture-feature-120-v1:**
- I've fixed up the bulk of the qa build issues. The only one remaining is a build issue with CentOS 6 where the forward declarations I've used for the Signature and TcpSession structs in detect-tag-pcap.h cause an error. It seems like this is due to an older version of gcc being used and the -pedantic flag. Before I make a fix to this I was wondering what your thoughts/guidance might be. I.e. are forward declarations avoided because of this?
- I've cleaned up some of the smaller fixes from the code review for v1, but haven't done any significant substance changes.
- I'll be working on the more substantive changes for v3, i.e. reducing the overhead when not enabled and moving away from the allocs as mentioned in v1 PR. I am also working on putting together some suricata verify cases for the .pcap output.
- Mainly with this V2 I just wanted to get a version out that passes the build tests (except for the aforementioned CentOS 6 case) so you could test things out if you wanted.

### DIFF
--- a/src/detect-tag-pcap.c
+++ b/src/detect-tag-pcap.c
@@ -1,0 +1,340 @@
+/* Copyright (C) 2007-2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#include "suricata-common.h"
+#include "detect-tag-pcap.h"
+#include "decode.h"
+#include "util-path.h"
+#include "util-time.h"
+#include "stream-tcp-private.h"
+#include <pthread.h>
+#include <errno.h>
+
+#ifdef __APPLE__
+#ifndef HOST_NAME_MAX
+#define HOST_NAME_MAX 255
+#endif
+#endif
+
+/* PCAP_SNAPLEN (snapshot length) is the amount of data available per capture.
+ * The default is 262144 bytes. Setting the snaplen to 0 will set it to
+ * this default value, but also allow for backwards compatibility for older
+ * versions of tcpdump. */
+#define PCAP_SNAPLEN 0
+
+typedef struct OutputConfig_ {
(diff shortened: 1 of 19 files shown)


src/detect-tag-pcap.c
```diff
@@ -0,0 +1,340 @@
...
+ * \brief Searches through the RB trees of tcp segments in both directions
+ *  (to server and to client) for the tcp segment with the specified pcap_cnt
+ *  value. If found, the segment is logged to output pcap file.
+ * \param session to be dumped to pcap.
+ * \param dump_handle pcap_dumper location.
+ * \param pcap_cnt A value added to tcp segment to keep track of the original
+ *  packet processing order.
+ * \return The tcp segment if it exists. Otherwise NULL.
+ */
+static TcpSegment *FindAndLogTcpSegment(TcpSession *session, pcap_dumper_t
```
victorjulien, line 299: I don't understand this logic of finding the oldest and then searching. Can we just walk using TCPSEG_RB_NEXT? That avoids a search for each segment. If the order needs to be preserved you should be able to walk both in parallel and use pcap_cnt to dump and progress the correct direction first?

src/log-pcap-stream.c
```diff
@@ -0,0 +1,309 @@
...
+ *  alerts, the alert file or event message are created and output.
+ *  For multiple packet alerts, the pcap file is created.
+ * \param tv thread-specific variables.
+ * \param td thread data containing the output context.
+ * \param p Pointer to current packet being processed.
+ */
+static void GeneratePcapFiles(ThreadVars *tv, StreamPcapLogThreadData *td,
+        const Packet *p)
+{
+    if (!(p->flags & PKT_HAS_FLOW)) {
```
victorjulien, line 95: it looks like this block cares about tags only, so then it would make sense to use p->flags & PKT_HAS_TAG

victorjulien:
Wrt the CentOS6 issue, I don't remember a specific policy. But as long as we support it we should compile cleanly. Not much guidance to offer, sorry.

scottfgjordan:
> 
> 
> Wrt the CentOS6 issue, I don't remember a specific policy. But as long as we support it we should compile cleanly. Not much guidance to offer, sorry.

No worries! Just thought I'd check with you first. Shouldn't be too big of a deal to correct.

scottfgjordan:
Closing this pull request as an updated version can be found here: https://github.com/OISF/suricata/pull/5217

PR #5165, opened 2020-07-10
Convert jsonbuilder/v1

inashivb:
https://redmine.openinfosecfoundation.org/issues/3756
https://redmine.openinfosecfoundation.org/issues/3754

### DIFF
--- a/rust/src/krb/log.rs
+++ b/rust/src/krb/log.rs
@@ -17,20 +17,18 @@
 
 // written by Pierre Chifflier  <chifflier@wzdftpd.net>
 
-use crate::json::*;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
 use crate::krb::krb5::{KRB5State,KRB5Transaction,test_weak_encryption};
 
-#[no_mangle]
-pub extern "C" fn rs_krb5_log_json_response(_state: &mut KRB5State, tx: &mut KRB5Transaction) -> *mut JsonT
+fn krb5_log_response(jsb: &mut JsonBuilder, tx: &mut KRB5Transaction) -> Result<(), JsonError>
 {
-    let js = Json::object();
     match tx.error_code {
         Some(c) => {
-            js.set_string("msg_type", "KRB_ERROR");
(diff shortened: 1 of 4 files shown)


rust/src/krb/log.rs
```diff
@@ -48,11 +46,19 @@ pub extern "C" fn rs_krb5_log_json_response(_state: &mut KRB5State, tx: &mut KRB
...
-    js.set_string("encryption", &encryption);
-    js.set_boolean("weak_encryption", tx.etype.map_or(false,test_weak_encryption));
-    return js.unwrap();
+    jsb.set_string("cname", &cname)?;
+    jsb.set_string("realm", &realm)?;
+    jsb.set_string("sname", &sname)?;
+    jsb.set_string("encryption", &encryption)?;
+    jsb.set_bool("weak_encryption", tx.etype.map_or(false,test_weak_encryption))?;
+
+    jsb.close()?;
```
victorjulien, line 55: what is closed here?

rust/src/snmp/log.rs
```diff
@@ -37,48 +37,54 @@ fn str_of_pdu_type(t:&PduType) -> Cow<str> {
...
             _           => ()
         }
         match tx.usm {
-            Some(ref s) => js.set_string("usm", s),
+            Some(ref s) => {jsb.set_string("usm", s)?;},
             _           => ()
         }
     }
-    js.unwrap()
+    jsb.close()?;
```
victorjulien, line 81: where is the matching open?

inashivb:
> I fear both have a close too many, can you double check?
> 
> Do we have SV tests that validate the output?

I didn't write any specially for this but the existing ones passed. Should there be more?

rust/src/snmp/log.rs
inashivb: There isn't. I thought I would get a panic for too many closes. So, didn't pay much attention. Fixing.

victorjulien:
The failure will only lead to panic with `--enable-debug-validation` enabled.

inashivb:
Replaced with #5171

PR #5171, opened 2020-07-11
Convert jsonbuilder/v2

inashivb:
https://redmine.openinfosecfoundation.org/issues/3756
https://redmine.openinfosecfoundation.org/issues/3754

### DIFF
--- a/rust/src/snmp/log.rs
+++ b/rust/src/snmp/log.rs
@@ -17,7 +17,7 @@
 
 // written by Pierre Chifflier  <chifflier@wzdftpd.net>
 
-use crate::json::*;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
 use crate::snmp::snmp::{SNMPState,SNMPTransaction};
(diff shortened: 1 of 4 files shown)


rust/src/snmp/log.rs
```diff
@@ -37,48 +37,53 @@ fn str_of_pdu_type(t:&PduType) -> Cow<str> {
...
         match tx.community {
-            Some(ref c) => js.set_string("community", c),
+            Some(ref c) => {jsb.set_string("community", c)?;},
             _           => ()
         }
         match tx.usm {
-            Some(ref s) => js.set_string("usm", s),
+            Some(ref s) => {jsb.set_string("usm", s)?;},
             _           => ()
         }
```
jasonish, line 79: Now that the `?` operator is used here, I think this would be more clean and readable like:
```
        if let Some(usm) = &tx.usm {
            jsb.set_string("usm", usm)?;
        }
```
likewise for the bit above..

inashivb:
Replaced with #5185

PR #5176, opened 2020-07-13
Transform urldecode 2689 v7

catenacyber:

Describe changes:

Modifies #5175 by fixed rebasing for unit tests

suricata-verify-pr: 146


victorjulien:
I think SV PR 146 is outdated, causing failures.

catenacyber:
> I think SV PR 146 is outdated, causing failures.

updated with https://github.com/OISF/suricata-verify/pull/278

victorjulien:
Can you rebase and do a new PR so the QA will run with the new SV PR?

PR #5187, opened 2020-07-15
Pcap dcerpc/v2

inashivb:
Commit to look at: cfd3c6f


catenacyber:
Are you expecting some review @inashivb ?

inashivb:
> Are you expecting some review @inashivb ?

Not exactly. This is for Victor to find a pcap that I can use for testing fragmented PDUs.

catenacyber:
Is it a pcap where

> the minor version number is 1 and the PFC_LAST_FRAG flag is not set

 ?

Do you have a Wireshark filter for this condition ?

PR #5205, opened 2020-07-21
Trailing fix/v1

inashivb:
Fix trailing char check with ByteExtractString.


victorjulien:
Merged in #5207, thanks!

PR #5213, opened 2020-07-24
Transform urldecode 2689 v8

catenacyber:

Describe changes:

Modifies #5176 with rebase as requested (to rerun with updated S-V PR)

suricata-verify-pr: 278


src/detect-transform-urldecode.c
```diff
@@ -0,0 +1,166 @@
+/* Copyright (C) 2019 Open Information Security Foundation
```
victorjulien, line 1: 2020

src/detect-http-client-body.c
```diff
@@ -258,6 +258,7 @@ static InspectionBuffer *HttpClientBodyGetDataCallback(DetectEngineThreadCtx *de
     StreamingBufferGetDataAtOffset(body->sb,
             &data, &data_len, offset);
     InspectionBufferSetup(buffer, data, data_len);
+    InspectionBufferApplyTransforms(buffer, transforms);
```
victorjulien, line 261: I think https://github.com/OISF/suricata/pull/5213/commits/39760336c22f043c060ac36d64d3dc04dd467e9f should be dropped for now. @jlucovsky is working on addressing to issues around applying transforms to HTTP.

src/detect-transform-urldecode.c
```diff
@@ -0,0 +1,166 @@
...
+#include "detect.h"
+#include "detect-engine.h"
+#include "detect-engine-prefilter.h"
+#include "detect-parse.h"
+#include "detect-transform-urldecode.h"
+
+#include "util-unittest.h"
+#include "util-print.h"
+//isxdigit
+#include <ctype.h>
```
victorjulien, line 37: this should be part of `suricata-common.h`.

src/detect-transform-urldecode.c
```diff
@@ -0,0 +1,166 @@
...
+
+static void TransformUrlDecode(InspectionBuffer *buffer, void *options)
+{
+    const uint8_t *input = buffer->inspect;
+    const uint32_t input_len = buffer->inspect_len;
+    uint8_t output[input_len]; // we can only shrink
+
+    uint32_t output_size = BufferUrlDecode(input, input_len, output);
+
+    InspectionBufferCopy(buffer, output, output_size);
```
victorjulien, line 117: can we avoid this step if the buffer is unchanged?

src/detect-transform-urldecode.h
```diff
@@ -0,0 +1,30 @@
+/* Copyright (C) 2019 Open Information Security Foundation
```
victorjulien, line 1: 2020

src/detect-http-client-body.c
catenacyber: ok

src/detect-transform-urldecode.c
catenacyber: Eheh, PR was submitted in 2019 ;-p

src/detect-transform-urldecode.h
catenacyber: ok

src/detect-transform-urldecode.c
catenacyber: indeed

src/detect-transform-urldecode.c
catenacyber: Ok.
This is the same in other transforms.
Should I create a optimization ticket in redmine ?

src/detect-http-client-body.c
catenacyber: S-V test no longer passes without this commit (and this PR does not make much sense either)
@jlucovsky what is next ?

PR #5217, opened 2020-07-25
Capture feature 120 v3

scottfgjordan:

Describe changes:
- This basic functionality should come at no expense when not enabled. 
- In this initial version, the tcp-segment enrichment is always occurring, regardless of whether packet capture is enabled or disabled. 

Suggested Review Order/Guide:
- log-pcap-stream.c/.h
- detect-tag-pcap.c/h
- pcap-helper.c/h
 - This is basically the same as the code for libpcaps pcap_dump(), but rewritten so that it could dump a packet in two parts, e.g. header and payload written separately. I stuck with the codestyle of libpcap for now. If you like this approach I can open up a request with libpcap to bring this into their API. Alternatively, I can rewrite it in Suricata's code style and it can be kept internal to Suricata for now.
- stream-tcp .c/.h files
- util-path.c
   - The only change here is an added function that undots a temporarily hidden file. This was originally being used detect-tag-pcap.c as a way to prevent incomplete captures from being handled, but was moved as it could have some general use. 

Previous Pull Requests: 

**Changes from capture-feature-120-v2:**
- Added in pcap-helper.c/h (also described above in suggested review order/guide) to allow for separate dumping of packet header and payload to remove a malloc in the previous version. This is basically the same as the code for libpcap's pcap_dump(), but rewritten so that it could dump a packet in two parts, e.g. header and payload written separately. I stuck with the codestyle of libpcap for now. If you like this approach I can open up a request with libpcap to bring this into their API. Alternatively, I can rewrite it in Suricata's code style and it can be kept internal to Suricata for now.
- Took your suggestions/comments on how I was handling the RB trees for the TcpSegments. The code is now much more simplified and is all done in one function. Thanks for the feedback on that!
- Corrected the CentOS6 build issue by removing forward declarations and just did a standard include of the .h files necessary.
- Initialized the pkt_hdr member of TcpSegment to start off with an allocation of 64 bytes and removed freeing of them until the Segment pool is cleaned up. If more space is needed for a pkt_hdr then a realloc is used. This helped cut down on the freeing/mallocing throughout the handling of TcpSegments. 
- Generally took away the checks for Ethernet and instead kept it generic. E.g. the data from GET_PKT_DATA(p)  to p->payload is copied into the pkt_hdr member. 
- Corrected (hopefully) a build issue that was failing with Mingyw by defining HOST_NAME_MAX in the case of _WIN32 or _WIN64. 
- For the use of PATH_MAX (4096 bytes), there's an instance for every alert that is captured. To cut down on the memory, I chopped this down to 512 bytes and specified in the suricata.yaml that the path should be less that 255 bytes, leaving the remaining bytes for the hostname and timestamps in the file name.

### DIFF
--- a/src/detect-tag-pcap.c
+++ b/src/detect-tag-pcap.c
@@ -1,0 +1,261 @@
+/* Copyright (C) 2007-2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#include "suricata-common.h"
+#include "detect-tag-pcap.h"
+#include "decode.h"
+#include "pcap-helper.h"
+#include "util-path.h"
+#include "util-time.h"
+#include "stream-tcp-private.h"
+#include <pthread.h>
+#include <errno.h>
+
+#ifndef HOST_NAME_MAX
+#ifdef __APPLE__
+#define HOST_NAME_MAX 255
+#elif defined(_WIN32) || defined(_WIN64)
+#define HOST_NAME_MAX 255
+#endif
+#endif
+
+/* PCAP_SNAPLEN (snapshot length) is the amount of data available per capture.
+ * The default is 262144 bytes. Setting the snaplen to 0 will set it to
+ * this default value, but also allow for backwards compatibility for older
+ * versions of tcpdump. */
+#define PCAP_SNAPLEN 0
+
+typedef struct OutputConfig_ {
+    char g_hostname[HOST_NAME_MAX];
+    const char *g_output_dir;
+} OutputConfig;
+
+OutputConfig g_output_config;
+
+static void DumpTcpSegment(TcpSession *session, TcpSegment *seg,
+                          pcap_dumper_t *dump_handle, bool client);
+
+/**
+ * \brief Initializes filename for output pcap log.
+ * \param output_directory Configurable filename from suricata.yaml.
+ */
+void InitializePcapLogFilenameSupport(const char *output_directory)
+{
+    if(gethostname(g_output_config.g_hostname, sizeof(g_output_config.g_hostname))
+        != 0) {
+        FatalError(SC_ERR_HOST_INIT, "Error looking up hostname in "
+                                     "detect-tag-pcap.c Error: %s",
+                                     strerror(errno));
+    }
+
+    if (output_directory == NULL) {
+        g_output_config.g_output_dir = ConfigGetLogDirectory();
+    } else {
+        g_output_config.g_output_dir = output_directory;
+    }
+}
+
+/**
+ *  \brief Fills the result_path_buf with a full file path that can be used
+ *   to create a file. InitializePcapLogFilenameSupport() must be run before
+ *   this is called.
+ *  \param result_path_buf buffer to hold the path string.
+ *  \param result_buf_size length of the filename buffer used.
+ *  \param p packet pointer.
+ *  \param signature that alerted.
+ *  \param thread_id unique id for the current thread.
+ *  \param unique_id counter incrementing over time to add entropy to filenames.
+ */
+void GenerateStreamFilepath(char *result_path_buf, size_t result_buf_size,
+        const Packet *p, const Signature *sig, int thread_id, uint32_t
+        unique_id)
+{
+    time_t time = p->ts.tv_usec;
+    struct tm local_tm;
+    struct tm *t = SCLocalTime(time, &local_tm);
+    if (unlikely(t == NULL)) {
+        FatalError(SC_ERR_TS, "Unable to create time structure, ts-error");
+    }
+
+    char fmt_time_buf[64];
+    CreateFormattedTimeString(t, "%Y%m%d_%H%M%S.%%06u", fmt_time_buf,
+            sizeof(fmt_time_buf));
+
+    char time_buf[64];
+    int ret = snprintf(time_buf, sizeof(time_buf),
+            fmt_time_buf, p->ts.tv_usec);
+    if (ret < 0 || (size_t) ret >= sizeof(time_buf)) {
+        SCLogError(SC_ERR_INVALID_NUM_BYTES, "Provided buffer size is too small"
+                                             "to create full time buffer."
+                                             " Error: %s", strerror(errno));
+    }
+
+    ret = snprintf(result_path_buf, result_buf_size,"%s/.%u-%u.%s.%u.%s.pcap",
+            g_output_config.g_output_dir, thread_id, unique_id,
+            g_output_config.g_hostname, sig->id, time_buf);
+
+    if (ret < 0 || (size_t)ret >= result_buf_size) {
+        FatalError(SC_ERR_INVALID_NUM_BYTES, "Provided buffer size is too "
+                                             "small to create PCAP file path."
+                                             " Error: %s", strerror(errno));
+    }
+}
+
+/**
+ *  \brief Registration for TaggedPcapEntry when writing the output to files.
+ *  \param p pointer to the packet being alerted on.
+ *  \param sig signature that triggered the alert.
+ *  \param thread_id id of the thread we're on; used for uniqueness in filename.
+ *  \param unique_id unique counter per thread that increments to increase
+ *  the entropy of filenames.
+ */
+TaggedPcapEntry *SetupTaggedPcap(const Packet *p, const Signature *sig, int
+        thread_id, int unique_id)
+{
+     TaggedPcapEntry *tpe = (TaggedPcapEntry*) SCMalloc(sizeof(*tpe));
+     if (unlikely(tpe == NULL)) {
+         return NULL;
+     }
+     tpe->pcap_file_path[0] = '\0';
+
+     GenerateStreamFilepath(tpe->pcap_file_path, PCAP_PATH_MAX, p, sig,
+             thread_id, unique_id);
+
+     tpe->pcap_dead_handle = pcap_open_dead(p->datalink, PCAP_SNAPLEN);
+     if (tpe->pcap_dead_handle == NULL) {
+         SCLogError(SC_ERR_PCAP_OPEN_OFFLINE, "Error opening dead pcap "
+                                              "handle: %s", pcap_geterr
+                                              (tpe->pcap_dead_handle));
+         SCFree(tpe);
+         exit(EXIT_FAILURE);
+     }
+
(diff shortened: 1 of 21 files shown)


src/detect-tag-pcap.h
```diff
@@ -0,0 +1,58 @@
...
+#include "log-pcap-stream.h"
+#include "pcap.h"
+#include "stream-tcp-private.h"
+
+#define PCAP_PATH_MAX 512
+
+typedef struct TaggedPcapEntry_ {
+    pcap_dumper_t *pcap_dumper;
+    pcap_t *pcap_dead_handle;
+    char pcap_file_path[PCAP_PATH_MAX];
```
victorjulien, line 43: AFAICS this is really only used in cleanup, not per packet. I wonder, can we just reconstruct the path then to avoid storing this here? Then we'd just store the much smaller things needed to construct the path here?

src/log-pcap-stream.c
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     *  The flow is locked during this method call. It's safe to read and
+     *  modify session tags.
+     */
+     DetectTagDataEntry *tags = TagGetFlowTag(p->flow);
+     /**
+      *  Log this packet to every tag's output PCAP stream. The detect-tag
+      *  code will handle cleanup and deletion of expired tags.
+      */
+      while (tags != NULL) {
```
victorjulien, line 123: indentation is drunk here :)

src/log-pcap-stream.c
```diff
@@ -0,0 +1,309 @@
...
+ * \brief Determines whether or not to log this packet.
+ * \param tv thread-specific variables.
+ * \param p Pointer to current packet being processed.
+ * \return TRUE if the packet should be logged.
+ * \return FALSE if we do not need to log the packet.
+ */
+static int StreamPcapLogCondition(ThreadVars *tv, const Packet *p)
+{
+    /* Flow is necessary for tag lookups. Reject invalid packets. */
+    if ((p->flags & PKT_HAS_FLOW && !(p->flags & PKT_IS_INVALID)) ||
```
victorjulien, line 191: This can be a single check: `(p->flags & (PKT_HAS_FLOW|PKT_IS_INVALID)) == PKT_HAS_FLOW)`

Not sure if the compiler would produce the same code, so net effect might be 0.

src/detect-tag-pcap.c
```diff
@@ -0,0 +1,261 @@
...
+{
+    struct pcap_pkthdr pcap_hdr;
+    uint32_t packet_len = seg->pktlen;
+    uint32_t payload_len = seg->payload_len;
+    uint32_t packet_header_len = packet_len - payload_len;
+
+    pcap_hdr.ts.tv_sec = seg->ts.tv_sec;
+    pcap_hdr.ts.tv_usec = seg->ts.tv_usec;
+
+    if (seg->pkt_hdr != NULL) {
```
jlucovsky, line 250: Suggest moving this check in modified form to the function entry.
```
if (seg->pkt_hdr == NULL) {
    return;
}
.
.
```

src/detect-tag-pcap.h
```diff
@@ -0,0 +1,58 @@
+/* Copyright (C) 2007-2020 Open Information Security Foundation
```
jlucovsky, line 1: nit: update range.

src/log-pcap-stream.c
```diff
@@ -0,0 +1,309 @@
+/* Copyright (C) 2007-2020 Open Information Security Foundation
```
jlucovsky, line 1: nit: range

src/log-pcap-stream.c
```diff
@@ -0,0 +1,309 @@
...
+            /* Is the alert from a tagged signature? */
+            if (p->alerts.alerts[x].s->sm_arrays[DETECT_SM_LIST_TMATCH] !=
+                NULL) {
+                TaggedPcapEntry *pcap_file = SetupTaggedPcap(p, p->alerts
+                .alerts->s, tv->id, td->count);
+                DumpTaggedPacket(pcap_file->pcap_dumper, p);
+                CleanUpTaggedPcap(pcap_file);
+            }
+        }
+        return;
```
jlucovsky, line 111: nit: `return` is unnecessary.

src/log-pcap-stream.c
```diff
@@ -0,0 +1,309 @@
...
+      }
+}
+
+/**
+ *  \brief StreamPcapLogRegister Register the logger to the output-packet
+ *   root logger.
+ */
+void StreamPcapLogRegister(void)
+{
+    SCLogNotice("StreamPcapLogRegister Enter");
```
jlucovsky, line 169: nit: `SCLogDebug`

src/log-pcap-stream.c
```diff
@@ -0,0 +1,309 @@
...
+    if ((p->flags & PKT_HAS_FLOW && !(p->flags & PKT_IS_INVALID)) ||
+        p->alerts.cnt > 0) {
+        return TRUE;
+    }
+    return FALSE;
+}
+
+/**
+ *  \brief StreamPcapLogThreadInit Initialize the thread data.
+ *  \param initdata Contains the output_ctx created by StreamPcapLogIitCtx
```
jlucovsky, line 200: nit: missing `tv`

src/log-pcap-stream.c
```diff
@@ -0,0 +1,309 @@
...
+{
+    // Create and initialize the thread data.
+    if (initdata == NULL) {
+        SCLogDebug("Error getting context for StreamLogPcap. \"initdata\" "
+                   "argument NULL");
+        return TM_ECODE_FAILED;
+    }
+    StreamOutputCtx *stream_output_ctx = GetStreamOutputCtx((OutputCtx*)
+            initdata);
+    StreamPcapLogThreadData *td = SCMalloc(sizeof(*td));
```
jlucovsky, line 216: suggest `SCCalloc` and performing allocation before calling `GetStreamOutputCtx`

src/pcap-helper.c
```diff
@@ -0,0 +1,75 @@
...
+#include "pcap-helper.h"
+
+void SplitPcapDump(u_char *user, struct pcap_pkthdr *libpcap_hdr,
+        const u_char *p1, bpf_u_int32 p1_len, const u_char *p2,
+        bpf_u_int32 p2_len) {
+    /*
+     * Check for overflow due to packet lengths.
+     */
+    if ((p1_len + p2_len) < p1_len) {
+        fprintf(stderr,"Error in SplitPcapDump. Overflow due to packet "
```
jlucovsky, line 28: `SCLogError` or warning?

src/pcap-helper.h
```diff
@@ -0,0 +1,36 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#ifndef __SURICATA_PCAP_HELPER_H__
+#define __SURICATA_PCAP_HELPER_H__
```
jlucovsky, line 19: nit: `__PCAP_HELPER_H__`

src/stream-tcp-reassemble.c
```diff
@@ -260,6 +260,15 @@ static int TcpSegmentPoolInit(void *data, void *initdata)
      * won't have uninitialized memory to consider. */
     memset(seg, 0, sizeof (TcpSegment));
 
+    /*
+     * Initialize seg->pkt_hdr to hold TCPSEG_PKT_HDR_DEFAULT_SIZE bytes of data
+     * for packet headers. The size of the packet header will be checked before
+     * copying data and seg->pkt_hdr will be grown if necessary.
+     */
+    seg->pkt_hdr = SCMalloc(sizeof(char) * TCPSEG_PKT_HDR_DEFAULT_SIZE);
```
jlucovsky, line 268: `SCCalloc`?

src/util-path.c
```diff
@@ -137,3 +137,51 @@ bool SCPathExists(const char *path)
...
+
+    /*
+     * Shift characters left by one until a '/' is found. The filename is
+     * temporarily hidden, i.e. prefixed with a '.', while packets are being
+     * dumped. This shifting is done to find that '.' and remove it so the
+     * file is no longer hidden.
+     */
+    char last_replaced_char = undotted_path[len];
+    size_t x = len;
+    while (undotted_path[--x] != '/' && x > 0) {
```
jlucovsky, line 174: Windows?

jlucovsky:
Can #5159 be closed?

scottfgjordan:
Thanks for the quick reviews! I'll get working on a V4 and responding to both of your comments shortly. @jlucovsky I will go ahead and close https://github.com/OISF/suricata/pull/5159 right away.

src/detect-tag-pcap.c
```diff
@@ -0,0 +1,261 @@
...
+
+     tpe->pcap_dumper = pcap_dump_open(tpe->pcap_dead_handle,
+             tpe->pcap_file_path);
+     if (tpe->pcap_dumper == NULL) {
+         SCLogError(SC_ERR_PCAP_OPEN_OFFLINE, "Failed to create tag output "
+                                              "file at %s. Error: %s",
+                                              tpe->pcap_file_path,
+                                              pcap_geterr(tpe->pcap_dead_handle));
+         SCFree(tpe);
+         exit(EXIT_FAILURE);
```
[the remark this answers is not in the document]
scottfgjordan: Yes it will. I'm thinking a return of NULL is probably a better idea.

src/log-pcap-stream.c
scottfgjordan: Good catch, thanks!

src/pcap-helper.c
scottfgjordan: I'll switch these to SCLogError. Initially I just wrote this in the same style as libpcap's dump_pcap function. I think I'll try and get this SplitPcapDump function added to their pcap API. For now though, I'll rewrite these pcap-helper files in the same code style as Suricata.

src/stream-tcp-private.h
```diff
@@ -58,12 +58,18 @@ int TcpSackCompare(struct StreamTcpSackRecord *a, struct StreamTcpSackRecord *b)
...
 
+#define TCPSEG_PKT_HDR_DEFAULT_SIZE 64
+
 typedef struct TcpSegment {
     PoolThreadReserved res;
     uint16_t payload_len;       /**< actual size of the payload */
     uint32_t seq;
     RB_ENTRY(TcpSegment) __attribute__((__packed__)) rb;
     StreamingBufferSegment sbseg;
+    struct timeval ts;
```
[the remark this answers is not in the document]
scottfgjordan: I like this idea, seems like a reasonable way to cut out the overhead when not enabled. I'll check out the other implementations you've mentioned and get working on a solution for TcpSegments.

src/detect-tag-pcap.h
scottfgjordan: This makes sense to me. Most of the length in that file name has to do with the output_directory and the potential of it being very long. This is already a global variable in detect-tag-pcap.c. I'll make this adjustment so that the TaggedPcapEntry would only need to contain the things necessary for the filename itself, rather than the absolute path.

scottfgjordan:
Just a quick update for both of you. I've made and tested out most of the changes both of you have requested. I am currently working out the TcpSegment Storage. I've got the packet capture working using the Setting/Getting of TcpSegment storage only and cut down the size of the TcpSegment significantly. Just working on the appropriate cleanup for the storage and performing a few more tests to ensure I've patched any leaks.

scottfgjordan:
Closing this down as v4 is now up.

PR #5220, opened 2020-07-27
pfring: fix compile warning

victorjulien:


src/runmode-pfring.c
```diff
@@ -264,10 +264,13 @@ static void *ParsePfringConfig(const char *iface)
                 }
             }
         } else {
-            if (StringParseUint16(&pfconf->threads, 10, 0, (const char *)threadsstr) < 0) {
+            uint16_t threads = 0;
```
catenacyber, line 267: Out of curiosity, why not change the definition of `threads` in `PfringIfaceConfig` structure ?

victorjulien: Would be good to do. This is just quick to get rid of the warning.

victorjulien:
Replaced by #5234

src/runmode-pfring.c
catenacyber: Ok thanks

PR #5215, opened 2020-07-24
MQTT parser - v9

satta:
Previous PR: #5208 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:  [#3549](https://redmine.openinfosecfoundation.org/issues/3549)

Describe changes (to previous PR):
- Split detection code for SUBSCRIBE/UNSUBSCRIBE into separate independent entities to fix potential bugs caused by trying to use common functionality.
- Fix documentation to correctly list  `mqtt.subscribe.topic` and `mqtt.unsubscribe.topic` as sticky buffers.
- Return value and error directly from `rs_mqtt_cstr_message_code()`.
- Introduce `MQTTTypeCode` enum and associated handling methods. Change message code handling throughout project, from parsing to logging, to use enum instead of explicit ints. This also gives us a better defined set of accepted values than just some range of ints.
- Reject message code 0 in probing parser as it is not valid MQTT.

PR #5229, opened 2020-07-27
MQTT parser - v10

satta:
Previous PR: #5215  

- Pass existing `JsonBuilder` object to function instead of returning a new object.
- Remove occurrences of re-defining `MAX_SUBSTRINGS`.
- Add `if` checks for `NULL` pointers before calling `SCFree()` in MQTT related detect C code.
- Use `SCCalloc()` instead of plain `SCMalloc()` when allocating structs.
- Wrap `RegisterTests` assignments in UNITTESTS `#ifdef`s.


satta:
macOS seems to have a build problem caused by crates.io download errors.

victorjulien:
MacOS build failure looks like a runner failure.

satta:
> MacOS build failure looks like a runner failure.

I think so too and am not going to do anything about it :)

rust/src/mqtt/detect.rs
```diff
@@ -0,0 +1,476 @@
...
+
+#[no_mangle]
+pub extern "C" fn rs_mqtt_tx_get_unsubscribe_topic(tx: &MQTTTransaction,
+    i: u16,
+    buf: *mut *const u8,
+    len: *mut u32)
+    -> u8
+{
+    for msg in tx.msg.iter() {
+        if let MQTTOperation::UNSUBSCRIBE(ref unsubv) = msg.op {
```
catenacyber, line 393: Can not there be more than one `UNSUBSCRIBE` in a transaction ?

catenacyber: This is my last question ;-)

catenacyber:
> I think so too and am not going to do anything about it :)

I launched it again

rust/src/mqtt/detect.rs
satta: Not according to the protocol, i.e. compliant brokers/clients would not create this. Forged or broken traffic might create such a situation if message IDs are used multiple times for separate UNSUBSCRIBE events.
I think I might really need the approach from https://github.com/OISF/suricata/pull/5208#discussion_r459466182 to make sure we get all of their topics if we want this function to work in this case as well (which I think we do!)

satta: See https://github.com/satta/suricata/commit/058c91b79cd43e587a49670eea0bd5e4580aa2ee

catenacyber: This commit looks good to me

satta: Cool, thanks for having a look. I will revisit and review the transaction code again to better define how the message ID ("packet identifier" in the spec) is used to group messages into transactions, in particular regarding ID reuse as addressed in the spec:

> The Packet Identifier becomes available for reuse after the Client has processed the corresponding acknowledgement packet. In the case of a QoS 1 PUBLISH this is the corresponding PUBACK; in the case of QoS 2 it is PUBCOMP. For SUBSCRIBE or UNSUBSCRIBE it is the corresponding SUBACK or UNSUBACK.

This means that we should make sure to never add to a 'completed' transaction where an acknowledgement has been seen. I'll make sure that is the case.

satta:
Next PR: #5248

PR #5235, opened 2020-07-28
transform: adds url_decode keyword

catenacyber:

Describe changes:

Does not make test https://github.com/OISF/suricata-verify/pull/278 pass cf https://github.com/OISF/suricata/pull/5213#discussion_r460776619 

Modifies #5213 with comments taken into account

suricata-verify-pr: 278


catenacyber:
Rather with unit tests disabled ;-)

PR #5239, opened 2020-07-29
datasets: reload static sets

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3760

### DIFF
--- a/src/datasets.c
+++ b/src/datasets.c
@@ -72,7 +72,7 @@ static Dataset *DatasetSearchByName(const char *name)
 {
     Dataset *set = sets;
     while (set) {
-        if (strcasecmp(name, set->name) == 0) {
+        if (strcasecmp(name, set->name) == 0 && set->hidden == false) {
             return set;
         }
         set = set->next;
@@ -476,11 +476,11 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
     set->type = type;
     if (save && strlen(save)) {
         strlcpy(set->save, save, sizeof(set->save));
-        SCLogDebug("name %s save '%s'", name, set->save);
+        SCLogInfo("name %s save '%s'", name, set->save);
     }
     if (load && strlen(load)) {
         strlcpy(set->load, load, sizeof(set->load));
-        SCLogDebug("set \'%s\' loading \'%s\' from \'%s\'", set->name, load, set->load);
+        SCLogInfo("set \'%s\' loading \'%s\' from \'%s\'", set->name, load, set->load);
     }
 
     char cnf_name[128];
@@ -513,8 +513,8 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
             break;
     }
 
-    SCLogDebug("set %p/%s type %u save %s load %s",
-            set, set->name, set->type, set->save, set->load);
+    SCLogInfo("set %p/%s type %u save %s load %s address %p",
+            set, set->name, set->type, set->save, set->load, set);
 
     set->next = sets;
     sets = set;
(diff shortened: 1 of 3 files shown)


src/datasets.c
```diff
@@ -532,6 +532,46 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
...
+            SCLogInfo("It is a static set or not defined in YAML, so we can continue...");
+        } else {
+            continue;
+        }
+        SCLogInfo("Soon to be hidden Dataset name: %s", set->name);
+        SCLogInfo("Soon to be hidden Dataset type: %d", set->type);
+        set->hidden = true;
+        set = set->next;
+    }
+    return 1;
```
inashivb, line 551: Probably don't need to return anything now?

src/datasets.c
```diff
@@ -532,6 +532,46 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
...
+}
+
+void DatasetRemoveCopies(void)
```
inashivb, line 554: Better name?

src/datasets.c
```diff
@@ -532,6 +532,46 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
     return NULL;
 }
 
+int DatasetReload(uint32_t de_vers)
+{
+    Dataset *set = sets;
```
victorjulien, line 537: need to wrap this function in the lock that protects `sets`

src/datasets.c
```diff
@@ -532,6 +532,46 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
...
+    // Reload if its a static set only
+    while (set != NULL) {
+        // Check if the set is static or dynamic
+        if ((set->save == NULL || strcmp(set->save, "") == 0) || (set->from_yaml == false)) {
+            SCLogInfo("It is a static set or not defined in YAML, so we can continue...");
+        } else {
+            continue;
```
victorjulien, line 544: endless loop: `set` is not set to `set->next` so never changes

src/datasets.c
victorjulien: agreed

src/datasets.c
victorjulien: `DatasetPostReloadCleanup`?

src/datasets.c
```diff
@@ -532,6 +532,46 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
...
+        }
+        SCLogInfo("destroying set %s, hidden: %d address: %p", set->name, set->hidden, set);
+        Dataset *next = set->next;
+        THashShutdown(set->hash);
+        SCFree(set);
+        set = next;
+    }
+    SCLogInfo("destroying old copies of datasets done: %p", sets);
+    sets = NULL;
```
victorjulien, line 571: this doesn't look right: we now no longer have any reference to the sets list

src/datasets.c
```diff
@@ -532,6 +532,46 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
     return NULL;
 }
 
+int DatasetReload(uint32_t de_vers)
```
victorjulien, line 535: version arg isn't used

src/datasets.c
```diff
@@ -532,6 +532,46 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
...
+{
+    Dataset *set = sets;
+    // Reload if its a static set only
+    while (set != NULL) {
+        // Check if the set is static or dynamic
+        if ((set->save == NULL || strcmp(set->save, "") == 0) || (set->from_yaml == false)) {
```
victorjulien, line 541: would suggest adding small util funcs `DatasetIsStatic` and `DatasetIsDynamic`

src/datasets.c
victorjulien: the list management here needs a bit more thought. Test with a mixed list of dynamic and static sets and walk it after the cleanup to see it is still consistent.

src/datasets.h
```diff
@@ -37,7 +37,8 @@ typedef struct Dataset {
     char name[DATASET_NAME_MAX_LEN + 1];
     enum DatasetTypes type;
     uint32_t id;
-
+    bool from_yaml;
+    bool hidden;
```
victorjulien, line 41: add comment to explain its purpose

src/detect-engine.c
```diff
@@ -4118,6 +4122,7 @@ int DetectEngineReload(const SCInstance *suri)
         return -1;
     }
     SCLogDebug("set up new_de_ctx %p", new_de_ctx);
+    DatasetRemoveCopies();
```
victorjulien, line 4125: this needs to move until after `DetectEngineReloadThreads` for sure, probably until after `DetectEnginePruneFreeList` to avoid removing sets that are still being referenced by the old instance of the detection engine

inashivb:
Replaced with #5250

PR #5250, opened 2020-07-31
datasets: reload static sets

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3760

### DIFF
--- a/src/datasets.c
+++ b/src/datasets.c
@@ -47,6 +47,7 @@ static inline void DatasetUnlockData(THashData *d)
     (void) THashDecrUsecnt(d);
     THashDataUnlock(d);
 }
+static bool DatasetIsStatic(Dataset *set);
 
 enum DatasetTypes DatasetGetTypeFromString(const char *s)
 {
@@ -72,7 +73,7 @@ static Dataset *DatasetSearchByName(const char *name)
 {
     Dataset *set = sets;
     while (set) {
-        if (strcasecmp(name, set->name) == 0) {
+        if (strcasecmp(name, set->name) == 0 && set->hidden == false) {
             return set;
         }
         set = set->next;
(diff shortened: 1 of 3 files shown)


src/detect-engine.c
```diff
@@ -4092,7 +4093,7 @@ int DetectEngineReload(const SCInstance *suri)
     old_de_ctx = DetectEngineGetCurrent();
     if (old_de_ctx == NULL)
         return -1;
-    SCLogDebug("get ref to old_de_ctx %p", old_de_ctx);
```
inashivb, line 4095: Oops. Deleted this by mistake. Fixing in next rev

src/datasets.c
```diff
@@ -532,6 +533,62 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
     return NULL;
 }
 
+static bool DatasetIsStatic(Dataset *set)
+{
+    // A set is static if it does not have any dynamic properties like
```
victorjulien, line 538: please default to `/* */` style comment blocks (multiple times in this PR)

src/datasets.c
```diff
@@ -532,6 +533,62 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
...
+    // save and/or state defined.
+    if (set->save == NULL || strcmp(set->save, "") == 0) {
```
victorjulien, line 540: I would like to add a condition: that load *is* set. It is possible to have a 'dynamic' set that just doesn't save the data its tracking.

src/datasets.c
```diff
@@ -532,6 +533,62 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
...
+
+void DatasetReload(void)
+{
+    // In order to reload the datasets, just mark the current sets as hidden
+    // and clean them up later.
+    // New datasets shall be created with the rule reload and do not require
+    // any intervention.
+    SCMutexLock(&sets_lock);
+    Dataset *set = sets;
+    SCLogInfo("Inside reload function. Sets: %p", sets);
```
victorjulien, line 554: debug leftover

src/datasets.c
```diff
@@ -532,6 +533,62 @@ Dataset *DatasetGet(const char *name, enum DatasetTypes type,
...
+    Dataset *prev = NULL;
+    while (cur) {
+        Dataset *next = cur->next;
+        if (cur->hidden == false) {
+            prev = cur;
+            cur = next;
+            continue;
+        }
+        // Delete the set in case it was hidden
+        if (prev != NULL) {
```
victorjulien, line 582: I think this will not work correctly if the list head is removed, as it looks like `sets` is not updated.

src/detect-engine.c
victorjulien, line 4095: can you undo this debug line removal? It looks unintentional

victorjulien: ah I see this now, nm :)

src/datasets.c
inashivb: Yeah don't know what I was looking at while testing it. I put a lock on sets for a reason! Coming up fixed in the next PR

inashivb:
Replaced with #5254

PR #5252, opened 2020-07-31
ikev1: add ikev1 parser - v4 (preliminary!)

fhonza:
Previous PR: [#5097](https://github.com/OISF/suricata/pull/5097)

Describe changes:
- suggestion for integrating ikev2 and ikev1 in a common parser (type is still ikev1 but will move this)
- use DetectU32Data helper for keyword ikev1.exchtype


fhonza:
@catenacyber could you have a look if this is going in the right direction?

There is now a common ike.rs for interfacing suricata and a separate ikev1.rs and ikev2.rs. The logic of the two parsers is unchanged in the end, have tested with several pcaps and verify tests should work too after small renaming.

3 Keywords are now working for both protocols, but i'll enable all of them later. Additionally i would harmonize log output between both parsers.

There are still some things from v3 not solved, I just wanted to make sure the basic path is good :)

catenacyber:
Thanks, this looks good :-)

I have one remark
Could you split the commits this way ?
- first rename ikev2.rs ike.rs
- then split ike.rs into ikev2.rs and common ike.rs
- then add ikev1.rs
And maybe follow this naming convention for keywords as well : `ike` for common keyword `ikev1` for v1-specific keywords, `ikev2` for v2-specific keywords

Do you see the idea ?

fhonza:
thx, with the commits i'll try my best.

So you want to preserve the historical order in which the stuff was added to suricata? :thinking: 

For keywords it's a good idea, i already wondered how to separate the different "features" of both parsers.

catenacyber:
> So you want to preserve the historical order in which the stuff was added to suricata? 🤔

Not exactly.
Let me try to rephrase.
We want each commit to be a logical unit.
So I see the units :
- change IKEv2 to generic IKE (renaming only)
- move code specific to V2 to specific V2 files
- add code specific to V1 (which will branch at some point in the generic code)

> For keywords it's a good idea, i already wondered how to separate the different "features" of both parsers.

Two ideas :
- have a `ike.majorversion` keyword (match all version when absent)
- or have the same keywords which can be `ike.keyword`, `ikev1.keyword` or `ikev2.keyword`

fhonza:
Thanks! I like your second idea more with the version-specific keywords, will implement that for the next PR.

Currently a bit busy on other stuff, hope to send the next PR next week...

fhonza:
@catenacyber do you know if it's still realistic to get this into 6.0 or was there already a "feature" freeze?

catenacyber:
> do you know if it's still realistic to get this into 6.0 or was there already a "feature" freeze?

This seems unlikely to me, but @victorjulien will know better

PR #5253, opened 2020-07-31
Http2 v17

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/1947

There are likely comments, but I feel this is ready

Describe changes:
- Adds basic HTTP2 parser

Still more work to be done :
 + more commit squashing
 
This needs more testing with more pcaps for
 + files (compression, chunked, etc...)
 + reuse of http1 keywords

Modifies #5251 with rebasing to get CI to pass suricata-verify tests

suricata-verify-pr: 287

Makes https://github.com/OISF/suricata-verify/pull/287 pass

PR #5263, opened 2020-08-03
HTTP2 - v18

jasonish:
Based on https://github.com/OISF/suricata/pull/5253/files.

- Rebased.
- Disables HTTP2 by default and marks it as experimental.
- Pull some fields out of the headers and place under the http object to be compatible with http logging.

Example output from S-V http2-basic:
https://gist.github.com/jasonish/21bd72ddd2c15b4434b7c45180d93f80#file-gistfile0-txt-L79

Example output from S-V http2-upgrade:
https://gist.github.com/jasonish/6636c079433caff7ede88b230e1149ee

suricata-verify-pr: 293

S-V is failing for some reason. Passes locally.


victorjulien:
In the last record of your gist the headers are in the json record before the basic things like uri and method and such. I think it'd be nicer to reverse them.

jasonish:
> In the last record of your gist the headers are in the json record before the basic things like uri and method and such. I think it'd be nicer to reverse them.

That would visually look better, but would require an extra pass over the headers, or using intermediate JSON builders. This is because we iterate the frames and log the headers as we do the iteration, which we have to do to pull out these fields.

victorjulien:
The Debian builds fail due to compile errors around the derive use of hash. Any idea what is going on?

rust/src/http2/logger.rs
```diff
@@ -0,0 +1,277 @@
...
+ * 02110-1301, USA.
+ */
+
+use super::http2::{HTTP2Frame, HTTP2FrameTypeData, HTTP2Transaction};
+use super::parser;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
+use std;
+use std::collections::HashMap;
+
+#[derive(std::hash::Hash, PartialEq, Eq)]
```
victorjulien, line 24: changing this to just `Hash` (instead of `std::hash::Hash`) seems to fix it on Debian

victorjulien:
Merged in #5284 with a few fixes. I also dropped the upgrade patches as they need more discussion.

PR #5267, opened 2020-08-04
Plugin support - v13

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5150

Changes from last PR:
- Address most comment in previous PR.
- Remove example plugins, they can be found at https://github.com/jasonish/suricata-example-plugins
- Install headers

For now, all headers are installed. This will allow plugins to be build
without the complete source code, for example when installed with
a package manager.

Redmine ticket:


victorjulien:
I'm not too excited about just installing all these headers. Can we, as part of this being 'experimental', just accept ppl need a copy of the source code for now? Then we can take some time to figure out how to do the headers thing in a cleaner way.

jasonish:
> I'm not too excited about just installing all these headers. Can we, as part of this being 'experimental', just accept ppl need a copy of the source code for now? Then we can take some time to figure out how to do the headers thing in a cleaner way.

This will probably mean that they have to have built it from source as well.  One area that could cause problems is how different features change the size of the Packet structure, so you'll have to make sure to build against the same Suricata source code you plan to load a plugin into (in the case of a capture one). 

But yes, we can do this.

From poking around other programs that seem to support plugins you do see 2 styles.  One that require the source. Some so much so that you extract the plugin into the source code, and rebuild. This might have been nginx or older nginx. 

The other approach that Zeek and Pidgin/Libpurple take is to just install nearly all the headers as I think its quite a major job to provide some sort of separation.  Taking a quick look at Zeek, it install 1169 headers, and only has 796 in its source (so autogenerated headers I'm guessing).

jasonish:
Oh, and this can be achieved by dropping the last commit. I'll have to update the example plugins to match tho.

victorjulien:
Merged in #5284 without the header install commit and with some compile fixes.

Thanks Jason!

PR #5269, opened 2020-08-05
rust: only run cbindgen if needed - v2

jasonish:
In short, fix "make && sudo make install" when sudo
provides a sanitized path (even with -E) and Rust is
installed with Rustup.

Only run cbindgen when necessary. This is a bit tricky. When
building a dist we want to unconditionally build the headers.
    
When going through a "make; sudo make install" type process,
cbindgen should not be run as the headers already exist, are
valid, and the environment under sudo is more often than
not suitable to pick up the Rust toolchains when installed
with rustup.
    
For the normal "make" case we have the gen/rust-bindings.h file
depend on library file, this will cause it to only be rebuilt
if the code was modified.
    
For "make dist" we unconditionally create "dist/rust-bindings.h".
This means the generated file could be in 2 locations, so update
configure.ac, and the library search find to find it.
   
The "gen/rust-bindings.h" should be picked up first if it exists,
for those who develop from a dist archive where "dist/rust-bindings.h"
also exists.
    
Not completely happy having the same file in 2 locations, but not
sure how else to get the dependency tracking correct.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2504,7 +2504,7 @@ fi
     fi
 
     RUST_LDADD="${RUST_SURICATA_LIB} ${RUST_LDADD}"
-    CFLAGS="${CFLAGS} -I\${srcdir}/../rust/gen"
+    CFLAGS="${CFLAGS} -I\${srcdir}/../rust/gen -I\${srcdir}/../rust/dist"
     AC_SUBST(RUST_SURICATA_LIB)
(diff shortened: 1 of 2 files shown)


jasonish:
There is a work-around that doesn't require a patch:
```
sudo -E PATH="$PATH" make install
```

victorjulien:
Merged in #5276, thanks!

PR #5270, opened 2020-08-05
dcerpc: handle gap for TCP streams

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3564

#suricata-verify-pr: 296

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -16,13 +16,13 @@
  */
 
 use std::mem::transmute;
-
 use crate::applayer::{AppLayerResult, AppLayerTxData};
 use crate::core;
 use crate::dcerpc::parser;
 use crate::log::*;
 use nom::error::ErrorKind;
 use nom::number::Endianness;
+use nom;
 use std::cmp;
 
 // Constant DCERPC UDP Header length
@@ -149,6 +149,8 @@ pub struct DCERPCTransaction {
     pub stub_data_buffer_reset_tc: bool,
     pub req_done: bool,
     pub resp_done: bool,
+    pub req_lost: bool,
+    pub resp_lost: bool,
     pub req_cmd: u8,
     pub resp_cmd: u8,
     pub tx_data: AppLayerTxData,
@@ -174,6 +176,8 @@ impl DCERPCTransaction {
             stub_data_buffer_reset_tc: false,
             req_done: false,
             resp_done: false,
+            req_lost: false,
+            resp_lost: false,
             req_cmd: DCERPC_TYPE_REQUEST,
             resp_cmd: DCERPC_TYPE_RESPONSE,
             tx_data: AppLayerTxData::new(),
@@ -299,6 +303,10 @@ pub struct DCERPCState {
     pub prev_dir: u8,
     pub prev_tx_call_id: u32,
     pub clear_bind_cache: bool,
+    pub ts_gap: bool,
+    pub tc_gap: bool,
+    pub ts_ssn_gap: bool,
+    pub tc_ssn_gap: bool,
 }
 
 impl DCERPCState {
@@ -319,6 +327,10 @@ impl DCERPCState {
             prev_dir: core::STREAM_TOSERVER,
             prev_tx_call_id: 0,
             clear_bind_cache: false,
+            ts_gap: false,
+            tc_gap: false,
+            ts_ssn_gap: false,
+            tc_ssn_gap: false,
         };
     }
 
@@ -410,9 +422,11 @@ impl DCERPCState {
         match direction {
             core::STREAM_TOSERVER => {
                 self.buffer_ts.clear();
+                self.ts_gap = false;
             }
             _ => {
                 self.buffer_tc.clear();
+                self.tc_gap = false;
             }
         }
         self.bytes_consumed = 0;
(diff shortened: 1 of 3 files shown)


rust/src/dcerpc/dcerpc.rs
```diff
@@ -810,10 +880,49 @@ impl DCERPCState {
...
+                            },
+                            _ => {
+                                self.tc_gap = false;
+                                break;
+                            }
+                        }
+                    },
+                    _ => {
+                        let mut consumed = cur_i.len();
+                        if consumed < 2 {
```
catenacyber, line 914: This deserved a comment.
I think one byte is kept because we need two bytes to tell if it is the start of another record, right ?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -810,10 +880,49 @@ impl DCERPCState {
...
+        if self.ts_gap || self.tc_gap {
+            SCLogDebug!("Trying to catch up after GAP (input {})", cur_i.len());
+            while cur_i.len() > 0 { // min record size
+                match self.search_dcerpc_record(cur_i) {
+                    Ok((_, pg)) => {
+                        SCLogDebug!("DCERPC record found");
+                        let offset = cur_i.len() - pg.len();
+                        if offset == 1 {
+                            cur_i = &cur_i[offset + 2..];
+                            continue; // see if we have another record in our data
```
catenacyber, line 899: Why continue ? is not the first record the good one to start ?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -510,6 +524,62 @@ impl DCERPCState {
...
+                }
+                tx.req_done = true;
+                tx.resp_done = true;
+            }
+        }
+    }
+
+    pub fn search_dcerpc_record<'a>(&mut self, i: &'a[u8]) -> nom::IResult<&'a[u8], &'a[u8]> {
+        let mut d = i;
+        while d.len() >= 2 {
```
catenacyber, line 574: Why change `d` and not have an index variable ?

rust/src/dcerpc/log.rs
```diff
@@ -22,7 +22,7 @@ use crate::jsonbuilder::{JsonBuilder, JsonError};
 fn log_dcerpc_header(
     jsb: &mut JsonBuilder, state: &DCERPCState, tx: &DCERPCTransaction,
 ) -> Result<(), JsonError> {
-    if tx.req_done == true {
+    if tx.req_done == true && tx.req_lost == false {
```
catenacyber, line 25: Is it not worth to log the request alone ?

catenacyber:
Thanks @inashivb for this PR. This is a good model to handle gaps

rust/src/dcerpc/dcerpc.rs
inashivb: I suppose that could be done. Does that lead to improved handling and/or performance?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -810,10 +880,49 @@ impl DCERPCState {
...
         let retval;
-        let input_len = input.len();
+        let mut cur_i = input;
+        let input_len = cur_i.len();
         let mut v: Vec<u8>;
         // Set any query's completion status to false in the beginning
         self.query_completed = false;
+
+        // Skip the record since this means that its in the middle of a known length record
+        if self.ts_gap || self.tc_gap {
```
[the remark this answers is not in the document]
inashivb: I suppose it should.

rust/src/dcerpc/dcerpc.rs
inashivb: Yes.

rust/src/dcerpc/log.rs
inashivb: I'm not sure I understand what you mean here. This check was added because in case the request was lost, we won't have the params that we try to access for logging.
Could you please explain what you think should have been done here?

rust/src/dcerpc/dcerpc.rs
inashivb: hmm I considered here that the newest record that we can find is the one that should be the one to start after the gap as it might be followed by its corresponding record. But, now that I think it only makes sense if we're looking for a request after the gap. Get the latest request and hope it is followed by its corresponding response.
Would this make sense?

rust/src/dcerpc/dcerpc.rs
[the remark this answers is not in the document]
catenacyber: Should we create a ticket for this then ?

rust/src/dcerpc/dcerpc.rs
catenacyber: I do not DCEPRC specifics, but what would make sens to me is to parse the maximum data, hence get the earliest record we can...

rust/src/dcerpc/dcerpc.rs
catenacyber: I guess.
The classical way is more using an index...

rust/src/dcerpc/log.rs
catenacyber: Indeed, you are right :-)

PR #5275, opened 2020-08-05
Feat/flow redesign/v163

victorjulien:
TODOs
- test with flow bypass (likely broken)
- flow spare pool can't shrink currently, once it maxes out mem use stays high
- it seems this introduced a bit more fragility wrt flow timeout detection with pcaps again

Large redesign of flow handling.

Goals:
- reduce locking
- take advantage of 'hot' caches
- better locality

Locking reduction:

New flow spare pool. The global pool is implmented as a list of blocks,
where each block has a 100 spare flows. Worker threads fetch a block at
a time, storing the block in the local thread storage.

Flow Recycler now returns flows to the pool is blocks as well.

Flow Recycler fetches all flows to be processed in one step instead of
one at a time.

Cache 'hot'ness

Worker threads now check the timeout of flows they evaluate during lookup.
The worker will have to read the flow into cache anyway, so the added
overhead of checking the timeout value is minimal. When a flow is considered
timed out, one of 2 things happens:

- if the flow is 'owned' by the thread it is handled locally. Handling means
  checking if the flow needs 'timeout' work.

- otherwise, the flow is added to a special 'evicted' list in the flow
  bucket where it will be picked up by the flow manager.

Flow Manager timing:

By default the flow manager now tries to do passes of the flow hash in
smaller steps, where the goal is to do full pass in 8 x the lowest timeout
value it has to enforce. So if the lowest timeout value is 30s, a full pass
will take 4 minutes. The goal here is to reduce locking overhead and not
get in the way of the workers.

In emergency mode each pass is full, and lower timeouts are used.

Timing of the flow manager is also no longer relying on pthread condition
variables, as these generally cause waking up much quicker than the desired
timout. Instead a simple (u)sleep loop is used.

Both changes reduce the number of hash passes a lot.

Emergency behavior

In emergency mode there a number of changes to the workers. In this scenario
the flow memcap is fully used up and it is unavoidable that some flows won't
be tracked.

1. flow spare pool fetches are reduced to once a second. This avoids locking
   overhead, while the chance of success was very low.

2. getting an active flow directly from the hash skips flows that had very
   recent activity to avoid the scenario where all flows get only into the
   NEW state before getting reused. Rather allow some to have a chance of
   completing.

3. TCP packets that are not SYN packets will not get a used flow, unless
   stream.midstream is enabled. The goal here is again to avoid evicting
   active flows unnecessarily.

Better Localily

Flow Manager injects flows into the worker threads now, instead of one or
two packets. Advantage of this is that the worker threads can get packets
from their local packet pools, avoiding constant overhead of packets returning
to 'foreign' pools.

Counters

A lot of flow counters have been added and some have been renamed.

Overall the worker threads increment 'flow.wrk.*' counters, while the flow
manager increments 'flow.mgr.*'.

Additionally, none of the counters are snapshots anymore, they all increment
over time. The flow.memuse and flow.spare counters are exceptions.

Misc

FlowQueue has been split into a FlowQueuePrivate (unlocked) and FlowQueue.
Flow no longer has 'prev' pointers and used a unified 'next' pointer for
both hash and queue use.


victorjulien:
commit 10c58e4 is where most of the action is. It's a squash of about 200 commits in what I call a 'journey branch'. Too big to cleanup :(

victorjulien:
Results I'm getting with t-rex:

**Single CPU system (Xeon e5-2690v2 10 core / 20 threads with Intel x520) in IPS mode (AF_PACKET)**

Replay at about 10G combined: `./t-rex-64 -f cap2/sfr_agg_tcp14_udp11_http200msec_new_high_new_nir_profile_ipg_mix.yaml -c 4 -m 10 -d 1800 --active-flows 524288 -p`

Suricata start up: `./src/suricata -c suricata-ips-x520.yaml --af-packet -v --disable-detection -k none --set flow.memcap=1gb --set flow.hash-size=500000 --set stream.memcap=2gb --set stream.reassembly.memcap=8gb -v`

master:
```
 Total-pkt-drop       : 583726814 pkts 
 Total-tx-pkt         : 3785561572 pkts 
 Total-rx-pkt         : 3201834758 pkts 
```

this branch:
```
 Total-pkt-drop       : 328925460 pkts 
 Total-tx-pkt         : 3785561572 pkts 
 Total-rx-pkt         : 3456636112 pkts 
```
So a small reduction of the drops.


**Dual CPU system (Xeon e5-2680v4 2x 14 core / 28 threads with Intel X710) in IPS mode (AF_PACKET)**

Replay at about 20G combined: `./t-rex-64 -f cap2/sfr_agg_tcp14_udp11_http200msec_new_high_new_nir_profile_ipg_mix.yaml -c 4 -m 20 -d 1800 --active-flows 524288 -p`

Suricata start up: `./src/suricata -c suricata-ips-x710-10G.yaml --af-packet -v --disable-detection -k none --set flow.memcap=1gb --set flow.hash-size=500000 --set stream.memcap=2gb --set stream.reassembly.memcap=8gb -l /home/victor/vm/logs/ --set outputs.1.eve-log.enabled=true --set outputs.1.eve-log.threaded=true`

master:
```
 Total-pkt-drop       : 4678314880 pkts 
 Total-tx-pkt         : 7571125457 pkts 
 Total-rx-pkt         : 2892810577 pkts 
```

this branch
```
 Total-pkt-drop       : 928154464 pkts 
 Total-tx-pkt         : 7571125457 pkts 
 Total-rx-pkt         : 6642970993 pkts 
```
So massive improvement.

victorjulien:
Replaced by #5281

PR #5291, opened 2020-08-12
Capture feature 120 v4

scottfgjordan:

Describe changes:
- The TcpSegment enrichment only will occur if the session-dump option is enabled in suricata.yaml nor will any allocations occur. 

Suggested Review Order/Guide:
- log-pcap-stream.c/.h
- detect-tag-pcap.c/h
- pcap-helper.c/h
 - This is basically the same as the code for libpcaps pcap_dump(), but rewritten so that it could dump a packet in two parts, e.g. header and payload written separately. This is currently written in Suricata codestyle and is internal to Suricata, but I'm open to approaching the libpcap team to get this functionality added to their pcap API.
- stream-tcp .c/.h files

Previous Pull Requests: 
https://github.com/OISF/suricata/pull/5217

**Changes from capture-feature-120-v3:**
- I've chopped down the size of a few structures in the code to improve performance:
  - The TaggedPcapEntry no longer holds the whole file path. Now it just contains the information necessary to produce the file path on the fly when it is needed. 
  - In the previous version, most of the enriched TcpSegment data was held by adding members to the TcpSegment struct, which resulted in a size increase of 32 bytes. In this version, I've moved that information to a new struct, TcpSegmentPcapHdrStorage, which now holds the necessary information. TcpSegment now only holds an additional pointer to this. The TcpSegmentPcapHdrStorage is also initialized with the TcpSegment pool, so there are still very few allocations during packet processing. I've also dropped the overhead when TcpSegment dumping is not enabled so that not only is the AddPacketData function not triggered, the TcpSegmentPcapHdrStorage space is also not allocated in the first place.
- Additionally, I've removed the temporarily hidden file aspect to the .pcap output. I didn't have a strong justification for doing this in the first place and felt like it wasn't really necessary or inherent to the functionality of the session capture feature. 
- Otherwise, most of the other comments from the previous PR were nits, which I corrected across the board. I think this feature is pretty close to being ready, so let me know what else is needed to get it across the finish line 😄

### DIFF
--- a/src/detect-tag-pcap.c
+++ b/src/detect-tag-pcap.c
@@ -1,0 +1,291 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#include "suricata-common.h"
+#include "detect-tag-pcap.h"
+#include "decode.h"
+#include "pcap-helper.h"
+#include "util-time.h"
+#include "stream-tcp-private.h"
+#include <pthread.h>
+#include <errno.h>
+
+#ifndef HOST_NAME_MAX
+#ifdef __APPLE__
+#define HOST_NAME_MAX 255
+#elif defined(_WIN32) || defined(_WIN64)
+#define HOST_NAME_MAX 255
+#endif
+#endif
+
+/* PCAP_SNAPLEN (snapshot length) is the amount of data available per capture.
+ * The default is 262144 bytes. Setting the snaplen to 0 will set it to
+ * this default value, but also allow for backwards compatibility for older
+ * versions of tcpdump. */
+#define PCAP_SNAPLEN 0
+
+typedef struct OutputConfig_ {
+    char g_hostname[HOST_NAME_MAX];
+    const char *g_output_dir;
+} OutputConfig;
+
+OutputConfig g_output_config;
+
+void GenerateStreamFilenameTimebuf(char *time_buf,size_t time_buf_size,
+                                   time_t packet_time);
+static void DumpTcpSegment(TcpSession *session, TcpSegment *seg,
+                          pcap_dumper_t *dump_handle, bool client);
+
+/**
+ * \brief Initializes filename for output pcap log.
+ * \param output_directory Configurable filename from suricata.yaml.
+ */
+void InitializePcapLogFilenameSupport(const char *output_directory)
+{
+    if(gethostname(g_output_config.g_hostname, sizeof(g_output_config.g_hostname))
+        != 0) {
+        FatalError(SC_ERR_HOST_INIT, "Error looking up hostname in "
+                                     "detect-tag-pcap.c Error: %s",
+                                     strerror(errno));
+    }
+
+    if (output_directory == NULL) {
+        g_output_config.g_output_dir = ConfigGetLogDirectory();
+    } else {
+        g_output_config.g_output_dir = output_directory;
+    }
+}
+
+/**
+ * \brief Fills the time_buf with a time string that will be used in the pcap
+ *  filename.
+ * \param time_buf buffer to hold the time string.
+ * \param time_buf_size length of the time buffer.
+ * \param packet_time time value that is used to generate the time string.
+ */
+void GenerateStreamFilenameTimebuf(char *time_buf, size_t time_buf_size,
+        time_t packet_time)
+{
+    time_t time = packet_time;
+    struct tm local_tm;
+    struct tm *t = SCLocalTime(time, &local_tm);
+
+    if (unlikely(t == NULL)) {
+        FatalError(SC_ERR_TS, "Unable to create time structure, ts-error");
+    }
+
+    char fmt_time_buf[64];
+    CreateFormattedTimeString(t, "%Y%m%d_%H%M%S.%%06u", fmt_time_buf,
+                              sizeof(fmt_time_buf));
+
+    int ret = snprintf(time_buf, time_buf_size,
+                       fmt_time_buf, packet_time);
+    if (ret < 0 || (size_t) ret >= time_buf_size) {
+        SCLogError(SC_ERR_INVALID_NUM_BYTES, "Provided buffer size is too "
+                                             "small to create full time "
+                                             "buffer. Error: %s",
+                   strerror(errno));
+    }
+}
+
+/**
+ *  \brief Fills the result_path_buf with a full file path that can be used
+ *   to create a file. InitializePcapLogFilenameSupport() must be run before
+ *   this is called.
+ *  \param result_path_buf buffer to hold the path string.
+ *  \param result_buf_size length of the filename buffer used.
+ *  \param p packet pointer.
+ *  \param signature that alerted.
+ *  \param thread_id unique id for the current thread.
+ *  \param unique_id counter incrementing over time to add entropy to filenames.
+ */
+void GenerateStreamFilepath(char *result_path_buf, size_t result_buf_size,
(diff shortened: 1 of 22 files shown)


src/detect-tag-pcap.c
```diff
@@ -0,0 +1,291 @@
...
+ *  \param p pointer to the packet being alerted on.
+ *  \param sig signature that triggered the alert.
+ *  \param thread_id id of the thread we're on; used for uniqueness in filename.
+ *  \param unique_id unique counter per thread that increments to increase
+ *  the entropy of filenames.
+ */
+TaggedPcapEntry *SetupTaggedPcap(const Packet *p, const Signature *sig, int
+        thread_id, int unique_id)
+{
+    char pcap_file_path[PATH_MAX];
```
scottfgjordan, line 148: This should be a dynamic allocation, will switch it in next version...

src/detect-tag-pcap.c
```diff
@@ -0,0 +1,291 @@
...
+ * 02110-1301, USA.
+ */
+
+#include "suricata-common.h"
+#include "detect-tag-pcap.h"
+#include "decode.h"
+#include "pcap-helper.h"
+#include "util-time.h"
+#include "stream-tcp-private.h"
+#include <pthread.h>
```
victorjulien, line 24: nit: these should be in `suricata-common.h` or be added there if not.

src/detect-tag-pcap.c
```diff
@@ -0,0 +1,291 @@
...
+#include <errno.h>
+
+#ifndef HOST_NAME_MAX
+#ifdef __APPLE__
```
victorjulien, line 28: whats the reason for having the multiple OS' here while all lead to the same result? Can we just use one?

src/detect-tag-pcap.c
```diff
@@ -0,0 +1,291 @@
...
+    tpe->unique_id = unique_id;
+    tpe->signature_id = sig->id;
+
+    GenerateStreamFilepath(pcap_file_path, PATH_MAX, tpe);
+
+    tpe->pcap_dead_handle = pcap_open_dead(p->datalink, PCAP_SNAPLEN);
+    if (tpe->pcap_dead_handle == NULL) {
+        SCLogError(SC_ERR_PCAP_OPEN_OFFLINE, "Error opening dead pcap "
+                                             "handle: %s", pcap_geterr
+                                             (tpe->pcap_dead_handle));
```
victorjulien, line 167: style: please have `pcap_geterr(tpe->pcap_dead_handle)` on a single line

src/detect-tag-pcap.c
```diff
@@ -0,0 +1,291 @@
...
+    }
+    /*
+     * Initialize tpe's id & time variables
+     */
+    tpe->time = p->ts.tv_usec;
+    tpe->thread_id = thread_id;
+    tpe->unique_id = unique_id;
+    tpe->signature_id = sig->id;
+
+    GenerateStreamFilepath(pcap_file_path, PATH_MAX, tpe);
```
victorjulien, line 161: this can `FatalError` while we are on the packet path. Can we make it more forgiving?

src/log-pcap-stream.c
```diff
@@ -0,0 +1,312 @@
...
+ * \param tv thread-specific variables.
+ * \param td thread data containing the output context.
+ * \param p Pointer to current packet being processed.
+ */
+static void GeneratePcapFiles(ThreadVars *tv, StreamPcapLogThreadData *td,
+        const Packet *p)
+{
+    if (!(p->flags & PKT_HAS_FLOW) && p->flags & PKT_HAS_TAG) {
+        /*
+         * Single packet protocols (eg. ICMP) won't have a flow created for
```
victorjulien, line 98: some ICMP will, e.g. echo/echo-reply

src/log-pcap-stream.c
```diff
@@ -0,0 +1,312 @@
...
+         * them. A tagged rule alert can still trigger on the packet meaning
+         * we'd want to produce a PCAP.
+         */
+        for (uint16_t x = 0; x < p->alerts.cnt; x++) {
+            /* Is the alert from a tagged signature? */
+            if (p->alerts.alerts[x].s->sm_arrays[DETECT_SM_LIST_TMATCH] !=
+                NULL) {
+                TaggedPcapEntry *pcap_file = SetupTaggedPcap(p, p->alerts
```
victorjulien, line 106: it looks like `SetupTaggedPcap` can return NULL in error cases, which would crash on the next line.

src/log-pcap-stream.c
```diff
@@ -0,0 +1,312 @@
...
+            }
+            if (tag_alert == NULL) {
+                /**
+                   * This case happens when a rule hits its threshold
+                   * settings. The alert doesn't get generated but the
+                   * tagging will still occur. Skip the tag.
+                   */
+                continue;
+            }
+            current_tag->pcap_file = SetupTaggedPcap(p, tag_alert->s, tv->id,
```
victorjulien, line 148: error handling vs pointer dereferencing

src/log-pcap-stream.c
```diff
@@ -0,0 +1,312 @@
...
+}
+
+/**
+ *  \brief Fill in stream-pcap logging struct from the provided ConfNode.
+ *  \param conf The configuration node for this output.
+ *  \return output_ctx: Output context (contains a StreamOutputCtx data member)
+ */
+static OutputInitResult StreamPcapLogInitCtx(ConfNode *conf)
+{
+    SCLogNotice("StreamPcapLogInitCtx enter");
```
victorjulien, line 247: debug artifact

src/log-pcap-stream.c
```diff
@@ -0,0 +1,312 @@
...
+}
+
+/**
+ * \brief StreamPcapLogFileDeInitCtx Free and close the output context
+ *  created by StreamPcapLogInitCtx.
+ * \param output_ctx The output context to free.
+ */
+static void StreamPcapLogFileDeInitCtx(OutputCtx *output_ctx)
+{
+    SCLogNotice("StreamPcapLogFileDeInitCtx Enter");
```
victorjulien, line 302: debug

src/stream-tcp-list.c
```diff
@@ -890,6 +895,53 @@ void StreamTcpPruneSession(Flow *f, uint8_t flags)
...
+        ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx)
+{
+    if (seg->pcap_hdr_storage == NULL || seg->pcap_hdr_storage->pkt_hdr == NULL) {
+        return;
+    }
+    if (GET_PKT_DATA(p) != NULL && GET_PKT_LEN(p) > p->payload_len) {
+        seg->pcap_hdr_storage->ts.tv_sec = p->ts.tv_sec;
+        seg->pcap_hdr_storage->ts.tv_usec = p->ts.tv_usec;
+        seg->pcap_hdr_storage->pktlen = GET_PKT_LEN(p);
+        seg->pcap_hdr_storage->pcap_cnt = p->pcap_cnt;
```
victorjulien, line 919: `Packet::pcap_cnt` is only going to be set when we're reading a pcap. For live traffic modes it isn't used.

src/detect-tag-pcap.c
scottfgjordan: Good point, I had added in the cases at separate times as I ran into build errors due to HOST_NAME_MAX not being defined. Will just define HOST_NAME_MAX regardless of Mac/Windows.

src/detect-tag-pcap.c
scottfgjordan: Will do, thanks!

src/detect-tag-pcap.c
scottfgjordan: Good catch, thanks!

src/stream-tcp-list.c
scottfgjordan: This is a good point that I hadn't realized as my development was tested against offline pcaps the whole time. The pcap_cnt member is only being used to determine the order of dumping, i.e. seeing which pcap was older. Working out a way to do this that will work for both live and offline modes shouldn't be an issue.

src/log-pcap-stream.c
scottfgjordan: Will change the comment, to better reflect this.

src/detect-tag-pcap.c
scottfgjordan: Sure thing! I'll switch it to a SCLogError and then just not open up the pcap_dumper in the case of an error.

src/log-pcap-stream.c
scottfgjordan: Good catch thanks. Not sure how I missed actually handling the returned NULL after working out the error cases... I will add a check error and log and return in the case of an error.

scottfgjordan:
Closing this down as a newer pull request can be found at: https://github.com/OISF/suricata/pull/5345

PR #5297, opened 2020-08-14
flows: add active flow counter

JoshuaLumb:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [1478](https://redmine.openinfosecfoundation.org/issues/1478)

Describe changes:
- Adds a global counter "flow_mgr.active_flows" tracking active flows The counter is global to support multiple flow manager threads.

### DIFF
--- a/src/flow-hash.c
+++ b/src/flow-hash.c
@@ -476,6 +476,7 @@ static inline void FlowUpdateCounter(ThreadVars *tv, DecodeThreadVars *dtv,
                 StatsIncr(tv, dtv->counter_flow_icmp6);
                 break;
         }
+        FlowManagerIncrActiveFlows(1);
 #ifdef UNITTESTS
     }
 #endif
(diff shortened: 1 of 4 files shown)


victorjulien:
I wonder if this could be implemented without adding a global atomic. These types of atomics are something we're trying to get rid of, as they are expensive especially on multi-NUMA systems.

In `master` we do already count flow creation (although per protocol) and also flow removal/destruction. So I wonder if we could utilize these counters to calculate the difference, which should be the number of active flows.

Maybe a PoC could be done by hardcoding it into `counters.c` somehow, taking the various relevant counter values and using it to set the 'active' flows counter...

Does this make sense to you?

victorjulien:
cc: @regit

victorjulien:
I have developed an alternate solution, that uses the `Stats` API by extending it to support not just `StatsIncr` but also `StatsDecr`. This way we get the same output, w/o the overhead of atomics. Will share a PR soon.

PR #5301, opened 2020-08-17
ikev1: add ikev1 parser - v5

fhonza:
Previous PR: (#5252)[https://github.com/OISF/suricata/pull/5252]

Describe changes:
- unification of IKEv1 and IKEv2 parsers under IKE
- updated documentation
- usage of DetectU32Data/DetectU8Data - Helpers for several keywords
- usage of InspectionBufferMultipleForListGet for vendor keyword
- several refactorings in the rust code
- added event for IKEv1 for extra data in the payload
- set rdir where possible in the proto-probing function (rust)
- check for hexadecimal content in the key-exchange -payload rules

suricata-verify-pr: [#301](https://github.com/OISF/suricata-verify/pull/301)


catenacyber:
@fhonza I edited your first comment to use the suricata-verify PR : `sed s/#suricata-verify-pr: /suricata-verify-pr: /`

doc/userguide/rules/ike-keywords.rst
```diff
@@ -0,0 +1,159 @@
...
+
+The keywords
+
+* ``ike.spi_initiator``
+* ``ike.spi_responder``
+* ``ike.chosen_sa_attribute``
+* ``ike.exchtype``
+* ``ikev1.vendor``
+* ``ikev1.server_key_exchange_payload``
+* ``ikev1.client_key_exchange_payload``
```
catenacyber, line 12: Could there be only `ikev1.key_exchange_payload` ? cf https://github.com/OISF/suricata/pull/5097#discussion_r477249870

catenacyber:
It looks like CI did not run. Could you rebase (to fix the simple conflict) and make another PR ?

rust/src/common.rs
```diff
@@ -72,3 +72,19 @@ pub unsafe extern "C" fn rs_cstring_free(s: *mut c_char) {
...
+    static CHARS: &'static [u8] = b"0123456789abcdef";
+
+    let mut output = Vec::with_capacity(input.len() * 2);
+    for &byte in input {
+        output.push(CHARS[(byte >>  4) as usize]);
+        output.push(CHARS[(byte & 0xf) as usize]);
+    }
+
+    // this should never fail, characters are from a defined set
+    unsafe {
```
catenacyber, line 87: Could we build the `String` directly instead of using a `Vec` to avoid this use of `unsafe` ?

rust/src/ike/detect.rs
```diff
@@ -0,0 +1,263 @@
...
+#[no_mangle]
+pub extern "C" fn rs_ike_state_get_exch_type(
+    tx: &mut IKETransaction,
+    exch_type: *mut u8,
+) -> u8 {
+    if exch_type == std::ptr::null_mut() {
+        return 0;
+    }
+
+    if tx.ike_version == 1 {
```
catenacyber, line 34: style: I wonder if we should rather use `match` than `if else if`

rust/src/ike/detect.rs
```diff
@@ -0,0 +1,263 @@
...
+    return 1;
+}
+
+#[no_mangle]
+pub extern "C" fn rs_ike_state_get_spi_responder(
+    tx: &mut IKETransaction,
+    buffer: *mut *const u8,
+    buffer_len: *mut u32,
+) -> u8 {
+    if buffer == std::ptr::null_mut() || buffer_len == std::ptr::null_mut() {
```
catenacyber, line 74: I think we do not need these checks

rust/src/ike/detect.rs
```diff
@@ -0,0 +1,263 @@
...
+pub extern "C" fn rs_ike_state_get_nonce(
+    tx: &mut IKETransaction,
+    buffer: *mut *const u8,
+    buffer_len: *mut u32,
+) -> u8 {
+    if buffer == std::ptr::null_mut() || buffer_len == std::ptr::null_mut() {
+        return 0;
+    }
+
+    if !tx.hdr.ikev1_header.nonce.is_empty() {
```
catenacyber, line 95: Should we not first check we have ike _v1_ here ?

rust/src/ike/logger.rs
```diff
@@ -0,0 +1,237 @@
...
+        _ => js.append_string("")?
+    };
+    Ok(())
+}
+
+fn log_ikev2(state: &IKEState, tx: &IKETransaction, jb: &mut JsonBuilder) -> Result<(), JsonError>
+{
+    jb.open_object("ike")?;
+    jb.set_uint("version_major", tx.hdr.ikev2_header.maj_ver as u64)?;
+    jb.set_uint("version_minor", tx.hdr.ikev2_header.min_ver as u64)?;
```
catenacyber, line 199: There are some common lines which can be factorized, are not they ?

src/detect-ike-exch-type.c
```diff
@@ -0,0 +1,165 @@
...
+#include "detect-engine-content-inspection.h"
+#include "detect-ike-exch-type.h"
+#include "app-layer-parser.h"
+#include "util-byte.h"
+#include "detect-engine-uint.h"
+
+#include "rust-bindings.h"
+
+/**
+ *   [ike.exchtype]:[<|>|<=|>=]<type>;
```
catenacyber, line 37: It seems to me `>=` is not yet supported by `DetectU8Data`
Could you add it ?

src/detect-ike-key-exchange-payload-length.c
```diff
@@ -0,0 +1,162 @@
...
+    if (sm == NULL)
+        goto error;
+
+    sm->type = DETECT_AL_IKE_KEY_EXCHANGE_PAYLOAD_LENGTH;
+    sm->ctx = (SigMatchCtx *)key_exchange_payload_length;
+
+    SigMatchAppendSMToList(s, sm, g_ike_key_exch_payload_length_buffer_id);
+    return 0;
+
+    error:
```
catenacyber, line 148: nit: remove indent before label

src/detect-ike-key-exchange-payload.c
```diff
@@ -0,0 +1,226 @@
...
+    for ( ; sm != NULL; sm = sm->next)
+    {
+        if (sm->type != DETECT_CONTENT)
+            continue;
+
+        const DetectContentData *cd = (DetectContentData *)sm->ctx;
+
+        for (size_t i = 0; i < cd->content_len; ++i)
+        {
+            if(!isxdigit(cd->content[i]))
```
catenacyber, line 144: Why is only hexadecimal expected here ?

src/tests/fuzz/fuzz_sigpcap.c
```diff
@@ -88,7 +88,6 @@ outputs:\n\
         - nfs\n\
         - smb\n\
         - tftp\n\
-        - ikev2\n\
```
catenacyber, line 91: this should be replace by `ike`

rust/src/ike/ikev1.rs
```diff
@@ -0,0 +1,158 @@
...
+use crate::applayer::*;
+use nom;
+use crate::ike::parser::*;
+use crate::ike::ike::{IKEState, IkeEvent};
+use std::collections::HashSet;
+
+#[derive(Default)]
+pub struct IkeV1Header {
+    pub maj_ver: Option<u8>,
+    pub min_ver: Option<u8>,
```
catenacyber, line 32: Are these fields not a duplicate of the common header ?

rust/src/ike/ikev1.rs
```diff
@@ -0,0 +1,158 @@
...
+                        &tx.hdr.ikev1_header.nonce,
+                        &tx.hdr.ikev1_header.vendor_ids,
+                        &tx.hdr.ikev1_transforms
+                    );
+                }
+
+                if rem.len() > 0 {
+                    // more data left unread than should be
+                    SCLogDebug!("Unread Payload Data");
+                    state.set_event(IkeEvent::PayloadExtraData);
```
catenacyber, line 139: There should be a files in the `rules` directory to match on these app-layer events

rust/src/ike/ike.rs
```diff
@@ -0,0 +1,628 @@
...
+        }
+        if let Some(state) = self.de_state {
+            core::sc_detect_engine_state_free(state);
+        }
+    }
+
+    pub fn set_progress(&mut self, direction: u8) {
+        let mut progress = 1; // set to 1 to log and initiator_spi alert
+        if self.hdr.ikev1_header.key_exchange.len() > 0 && direction == STREAM_TOSERVER {
+            progress = 3;
```
catenacyber, line 157: Could we have a name for this magic 3 constant ?

catenacyber: And reuse this name in the C code for logging cf `SshStateBannerDone`

rust/src/ike/ike.rs
```diff
@@ -0,0 +1,628 @@
...
+    }
+}
+
+/// Probe to see if this input looks like a request or response.
+fn probe(input: &[u8], rdir: *mut u8) -> bool {
+    match parse_isakmp_header(input) {
+        Ok((_, isakmp_header)) => {
+            if isakmp_header.maj_ver == 1 {
+                if isakmp_header.resp_spi == 0 {
+                    unsafe {*rdir = STREAM_TOSERVER;}
```
catenacyber, line 291: I think the direction should be reversed only if it is not the right one

rust/src/ike/ike.rs
```diff
@@ -0,0 +1,628 @@
...
+        return unsafe { ALPROTO_FAILED };
+    }
+
+    if input_len > 1 && input != std::ptr::null_mut() {
+        let slice = build_slice!(input, input_len as usize);
+        if probe(slice, rdir) {
+            return unsafe { ALPROTO_IKE };
+        }
+    }
+    return ALPROTO_UNKNOWN;
```
catenacyber, line 354: Why not `ALPROTO_FAILED` ?

rust/src/ike/ike.rs
```diff
@@ -0,0 +1,628 @@
...
+pub extern "C" fn rs_ike_probing_parser(
+    _flow: *const Flow,
+    _direction: u8,
+    input: *const u8,
+    input_len: u32,
+    rdir: *mut u8
+) -> AppProto {
+    if input_len < 28 {
+        // at least the ISAKMP_HEADER must be there
+        return unsafe { ALPROTO_FAILED };
```
catenacyber, line 345: Could you add a comment that this is not `ALPROTO_UNKNOWN` because we are over UDP ?

rust/src/ike/ike.rs
```diff
@@ -0,0 +1,628 @@
...
+use std::collections::HashSet;
+
+static mut ALPROTO_IKE: AppProto = ALPROTO_UNKNOWN;
+
+#[repr(u32)]
+pub enum IkeEvent {
+    MalformedData = 0,
+    NoEncryption,
+    WeakCryptoEnc,
+    WeakCryptoPRF,
```
catenacyber, line 43: Are these events part of the new PR ? Or were they already here ?

catenacyber (changes requested):
The overall design is looking good.
There are still some iterations to do, I guess...

rust/src/ike/detect.rs
fhonza: hmmm first i had it with a match-statement, but with the obligatory "_" it had 3 options and found the if/else way somehow nicer...no strong opinion here....

rust/src/ike/detect.rs
fhonza: have added them everywhere after review from @chifflier ....  (this was in the first PR for the parser)

rust/src/ike/detect.rs
fhonza: Not sure if needed, the data structure is only filled if we have a v1 communication. Perhaps reasonable if integer comparision faster than check for empty string? (not sure here)

src/detect-ike-exch-type.c
fhonza: yes have added it and will appear in v6 :)

rust/src/ike/ikev1.rs
fhonza: the existing implementation of ikev2 uses the data structures from an external lib (https://github.com/rusticata/ipsec-parser).
This implementation is bound to the IKEv2-Types.
I just didn't want to reimplement/change too much stuff there. Thats why data structures of both versions are more separated than needed. Elsewhise this could be a common Header....

src/detect-ike-key-exchange-payload.c
fhonza: You are right pcre is possible as well, i'll add the reserved chars and rename the test.

rust/src/ike/ike.rs
fhonza: this is more coincidence, haven't debugged what happens after returning these constants :)

rust/src/ike/ike.rs
fhonza: no, were already there, i've just added PayloadExtraData.....

src/detect-ike-key-exchange-payload.c
catenacyber: But why not any character from 0 to 255 ?

rust/src/ike/ikev1.rs
catenacyber: That would be nice if this `maj_ver` was not present in `IkeV1Header` and was present in some common `IkeHeader `, even if it is also redundant in IKEv2

rust/src/ike/detect.rs
catenacyber: > the data structure is only filled if we have a v1 communication

 Do not we risk some kind of null dereference then ?

That seems reasonable to me to add an integer comparison...

rust/src/ike/detect.rs
catenacyber: @chifflier, why did you ask for this ?
These checks are not done in the other parsers, are they ? cf `rs_ssh_tx_get_protocol` for instance

rust/src/ike/detect.rs
catenacyber: No strong opinion neither...

fhonza: then i'll leave it that way :)

rust/src/ike/detect.rs
fhonza: ok

src/detect-ike-key-exchange-payload.c
fhonza: yes....sounds simple, ok

rust/src/ike/ikev1.rs
fhonza: ok i'll add the common parts one layer above, found the redundancy initially indeed not so nice with ikev2.

rust/src/ike/detect.rs
catenacyber: > If Suricata ensures that pointers are never null

I think that is the case now...
@jasonish thoughts on this ?

jasonish: I'm fine with it, even though I see it as part of the contract here that these should not be NULL.  Perhaps in the future we should start adding bug_on type assertions here to catch the possibility of these being null in CI.

catenacyber: Yes `bug_on` or `debug_validate_bug_on` seem to be better than testing and returning 0 @fhonza

src/detect-ike-key-exchange-payload.c
fhonza: checked DetectContentData->content, it's type is *uint8_t
It should not be possible to be < 0 or > 255. (got a compiler warning as well that this cannot happen).

Better remove this test? Alternative would be constraint hex + pcre-chars .....

catenacyber: > Alternative would be constraint hex + pcre-chars 

I am trying to figure out why you want this restriction.
Why could not there be `///` in the content for ikev1 exchange payload instance ?

fhonza: The exchange payload is stored in rust as a hex-encoded string. Afaik exact match with content: and "fuzzy" match with pcre: are supported by sticky buffers.
Just wanted to restrict the rule-content to values which are possible. Or does this make no sense?

catenacyber: > The exchange payload is stored in rust as a hex-encoded string

Thanks, this is the interesting point I was missing.

So now, why is this stored as a hex-encoded string ?
We could use the raw data, and avoid to have this content validation function.

fhonza: The key-exchange-payload is the public Diffie-Hellman key, which is just a "byte blob", there is no encoding defined in the protocol. I chose hex-encoding because it's the same encoding used in wireshark, making dev/validation easier for me.

Also i found it nicer to have the eve.json output "compatible" with the rules....

rust/src/ike/ike.rs
fhonza: checked on the whole progress feature and will drop it completely (always keep 1), the change a few PR's back of pushing alert-values into transactions and not alerting from state-values solved the problems i had...

src/detect-ike-key-exchange-payload.c
fhonza: ....checked other protocols/rules, i'll store nonce/payload as byte-array and will transform to hex just for the logger.

Didn't know that |xx| is supported in the rule-syntax for writing "arbitrary byte matches".

first parser i'm implementing completely myself here, feeling a bit noobish from time to time.....

catenacyber: You are doing great :-)

fhonza: thx :)

PR #5315, opened 2020-08-21
Plugins - Remove RustContext and dynamically link at runtime - v1

jasonish:
The root of this problem is that Rust code that is part of a plugin will not see the function pointers that are passed in the `RustContext` object during `rs_init()`.  Right now there are not many functions that a plugin would need to call, but logging is one of these. There are 3 ways to fix this:

1) Get rid of the RustContext. For Rust tests to pass, we'd also have to link with the Suricata C code compiled to a library. This complicates the build a little as we'd have to build all C, and all Rust before running the Rust tests.  I did get this to work, but not successfully with ASAN, but ultimately its where we want to end up, and probably what we have to get to for a "libsuricata.so" type thing.
2) On plugin initialiation, the plugin could execute some boilerplate that essentially sets up this context. This might have been the simpler approach, but I really didn't want to introduce this boilerplate into the plugin.  Suricata can't call this code on behalf of the plugin, the plugin must call it iself. Like some "BootStrapPlugin" method. It would only need to be Rust.
3) The approach I took which removes the boilerplate and removes the RustContext and sets up the function pointers as `lazy_static` references.  Essentially on first use, say `SCLogMessage`, we use `dlsym` to find this symbol then link up the function pointer, otherwise set it to no. So there is a small cost on first use. This isn't that different from what we do know where we check if the RustContext is set or not, and call the function based on that.

These changes allow a pure Rust plugin to be written:
https://github.com/jasonish/suricata-example-plugins/blob/rust-plugin/rust/src/lib.rs

Its still not that pleasant to look at, but its easier fixes to clean that up.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -19,6 +19,9 @@ env:
 
   DEFAULT_CFLAGS: "-Wall -Wextra -Werror -Wno-unused-parameter -Wno-unused-function"
 
+  # Apt sometimes likes to ask for user input, this will prevent that.
+  DEBIAN_FRONTEND: "noninteractive"
+
 jobs:
 
   prep:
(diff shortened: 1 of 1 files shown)


victorjulien:
Is this PR correct? I only see github-ci changes?

jasonish:
Oops. Crossed the streams there somehow. Real PR: https://github.com/OISF/suricata/pull/5320

PR #3889, opened 2019-05-27
suricatasc: fix reconnect

regit:
The reconnection system in suricatasc was not working properly. Restarting suricata was causing an error in suricatasc when trying to reopen the unix socket.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Fix reconnect

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/454
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/235

PR #5316, opened 2020-08-23
af-packet: increase block size default order to accommodate bigger frames

odedlaz:

https://redmine.openinfosecfoundation.org/issues/3889

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#3889](https://redmine.openinfosecfoundation.org/issues/3889)

Describe changes:

- af-packet: increase block size default order to accomodate bigger frames

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/source-af-packet.h
+++ b/src/source-af-packet.h
@@ -74,9 +74,9 @@ struct ebpf_timeout_config {
 
 /* In kernel the allocated block size is allocated using the formula
  * page_size << order. So default value is using the same formula with
- * an order of 3 which guarantee we have some room in the block compared
- * to standard frame size */
-#define AFP_BLOCK_SIZE_DEFAULT_ORDER 3
+ * an order of 5 which guarantee we have some room in the block compared
+ * to standard frame size and default localhost frame size */
+#define AFP_BLOCK_SIZE_DEFAULT_ORDER 5
 
 typedef struct AFPIfaceConfig_
 {


regit:
This MR is proposing a fix for loopback sniffing in tpacket_v2. It is changing the default block size in tpacket_v2 from 32768 to 131072. This way, the frame from loopback fits in. By the way, this may be not optimal at all because as we need room for metadata then only one full loopback frame is filling in a block. So switching to order 6 may be a better choice here.

This has a possible impact on performance as increasing the order augment the likelihood of a switch from page allocation to a regular kernel malloc (in af packet kernel code). I would feel confident with this patch with some feedback on performance for both low end and high end devices.

I'm wondering if switching to tpacket_v3 and using a correct block size when sniffing loopback could not be a better way to fix this issue. But this would mean we need a magic fix when using `-i lo` as it should switch to tpacket_v3 and increase the default block size.

odedlaz:
@regit, unless I'm mistaken this change shouldn't affect performance for MTU <= 1500, since the [smallest buffer size is selected](https://github.com/OISF/suricata/blob/928d4820f9e9ac26405dd3e4d2b6dfa56ec3255a/src/source-af-packet.c#L1875).

I can go ahead and increase the default order to `6`, but I haven't spotted any issues with `131072` - I ran suricata on `localhost` and everything worked fine. Also, I can introduce the `v3` fix. WDYT?

regit:
> @regit, I can reverse the for loop such that the [smallest buffer size will be selected](https://github.com/OISF/suricata/blob/928d4820f9e9ac26405dd3e4d2b6dfa56ec3255a/src/source-af-packet.c#L1875) instead of the current logic. WDYT?

I don't think it is a good idea. Bigger block size gives some continuity in memory making processing faster.

> I can go ahead and increase the default order to `6`, but I haven't spotted any issues with `131072` - I ran suricata on `localhost` and everything worked fine. 

Yes it will work fine, this will just be not optimal with one 65536 packet per block size of 131072.

> Also, I can introduce the `v3` fix.

It would be nice as this will bring better performances too. Even if I am not sure we care about performances on loopback catpure).

victorjulien:
No activity for a year, and it seems @regit isn't sold on this. So closing. Feel free to open a rebased version if this should still be considered.

PR #5326, opened 2020-08-25
plugins - rust plugin bootstrap support - v3

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5320

Abandon the idea of linking things up at runtime. It depends
on plugin style support, even if we haven't figure out plugins
on the target platform, like Windows.

Instead Rust plugins will have to call a "suricata::plugin::init()"
method to bootstrap their support on plugins.

Further, plugins must now define a function, SCPluginRegister() which
returns an SCPlugin struct instead of having an SCPluginSpec variable
name that the plugin loader expects to find. This make it much
simpler to define a plugin in Rust.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2516,6 +2516,12 @@ fi
 	[])
     AC_MSG_RESULT(yes)
 
+    RUST_FEATURES=""
+    AS_VERSION_COMPARE([$rustc_version], [1.38.0],
+        [],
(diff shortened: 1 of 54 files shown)


jasonish:
CI issue fixed. Next PR: https://github.com/OISF/suricata/pull/5343

PR #5329, opened 2020-08-27
Http2 upgrade v1

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3866

Describe changes:
- Adds mechanism to change protocol from HTTP1 to HTTP2

suricata-verify-pr: 308

Should make https://github.com/OISF/suricata-verify/pull/308 pass

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -31,6 +31,8 @@
 #include "app-layer-protos.h"
 #include "app-layer-expectation.h"
 #include "app-layer-ftp.h"
+#include "app-layer-htp.h"
+#include "app-layer-http2.h"
 #include "app-layer-detect-proto.h"
 #include "stream-tcp-reassemble.h"
 #include "stream-tcp-private.h"
@@ -630,13 +632,25 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
         }
     } else if (alproto != ALPROTO_UNKNOWN && FlowChangeProto(f)) {
         f->alproto_orig = f->alproto;
+        void * tx = NULL;
+        if (f->alproto_orig == ALPROTO_HTTP && f->alproto_expect == ALPROTO_HTTP2) {
+            //get tx and remove it from HTTP
+            tx = HtpGetTxForH2(f->alstate);
+        }
         SCLogDebug("protocol change, old %s", AppProtoToString(f->alproto_orig));
         AppLayerProtoDetectReset(f);
         /* rerun protocol detection */
         if (TCPProtoDetect(tv, ra_ctx, app_tctx, p, f, ssn, stream,
                            data, data_len, flags) != 0) {
             SCLogDebug("proto detect failure");
+            if (tx != NULL) {
+                HtpFreeTxFromH2(tx);
+            }
             goto failure;
+        }
+        if (f->alproto_orig == ALPROTO_HTTP && f->alproto_expect == ALPROTO_HTTP2 && tx != NULL) {
+            HTTP2MimicHttp1Request(tx, f->alstate);
+            HtpFreeTxFromH2(tx);
         }
         SCLogDebug("protocol change, old %s, new %s",
                 AppProtoToString(f->alproto_orig), AppProtoToString(f->alproto));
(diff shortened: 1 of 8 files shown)


victorjulien:
The commit messages are very sparse wrt explaining the implementation. Can you explain it?

catenacyber:
This comes mostly from https://tools.ietf.org/html/rfc7540#section-3.2

When we have a request with headers
```
Connection: Upgrade, HTTP2-Settings
Upgrade: h2c
HTTP2-Settings: <base64url encoding of HTTP/2 SETTINGS payload>
```
and a response with status 101 such as 
```
HTTP/1.1 101 Switching Protocols
Connection: Upgrade
Upgrade: h2c
```

We can change from HTTP1 to HTTP2, this is commit http2: first connection upgrade from http1

Furthermore, the server then does a HTTP2 reply to the HTTP1 request (hence the commit http2: mimic HTTP1 request from upgrade) so that we have a full HTTP2 transaction with response and mimicked request

Last, the HTTP1 request contains HTTP2 settings in its header, and we may want to match on them, hence the commit http2: settings from http1 upgrade

Is it clearer ?

src/app-layer.c
```diff
@@ -630,14 +632,26 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
         }
     } else if (alproto != ALPROTO_UNKNOWN && FlowChangeProto(f)) {
         f->alproto_orig = f->alproto;
+        void * tx = NULL;
+        if (f->alproto_orig == ALPROTO_HTTP && f->alproto_expect == ALPROTO_HTTP2) {
+            //get tx and remove it from HTTP
+            tx = HtpGetTxForH2(f->alstate);
```
catenacyber, line 638: 1) We must save the HTTP1 request before it gets freed

src/app-layer.c
```diff
@@ -630,14 +632,26 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
...
+        }
         SCLogDebug("protocol change, old %s", AppProtoToString(f->alproto_orig));
         AppLayerProtoDetectReset(f);
```
catenacyber, line 641: 2) All HTTP1 data gets freed here, except the transaction we hacked

src/app-layer.c
```diff
@@ -630,14 +632,26 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
...
         /* rerun protocol detection */
         if (TCPProtoDetect(tv, ra_ctx, app_tctx, p, f, ssn, stream,
                            data, data_len, flags) != 0) {
             SCLogDebug("proto detect failure");
+            if (tx != NULL) {
+                HtpFreeTxFromH2(tx);
```
catenacyber, line 647: 3. Do not leak it if this is not HTTP2 afterwards

src/app-layer.c
```diff
@@ -630,14 +632,26 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
...
+            }
             goto failure;
         }
+        if (f->alproto_orig == ALPROTO_HTTP && f->alproto_expect == ALPROTO_HTTP2 && tx != NULL) {
+            HTTP2MimicHttp1Request(tx, f->alstate);
+            HtpFreeTxFromH2(tx);
```
catenacyber, line 653: 4. Use it to create a fake HTTP2 request, and then finally free it

victorjulien:
Yeah I understand the flow better now, thanks.

So currently we have a parser that requests a protocol update. Then its state is freed and we rerun protocol detection and setup the new state for the new protocol, meaning (almost) all context is lost. I'm thinking we can rearrange things so that the original state remains valid for just a bit longer, so that the first setup for HTTP2 can still access the state for HTTP1. Then when the HTTP2 is setup we can free the HTTP1 state. This would mean we wouldn't have strange lifetime issue with the HTTP1 state already freed except for that one TX from it. Open question is where to keep the HTTP1 `alstate` in this process.

catenacyber:
> I'm thinking we can rearrange things so that the original state remains valid for just a bit longer, so that the first setup for HTTP2 can still access the state for HTTP1. Then when the HTTP2 is setup we can free the HTTP1 state. This would mean we wouldn't have strange lifetime issue with the HTTP1 state already freed except for that one TX from it. Open question is where to keep the HTTP1 alstate in this process.

This looks like a good idea.
I will try to implement this in the next PR

PR #5336, opened 2020-08-28
Reverted app-layer-expectation.c to the original suricata-6.0.0-beta1…

jbarfield:
… version and revised the queue.h for a cleaner solution to the previous build failures on FreeBSD 12.1 and 13 with clang10.0

Describe changes:
Updated the OISF included 1994 bsd src/queue.h header file to resolve build issues with Clang10 on both FreeBSD12.1-Release and FreeBSD13-Current. This fixes app-layer-expectation.c. If you noticed our previous pull request has now been deleted. It was chosen rather than to modify app-layer-expectation.c to the fix the queue.h header definitions instead which in turn should have also fixed the ctime errors reported after our initial commit/pull.

### DIFF
--- a/src/queue.h
+++ b/src/queue.h
@@ -30,10 +30,8 @@
  * SUCH DAMAGE.
  *
  *	@(#)queue.h	8.5 (Berkeley) 8/20/94
- */
-
-#ifndef	_SYS_QUEUE_H_
-#define	_SYS_QUEUE_H_
+ *      extended		       8/28/2020
+*/
 
 /*
  * This file defines five types of data structures: singly-linked lists,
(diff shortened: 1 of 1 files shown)


victorjulien:
If you have no intention of contributing the changes back to the main repo I think we should move the conversation elsewhere. Either a ticket or a forum post.

Anyway, I just tested a build with clang 10 on freebsd 12.1 and it compiles w/o warning. So please include the details to make sure it can be reproduced.

Thanks!

jbarfield:
My intention is to contribute this work to the main repo. I must have missed something in the process. What did I do incorrectly on this request? Or do you mean the queue.h repo directly? Either way this code update is free to distribute.

victorjulien:
Ah ok so then I guess I misunderstood the message in the previous PR. Signing the CLA is not optional. If you can't or don't want to sign it, we can't accept the PR. Thats why I suggested turning it into a ticket.

PR #5342, opened 2020-08-31
dns: simply skips zero-sized dns requests/responses

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3895

Describe changes:
- simply skips zero-sized dns requests/responses (instead of returning incomplete)

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -564,6 +564,9 @@ impl DNSState {
                 } else {
                     return AppLayerResult::err();
                 }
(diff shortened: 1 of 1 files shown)


catenacyber:
Should we set an event if this happens?

catenacyber:
Any ideas why Is CI not running ?

> Should we set an event if this happens?

I do not know for this one... It is easy to add it now,

victorjulien:
Merged in #5349, thanks!

PR #5345, opened 2020-09-01
Capture feature 120 v5

scottfgjordan:

Describe changes:
- The TcpSegment enrichment only will occur if the session-dump option is enabled in suricata.yaml nor will any allocations occur. 

Suggested Review Order/Guide:
- log-pcap-stream.c/.h
- detect-tag-pcap.c/h
- pcap-helper.c/h
 - This is basically the same as the code for libpcaps pcap_dump(), but rewritten so that it could dump a packet in two parts, e.g. header and payload written separately. This is currently written in Suricata codestyle and is internal to Suricata, but I'm open to approaching the libpcap team to get this functionality added to their pcap API.
- stream-tcp .c/.h files

Previous Pull Requests: 
https://github.com/OISF/suricata/pull/5217
https://github.com/OISF/suricata/pull/5291

**Changes from capture-feature-120-v4:**
- Corrected various minor comments from previous pull request.
- Allowed for more gentle and forgiving error handling during the setup with respect to file creation errors for the pcap output. Will no longer be a FatalError in the case of issues with the file path.
- No longer tracks pcap_cnt to determine the order of packet dumping as this variable was only available during offline pcap processing mode. Now relies on the timeval comparison function TIMEVAL_EARLIER to determine dump order for the packets.


doc/userguide/output/pcap-alert-output.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Suricata is able to provide a packet capture output upon the generation of an
+alert. This feature makes use of the "tag" keyword in a signature by searching
+for and dumping tagged packets that either have generated an alert or belong to
+the same session of a packet that has generated an alert. This feature must be
+enabled through the suricata.yaml file, but also requires a signature to have
+the "tag:session;" keyword for it to perform the capture. Signatures without
+"tag:session;" will not trigger a capture.
+
+Additionally, there is an option to enabled a session-dump in the suricata.yaml
```
norg, line 12: *enable* (remove the d)

doc/userguide/output/pcap-alert-output.rst
```diff
@@ -0,0 +1,53 @@
...
+file. This causes all available tcp segments in the relevant session at the time
+of alert to be dumped to the capture file. This is recommended, as otherwise
+often the packet that generates an alert may not be captured due to a lag
+between the processing of a packet and the generation of an alert. This
+behaviour was only found in tcp based traffic. As a result, the session-dump
+option only relates to tcp based traffic.
+
+By default, the stream-pcap-log is not enabled.
```
norg, line 20: Does this have an noticeable impact on performance? If yes, we should mention that like we do with profiling for example.

doc/userguide/output/pcap-alert-output.rst
scottfgjordan: Good catch thanks! Will fix that up.

doc/userguide/output/pcap-alert-output.rst
scottfgjordan: I can do a bit more analysis on the exact performance implications. There would be two levels of cost/performance differences:
1.  The stream-pcap-log function is enabled, but no tcp session dumping: This would just add in the cost associated with the pcap file setups and the file writes, i.e. the pcap dumps to file.
2. The stream-pcap-log function is enabled and tcp session dumping is too: Same costs associated with the above + some additional information to be held on to in the tcp segments so some additional memory allocations (performed with segment pool initialization) for the storage of this data and then also memcpy'ing the data from packet header into this storage as well. There could also be the case where a Realloc is performed if the packet header storage size is too small.

scottfgjordan: Is there a particular test framework you use / would suggest for evaluating the performance? I have some suricata verify cases that I could put together, but that's more so for validation of the output rather than performance evaluation right?

scottfgjordan:
I've responded to the two comments posted a while back, sorry for the delay in response I think I must have missed them when they were first posted.

doc/userguide/output/pcap-alert-output.rst
norg: I would recommend testing with t-rex which is a solid way of testing Suricata performance.

regit:
This is a really an interesting work. I have a few high level remarks:

On the concept itself, I really like the overall idea but:
- I don't think not reusing the log pcap module is a good idea. For example, a problem such as the log rotation on pcap size would need to be reimplemented if we want to deal with huge flows.
- Getting one single pcap file per stream in the same directory could end up in getting some millions on files in the same directory. This would lead to serious performance problem there when accessing the directory.
- It is also frequent to have mutiple alerts on the same packet so we will have duplicate.
- In some case, it would still be useful to have this one pcap file per alert but I did not get how the analyst can pass from the alert to the pcap. Maybe updating the alert with the key or filename used when a pcap is stored could be useful

On the form:
- one single commit on code result in something too big. It would be interesting to have at least a commit on the TCP segment update and one on the pcap logging addition
- detect-tag-pcap has some logging code (DumpTaggedPacket, ...), it would be appropriate to move that to the log log-pcap-stream

scottfgjordan:
Hi there,
Thanks for the feedback on this! I've been meaning to put some more time into this project and get some testing done with T-Rex and your feedback has got me focused again. On your comments:

> I don't think not reusing the log pcap module is a good idea. For example, a problem such as the log rotation on pcap size would need to be reimplemented if we want to deal with huge flows.

I will take a deeper look into this to see how I can better integrate this capture feature with the log-pcap module.

> Getting one single pcap file per stream in the same directory could end up in getting some millions on files in the same directory. This would lead to serious performance problem there when accessing the directory.

Agreed on this, I started with outputting the pcap file to a single directory just because it's easy for testing, but definitely not a solution that would scale very well. Curious what you might suggest as an alternative. 

> It is also frequent to have mutiple alerts on the same packet so we will have duplicate.

and

> In some case, it would still be useful to have this one pcap file per alert but I did not get how the analyst can pass from the alert to the pcap. Maybe updating the alert with the key or filename used when a pcap is stored could be useful.

This I'm not too sure about for how we'd want to craft a solution. As you mention below, it can be useful to have this one pcap file per alert. I think an approach where we get a tighter coupling between alerts and pcaps would be useful as you mention and is a good first step. I'll look in to some different approaches to this. 

> one single commit on code result in something too big. It would be interesting to have at least a commit on the TCP segment update and one on the pcap logging addition

I'll do this. Would you also prefer that they be separate pull requests as well? I.e. first I'll make a pull request for basic pcap logging, get that merged in, and then sort out the TCP segment updates in another PR? 

> detect-tag-pcap has some logging code (DumpTaggedPacket, ...), it would be appropriate to move that to the log log-pcap-stream

Will do.

regit:
I'm currently cooking a pull request that will implement conditional pcap capture (including the stream idea used here).

scottfgjordan:
> 
> 
> I'm currently cooking a pull request that will implement conditional pcap capture (including the stream idea used here).

Hey there, I've been working on something that exists in the pcap-log module. Sorry about not keeping you updated, hopefully there isn't too much overlap in effort. I'll get it out ASAP to see if there's some value there on your end.

scottfgjordan:
Closed as it there are new PRs https://github.com/OISF/suricata/pull/5828 and https://github.com/OISF/suricata/pull/5825

PR #5339, opened 2020-08-30
configure: Improved cross compilation support

jlucovsky:
Continuation of #5338 

This PR changes the run-time check for PCRE's "JIT exec" from a run-time to a compile time check for better support in cross compilation environments. Additionally, the run-time check to determine the size of a Lua integer is now guarded so it won't run on cross-compilation environments.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3872](https://redmine.openinfosecfoundation.org/issues/3872)

Describe changes:
- Uses `AC_LINK_ELSEIF`

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5348, opened 2020-09-01
configure: fix test for rust headers for cross compile - v1

jasonish:
As AC_CHECK_FILES cannot be used when cross compiling, just use
"if test -f ..." for the Rust headers.

Related comment:
https://github.com/OISF/suricata/pull/5339#issuecomment-684803566


victorjulien:
Merged through #5350

PR #5353, opened 2020-09-03
ikev1: add ikev1 parser - v6

fhonza:
Previous PR: [#5301](https://github.com//pull/5301)

Describe changes:
- tool function byte-array to hex-string without unsafe
- using debug_validate_bug_on macro for input checking pointers from C
- check ike version explicitly in detect.rs functions
- add >= and <= to DetectU8Data/DetectU32Data
- store nonce/payload data as raw-bytes
- removed transaction progress
- small refactorings


fhonza:
@catenacyber have not added a new suricata-verify PR because i've gotten into a conflict i'm not certain how to solve correctly.

There is an existing test for ikev2 which checks for ```alert ikev2 any any -> any any (msg:"SURICATA....``` 
Since both parsers are now united under ```ike``` these alerts won't work anymore.

Have debugged a bit, the parsers are registered in a loop in app-layer-detect-proto.c -> AppLayerProtoDetectRegisterProtocol() using for ike the ```rs_ike_register_parser()``` - Function which registers itself as "```ike```"

As i've just learned suricata-verify needs to work for older suricata versions as well.

Do you have an idea how to solve this, perhaps enable aliases somehow? Or would it be even ok in this case if existing rules change?

catenacyber:
> There is an existing test for ikev2 which checks for alert ikev2 any any -> any any (msg:"SURICATA....

How would you write such a rule now ?
`alert ike any any -> any any (msg:"SURICATA", ike.version:2, ....` ?

> As i've just learned suricata-verify needs to work for older suricata versions as well.

You can specify a `min-version: 6.0.0` for a test about IKEv1
But yes, we maintain suricata versions 4, 5 and 6

fhonza:
> How would you write such a rule now ?
> `alert ike any any -> any any (msg:"SURICATA", ike.version:2, ....` ?

Yes, protocol would be now `ike`. I've not added a ike.version-keyword, the keywords for both protocols are called `ike.xxx` and the ones specific for ikev1 `ikev1.xxx`, ikev2 has currently only the app-layer-event alerts.
The full updated rule is now:
`alert ike any any -> any any (msg:"SURICATA IKEv2 weak cryptographic parameters (Diffie-Hellman)"; flow:to_server; app-layer-event:ike.weak_crypto_dh; classtype:protocol-command-decode; sid:1; rev:1;)`

> You can specify a `min-version: 6.0.0` for a test about IKEv1

Thanks, didn't know it!

catenacyber:
So, yes it seems we need something like an alias translation `ikev2` to `ike` and maybe emitting a warning poor rule.
I do not know where is this code, but that should not be too hard to find starting from `SigInit`

fhonza:
yeah debugged a bit, the registration happens in app-layer-detect-proto.c, a data structure named alpd_ctx.alproto_names is filled with the different parsers.

On the other side when parsing the rules there is app-layer.c -> AppLayerGetProtoByName() which checks this alpd_ctx.alproto_names-structure and returns the fitting parser for the protoname.

I'll try to create two entries in alpd_ctx.alproto_names and see how far i come....

catenacyber:
so @fhonza should I review this PR or wait for the next one ?

fhonza:
How you like, the things you found from the last PR should be solved now.

If it's easier for you i can link v7 to v5 later, the you need to review only one PR.

At least i hope that i find a good solution for this alias thing :)

catenacyber:
I will not put it on top of priority list for now then ;-)

PR #5354, opened 2020-09-03
Http2 upgrade v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3866

Describe changes:
- Adds mechanism to change protocol from HTTP1 to HTTP2

suricata-verify-pr: 320

Makes https://github.com/OISF/suricata-verify/pull/320 pass

Modifies #5329 by having the HTTP state surviving a bit longer :
- `AppLayerProtoDetectReset` does not call `AppLayerParserStateCleanup`
- Adds `AppLayerParserStateProtoCleanup` which uses different parameters (from the previous protocol)
- we do not need to keep track of `upgrade_tx`
- we do not need to close connection to get HTTP1 logged after `AppLayerRequestProtocolChange` 
- added a defensive programming check of `NULL` http1 transaction in `HTTP2MimicHttp1Request`
+ rebased and each commit clang-formatted 

So, the process is now 
- save context, ie`alstate`
- run `AppLayerProtoDetectReset` resetting but not cleaning
- run `TCPProtoDetect` protocol detection
- run the tasks needing both old original and new expected context ie `HTTP2MimicHttp1Request`
- run `AppLayerParserStateProtoCleanup` cleaning

The latest commit needs likely to be squashed into a previous one, but I thought this is easier to review for now

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -31,6 +31,8 @@
 #include "app-layer-protos.h"
 #include "app-layer-expectation.h"
 #include "app-layer-ftp.h"
+#include "app-layer-htp.h"
+#include "app-layer-http2.h"
 #include "app-layer-detect-proto.h"
 #include "stream-tcp-reassemble.h"
 #include "stream-tcp-private.h"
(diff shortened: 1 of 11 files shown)


victorjulien:
Converted to draft as it tests an outdated SV PR as well as due to messy git history.

src/app-layer.c
```diff
@@ -631,13 +633,21 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
...
+        // we delay AppLayerParserStateCleanup because we may need previous parser state
         AppLayerProtoDetectReset(f);
         /* rerun protocol detection */
         if (TCPProtoDetect(tv, ra_ctx, app_tctx, p, f, ssn, stream,
                            data, data_len, flags) != 0) {
             SCLogDebug("proto detect failure");
+            AppLayerParserStateProtoCleanup(f->protomap, f->alproto_orig, alstate_orig, alparser);
             goto failure;
         }
+        if (f->alproto_orig == ALPROTO_HTTP && f->alproto_expect == ALPROTO_HTTP2) {
```
victorjulien, line 647: the goal should be to get rid of this and instead move this to the http2 implementation, which should then be able to access the old http1 state to do its initial setup

catenacyber: Ok, so adding a field `alstate_orig` to the Flow structure ?

victorjulien: Thats the easiest way, but I'm not completely happy with it. It means we'll spend 8 extra bytes for each extra flow. I wonder if we can do something like have a special flag that indicates that we're going to perform an upgrade, so `f->alstate` contains the old state, which will be replaced by the new state in `AppLayerParserParse`. Since we need it only at one moment, perhaps we can expose the old state to the new parsers state creation function via a simple thread local variable. A bit hacky as well though.

catenacyber: > It means we'll spend 8 extra bytes for each extra flow.

Indeed, and this current PR does not need those extra memory consumption

> which will be replaced by the new state in AppLayerParserParse

We need this new state in `HTTP2MimicHttp1Request`

So, if this is to be called from http2 implementation, this can only be in its `AppLayerParserParse`
So, we have alstate, pstate, and `alp_tctx->alproto_local_storage[f->protomap][alproto]` available
Is this last one your suggestion so that HTTP2 parser when checking a flag, we access there somehow the HTTP1 state

I am not sure it is better in the HTTP2 implementation, than in this condition with `FlowChangeProto`

Another solution is to call a new generic function `AppLayerParserProtoChange(f, alstate_orig)` and make it call some new parser's defined function `ProtoChange`

victorjulien:
Replaced by #5362

catenacyber:
Actually, this is still in competition with #5390

PR #5366, opened 2020-09-04
http/eve: use set_string_from_bytes where appropriate - v1

jasonish:
The JsonBuilder set_string_from_bytes will escape unprintable
characters in the output stream. Using BytesToStringBuffer
can generate invalid UTF-8 which prevents the conversion from
a C string to a Rust string.

### DIFF
--- a/src/output-json-http.c
+++ b/src/output-json-http.c
@@ -204,10 +204,8 @@ static void EveHttpLogJSONBasic(JsonBuilder *js, htp_tx_t *tx)
 {
     /* hostname */
     if (tx->request_hostname != NULL) {
-        const size_t size = bstr_len(tx->request_hostname) * 2 + 1;
-        char string[size];
-        BytesToStringBuffer(bstr_ptr(tx->request_hostname), bstr_len(tx->request_hostname), string, size);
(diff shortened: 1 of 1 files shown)


catenacyber:
Can we even remove the `BytesToStringBuffer` definition ?

jasonish:
> Can we even remove the `BytesToStringBuffer` definition ?

No, not yet.  We still have to use it here:

https://github.com/jasonish/suricata/blob/http-eve-set-string-from-bytes/v1/src/output-json-http.c#L250

As the converted string is modified before passing to JsonBuilder.

catenacyber:
> > Can we even remove the `BytesToStringBuffer` definition ?
> 
> No, not yet. We still have to use it here:
> 
> https://github.com/jasonish/suricata/blob/http-eve-set-string-from-bytes/v1/src/output-json-http.c#L250
> 
> As the converted string is modified before passing to JsonBuilder.

We could have a workaround to use something else than `strchr`, but this is no big deal

victorjulien:
Merged in #5376, thanks!

Can you add a ticket for this issue?

PR #5369, opened 2020-09-04
rules - validate rule input as utf-8 - v1

jasonish:
Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3850

- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/495
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/853


victorjulien:
The formatting check is a bit ... greedy. It tries to reformat more than the changed lines, adding the entire enum it seems. @roligugus any idea if we can have it be less greedy?

victorjulien:
Merged in #5376, thanks!

roligugus:
> The formatting check is a bit ... greedy. It tries to reformat more than the changed lines, adding the entire enum it seems. @roligugus any idea if we can have it be less greedy?

@victorjulien Sorry for being late to the party. Vacation, kids, school start and work picking up. Seen that you've already merged.
Just played around a bit. Behaviour seems that it adjust the whole enum for any changes in the enum as you've seen here with the comments being fixed. Oddly enough, it does not do that for structs.

Unfortunately, would not know of any knob to change that. :(

Probably would require into looking into clang-format source code to see if that behaviour could be tailored somehow. I'll throw that onto my clang-format pile to investigate "weird behaviour". No ETA on that though.

roligugus:
> Probably would require into looking into clang-format source code to see if that behaviour could be tailored somehow. I'll throw that onto my clang-format pile to investigate "weird behaviour". No ETA on that though.

@victorjulien It's a feature/bug/behaviour of the clang-format (C++) binary. One can reproduce it with calling `clang-format src/util-error.h -lines x:x` where x is the one line number that was changed in this PR (368). I.e. there isn't any work-around that we can do by copying more of clang's clang-format scripts and fixing them. This needs to be fixed in the clang-format C++ code base.

PR #5374, opened 2020-09-06
Af packet fixes 6

regit:
Update bypass code to deal with flow engine update.

- https://redmine.openinfosecfoundation.org/issues/3861
- https://redmine.openinfosecfoundation.org/issues/3902

Describe changes:
- Fix 2 description texts
- Update code to new API

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/539
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/315

### DIFF
--- a/src/flow-hash.c
+++ b/src/flow-hash.c
@@ -849,20 +849,25 @@ static inline int FlowCompareKey(Flow *f, FlowKey *key)
  *  \retval f *LOCKED* flow or NULL
  */
 
-Flow *FlowGetFromFlowKey(FlowKey *key, struct timespec *ttime, const uint32_t hash)
+Flow *FlowGetFromFlowKey(ThreadVars *th_v, FlowLookupStruct *fls, FlowKey *key, struct timespec *ttime,
+                         const uint32_t hash)
 {
     Flow *f = FlowGetExistingFlowFromHash(key, hash);
 
     if (f != NULL) {
         return f;
     }
-    /* TODO use spare pool */
-    /* now see if we can alloc a new flow */
-    f = FlowAlloc();
+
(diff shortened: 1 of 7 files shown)


src/flow-hash.c
```diff
@@ -849,20 +849,25 @@ static inline int FlowCompareKey(Flow *f, FlowKey *key)
...
 
     if (f != NULL) {
         return f;
     }
-    /* TODO use spare pool */
-    /* now see if we can alloc a new flow */
-    f = FlowAlloc();
+
+    /* get a flow from the spare queue */
+    f = FlowQueuePrivateGetFromTop(&fls->spare_queue);
```
victorjulien, line 862: this is still missing a step. The pool remains empty. The spare pool is now a per thread construct, where it gets a 'budget' from the master pool. See https://github.com/OISF/suricata/blob/master/src/flow-hash.c#L551 in the normal lookup path.

regit: OK, there is a problem with FlowSpareSync: What am I suppose to use for packet, should I created a pseudo packet ? or should I cook a derivative of FlowSpareSync that don't need a Packet ?

regit: By looking at FlowSpareSync I could maybe just patch it to support a NULL value for Packet as it is only used for emergency mode. This way, we would have a regular call here.

regit: Hum I could just call it with emerg set to false and this should pass.

victorjulien: Or create your own version of the FlowSpareSync func.

PR #5377, opened 2020-09-07
Af packet fixes 6 v2

regit:
Update of #5374 

Describe changes:
- Remove patch on bypass callback
- Real usage of flow pool system 

- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/540
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/316

### DIFF
--- a/src/flow-hash.c
+++ b/src/flow-hash.c
@@ -506,9 +506,7 @@ static inline Flow *FlowSpareSync(ThreadVars *tv, FlowLookupStruct *fls,
         f = FlowQueuePrivateGetFromTop(&fls->spare_queue);
         spare_sync = true;
     }
-#ifdef UNITTESTS
     if (tv && fls->dtv) {
-#endif
         if (spare_sync) {
             if (f != NULL) {
                 StatsAddUI64(tv, fls->dtv->counter_flow_spare_sync_avg, fls->spare_queue.len+1);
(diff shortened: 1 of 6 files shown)


src/flow-hash.c
```diff
@@ -506,9 +506,7 @@ static inline Flow *FlowSpareSync(ThreadVars *tv, FlowLookupStruct *fls,
         f = FlowQueuePrivateGetFromTop(&fls->spare_queue);
         spare_sync = true;
     }
-#ifdef UNITTESTS
```
victorjulien, line 509: why not support the counters from this path as well? Seems useful to me.

regit: Currently the only user of flow bypass manager is af-packet and with the new implementation (5.1) the thread starts run the init function and leave as the recurring check function is not implemented. The currently used init function read the pinned maps and create the flows based on the data collected. So we could have a counter but it will be for a one shot function at start.

victorjulien: I don't understand this, what is version 5.1 here?

So when does this thing run? Does it even make sense to use a spare pool here?

regit: Oups, please read Suricata 5.0.x not 5.1.

It runs at start of Suricata to load the flows from the flow bypass table stored in kernel.This is only active when the eBPF maps are pinned to resist a Suricata restart.

victorjulien: If its not performance critical lets keep things as simple as possible.

regit: OK, in this case I think we have no active code to add here. I will keep the doc update for later to not add more on top of your current work.

PR #5383, opened 2020-09-08
http2: returns error in case of index 0

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3907

Describe changes:
- Checks that index value for header is 0 and returns an error in this case
- adds a unit test against that behavior


catenacyber:
Merged in https://github.com/OISF/suricata/pull/5388

PR #5385, opened 2020-09-08
mqtt: limit size of variable integer

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3906

Describe changes:
- Limits the variable integer size to 4 bytes during MQTT parsing (as it is a variable uint32 up to 0xFFFFFF) 

@satta what do you think ?
I vaguely remember we discussed a bit about this `assert`

### DIFF
--- a/rust/src/mqtt/parser.rs
+++ b/rust/src/mqtt/parser.rs
@@ -52,17 +52,6 @@ fn convert_varint(continued: Vec<u8>, last: u8) -> u32 {
     return value;
 }
 
-#[inline]
-fn varint_length(val: usize) -> usize {
-  match val {
-      0 ..= 127 => 1,
-      128 ..= 16383 => 2,
-      16384 ..= 2097151 => 3,
-      2097152 ..= 268435455 => 4,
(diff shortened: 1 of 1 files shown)


satta:
I think this makes sense. I tried to think hard about a case where after successful parsing of the remaining length, the change could lead to a problem, but could not find one. One thing to keep in mind is that the range of the remaining length is limited to less than the full unsigned 32 bits, since one bit per byte is used as a continuation signal (http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/csprd02/mqtt-v3.1.1-csprd02.html#_Toc385349759). So I would add some tests for the boundary cases around this limit. You can cherry-pick https://github.com/satta/suricata/commit/440d9957c7fa6d44f29d04684fd4011fb7a89c5f if you want.

catenacyber:
Thanks @satta 
This got merged when you were commenting.
Do you want to make a PR with your commit ?

catenacyber:
Merged in #5388

satta:
Sure, no problem. See #5391

PR #5365, opened 2020-09-04
http: disables lzma by default for HTTP

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3776

Describe changes:
- Disables lzma by default

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

libhtp-pr: 302

direct reference OISF/libhtp#302

Modifies #5364 with rebase (just to launch CI again with another libhtp PR, which itself is just a rebase)

PR #5389, opened 2020-09-09
http: disables lzma by default for HTTP

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3776

Describe changes:
- Disables lzma by default

libhtp-pr: 303

direct reference OISF/libhtp#303

Modifies #5365 with using new libhtp `htp_config_set_lzma_layers`

clang-format did not like nice spaces for
`+#define HTP_CONFIG_DEFAULT_LZMA_LAYERS 0U`

### DIFF
--- a/doc/userguide/configuration/suricata-yaml.rst
+++ b/doc/userguide/configuration/suricata-yaml.rst
@@ -1358,8 +1358,8 @@ use of libhtp.
        # Default value of randomize-inspection-range is 10.
        #randomize-inspection-range: 10
 
-       # Can disable LZMA decompression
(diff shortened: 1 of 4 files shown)


catenacyber:
It looks like commit-check does not use libhtp PR, hence its failure

victorjulien:
> It looks like commit-check does not use libhtp PR, hence its failure

If you introduce a new API call should we not check for it in configure?

catenacyber:
> If you introduce a new API call should we not check for it in configure?

Good idea, doing so

PR #5382, opened 2020-09-08
log: Ensure threaded eve honors SIGHUP

jlucovsky:
This commit ensures that all logging contexts register for the file
rotation mechanism (SIGHUP and configured).

Describe changes:
- Register file rotation flag when creating context
- Set `is_regular` correctly

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/3915

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5394, opened 2020-09-09
log: Ensure threaded eve honors SIGHUP

jlucovsky:
Continuation of #5382 

This commit ensures that all logging contexts register for the file
rotation mechanism (SIGHUP and configured) and that the suffixes use for threaded EVE output are continuous.

Describe changes:
- Ensure EVE suffixes are continuuous.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/3915

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -736,8 +736,10 @@ static bool LogFileNewThreadedCtx(LogFileCtx *parent_ctx, const char *log_path, 
     thread->threaded = false;
     thread->parent = parent_ctx;
     thread->id = thread_id;
+    thread->is_regular = true;
     thread->Write = SCLogFileWriteNoLock;
     thread->Close = SCLogFileCloseNoLock;
+    OutputRegisterFileRotationFlag(&thread->rotation_flag);
 
     parent_ctx->threads->lf_slots[thread_id] = thread;
     return true;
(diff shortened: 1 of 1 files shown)


src/util-logopenfile.c
```diff
@@ -713,15 +716,22 @@ LogFileCtx *LogFileEnsureExists(LogFileCtx *parent_ctx, int thread_id)
  */
 static bool LogFileNewThreadedCtx(LogFileCtx *parent_ctx, const char *log_path, const char *append, int thread_id)
 {
+    static bool eve_suffix_init = false;
```
victorjulien, line 719: is this shared between threads as well? If so it should probably be an atomic too?

jasonish:
First thing I noticed is that this starts with eve.json.0, which is fine on its own.  However, our pcap.log starts at 1. I wonder if we should be consistent here.

src/util-logopenfile.c
jasonish: Looks like it is.

jasonish (commented):
Tests good, but it looks like that one var should be atomic.

@victorjulien Any thoughts on the thread id numbering? I think I'd like to be consistent with the pcap-log and start at 1. But OK to leave it at 0 and never think of it again.

victorjulien:
> @victorjulien Any thoughts on the thread id numbering? I think I'd like to be consistent with the pcap-log and start at 1. But OK to leave it at 0 and never think of it again.

I prefer it to be consistent, so lets start at 1.

jlucovsky:
Continued in #5399

PR #5397, opened 2020-09-09
Datasets memcap/v1

inashivb:
Allow memcap and hashsize to be set via rules or via yaml. In case of `load`, set the memcap to highest possible value unless defined.

### DIFF
--- a/src/datasets.c
+++ b/src/datasets.c
@@ -177,7 +177,7 @@ static int DatasetLoadMd5(Dataset *set)
         /* straight black/white list */
         if (strlen(line) == 33) {
             line[strlen(line) - 1] = '\0';
-            SCLogDebug("line: '%s'", line);
+            //SCLogDebug("line: '%s'", line);
 
             uint8_t hash[16];
             if (HexToRaw((const uint8_t *)line, 32, hash, sizeof(hash)) < 0)
@@ -246,7 +246,7 @@ static int DatasetLoadSha256(Dataset *set)
         /* straight black/white list */
         if (strlen(line) == 65) {
             line[strlen(line) - 1] = '\0';
-            SCLogDebug("line: '%s'", line);
+//            SCLogDebug("line: '%s'", line);
 
             uint8_t hash[32];
             if (HexToRaw((const uint8_t *)line, 64, hash, sizeof(hash)) < 0)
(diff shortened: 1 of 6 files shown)


src/detect-datarep.c
```diff
@@ -169,6 +170,18 @@ static int DetectDatarepParse(const char *str,
                 SCLogDebug("load %s", val);
                 strlcpy(load, val, load_size);
             }
+            if (strcmp(key, "memcap") == 0) {
+                if (StringParseUint64(memcap, 10, 0, val) < 0) {
```
victorjulien, line 174: This should probably be `ParseSize*` as then we can use things like `1gb`.

src/datasets.c
```diff
@@ -357,6 +357,8 @@ static int DatasetLoadString(Dataset *set)
         }
     }
 
+    // TODO figure out how to reset memcap after load is finished, a reset function? a macro?
```
victorjulien, line 360: Maybe just a small util func like `THashConsolidateMemcap` which takes the memuse and sets memcap to that value. You can call it directly after the load completes.

src/detect-dataset.c
```diff
@@ -104,7 +105,8 @@ static int DetectDatasetParse(const char *str,
         char *name, int name_len,
         enum DatasetTypes *type,
         char *load, size_t load_size,
-        char *save, size_t save_size)
+        char *save, size_t save_size,
+        uint64_t *memcap, uint32_t *hashsize) // TODO datatype for hashsize?
```
victorjulien, line 109: `uint32_t` is what I'd expect

src/util-thash.c
```diff
@@ -291,10 +291,13 @@ static void THashInitConfig(THashTableContext *ctx, const char *cnf_prefix)
 }
 
 THashTableContext* THashInit(const char *cnf_prefix, size_t data_size,
-    int (*DataSet)(void *, void *),
+     int (*DataSet)(void *, void *),
      void (*DataFree)(void *),
      uint32_t (*DataHash)(void *),
-     bool (*DataCompare)(void *, void *))
+     bool (*DataCompare)(void *, void *),
+     bool reset_memcap,
```
victorjulien, line 298: doesn't look like its used, or needed either

inashivb: This is for the special case of `load`, since we want to make sure that everything from the dataset file is loaded.
if memcap is defined in rule/yaml, set that
else check if load was the cmd (with this param) and let the memcap be `UINT64_MAX`
if it wasn't a load cmd, set memcap to `THASH_DEFAULT_MEMCAP`.

Should this not be done?

victorjulien: ah yes I missed the actual use, please ignore my comment

inashivb:
Replaced with #5398

PR #5398, opened 2020-09-10
Datasets memcap/v2

inashivb:
Known fixes to be done:
- Unnecessary Debug statements
- TODO corresponding to `GetDefaultMemcap`
- Commit history and messages
- Docs updation

### DIFF
--- a/src/datasets.c
+++ b/src/datasets.c
@@ -177,7 +177,7 @@ static int DatasetLoadMd5(Dataset *set)
         /* straight black/white list */
         if (strlen(line) == 33) {
             line[strlen(line) - 1] = '\0';
-            SCLogDebug("line: '%s'", line);
+            //SCLogDebug("line: '%s'", line);
 
             uint8_t hash[16];
             if (HexToRaw((const uint8_t *)line, 32, hash, sizeof(hash)) < 0)
@@ -246,7 +246,7 @@ static int DatasetLoadSha256(Dataset *set)
         /* straight black/white list */
         if (strlen(line) == 65) {
             line[strlen(line) - 1] = '\0';
-            SCLogDebug("line: '%s'", line);
+//            SCLogDebug("line: '%s'", line);
 
             uint8_t hash[32];
             if (HexToRaw((const uint8_t *)line, 64, hash, sizeof(hash)) < 0)
@@ -314,7 +314,7 @@ static int DatasetLoadString(Dataset *set)
         char *r = strchr(line, ',');
         if (r == NULL) {
             line[strlen(line) - 1] = '\0';
-            SCLogDebug("line: '%s'", line);
+//            SCLogDebug("line: '%s'", line);
 
             // coverity[alloc_strlen : FALSE]
             uint8_t decoded[strlen(line)];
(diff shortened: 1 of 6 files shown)


src/datasets.c
```diff
@@ -593,11 +609,45 @@ void DatasetPostReloadCleanup(void)
     SCMutexUnlock(&sets_lock);
 }
 
+static void GetDefaultMemcap(uint64_t *memcap, uint32_t *hashsize) // TODO Do something about this name! OR Save default memcap in dataset struct?
```
inashivb, line 612: @victorjulien could you please check if this is a decent way of retrieving defaults from yaml? This shall be called per rule and once during Init.

victorjulien: might be able to simplify this to something like:

```c
const char *str = NULL;
if (ConfGetValue("datasets.defaults.memcap", &str) == 1) {
    if (ParseSizeStringU64(str, memcap) < 0) {
        FatalError(SC_ERR_FATAL, "memcap value cannot be deduced: %s", str);
    }
}
```
I think the yaml part is a map, not a list, so we can avoid the loop.

inashivb: OK. Thank you!

src/util-thash.c
```diff
@@ -320,6 +329,14 @@ THashTableContext* THashInit(const char *cnf_prefix, size_t data_size,
     return ctx;
 }
 
+/* \brief Set memcap to current memuse
+ * */
+void THashConsolidateMemcap(THashTableContext *ctx)
+{
+    ctx->config.memcap = SC_ATOMIC_GET(ctx->memuse);
```
victorjulien, line 336: wondering if we should update this to make the memcap something like `MAX(SC_ATOMIC_GET(ctx->memuse), default_memcap)`. If the configs memcap is higher than the memuse of the loaded set, use the memcap setting instead.

inashivb: Makes sense.

src/datasets.c
```diff
@@ -593,11 +609,45 @@ void DatasetPostReloadCleanup(void)
...
+{
+    ConfNode *datasets = ConfGetNode("datasets");
+    if (datasets != NULL) {
```
jlucovsky, line 615: nit/style: If `datasets == NULL`, return. This will remove one level of indentation

src/detect-dataset.c
```diff
@@ -181,7 +184,7 @@ static int DetectDatasetParse(const char *str,
                         "'load' can only appear once");
                     return -1;
                 }
-                SCLogDebug("load %s", val);
+                SCLogDebug("load %s size: %lu", val, load_size);
```
victorjulien, line 187:
```
Invalid printf with size_t (not casted to uintmax_t) at /builds/inliniac/suricata-ci/suricata/src/detect-dataset.c:187
Makefile:545: recipe for target 'check' failed
```
To print a `sizet_t` use `SCLogDebug("load %s size: %"PRIuMAX, val, (uintmax_t)load_size);`

inashivb:
Replaced with #5401

PR #5401, opened 2020-09-10
Datasets memcap/v3

inashivb:

Rule example:
```
```

`suricata.yaml` example:
```
datasets:
  defaults:
    memcap: 100mb
    hashsize: 2048
  ua-seen:
    type: string
    load: datasets.csv
    memcap: 20mb
    hashsize: 2048
```

### DIFF
--- a/src/datasets.c
+++ b/src/datasets.c
@@ -216,6 +216,7 @@ static int DatasetLoadMd5(Dataset *set)
                     (uint32_t)strlen(line), line);
         }
     }
+    THashConsolidateMemcap(set->hash);
 
     fclose(fp);
     SCLogConfig("dataset: %s loaded %u records", set->name, cnt);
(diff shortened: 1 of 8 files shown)


src/datasets.c
```diff
@@ -593,11 +605,29 @@ void DatasetPostReloadCleanup(void)
     SCMutexUnlock(&sets_lock);
 }
 
+static void GetDefaultMemcap(uint64_t *memcap, uint32_t *hashsize)
+{
+    const char *str = NULL;
+    if (ConfGetValue("datasets.defaults.memcap", &str) == 1) {
+        if (ParseSizeStringU64(str, memcap) < 0) {
+            FatalError(SC_ERR_FATAL, "memcap value cannot be deduced: %s", str);
```
inashivb, line 613: Should we fail here or fallback to defaults in the code with a warning?

inashivb: Yes OK.

jasonish: Can we store the parsed values instead of re-parsing them every time?

jasonish (commented):
Nothing jumps out at me.  Tests OK.

inashivb:
Thanks, @jasonish! I created #5404 which deals with invalid memcap and hashsize values with warnings instead of fatal errors, could you please have a look there?

PR #5404, opened 2020-09-10
Datasets memcap/v4

inashivb:

Rule example:
```
```

`suricata.yaml` example:
```
datasets:
  defaults:
    memcap: 100mb
    hashsize: 2048
  ua-seen:
    type: string
    load: datasets.csv
    memcap: 20mb
    hashsize: 2048
```

### DIFF
--- a/src/datasets.c
+++ b/src/datasets.c
@@ -421,8 +421,8 @@ Dataset *DatasetFind(const char *name, enum DatasetTypes type)
     return set;
 }
 
-Dataset *DatasetGet(const char *name, enum DatasetTypes type,
-        const char *save, const char *load, uint64_t memcap, uint32_t hashsize)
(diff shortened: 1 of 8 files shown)


src/datasets.c
```diff
@@ -636,29 +676,55 @@ int DatasetsInit(void)
                 }
             }
 
+            ConfNode *set_memcap = ConfNodeLookupChild(iter, "memcap");
+            if (set_memcap) {
+                if (ParseSizeStringU64(set_memcap->val, &memcap) < 0) {
+                    SCLogWarning(SC_ERR_SIZE_PARSE,
+                            "memcap value cannot be"
+                            " deduced: %s, resetting to default",
+                            set_memcap->val);
```
jasonish, line 685: very minor: but I think we should stick with "invalid" or "parse" error here.

inashivb:
Replaced with #5406

PR #5406, opened 2020-09-10
Datasets memcap/v5

inashivb:

Rule example:
```
```

`suricata.yaml` example:
```
datasets:
  defaults:
    memcap: 100mb
    hashsize: 2048
  ua-seen:
    type: string
    load: datasets.csv
    memcap: 20mb
    hashsize: 2048
```


victorjulien:
Merged in #5407, thanks!

PR #5416, opened 2020-09-11
Enum dir rust/v1

inashivb:

### DIFF
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -39,6 +39,12 @@ pub const STREAM_GAP:      u8 = 0x10;
 pub const STREAM_DEPTH:    u8 = 0x20;
 pub const STREAM_MIDSTREAM:u8 = 0x40;
 
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum Direction {
+    ToServer = 0x04,
+    ToClient = 0x08,
+ }
+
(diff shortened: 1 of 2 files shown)


inashivb:
@jasonish @victorjulien do you like this?
Any changes I should make?
Should I do the same for other protocols as well?

catenacyber:
Looks a good idea.
Can it be exported to C ? Like `SshStateFinished` for an example

catenacyber:
(so as to have only one definition)

rust/src/core.rs
```diff
@@ -39,6 +39,12 @@ pub const STREAM_GAP:      u8 = 0x10;
 pub const STREAM_DEPTH:    u8 = 0x20;
 pub const STREAM_MIDSTREAM:u8 = 0x40;
 
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum Direction {
+    ToServer = 0x04,
+    ToClient = 0x08,
+ }
+
```
jasonish, line 47: Can you try adding repr(C) here as well?  Then maybe it can be the source of truth and used from C as well.

jasonish (changes requested):
I'd like to see this rebased.  Its a good candidate to become an enum on the Rust side I think.

inashivb:
Replaced w #6294

PR #5420, opened 2020-09-16
doc/userguide: include man page even when not including pdf - v1

jasonish:
Fix a mistake in Makefile.am where the man page was only being added to the
distribution if the PDF was also created. It should be included even if the PDF
cannot be included.


norg:
LGTM

norg (approved):

victorjulien:
Merged in #5442, thanks!

PR #5421, opened 2020-09-17
fuzz: better configure check for rust nightly

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None

Describe changes:
- Improves configure check for Rust nightly version, in order to build with MSAN if needed

It seems #5411 configure/bash magic was not good enough for oss-fuzz Docker, even if it worked locally on my Mac

PR #5411, opened 2020-09-11
Fuzz msan v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None

Describe changes:
- Enables MSAN by recompiling rust std library with fuzz targets
- Fixes a use of uninitialized value

Cf google/oss-fuzz#3469 (comment)

Follows https://github.com/OISF/suricata/pull/5409 with checking rust compiler is nightly

PR #5422, opened 2020-09-17
fuzz: better configure checks for MSAN building

catenacyber:
None

Describe changes:
- Improves configure check for Rust nightly version, in order to build with MSAN if needed

It seems #5411  was not good enough for oss-fuzz Docker

Replaces #5421 with version working bot locally and on oss-fuzz with https://github.com/catenacyber/oss-fuzz/commit/4c922f1cc0ea3a9683ca6e3d9e1218e6f28956df

Let's check if CI is happy with other systems

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -472,47 +472,6 @@
 # options
 
 
-    AC_ARG_ENABLE(fuzztargets,
-        AS_HELP_STRING([--enable-fuzztargets], [Enable fuzz targets]),[enable_fuzztargets=$enableval],[enable_fuzztargets=no])
-    AM_CONDITIONAL([BUILD_FUZZTARGETS], [test "x$enable_fuzztargets" = "xyes"])
-    AM_CONDITIONAL([RUST_BUILD_STD], [test "x$enable_fuzztargets" = "xyes" && echo $rust_compiler_version | grep -q nightly])
-    AC_PROG_CXX
-    AS_IF([test "x$enable_fuzztargets" = "xyes"], [
-        AC_DEFINE([FUZZ], [1], [Fuzz targets are enabled])
-        AC_DEFINE([AFLFUZZ_NO_RANDOM], [1], [Disable all use of random functions])
-        CFLAGS_ORIG=$CFLAGS
-        CFLAGS="-Werror"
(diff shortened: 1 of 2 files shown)


jasonish:
Would it be useful to duplicate the ubuntu-18-04-fuzz CI test, use Rust nightly then do a post-configure check that this was setup correctly?

catenacyber:
> Would it be useful to duplicate the ubuntu-18-04-fuzz CI test, use Rust nightly then do a post-configure check that this was setup correctly?

For at least one with fuzzing, I think it would be useful to have Rust nightly, as it is needed to get the best out of fuzzing Rust.
Between the Debian one, or the Ubuntu one, I do not know if it makes a difference

jasonish:
> > Would it be useful to duplicate the ubuntu-18-04-fuzz CI test, use Rust nightly then do a post-configure check that this was setup correctly?
> 
> For at least one with fuzzing, I think it would be useful to have Rust nightly, as it is needed to get the best out of fuzzing Rust.
> Between the Debian one, or the Ubuntu one, I do not know if it makes a difference

I'd probably go with a new Ubuntu 20.04 test.

catenacyber:
Ok, so... do you want this new ubuntu test qs part of this PR ?

jasonish:
> Ok, so... do you want this new ubuntu test qs part of this PR ?

Doesn't really matter.  But should probably be based on this PR.

catenacyber:
@jasonish do you manage to install/run dustup in Github CI builds ?
Or do you know an equivalent of `rustup component add rust-src` without using rustup ?

jasonish:
> @jasonish do you manage to install/run dustup in Github CI builds ?
> Or do you know an equivalent of `rustup component add rust-src` without using rustup ?

Rustup is used in GitHub CI builds, see: https://github.com/OISF/suricata/blob/master/.github/workflows/builds.yml#L916 for an example.

catenacyber:
> Rustup is used in GitHub CI builds

I see its download, but then I cannot use it as it is not in the PATH cf 
https://github.com/catenacyber/suricata/runs/1165320022?check_suite_focus=true

victorjulien:
Merged in #5452, thanks!

PR #5423, opened 2020-09-17
doc/userguide: add info about --set and lists - v1

jasonish:
Add more info about --set and lists.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2211


norg:
LGTM, although Travis CI fails but that doesn't look like being related to the commit.

victorjulien:
Merged in #5452, thanks!

PR #5445, opened 2020-09-25
applayer: fix a leak in protocol change

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3924

Describe changes:
- Fixes memory leak in case of protocol change (and potential use after free)
- Adds debug validation for other cases

Leak happened when 
1. there was a protocol change (for instance from HTTP to TLS with `CONNECT`, but it could be other protocols)
2. Call to `TCPProtoDetect` made these changes
2.1 Its call to `AppLayerProtoDetectGetProto` changed `f->alproto` to `ALPROTO_TLS`
2.2 Its call to `AppLayerParserParse` (parsing the TLS data) returned -1 error 
3. We cleant the previous memory state with `AppLayerParserStateProtoCleanup`
4. We forgot the new TLS state by setting `f->alstate = NULL;`  which got leaked

Fix is to tie together the memory cleaning and setting `f->alstate = NULL;`

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -1196,7 +1196,7 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow 
     if (flags & STREAM_GAP) {
         if (!(p->option_flags & APP_LAYER_PARSER_OPT_ACCEPT_GAPS)) {
             SCLogDebug("app-layer parser does not accept gaps");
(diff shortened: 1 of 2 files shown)


pevma:
So far so good, seeing no leaks  with :
```
This is Suricata version 6.0.0-dev (2b61596cd 2020-09-25)
```

pevma:
No leaks  - looks good from that perspective :)

PR #5444, opened 2020-09-25
log/eve: Threaded filename change: eve.N.json

jlucovsky:
Continuation of #5443 

This commit changes the name of the file used with threaded eve logging
to better support log rotation

Instead of using "eve.json.N" and creating potential issues with log
rotation (which also uses a ".N" suffix), the eve logs will be named
"eve.N.json" when threaded.

Describe changes:
- Applies clang-format suggestions

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5443, opened 2020-09-25
log/eve: Threaded filename change: eve.N.json

jlucovsky:
This commit changes the name of the file used with threaded eve logging
to better support log rotation

Instead of using "eve.json.N" and creating potential issues with log
rotation (which also uses a ".N" suffix), the eve logs will be named
"eve.N.json" when threaded.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Add utility function to create filename when eve json output is threaded.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5447, opened 2020-09-26
log/eve: Threaded filename change: eve.N.json

jlucovsky:
Continuation of #5444

This commit changes the name of the file used with threaded eve logging
to better support log rotation

Instead of using "eve.json.N" and creating potential issues with log
rotation (which also uses a ".N" suffix), the eve logs will be named
"eve.N.json" when threaded.

Describe changes:
- Updates documentation and yaml configuration template
- Improved handling of filenames without extensions.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/util-logopenfile.c
+++ b/src/util-logopenfile.c
@@ -44,8 +44,8 @@
 static bool LogFileNewThreadedCtx(LogFileCtx *parent_ctx, const char *log_path, const char *append, int i);
 static bool SCLogOpenThreadedFileFp(const char *log_path, const char *append, LogFileCtx *parent_ctx, int slot_count);
 
-// Threaded eve.json suffixes
-static SC_ATOMIC_DECL_AND_INIT_WITH_VAL(uint32_t, eve_file_suffix, 1);
+// Threaded eve.json identifier
+static SC_ATOMIC_DECL_AND_INIT_WITH_VAL(uint32_t, eve_file_id, 1);
 
 #ifdef BUILD_WITH_UNIXSOCKET
 /** \brief connect to the indicated local stream socket, logging any errors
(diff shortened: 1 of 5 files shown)


jlucovsky:
appveyor failed while building Rust components: https://ci.appveyor.com/project/OISF-CI/suricata/builds/35420209/job/b7ljltwrl0p8cfn0#L1321

jlucovsky:
@jasonish I'll rebuild a new PR to exclude the transforms rst file that was incorrectly added to a commit. However, I'll await any comments from you before doing so.

src/util-logopenfile.c
```diff
@@ -708,6 +708,37 @@ LogFileCtx *LogFileEnsureExists(LogFileCtx *parent_ctx, int thread_id)
...
+                    "Invalid filename for threaded mode \"%s\"; "
+                    "filenames must include an extension, e.g: \"name.ext\"",
+                    original_name);
+        }
+        SCFree(tname);
+    } else {
+        snprintf(threaded_name, len, "%s.%d", original_name, unique_id);
+    }
+    return true;
+}
```
jasonish, line 740: This is probably an edge case.  But if I just change the filename, say to "eve". We hit the else case, and we use filenames like "eve.1", "eve.2", and so on.

If I add the command line option, `-l ./log`, then we find a dot, but not in the basename, and we hit the error condition.

I'm OK with ending up with filenames like "eve.1", but I think the check for a "." should be on the basename only, not the full path.

jlucovsky: Good points. I'll make an update.

jlucovsky:
Continued in #5443

PR #5482, opened 2020-10-12
detect/mark:Fix unittest into FAIL/PASS API

JananiRamjee:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [ .] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ .] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ .] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4052

Describe changes:
- Removed initial test methods in detect-mark.c
- Used FAIL or PASS macros described in util-unittest.h
- FAIL/PASS macros determine conditionals for return value(0 or 1 respectively)

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5485, opened 2020-10-12
util/unittest:Add PASS_IF_NULL macro to the FAIL/PASS API

JananiRamjee:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4066

Describe changes:
- Added PASS_IF_NULL macro to util-unittest.h
- pass if the value provided to it is NULL.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/util-unittest.h
+++ b/src/util-unittest.h
@@ -119,6 +119,17 @@ extern int unittests_fatal;
         PASS;              \
     } while (0)
 
+/**
+ * \brief Pass the test if expression evaluates to true.
+ *
+ * Only to be used at the end of a function instead of returning the
+ * result of an expression.
+ */
+#define PASS_IF_NULL(expr)                                                                         \
+    do {                                                                                           \
+        PASS_IF(NULL == expr);                                                                     \
+    } while (0)
+
 #endif /* __UTIL_UNITTEST_H__ */
 
 /**


JananiRamjee:
Hello, 
Requesting feedback on the pull request. Thank you!

src/util-unittest.h
```diff
@@ -119,6 +119,17 @@ extern int unittests_fatal;
         PASS;              \
     } while (0)
 
+/**
+ * \brief Pass the test if expression evaluates to true.
+ *
+ * Only to be used at the end of a function instead of returning the
```
inashivb, line 125: Hmm but we do not seem to require to use it at the end of the function in detect-mark.c, right?

src/util-unittest.h
```diff
@@ -119,6 +119,17 @@ extern int unittests_fatal;
...
+ * result of an expression.
+ */
+#define PASS_IF_NULL(expr)                                                                         \
```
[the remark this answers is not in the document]
JananiRamjee: Even i noticed that. I did put it close, but after running git clang-format, the `\` along with the do statement shifts away. Will try once again to align properly

src/util-unittest.h
JananiRamjee: I added that since the PASS_IF macro also had that description in util-unittest.h, though it wasn't required to use at the end of function in detect-mark.c. Please do correct me if I am wrong.

src/util-unittest.h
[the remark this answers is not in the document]
inashivb: Oh OK. Then maybe its a clang thing.

src/util-unittest.h
inashivb: hmm I see that `PASS_IF` is only used at one place and it really is used at the end of the function. Although there could be places when it is required in the middle of a function. This makes me wonder if PASS_* statements are always expected be at the end of the function (which means detect-mark test should be modified in that manner) or should this comment be updated. @victorjulien ?

JananiRamjee: Should i update the comment and make the new pull request?

inashivb: hmm I'd wait for Victor to respond. @victorjulien ?

JananiRamjee: @victorjulien, please advice on whether to update the comment.

victorjulien: The goal of new (and updated) unittests should be that they run from start to finish (top to bottom) in the pass case. So a pass mid-function violates that. We don't want any logic that runs in a 'fail' case, other than checking for the fail itself. No cleanup, no additional checks in the failure path.

JananiRamjee: So, a pass function must ultimately run after ensuring that there are no failures, right? So, shall i add a comment as below @victorjulien 

> Only to be used to check pass case after checking all fail cases

JananiRamjee: @inashivb @victorjulien may I update with the above comment?
Please advice

victorjulien: I'm questioning my own logic about the `PASS_IF`. If the "pass" should be last, this should also be true for `PASS_IF`. But what happens if the condition is false? Do we implicitly return `FAIL` in this case? Or would we need something like
```c
PASS_IF(r == 0);
FAIL;
```
But that would violate the "pass should be last rule". We could also do something like `PASS_IF_OR_FAIL(r == 0)` but that becomes a bit too convoluted I think.

So I think I would prefer not introducing a `PASS_IF` after all and just write the tests in such a way that when we reach the end we can issue a `PASS`.

@jasonish any thoughts?

jasonish: PASS_IF is terminal isn't it? It just bundles together a FAIL_IF and a PASS.  I see its only used once. I say we get rid of it. While there is nothing wrong with it, everything else is FAIL type expressions, then we have this one PASS_IF expression which you have to invert your logic for making it hard to read.

JananiRamjee: Please advice on what is to be done further for the issue. Shall i remove the PASS_IF macros and also update #5482 accordingly?

victorjulien: Yeah lets not introduce `PASS_IF`. Sorry we put you on the wrong track.

victorjulien:
Closing as we changed our minds on this feature. Thanks for your work on it!

PR #5490, opened 2020-10-15
github-ci: remove deprecated set-env and add-path - v1

jasonish:
Use the new methods for setting an environment variable and
updating the PATH.

Addresses the warnings we are getting from GitHub CI:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


victorjulien:
Merged in #5533, thanks!

PR #5494, opened 2020-10-16
af-packet: use configured cluster-id when checking for fanout - v1

jasonish:
When testing for fanout support a cluster-id of 1 was always being
used instead of the configured cluster-id. This limited fanout
support to only one Suricata instance.

Instead of hardcoding an ID of 1, use the configured cluster-id.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/3419

- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/859
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/501

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -1977,7 +1977,7 @@ mmap_err:
 /** \brief test if we can use FANOUT. Older kernels like those in
  *         CentOS6 have HAVE_PACKET_FANOUT defined but fail to work
  */
-int AFPIsFanoutSupported(int cluster_id)
+int AFPIsFanoutSupported(uint16_t cluster_id)
 {
 #ifdef HAVE_PACKET_FANOUT
     int fd = socket(AF_PACKET, SOCK_RAW, htons(ETH_P_ALL));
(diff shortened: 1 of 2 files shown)


jlucovsky:
Should `cluster_id` within `AFPThreadVars` be `uint16_t` instead of `int`?

regit (approved):
Looks good to me.

src/source-af-packet.c
```diff
@@ -1977,16 +1977,15 @@ static int AFPSetupRing(AFPThreadVars *ptv, char *devname)
...
 {
 #ifdef HAVE_PACKET_FANOUT
     int fd = socket(AF_PACKET, SOCK_RAW, htons(ETH_P_ALL));
     if (fd < 0)
         return 0;
 
     uint32_t mode = PACKET_FANOUT_HASH | PACKET_FANOUT_FLAG_DEFRAG;
-    uint16_t id = 1;
-    uint32_t option = (mode << 16) | (id & 0xffff);
+    uint32_t option = (mode << 16) | (cluster_id & 0xffff);
```
victorjulien, line 1988: What is the effect of `<u16> & 0xffff`? Would it ever be something else than just `<u16>`?

jasonish: Ah, no.. With the type of the function changed that could not be removed.

victorjulien (changes requested):
Minor comments to be addressed.

jasonish:
Continued at https://github.com/OISF/suricata/pull/6340.

PR #5497, opened 2020-10-18
Task 4053, convert unit tests to FAIL/PASS API

jufajardini:
detect/msg: convert to FAIL/PASS API

https://redmine.openinfosecfoundation.org/issues/4053

Describe changes:
- detect/msg: convert to FAIL/PASS API
- Partially closes redmine ticket 2975

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-msg.c
+++ b/src/detect-msg.c
@@ -125,89 +125,66 @@ error:
 #ifdef UNITTESTS
 static int DetectMsgParseTest01(void)
 {
-    int result = 0;
     Signature *sig = NULL;
     const char *teststringparsed = "flow stateless to_server";
     DetectEngineCtx *de_ctx = DetectEngineCtxInit();
-    if (de_ctx == NULL)
-        goto end;
+    FAIL_IF(de_ctx == NULL);
 
     FILE *fd = SCClassConfGenerateValidDummyClassConfigFD01();
     SCClassConfLoadClassficationConfigFile(de_ctx, fd);
 
     sig = SigInit(de_ctx, "alert tcp any any -> any any (msg:\"flow stateless to_server\"; flow:stateless,to_server; content:\"flowstatelesscheck\"; classtype:bad-unknown; sid: 40000002; rev: 1;)");
-    if(sig == NULL)
-        goto end;
+    FAIL_IF(sig == NULL);
 
-    if (strcmp(sig->msg, teststringparsed) != 0) {
-        printf("got \"%s\", expected: \"%s\": ", sig->msg, teststringparsed);
-        goto end;
-    }
+    FAIL_IF(strcmp(sig->msg, teststringparsed) != 0);
 
-    result = 1;
-end:
     if (sig != NULL)
         SigFree(de_ctx, sig);
     if (de_ctx != NULL)
         DetectEngineCtxFree(de_ctx);
-    return result;
+
+    PASS;
 }
 
 static int DetectMsgParseTest02(void)
 {
-    int result = 0;
     Signature *sig = NULL;
     const char *teststringparsed = "msg escape tests wxy'\"\\;:";
     DetectEngineCtx *de_ctx = DetectEngineCtxInit();
-    if (de_ctx == NULL)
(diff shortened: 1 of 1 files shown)


victorjulien:
In the git commit message, "detect/msg: convert to FAIL/PASS API" should be the subject line. Referencing to a ticket would be something like "Issue #xxxx" in the message body

src/detect-msg.c
```diff
@@ -125,89 +125,66 @@ static int DetectMsgSetup (DetectEngineCtx *de_ctx, Signature *s, const char *ms
 #ifdef UNITTESTS
 static int DetectMsgParseTest01(void)
 {
-    int result = 0;
     Signature *sig = NULL;
     const char *teststringparsed = "flow stateless to_server";
     DetectEngineCtx *de_ctx = DetectEngineCtxInit();
-    if (de_ctx == NULL)
-        goto end;
+    FAIL_IF(de_ctx == NULL);
```
victorjulien, line 131: We have a slightly cleaner macro for checking pointers: `FAIL_IF_NULL(ptr)`.

src/detect-msg.c
```diff
@@ -125,89 +125,66 @@ static int DetectMsgSetup (DetectEngineCtx *de_ctx, Signature *s, const char *ms
...
 
-    if (strcmp(sig->msg, teststringparsed) != 0) {
-        printf("got \"%s\", expected: \"%s\": ", sig->msg, teststringparsed);
-        goto end;
-    }
+    FAIL_IF(strcmp(sig->msg, teststringparsed) != 0);
 
-    result = 1;
-end:
     if (sig != NULL)
```
victorjulien, line 141: Since `sig` can now never be `NULL` here, we should get rid of the check.

In general the goal of using the new API is to get branchless tests.

src/detect-msg.c
```diff
@@ -125,89 +125,66 @@ static int DetectMsgSetup (DetectEngineCtx *de_ctx, Signature *s, const char *ms
...
     const char *teststringparsed = "flow stateless to_server";
     DetectEngineCtx *de_ctx = DetectEngineCtxInit();
-    if (de_ctx == NULL)
-        goto end;
+    FAIL_IF(de_ctx == NULL);
 
     FILE *fd = SCClassConfGenerateValidDummyClassConfigFD01();
     SCClassConfLoadClassficationConfigFile(de_ctx, fd);
 
     sig = SigInit(de_ctx, "alert tcp any any -> any any (msg:\"flow stateless to_server\"; flow:stateless,to_server; content:\"flowstatelesscheck\"; classtype:bad-unknown; sid: 40000002; rev: 1;)");
```
victorjulien, line 136: extra points for: declare `sig` here. This leads to more compact tests.

victorjulien (changes requested):

src/detect-msg.c
jufajardini: Understood! I saw that option at first, but wasn't really sure it was ready for using, because of the issue open that seems to be related to that...

src/detect-msg.c
jufajardini: Got it, thanks. Following this logic, could lines 
```
FILE *fd = SCClassConfGenerateValidDummyClassConfigFD01();
SCClassConfLoadClassficationConfigFile(de_ctx, fd);
```
Be moved to the end, where there used to be the check
```
if(sig != NULL)
```
I don't see fd being directly used anywhere else, which makes me believe they could. I have checked SCClassConfLoadClassficationConfigFile and it seems it will only use de_ctx, not change it. So it seems that this is something that could sort of be together with the other statements near 
```
PASS
```
Does that make sense?

jufajardini:
Thanks for the overall comments. I believe I have done the appropriate changes, and I'll make a new PR now.

PR #5498, opened 2020-10-18
[Draft] detect: Improve support for larger address values.

jlucovsky:
[Draft PR] SV tests will be created once comments on the initial approach have been addressed/discussed.

This PR modifies the address rule value parsing to permit larger address values in the rule variables section.

When a value larger than the default (8192 bytes) is to be used, adjust the size of `detect.max-address-size`.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3887](https://redmine.openinfosecfoundation.org/issues/3887)

Describe changes:
- Determine the largest address size permitted from the configuration settings
- Set the size of the container to hold the address being parsed using this value or a default.
- New test cases to ensure maximum value is honored.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:


victorjulien:
Could we simply make it dynamic/unlimited?

jlucovsky:
> Could we simply make it dynamic/unlimited?

Of course but I suspect fuzzing will (eventually) raise an issue.

How about using a reasonably sized maximum permitted value with a suitable default max value?

victorjulien:
I'm not too excited about adding a configuration parameter for this. Wrt the fuzzing, I think that is a concern, although I think by just adding more variables we can get there as well. E.g.
```
A: [ addresses to max size ]
B: [ addresses to max size ]
C: [ $A, $B ]
```

jasonish:
Not of fan of adding a configuration variable here either. 

Is there a better way to support large lists of IP addresses? Like I dataset? I have a PR somewhere around here where you could do:
`alert any @filename any -> any any ...` where filename would contain an IP address (CIDR) per line and construct it into a list while loading... But it also seems that just quadrupling the size would go along way as well.

jlucovsky:
Ok .... I'll remove the configuration setting and rebuild a PR.

jlucovsky:
Continued in #5598

PR #5499, opened 2020-10-18
rust/applayer - derive utility functions for app-layer event enums - v3

jasonish:
This PR shows how we can use a procedural macro to derive code that is common across all Rust app-layers that have an enum of events.

Given an enum like:
```
#[derive(Debug, PartialEq, AppLayerEvent)]
pub enum DNSEvent {
    MalformedData,
    NotRequest,
    NotResponse,
    ZFlagSet,
}
```
the following utility functions will be generated:
- from_id
- as_i32
- to_cstring
- from_cstring

These are functions that each parser pretty much implements itself.

We can take this one step further and use generic functions for `get_event_info_by_id` and `get_event_info`. However, as generic functions can't be extern "C" we do need a little wrapper around them.. For example:
```
#[no_mangle]
pub extern "C" fn rs_mqtt_state_get_event_info_by_id(
    event_id: std::os::raw::c_int,
    event_name: *mut *const std::os::raw::c_char,
    event_type: *mut core::AppLayerEventType,
) -> i8 {
    get_event_info_by_id::<MQTTEvent>(event_id, event_name, event_type)
}
```

But they can used for any event enum that derives (or implements the trait) `AppLayerEvent`.

At some point I'd like to get rid of the wrapper used here but that would require some much larger changes to the Rust parser.

- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/862
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/504


jasonish:
Commits will have to be refactored some to pass the commit-check check.

victorjulien:
Could the wrapper func(s) be hidden (or further simplified) by some form of macro?

jasonish:
> Could the wrapper func(s) be hidden (or further simplified) by some form of macro?

Addressed in the next PR: https://github.com/OISF/suricata/pull/5501

PR #5500, opened 2020-10-19
ikev1: add ikev1 parser - v10

fhonza:
Previous PR: #5456 

Describe changes:

* restructured commits
* documentation update

suricata-verify-pr: 343


fhonza:
@catenacyber hi, the git-clang-format format want's these 3 changes:
 src/app-layer-protos.c | 3 ++-
 src/rust-context.h     | 2 +-
 src/util-profiling.c   | 2 +-

This would lead to a different "style" than the surrounding code, example for util-profiling.c:
```
         CASE_CODE (LOGGER_JSON_KRB5);
-        CASE_CODE (LOGGER_JSON_IKE);
+        CASE_CODE(LOGGER_JSON_IKE);
         CASE_CODE (LOGGER_JSON_FTP);
```
Seems that these files were not run with clang-format before?

jasonish:
One thing I noticed in the output is that the `event_type` is `ike`, but the event object will be `ikev1` or `ikev2` depending on the protocol version which breaks our pattern of the `event_type` matching the event object.

Are there any incompatibilities between these 2 objects? For example, where there is the same field name with a different data type?

Can they just be under the `ike` object as we already have a `version_major`?  Even if there are incompatibilities a better approach might to be create `ikev1` and `ikev2` objects within the `ike` object.  But probably better if we can avoid that.

jasonish:
Re: Documentation

Can you please separate the documentation commits from the code commits?

We're also going to need an upgrade guide. By that I mean some documentation on the the field name changes in the ike output.  The upgrade section of the documentation could use some re-working around this type of upgrade documentation, so maybe just a section in `upgrade.rst` for now, which we'll rework later on.

fhonza:
> One thing I noticed in the output is that the `event_type` is `ike`, but the event object will be `ikev1` or `ikev2` depending on the protocol version which breaks our pattern of the `event_type` matching the event object.
> 
> Are there any incompatibilities between these 2 objects? For example, where there is the same field name with a different data type?
> 
> Can they just be under the `ike` object as we already have a `version_major`? Even if there are incompatibilities a better approach might to be create `ikev1` and `ikev2` objects within the `ike` object. But probably better if we can avoid that.

The proposed ikev1 parser and the existing ikev2 parser are two completely independent implementations with a different featureset, although there is some overlap. I just wanted to minimize changes of existing contributions. In the next PR i can try to unify the logging output under `ike`, but there will be several fields missing in case of ikev2.
(for ikev1 I'll check whats missing to be on the same level as ikev2)

And i'll separate the documentation-commits & check upgrade.rst

jasonish:
Missing fields is fine, as long as the ones that are in common have the same meaning.

If that condition can't be met, it might be better to stick with event_type's of "ikev1" and "ikev2" so the object names match, I'd rather not have that but its preferable to an "event_type" of "ike" with no associated "ike" object.

fhonza:
> Missing fields is fine, as long as the ones that are in common have the same meaning.
> 
> If that condition can't be met, it might be better to stick with event_type's of "ikev1" and "ikev2" so the object names match, I'd rather not have that but its preferable to an "event_type" of "ike" with no associated "ike" object.

Yes that's my plan now that the same fields will have the same data-type and semantic, there will be a few fields which are exclusively for each protocol.
Thought to put these into a ikev1/ikev2-Field below ike, hope that's ok.

PR #5510, opened 2020-10-22
Smbshare keyword v2

norg:

https://redmine.openinfosecfoundation.org/issues/3236

Describe changes:
- add legacy smb_share keyword output
- add legacy smb_named_pipe keyword output
- includes clang-format changes
- split commits

- PR norg: https://buildbot.openinfosecfoundation.org/builders/norg/builds/77
- PR norg-pcap: https://buildbot.openinfosecfoundation.org/builders/norg-pcap/builds/75


src/detect-smb-share.c
```diff
@@ -143,15 +151,19 @@ void DetectSmbShareRegister(void)
...
     sigmatch_table[KEYWORD_ID].alias = KEYWORD_NAME_LEGACY;
     sigmatch_table[KEYWORD_ID].Setup = DetectSmbShareSetup;
-    sigmatch_table[KEYWORD_ID].flags |= SIGMATCH_NOOPT|SIGMATCH_INFO_STICKY_BUFFER;
+    sigmatch_table[KEYWORD_ID].flags |= SIGMATCH_NOOPT | SIGMATCH_INFO_STICKY_BUFFER;
     sigmatch_table[KEYWORD_ID].desc = "sticky buffer to match on SMB share name in tree connect";
 
-    DetectAppLayerMpmRegister2(BUFFER_NAME, SIG_FLAG_TOSERVER, 2,
-            PrefilterGenericMpmRegister, GetShareData,
-            ALPROTO_SMB, 1);
+    sigmatch_table[KEYWORD_ID_LEGACY].name = KEYWORD_NAME_LEGACY;
```
victorjulien, line 157: I don't see the value of this. We already have `alias` in the original code. We can use that for output and such without growing the sigmatch_table

victorjulien (changes requested):
I'd like to see a different approach

src/detect-smb-share.c
norg: Please correct me if I'm wrong but alias itself won't be enough since when I run `list-keywords` it will call `SigTableList` which would just use those keywords with a `.name` to be printed and not when passed with the `.alias` name.

So for example:
```
--list-keywords=tls_cert_subject 
Non existing keyword
```

Even though it's an alias for `tls.cert_subject`. Maybe we want to print `.alias` as well with the `--list-keywords` and reference to the actual `name`? That would avoid several changes on the protocol code where I would argue we would have to add the old/alternative names to the table with `.name`.

PR #5512, opened 2020-10-26
ikev1: add ikev1 parser - v11

fhonza:
Previous PR: #5500

Describe changes:

* unify logging for ikev1/ikev2
* updated documentation into separate commit
* clang-format applied

suricata-verify-pr: 351


catenacyber:
https://github.com/OISF/suricata-verify/pull/351 for GitHub mentions

src/util-profiling.c
```diff
@@ -1284,57 +1326,57 @@ const char * PacketProfileDetectIdToString(PacketProfileDetectId id)
  *
  * \retval string equivalent for the LoggerId id
  */
-const char * PacketProfileLoggertIdToString(LoggerId id)
+const char *PacketProfileLoggertIdToString(LoggerId id)
```
catenacyber, line 1329: Could we restrict the clang-format to the changes in the file and not the whole file ?

That should be something like 
- `git add`
- `git clang-format`
- `git commit` ie commits without the clang-format changes
- `git commit -a` ie commits the clang-format changes

doc/userguide/rules/ike-keywords.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Match against the length of the public key exchange payload (e.g. Diffie-Hellman) of the server or client.
+
+This keyword takes a numeric argument after a colon and supports additional qualifiers, such as:
+
+* ``=`` (equal)
+* ``>`` (greater than)
+* ``<`` (less than)
+* ``>=`` (greater than or equal)
+* ``<=`` (less than or equal)
```
catenacyber, line 110: There is also a range cf `DETECT_UINT_RA`

catenacyber:
Running `git grep -i ikev2`
there are still some `IKEv2` patterns in file rules/ipsec-events.rules, for some rules titles
Should they be changed ?

Also, did you run a specific command for commit `ikev1: rename ikev2 to common ike` ?

catenacyber:
Travis is failing because it is not using S-V PR
@jasonish should we change something in our CI ?

rules/ipsec-events.rules
```diff
@@ -4,16 +4,17 @@
...
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 no encryption (AH)"; flow:to_client; app-layer-event:ikev2.no_encryption; classtype:protocol-command-decode; sid:2224008; rev:1;)
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 invalid proposal"; flow:to_server; app-layer-event:ikev2.invalid_proposal; classtype:protocol-command-decode; sid:2224009; rev:1;)
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 invalid proposal selected"; flow:to_client; app-layer-event:ikev2.invalid_proposal; classtype:protocol-command-decode; sid:2224010; rev:1;)
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 unknown proposal"; flow:to_server; app-layer-event:ikev2.unknown_proposal; classtype:protocol-command-decode; sid:2224011; rev:1;)
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 unknown proposal selected"; flow:to_client; app-layer-event:ikev2.unknown_proposal; classtype:protocol-command-decode; sid:2224012; rev:1;)
+alert ike any any -> any any (msg:"SURICATA IKEv2 malformed request data"; flow:to_server; app-layer-event:ike.malformed_data; classtype:protocol-command-decode; sid:2224000; rev:1;)
```
catenacyber, line 7: Should this change be part of this commit ?

rust/src/ike/ikev1.rs
```diff
@@ -0,0 +1,157 @@
...
+
+use std;
+use crate::core::{STREAM_TOSERVER};
+use crate::applayer::*;
+use nom;
+use crate::ike::parser::*;
+use crate::ike::ike::{IKEState, IkeEvent};
+use std::collections::HashSet;
+use crate::common::to_hex;
+
```
catenacyber, line 28: Could you run `rustfmt rust/src/ike/*.rs` on some of these files ?

catenacyber: @jasonish  do you know if we could add this `rustfmt` as a CI check like clang-format ?

src/output-json-ike.c
```diff
@@ -179,6 +199,4 @@ void JsonIKELogRegister(void)
     OutputRegisterTxSubModule(LOGGER_JSON_IKE, "eve-log", "JsonIKELog", "eve-log.ike",
             OutputIKELogInitSub, ALPROTO_IKE, JsonIKELogger, JsonIKELogThreadInit,
             JsonIKELogThreadDeinit, NULL);
-
-    SCLogDebug("IKE JSON logger registered.");
```
catenacyber: Why did you remove this ?

src/output-json-ike.c
```diff
@@ -150,6 +169,7 @@ static TmEcode JsonIKELogThreadInit(ThreadVars *t, const void *initdata, void **
     }
 
     *data = (void *)thread;
+
```
catenacyber, line 172: nit

src/output-json-ike.c
```diff
@@ -135,12 +152,14 @@ static TmEcode JsonIKELogThreadInit(ThreadVars *t, const void *initdata, void **
 
     if (initdata == NULL) {
         SCLogDebug("Error getting context for EveLogIKE.  \"initdata\" is NULL.");
-        goto error_exit;
+        SCFree(thread);
```
catenacyber, line 155: Why this change ?
Was this a bug in IKEv2 ?

catenacyber (commented):
This looks good enough to me.
I think it deserves a second reviewer since this is a big contribution

catenacyber:
Maybe we could merge some of the first commits like
- detect: add comparison-mode LTE/GTE for Detect(U32/u8)Data
- util: add function converting u8-array into a hex-String
- detect: added support for protocol-aliases
- ikev1: rename ikev2 to common ike
I think we did this for HTTP2 and it was then simpler to review the core part

src/util-profiling.c
fhonza: hi, that was even the normal behaviour of git-clang-format to format only the modified lines by the commit.
There was a comment a few PR's back to fix these changes because they looked differently than the "surrounding code".

The only solution I saw to make the pipeline run and resolve the comment was to clang-format the whole file :)

doc/userguide/rules/ike-keywords.rst
fhonza: Yes you are right! Forgot to add this after switching to this helper....

catenacyber: not sure it makes sense for these keywords, but it is possible ;-)

src/util-profiling.c
catenacyber: Fair enough for me :-)

fhonza:
> Running `git grep -i ikev2`
> there are still some `IKEv2` patterns in file rules/ipsec-events.rules, for some rules titles
> Should they be changed ?
> 
> Also, did you run a specific command for commit `ikev1: rename ikev2 to common ike` ?

To be honest I'm not sure what the entries in rules/ipsec-events.rules are used for. Are they only examples?
Without knowing it I changed as less as possible...

For the renaming commit I used sed like you suggested, worked well, had only some problems in the rust-code (the rusticata dependencies)

src/output-json-ike.c
fhonza: hmmm not sure, from having a look in the code both should work, can throw a few of my pcaps against it to test this change.

src/output-json-ike.c
fhonza: Got the impression that newer parsers don't log that, don't know if there is a best practice....

catenacyber: Ok for me

src/output-json-ike.c
catenacyber: I would rather keep the code unmodified if there is no problem

catenacyber:
> To be honest I'm not sure what the entries in rules/ipsec-events.rules are used for. Are they only examples?
Without knowing it I changed as less as possible...

They are not examples.
They are real rules using app-layer events.
I think they are enabled by default

> For the renaming commit I used sed like you suggested, worked well, had only some problems in the rust-code (the rusticata dependencies)

Could you put in the commit comment the command you typed ? (lots of comm in this sentence...)

jasonish:
> @jasonish should we change something in our CI ?

This would require doing pull-request parsing over at Travis-CI as well. GitHub CI for master now covers everything Travis did, so we can either ignore Travis, or better yet, remove it from master.

rust/src/ike/ikev1.rs
jasonish: > @jasonish do you know if we could add this `rustfmt` as a CI check like clang-format ?

No, not really.  Rustfmt can only operate on complete files, not just changes like clang.  So we can't do much until we reformat all the code.

catenacyber:
> so we can either ignore Travis, or better yet, remove it from master.

Oh yeah let's remove it and delete some code :-)

rust/src/ike/ikev1.rs
catenacyber: Thanks Jason.
Franck, can you run rustfmt on some of these whole files ?

fhonza: yes, i'm just waiting a bit for further comments and would incorporate everything in v12?

PR #5514, opened 2020-10-27
readthedocs: Replace dead link from public-data-sets.rst

lukashino:

- [X] have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Describe changes:
- On ll.mit.edu it is not possible to find any dataset in cyber section (or any section that would give plausible data).
- Site mawi.wide.ad.jp is dead

### DIFF
--- a/doc/userguide/public-data-sets.rst
+++ b/doc/userguide/public-data-sets.rst
@@ -3,7 +3,7 @@ Public Data Sets
 
 Collections of pcaps for testing and profiling.
 
-DARPA sets: http://www.ll.mit.edu/mission/communications/cyber/CSTcorpora/ideval/data/
+DARPA sets: https://www.ll.mit.edu/r-d/datasets?author=All&rdarea=All&rdgroup=All&keywords=cyber&tag=All&items_per_page=10
 
 MAWI sets (pkt headers only, no payloads): http://mawi.wide.ad.jp/mawi/samplepoint-F/2012/
 


doc/userguide/public-data-sets.rst
```diff
@@ -3,10 +3,6 @@ Public Data Sets
 
 Collections of pcaps for testing and profiling.
 
-DARPA sets: http://www.ll.mit.edu/mission/communications/cyber/CSTcorpora/ideval/data/
```
jasonish, line 6: I wonder if this was just moved: http://www.ll.mit.edu/r-d/datasets/1998-darpa-intrusion-detection-evaluation-dataset

lukashino: Interesting, I haven't found that and still can not if I go the [usual way of looking through datasets and refining them through Cyber Security filter on R&D Areas/Groups](http://www.ll.mit.edu/r-d/datasets?author=All&rdarea=63&rdgroup=All&tag=All&items_per_page=10). 

But later I found them by searching "cyber" directly in the searchbox [here](http://www.ll.mit.edu/r-d/datasets?author=All&rdarea=All&rdgroup=All&keywords=cyber&tag=All&items_per_page=10 ). 

Maybe we can just change the URL to the direct search results of datasets. 
I believe it can be more apparent for other users.

lukashino: @jasonish What do you think, would changing the ll.mit.edu general search URL to the direct search result of sample data be a good way to go around this problem?

jasonish: For now I think so. In the longer term we may decide that static documentation like this is not the best place for "live" links that can change. But this is good for now.

lukashino:
Revisited this PR after quite a while. 
I found out that the MAWI sets link is not dead anymore...
I replaced the Darpa link with a search result as we discussed earlier

victorjulien (approved):

victorjulien:
Merged in #6324, thanks!

PR #3763, opened 2019-04-01
doc: Add manpages for suricatasc and suricatactl

shivan1b:
Add the missing manpages and the corresponding Sphinx configuration
for the command line tools `suricatasc` and `suricatactl`.

These manpages have been given the structure following the manpages for `git-submodules`.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/884

PR #4065, opened 2019-07-23
Next/20190722/v2

victorjulien:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3004

Describe changes:
- fixes for https://redmine.openinfosecfoundation.org/issues/3004
- #4063

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/281
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/284

PR #5518, opened 2020-10-27
Prelude: Fix #3763 and #4065 (Redmine tickets)

ToToL:
Describe changes:
- #4065 : Fix missing file config.h
      Since 900f1522b444e8391250683d48855ceb3d23f225 , HAVE_CONFIG_H is defined but config.h is not available.
- #3763 : Update from libjansson to JsonBuilder when getting information from output-json-http for example.
      JsonBuilder does not provides functions to access JSON data. Loading it with libjansson and access data.
- Update m4/libprelude.m4 to detect libprelude through pkg-config because libprelude-config is deprecated

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1192,7 +1192,7 @@
     CFLAGS="${CFLAGS} -Wno-error=unused-result"
 
     AS_IF([test "x$enable_prelude" = "xyes"], [
-        AM_PATH_LIBPRELUDE(0.9.9, , AC_MSG_ERROR(Cannot find libprelude: Is libprelude-config in the path?), no)
+        AM_LIBPRELUDE(4.0.0, , AC_MSG_ERROR(Cannot find libprelude: Is libprelude-config in the path?), no)
         if test "x${LIBPRELUDE_CFLAGS}" != "x"; then
             CPPFLAGS="${CPPFLAGS} ${LIBPRELUDE_CFLAGS}"
         fi
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
     AC_INIT([suricata],[7.0.0-dev])
     m4_ifndef([AM_SILENT_RULES], [m4_define([AM_SILENT_RULES],[])])AM_SILENT_RULES([yes])
-    AC_CONFIG_HEADERS([src/autoconf.h])
+    AC_CONFIG_HEADERS([src/config.h])
     AC_CONFIG_SRCDIR([src/suricata.c])
     AC_CONFIG_MACRO_DIR(m4)
     AM_INIT_AUTOMAKE([tar-ustar subdir-objects])
--- a/src/alert-prelude.c
+++ b/src/alert-prelude.c
@@ -676,15 +676,20 @@ static int JsonToAdditionalData(const char * key, json_t * value, idmef_alert_t 
 static void PacketToDataProtoHTTP(const Packet *p, const PacketAlert *pa, idmef_alert_t *alert)
 {
     json_t *js;
+    json_error_t error;
+    JsonBuilder *jsb;
 
-    js = JsonHttpAddMetadata(p->flow, pa->tx_id);
-    if (js == NULL)
+    jsb = jb_new_object();
+    EveHttpAddMetadata(p->flow, pa->tx_id, jsb);
+    jb_close(jsb);
+    js = json_loadb((char *)jb_ptr(jsb), jb_len(jsb), 0, &error);
+    jb_free(jsb);
+    if (unlikely(js == NULL))
         return;
 
     JsonToAdditionalData(NULL, js, alert);
 
     json_decref(js);
-
 }
 
 /**
@@ -696,12 +701,21 @@ static void PacketToDataProtoHTTP(const Packet *p, const PacketAlert *pa, idmef_
  */
 static void PacketToDataProtoHTTP2(const Packet *p, const PacketAlert *pa, idmef_alert_t *alert)
 {
-    void *http2_state = FlowGetAppState(f);
+    void *http2_state = FlowGetAppState(p->flow);
+    json_error_t error;
+    JsonBuilder *jsb;
+
     if (http2_state) {
         void *tx_ptr = rs_http2_state_get_tx(http2_state, pa->tx_id);
-        json_t *js = rs_http2_log_json(tx_ptr);
+        jsb = jb_new_object();
+        rs_http2_log_json(tx_ptr, jsb);
+        jb_close(jsb);
+
+        json_t *js = json_loadb((char *)jb_ptr(jsb), jb_len(jsb), 0, &error);
+        jb_free(jsb);
         if (unlikely(js == NULL))
             return;
+
         JsonToAdditionalData(NULL, js, alert);
         json_decref(js);
     }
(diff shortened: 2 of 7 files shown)


jasonish (changes requested):
Fails CI, and to build locally eve with prelude disabled.

ToToL:
Fails CI seems to not be related to my PR, right ?

If it is, can you help me on this ?

Thanks

Regards

configure.ac
```diff
@@ -1,6 +1,11 @@
     AC_INIT([suricata],[6.0.1-dev])
     m4_ifndef([AM_SILENT_RULES], [m4_define([AM_SILENT_RULES],[])])AM_SILENT_RULES([yes])
     AC_CONFIG_HEADERS([src/autoconf.h])
+
+    # FIXME: Undef HAVE_CONFIG_H since config.h is now not used
+    # Commit: 900f1522b444e8391250683d48855ceb3d23f225
+    AC_CONFIG_COMMANDS_PRE([test "X$DEFS" = X-DHAVE_CONFIG_H && DEFS=])
```
jasonish, line 7: Its related to this.  We still use HAVE_CONFIG_H, the file it loads is not autoconf.h instead of config.h though. Setting this appears to undef HAVE_CONFIG_H, sudo autoconf.h is not longer loaded, and its required.

ToToL:
Hello
Thanks for pointing me out that.
There are still fails CI, is it normal ?
Thanks for your help
Regards

jasonish:
> Hello
> Thanks for pointing me out that.
> There are still fails CI, is it normal ?
> Thanks for your help
> Regards

This time it looks like an issue in the CI, I've restarted the builds.

src/alert-prelude.c
```diff
@@ -696,10 +700,16 @@ static void PacketToDataProtoHTTP(const Packet *p, const PacketAlert *pa, idmef_
...
+    json_error_t error;
+    JsonBuilder *jsb;
+
     if (http2_state) {
         void *tx_ptr = rs_http2_state_get_tx(http2_state, pa->tx_id);
-        json_t *js = rs_http2_log_json(tx_ptr);
+        jsb = jb_new_object();
+        rs_http2_log_json(tx_ptr, jsb);
+        jb_close(jsb);
+        json_t *js = json_loadb((char *)jb_ptr(jsb), jb_len(jsb), 0, &error);
```
victorjulien, line 712: I don't like this approach. If the prelude output is essentially going to parse the eve-json string, why not move it into an external process? It seems that eve contains enough info for that, and if anything is missing we can consider adding it.

victorjulien (changes requested):

src/alert-prelude.c
ToToL: Hello,

Not essentially, just for extra data. If there is some already processed data, why don't add them as additional data?

If you really don't want this, I see two options :
* In the http code, split functions so Prelude can call them and get data without parsing JSON
* Remove these additionals data for the Prelude connector

Regards

configure.ac
```diff
@@ -1,6 +1,11 @@
     AC_INIT([suricata],[6.0.1-dev])
     m4_ifndef([AM_SILENT_RULES], [m4_define([AM_SILENT_RULES],[])])AM_SILENT_RULES([yes])
     AC_CONFIG_HEADERS([src/autoconf.h])
+
+    # FIXME: Rename HAVE_CONFIG_H to HAVE_AUTOCONF_H
```
victorjulien, line 5: fixme

src/alert-prelude.c
```diff
@@ -676,15 +676,19 @@ static int JsonToAdditionalData(const char * key, json_t * value, idmef_alert_t
 static void PacketToDataProtoHTTP(const Packet *p, const PacketAlert *pa, idmef_alert_t *alert)
 {
     json_t *js;
+    json_error_t error;
+    JsonBuilder *jsb;
 
-    js = JsonHttpAddMetadata(p->flow, pa->tx_id);
+    jsb = jb_new_object();
```
victorjulien, line 682: `jb_new_object` allocates memory, which is not freed in this function (or in any of the others). End with a call to `jb_free`.

Also need error checking.

src/alert-prelude.c
victorjulien: I think for 6 (`master-6.0.x`) we can accept this approach, but for the suricata 7 side (`master`) I would like to get rid of it. Then there are several options: 1) see if it fits our new output plugin support 2) make it a completely separate tool that post processes EVE.

It seems that there aren't many users anyway. The only report we've had was on the debian package not building. But no actual users complaining yet.

victorjulien (changes requested):

configure.ac
ToToL: done

src/alert-prelude.c
ToToL: done

ToToL: >I think for 6 (master-6.0.x) we can accept this approach, but for the suricata 7 side (master) I would like to get rid of it. Then there are several options: 1) see if it fits our new output plugin support 2) make it a completely separate tool that post processes EVE.
Next version of IDMEF will be in JSON I think so all of this will be easy :)

>It seems that there aren't many users anyway. The only report we've had was on the debian package not building. But no actual users complaining yet.
No report because no bug :) ? I can't find the bug you are talking about. Most of users that use the prelude support are on Centos/Redhat/Fedora.

ToToL:
CI is failing but it doesn't seems to be linked with my PR

Let me know how to go forward on this PR

catenacyber:
@ToToL I just restarted the CI jobs

ToToL:
I don't know if 4c176b7 is good because it is seems to be like reverting 900f1522b444e8391250683d48855ceb3d23f225 . If it is not the right way, let me know how to fix this.
The issue is suricata is using AC_CONFIG_HEADERS that defined HAVE_CONFIG_H but without providing config.h

Thanks for your review

catenacyber:
It looks like you fixed CI :-)
I leave Victor and Jason to comment on the prelude stuff that I do not know well enough.

ToToL:
@victorjulien : is it ok now for you ?

Thanks for your review

Regards

jasonish:
> The issue is suricata is using AC_CONFIG_HEADERS that defined HAVE_CONFIG_H but without providing config.h

This change from config.h to autoconf.h was on purpose.  I'm not sure what the issue is with HAVE_CONFIG_H, but using autoconf.h as the filename?  Outside of inconsistent naming, there should be no technical issue right?

Oh, I see the issue.  The libprelude headers has a `#ifdef HAVE_CONFIG_H` and then includes `config.h`.  This is interesting as its pulling the Suricata `config.h` which I'm pretty sure its not intending to do.  But isn't this a libprelude problam and not a Suricata problem then?

ToToL:
Thanks for your comment

In libprelude, when you have config.h file, you have HAVE_CONFIG_H defined

In suricata, when you have autoconf.h file, you have HAVE_CONFIG_H defined    <= quite strange no ? it should be HAVE_AUTOCONF_H but because suricata use AC_CONFIG_HEADERS, HAVE_CONFIG_H is defined. My first patch was to correct this and keep autoconf.h file but @victor ask for fixing

libprelude way seems more logical to me

victor:
No, I didn't ask anything

configure.ac
ToToL: > fixme

@victor : here

ToToL:
victorjulien 24 days ago
fixme

=> Here

configure.ac
victor: nope. You probably meant @victorjulien ?

jasonish:
> In libprelude, when you have config.h file, you have HAVE_CONFIG_H defined

Yeah, but I'm pretty sure the intention was not to pull in the Suricata `config.h`, but instead of the libprelude config.h.  I believe this is an error on libpreludes part on leaking this into a public header file, and we shouldn't have to name our files to satisfy libprelude. It seems to common pattern that a library would do is to rename config file to something like `prelude_config.h`.

Anyways, an alternative fix would be to do the following in `src/alert-prelude.c`:
```
+#undef HAVE_CONFIG_H
 #include <libprelude/prelude.h>
 ```
 as even if we have HAVE_CONFIG_H and an associated config.h, I'm pretty sure thats not the config.h it actually wants or should use.

File: src/alert-prelude.c [hunk 5]
```diff
@@ -744,14 +763,23 @@ static void PacketToDataProtoTLS(const Packet *p, const PacketAlert *pa, idmef_a
...
 
     if (ssh_state == NULL)
         return;
 
+    jsb = jb_new_object();
+
     void *tx_ptr = rs_ssh_state_get_tx(ssh_state, 0);
     BUG_ON(tx_ptr == NULL);
-    js = rs_ssh_log_json(tx_ptr);
+    rs_ssh_log_json(tx_ptr, jsb);
```
jasonish, line 778: As @victorjulien mentioned about error checking, rs_ssh_log_json is one method that could fail. So should be checked.

src/alert-prelude.c
jasonish: Agreed on this approach for 6.. But I think we should remove it from master sooner than later.  Might get a plugin going, or better yet, seems like some log watching agent is the best for this kind of stuff, at least in my opinion.

PR #5525, opened 2020-10-30
dns: parse additional RR types

satta:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: 
https://redmine.openinfosecfoundation.org/issues/3663

Describe changes:

The DNS parser code was extended to also parse the payloads for `NULL`, `NS` and `SRV` records. These could be useful to detect and document DNS tunneling activity, among other use cases.

New EVE-JSON examples:

```json
  "dns": {
    "version": 2,
    "type": "answer",
    "id": 1064,
    "flags": "8400",
    "qr": true,
    "aa": true,
    "rrname": "sbhi1c.pirate.sea",
    "rrtype": "NULL",
    "rcode": "NOERROR",
    "answers": [
      {
        "rrname": "sbhi1c.pirate.sea",
        "rrtype": "NULL",
        "ttl": 0,
        "rdata": "Base128"
      }
    ],
    "grouped": {
      "NULL": [
        "Base128"
      ]
    }
  }
```

```json
  "dns": {
    "version": 2,
    "type": "answer",
    "id": 6158,
    "flags": "81a0",
    "qr": true,
    "rd": true,
    "ra": true,
    "rrname": "isc.org",
    "rrtype": "NS",
    "rcode": "NOERROR",
    "answers": [
      {
        "rrname": "isc.org",
        "rrtype": "NS",
        "ttl": 7127,
        "rdata": "ns1.isc.org"
      },
      {
        "rrname": "isc.org",
        "rrtype": "NS",
        "ttl": 7127,
        "rdata": "ns3.isc.org"
      },
      {
        "rrname": "isc.org",
        "rrtype": "NS",
        "ttl": 7127,
        "rdata": "ns.isc.afilias-nst.info"
      },
      {
        "rrname": "isc.org",
        "rrtype": "NS",
        "ttl": 7127,
        "rdata": "ns2.isc.org"
      }
    ],
    "grouped": {
      "NS": [
        "ns1.isc.org",
        "ns3.isc.org",
        "ns.isc.afilias-nst.info",
[...]

PR #5529, opened 2020-11-03
dns: parse SRV records - v2

satta:
Previous PR: https://github.com/OISF/suricata/pull/5525

- Remove `NS` and `NULL` handling.
- Restructure commits.
- Use `NS` parsing function using `rest()`.

### DIFF
--- a/doc/userguide/output/eve/eve-json-format.rst
+++ b/doc/userguide/output/eve/eve-json-format.rst
@@ -422,6 +422,14 @@ More complex DNS record types may log additional fields for resource data:
   * "algo": Algorithm number (ex: 1 for RSA, 2 for DSS)
   * "type": Fingerprint type (ex: 1 for SHA-1)
 
+* "srv": section containing fields for the SRV (location of services) record type
+
+  * "target": Domain name of the target host (ex: ``foo.bar.baz``)
+  * "priority": Target priority (ex: 20)
+  * "weight": Weight for target selection (ex: 1)
+  * "port": Port on this target host of this service (ex: 5060)
+  * "raw": Raw resource data bytes (e.g. for DNS tunneling detection)
+
 One can control which RR types are logged by using the "types" field in the
 suricata.yaml file. If this field is not specified, all RR types are logged.
 More than 50 values can be specified with this field as shown below:
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -259,6 +259,20 @@ pub struct DNSRDataSSHFP {
     pub fingerprint: Vec<u8>,
 }
 
+#[derive(Debug,PartialEq)]
+pub struct DNSRDataSRV {
+    /// Priority
+    pub priority: u16,
+    /// Weight
+    pub weight: u16,
+    /// Port
+    pub port: u16,
+    /// Target
+    pub target: Vec<u8>,
+    /// Raw value (for tunneling detection etc.)
+    pub raw: Vec<u8>,
+}
+
 /// Represents RData of various formats
 #[derive(Debug,PartialEq)]
 pub enum DNSRData {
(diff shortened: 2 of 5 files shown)


jasonish:
Looks good. Do you have a PCAP we can use for an S-V test. None of the pcaps in S-V at this time seem to output any SRV records.

Thanks.

satta:
> Looks good. Do you have a PCAP we can use for an S-V test. None of the pcaps in S-V at this time seem to output any SRV records.

Done, see https://github.com/OISF/suricata-verify/pull/356

jasonish:
Thanks. All looks good. Just unsure about the "raw" in the SRV records, its not something we've typically logged. Do you find this useful to have in your event feed?

satta:
> Just unsure about the "raw" in the SRV records, its not something we've typically logged. Do you find this useful to have in your event feed?

The idea is to have an uninterpreted byte log to possibly reconstruct the traffic for both directions of a tunnel from the EVE. That would require serializing the parsed structure back to bytes without the `raw` field, that's why I added it. If we can find a more idiomatic solution for this, I'd be happy to implement it ;)

jasonish:
> The idea is to have an uninterpreted byte log to possibly reconstruct the traffic for both directions of a tunnel from the EVE. That would require serializing the parsed structure back to bytes without the `raw` field, that's why I added it. If we can find a more idiomatic solution for this, I'd be happy to implement it ;)

How much of this data would exist in a true tunnel? I'd worry about performance when a tunnel is opened due to the amount of data that may need to be written to the log file.  Would base64 be better rather than our somewhat hacky method of encoding non-printable characters?

satta:
> > The idea is to have an uninterpreted byte log to possibly reconstruct the traffic for both directions of a tunnel from the EVE. That would require serializing the parsed structure back to bytes without the `raw` field, that's why I added it. If we can find a more idiomatic solution for this, I'd be happy to implement it ;)
> 
> How much of this data would exist in a true tunnel? I'd worry about performance when a tunnel is opened due to the amount of data that may need to be written to the log file. 

No idea but I would guess it's pretty limited. Maybe it also would be best to make this field optional, to reduce content duplication if raw DNS rdata dumping is not required?

> Would base64 be better rather than our somewhat hacky method of encoding non-printable characters?

Maybe. It would probably best to stick with the format also used for the `payload` fields.

jasonish:
> > > The idea is to have an uninterpreted byte log to possibly reconstruct the traffic for both directions of a tunnel from the EVE. That would require serializing the parsed structure back to bytes without the `raw` field, that's why I added it. If we can find a more idiomatic solution for this, I'd be happy to implement it ;)
> > 
> > 
> > How much of this data would exist in a true tunnel? I'd worry about performance when a tunnel is opened due to the amount of data that may need to be written to the log file.
> 
> No idea but I would guess it's pretty limited. Maybe it also would be best to make this field optional, to reduce content duplication if raw DNS rdata dumping is not required?
> 
> > Would base64 be better rather than our somewhat hacky method of encoding non-printable characters?
> 
> Maybe. It would probably best to stick with the format also used for the `payload` fields.

Ok. How about we kill this raw field as part of this PR.. I can do that if you don't mind, I have some changes already applied on top of this around clippy cleanups and a rustfmt of the DNS module I'm hoping to get in soon.

Then make it optional. I would be nice to see a pcap of a real tunnel and just how much data would get logged in that case.

satta:
[...]
> Ok. How about we kill this raw field as part of this PR.. I can do that if you don't mind, I have some changes already applied on top of this around clippy cleanups and a rustfmt of the DNS module I'm hoping to get in soon.

OK, let's do that then.

> Then make it optional. I would be nice to see a pcap of a real tunnel and just how much data would get logged in that case.

The OP of the original Redmine ticket (https://redmine.openinfosecfoundation.org/issues/3663) posted one. Obviously, in the end the total amount of logged data depends on the full amount of data transferred over the tunnel, but the amount of data logged within a _single_ DNS answer is sure limited.

victorjulien:
Merged through #5641, thanks!

PR #5538, opened 2020-11-05
Fuzz ideal v2

catenacyber:
Should I create one ?

Describe changes:
- adds fuzzing status badge in README
- adds CIfuzz GitHub workflow, to run fuzzers for 10 minutes as part of CI
- uses yaml config for applayer fuzz target (so that every protocol such as DNP3 is enabled) 

Replaces #5535 with adding more commits with different improvements


catenacyber:
Each commit is independent, but they all contribute to an oss-fuzz ideal integration

catenacyber:
Another solution to enable DNP3 in appealer fuzz target would be to change the condition `RunmodeIsUnittests()` in `RegisterDNP3Parsers` (and other parsers)

PR #5537, opened 2020-11-05
decode: add icmpv4.hdr keyword

jlucovsky:
Continuation of #5536 

This PR adds a keyword for the ICMPv4 header -- `icmpv4.hdr`. 

This sticky buffer returns ICMPv4 header bytes; the size of the header is determined by the ICMP message type. Some messages, like router advertisements, are variable lengths.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4090](https://redmine.openinfosecfoundation.org/issues/4090)

Describe changes:
- NULL check of `icmpv4h` in `GetData`

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

suricata-verify-pr: 357
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5540, opened 2020-11-06
decode: add icmpv4.hdr keyword 

jlucovsky:
Continuation of #5537

This PR adds a keyword for the ICMPv4 header -- `icmpv4.hdr`. 

This sticky buffer returns ICMPv4 header bytes; the size of the header is determined by the ICMP message type. Some messages, like router advertisements, are variable lengths.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4090](https://redmine.openinfosecfoundation.org/issues/4090)

Describe changes:
- Early exit for IPv6 packets in `GetData`

suricata-verify-pr: 357
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-icmp-hdr.c
+++ b/src/detect-icmp-hdr.c
@@ -1,0 +1,124 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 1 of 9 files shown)


src/detect-icmp-hdr.c
```diff
@@ -0,0 +1,124 @@
...
+
+    return 0;
+}
+
+static InspectionBuffer *GetData(DetectEngineThreadCtx *det_ctx,
+        const DetectEngineTransforms *transforms, Packet *p, const int list_id)
+{
+    SCEnter();
+
+    if (p->ip6h != NULL) {
```
catenacyber, line 98: Why do you test for `ip6h` instead of `icmpv4h ` ?

jlucovsky:
Continued in #5546

src/detect-icmp-hdr.c
jlucovsky: This will be called for ICMP packets from v6 and v4. This check filters those from ICMP v6

catenacyber: IMHO, we should test `icmpv4h`, not `ip6h`, because it is what gets used afterwards

If there is ICMPv6, `icmpv6h` should be set instead of `icmpv4h`
Furthermore, `DETECT_PROTO_IPV4` should prevent it, right ?

PR #5539, opened 2020-11-05
Fuzz ideal v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
Should I create one ?

Describe changes:
- adds fuzzing status badge in README
- adds CIfuzz GitHub workflow, to run fuzzers for 10 minutes as part of CI
- uses yaml config for applayer fuzz target (so that every protocol such as DNP3 is enabled) 

Replaces #5538 with clang-format
CIFuzz success depends on https://github.com/google/oss-fuzz/pull/4606

PR #5541, opened 2020-11-06
Fuzz ideal v4

catenacyber:
Should I create one ?

Describe changes:
- adds fuzzing status badge in README
- adds CIfuzz GitHub workflow, to run fuzzers for 10 minutes as part of CI
- uses yaml config for applayer fuzz target (so that every protocol such as DNP3 is enabled) 
- improves sigpcap target by having more than one valid packet with its datalink

Replaces #5539 with improving fuzz_sigpcap target


victorjulien:
Minor typo "fuzz: better packet recycling in sigpcap tqrget"  should be "target"

PR #4109, opened 2019-08-20
Ebpf xdp update 5.0 v6

regit:
Update of #4037 addressing comments

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Describe changes:
- Address comments in doc and code

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output:
- PR regit: https://buildbot.openinfosecfoundation.org/builders/regit/builds/487
- PR regit-pcap: https://buildbot.openinfosecfoundation.org/builders/regit-pcap/builds/265

PR #5551, opened 2020-11-08
eve: do not access flow storage in packet context

satta:

https://redmine.openinfosecfoundation.org/issues/4109

Describe changes:
- Make sure not to access the flow storage (e.g. keeping a `MacSet`) before making sure we have a flow to begin with. We can,
for example, run into an alert without a flow with `ip` rules, in which case the flow might be NULL. See Redmine issue #4109.

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -439,7 +439,7 @@ static void EveAddMetadata(const Packet *p, const Flow *f, JsonBuilder *js)
     }
 }
 
-int CreateJSONEther(JsonBuilder *parent, const Packet *p, const MacSet *ms);
+int CreateJSONEther(JsonBuilder *parent, const Packet *p, const Flow *f);
 
 void EveAddCommonOptions(const OutputJsonCommonSettings *cfg,
         const Packet *p, const Flow *f, JsonBuilder *js)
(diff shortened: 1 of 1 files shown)


src/output-json.c
```diff
@@ -798,15 +796,23 @@ static int MacSetIterateToJSON(uint8_t *val, MacSetSide side, void *data)
     return 0;
 }
 
-int CreateJSONEther(JsonBuilder *js, const Packet *p, const MacSet *ms)
+int CreateJSONEther(JsonBuilder *js, const Packet *p, const Flow *f)
 {
-    jb_open_object(js, "ether");
     if (unlikely(js == NULL))
```
victorjulien, line 801: why not check for `f` here as well? avoids having to do the cleanup from line 808

satta: Because it can happen that `f` is NULL -- this was the reason for the bug altogether. If `f` is NULL, then we want to use only packet metadata (lines 806-833). Line 808 is only there to make sure that we don't crash again in case `CreateJSONEther()` is called without neither `p` nor `f` being valid (I have no idea if that is possible). It should never happen.

If we check `f` as early you suggest, we will not be able to handle the packet case. Or am I not understanding correctly what your suggestion is?

victorjulien: I missed the packet case. Merged as is, thanks!

PR #5569, opened 2020-11-16
optimize rust logging macros - v1

jasonish:
Expand Rust logging macros after the log level is checked.
Currently too much work is done before expansion which
can have an affect at runtime as SCLogDebug calls
occur during the packet path.

Should not have an affect if compiled without a debug.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/4114


victorjulien:
Merged in #5575, thanks!

PR #5556, opened 2020-11-11
quic: Add GQUIC App Layer - v1

sharksforarms:
- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3440

Describe changes:
- Parses gquic version Q043-Q046 and logs a CYU hash
- Added a mix of unit and suricata-verify pcap tests
- Fails gracefully with unknown QUIC versions
- Other versions can be added later w/ pcaps + testing
- Logs a cyu hash and string to eve.json
  - Example ```{"timestamp":"2020-09-02T15:58:21.139196+0000","flow_id":1610311033233340,"pcap_cnt":1,"event_type":"quic","src_ip":"10.0.2.15","src_port":51989,"dest_ip":"172.217.164.142","dest_port":443,"proto":"UDP","quic":{"cyu":{"hash":"910a5e3a4d51593bd59a44611544f209","string":"46,PAD-SNI-VER-CCS-UAID-TCID-PDMD-SMHL-ICSL-NONP-MIDS-SCLS-CSCT-COPT-IRTT-CFCW-SFCW"}}}```

References:
   - https://docs.google.com/document/d/1WJvyZflAO2pq77yOLbp9NsGjC1CHetAXV8I0fQe-B_U/edit
   - https://docs.google.com/document/d/1g5nIXAIkN_Y-7XJW5K45IblHd_L2f5LTaDUDwvZ5L6g/edit
   - https://www.slideshare.net/shigeki_ohtsu/quic-overview
   - https://tools.ietf.org/html/draft-ietf-quic-transport-17#section-19.8
   - https://github.com/salesforce/GQUIC_Protocol_Analyzer/blob/master/src/gquic-protocol.pac
   - cyu: https://engineering.salesforce.com/gquic-protocol-analysis-and-fingerprinting-in-zeek-a4178855d75f

suricata-verify:
https://github.com/OISF/suricata-verify/pull/364

Questions:
- Should it be enabled by default in suricata.yml
[...]

PR #5574, opened 2020-11-17
quic: Add QUIC App Layer - v2

sharksforarms:
Previous PR: https://github.com/OISF/suricata/pull/5556

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3440

Describe changes:
- Parses gquic version Q043-Q046 and logs a CYU hash
- Added a mix of unit and suricata-verify pcap tests
- Fails gracefully with unknown QUIC versions
- Other versions can be added later w/ pcaps + testing
- Logs a cyu hash and string to eve.json
  - Example ```{"timestamp":"2020-09-02T15:58:21.139196+0000","flow_id":1610311033233340,"pcap_cnt":1,"event_type":"quic","src_ip":"10.0.2.15","src_port":51989,"dest_ip":"172.217.164.142","dest_port":443,"proto":"UDP","quic":{"cyu":{"hash":"910a5e3a4d51593bd59a44611544f209","string":"46,PAD-SNI-VER-CCS-UAID-TCID-PDMD-SMHL-ICSL-NONP-MIDS-SCLS-CSCT-COPT-IRTT-CFCW-SFCW"}}}```

References:
   - https://docs.google.com/document/d/1WJvyZflAO2pq77yOLbp9NsGjC1CHetAXV8I0fQe-B_U/edit
   - https://docs.google.com/document/d/1g5nIXAIkN_Y-7XJW5K45IblHd_L2f5LTaDUDwvZ5L6g/edit
   - https://www.slideshare.net/shigeki_ohtsu/quic-overview
   - https://tools.ietf.org/html/draft-ietf-quic-transport-17#section-19.8
   - https://github.com/salesforce/GQUIC_Protocol_Analyzer/blob/master/src/gquic-protocol.pac
   - cyu: https://engineering.salesforce.com/gquic-protocol-analysis-and-fingerprinting-in-zeek-a4178855d75f

suricata-verify:
https://github.com/OISF/suricata-verify/pull/368

Questions:
- Should it be enabled by default in suricata.yml
- Should port 80 be added as per https://github.com/salesforce/GQUIC_Protocol_Analyzer/blob/e62494ad437539e4b88d1b917e7cc16e4acaca91/scripts/Salesforce/GQUIC/main.zeek#L33

TODO:
- Documentation, will do this in a version+1 PR once code is ok


catenacyber:
If you now rebase on top of master, the CI for commit-check and formatting-check should run normally

catenacyber:
> Should port 80 be added ?

I think so.

> Should it be enabled by default in suricata.yml ?

Victor already replied in previous PR ;-)

rust/src/quic/cyu.rs
```diff
@@ -0,0 +1,165 @@
...
+            QuicVersion::Q045 => Some("44"),
+            QuicVersion::Q046 => Some("46"),
+            _ => None,
+        }?;
+
+        if frames.is_empty() {
+            return None;
+        }
+
+        for frame in frames {
```
catenacyber, line 30: I do not know the protocol here.
But this construction feels strange : iterating for frames and returning the first cyu_string
Could not there be a later frame with a different cyu ?
Would not it be a problem ?

rust/src/quic/error.rs
```diff
@@ -0,0 +1,44 @@
...
+ * 02110-1301, USA.
+ */
+
+#[derive(Debug, PartialEq)]
+pub enum QuicError {
+    InvalidPacket,
+    Parse(String),
+}
+
+macro_rules! nom_to_quicerror {
```
catenacyber, line 24: Could you comment to explain the purpose here ?

rust/src/quic/frames.rs
```diff
@@ -0,0 +1,249 @@
...
+    let tag_as_str = std::str::from_utf8(tag)
+        .map_err(|e| {
+            QuicError::Parse(format!(
+                "Could not parse stream tag as str: {}",
+                e.to_string()
+            ))
+        })?
+        .trim_matches(char::from(0)); // trim null padding
+
+    let tag = StreamTag::from_str(tag_as_str).map_err(|e| {
```
catenacyber, line 93: Maybe `http2_parse_frame_setting` syntax is easier with something like 
`map_opt!(be_u16, num::FromPrimitive::from_u16)` (u32 in your case)

rust/src/quic/logger.rs
```diff
@@ -0,0 +1,38 @@
...
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use super::quic::QuicTransaction;
+use crate::jsonbuilder::{JsonBuilder, JsonError};
+
+fn log_template(tx: &QuicTransaction, js: &mut JsonBuilder) -> Result<(), JsonError> {
+    js.set_uint("version", u32::from(tx.header.version.clone()).into())?;
```
catenacyber, line 22: Why do we need clone and into ?
Could we not use ` as u64`  ?

rust/src/quic/quic.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    let slice = build_slice!(input, input_len as usize);
+
+    match QuicHeader::from_bytes(slice, DEFAULT_DCID_LEN) {
+        Ok((_, _)) => {
+            return unsafe { ALPROTO_QUIC };
+        }
+        _ => {
+            return ALPROTO_UNKNOWN;
```
catenacyber, line 209: Can we never fail ? ie return `ALPROTO_FAILED`

Maybe we should return `ALPROTO_UNKNOWN ` only in case `Err(nom::Err::Incomplete(_))`

src/detect-quic-cyu-hash.c
```diff
@@ -0,0 +1,89 @@
...
+    DetectAppLayerInspectEngineRegister2(BUFFER_NAME, ALPROTO_QUIC, SIG_FLAG_TOSERVER, 0,
+            DetectEngineInspectBufferGeneric, GetData);
+
+    DetectAppLayerMpmRegister2(BUFFER_NAME, SIG_FLAG_TOSERVER, 2, PrefilterGenericMpmRegister,
+            GetData, ALPROTO_QUIC, 1);
+
+    DetectBufferTypeSetDescriptionByName(BUFFER_NAME, BUFFER_DESC);
+
+    g_buffer_id = DetectBufferTypeGetByName(BUFFER_NAME);
+
```
catenacyber, line 87: Can you validate the signature is valid ? ie the content contains only hexadecimal bytes ? cf `DetectSshHasshHashValidateCallback`

rust/src/quic/parser.rs
```diff
@@ -0,0 +1,330 @@
...
+            || *self == QuicVersion::Q044
+            || *self == QuicVersion::Q045
+            || *self == QuicVersion::Q046
+    }
+}
+
+impl From<QuicVersion> for u32 {
+    fn from(from: QuicVersion) -> Self {
+        match from {
+            QuicVersion::Q043 => 0x51303433,
```
catenacyber, line 58: I do not know if for detection, we should translate 3 -> 0x51303433...

catenacyber (commented):
Thanks for this contribution.
This looks good overall.
I left some comments for details, and this still needs a bigger review

rust/src/quic/cyu.rs
sharksforarms: I assumed there would only be a single STREAM frame per packet (is the case in the testing pcaps) but this may not be the case if multiple connections are multiplexed. I'll make this return a Vec instead.

rust/src/quic/parser.rs
sharksforarms: not sure I understand what you mean

rust/src/quic/quic.rs
sharksforarms: Good idea!

rust/src/quic/parser.rs
catenacyber: For the signature keyword `quic.version` do we match a packet with `QuicVersion::Q043` if the signature has `quic.version: 3` or `quic.version: 0x51303433` ?

sharksforarms: Ah, yes. Right now it's parsing the keyword in C using `DetectU32Parse`

rust/src/quic/logger.rs
sharksforarms: `clone` is to call the from implementation: `from(from: QuicVersion)`
`into()` calls the from implementation for u32 -> u64 in this case. I prefer to avoid using `as` to avoid accidental precision loss, both are equivalent here.

PR #5576, opened 2020-11-18
detect/bsize: Validate against content buffer when available 

jlucovsky:
Continuation of #5028

Describe changes:
- Address review comments
- Rebase

suricata-verify-pr: 238
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-bsize.c
+++ b/src/detect-bsize.c
@@ -38,7 +38,7 @@
 
 /*prototypes*/
 static int DetectBsizeSetup (DetectEngineCtx *, Signature *, const char *);
-static void DetectBsizeFree (DetectEngineCtx *, void *);
+static void DetectBsizeFree(DetectEngineCtx *, void *);
 #ifdef UNITTESTS
 static void DetectBsizeRegisterTests (void);
 #endif
@@ -64,6 +64,8 @@ void DetectBsizeRegister(void)
 #define DETECT_BSIZE_GT 1
 #define DETECT_BSIZE_RA 2
 #define DETECT_BSIZE_EQ 3
+
+const char *bsize_mode_strings[] = { "<", ">", "<>", "=" };
 
 typedef struct DetectBsizeData {
     uint8_t mode;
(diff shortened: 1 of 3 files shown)


src/detect-bsize.c
```diff
@@ -291,6 +359,13 @@ static int DetectBsizeSetup (DetectEngineCtx *de_ctx, Signature *s, const char *
     DetectBsizeData *bsz = DetectBsizeParse(sizestr);
     if (bsz == NULL)
         goto error;
+
+    if (s->init_data->smlists[list] != NULL) {
```
catenacyber, line 363: Should this should be done with `DetectBufferTypeRegisterValidateCallback` ?

src/detect-bsize.c
```diff
@@ -65,6 +65,8 @@ void DetectBsizeRegister(void)
 #define DETECT_BSIZE_RA 2
 #define DETECT_BSIZE_EQ 3
 
+const char *bsize_mode_strings[] = { "<", ">", "<>", "=" };
```
catenacyber, line 68: Is this variable used ?

PS : could we reuse stuff from detect-engine-uint.h here ?

jlucovsky:
Continued in #5585

src/detect-bsize.c
jlucovsky: It's not used -- will remove.

catenacyber: Ok.
So, could we reuse stuff from detect-engine-uint.h here ?

src/detect-bsize.c
jlucovsky: What would be the advantage of doing so?

catenacyber: Your fix seems better :-)

PR #5578, opened 2020-11-18
rules/mqtt: renumber mqtt events to avoid conflict with ssh - v1

jasonish:
Both SSH and MQTT events were in the 2228000 range. As SSH was
added first, renumber MQTT events into the 2229000 range which is
free.

### DIFF
--- a/rules/mqtt-events.rules
+++ b/rules/mqtt-events.rules
@@ -4,12 +4,12 @@
 #    http://doc.emergingthreats.net/bin/view/Main/SidAllocation and
 #    https://redmine.openinfosecfoundation.org/projects/suricata/wiki/AppLayer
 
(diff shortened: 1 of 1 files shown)


victorjulien:
Can we add a QA check to avoid this in the future? A `-S rules/*.rules -T` of some kind?

catenacyber:
Thanks, I had missed that

jasonish:
Duplicate SID checking done here: https://github.com/OISF/suricata/pull/5587

It uses sid, sort, uniq to parse out the sids and check for duplicates.  A little easier than updating Suricata configuration to enable all parsers, test, parse output, etc.

catenacyber:
Is there an easy way to enable all parsers in Suricata ?

victorjulien:
@catenacyber I don't think so, see also https://github.com/OISF/suricata/pull/5586#issuecomment-731271909

PR #5580, opened 2020-11-19
app-layer: do not try to parse gaps during protocol change

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3924

Describe changes:
- do not try to parse gaps during protocol change

As this will leak the flow alstate because `AppLayerParserParse` relies on `FlowChangeProto` to know if it should allocate a new alstate

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -618,6 +618,11 @@ int AppLayerHandleTCPData(ThreadVars *tv, TcpReassemblyThreadCtx *ra_ctx,
                 goto failure;
             }
         }
+        if (FlowChangeProto(f)) {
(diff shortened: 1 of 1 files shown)


victorjulien:
In the normal protocol detection we consider both sides, and if one side starts with a gap we can still detect the protocol. Is this impossible for protocol upgrades?

catenacyber:
I do not think this is impossible, but this needs some more refactoring of the protocol change code.
For instance, other cases are not dealt with and may lead to evasion if there is inconsistency between the protocol used by the client and the one used by the server.
Do you want it now or make a ticket for later ?

PR #5587, opened 2020-11-20
rules/mqtt: renumber mqtt events to avoid conflict with ssh - v2

jasonish:
Both SSH and MQTT events were in the 2228000 range. As SSH was
added first, renumber MQTT events into the 2229000 range which is
free.

Previous PR: https://github.com/OISF/suricata/pull/5578

Changes from last PR:
- Add a duplicate SID check to the build that prepares the
  distribution package for many of the other builds.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -116,6 +116,15 @@ jobs:
 
       - uses: actions/checkout@v2
 
+      # Prebuild check for duplicat SIDs
+      - name: Check for duplicate SIDs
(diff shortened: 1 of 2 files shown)


jasonish:
Example failure can be seen with the check commit on master:
https://github.com/jasonish/suricata/runs/1432221778?check_suite_focus=true

Screenshot from 2020-11-20 11-07-55

victorjulien:
Merged in #5600, thanks!

PR #5592, opened 2020-11-23
dnp3: set byte order when logging dnp3 src and dst - v1

jasonish:
DNP3 uses little endian on the wire, for the most part this
is handled as the messages are deserialize. However, the link
header is a cast over raw data, so swap these bytes as they
are being logged.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/4173


victorjulien:
Response side will also need this.

jasonish:
Oops.. Fixed here: https://github.com/OISF/suricata/pull/5594

PR #5594, opened 2020-11-23
dnp3: set byte order when logging dnp3 src and dst - v2

jasonish:
DNP3 uses little endian on the wire, for the most part this
is handled as the messages are deserialize. However, the link
header is a cast over raw data, so swap these bytes as they
are being logged.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/4173

Previous PR: #5592 

Changes from previous PR:
- Fix src and dst in response logging as well.

### DIFF
--- a/src/output-json-dnp3.c
+++ b/src/output-json-dnp3.c
@@ -152,8 +152,8 @@ void JsonDNP3LogRequest(JsonBuilder *js, DNP3Transaction *dnp3tx)
     JsonDNP3LogLinkControl(js, dnp3tx->request_lh.control);
     jb_close(js);
 
-    jb_set_uint(js, "src", dnp3tx->request_lh.src);
-    jb_set_uint(js, "dst", dnp3tx->request_lh.dst);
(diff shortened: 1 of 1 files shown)


src/output-json-dnp3.c
```diff
@@ -152,8 +152,8 @@ void JsonDNP3LogRequest(JsonBuilder *js, DNP3Transaction *dnp3tx)
     JsonDNP3LogLinkControl(js, dnp3tx->request_lh.control);
     jb_close(js);
 
-    jb_set_uint(js, "src", dnp3tx->request_lh.src);
-    jb_set_uint(js, "dst", dnp3tx->request_lh.dst);
+    jb_set_uint(js, "src", DNP3_SWAP16(dnp3tx->request_lh.src));
```
catenacyber, line 155: Should it be done during parsing ?

jasonish: Normally I'd say yes, but currently we just cast the struct over the bytes and we don't use the src/dst in a way where the byte order matters.

PR #5616, opened 2020-11-30
Fuzz improve fixes v3

catenacyber:

Describe changes:
- improves detect proto fuzz target (to be matched as an assert by oss-fuzz)
- fix one warning when compiling with unit tests and `DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`
- limit memcap for datasets when fuzzing to avoid OOMs
- check PacketCopyData return value in fuzz sigpcap target
- rust : do not compile with coverage and build-std

Modifies #5615 with cocci check fix


victorjulien:
Merged in #5618, thanks!

PR #5617, opened 2020-11-30
windows rust fixups - v1

jasonish:
First, move libraries from LDFLAGS to LDADD on Windows. This puts them in the order expected. Not quite sure why it worked before, but I think it had something to do with libtool getting the order wrong when presented with a library name of "suricata.lib" instead of "libsuricata.a", and by fluke it worked itself out.

Secondly, as of Rust 1.44, static libraries on windows-gnu build systems with be named "libsuricata.a", and not "suricata.lib" meaning we no longer have to special case the library name on Windows. But for compatibility reasons we will still rename suricata.lib to libsuricata.a if found.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -306,7 +306,6 @@
             WINDOWS_PATH="yes"
             PCAP_LIB_NAME="wpcap"
             AC_DEFINE([HAVE_NON_POSIX_MKDIR], [1], [mkdir is not POSIX compliant: single arg])
-            RUST_SURICATA_LIBNAME="suricata.lib"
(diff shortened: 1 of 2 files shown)


victorjulien:
It now compiles on mingw again, thanks.

This output does look a bit suspicious, with the path just showing /<file>
Screenshot from 2020-12-01 05-56-41

victorjulien:
Can fix it with
Screenshot from 2020-12-01 06-14-33
Screenshot from 2020-12-01 06-14-07

victorjulien:
@jasonish if that looks reasonable to you I'll just amend it into your commit

jasonish:
> @jasonish if that looks reasonable to you I'll just amend it into your commit

Yup, thats reasonable. Thanks. Missed that.

victorjulien:
Merged in #5618, thanks!

PR #5634, opened 2020-12-07
Next/20201207/v4

victorjulien:
Describe changes:
- app-layer API cleanup: don't use callback for "completion value"
- remove buffer inspect api v1. Move remaining users to v2.
- include #5633

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
Is down.

PR #5633, opened 2020-12-07
github-ci: cbindgen optimization and updates - v1

jasonish:
- Remove Fedora 31 (its now EOL)
- Add Fedora 33
- In prep, remove libhtp, suricata-update and suricata-verify directories before artifact upload, reduces time of upload from minutes to seconds
- Create a new prep build that builds cbindgen statically for Linux x86_64.
- In Linux x86_64 builds, use the prepared cbindgen

PR #5635, opened 2020-12-07
Next/20201207/v5

victorjulien:
Describe changes:

replace #5634 to fix commit checks
app-layer API cleanup: don't use callback for "completion value"
remove buffer inspect api v1. Move remaining users to v2.
include #5633

PRScript output (if applicable):
Is down.


victorjulien:
Replaced by #5636

PR #5637, opened 2020-12-07
github-ci: add sccache to per-commit check - v1

jasonish:
Also use the pre-build cbindgen binary.
Hopefully speeds up the build process.

I've dumped the sccache stats at the end to show that its being used.


victorjulien:
Merged through #5644, thanks!

PR #5650, opened 2020-12-10
fuzz: adds MSAN to CIFuzz

catenacyber:

Describe changes:
- Adds memory sanitizer to CIFuzz

cf https://github.com/google/oss-fuzz/issues/4614
cc @jonathanmetzman

Let's see if CI works

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -6,7 +6,7 @@ jobs:
    strategy:
      fail-fast: false
      matrix:
-       sanitizer: [address, undefined]
+       sanitizer: [address, undefined, memory]
    steps:
    - name: Build Fuzzers (${{ matrix.sanitizer }})
      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master


catenacyber:
So, not working

victorjulien:
How is it expected to work? It somehow builds in a place that has special compiled versions of our dependencies like pcre?

catenacyber:
> It somehow builds in a place that has special compiled versions of our dependencies like pcre?

Yes, it should have prebuilt MSAN-compiled versions of some libraries like pcre
That is how oss-fuzz does it. CIFuzz is getting fixed to do the same apparently

jonathanmetzman:
> > It somehow builds in a place that has special compiled versions of our dependencies like pcre?
> 
> Yes, it should have prebuilt MSAN-compiled versions of some libraries like pcre
> That is how oss-fuzz does it. CIFuzz is getting fixed to do the same apparently

+1 Maybe we can close this PR. I can reopen a similar one and use it to debug a branch I'm working on when I next get a chance to fix this (probably after the holidays).

PR #5660, opened 2020-12-12
Http2 keywords 4067 v4

catenacyber:

Describe changes:
- Adds signature protocol extending flow protocol (one signature protocol matches multiple flow protocols cf `AppProtoEquals`)
- Adds `ALPROTO_HTTP_ANY` (as such a signature protocol) for signatures to match on both HTTP1 and HTTP2
- `http.uri` keyword now matches on HTTP2 traffic

suricata-verify-pr: 377

TL;DR
Is this the good way to go ?

The big change is in this `ALPROTO_HTTP_ANY` (first commit)
It comes from writing the rules, cf S-V test in https://github.com/OISF/suricata-verify/pull/377
I think it is good to allow these rules
```
# match on any HTTP version
alert http any any -> any any (http.uri; content:"/doc/manual/html/index.html"; sid:10;)
# match on HTTP 1 only
alert http1 any any -> any any (http.uri; content:"/doc/manual/html/index.html"; sid:11;)
# match on HTTP 2 only
alert http2 any any -> any any (http.uri; content:"/doc/manual/html/index.html"; sid:12;)
# match on whatever
alert ip any any -> any any (http.uri; content:"/doc/manual/html/index.html"; sid:13;)
```

As a consequence, the `Signature` structure shall contain this detailed information  about the app-layer protocol
That meant either
- adding a new (signature) protocol to `AppProtoEnum`
- adding a new field to `struct Signature_` to complete `AppProto alproto`
- merging HTTP2 and HTTP1 all in `ALPROTO_HTTP`

The last solution does not feel good to me, as it makes sense to have 2 different `ALPROTO_` related to the flow as they are 2 completely different protocols/encoding even if they offer the (almost) same functionalities
Adding a field implied consuming more memory...
So I went with the first solution.

I renamed the `struct Signature_` field to `alolproto` to see where it gets used, as this is a Signature protocol, which may correspond to multiple Flow protocol. That is : a Flow has only one protocol. A Signature may have multiple protocols
I created a function `AppProtoEquals` that says if a Signature protocol matches a Flow protocol
I added `ALPROTO_HTTP_ANY`
Then, this was just inspecting every use of `alolproto` and translating it to its new meaning

What remains to be done :
- get validation about design first
- rename `alproto` to `sig_alproto`. Thoughts about a better name ?
- CI (clang-format)
- document how to translate more keywords to using `ALPROTO_HTTP_ANY` (simple enough) and do it
- squash last commit

Modifies #5659 by removing compile warning

### DIFF
--- a/src/app-layer-protos.c
+++ b/src/app-layer-protos.c
@@ -120,6 +120,9 @@ const char *AppProtoToString(AppProto alproto)
         case ALPROTO_HTTP2:
             proto_name = "http2";
             break;
+        case ALPROTO_HTTP_ANY:
+            proto_name = "http_any";
+            break;
         case ALPROTO_FAILED:
             proto_name = "failed";
             break;
@@ -138,7 +141,8 @@ AppProto StringToAppProto(const char *proto_name)
 {
     if (proto_name == NULL) return ALPROTO_UNKNOWN;
 
-    if (strcmp(proto_name,"http")==0) return ALPROTO_HTTP;
+    if (strcmp(proto_name,"http")==0) return ALPROTO_HTTP_ANY;
+    if (strcmp(proto_name,"http1")==0) return ALPROTO_HTTP;
     if (strcmp(proto_name,"ftp")==0) return ALPROTO_FTP;
     if (strcmp(proto_name,"smtp")==0) return ALPROTO_SMTP;
     if (strcmp(proto_name,"tls")==0) return ALPROTO_TLS;
--- a/rust/src/http2/detect.rs
+++ b/rust/src/http2/detect.rs
@@ -478,6 +478,29 @@ pub unsafe extern "C" fn rs_http2_tx_get_header_name(
     return 0;
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn rs_http2_tx_get_uri(
+    tx: &mut HTTP2Transaction, buffer: *mut *const u8, buffer_len: *mut u32,
+) -> u8 {
+    for i in 0..tx.frames_ts.len() {
+        match &tx.frames_ts[i].data {
+            HTTP2FrameTypeData::HEADERS(hd) => {
+                for j in 0..hd.blocks.len() {
+                    if hd.blocks[j].name == ":path".as_bytes().to_vec() {
+                        let value = &hd.blocks[j].value;
+                        *buffer = value.as_ptr(); //unsafe
+                        *buffer_len = value.len() as u32;
+                        return 1;
+                    }
+                }
+            }
+            _ => {}
+        }
+    }
+
+    return 0;
+}
+
 fn http2_escape_header(hd: &parser::HTTP2FrameHeaders, i: u32) -> Vec<u8> {
     //minimum size + 2 for escapes
     let normalsize = hd.blocks[i as usize].value.len() + 2 + hd.blocks[i as usize].name.len() + 2;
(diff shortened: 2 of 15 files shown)


victorjulien:
I think the approach makes sense. It's unfortunate that we need to do the more expensive checks in places, but this is not different than we do with smb+dcerpc, its just going to be visible in more places.

A detail that I'd like to consider is this: rename ALPROTO_HTTP to ALPROTO_HTTP1 and use ALPROTO_HTTP as ALPROTO_HTTP_ANY.

victorjulien:
@jasonish can you also have a look at this PR?

catenacyber:
Thanks for the answer. Will work on it

> rename ALPROTO_HTTP to ALPROTO_HTTP1 and use ALPROTO_HTTP as ALPROTO_HTTP_ANY.

No problemo

> It's unfortunate that we need to do the more expensive checks in places

Do you worry about the CPU time using `AppProtoEquals` instead of single equality ?

src/app-layer-protos.c
```diff
@@ -120,6 +120,9 @@ const char *AppProtoToString(AppProto alproto)
         case ALPROTO_HTTP2:
             proto_name = "http2";
             break;
+        case ALPROTO_HTTP_ANY:
+            proto_name = "http_any";
+            break;
```
jasonish, line 125: As Victor mentioned elsewhere on naming..  Should this just output "http", then then HTTP1 and HTTP2 output "http1" and "http2"?

jasonish:
> I think the approach makes sense. It's unfortunate that we need to do the more expensive checks in places, but this is not different than we do with smb+dcerpc, its just going to be visible in more places.
> 
> A detail that I'd like to consider is this: rename ALPROTO_HTTP to ALPROTO_HTTP1 and use ALPROTO_HTTP as ALPROTO_HTTP_ANY.

The approach makes sense to me.  When originally talking about "alert http" being version agnostic, I had a thought that such a rule would kind of be a macro..  Evaluate it against http1, if it parses, add it.  Evaluate against http2, if it parses, add it.. So essentially the rule would be added once for each proto.  Could move some of the expense to initialization time instead of runtime.  But I'm not sure if the expense is worth worrying about?

victorjulien:
> > I think the approach makes sense. It's unfortunate that we need to do the more expensive checks in places, but this is not different than we do with smb+dcerpc, its just going to be visible in more places.
> > A detail that I'd like to consider is this: rename ALPROTO_HTTP to ALPROTO_HTTP1 and use ALPROTO_HTTP as ALPROTO_HTTP_ANY.
> 
> The approach makes sense to me. When originally talking about "alert http" being version agnostic, I had a thought that such a rule would kind of be a macro.. Evaluate it against http1, if it parses, add it. Evaluate against http2, if it parses, add it.. So essentially the rule would be added once for each proto. Could move some of the expense to initialization time instead of runtime. But I'm not sure if the expense is worth worrying about?

I think this would actually put a lot more cost to runtime, as we'd have multiple instances of the same rule in memory, in the rule trees, etc. Likely not worth the small extra cost of the alproto comparison.

catenacyber:
Thanks for this proposal Jason, I had not thought about it
With this proposal, I do not see how we could `DetectSignatureSetAppProto` in detect-http-uri.c and such

src/app-layer-protos.c
catenacyber: Doing this simple change makes S-V test http-connect-fail `for filter {'count': 1, 'match': {'app_proto_orig': 'http', 'event_type': 'flow', 'app_proto': 'dcerpc'}}` because `app_proto_orig` would now be http1
I assumed it is better not to break the API like I did with `AppLayerEventGetProtoByName` in detect-app-layer-event.c

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5668

PR #5664, opened 2020-12-15
Added reset of the ReleasePacket field of the Packet struct

ivov160:
Description:
- Added reset of the ReleasePacket field of the Packet struct

PACKET_REINIT macro - used in the packet pool to reset structure fields but does not reset the ReleasePacket.
Because of this, a reassembled packet (see Defrag4Reassemble) used only in the detect module
can be passed to the capture module (example: netmap and others: af-packet, nfq).

I tested with a netmap. Test case: 
* run suricata in ips mode;
* place suricata in the middle between server and client iperf;
* start iperf (so that packets from the packet pool are reused);
* run ping with large packet size (ping 1.1.1.1 -s 5120).

Other sources have also been fixed. I can add BUG_ON to af-packer, nfq, napatech to check the source of the package (PKT_SRC_WIRE), but not sure about it.

Please correct me if I misunderstood how the defragmentation and detection modules work with packets.

***

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/source-netmap.c
+++ b/src/source-netmap.c
@@ -542,6 +542,8 @@ static void NetmapReleasePacket(Packet *p)
 {
     NetmapThreadVars *ntv = (NetmapThreadVars *)p->netmap_v.ntv;
 
+    BUG_ON(p->pkt_src != PKT_SRC_WIRE);
+
     if ((ntv->copy_mode != NETMAP_COPY_MODE_NONE) && !PKT_IS_PSEUDOPKT(p)) {
         NetmapWritePacket(ntv, p);
     }
--- a/src/decode.h
+++ b/src/decode.h
@@ -756,73 +756,74 @@ void CaptureStatsSetup(ThreadVars *tv, CaptureStats *s);
 /**
  *  \brief Recycle a packet structure for reuse.
  */
-#define PACKET_REINIT(p) do {             \
-        CLEAR_ADDR(&(p)->src);                  \
-        CLEAR_ADDR(&(p)->dst);                  \
-        (p)->sp = 0;                            \
-        (p)->dp = 0;                            \
-        (p)->proto = 0;                         \
-        (p)->recursion_level = 0;               \
-        PACKET_FREE_EXTDATA((p));               \
-        (p)->flags = (p)->flags & PKT_ALLOC;    \
-        (p)->flowflags = 0;                     \
-        (p)->pkt_src = 0;                       \
-        (p)->vlan_id[0] = 0;                    \
-        (p)->vlan_id[1] = 0;                    \
-        (p)->vlan_idx = 0;                      \
-        (p)->ts.tv_sec = 0;                     \
-        (p)->ts.tv_usec = 0;                    \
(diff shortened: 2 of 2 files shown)


victorjulien:
Can you split this into at least 2 commits? One that changes the macro to the `do { .. } while(0)` loop, the other making the actual fix?

victorjulien:
Btw, I wonder if it should just be a function. The logic is expensive anyway, so might not make sense to have it be a macro.

src/source-netmap.c
```diff
@@ -542,6 +542,8 @@ static void NetmapReleasePacket(Packet *p)
 {
     NetmapThreadVars *ntv = (NetmapThreadVars *)p->netmap_v.ntv;
 
+    BUG_ON(p->pkt_src == PKT_SRC_WIRE);
```
victorjulien, line 545: I don't understand this, why is it a bug to have the `PKT_SRC_WIRE` value here?

ivov160: This is my typo, of course it should be BUG_ON (p-> pkt_src! = PKT_SRC_WIRE); I added a commit to fix this.

ivov160:
> Btw, I wonder if it should just be a function. The logic is expensive anyway, so might not make sense to have it be a macro.

ivov160:
> Can you split this into at least 2 commits? One that changes the macro to the `do { .. } while(0)` loop, the other making the actual fix?

Ok. I split the commit in two and rebased the PR.

ivov160:
> Btw, I wonder if it should just be a function. The logic is expensive anyway, so might not make sense to have it be a macro.

I agree. This will also help fix compilation for all variants. Most builds now fail because the PacketPoolReturnPacket function is not defined in decode.h. If rewrite REINIT_PACKET as a function and move it to decode.c the problem will be fixed.

norg:
Could you create a fresh PR with all the fixes included?

victorjulien:
Looking at this again I'm not sure I understand how we could end up with a case where the `Packet::ReleasePacket` pointer is wrong. Could you explain the steps that lead to this?

victorjulien:
I think I see an issue now. Steps:
# afpacket (or netmap) gets a packet from the pool, sets its custom `ReleasePacket` callback
# at the end of the pipeline the packet is returned to the pool but this leaves the `ReleasePacket` untouched
# defrag (or tunnel) gets this packet from the packet pool. `ReleasePacket` is still untouched, meaning its set to whatever afpacket set it to.

I think this is indeed an issue that needs fixing, although also wonder why we're not seeing lots of crashes. Perhaps the main reason is that at least in workers mode the capture method essentially reuses the same packet all the time, so defrag wouldn't normally get a packet that was used by afpacket.

victorjulien:
Think this was fixed with https://github.com/OISF/suricata/commit/07ce871da4b385f6f1d86310f5dcdf00e6c79692

PR #5668, opened 2020-12-17
Http2 keywords 4067 v5

catenacyber:

Describe changes:
- Adds signature protocol extending flow protocol (one signature protocol matches multiple flow protocols cf `AppProtoEquals`)
- Adds `ALPROTO_HTTP_ANY` (as such a signature protocol) for signatures to match on both HTTP1 and HTTP2
- Rename `ALPROTO_HTTP` to `ALPROTO_HTTP1` (and `ALPROTO_HTTP` is now for any version)
- `http.uri` keyword now matches on HTTP2 traffic (as well as `http.uri.raw`)
- `http.user_agent` keyword now matches on HTTP2 traffic

suricata-verify-pr: 383

Makes S-V test in https://github.com/OISF/suricata-verify/pull/383 pass

What remains to be done :
- rename `alproto` to `sig_alproto`. Thoughts about a better name ?
- Commits squashing  (commit-check)
- CI (commit-check)
- Do all the keywords

Modifies #5660 by 
- splitting in 2 commits : one introducing `AppProtoEquals` and the next one adding `ALPROTO_HTTP_ANY`
- renaming ALPROTO_HTTP* constants
- defining generic `http2_frames_get_header_value` for code reuse
- http.user_agent keyword now works for HTTP2
- http.uri.raw keyword now works for HTTP2


catenacyber:
> Do all the keywords

What about urilen ?
It seems to be usable by SIP cf `git grep DetectUrilenValidateContent`

Otherwise, it seems the keywords are all in detect-http*.c, right ?

src/detect-http-uri.c
```diff
@@ -153,6 +153,13 @@ void DetectHttpUriRegister (void)
     DetectAppLayerMpmRegister2("http_raw_uri", SIG_FLAG_TOSERVER, 2, PrefilterGenericMpmRegister,
             GetRawData, ALPROTO_HTTP1, HTP_REQUEST_LINE);
 
+    // no difference between raw and decoded uri for HTTP2
```
catenacyber, line 156: What do we want for this ?
I guess the good way to go is to test urlencoded uris against a real web server...

src/detect-http-ua.c
```diff
@@ -178,6 +181,29 @@ static InspectionBuffer *GetData(DetectEngineThreadCtx *det_ctx,
     return buffer;
 }
 
+static InspectionBuffer *GetData2(DetectEngineThreadCtx *det_ctx,
```
catenacyber, line 184: Do we want to have a generic function for this ?
Having a callback as an argument to get the buffer and its length

catenacyber:
The way to add another keyword is to :

First create a S-V check with a signature on HTTP2 traffic with this keyword cf https://github.com/OISF/suricata-verify/pull/383/files#diff-1f4ea8592d784dced51ca29678925c0c2d6413c6b4f8685fde43bf86142bb244R5
`alert http2 any any -> any any (http.user_agent; content:"curl"; sid:20;)`

Then change Suricata to get S-V test to pass, that is following the 4 commits in this PR about  http.user_agent

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5669

PR #5671, opened 2020-12-18
issue 4216 - filestore: fix global counter init in unix socket mode - v1

jasonish:
Redmine issue:
https://redmine.openinfosecfoundation.org/issues/4216

Move initialization of filestore global counter to PreRunInit,
so they get registered during program initialization, or as
required in unix-socket mode, initialized for each file run.

Will do 5.0.x backport as well.

### DIFF
--- a/src/output-filestore.c
+++ b/src/output-filestore.c
@@ -527,9 +527,6 @@ static OutputInitResult OutputFilestoreLogInitCtx(ConfNode *conf)
         }
     }
 
-    StatsRegisterGlobalCounter("file_store.open_files",
-            OutputFilestoreOpenFilesCounter);
-
     result.ctx = output_ctx;
     result.ok = true;
     SCReturnCT(result, "OutputInitResult");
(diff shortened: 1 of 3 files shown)


src/output-filestore.c
```diff
@@ -549,3 +546,10 @@ void OutputFilestoreRegister(void)
     SC_ATOMIC_SET(filestore_open_file_cnt, 0);
 #endif
 }
+
+void OutputFilestoreRegisterGlobalCounters(void)
+{
+#ifdef HAVE_NSS
```
victorjulien, line 552: file extract does not depend on nss, so why would this counter depend on it?

jasonish: This was previously called behind HAVE_NSS guard, as the filestore output does require NSS as sha256 is used during the filename processing.

victorjulien:
I've merged this into `master-6.0.x` through #5860 as the code has changed in `master` with the rust hashing work. Can you do a branch for master as well?

victorjulien:
Ping @jasonish

jasonish:
> I've merged this into `master-6.0.x` through #5860 as the code has changed in `master` with the rust hashing work. Can you do a branch for master as well?

I think just the guard has to be removed is all.  CI running, PR to come.

PR #5674, opened 2020-12-22
Multi-threaded EVE support for output plugins

jlucovsky:
This commit includes draft changes to support EVE 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4126](https://redmine.openinfosecfoundation.org/issues/4126)

Describe changes:
These changes build upon the experimental file output plugin support by
adding calls when a thread attaches and detaches from an output context:
- Thread initialization function returns per-thread context
- Thread deinitialization function uses context to remove resources

I've modified @jasonish's output plugin with a multi-threaded variant: https://github.com/jlucovsky/suricata-example-plugins/tree/output/1/eve-file-type

Configure Suricata to use this plugin by adding the following to `suricata.yaml`

```
plugins:
    threaded-filetype-plugin: /path/to/plugin.so
```

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2013-2020 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -842,6 +842,8 @@ static TmEcode JsonAlertLogThreadDeinit(ThreadVars *t, void *data)
 
     MemBufferFree(aft->json_buffer);
     MemBufferFree(aft->payload_buffer);
+
+    LogFileDeInitCtx(aft->json_output_ctx->file_ctx, t->id);
 
     /* clear memory */
     memset(aft, 0, sizeof(JsonAlertLogThread));
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -1186,7 +1186,7 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
             ConfNode *plugin_conf = ConfNodeLookupChild(conf,
                 json_ctx->plugin->name);
             void *plugin_data = NULL;
-            if (json_ctx->plugin->Open(plugin_conf, &plugin_data) < 0) {
+            if (json_ctx->plugin->Init(plugin_conf, false, &plugin_data) < 0) {
                 LogFileFreeCtx(json_ctx->file_ctx);
                 SCFree(json_ctx);
                 SCFree(output_ctx);
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2020 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -1105,20 +1105,19 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
             json_ctx->file_ctx->prefix_len = strlen(prefix);
         }
 
+        /* Threaded file output */
+        const ConfNode *threaded = ConfNodeLookupChild(conf, "threaded");
+        if (threaded && threaded->val && ConfValIsTrue(threaded->val)) {
+            SCLogConfig("Threaded EVE logging configured");
+            json_ctx->file_ctx->threaded = true;
+        } else {
+            json_ctx->file_ctx->threaded = false;
+        }
+
         if (json_ctx->json_out == LOGFILE_TYPE_FILE ||
             json_ctx->json_out == LOGFILE_TYPE_UNIX_DGRAM ||
             json_ctx->json_out == LOGFILE_TYPE_UNIX_STREAM)
         {
-            if (json_ctx->json_out == LOGFILE_TYPE_FILE) {
-                /* Threaded file output */
-                const ConfNode *threaded = ConfNodeLookupChild(conf, "threaded");
-                if (threaded && threaded->val && ConfValIsTrue(threaded->val)) {
-                    SCLogConfig("Enabling threaded eve logging.");
-                    json_ctx->file_ctx->threaded = true;
-                } else {
-                    json_ctx->file_ctx->threaded = false;
-                }
-            }
 
             if (SCConfLogOpenGeneric(conf, json_ctx->file_ctx, DEFAULT_LOG_FILENAME, 1) < 0) {
                 LogFileFreeCtx(json_ctx->file_ctx);
@@ -1175,6 +1174,7 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
             }
 
             if (SCConfLogOpenRedis(redis_node, json_ctx->file_ctx) < 0) {
+                SCFree(json_ctx->file_ctx->sensor_name);
                 LogFileFreeCtx(json_ctx->file_ctx);
                 SCFree(json_ctx);
                 SCFree(output_ctx);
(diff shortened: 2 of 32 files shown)


src/output-json.c
```diff
@@ -1190,27 +1189,42 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
...
+            if (append == NULL)
+                append = DEFAULT_LOG_MODE_APPEND;
+
+            SCPluginFileOptions opts = { .threaded = json_ctx->file_ctx->threaded,
+                .log_dir = ConfigGetLogDirectory(),
+                .append = append };
+            if (json_ctx->file_ctx->threaded) {
+                if (!SCLogOpenThreadedFileFp(opts.log_dir, opts.append, json_ctx->file_ctx, 1)) {
+                    SCLogNotice("can't open");
+                }
```
jasonish, line 1215: The naming suggests we're opening file pointers here. Is that the case?

jlucovsky: No -- I'll rename the function to be more indicative of what it's doing -- preparing slots for thread opens.

src/output-json-alert.c
```diff
@@ -836,6 +836,8 @@ static TmEcode JsonAlertLogThreadDeinit(ThreadVars *t, void *data)
     MemBufferFree(aft->json_buffer);
     MemBufferFree(aft->payload_buffer);
 
+    LogFileDeInitCtx(aft->json_output_ctx->file_ctx, t->id);
```
jasonish, line 839: Do we do this in each logger to perform cleanup at shutdown we should have always been doing?  Its just clearer when doing in plugin context?

jasonish: Hmm.. Nevermind. We have to do it in all possible loggers as its the only place to run some generic cleanup right? Thats unfortunate, but shouldn't block us on this. Just points at a pain point in the logging code I think.

src/suricata-plugin.h
```diff
@@ -41,16 +41,30 @@ typedef struct SCPlugin_ {
     void (*Init)(void);
 } SCPlugin;
 
+typedef struct SCPluginFileOptions_ {
+    bool threaded;
+    const char *filename; /* Configured file name */
+    const char *log_dir;
+    const char *append;
```
jasonish, line 48: I'm not to keen on seeing file specific things in here.  For the log dir `ConfGet("default-log-dir", ...)` can be used. This is easier now with the libsuricata PR in as you can do a `make install-headers` and you have all the headers available.

The `ConfNode` is already passed in which can be used to discover the filename, and append mode, if needed.

Threaded should be considered separately. The `output` plugin *has* to take this value into consideration right?  If the plugin doesn't support threading, but threaded is true things are going to fail as the threading happens at a higher (or lowever - whatever) level.  Making it a parameter makes it more clear it needs to be considered.

Other thought: could the plugin choose to reject the threaded value and still work OK? I'm guessing not, at the very least it would need to know its multi-threaded to do some lock operation.

src/suricata-plugin.h
```diff
@@ -41,16 +41,30 @@ typedef struct SCPlugin_ {
...
+    /* Called on first open; no other calls will be made if unthreaded */
+    int (*Open)(ConfNode *conf, SCPluginFileOptions *opt, void **ctx);
+    /* Write - Called on each write to the object */
+    int (*Write)(const char *buffer, int buffer_len, void *ctx, void *thread_ctx);
+    /* Close - Called on final close */
     void (*Close)(void *ctx);
+    /* Open - Called for each thread using the file object */
+    int (*ThreadOpenFunc)(void *ctx, int thread_id, void **thread_ctx);
+    /* Close - Called for each thread using the file object */
+    int (*ThreadCloseFunc)(void *ctx, void *thread_ctx);
```
jasonish, line 67: The naming still confuses me...  Maybe its the multiple opens..
* Open -> Init: as its a one time initializer, could open files...
* ThreadOpenFunc -> ThreadInit

Then we document... If not threaded, you will get a one time call to Init.. Open your file, database connection socket, etc. here... If threading is enabled this is where you will get your configuration data, another call for each thread will be made the ThreadInit where you can open files, databases, sockets, whatever..

jasonish, line 67: Naming... In Suricata was use the Func function for some funcs, for function pointer variable names and for function point types.  I propose that in the plugin API we save the "Func" suffext for function type declartions, but not for the variable names..

We could `typedef int ThreadOpenFunc(...)`, but for end-users its often just easier to see the function definitions right in the initialization structure as well.

Just thinking out loud here.. Thoughts?

jasonish (changes requested):
I think a lot of it comes to naming.

In my tests, it works well and I haven't found any issues yet. Would like to test with something like a Redis output, perhaps written in Rust.

src/output-json-alert.c
jlucovsky: This call is necessary .. for the reasons stated.

src/suricata-plugin.h
jlucovsky: I'll make the `conf` related changes -- good point.

Regarding threaded -- if threading is requested and the plugin can't support it, then the plugin should "fail" (and not work). Making it an explicit param is a good idea -- with failure if there's a mismatch.

src/suricata-plugin.h
jlucovsky: I don't mind changing the name to avoid confusion .... `ThreadOpen` make sense?

src/util-logopenfile.c
```diff
@@ -767,37 +765,45 @@ static bool LogFileNewThreadedCtx(LogFileCtx *parent_ctx, const char *log_path,
...
+                    thread_id);
+            goto error;
+        }
+        thread->is_regular = true;
+        thread->Write = SCLogFileWriteNoLock;
+        thread->Close = SCLogFileCloseNoLock;
+        OutputRegisterFileRotationFlag(&thread->rotation_flag);
+    } else if (parent_ctx->type == LOGFILE_TYPE_PLUGIN) {
+        thread->plugin.plugin->ThreadInit(
+                thread->plugin.ctx, thread_id, &thread->plugin.thread_ctx);
```
jasonish, line 791: I think we should check that this is not NULL.  Probably a fatal error instead of crashing.

jlucovsky: Hmm .... we could also sanity check the returned `plugin_data` structure from `SCPluginFindFileType` to ensure the proper function pointer slots are filled?

jasonish: That would work as well. Do we know at that point if ThreadInit and ThreadDeinit are required?

jlucovsky: We do if `Init` succeeds and `threaded = true`.

src/suricata-plugin.h
```diff
@@ -45,12 +45,22 @@ typedef struct SCPlugin_ {
...
-    int (*Write)(const char *buffer, int buffer_len, void *ctx);
-    void (*Close)(void *ctx);
+    /* Init Called on first access */
+    int (*Init)(ConfNode *conf, bool threaded, void **init_data);
+    /* Write - Called on each write to the object */
+    int (*Write)(const char *buffer, int buffer_len, void *init_data, void *thread_data);
+    /* Close - Called on final close */
+    void (*Deinit)(void *init_data);
+    /* ThreadInit - Called for each thread using file object*/
+    int (*ThreadInit)(void *init_data, int thread_id, void **thread_data);
```
jasonish, line 61: Seeing Victor's comments on thread local storage, is this needed? Or could thread_init just stuff its data in thread local storage?

jlucovsky: We could use TLS in lieu and yes, for the second question. Both come with extra work, which is fine.

Strictly speaking, the changes victor proposed are somewhat orthogonal but related as there as lots of places in suri that _could be_ modified to use tls over the pattern used here (and elsewhere in suri).

jasonish:
On the whole I think this is good to take out of draft.  We need this feature.

I'd like to discuss TLS further.. Does it make sense? Given this is a brand new feature, would TLS make more sense? Easier to use? Any drawbacks? My simple tests suggest that it works well in plugin scope, at least from C.  I still need to check out how Rust handles thread locals, as well as how it integrates with FFI.

jlucovsky:
> On the whole I think this is good to take out of draft. We need this feature.
> 
> I'd like to discuss TLS further.. Does it make sense? Given this is a brand new feature, would TLS make more sense? Easier to use? Any drawbacks? My simple tests suggest that it works well in plugin scope, at least from C. I still need to check out how Rust handles thread locals, as well as how it integrates with FFI.

I think TLS makes sense ... perhaps we could keep this in draft form until the Rust/FFI investigation completes?

jasonish:
> I think TLS makes sense ... perhaps we could keep this in draft form until the Rust/FFI investigation completes?

Actually I'm not sure if TLS would work.  We can have multiple instances of eve loggers with a different configuration, perhaps the same plugin but configured differently. This probably wouldn't work with TLS.

jlucovsky:
Continued in #6017

PR #5676, opened 2020-12-22
filestore: use Rust to provde md5/sha1/sha256 support - v0

jasonish:
This is a draft to show how we can provide sha1/sha256/md5 support
with Rust. This support uses crates from the RustCrypto project
that provide pure Rust implementations for many hash algorithms.

Some simple testing with sha256 showed that it was about 10%
faster than libnss.

This PR only touches that filestore and related detect. But it
should be trivial update all sha1/sha256/md5 use to these functions.

I added the RustCrypto md5 crate as md5c as it conflicts with an
md5 crate we already use. My plan is to replace the one we already
use with the one from the RustCrypto project, as all hashes from
that project use the same Digest trait allowing us to reduce some
code.

Testing with a Suricata with no NSS support, all Suricata-Verify
filestore tests pass.

TODO:
- Suricata-Verify will need some tweaks as it will require NSS on 6.0
  and less. But many tests will no longer require NSS on 7.0+.
- Replace all md5/sha1/sha256 usage in C with these Rust functions.

Issues;
- The RustCrypto project has a MSRV (minimum support Rust version) of 1.41 at this time. They appear to test this, document it, and stick to their MSRV per minor version.

Update:
- Enable JA3 without NSS support.

### DIFF
--- a/rust/src/ffi/hashing.rs
+++ b/rust/src/ffi/hashing.rs
@@ -16,7 +16,7 @@
  */
 
 use digest::Digest;
-use md5c::Md5;
+use md5::Md5;
 use sha1::Sha1;
 use sha2::Sha256;
 
--- a/rust/src/ffi/hashing.rs
+++ b/rust/src/ffi/hashing.rs
@@ -15,7 +15,9 @@
  * 02110-1301, USA.
  */
 
-use sha2::{Digest, Sha256};
+use digest::Digest;
+use sha1::Sha1;
+use sha2::Sha256;
 
 pub const SC_SHA256_LEN: usize = 32;
 
@@ -32,17 +34,13 @@ pub extern "C" fn SCSha256New() -> *mut SCSha256 {
 
 #[no_mangle]
 pub unsafe extern "C" fn SCSha256Update(hasher: &mut SCSha256, bytes: *const u8, len: u32) {
-    let data = std::slice::from_raw_parts(bytes, len as usize);
-    hasher.0.update(data);
+    update(&mut hasher.0, bytes, len);
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn SCSha256Finalize(hasher: &mut SCSha256, out: *mut u8, len: u32) {
     let hasher: Box<SCSha256> = Box::from_raw(hasher);
-    let result = hasher.0.finalize();
-    let output = std::slice::from_raw_parts_mut(out, len as usize);
-    // This will panic if the sizes differ.
-    output.copy_from_slice(&result);
+    finalize(hasher.0, out, len);
 }
 
 /// C function to finalize the Sha256 hasher to a hex string.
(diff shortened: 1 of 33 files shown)


victorjulien:
Awesome work Jason!

It seems MSRV of 1.41 is ok, esp for the 7 branch.

One thing that we might have to review is if the hashing support can be disabled everywhere. I know right now ppl build suri w/o nss to globally disable hashing. I *think* they do this for performance reasons, but maybe there are other use cases too.

rust/src/ffi/hashing.rs
```diff
@@ -0,0 +1,71 @@
...
+    let output = std::slice::from_raw_parts_mut(out, len as usize);
+    output.copy_from_slice(&result);
+}
+
+/// C function to finalize the Sha256 hasher to a hex string.
+///
+/// Notes:
+/// - The output pointer is of type *mut u8 as I'm not sure how to perform the copy
+///   from the hex string to C string efficiently. I believe the code is still correct
+///   but C function signature isn't ideal bing *uint8_t instead of *char.
```
victorjulien, line 52: can we just change the type we pass to this function?

victorjulien: nit: s/bing/being/g ?

victorjulien:
As a next clean up step it would be nice to remove our `util-crypt.[ch]` sha code and see if there are any consumers left and convert them over as well.

rust/src/ffi/hashing.rs
jasonish: Turned out to be trivial to have c_char in the function, the cast it to u8 which gives us the function signature we want.

jasonish:
> Awesome work Jason!
> 
> It seems MSRV of 1.41 is ok, esp for the 7 branch.
> 
> One thing that we might have to review is if the hashing support can be disabled everywhere. I know right now ppl build suri w/o nss to globally disable hashing. I _think_ they do this for performance reasons, but maybe there are other use cases too.

Interesting. Most hashing was wrapped in HAVE_NSS disabling NSS accomplished the job, with the exception of the ssh parser that used a Rust md5 implementation, and then community ID, and then the SSL app-layer which uses ComputeSHA1 which works without NSS.

Options could be a compile time flag, `--disable-hashing`, this wouldn't compile it out, but we could use it to put guards around where we had HAVE_NSS, but still use it for community-id, ssl, etc - where it can't be disabled now.

I guess the other option is a global run time flag, a bit more involved though.

It looks like the main culprit here would be the filestore.It looks like hashing is off by default, however, a rule can come a long and turn that hashing on provided nss support is built in. And I don't see a current way to disable that.

jasonish:
> Awesome work Jason!
> 
> It seems MSRV of 1.41 is ok, esp for the 7 branch.
> 
> One thing that we might have to review is if the hashing support can be disabled everywhere. I know right now ppl build suri w/o nss to globally disable hashing. I _think_ they do this for performance reasons, but maybe there are other use cases too.

What a global flag could look like to disable hashing:

https://github.com/OISF/suricata/commit/233c6591a2c95902fb4684cf674f2bdf5c080d18

victorjulien:
> > Awesome work Jason!
> > It seems MSRV of 1.41 is ok, esp for the 7 branch.
> > One thing that we might have to review is if the hashing support can be disabled everywhere. I know right now ppl build suri w/o nss to globally disable hashing. I _think_ they do this for performance reasons, but maybe there are other use cases too.
> 
> What a global flag could look like to disable hashing:
> 
> [233c659](https://github.com/OISF/suricata/commit/233c6591a2c95902fb4684cf674f2bdf5c080d18)

Wonder if we should make it dynamic while at it. There'd be 2 components to it: an atomic flag to check instead of this `bool`, and trickier proper handling of enabling/disabling in the middle of a flow/file or whatever handling. The case I worry about is that an operation requiring a "start" and "end" would call "end" w/o "start" having run due to it still being disabled. Also a bit hard to test this reliably...

jasonish:
> Wonder if we should make it dynamic while at it. There'd be 2 components to it: an atomic flag to check instead of this `bool`, and trickier proper handling of enabling/disabling in the middle of a flow/file or whatever handling. The case I worry about is that an operation requiring a "start" and "end" would call "end" w/o "start" having run due to it still being disabled. Also a bit hard to test this reliably...

What would be the usecase for toggling it at runtime?

victorjulien:
Part of tuning/perf analysis is what comes to mind, but I don't really have a specific use case. More https://redmine.openinfosecfoundation.org/issues/3305

PR #5683, opened 2020-12-26
eve/log: add src and dest info to flow in alert

regit:
When looking at an alert event, it was impossible to determine which
side from src or dest IP in the alert was the client and wich side
was the server with regards to the underlying flow. This was a problem
when you try to known who belongs a metadata property such as a HTTP
hostname or a TLS JA3.

This patch updates the code to add src and dest IP in the flow
subobject as well as src and dst port. This way, we can now which
side is the client and which side is the server.

The result is looking like:
```JSON
{
  "event_type": "alert",
  "src_ip": "22.47.184.196",
  "src_port": 81,
  "dest_ip": "192.168.1.47",
  "dest_port": 1063,
  "proto": "TCP",
  "tx_id": 0,
  "alert": {
    "signature_id": 2018959,
    "rev": 3,
  },
  "app_proto": "http",
  "flow": {
    "pkts_toserver": 22,
    "pkts_toclient": 35,
    "bytes_toserver": 1370,
    "bytes_toclient": 48852,
    "start": "2009-10-28T10:01:46.755232+0100",
    "src_ip": "192.168.1.47",
    "dest_ip": "22.47.184.196",
    "src_port": 1063,
    "dest_port": 81
  }
}
```

Describe changes:
- Update flow sub object in alert log

### DIFF
--- a/src/output-json-flow.c
+++ b/src/output-json-flow.c
@@ -64,6 +64,32 @@ typedef struct JsonFlowLogThread_ {
     MemBuffer *buffer;
 } JsonFlowLogThread;
 
+static void EveBuildFlowIPParams(
+        const Flow *f, char *srcip, size_t srclen, char *dstip, size_t dstlen, Port *sp, Port *dp)
+{
+    if ((f->flags & FLOW_DIR_REVERSED) == 0) {
+        if (FLOW_IS_IPV4(f)) {
+            PrintInet(AF_INET, (const void *)&(f->src.addr_data32[0]), srcip, srclen);
+            PrintInet(AF_INET, (const void *)&(f->dst.addr_data32[0]), dstip, dstlen);
+        } else if (FLOW_IS_IPV6(f)) {
+            PrintInet(AF_INET6, (const void *)&(f->src.address), srcip, srclen);
+            PrintInet(AF_INET6, (const void *)&(f->dst.address), dstip, dstlen);
+        }
+        *sp = f->sp;
+        *dp = f->dp;
+    } else {
+        if (FLOW_IS_IPV4(f)) {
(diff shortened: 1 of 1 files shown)


jlucovsky:
LGTM

catenacyber:
Another (less json verbose) way would be to add some json field `server` or `direction` in the `alert` part to tell if the alert source is the flow source or the flow server...

regit:
> Another (less json verbose) way would be to add some json field `server` or `direction` in the `alert` part to tell if the alert source is the flow source or the flow server...

I thought about that too but this was making queries like `give me all alerts where IP 1.2.3.4 is a server` more complex than with the verbose solution.

jasonish:
> > Another (less json verbose) way would be to add some json field `server` or `direction` in the `alert` part to tell if the alert source is the flow source or the flow server...
> 
> I thought about that too but this was making queries like `give me all alerts where IP 1.2.3.4 is a server` more complex than with the verbose solution.

This seems to come up every so often.  Should the JSON be in format that is an easily queryable row? This leads to a denormalized form often with duplication of data.  Or should the JSON be limited to providing the info to create easily queryable data which will give us smaller JSON records, but at the expense of requiring post-processing depending on your database.

The less verbose form could require a rather simple join in Postgres, but does make an Elasticsearch for query much more complex.

victorjulien:
I don't like the duplication. I prefer a field indicating the direction.

regit:
> I don't like the duplication. I prefer a field indicating the direction.

So if we add a direction flag with values in to_client, to_server and want to answer to the following question:

What servers are involved in attack ?

In pseudo code we will have:

```
If direction == to_client then SERVER_IP = src_ip else SERVER_IP = dest_ip
```

It should be doable in SQL but I don't see how we can do this in Elasticsearch.

regit:
Replaced by #6066

PR #5691, opened 2020-12-29
hashing: use Rust to provide sha256/sha1/md5 hashing funcionality - v2

jasonish:
Replaces libnss with C bindings to Rust implementations of sha256/sha1/md5 from the RustCrypto project.
- Hashing that was guarded by `#ifdef HAVE_NSS` is now wrapped in a check for `g_disable_hashing`.  The idea is that `g_disable_hashing` can be set to true to get the same behaviour as building without libnss.
- I have not added a way to set this `g_disable_hashing` yet.
- This bumps the MSRV to 1.41, so our CI test for Rust 1.34 fails.

Previous PR:
- https://github.com/OISF/suricata/pull/5676

Changes from last PR:
- Removes requirement of libnss and libnspr completely.


src/util-ja3.c
```diff
@@ -235,19 +233,15 @@ char *Ja3GenerateHash(JA3Buffer *buffer)
...
 
-    unsigned char md5[MD5_LENGTH];
-    HASH_HashBuf(HASH_AlgMD5, md5, (unsigned char *)buffer->data, buffer->used);
+    unsigned char md5[SC_MD5_LEN];
+    SCMd5HashBuffer((unsigned char *)buffer->data, buffer->used, md5, sizeof(md5));
 
     int i, x;
-    for (i = 0, x = 0; x < MD5_LENGTH; x++) {
+    for (i = 0, x = 0; x < SC_MD5_LEN; x++) {
         i += snprintf(ja3_hash + i, MD5_STRING_LENGTH - i, "%02x", md5[x]);
```
catenacyber, line 241: Maybe we can enjoy this opportunity to remove this dangerous pattern `i += snprintf(buffer + i,` cf https://access.redhat.com/blogs/766093/posts/1976193

catenacyber:
This looks great by the way

src/util-ja3.c
jasonish: Yeah, I have to give it another pass of where I can use the `FinalizeToHex` functions I put into the Rust hashing module. It would replace this one easily as we're only interested in the hex value as a string.

jasonish:
Next version: https://github.com/OISF/suricata/pull/5695

src/util-ja3.c
jasonish: Addressed in https://github.com/OISF/suricata/pull/5695. With the exception of the TLS fingerprint which uses a ":" separated hex string, so even with Rust we'd have to create a custom formatter for. Perhaps something to consider, but not a simple drop-in replacement.

PR #5694, opened 2020-12-30
mime: postpone md5 calculation to parse complete - v1

jasonish:
Instead of calculating the MD5 at the end of every part, only
compute it when parsing is complete.

With libnss, the hash never updates after the first HASH_End, so
the md5 of only the first part of the body is logged, rather than
the md5 of all the parts.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/4245

suricata-verify-pr: 391

### DIFF
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -2094,13 +2094,6 @@ static int ProcessBodyComplete(MimeDecParseState *state)
         }
     }
 
-#ifdef HAVE_NSS
-    if (state->md5_ctx) {
-        unsigned int len = 0;
-        HASH_End(state->md5_ctx, state->md5, &len, sizeof(state->md5));
-    }
-#endif
-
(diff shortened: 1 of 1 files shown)


jasonish:
@regit Can you take a look?

src/util-decode-mime.c
```diff
@@ -2547,6 +2540,13 @@ int MimeDecParseComplete(MimeDecParseState *state)
         return ret;
     }
 
+#ifdef HAVE_NSS
+    if (state->md5_ctx) {
+        unsigned int len = 0;
```
catenacyber, line 2545: This `len` feels weird... Is this something from NSS ?

jasonish: Yeah. I think it returns to you the size of the digest.

victorjulien:
Applied to `master-6.0.x` through #5726, thanks!

PR #5695, opened 2020-12-30
hashing: use Rust to provide sha256/sha1/md5 hashing funcionality - v4

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5691

Changes from last PR:
- Fix commit check.
- Remove some += snprintf hex string building where we can.  We can't in one for TLS fingerprints that use a ":" separated hex string.
- Fix for https://redmine.openinfosecfoundation.org/issues/4245


src/app-layer-ssl.c
```diff
@@ -3022,18 +3021,17 @@ void RegisterSSLParsers(void)
...
-                         "no MD5 calculation support built in (LibNSS), disabling JA3");
-            SC_ATOMIC_SET(ssl_config.enable_ja3, 0);
-        }
-#else
-        if (RunmodeIsUnittests()) {
-            SC_ATOMIC_SET(ssl_config.enable_ja3, 1);
+        if (g_disable_hashing) {
+            if (SC_ATOMIC_GET(ssl_config.enable_ja3)) {
+                SCLogWarning(SC_WARN_NO_JA3_SUPPORT,
+                        "no MD5 calculation support built in (LibNSS), disabling JA3");
```
victorjulien, line 3027: message refers to libnss

victorjulien: Other than the Changelog, there are a few more to remove:
```
git grep -i libnss
.github/workflows/builds.yml:                libnss3-dev \
.github/workflows/builds.yml:                libnss3-dev \
.github/workflows/builds.yml:                libnss3-dev \
.travis.yml:      - libnss3-dev
ChangeLog:Bug #3465: build-info and configure wrongly display libnss status
ChangeLog:Bug #1427: configure with libnss and libnspr
ChangeLog:Optimization #2461: Let user to explicit disable libnss and libnspr support
ChangeLog:- Improve libnss detection, thanks to Christian Kreibich
doc/userguide/install.rst:  libnss, libgeoip, liblua5.1, libhiredis, libevent
doc/userguide/install.rst:                    libnss3-dev libgeoip-dev liblua5.1-dev libhiredis-dev libevent-dev \
doc/userguide/rules/transforms.rst:.. note:: depends on libnss being compiled into Suricata
doc/userguide/rules/transforms.rst:.. note:: depends on libnss being compiled into Suricata
doc/userguide/rules/transforms.rst:.. note:: depends on libnss being compiled into Suricata
src/app-layer-ssl.c:                        "no MD5 calculation support built in (LibNSS), disabling JA3");
```

src/detect-transform-md5.c
```diff
@@ -82,6 +71,11 @@ static int DetectTransformToMd5Setup (DetectEngineCtx *de_ctx, Signature *s, con
 static int DetectTransformToMd5Setup (DetectEngineCtx *de_ctx, Signature *s, const char *nullstr)
 {
     SCEnter();
+    if (g_disable_hashing) {
+        SCLogError(SC_ERR_HASHING_DISABLED, "MD5 hashing has been disabled, "
```
victorjulien, line 75: Wonder if we need to be a bit more relax here. This would fail loading the set, but if we want rule writers to be comfortable including support like this into their set we may only want to fail if we're in strict mode?

src/output-json-email-common.c
```diff
@@ -53,6 +53,10 @@
 #include "output-json.h"
 #include "output-json-email-common.h"
 
+#ifdef HAVE_NSS
```
victorjulien, line 56: To be removed

src/util-decode-mime.h
```diff
@@ -197,10 +199,9 @@ typedef struct MimeDecParseState {
     uint8_t bvremain[B64_BLOCK];  /**< Remainder from base64-decoded line */
     uint8_t bvr_len;  /**< Length of remainder from base64-decoded line */
     uint8_t data_chunk[DATA_CHUNK_SIZE];  /**< Buffer holding data chunk */
-#ifdef HAVE_NSS
-    HASHContext *md5_ctx;
-    uint8_t md5[MD5_LENGTH];
-#endif
+    SCMd5 *md5_ctx;
+    char md5hex[SC_MD5_HEX_LEN + 1];
```
victorjulien, line 203: I'm not really happy with growing the data structure like this. Can we do it some other way?

src/util-file.h
```diff
@@ -25,14 +25,23 @@
...
 
-#ifdef HAVE_NSS
-#include <sechash.h>
-#endif
-
 #include "conf.h"
 
 #include "util-streaming-buffer.h"
 
+/* Hack: Pulling rust.h to get the SCSha256 causes all sorts of problems with
```
victorjulien, line 32: Any ideas for resolving this?

src/util-lua-common.c
```diff
@@ -715,42 +715,41 @@ static int LuaCallbackLogError(lua_State *luastate)
...
-            strlcat(md5, one, sizeof(md5));
+    char *md5ptr = NULL;
+    char *sha1ptr = NULL;
+    char *sha256ptr = NULL;
+    if (!g_disable_hashing) {
+        char md5[33] = "";
+        md5ptr = md5;
+        if (file->flags & FILE_MD5) {
+            size_t x;
+            for (x = 0; x < sizeof(file->md5); x++) {
```
victorjulien, line 726: Can we use a `ToHex` func here as well?

victorjulien:
While trying to build this I'm getting
```
error: no matching package named `digest` found
```

jasonish:
> While trying to build this I'm getting
> 
> ```
> error: no matching package named `digest` found
> ```

Strange.  Its also a dependency of the sha-1, sha256 and md5 crates? Could it be a local issue as it works fine on CI? In this intel?

src/util-decode-mime.h
jasonish: This is simplest.  We could:
- Go back to how it was before and just keep the binary hash.  Use snprintf while logging.
- FinalizeToHex in logging.  But this would allow the logger to run once.  So I'd have to add a "Clone" method to clone the context in the logger, then that could finalize to hex. This postpones the hex til logging, and still allows multiple loggers, at the expense of copying the context. This would shrink the struct as no buffer would be required at all.
Anything thing else would kind of be out of scope for this PR I think.

src/util-file.h
jasonish: No, spent a lot of time trying to get `rust.h` included here but it just leads to too many conflicts.. I think because this is included in detect.h, but the Rust headers depends on some app-layer stuff. Maybe its something some forward declarations in just the right spot could solve? Can revisit when the rest this is OK, as like a said, a lot of time spent here so far with no gain.

src/util-decode-mime.h
victorjulien: Perhaps for now revert to the binary hash + snprintf. Seems less expensive than the Clone method.

victorjulien:
> > While trying to build this I'm getting
> > ```
> > error: no matching package named `digest` found
> > ```
> 
> Strange. Its also a dependency of the sha-1, sha256 and md5 crates? Could it be a local issue as it works fine on CI? In this intel?

Was on my workstation. For some reason it does work now. No idea why.

src/util-lua-common.c
jasonish: No, not without a larger refactor.  The checksum has already been finalized, so the context is gone.  We also have the same issue in the fileinfo logging.

Same argument applies here to the email body md5..  We could keep the context alive by not calling finalize.  Then each time we log it we clone it, then finalize to hex.

Computationally it seems to make more sense to store the hex buffer once and reuse it, like in the email. But that leads to struct growth. The alternative leads to more CPU usage as the checksum is potentially cloned, finalized and formatted multiple times.

victorjulien:
This fully passes my QA, nice work.

I was wondering: have you checked that we have tests for each type of user?

jasonish:
> I was wondering: have you checked that we have tests for each type of user?

Yeah, for each conversion I made sure some test failed on a bad hash.  I only caught 2 cases that we didn't cover:
- community id
- the smtp issue where the hash turned out to be wrong

src/util-decode-mime.h
jasonish: Ok, this has been reverted back to the snprintf loop. The relevant commit only changes the subject hashing now.

src/detect-transform-md5.c
jasonish: Is this an issue? I would expect people who do disable hashing to be in the small minority, and probably running more custom setups, so tweaking their input rules is probably a non-issue.

jasonish:
Latest version at https://github.com/OISF/suricata/pull/5709.

PR #5669, opened 2020-12-17
Http2 keywords 4067 v6

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4067

Describe changes:
- Adds signature protocol extending flow protocol (one signature protocol matches multiple flow protocols cf `AppProtoEquals`)
- Adds `ALPROTO_HTTP_ANY` (as such a signature protocol) for signatures to match on both HTTP1 and HTTP2
- Rename `ALPROTO_HTTP` to `ALPROTO_HTTP1` (and `ALPROTO_HTTP` is now for any version)
- `http.uri` keyword now matches on HTTP2 traffic (as well as `http.uri.raw`)
- `http.user_agent` keyword now matches on HTTP2 traffic
- `http.stat_msg` keyword now matches on HTTP2 traffic

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

suricata-verify-pr: 384

Makes S-V test in https://github.com/OISF/suricata-verify/pull/384 pass

What remains to be done :
- rename `alproto` to `sig_alproto`. Thoughts about a better name ?
- Commits squashing  (commit-check)
- CI (commit-check)
- Do all the keywords

Modifies #5668 by 
- running `git rebase master -x "git clang-format HEAD^"`
- adding `http.stat_msg` commit

PR #5697, opened 2021-01-03
ikev1: add ikev1 parser - v13

satta:
Submitting a new PR to keep code in sync with build breaking master changes and address some remarks. Just to get things moving again.

Previous PR: https://github.com/OISF/suricata/pull/5552

Describe changes:
- Change all remaining use of inspect v1 API to v2 (`DetectAppLayerInspectEngineRegister2`) to make code build with current master again
- Move addition of `EveIKEAddMetadata` into separate commit
- Adjust dates in `src/output-json-ike.h` comment
- Undo rename of variable in `src/output-json-ike.c` as requested.

suricata-verify-pr: 351


satta:
Ah,. I just noticed that I likely did not use the correct syntax for `suricata-verify-pr` to be used automatically. Will force-push to re-trigger actions.

satta:
CI completing now. Looking forward to your comments!

catenacyber:
Thanks @satta
About the protocol alias commit, what do you think about the first three commits of https://github.com/OISF/suricata/pull/5669 ?

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -1811,3 +1811,118 @@ Example of HTTP2 logging, of a request and response:
...
+      "vendor_ids": [
+        "4048b7d56ebce88525e7de7f00d6c2d3",
+        "4a131c81070358455c5728f20e95452f",
+        "afcad71368a1f1c96b8696fc77570100",
+        "7d9419a65310ca6f2c179d9215529d56",
+        "cd60464335df21f87cfdb2fc68b6a448",
+        "90cb80913ebb696e086381b5ec427b1f"
+      ]
+    },
+    "ikev2": {
```
catenacyber, line 1920: How come there is ikev2 if we have `"version_major": 1` above ?

src/output-json-ike.h
```diff
@@ -1,4 +1,4 @@
-/* Copyright (C) 2015 Open Information Security Foundation
+/* Copyright (C) 2015-2020 Open Information Security Foundation
```
catenacyber, line 1: Do we need to change this file ?

doc/userguide/output/eve/eve-json-format.rst
satta: Could be an old example in this documentation. Will double-check.

src/output-json-ike.c
```diff
@@ -107,18 +105,24 @@ static OutputInitResult OutputIKELogInitSub(ConfNode *conf, OutputCtx *parent_ct
         return result;
     }
     ikelog_ctx->file_ctx = ajt->file_ctx;
-    ikelog_ctx->cfg = ajt->cfg;
```
catenacyber: Why do we remove this line ?

src/output-json-ike.c
```diff
@@ -63,21 +66,16 @@ typedef struct LogIKELogThread_ {
...
 {
-    IKETransaction *iketx = tx;
     LogIKELogThread *thread = thread_data;
-
     JsonBuilder *jb = CreateEveHeader((Packet *)p, LOG_DIR_PACKET, "ike", NULL);
     if (unlikely(jb == NULL)) {
         return TM_ECODE_FAILED;
     }
 
-    EveAddCommonOptions(&thread->ikelog_ctx->cfg, p, f, jb);
```
catenacyber: Why do we remove this line ?

src/output-json-ike.c
```diff
@@ -46,12 +47,14 @@
 
 #include "app-layer-ike.h"
 #include "output-json-ike.h"
-
```
catenacyber, line 49: nit

satta:
> About the protocol alias commit, what do you think about the first three commits of #5669 ?

The idea is to decouple signature protocols specified in rules from parser protocols, allowing one rule protocol to match more than one parser protocol? That would also work I guess. Would need some extra digging to adapt this here.

src/detect-ike-chosen-sa.c
```diff
@@ -0,0 +1,283 @@
...
+
+#include "rust-bindings.h"
+
+/**
+ *   [ike.chosen_sa_attribute]:<sa_attribute>=<type>;
+ */
+
+// support the basic attributes, which are parsed as integer and life_duration, if variable length
+// is 4 it is stored as integer too
+#define PARSE_REGEX                                                                                \
```
catenacyber, line 42: I would love to have rust parsing code instead of regex... But it is not preventing the merge for me

rust/src/ike/logger.rs
```diff
@@ -0,0 +1,287 @@
...
+        jb.close()?; // vendor_ids
+    }
+    jb.close()?;
+
+    return Ok(());
+}
+
+fn append_payload_type_extended(js: &mut JsonBuilder, pt: &u8) -> Result<(), JsonError> {
+    match *pt {
+        1 => js.append_string(IsakmpPayloadType::SecurityAssociation.to_string().as_str())?,
```
catenacyber, line 238: Is there a more idiomatic rust way to do this ?

src/output-json-ike.h
satta: Probably not in this commit. Frank just added himself as additional author, but in fact only removed the 'v2' suffix from the functions in the previous commit. I can simply remove that if you want.

rust/src/ike/logger.rs
```diff
@@ -0,0 +1,287 @@
...
+    jb.set_uint("message_id", tx.hdr.msg_id as u64)?;
+
+    if tx.ike_version == 1 {
+        if let Some(exchange_type) = tx.hdr.ikev1_header.exchange_type {
+            jb.set_uint("exchange_type", exchange_type as u64)?;
+
+            if (flags & LOG_EXTENDED) == LOG_EXTENDED {
+                match exchange_type {
+                    1 => jb.set_string(
+                        "exchange_type_verbose",
```
catenacyber, line 68: More rust idiomatic way to do this to avoid copy/pasting `exchange_type_verbose` ?

catenacyber (approved):
For me, this is good for merge even if there is always room for improvement.
Maybe someone else can take a look as well ?

src/output-json-ike.h
catenacyber: Yes, let's minimize the changes

src/output-json-ike.c
satta: Probably because the `cfg` field was also removed from the `LogIKEFileCtx`, see previous PR. 
See https://github.com/OISF/suricata/pull/5552/commits/7c99aa30cfee4cbc94a542703e8ddaea65d7f481#diff-a67ed7262d3c2ec79597649dc8dbf9aaab3f86b51f06de1b790bff4f791f46d1L54

catenacyber:
>> About the protocol alias commit, what do you think about the first three commits of #5669 ?
>The idea is to decouple signature protocols specified in rules from parser protocols, allowing one rule protocol to match more than one parser protocol? That would also work I guess. Would need some extra digging to adapt this here.

I am not sure it is relevant here.
The question is what is the purpose of adding protocol aliases ?

Is it to support both signatures with `alert ike` and `alert ikev2` right ?
Because existant signatures use `alert ikev2`.
And we want (some of) these old signatures to trigger on ikev1 traffic, right ?
If so maybe we can just add a line to `StringToAppProto`...

src/output-json-ike.c
catenacyber: Makes sense

src/output-json-ike.c
satta: See above, `cfg` was removed. I am not sure why it was removed, though. If there is nothing in the way of reinstating it, I would be happy to do so.

src/output-json-ike.c
satta: An you want that newline _put back in_. OK.

catenacyber: Yes, just minimize the changes :-)

satta: Done

src/output-json-ike.h
satta: Done

satta:
> > > About the protocol alias commit, what do you think about the first three commits of #5669 ?
> > > The idea is to decouple signature protocols specified in rules from parser protocols, allowing one rule protocol to match more than one parser protocol? That would also work I guess. Would need some extra digging to adapt this here.
> 
> I am not sure it is relevant here.
> The question is what is the purpose of adding protocol aliases ?
> 
> Is it to support both signatures with `alert ike` and `alert ikev2` right ?
> Because existant signatures use `alert ikev2`.

I think this is the idea. The commit adds a new mapping between protocol string (e.g. `ike`) and alias (e.g. `ikev2`, see https://github.com/OISF/suricata/pull/5697/files#diff-8e5a3a135d258caaf98cf08bf0c585d811c898384a91b9ce96bcdf491bc2c383R576), then modifies `AppLayerProtoDetectGetProtoByName()` to also consider the mapped protocol string when its alias is given (i.e. return the protocol number for `ike` when `ikev2` is passed). I wonder if that can be done using your approach as well.

> And we want (some of) these old signatures to trigger on ikev1 traffic, right ?

I think so. AFAICS from looking at the code, the old proto `ikev2` should simply mean `ike` now, there is no more distinction on the keyword level. Correct, @fhonza?

> If so maybe we can just add a line to `StringToAppProto`...

Can you give an example, please?

doc/userguide/output/eve/eve-json-format.rst
satta: Changed.

rust/src/ike/logger.rs
jasonish: It looks like `IsakmpPayloadType` implements `FromPrimitive`, so it has a `from_u8` method that returns on option. That would make this method much smaller.

rust/src/ike/logger.rs
satta: I guess, one could only determine the value string in the `match` and then call `jb.set_string()` only once afterwards. Will do that in the next PR.

jasonish: Yeah, probably similar to above if `ExchangeType` has the `FromPrimitive` derive added to it as well.

rust/src/ike/logger.rs
satta: Done. Much better:
```rust
fn append_payload_type_extended(js: &mut JsonBuilder, pt: &u8) -> Result<(), JsonError> {
    if let Some(v) = IsakmpPayloadType::from_u8(*pt) {
        js.append_string(&format!("{:?}", v))?;
    }
    Ok(())
}
```

satta:
Next PR: #5711

rust/src/ike/logger.rs
satta: Done as well. Thanks for the hint!

fhonza:
> I think so. AFAICS from looking at the code, the old proto `ikev2` should simply mean `ike` now, there is no more distinction between versions on the keyword level. Correct, @fhonza?

yes the alias is just to make the already existing ikev2-signatures to work with the new `ike` proto. There is no distinction in the keywords, most of them will just work with both versions.

catenacyber:
So, I think the alias commit can be replaced by having in `StringToAppProto` these 2 lines :
```
    if (strcmp(proto_name,"ikev2")==0) return ALPROTO_IKE;
    if (strcmp(proto_name,"ike")==0) return ALPROTO_IKE;
```

fhonza:
> So, I think the alias commit can be replaced by having in `StringToAppProto` these 2 lines :
> 
> ```
>     if (strcmp(proto_name,"ikev2")==0) return ALPROTO_IKE;
>     if (strcmp(proto_name,"ike")==0) return ALPROTO_IKE;
> ```

This was my first idea too but that doesn't work. You are getting to that piece of logic when you are registering the
parser in `rs_ike_register_parser` from rust.
After that it's added in `app-layer-detect-proto.c` to some data structure: `alpd_ctx.alproto_names[alproto] = alproto_name`

There you can have for each protocol only one "name", this is why i've added the aliases.

Perhaps it's possible to use a complete different data structure there, but i feared a larger rewrite then. The alias thing was
imho "least invasive"

catenacyber:
@fhonza you can check the changes I made to `AppLayerProtoDetectGetProtoByName` in #5669

With this, the simple change to `StringToAppProto` should work

PR #5700, opened 2021-01-04
pgsql: initial support

jufajardini:
- pgsql/parser: add startup message structure, initial nom parser and
corresponding unit test
- pgsql/mod: list pgsql modules that must be compiled
- lib: add pgsql to the module's list

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4241

Describe changes:
- pgsql/parser: add startup message structure, initial nom parser and
corresponding unit test
- pgsql/mod: list pgsql modules that must be compiled
- lib: add pgsql to the module's list

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5706, opened 2021-01-06
PostgreSQL: initial support

jufajardini:
- pgsql/parser: add startup message structure, initial nom parser and
corresponding unit tests
- pgsql/mod: list pgsql modules that must be compiled
- lib: add pgsql to the module's list

Link to previous PR:
https://github.com/OISF/suricata/pull/5700

Describe changes:
- add constants to represent the dummy protocol version used for the SSL handshake
- add tests for invalid input (try to make them off by one, where possible)
- adjust readability in some places
- get rid of unused imports

Regarding the next steps, I was wondering if I should start following the approach Jason mentioned, working on the outer limit that will check length etc, since that would affect slightly the nom parsers logic (right?). Or do I write such nom parsers now (like a parser for partial record), assuming that future outer logic and flow control, that will be added soon?

There are still some failed checks with fedora 32 and 33 (debug, clang, asan, wshadow, rust-strict). I've looked into them, but from what I understand they mention unused code that is actually being used (except for the two new constants that I have just added), so I don't get why they're giving trouble...

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/rust/src/pgsql/parser.rs
+++ b/rust/src/pgsql/parser.rs
@@ -1,0 +1,191 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+pub const PGSQL_LENGTH_SIZE: usize = 4;
+pub const PGSQL_PROTO_SIZE: usize = 4;
+
+// dummy code used for the frontend to signal an SSL handshake (04 d2 16 2f)
+pub const PGSQL_SSL_DUMMY_PROTO_MAJOR: u16 = 1234;
+pub const PGSQL_SSL_DUMMY_PROTO_MINOR: u16 = 5679;
+
+// PostgreSQL Startup message packet
+#[derive(Debug, PartialEq)]
+pub struct PGSQLStartupPacket<'a> {
+    pub length: u32,
+    pub proto_version_major: u16,
+    pub proto_version_minor: u16,
+    pub data: &'a[u8],
+}
+
+impl<'a> PGSQLStartupPacket<'a> {
+    
+    pub fn is_valid(&self) -> bool {
+        // currently, the protocol support only three major versions
+        let valid_proto: bool = match self.proto_version_major {
+            1 | 2 | 3 => true,
+            _ => false,
+        };
+        valid_proto
+    }
+}
+
+named!(pub parse_pgsql_startup_packet<PGSQLStartupPacket>,
+    do_parse!(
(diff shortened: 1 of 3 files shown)


rust/src/pgsql/parser.rs
```diff
@@ -0,0 +1,191 @@
...
+            Err(nom::Err::Incomplete(needed)) => {
+                panic!("Should be Ok, but needs {:?}", needed);
+            }
+            _ => {
+                panic!("Unexpected behavior!");
+            }
+        }
+
+        // A non-valid Startup message. Data length < length field indicates
+        let invalid_data_length_buff:&[u8] = &[
```
victorjulien, line 167: can you also add a test like this with a length field that is < 8? I suspect this will underflow the parser and trigger a panic.

rust/src/pgsql/parser.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+}
+
+named!(pub parse_pgsql_startup_packet<PGSQLStartupPacket>,
+    do_parse!(
+        len: bits!(take_bits!(32u32))
+        >> proto_version: bits!(tuple!(
+            take_bits!(16u16),
+            take_bits!(16u16)))
+        >> data: take!(len as usize - (PGSQL_LENGTH_SIZE + PGSQL_PROTO_SIZE))
```
victorjulien, line 52: if len < 8 this would underflow, right?

rust/src/pgsql/parser.rs
```diff
@@ -0,0 +1,191 @@
...
+            1 | 2 | 3 => true,
+            _ => false,
+        };
+        valid_proto
+    }
+}
+
+named!(pub parse_pgsql_startup_packet<PGSQLStartupPacket>,
+    do_parse!(
+        len: bits!(take_bits!(32u32))
```
victorjulien, line 48: why is this using `bits!(takebits!(` ? Is it not possible to use `le_u32` or `be_u32`?

rust/src/pgsql/parser.rs
```diff
@@ -0,0 +1,191 @@
+/* Copyright (C) 2020 Open Information Security Foundation
```
victorjulien, line 1: nit: 2021

rust/src/pgsql/parser.rs
victorjulien: The difference with the test at line 111 would be that there the nom parser exists with `Incomplete` before it hits that code.

rust/src/pgsql/parser.rs
jufajardini: Yes. I must throw an error here, right? Or I could check len first, then through the error. I've seen in other places (like SMB, if I recall correctly) that this length is also checked before calling the parser. Even so, would the safest approach be to also check it here?

rust/src/pgsql/parser.rs
jufajardini: The part in lines 122 to 144 weas supposed to do that. But I wasn't sure how to handle the result, so I added an                 `assert!(invalid_length_field_buff.len() < (PGSQL_PROTO_SIZE + PGSQL_LENGTH_SIZE));` to the Incomplete arm of the match, for that's where my test goes...

rust/src/pgsql/parser.rs
jufajardini: Thanks, to be very honest I tried using take_bits! and got an error, looked for other pieces of code and saw that structure. The 32u32 bothered me, but I didn't really know I could have just used len: be_u32 instead. I'm still quite new to Rust...

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/5721

PR #5699, opened 2021-01-04
hashing: use Rust to provide sha256/sha1/md5 hashing funcionality - v5

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5695

Changes from last PR:
- Revert modification to MimeDecParseState to avoid struct grown, leave snprintf loop in place.
- Remove remaining libnss references.

suricata-verify-pr: 393

PR #5709, opened 2021-01-06
hashing: use Rust to provide sha256/sha1/md5 hashing funcionality - v6

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5699

Changes from last PR:
- Fix crash in md5 after revert (https://github.com/OISF/suricata/pull/5699#issuecomment-754197340)
- Add `--disable-hashing` as a command line option.

Still todo:
- More graceful handling when `--disable-hashing` is used. Still need clarification.

suricata-verify-pr: 393


src/util-decode-mime.c
```diff
@@ -2269,17 +2264,16 @@ static int ProcessMimeBody(const uint8_t *buf, uint32_t len,
...
-    if (MimeDecGetConfig()->body_md5) {
-        if (state->body_begin == 1) {
-            if (state->md5_ctx == NULL) {
-                state->md5_ctx = HASH_Create(HASH_AlgMD5);
-                HASH_Begin(state->md5_ctx);
+    if (!g_disable_hashing) {
+        if (MimeDecGetConfig()->body_md5) {
+            if (state->body_begin == 1) {
+                if (state->md5_ctx == NULL) {
+                    state->md5_ctx = SCMd5New();
```
victorjulien, line 2271: can this fail?

src/util-decode-mime.h
```diff
@@ -197,10 +197,9 @@ typedef struct MimeDecParseState {
     uint8_t bvremain[B64_BLOCK];  /**< Remainder from base64-decoded line */
     uint8_t bvr_len;  /**< Length of remainder from base64-decoded line */
     uint8_t data_chunk[DATA_CHUNK_SIZE];  /**< Buffer holding data chunk */
-#ifdef HAVE_NSS
-    HASHContext *md5_ctx;
-    uint8_t md5[MD5_LENGTH];
-#endif
+    SCMd5 *md5_ctx;
```
victorjulien, line 200: Is this an opague type? Wonder if we could not make it a pointer so prevent an allocation we'd do every time?

src/util-decode-mime.c
jasonish: No, not in a way we can test anyways.  Much like any pointer that is created in Rust and passed back to see.

The same actually applies to the jsonbuilder stuff.  Just looking now, we do check for null with the jsonbuilder stuff, but AFAIK, it will never actually return a NULL, as the Box::into_raw we use never will return NULL.  We could add the NULL check, but its just non-functional sugar, but I admit, it does look better.

src/util-decode-mime.h
jasonish: Its an opaque type, but its a pointer returned from a rust constructor.

src/suricata.c
```diff
@@ -1177,68 +1177,50 @@ static TmEcode ParseCommandLine(int argc, char** argv, SCInstance *suri)
...
-        {"pcap-file-delete", 0, 0, 0},
-        {"pcap-file-recursive", 0, 0, 0},
-        {"simulate-ips", 0, 0 , 0},
-        {"no-random", 0, &g_disable_randomness, 1},
-        {"strict-rule-keywords", optional_argument, 0, 0},
-
-        {"capture-plugin", required_argument, 0, 0},
-        {"capture-plugin-args", required_argument, 0, 0},
+    struct option long_opts[] = { { "dump-config", 0, &dump_config, 1 },
+        { "dump-features", 0, &dump_features, 1 }, { "pfring", optional_argument, 0, 0 },
```
victorjulien, line 1181: I feel this is not an improvement, at all. Wonder if we can make the `.clang-format` file to not do this... @roligugus any ideas? Main complaint is not having each option on its own line.

src/suricata.c
```diff
@@ -233,6 +233,10 @@ int g_disable_randomness = 1;
   * comparing flows */
 uint16_t g_vlan_mask = 0xffff;
 
+/* flag to disable hashing almost globally, to be similar to disabling nss
+ * support */
+bool g_disable_hashing = false;
```
victorjulien, line 233: I feel the disable hashing logic should be in its own commit, not as part of a transform update.

.travis.yml
```diff
@@ -35,7 +35,6 @@ addons:
     # Define the required packages without libjansson so we can do a
```
victorjulien, line 35: Merge this with `github-ci: don't install libnss/libnspr packages`?

src/util-file.c
```diff
@@ -863,14 +883,16 @@ static File *FileOpenFile(FileContainer *ffc, const StreamingBufferConfig *sbcfg
...
 
-    if (!(ff->flags & FILE_NOMD5) || g_file_force_md5) {
-        ff->md5_ctx = SCMd5New();
-    }
-    if (!(ff->flags & FILE_NOSHA1) || g_file_force_sha1) {
-        ff->sha1_ctx = SCSha1New();
-    }
-    if (!(ff->flags & FILE_NOSHA256) || g_file_force_sha256) {
-        ff->sha256_ctx = SCSha256New();
+    if (!g_disable_hashing) {
```
victorjulien, line 886: Did you see this be necessary or is this more to be on the safe side?

src/suricata.c
jasonish: I couldn't find any options to fix this in the clang-format docs, so for now I've added `// clang-format off/on` around this section. I'm still reformatting the `if else if` stuff as it is being reformatted for the better.

src/util-file.c
jasonish: In master this section is wrapped in an `ifdef HAVE_NSS` guard, so it makes sense to wrap this in a `g_disable_hashing` guard.

.travis.yml
jasonish: We've typically never combined github-ci and travis-ci changes into the same commit.

src/util-file.c
victorjulien: My expectation would that this is redundant. I would exist the `FILE_NO<hash>` flags to be set and the `g_file_force_<hash>` to be unset. I think I've confirmed the latter, but not yet sure about the former.

jasonish: Yes, it appears to be redundant in my testing.

jasonish:
Next revision: https://github.com/OISF/suricata/pull/5709

PR #5711, opened 2021-01-07
ikev1: add ikev1 parser - v14

satta:
Previous PR: https://github.com/OISF/suricata/pull/5697

Describe changes:
- Minimize unnecessary minor changes to the code (whitespace, author additions for minor changes).
- Make Rust code more idiomatic by removing large `match` case lists and replacing them with `FromPrimitive` enum casting plus the existing display string definitions.
- Improve documentation wording and example EVE output.

suricata-verify-pr: 351


catenacyber (approved):
Looks good enough to me.

The CI failure seems unrelated : Rust `sys`crate change...

As an improvement, 
I think the alias commit can be replaced by having in StringToAppProto these 2 lines :
```
    if (strcmp(proto_name,"ikev2")==0) return ALPROTO_IKE;
    if (strcmp(proto_name,"ike")==0) return ALPROTO_IKE;
```

fhonza:
> Looks good enough to me.
> 
> The CI failure seems unrelated : Rust `sys`crate change...
> 
> As an improvement,
> I think the alias commit can be replaced by having in StringToAppProto these 2 lines :
> 
> ```
>     if (strcmp(proto_name,"ikev2")==0) return ALPROTO_IKE;
>     if (strcmp(proto_name,"ike")==0) return ALPROTO_IKE;
> ```

This was my first idea too but that doesn't work. You are getting to that piece of logic when you are registering the
parser in `rs_ike_register_parser` from rust.
After that it's added in `app-layer-detect-proto.c` to some data structure: `alpd_ctx.alproto_names[alproto] = alproto_name`

There you can have for each protocol only one "name", this is why i've added the aliases.

Perhaps it's possible to use a complete different data structure there, but i feared a larger rewrite then. The alias thing was
imho "least invasive"

catenacyber:
You can check the changes I made to AppLayerProtoDetectGetProtoByName in #5669
(you can replace `AppProtoEquals(b, a)` with `b == a`)

With this, the simple change to StringToAppProto should work

satta:
We can just change this in a separate PR once your PR gets merged, right?

catenacyber:
> We can just change this in a separate PR once your PR gets merged, right?

Of course.
But I do not know which PR will get merged first.
And this was also a request for comments on my commit ;-) Do you think it is the good way to go ?

satta:
> > We can just change this in a separate PR once your PR gets merged, right?
> 
> Of course.
> But I do not know which PR will get merged first.

True, but either way we can clean it up later. Difficult when PRs have dependencies but I'd be happy to propose an adaptation once your changes get merged.

> And this was also a request for comments on my commit ;-) Do you think it is the good way to go ?

I think they make sense -- providing an additional level of indirection beyond simple string comparison. Just to check: The change in https://github.com/OISF/suricata/pull/5669/commits/f85d079113bf2ce71b74202bd15cf0e93f3d9078#diff-828364efb39780c8edef35141add91767c48a8774c2935f38a75c8eda4e33c25R2105 is the key one that makes this work, right? Because it gets `StringToAppProto()` in the loop where it wasn't before and led to Frank having to to the alias workaround?
Sorry but I don't know if we're missing anything without digging super deep into the call chain there in all cases...

fhonza:
> And this was also a request for comments on my commit ;-) Do you think it is the good way to go ?

Thanks this should solve the problem too! I like your solution, it's shorter and imho comparing the enum-values in the loop is nicer.

Didn't debug just looked at the code but pretty confident that this should work ;)

catenacyber:
>  Because it gets StringToAppProto() in the loop where it wasn't before and led to Frank having to to the alias workaround?

Exactly

> Didn't debug just looked at the code but pretty confident that this should work ;)

@fhonza could you try it in a new PR ?

fhonza:
> @fhonza could you try it in a new PR ?

Can you clarify?

Your modification to AppLayerProtoDetectGetProtoByName() with the StringToAppProto is not yet merged.
There is only required that "ikev2" and "ike" map both to proto ike, i can try it locally here if it would work with your modified logic.

I would suggest to handle it like written above by @satta depending which PR gets merged first....

satta:
Yeah. One could cherry-pick these commits and then prepare a new PR including them, but I guess that would make things complicated to merge if the other PR gets merged first, also containing these commits.
I'd say let's keep this in as it is with no overlap, and once both are merged we clean this one up and adapt it to the other style. What do you say?

catenacyber:
I say this PR looks good to me :-) and we can optimize later

norg:
Those failures in the builds don't seem to be related to the PR itself, right?

satta:
I don't think so, at least they do not seem to be related to Rust things changed in this PR. Maybe someone could restart the CI runs?

catenacyber:
> I don't think so, at least they do not seem to be related to Rust things changed in this PR. Maybe someone could restart the CI runs?

This needs a rebase to fix the conflicts (and S-V rebase as well to get the tests to pass)

satta:
Ah I see, will do this later today.

satta:
Next PR: https://github.com/OISF/suricata/pull/5941

PR #5642, opened 2020-12-08
Http2 files v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4116

Describe changes:
- Files decompression for HTTP2

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

suricata-verify-pr: 375

Makes https://github.com/OISF/suricata-verify/pull/375 pass

Modifies #5640 with comment about why need to implement `Read` for `HTTP2cursor`

PR #5713, opened 2021-01-08
Http2 files v4

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4116

Describe changes:
- Files decompression for HTTP2

suricata-verify-pr: 397

Makes https://github.com/OISF/suricata-verify/pull/397 pass (includes a framework change to have feature selection by check and not only by whole test)

Modifies #5642 with
- Making HTTP2 decompression a configure/compile-time option/feature
- Doing so, putting the decompression logic in a specific file

There is probably some commits squashing to be done.
Do you want a single commit ? Or one adding the decompression, and one making it optional ?

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -641,7 +641,7 @@ jobs:
           chmod 755 $HOME/.cargo/bin/cbindgen
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - run: ./autogen.sh
-      - run: CFLAGS="$DEFAULT_CFLAGS -DNDEBUG" ./configure --enable-unittests
+      - run: CFLAGS="$DEFAULT_CFLAGS -DNDEBUG" ./configure --enable-unittests --enable-http2-decompression
       - run: make -j2
       - run: make check
       - run: make dist
--- a/rust/src/http2/decompression.rs
+++ b/rust/src/http2/decompression.rs
@@ -1,0 +1,216 @@
+/* Copyright (C) 2021 Open Information Security Foundation
+*
+* You can copy, redistribute or modify this Program under the terms of
+* the GNU General Public License version 2 as published by the Free
+* Software Foundation.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* version 2 along with this program; if not, write to the Free Software
+* Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+* 02110-1301, USA.
+*/
+
+use crate::core::STREAM_TOCLIENT;
+use brotli;
+use flate2::read::GzDecoder;
+use std;
+use std::io;
+use std::io::{Cursor, Read, Write};
+
+pub const HTTP2_DECOMPRESSION_CHUNK_SIZE: usize = 0x1000; // 4096
+
+#[repr(u8)]
+#[derive(Copy, Clone, PartialOrd, PartialEq, Debug)]
+pub enum HTTP2ContentEncoding {
+    HTTP2ContentEncodingUnknown = 0,
+    HTTP2ContentEncodingGzip = 1,
+    HTTP2ContentEncodingBr = 2,
+    HTTP2ContentEncodingUnrecognized = 3,
+}
+
+//a cursor turning EOF into blocking errors
+pub struct HTTP2cursor {
+    pub cursor: Cursor<Vec<u8>>,
+}
+
+impl HTTP2cursor {
+    pub fn new() -> HTTP2cursor {
+        HTTP2cursor {
+            cursor: Cursor::new(Vec::new()),
+        }
+    }
+
+    #[cfg(feature = "debug-validate")]
+    pub fn position(&self) -> u64 {
+        return self.cursor.position();
+    }
+
+    pub fn set_position(&mut self, pos: u64) {
+        return self.cursor.set_position(pos);
+    }
+}
+
+// we need to implement this as flate2 and brotli crates
+// will read from this object
+impl Read for HTTP2cursor {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        //use the cursor, except it turns eof into blocking error
+        let r = self.cursor.read(buf);
+        match r {
+            Err(ref err) => {
+                if err.kind() == io::ErrorKind::UnexpectedEof {
+                    return Err(io::ErrorKind::WouldBlock.into());
+                }
+            }
+            Ok(0) => {
+                //regular EOF turned into blocking error
+                return Err(io::ErrorKind::WouldBlock.into());
+            }
+            Ok(_n) => {}
+        }
+        return r;
+    }
+}
+
+pub enum HTTP2Decompresser {
+    UNASSIGNED,
+    GZIP(GzDecoder<HTTP2cursor>),
+    BROTLI(brotli::Decompressor<HTTP2cursor>),
+}
+
+struct HTTP2DecoderHalf {
+    encoding: HTTP2ContentEncoding,
+    decoder: HTTP2Decompresser,
+}
+
+pub trait GetMutCursor {
+    fn get_mut(&mut self) -> &mut HTTP2cursor;
+}
+
+impl GetMutCursor for GzDecoder<HTTP2cursor> {
+    fn get_mut(&mut self) -> &mut HTTP2cursor {
+        return self.get_mut();
+    }
+}
+
+impl GetMutCursor for brotli::Decompressor<HTTP2cursor> {
+    fn get_mut(&mut self) -> &mut HTTP2cursor {
+        return self.get_mut();
+    }
+}
+
+fn http2_decompress<'a>(
+    decoder: &mut (impl Read + GetMutCursor), input: &'a [u8], output: &'a mut Vec<u8>,
+) -> io::Result<&'a [u8]> {
+    match decoder.get_mut().cursor.write_all(input) {
+        Ok(()) => {}
+        Err(e) => {
+            return Err(e);
+        }
+    }
+    let mut offset = 0;
+    decoder.get_mut().set_position(0);
+    output.resize(HTTP2_DECOMPRESSION_CHUNK_SIZE, 0);
+    loop {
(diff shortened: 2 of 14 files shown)


.github/workflows/builds.yml
```diff
@@ -641,7 +641,7 @@ jobs:
           chmod 755 $HOME/.cargo/bin/cbindgen
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - run: ./autogen.sh
-      - run: CFLAGS="$DEFAULT_CFLAGS -DNDEBUG" ./configure --enable-unittests
+      - run: CFLAGS="$DEFAULT_CFLAGS -DNDEBUG" ./configure --enable-unittests --enable-http2-decompression
```
catenacyber, line 644: Just picked a random build with make check and S-V

rust/src/http2/http2.rs
```diff
@@ -160,6 +168,46 @@ impl HTTP2Transaction {
...
+                self.decoder.http2_encoding_fromvec(&blocks[i].value, dir);
+            }
+        }
+    }
+
+    // dir may be unused without the decompression feature
+    fn decompress<'a>(
+        &'a mut self, input: &'a [u8], _dir: u8, sfcm: &'static SuricataFileContext, over: bool,
+        files: &mut FileContainer, flags: u16,
+    ) -> io::Result<()> {
```
catenacyber, line 187: Spent a lot of time in the following lines
`_dir` looks a bit ugly 
We need to define output here so that its lifetime can be used in `self.ft.new_chunk ` the same as `input`

src/suricata.c
```diff
@@ -702,6 +702,9 @@ static void PrintBuildInfo(void)
 #ifdef HAVE_NSS
     strlcat(features, "HAVE_NSS ", sizeof(features));
 #endif
+#ifdef HTTP2_DECOMPRESSION
+    strlcat(features, "HTTP2_DECOMPRESSION ", sizeof(features));
+#endif
```
catenacyber, line 707: Are there other places I need to check ?

catenacyber:
CI fails due to Rust `syn`:-(

rust/src/http2/http2.rs
```diff
@@ -160,6 +168,46 @@ impl HTTP2Transaction {
...
+        #[cfg(feature = "decompression")]
+        let mut output = Vec::with_capacity(decompression::HTTP2_DECOMPRESSION_CHUNK_SIZE);
+        #[cfg(feature = "decompression")]
+        let decompressed = self.decoder.decompress(input, &mut output, _dir)?;
+        #[cfg(not(feature = "decompression"))]
```
victorjulien, line 192: instead of these conditions I think its also possible to define a method twice. One would support the decoder, other would just pass through data?

SMB has an example
```rust
#[cfg(not(feature = "debug"))]
fn debug_add_progress(_js: &mut JsonBuilder, _tx: &SMBTransaction) -> Result<(), JsonError> { Ok(()) }

#[cfg(feature = "debug")]
fn debug_add_progress(jsb: &mut JsonBuilder, tx: &SMBTransaction) -> Result<(), JsonError> {
    jsb.set_bool("request_done", tx.request_done)?;
    jsb.set_bool("response_done", tx.response_done)?;
    Ok(())
}
```

rust/src/http2/http2.rs
jasonish: `_dir` is fine by me...

rust/src/http2/http2.rs
jasonish: Agreed.

jasonish (commented):
So for decompress we first copy the data into a vector backed cursor. And we do this to get a WouldBlock type error instead of an EOF error.  Do the decompressors do something different on these error types that we can't just use the EOF error as well?

I think I'd like to see some unit tests around short buffers and such.

rust/src/http2/http2.rs
catenacyber: > instead of these conditions I think its also possible to define a method twice. One would support the decoder, other would just pass through data?

Indeed, it is possible to define a method twice as is done for `handle_headers`

But I do not see how it can be implemented in this case because of `output` lifetime which needs to reach `self.ft.new_chunk`

Ie this patch fails compilation with decompression enabled with 
```
error[E0515]: cannot return value referencing local variable `output`
   --> src/http2/http2.rs:191:16
    |
191 |         return self.decoder.decompress(input, &mut output, _dir);
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-----------^^^^^^^
    |                |                              |
    |                |                              `output` is borrowed here
    |                returns a value referencing data owned by the current function


```

rust/src/http2/decompression.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl HTTP2DecoderHalf {
+    pub fn new() -> HTTP2DecoderHalf {
+        HTTP2DecoderHalf {
+            encoding: HTTP2ContentEncoding::HTTP2ContentEncodingUnknown,
+            decoder: HTTP2Decompresser::UNASSIGNED,
+        }
+    }
+
+    pub fn http2_encoding_fromvec(&mut self, input: &Vec<u8>) {
```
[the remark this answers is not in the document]
catenacyber: ok

rust/src/http2/decompression.rs
```diff
@@ -0,0 +1,216 @@
...
+        //use first encoding...
+        if self.encoding == HTTP2ContentEncoding::HTTP2ContentEncodingUnknown {
+            if *input == "gzip".as_bytes().to_vec() {
```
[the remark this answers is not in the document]
catenacyber: Wow !

rust/src/http2/decompression.rs
```diff
@@ -0,0 +1,216 @@
...
+                self.encoding = HTTP2ContentEncoding::HTTP2ContentEncodingGzip;
+                self.decoder = HTTP2Decompresser::GZIP(GzDecoder::new(HTTP2cursor::new()));
+            } else if *input == "br".as_bytes().to_vec() {
```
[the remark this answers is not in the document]
catenacyber: indeed

rust/src/http2/decompression.rs
```diff
@@ -0,0 +1,216 @@
...
+            Ok(_n) => {}
+        }
+        return r;
+    }
+}
+
+pub enum HTTP2Decompresser {
+    UNASSIGNED,
+    GZIP(GzDecoder<HTTP2cursor>),
+    BROTLI(brotli::Decompressor<HTTP2cursor>),
```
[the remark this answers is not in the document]
catenacyber: ok

rust/src/http2/decompression.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl HTTP2Decoder {
+    pub fn new() -> HTTP2Decoder {
+        HTTP2Decoder {
+            decoder_tc: HTTP2DecoderHalf::new(),
+            decoder_ts: HTTP2DecoderHalf::new(),
+        }
+    }
+
+    pub fn http2_encoding_fromvec(&mut self, input: &Vec<u8>, dir: u8) {
```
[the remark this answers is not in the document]
catenacyber: ok

catenacyber:
> So for decompress we first copy the data into a vector backed cursor. And we do this to get a WouldBlock type error instead of an EOF error. Do the decompressors do something different on these error types that we can't just use the EOF error as well?

The decompressors wait for more data later on WouldBlock and fail right now on EOF
cf https://github.com/rust-lang/flate2-rs/blob/2a6dc3b66fe4313da1a58ebf6b4442998a39b841/src/gz/bufread.rs#L443
Do you have another idea ?

jasonish:
> > So for decompress we first copy the data into a vector backed cursor. And we do this to get a WouldBlock type error instead of an EOF error. Do the decompressors do something different on these error types that we can't just use the EOF error as well?
> 
> The decompressors wait for more data later on WouldBlock and fail right now on EOF
> cf https://github.com/rust-lang/flate2-rs/blob/2a6dc3b66fe4313da1a58ebf6b4442998a39b841/src/gz/bufread.rs#L443
> Do you have another idea ?

So EOF is an unrecoverable error as far as the decompressors are concerned?

catenacyber:
> So EOF is an unrecoverable error as far as the decompressors are concerned?

Yes, without turning `Ok(0)` into `Err(io::ErrorKind::WouldBlock.into())`, `decoder.read` in `http2_decompress` returns `Err(Custom { kind: UnexpectedEof, error: "failed to fill whole buffer" })` without returning the partially decompressed data

catenacyber:
Hacky patch to test this on S-V http2-bugfixes 
```
             HTTP2Decompresser::Gzip(ref mut gzip_decoder) => {
+if input.len() > 100 {
+let r = http2_decompress(gzip_decoder, &input[..100], output);
+println!("lol r r2 {:?} {:?}", r, input.len());
+println!("lol r output {:?}", &output);
+let r2 = http2_decompress(gzip_decoder, &input[100..], output);
+println!("lol r r2 {:?}", r2);
+println!("lol r output {:?}", &output);
+}
```

jasonish:
> > So EOF is an unrecoverable error as far as the decompressors are concerned?
> 
> Yes, without turning `Ok(0)` into `Err(io::ErrorKind::WouldBlock.into())`, `decoder.read` in `http2_decompress` returns `Err(Custom { kind: UnexpectedEof, error: "failed to fill whole buffer" })` without returning the partially decompressed data

Can we still retry with a new larger chunk of data? Or is the decoder dead? Not saying we should, just thinking out loud.

catenacyber:
> Can we still retry with a new larger chunk of data? Or is the decoder dead? Not saying we should, just thinking out loud.

We can recreate a new decoder to retry with a larger chunk of data.
But the current decoder is dead and we cannot supply it with new input (when flate2 reaches the error state, it never gets out of it)

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5775

jasonish:
> > Can we still retry with a new larger chunk of data? Or is the decoder dead? Not saying we should, just thinking out loud.
> 
> We can recreate a new decoder to retry with a larger chunk of data.
> But the current decoder is dead and we cannot supply it with new input (when flate2 reaches the error state, it never gets out of it)

Fair enough. Makes sense now to have the Cursor + Write impl.  Would just like to see this commented in the code as to why we haves these wrappers and implementation.

catenacyber:
@jasonish https://github.com/OISF/suricata/pull/5780 got merged.
Do you think it is worth to open a new PR to add some comments ?

PR #5716, opened 2021-01-11
hashing: use Rust to provide sha256/sha1/md5 hashing funcionality - v7

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5709

Changes from last PR:
- Split intoduction of `g_disable_hashing` from `transform/md5` conversion to Rust MD5.
- Merge `g_disable_hashing` commit with addition of command line option.
- Remove redundant use of `g_disable_hashing` in `util-file.c`.
- (misc) Update MacOS and Debian github-ci jobs to use Python 3 (can split into another PR).


victorjulien:
Already cherry-picked some general cleanup and CI improvement commits in #5720.

src/output-json-email-common.c
```diff
@@ -131,15 +131,10 @@ static void EveEmailLogJSONMd5(OutputJsonEmailCtx *email_ctx, JsonBuilder *js, S
         }
         field = MimeDecFindField(entity, "subject");
         if (field != NULL) {
-            unsigned char md5[SC_MD5_LEN];
-            char smd5[256];
+            char smd5[SC_MD5_HEX_LEN + 1];
             char *value = BytesToString((uint8_t *)field->value , field->value_len);
```
victorjulien, line 135: this is not in your change, but it caught my eye: why would we even do `BytesToString` here if we then don't use the value as a string but as an array of chars? Seems we could just pass `field->value` and `field->value_len` directly to `SCMd5HashBufferToHex`

victorjulien: I have an easy patch to improve this, but do we have a test that covers it?

victorjulien: I see SV pr/393 adds one.

victorjulien:
Running this through QA currently. If it passes I'll merge it tomorrow.

victorjulien:
Merged as part of #5722, thanks Jason!

PR #5724, opened 2021-01-14
lua: fix coverity issue with out of scope variable - v1

jasonish:
Fix usage of out-of-scope variables. Introduced with the hashing
and adding the guard of g_disable_hashing.

To fix, just remove the guard so all variables are in scope. Hashes
are not initialized here so there is no need for the guard.

Coverity CID: 1471768


victorjulien (approved):

victorjulien:
Merged in #5727, thanks!

PR #5749, opened 2021-01-19
Cleanup/v16

victorjulien:
Also test otherwise unrelated SV pr 410

suricata-verify-pr: 410

Try if we can get more of the tests files into codecov.


victorjulien:
Replaced by #5751

PR #5761, opened 2021-01-21
fuzz: run OSS-Fuzz corpus and track coverage

victorjulien:


catenacyber:
> Our goal is not to find crashes, but to max out coverage

The goal is also to find bugs in my humble opinion !

Other information : public corpuses are 3 months old... they will improve (src/tests/fuzz/confyaml.c  was added in november so DNP3 fuzzing corpus will grow in February)

PR #5763, opened 2021-01-21
Add negation support for TLS and corresponding unit tests

Fredamabob:

Link to ticket: https://redmine.openinfosecfoundation.org/issues/2269

Describe changes:
- Added support for negation operator when specifying a TLS version. SSL versions (they use separate parsers for SSL and TLS) already supported it.
- Added test cases to confirm that a negated rule behaves properly.

### DIFF
--- a/doc/userguide/rules/tls-keywords.rst
+++ b/doc/userguide/rules/tls-keywords.rst
@@ -153,11 +153,12 @@ Example::
 tls.version
 -----------
 
-Match on negotiated TLS/SSL version.
+Match on negotiated TLS/SSL version. Supports negation of the version match.
 
-Supported values: "1.0", "1.1", "1.2", "1.3"
+Supported values: "1.0", "1.1", "1.2", "1.3", "!1.0", "!1.1", "!1.2", "!1.3"
 
 It is also possible to match versions using a hex string.
+These can also be inverted.
 
 Examples::
 
--- a/src/detect-tls-version.c
+++ b/src/detect-tls-version.c
@@ -54,7 +54,7 @@
 /**
  * \brief Regex for parsing "id" option, matching number or "number"
  */
-#define PARSE_REGEX  "^\\s*([A-z0-9\\.]+|\"[A-z0-9\\.]+\")\\s*$"
+#define PARSE_REGEX "^\\s*(!?[A-z0-9\\.]+|\"[A-z0-9\\.]+\")\\s*$"
 
 static DetectParseRegex parse_regex;
 
@@ -135,6 +135,11 @@ static int DetectTlsVersionMatch (DetectEngineThreadCtx *det_ctx,
         ret = 1;
     }
 
+    if (tls_data->flags & SIGMATCH_HANDLE_NEGATION) {
+        /* Inverse match result for negated sig */
+        ret = (ret == 0) ? 1 : 0;
+    }
+
     SCReturnInt(ret);
 }
 
@@ -185,6 +190,11 @@ static DetectTlsVersionData *DetectTlsVersionParse (DetectEngineCtx *de_ctx, con
             tmp_str += 1;
         }
 
+        if (tmp_str[0] == '!') {
+            tls->flags |= SIGMATCH_HANDLE_NEGATION;
+            tmp_str++;
+        }
+
         if (strncmp("1.0", tmp_str, 3) == 0) {
             temp = TLS_VERSION_10;
         } else if (strncmp("1.1", tmp_str, 3) == 0) {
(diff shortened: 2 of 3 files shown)


catenacyber:
Thanks for this @Fredamabob 
I would love to have Rust replace this `PARSE_REGEX`
Would you be interested in doing that ?

Fredamabob:
> Thanks for this @Fredamabob
> I would love to have Rust replace this `PARSE_REGEX`
> Would you be interested in doing that ?

Hey, thanks for the feedback. I think rewriting it in Rust would be nice - but a larger task than expected. Do you think it'd be better to go with smaller changes to be safer, and open up another ticket for that?

catenacyber:
> Do you think it'd be better to go with smaller changes to be safer, and open up another ticket for that?

This PR is ok as is.
The ticket to track rustification is https://redmine.openinfosecfoundation.org/issues/3195

doc/userguide/rules/tls-keywords.rst
```diff
@@ -153,11 +153,12 @@ Example::
 tls.version
 -----------
 
-Match on negotiated TLS/SSL version.
+Match on negotiated TLS/SSL version. Supports negation of the version match.
 
-Supported values: "1.0", "1.1", "1.2", "1.3"
+Supported values: "1.0", "1.1", "1.2", "1.3", "!1.0", "!1.1", "!1.2", "!1.3"
```
catenacyber, line 158: Does this support multiple values ?
Like either 1.1 or 1.2 ?

Fredamabob: Yes. It will match on values 1.0->1.3, and !1.0->!1.3

catenacyber: That was not what I was asking.
Is it possible with this keyword to match on TLS flows which will either be TLS 1.1 or TLS 1.2 (but not TLS 1.3, not TLS 1.0 and not SSLv2 or 3) ?

catenacyber: I also found out recently about `SIGMATCH_HANDLE_NEGATION` 
Is this related ?

Fredamabob: Ah alright, my bad. After a bit of testing with suricata-verify, it seems that the answer may be no...
`SIGMATCH_HANDLE_NEGATION` appears to be very similar to the approach implemented here, it may be related - I am only aware of this now that you have mentioned it.

catenacyber: friendly ping @Fredamabob  are you planning to work in this ?

Fredamabob: @catenacyber Hi, sorry about the late reply. I can look into this if you'd like, but it might take awhile for me to get back to it.

catenacyber: Ok, no rush, just wanted to know if this PR still alive :-)

Fredamabob: Sorry for the long delay. I have updated it to use the flag SIGMATCH_HANDLE_NEGATION instead of using a new flag defined in detect-tls-version.h.

src/detect-tls-version.c
```diff
@@ -163,11 +168,9 @@ static DetectTlsVersionData *DetectTlsVersionParse (DetectEngineCtx *de_ctx, con
     if (ret > 1) {
         char ver_ptr[64];
         char *tmp_str;
-        pcre2len = sizeof(ver_ptr);
-        res = pcre2_substring_copy_bynumber(
-                parse_regex.match, 1, (PCRE2_UCHAR8 *)ver_ptr, &pcre2len);
+        res = pcre_copy_substring((char *)str, ov, MAX_SUBSTRINGS, 1, ver_ptr, sizeof(ver_ptr));
```
catenacyber, line 171: You should now use PCRE2 which got merged in master, and no longer PCRE1...
Basically, that means you should leave untouched `DetectParsePcreExec` and such... and your PR should work

doc/userguide/rules/tls-keywords.rst
catenacyber: Thanks for this improvement, I think CI will fail because of the PCRE1/PCRE2 thing cf other comment. Could you please take care of that ?

src/tests/detect-tls-version.c
```diff
@@ -236,15 +252,106 @@ static int DetectTlsVersionTestDetect02(void)
     PASS;
 }
 
+/*
+ * Test that a TLS version 1.0 packet will match on a !1.1 version signature
+ */
+static int DetectTlsVersionTestDetect03(void)
```
victorjulien, line 258: we're actively trying to move away from these types of unittests in favor of suricata-verify tests. The unittests that mimic traffic and setting up all the layers (packet, stream, flow, etc) are generally taking too many shortcuts that hurt us later.

victorjulien (changes requested):
Test change requested. Also generally please implement changes in a new branch and PR. This makes it easier to review. Thanks!

PR #5766, opened 2021-01-21
Ossfuzz corp/v4

victorjulien:


qa/run-ossfuzz-corpus.sh
```diff
@@ -0,0 +1,13 @@
...
+ls src/tests/fuzz/fuzz_*.c | sed 's/\.c//' | while read ftarget
+do
+    target=$(basename $ftarget)
+    echo "target $target"
+    #download public corpus
+    rm -f public.zip
+    wget --quiet "https://storage.googleapis.com/suricata-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/suricata_$target/public.zip"
+    rm -rf corpus_$target
+    unzip -q public.zip -d corpus_$target
+    #run target on corpus. Don't fail CI if the target fails.
```
catenacyber, line 11: Why not fail the CI if the target fails ?
(because we need to fix some bugs first ?)

catenacyber:
Will you want to cherry-pick https://github.com/OISF/suricata/pull/5613/commits/f520bc88eff8f9100a49089143a39d716522274e ?

qa/run-ossfuzz-corpus.sh
victorjulien: Right now I want to understand how we can max out the coverage, and it seems its not yet working as expected. With fuzz enabled we have only a minimal amount of extra code, yet coverage drops? I think when I'm confident the cov is handled correctly I want to either update the test to fail or add a second test that can fail

victorjulien: The math doesn't add up for me, maybe I'm overlooking something. I seems like cov for the unittests and verify flags didn't change, and we added a new flag with some cov. So how did things go down?

catenacyber: Can you visualize the diff in coverage ?

victorjulien: Yeah you can look here https://app.codecov.io/gh/OISF/suricata/compare/5766

catenacyber: Indeed, it does not seem consistent and now buggy with error "Missing base report." (was ok two minutes ago)

victorjulien:
> Will you want to cherry-pick [f520bc8](https://github.com/OISF/suricata/commit/f520bc88eff8f9100a49089143a39d716522274e) ?

I tried in #5768, but it leads to many build failures.

catenacyber:
This one seems to work 
https://github.com/catenacyber/suricata/pull/2/commits/5cf523ed14c7fc47a22428d54b666c10a51e294f

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5768 ?

PR #5232, opened 2020-07-28
protodetect: debug validation when multiple patterns match

catenacyber:
This PR is for discussion rather than straight merging.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2757

Describe changes:
- Adds debug validation when multiple patterns are found for protocol detection

Modifies #4607 : debug validation check instead of handling unreachable case
Modifies #5231 with fixed compilation

PR #5769, opened 2021-01-22
Next/20210122/v1

victorjulien:
#5232 plus disable a debug validation that is too restrictive and for which there is no alternative.


catenacyber:
Should use #5773 instead of #5232

PR #5775, opened 2021-01-23
Http2 files v5

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4116

Describe changes:
- Files decompression for HTTP2

suricata-verify-pr: 398

Makes https://github.com/OISF/suricata-verify/pull/398 pass (includes a framework change to have feature selection by check and not only by whole test)

Modifies #5713 with rebasing and fixing conflicts

There can be more commits squashing, instead of one commit adding the decompression, and one making it optional ?

PR #5776, opened 2021-01-23
Http2 files v6

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4116

Describe changes:
- Files decompression for HTTP2

suricata-verify-pr: 398

Makes https://github.com/OISF/suricata-verify/pull/398 pass (includes a framework change to have feature selection by check and not only by whole test)

Modifies #5775 with rightly fixed conflict, and commits squashed so that each one compiles

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -723,7 +723,7 @@ jobs:
           chmod 755 $HOME/.cargo/bin/cbindgen
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - run: ./autogen.sh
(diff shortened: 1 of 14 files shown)


victorjulien:
So I think in `master` we don't need to make it optional at configure time, only in the `master-6.0.x` branch. Can you make a branch for master that is not conditional and see if you can do a PR of this work against `master-6.0.x`?

catenacyber:
> So I think in master we don't need to make it optional at configure time, only in the master-6.0.x branch. Can you make a branch for master that is not conditional 

Done in #5780

> and see if you can do a PR of this work against master-6.0.x?

to be done later

PR #5658, opened 2020-12-12
detect/http-client-body: Improved support for shared bufs

jlucovsky:
This commit improves support for shared buffer usage, i.e., when
multiple rules share the HTTP client body and apply different
combinations of transforms and fast_patterns (or none).

Redmine issue: [4199](https://redmine.openinfosecfoundation.org/issues/4199)

Describe changes:
- Improved shared buffer usage with HTTP client body buffers

suricata-verify-pr: 379
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5778, opened 2021-01-24
detect/http-client-body: Improved support for shared bufs

jlucovsky:
Continuation of #5658 

This commit improves support for shared buffer usage, i.e., when
multiple rules share the HTTP client body and apply different
combinations of transforms and fast_patterns (or none).

Redmine issue: [4199](https://redmine.openinfosecfoundation.org/issues/4199)

Describe changes:
- Adds per-TX cleanup when inspection has completed for that tx.

suricata-verify-pr: 420
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-http-client-body.c
+++ b/src/detect-http-client-body.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2018 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -40,6 +40,7 @@
 #include "detect-engine-mpm.h"
 #include "detect-engine-state.h"
 #include "detect-engine-prefilter.h"
+#include "detect-engine-content-inspection.h"
 #include "detect-content.h"
 #include "detect-pcre.h"
 
@@ -58,6 +59,9 @@
 #include "detect-http-client-body.h"
 #include "stream-tcp.h"
 
+static int DetectEngineInspectBufferHttpRequest(DetectEngineCtx *, DetectEngineThreadCtx *,
+        const DetectEngineAppInspectionEngine *, const Signature *, Flow *, uint8_t, void *, void *,
+        uint64_t);
 static int DetectHttpClientBodySetup(DetectEngineCtx *, Signature *, const char *);
 static int DetectHttpClientBodySetupSticky(DetectEngineCtx *de_ctx, Signature *s, const char *str);
 #ifdef UNITTESTS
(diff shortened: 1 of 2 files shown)


victorjulien:
Can you confirm that CI ran with the SV PR? I think the body was edited later to add that reference in?

jlucovsky:
> Can you confirm that CI ran with the SV PR? I think the body was edited later to add that reference in?

They did not run with the sv pr. Restarting jobs ...

src/detect-http-client-body.c
```diff
@@ -261,12 +266,71 @@ static InspectionBuffer *HttpClientBodyGetDataCallback(DetectEngineThreadCtx *de
...
      * the window sizes when freeing data */
-    body->body_inspected = body->content_len_so_far;
-    SCLogDebug("body->body_inspected now: %"PRIu64, body->body_inspected);
 
-    SCReturnPtr(buffer, "InspectionBuffer");
+    if (r == 1) {
+        return DETECT_ENGINE_INSPECT_SIG_MATCH;
+    }
+
+    if (flags & STREAM_TOSERVER) {
```
victorjulien, line 324: this should be implied by the fact that this function is called for requests

src/detect-http-client-body.c
```diff
@@ -261,12 +266,71 @@ static InspectionBuffer *HttpClientBodyGetDataCallback(DetectEngineThreadCtx *de
...
+        if (AppLayerParserGetStateProgress(IPPROTO_TCP, ALPROTO_HTTP, txv, flags) >
+                HTP_REQUEST_BODY)
+            return DETECT_ENGINE_INSPECT_SIG_CANT_MATCH;
+    } else {
+        if (AppLayerParserGetStateProgress(IPPROTO_TCP, ALPROTO_HTTP, txv, flags) >
```
victorjulien, line 329: this code should be unreachable

src/detect.c
```diff
@@ -1523,6 +1549,10 @@ static void DetectRunTx(ThreadVars *tv,
 
             StoreDetectFlags(&tx, flow_flags, ipproto, alproto, new_detect_flags);
         }
+        if (new_detect_flags & APP_LAYER_TX_INSPECTED_FLAG) {
```
victorjulien, line 1552: this will only trigger when the tx is complete I think, which means with very large bodies we'll wait too long. I think we need to check per inspection round something like
```c
if (det_ctx->http_body_progress_updated) {
   body->body_inspected = det_ctx->http_body_progress;
   body->http_body_progress_updated = false;
}
```
`det_ctx::http_body_progress_updated` should then only be set if we actually did body inspection (so set by `DetectEngineInspectBufferHttpRequest` and friends after the various limits have been checked)

victorjulien:
I think it would be helpful to add a some SV tests that use large (multi megabyte) HTTP POSTs, as this will make it easier to see limitations of the approaches.

src/detect-http-client-body.c
jlucovsky: ok

src/detect-http-client-body.c
jlucovsky: ok

src/detect.c
jlucovsky: > we need to check per inspection round something like
This means after each call to `DetectRunTxInspectRule`?

victorjulien: No I think at the end of each tx iteration in `DetectRunTx`

jlucovsky:
Continued in #5820

PR #5750, opened 2021-01-19
modbus: move parser from c to rust

cccs-sadugas:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
- [3957](https://redmine.openinfosecfoundation.org/issues/3957)

Describe changes:
- Moves the C parsing logic into a rust module and removes unused C files.
- Moves the modbus detection keyword to a rust module.
- Adds and documents eve json logging, which did not seem to exist previously.
- Uses [sawp-modbus](https://crates.io/crates/sawp-modbus), our new modbus parser for integration in suricata and other projects.

Example of eve log for modbus alert:
```json
{
    "timestamp": "2004-08-26T12:12:19.148970+0000",
    "flow_id": 337987411284003,
    "pcap_cnt": 53,
    "event_type": "alert",
    "src_ip": "10.0.0.9",
    "src_port": 3082,
    "dest_ip": "10.0.0.3",
    "dest_port": 502,
    "proto": "TCP",
    "tx_id": 0,
    "alert": {
        "action": "allowed",
        "gid": 1,
        "signature_id": 1,
        "rev": 1,
        "signature": "Modbus function number test",
        "category": "",
        "severity": 3
    },
    "modbus": {
        "id": 1,
        "request": {
            "transaction_id": 1,
            "protocol_id": 0,
            "unit_id": 10,
            "function_raw": 1,
            "function_code": "RdCoils",
            "access_type": "READ | COILS",
            "category": "PUBLIC_ASSIGNED",
            "error_flags": "NONE",
            "read": {
                "address": 0,
                "quantity": 1
            }
        },
        "response": {
            "transaction_id": 1,
            "protocol_id": 0,
            "unit_id": 10,
            "function_raw": 1,
            "function_code": "RdCoils",
            "access_type": "READ | COILS",
            "category": "PUBLIC_ASSIGNED",
            "error_flags": "NONE",
            "read": {
                "data": "\u0000"
            }
        }
    },
    "app_proto": "modbus",
    "flow": {
        "pkts_toserver": 4,
[...]

PR #5790, opened 2021-01-26
modbus: move parser from c to rust - v2

cccs-sadugas:

Describe changes since last PR:
- See #5750 for description.
- Ran `clang-format` to format the code.
- Fixed the fuzzing crash by including the unconsumed input in `AppLayerResult::incomplete`'s `needed` argument.
- Added test cases to improve code coverage of the C module.
- Fixed MSRV to be `1.41.1` and updated sawp dependency to `0.1.1-rc.2` for this.

suricata-verify-pr: 411


cccs-sadugas:
@victorjulien The MSRV is now `1.41.1` as suggested in #5750 but it seems debian builds are still failing. Build logs show that `RUST_VERSION_KNOWN`  of `1.37.0` is installed. Should we aim to support `1.37.0`? Is this intended?
```
RUST_VERSION_KNOWN: 1.37.0
RUST_VERSION_MIN: 1.41.1
```

rust/src/modbus/log.rs
```diff
@@ -0,0 +1,148 @@
+/* Copyright (C) 2020 Open Information Security Foundation
```
catenacyber, line 1: nit : 2021

src/output-json-alert.c
```diff
@@ -542,6 +543,12 @@ static void AlertAddAppLayer(const Packet *p, JsonBuilder *jb,
         case ALPROTO_RDP:
             AlertJsonRDP(p->flow, tx_id, jb);
             break;
+        case ALPROTO_MODBUS:
```
catenacyber, line 546: Is this an addition ? (and not just the translation of the C parser)
If so, it deserves a specific commit

src/util-profiling.c
```diff
@@ -1309,6 +1309,7 @@ const char * PacketProfileLoggertIdToString(LoggerId id)
         CASE_CODE (LOGGER_JSON_HTTP);
         CASE_CODE (LOGGER_JSON_DHCP);
         CASE_CODE (LOGGER_JSON_KRB5);
+        CASE_CODE(LOGGER_JSON_MODBUS);
```
catenacyber, line 1312: Did clang-format not like a space after `CASE_CODE` ?

src/detect-modbus.c
```diff
@@ -570,442 +171,179 @@ void DetectModbusRegister(void)
...
 
 /**
  * \brief this function registers unit tests for DetectModbus
  */
-void DetectModbusRegisterTests(void)
+static void DetectModbusRegisterTests(void)
 {
-    UtRegisterTest("DetectModbusTest01 - Testing function",
-                   DetectModbusTest01);
-    UtRegisterTest("DetectModbusTest02 - Testing function and subfunction",
```
catenacyber, line 986: Why did you remove these unit tests ?
They should be useful to check if we still have the same behavior, right ?

src/util-profiling.c
cccs-sadugas: Yes, this was `clang-format`'s doing.

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -1811,3 +1811,110 @@ Example of HTTP2 logging, of a request and response:
       ]
     }
   }
+
+Event type: Modbus
+------------------
+
```
catenacyber, line 1817: This documentation addition could belong to its own commit

rust/src/modbus/detect.rs
```diff
@@ -0,0 +1,1345 @@
...
+use lazy_static::lazy_static;
+use regex::Regex;
+use sawp_modbus::{AccessType, CodeCategory, Data, FunctionCode, Message};
+use std::ffi::CStr;
+use std::ops::{Range, RangeInclusive};
+use std::os::raw::{c_char, c_void};
+use std::str::FromStr;
+
+lazy_static! {
+    static ref ACCESS_RE: Regex = Regex::new("^\\s*\"?\\s*access\\s*(read|write)\\s*(discretes|coils|input|holding)?(?:,\\s*address\\s+([<>]?\\d+)(?:<>(\\d+))?(?:,\\s*value\\s+([<>]?\\d+)(?:<>(\\d+))?)?)?\\s*\"?\\s*$").unwrap();
```
catenacyber, line 28: Could we use nom instead of regexes to parse the rule keyword value ?

rust/src/modbus/detect.rs
```diff
@@ -0,0 +1,1345 @@
...
+                .unwrap();
+}
+
+#[derive(Debug, PartialEq)]
+pub struct DetectModbus {
+    category: Option<CodeCategory>,
+    function: Option<FunctionCode>,
+    subfunction: Option<u16>,
+    access_type: Option<AccessType>,
+    unit_id: Option<Range<u16>>,
```
catenacyber, line 44: Interesting, did you see `DetectU32Data` in HTTP2 ?
What are the pros and cons for `Range` ?

catenacyber (commented):
This looks awesome.
I did not fully review it but I already asked some questions...

src/output-json-alert.c
cccs-sadugas: Yes, it is an addition and actually resolves a separate redmine ticket which I noticed after we implemented this: [2096](https://redmine.openinfosecfoundation.org/issues/2096) as opposed to [3957](https://redmine.openinfosecfoundation.org/issues/3957).

src/detect-modbus.c
victorjulien: Maybe they can be converted to rust tests? (`cargo test` style)

rust/src/modbus/detect.rs
victorjulien: I think @jasonish (or what it @chifflier?) was saying that for string parsing `nom` isn't the best choice. Not sure if that is still the consensus?

victorjulien:
> @victorjulien The MSRV is now `1.41.1` as suggested in #5750 but it seems debian builds are still failing. Build logs show that `RUST_VERSION_KNOWN` of `1.37.0` is installed. Should we aim to support `1.37.0`? Is this intended?
> 
> ```
> RUST_VERSION_KNOWN: 1.37.0
> RUST_VERSION_MIN: 1.41.1
> ```

Our goal is to have MSRV of 1.41.1, so that 1.37 should probably be bumped to that. cc @jasonish

src/detect-modbus.c
cccs-sadugas: All of these tests, and more, are now part of the rust module. Is there benefit to adding more tests to this C module?

See [rust/src/modbus/detect.rs](https://github.com/cccs-sadugas/suricata/blob/rust-modbus-v2/rust/src/modbus/detect.rs#L492).

Overall our testing strategy is as follows:
- tests related to the parsing of data are in `sawp-modbus` (81% coverage)
- tests related to parsing keywords and other suricata functionality are in their respective rust module (not sure these are recognized by codecov?)
- tests related to integration are in the C wrapper (minimal, to avoid duplication with unit tests)
- system tests are in suricata-verify
- we are currently soak testing on a staging environment to see if there are any defects

rust/src/modbus/detect.rs
cccs-sadugas: What would be the advantage of using `nom` over `regex`? The regex is a port of the pcre regex in the prior C module.

`nom` does support `&str` as an input instead of bytes IIRC. 

I can only find [this example](https://github.com/OISF/suricata/blob/master/rust/src/asn1/parse_rules.rs#L124) from asn1 keyword parsing.

rust/src/modbus/detect.rs
cccs-sadugas: `DetectU32Data` looks very similar to the way it was implemented in the prior [modbus C module](https://github.com/OISF/suricata/blob/master/src/detect-modbus.h#L50).

Our goal with using `Range` was to use functions like `range.contains()` and make it more idiomatic but in the end we've had to handle all the edge cases to make it compatible with the prior C module (i.e. Rust ranges are half-open).

Should we move to a `DetectModbusValue` style structure?

out-of-scope: If this is used across modules, should it be in a common module?

jasonish:
> Our goal is to have MSRV of 1.41.1, so that 1.37 should probably be bumped to that. cc @jasonish

Yes, it should be bumped up. Was just missed in the.. hashing PR I think.

rust/src/modbus/detect.rs
jasonish: I found using nom was ok for parsing a rule, so its not that bad.  I think it comes down to whatever is simpler to maintain..  For simple stuff a regex might be simpler, but we do have some regex that are really hard to maintain when changes are needed, and those might be better with nom, or even simple word splitting and parsing.

catenacyber: > What would be the advantage of using nom over regex? 

It seems more readable/maintainable to me
So, this is subjective, and I am good either way.

> I can only find this example from asn1 keyword parsing.

Another example is `rs_http2_detect_settingsctx_parse`

rust/src/modbus/detect.rs
catenacyber: The point would be to have a common module to reduce code duplication...
I think this is not a blocking point and you should go as you feel best

src/detect-modbus.c
catenacyber: >  Is there benefit to adding more tests to this C module?

I did not mean to add more C tests.
Translating the C tests to Rust is ok :-)

But I would like it (the moving of C tests ro Rust) to be in its own commit after the parser itself went from C to Rust.
This way, we can check easily that the translation of the parser from C to Rust did not break these tests.
When I translated SSH from C to Rust, these C tests helped me get some edge cases right

victorjulien: > * tests related to parsing keywords and other suricata functionality are in their respective rust module (not sure these are recognized by codecov?)

No, our codecov integration doesn't support Rust yet. Still need to look into that.

cccs-sadugas:
> 
> 
> > Our goal is to have MSRV of 1.41.1, so that 1.37 should probably be bumped to that. cc @jasonish
> 
> Yes, it should be bumped up. Was just missed in the.. hashing PR I think.

Should I bump [the value](https://github.com/OISF/suricata/blob/master/.github/workflows/builds.yml#L26) up in this PR?

```
# A known good Rust version we should test against.
RUST_VERSION_KNOWN: "1.37.0"  ->  "1.41.1"

# The minimum version of Rust supported.
RUST_VERSION_MIN: "1.41.1"
```

Not sure what is meant by a known good version of Rust, but I think it's the variable used for debian 9 & 10 builds.

jasonish:
> Not sure what is meant by a known good version of Rust, but I think it's the variable used for debian 9 & 10 builds.

At some point we would run into 3 issues..
1) A minimum supported version of Rust. We don't want to add new code that will fail on some known minimum version.
2) The latest Rust stable would change something that would cause our build to fail. Usually promoting something to a compile error.  Obviously we want to know about this and fix, but it still useful to test against a known recent version of Rust. We fix these issues pretty quickly, and they happen less and less.
3) So we added the known good version, so we could still get a build with unit tests and verification running, etc.

cccs-sadugas:
> 
> 
> > Not sure what is meant by a known good version of Rust, but I think it's the variable used for debian 9 & 10 builds.
> 
> At some point we would run into 3 issues..
> 
>     1. A minimum supported version of Rust. We don't want to add new code that will fail on some known minimum version.
> 
>     2. The latest Rust stable would change something that would cause our build to fail. Usually promoting something to a compile error.  Obviously we want to know about this and fix, but it still useful to test against a known recent version of Rust. We fix these issues pretty quickly, and they happen less and less.
> 
>     3. So we added the known good version, so we could still get a build with unit tests and verification running, etc.

Thank you for the insight. If this is the case I can update the known version in a future version of this pull request. I would choose the latest stable version known to pass compilation, testing and verification (and maybe expand the documentation to reflect this).

cccs-sadugas:
Thank you for prompt feedback and discussion. Another version of this PR will be coming soon :).

PR #5789, opened 2021-01-25
Dataset ops v1.1

regit:
Update of #5754 fixing some tests.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Fix a non void function not returning a value in case of error.
- Rebase on latest master

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5797, opened 2021-01-28
Dataset ops v1.2

regit:
Update of #5789 addressing comments.

Describe changes:
- Rebase on latest master
- Fix documentation
- Rename `dataset-exist` to `dataset-lookup`

### DIFF
--- a/src/runmode-unix-socket.c
+++ b/src/runmode-unix-socket.c
@@ -744,6 +744,15 @@ TmEcode UnixSocketDatasetRemove(json_t *cmd, json_t* answer, void *data)
     }
 }
 
+TmEcode UnixSocketDatasetDump(json_t *cmd, json_t *answer, void *data)
+{
+    SCEnter();
+    SCLogDebug("Going to dump datasets");
+    DatasetsSave();
+    json_object_set_new(answer, "message", json_string("datasets dump done"));
+    SCReturnInt(TM_ECODE_OK);
+}
+
 /**
  * \brief Command to add a tenant handler
  *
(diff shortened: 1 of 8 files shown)


catenacyber:
@inashivb as this is datasets, should you be reviewing this ?

inashivb:
> @inashivb as this is datasets, should you be reviewing this ?
Sure. I'll test the commands.

src/runmode-unix-socket.c
```diff
@@ -744,6 +744,100 @@ TmEcode UnixSocketDatasetRemove(json_t *cmd, json_t* answer, void *data)
     }
 }
 
+TmEcode UnixSocketDatasetDump(json_t *cmd, json_t *answer, void *data)
+{
+    SCEnter();
+    SCLogDebug("Going to dump datasets");
+    DatasetsSave();
```
inashivb, line 751: Should we check if `save` command is used in the loaded datasets? I used it with a `load` command dataset and got the output
```
Success:
"datasets dump done"
```
but it seems to move to next entry if `save` is not found so it did not seem to do anything.

inashivb (commented):
Looks great. :) Test well too. One question inline though.

suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.memuse | 531654848 | 505591168 | 105.16% | 

Pipeline 6126

src/runmode-unix-socket.c
regit: I don't think it is an issue as it is just a noop in this case. For example, you can use the command to dump periodically the datasets (via cron) and in this case you don't really control which are the datasets used and don't really want a warning.

PR #5800, opened 2021-01-29
Indication that Suricata is ready for packets.

rtkjweeks:
Depending on the configuration used, the time it takes for
Suricata to initialize and have worker threads ready to
process packets can vary dramatically.

In order to facility quicker end-to-end/integration
tests, having an indication for when this state has
been reached would be beneficial.

This change first waits for workers to unpause, and then
waits for them to actively enter their packet polling loops
before outputting an log string which test frameworks
can use to determine that Suricata is ready to process packets.

Describe changes:
-
-
-

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -1519,6 +1519,9 @@ TmEcode ReceiveAFPLoop(ThreadVars *tv, void *data, void *slot)
     fds.fd = ptv->socket;
     fds.events = POLLIN;
 
+    // Indicate that the thread is actually running its application level code (i.e., it can poll
+    // packets)
+    TmThreadsSetFlag(tv, THV_RUNNING);
     while (1) {
         /* Start by checking the state of our interface */
         if (unlikely(ptv->afp_state == AFP_STATE_DOWN)) {
--- a/src/source-pfring.c
+++ b/src/source-pfring.c
@@ -360,6 +360,9 @@ TmEcode ReceivePfringLoop(ThreadVars *tv, void *data, void *slot)
         SCReturnInt(TM_ECODE_FAILED);
     }
 
+    // Indicate that the thread is actually running its application level code (i.e., it can poll
+    // packets)
+    TmThreadsSetFlag(tv, THV_RUNNING);
     while(1) {
         if (suricata_ctl_flags & SURICATA_STOP) {
             SCReturnInt(TM_ECODE_OK);
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -232,6 +232,10 @@ uint16_t g_vlan_mask = 0xffff;
  * support */
 bool g_disable_hashing = false;
 
+/* What for worker threads to indicate they're running before indicating that
+ * Suricata has started */
+bool g_sync_workers = false;
+
 /** Suricata instance */
 SCInstance suricata;
 
@@ -631,6 +635,8 @@ static void PrintUsage(const char *progname)
     printf("\t--reject-dev <dev>                   : send reject packets from this interface\n");
 #endif
     printf("\t--set name=value                     : set a configuration value\n");
+    printf("\t--sync-workers                       : wait for all workers to start before logging "
+           "that Suricata has started\n");
     printf("\n");
     printf("\nTo run the engine with default configuration on "
             "interface eth0 with signature file \"signatures.rules\", run the "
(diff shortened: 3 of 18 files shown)


src/source-af-packet.c
```diff
@@ -1541,6 +1541,8 @@ TmEcode ReceiveAFPLoop(ThreadVars *tv, void *data, void *slot)
          * us from alloc'ing packets at line rate */
         PacketPoolWait();
 
+        // Indicate that the thread is actually running its application level code (i.e., it can poll packets)
+        TmThreadsSetFlag(tv, THV_RUNNING);
```
victorjulien, line 1545: `TmThreadSetFlag` wraps around an atomic operation, so it is relatively expensive. Running that from this loop seems excessive.

src/source-pfring.c
```diff
@@ -387,6 +387,8 @@ TmEcode ReceivePfringLoop(ThreadVars *tv, void *data, void *slot)
             pkt_buffer = GET_PKT_DIRECT_DATA(p);
         }
 
+        // Indicate that the thread is actually running its application level code (i.e., it can poll packets)
+        TmThreadsSetFlag(tv, THV_RUNNING);
```
victorjulien, line 391: Same here, calling this per packet is excessive.

src/tm-threads.c
```diff
@@ -1842,6 +1842,64 @@ void TmThreadContinueThreads()
...
+void TmThreadEnsureUnpaused()
+{
+    ThreadVars *tv = NULL;
+    int i = 0;
+    int pausedThreads = 0;
+
+    SCMutexLock(&tv_root_lock);
+
+    do {
+        pausedThreads = 0;
```
victorjulien, line 1854: style: `paused_threads`

src/tmqh-flow.c
```diff
@@ -98,6 +99,8 @@ Packet *TmqhInputFlow(ThreadVars *tv)
 
     StatsSyncCountersIfSignalled(tv);
 
+    TmThreadsSetFlag(tv, THV_RUNNING);
```
victorjulien, line 102: another per packet atomic set for something that is essentially something that should only happen at startup

src/tmqh-packetpool.c
```diff
@@ -369,6 +369,7 @@ void PacketPoolDestroy(void)
 
 Packet *TmqhInputPacketpool(ThreadVars *tv)
 {
+    TmThreadsSetFlag(tv, THV_RUNNING);
```
victorjulien, line 372: same here

src/tmqh-simple.c
```diff
@@ -50,6 +51,8 @@ Packet *TmqhInputSimple(ThreadVars *t)
 
     StatsSyncCountersIfSignalled(t);
 
+    TmThreadsSetFlag(t, THV_RUNNING);
```
victorjulien, line 54: same

src/suricata.c
```diff
@@ -2800,8 +2801,10 @@ int SuricataMain(int argc, char **argv)
     SC_ATOMIC_SET(engine_stage, SURICATA_RUNTIME);
     PacketPoolPostRunmodes();
 
-    /* Un-pause all the paused threads */
+    /* Un-pause all the paused threads, and wait for them to start */
     TmThreadContinueThreads();
+    TmThreadEnsureUnpaused();
+    TmThreadEnsureRunning();
```
victorjulien, line 2807: this adds 2 loops that (potentially) sleep, how does this affect start up time? I fear it can add quite a bit of time for runs with many cores.

victorjulien (changes requested):
Hi Jeff, I think addressing this is a good idea. Some feedback below and in the comments:

I have several performance concerns with this (see comments).

Also needs an investigation into the CI failures.

src/source-af-packet.c
rtkjweeks: Yeah; I was trying to get them as close as possible to the first poll for packets.  I'll pull them out to before the loop.

src/suricata.c
rtkjweeks: Hmm... I can look into the affect it has.

I think it shouldn't actually affect packet processing, as the threads themselves continue running as usual, this just creates a sync point, to ensure all the threads are up and running, before executing the remaining calls (PostRunStartedDetectSetup(), SCPledge() and SuricataMainLoop()), which largely just setup and monitor for signals, I believe, right?

That said; I feel like this is mostly useful in an integration/end-to-end test scenario, so maybe this sync point should be command line configurable?  So, in production, you can bypass it if you want.

What do you think?

rtkjweeks:
Closing per v2 PR: https://github.com/OISF/suricata/pull/5842

PR #5619, opened 2020-12-01
 detect: Improve support for larger address values.

jlucovsky:
Continuation of pr #5604

This PR modifies the address rule value parsing to permit larger address values in the rule variables section.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3887](https://redmine.openinfosecfoundation.org/issues/3887)

Describe changes:
- Reduce heap allocations during address parsing.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5808, opened 2021-01-31
detect: Improve support for larger address values.

jlucovsky:
Continuation of pr #5619 

This PR modifies the address rule value parsing to permit larger address values in the rule variables section.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3887](https://redmine.openinfosecfoundation.org/issues/3887)

Describe changes:
- Address review comments.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2020 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -54,6 +54,10 @@ static void DetectAddressPrint(DetectAddress *);
 static int DetectAddressCutNot(DetectAddress *, DetectAddress **);
 static int DetectAddressCut(DetectEngineCtx *, DetectAddress *, DetectAddress *,
                             DetectAddress **);
+static int DetectAddressParse2(const DetectEngineCtx *de_ctx, DetectAddressHead *gh,
+        DetectAddressHead *ghn, const char *s, int negate, ResolvedVariablesList *var_list,
+        int recur);
+
 int DetectAddressMergeNot(DetectAddressHead *gh, DetectAddressHead *ghn);
 
 /**
@@ -715,6 +719,10 @@ static int DetectAddressSetup(DetectAddressHead *gh, const char *s)
  * \brief Parses an address string and updates the 2 address heads with the
  *        address data.
  *
+ * Note that this function should only be called by the wrapping function
+ * DetectAddressParse2. The wrapping function provides long address handling
+ * when the address size exceeds a threshold value.
+ *
  * \todo We don't seem to be handling negated cases, like [addr,![!addr,addr]],
  *       since we pass around negate without keeping a count of ! with depth.
  *       Can solve this by keeping a count of the negations with depth, so that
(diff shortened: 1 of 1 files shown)


src/detect-engine-address.c
```diff
@@ -1009,6 +1014,38 @@ static int DetectAddressParse2(const DetectEngineCtx *de_ctx,
...
+static int DetectAddressParse2(const DetectEngineCtx *de_ctx, DetectAddressHead *gh,
+        DetectAddressHead *ghn, const char *s, int negate, ResolvedVariablesList *var_list,
+        int recur)
+{
+    int rc;
+#define MAX_ADDRESS_LENGTH 8192
+
+    size_t address_length = strlen(s);
+    if (address_length > (MAX_ADDRESS_LENGTH - 1)) {
+        char *address = SCCalloc(1, address_length);
```
catenacyber, line 1032: Is there a way to get rid of `address` and use only `s` ?

catenacyber (commented):
Looks good to me...
Wonder if a deeper rewriting would be beneficial...

src/detect-engine-address.c
jlucovsky: `address` is a return parameter ... built as `s` is parsed.

catenacyber: Ok, what is this return parameter for ?

jlucovsky: Recursion is possible when parsing addresses; when this happens `address` contains the subset of the original address to be parsed in the next step.

catenacyber: What recursion ? Between variables like `EXTERNAL_NET` and their values ?

jlucovsky: See the calls in `DetectAddressParseInternal` that invoke `DetectAddressParse2` for example.

catenacyber: I think I do not get it :-/
So I would upvote code simplicity over saving an allocation...

src/detect-engine-address.c
```diff
@@ -755,11 +760,13 @@ static int DetectAddressParse2(const DetectEngineCtx *de_ctx,
...
+    size_t size = strlen(s);
+    for (u = 0, x = 0; u < size && x < address_length; u++) {
+        if (x == (address_length - 1)) {
+            SCLogError(SC_ERR_ADDRESS_ENGINE_GENERIC,
+                    "Hit the address buffer"
+                    " limit for the supplied address.  Invalidating sig.  "
+                    "Please file a bug report on this.");
             goto error;
         }
         address[x] = s[u];
```
victorjulien, line 772: I'm confused about the logic here, and why we'd need a 'full size' `address` here. AFAICS we use it as a temporary place to put the address from `s` before passing it on. But does it need to be so big for that? Or could we have a smaller (local) temp var that is just reset after loop interations?

victorjulien: I suppose its because this parses anything between `[...]` which can have any length.

src/detect-engine-address.c
victorjulien: From what I can gather (and remember), `address` is really only used to tokenize the input string on the fly. Other than making a copy the only modifications I can spot are inserting `\0` chars to split the string before passing the parts on to further parsing logic. If this is indeed the case another approach could be to pass arrays of chars with an explicit length around. It does look like this is quite intrusive as we call out into various functions with the address part strings as the argument.

victorjulien (approved):

victorjulien:
Merged in #6071, thanks!

PR #5810, opened 2021-02-01
modbus: move parser from c to rust - v4

cccs-sadugas:

Changes since last PR:
- Change copyright year to 2021.
- Separate commits for rust implementation, moving regression tests, eve logging and docs.
- Change known good version to be more recent and added docs for this variable.
- Update `sawp` to `0.1.1-rc.3`.

Regression Testing:
- `DetectEngineInspectModbusTest04`
  - `0x12` is now an assigned function code so we are testing with `0x3F`
- `DetectModbusTest04`
  - bitflags’s “not” operator isn’t the same as C so `~MODBUS_CAT_PUBLIC_ASSIGNED` won't work. Bitflags only flips the bits that have a flag value.

Todo:
- Refactor `DetectModbusValue` (and other rust struct like `DetectU32Data`) into a common module in a separate PR.
- Our long term goal is to stop exposing `nom` and `bitflags` through our public SAWP API so that we aren't tightly coupled to `nom` and `bitflags` versions. Expect this change in a future PR.

Link to previous PR:
- #5790

suricata-verify-pr: 411


src/detect-modbus.c
```diff
@@ -572,6 +176,72 @@ void DetectModbusRegister(void)
...
+
+        if (modbus->data)
+            SCFree(modbus->data);
+
+        SCFree(modbus);
+    }
+}
+
 /** \test Signature containing a function. */
 static int DetectModbusTest01(void)
```
cccs-sadugas: @catenacyber We tried to change the unit tests as little as possible in this commit to test the rust implementation using C tests before moving them to Rust. There are some minor changes, as noted in the description of the PR. More tests are also in `detect-engine-modbus.c`. Unfortunately, we could not do the same with the `app-layer-modbus.c` tests because this would be significantly more effort.

rust/src/modbus/modbus.rs
```diff
@@ -0,0 +1,1146 @@
...
+) -> *mut std::os::raw::c_void {
+    let state = cast_pointer!(state, ModbusState);
+    match state.get_tx(tx_id) {
+        Some(tx) => (tx as *mut ModbusTransaction) as *mut std::os::raw::c_void,
+        None => std::ptr::null_mut(),
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn rs_modbus_state_progress_completion_status(
```
victorjulien, line 339: unused function

victorjulien: this logic was changed in the git master recently

src/output-json-modbus.c
```diff
@@ -0,0 +1,173 @@
...
+    LogModbusFileCtx *modbuslog_ctx;
+    MemBuffer *buffer;
+} JsonModbusLogThread;
+
+static int JsonModbusLogger(ThreadVars *tv, void *thread_data, const Packet *p, Flow *f,
+        void *state, void *tx, uint64_t tx_id)
+{
+    JsonModbusLogThread *thread = thread_data;
+
+    JsonBuilder *js = CreateEveHeader(p, LOG_DIR_PACKET, "modbus", NULL);
```
victorjulien, line 53: since we log the bidirectional tx I think this should be `LOG_DIR_FLOW`

rust/src/modbus/detect.rs
```diff
@@ -0,0 +1,1345 @@
...
+    }
+}
+
+/// Compares a transaction to a signature to determine whether the transaction
+/// matches the signature. If it does, 1 is returned; otherwise 0 is returned.
+#[no_mangle]
+pub extern "C" fn rs_modbus_inspect(tx: &ModbusTransaction, modbus: &DetectModbusRust) -> u8 {
+    let req = match &tx.request {
+        Some(r) => r,
+        None => return 0,
```
victorjulien, line 196: So we can only run inspection if we saw a request? Since the parser has indicated it accepts GAPS, we might see response only transactions, right? Both in packet loss and midstream cases. So I wonder if there is anything we could salvage from just the responses that is worth inspecting here?

victorjulien (commented):
Incomplete review, some initial feedback/questions.

src/detect-modbus.c
catenacyber: Thank you for this.

How did you change the test `DetectModbusTest04` ?
Why was it not possible for app-layer-modbus.c ?
Reading `ModbusParserTest01`, was it because of accessing internal transaction fields ? like ` FAIL_IF_NOT(tx->read.address == 0x7890);` ?

cccs-sadugas: > How did you change the test `DetectModbusTest04` ?

The actual diff is [here](https://github.com/OISF/suricata/pull/5810/commits/a51e82e508c8df47b7b3959e9e2046997e7a1919#diff-a8bc8a6c2bef032dcc859072e8031de5cd492227114fb74cd9079b569edf9400R357).
Instead of comparing to `~MODBUS_CAT_PUBLIC_ASSIGNED` we are checking all the flags individually.
```rust

    FAIL_IF(modbus->category & MODBUS_CAT_PUBLIC_ASSIGNED);
    FAIL_IF_NOT(modbus->category & MODBUS_CAT_PUBLIC_UNASSIGNED);
    FAIL_IF_NOT(modbus->category & MODBUS_CAT_USER_DEFINED);
    FAIL_IF_NOT(modbus->category & MODBUS_CAT_RESERVED); 
```
> Why was it not possible for app-layer-modbus.c ?

I'll see what I can do :).

> Reading `ModbusParserTest01`, was it because of accessing internal transaction fields ? like ` FAIL_IF_NOT(tx->read.address == 0x7890);` ?

Yes, I will have to create temporary ffi accessors for the test cases to work like in `detect-modbus.c`. I'll see how much effort this is.

catenacyber: Looks good for `DetectModbusTest04 `
Thanks for the excellent explanation.

Temporary ffi accessors seem indeed too much work...
Is there a better way @jasonish ?
Is it possible to share the structure between Rust and C ?

In the meantime, how did you translate this specific `ModbusParserTest01` ? (I do not see such a name in the rust tests)

rust/src/modbus/detect.rs
```diff
@@ -0,0 +1,1345 @@
...
+use lazy_static::lazy_static;
+use regex::Regex;
+use sawp_modbus::{AccessType, CodeCategory, Data, FunctionCode, Message};
+use std::ffi::CStr;
+use std::ops::{Range, RangeInclusive};
+use std::os::raw::{c_char, c_void};
+use std::str::FromStr;
+
+lazy_static! {
+    static ref ACCESS_RE: Regex = Regex::new("^\\s*\"?\\s*access\\s*(read|write)\\s*(discretes|coils|input|holding)?(?:,\\s*address\\s+([<>]?\\d+)(?:<>(\\d+))?(?:,\\s*value\\s+([<>]?\\d+)(?:<>(\\d+))?)?)?\\s*\"?\\s*$").unwrap();
```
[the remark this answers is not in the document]
cccs-sadugas: It's possible using a backslash.
```rust
let string = "line 1 \
    line 2 \
    line 3";
```

Another option could be decomposing the regex into smaller parts.
```rust
Regex::new(&format!("{}{}{}{}", function, access, address, value))
```

src/detect-modbus.c
cccs-sadugas: I think it will be similar effort either way. The advantage of writing accessors is that we don't have to refactor the C tests to make them compatible with the rust structs (the diff remains as small as possible).

That test was moved to [rust/src/modbus/modbus.rs](https://github.com/OISF/suricata/pull/5810/files?file-filters%5B%5D=.am&file-filters%5B%5D=.c&file-filters%5B%5D=.h&file-filters%5B%5D=.rs#diff-57a6cf267e4fb39aa835b56cc817f0c7dccbafc37de20f4c90eaf8eb216a5958R683). It does the same function, address, qt check, calls parse again for the response, adds a couple more checks and also checks transaction length.

All tests in that file should be the same order as the C tests.
- `ModbusParserTest12` wasn't added because it calls `UTHBuildPacket`whereas we are testing the parse function.
- `ModbusParserTest18` wasn't added because there is no stream depth in the new module. Not sure how this is handled by `RustParser`.

rust/src/modbus/detect.rs
cccs-sadugas: We can certainly improve inspection.

I assume you mean that we passed the flag `APP_LAYER_PARSER_OPT_ACCEPT_GAPS`. I'm not sure what exactly this flag does because I can't find documentation for it.

The C parser did not accept gaps. The rust parser can accept gaps but messages must be encapsulated in a single packet. E.g. given,
- packet 1: message 1
- packet 2: first half of message 2
- packet 3: second half of message 2
- packet 4: message 3
- packet 5: message 4

This will occur:
- Losing no packets -- parsing packet 2 will result in an incomplete error (`AppLayerResult::incomplete(_)`).
- Losing only packet 2 or only packet 3 -- will result in an unrecoverable error (`AppLayerResult::err()`).
- Losing packet 4 -- would be fine.

We are considering adding more complex gap handling in the future, allowing loss of packet 3 to output a partial packet 2 and recover in time for packet 4 and 5.

src/detect-modbus.c
catenacyber: Ok, so `read_coils` is the new `ModbusParserTest01` right ?

For `ModbusParserTest12` what is the replacement ? A suricata-verify test to check that we get an alert ?

`ModbusParserTest18` may need some more investigation... The `reassembly_depth` seems generic about TCP...

cccs-sadugas:
New version of pull request is out: #5849

src/detect-modbus.c
cccs-sadugas: All of the `ModbusParserTest*` were re-introduced in #5849 [app-layer-modbus.c](https://github.com/OISF/suricata/blob/507f811f9d8be83638e7c2e4793fc65f69929e01/src/app-layer-modbus.c#L267).

I am not sure how `reassembly_depth` works with rust parsers and started looking at `stream-tcp-reassmble.c`. If you have any relevant documentation that would be helpful.

catenacyber: You can take a look at https://suricata.readthedocs.io/en/suricata-6.0.0/configuration/suricata-yaml.html?highlight=reassembly%20depth#stream-engine

It should work the same for C and Rust parsers (as it is in TCP handling, not in the app-layer handling)

PR #5811, opened 2021-02-02
Ossfuzz corp v6

catenacyber:

Describe changes:
- Runs CI on public corpuses of fuzz targets for coverage and ASAN bugs
- simple fuzz driver can run on flat directories (not only files)
- fix bug of uninitialized memory in `bytemath` keyword parsing

### DIFF
--- a/src/decode-ethernet.c
+++ b/src/decode-ethernet.c
@@ -38,6 +38,8 @@
 #include "util-unittest.h"
 #include "util-debug.h"
 
+#define MAX_ETH_OFFSET 256
+
 int DecodeEthernet(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
                    const uint8_t *pkt, uint32_t len)
 {
@@ -48,6 +50,9 @@ int DecodeEthernet(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
         return TM_ECODE_FAILED;
     }
 
+    if (unlikely(p->ethh != NULL) && ((uint8_t *)p->ethh) - p->ext_pkt > MAX_ETH_OFFSET) {
+        return TM_ECODE_FAILED;
+    }
     p->ethh = (EthernetHdr *)pkt;
     if (unlikely(p->ethh == NULL))
         return TM_ECODE_FAILED;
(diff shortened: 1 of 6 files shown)


src/decode-ethernet.c
```diff
@@ -38,6 +38,8 @@
 #include "util-unittest.h"
 #include "util-debug.h"
 
+#define MAX_ETH_OFFSET 256
```
catenacyber, line 41: This is a simple fix.

Another fix could be to count recursion levels in every `DecodeX` function

A better fix would to have a generic `DecodeFirstLayers` function which will loop calling every `DecodeX` function, and these `DecodeX` functions should now return the number of consumed bytes, and the next layer to decode... So this would be a deeper bigger change

catenacyber:
CIFuzz seems to have unrelated problems with its rust...

catenacyber:
> CIFuzz seems to have unrelated problems with its rust...

I suspect this is due to rustc 1.51.0-nightly (0e63af5da 2021-02-01)
Let's wait a bit more to see if oss-fuzz manages to build with it

src/decode-ethernet.c
```diff
@@ -48,6 +50,9 @@ int DecodeEthernet(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
         return TM_ECODE_FAILED;
     }
 
+    if (unlikely(p->ethh != NULL) && ((uint8_t *)p->ethh) - p->ext_pkt > MAX_ETH_OFFSET) {
```
victorjulien, line 53: shouldn't use `p->ext_pkt`, only use `GET_PKT_DATA(p)`.

This solution is also too limiting I think, we should probably count the total number of layers so that too much recursion can't happen anywhere

catenacyber: > we should probably count the total number of layers

Can we add a field to the structure `Packet` ?

Or do we need to get into the trouble fo https://github.com/OISF/suricata/pull/5811#discussion_r568441311 ?

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5837

PR #5812, opened 2021-02-02
libsuricata - baby steps - v1.0

jasonish:
This PR moves one C method called by Rust into the "rust context" of C function pointers prevent a circular reference that prevented us from using autotools/libtool convenience wrappers. Then we make use of this convenience library for Suricata and all the fuzz applications, which is how we should have done this in the first place.

This is a precursor to upcoming lib work, but want to see this simple step pass all QA.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.o
 *.lo
+*.la
 *.in
 *.[ch]e
 *.log
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -22,6 +22,7 @@ use crate::core::{DetectEngineState,Flow,AppLayerEventType,AppLayerDecoderEvents
 use crate::filecontainer::FileContainer;
 use crate::applayer;
 use std::os::raw::{c_void,c_char,c_int};
+use crate::core::SC;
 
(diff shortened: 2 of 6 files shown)


victorjulien:
Codecov and cifuzz are having issues. Codecov is likely unrelated (they report degraded service on their site) and cifuzz is also having issues in other PRs.

victorjulien:
Think this looks good. How many others are there to do?

catenacyber:
Looks awesome :-)

jasonish:
This is the first platform neutral step which is more or less what we
wanted when we added the guzzlers but couldn’t due to a circular reference.

The next steps will deal with some platform specific stuff that may take
longer to iron out.

jasonish:
> Think this looks good. How many others are there to do?

Hold off on any merging/QA on this.  I'm trying something else where we use a new Rust project that pulls in the C static lib and the Rust static lib and creates one shared library.  It seems to be working well with one issue.  In order for a C function to be exported in the shared library, it needs a dummy wrapper Rust function around it, otherwise the optimizer removes it from the output. But I'm hoping there is a work-around for this.

PR #5818, opened 2021-02-05
fuzz: adds structure aware target

catenacyber:

Describe changes:
- adds structure aware fuzz target (aka fuzz suricata, not libpcap)

The point is to have a more efficient fuzzing discovering more code coverage and finding more bugs

To do so, it uses a fuzz-friendly format instead of libpcap
To test it you need the fuzzpcap dependency which can generate either generic pcap, or pcaps with only one tcp stream (computing sequence and acks automatically)
There is also a signature generation in this target, meant to produce more valid signatures.

Will work with oss-fuzz with https://github.com/catenacyber/oss-fuzz/tree/suricataware

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2518,6 +2518,8 @@ fi
     ])
     AC_SUBST(RUST_FEATURES)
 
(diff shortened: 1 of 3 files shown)


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5965

PR #5821, opened 2021-02-05
Rust cleanup v1

catenacyber:
None

Describe changes:
- Remove some dead code
- Fix rust nightly warning by using `response_gap` in template parser

There are some more unused rust exported function :
- SCSha256FinalizeToHex
- SCMd5FinalizeToHex
- jb_append_float
Should we remove them ? Or will they be useful soon ?

### DIFF
--- a/rust/src/applayertemplate/template.rs
+++ b/rust/src/applayertemplate/template.rs
@@ -177,6 +177,17 @@ impl TemplateState {
             return AppLayerResult::ok();
         }
 
+        if self.response_gap {
+            if probe(input).is_err() {
+                // The parser now needs to decide what to do as we are not in sync.
+                // For this template, we'll just try again next time.
+                return AppLayerResult::ok();
+            }
+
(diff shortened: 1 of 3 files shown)


jasonish:
`jb_append_float` was added more for API completeness.  Basically, as an API I find it not to have a `set_float`, but not an `append_float` as a complete API would have the same variants here for completeness.  But it we don't use it, should we remove it?

Its similar with the hashing.. I think they were used in certain versions of the PR, but some reverts to avoid other struct growth has them not being used.  But again, they are kind there for completeness should they be needed.

To me they are unused, but they are not out of place, so I don't know what is best.  At least jb_append_float has a unit test.  Maybe I should add unit tests for the unused hashing functions as well.

catenacyber:
> To me they are unused, but they are not out of place, so I don't know what is best.

I feel the same about these ones

victorjulien:
Merged in #6042, thanks!

PR #5822, opened 2021-02-05
libsuricata - v2.0

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5812

This PR compiles the Suricata C source into a .a (non-libtool) convenience library which is used along with the Rust .a library to create the Suricata binary as well as the fuzz targets. This is generally more idiomatic than specifying all the source files to each outputted binary which we had to do due to a circular reference which is also fixed.

The 2 .a Suricata static libraries are now also install as part of Suricata, libsuricata_c.a and libsuricata_rust.a.  There is no way with autotools to combine the 2, however we could probably explode them and re-archive them into a single library with a script.

For demonstration purposes I also add a "make install-headers" target that depends on gnu make for now. This is a bit of a hack to prevent the headers from always being install which is the norm for autoconf/automake projects.

So after a normal `make install`, you can go into `src/` and run `make install-headers`, then you can create a small program like the following:
```
#include <suricata.h>

int main(int argc, char **argv) {
	return SuricataMain(argc, argv);
}
```
Run `suricata --build-info` to get the `LDFLAGS` required (this should actually be renamed LIBS now that I think about it) and then you can compile the above program with the following command:
```
gcc -o test -DHAVE_CONFIG_H test.c -I/usr/local/include/suricata \
    /usr/local/lib/libsuricata_c.a /usr/local/lib/libsuricata_rust.a /usr/local/lib/libhtp.a $LIBS...
```

On my system this looks like:
```
gcc -o test -DHAVE_CONFIG_H test.c -I/usr/local/include/suricata /usr/local/lib/libsuricata_c.a \
    /usr/local/lib/libsuricata_rust.a /usr/local/lib/libhtp.a -llz4 -lmagic -lcap-ng -lpcap -lnet -ljansson \
    -lpthread -lyaml -lhs -lpcre  -ldl -lrt -lm -lz -lhs  -lpcap
```

We're not far off from being able to combine the 2 .a libraries into a shared library. This does require always building with `-fPIC`, all Rust output is `-fPIC` by default, and while its trivial to add to the C code as well, its causing odd build failures on Mac, Windows, and I think Ubuntu 16.04. So I'm still looking at that, but of course one option is to make it Linux only for now.


jasonish:
> Run `suricata --build-info` to get the `LDFLAGS` required (this should actually be renamed LIBS now that I think about it) and then you can compile the above program with the following command:

TODO: Add a `suricata-config` tool like `python-config`, `pcap-config` where `suricata-config --libs` would output:
```
 -llz4 -lmagic -lcap-ng -lpcap -lnet -ljansson -lpthread -lyaml -lhs -lpcre  -ldl -lrt -lm -lz -lhs  -lpcap
```
and likewise `--cflags` would be:
```
-DHAVE_CONFIG_H
```
(silly example, I think we can do away with that specific one, but you get the idea).

jasonish:
Update at https://github.com/OISF/suricata/pull/5829.

jasonish:
A library user can build with a command like:
```
gcc -fsanitize=address -o test test.c $(/opt/suricata/bin/suricata-config --cflags --libs) /usr/local/lib/libhtp.a
```
TODO: `suricata-config` should be libhtp aware and include it in its output.

victorjulien:
Replaced by #5829

PR #5824, opened 2021-02-07
PostgreSQL: initial support (MVP)

jufajardini:
PostgreSQL: add SSL handshake support

Describe changes:
- PostgreSQL: this is an attempt to have the SSL handshake part of the postgresql FE/BE protocol covered, using the aid of the setup-applayer-py script, as a means to understand how the whole idea of parsing, identifying and processing packets/stream works, with the minimum real-case scenario I could think of. This isn't complete yet, but parses the possible request (SSL encrytion request message) and two of the three possible responses ('S' or 'N', not covering the case of an error message, yet). There is some tentative work with the probing phase, and although there is an error type, it is not being properly used to let the engine know when it show give up and when it should wait for more data.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:


victorjulien:
Do you have a Suricata-Verify test that goes with this?

If not, a pcap to play with it?

jufajardini:
Oh, I haven't thought of adding the suricata-verify, I'll do that next time. 

Here are two pcaps, I don't know if this is doing much, already, though:
- This is just one interaction: https://drive.google.com/file/d/1XKn_UDGdrvGS5euGcZczlfkzy3plP60i/view?usp=sharing
- This has the handshake and then other pgsql messages: https://drive.google.com/file/d/17w1whhEfnWV9w4r7Doqi_zi3OA6TqCJb/view?usp=sharing

rust/src/applayerpgsql/parser.rs
```diff
@@ -0,0 +1,155 @@
...
+    ));
+
+fn parse_len(input: &str) -> Result<u32, std::num::ParseIntError> {
+    input.parse::<u32>()
+}
+
+named!(pub parse_message<String>,
+       do_parse!(
+           len:  map_res!(
+                 map_res!(take_until!(":"), std::str::from_utf8), parse_len) >>
```
victorjulien, line 88: I'm not understanding what is happening here. It parses the 4 byte length field, but then also takes something until a `:` is found? Shouldn't we just to something like `take!(len)`?

When testing with `postgresql-13-simple-select.pcap` I see this return `incomplete` all the time.

jufajardini: Sorry, that's my fault, I haven't worked on this function, but now I can see that I certainly should, before sharing. (This also helps me understand when I can expect to see things "changing" if I run suricata after changing things here, and when not much will happen...) I will do that, and submit a new PR.

jufajardini:
Closed with: https://github.com/OISF/suricata/pull/5830

PR #5825, opened 2021-02-07
Pcap conditional v2.0.0

regit:
This patchset implement conditional pcap logging. Pcap logging can now be triggered for alerted flows and for flow tagged with the `tag` keyword. It uses some concept and code from #5345 for the stream logging that is used to log packets containing the alerting data.

Describe changes:
- Implement conditional logging
- Optionally add packet header to tcp segment
- Use stream data to log alert packet
- Log alerted tunnel packets
- Add tag support for conditional logging

### DIFF
--- a/src/log-pcap.c
+++ b/src/log-pcap.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2014 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -14,7 +14,6 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
  * 02110-1301, USA.
  */
-
 
 /**
  * \file
--- a/src/log-pcap.c
+++ b/src/log-pcap.c
@@ -26,7 +26,9 @@
  */
 
 #include "suricata-common.h"
+#include "util-buffer.h"
 #include "util-fmemopen.h"
+#include "stream-tcp-util.h"
 
 #ifdef HAVE_LIBLZ4
 #include <lz4frame.h>
@@ -186,6 +188,7 @@ typedef struct PcapLogData_ {
 
 typedef struct PcapLogThreadData_ {
     PcapLogData *pcap_log;
+    MemBuffer *buf;
 } PcapLogThreadData;
 
 /* Pattern for extracting timestamp from pcap log files. */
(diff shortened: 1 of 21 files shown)


regit:
Working on fixing tests.

src/log-pcap.c
```diff
@@ -463,6 +497,85 @@ static void PcapLogUnlock(PcapLogData *pl)
...
+        MemBufferWriteRaw(pctx->buf, seg->pcap_hdr_storage->pkt_hdr, seg->pcap_hdr_storage->pktlen);
+        MemBufferWriteRaw(pctx->buf, buf, buflen);
+
+        PcapWrite(pctx->pl, pctx->connp, (uint8_t *)pctx->buf->buffer,
+                seg->pcap_hdr_storage->pktlen + buflen);
+    }
+    return 1;
+}
+
+static void PcapLogDumpSegments(
```
scottfgjordan, line 562: Still working on a full review, but playing around with your branch a bit... It looks like this is only used in a single direction correct? Comparing outputs between our two approaches it looks like this results in a few missing packets. See https://github.com/scottfgjordan/suricata/blob/3af87b26638cfcc7b64eee532a54bbebbc32dc62/src/detect-tag-pcap.c#L244 for how I went about this. I think otherwise the results between the two should be pretty similar.

regit: Yes, this patchset just output one side that will contain the data that did trigger the alert. I still need to be convinced we can have useful data when dumping both sides.

scottfgjordan: I did a small test just comparing the capture results for a basic signature that would alert on an HTTP GET request to compare how the results differ between the two approaches. Links to the captures can be found here: https://wiki.wireshark.org/SampleCaptures (tcp-ecn-sample.pcap and http.pcap). The signature I used was simple: alert tcp any any -> any any (msg:"ALERTING ON HTTP GET REQUEST"; http.method; content:"GET"; tag:session; sid:1; rev:1;).

The ideal results in the situation would be a capture starting from the packet with the GET request and then capturing all packets in the session from that point. The results are as follows:

For Pcap conditional v2.0.0
- Conditional: alerts
  - tcp-ecn-sample.pcap: Grabs the packet that caused the alert (Note: there is a duplication of the packet that causes the alert) then misses the next two packets immediately after this (in the opposite direction) which contain the HTTP OK response. 
  - http.pcap: Grabs the packet that caused the alert (duplicated again) and then misses the next 3 packets (2 in the opposite direction of alert and 1 in the same) resulting in losing out on the HTTP OK response and some additional data.
- Conditional: tag
  - tcp-ecn-sample.pcap: Misses the packet that causes the alert and the following 2 packets. Picks it up from there.
  - http.pcap: Misses the packet that causes the alert and the following 3 packets. Picks it up from there. 
For capture feature 120 v5
No Session Dumping:
- tcp-ecn-sample.pcap:  Misses the packet that causes the alert and the following packet. Picks it up from there
- http.pcap: Misses the packet that causes the alert and the following 2 packets. Picks it up from there. 
Session Dumping Enabled:
- tcp-ecn-sample.pcap: Grabs the packet that causes the alert and everything following.
- http.pcap: Grabs the packet that causes the alert and everything following. 

A small scale example, but hopefully does a bit of convincing you of some the value with dumping the tcp session and the value of doing both directions.

scottfgjordan (commented):
In general I like the approach you took compared to what I had done in https://github.com/OISF/suricata/pull/5345 It was interesting to review to see how I could have better integrated into Suricata's existing code base/functionality and I found the code quality to be high. 

Besides the comments I've given about the tcp-session dumping, the only other thing I would mention is the logging approach. The current approach results in all alerts being dumped to the same pcap file, which I don't think is a very useful way to present the data. I really think that presenting each captured alert as its own pcap alongside an .alert file of the same name in the same directory (which wasn't something I had implemented, but did mention in https://github.com/OISF/suricata/pull/5828) would be the best approach. I know you had concern about many files in a single directory, but I think this could be mitigated as I had shown in my aforementioned pull request, where some organization and file count limits are put in place to prevent an overwhelming number of files in a directory.

In terms of how to proceed, I am willing to provide any assistance you want to complete this feature, so please let me know how you'd like to move forward.

src/log-pcap.c
regit: Ok, thanks a lot for feedback and tests. I'm gonna replay them and try to see what is needed.

victorjulien:
Replaced by #5858

jasonish:
> It uses some concept and code from #5345 for the stream logging that is used to log packets containing the alerting data.

I don't know the Arkime tool well, but do we also provide a way of logging these conditional pcaps in a way that could be picked up by Arkime?

PR #5829, opened 2021-02-09
libsuricata - v3.0

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5822

Changes from last PR:
- Always build with `-fPIC` on Linux to support shared libraries.
- Add Makefile target to build and install shared library.
- Add `suricata-config` program to output `--libs` and `--cflags` for use in build scripts of Suricata libray users.

Buliding the Shared Library:
- We can't use libtool as it doesn't understand the static library output from Rust, as its not a libtool library, so we have to do a bit of a kludge.
- `make solib` and `make solib-install` will build and install the shared library on Linux.
- I consider this a bit of a hack to make the library used now. I hope to come up with something better.


jasonish:
A library user can build with a command like:
```
gcc -fsanitize=address -o test test.c $(/opt/suricata/bin/suricata-config --cflags --libs) /usr/local/lib/libhtp.a
```
TODO: `suricata-config` should be libhtp aware and include it in its output.

victorjulien:
I see a build failure on OpenBSD 6.8
```
checking zlib.h usability... yes
checking zlib.h presence... yes
checking for zlib.h... yes
checking for inflate in -lz... no
   ERROR!  zlib library not found, go get it
   Debian/Ubuntu: apt install zlib1g-dev
   Fedora: dnf install zlib-devel
   CentOS/RHEL: yum install zlib-devel
ERROR: Job failed: exit code 1
```

jasonish:
> I see a build failure on OpenBSD 6.8

Don't suppose there is an OpenBSD 6.8 ready for logging into?

victorjulien:
Not easily, no. This is in `config.log`, I see it also when it checks for plugin support.
Screenshot from 2021-02-10 08-33-28

jasonish:
OpenBSD issue should be fixed here: https://github.com/OISF/suricata/pull/5835

PR #5835, opened 2021-02-10
libsuricata - v3.1

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5829

Changes from last PR:
- Remove the commit that removed RUST_LDADD, it introduces library order issues in some places.
- Some LDFLAGS/LIBS fixups (-llua should be in LIBS)
- Make suricata-config libhtp aware

Compiling my test program against libsuricata now looks like:
```
gcc -o test test.c -fsanitize=address $(/opt/suricata/bin/suricata-config --cflags --libs)
```
(the -fsanitize can be dropped if not built with asan).

Not really a fan of the name `suricata-config`, perhaps `libsuricata-config`?


victorjulien:
Any thoughts on turning https://github.com/OISF/suricata/pull/5835/commits/61fea274a9800b387985be89dcf31312dc68e1c0 into a non-hack?

jasonish:
> Any thoughts on turning https://github.com/OISF/suricata/pull/5835/commits/61fea274a9800b387985be89dcf31312dc68e1c0 into a non-hack?

The autotools way is to unconditionally install all headers. So unless we do that we will have to do some sort of kludge/hack. 

I would like to remove the gmake'ism tho. Then at least it's portable.

victorjulien:
This now fully passed my builds, including OpenBSD.

Can we also have some CI tests for the library install and headers install? At a minimum that headers are not installed by default, are installed when requested? And same for the lib?

jasonish:
> Can we also have some CI tests for the library install and headers install? At a minimum that headers are not installed by default, are installed when requested? And same for the lib?

Should probably not install the static libs by default like they are now.  So will look at that as well.

victorjulien:
Already cherry-picked the first 2 commits.

victorjulien:
Replaced by #5847

PR #5836, opened 2021-02-11
util/thash: fix memcap consolidate function

regit:
The function THashConsolidateMemcap is used to allow to load a dataset even when the memcap is not set. But the implementation was in fact resetting the memcap value to the max of memory usage after loading and default memcap. As a result, the function was resetting memcap to the default memcap even if a huge memcap was set in the dataset definition. In the case of dataset where we add to the set it was leading to memcap limit hitting despite the settings of memcap by the user.

This patch udpates the code to set the final memcap value to the max of memory usage after loading and set memcap.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4275

Describe changes:
- Fix THashConsolidateMemcap function

### DIFF
--- a/src/util-thash.c
+++ b/src/util-thash.c
@@ -336,7 +336,7 @@ THashTableContext *THashInit(const char *cnf_prefix, size_t data_size,
  * */
 void THashConsolidateMemcap(THashTableContext *ctx)
 {
-    ctx->config.memcap = MAX(SC_ATOMIC_GET(ctx->memuse), THASH_DEFAULT_MEMCAP);
+    ctx->config.memcap = MAX(SC_ATOMIC_GET(ctx->memuse), ctx->config.memcap);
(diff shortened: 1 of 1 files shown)


pevma:
Taking it for a spin.

victorjulien:
Codecov suggests this is a line that is not used in any of our tests. We do have dataset tests though. Do you understand what is going on?

regit:
> Codecov suggests this is a line that is not used in any of our tests. We do have dataset tests though. Do you understand what is going on?

The test added in https://github.com/OISF/suricata-verify/pull/322 should cover this line of code.

pevma:
The fix is working as expected - behaves much better

victorjulien:
Merged as part of #5891, thanks.

PR #5837, opened 2021-02-11
Ossfuzz corp v7

catenacyber:

Describe changes:
- Runs CI on public corpuses of fuzz targets for coverage and ASAN bugs
- simple fuzz driver can run on flat directories (not only files)
- fix bug of uninitialized memory in `bytemath` keyword parsing

Modifies #5811 by counting the total number of layers so that too much recursion can't happen anywhere (and adding a field to the structure Packet to do so)

### DIFF
--- a/src/decode-ethernet.c
+++ b/src/decode-ethernet.c
@@ -38,6 +38,8 @@
 #include "util-unittest.h"
 #include "util-debug.h"
 
+#define MAX_ETH_OFFSET 256
+
 int DecodeEthernet(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
                    const uint8_t *pkt, uint32_t len)
 {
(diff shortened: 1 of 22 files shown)


src/decode-ethernet.c
```diff
@@ -38,6 +38,8 @@
 #include "util-unittest.h"
 #include "util-debug.h"
 
+#define MAX_ETH_OFFSET 256
```
victorjulien, line 41: is this still used?

src/decode.h
```diff
@@ -887,6 +891,15 @@ void CaptureStatsSetup(ThreadVars *tv, CaptureStats *s);
      ((p)->action |= a)); \
 } while (0)
 
+#define PKT_MAX_DECODED_LAYERS 16
+
+#define PACKET_INCREASE_CHECK_LAYERS(p) do {                                        \
```
victorjulien, line 896: I think adding a macro to hides a return statement inside is not a great solution. I understand the desire to keep things tidy, but this is not the way.

src/decode-ethernet.c
catenacyber: Indeed, removing in next PR

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5838

src/decode.h
catenacyber: ok :-)

PR #5802, opened 2021-01-29
Improve midstream pickup of TCP flows at SYN/ACK stage

rtkjweeks:
If midstream pickup is enabled for suricata, the first packet seen for
a flow is assumed to come from the client endpoint. This approach has a
50% chance of being correct. If the first packet to be seen for a TCP
flow is a SYN/ACK then it must have been sent by the server endpoint,
not the client. This change leverages that fact to improve the chances
of flows that are picked up midstream having the correct cardinality.

The FlowInit method now checks to see if the first packet is a TCP
SYN/ACK and, if it is, sets the source and destination addresses for
the flow accordingly. This subsequently means that the
StreamTcpPacketSwitchDir method is no longer required, which in turn
*should* slightly reduce the computational overhead for affected flows.

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
-
-
-

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5844, opened 2021-02-12
Improve midstream pickup of TCP flows at SYN/ACK stage

rtkjweeks:
If midstream pickup is enabled for suricata, the first packet seen for
a flow is assumed to come from the client endpoint. This approach has a
50% chance of being correct. If the first packet to be seen for a TCP
flow is a SYN/ACK then it must have been sent by the server endpoint,
not the client. This change leverages that fact to improve the chances
of flows that are picked up midstream having the correct cardinality.

The FlowInit method now checks to see if the first packet is a TCP
SYN/ACK and, if it is, sets the source and destination addresses for
the flow accordingly. This subsequently means that the
PacketSwap/FlowSwap methods are no longer required, which in turn
*should* slightly reduce the computational overhead for affected flows.

- [ ] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/

Previous PR:
https://github.com/OISF/suricata/pull/5802

Describe changes:
-
-
-

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/stream-tcp.c
+++ b/src/stream-tcp.c
@@ -933,11 +933,6 @@ static int StreamTcpPacketStateNone(ThreadVars *tv, Packet *p,
             StatsIncr(tv, stt->counter_tcp_midstream_pickups);
         }
 
-        /* reverse packet and flow */
-        SCLogDebug("reversing flow and packet");
-        PacketSwap(p);
-        FlowSwap(p->flow);
-
         /* set the state */
         StreamTcpPacketSetState(p, ssn, TCP_SYN_RECV);
         SCLogDebug("ssn %p: =~ midstream picked ssn state is now "
--- a/src/flow-timeout.c
+++ b/src/flow-timeout.c
@@ -348,7 +348,9 @@ int FlowForceReassemblyNeedReassembly(Flow *f)
  */
 void FlowForceReassemblyForFlow(Flow *f)
 {
-    const int thread_id = (int)f->thread_id[0];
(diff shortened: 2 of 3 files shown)


src/stream-tcp.c
```diff
@@ -933,11 +933,6 @@ static int StreamTcpPacketStateNone(ThreadVars *tv, Packet *p,
             StatsIncr(tv, stt->counter_tcp_midstream_pickups);
         }
 
-        /* reverse packet and flow */
```
catenacyber, line 936: Why do these lines get removed ?

By the way, there is one CI failure about S-V test http-sticky on Fedora... to be investigated

rtkjweeks: They're no longer required; in the case that we start a flow on the SYN-ACK the flow is now created in the proper client->server direction, rather then server->client, which required this flipping after the fact.

catenacyber: ok thanks

rtkjweeks:
Test failures are unrelated; not sure how I'm supposed to handle those (I think they're fixed in master; should I rebase them in?  If so, do I need to create a new PR as well?)

catenacyber:
Happy new year and friendly ping @rtkjweeks

victorjulien (changes requested):
Needs a rebase, thanks!

PR #5847, opened 2021-02-12
libsuricata - v4.3

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/5835

Changes from last PR:
- rename suricata-config to libsuricata-config
- only install libsuricata-config on "make install-library"
- use a more portable way of installing headers

To install the headers: `make install-headers`
To install the library: `make install-library`

The shared library is only enabled on Linux and will be installed by `make install-library` along with the static libs.


.gitignore
```diff
@@ -65,3 +65,4 @@ test.sh
 .vscode/*
 /suricata-update/*
 !/suricata-update/Makefile.am
+/suricata-config
```
victorjulien, line 68: should this now be changed to `/libsuricata-config`?

jasonish: Yes

jasonish:
Continued at https://github.com/OISF/suricata/pull/5866.

PR #5854, opened 2021-02-13
detect/content: Consider distance in validation

jlucovsky:
This commit modifies the validation callback to include the distance
during validation.

Values of distance that cause the right edge to be exceeded are
considered an error and the signature will be rejected.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [2982](https://redmine.openinfosecfoundation.org/issues/2982)

Describe changes:
- Incorporate `distance` into the validation check.

suricata-verify-pr: 436
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-content.c
+++ b/src/detect-content.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2019 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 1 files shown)


src/detect-content.c
```diff
@@ -407,21 +407,22 @@ bool DetectContentPMATCHValidateCallback(const Signature *s)
...
-        uint32_t right_edge = cd->content_len + cd->offset;
         if (cd->content_len > max_right_edge) {
             SCLogError(SC_ERR_INVALID_SIGNATURE,
                     "signature can't match as content length %u is bigger than dsize %u.",
                     cd->content_len, max_right_edge);
-            return FALSE;
+            return false;
         }
+        /* Incorporate the distance when comparing with the right edge */
+        uint32_t right_edge = cd->content_len + cd->offset + cd->distance;
```
victorjulien, line 417: If there is  distance keyword, it means there is a previous content keyword as well. Ideally we'd also take this into account. I think ultimately we'd need logic to take all the relevant keywords and consider them all. But this would be quite complex.

Minor note here: distance is signed and can be negative. It can be used to look before the current pattern I think, so it might lead to underflow of `right_edge`

jlucovsky:
Continued in #5881

PR #5863, opened 2021-02-15
protocol detection: midstream - set done flag only if parsers for bot…

ilya-bakhtin:
…h directions are not found

  in a case of midstream parsers from other direction are tried if nothing is found for the initial one.
  done flag must be set if nothing is found in both directions.
  otherwise processing of incomplete data is terminated at the very first try.

Currently, done flag is set when parsers for initial directions are not found. It prevents further protocol detection on a flow that starts from the server message and this message is split between TCP segments. Normally probing parser returns incomplete and detection is retried until the max depth of probing parser is reached. But in the case of midstream, the direction is immediately marked as done so detection is terminated.

The fix allows detection from midstream to continue the same way as for normally started flow.

The test is already prepared in suricata-verify with the same branch name midstream-detect

suricata-verify-pr: 438
#suricata-verify-repo:
#suricata-verify-branch: midstream-detect
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -534,6 +534,7 @@ static AppProto AppLayerProtoDetectPPGetProto(Flow *f,
     uint8_t dir = idir;
     uint16_t dp = f->protodetect_dp ? f->protodetect_dp : FLOW_GET_DP(f);
     uint16_t sp = FLOW_GET_SP(f);
+    bool probe_is_found = false;
 
 again_midstream:
     if (idir != dir) {
@@ -596,9 +597,9 @@ again_midstream:
     if (pe1 == NULL && pe2 == NULL && pe0 == NULL) {
         SCLogDebug("%s - No probing parsers found for either port",
                 (dir == STREAM_TOSERVER) ? "toserver":"toclient");
-        if (dir == idir)
-            FLOW_SET_PP_DONE(f, dir);
         goto noparsers;
+    } else {
+        probe_is_found = true;
     }
 
     /* run the parser(s): always call with original direction */
@@ -634,8 +635,8 @@ again_midstream:
         }
     }
 
- noparsers:
-    if (stream_config.midstream == true && idir == dir) {
+noparsers:
+    if (stream_config.midstream && idir == dir) {
         if (idir == STREAM_TOSERVER) {
             dir = STREAM_TOCLIENT;
         } else {
@@ -644,6 +645,8 @@ again_midstream:
         SCLogDebug("no match + midstream, retry the other direction %s",
                 (dir == STREAM_TOSERVER) ? "toserver" : "toclient");
         goto again_midstream;
+    } else if (!probe_is_found) {
+        FLOW_SET_PP_DONE(f, idir);
     }
 
  end:
--- a/src/stream-tcp.c
+++ b/src/stream-tcp.c
@@ -418,7 +418,9 @@ void StreamTcpInitConfig(char quiet)
         SCLogConfig("stream \"memcap\": %"PRIu64, SC_ATOMIC_GET(stream_config.memcap));
     }
 
-    ConfGetBool("stream.midstream", &stream_config.midstream);
+    int imidstream;
+    ConfGetBool("stream.midstream", &imidstream);
+    stream_config.midstream = imidstream != 0;
 
     if (!quiet) {
         SCLogConfig("stream \"midstream\" session pickups: %s", stream_config.midstream ? "enabled" : "disabled");
@@ -919,8 +921,7 @@ static int StreamTcpPacketStateNone(ThreadVars *tv, Packet *p,
 
     /* SYN/ACK */
     } else if ((p->tcph->th_flags & (TH_SYN|TH_ACK)) == (TH_SYN|TH_ACK)) {
-        if (stream_config.midstream == FALSE &&
-                stream_config.async_oneside == FALSE)
+        if (!stream_config.midstream && stream_config.async_oneside == FALSE)
             return 0;
 
         if (ssn == NULL) {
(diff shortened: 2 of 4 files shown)


ilya-bakhtin:
It seems something is wrong with http-sticky-start test on F32 and F33.
It's not properly performed on a set of absolutely different PRs.

catenacyber:
> It seems something is wrong with http-sticky-start test on F32 and F33.

Yes, these tests fail with some kind of timeout for every PR

victorjulien:
I started this https://forum.suricata.io/t/github-ci-status/1095/5

src/app-layer-detect-proto.c
```diff
@@ -634,8 +635,8 @@ static AppProto AppLayerProtoDetectPPGetProto(Flow *f,
         }
     }
 
- noparsers:
-    if (stream_config.midstream == true && idir == dir) {
+noparsers:
+    if (stream_config.midstream != false && idir == dir) {
```
victorjulien, line 639: `!= false` instead of `== true`?

ilya-bakhtin: Sure.
In this place, it probably does not matter. But false is exactly zero and true is exactly 1.
stream_config.midstream is declared as int.
Any nonzero value is treated as true but e.g. 123 != true literally since true is 1.
So if we are writing in C without embedded bool type then '!= false' is much more correct than '== true'

victorjulien: I quick scan of the code suggests its always set to 1 or 0, so we should probably convert the type to `bool` and use it accordingly.

ilya-bakhtin: Anyhow a condition "if (some_var == true)" looks confusing and dangerous in C language.
I would prefer "if (some_var != false)" or simply "if (some_var)"

I did not quite understand if you want me to apply the mentioned stream_config.midstream type change. I definitely could do that in the scope of this PR, however it makes sense only if the probing parsers change itself looks good for you.

catenacyber:
This looks relevant.
Do you have a Suricata-verify test for this ?

src/app-layer-detect-proto.c
```diff
@@ -534,6 +534,7 @@ static AppProto AppLayerProtoDetectPPGetProto(Flow *f,
     uint8_t dir = idir;
     uint16_t dp = f->protodetect_dp ? f->protodetect_dp : FLOW_GET_DP(f);
     uint16_t sp = FLOW_GET_SP(f);
+    int found_probes = 0;
```
catenacyber, line 537: Should we use a simple bool ?

src/app-layer-detect-proto.c
catenacyber: I think it makes sens to do the change (convert `stream_config.midstream` from int to bool) in this PR (in another commit). 

> however it makes sense only if the probing parsers change itself looks good for you.

indeed, I would like to have a suricata-verify test first

catenacyber (changes requested):
Needs a suricata-verify test for checking

ilya-bakhtin:
> Needs a suricata-verify test for checking

It's already available https://github.com/OISF/suricata-verify/pull/438

src/app-layer-detect-proto.c
ilya-bakhtin: It's already available OISF/suricata-verify#438

src/app-layer-detect-proto.c
ilya-bakhtin: sure bool is enough

src/app-layer-detect-proto.c
catenacyber: Thanks, can you mention it in the PR comment
Instead of leaving `#suricata-verify-pr:`, please put `suricata-verify-pr: 438`
This way, CI uses this S-V PR to run the tests :-)

catenacyber (changes requested):
Looks quite good
Just need some nits to be fixed and make the new PR run CI with the corresponding S-V test

ilya-bakhtin:
well, http-sticky-start does not work so far but 'suricata-verify-pr: 438' does :-)

catenacyber (approved):
Thanks, looks good to me

ilya-bakhtin:
Thanks,
So i've added another commit changing stream_config.midstream type to bool

src/stream-tcp.c
```diff
@@ -418,7 +418,9 @@ void StreamTcpInitConfig(char quiet)
         SCLogConfig("stream \"memcap\": %"PRIu64, SC_ATOMIC_GET(stream_config.memcap));
     }
 
-    ConfGetBool("stream.midstream", &stream_config.midstream);
+    int imidstream;
+    ConfGetBool("stream.midstream", &imidstream);
```
catenacyber, line 422: Thanks for this...
Should `ConfGetBool` use a bool instead ?
Maybe not for this PR, but we can file a redmine ticket about this

ilya-bakhtin: Well, it could be good to modify ConfGetBool so.
However, it will result in massive changes in corresponding variables/structure members types. Also, it would be reasonable to get rid of conditions (something ==/!= TRUE) and change them to (something) or (!something)
So the proposed modification is good but it should be made in the scope of a special ticket and very carefully.

catenacyber: Thanks for your comment

victorjulien:
I'm merging this in my release staging branch, thanks.

In future PRs can you please make sure the commit messages are compliant with our guidelines and also create a new branch+PR for updates? Thanks.

victorjulien:
Merged as part of #5891, thanks.

PR #5872, opened 2021-02-18
Http range file v7

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/1576

Describe changes:
- Uses a generic container (a thread-safe hash table whose key is the url/filename) to store file parts with HTTP Range header
- handles reassembly of these parts, while enforcing a global memcap

I bet there is a lot to say about this... comments to follow in the code

suricata-verify-pr: 443

PR #5873, opened 2021-02-18
Http range file v8

catenacyber:

Describe changes:

I bet there is a lot to say about this... comments to follow in the code

suricata-verify-pr: 443

Modifies #5872 with
- clang-format
- rebase
- use URL (and not just URI) as hash table key
- do not bother to reassemble if one range is complete (made S-V test http-evader 501 fail because we matched twice instead of once)


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5874

PR #5874, opened 2021-02-18
Http range file v9

catenacyber:

Describe changes:

I bet there is a lot to say about this... comments to follow in the code

suricata-verify-pr: 443

Modifies #5873 with
- fixing cocci by checking malloc returns
- fixing windows by casting `tv_sec` into uint32


src/app-layer-htp-file.c
```diff
@@ -225,34 +228,108 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+        const uint8_t *data, uint32_t data_len, uint64_t txid, bstr *rawvalue, HtpTxUserData *htud)
 {
     SCEnter();
+    uint16_t flags;
 
     if (s == NULL) {
         SCReturnInt(-1);
     }
 
+    HtpContentRange crparsed;
```
victorjulien, line 241: should this whole logic be wrapped in a `if (STREAM_TOCLIENT) {`? It seems all calls to `HTPFileOpen` assume toclient

src/app-layer-htp-file.c
```diff
@@ -225,34 +228,108 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+    if (HTPParseContentRange(rawvalue, &crparsed) != 0) {
```
victorjulien, line 242: Can this parse and check func move into a util func?

src/app-layer-htp-file.c
```diff
@@ -225,34 +228,108 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
+        keyurl = SCMalloc(keylen);
+        if (keyurl == NULL) {
+            SCReturnInt(-1);
+        }
+        memcpy(keyurl, bstr_ptr(tx->request_hostname), bstr_len(tx->request_hostname));
+        memcpy(keyurl + bstr_len(tx->request_hostname), filename, filename_len);
+        keyurl[keylen - 1] = 0;
+    } else {
+        // do not reassembly file without host info
+        SCReturnInt(-1);
```
victorjulien, line 309: what does this mean and what should the consequences be?

src/app-layer-htp-file.c
```diff
@@ -225,34 +228,108 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
     }
-    SCReturnInt(retval);
+    s->file_range_container = ContainerUrlRangeGet(keyurl, keylen, &s->f->lastts);
```
victorjulien, line 311: is the filename the only thing we could use as a key? I think in the past I've looked at using the etag header value for this. It seems a file name could easily lead to collisions, esp if we're protecting a server that serves files to many clients.

src/app-layer-htp.c
```diff
@@ -99,113 +99,74 @@ static uint64_t htp_state_memuse = 0;
...
-        HTTP_DECODER_EVENT_GZIP_DECOMPRESSION_FAILED},
-    { "REQUEST_FIELD_MISSING_COLON",
-        HTTP_DECODER_EVENT_REQUEST_FIELD_MISSING_COLON},
-    { "RESPONSE_FIELD_MISSING_COLON",
-        HTTP_DECODER_EVENT_RESPONSE_FIELD_MISSING_COLON},
-    { "INVALID_REQUEST_CHUNK_LEN",
-        HTTP_DECODER_EVENT_INVALID_REQUEST_CHUNK_LEN},
-    { "INVALID_RESPONSE_CHUNK_LEN",
-        HTTP_DECODER_EVENT_INVALID_RESPONSE_CHUNK_LEN},
+SCEnumCharMap http_decoder_event_table[] = {
```
victorjulien, line 102: please put the reformat into its own commit

src/containers.c
```diff
@@ -0,0 +1,314 @@
...
+
+static int ContainerUrlRangeSet(void *dst, void *src)
+{
+    ContainerUrlRange *src_s = src;
+    ContainerUrlRange *dst_s = dst;
+    dst_s->len = src_s->len;
+    dst_s->key = SCMalloc(dst_s->len);
+    BUG_ON(dst_s->key == NULL);
+    memcpy(dst_s->key, src_s->key, dst_s->len);
+    dst_s->files = FileContainerAlloc();
```
victorjulien, line 50: error check

src/containers.c
```diff
@@ -0,0 +1,314 @@
...
+#include "util-misc.h"        //ParseSizeStringU64
+#include "util-thash.h"       //HashTable
+#include "util-memcmp.h"      //SCBufferCmp
+#include "util-hash-string.h" //StringHashDjb2
+#include "util-validate.h"    //DEBUG_VALIDATE_BUG_ON
+
+typedef struct ContainerList {
+    THashTableContext *ht;
+    uint32_t timeout;
+} ContainerList;
```
victorjulien, line 35: this name is confusing. It's a hash table, not a list

src/containers.c
```diff
@@ -0,0 +1,314 @@
...
+                    h->next->prev = h->prev;
+                if (hb->head == h)
+                    hb->head = h->next;
+                if (hb->tail == h)
+                    hb->tail = h->prev;
+                h->next = NULL;
+                h->prev = NULL;
+                ContainerUrlRange *c = h->data;
+                FileCloseFile(c->files, NULL, 0, FILE_TRUNCATED);
+                // TODOask can we log it somehow ?
```
victorjulien, line 173: Not easily. In flow logging we have logic for this, but that is not trivial. Flow hash handling has also been redesigned after this thash thing was based on it.

src/containers.c
```diff
@@ -0,0 +1,314 @@
...
+    if (!(THASH_CHECK_MEMCAP(ContainerUrlRangeList.ht, buflen))) {
+        // TODOask release memory for others RangeContainerFree(c);
+        // skips this range
+        c->toskip = buflen;
+        return -1;
+    }
+    (void)SC_ATOMIC_ADD(ContainerUrlRangeList.ht->memuse, buflen);
+    RangeContainer *range = SCCalloc(1, sizeof(RangeContainer));
+    BUG_ON(range == NULL);
+    range->buffer = SCMalloc(buflen);
```
victorjulien, line 228: error handling

src/containers.c
```diff
@@ -0,0 +1,314 @@
...
+        range->next = next->next;
+        next->next = range;
+    }
+    c->current = range;
+    return 0;
+}
+
+int ContainerUrlRangeAppendData(ContainerUrlRange *c, const uint8_t *data, size_t len)
+{
+    if (c->current) {
```
victorjulien, line 249: can you document what is happening in this function? It's not clear what c->current, c->toskip and the default case mean here

src/containers.c
```diff
@@ -0,0 +1,314 @@
...
+        c->current = NULL;
+        return NULL;
+    } // else {
+    File *f = c->files->tail;
+
+    // just finished appending to a file, have we reached a saved range ?
+    RangeContainer *range = c->ranges;
+    while (range && f->size >= range->start) {
+        if (f->size == range->start) {
+            FileAppendData(c->files, range->buffer, range->offset);
```
victorjulien, line 288: error check?

src/containers.h
```diff
@@ -0,0 +1,51 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+#ifndef __CONTAINERS_H__
```
victorjulien, line 18: I'm not a fan of this very generic `Container` naming for something that is quick specific. Would suggest `RangeFile` or `ByteRangeFiles` or something along those lines so it is more descriptive.

src/containers.h
```diff
@@ -0,0 +1,51 @@
...
+#define __CONTAINERS_H__
+
+void ContainersInit(void);
+void ContainersDestroy(void);
+uint32_t ContainersTimeoutHash(struct timeval *ts);
+
+void *ContainerUrlRangeGet(const uint8_t *key, size_t keylen, struct timeval *ts);
+
+typedef struct RangeContainer {
```
victorjulien, line 27: please add docs to the fields in these structs

src/containers.c
```diff
@@ -0,0 +1,314 @@
...
+    ContainerUrlRange lookup;
+    // cast so as not to have const in the structure
+    lookup.key = (uint8_t *)key;
+    lookup.len = keylen;
+    struct THashDataGetResult res = THashGetFromHash(ContainerUrlRangeList.ht, &lookup);
+    if (res.data) {
+        // nothing more to do if (res.is_new)
+        ContainerUrlRangeUpdate(res.data->data, ts->tv_sec + ContainerUrlRangeList.timeout);
+        THashDecrUsecnt(res.data);
+        THashDataUnlock(res.data);
```
victorjulien, line 196: data is now no long protected by lock, but caller gets a reference to it and proceeds to do things with it. This seems to be a thread safety issue.

victorjulien (changes requested):
First review round. My biggest issue is that I don't see how thread safety is handled.

src/app-layer-htp-file.c
catenacyber: `HTPFileOpenWithRange` is only called from `HtpResponseBodyHandle` so only `STREAM_TOCLIENT

src/app-layer-htp-file.c
catenacyber: sure

src/app-layer-htp-file.c
catenacyber: If we do not have the host name, we should fall back to `HTPFileOpen` (instead of returning -1, thanks)

src/app-layer-htp-file.c
catenacyber: It is the URL, not the file name...
Do you think it is relevant ?

src/containers.c
catenacyber: ok

src/containers.c
catenacyber: ok

src/containers.c
catenacyber: ok

src/containers.h
catenacyber: Indeed, I was hesitating between genericity and specificity...
Should I just keep things specific and not think about containers for other keys/values ?

src/containers.c
catenacyber: ok

src/containers.c
catenacyber: trying something

src/containers.h
catenacyber: ok

src/app-layer-htp.c
catenacyber: ok

src/containers.c
catenacyber: You are right.
There are 2 problems : 
- the fact that `ContainerUrlRange` has no inner lock/mutex and that multiple flows may use it (read/write)
Solution is to add a lock to the structure and use it
- The fact that the item of the hash table is saved in a flow structure cf `file_range_container` and that it may timeout and get removed/freed of the hash table while remaining in the flow structure
Solution may be to add a number of references to the structure

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5979

src/containers.c
catenacyber: Done the easy part about the number of references

Now remains the hard part with multiple flows writing concurrently to the file's ranges
It looks like `ContainerUrlRange` needs to be split in 2 :
- one in the hash table where I should open and close a range
- one lockless structure per flow for the current range state (ie fields `toskip`, `current`) when appending data

PR #5888, opened 2021-02-22
Bug 4198/v3

inashivb:

suricata-verify-pr: 448

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -16,7 +16,7 @@
  */
 
 use std::mem::transmute;
-use crate::applayer::{AppLayerResult, AppLayerTxData};
+use crate::applayer::*;
 use crate::core::{self, sc_detect_engine_state_free};
 use crate::dcerpc::parser;
 use nom::error::ErrorKind;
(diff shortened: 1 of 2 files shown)


rust/src/dcerpc/dcerpc.rs
```diff
@@ -1197,7 +1205,15 @@ pub extern "C" fn rs_dcerpc_parse_response(
     if input_len > 0 {
         if input != std::ptr::null_mut() {
             let buf = build_slice!(input, input_len as usize);
-            return state.handle_input_data(buf, core::STREAM_TOCLIENT);
+            let ret = state.handle_input_data(buf, core::STREAM_TOCLIENT);
+            let tx_id = state.tx_id - 1;
```
victorjulien, line 1209: Is this correct? I would expect a tx to be created on the request side, not the response. So here I would expect the `state.tx_id` to stay the same.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -1177,14 +1177,22 @@ pub extern "C" fn rs_dcerpc_parse_request(
     }
     if input_len > 0 && input != std::ptr::null_mut() {
         let buf = build_slice!(input, input_len as usize);
-        return state.handle_input_data(buf, core::STREAM_TOSERVER);
+        let ret = state.handle_input_data(buf, core::STREAM_TOSERVER);
+        let tx_id = state.tx_id - 1;
+        let req_done = if let Some(tx) = state.get_tx(tx_id) {
```
victorjulien, line 1182: for the record: if this works it needs to be cleaned up to avoid the tx lookup

inashivb: OK. On it.

inashivb:
Replaced with #5889

PR #5889, opened 2021-02-22
Bug 4198/v4

inashivb:

suricata-verify-pr: 449

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -350,6 +350,8 @@ pub struct DCERPCState {
     pub tc_ssn_gap: bool,
     pub ts_ssn_trunc: bool, /// true if Truncated in this direction
     pub tc_ssn_trunc: bool,
+    pub req_done: bool,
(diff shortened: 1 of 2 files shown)


rust/src/dcerpc/dcerpc.rs
```diff
@@ -376,6 +378,8 @@ impl DCERPCState {
             tc_ssn_gap: false,
             ts_ssn_trunc: false,
             tc_ssn_trunc: false,
+            req_done: false,
```
victorjulien, line 381: they are initialized to `false` but can never again be false after they've once been `true`.

In general I don't like this indirect approach. I would much prefer a solution where `AppLayerParserTriggerRawStreamReassembly` is called the moment a req or resp is marked as done. So probably a small method that calls `AppLayerParserTriggerRawStreamReassembly` when `tx::req_done` is set?

inashivb: OK thanks.

victorjulien:
Replaced by #5894

PR #5893, opened 2021-02-22
protodetect: improve midstream handling

victorjulien:
Set "done flag" only if parsers for both directions are not found in a
case of midstream parsers from other direction are tried if nothing is found
for the initial one. "done flag" must be set if nothing is found in both
directions. Otherwise processing of incomplete data is terminated at the very
first try.

(cherry picked from commit 5285163d8f31dc89a4ab96b0842099f9792e29be)

suricata-verify-pr: 451


victorjulien:
Replaced by #5896

PR #5850, opened 2021-02-12
decode: limits the number of decoded layers

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4125

Describe changes:
- fix bug by avoiding over recursion between `DecodeEthernet` and `DecodeMPLS`

Modifies #5838 by rebasing and keeping the last commit to be merged

PR #5904, opened 2021-02-23
Fuzz limit decoded layers v2

catenacyber:

Describe changes:

Modifies #5850 by setting an event

suricata-verify-pr: 455

https://github.com/OISF/suricata-verify/pull/455


src/decode-esp.c
```diff
@@ -65,6 +65,9 @@ int DecodeESP(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p, const uint8_t *p
         CLEAR_ESP_PACKET(p);
         return TM_ECODE_FAILED;
     }
+    if (!PacketIncreaseCheckLayers(p)) {
+        return TM_ECODE_FAILED;
```
victorjulien, line 69: In the error path above we call `CLEAR_ESP_PACKET`. We should probably do the same here.

victorjulien: In the decoders before this we call this before the actual decoding is done, but here it is run after `DecodeESPPacket`. Should we check before?

src/decode-ipv4.c
```diff
@@ -527,6 +527,9 @@ int DecodeIPV4(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
         CLEAR_IPV4_PACKET((p));
         return TM_ECODE_FAILED;
     }
+    if (!PacketIncreaseCheckLayers(p)) {
+        return TM_ECODE_FAILED;
```
victorjulien, line 531: call before `DecodeIPV4Packet`? If not at least call `CLEAR_IPV4_PACKET`?

src/decode-ipv6.c
```diff
@@ -571,6 +571,9 @@ int DecodeIPV6(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p, const uint8_t *
         CLEAR_IPV6_PACKET(p);
         return TM_ECODE_FAILED;
     }
+    if (!PacketIncreaseCheckLayers(p)) {
+        return TM_ECODE_FAILED;
```
victorjulien, line 575: same questions as for ipv4

src/decode-template.c
```diff
@@ -62,6 +62,9 @@ int DecodeTEMPLATE(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
         //ENGINE_SET_EVENT(p,TEMPLATE_HEADER_TOO_SMALL);
         return TM_ECODE_FAILED;
     }
+    if (!PacketIncreaseCheckLayers(p)) {
```
victorjulien, line 65: A comment explaining what is happening would be good here. The template should be verbose about such things.

src/decode.h
```diff
@@ -610,6 +610,9 @@ typedef struct Packet_
      */
     struct PktPool_ *pool;
 
+    /* make sure we have no too deep recursion */
```
victorjulien, line 613: Would like to see something that explains more: "count layers of packet decoding, to avoid loops and too many layers cause issues with performance and stability (stack exhaustion)"

src/decode.h
```diff
@@ -1161,6 +1164,18 @@ void DecodeUnregisterCounters(void);
 
 #define PKT_SET_SRC(p, src_val) ((p)->pkt_src = src_val)
 
+#define PKT_MAX_DECODED_LAYERS 16
```
victorjulien, line 1167: How did you come to 16?

src/decode-esp.c
catenacyber: > Should we check before?

Looks better indeed

src/decode-ipv4.c
catenacyber: indeed before

src/decode-ipv6.c
catenacyber: same as ip4, should be done before

src/decode-template.c
catenacyber: ok

src/decode.h
catenacyber: Looked reasonable to me, nothing scientific...
Do you have a better idea ?

src/decode.h
catenacyber: ok

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5917

PR #5906, opened 2021-02-23
Bug 4198/v8

inashivb:

suricata-verify-pr: 449

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -17,7 +17,7 @@
 
 use std::mem::transmute;
 use crate::applayer::*;
-use crate::core::{self, sc_detect_engine_state_free};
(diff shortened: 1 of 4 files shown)


rust/src/dcerpc/dcerpc.rs
```diff
@@ -1197,6 +1218,7 @@ pub extern "C" fn rs_dcerpc_parse_response(
     if input_len > 0 {
         if input != std::ptr::null_mut() {
             let buf = build_slice!(input, input_len as usize);
+            state.flow = Some(flow);
```
jasonish, line 1221: The flow would also be the same on subsequent calls right? Maybe we should do:
```
if state.flow.is_none() {
    state.flow = Some(flow);
```

victorjulien:
Merged in #5913, thanks!

PR #5924, opened 2021-02-24
Next/20210224/v2

victorjulien:
#5922 with formatting fixups

suricata-verify-pr: 460


victorjulien:
Went with a different solution in #5932

PR #5923, opened 2021-02-24
Detect prefilter setupu8 4335 v5

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4112
https://redmine.openinfosecfoundation.org/issues/4335

Describe changes:
- Adds error for impossible rules such as >255 for uint8 fields
- converts icode to use the generic (many more keywords to do)
  * allows <> syntax for uint ranges to support existing syntax
  * corrects reverted ranges and issues a warning (so 20<>8 becomes 8<>20 ) to support existing syntax

Modifies #5920 by getting CI to pass with //fallthrough comment and using the new test

PR #5937, opened 2021-02-26
Detect prefilter setupu8 4335 v6

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4335

Describe changes:
- Adds error for impossible rules such as >255 for uint8 fields
- converts icode to use the generic (many more keywords to do)
  * allows <> syntax for uint ranges to support existing syntax
  * corrects reverted ranges and issues a warning (so 20<>8 becomes 8<>20 ) to support existing syntax

Modifies #5923 by formatting and rewording commit

### DIFF
--- a/src/detect-engine-uint.c
+++ b/src/detect-engine-uint.c
@@ -151,24 +151,33 @@ DetectU32Data *DetectU32Parse (const char *u32str)
         switch(arg2[0]) {
             case '<':
             case '>':
-                if (strlen(arg3) == 0)
-                    return NULL;
+                if (strlen(arg2) == 1) {
+                    if (strlen(arg3) == 0)
+                        return NULL;
 
-                if (ByteExtractStringUint32(&u32da.arg1, 10, strlen(arg3), arg3) < 0) {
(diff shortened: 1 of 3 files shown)


src/detect-engine-uint.c
```diff
@@ -373,6 +454,10 @@ DetectU8Data *DetectU8Parse (const char *u8str)
             return NULL;
         }
     }
+    if (DetectU8Validate(&u8da)) {
+        SCLogError(SC_ERR_INVALID_VALUE, "Impossible value for uint8 condition");
```
victorjulien, line 458: can we provide a more helpful error here? Something that includes the keyword name and perhaps also the original value?

catenacyber: Ok for the original value.
For the keyword name, should it be the caller of `DetectU32Parse` (such as `DetectHTTP2windowSetup`) which should print another message or should it transmit the keyword name to `DetectU32Parse` just for this debugging purpose ?

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5966

PR #5941, opened 2021-02-28
ikev1: add ikev1 parser - v15

satta:
Previous PR: https://github.com/OISF/suricata/pull/5711

Describe changes:
- Rebase against master and resolve conflicts.

suricata-verify-pr: 465


victorjulien:
Can you fix the formatting differently? Instead of a fix commit do:
`git rebase master -x 'git clang-format HEAD~'`. This will fix the formatting per commit.

src/Makefile.am
```diff
@@ -12,7 +12,7 @@ noinst_HEADERS = \
 	alert-fastlog.h \
 	alert-prelude.h \
 	alert-syslog.h \
-	app-layer-dcerpc-common.h \
```
victorjulien, line 15: this looks suspicious

rules/ipsec-events.rules
```diff
@@ -4,16 +4,16 @@
...
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 no encryption (AH)"; flow:to_client; app-layer-event:ikev2.no_encryption; classtype:protocol-command-decode; sid:2224008; rev:1;)
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 invalid proposal"; flow:to_server; app-layer-event:ikev2.invalid_proposal; classtype:protocol-command-decode; sid:2224009; rev:1;)
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 invalid proposal selected"; flow:to_client; app-layer-event:ikev2.invalid_proposal; classtype:protocol-command-decode; sid:2224010; rev:1;)
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 unknown proposal"; flow:to_server; app-layer-event:ikev2.unknown_proposal; classtype:protocol-command-decode; sid:2224011; rev:1;)
-alert ikev2 any any -> any any (msg:"SURICATA IKEv2 unknown proposal selected"; flow:to_client; app-layer-event:ikev2.unknown_proposal; classtype:protocol-command-decode; sid:2224012; rev:1;)
+alert ike any any -> any any (msg:"SURICATA IKE malformed request data"; flow:to_server; app-layer-event:ike.malformed_data; classtype:protocol-command-decode; sid:2224000; rev:1;)
```
victorjulien, line 7: this sigs should all have a rev bump with the change

rust/src/common.rs
```diff
@@ -72,3 +72,12 @@ pub unsafe extern "C" fn rs_cstring_free(s: *mut c_char) {
     }
     drop(CString::from_raw(s));
 }
+
+/// Convert an u8-array of data into a hexadecimal representation
+pub fn to_hex(input: &[u8]) -> String {
```
victorjulien, line 77: @jasonish any thoughts on this name?

rust/src/ike/detect.rs
```diff
@@ -0,0 +1,227 @@
...
+    unsafe {
+        *buf = ptr::null();
+        *len = 0;
+    }
+
+    return 0;
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_ike_state_get_sa_attribute(
```
victorjulien, line 139: for an unsafe function, this is a bit long. Can we split the meat of the function into a safe function?

rust/src/ike/ike.rs
```diff
@@ -54,547 +62,402 @@ impl IkeEvent {
...
+#[no_mangle]
+pub extern "C" fn rs_ike_probing_parser(
+    _flow: *const Flow, direction: u8, input: *const u8, input_len: u32, rdir: *mut u8,
+) -> AppProto {
+    if input_len < 28 {
+        // at least the ISAKMP_HEADER must be there, not ALPROTO_UNKNOWN because over UDP
+        return unsafe { ALPROTO_FAILED };
+    }
+
+    if input_len > 1 && input != std::ptr::null_mut() {
```
victorjulien, line 332: `input_len` already validated to be `>= 28` above.

rust/src/common.rs
victorjulien: Maybe it should be something like `slice_to_hex`

victorjulien (changes requested):
Some minor changes requirement, as well as passing CI of course.

I will fast track this when we reach a clean PR state.

satta:
Thanks for the review and the concise requests. Will look at it soon.

src/Makefile.am
satta: Indeed, most likely an editing oops made during merge conflict resolution. Will fix

rust/src/common.rs
satta: How about `byteslice_to_hex`?

rust/src/ike/ike.rs
satta: Correct, we can remove the `input_len > 1` part of this check.

rust/src/ike/detect.rs
satta: I think so. I was able to narrow this down to just two unsafe lines of code.

satta:
> Can you fix the formatting differently? Instead of a fix commit do:
> `git rebase master -x 'git clang-format HEAD~'`. This will fix the formatting per commit.

Done -- and quite handy to know as well! Doesn't work with clang versions <9 though (like the one available in Buster) so I had to do this in a chroot.

rules/ipsec-events.rules
satta: Done.

rust/src/common.rs
jasonish: I think the name is OK if its the one true format of hex we want to support.  As most things we'd want to print as hex will already by a Vec<u8> or a &[u8], or have some sort of as_bytes() method (strings).

How does this differ from doing the builtin `let hex_string = format!("{:x}", some_byte_slice);`?

satta: Your example does not build for me:
```rust
let byteslice: &[u8] = tx.hdr.ikev1_header.key_exchange.as_ref();
let hex_string = format!("{:x}", byteslice);
SCLogInfo!("{:?}", hex_string);
```

```
    |
130 |                     let hex_string = format!("{:x}", byteslice);
    |                                                      ^^^^^^^^^ the trait `std::fmt::LowerHex` is not implemented for `[u8]`
    |
    = note: required because of the requirements on the impl of `std::fmt::LowerHex` for `&[u8]`
    = note: required by `std::fmt::LowerHex::fmt`

```

but I just noticed that there is also `to_hex_string(bytes: &[u8]) -> String` in `rust/src/krb/krb5.rs` and that differs only by uppercase vs. lowercase output. Maybe it makes sense to consolidate these in one place...

jasonish: Ah, your right.  The hashing module does this, but the hasing functions return a GenericArray which implement `LowerHex`...  So I'm happy with this as as.

As for `to_hex_string`in krb5, it doesn't appear to be used anywhere.

satta: Okay, thanks for your insight! :+1:

satta:
Next PR: #5955

PR #5947, opened 2021-03-02
duplicate sid for 2200111 rev 1 conflicts with rev 2

ct0br0:
Duplicate SID 2200111 conflicts with rev 2 that is already defined.

### DIFF
--- a/rules/decoder-events.rules
+++ b/rules/decoder-events.rules
@@ -143,6 +143,6 @@ alert pkthdr any any -> any any (msg:"SURICATA ERSPAN too many vlan layers"; dec
 # Cisco Fabric Path/DCE
 alert pkthdr any any -> any any (msg:"SURICATA DCE packet too small"; decode-event:dce.pkt_too_small; classtype:protocol-command-decode; sid:2200110; rev:2;)
 
-alert pkthdr any any -> any any (msg:"SURICATA packet with too many layers"; decode-event:too_many_layers; classtype:protocol-command-decode; sid:2200111; rev:1;)
-# next sid is 2200112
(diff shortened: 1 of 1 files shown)


rules/decoder-events.rules
```diff
@@ -143,6 +143,3 @@ alert pkthdr any any -> any any (msg:"SURICATA ERSPAN too many vlan layers"; dec
 # Cisco Fabric Path/DCE
 alert pkthdr any any -> any any (msg:"SURICATA DCE packet too small"; decode-event:dce.pkt_too_small; classtype:protocol-command-decode; sid:2200110; rev:2;)
 
-alert pkthdr any any -> any any (msg:"SURICATA packet with too many layers"; decode-event:too_many_layers; classtype:protocol-command-decode; sid:2200111; rev:1;)
```
victorjulien, line 146: we still need this rule, just need it to have a unique sid. Also, we should probably make sure it has the same sid across our supported branches. CI checks for both would be great :)

ct0br0: increased sid, looks like that's what it is in other branches.

will look into github CI for testing rules

ct0br0:
Took another look and realized the SIDs are not quite sequential.
After talking with Jason Ish it makes a bit more sense

This seems to be the SID for this rule.

```
master-5.0.x.rules
alert pkthdr any any -> any any (msg:"SURICATA packet with too many layers"; decode-event:too_many_layers; classtype:protocol-command-decode; sid:2200112; rev:1;)

master-6.0.x.rules
alert pkthdr any any -> any any (msg:"SURICATA packet with too many layers"; decode-event:too_many_layers; classtype:protocol-command-decode; sid:2200116; rev:1;)
```

victorjulien:
Merged into `master-5.0.x`, thanks.

PR #5948, opened 2021-03-03
smtp/mime: configurable url scheme extraction v5

abungay:
Previous PR: #5719

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2054

Suricata-verify:
https://github.com/OISF/suricata-verify/pull/470

Describe changes:
- adds option 'extract-urls-schemes' to suricata.yaml to specify schemes to extract URLs instead of only 'http://' prefixed URL's getting parsed from mime bodies of SMTP traffic when using 'extract-urls: yes'
- adds option 'log-url-scheme' to suricata.yaml to specify whether to log the scheme portion of the extracted URLs, e.g. log the extra 'http://' prefix along with URLs

Changes since last PR:
- append "://" suffix to schemes from config using snprintf at initializing time once instead of at search time every time
- add extra comments
- add extra error checks in test cases incase SCStrdup or ConfNodeNew fail
- add extra commit to use bool instead of int where applicable in related SMTP source files
- use FatalError instead of exit(EXIT_FAILURE)
- remove max scheme length constant, not limiting the length of scheme names in config anymore
- update Copyright of changed files to say 2021 as year
- update commits author format to use "Firstname Lastname <email>"

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

suricata-verify-pr: 470

PR #5952, opened 2021-03-03
smtp/mime: configurable url scheme extraction v6

abungay:
Previous PR: #5948

https://redmine.openinfosecfoundation.org/issues/2054

Suricata-verify:
https://github.com/OISF/suricata-verify/pull/470

Describe changes:
- adds option 'extract-urls-schemes' to suricata.yaml to specify schemes to extract URLs instead of only 'http://' prefixed URL's getting parsed from mime bodies of SMTP traffic when using 'extract-urls: yes'
- adds option 'log-url-scheme' to suricata.yaml to specify whether to log the scheme portion of the extracted URLs, e.g. log the extra 'http://' prefix along with URLs

Changes since last PR:
- fix coccinelle qa script failure for CI

suricata-verify-pr: 470

### DIFF
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -1,5 +1,5 @@
 /* Copyright (C) 2012 BAE Systems
- * Copyright (C) 2020 Open Information Security Foundation
+ * Copyright (C) 2020-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -72,7 +72,6 @@
 #define QP_STR            "quoted-printable"
 #define TXT_STR           "text/plain"
 #define HTML_STR          "text/html"
-#define URL_STR           "http://"
 
 /* Memory Usage Constants */
 #define STACK_FREE_NODES  10
@@ -82,7 +81,7 @@
 #define MAX_IP6_CHARS  39
 
 /* Globally hold configuration data */
-static MimeDecConfig mime_dec_config = { 1, 1, 1, 0, MAX_HEADER_VALUE };
+static MimeDecConfig mime_dec_config = { 1, 1, 1, NULL, false, 0, MAX_HEADER_VALUE };
 
 /* Mime Parser String translation */
 static const char *StateFlags[] = { "NONE",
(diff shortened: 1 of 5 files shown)


src/util-decode-mime.c
```diff
@@ -1026,88 +1024,101 @@ static int FindUrlStrings(const uint8_t *line, uint32_t len,
...
-            if (tok == fptr) {
-                SCLogDebug("Found url string");
-
-                /* First copy to temp URL string */
-                tempUrl = SCMalloc(urlStrLen + tokLen);
-                if (unlikely(tempUrl == NULL)) {
-                    SCLogError(SC_ERR_MEM_ALLOC, "Memory allocation failed");
-                    return MIME_DEC_ERR_MEM;
-                }
+    TAILQ_FOREACH (scheme, &mdcfg->extract_urls_schemes->head, next) {
```
victorjulien, line 1039: have you considered using mpm (multi pattern matching) here instead? That way we'd avoid searching a buffer 4 times if we have 4 patterns defined.

abungay: Decided the simple loop here for simplicity but mpm is a good idea :) Do you have any examples from the suricata source code of how to implement mpm in C?

victorjulien: The parser already uses Mpm to match on response codes. Check the src/app-layer-smtp.c file for
```c
static int SMTPProcessReply(SMTPState *state, Flow *f,
                            AppLayerParserState *pstate,
                            SMTPThreadCtx *td)
{
   ...
    PmqReset(td->pmq);
    int mpm_cnt = mpm_table[SMTP_MPM].Search(smtp_mpm_ctx, td->smtp_mpm_thread_ctx,
                                             td->pmq, state->current_line,
                                             3);
    if (mpm_cnt == 0) {
        /* set decoder event - reply code invalid */
        SMTPSetEvent(state, SMTP_DECODER_EVENT_INVALID_REPLY);
        SCLogDebug("invalid reply code %02x %02x %02x",
                state->current_line[0], state->current_line[1], state->current_line[2]);
        SCReturnInt(-1);
    }
    enum SMTPCode reply_code = smtp_reply_map[td->pmq->rule_id_array[0]].enum_value;
```

The setup logic is in `SMTPSetMpmState`.

abungay: I took at look at how Suricata does MPM and correct me if I'm wrong but it seems that you can only see which pattern matched from the prefilter object pmq and the count of matches in the buffer, but you can't get indexes within the buffer where the pattern matched. This existing URL extraction code for SMTP needs to find the starting index of the match to properly extract the URL so I'm confused about how to use Suricata's MPM implementation here.

abungay:
Opened new PR v7 - https://github.com/OISF/suricata/pull/5952

PR #5955, opened 2021-03-04
ikev1: add ikev1 parser - v16

satta:
Previous PR: https://github.com/OISF/suricata/pull/5941

Describe changes:
- Move clang-format actions out of separate commit.
- Reduce amount of `unsafe` code in `rs_ike_state_get_sa_attribute()`.
- Revision bump on ipsec rules.
- Fix issue in header list.
- Remove unnecessary checks.

suricata-verify-pr: 465


victorjulien:
Merged through #5961, thanks a lot @fhonza and @satta !

satta:
Thank _you_ :)

fhonza:
cool, thanks!

rust/src/ike/logger.rs
```diff
@@ -0,0 +1,227 @@
...
+        }
+    } else if tx.ike_version == 2 {
+        jb.set_uint("exchange_type", tx.hdr.ikev2_header.exch_type.0 as u64)?;
+    }
+
+    if tx.ike_version == 1 {
+        let mut index = 0;
+        for server_transform in &state.ikev1_container.server.transforms {
+            if index >= 1 {
+                debug_validate_bug_on!(true);
```
catenacyber, line 79: Hi @satta 
Why is there this assertion ?
It looks like nothing in the code prevents to have multiple transforms, coming from multiple transactions/packets
cf https://redmine.openinfosecfoundation.org/issues/4433

fhonza: Hi, the client is asking the server with a list of proposals, the server is choosing one of them and always sends back that one choice it made. That's the protocol-logic according to the RFC.

Since the packets in both directions have the same format I found it "nicer" to have the same parser logic for both directions and assert afterwards. Perhaps that could even happen earlier while parsing....or it's even something to alert on...not sure here if it can be used in a malicious way.

Unfortunately I have no login to the redmine....

catenacyber: Thanks for the quick feedback, I sent you a pcap reproducer by mail.

From your answer, I guess we want to set an event (like `state.set_event(IkeEvent::PayloadExtraData);` but with another event type), instead of using `debug_validate_bug_on`

fhonza: Thanks, I'll have a look!

Yes perhaps is a "softer" logic here better :)

satta: Yep, I think an event would be better, too.

PR #5964, opened 2021-03-10
ssl: test removing pattern

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4389

Describe changes:
- Remove one pattern for SSL recognition

As this pattern does not start at offset 0 and I do not know where it comes from

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -2764,12 +2764,6 @@ static int SSLStateGetEventInfoById(int event_id, const char **event_name,
 
 static int SSLRegisterPatternsForProtocolDetection(void)
 {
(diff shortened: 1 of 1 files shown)


catenacyber:
codecov is unhappy because this PR is covered code removal...

victorjulien:
It took a while to trace this back, but it was introduced for SSLv2 support a long time ago: 5fe1dc1d2449109e34d1f3ef1dddb26ec3faefa9
So we need to make sure we test sslv2 and that protocol detection works before and after this PR.

catenacyber:
Indeed, replaced by https://github.com/OISF/suricata/pull/5973

PR #5969, opened 2021-03-10
install: better warning on install-full and don't fail - v1

jasonish:
"make install" exiting with an error when suricata-update is not
bundled has come up a few times now. Instead, I don't think
we should exit with an error, but just add some extra text
as to why rules were not installed.


victorjulien:
Merged in #5977, thanks!

PR #5971, opened 2021-03-11
Detect/pd only/v3

victorjulien:
Fix dropping on "PD only" sigs. These drops were only applied to a single packet, not to the flow.

Minor cleanups.

suricata-verify-pr: 476


victorjulien:
Replaced by #5975

PR #5973, opened 2021-03-11
sslv2: precise detection pattern with probing parser

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4389

Describe changes:
- Precise one pattern for SSL v2 recognition

As this pattern does not start at offset 0, it needs a better check

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -2764,9 +2764,8 @@ static int SSLStateGetEventInfoById(int event_id, const char **event_name,
 
 static int SSLRegisterPatternsForProtocolDetection(void)
 {
-    if (AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP, ALPROTO_TLS,
-                                               "|01 00 02|", 5, 2, STREAM_TOSERVER) < 0)
-    {
+    if (AppLayerProtoDetectPMRegisterPatternCSwPP(IPPROTO_TCP, ALPROTO_TLS, "|01 00 02|", 5, 2,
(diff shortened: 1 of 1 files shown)


victorjulien:
What is wrong with a pattern match a few bytes in? It still uses a fixed offset.

victorjulien:
Merged in #5977, thanks!

catenacyber:
Sorry @victorjulien this was not part of #5977

src/app-layer-ssl.c
```diff
@@ -2764,9 +2764,8 @@ static int SSLStateGetEventInfoById(int event_id, const char **event_name,
 
 static int SSLRegisterPatternsForProtocolDetection(void)
 {
-    if (AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP, ALPROTO_TLS,
-                                               "|01 00 02|", 5, 2, STREAM_TOSERVER) < 0)
-    {
+    if (AppLayerProtoDetectPMRegisterPatternCSwPP(IPPROTO_TCP, ALPROTO_TLS, "|01 00 02|", 5, 2,
+                STREAM_TOSERVER, SSLProbingParser, 5, 5) < 0) {
```
jlucovsky, line 2768: Both `pp_min_depth` and `pp_max_depth` are `5` ... seems odd that `min == max`

catenacyber: You are right, other `SSLProbingParser` use 0 and 3

In this case, that does not make much difference since we have the pattern which needs at least 5 bytes

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5990

PR #5979, opened 2021-03-16
Http range file v10

catenacyber:

Describe changes:

suricata-verify-pr: 443

Modifies #5874 with comments taken into account

Still TODO : thread safety with lock(s) in `ContainerUrlRange` itself


victorjulien:
"format: app-layer-htp.c" should be more specific. Something like: `app-layer/htp: format event table`

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/5994

PR #5982, opened 2021-03-17
dns logging cleanup - v1

jasonish:
The first commit removes DNS from registering 2 loggers, one for to_server and one for to_client. While this still works, it is not required and was only required for the C DNS parser that was stateful.

Finally, remove a few logger IDs that are no longer used.


src/output-json-dns.c
```diff
@@ -415,6 +415,17 @@ static int JsonDnsLoggerToClient(ThreadVars *tv, void *thread_data,
...
 
+static int JsonDnsLogger(ThreadVars *tv, void *thread_data, const Packet *p, Flow *f, void *alstate,
+        void *txptr, uint64_t tx_id)
+{
+    if (rs_dns_tx_is_request(txptr)) {
+        return JsonDnsLoggerToServer(tv, thread_data, p, f, alstate, txptr, tx_id);
+    } else if (rs_dns_tx_is_response(txptr)) {
+        return JsonDnsLoggerToClient(tv, thread_data, p, f, alstate, txptr, tx_id);
+    }
+    return TM_ECODE_OK;
```
jlucovsky, line 426: If the request is neither a request nor response, should an error be returned instead?

jasonish: Interesting question as I believe this return value is ignored.  @victorjulien any thoughts? We can't log an error on error (or maybe one) as errors are going to be repetitive here I bet.  But perhaps a stat for logging errors could make sense?

victorjulien: The loggers should not ever return errors based on anomalies in traffic. These should be handled in the parser. It *may* be useful to have a special error record for some cases, although I think this is really what the anomaly log facility is for. So really we could change the type here to void I think.

jasonish: With respect to errors here I was thinking more about errors in the actual building of logs unrelated to the protocol decode.  I don't think we can SCLogError, it could be to frequent.  But might be interesting to know they are happening.

Anyways, in this specific case we could BUG_ON if its not a request or a reply, the parser should have taken care that a transaction can't be created that is not a request or a response.

jasonish: But in other words, that might be tasks for another issue/pr/day..

victorjulien:
Merged in #6018

PR #5983, opened 2021-03-18
scripts/setup-app-layer: fix Makefile.am patch

jufajardini:
adjust lines for patching /src/Makefile.am, as currently generated
Makefile wasn't building Suricata.
Add suggestion to run "./configure" before running "make"

there isn't one. Should I create in this case?

Describe changes:
- after a recent change in the Makefile.am file, setup-app-layer patches to this file stopped working, as an indent was inserted in the line it should recognize. To fix this, added that space where the script looks for the pattern to be replaced.
- also added the suggestion to run `./configure` before make, so a new suricata.yaml file will be generated, too.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/scripts/setup-app-layer.py
+++ b/scripts/setup-app-layer.py
@@ -90,7 +90,7 @@ def patch_makefile_am(protoname):
     output = io.StringIO()
     with open("src/Makefile.am") as infile:
         for line in infile:
-            if line.startswith("app-layer-template.c"):
+            if line.startswith("	app-layer-template.c"):
                 output.write(line.replace("template", protoname.lower()))
             output.write(line)
     open("src/Makefile.am", "w").write(output.getvalue())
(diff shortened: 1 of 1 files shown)


scripts/setup-app-layer.py
```diff
@@ -90,7 +90,7 @@ def patch_makefile_am(protoname):
     output = io.StringIO()
     with open("src/Makefile.am") as infile:
         for line in infile:
-            if line.startswith("app-layer-template.c"):
+            if line.startswith("	app-layer-template.c"):
```
jasonish, line 93: So before this break the line would look like:
```
app-layer-template.c app-layer-template.h
```
But now the headers are in a separate var, so this only applies to the .c file. Do you want to take a stab and fixing this up for the header as well?

jufajardini: I was wondering about that difference, true! Sure, I like fixing things. Thanks for pointing that out! :)

catenacyber: Should we rather add `lstrip` like `line.lstrip().startswith("app-layer-template.c"):` ?

jufajardini: That sounds better, indeed. I wasn't sure my solution was good, just knew it needed some fixing...

jufajardini:
Closed with [PR 5895](https://github.com/OISF/suricata/pull/5985)

PR #5986, opened 2021-03-21
plugin/syslog: Make syslog an internal plugin

jlucovsky:
Conversion from draft pr.

This PR changes the syslog output type into a plugin -- an internal plugin -- to take advantage of the EVE output plugin framework. 

Describe changes:
- Refactor internal output file type handling code
- Move syslog output code into an internal plugin

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -413,6 +413,7 @@ noinst_HEADERS = \
 	output-json-template-rust.h \
 	output-json-tftp.h \
 	output-json-tls.h \
+	output-eve-syslog.h \
 	output-lua.h \
 	output-packet.h \
 	output-stats.h \
@@ -990,6 +991,7 @@ libsuricata_c_a_SOURCES = \
 	output-json-template-rust.c \
 	output-json-tftp.c \
 	output-json-tls.c \
+	output-eve-syslog.c \
 	output-lua.c \
 	output-packet.c \
 	output-stats.c \
--- a/src/output-eve-syslog.c
+++ b/src/output-eve-syslog.c
@@ -1,0 +1,117 @@
+/* vi: set et ts=4: */
+/* Copyright (C) 2021 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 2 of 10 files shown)


src/output-plugin-syslog.c
```diff
@@ -0,0 +1,129 @@
...
+    .name = OUTPUT_NAME,
+    .author = "OISF Dev",
+    .license = "GPLv2",
+    .Init = SyslogPluginInitialize
+};
+
+SCPlugin *SyslogPluginRegister(void)
+{
+	return &SyslogPluginRegistration;
+}
```
jasonish, line 128: I think we can skip this bit. This is only required for discovery of externally loadable modules. In this case We could go straight to `SyslogPluginInitialize`.

jasonish:
My first thats are that internal plugins should probably not be called plugins, and this will affect some of the naming we have in place already. For example: `SCPluginRegisterFileType` is a bad name if you consider it a generic registration function for registering a file type.  Perhaps this should have been renamed `SCRegisterEveFileType` which makes a lot more sense if we're going to use it internally as well.

I'd like to think of plugin support more as allowing dynamically loadable modules to use the internal registration functions that already exist in Suricata.

Can this be refactored in such a way where its not so much like a plugin?  Still use `SCPluginRegisterFileType` to register the output type (can change this API later), but when it comes to naming of the syslog "file type", drop the use of the name plugin?

src/output-plugin-syslog.c
jlucovsky: Agreed.

jlucovsky:
> My first thats are that internal plugins should probably not be called plugins, and this will affect some of the naming we have in place already. For example: `SCPluginRegisterFileType` is a bad name if you consider it a generic registration function for registering a file type. Perhaps this should have been renamed `SCRegisterEveFileType` which makes a lot more sense if we're going to use it internally as well.
> 
> I'd like to think of plugin support more as allowing dynamically loadable modules to use the internal registration functions that already exist in Suricata.
> 
> Can this be refactored in such a way where its not so much like a plugin? Still use `SCPluginRegisterFileType` to register the output type (can change this API later), but when it comes to naming of the syslog "file type", drop the use of the name plugin?

Good suggestion. 

The intent behind this change was to move the eve output types into a format more like a plugin -- agreed that the term "plugin" should be different -- so that there'd be less chance for functionality gaps between internal output types in this format and external plugins.

src/output-plugin-syslog.c
jlucovsky: Well -- the syslog module must be discoverable ... currently, that happens by inclusion on the plugin list.

jasonish: Can't `SyslogInitialize` just be called at startup when we register everything else? The plugin registration is more of a hardcoded name we can use to discover the registration function in a dynamically loaded module, that step isn't really needed for internal modules.

jlucovsky: Updated with latest commit.

jlucovsky:
Continued in #6121

PR #5994, opened 2021-03-23
Http range file v11

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/1576

Describe changes:
- Uses a generic container (a thread-safe hash table whose key is the url/filename) to store file parts with HTTP Range header
- handles reassembly of these parts, while enforcing a global memcap

suricata-verify-pr: 443

Modifies #5979 with thread safety with lock(s) in `ContainerUrlRange` itself

TODO: commit squashing 
Renaming to more specific ?
Test concurrency ? Is there a generic framework for this ?

PR #5996, opened 2021-03-23
Http range file v12

catenacyber:

Describe changes:

suricata-verify-pr: 443

Modifies #5994 with fixing S-V  with corner cases where we started skipping and finished skipping everything...

TODO: commit squashing 
Renaming to more specific ?
Test concurrency ? Is there a generic framework for this ?


src/app-layer-htp-file.c
```diff
@@ -225,34 +261,89 @@ int HTPParseContentRange(bstr * rawvalue, HtpContentRange *range)
...
-    SCReturnInt(retval);
+    s->file_range = ContainerUrlRangeOpenFile(file_range_container, crparsed.start, crparsed.end,
+            crparsed.size, &s->cfg->response.sbcfg, filename, filename_len, flags);
+    if (s->file_range == NULL) {
+        // probably reached memcap
+        return HTPFileOpen(
+                s, filename, (uint32_t)filename_len, data, data_len, txid, STREAM_TOCLIENT);
+    }
+
+    if (ContainerUrlRangeAppendData(s->file_range, data, data_len) < 0) {
```
victorjulien, line 342: It looks like `s->file_range` is not holding a mutex here, and `ContainerUrlRangeAppendData` doesn't seem to take it either. The pattern we use in the flow engine for example, is that lookup/creation returns a locked object. We then do our work on it and release the lock when we're done.

catenacyber: > It looks like `s->file_range` is not holding a mutex here, and `ContainerUrlRangeAppendData` doesn't seem to take it either. 

Indeed

> The pattern we use in the flow engine for example, is that lookup/creation returns a locked object. We then do our work on it and release the lock when we're done.

That is what is meant to happen here
`ContainerUrlRangeAppendData` is not part of the `work` to do as it operates on `ContainerUrlRangeFile`
The work to do is to return a `ContainerUrlRangeFile` which can be used without locks as each flow will have a different one (as opposed to `ContainerUrlRange` which has the lock for lookup/creation/work)
Only one flow will have the pointer to the `FileContainer` structure while others have buffers (or just an indication to skip the content already seen)
Then `ContainerUrlRangeFileClose` uses the lock from the structure `ContainerUrlRange` to kind of synchronize together a finished range chunk...

Is it clearer ?

ct0br0:
FYI seg faults in QA
3049 - tlpr1

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6207

PR #5938, opened 2021-02-26
PostgreSQL: initial support (MVP)

jufajardini:
PostgreSQL: add SSL handshake support
-- applayerpgsql/pgsql: add unittests for parse_response
-- applayerpgsql/parser: add parsers for request and response messages
-- applayerpgsql/pgsql: add initial parsing for error/notice messages
field types/values
-- this parser does NOT yet handle what happens after the handshake

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)   

Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/4241

Previous PR: https://github.com/OISF/suricata/pull/5877

Describe changes:
- pgsql/parser.rs: add nom parsers for the Error response message.
--- add nom parser for Error response field SQLStateCode
--- make u8 enum values human friendly/readable
--- make `match` for Error Field types more specific
--- add unittests for the new nom parsers
- pgsql/pgsql.rs: adjust APP_LAYER_PARSER_EOF to comply with new format TC/TS;
--- change is_ssl_request to is_ssl_hanshake (still unsure if it will be needed, tho)

Current challenge/step is to understand/decide how to have generic types for `PGSQLTransaction.request` and `PGSQLTransaction.response`. I must decide on a solution, so Suricata can actually start parsing the Error Response message, too. Thought of a few things, but realised it would be better to share what I have first, and then keep moving forward. 

#suricata-verify-pr:
#suricata-verify-repo: https://github.com/jufajardini/suricata-verify/tree/sslhandshake-v0
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #5997, opened 2021-03-24
PostgreSQL: add initial support (WIP)

jufajardini:

Last PR (not exactly a continuation, though): https://github.com/OISF/suricata/pull/5938

Describe changes:
- Reboot the code: restart from scratch, using again setup-app-layer, to get rid of messy spaghetti code;
- Patch common files;
- rust/src/applayerpgsql: add nom parsers for decoding SSL Requests and Responses;
--- there's a TODO indicating a likely source of problem in how it parses SSL Responses, because other checks are necessary to make sure it is indeed an SSL Response instead of a longer, regular message.
- use a data model that allows adding more message types more easily
- Add nom parser for real startup messages.
- Adapt pgsql.rs to work with current pgsql nom parsers.
- This is parser parses SSL Requests, SSL Responses, and basic Startup Messages (parameter order still fixed) 

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch: https://github.com/jufajardini/suricata-verify/tree/ssl-handshake-startup-message-v0
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:


rust/src/applayerpgsql/parser.rs
```diff
@@ -0,0 +1,426 @@
...
+struct PgsqlParameter {
+    param_name: Vec<u8>,
+    param_value: Vec<u8>,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct PgsqlStartupParameters {
+    user: PgsqlParameter,
+    database: Option<PgsqlParameter>,
+    // TODO size-wise, being Option<Vec> or just Vec here is the same. I wonder if
```
victorjulien, line 36: how did you reach this conclusion?

rust/src/applayerpgsql/pgsql.rs
```diff
@@ -0,0 +1,652 @@
...
+    let eof = unsafe {
+        if AppLayerParserStateIssetFlag(pstate, APP_LAYER_PARSER_EOF_TS) > 0 {
+            true
+        } else {
+            false
+        }
+    };
+
+    if eof {
+        // If needed, handled EOF, or pass it into the parser.
```
victorjulien, line 360: with eof we can still have data I think, so we'd have to process that

src/Makefile.am
```diff
@@ -40,6 +40,7 @@ noinst_HEADERS = \
 	app-layer-nfs-udp.h \
 	app-layer-ntp.h \
 	app-layer-parser.h \
+    app-layer-pgsql.h \
```
victorjulien, line 43: formatting looks off

victorjulien:
I think adding more tests is the critical next step, as it will help you define the parsers and their data structures.

victorjulien:
I'd also like these PRs to already fully pass our CI checks. In general leaving warnings in dev code is something I don't like. In many cases it has given me clues about things being broken than I missed because they got lost in the noise.

src/Makefile.am
jufajardini: Thank you, and I'm sorry for having to have that one pointed out >_<' . In VSCode they looked the same...

rust/src/applayerpgsql/parser.rs
jufajardini: As I was concerned about this, I searched and found an answer in StackOverflow that shared a macro where one could make such comparisons (here: https://stackoverflow.com/questions/16504643/what-is-the-overhead-of-rusts-option-type)
I used that macro to compare some possibilities, and that seemed to be the case, to me. Here are some of the comparisons I've tried: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=076232646a400fa0fe16a2fe0d53917a

(longer story here: https://jufajardini.wordpress.com/2021/03/10/til-sizes-of-structs-with-option-fields-vs-normal-attributes/)

jasonish: `Vec<u8>` and `Option<Vec<u8>` do show the same size with `size_of`, but I think there is a piece missing.  `Vec<u8>` requires a `Vec::new()` which is going to use some more memory that doesn't appear to show up in `size_of`.

I wonder if a good rule of thumb for us would be:
- If under normal conditions, something will always be in the Vec, just initialize it.
- If its common enough that the list could be empty, wrap it in Option and allocate it as needed?

jufajardini: Thanks for the clarification. I think that approach makes sense!
I was thinking yesterday that I could actually have something much simpler, like just one Vec where all parameters will be stored, after parsing, but not sure if that's good, and had some trouble in making that work with my current parsers. But if you'd consider that better, I'd find out how other implementations have done it, and make it work.

jufajardini: In fact, considering memory allocation, having just one vec seems to make much more sense in my head, but as I'm still learning, I'll wait for more knowledgeable considerations x)

victorjulien: While this is an interesting discussion, I would like to suggest to not worry about such things too much in this phase of the project. These kind of micro optimizations can easily add unnecessary complexity.

jufajardini: Fair, and important advice, to someone like me :sweat_smile: I'll stay on the path with adding more tests, as you guided me to.

jufajardini:
Closed with PR https://github.com/OISF/suricata/pull/6012

PR #5998, opened 2021-03-24
Research/bug 3703 file bug/v9

victorjulien:
Address issue 3703 by logging files/filedata in the packet direction only. This leads to missing metadata in file records as txs are cleared before the files are logged. So address this by creating file accouning per tx.

suricata-verify-pr: 481


victorjulien:
Replaced by #5999

PR #5999, opened 2021-03-25
Bug 3703 file bug/v10

victorjulien:
Address issue 3703 by logging files/filedata in the packet direction only. This leads to missing metadata in file records as txs are cleared before the files are logged. So address this by creating file accouning per tx.

suricata-verify-pr: 482

Replaces #5998 fixing formatting.

### DIFF
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -1,4 +1,4 @@
-/* Copyright (C) 2017-2020 Open Information Security Foundation
+/* Copyright (C) 2017-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -58,6 +58,10 @@ pub struct AppLayerTxData {
     /// logger flags for tx logging api
     logged: LoggerFlags,
 
+    /// track file open/logs so we can know how long to keep the tx
+    pub files_opened: u32,
+    pub files_logged: u32,
+
     /// detection engine flags for use by detection engine
     detect_flags_ts: u64,
     detect_flags_tc: u64,
@@ -68,9 +72,17 @@ impl AppLayerTxData {
         Self {
             config: AppLayerTxConfig::new(),
             logged: LoggerFlags::new(),
+            files_opened: 0,
+            files_logged: 0,
             detect_flags_ts: 0,
             detect_flags_tc: 0,
         }
+    }
+    pub fn init_files_opened(&mut self) {
+        self.files_opened = 1;
+    }
+    pub fn incr_files_opened(&mut self) {
+        self.files_opened += 1;
     }
 }
 
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -61,6 +61,7 @@ pub struct AppLayerTxData {
     /// track file open/logs so we can know how long to keep the tx
     pub files_opened: u32,
     pub files_logged: u32,
+    pub files_stored: u32,
 
     /// detection engine flags for use by detection engine
     detect_flags_ts: u64,
@@ -74,6 +75,7 @@ impl AppLayerTxData {
             logged: LoggerFlags::new(),
             files_opened: 0,
             files_logged: 0,
+            files_stored: 0,
             detect_flags_ts: 0,
             detect_flags_tc: 0,
         }
--- a/rust/src/http2/http2.rs
+++ b/rust/src/http2/http2.rs
@@ -157,6 +157,8 @@ impl HTTP2Transaction {
     }
 
     pub fn free(&mut self) {
+        debug_validate_bug_on!(self.tx_data.files_opened > 1);
+        debug_validate_bug_on!(self.tx_data.files_logged > 1);
         if self.events != std::ptr::null_mut() {
             core::sc_app_layer_decoder_events_free_events(&mut self.events);
         }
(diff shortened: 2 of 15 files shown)


rust/src/http2/http2.rs
```diff
@@ -809,6 +809,7 @@ impl HTTP2State {
                                     let mut tx_same = &mut self.transactions[index - 1];
                                     tx_same.ft.tx_id = tx_same.tx_id - 1;
                                     let (files, flags) = self.files.get(dir);
+                                    tx_same.tx_data.init_files_opened(); // TODO do we have a place for doing this once?
```
victorjulien, line 814: @catenacyber I did not field a single place where we can initialize a file in the tx. This works because it just keeps the file cnt at 1, but its called for each chunk of file data.

rust/src/applayer.rs
```diff
@@ -58,6 +58,11 @@ pub struct AppLayerTxData {
     /// logger flags for tx logging api
     logged: LoggerFlags,
 
+    /// track file open/logs so we can know how long to keep the tx
+    pub files_opened: u32,
```
victorjulien, line 62: One thing I don't like is that this increases this per tx structure with 16 bytes (12 for the fields and 4 for padding):
```c
typedef struct {
	AppLayerTxConfig           config;               /*     0     1 */

	/* XXX 3 bytes hole, try to pack */

	LoggerFlags                logged;               /*     4     4 */
	uint32_t                   files_opened;         /*     8     4 */
	uint32_t                   files_logged;         /*    12     4 */
	uint32_t                   files_stored;         /*    16     4 */

	/* XXX 4 bytes hole, try to pack */

	uint64_t                   detect_flags_ts;      /*    24     8 */
	uint64_t                   detect_flags_tc;      /*    32     8 */

	/* size: 40, cachelines: 1, members: 7 */
	/* sum members: 33, holes: 2, sum holes: 7 */
	/* last cacheline: 40 bytes */
} AppLayerTxData;
```
vs master:
```c
typedef struct {
	AppLayerTxConfig           config;               /*     0     1 */

	/* XXX 3 bytes hole, try to pack */

	LoggerFlags                logged;               /*     4     4 */
	uint64_t                   detect_flags_ts;      /*     8     8 */
	uint64_t                   detect_flags_tc;      /*    16     8 */

	/* size: 24, cachelines: 1, members: 4 */
	/* sum members: 21, holes: 1, sum holes: 3 */
	/* last cacheline: 24 bytes */
} AppLayerTxData;
```

victorjulien: The reason I've gone with `u32` is that in SMTP (attachments) and HTTP (multipart) we can potentially have a very large number of files per tx.

jlucovsky (approved):

rust/src/http2/http2.rs
catenacyber: Indeed, there is no initialization, just each chunk of HTTP2 `DATA` is part of the file 
Would you like some condition like `if isFirstDataChunk(files)` to run this line ?

rust/src/http2/http2.rs
```diff
@@ -157,6 +157,8 @@ impl HTTP2Transaction {
     }
 
     pub fn free(&mut self) {
+        debug_validate_bug_on!(self.tx_data.files_opened > 1);
```
catenacyber, line 160: Cannot we have 2 files ? 1 to client and 1 to server ?

catenacyber:
> logging files/filedata in the packet direction only

Why did we log files when the packet was in the other direction ?

rust/src/http2/http2.rs
victorjulien: Yeah I think this would be useful

rust/src/http2/http2.rs
victorjulien: Yes you are correct. So if we have a way to only increment the file count when the file starts, we can use `incr_files_opened()` and check for max value of 2 here.

catenacyber: It makes me wonder if the `ft: FileTransferTracker` in `HTTP2Transaction` should be duplicated as ft_ts and ft_tc

rust/src/http2/http2.rs
catenacyber: Guess it would be `if files.tail == ptr::null_mut() {`
Or `self.ft.file_open` but then, maybe `ft` needs to ne duplicated in `ft_tc` and `ft_ts`

rust/src/http2/http2.rs
victorjulien: Yes, I think so. HTTP1 uses this logic as well. A file container per direction.

rust/src/http2/http2.rs
victorjulien: I think we need that `ft_ts` and `ft_tc` anyway, so `self.ft.file_open` might work.

catenacyber: So, how do you want to proceed, should I create a PR on top of these commits ?

rust/src/http2/http2.rs
victorjulien: Can you implement this split in a separate PR? We'd want that to go into 6 as well. I'll then rework my branch on top of that.

victorjulien:
> > logging files/filedata in the packet direction only
> 
> Why did we log files when the packet was in the other direction ?

Probably to work around the complexity of the solution in this PR, but thats a guess. Its been like this for a long time.

rust/src/http2/http2.rs
catenacyber: https://github.com/OISF/suricata/pull/6080

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |
| .app_layer.flow.ftp-data | 157 | 500 | 31.0% |

suricata-qa:
WARNING:
|--|--|--|--|
| | tlpr1_stats_chk |
| .app_layer.flow.ftp-data | 157 | 500 | 31.0% |

victorjulien:
Merged in #6324

PR #6001, opened 2021-03-25
Draft: Eve cleanup - round 1

jasonish:
The first change was to have CreateEveHeader add the common options
as this was left out in a few loggers. While update all the loggers
that use CreateEveHeader, remove redundant code, in particular
from loggers that don't need to use their own context but
can use the generic one.

The idea is to make loggers much easier to register, which will ease registration from Rust, and parser plugins.


victorjulien:
Looks good to me.

The missing coverage is something I'd like to see addressed as its now so clearly laid out. Including a test for the template parsers?

jasonish:
Continue at https://github.com/OISF/suricata/pull/6006.

Nothing extra done with respect to coverage.  But the new PR does use an S-V PR that has tests for the template parsers.

PR #6007, opened 2021-03-29
fuzz: specify protocol with fuzz target name

catenacyber:

Describe changes:
- Make fuzz target about app-layer force a protocol based on its name

So that oss-fuzz can have one fuzz target for each protocol, just by copying and using another name
Tested with my oss-fuzz branch suricataproto

### DIFF
--- a/src/tests/fuzz/fuzz_applayerparserparse.c
+++ b/src/tests/fuzz/fuzz_applayerparserparse.c
@@ -15,6 +15,7 @@
 #define HEADER_LEN 6
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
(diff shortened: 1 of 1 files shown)


victorjulien:
Can you add more explanation to the git commit, including ticket ref?

Would it make sense to have our buildsys output these binaries based on the app protocols we have in the code?

catenacyber:
> Would it make sense to have our buildsys output these binaries based on the app protocols we have in the code?

I do not think so.
It is the same binary.
The hack is oss-fuzz is to use the binary name, as we use the env variable, to find out which app-layer we will be fuzzing.

catenacyber:
Force-pushed with reworded commit

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6029

PR #6033, opened 2021-04-07
Tcp protodetect bail 4171 v2

catenacyber:

Describe changes:
- `TCPProtoDetectCheckBailConditions` used `STREAM_RIGHT_EDGE` which is not the size already processed for protocol detection, but the far edge of the stream where reassembly is not finished yet. Fix is to use `next_seq` rather than `segs_right_edge`
- `HandleEstablishedPacketToServer` was updating too much `next_seq` in the case we have reassembly to do after a gap. Fix is to update only up to `last_ack` in this specific case

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -178,6 +178,9 @@ static void DisableAppLayer(ThreadVars *tv, Flow *f, Packet *p)
             f, f->alproto, f->alproto_ts, f->alproto_tc);
 }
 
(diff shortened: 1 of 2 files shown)


catenacyber:
This is blocking fuzz_sigpcap_aware

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6069

PR #6030, opened 2021-04-07
Protodetect enip dns 4388 v4

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4388

Describe changes:
- Improves probing parsers for nfs, enip and dns

so as to avoid protocol detection confusion with TCP splitting

PR #5759, opened 2021-01-21
Protocol detection 2757 v6

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/2757

Describe changes:
- Use both directions for protocol detection over UDP (as is done with TCP)
- Changes `AppLayerProtoDetectPPGetProto` so that first packet does not bit mask away a protocol for both directions when config option midstream is activated 

The main goal was the first one (use both directions for UDP)
And then I came across this unwanted behavior from `AppLayerProtoDetectPPGetProto` :
When the first packet is no valid DNS but on port 53 (for instance a junk request), second packet (error response from server) does not get checked for DNS as first packet bit masked away DNS for both directions (ie `probing_parser_toclient_alproto_masks` and `probing_parser_toserver_alproto_masks`)

There are other open PRs regarding
https://redmine.openinfosecfoundation.org/issues/2757
such as "Support when multiple protocols have patterns matching (such as USER for FTP and IRC)"

Modifies #4681
- do not check forever if one side says `ALPROTO_UNKNOWN` and the other one says `ALPROTO_DCERPC`
- call `AppLayerParserParse` if the current side failed protocol detection, but the other side gave us previously the protocol to parse
Modifies #5758 with clang-format

suricata-verify-pr: 417

PR #6035, opened 2021-04-07
Protodetect enip dns 4388 v5

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4388

Describe changes:
- Improves probing parsers for nfs, enip and dns

so as to avoid protocol detection confusion with TCP splitting

Modifies #6030 with formatting fixes

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -701,23 +701,45 @@ impl DNSState {
     }
 }
 
+fn probe_header_validity(header : DNSHeader, rlen: usize) -> (bool, bool, bool) {
+    let opcode = ((header.flags >> 11) & 0xf) as u8;
+    if opcode >= 7 {
+        //unassigned opcode
+        return (false, false, false);
+    }
+    if header.additional_rr as usize > rlen {
+        //not enough data for additional records
+        return (false, false, false);
+    }
+    let is_request = header.flags & 0x8000 == 0;
+    return (true, is_request, false);
+}
+
 /// Probe input to see if it looks like DNS.
-fn probe(input: &[u8]) -> (bool, bool) {
-    match parser::dns_parse_request(input) {
-        Ok((_, request)) => {
-            let is_request = request.header.flags & 0x8000 == 0;
-            return (true, is_request);
+fn probe(input: &[u8], dlen: usize) -> (bool, bool, bool) {
+    let i2 = if input.len() <= dlen { input } else { &input[..dlen] };
+    match parser::dns_parse_request(i2) {
+        Ok((rem, request)) => {
+            return probe_header_validity(request.header, rem.len());
         },
-        Err(_) => (false, false),
+        Err(nom::Err::Incomplete(_)) => {
+            match parser::dns_parse_header(input) {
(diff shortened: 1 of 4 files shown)


catenacyber:
Force-pushed to fix CIFuzz other variant by improving further enip (thought I was on a new branch)

victorjulien:
@jasonish can you review this?

rust/src/dns/dns.rs
```diff
@@ -701,23 +701,45 @@ impl DNSState {
...
+fn probe_header_validity(header : DNSHeader, rlen: usize) -> (bool, bool, bool) {
+    let opcode = ((header.flags >> 11) & 0xf) as u8;
+    if opcode >= 7 {
+        //unassigned opcode
+        return (false, false, false);
+    }
+    if header.additional_rr as usize > rlen {
+        //not enough data for additional records
+        return (false, false, false);
+    }
```
jasonish, line 713: We can end up here if `dns_parse_request` fails with incomplete, but `dns_parse_header` passes.  In the case of TCP could we just need more data here?

jasonish: Nevermind, still results in the probe returning ALPROTO_UNKNOWN, which should lead to the parser being called again when more data is available. Is that correct @victorjulien .

jasonish:
My one concern is that this would eliminate the possibility to alert on something like a DNS request that uses a bad opcode.  But not sure where to draw the line such that we don't accidentally parse something as DNS which is semantically not DNS, or being able to detect broken DNS implementations, or someone doing goofy stuff. Same goes for ENIP.

rust/src/dns/dns.rs
catenacyber: > the probe returning ALPROTO_UNKNOWN, which should lead to the parser being called again when more data is available

That seems correct to me

catenacyber:
> My one concern is that this would eliminate the possibility to alert on something like a DNS request that uses a bad opcode. But not sure where to draw the line such that we don't accidentally parse something as DNS which is semantically not DNS, or being able to detect broken DNS implementations, or someone doing goofy stuff. Same goes for ENIP.

I guess that, in case of a first junk message, we need to resort on the other direction to get the correct protocol

cf https://github.com/OISF/suricata/pull/5759

rust/src/dns/dns.rs
catenacyber: > We can end up here if `dns_parse_request` fails with incomplete, but `dns_parse_header` passes. In the case of TCP could we just need more data here?

It now looks to me like there was a bug in `rlen` value

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6182

jasonish:
> My one concern is that this would eliminate the possibility to alert on something like a DNS request that uses a bad opcode. But not sure where to draw the line such that we don't accidentally parse something as DNS which is semantically not DNS, or being able to detect broken DNS implementations, or someone doing goofy stuff. Same goes for ENIP.

@victorjulien Any thoughts on protocol accuracy vs detection of a protocol doing wrong things?

PR #6029, opened 2021-04-07
Fuzz protocols v2

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4125

Describe changes:
- Make fuzz target about app-layer force a protocol based on its name

So that oss-fuzz can have one fuzz target for each protocol, just by copying and using another name
Tested with my oss-fuzz branch suricataproto

Modifies #6007 by adding `stream.midstream = true` especially for fuzz_sigpcap

PR #6040, opened 2021-04-09
Next/20210409/v4

victorjulien:
#5821 #6029 

https://github.com/OISF/suricata/commit/b4e510a1bbba035f6ead05211030d51244626543 proposes to use specific types for the flow storage id's. This would have to be implemented as well for the other storage api users.


victorjulien:
Replaced by #6041

PR #6041, opened 2021-04-09
Next/20210409/v5

victorjulien:
#6040 with formatting fixes

#5821 #6029

b4e510a proposes to use specific types for the flow storage id's. This would have to be implemented as well for the other storage api users.


victorjulien:
Replaces #6042

PR #6052, opened 2021-04-12
Pcre2 3194 v11

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3194

Describe changes:
- migrate all pcre1 code to pcre2
- fuzz: builds fuzz_sigpcap_aware only when asked (ie ./configure --enable-fuzztargets)
- qa : remove some obsolete files
- pcrexform : checks that we have exactly one substring capture in a signature
- transforms : use generic `Free` functions everywhere

What remains to be done :
- fix `make check` : wrapper to turn `PCRE2_ERROR_UNSET` into an empty string ? or individual changes as in the last commit
- answer TODOs (ie test jit ? keep `PCRE` as suricata feature and add `PCRE2` )
- decide if this is priority in front of https://redmine.openinfosecfoundation.org/issues/4112 https://github.com/OISF/suricata/pull/5966 ie using generic uint parsing and other functional changes to the use pcre1/2

suricata-verify-pr: 480

https://github.com/OISF/suricata-verify/pull/480

Replaces https://github.com/OISF/suricata/pull/6045
- migrate all to pcre2 and remove pcre1
- fix compile warning

Used dummy python script to help get there
```
import sys

f = open(sys.argv[1])
l = f.readline()
while True:
    if "pcre2_substring_copy_bynumber(" in l:
        eol = ';' in l
        while not eol:
            l = l.rstrip() + f.readline().lstrip()
            eol = ';' in l
        b = l[l.find("(")+1:l.rfind(")")]
        a = b.split(",")
        print("pcre2len = %s;" % a[-1])
        a[-2] = "(PCRE2_UCHAR8 *)" + a[-2]
        a = ["parse_regex.match"] + a[3:-1]+ ["&pcre2len"]
        c = l[:l.find("(")+1] + ",".join(a) + l[l.rfind(")"):]
        sys.stdout.write(c)
    else:
        sys.stdout.write(l)
    l = f.readline()
    if l == "":
        break
```

PR #5966, opened 2021-03-10
Detect prefilter setupu8 4335 v7

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4112
https://redmine.openinfosecfoundation.org/issues/4335

Describe changes:
- Adds error for impossible rules such as >255 for uint8 fields
- converts icode to use the generic (many more keywords to do)
  * allows <> syntax for uint ranges to support existing syntax
  * corrects reverted ranges and issues a warning (so 20<>8 becomes 8<>20 ) to support existing syntax

Modifies #5973 by rebasing and improving error message

PR #6054, opened 2021-04-13
Pcre2 3194 v13

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3194

Describe changes:
- migrate all pcre1 code to pcre2
- fuzz: builds fuzz_sigpcap_aware only when asked (ie ./configure --enable-fuzztargets)
- qa : remove some obsolete files
- pcrexform : checks that we have exactly one substring capture in a signature
- transforms : use generic `Free` functions everywhere
suricata-verify-pr: 480

https://github.com/OISF/suricata-verify/pull/480

Replaces https://github.com/OISF/suricata/pull/6052
- commits reordered
- fix make check with handling `PCRE2_ERROR_UNSET

This is ready for review (and it needs a big review and testing)
Here are the big questions beyond details :
- Should I make a PR with the first 4 preliminary commits ? (like transform: use generic Free function)
- Should we include in this PR the removal of the usage of PCRE for some keywords parsing like https://redmine.openinfosecfoundation.org/issues/4112 #5966 
- There are some changes in the behavior between PCRE1 and PCRE2 (in the patterns `\i`, `[\w-.]`, or one the behavior to get substrings (now returning an error instead of an empty string if the capture group is not there). How would we deal with this ?


catenacyber:
commit-check fails because it tries to build intermediate commits without libpcre1 as the before-last commit removes libpcre1 from `.github/workflows/commits.yml`

catenacyber:
https://github.com/google/oss-fuzz/pull/5606 should make CIFuzz happier

victorjulien:
> commit-check fails because it tries to build intermediate commits without libpcre1 as the before-last commit removes libpcre1 from `.github/workflows/commits.yml`

Can this be fixed by shuffling the commits? I want to be sure the tree builds at every commit.

victorjulien:
I suppose CIFuzz could already be updated to include pcre2 as well?

catenacyber:
> Can this be fixed by shuffling the commits? I want to be sure the tree builds at every commit.

Nope, but this can be fixed by splitting in 2 PRs, keeping the last commit removing pcre for another independent PR

catenacyber:
> I suppose CIFuzz could already be updated to include pcre2 as well?

CIFuzz (oss-fuzz) was updated to include pcre2.
Re-running...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6138

PR #6055, opened 2021-04-13
modbus: move parser from c to rust - v10

cccs-sadugas:

Describe changes:
- Rebase master.
- Update sawp from `0.2.0` to `0.4.0` and use new `sawp-flags` crate to handle bit flags and remove dependency on `bitflags` crate.

Link to parser:
- [sawp-modbus-0.4.0](https://github.com/CybercentreCanada/sawp/blob/sawp-modbus-0.4.0/sawp-modbus/src/lib.rs)

Link to previous PR:
- #5909

suricata-verify-pr: 486


victorjulien:
The many CI failures are likely caused by using an outdated Suricata-Verify PR.

catenacyber:
I think CIFuzz found a timeout in signatures parsing, but did not manage to report it cf https://github.com/google/oss-fuzz/issues/5621

cccs-sadugas:
Rebased to `suricata-verify-pr: 486`. All tests pass locally.

catenacyber:
I will look into this more closely, probably next week

.travis.yml
```diff
@@ -80,9 +80,11 @@ env:
     # The version of Rust that will be used if not otherwise set.
     - RUST_VERSION="stable"
     # The minimum version of Rust supported.
-    - RUST_VERSION_MIN="1.34.2"
```
catenacyber, line 83: @jasonish would we not remove .travis.yml ?

doc/userguide/output/eve/eve-json-format.rst
```diff
@@ -1920,3 +1920,110 @@ Example of IKE logging:
...
+
+* "address": Starting address of content modification
+* "and_mask": And mask to modify content with
+* "or_mask": Or mask to modify content with
+
+Other Write fields
+~~~~~~~~~~~~~~~~~~
+
+* "address": Starting address to write to
+* "data": Data to write
```
catenacyber, line 1993: Could this become too verbose ?

src/app-layer-modbus.c
```diff
@@ -1580,12 +1582,14 @@ static uint8_t readCoilsRsp[] = {/* Transaction ID */    0x00, 0x00,
...
-                                      /* Unit ID */           0x00,
-                                      /* Function code */     0x81,
-                                      /* Exception code */    0x05};
+static uint8_t readCoilsErrorRsp[] = {
+    /* Transaction ID */ 0x00, 0x00,
+    /* Protocol ID */ 0x00, 0x00,
+    /* Length */ 0x00, 0x03,
+    /* Unit ID */ 0x00,
+    /* Function code */ 0x81,
+    /* Exception code */ 0xFF
```
catenacyber, line 111: > readCoilsErrorRsp: changed to different invalid response code.

Is 0xFF an invalid response code ?
If so, can it appear in the comment ?

src/detect-engine-modbus.c
```diff
@@ -336,11 +117,13 @@ static uint8_t encapsulatedInterfaceTransport[] = {
...
-                               /* Protocol ID */        0x00, 0x00,
-                               /* Length */             0x00, 0x02,
-                               /* Unit ID */            0x00,
-                               /* Function code */      0x12};
+static uint8_t unassigned[] = {
+    /* Transaction ID */ 0x00, 0x0A,
+    /* Protocol ID */ 0x00, 0x00,
+    /* Length */ 0x00, 0x02,
+    /* Unit ID */ 0x00,
+    /* Function code */ 0x3F
```
catenacyber: This change should belong to the commit about fixing tests, right ?

rust/src/modbus/detect.rs
```diff
@@ -0,0 +1,563 @@
...
+use super::modbus::ModbusTransaction;
+use lazy_static::lazy_static;
+use regex::Regex;
+use sawp_modbus::{AccessType, CodeCategory, Data, Flags, FunctionCode, Message};
+use std::ffi::CStr;
+use std::ops::{Range, RangeInclusive};
+use std::os::raw::{c_char, c_void};
+use std::str::FromStr;
+
+lazy_static! {
```
catenacyber, line 27: I wonder if regular expressions are the best way to go here...
What do you think ?
see also https://redmine.openinfosecfoundation.org/issues/4112

src/app-layer-modbus.h
```diff
@@ -34,25 +34,6 @@
...
-    MODBUS_DECODER_EVENT_INVALID_UNIT_IDENTIFIER,
-    MODBUS_DECODER_EVENT_INVALID_FUNCTION_CODE,
-    MODBUS_DECODER_EVENT_INVALID_VALUE,
-    MODBUS_DECODER_EVENT_INVALID_EXCEPTION_CODE,
-    MODBUS_DECODER_EVENT_VALUE_MISMATCH,
-    MODBUS_DECODER_EVENT_FLOODED,
-};
-
 /* Modbus Function Code Categories. */
 #define MODBUS_CAT_NONE                 0x0
```
catenacyber, line 38: These seem unused and could be removed, right ?

src/detect-modbus.c
```diff
@@ -509,15 +90,15 @@ static DetectModbus *DetectModbusUnitIdParse(DetectEngineCtx *de_ctx, const char
...
+    DetectModbusRust *modbus = NULL;
     SigMatch        *sm = NULL;
 
     if (DetectSignatureSetAppProto(s, ALPROTO_MODBUS) != 0)
         return -1;
 
-    if ((modbus = DetectModbusUnitIdParse(de_ctx, str)) == NULL) {
-        if ((modbus = DetectModbusFunctionParse(de_ctx, str)) == NULL) {
-            if ((modbus = DetectModbusAccessParse(de_ctx, str)) == NULL) {
+    if ((modbus = rs_modbus_parse_unit_id(str)) == NULL) {
```
catenacyber, line 97: Should we have only one rust wrapper function for these 3 possibilities ?

src/app-layer-modbus.c
```diff
@@ -37,1503 +37,21 @@
...
-#define MODBUS_TYP_READ_WRITE_MULTIPLE  (MODBUS_TYP_READ | MODBUS_TYP_WRITE | MODBUS_TYP_MULTIPLE)
-
-/* Macro to convert quantity value (in bit) into count value (in word): count = Ceil(quantity/8) */
-#define CEIL(quantity) (((quantity) + 7)>>3)
-
-/* Modbus Default unreplied Modbus requests are considered a flood */
-#define MODBUS_CONFIG_DEFAULT_REQUEST_FLOOD 500
-
 /* Modbus default stream reassembly depth */
 #define MODBUS_CONFIG_DEFAULT_STREAM_DEPTH 0
```
catenacyber, line 45: Only for unit tests, right ?

rust/src/modbus/detect.rs
```diff
@@ -0,0 +1,563 @@
...
+    }
+}
+
+/// Compares a transaction to a signature to determine whether the transaction
+/// matches the signature. If it does, 1 is returned; otherwise 0 is returned.
+#[no_mangle]
+pub extern "C" fn rs_modbus_inspect(tx: &ModbusTransaction, modbus: &DetectModbusRust) -> u8 {
+    let msg = match &tx.request {
+        Some(r) => r,
+        None => match &tx.response {
```
catenacyber, line 206: Does this mean that If a transaction has a request and a response, we only match on the request ?

rust/src/modbus/detect.rs
```diff
@@ -0,0 +1,563 @@
...
+    .unwrap();
+    static ref UNIT_RE: Regex = Regex::new(
+        "^\\s*\"?\\s*unit\\s+([<>]?\\d+)\
+        (?:<>(\\d+))?(?:,\\s*(.*))?\\s*\"?\\s*$"
+    )
+    .unwrap();
+}
+
+#[derive(Debug, PartialEq)]
+pub struct DetectModbusRust {
```
catenacyber, line 49: Should this be more like a C union (a Rust enum) as we cannot have the whole fields set at one time ?
(it is either units, or function ...)

rust/src/modbus/modbus.rs
```diff
@@ -0,0 +1,1577 @@
...
+    }
+
+    pub fn new_tx(&mut self) -> ModbusTransaction {
+        self.tx_id += 1;
+        let mut tx = ModbusTransaction::new(self.tx_id);
+
+        // Check flood limit
+        if REQUEST_FLOOD != 0 && self.tx_id > REQUEST_FLOOD {
+            tx.set_event(ModbusEvent::Flooded);
+            self.givenup = true;
```
catenacyber, line 210: I think we should stop allocating transactions if there is a flood

catenacyber (commented):
This looks good.
I think these are my last questions

doc/userguide/output/eve/eve-json-format.rst
cccs-sadugas: Do you mean that outputing all the data could be too verbose? There are three options:
- don't output any data, maybe only the length
- leave as is, always output data
- add config option to choose what to do

src/app-layer-modbus.c
cccs-sadugas: Yes, it used to be `0x05` which is now a known exception code. We can add a comment for `0xFF`.

src/detect-engine-modbus.c
cccs-sadugas: Yes, that's possible. The code `0x3F` should be unassigned on both the C and Rust parsers.

rust/src/modbus/detect.rs
cccs-sadugas: I think these are two separate discussions. Both were discussed previously in #5790.

**Use of regexes versus nom to parse arguments**
https://github.com/OISF/suricata/pull/5790#discussion_r564789021

**Use of Range Vs. DetectU32Data**
https://github.com/OISF/suricata/pull/5790#discussion_r564790700

My opinion for both would be to leave as-is for now in the interest of merging this PR sooner. Create a generic module for `DetectU32Data` along with a nom parser for arguments as stated in [4112](https://redmine.openinfosecfoundation.org/issues/4112). Once it is available, refactor modbus to use it.

src/app-layer-modbus.c
cccs-sadugas: > If so, can it appear in the comment ?

Do you mean in the git commit comment or as a code comment on the struct?

src/app-layer-modbus.h
cccs-sadugas: Yes, good find. I think they popped back up during a rebase.

src/app-layer-modbus.c
cccs-sadugas: Yes, so we could probably move this inside the `UNITTESTS` block.

rust/src/modbus/detect.rs
cccs-sadugas: The [modbus keyword ](https://suricata.readthedocs.io/en/suricata-6.0.2/rules/modbus-keyword.html?highlight=modbus)supports unit, access and address inspection. These fields are guaranteed to be the same because of how [matches ](https://docs.rs/sawp-modbus/0.4.0/src/sawp_modbus/lib.rs.html#874-1045) associates a response to a request.

rust/src/modbus/detect.rs
cccs-sadugas: I think we can have more than one field set according to the [modbus keyword documentation](https://suricata.readthedocs.io/en/suricata-6.0.2/rules/modbus-keyword.html?highlight=modbus). This is why we chose a `struct`.
```
modbus: unit 10<>20, access write holding, address 500, value >200    
```

rust/src/modbus/detect.rs
cccs-sadugas: Looking at it again there is also `value`, which may only be present in either the request or response.

doc/userguide/output/eve/eve-json-format.rst
catenacyber: > Do you mean that outputing all the data could be too verbose?

Indeed, that is what I am wondering.
And there are indeed the 3 options.
What is your feeling about the verbosity of these ?

src/app-layer-modbus.c
catenacyber: a comment in the struct :
`/* Exception code */ 0xFF` looks like 0xFF is a valid exception code

rust/src/modbus/detect.rs
catenacyber: Thanks for these explanations.
Could you put comments in the code about this ?
So that future extensions to the modbus keyword do not break these expectations...

rust/src/modbus/detect.rs
catenacyber: Ok.

But then `rs_modbus_inspect` seems to do too early `return` like `return inspect_data(msg, modbus) as u8;`
Cannot we have `modbus.access_type` and `modbus.category` set ?

doc/userguide/output/eve/eve-json-format.rst
cccs-sadugas: Max length of data will be `255` bytes ad determined by the byte count field, so I would say it is fine for now but an improvement (i.e. new ticket) would be to make this configurable?

rust/src/modbus/detect.rs
catenacyber: Thanks for bringing back the pointers. I had a feeling of déjà-vu when saying this.
Fair enough for me

doc/userguide/output/eve/eve-json-format.rst
catenacyber: So, ok for now, and no need to create a new ticket unless the need arises.
(my question is not about the size of one single entry, but also about how often these entries may pop up)

rust/src/modbus/detect.rs
cccs-sadugas: This functionality was ported from the C parser and refactored to use early returns instead of nested ifs but it should have the same effect.

See [detect-engine-modbus.c#L199](https://github.com/OISF/suricata/blob/master/src/detect-engine-modbus.c#L199)

doc/userguide/output/eve/eve-json-format.rst
cccs-sadugas: Some real world applications have a very small ratio of large entries. Sample pcaps found on the web may have 50% large entries.

cccs-sadugas:
Replaced by #6090

rust/src/modbus/detect.rs
catenacyber: The C behavior looks confusing to me.

Is it possible to have a signature like `alert modbus any any -> any any (msg:"Modbus function number test"; modbus: function 1, access read coils; sid:1; rev:1;)` ?
If not, why ?

cccs-sadugas: I think the documentation is a bit more clear than the C code: https://suricata.readthedocs.io/en/suricata-6.0.0/rules/modbus-keyword.html

The function and access keywords cannot be used together, probably because one is more specific than the other. I.e. `function 1` is more specific than `access read coils`.

catenacyber: Ok thanks :-)

PR #6061, opened 2021-04-16
ike: set event for multiple server proposals

fhonza:
https://redmine.openinfosecfoundation.org/issues/4433

Describe changes:
- in case of multiple chosen server transforms set an event, log them and don't panic on debug-validate

### DIFF
--- a/rust/src/ike/ikev1.rs
+++ b/rust/src/ike/ikev1.rs
@@ -136,6 +136,11 @@ pub fn handle_ikev1(
                         &tx.hdr.ikev1_header.vendor_ids,
                         &tx.hdr.ikev1_transforms,
                     );
+
+                    if state.ikev1_container.server.transforms.len() > 1 {
+                        SCLogDebug!("More than one chosen server proposal");
+                        state.set_event(IkeEvent::MultipleServerProposal);
+                    }
                 }
 
                 if rem.len() > 0 {
--- a/rust/src/ike/logger.rs
+++ b/rust/src/ike/logger.rs
@@ -73,14 +73,19 @@ fn log_ike(
     }
 
     if tx.ike_version == 1 {
-        let mut index = 0;
-        for server_transform in &state.ikev1_container.server.transforms {
-            if index >= 1 {
-                debug_validate_bug_on!(true);
-                break;
+        if state.ikev1_container.server.transforms.len() > 0 {
+            // log the first transform as the chosen one
+            add_attributes(&state.ikev1_container.server.transforms[0], jb)?;
+        }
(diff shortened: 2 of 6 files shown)


victorjulien:
The commit check CI issue looks like something we can ignore.

rust/src/ike/ikev1.rs
```diff
@@ -136,6 +136,11 @@ pub fn handle_ikev1(
                         &tx.hdr.ikev1_header.vendor_ids,
                         &tx.hdr.ikev1_transforms,
                     );
+
+                    if state.ikev1_container.server.transforms.len() > 1 {
```
catenacyber, line 140: Should we set the event only once we cross the threshold ?
Something like `state.ikev1_container.server.transforms.len() <= 1 ` and `tx.hdr.ikev1_transforms.len() + state.ikev1_container.server.transforms.len()  > 1` before calling update ?

rust/src/ike/logger.rs
```diff
@@ -73,14 +73,19 @@ fn log_ike(
...
-            if index >= 1 {
-                debug_validate_bug_on!(true);
-                break;
+        if state.ikev1_container.server.transforms.len() > 0 {
+            // log the first transform as the chosen one
+            add_attributes(&state.ikev1_container.server.transforms[0], jb)?;
+        }
+        if state.ikev1_container.server.transforms.len() > 1 {
+            // in case we have multiple server transforms log them in a list
+            jb.open_array("server_proposals")?;
```
catenacyber, line 82: I wonder if we should have an S-V test for this

catenacyber (commented):
I think we would prefer to have only one single event per flow when we cross the threshold, and not one event every time we are over the threshold...
What do you think @satta ?

satta:
> I think we would prefer to have only one single event per flow when we cross the threshold, and not one event every time we are over the threshold...
> What do you think @satta ?

Yes, right. I think one event should be enough when we reach >1 proposals.

rust/src/ike/ikev1.rs
fhonza: yes, good idea

rust/src/ike/logger.rs
catenacyber: Maybe the doc should be updated as well...

fhonza: > I wonder if we should have an S-V test for this

Not sure about that, at least it seems to be a rare case. (not in the rfc and the first pcap I saw with this property)

satta:
> typo in `parameters`

Ah, yep. Will fix that in next PR.

PR #6064, opened 2021-04-17
Multi-threaded EVE support for output plugins

jlucovsky:
Continuation of #6017

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4126](https://redmine.openinfosecfoundation.org/issues/4126)

Describe changes:
These changes build upon the experimental file output plugin support by
adding calls when a thread attaches and detaches from an output context:
- Address review comments
- Remove unneeded per-thread de-initialization call from each output file.
- 
I've modified @jasonish's output plugin with a multi-threaded variant: https://github.com/jlucovsky/suricata-example-plugins/tree/output/1/eve-file-type

Configure Suricata to use this plugin by adding the following to `suricata.yaml`

```
plugins:
    - /path/to/plugin.so
```

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #6065, opened 2021-04-17
Multi-threaded EVE support for output plugins 

jlucovsky:
Continuation of #6064

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4126](https://redmine.openinfosecfoundation.org/issues/4126)

Describe changes:
These changes build upon the experimental file output plugin support by
adding calls when a thread attaches and detaches from an output context:
- Address review comments
- Remove unneeded per-thread de-initialization call from each output file.
- clang-format fix

I've modified @jasonish's output plugin with a multi-threaded variant: https://github.com/jlucovsky/suricata-example-plugins/tree/output/1/eve-file-type

Configure Suricata to use this plugin by adding the following to `suricata.yaml`

```
plugins:
    - /path/to/plugin.so
```

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -1109,20 +1109,19 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
             json_ctx->file_ctx->prefix_len = strlen(prefix);
         }
 
+        /* Threaded file output */
+        const ConfNode *threaded = ConfNodeLookupChild(conf, "threaded");
+        if (threaded && threaded->val && ConfValIsTrue(threaded->val)) {
+            SCLogConfig("Threaded EVE logging configured");
+            json_ctx->file_ctx->threaded = true;
+        } else {
+            json_ctx->file_ctx->threaded = false;
+        }
+
         if (json_ctx->json_out == LOGFILE_TYPE_FILE ||
             json_ctx->json_out == LOGFILE_TYPE_UNIX_DGRAM ||
             json_ctx->json_out == LOGFILE_TYPE_UNIX_STREAM)
         {
-            if (json_ctx->json_out == LOGFILE_TYPE_FILE) {
-                /* Threaded file output */
-                const ConfNode *threaded = ConfNodeLookupChild(conf, "threaded");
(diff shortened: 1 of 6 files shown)


src/output-json.c
```diff
@@ -1179,27 +1178,51 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
...
+                append = DEFAULT_LOG_MODE_APPEND;
+            }
+            if (json_ctx->file_ctx->threaded) {
+                if (!SCLogOpenThreadedFile(
+                            ConfigGetLogDirectory(), append, json_ctx->file_ctx, 1)) {
+                    SCFree(json_ctx);
+                    SCFree(output_ctx);
+                    return result;
+                }
+            }
```
jasonish, line 1203: Is this needed for plugins?

jlucovsky: Yes -- this is where Suricata keeps track of the per-thread data blobs. `lf_slots` contains a blob for each thread -- the plugin threadinit returns the data blob tracked by Suricata.

jasonish: Ok.  I commented it out and things seem to work OK.  I think it needs a comment, cause the `append`, which is not really needed in the plugin block and naming make it sound out of place in this section of code.

jasonish:
Other than the one clarification, I think we're good to go on this.  As it touches things at a higher level than just plugins, I've test stuff like normal file rotation while threaded, and all seems good.

src/output-json.c
jasonish: > Ok. I commented it out and things seem to work OK. I think it needs a comment, cause the `append`, which is not really needed in the plugin block and naming make it sound out of place in this section of code.

My error.. Must have had threading off when I gave this a try.

jlucovsky:
Continued in #6073

PR #6069, opened 2021-04-20
Tcp protodetect bail 4171 v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4171

Describe changes:
- `TCPProtoDetectCheckBailConditions` used `STREAM_RIGHT_EDGE` which is not the size already processed for protocol detection, but the far edge of the stream where reassembly is not finished yet. Fix is to use `next_seq` rather than `segs_right_edge`
- `HandleEstablishedPacketToServer` was updating too much `next_seq` in the case we have reassembly to do after a gap. Fix is to update only up to `last_ack` in this specific case
- Remove `midstream` case from debug assertion

PR #6072, opened 2021-04-22

victorjulien:
For size limit checks consider only available data at the stream start
and before any GAPS.

The old check would consider too much data if there were temporary gaps,
like when a data packet was in-window but (far) ahead of the expected
segment.

Alternative to #6069

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -205,11 +205,9 @@ static void TCPProtoDetectCheckBailConditions(ThreadVars *tv,
         return;
     }
 
-    const uint64_t size_ts = STREAM_HAS_SEEN_DATA(&ssn->client) ?
-        STREAM_RIGHT_EDGE(&ssn->client) : 0;
-    const uint64_t size_tc = STREAM_HAS_SEEN_DATA(&ssn->server) ?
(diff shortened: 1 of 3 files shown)


victorjulien:
@catenacyber this takes a simpler approach by only setting the available data size to what we have ready at the start of the stream. It will only consider data up to a GAP. I'm not entirely sure if that makes sense here.

catenacyber:
It looks good. But I do not know enough about the streaming buffer/ red and black tree to tell if it is perfect.

victorjulien:
I think the RB logic is fine. The thing I'm doubtful about is how GAPs are not considered. If we get a GAP < 100k we can not trigger the size conditions in that direction. It would be great to have a set of tests for all the "bail" cases here.

catenacyber:
> I think the RB logic is fine. The thing I'm doubtful about is how GAPs are not considered. If we get a GAP < 100k we can not trigger the size conditions in that direction. It would be great to have a set of tests for all the "bail" cases here.

You could use the 2 different sizes (the previous one and the new one) with different thresholds to bail out (and debug validate for the relevant cases)

PR #5340, opened 2020-08-30
Updated queue.h to fix build errors using clang10 on FreeBSD13.0-CURRENT

jbarfield:
Updated "queue.h" headers to fix build errors using clang10 on FreeBSD13-CURRENT.

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [NA] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: NA

Describe changes:

Updated Suricata included "queue.h" headers to resolve build errors on FreeBSD12.1-STABLE & 13-CURRENT & HardenedBSD-12.1-STABLE. 

The original included "queue.h" from 6.0.0-beta1 had numerous errors regarding types, identifiers, and pointers making it impossible to build from source on the aforementioned OS platform. We initially updated "app-layer-expectation.c functions directly" but it was discovered by OISF that there were new build errors since implemented in a function named ctime. After further research into the issue it was determined that "queue.h" was the culprit of build issues across the app-layers and hence the need for an updated "queue.h" attached to this pull request. 

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

I am having problems getting the pr script to run on FreeBSD12 and can't find much documentation about it. I've run it with both python 2.7 and python 3.7 with mixed results but it never runs. It either complains about docker or complains that a remote push has not happened with the branch. Very strange behaviors have been observed. I am assuming that this is meant for linux/docker. 

I hope that I have properly submitted this pull request this time. 

Have a great day!

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #6075, opened 2021-04-22
misc: include queue.h before other headers - v1

jasonish:
At least on FreeBSD, some other include is including "sys/queue.h"
which results in FreeBSDs /usr/include/sys/queue.h being picked
up and setting `__SYS_QUEUE_H__` so our queue.h is not picked up.

But the FreeBSD queue.h does not have the `CIRCLEQ` definitions. To
fix just include our queue.h first, which also sets `__SYS_QUEUE_H__`
preventing the system one from being picked up.

This is a different take on https://github.com/OISF/suricata/pull/5340.

### DIFF
--- a/src/app-layer-expectation.c
+++ b/src/app-layer-expectation.c
@@ -52,6 +52,7 @@
  * \author Eric Leblond <eric@regit.org>
  */
 
+#include "queue.h"
 #include "suricata-common.h"
 #include "debug.h"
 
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged in #6078, thanks!

jasonish:
Redmine issue: https://redmine.openinfosecfoundation.org/issues/4442

PR #6076, opened 2021-04-22
tx: fix unidirection transaction cleanup (dns tx fix) - v4

jasonish:
Fir commit fixes unidirection transaction handling. We only want to mark a TX as skipped if skipped in both the TS and TC side if unidirectional.  If one side is seen, then its complete.  Instead it has been marked as skipped, so the starting TX ID on cleanup is never incremented, resulting in a larger number of iterations as the transaction grow on a session.

This was easily seen with DNS when multiple transactions occurred on the same flow before the flow timed out. In many cases this would not be noticed, but is really notiecable when there are devices present that blast DNS requests over the same session.

Note: You must have some rules enabled to hit this case, it doesn't appear to show up if you have no rules enabled.

The second commit just removes the DNS flood protection which appears is not needed, as it only kicks in when it should not kick in.

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/4437

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -114,19 +114,6 @@ pub const DNS_RCODE_BADALG:   u16 = 21;
 pub const DNS_RCODE_BADTRUNC: u16 = 22;
 
 
-/// The maximum number of transactions to keep in the queue pending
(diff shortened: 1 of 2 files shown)


victorjulien:
Thanks!

PR #6092, opened 2021-04-28
detect: runs pkt checks every time

catenacyber:

Describe changes:

suricata-verify-pr: 493

Analyzing what happens on suricata-verify test 
(maybe as the flow bit is not set at the start of the transaction), we run `DetectRunStoreStateTx` at packet 5
At packet 6, we skip `DetectEnginePktInspectionRun` but we do not have any file yet, so we do not trigger sid 3
At packet 7 we set the flow bit Sith the signature sid 1
We do not trigger sid 3 at packet 7 because `stored_flags == NULL` (packet in opposite direction) in `DetectRunTxInspectRule` and so we call `DetectEnginePktInspectionRun` which says no match as the flow bit is set
And we do trigger sid 3 at packet 8 because `stored_flags!=NULL` (because we called `DetectRunStoreStateTx` at packet 5), so we do not call `DetectEnginePktInspectionRun` to check the flow bit, so we just check the filemagic which says match

Interestingly, the bug does not happen for the flow bit isset because we do `s->flags |= SIG_FLAG_REQUIRE_FLOWVAR` for this case and this flag is checked by `DetectRunInspectRuleHeader`

Feedback welcome here

PS : This was the oldest S-V PR

### DIFF
--- a/src/detect.c
+++ b/src/detect.c
@@ -1084,15 +1084,15 @@ static bool DetectRunTxInspectRule(ThreadVars *tv,
             TRACE_SID_TXS(s->id, tx, "DetectRunInspectRuleHeader() no match");
             return false;
         }
-        if (DetectEnginePktInspectionRun(tv, det_ctx, s, f, p, NULL) == false) {
-            TRACE_SID_TXS(s->id, tx, "DetectEnginePktInspectionRun no match");
-            return false;
-        }
         /* stream mpm and negated mpm sigs can end up here with wrong proto */
         if (!(AppProtoEquals(s->alproto, f->alproto) || s->alproto == ALPROTO_UNKNOWN)) {
             TRACE_SID_TXS(s->id, tx, "alproto mismatch");
             return false;
         }
+    }
+    if (DetectEnginePktInspectionRun(tv, det_ctx, s, f, p, NULL) == false) {
+        TRACE_SID_TXS(s->id, tx, "DetectEnginePktInspectionRun no match");
+        return false;
     }
 
     const DetectEngineAppInspectionEngine *engine = s->app_inspect;


catenacyber:
https://github.com/OISF/suricata-verify/pull/492

src/detect.c
```diff
@@ -1084,16 +1084,16 @@ static bool DetectRunTxInspectRule(ThreadVars *tv,
...
-            TRACE_SID_TXS(s->id, tx, "DetectEnginePktInspectionRun no match");
-            return false;
-        }
         /* stream mpm and negated mpm sigs can end up here with wrong proto */
         if (!(AppProtoEquals(s->alproto, f->alproto) || s->alproto == ALPROTO_UNKNOWN)) {
             TRACE_SID_TXS(s->id, tx, "alproto mismatch");
             return false;
         }
     }
+    if (DetectEnginePktInspectionRun(tv, det_ctx, s, f, p, NULL) == false) {
```
victorjulien, line 1093: This does not look correct. I did not understand the explanation, though. Maybe some of the checks are in the wrong inspection list?

catenacyber: >  I did not understand the explanation, though.

What did you not understand in the sequence of events produced by the packets ?

> Maybe some of the checks are in the wrong inspection list?

Not sure what an inspection list refers to. Is it something like `DETECT_SM_LIST_MATCH` ?
Then it is `DETECT_SM_LIST_MATCH` for `flowbits` and `g_file_match_list_id` for `file magic`

Otherwise, `flowbits` is run by `DetectEnginePktInspectionRun` because it has a `Match` function, and no `AppLayerTxMatch

> This does not look correct

I thought this was just a performance optimization not to run the check every time
I hoped the CI would catch this if it were not correct.
What should crash in this case ?

victorjulien: The concern is performance. Running inspection of things per packet instead of once per tx is potentially a huge cost.

The issue here seems to be that the flowbit logic isn't run at the expected moment. Can you look into ways to get it to run it at the right time w/o brute forcing it?

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6126

PR #6093, opened 2021-04-29
detect: try to get tx_id for packet alerts

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3354

Describe changes:
- Try to get a better tx_id for packet alerts, so as to get the right app-layer metadata

Uses `AppLayerParserGetTransactionInspectId` to do so.

I am not sure if there is a better way. For HTTP2, we may have multiple transactions going on at the same moment...

suricata-verify-pr: 495

https://github.com/OISF/suricata-verify/pull/495

### DIFF
--- a/src/detect.c
+++ b/src/detect.c
@@ -804,8 +804,14 @@ static inline void DetectRulePacketRules(
 
         if (!(sflags & SIG_FLAG_NOALERT)) {
             /* stateful sigs call PacketAlertAppend from DeStateDetectStartDetection */
-            if (!state_alert)
-                PacketAlertAppend(det_ctx, s, p, 0, alert_flags);
+            if (!state_alert) {
+                uint64_t tx_id = 0;
+                if (pflow && pflow->alparser) {
+                    tx_id = AppLayerParserGetTransactionInspectId(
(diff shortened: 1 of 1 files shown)


src/detect.c
```diff
@@ -804,8 +804,14 @@ static inline void DetectRulePacketRules(
 
         if (!(sflags & SIG_FLAG_NOALERT)) {
             /* stateful sigs call PacketAlertAppend from DeStateDetectStartDetection */
-            if (!state_alert)
-                PacketAlertAppend(det_ctx, s, p, 0, alert_flags);
+            if (!state_alert) {
```
victorjulien, line 807: if the alert doesn't happen in a tx, which is the case here, we can't know the tx id for sure

catenacyber: Do you mean that, in this case, we should not have any app-layer metadata as they come from a transaction ?

victorjulien: Essentially yes. I guess we could see if there is a single tx and use that as a special case. But if there are more txs we'd risk logging the wrong tx causing confusion.

catenacyber: Thanks, closing this for now, and moving to the redmine ticket for further discussion

PR #6103, opened 2021-05-04
dcerpc: unify applayer registration logic

inashivb:
https://redmine.openinfosecfoundation.org/issues/3315


inashivb:
Hi @jasonish !
I was at first trying to add a pattern probe function the the parser register function but then Victor pointed that it misses cases like SMB where more than one patterns are required to be registered. We have now done it differently and want your opinion if this is in fact the right way to add a pattern probe function.

victorjulien:
It's not just that there are multiple patterns and probing parsers that are registered, its also that the registrations have quite a few settings too. So putting all these in the RustParser table would make that very big.

jasonish:
I wouldn't try to fit them in the registration struct.  Separate function calls after that would be ok.  Maybe some days we'll have a nice fluent builder pattern.

inashivb:
> I wouldn't try to fit them in the registration struct. Separate function calls after that would be ok. Maybe some days we'll have a nice fluent builder pattern.

Thank you!

inashivb:
Replaced with #6122

PR #6105, opened 2021-05-04
yaml: treat some unquoted values as null (per spec) - v1

jasonish:
Per the YAML spec, the following values when present unquoted
should be equivalent to null:
- ~
- NULL
- Null
- null

Redmine issue:
https://redmine.openinfosecfoundation.org/issues/4472


jlucovsky (approved):

victorjulien:
Merged in #6109, thanks!

PR #6114, opened 2021-05-11
MQTT: improve handling of many topics -- v2

satta:
Previous PR: #6113 

Describe changes:
- Use default value for limits if not specified in the configuraton.
- Do not use separate commit to fix formatting.

### DIFF
--- a/rust/src/mqtt/detect.rs
+++ b/rust/src/mqtt/detect.rs
@@ -353,7 +353,7 @@ pub extern "C" fn rs_mqtt_tx_get_publish_message(
 
 #[no_mangle]
 pub extern "C" fn rs_mqtt_tx_get_subscribe_topic(tx: &MQTTTransaction,
-    i: u16,
+    i: u32,
     buf: *mut *const u8,
     len: *mut u32)
     -> u8
@@ -386,7 +386,7 @@ pub extern "C" fn rs_mqtt_tx_get_subscribe_topic(tx: &MQTTTransaction,
 
 #[no_mangle]
 pub extern "C" fn rs_mqtt_tx_get_unsubscribe_topic(tx: &MQTTTransaction,
-    i: u16,
+    i: u32,
     buf: *mut *const u8,
     len: *mut u32)
     -> u8
--- a/src/detect-mqtt-subscribe-topic.c
+++ b/src/detect-mqtt-subscribe-topic.c
@@ -78,8 +78,8 @@ static InspectionBuffer *MQTTSubscribeTopicGetData(DetectEngineThreadCtx *det_ct
 
     const uint8_t *data;
     uint32_t data_len;
-    if (rs_mqtt_tx_get_subscribe_topic(cbdata->txv, (uint16_t)cbdata->local_id,
-                &data, &data_len) == 0) {
+    if (rs_mqtt_tx_get_subscribe_topic(cbdata->txv, (uint32_t)cbdata->local_id, &data, &data_len) ==
+            0) {
         return NULL;
     }
 
--- a/src/detect-mqtt-subscribe-topic.c
+++ b/src/detect-mqtt-subscribe-topic.c
@@ -82,7 +82,7 @@ static InspectionBuffer *MQTTSubscribeTopicGetData(DetectEngineThreadCtx *det_ct
             0) {
         return NULL;
     }
-
+    buffer->inspect = data;
     InspectionBufferSetup(det_ctx, list_id, buffer, data, data_len);
     InspectionBufferApplyTransforms(buffer, transforms);
 
--- a/src/detect-mqtt-subscribe-topic.c
+++ b/src/detect-mqtt-subscribe-topic.c
@@ -58,6 +58,8 @@ static int DetectMQTTSubscribeTopicSetup(DetectEngineCtx *, Signature *, const c
 
 static int g_mqtt_subscribe_topic_buffer_id = 0;
 
+static int subscribe_topic_match_limit = 100;
+
 struct MQTTSubscribeTopicGetDataArgs {
     int local_id;
     void *txv;
@@ -102,7 +104,7 @@ static int DetectEngineInspectMQTTSubscribeTopic(
         transforms = engine->v2.transforms;
     }
 
-    while(1) {
+    while ((subscribe_topic_match_limit == 0) || local_id < subscribe_topic_match_limit) {
         struct MQTTSubscribeTopicGetDataArgs cbdata = { local_id, txv, };
         InspectionBuffer *buffer = MQTTSubscribeTopicGetData(det_ctx,
             transforms, f, &cbdata, engine->sm_list, false);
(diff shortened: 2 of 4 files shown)


src/detect-mqtt-subscribe-topic.c
```diff
@@ -153,7 +155,7 @@ static void PrefilterTxMQTTSubscribeTopic(DetectEngineThreadCtx *det_ctx,
     const int list_id = ctx->list_id;
 
     int local_id = 0;
-    while(1) {
+    while ((subscribe_topic_match_limit == 0) || local_id < subscribe_topic_match_limit) {
```
victorjulien, line 158: this enforces the limit in detection, but will still parse all the topics and also still log all of them. Is that desired? I'd expect a limit like this to be enforced during parsing to avoid resource starvation during all processing.

src/detect-mqtt-subscribe-topic.c
```diff
@@ -82,7 +82,7 @@ static InspectionBuffer *MQTTSubscribeTopicGetData(DetectEngineThreadCtx *det_ct
             0) {
         return NULL;
     }
-
+    buffer->inspect = data;
```
victorjulien, line 87: these are wrong. I will solve this issue differently, no action on your side required

src/detect-mqtt-subscribe-topic.c
```diff
@@ -78,8 +78,8 @@ static InspectionBuffer *MQTTSubscribeTopicGetData(DetectEngineThreadCtx *det_ct
 
     const uint8_t *data;
     uint32_t data_len;
-    if (rs_mqtt_tx_get_subscribe_topic(cbdata->txv, (uint16_t)cbdata->local_id,
-                &data, &data_len) == 0) {
+    if (rs_mqtt_tx_get_subscribe_topic(cbdata->txv, (uint32_t)cbdata->local_id, &data, &data_len) ==
```
victorjulien, line 83: this uint32_t cast looks suspicious. Why not update the `local_id` var to actually be `uint32_t` in `MQTTSubscribeTopicGetDataArgs`?

src/detect-mqtt-unsubscribe-topic.c
```diff
@@ -78,8 +78,8 @@ static InspectionBuffer *MQTTUnsubscribeTopicGetData(DetectEngineThreadCtx *det_
 
     const uint8_t *data;
     uint32_t data_len;
-    if (rs_mqtt_tx_get_unsubscribe_topic(cbdata->txv, (uint16_t)cbdata->local_id,
-                &data, &data_len) == 0) {
+    if (rs_mqtt_tx_get_unsubscribe_topic(
+                cbdata->txv, (uint32_t)cbdata->local_id, &data, &data_len) == 0) {
```
victorjulien, line 84: same here

src/detect-mqtt-subscribe-topic.c
satta: From my point of view, this is desired. For logging, I would always want to see everything that is going on, as I would consider this the main use case for this parser. If you disagree, we could surely also make this an option in the `app-layer.mqtt` config section.

src/detect-mqtt-subscribe-topic.c
satta: Got it -- I just added this to avoid a different issue pointed out by @catenacyber

src/detect-mqtt-subscribe-topic.c
satta: Good point. I guess that's what you get for starting off other code, you don't see the control you actually have. Of course that's the cleanest way to do it.

src/detect-mqtt-unsubscribe-topic.c
satta: see above

src/detect-mqtt-subscribe-topic.c
satta: OK with leaving parsing as is for now? Then I will open a new PR.

catenacyber: I am not sure that having many topics in one transaction is the most expensive way, ie to get resource starvation. I think it is more effective, from an adversary, to send the same packet with a single topic multiple times...
What do you think @satta ?

catenacyber (changes requested):
The uint32_t cast needs indeed to go

src/detect-mqtt-subscribe-topic.c
satta: Done for next PR

satta:
> The uint32_t cast needs indeed to go

Done. Just waiting for a decision about Victor's point above before sending a new PR.

src/detect-mqtt-subscribe-topic.c
satta: I tend to agree but have not measured it.
I just wanted to say that for logging I would really _not_ expect any truncation beyond the cutoff based on raw message size.

victorjulien: is there a max number of topics defined by the protocol?

satta: There is, and it is implicitly defined by the maximal size of a "message" and the space requirement per topic. See https://github.com/OISF/suricata/pull/5622#discussion_r628123516 -- about 67M topics max in the SUBSCRIBE case, for example.

rust/src/mqtt/detect.rs
```diff
@@ -353,7 +353,7 @@ pub extern "C" fn rs_mqtt_tx_get_publish_message(
 
 #[no_mangle]
 pub extern "C" fn rs_mqtt_tx_get_subscribe_topic(tx: &MQTTTransaction,
-    i: u16,
+    i: u32,
```
jlucovsky, line 356: Is the message id 32 or 16 bits?

This states 16-bits but perhaps I'm misreading?  https://public.dhe.ibm.com/software/dw/webservices/ws-mqtt/mqtt-v3r1.html#msg-id

satta: You are referring to the internal message ID used to connect multiple items in a transaction. The `i` here, however, is a local ID coming from the C side of the detect code: https://github.com/OISF/suricata/blob/9ef13d309bae377b10e0554b8d37471518af1580/src/detect-mqtt-subscribe-topic.c#L83

There's no direct connection to the message ID, other than that the range of the message ID should set a limit for the the maximum number of items to be traversed. But I wanted to rather be safe than sorry and just use an `int` here.

satta: Sorry, did not notice your 👍🏻  here -- does this mean we can resolve this question?

src/detect-mqtt-subscribe-topic.c
satta: @victorjulien @catenacyber can we resolve this discussion or do you think I should also limit topic parsing in addition to detection?

rust/src/mqtt/detect.rs
catenacyber: For me, it is ok.
@jlucovsky is it ok for you ?

src/detect-mqtt-subscribe-topic.c
catenacyber: For me it is ok

satta:
New PR: #6190

PR #6119, opened 2021-05-14
flow: provide flags accessor function - v1

jasonish:
Add an accessor function for flow flags. To be used by Rust where the flow struct is an opaque data type.


victorjulien:
Looks good to me. I think @inashivb will include it in the PR what depends on it, so I'm closing this here.

inashivb:
Used these and added some more required fns in the same commit in PR https://github.com/OISF/suricata/pull/6206

PR #6123, opened 2021-05-17
doc: update sphinx api to use add_css_file

catenacyber:
None

Describe changes:
- update sphinx api to use `add_css_file` instead of `add_stylesheet`

cf https://github.com/sphinx-doc/sphinx/issues/7747

### DIFF
--- a/doc/devguide/conf.py
+++ b/doc/devguide/conf.py
@@ -129,7 +129,7 @@ if not on_rtd:
     except:
         html_theme = 'default'
     def setup(app):
-        app.add_stylesheet('css/suricata.css')
+        app.add_css_file('css/suricata.css')
 else:
     html_context = {
         'css_files': [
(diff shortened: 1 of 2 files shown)


catenacyber:
Looks like CI is using old sphinxes...
How do we do python code compatible with both old and newest version ?

jasonish:
> How do we do python code compatible with both old and newest version ?

I think we need to settle on whatever Python we have available to us in the distrubtion we choose to build the release tarballs.  As this will require running Sphinx to generate the man pages we need to ship in the release tarballs.

catenacyber:
Is there a way _not_ to build the docs ?

Replaced by https://github.com/OISF/suricata/pull/6127

jasonish:
> Is there a way _not_ to build the docs ?

Not if you have Sphinx installed.  So I guess `--without-docs` might be a good idea (or should that be a `--disable-docs`).

PR #6129, opened 2021-05-19
suricata-plugin.h: don't include autoconf.h - v1

jasonish:
It is not required here and just creates double inclusion in some
scenarios.


victorjulien:
Merged in #6130, thanks!

PR #6141, opened 2021-05-25
Http2 keywords 60x v4

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4067

Describe changes:
- Allows some HTTP1 keywords one HTTP2 traffic

Only if the signature is `alert http2`
`alert http` or `alert ip` remain only HTTP1

https://github.com/OISF/suricata/pull/6039 with rebasing and adding the latest keywords cf #6087

The first commit is the one really different from the current dev branch
The other ones are almost the same (except there is no `ALPROTO_HTTP1`)

I guess we need to have `alert http` work on HTTP2 to get good S-V tests

PR #6143, opened 2021-05-25
6.0.x - build fixups - v1

jasonish:
- makefile: don't include the whole test/ directory
- rust: pin memchr to ~2.3

PR #6140, opened 2021-05-25
Http2 decompression 60x 4501 v1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4501
Backport of https://redmine.openinfosecfoundation.org/issues/4277 (which kinds of belong to 4501)
https://redmine.openinfosecfoundation.org/issues/4417
Backport of https://redmine.openinfosecfoundation.org/issues/4494

Describe changes:
- Handles file decompression in HTTP2
- Fix HTTP2 dynamic headers table eviction 
- Remove wrong assertion

No need to add a new suricata-verify PR as the check is done through suricata feature `HTTP2_DECOMPRESSION`

PR #6144, opened 2021-05-26
Http2 keywords 60x v5

catenacyber:

Describe changes:
- Allows some HTTP1 keywords one HTTP2 traffic

A configuration option allows to have all HTTP1 rules working on HTTP2 traffic without specifying it for each rule

Modifies https://github.com/OISF/suricata/pull/6141 with 
- allowing `alert http` and `alert ip` to work on HTTP2 traffic

With Signature alproto and Flow alproto, the options were 
- have only one alproto for HTTP1 and HTTP2, but this needed a major rework (as I see it)
- Duplicate signatures so that a HTTP1 signature has also a HTTP2 version, as is done with bidirectional signatures, but that linked list code has been a trap for bugs, so I avoided touching it
- Have a signature alproto match multiple Flow alproto : That is the solution I chose as in master. The difference with master is that we do not have HTTP_ANY, just a config option to tell that all HTTP1 rules should try to match on HTTP2 traffic

suricata-verify-pr: 503

Makes S-V test in OISF/suricata-verify#503 pass

The S-V tests `http2-keywords` and `http2-keywords2` still require version 7 because they have rules with `alert http1` which is not recognized here.
What could we do about it ?
- have a signature keyword to load the rule only if the suricata version is good for it ?
- allow to load some rules even if some other rules error ?


catenacyber:
Also fixed the commit-check failure by squashing 2 commits together

catenacyber:
CIFuzz failure should be fixed by #6140

catenacyber:
Needs #6143 to get the failing builds to work

src/app-layer-parser.c
```diff
@@ -215,10 +215,16 @@ void AppLayerParserStateFree(AppLayerParserState *pstate)
...
 }
 
+bool g_config_http1keywords_http2traffic = false;
+
 int AppLayerParserSetup(void)
 {
     SCEnter();
     memset(&alp_ctx, 0, sizeof(alp_ctx));
+    int value = 0;
+    if (ConfGetBool("app-layer.protocols.http.rules-on-http2", &value) == 1 && value == 1) {
```
victorjulien, line 225: I think it would make more sense to make this part of `app-layer.protocols.http2`.

catenacyber: indeed, will do

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6152

PR #6133, opened 2021-05-21
plugin/syslog: Make syslog an internal plugin

jlucovsky:
Continuation of #6132 

This PR changes the syslog output type into a plugin -- an internal plugin -- to take advantage of the EVE output plugin framework. 

Describe changes:
- clang format issue

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #6145, opened 2021-05-26
Eve/Syslog - Refactor to use the file type plugin API (but not as a plugin)

jasonish:
Builds on Jeff's work to convert the Eve syslog output to an "internal"
plugin. This PR just renames a few things as I don't think "internal
plugins" should be a thing. Instead the idea of plugins should be
exposing internal APIs via plugin interface, and here we are more or
less converting the syslog file type to use new APIs that were
introduced as part of the plugin loading.

Thoughts?

Jeff's PR: https://github.com/OISF/suricata/pull/6133
Changes required for existing plugins: https://github.com/jasonish/suricata-example-plugins/pull/2/files

### DIFF
--- a/output-eve-syslog.c
+++ b/output-eve-syslog.c
@@ -1,0 +1,114 @@
+/* vi: set et ts=4: */
+/* Copyright (C) 2021 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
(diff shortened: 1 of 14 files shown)


jlucovsky:
I have no opposition to the renaming that you've done.

jasonish:
Pulled out of draft status.

jlucovsky (approved):
LGTM

jasonish:
Continued here: https://github.com/OISF/suricata/pull/6161

PR #6150, opened 2021-05-31
Http2 keywords norm 4067 v1

catenacyber:

Describe changes:
- http2: turn `Host header into `authority` during upgrade
- http2: `http.host` normalized keyword now works for HTTP2
- http2: `http.header_names` normalized keyword now works for HTTP2
- http2: `http.header` keyword now works for HTTP2 as well as `http.header.raw`
- http2: concatenate one headers multiple values
- http2: generic http2_header_blocks (bonus as some cases were forgotten for continuation and push_promise)

suricata-verify-pr: 504

We do not escape linefeeds or `:` in HTTP2 header names even if they can syntactically be present. Is it ok ?

### DIFF
--- a/rust/src/http2/detect.rs
+++ b/rust/src/http2/detect.rs
@@ -539,6 +539,57 @@ pub unsafe extern "C" fn rs_http2_tx_get_host(
     return 0;
 }
 
+fn http2_lower(value: &[u8]) -> Option<Vec<u8>> {
+    for i in 0..value.len() {
+        if value[i].is_ascii_uppercase() {
+            // we got at least one upper character, need to transform
+            let mut vec: Vec<u8> = Vec::with_capacity(value.len());
+            vec.extend_from_slice(value);
+            for j in i..vec.len() {
+                vec[j].make_ascii_lowercase();
+            }
+            return Some(vec);
+        }
+    }
+    return None;
+}
+
+fn http2_normalize_host(value: &[u8]) -> (Option<Vec<u8>>, usize) {
+    match value.iter().position(|&x| x == ':' as u8) {
+        Some(i) => {
+            return (http2_lower(&value[..i]), i);
+        }
+        None => {
+            return (http2_lower(value), value.len());
+        }
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_http2_tx_get_host_norm(
+    tx: &mut HTTP2Transaction, buffer: *mut *const u8, buffer_len: *mut u32,
+) -> u8 {
+    if let Ok(value) = http2_frames_get_header_value(&tx.frames_ts, ":authority") {
+        let r = http2_normalize_host(value);
+        match r.0 {
+            Some(normval) => {
+                tx.escaped.push(normval);
+                let idx = tx.escaped.len() - 1;
+                let resvalue = &tx.escaped[idx];
+                *buffer = resvalue.as_ptr(); //unsafe
+                *buffer_len = r.1 as u32;
+                return 1;
+            }
+            None => {
+                *buffer = value.as_ptr(); //unsafe
+                *buffer_len = r.1 as u32;
+                return 1;
+            }
+        }
+    }
+    return 0;
+}
+
 #[no_mangle]
 pub unsafe extern "C" fn rs_http2_tx_get_useragent(
     tx: &mut HTTP2Transaction, buffer: *mut *const u8, buffer_len: *mut u32,
@@ -865,5 +916,43 @@ pub extern "C" fn rs_http2_tx_add_header(
         http2_tx_set_header(state, ":authority".as_bytes(), slice_value)
     } else {
         http2_tx_set_header(state, slice_name, slice_value)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    fn test_http2_normalize_host() {
+        let buf0 = "aBC.com:1234".as_bytes();
+        let r0 = http2_normalize_host(buf0);
+        match r0.0 {
+            Some(r) => {
+                assert_eq!(r, "abc.com".as_bytes().to_vec());
+            }
+            None => {
+                panic!("Result should not have been None");
+            }
+        }
+        let buf1 = "oisf.net".as_bytes();
+        let r1 = http2_normalize_host(buf1);
+        match r1.0 {
+            Some(r) => {
+                panic!("Result should not have been None, not {:?}", r);
+            }
+            None => {}
+        }
+        assert_eq!(r1.1, "oisf.net".len());
+        let buf2 = "localhost:3000".as_bytes();
+        let r2 = http2_normalize_host(buf2);
+        match r2.0 {
+            Some(r) => {
+                panic!("Result should not have been None, not {:?}", r);
+            }
+            None => {}
+        }
+        assert_eq!(r2.1, "localhost".len());
     }
 }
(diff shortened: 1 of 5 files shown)


catenacyber:
What remains to be asked to signature writers :
- http.request_body and http.response_body, covered by file_data. Should we have these specifically ?
- http.request_line and http.response_line do not exist in HTTP2, should we emulate them ? What about http.start ?
- http.protocol and http.stat_msg are implicit, should we emulate them ?

rust/src/http2/detect.rs
```diff
@@ -861,6 +861,8 @@ pub extern "C" fn rs_http2_tx_add_header(
     let slice_value = build_slice!(value, value_len as usize);
     if slice_name == "HTTP2-Settings".as_bytes() {
         http2_tx_set_settings(state, slice_value)
+    } else if slice_name == "Host".as_bytes() {
```
victorjulien, line 1077: I'm not sure if a header key like "Host" is case sensitive, so would a case-insensitive check make sense here?

rust/src/http2/detect.rs
```diff
@@ -539,6 +539,57 @@ pub unsafe extern "C" fn rs_http2_tx_get_host(
...
+) -> u8 {
+    if let Ok(value) = http2_frames_get_header_value(&tx.frames_ts, ":authority") {
+        let r = http2_normalize_host(value);
+        match r.0 {
+            Some(normval) => {
+                tx.escaped.push(normval);
+                let idx = tx.escaped.len() - 1;
+                let resvalue = &tx.escaped[idx];
+                *buffer = resvalue.as_ptr(); //unsafe
+                *buffer_len = r.1 as u32;
```
victorjulien, line 613: can you add an explanation of why we use `r.1` as a size var here? I don't find it to be self explaining

catenacyber: indeed, `http2_normalize_host` returns a tuple with r.1 being the size
This is useful when we only take a substring of value (before the port)

In this case, this is equal to `normval.len()`
Do you prefer to use `normval.len()` ?

Adding a comment for now

rust/src/http2/detect.rs
catenacyber: Looks relevant to me, adding it

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6169

rust/src/http2/detect.rs
```diff
@@ -676,6 +676,44 @@ fn http2_escape_header(hd: &parser::HTTP2FrameHeaders, i: u32) -> Vec<u8> {
     return vec;
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn rs_http2_tx_get_header_names(
+    tx: &mut HTTP2Transaction, direction: u8, buffer: *mut *const u8, buffer_len: *mut u32,
+) -> u8 {
+    let mut vec = Vec::new();
+    vec.push('\r' as u8);
```
victorjulien, line 712: can we do this in a single step? Both '\r's but perhaps even combine it with the construction?

rust/src/http2/detect.rs
```diff
@@ -676,6 +676,44 @@ fn http2_escape_header(hd: &parser::HTTP2FrameHeaders, i: u32) -> Vec<u8> {
...
+                    vec.extend_from_slice(&hd.blocks[j].name);
+                    vec.push('\r' as u8);
+                    vec.push('\n' as u8);
+                }
+            }
+            _ => {}
+        }
+    }
+    if vec.len() > 2 {
+        vec.push('\r' as u8);
```
victorjulien, line 730: one push for both chars?

src/detect-http-raw-header.c
```diff
@@ -105,6 +112,16 @@ void DetectHttpRawHeaderRegister(void)
             PrefilterMpmHttpHeaderRawResponseRegister, NULL, ALPROTO_HTTP1,
             0); /* progress handled in register */
 
+    DetectAppLayerInspectEngineRegister2("http_raw_header", ALPROTO_HTTP2, SIG_FLAG_TOSERVER,
+            HTTP2StateDataClient, DetectEngineInspectBufferGeneric, GetData2);
+    DetectAppLayerInspectEngineRegister2("http_raw_header", ALPROTO_HTTP2, SIG_FLAG_TOCLIENT,
+            HTTP2StateDataServer, DetectEngineInspectBufferGeneric, GetData2);
+
+    DetectAppLayerMpmRegister2("http_raw_header", SIG_FLAG_TOSERVER, 2,
+            PrefilterMpmHttp2HeaderRawRequestRegister, NULL, ALPROTO_HTTP2, HTTP2StateDataClient);
```
victorjulien, line 121: I don't see the reason for not using `PrefilterGenericMpmRegister` here instead? The prefilter func below looks standard

src/detect-http-header.c
```diff
@@ -212,6 +234,66 @@ static int DetectEngineInspectBufferHttpHeader(
     return DETECT_ENGINE_INSPECT_SIG_NO_MATCH;
 }
 
+/** \internal
+ *  \brief custom inspect function to utilize the cached headers
+ */
+static int DetectEngineInspectBufferHttp2Header(DetectEngineCtx *de_ctx,
```
victorjulien, line 240: I think this should work the same as `DetectEngineInspectBufferGeneric`.

src/detect-http-header.c
```diff
@@ -263,6 +345,48 @@ static void PrefilterMpmHttpHeader(DetectEngineThreadCtx *det_ctx,
...
 
+/** \brief Generic Mpm prefilter callback
+ *
+ *  \param det_ctx detection engine thread ctx
+ *  \param p packet to inspect
+ *  \param f flow to inspect
+ *  \param txv tx to inspect
+ *  \param pectx inspection context
+ */
+static void PrefilterMpmHttp2Header(DetectEngineThreadCtx *det_ctx, const void *pectx, Packet *p,
```
victorjulien, line 356: use `PrefilterMpm` instead?

src/detect-http-header.c
```diff
@@ -326,6 +450,27 @@ static int PrefilterMpmHttpHeaderRequestRegister(DetectEngineCtx *de_ctx,
     return r;
 }
 
+static int PrefilterMpmHttp2HeaderRequestRegister(DetectEngineCtx *de_ctx, SigGroupHead *sgh,
```
victorjulien, line 453: same as `PrefilterGenericMpmRegister`?

rust/src/http2/detect.rs
catenacyber: I understand you want `vec.extend([b'\r', b'\n'.iter().copied());` looking at the beginning of https://doc.rust-lang.org/std/vec/struct.Vec.html

Going for this

rust/src/http2/detect.rs
catenacyber: ok same as above

rust/src/http2/detect.rs
catenacyber: Changing it in other places as well

src/detect-http-raw-header.c
catenacyber: Great :-)

src/detect-http-header.c
catenacyber: ok trying this

src/detect-http-header.c
catenacyber: same as above

src/detect-http-header.c
catenacyber: And still ok

rust/src/http2/detect.rs
jasonish: Clippy these days suggest to use `vec![]` instead of `new` followed by a `push`.

PR #6152, opened 2021-05-31
Http2 keywords 60x v6

catenacyber:

Describe changes:
- Allows some HTTP1 keywords one HTTP2 traffic

A configuration option allows to have all HTTP1 rules working on HTTP2 traffic without specifying it for each rule

Modifies https://github.com/OISF/suricata/pull/6144 with 
- rebase
- using `git cherry-pick -x` for the cherry-picked commits to get `cherry-picked from` in the commit. Some commits are specific to version 6 and are not cherry-picked
- using `app-layer.protocols.http2.http1-rules` in configuration instead of `app-layer.protocols.http.rules-on-http2`

With Signature alproto and Flow alproto, the options were 
- have only one alproto for HTTP1 and HTTP2, but this needed a major rework (as I see it)
- Duplicate signatures so that a HTTP1 signature has also a HTTP2 version, as is done with bidirectional signatures, but that linked list code has been a trap for bugs, so I avoided touching it
- Have a signature alproto match multiple Flow alproto : That is the solution I chose as in master. The difference with master is that we do not have HTTP_ANY, just a config option to tell that all HTTP1 rules should try to match on HTTP2 traffic

suricata-verify-pr: 506

Makes S-V test in OISF/suricata-verify#506 pass

The S-V tests `http2-keywords` and `http2-keywords2` still require version 7 because they have rules with `alert http1` which is not recognized here.
What could we do about it ?
- have a signature keyword to load the rule only if the suricata version is good for it ?
- allow to load some rules even if some other rules error ?


catenacyber:
CIFuzz wants https://redmine.openinfosecfoundation.org/issues/4417 from #6153

victorjulien:
Merged in #6212, thanks!

PR #6160, opened 2021-06-02

catenacyber:

Describe changes:
- Use function `StreamDataAvailableForProtoDetect` instead of `STREAM_RIGHT_EDGE` to bail out of protocol detection

Replaces #6072 (and #6069) by iterating over the whole red and black tree (which should not be that fulll ?) to get the total length of received tcp segments (and not only the first one)

suricata-verify-pr: 488
cf OISF/suricata-verify#488

### DIFF
--- a/src/stream-tcp-reassemble.c
+++ b/src/stream-tcp-reassemble.c
@@ -587,6 +587,31 @@ static uint32_t StreamTcpReassembleCheckDepth(TcpSession *ssn, TcpStream *stream
     SCReturnUInt(0);
 }
 
+uint32_t StreamDataAvailableForProtoDetect(TcpStream *stream)
+{
+    if (RB_EMPTY(&stream->sb.sbb_tree)) {
+        if (stream->sb.stream_offset != 0)
+            return 0;
+
+        return stream->sb.buf_offset;
+    } else {
+        DEBUG_VALIDATE_BUG_ON(stream->sb.head == NULL);
+
+        if (stream->sb.head->offset != 0)
+            return 0;
+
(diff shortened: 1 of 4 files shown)


src/stream-tcp-reassemble.c
```diff
@@ -587,6 +587,30 @@ static uint32_t StreamTcpReassembleCheckDepth(TcpSession *ssn, TcpStream *stream
...
+        DEBUG_VALIDATE_BUG_ON(stream->sb.head == NULL);
+
+        if (stream->sb.head->offset != 0)
+            return 0;
+
+        // sums all lengths of different TCP segments
+        // so as to bail out in case of missing data
+        uint32_t r = 0;
+        StreamingBufferBlock *sbb = NULL;
+        RB_FOREACH(sbb, SBB, &stream->sb.sbb_tree) {
```
victorjulien, line 607: So generally we want to avoid walking the whole tree as this can lead to big perf issues if we get many segments. This is why it switched from list to tree some years ago. So I wonder if we can get something like this w/o walking the whole thing. I guess one thing is that we could track the data size in the `StreamingBuffer` on insert/removal/update so we'll never do this walk. But that will be more involved and somewhat risky as a change.

catenacyber: Indeed, I was wondering about performance.
I have another idea coming in another PR.

> So I wonder if we can get something like this w/o walking the whole thing

We have Flow `todstbytecnt`(which counts not only app-layer)

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6164

PR #6161, opened 2021-06-02
Draft: Plugins, refactors, public API

jasonish:
Builds on Jeff's PR https://github.com/OISF/suricata/pull/6133 to use the plugin API for registering eve filetypes internally. First we  that idea and just make it feel less like a plugin, as its internal. Then refactor a bit to expose that new internal API for registering filetypes as a plugin. Probably how it should have been done in the first place. Also includes some minor fixups to that PR.

Also includes some include file cleanup with an attempt to seprate some public/private API.  Headers that are designed to be public will be installed in `/usr/include/suricata`, while all other headers will be installed in `/usr/include/suricata/private`. The idea here is not to prevent access to private API, but just make it clear to users, and to us, which APIs we've exposed as public and that will need extra documentation when we break.

For example, the eve filetype template plugin now looks like:
```
/* Required by suricata-plugin.h at this time. Will be fixed. */
#include <suricata/private/queue.h>

#include <suricata/suricata-plugin.h>
#include <suricata/output-json-filetypes.h>
#include <suricata/private/util-debug.h>
```

Here the plugin interface and registering new JSON/EVE filetypes is public interface. Obvious SCLog* make good candidates for public interfaces.  Queue.h might, but its just waiting on a refactor of the capture plugin interface to remove it as a requirement from being included.

This header idea was taken from some documentation I saw on Qt projects that do similar, even though many C projects just install all headers.  But I think the separation is good as it makes it clear what we need to be careful of changing. I don't plan on this settings APIs in stone, but more a queue that a changelog entry is required on change. Also its bad to expose HAVE_CONFIG_H in public headers, as a user of our library may have their own and this can lead to undesirable side affects.

I have a companion plugin to this work which is a Redis output plugin written in Rust: https://github.com/jasonish/eve-redis-plugin/blob/main/src/ffi.rs

### DIFF
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -91,6 +91,10 @@ void OutputJsonRegister (void)
 
     traffic_id_prefix_len = strlen(TRAFFIC_ID_PREFIX);
     traffic_label_prefix_len = strlen(TRAFFIC_LABEL_PREFIX);
+
+    // Register output file types that use the new eve filetype registration
+    // API.
+    SyslogInitialize();
 }
 
 json_t *SCJsonString(const char *val)
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -1133,7 +1133,7 @@ OutputInitResult OutputJsonInitCtx(ConfNode *conf)
         enum LogFileType log_filetype = FileTypeFromConf(output_s);
         if (log_filetype == LOGFILE_TYPE_NOTSET) {
 #ifdef HAVE_PLUGINS
-            SCPluginFileType *plugin = SCPluginFindFileType(output_s);
+            SCEveFileType *plugin = SCPluginFindFileType(output_s);
             if (plugin != NULL) {
                 log_filetype = LOGFILE_TYPE_PLUGIN;
                 json_ctx->plugin = plugin;
--- a/src/output-json.c
+++ b/src/output-json.c
@@ -48,6 +48,7 @@
 
 #include "output.h"
 #include "output-json.h"
+#include "output-json-filetypes.h"
 
 #include "util-byte.h"
 #include "util-privs.h"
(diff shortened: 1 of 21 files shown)


jasonish:
> #include <suricata/output-json-filetypes.h>

Probably should rename this file `output-eve-filetypes`.

src/output-json.c
```diff
@@ -1002,6 +1004,76 @@ int OutputJsonBuilderBuffer(JsonBuilder *js, OutputJsonThreadCtx *ctx)
...
+        }
+    }
+#endif
+    else if (log_filetype == LOGFILE_TYPE_PLUGIN) {
+        if (json_ctx->file_ctx->threaded) {
+            /* Prepare for threaded log output. */
+            if (!SCLogOpenThreadedFile(NULL, NULL, json_ctx->file_ctx, 1)) {
+                return -1;
+            }
+        }
```
jasonish, line 1065: @jlucovsky Can you please check on this.

jlucovsky: This seems correct.

jasonish: Thanks. Squashed into prior commit.

suricata-qa:
|--|--|--|--|
| | generic_stats_chk |
| .tcp.reassembly_memuse | 2293760 | 1966080 | 86.0% |

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |
| .flow.spare | 6919992 | 5890991 | 85.0% |
| .flow.memuse | 1273244608 | 1130221888 | 89.0% |

suricata-qa:
WARNING:
|--|--|--|--|
| | generic_stats_chk |
| .tcp.reassembly_memuse | 2293760 | 1966080 | 86.0% |

jasonish:
Continued at https://github.com/OISF/suricata/pull/6357.

PR #6164, opened 2021-06-03
Tcp protodetect bail 4171 v5

catenacyber:

Describe changes:
- Use function `StreamDataAvailableForProtoDetect` and flow bytes count instead of `STREAM_RIGHT_EDGE` to bail out of protocol detection

Replaces #6160 by not iterating over the whole red and black tree for performance reasons, and using flow bytes count to get a sense data length beyond gaps

suricata-verify-pr: 488
cf OISF/suricata-verify#488

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -205,11 +205,9 @@ static void TCPProtoDetectCheckBailConditions(ThreadVars *tv,
         return;
     }
 
-    const uint64_t size_ts = STREAM_HAS_SEEN_DATA(&ssn->client) ?
-        STREAM_RIGHT_EDGE(&ssn->client) : 0;
-    const uint64_t size_tc = STREAM_HAS_SEEN_DATA(&ssn->server) ?
-        STREAM_RIGHT_EDGE(&ssn->server) : 0;
-    SCLogDebug("size_ts %"PRIu64", size_tc %"PRIu64, size_ts, size_tc);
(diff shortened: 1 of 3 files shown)


src/app-layer.c
```diff
@@ -205,63 +205,61 @@ static void TCPProtoDetectCheckBailConditions(ThreadVars *tv,
...
-    SCLogDebug("size_ts %"PRIu64", size_tc %"PRIu64, size_ts, size_tc);
-
-    DEBUG_VALIDATE_BUG_ON(size_ts > 1000000UL);
-    DEBUG_VALIDATE_BUG_ON(size_tc > 1000000UL);
+    const uint32_t size_ts = StreamDataAvailableForProtoDetect(&ssn->client);
+    const uint32_t size_tc = StreamDataAvailableForProtoDetect(&ssn->server);
+    // We also use the byte count from the flow so that a dropped packet
+    // or a gap finally bails out even if we do not have enough data
+    // in the beginning
+    const uint32_t flow_size_ts = (uint32_t)f->todstbytecnt;
```
victorjulien, line 213: I fear this isn't going to work. The flow byte counts do not take things like overlaps into account, they just count the bytes of any packet that belongs to the flow. So someone could resent the same segment in a loop until the limits is reached causing us to abort the protocol detection.

catenacyber: Right

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6167

PR #6167, opened 2021-06-04

catenacyber:

Describe changes:
- Adds a field `data_protodetect_seen` to structure `TcpStream` to count data for protocol detection bailout
- Places this field in an union so as not to consume more memory
- Uses flag `STREAMTCP_STREAM_FLAG_APPPROTO_DETECTION_COMPLETED` to know which field to use in the union (ie do not use `data_required` as long as we did not complete protocol detection)
- Updates this field `data_protodetect_seen` in `StreamTcpReassembleInsertSegment` (so long as we did not finish protocol detection)
- Resets this union on protocol detection success (does not seem needed in case of protocol detection failure)

Replaces #6164 by not using flow bytes count but a field in the tcp stream, and the difficult trick is to place it in an union so as not to consume more memory

suricata-verify-pr: 488
cf OISF/suricata-verify#488

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -205,11 +205,9 @@ static void TCPProtoDetectCheckBailConditions(ThreadVars *tv,
         return;
     }
 
-    const uint64_t size_ts = STREAM_HAS_SEEN_DATA(&ssn->client) ?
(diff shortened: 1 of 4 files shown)


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6189

PR #6166, opened 2021-06-04
rust: fix app-layer parser flags

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4508

Describe changes:
- fix rust app-layer parser flags so that they have the same values as in C

This especially allows for SSH bypass to work

suricata-verify-pr: 507
https://github.com/OISF/suricata-verify/pull/507

@jasonish could cbindgen export the constants 
cf `item_types = ["enums","structs","opaque","functions"]` in rust/cbindgen.toml 
This way, we will not meet this kind of problem again...

PR #6168, opened 2021-06-04
rust: fix app-layer parser flags

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4508

Describe changes:
- fix rust app-layer parser flags so that they have the same values as in C

This especially allows for SSH bypass to work

suricata-verify-pr: 507
https://github.com/OISF/suricata-verify/pull/507

@jasonish could cbindgen export the constants 
cf `item_types = ["enums","structs","opaque","functions"]` in rust/cbindgen.toml 
This way, we will not meet this kind of problem again...

Replaces #6166


catenacyber:
Merged in #6172

jasonish:
> @jasonish could cbindgen export the constants
> cf `item_types = ["enums","structs","opaque","functions"]` in rust/cbindgen.toml
> This way, we will not meet this kind of problem again...

Might be a good idea to explore.
1) cbindgen doesn't like our BIT_ macros, maybe there is a work-around, maybe not.  In Rust bit flags are often written out like 0b00000001.
2) This leads us more to a big super include.  Should we care?  I noticed I can run cbindgen aginst particular .rs files to make more granular header files which might be something to explore as well.

catenacyber:
> Might be a good idea to explore.
> 
> 1. cbindgen doesn't like our BIT_ macros, maybe there is a work-around, maybe not.  In Rust bit flags are often written out like 0b00000001.
> 2. This leads us more to a big super include.  Should we care?  I noticed I can run cbindgen aginst particular .rs files to make more granular header files which might be something to explore as well.

Should we then create a redmine ticket for this ? (is there already one ?)

jasonish:
> > Might be a good idea to explore.
> > 
> > 1. cbindgen doesn't like our BIT_ macros, maybe there is a work-around, maybe not.  In Rust bit flags are often written out like 0b00000001.
> > 2. This leads us more to a big super include.  Should we care?  I noticed I can run cbindgen aginst particular .rs files to make more granular header files which might be something to explore as well.
> 
> Should we then create a redmine ticket for this ? (is there already one ?)

I don't think there is one covering this idea.  Not a bad idea tho.  Timing wise I think it might be something good to look at after 7.0 drops.

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4517

PR #6175, opened 2021-06-04
Smtp rfc2231 4386 v1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4386

Describe changes:
- Move `FindMimeHeaderTokenRestrict` from C to rust
- mime :  handles multiple sections for a parameter as per RFC2231

Also fixes the evasions where the token name is present in a value as described in rust unit tests
- `attachment; dummy="filename=wrong"; filename=real`
- `attachment; notafilename=wrong; filename=good;`

suricata-verify-pr: 509
https://github.com/OISF/suricata-verify/pull/509

PR #6177, opened 2021-06-04
Smtp rfc2231 4386 v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4386

Describe changes:
- Move `FindMimeHeaderTokenRestrict` from C to rust
- mime :  handles multiple sections for a parameter as per RFC2231

Also fixes the evasions where the token name is present in a value as described in rust unit tests
- `attachment; dummy="filename=wrong"; filename=real`
- `attachment; notafilename=wrong; filename=good;`

suricata-verify-pr: 509
https://github.com/OISF/suricata-verify/pull/509

Modifies #6175 by fixing compile warning on Fedora and the rust failure on old ubuntu

### DIFF
--- a/rust/src/mime/mod.rs
+++ b/rust/src/mime/mod.rs
@@ -1,0 +1,168 @@
+/* Copyright (C) 2021 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+use std;
+
+use nom::combinator::rest;
+use nom::error::ErrorKind;
+use nom::Err;
+use nom::IResult;
+
+#[derive(Clone, Debug)]
+pub struct MIMEHeaderToken<'a> {
+    pub name: &'a [u8],
+    pub value: &'a [u8],
+}
+
+#[derive(Clone)]
+pub struct MIMEHeaderTokens<'a> {
+    pub tokens: Vec<MIMEHeaderToken<'a>>,
+}
+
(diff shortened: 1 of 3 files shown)


inashivb:
Mime bindings created by cbindgen (done by Jason here: https://github.com/jasonish/suricata/tree/rust-mime-bindings/v1 ) can also go cleanly with this work, it seems.
@jasonish would it be the right thing to do?

rust/src/mime/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+#[derive(Clone)]
+pub struct MIMEHeaderTokens<'a> {
+    pub tokens: Vec<MIMEHeaderToken<'a>>,
+}
+
+pub fn mime_parse_value_delimited(input: &[u8]) -> IResult<&[u8], &[u8]> {
+    let (i2, _) = tag!(input, "\"")?;
+    let (i3, value) = take_until!(i2, "\"")?;
+    let (i4, _) = tag!(i3, "\"")?;
+    return Ok((i4, value));
```
jasonish, line 40: Its OK, and idiomatic to reuse the same variable name if you don't need the previous value, ie)
```
let (input, _) = tag!(input, "\"")?;
let (input, value) = take_until(...
```

rust/src/mime/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    return Err(());
+}
+
+// TODO ? export with "constants" in cbindgen
+// and use in outbuf definition for rs_mime_find_header_token
+// but other constants are now defined twice in rust and in C
+const RS_MIME_MAX_TOKEN_LEN: usize = 255;
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_mime_find_header_token(
```
jasonish, line 155: The `rs_` prefix served its need when we were new to Rust and Rust was optional, but I don't really like it anymore.  I wonder if our `extern "C"` functions should follow our C conventions now as that is the user of the code. For example, this would become `MimeFindHeaderToken`.  Thoughts? @victorjulien

rust/src/mime/mod.rs
catenacyber: Ok, but then I need input to be `mut` right ?

jasonish: > Ok, but then I need input to be `mut` right ?

No, shouldn't be.  Redefining is different than mutating.  Rust allowed this for patterns like:
```
let x = Something{}...
// Do some stuff with X..
let x = Arc::new(Mutex::new(x));
// Pass x of to threads or something
```
When dealing with input like above, I think it also fits a pattern where redefining is actually nice.

catenacyber: ok done in next version

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6194

PR #6179, opened 2021-06-04
Issue 4171/v4

victorjulien:
Replaces #6072, fixes #6178

Differs from #6167 in that it tracks the actual stored data, at minimal overhead, in the stream buffer.

PR #6180, opened 2021-06-04
Issue 4171/v5

victorjulien:
Replaces #6072, fixes #6179

Differs from #6167 in that it tracks the actual stored data, at minimal overhead, in the stream buffer.

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2020 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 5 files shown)


src/app-layer.c
```diff
@@ -205,11 +205,9 @@ static void TCPProtoDetectCheckBailConditions(ThreadVars *tv,
         return;
     }
 
-    const uint64_t size_ts = STREAM_HAS_SEEN_DATA(&ssn->client) ?
-        STREAM_RIGHT_EDGE(&ssn->client) : 0;
-    const uint64_t size_tc = STREAM_HAS_SEEN_DATA(&ssn->server) ?
-        STREAM_RIGHT_EDGE(&ssn->server) : 0;
-    SCLogDebug("size_ts %"PRIu64", size_tc %"PRIu64, size_ts, size_tc);
+    const uint32_t size_ts = StreamDataAvailableForProtoDetect(&ssn->client);
```
victorjulien, line 208: I wonder if we still should take the diff between last_ack and isn into account here as well. Otherwise we might get all in-window data with some missing at the start. If the window is big enough this might trigger the bail conditions.

catenacyber: Maybe I do not understand what you mean.
But I do not think we should take the diff between last_ack and isn into account here.
It can suddenly get really big in one packet if the TCP window is big enough (that is what oss-fuzz found).
That seems fragile, if the attacker can inject one TCP packet to evade protocol detection...
Anyways in this case, we should not use `DEBUG_VALIDATE_BUG_ON` on its value

victorjulien: The scenario that I worry about is:

1. window size is bigger than the 100k we use in this function to make certain decisions
2. attacker makes sure first bytes in the stream are missing, then fills the window to the point that we reach the 100k
3. suricata gives up on protocol detect
4. attacker sends missing bytes
5. evasion

Maybe we should take the window into account, although this can be very big.

catenacyber: Indeed you are right : tcp window should be used instead of hard-coded constants (but keep a minimum, and configure a maximum)

Should I add this on top of this PR ?

victorjulien: Yeah please give it a try. I wonder how to deal with the potential of huge windows, as with huge windows we might use too many resources.

catenacyber: Suggestion in https://github.com/OISF/suricata/pull/6189

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6189

PR #6183, opened 2021-06-07
Http2 keywords norm 4067 v4

catenacyber:

Describe changes:
- http2: turn `Host header into `authority` during upgrade
- http2: `http.host` normalized keyword now works for HTTP2
- http2: `http.header_names` normalized keyword now works for HTTP2
- http2: `http.header` keyword now works for HTTP2 as well as `http.header.raw`
- http2: concatenate one headers multiple values
- http2: generic http2_header_blocks (bonus as some cases were forgotten for continuation and push_promise)

suricata-verify-pr: 504
https://github.com/OISF/suricata-verify/pull/504

We do not escape linefeeds or `:` in HTTP2 header names even if they can syntactically be present. Is it ok ?

Modifies #6176 by taking the review into account, ie using `vec!`

### DIFF
--- a/rust/src/http2/detect.rs
+++ b/rust/src/http2/detect.rs
@@ -539,6 +539,60 @@ pub unsafe extern "C" fn rs_http2_tx_get_host(
     return 0;
 }
 
+fn http2_lower(value: &[u8]) -> Option<Vec<u8>> {
+    for i in 0..value.len() {
+        if value[i].is_ascii_uppercase() {
+            // we got at least one upper character, need to transform
+            let mut vec: Vec<u8> = Vec::with_capacity(value.len());
+            vec.extend_from_slice(value);
+            for j in i..vec.len() {
+                vec[j].make_ascii_lowercase();
+            }
+            return Some(vec);
+        }
+    }
+    return None;
+}
+
+// returns a tuple with the value and its size
+fn http2_normalize_host(value: &[u8]) -> (Option<Vec<u8>>, usize) {
+    match value.iter().position(|&x| x == ':' as u8) {
+        Some(i) => {
+            return (http2_lower(&value[..i]), i);
+        }
+        None => {
+            return (http2_lower(value), value.len());
+        }
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_http2_tx_get_host_norm(
+    tx: &mut HTTP2Transaction, buffer: *mut *const u8, buffer_len: *mut u32,
+) -> u8 {
+    if let Ok(value) = http2_frames_get_header_value(&tx.frames_ts, ":authority") {
+        let r = http2_normalize_host(value);
+        // r is a tuple with the value and its size
+        // this is useful when we only take a substring (before the port)
+        match r.0 {
+            Some(normval) => {
+                tx.escaped.push(normval);
+                let idx = tx.escaped.len() - 1;
+                let resvalue = &tx.escaped[idx];
+                *buffer = resvalue.as_ptr(); //unsafe
+                *buffer_len = r.1 as u32;
+                return 1;
+            }
+            None => {
+                *buffer = value.as_ptr(); //unsafe
+                *buffer_len = r.1 as u32;
+                return 1;
+            }
+        }
+    }
+    return 0;
+}
+
 #[no_mangle]
 pub unsafe extern "C" fn rs_http2_tx_get_useragent(
     tx: &mut HTTP2Transaction, buffer: *mut *const u8, buffer_len: *mut u32,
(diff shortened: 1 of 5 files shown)


suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:
Information:
|--|--|--|--|
| | tlpr1_stats_chk |

rust/src/http2/detect.rs
```diff
@@ -604,68 +684,193 @@ pub unsafe extern "C" fn rs_http2_tx_get_header_value(
...
+    };
+    for i in 0..frames.len() {
+        if let Some(blocks) = http2_header_blocks(&frames[i]) {
+            for j in 0..blocks.len() {
+                // we do not escape linefeeds nor : in headers names
+                vec.extend_from_slice(&blocks[j].name);
+                vec.extend([b':', b' '].iter().copied());
+                vec.extend_from_slice(&blocks[j].value);
+                vec.extend([b'\r', b'\n'].iter().copied());
+            }
```
jasonish, line 825: This is a bit of a nit, but I believe its more idiomatic to use `.iter()` here as its generally faster than indexing into an array which occurs runtime bound checks.  Instead this is recommended:
```
for block in blocks.iter() {
    vec.extend_from_side(&block.name);
```
the clippy warning is pretty terse tho:
```
Just iterating the collection itself makes the intent more clear and is probably faster.
```

rust/src/http2/detect.rs
```diff
@@ -531,19 +562,73 @@ pub unsafe extern "C" fn rs_http2_tx_get_method(
...
+        let r = http2_normalize_host(value);
+        // r is a tuple with the value and its size
+        // this is useful when we only take a substring (before the port)
+        match r.0 {
+            Some(normval) => {
+                tx.escaped.push(normval);
+                let idx = tx.escaped.len() - 1;
+                let resvalue = &tx.escaped[idx];
+                *buffer = resvalue.as_ptr(); //unsafe
+                *buffer_len = r.1 as u32;
```
jasonish, line 614: Its a little odd to see a mutation like this in a `getter`. Can you add a comment as to why?

jasonish: Ok, I see why.  Still a little comment about having the TX take ownership maybe?

rust/src/http2/detect.rs
```diff
@@ -604,68 +684,193 @@ pub unsafe extern "C" fn rs_http2_tx_get_header_value(
...
+        &tx.frames_ts
+    } else {
+        &tx.frames_tc
+    };
+    for i in 0..frames.len() {
+        if let Some(blocks) = http2_header_blocks(&frames[i]) {
+            for j in 0..blocks.len() {
+                // we do not escape linefeeds in headers names
+                vec.extend_from_slice(&blocks[j].name);
+                vec.extend([b'\r', b'\n'].iter().copied());
```
jasonish, line 722: Why not `vec.extend_from_slice(&[b'\r', b'\n']);`?

jasonish:
Can some or all of f7291ad207ef3706f3052d438058599473ad98eb be squashed into the commits it fixes?

jasonish:
> ERROR: ASAN TRIP in tlpr1_asan_suri QA test

Looks like a rebase clears this up.

rust/src/http2/detect.rs
```diff
@@ -604,68 +684,193 @@ pub unsafe extern "C" fn rs_http2_tx_get_header_value(
...
+        }
+    }
+    while end > 0 {
+        if value[end - 1] == b' ' || value[end - 1] == b'\t' {
+            end -= 1;
+        } else {
+            break;
+        }
+    }
+    return &value[start..end];
```
jasonish, line 770: This appears to be panic when called with a buffer that is only spaces.  Not sure if we can get here with such a buffer, but should still probably handle that case?

rust/src/http2/detect.rs
```diff
@@ -604,68 +684,193 @@ pub unsafe extern "C" fn rs_http2_tx_get_header_value(
     return 0;
 }
 
-fn http2_escape_header(hd: &parser::HTTP2FrameHeaders, i: u32) -> Vec<u8> {
+fn http2_escape_header(blocks: &Vec<parser::HTTP2FrameHeaderBlock>, i: u32) -> Vec<u8> {
```
jasonish, line 687: This should probably be `blocks: &[parser::HTTP2FrameHeaderBlock]`.

jasonish (changes requested):
Looking good. Some mostly minor things and a rebase necessary, but I think you've already started on that.

rust/src/http2/detect.rs
catenacyber: both works, changing for your version :-)

rust/src/http2/detect.rs
catenacyber: Right, thanks for spotting this.
Adding a test case

catenacyber:
> Can some or all of f7291ad207ef3706f3052d438058599473ad98eb be squashed into the commits it fixes?

I do not see this commit f7291ad207ef3706f3052d438058599473ad98eb

rust/src/http2/detect.rs
catenacyber: Looks better indeed

rust/src/http2/detect.rs
catenacyber: Indeed, The tx needs to take ownership so that the buffer outlives the rust function and can be used in its C caller

Adding a comment

rust/src/http2/detect.rs
catenacyber: ok thanks

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6270

catenacyber:
Thanks a lot for this insightful review Jason 👍

jasonish:
> > Can some or all of f7291ad207ef3706f3052d438058599473ad98eb be squashed into the commits it fixes?
> 
> I do not see this commit f7291ad207ef3706f3052d438058599473ad98eb

Oops, I was looking at a local rebase, I meant this commit: https://github.com/OISF/suricata/pull/6183/commits/66228f90e25607c8aeebda7a73b92a720dffa63e

catenacyber:
Ok, so now, I put a new rust-style commit in the end.
It changes more spots than this PR, like `http2_detect_sizeupdate_match`

PR #6182, opened 2021-06-05
Protodetect enip dns 4388 v6

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4388

Describe changes:
- Improves probing parsers for nfs, enip and dns

so as to avoid protocol detection confusion with TCP splitting

Modifies #6035 with fix in dns `probe_header_validity` value for rlen (length of the total DNS message should allow place for all the additional records)

PR #6185, opened 2021-06-07
Protodetect enip dns 4388 v7

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4388

Describe changes:
- Improves probing parsers for nfs, enip and dns

so as to avoid protocol detection confusion with TCP splitting

Modifies #6182 with 
- more precision for dns : the sum of the number of records cannot exceed the message length
- more precision for enip for `LIST_INTERFACES`
(found by fuzzing the fixing branch)

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -659,23 +659,50 @@ impl DNSState {
     }
 }
 
+fn probe_header_validity(header : DNSHeader, rlen: usize) -> (bool, bool, bool) {
+    let opcode = ((header.flags >> 11) & 0xf) as u8;
+    if opcode >= 7 {
+        //unassigned opcode
+        return (false, false, false);
+    }
+    if header.additional_rr as usize
+        + header.answer_rr as usize
+        + header.authority_rr as usize
+        + header.questions as usize
+        > rlen
+    {
+        //not enough data for additional records
+        return (false, false, false);
+    }
+    let is_request = header.flags & 0x8000 == 0;
+    return (true, is_request, false);
+}
+
 /// Probe input to see if it looks like DNS.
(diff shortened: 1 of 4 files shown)


rust/src/dns/dns.rs
```diff
@@ -659,23 +659,50 @@ impl DNSState {
...
+    }
+    if header.additional_rr as usize
+        + header.answer_rr as usize
+        + header.authority_rr as usize
+        + header.questions as usize
+        > rlen
+    {
+        //not enough data for additional records
+        return (false, false, false);
+    }
```
jasonish, line 676: As these are counts not sizes, its not accurate, but also won't false positive.  Should the size of the header be calculated in as well?  If sizeof(header) + addtional_rr +answer_rr > rlen...  Might be a bit more accurate.  I think we could also assume a minumum of 2 bytes per expected record being the CLASS is a u16.

catenacyber: Indeed, these are counts, not sizes.
And I do not count the header size.
But as each of these has at least one byte, we can do this test to exclude some cases
 
So, should I include the header size in this sum ?

catenacyber: > I think so. Just to confirm, rlen is the size of the record. In TCP this would be the first 2 bytes, in UDP its the payload size.

Exactly, you can check the callers do it right

> So a test to make sure its greater than or equal to:
> 
> ```
> sizeof(header) + (additional_rr * 2) + (answer_rr * 2) + (authority_rr * 2) + (questions * 2)
> ```
> 
> should be safe and accurate, right?

Ok, will do this

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6195

PR #6188, opened 2021-06-09
jsonbuilder: avoids one allocation in encode_string

catenacyber:
None specific, part of https://redmine.openinfosecfoundation.org/issues/4125

Describe changes:
- Improves performance of jsonbuilder `encode_string` by avoiding an allocation

Profiling was done on running fuzz_sigpcap_aware on its public corpus.
On my machine, `encode_string` takes about 17 seconds (I removed the inlining to get this)
With this patch, it takes about 2 seconds

### DIFF
--- a/rust/src/jsonbuilder.rs
+++ b/rust/src/jsonbuilder.rs
@@ -166,7 +166,7 @@ impl JsonBuilder {
             State::None => {
                 debug_validate_fail!("invalid state");
                 Err(JsonError::InvalidState)
-            },
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


rust/src/jsonbuilder.rs
```diff
@@ -166,7 +166,7 @@ impl JsonBuilder {
             State::None => {
                 debug_validate_fail!("invalid state");
                 Err(JsonError::InvalidState)
-            },
```
catenacyber, line 169: rustfmt did this

rust/src/jsonbuilder.rs
```diff
@@ -503,61 +503,34 @@ impl JsonBuilder {
...
-                offset += 1;
-                buf[offset] = b'0';
-                offset += 1;
-                buf[offset] = b'0';
-                offset += 1;
-                buf[offset] = HEX[(x >> 4 & 0xf) as usize];
-                offset += 1;
-                buf[offset] = HEX[(x & 0xf) as usize];
-                offset += 1;
+                let mut buf = vec![b'\\', b'u', b'0', b'0', 0, 0];
```
jasonish, line 517: This could be `let but = &mut &[b'\\', ...]` for maybe even tinier bit more performance..

jasonish:
Nice to see this is faster.. Its also more what this should look like anyways.

rust/src/jsonbuilder.rs
catenacyber: Then I get the compile error :
```
error[E0594]: cannot assign to `buf[_]` which is behind a `&` reference
   --> src/jsonbuilder.rs:518:17
    |
518 |                 buf[4] = HEX[(x >> 4 & 0xf) as usize];
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot assign
```

jasonish:
Just to some up some rather manual bench marking.  This is much faster for small values, but for values over 1000 characters what is in master appears to be faster, especially as more and more bytes need to be encoded.  For some configurations that might be faster such as alerts only, no payload logging, etc.  But as soon as you added payload logging to your alerts master will perform better.

rust/src/jsonbuilder.rs
jasonish: Oops, that should be:
```
let buf = &mut [b'\\', ...];
```

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6196

rust/src/jsonbuilder.rs
catenacyber: Thanks Jason

PR #6189, opened 2021-06-09
Tcp protodetect bail 4171 v7

catenacyber:

Describe changes:
- tracks the actual stored data, at minimal overhead, in the stream buffer.

Replaces #6167 and #6180 by adding on top of #6180 a commit to 
- take TCP window into account
- still have a limit from the configuration
- bail out if we really have too much data as that can actually happen

suricata-verify-pr: 488
cf OISF/suricata-verify#488


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6244

PR #6196, opened 2021-06-11
jsonbuilder: avoids one allocation in encode_string

catenacyber:
None specific, part of https://redmine.openinfosecfoundation.org/issues/4125

Describe changes:
- Improves performance of jsonbuilder `encode_string` by avoiding an allocation

Profiling was done on running fuzz_sigpcap_aware on its public corpus.
On my machine, `encode_string` takes about 17 seconds (I removed the inlining to get this)
With this patch, it takes about 2 seconds

Modifies #6188 with doing `vec!` only once outside of loop

### DIFF
--- a/rust/src/jsonbuilder.rs
+++ b/rust/src/jsonbuilder.rs
@@ -166,7 +166,7 @@ impl JsonBuilder {
             State::None => {
                 debug_validate_fail!("invalid state");
                 Err(JsonError::InvalidState)
-            },
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


jasonish:
In my real crude benchmark test, this PR is faster for Rust code compiled in debug mode, however, once built with `--release` as most production builds wiill do, master is faster.  Up to 4x faster in the worse case condition of every character needing to be escaped.

catenacyber:
Closing for now

Bounds check seem to be the reason for the performance difference...

PR #6203, opened 2021-06-18
src/detect: fix bug where rule w/o sid is accepted

jufajardini:
Before, if Suricata parsed a rule without a 'sid' option, instead of
failing that rule, the rule was parsed and attributed a sid 0.
Changes to:
src/detect-parse:
- add logic to filter out rules without sid;
- change unittest which didn't have a sid, but used to pass.
src/detect-sid: add unittest for rules without sid or with sid: 0

Previous PR: https://github.com/OISF/suricata/pull/6202. New PR to make sure CI uses S-V PR, this time.

Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/4491

suricata-verify-pr:510

PR #6204, opened 2021-06-18
src/detect: fix bug where rule w/o sid is accepted

jufajardini:
Before, if Suricata parsed a rule without a 'sid' option, instead of
failing that rule, the rule was parsed and attributed a sid 0.
Changes to:
src/detect-parse:
- add logic to filter out rules without sid;
- change unittest which didn't have a sid, but used to pass.
src/detect-sid: add unittest for rules without sid or with sid: 0

https://redmine.openinfosecfoundation.org/issues/4491

Previous PR https://github.com/OISF/suricata/pull/6203
(same changes, just trying to make CI read the S-V PR...)

suricata-verify-pr: 510
https://github.com/OISF/suricata-verify/pull/510

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -1243,6 +1243,16 @@ static int SigParse(DetectEngineCtx *de_ctx, Signature *s,
         memset(input, 0x00, buffer_size);
         memcpy(input, parser->opts, strlen(parser->opts)+1);
 
+        /* Make sure that options have a "sid" field at all.
+         * This is necessary because when Suricata doesn't parse "sid",
+         * it won't call the functions to check and validate that
+         * field, and Suri won't consider there's something wrong,
+         * even if sid was missing. */
+        if (strstr(input, "sid:") == NULL) {
+            SCLogError(SC_ERR_INVALID_SIGNATURE, "No `sid` option found.");
+            SCReturnInt(-1);
+        }
+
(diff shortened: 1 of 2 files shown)


src/detect-parse.c
```diff
@@ -1243,6 +1243,16 @@ static int SigParse(DetectEngineCtx *de_ctx, Signature *s,
         memset(input, 0x00, buffer_size);
         memcpy(input, parser->opts, strlen(parser->opts)+1);
 
+        /* Make sure that options have a "sid" field at all.
```
jlucovsky, line 1246: Suggest this check occurs after the `if` stmt on line 1239. Fail early before preparing `input` (zero-fill, then copy).

Also suggest error message be adjusted ... "Signature missing required value \"sid\""

jlucovsky:
Suggest that a s-v test be added to validate the sid requirement.

jufajardini:
> Suggest that a s-v test be added to validate the sid requirement.
Thoughts about this:
Wouldn't all the s-v tests that have rules currently work as a test for the sid requirement? Because if any of those don't have a sid (like the mqtt-binary-message one), that test would fail, from now on.
Also, considering that this is a run-time error, instead of an output one, is an s-v test the best approach?

jufajardini:
Followed by: https://github.com/OISF/suricata/pull/6210

PR #6206, opened 2021-06-19
Unify reg smb/v1

inashivb:


src/app-layer-smb.c
```diff
@@ -230,95 +38,11 @@ static SuricataFileContext sfc = { &sbcfg };
...
-
-    /** SMB */
-    if (AppLayerProtoDetectConfProtoDetectionEnabled("tcp", proto_name)) {
-        AppLayerProtoDetectRegisterProtocol(ALPROTO_SMB, proto_name);
-        if (SMBRegisterPatternsForProtocolDetection() < 0)
-            return;
-
-        rs_smb_init(&sfc);
-
-        if (RunmodeIsUnittests()) {
```
inashivb, line 247: If there are no C tests is this conditional still required? If SMB is used by other protos, is that when we need this?
This has not been translated to Rust

src/app-layer-smb.c
```diff
@@ -230,95 +38,11 @@ static SuricataFileContext sfc = { &sbcfg };
...
-        ConfNode *p = ConfGetNode("app-layer.protocols.smb.stream-depth");
-        if (p != NULL) {
-            uint32_t value;
-            if (ParseSizeStringU32(p->val, &value) < 0) {
-                SCLogError(SC_ERR_SMB_CONFIG, "invalid value for stream-depth %s", p->val);
-            } else {
-                stream_depth = value;
-            }
-        }
-        SCLogConfig("SMB stream depth: %u", stream_depth);
```
inashivb, line 315: Although there is just TCP registration here, to be sure, is that all we need?

rust/src/smb/smb.rs
```diff
@@ -52,6 +52,12 @@ use crate::smb::events::*;
 use crate::smb::files::*;
 use crate::smb::smb2_ioctl::*;
 
+pub const MIN_REC_SIZE: u16 = 32+4; // SMB hdr + nbss hdr
+pub const FLOW_DIR_REVERSED: u32 = BIT_U32!(26);
```
victorjulien, line 57: this should probably move somewhere generic as its a flag from the flow engine

rust/src/smb/smb.rs
```diff
@@ -2007,11 +2033,15 @@ pub extern "C" fn rs_smb_state_get_tx(state: &mut SMBState,
 // for use with the C API call StateGetTxIterator
 #[no_mangle]
 pub extern "C" fn rs_smb_state_get_tx_iterator(
-                                      state: &mut SMBState,
-                                      min_tx_id: u64,
-                                      istate: &mut u64)
-                                      -> applayer::AppLayerGetTxIterTuple
+_ipproto: u8,
```
victorjulien, line 2037: formatting looks a bit off

src/app-layer-smb.c
victorjulien: I don't see a reason to keep it if there are no tests.

src/app-layer-smb.c
victorjulien: Yes we support smb over tcp only.

victorjulien:
Minor issues but looks good otherwise. Passes my QA as well.

inashivb:
Replaced with #6209

PR #6213, opened 2021-06-21
counters: only print alerts if stats are enabled

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4537

### DIFF
--- a/src/counters.c
+++ b/src/counters.c
@@ -851,7 +851,9 @@ static void StatsLogSummary(void)
         }
     }
     SCMutexUnlock(&stats_table_mutex);
-    SCLogInfo("Alerts: %"PRIu64, alerts);
+    if (stats_enabled) {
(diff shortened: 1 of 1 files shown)


src/counters.c
```diff
@@ -851,7 +851,9 @@ static void StatsLogSummary(void)
         }
     }
     SCMutexUnlock(&stats_table_mutex);
-    SCLogInfo("Alerts: %"PRIu64, alerts);
+    if (stats_enabled) {
```
victorjulien, line 854: I think the check should be at the start of the function. Something like
```c
if (!stats_enabled)
   return;
```
Reason is that I think we shouldn't touch the stats internals if stats aren't enabled.

inashivb: OK. Got it.

inashivb:
Replaced with #6221

PR #6162, opened 2021-06-03
Documentation: transaction dev guide v5

jufajardini:
Describe changes:
- Add contents and examples about rule matching
- Fix typos
- Change formatting

Link to Redmine issue: https://redmine.openinfosecfoundation.org/issues/4396

Previous PR: https://github.com/OISF/suricata/pull/6068

PR #6220, opened 2021-06-22
devguide/transaction documentation v6

jufajardini:
Describe changes:

- add SequenceDiagrams for TLS, HTTP2 and DNS transactions, generated by mscgen
- refine grammar and some wordings
- add examples for simpler protocols
- use native markup for `table of contents`

Link to Redmine issue: https://redmine.openinfosecfoundation.org/issues/4396

Previous PR: https://github.com/OISF/suricata/pull/6162

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1024,6 +1024,7 @@ jobs:
       - name: Install packages for generating documentation
         run: |
           DEBIAN_FRONTEND=noninteractive apt -y install \
+                mscgen \
                 sphinx-doc \
                 sphinx-common \
                 texlive-latex-base \
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1031,6 +1031,7 @@ jobs:
                 texlive-fonts-recommended \
                 texlive-fonts-extra \
                 texlive-latex-extra
+      - run:  pip install sphinxcontrib-mscgen
       - name: Install Coccinelle
         run: |
           add-apt-repository -y ppa:npalix/coccinelle
--- a/doc/devguide/Makefile.am
+++ b/doc/devguide/Makefile.am
@@ -7,6 +7,7 @@ EXTRA_DIST = \
 	extending/index.rst \
 	extending/app-layer/index.rst \
 	extending/app-layer/parser.rst \
+	extending/app-layer/transactions.rst \
 	extending/capture/index.rst \
 	extending/output/index.rst \
 	internals/engines/index.rst \
--- a/doc/devguide/conf.py
+++ b/doc/devguide/conf.py
@@ -13,6 +13,7 @@ import sys
 import os
 import shlex
 import re
+import sphinxcontrib
 
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
@@ -29,7 +30,7 @@ on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = []
+extensions = ['sphinxcontrib.mscgen']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/doc/devguide/conf.py
+++ b/doc/devguide/conf.py
@@ -13,7 +13,7 @@ import sys
 import os
 import shlex
 import re
-import sphinxcontrib
+import sphinxcontrib.mscgen
 
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
--- a/doc/devguide/extending/app-layer/index.rst
+++ b/doc/devguide/extending/app-layer/index.rst
@@ -5,3 +5,4 @@ App-Layer
    :maxdepth: 2
 
    parser.rst
+   transactions.rst
--- a/doc/devguide/extending/app-layer/transactions.rst
+++ b/doc/devguide/extending/app-layer/transactions.rst
@@ -66,7 +66,7 @@ _`Progress Tracking`
 
 As a rule of thumb, transactions will follow a request-response model: if a transaction has had a request and a response, it is complete.
 
-But if a protocol has situations where a request or response won’t expect or generate a message from its counter-part,
+But if a protocol has situations where a request or response won’t expect or generate a message from its counterpart,
 it is also possible to have uni-directional transactions. In such cases, transaction is set to complete at the moment of
 creation.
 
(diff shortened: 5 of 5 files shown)


jufajardini:
Regarding mscgen: works locally adding it as a sphinx extension in conf.py But tbh I'm not sure what else can I do to make sure it would work with readthedocs... Searching about it online didn't help, maybe I didn't know how to search...

jufajardini:
Currently trying to figure out how to pass mscgen to the CI. Tried builds.yml, but doesn't seem to work....

jasonish:
There are a few issues to consider when adding a dependency on mscgen.

The Sphinx mscgen module doesn't appear to be ubiquitous in package managers across all the major distributions (I might be wrong, might be looking for the long name). Which means installing from pip, something many people installing Suricata won't want to do, or can't to.

Because of this, we'd have to add ./configure checking for mscgen, and the sphinx module before enabling the documentation to be built as it it will fail to build on systems where it currently does build.

While it does feel archaic, its probably simpler to generate these images and commit the image artifact and pull that into Sphinx, if possible, to avoid add a new built-time dependency.

jufajardini:
> There are a few issues to consider when adding a dependency on mscgen.
> 
> The Sphinx mscgen module doesn't appear to be ubiquitous in package managers across all the major distributions (I might be wrong, might be looking for the long name). Which means installing from pip, something many people installing Suricata won't want to do, or can't to.
> 
> Because of this, we'd have to add ./configure checking for mscgen, and the sphinx module before enabling the documentation to be built as it it will fail to build on systems where it currently does build.
> 
> While it does feel archaic, its probably simpler to generate these images and commit the image artifact and pull that into Sphinx, if possible, to avoid add a new built-time dependency.

I see, it is possible to include images with Sphinx, too: it's just a matter of deciding what we prefer. Victor suggested that a workaround would be having a script to generate the images that would, later on, be pulled into Sphinx. One thing that isn't clear to me with that solution is how would it solve the issue of the mscgen dependency... (or it wouldn't?)

jasonish:
> > There are a few issues to consider when adding a dependency on mscgen.
> > The Sphinx mscgen module doesn't appear to be ubiquitous in package managers across all the major distributions (I might be wrong, might be looking for the long name). Which means installing from pip, something many people installing Suricata won't want to do, or can't to.
> > Because of this, we'd have to add ./configure checking for mscgen, and the sphinx module before enabling the documentation to be built as it it will fail to build on systems where it currently does build.
> > While it does feel archaic, its probably simpler to generate these images and commit the image artifact and pull that into Sphinx, if possible, to avoid add a new built-time dependency.
> 
> I see, it is possible to include images with Sphinx, too: it's just a matter of deciding what we prefer. Victor suggested that a workaround would be having a script to generate the images that would, later on, be pulled into Sphinx. One thing that isn't clear to me with that solution is how would it solve the issue of the mscgen dependency... (or it wouldn't?)

There are 2 ways around it.  Have the scripts, all the source input in git, but pre-generate the images and check them into git. This means mscgen is only required when updating the images.  But it also go against the idea that anything that can be generated during the build should be.

But as `mscgen` does seem more common in package managers than the Sphinx module, maybe we could hook it into the build system to generate the images before the docs.  We'd still want a test in `configure.ac` to look for `mscgen` as a requirement for building the docs tho.

jufajardini:
> > > There are a few issues to consider when adding a dependency on mscgen.
> > > The Sphinx mscgen module doesn't appear to be ubiquitous in package managers across all the major distributions (I might be wrong, might be looking for the long name). Which means installing from pip, something many people installing Suricata won't want to do, or can't to.
> > > Because of this, we'd have to add ./configure checking for mscgen, and the sphinx module before enabling the documentation to be built as it it will fail to build on systems where it currently does build.
> > > While it does feel archaic, its probably simpler to generate these images and commit the image artifact and pull that into Sphinx, if possible, to avoid add a new built-time dependency.
> > 
> > 
> > I see, it is possible to include images with Sphinx, too: it's just a matter of deciding what we prefer. Victor suggested that a workaround would be having a script to generate the images that would, later on, be pulled into Sphinx. One thing that isn't clear to me with that solution is how would it solve the issue of the mscgen dependency... (or it wouldn't?)
> 
> There are 2 ways around it. Have the scripts, all the source input in git, but pre-generate the images and check them into git. This means mscgen is only required when updating the images. But it also go against the idea that anything that can be generated during the build should be.
> 
> But as `mscgen` does seem more common in package managers than the Sphinx module, maybe we could hook it into the build system to generate the images before the docs. We'd still want a test in `configure.ac` to look for `mscgen` as a requirement for building the docs tho.

I think I like the second approach, thanks, I'll try that out...

inashivb:
@jasonish wondering if writing a new job or adding to an existing the following can be a solution:
Getting the source code using curl and doing a `setup.py install` provided we install mscgen with `apt` (which is available and maintained)? I see that we use curl to get rust, and we already install setuptools which is all that seems to be required by `mscgen/setup.py` (https://github.com/sphinx-contrib/mscgen/blob/master/setup.py)

jufajardini:
With the check for mscgen in configure.ac, we could still face the situation where the person tries to build the devguide, nonetheless, right?
I'm wondering what should be done in that case - let the document be generated with broken images sounds bad, but, in my researches, I couldn't find a way to circumvent this from a Sphinx perspective - unless skipping building the devguide altogether...

jasonish:
> With the check for mscgen in configure.ac, we could still face the situation where the person tries to build the devguide, nonetheless, right?
> I'm wondering what should be done in that case - let the document be generated with broken images sounds bad, but, in my researches, I couldn't find a way to circumvent this from a Sphinx perspective - unless skipping building the devguide altogether...

This shouldn't be much of a problem.  If `sphinx-build` is not found during `./configure` we disable building of the Sphinx documentation as part of a normal `make`.  Nothing is stopping a user from going into the documentation and calling Sphinx directly tho.  So as long as we prevent it from happening as part of the normal make if `mscgen` is not found, then we're good from our end of things.

jasonish:
> Getting the source code using curl and doing a `setup.py install` provided we install mscgen with `apt` (which is available and maintained)? I see that we use curl to get rust, and we already install setuptools which is all that seems to be required by `mscgen/setup.py` (https://github.com/sphinx-contrib/mscgen/blob/master/setup.py)

We can of course do whatever we want in our own CI. But when it comes to end users we shouldn't rely on them having to use `pip` or `curl` to get dependencies installed to build Suricata.

jufajardini:
> > With the check for mscgen in configure.ac, we could still face the situation where the person tries to build the devguide, nonetheless, right?
> > I'm wondering what should be done in that case - let the document be generated with broken images sounds bad, but, in my researches, I couldn't find a way to circumvent this from a Sphinx perspective - unless skipping building the devguide altogether...
> 
> This shouldn't be much of a problem. If `sphinx-build` is not found during `./configure` we disable building of the Sphinx documentation as part of a normal `make`. Nothing is stopping a user from going into the documentation and calling Sphinx directly tho. So as long as we prevent it from happening as part of the normal make if `mscgen` is not found, then we're good from our end of things.

Ok, that makes sense. Thanks again!

inashivb:
> 
> We can of course do whatever we want in our own CI. But when it comes to end users we shouldn't rely on them having to use `pip` or `curl` to get dependencies installed to build Suricata.

I apologize. I did not see the context properly.

jufajardini:
Followed by: https://github.com/OISF/suricata/pull/6233

PR #6207, opened 2021-06-21
Http range file v13

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/1576

Describe changes:
- Uses a generic container (a thread-safe hash table whose key is the url/filename) to store file parts with HTTP Range header
- handles reassembly of these parts, while enforcing a global memcap

suricata-verify-pr: 443

Modifies #5996 with handling gaps which are now handled by HTTP1 (ie `data == NULL && len > 0`)
They are handled by the `FileAppendData` util function

TODO: commit squashing 
Renaming to more specific ?
Test concurrency ? Is there a generic framework for this ?

PR #6226, opened 2021-06-24
Http range file v14

catenacyber:

Describe changes:

suricata-verify-pr: 443

Modifies #6207 with 
- Only inserting a range in the linked list when it is completed with data (and not at is creation)
- Using the mutex when timing out a `ContainerUrlRange`

TODO: commit squashing 
Renaming to more specific ?
More testing of concurrency ? Is there a generic framework for this ?


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6228

PR #6227, opened 2021-06-24
[DRAFT] Add base64 encoders to JsonBuilder - v1

jasonish:
Add methods `jb_set_base64` and `jb_append_base64` to JsonBuilder to set/append byte array values as base64. This avoids the intermediate buffer that we currently use, and the Rust base64 encode we use can encode directly to the underlying JsonBuilder buffer (internally it appears to use a 1024 byte intermediate buffer that is flushed to the output string when full).

Finally convert all occurrences of `Base64Encode` directly before using it in JsonBuilder with these new methods.

Would be good to have a performance test done. I haven't done my own micro-benchmarks yet, but may do so soon.

### DIFF
--- a/rust/Cargo.toml.in
+++ b/rust/Cargo.toml.in
@@ -49,6 +49,7 @@ sha-1 = "~0.9.2"
 md-5 = "~0.9.1"
 regex = "~1.4.2"
 lazy_static = "~1.4.0"
+base64 = "~0.13.0"
 
 [dev-dependencies]
 test-case = "~1.1.0"
--- a/rust/src/jsonbuilder.rs
+++ b/rust/src/jsonbuilder.rs
@@ -311,6 +311,30 @@ impl JsonBuilder {
         }
     }
 
+    /// Add a string to an array.
+    pub fn append_base64(&mut self, val: &[u8]) -> Result<&mut Self, JsonError> {
+        match self.current_state() {
+            State::ArrayFirst => {
+                self.buf.push('"');
+                base64::encode_config_buf(val, base64::STANDARD, &mut self.buf);
+                self.buf.push('"');
+                self.set_state(State::ArrayNth);
+                Ok(self)
+            }
+            State::ArrayNth => {
+                self.buf.push(',');
+                self.buf.push('"');
+                base64::encode_config_buf(val, base64::STANDARD, &mut self.buf);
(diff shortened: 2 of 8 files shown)


victorjulien:
Yeah I'd like to see some benchmarks, esp since this probably introduces heap allocations where none previously existed.

jasonish:
> Yeah I'd like to see some benchmarks, esp since this probably introduces heap allocations where none previously existed.

While we don't make much use of it, the JsonBuilders do support a `reset` which means they can be reused much like our current intermediate output buffer, so as they grow, the re-allocs will be less and less over the runtime of the program.

Anyways, will do some get some micro benchmarks done, and then we should also see if anything shows up in T-rex or something.

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

jasonish:
I created a crude benchmark that can be seen here: https://github.com/jasonish/scratch-suricata-base64-json-bench/blob/main/bench.c

The output is like:

```
Rust: 2.475810
C:   3.277675
Rust: 2.392453
C:   3.163671
Rust: 2.689030
C:   3.702352
```

Suggesting the Rust variant is faster.

From what I can tell the Rust variant uses a 1024 intermediate buffer on the stack compared to the C version that uses a full size buffer.  When this buffer is flushed, it is flushed directly to the underlying JsonBuilder buffer which may grow (on the heap), but this growth happens in both versions. So I'm not convinced there are heap allocations where none existed before.

I did have to add a new function `jb_append_string_safe` which appends a string without running it through escaping as a base64 string is already safe.  This brought the C times above down from 11s into the 3s.

- Take a 8192 size byte buffer.
- In loop with 5000 iterations, create a JsonBuilder array and base64 encode the buffer into the array 100 times.

suricata-qa:
|--|--|--|--|
| | generic_stats_chk |
| .tcp.reassembly_memuse | 2293760 | 1966080 | 86.0% |

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:
WARNING:
|--|--|--|--|
| | tlpr1_stats_chk |

Pipeline 3461

jasonish:
Replace by https://github.com/OISF/suricata/pull/6342.

PR #6229, opened 2021-06-24
mqtt: disable by default

catenacyber:
None

Describe changes:
- Effectively disables MQTT by default

#suricata-verify-pr: TODO

cc @satta do we want MQTT enabled or disabled by default ?
It is currently enabled (as the line in suricata.yaml.in is commented out), but the comment states otherwise...

### DIFF
--- a/suricata.yaml.in
+++ b/suricata.yaml.in
@@ -724,7 +724,7 @@ app-layer:
         dp: 5900, 5901, 5902, 5903, 5904, 5905, 5906, 5907, 5908, 5909
     # MQTT, disabled by default.
     mqtt:
-      # enabled: no
+      enabled: no
       # max-msg-length: 1mb
     krb5:
       enabled: yes


catenacyber:
Other option is to change the comment ;-)

By disabling MQTT by default, 4 out of 30 of the S-V MQTT tests fail

satta:
> Other option is to change the comment ;-)
> 
> By disabling MQTT by default, 4 out of 30 of the S-V MQTT tests fail

I'll look at it! Wanted to take some time this weekend to do the MQTT updates anyway, like the directional tx stuff.  I can polish this as well then. I think it should be fine to disable by default, but then that should not fail the S-V tests.

catenacyber:
> but then that should not fail the S-V tests.

Most S-V tests have a suricata.yaml file to enable mqtt, but not all of them...

satta:
Ah, So I just realized that the default is enabled if there is no entry in `app-layer.protocols.mqtt.enabled`. In this light I would rather enable by default and change the comment if that's OK. I thought it was disabled if there is no setting and then reflect that by keeping the `enabled: no` commented out. I think we should set `enabled: yes` here and change the comment.  Should I open a PR?

I will look at forcing MQTT support in the S-V tests anyway so they don't fail on changing the default. I'll open a separate PR for that there.

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6235

PR #6230, opened 2021-06-24
App maxtx 4530 v2

catenacyber:

Describe changes:
- limits the number of active transactions per flow

Default value is 4096 and configurable.

These protocols include HTTP2, MQTT, Modbus (tx id is only u16), SMB ?..

These maxed-out transactions should get logged even if they are incomplete

I do not find elegant that `AppLayerParserUpdateDisrupt` sets a fake STREAM_DEPTH
Do you see a better way ?
Casting u8 to a u16 and adding a new flag ?

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -227,7 +227,7 @@ int AppLayerParserSetup(void)
 }
 
 // default value is 4096
-uint32_t g_alparser_tx_max = 0x1000;
+uint32_t g_alparser_tx_max = 0x100;
 
 void AppLayerParserPostStreamSetup(void)
 {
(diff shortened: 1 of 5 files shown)


src/app-layer-parser.c
```diff
@@ -221,6 +226,9 @@ int AppLayerParserSetup(void)
     SCReturnInt(0);
 }
 
+// default value is 4096
+uint32_t g_alparser_tx_max = 0x100;
```
victorjulien, line 230: please use decimals for better readability.

also, 0x100 is not 4096, but 256

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6251

PR #6234, opened 2021-06-26
dcerpc: use null for default ports

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4475

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -1378,10 +1378,9 @@ pub const PARSER_NAME: &'static [u8] = b"dcerpc\0";
 
 #[no_mangle]
 pub unsafe extern "C" fn rs_dcerpc_register_parser() {
-    let default_port = CString::new("[0:65355]").unwrap();
     let parser = RustParser {
         name: PARSER_NAME.as_ptr() as *const std::os::raw::c_char,
-        default_port: default_port.as_ptr(),
+        default_port: b"\0".as_ptr() as *const std::os::raw::c_char,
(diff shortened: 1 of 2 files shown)


suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

rust/src/dcerpc/dcerpc.rs
```diff
@@ -1378,10 +1378,9 @@ pub const PARSER_NAME: &'static [u8] = b"dcerpc\0";
 
 #[no_mangle]
 pub unsafe extern "C" fn rs_dcerpc_register_parser() {
-    let default_port = CString::new("[0:65355]").unwrap();
     let parser = RustParser {
         name: PARSER_NAME.as_ptr() as *const std::os::raw::c_char,
-        default_port: default_port.as_ptr(),
+        default_port: b"\0".as_ptr() as *const std::os::raw::c_char,
```
jasonish, line 1383: This is creating an empty string, not null.  `std::ptr::null()` can be used here for a real null value.

jasonish (changes requested):
While an empty string might be good enough to get the same result, we should probably use a real null.

rust/src/dcerpc/dcerpc.rs
inashivb: Thanks. I thought this was null...

inashivb:
Replaced with #6234

PR #6241, opened 2021-06-30
devguide/transactions: add template examples

jufajardini:
Transactions documentation - v8

Describe changes:
- add examples from the Template protocol (including Sequence Diagrams)

(waiting for final approval to figure out the squashing process... If that's not ideal, I can fix that now.)

Link to Redmine issue: https://redmine.openinfosecfoundation.org/issues/4396

Previous PR: #6233

PR #6245, opened 2021-07-01
devguide/transactions: fix image filename

jufajardini:
Transactions documentation - v9

Describe changes:
- Previous PR somehow went through with an image name wrong. Fixing that.

(waiting for final approval to figure out the squashing process... If that's not ideal, I can fix that now.)

Link to Redmine issue: https://redmine.openinfosecfoundation.org/issues/4396

Previous PR: #6241

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2322,6 +2322,18 @@ fi
         AC_DEFINE([CLS],[64],[L1 cache line size])
     fi
 
+# mscgen for devguide images
+    AC_PATH_PROG([HAVE_MSCGEN], mscgen, "no")
+    if test "$HAVE_MSCGEN" = "no"; then
+        enable_mscgen=no
+        echo "WARNING! mscgen package not installed."
+        echo "    Devguide images won't be generated!"
+        echo "    Get mscgen package:"
+        echo "    https://www.mcternan.me.uk/mscgen/"
+        echo "    or install it from your distribution"
+    fi
+    AM_CONDITIONAL([HAVE_MSCGEN], [test "x$enable_mscgen" != "xno" ])
+
 # sphinx for documentation
     AC_PATH_PROG(HAVE_SPHINXBUILD, sphinx-build, "no")
     if test "$HAVE_SPHINXBUILD" = "no"; then
--- a/doc/devguide/extending/app-layer/img/generate-images.sh
+++ b/doc/devguide/extending/app-layer/img/generate-images.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+#
+# Script to generate Sequence Diagram images with mscgen
+#
+
+cd extending/app-layer/img
+
+for FILE in *.msc ; do
+    # call mscgen and convert each file in images dir
+    mscgen -T png -F Arial $FILE
+    # if command fails, lets inform about that
+    if [ $? -ne 0 ]; then
+        echo "$FILE couldn't be converted in the devguide"
+    fi
+done
+
+exit 0
--- a/doc/devguide/extending/app-layer/transactions.rst
+++ b/doc/devguide/extending/app-layer/transactions.rst
@@ -66,7 +66,7 @@ _`Progress Tracking`
 
 As a rule of thumb, transactions will follow a request-response model: if a transaction has had a request and a response, it is complete.
 
-But if a protocol has situations where a request or response won’t expect or generate a message from its counter-part,
+But if a protocol has situations where a request or response won’t expect or generate a message from its counterpart,
 it is also possible to have uni-directional transactions. In such cases, transaction is set to complete at the moment of
 creation.
 
@@ -86,7 +86,7 @@ In Summary - Transactions and State
 
 - Initial state value: ``0``
 - Simpler scenarios: state is simply an int.  ``1`` represents transaction completion, per direction.
-- Complex Transaction State in Suricata: ``enum`` (Rust: ``i32``). Completion is indicated by the highest enum value (some examples are: SSH, HTTP, DNS, SBM).
+- Complex Transaction State in Suricata: ``enum`` (Rust: ``i32``). Completion is indicated by the highest enum value (some examples are: SSH, HTTP, DNS, SMB).
 
 _`Examples`
 ===========
--- a/doc/devguide/extending/app-layer/transactions.rst
+++ b/doc/devguide/extending/app-layer/transactions.rst
@@ -114,12 +114,14 @@ The engine interacts with transactions state using a set of callbacks the parser
 In Summary - Transactions and State
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- Initial state value: ``0``
+- Initial state value: ``0``.
 - Simpler scenarios: state is simply an int.  ``1`` represents transaction completion, per direction.
 - Complex Transaction State in Suricata: ``enum`` (Rust: ``i32``). Completion is indicated by the highest enum value (some examples are: SSH, HTTP, DNS, SMB).
 
 _`Examples`
 ===========
+
+This section shares some examples from Suricata codebase, to help visualize how Transactions work and are handled by the engine.
 
 Enums
 ~~~~~
@@ -144,7 +146,6 @@ From src/app-layer-ftp.h:
         FTP_STATE_PORT_DONE,
         FTP_STATE_FINISHED,
     };
-
 
 API Callbacks
 ~~~~~~~~~~~~~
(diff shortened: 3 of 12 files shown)


suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

configure.ac
```diff
@@ -2322,6 +2322,18 @@ fi
         AC_DEFINE([CLS],[64],[L1 cache line size])
     fi
 
+# mscgen for devguide images
+    AC_PATH_PROG([HAVE_MSCGEN], mscgen, "no")
```
catenacyber, line 2326: I do not know if this should be part of this PR, but `disable-docs` would be a nice configure option

doc/devguide/extending/app-layer/img/generate-images.sh
```diff
@@ -0,0 +1,17 @@
...
+#
+
+cd extending/app-layer/img
+
+for FILE in *.msc ; do
+    # call mscgen and convert each file in images dir
+    mscgen -T png -F Arial $FILE
+    # if command fails, lets inform about that
+    if [ $? -ne 0 ]; then
+        echo "$FILE couldn't be converted in the devguide"
```
catenacyber, line 13: Should we not exit there ?

doc/devguide/extending/app-layer/transactions.rst
```diff
@@ -0,0 +1,299 @@
+************
+Transactions
+************
+
+.. contents:: Table of Contents
+
+_`General Concepts`
+===================
+
+Transactions are abstractions that help detecting and logging in Suricata. They also help during the detection phase,
```
catenacyber, line 10: Why `also` ? You already mentioned `detecting`

doc/devguide/extending/app-layer/transactions.rst
```diff
@@ -0,0 +1,299 @@
...
+when dealing with protocols that can have large PDUs, like TCP, in controlling state for partial rule matching, in case of rules that mention more than one field.
```
catenacyber, line 11: This does not seem clear to me as an introduction. I would rather begin with an example like a HTTP request and its response constitute one transaction

doc/devguide/extending/app-layer/transactions.rst
```diff
@@ -0,0 +1,299 @@
...
+             }
+        .
+        .
+        .
+    }
+
+Rule Matching
+~~~~~~~~~~~~~
+
+Transaction progress is also used for certain keywords to know what is the minimum state before we can expect a match: until that, Suricata won't even try to look for the patterns.
```
catenacyber, line 65: Would it be nice to have the list of such keywords when running `./src/suricata --list-keywords=all`

doc/devguide/extending/app-layer/transactions.rst
```diff
@@ -0,0 +1,299 @@
...
+- al, applayer: application layer
+- alproto: application layer protocol
+- alstate: application layer state
+- engine: refers to Suricata core detection logic
+- flow: a bidirectional flow of packets with the same 5-tuple elements (protocol, source ip, destination ip, source port, destination port. Vlans can be added as well)
+- PDU: Protocol Data Unit
+- rs: rust
+- tc: to client
+- ts: to server
+- tx: transaction
```
catenacyber, line 299: What about the files ?
I think it is worth mentioning the current work to have files be part of the transaction cf https://redmine.openinfosecfoundation.org/issues/4444

It may also be worth mentioning the current work to put a limit on the number of transactions per flow cf https://redmine.openinfosecfoundation.org/issues/4530

doc/devguide/extending/app-layer/transactions.rst
```diff
@@ -0,0 +1,299 @@
...
+progresses. A state will start at 0. The higher its value, the closer the transaction would be to completion.
+
+The engine interacts with transactions state using a set of callbacks the parser registers. State is defined per flow direction (``STREAM_TOSERVER`` / ``STREAM_TOCLIENT``).
+
+In Summary - Transactions and State
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Initial state value: ``0``.
+- Simpler scenarios: state is simply an int.  ``1`` represents transaction completion, per direction.
+- Complex Transaction State in Suricata: ``enum`` (Rust: ``i32``). Completion is indicated by the highest enum value (some examples are: SSH, HTTP, DNS, SMB).
```
catenacyber, line 119: HTTP2 also, they have a state diagram in their RFC

catenacyber (commented):
Thanks for this.
Is there some URL where we can see the nice thing with generated images ?

doc/devguide/extending/app-layer/transactions.rst
jufajardini: Could it be like a linked document, which I'd mention here?

doc/devguide/extending/app-layer/transactions.rst
jufajardini: This makes me realize I'm confused with DNS. That's because I'm mentioning it here as an example of a complex transaction state, while I also mention it before as an instance of a simpler protocol. Is that ok, could others get confused by that, as well? In that case, what can we do to dispel such confusion?

doc/devguide/extending/app-layer/transactions.rst
jufajardini: How do you suggest we add those? As a work in progress changes section? Or one with advanced topics? Or something else entirely?

doc/devguide/extending/app-layer/transactions.rst
catenacyber: I would rather like something automatically generated...

doc/devguide/extending/app-layer/transactions.rst
catenacyber: work in progress changes section seems good for me

doc/devguide/extending/app-layer/transactions.rst
catenacyber: You can keep DNS as a simplicity example when it fits

doc/devguide/extending/app-layer/img/generate-images.sh
jufajardini: I'm not sure about that, what if that was just with a specific file, because of bad naming, for instance?

doc/devguide/extending/app-layer/transactions.rst
jufajardini: Ok, could we generate that list automatically, but have the list on a different page, to prevent this one from getting too long?

doc/devguide/extending/app-layer/img/generate-images.sh
catenacyber: This looks like not supposed to happen, so let's stop the work so that the user notices ;-)

doc/devguide/extending/app-layer/transactions.rst
catenacyber: Yes, but my point was that this requires another development work first.
I do not know how to get that list

jufajardini: Oh, ok, sorry, I had gotten it wrongly at first. So we need something that allows us to list what keywords need that minimum stage, before matching? Should I create a ticket for that?

doc/devguide/extending/app-layer/img/generate-images.sh
jufajardini: That makes way more sense 😅  sorry for sounding stubborn sometimes. Patched.

doc/devguide/extending/app-layer/transactions.rst
catenacyber: > So we need something that allows us to list what keywords need that minimum stage, before matching?

I would first like to have the list of keywords that are matched within a transaction (and the what the other ones are matched on)

jufajardini:
> Thanks for this.
> Is there some URL where we can see the nice thing with generated images ?

I can set up something locally, would that work?

catenacyber:
> I can set up something locally, would that work?

No need if it is not ready yet, I will make it work locally

doc/devguide/extending/app-layer/transactions.rst
jufajardini: Ok, I think I must dig more in order to get a better grasp at this...

jufajardini:
Addressed topics that are settled (somewhat) in the following PR: https://github.com/OISF/suricata/pull/6265

PR #6247, opened 2021-07-02
lib: callback on tcp data

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4431

Describe changes:
- Adds a callback on tcp data (ie `AppLayerHandleTCPData`) to be used in libsuricata 

Built with `export CFLAGS=-DLIBSURICATA_BUILD=1`
Ngrep POC is available at https://github.com/catenacyber/ngrep-libsuricata

The other alternative I see to have some ngrep over libsuricata, benefiting from suricata TCP reassembly, is to have a big architectural change in the tcp reassembly to have sequentially tcp reassembly, then app-layer processing on data supplied by tcp reassembly,
ie to call `AppLayerHandleTCPData` (or `StreamTcpReassembleAppLayer`) directly from `FlowWorker` after `FlowWorkerStreamTCPUpdate`, instead of having `AppLayerHandleTCPData` getting called in a 9-deep functions stack
This alternative is complex, and should also take into account the fact that suricata TCP reassembly depends on app-layer processing (sic !) for the case with a request like `GOT / HTTP/1.1` (not GET) and response like `HTTP/1.1 200 OK` where protocol detection recognizes HTTP1 in the response (and nothing in the request), but then wants to parse the HTTP1-ish request before parsing the response. cf `AppLayerTest05`

### DIFF
--- a/src/app-layer.c
+++ b/src/app-layer.c
@@ -570,6 +570,10 @@ static int TCPProtoDetect(ThreadVars *tv,
     SCReturnInt(0);
 }
 
+#ifdef LIBSURICATA_BUILD
+libsuricata_tcp_cb AppLayerTCPDataCallback = NULL;
+#endif
+
 /** \brief handle TCP data for the app-layer.
  *
  *  First run protocol detection and then when the protocol is known invoke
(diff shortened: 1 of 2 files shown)


jasonish:
Not too keen on the compile time option.  In the bigger picture of a `libsuricata`, Suricata itself is the main consumer so I think we need to avoid scenarios where:
- Only using libsuricata for Suricata itself, compile as normal.
- Using libsuricata for something else, add these compile time options as well.

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

catenacyber:
Continued in https://github.com/OISF/suricata/pull/6252

PR #6251, opened 2021-07-05
App maxtx 4530 v3

catenacyber:

Describe changes:
- limits the number of active transactions per flow

Default value is 4096 and configurable.

These protocols include HTTP2, MQTT, Modbus (tx id is only u16), SMB ?..

These maxed-out transactions should get logged even if they are incomplete

I do not find elegant that `AppLayerParserUpdateDisrupt` sets a fake STREAM_DEPTH
Do you see a better way ?
Casting u8 to a u16 and adding a new flag ?

Replaces #6230 with better decimal readability

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -164,6 +164,11 @@ struct AppLayerParserState_ {
 
     uint64_t min_id;
 
+    /* There are incomplete transactions which should get removed
+     * between min_id and purge_before_tx_id
+     */
+    uint64_t purge_before_tx_id;
+
     /* Used to store decoder events. */
     AppLayerDecoderEvents *decoder_events;
 };
@@ -221,6 +226,9 @@ int AppLayerParserSetup(void)
     SCReturnInt(0);
 }
 
+// default value is 0x1000
+uint32_t g_alparser_tx_max = 4096;
+
 void AppLayerParserPostStreamSetup(void)
 {
     AppProto alproto = 0;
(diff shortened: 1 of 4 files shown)


suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:
Information:
|--|--|--|--|
| | tlpr1_stats_chk |

suricata-qa:

Pipeline

src/app-layer-parser.c
```diff
@@ -164,6 +164,11 @@ struct AppLayerParserState_ {
 
     uint64_t min_id;
 
+    /* There are incomplete transactions which should get removed
+     * between min_id and purge_before_tx_id
+     */
+    uint64_t purge_before_tx_id;
```
victorjulien, line 170: I would like to see if we can avoid adding this here.

src/app-layer-parser.c
```diff
@@ -221,6 +226,9 @@ int AppLayerParserSetup(void)
     SCReturnInt(0);
 }
 
+// default value is 0x1000
+uint32_t g_alparser_tx_max = 4096;
```
victorjulien, line 230: thinking about this a bit, it seems to make sense to make this a per alproto setting, where some protocols will use much lower defaults

src/app-layer-parser.c
```diff
@@ -914,6 +944,7 @@ void AppLayerParserTransactionsCleanup(Flow *f)
     bool skipped = false;
     const bool is_unidir =
             AppLayerParserGetOptionFlags(f->protomap, f->alproto) & APP_LAYER_PARSER_OPT_UNIDIR_TXS;
+    uint64_t nbtx = 0;
```
victorjulien, line 947: I guess we could avoid the double loop if we'd have a proper way to get an actual tx cnt from the parser (instead of the misnomed call to get the max id)

victorjulien:
I think we need something like this, but I'm not liking the implementation yet. Needs some more thought. I'll also think about it a bit more.

catenacyber:
> I think we need something like this, but I'm not liking the implementation yet. Needs some more thought. I'll also think about it a bit more.

Thanks for the comments
I will look into them and give the whole thing some thought as well...

victorjulien:
I guess one way to do it would be to move the logic to the parsers themselves. This would make some sense as I can imagine that we want specific settings per protocol (and maybe even different logic, e.g. which txs are evicted... file tx staying in even if older?) and would also be easy to implement as the parser knows its own tx count. Additionally it would allow the parser to just set the tx progress values to 'complete' so it is considered complete by the rest of the engine w/o the need for updates to the core logic.

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6661

PR #6259, opened 2021-07-07
Fuzz perf v2

catenacyber:

Describe changes:
- avoid quadratic complexity in `DetectAppLayerInspectEngineCopyListToDetectCtx` and `DetectMpmInitializeAppMpms`
- use smaller hash tables when compiled in unsafe fuzzing mode (faster for reset)
- avoids useless reads of config when compiled for fuzzing
- no eve log for fuzzing

This is a draft PR for discussion :
- The first 2 commits seem good in every case for me.
- The additions of `#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` make the code more complex even if fuzzing is faster
- The removal of eve log will cover less code, but anomaly logging is expensive while fuzzing...

cf https://github.com/google/oss-fuzz/pull/5995


catenacyber:
Hash tables seem to rely on the hash function to respect their sizes... This looks dangerous...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6260

PR #6263, opened 2021-07-08
Unify reg nfs+smb/v1

inashivb:


suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |
| .tcp.insert_list_fail | 0 | 255 | 0.0% |

rust/src/nfs/nfs.rs
```diff
@@ -1859,3 +1928,156 @@ pub extern "C" fn rs_nfs_setfileflags(direction: u8, ptr: *mut NFSState, flags:
...
+                parser.name, ALPROTO_NFS, 0, NFS_MIN_FRAME_LEN,
+                rs_nfs_probe_udp_ts, rs_nfs_probe_udp_tc) == 0 {
+            SCLogDebug!("No NFSUDP app-layer configuration, enabling NFSUDP
+                        detection UDP detection on port {:?}.",
+                        default_port);
+            AppLayerProtoDetectPPRegister(IPPROTO_UDP as u8,
+                default_port.as_ptr(), ALPROTO_NFS, 0,
+                NFS_MIN_FRAME_LEN, STREAM_TOSERVER,
+                rs_nfs_probe_udp_ts, rs_nfs_probe_udp_tc);
+        }
```
jasonish, line 2070: I don't think this bit is required as its already done inside `AppLayerRegisterProtocolDetection`.

jasonish: Oh, but the probe functions aren't set in the registration struct.  If there is a reason for doing this, instead of doing it how it is in `dns.rs` for example, can you elaboroate with a comment as to why?

catenacyber: I agree with Jason, `rs_nfs_probe_udp_ts ` should be defined as `probe_ts` in `rs_nfs_register_parser`
Thus, we likely do not need `AppLayerProtoDetectPPRegister` in rust

rust/src/core.rs
```diff
@@ -64,6 +64,9 @@ macro_rules!BIT_U64 {
     ($x:expr) => (1 << $x);
 }
 
+// Flow flags
+pub const FLOW_DIR_REVERSED: u32 = BIT_U32!(26);
```
catenacyber, line 68: @jasonish we really need to use rust constants on the C side...

rust/src/smb/files.rs
```diff
@@ -111,7 +111,7 @@ impl SMBState {
     }
     fn setfileflags(&mut self, direction: u8, flags: u16) {
         SCLogDebug!("direction: {}, flags: {}", direction, flags);
-        if direction == 1 {
+        if direction == STREAM_TOCLIENT {
```
catenacyber, line 115: This commit is not complete as `rs_smb_setfileflags` still use 0 or 1

rust/src/smb/smb.rs
```diff
@@ -26,175 +26,186 @@
...
 #[no_mangle]
-pub extern "C" fn rs_smb_init(context: &'static mut SuricataFileContext)
-{
+pub extern "C" fn rs_smb_init(context: &'static mut SuricataFileContext) {
     unsafe {
         SURICATA_SMB_FILE_CONFIG = Some(context);
     }
 }
 
-pub const SMB_NTSTATUS_SUCCESS:                    u32 = 0;
```
catenacyber, line 65: The commit name is not about rustfmt.
Could we remove the formatting out of this commit ?

rust/src/smb/smb.rs
```diff
@@ -2236,3 +2237,143 @@ pub extern "C" fn rs_smb_state_get_event_info(
     }
     0
 }
+
+pub extern "C" fn smb3_probe_tcp(f: *const Flow, dir: u8, input: *const u8, len: u32, rdir: *mut u8) -> u16 {
+    let retval = rs_smb_probe_tcp(f, dir, input, len, rdir);
+    let f = cast_pointer!(f, Flow);
+    unsafe {
```
catenacyber, line 2244: This looks like a big unsafe...

src/app-layer-smb.c
```diff
@@ -28,199 +28,6 @@
...
-
-static AppLayerResult SMBTCPParseRequest(Flow *f, void *state,
-        AppLayerParserState *pstate, const uint8_t *input, uint32_t input_len,
-        void *local_data, const uint8_t flags)
-{
-    SCLogDebug("SMBTCPParseRequest");
-    uint16_t file_flags = FileFlowToFlags(f, STREAM_TOSERVER);
-    rs_smb_setfileflags(0, state, file_flags|FILE_USE_DETECT);
-
-    if (input == NULL && input_len > 0) {
```
catenacyber, line 41: I do not see this gap handling code in rust in this commit where it gets removed from C... Am I missing something ?

rust/src/nfs/nfs.rs
```diff
@@ -1799,57 +1824,83 @@ pub extern "C" fn rs_nfs_probe_ms(
...
                 SCLogDebug!("nfs_probe_dir said STREAM_TOCLIENT");
             }
-            let r = nfs_probe(slice, adirection);
-            if r == 1 {
-                SCLogDebug!("nfs_probe success: dir {:02x} adir {:02x}", direction, adirection);
-                if (direction & (STREAM_TOSERVER|STREAM_TOCLIENT)) != adirection {
-                    unsafe { *rdir = adirection; }
-                }
-                return 1;
+            match nfs_probe(slice, adirection) {
```
catenacyber, line 1845: I wonder if `nfs_probe` should return `ALPROTO_NFS`. instead of 1... Thoughts ?

rust/src/nfs/nfs.rs
```diff
@@ -39,6 +39,8 @@ use crate::nfs::nfs3_records::*;
 
 pub static mut SURICATA_NFS_FILE_CONFIG: Option<&'static SuricataFileContext> = None;
 
+pub const NFS_MIN_FRAME_LEN: u16 = 32;
```
catenacyber, line 43: Why a specific commit for the definition (and not when it gets used)

rust/src/nfs/nfs.rs
```diff
@@ -21,13 +21,14 @@ use std;
...
 use std::mem::transmute;
 use std::collections::{HashMap};
-use std::ffi::CStr;
+use std::ffi::{CStr, CString};
 
 use nom;
 
 use crate::applayer;
-use crate::applayer::{AppLayerResult, AppLayerTxData};
+use crate::applayer::{AppLayerResult, AppLayerTxData, *};
```
catenacyber, line 29: Guess you do not need the first id you import `*`

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |
| .tcp.insert_list_fail | 0 | 255 | 0.0% |

suricata-qa:
|--|--|--|--|
| | tlpr1_stats_chk |
| .tcp.insert_list_fail | 0 | 255 | 0.0% |

rust/src/nfs/nfs.rs
inashivb: Can we also use `AppLayerProtoDetectPPParseConfPorts` somehow as a part of the registration struct? I used it this way because I did not know how to incorporate conf port check here...

src/app-layer-smb.c
inashivb: It is added in the commit before this removal.

inashivb:
Replaced with #6285

PR #6264, opened 2021-07-12
 Integrate DPDK capture interface into Suricata v2

lukashino:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/4507) ticket:

Describe changes:
- Add STAILQ implementation (required for DPDK).
- Integrate DPDK capture interface as an alternative to AF_PACKET or others.
- Focus implementation on workers runmode.
- Support both IDS and IPS modes.
- Follow AF_PACKET capture interface architecture and apply DPDK best practices for development.
- Update documentation to reflect new changes.

Work has been based on DPDK 21.05 but I believe running it on earlier versions would be also possible.

This is a follow-up pull request of #6159. I have reformatted all source file changes according to clang. However the check will probably not pass regardless because there are still many source files not following clang style guides. Please correct me if I am wrong.

PR #6269, opened 2021-07-15
Integrate dpdk runmode v3

lukashino:

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/4507) ticket:

This is a follow-up pull request of #6264. I have progressed according to @catenacyber comments. 

Describe changes:
- Add STAILQ implementation (required for DPDK).
- Integrate DPDK capture interface as an alternative to AF_PACKET or others.
- Focus implementation on workers runmode.
- Support both IDS and IPS modes.
- Follow AF_PACKET capture interface architecture and apply DPDK best practices for development.
- Update documentation to reflect new changes.

Work has been based on DPDK 21.05 but I believe running it on earlier versions would be also possible.


src/queue.h
```diff
@@ -254,6 +254,93 @@ struct {								\
 	_Q_INVALIDATE((elm)->field.le_next);				\
 } while (0)
 
+/*
+ * Singly-linked Tail queue declarations.
+ */
+#define STAILQ_HEAD(name, type)                                                                    \
```
catenacyber, line 260: I wonder if it is worth, or if using `TAILQ` is good enough...
What do you think ?

src/runmode-dpdk.c
```diff
@@ -0,0 +1,1278 @@
...
+#include "util-byte.h"
+
+#ifdef HAVE_DPDK
+#include <rte_log.h>
+#include <rte_mempool.h>
+#include <rte_eal.h>
+#include <rte_lcore.h>
+
+#define RSS_HKEY_LEN 40
+static uint8_t rss_hkey[] = { 0x6D, 0x5A, 0x6D, 0x5A, 0x6D, 0x5A, 0x6D, 0x5A, 0x6D, 0x5A, 0x6D,
```
catenacyber, line 47: What is this hardcoded value ?

src/util-runmodes.c
```diff
@@ -49,6 +49,14 @@
 
 #include "flow-hash.h"
 
+#ifdef HAVE_DPDK
+#include <rte_lcore.h>
+#include <rte_ethdev.h>
+#endif /* HAVE_DPDK */
+
+// equal to -1 to get the first lcore id
+int last_spawned_lcore = -1;
```
catenacyber, line 58: Should be in the ifdef ?

catenacyber (commented):
I do not have anything else to add.
Someone with a better knowledge of packet sources should review this

src/runmode-dpdk.c
norg: This is the haskey for RSS symmetric hashing. That is also used in our guides for high performance, see https://suricata.readthedocs.io/en/latest/performance/packet-capture.html

src/queue.h
lukashino: DPDK requires STAILQ for its function. It loads `sys/queue.h` and since Suricata redefines the file, DPDK misses STAILQ. I know from my earlier experiments that putting DPDK headers at the top of each file kind of worked but if I am not mistaken, it just loaded `sys/queue.h` sooner than Suricata could. I made an attempt to find a better solution. 

For that reason, I added STAILQ implementation to `src/queue.h`. I am currently not aware of possibility to change STAILQ to TAILQ in DPDK. I am, however, open to put it in `#ifdef HAVE_DPDK` clause if required.

src/runmode-dpdk.c
lukashino: This value is a special hash key that configures receive side scaling in a way that it puts packets of one flow (but from both directions) into one NIC queue. Packets are symmetrically scattered to NIC queues using a hash function. 
Suricata workers then fetch packets from these NIC queues for further processing. Configuring RSS with this hash key ensures packets of one flow land on the same worker every time, thus not requiring multi-thread communication. Other hash keys usually do not have this property.

Other receive runmodes (e.g. AF_PACKET) take the configured NIC and interact with them without any configuration. However, in DPDK each application that uses a NIC is required to configure the NIC by itself. For that reason, it was necessary to implement also the configuration phase (you would normally do that with `ethtool` prior to running Suricata). 

More on the RSS hash key value: https://www.ndsl.kaist.edu/~kyoungsoo/papers/TR-symRSS.pdf

src/util-runmodes.c
lukashino: True, it is irrelevant to all runmodes other than the DPDK runmode. I will move it there, thanks!

src/queue.h
catenacyber: Fair enough for me :-)

vipinpv85: @lukashino the manual changes are not required. There is a better way to include the missing elements. Please revisit the same

src/runmode-dpdk.c
vipinpv85: @lukashino in port_conf you are assigning RTE_IP_RSS, but in the hash key you have defines suing layer 4. there is conflict in what is been explained and used, Please rework

src/source-dpdk.h
```diff
@@ -0,0 +1,91 @@
...
+
+typedef struct DPDKIfaceConfig_ {
+#ifdef HAVE_DPDK
+    char iface[RTE_ETH_NAME_MAX_LEN];
+    uint16_t port_id;
+    uint16_t socket_id;
+    /* number of threads - zero means all available */
+    int threads;
+    /* IPS mode */
+    DpdkCopyModeEnum copy_mode;
```
vipinpv85, line 53: in IPS mode you are not copying packets from the Ingress interface, It is in IDS mode one needs to mirror the packet (if required). But normally IDS is deployed on mirrored traffic, if are you planning to mirror the packet there is no code found to mirror via SW or using HW. please rework.

src/tm-threads.c
```diff
@@ -1291,7 +1338,17 @@ static int TmThreadKillThread(ThreadVars *tv)
     }
 
     /* join it and flag it as dead */
-    pthread_join(tv->t, NULL);
+#ifdef HAVE_DPDK
+    if (tv->lcore_id != LCORE_ID_ANY && tv->type == TVT_PPT) {
+        /* wait for DPDK worker threads */
+        rte_eal_wait_lcore(tv->lcore_id);
```
vipinpv85, line 1344: the DPDK API can only be invoked from the main thread. It can not be called from sibling thread http://doc.dpdk.org/api/rte__launch_8h.html#ae9500e1d35bd4cfb95d18c0be863cb1e

src/tm-threads.c
```diff
@@ -1713,16 +1770,32 @@ TmEcode TmThreadSpawn(ThreadVars *tv)
...
 
-    /* Initialize and set thread detached attribute */
-    pthread_attr_init(&attr);
+#ifdef HAVE_DPDK
+    /* only DPDK worker threads are launched using rte_eal_remote_launch */
+    if (tv->lcore_id != LCORE_ID_ANY) {
+        struct dpdk_callback cb = { .func = tv->tm_func, .tv = tv };
 
-    pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_JOINABLE);
+        int rc = rte_eal_remote_launch(&TmThreadSpawnDpdkThread, (void *)&cb, tv->lcore_id);
```
vipinpv85, line 1778: Suricata provides API calls which launches on the worker therads as defined in YAML file. please rework this logic

src/tm-threads.c
```diff
@@ -220,6 +239,27 @@ static int TmThreadTimeoutLoop(ThreadVars *tv, TmSlot *s)
...
+ * With DPDK enabled, the function encapsulates pthread exit with a check for DPDK worker threads.
+ * If the thread has lcore_id attribute set to a different value than LCORE_ID_ANY, it does not
+ * exit because it is identified as a DPDK worker. DPDK worker threads are joined at a single point.
+ * Threads other than DPDK workers exit with pthred_exit.
+ *
+ */
+static void TmThreadExit(ThreadVars *tv, int64_t exit_code)
+{
+#ifdef HAVE_DPDK
+    if (tv->lcore_id == LCORE_ID_ANY) {
```
vipinpv85, line 255: I have disagree with this flow in which all threads are put under DPDK, this is similar to the early work done https://github.com/vipinpv85/DPDK-Suricata_3.0. The same has been modified not to do so. Please rework

src/tm-threads.c
```diff
@@ -43,6 +43,11 @@
 #include "util-signal.h"
 #include "queue.h"
 
+#ifdef HAVE_DPDK
+#include <rte_launch.h>
```
vipinpv85, line 47: combine all DPDK header inclsuion into a common file.

src/source-dpdk.h
lukashino: I think there is a slight misunderstanding here. During my development, I tried to follow AF_PACKET capture runmode to keep consistent with the architecture/naming conventions. This field does not in fact represent packet mirroring. [The same way it is used in AF_PACKET](https://github.com/OISF/suricata/blob/b8499de498f0e31bb31097ac8b959e3877573de0/src/source-af-packet.h#L68). This field holds information whether Suricata is in None (IDS) / TAP(IPS) / True IPS mode. 
And if Suricata is in either TAP or IPS mode, the packets are transmitted in packet release function. ([like here](https://github.com/OISF/suricata/blob/b8499de498f0e31bb31097ac8b959e3877573de0/src/source-af-packet.c#L828)). If Suricata is in IDS mode, the packet is only released.
So to wrap up, there is no packet mirroring and copy_mode only represents the mode Suricata is in. DPDK runmode does not copy packets anywhere. The name was only derived from AF_PACKET implementation to keep implementations consistent.

src/tm-threads.c
lukashino: This function is only my wrapper to end DPDK worker threads only. I added the function in the latest PR. Possibly I can rename the function. 
But do I understand you correctly that you would like to run only worker threads under DPDK? If so, it is already implemented this way.

src/queue.h
lukashino: I am not sure what you mean by that. I might be missing something. Could you please elaborate how that can be done in a better way?

src/runmode-dpdk.c
lukashino: Thanks, I've changed it to ETH_RSS_TCP | ETH_RSS_UDP
If I am not mistaken, this should be correct.

src/tm-threads.c
lukashino: That is true indeed. But this piece of code is always called from the main thread. Function TmThreadKillThread is called from the main thread. The function accepts ThreadVars structure of the thread to be killed. So in this case, in contrast to functions like in receive modules (e.g. receive loop), `tv` (ThreadVars structure) does not represent the thread itself but the thread it is supposed to kill. Above mentioned "if" statement then checks if the killed thread is really a DPDK worker thread. 
Attribute `tv->lcore_id` is set to a different value only if it is a DPDK thread (the structure has value LCORE_ID_ANY by default and only if the thread is spawned under DPDK, the lcore_id is assigned.) In current circumstances, only worker threads are spawned under DPDK. Other threads remain as pthreads. Therefore, checking value of `tv->type` serves as an extra check to make sure the thread is a DPDK worker thread.

src/tm-threads.c
lukashino: I think I am a bit lost here. Could you please be more specific? I am not sure what API calls or what parts of YAML file you mean. 
`TmThreadSpawnDpdkThread` is a wrapper to run tv->tm_func because `rte_eal_remote_launch` accepts a different function pointer compared to `pthread_create`. It does not do anything more than that. 
CPU affinity is reassigned later in the thread initialization process if defined differently than `tv->lcore_id`.
Thanks for explanation.

src/tm-threads.c
lukashino: Done.

lukashino:
@vipinpv85 Thanks for your comments, I have made minor adjustments discussed in this PR. To keep it updated, I have created a new PR #6272 with all the changes included.

PR #6271, opened 2021-07-19
ipv6: decoder event on invalid length

catenacyber:

Describe changes:
- Adds a decoder event when an IPv6 fragment has an invalid length according to the RFC

suricata-verify-pr: 519
https://github.com/OISF/suricata-verify/pull/519

### DIFF
--- a/src/decode-ipv6.c
+++ b/src/decode-ipv6.c
@@ -453,6 +453,12 @@ DecodeIPV6ExtHdrs(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
                     plen -= hdrextlen;
                     break;
                 }
+                if (p->ip6eh.fh_more_frags_set != 0 && plen & 7) {
+                    // cf https://datatracker.ietf.org/doc/html/rfc2460#section-4.5
+                    // each, except possibly the last ("rightmost") one,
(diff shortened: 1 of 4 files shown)


suricata-qa:

src/decode-ipv6.c
```diff
@@ -453,6 +453,12 @@ DecodeIPV6ExtHdrs(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p,
                     plen -= hdrextlen;
                     break;
                 }
+                if (p->ip6eh.fh_more_frags_set != 0 && plen & 7) {
```
victorjulien, line 456: what is the `& 7` here?

catenacyber: That is the check that each fragment has a length being an integer multiple of 8 octets long.

catenacyber: I can write `plen % 8 != 0` if that is more readable and the compiler should get it optimized, right ?

victorjulien: Yeah that helps my tiny brain to also understand it ;)

catenacyber: Done

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6310

PR #6281, opened 2021-07-28
postgresql: add initial support (WIP) - checkpoint

jufajardini:
Link to redmine ticket:

Previous PR: https://github.com/OISF/suricata/pull/6062

Describe changes:
This version covers most of the messages related to startup phase and simple query, except for `NegotiateProtocolVersion` - must understand how this happens, still.

The logger has filterable fields (eve.json), but might be too verbose at this point.

Considerations:
- Transactions need refining, to ensure messages are in the correct transaction.
- Doesn't upgrade to `TLS` (STARTTLS) yet.
- In same cases, it seems that some messages are not logged (happens with pgsql-jdbc.pcap, available in gdrive). This may be due to them being part of extended query protocol message flow. Shall investigate further. 

Mostly tested with the pcaps available here:
https://drive.google.com/drive/folders/1BPJICt_ybkofDUPrMdGglvfQeefFCQF_?usp=sharing


jufajardini:
the commit-check fail is, as far as I could understand, from a commit when the logger hadn't been converted into JsonBuilder yet...

suricata-qa:
ERROR: QA failed on generic_collect.

Pipeline 3639

rust/src/pgsql/log.rs
```diff
@@ -0,0 +1,420 @@
...
+            PgsqlErrorNoticeFieldType::Line => {
+                jb.set_string_from_bytes("line", &field.field_value)?;
+            },
+            PgsqlErrorNoticeFieldType::Routine => {
+                jb.set_string_from_bytes("routine", &field.field_value)?;
+            },
+            PgsqlErrorNoticeFieldType::TerminatorToken => {
+                // do nothing, as that's the terminator token ?
+            },
+            PgsqlErrorNoticeFieldType::UnknownFieldType => {
```
victorjulien, line 388: is this unknown to suri, or unknown in the pgsql protocol spec?

rust/src/pgsql/parser.rs
```diff
@@ -0,0 +1,2088 @@
...
+            'F' => PgsqlErrorNoticeFieldType::File,
+            'L' => PgsqlErrorNoticeFieldType::Line,
+            'R' => PgsqlErrorNoticeFieldType::Routine,
+            '\u{0}' => PgsqlErrorNoticeFieldType::TerminatorToken,
+            _ => PgsqlErrorNoticeFieldType::UnknownFieldType, // adding this because documentation says "frontends should silently ignore fields of unrecognized type."
+        }
+    }
+}
+
+impl From<u8> for PgsqlErrorNoticeFieldType {
```
victorjulien, line 394: can these 2 funcs somehow be merged, or use a common util func?

rust/src/pgsql/pgsql.rs
```diff
@@ -0,0 +1,957 @@
...
+    }
+}
+
+impl Drop for PgsqlTransaction {
+    fn drop(&mut self) {
+        self.free();
+    }
+}
+#[derive(Debug, PartialEq)]
+pub enum PgsqlStateProgress {
```
victorjulien, line 91: this looks quite similar to `PgsqlTransactionState`, is there a need for both to exist?

rust/src/pgsql/pgsql.rs
```diff
@@ -0,0 +1,957 @@
...
+        if  self.state_progress == PgsqlStateProgress::IdleState ||
+            self.state_progress == PgsqlStateProgress::StartupMessageReceived ||
+            self.state_progress == PgsqlStateProgress::ConnectionCompleted ||
+            self.state_progress == PgsqlStateProgress::ReadyForQueryReceived ||
+            self.state_progress == PgsqlStateProgress::SslRejectedReceived {
+                let tx = self.new_tx();
+                self.transactions.push(tx);
+            }
+            // If we don't have to create a new transaction, just return the current one
+            return self.transactions.last_mut().unwrap();
```
victorjulien, line 185: is it possible that there is no tx yet, so this `unwrap` would fail?

rust/src/pgsql/pgsql.rs
```diff
@@ -0,0 +1,957 @@
...
+            match self.state_progress {
+                PgsqlStateProgress::IdleState |
+                PgsqlStateProgress::ConnectionCompleted |
+                PgsqlStateProgress::SslRejectedReceived |
+                PgsqlStateProgress::SslAcceptedReceived |
+                PgsqlStateProgress::ReadyForQueryReceived |
+                PgsqlStateProgress::CommandCompletedReceived => {
+                    match parser::parse_request(start) {
+                        // TODO Question should I also add "&& self.transactions.is_empty()"
+                        Ok((rem, request)) => {
```
victorjulien, line 225: think it would be good to move this block into a util func to keep the main loop more readable

victorjulien (commented):
Some comments and questions. Nothing major. Nice work :)

rust/src/pgsql/log.rs
jufajardini: The specification lists existing field types, but states that more types can be added in the future - therefore, frontends should silently ignore unrecognized field types.

rust/src/pgsql/pgsql.rs
jufajardini: They're indeed similar, and I'm still trying to figure that out, to be honest. First, we thought that progress would be tracked per transaction, then, due to how postgresql protocol transactions work, we thought that maybe tracking it just in state would be enough. Currently, basically only `PgsqlStateProgress` is being used. But I would say I must investigate the logs a bit more, to understand if Suri can interpret the traffic correctly with `StateProgress` only.

rust/src/pgsql/parser.rs
jufajardini: Checking this properly, now, those two cannot, as this `From<u8>` is for the `ErrorMessages` and `NoticeMessages` types, not the front end messages.

The From functions are also used by the nom parsers, in the cases where that makes sense, and return types compatible with said parsers. Whereas the get_message_type is used by the app-layer parsing the requests and responses. So I'm guessing that some documentation is needed here, to explain that (plus proper naming the `get_message_type`one)...

victorjulien:
@jasonish can you have a look at the EVE output and see if that is what we'd want vs how verbose it is?

jasonish:
With respect to logging... I find the multiple requests and responses in one of the samples a little confusing:
```json
{
  "timestamp": "2020-12-30T15:36:39.846111+0000",
  "flow_id": 131952644348043,
  "pcap_cnt": 41,
  "event_type": "pgsql",
  "src_ip": "10.16.1.10",
  "src_port": 40816,
  "dest_ip": "10.16.1.11",
  "dest_port": 5432,
  "proto": "TCP",
  "pgsql": {
    "transaction_id": 3,
    "requests": [
      {
        "startup_message": {
          "protocol_version": "3.0",
          "startup_parameters": {
            "user": "indexer",
            "database": "indexer",
            "application_name": "psql",
            "client_encoding": "UTF8"
          }
        }
      },
      {
        "password_message": {
          "payload": "md5e4cfa9552468cae5d48ca2822ca36e22"
        }
      }
    ],
    "responses": [
      {
        "authentication_md5_password": {
          "payload": "\\x9fi\u001a\\x8e"
        }
      },
      {
        "authentication_ok": {
          "payload": ""
        }
      },
      {
        "parameter_status": {
          "parameter_name": "application_name",
          "parameter_value": "psql"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "client_encoding",
          "parameter_value": "UTF8"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "DateStyle",
          "parameter_value": "ISO, MDY"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "integer_datetimes",
          "parameter_value": "on"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "IntervalStyle",
          "parameter_value": "postgres"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "is_superuser",
          "parameter_value": "on"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "server_encoding",
          "parameter_value": "UTF8"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "server_version",
          "parameter_value": "13.0 (Debian 13.0-1.pgdg100+1)"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "session_authorization",
          "parameter_value": "indexer"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "standard_conforming_strings",
          "parameter_value": "on"
        }
      },
      {
        "parameter_status": {
          "parameter_name": "TimeZone",
          "parameter_value": "Etc/UTC"
        }
      },
      {
        "backend_key_data": {
          "process_id": 61,
          "secret_key": 3152142766
        }
      }
    ]
  }
}
```
I think the above is 2 distinct transactions and should be logged as such.  One request, one response per transaction.  Might make state handling simpler, as well as querying the database and the other end of this.

jasonish:
I'm still not sure on this style of parameter logging:
```json
      {
        "parameter_status": {
          "parameter_name": "standard_conforming_strings",
          "parameter_value": "on"
        }
      },
```

I need to think about it more on the event management side of things, but for startes I think we should rename the `name` and `value` fields to simply `name` and `value`, ie)
```
      {
        "parameter_status": {
          "name": "standard_conforming_strings",
          "value": "on"
        }
      },
```

jasonish:
```json
{
  "timestamp": "2020-12-30T15:36:39.846111+0000",
  "flow_id": 131952644348043,
  "pcap_cnt": 41,
  "event_type": "pgsql",
  "src_ip": "10.16.1.10",
  "src_port": 40816,
  "dest_ip": "10.16.1.11",
  "dest_port": 5432,
  "proto": "TCP",
  "pgsql": {
    "transaction_id": 6,
    "response": [
      {
        "ready_for_query": {
          "transaction_status": "I"
        }
      }
    ]
  }
}
```
I think this ready for query belongs in the previous transaction, at least when the request is a command/query string.  From what I can tell from the protocol documentation, you could get multiple `CommandComplete`, but you'll only get a single `ReadyForQuery` which should be at the very end, at which point the server starts waiting.

rust/src/pgsql/log.rs
```diff
@@ -0,0 +1,420 @@
...
+                params}) =>
+        {
+            let proto = format!("{}.{}", proto_major, proto_minor);
+            js.set_string("protocol_version", &proto)?;
+            let jb = log_pgsql_parameters(params)?;
+            js.set_object("startup_parameters", &jb)?;
+        },
+        PgsqlFEMessage::SslRequest(_) =>
+        {
+            js.set_string("message_type", "SSL Request")?;
```
jasonish, line 86: I think this should be `SSLRequest` as it seems to be consistently like that in the Postgres docs.

rust/src/pgsql/log.rs
```diff
@@ -0,0 +1,420 @@
...
+        PgsqlBEMessage::AuthenticationSASLContinue(_) => {
+            js.set_string("authentication_SASL_continue", &res.to_string())?;
+        },
+        PgsqlBEMessage::ParameterStatus(ParameterStatusMessage{
+            identifier: _,
+            length: _,
+            param,
+        }) => {
+            js.set_string_from_bytes("parameter_name", &param.param_name)?;
+            js.set_string_from_bytes("parameter_value", &param.param_value)?;
```
jasonish, line 221: As already mentioned, simply `name` and `value`?

rust/src/pgsql/log.rs
```diff
@@ -0,0 +1,420 @@
...
+        PgsqlBEMessage::CommandComplete(
+            RegularPacket{
+                identifier: _,
+                length: _,
+                payload,
+            }) =>
+        {
+            // TODO this may result in not so pretty strings. not sure what to do
+            //  example "SELECT 3\u0000"
+            js.set_string_from_bytes("payload", payload)?;
```
jasonish, line 241: Right.  Postgres was clearly written by C developers sending NUL terminated strings over the wire.  I wonder if that should be prettied up here? Anyways, minor nit for now.

jufajardini:
> ```json
> {
>   "timestamp": "2020-12-30T15:36:39.846111+0000",
>   "flow_id": 131952644348043,
>   "pcap_cnt": 41,
>   "event_type": "pgsql",
>   "src_ip": "10.16.1.10",
>   "src_port": 40816,
>   "dest_ip": "10.16.1.11",
>   "dest_port": 5432,
>   "proto": "TCP",
>   "pgsql": {
>     "transaction_id": 6,
>     "response": [
>       {
>         "ready_for_query": {
>           "transaction_status": "I"
>         }
>       }
>     ]
>   }
> }
> ```
> 
> I think this ready for query belongs in the previous transaction, at least when the request is a command/query string. From what I can tell from the protocol documentation, you could get multiple `CommandComplete`, but you'll only get a single `ReadyForQuery` which should be at the very end, at which point the server starts waiting.

I believe you are right. This is more likely the case that the transaction id tracking/ progressing is still a bit wonky, but the idea is to have transactions like this, for `Simple Query` (sorry for the huge img):
 
SimpleQuery

jasonish (changes requested):
Overall I love the verbosity but fear it might be too much, but also don't want to go crazy with configuration options, but I will continue thinking about that.

I noticed that Zeek does not log full responses, but probably as its just too verbose, but logs response success/failure (for MySQL) anyways. So that might be one option to consider, where response logging is off by default, but can be enabled.

jufajardini:
> With respect to logging... I find the multiple requests and responses in one of the samples a little confusing:
> 
> <-- suppressed for readability --> 
> 
> I think the above is 2 distinct transactions and should be logged as such. One request, one response per transaction. Might make state handling simpler, as well as querying the database and the other end of this. Or I guess there could be cases where one request results in multiple responses. But I don't think multiple requests should get bundled into the same TX.

I see. From the Postgresql documentation, I had understood the StartupPhase message flow as shown below:


Would breaking the transaction before `PasswordMessage` be a better approach, then?

jufajardini:
> Overall I love the verbosity but fear it might be too much, but also don't want to go crazy with configuration options, but I will continue thinking about that.
> 
> I noticed that Zeek does not log full responses, but probably as its just too verbose, but logs response success/failure (for MySQL) anyways. So that might be one option to consider, where response logging is off by default, but can be enabled.

Many thanks for the comprehensive review! I'll give it a try with a more zeek-like approach for the responses.

jasonish:
> > With respect to logging... I find the multiple requests and responses in one of the samples a little confusing:
> > <-- suppressed for readability -->
> > I think the above is 2 distinct transactions and should be logged as such. One request, one response per transaction. Might make state handling simpler, as well as querying the database and the other end of this. Or I guess there could be cases where one request results in multiple responses. But I don't think multiple requests should get bundled into the same TX.
> 
> I see. From the Postgresql documentation, I had understood the StartupPhase message flow as shown below:
> 
> 
> Would breaking the transaction before `PasswordMessage` be a better approach, then?
> 
> 

Yes, your suggested one is better.  I see it as 2 distinct transactions.  

1) C: Hello, I'm Joe S: But are you really Joe
2) C: I'm really joe, here's proof.  S: Hi Joe, I'm server and here are my details (OR: you're not joe, go away)

suricata-qa:
ERROR: QA failed on generic_collect.

Pipeline 3639

jufajardini:
Tried to come up with an "RFC" for a less verbose logger for the first merge...
`AuthenticationOk` and server details transaction would look like this:
image
`SimpleQuery` transaction would look like this:
image

May I go for something like this, or even less, for now?

jasonish:
> May I go for something like this, or even less, for now?

I'm unsure about this for now.  The response lists contains 2 data types, strings and objects.  While allowed by JSON, we know this to trip up various SIEMs. So in that respect I think your first attempt was better.  I'd probably focus on the multiple requests in a tx, and just some naming for now.  Then take another look at what we can remove in a non-extended logging format.

jufajardini:
> > May I go for something like this, or even less, for now?
> 
> I'm unsure about this for now. The response lists contains 2 data types, strings and objects. While allowed by JSON, we know this to trip up various SIEMs. So in that respect I think your first attempt was better. I'd probably focus on the multiple requests in a tx, and just some naming for now. Then take another look at what we can remove in a non-extended logging format.

Roger that, thanks! :)

jufajardini:
Feedbacks incorporated in: https://github.com/OISF/suricata/pull/6376

PR #6285, opened 2021-08-06
Unify reg nfs+smb/v2

inashivb:


rust/src/smb/smb.rs
```diff
@@ -2189,3 +2190,143 @@ pub extern "C" fn rs_smb_state_get_event_info(event_name: *const std::os::raw::c
...
+        SCLogDebug!("Rust SMB parser registered.");
+        let retval = conf_get("app-layer.protocols.smb.stream-depth");
+        if let Some(val) = retval {
+            let val = val.parse::<i32>().unwrap();
+            if val < 0 {
+                SCLogError!("invalid value for stream-depth");
+            } else {
+                stream_depth = val as u32;
+           }
+            AppLayerParserSetStreamDepth(IPPROTO_TCP as u8, ALPROTO_SMB, stream_depth);
```
catenacyber, line 2327: @inashivb This is now called only if we have `app-layer.protocols.smb.stream-depth` in the config, which was not the case before 

Suggested fix is 
```
diff --git a/rust/src/smb/smb.rs b/rust/src/smb/smb.rs
index 054e65ab8..482ae8a78 100644
--- a/rust/src/smb/smb.rs
+++ b/rust/src/smb/smb.rs
@@ -2288,8 +2288,8 @@ pub unsafe extern "C" fn rs_smb_register_parser() {
             } else {
                 stream_depth = val as u32;
            }
-            AppLayerParserSetStreamDepth(IPPROTO_TCP as u8, ALPROTO_SMB, stream_depth);
         }
+        AppLayerParserSetStreamDepth(IPPROTO_TCP as u8, ALPROTO_SMB, stream_depth);
     } else {
         SCLogDebug!("Protocol detector and parser disabled for SMB.");
     }
```

inashivb: Waiting on [this](https://github.com/OISF/suricata/pull/6364) to get merged/approved then do this fix. Thanks, @catenacyber !

catenacyber: Ok, letting you handle this...

PR #6289, opened 2021-08-10
Smtp refactoring/v1

inashivb:
A small approachable demo of conversion of SMTP to Rust while keeping File extraction and MIME in C as it is.

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -482,18 +482,6 @@ static int SMTPCleanCloseFile(const uint8_t *chunk, uint32_t len,
     return ret;
 }
 
-static void SMTPSetMinInspectDepth(uint32_t ctnt_min_size,
-        uint64_t ts_data_cnt, uint64_t ts_last_ds,
-        Flow *flow, uint16_t dir, bool trigger_reassembly)
-{
-    uint32_t depth = ctnt_min_size + ts_data_cnt - ts_last_ds;
-    if (trigger_reassembly) {
-        AppLayerParserTriggerRawStreamReassembly(flow, dir);
-    }
-    SCLogDebug("StreamTcpReassemblySetMinInspectDepth STREAM_TOSERVER %u", depth);
-    StreamTcpReassemblySetMinInspectDepth(flow->protoctx, dir, depth);
-}
-
 int SMTPProcessDataChunk(const uint8_t *chunk, uint32_t len,
         MimeDecParseState *state)
 {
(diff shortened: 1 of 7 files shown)


suricata-qa:

Pipeline 3745

src/app-layer-smtp.c
```diff
@@ -405,6 +405,83 @@ static void SMTPNewFile(SMTPTransaction *tx, File *file)
...
 }
 
+static int SMTPFileContainerAlloc(SMTPState *smtp_state)
+{
+    /* Make sure file container allocated */
+    if (smtp_state->files_ts == NULL) {
+        smtp_state->files_ts = FileContainerAlloc();
+        if (smtp_state->files_ts == NULL) {
+            SCLogError(SC_ERR_MEM_ALLOC, "Could not create file container");
+            return MIME_DEC_ERR_MEM;
```
victorjulien, line 415: I further step of separating the various api's would to not use `MIME_DEC...` for returns here.

src/app-layer-smtp.c
```diff
@@ -421,131 +498,52 @@ int SMTPProcessDataChunk(const uint8_t *chunk, uint32_t len,
...
-        /* Make sure file container allocated */
-        if (smtp_state->files_ts == NULL) {
-            smtp_state->files_ts = FileContainerAlloc();
-            if (smtp_state->files_ts == NULL) {
-                ret = MIME_DEC_ERR_MEM;
-                SCLogError(SC_ERR_MEM_ALLOC, "Could not create file container");
-                SCReturnInt(ret);
-            }
+        if (SMTPFileContainerAlloc(smtp_state) == MIME_DEC_ERR_MEM) {
+            SCReturnInt(-2);
```
victorjulien, line 502: this is a bit odd. We use `MIME_DEC_ERR_MEM` in our util func, but do not use `MIME_DEC_...` to return from a mime api callback?

src/app-layer-smtp.c
```diff
@@ -421,131 +498,52 @@ int SMTPProcessDataChunk(const uint8_t *chunk, uint32_t len,
...
-
-            if (FileOpenFileWithId(files, &smtp_config.sbcfg, smtp_state->file_track_id++,
-                        (uint8_t *) entity->filename, entity->filename_len,
-                        (uint8_t *) chunk, len, flags) != 0) {
-                ret = MIME_DEC_ERR_DATA;
-                SCLogDebug("FileOpenFile() failed");
-            }
-            SMTPNewFile(smtp_state->curr_tx, files->tail);
-
+            set_min_inspect_depth(flow, smtp_config.content_inspect_min_size,
```
victorjulien, line 513: nit: function name style

victorjulien:
I think this makes sense. It helps separate the various layers. I find the example of Rust code not a very convincing one yet, as it seems to be just C logic :) But I think this is the way forward.

inashivb:
Moving forward the work with this while keeping comments in mind. Closing the PR

PR #6295, opened 2021-08-13
Feature/http range/v16

victorjulien:
#6293 with a few changes:
- renaming of some parts of the API
- malloc failure error handling

suricata-verify-pr: 517

PR #6297, opened 2021-08-17
Feature/http range/v20

victorjulien:
#6295, changing:

fix memory leaks
fix reference counting
misc improvements

suricata-verify-pr: 517


victorjulien:
Replaced by #6298

PR #6298, opened 2021-08-17
Feature/http range/v21

victorjulien:
#6295, changing:

fix memory leaks
fix reference counting
misc improvements
fixes formatting in #6297

suricata-verify-pr: 517


suricata-qa:

Pipeline 3817

src/detect-engine.c
```diff
@@ -1078,7 +1078,9 @@ void InspectionBufferInit(InspectionBuffer *buffer, uint32_t initial_size)
 void InspectionBufferSetupMulti(InspectionBuffer *buffer, const DetectEngineTransforms *transforms,
         const uint8_t *data, const uint32_t data_len)
 {
+#ifdef DEBUG_VALIDATION
```
catenacyber, line 1081: How is this related to HTTP byte ranges ?

src/util-thash.c
```diff
@@ -186,6 +187,8 @@ static THashData *THashDataAlloc(THashTableContext *ctx)
 static void THashDataFree(THashTableContext *ctx, THashData *h)
 {
     if (h != NULL) {
+        DEBUG_VALIDATE_BUG_ON(SC_ATOMIC_GET(h->use_cnt) != 0);
```
catenacyber, line 190: Nice :-)
Would have helped me when I started

src/app-layer-htp-file.c
```diff
@@ -332,17 +332,53 @@ int HTPFileOpenWithRange(HtpState *s, const uint8_t *filename, uint16_t filename
...
+    /* we're done, so unlock. But since we have a reference in s->file_range keep use_cnt. */
+    THashDataUnlock(file_range_container->hdata);
     SCReturnInt(0);
 }
 
+static void HTPFileStoreChunkHandleRange(
+        HttpRangeContainerBlock *c, const uint8_t *data, uint32_t data_len)
+{
+    if (c->container) {
+        THashDataLock(c->container->hdata);
```
catenacyber, line 368: Why would we need this lock ?

The point of my architecture was to have the concurrency only in open and close, so that append is simply protected by the lock of its flow.
To say it another way, we should be able to append data to 2 different buffers at the same time.
Am I missing something ?

catenacyber (commented):
Many great improvements.

My big question is about the lock to append data...

src/app-layer-htp-range.h
```diff
@@ -55,18 +55,18 @@ typedef struct HttpRangeContainerFile {
...
     /** total epxected size of the file in ranges */
     uint64_t totalsize;
-    /** file flags */
-    uint16_t flags;
     /** file container, with only one file */
     FileContainer *files;
     /** red and black tree list of ranges which came out of order */
     struct HTTP_RANGES fragment_tree;
+    /** file flags */
+    uint16_t flags;
```
catenacyber, line 69: Is there a tool for this that we can use in CI ?

src/app-layer-htp-file.c
```diff
@@ -397,6 +432,38 @@ int HTPFileStoreChunk(HtpState *s, const uint8_t *data, uint32_t data_len,
...
+        }
+        SCLogDebug("c->container->files->tail %p", c->container->files->tail);
+        THashDataUnlock(c->container->hdata);
+    } else {
+        if (c->toskip > 0) {
+            if (HttpRangeProcessSkip(c, data, data_len) < 0) {
+                SCLogDebug("Failed to append data");
+            }
+        }
+        File *ranged = HttpRangeClose(c, flags);
```
catenacyber, line 459: Is there anything to close if `c->container == NULL` ?

src/app-layer-htp-range.c
```diff
@@ -66,23 +68,33 @@ static int ContainerUrlRangeSet(void *dst, void *src)
     HttpRangeContainerFile *dst_s = dst;
     dst_s->len = src_s->len;
     dst_s->key = SCMalloc(dst_s->len);
-    BUG_ON(dst_s->key == NULL);
+    if (dst_s->key == NULL)
```
catenacyber, line 71: This was some copied/pasted code, so I guess the original can have the same improvements...

src/app-layer-htp-range.h
```diff
@@ -69,8 +69,8 @@ typedef struct HttpRangeContainerFile {
     uint16_t flags;
     /** wether a range file is currently appending */
     bool appending;
-    /** mutex */
-    SCMutex mutex;
+    /** error condition for this range. Its up to timeout handling to cleanup */
+    bool error;
```
catenacyber, line 73: This is nice.
Should we update `expire` so that they get cleant faster ?

src/app-layer-htp.c
```diff
@@ -374,6 +374,8 @@ void HTPStateFree(void *state)
         htp_connp_destroy_all(s->connp);
     }
 
+    HttpRangeFreeBlock(s->file_range);
```
catenacyber, line 377: Should we not try to keep this buffer even if it was incomplete ?
That means calling `HTPFileCloseHandleRange`

src/app-layer-htp-range.c
```diff
@@ -444,12 +491,36 @@ File *ContainerUrlRangeClose(HttpRangeContainerBlock *c, uint16_t flags)
...
+    SCLogDebug("returning f %p (c:%p container:%p)", f, c, c->container);
     return f;
 }
+
+static void HttpRangeBlockDerefContainer(HttpRangeContainerBlock *b)
+{
+    if (b && b->container) {
+        DEBUG_VALIDATE_BUG_ON(SC_ATOMIC_GET(b->container->hdata->use_cnt) == 0);
+        THashDecrUsecnt(b->container->hdata);
+        DEBUG_VALIDATE_BUG_ON(SC_ATOMIC_GET(b->container->hdata->use_cnt) > (uint32_t)INT_MAX);
```
catenacyber, line 509: Is this debug validation useful ? (the one just above looks clearly useful to me)

src/app-layer-htp-range.c
```diff
@@ -221,110 +243,139 @@ void *HttpRangeContainerUrlGet(const uint8_t *key, size_t keylen, struct timeval
...
+    }
+
+    (void)SC_ATOMIC_ADD(ContainerUrlRangeList.ht->memuse, buflen);
     range->buffer = SCMalloc(buflen);
-    BUG_ON(range->buffer == NULL);
+    if (range->buffer == NULL) {
+        c->error = true;
+        SCFree(curf);
+        SCFree(range);
+        (void)SC_ATOMIC_SUB(ContainerUrlRangeList.ht->memuse, buflen);
```
catenacyber, line 313: We could avoid this by having `SC_ATOMIC_ADD` a bit later

src/app-layer-htp-range.h
victorjulien: Yeah that is a good idea.

src/app-layer-htp.c
victorjulien: `s->file_range` contains the flow specific data, so it needs to be freed with the state. Normally if we'd get here we would have called the "close" logic before it, so that would be responsible for handling any data that is still useful somehow.

src/app-layer-htp-range.c
victorjulien: I added it when debugging an issue where I decremented more than incremented. I agree it seems redundant.

src/app-layer-htp-range.c
victorjulien: Yeah, but the idea of the memcap updates it that we keep it as close to the actual allocation as possible to keep the accounting as accurate as we can. The error path should be very rare anyway.

src/app-layer-htp-range.h
victorjulien: Actually I think its cleaner to add a check for `cu->error` in `ContainerValueRangeTimeout`. Trying that.

src/app-layer-htp-range.c
victorjulien: Agreed. Can you do that or at least create a ticket for it to track it?

src/app-layer-htp-file.c
victorjulien: Good catch. I had been meaning to revisit, but forgot. No it looks like this can go.

src/app-layer-htp-range.h
victorjulien: There is `pahole` that I use, at least for more complex layouts. Its a bit fragile though. Not sure if using it in CI would work.

src/app-layer-htp-file.c
victorjulien: The idea is that any access to the shared data, so `HttpRangeContainerFile` and everything inside it, needs to happen under lock as we can have multiple flows trying to work with it at the same time.

When we create a new range the hash table returns a locked `HttpRangeContainerFile` to us. For any updates following that we need to lock it on append and close directly.

src/detect-engine.c
victorjulien: Its not, but i noticed it in my QA. Its in its own commit for that reason

victorjulien:
Replaced by #6308

src/app-layer-htp-range.c
catenacyber: Looks like there is only one occurence for `git grep "BUG_ON(dst_s"` in datasets-string.c `StringSet`

Should I create add a commit on top of #6308 and create a new PR ?

src/app-layer-htp-file.c
catenacyber: I will read this again in the morning right after ☕

catenacyber: Maybe we need a live talk about it.

> so HttpRangeContainerFile and everything inside it, needs to happen under lock as we can have multiple flows trying to work with it at the same time.

I totally agree.

But `HTPFileStoreChunkHandleRange ` is about `HttpRangeContainerBlock`, not `HttpRangeContainerFile` which I think does not need a lock, as it is accessed by only one flow.
`HttpRangeContainerBlock` just checks its `container` filled, as you are doing here without a lock, and only one of them is allowed to `FileAppendData(c->container->files, data, len)` cf `appending` set in `HttpRangeOpenFileAux
It is not at risk by the timeout since timeout will occur only if the use count is zero (no more flows have a reference to the`HttpRangeContainerFile`)

PR #6300, opened 2021-08-18
[draft] rust: apply rust format - v1

jasonish:
Based on PR: https://github.com/OISF/suricata/pull/6299

- Applies the fixes that clippy can do automatically.
- Run rustfmt on the Rust codebase.

This PR is just to get a look at what the code might look like.  There might various formatting options we want to tweak. But in general I think Rust developers are starting to expect to be able to use the Rust formatting tools as part of their workflow and it would be nice to be able to do on our code as well.


rust/src/asn1/parse_rules.rs
```diff
@@ -74,7 +74,7 @@ pub unsafe extern "C" fn rs_detect_asn1_parse(input: *const c_char) -> *mut Dete
 ///
 /// ptr must be a valid object obtained using `rs_detect_asn1_parse`
 #[no_mangle]
-pub unsafe extern "C" fn rs_detect_asn1_free(ptr: *mut DetectAsn1Data) {
```
victorjulien, line 77: is this a formatting change?

rust/src/applayertemplate/template.rs
```diff
@@ -578,20 +528,55 @@ mod test {
         let buf = b"5:Hello3:bye";
 
         let r = state.parse_request(&buf[0..0]);
-        assert_eq!(r, AppLayerResult{ status: 0, consumed: 0, needed: 0});
+        assert_eq!(
```
victorjulien, line 531: can this be avoided? Its so unnecessarily verbose

rust/src/conf.rs
```diff
@@ -113,9 +100,7 @@ impl ConfNode {
 
         unsafe {
             let s = CString::new(key).unwrap();
-            if ConfGetChildValueBool(self.conf,
-                                     s.as_ptr(),
-                                     &mut vptr) != 1 {
+            if ConfGetChildValueBool(self.conf, s.as_ptr(), &mut vptr) != 1 {
```
victorjulien, line 103: I don't get why some things like the test stuff above are expanded into many line bloated things, while this is made more compact. I like the more compact stuff like this

rust/src/dcerpc/dcerpc_udp.rs
```diff
@@ -73,22 +73,29 @@ impl DCERPCUDPState {
...
                 found = true;
                 index = i;
                 SCLogDebug!("tx {} progress {}/{}", tx.id, tx.req_done, tx.resp_done);
                 break;
             }
         }
         if found {
-            SCLogDebug!("freeing TX with ID {} TX.ID {} at index {} left: {} max id: {}",
-                            tx_id, tx_id+1, index, self.transactions.len(), self.tx_id);
+            SCLogDebug!(
```
victorjulien, line 91: this way having `SCLogDebug!` statements becomes really disruptive wrt following the code flow. These need to be compact.

rust/src/dcerpc/mod.rs
```diff
@@ -17,6 +17,6 @@
 
 pub mod dcerpc;
 pub mod dcerpc_udp;
-pub mod parser;
 pub mod detect;
 pub mod log;
+pub mod parser;
```
victorjulien, line 22: are these rustfmt changes?

suricata-qa:

Pipeline 3843

rust/src/dcerpc/mod.rs
jasonish: Yes, by default it will re-order imports alphabetically: https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=imports#reorder_imports

rust/src/asn1/parse_rules.rs
jasonish: Yes.  By default rustfmt will convert all `extern {` (without a language) to `extern "C"` just to be explicit by default.  This causes issues in an extern block that creates bindings to C functions, for example:
```
extern "C" {
    pub fn AppLayerRegisterProtocolDetection(
        parser: *const RustParser, enable_default: c_int,
    ) -> AppProto;
    pub fn AppLayerRegisterParserAlias(parser_name: *const c_char, alias_name: *const c_char);
}
```
When the `"C"` is there, `cbindgen` will go and re-export these methods in the generated header which our C code doesn't like.  And `cbindgen` has no option to turn them off.  This does not happen when the "C" is left out, and "C" is the default when left out.

This is why I had to add the following to `rustfmt.toml`:
```
force_explicit_abi = false
```

The other options is to explicitly ignore every function we don't want re-exported in the cbindgen configuration, but thats a little tedious.

rust/src/applayertemplate/template.rs
jasonish: This appears to be dependent on the size of the struct literal.  With Rust stable we can use `struct_lit_width = 36` (the default is 18 to get these particular lines to format like:
```
assert_eq!(r, AppLayerResult { status: 0, consumed: 0, needed: 0 });
```
bigger ones will continue to wrap like the above.

This works in current Rust stable, but not as far back as 1.41.1.

rust/src/conf.rs
jasonish: Struct literals and function arguments are treated differently.

rust/src/dcerpc/dcerpc_udp.rs
jasonish: We can prevent the SCLog* macros from any formatting at all.  But otherwise there are no options to fine tune this.

rust/src/applayertemplate/template.rs
jasonish: This option requires Rust 1.53.0 or newer (2021-06-17).  But we can instruct rustfmt to skip assert_eq macros.

rust/src/dcerpc/dcerpc_udp.rs
jasonish: Unfortunately the feature to disable formatting of macro calls by name is only available in nightly rust.  The nom macros get messed up as well, however we can put an ignore specifically before them, but thats a little tedious before our logging macros.

This behaviour does appear to be idiomatic for lines this long.. 5 of 5 random Rust projects I checked are `rustfmt --check` clean and do this.

jasonish: Another work-around to prevent formatting inside a macro..
```
SCLogDebug{"freeing TX with ID...
```

victorjulien: I think having verbose debug statements would be an anti-pattern. They'd hinder the ease of following the program logic.

rust/src/asn1/parse_rules.rs
victorjulien: Not sure I fully get it. Did rustfmt make these changes despite `force_explicit_abi = false` or are you making that change in a follow up PR?

victorjulien: (in general a reformatting tool doing things that have effects beyond formatting is ... ugh. It matters because it means I now distrust the tool in terms of not producing side effects)

jasonish: Rust format made these changes.  I've instructed it to just use `extern {` and not `extern "C" {`. This is side affect free to Rust and purely a matter of style.  An `extern "WASM"` would not be changed.

This is just a matter of consistency.  I'd rather be consistent on `extern "C" {`, but cbindgen is broken a bit here, which is another reason to be consistent about it as well.

This is a change I'd like to make even without going ahead with rustfmt.

rust/src/dcerpc/dcerpc_udp.rs
jasonish: @victorjulien It looks like our clang format has the same issue: https://github.com/OISF/suricata/blob/master/src/detect-file-data.c#L512

PR #4478, opened 2020-01-09
Napatech bypass v1.09

pjyoung1:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to redmine ticket:
[https://redmine.openinfosecfoundation.org/issues/2818](https://redmine.openinfosecfoundation.org/issues/2818)

Describe changes:
```
Added hardware based bypass support on Napatech SmartNICs
Added support for inline processing using Napatech SmartNICs
```
This PR Addresses issues identified in the review of [PR 4472 ](https://github.com/OISF/suricata/pull/4472)
This reflects changes since PR  [Napatech bypass v1.07]https://github.com/OISF/suricata/pull/4472)

PR #6301, opened 2021-08-19
threading: don't pass locked flow between threads

victorjulien:
Previously the flow manager would share evicted flows with the workers
while keeping the flows mutex locked. This reduced the number of unlock/
lock cycles while there was guaranteed to be no contention.

This turns out to be undefined behavior. A lock is supposed to be locked
and unlocked from the same thread. It appears that FreeBSD is stricter on
this than Linux.

This patch addresses the issue by unlocking before handing a flow off
to another thread, and locking again from the new thread.

Issue was reported and largely analyzed by Bill Meeks.

Bug: #4478

https://redmine.openinfosecfoundation.org/issues/4478

### DIFF
--- a/src/flow-hash.c
+++ b/src/flow-hash.c
@@ -692,6 +692,7 @@ static inline void MoveToWorkQueue(ThreadVars *tv, FlowLookupStruct *fls,
         f->fb = NULL;
         f->next = NULL;
         FlowQueuePrivateAppendFlow(&fls->work_queue, f);
+        FLOWLOCK_UNLOCK(f);
     } else {
         /* implied: TCP but our thread does not own it. So set it
          * aside for the Flow Manager to pick it up. */
--- a/src/flow-manager.c
+++ b/src/flow-manager.c
@@ -333,6 +333,7 @@ static uint32_t ProcessAsideQueue(FlowManagerTimeoutThread *td, FlowTimeoutCount
                 FlowForceReassemblyNeedReassembly(f) == 1)
         {
             FlowForceReassemblyForFlow(f);
+            FLOWLOCK_UNLOCK(f);
             /* flow ownership is passed to the worker thread */
 
             /* flow remains locked */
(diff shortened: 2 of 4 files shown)


src/flow-manager.c
```diff
@@ -333,6 +333,7 @@ static uint32_t ProcessAsideQueue(FlowManagerTimeoutThread *td, FlowTimeoutCount
                 FlowForceReassemblyNeedReassembly(f) == 1)
         {
             FlowForceReassemblyForFlow(f);
+            FLOWLOCK_UNLOCK(f);
             /* flow ownership is passed to the worker thread */
 
             /* flow remains locked */
```
regit, line 339: Comment is not valid anymore.

src/flow-timeout.c
```diff
@@ -401,6 +401,7 @@ static inline void FlowForceReassemblyForHash(void)
                 RemoveFromHash(f, prev_f);
                 f->flow_end_flags |= FLOW_END_FLAG_SHUTDOWN;
                 FlowForceReassemblyForFlow(f);
```
regit, line 403: Should we not put the unlock in the FlowForceReassemblyForFlow ? As it is only queuing the flow in another thread this should be systematic.

src/flow-hash.c
```diff
@@ -692,6 +692,7 @@ static inline void MoveToWorkQueue(ThreadVars *tv, FlowLookupStruct *fls,
         f->fb = NULL;
         f->next = NULL;
         FlowQueuePrivateAppendFlow(&fls->work_queue, f);
+        FLOWLOCK_UNLOCK(f);
```
regit, line 695: This looks ok but got me confused at first. FlowQueuePrivateAppendFlow is used without unlocking on some other part of the code. Only work_queue has the lock at start because of injection from external threads. Should we use a dedicated function for work_queue to avoid confusion ?

regit (commented):
Some minor comments. It looks ok to me.

suricata-qa:
WARNING:
|--|--|--|--|
| | tlpr1_alerts_cmp |


Pipeline 3851

src/flow-timeout.c
victorjulien: `FlowForceReassemblyForFlow` accesses a field in the flow, so technically it should do so under lock. I don't think it would matter much either way in this case.

src/flow-hash.c
victorjulien: `FlowQueue` API uses its own lock logic independent of the `Flow` objects with passes around.

victorjulien:
Merged in #6306, with comment fixup.

PR #6302, opened 2021-08-20
Enum dir rust/v3

inashivb:

### DIFF
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -39,6 +39,35 @@ pub const STREAM_GAP:      u8 = 0x10;
 pub const STREAM_DEPTH:    u8 = 0x20;
 pub const STREAM_MIDSTREAM:u8 = 0x40;
 
+#[repr(C)]
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum Direction {
+    ToServer = 0x04,
+    ToClient = 0x08,
+}
+
+impl Default for Direction {
+    fn default() -> Self { Direction::ToServer }
+}
+
+impl From<u8> for Direction {
+    fn from(d: u8) -> Self {
+        match d {
+            0x04 => { Direction::ToServer },
+            0x08 => { Direction::ToClient },
+            _ => {
+                if d & 0b0000_0100 != 0 {
+                    Direction::ToServer
+                } else if d & 0b0000_1000 != 0 {
+                    Direction::ToClient
+                } else {
+                    panic!("Unknown direction!!");
+                }
+            }
+        }
+    }
+}
+
 // Application layer protocol identifiers (app-layer-protos.h)
 pub type AppProto = u16;
 
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -321,8 +321,8 @@ pub struct DCERPCState {
     pub bytes_consumed: u16,
     pub tx_id: u64,
     pub query_completed: bool,
-    pub data_needed_for_dir: u8,
-    pub prev_dir: u8,
+    pub data_needed_for_dir: Direction,
+    pub prev_dir: Direction,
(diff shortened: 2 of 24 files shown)


rust/src/core.rs
```diff
@@ -39,6 +39,35 @@ pub const STREAM_GAP:      u8 = 0x10;
...
+#[repr(C)]
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum Direction {
+    ToServer = 0x04,
+    ToClient = 0x08,
+}
+
+impl Default for Direction {
+    fn default() -> Self { Direction::ToServer }
+}
```
jasonish, line 51: I'm a little worried about using a default that has meaning.  Would adding Unknown to the enum as 0 make sense and using that as the default?

rust/src/core.rs
```diff
@@ -39,6 +39,35 @@ pub const STREAM_GAP:      u8 = 0x10;
...
+impl From<u8> for Direction {
+    fn from(d: u8) -> Self {
+        match d {
+            0x04 => { Direction::ToServer },
+            0x08 => { Direction::ToClient },
+            _ => {
+                if d & 0b0000_0100 != 0 {
+                    Direction::ToServer
+                } else if d & 0b0000_1000 != 0 {
+                    Direction::ToClient
```
jasonish, line 62: Might be cleaner to just use the bit fields, and define them as constants somewhere.

rust/src/core.rs
```diff
@@ -39,6 +39,35 @@ pub const STREAM_GAP:      u8 = 0x10;
...
+                } else {
+                    panic!("Unknown direction!!");
```
jasonish, line 64: Convert to a TryFrom.  The caller can then decide to panic?  I'm not sure how intrusive this would be, but I don't think an impl from From should panic.

jasonish (changes requested):
The users of the new enum look OK.

rust/src/core.rs
jasonish: I guess a lot of our code just checks for one direction, the blindly assumes the else branch is the other direction. Any thoughts here @victorjulien ?

suricata-qa:
WARNING:
|--|--|--|--|
| | tlpr1_alerts_cmp |


Pipeline 3872

rust/src/core.rs
victorjulien: I think this could be done using `debug_validate_bug_on`. We want to catch improper uses of the direction, but I don't think we need to do dynamic error checking on every path

catenacyber:
Looks great to me...
The SSH commit should be taken back in as it fixes https://redmine.openinfosecfoundation.org/issues/4563
It needs to fix a Suricata-verify test 
```
diff --git a/tests/ssh-banner-only/test.yaml b/tests/ssh-banner-only/test.yaml
index 5c4b908..c138be0 100644
--- a/tests/ssh-banner-only/test.yaml
+++ b/tests/ssh-banner-only/test.yaml
@@ -20,4 +20,4 @@ checks:
       count: 1
       match:
         event_type: alert
-        alert.signature_id: 1
+        alert.signature_id: 2
```
As they are indeed 2 alerts : one from the client and one from the server

rust/src/core.rs
victorjulien: There should be no other direction than these 2, so any other setting is a bug. Would hate to have this dynamically checked all the time as this *should* be unreachable and thus untested code paths.

rust/src/core.rs
victorjulien: > @victorjulien If debug validation is not enabled would it be OK to default to `ToServer`? Otherwise there is probably no other option but to panic. If we did use a `Result` for the return type, what would be caller due on err? Probably panic.

This is the solution I'd want. In `debug_validation` mode panic, otherwise default to toserver

inashivb:
Replaced w #6343

PR #6296, opened 2021-08-15
[DRAFT] netmap: Netmap API 14 support

jlucovsky:
Continuation of #6288 

This commit provides support for Netmap API version 14+.

We thank @bmeeks8 for his contributions -- much of the content in this
commit is his.

This commit is in *DRAFT* mode and should not be used in production
systems.

The most notable change is that multiple threads are created it threads
is configured as "auto".

Multiple thread support is working for
- IPS with Host method, e.g., interface-name and interface-name^
- IDS mode

Multiple thread support is not yet working for other scenarios.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #6305, opened 2021-08-23
[Draft] netmap: Netmap API 14 support

jlucovsky:
Continuation of #6296 

This commit provides support for Netmap API version 14+.

We thank @bmeeks8 for his contributions -- much of the content in this
pull request are the result of his efforts.

This commit is in *DRAFT* mode and should not be used in production
systems.

Updates
- Multiple threads will be created to process packets iff the network interface supports multiple queues
- And, multiple threads have been configured, e.g., using `threads: auto` in the `netmap` section.
- Netmap V14 API used if available *and* libnetmap.h exists; it's required for multi-thread support.

This has been tested with runmodes `autofp` and `workers`.

Multiple thread support is working for
- IPS with Host method, e.g., interface-name and interface-name^
- IDS mode

In IPS mode and when using physical interfaces, `threads: 1` required as packets are not yet fully distributed to multiple
queues.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1580,7 +1580,17 @@
         #endif
         ])], [have_gtv13_netmap="yes"])
         if test "x$have_gtv13_netmap" = "xyes"; then
-            have_netmap_version="> v13"
+            have_netmap_version="v14+"
+            AC_CHECK_HEADER(libnetmap.h,,[AC_MSG_ERROR(libnetmap.h not found ...)],)
+            LIBNETMAP=""
+            AC_SEARCH_LIBS([nmport_open],[netmap],,[LIBNETMAP="no"])
+            if test "$LIBNETMAP" = "no"; then
+                echo
+                echo "   ERROR!  libnetmap library not found!"
+                echo
+                exit 1
+            fi
+            AC_DEFINE([HAVE_NETMAP_V14],[1],(NETMAP API v14 support enabled))
         fi
   ])
 
--- a/src/runmode-netmap.c
+++ b/src/runmode-netmap.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2014-2018 Open Information Security Foundation
+/* Copyright (C) 2014-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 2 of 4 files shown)


suricata-qa:
WARNING:
|--|--|--|--|
| | tlpr1_alerts_cmp |


Pipeline 3888

victorjulien:
So it seems FreeBSD 11 is almost EOL, leaving us with 12+. My fully updated 12.2 VM has Netmap API 14. Should we just make API 14 the only supported version?

bmeeks8:
There may be another more problematic issue that needs addressing. After some research, it appears the libnetmap wrapper library is present only in the FreeBSD 12.2-STABLE branch and up (including 13.0-RELEASE). That wrapper is not part of FreeBSD 12.2-RELEASE and lower. The inclusion of the Netmap v14 API is somewhat divorced from the libnetmap wrapper library.

So that means the _nmport_prepare()_, _nmport_open_desc()_,  _nmport_inject()_ and _nmport_close()_ functions used from that wrapper library are not available in the 12.x RELEASE versions of FreeBSD. You can still do multiple host rings, but you must call the lower-level netmap functions in order to do so. And the parsing of the device descriptor string (the "netmap:xxx" or "vale:xxx" device name with the optional arguments) would have to be done within Suricata code. Then the proper values and bit flags set in the _nm_req_ structure prior to calling the _ioctl()_ function to open the netmap device.

If Suricata restricts itself to using simple "netmap:" or "vale:" device types, creating the parser code would not be too difficult. There is a template to follow in the source code for the libnetmap library.

jlucovsky:
We discussed how to handle the "wrapper" library used by Suricata. There are 2 paths forward

For `master` (7.0), Netmap V14 with the wrapper library is required
For `6.0.x`, an additional configuration flag (name tbd) will be required to select Netmap V14 (the wrapper library will be required). Without the additional configuration flag, functionality from V14 will not be used.

bmeeks8:
> We discussed how to handle the "wrapper" library used by Suricata. There are 2 paths forward
> 
> For `master` (7.0), Netmap V14 with the wrapper library is required
> For `6.0.x`, an additional configuration flag (name tbd) will be required to select Netmap V14 (the wrapper library will be required). Without the additional configuration flag, functionality from V14 will not be used.

I think that is a good approach. Over time the wrapper library should become standard in the various OS releases supporting netmap.

PR #6316, opened 2021-08-30
detect: debug validation for list ids overflows

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4403

Describe changes:
- Adds debug validation for list id integer overflow

https://github.com/OISF/suricata/pull/5932/commits/975062cf401f79c00abf728d923c65aabd143af2#diff-99eda33658bd0778da7bf89acbb4e7bbdb9ce82b0ab93486e1643691925f4091L600 has changed `sm_list` from `int` to `int16_t`
But there are no checks against integer overflow on this shorter integer

### DIFF
--- a/src/detect-engine-mpm.c
+++ b/src/detect-engine-mpm.c
@@ -115,6 +115,7 @@ void DetectAppLayerMpmRegister2(const char *name,
     am->name = name;
     snprintf(am->pname, sizeof(am->pname), "%s", am->name);
     am->direction = direction;
+    DEBUG_VALIDATE_BUG_ON(sm_list < 0 || sm_list > INT16_MAX);
     am->sm_list = sm_list;
     am->sm_list_base = sm_list;
     am->priority = priority;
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 3960

catenacyber:
And after a long quadratic time, I get the debug assert to trigger...
Let's see if oss-fuzz is getting it the same way

victorjulien:
The id increases with the number of rules using unique combinations of transactions, I think. Is that what you're seeing?

victorjulien:
Merged in #6324, thanks!

catenacyber:
Let's now wait for https://github.com/google/oss-fuzz/pull/6362 to get oss-fuzz back to build latest suricata and see if this is the bug getting triggered

catenacyber:
> The id increases with the number of rules using unique combinations of transactions, I think. Is that what you're seeing?

The unique combination of _transforms_ ...
 That is indeed what I am seeing (with the script from the redline ticket to generate 65k rules with different combinations of transforms)
Maybe there is another way to increase this id...

PR #6337, opened 2021-09-03
Fixed threshold rule to match description

kjakan:
Describe changes:
- Fixed threshold.conf rule examples to match description, based on documentation (https://suricata.readthedocs.io/en/suricata-6.0.3/rules/thresholding.html)

### DIFF
--- a/threshold.config
+++ b/threshold.config
@@ -20,10 +20,10 @@
 # https://suricata.readthedocs.io/en/latest/configuration/global-thresholds.html#global-thresholds-vs-rule-thresholds
 
 # Limit to 10 alerts every 10 seconds for each source host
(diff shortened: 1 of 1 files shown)


catenacyber:
The doc change seems appropriate indeed cf https://suricata.readthedocs.io/en/suricata-6.0.0/rules/thresholding.html#type-threshold

jufajardini (approved):
This looks indeed good, as Philippe has mentioned. I've noticed that we don't have examples for the `threshold` type and that we have a sig_id 0, which isn't valid. So I think this can be approved, but we need a follow-up work to fix these aspects. :)

victorjulien:
Merged in #8861, thanks!

PR #6339, opened 2021-09-03
rules: fix flow drop for ip-only sig with noalert

victorjulien:
A simple drop rules like drop tcp any any -> any any (noalert; sid:1; rev:1;)
fails to drop packets beyond the first in each direction of the flow.

The cause is that the rule is just evaluated once per flow direction, by the
"IP-only" detection path. In this case the 'noalert' presence causes the drop
not to be applied to the flow.

Bug: #4663

suricata-verify-pr: 528

### DIFF
--- a/src/detect-engine-iponly.c
+++ b/src/detect-engine-iponly.c
@@ -1115,14 +1115,15 @@ void IPOnlyMatchPacket(ThreadVars *tv,
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


suricata-qa:

Pipeline 4048

victorjulien:
Replaced by #6368

PR #6342, opened 2021-09-05
base64: use Rust base64 encoder - v2

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/6227

Changes from last PR:
- Replace our C Base64Encode with a Rust wrapper.

Add methods jb_set_base64 and jb_append_base64 to JsonBuilder to set/append byte array values as base64. This avoids the intermediate buffer that we currently use, and the Rust base64 encode we use can encode directly to the underlying JsonBuilder buffer (internally it appears to use a 1024 byte intermediate buffer that is flushed to the output string when full).

Finally convert all occurrences of Base64Encode directly before using it in JsonBuilder with these new methods.


suricata-qa:

Pipeline 4051

src/log-tlsstore.c
```diff
@@ -123,7 +122,7 @@ static void LogTlsLogPem(LogTlsStoreLogThread *aft, const Packet *p, SSLState *s
     }
 
     TAILQ_FOREACH(cert, &state->server_connp.certs, next) {
-        pemlen = BASE64_BUFFER_SIZE(cert->cert_len);
+        pemlen = Base64EncodeBufferSize(cert->cert_len);
```
victorjulien, line 125: I'm seeing a difference in the stored pem files. I'll share my example offline

victorjulien (changes requested):
TLS log issue (shared privately)

jasonish:
Issue should be fixed now. New PR: https://github.com/OISF/suricata/pull/6346

PR #6349, opened 2021-09-07
Enum dir rust/v5

inashivb:

suricata-verify-pr: 529

### DIFF
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -34,8 +34,6 @@ pub const APP_LAYER_EVENT_TYPE_PACKET      : i32 = 2;
 // From stream.h.
 pub const STREAM_START:    u8 = 0x01;
 pub const STREAM_EOF:      u8 = 0x02;
-pub const STREAM_TOSERVER: u8 = 0x04;
-pub const STREAM_TOCLIENT: u8 = 0x08;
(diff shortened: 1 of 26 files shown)


rust/src/nfs/nfs.rs
```diff
@@ -474,7 +474,7 @@ impl NFSState {
                     if self.ts > f.post_gap_ts {
                         tx.request_done = true;
                         tx.response_done = true;
-                        let (files, flags) = self.files.get(tx.file_tx_direction);
+                        let (files, flags) = self.files.get(tx.file_tx_direction.into());
```
jasonish, line 477: Clippy is complaining that this conversion is not needed.

rust/src/core.rs
```diff
@@ -33,11 +34,41 @@ pub const APP_LAYER_EVENT_TYPE_PACKET      : i32 = 2;
...
+                } else if d & DIR_TOCLIENT != 0 {
+                    Direction::ToClient
+                } else {
+                    debug_validate_fail!("Unknown direction!!");
+                    Direction::ToServer
+                }
+            }
+        }
+    }
+}
```
jasonish, line 71: I think we can drop the match and just do the if/else here.

inashivb:
Replaced w #6358

PR #6350, opened 2021-09-07
default null port in parser registration table

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4475

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -21,6 +21,7 @@ use crate::dcerpc::parser;
 use nom::error::ErrorKind;
 use nom::number::Endianness;
 use nom;
+use std;
 use std::cmp;
 use std::ffi::CString;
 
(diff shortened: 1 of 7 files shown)


catenacyber:
How is default_port used anyways ?

inashivb:
> How is default_port used anyways ?

From what I understand, the control goes to https://github.com/OISF/suricata/blob/master/src/app-layer-register.c#L78 and then `AppLayerProtoDetectPPRegister` uses it later to detect on that port for that particular proto. So, unless the default port is Null in the registration table and there is an explicit call to `AppLayerProtoDetectPPRegister`, the port is overlooked. e.g. https://github.com/OISF/suricata/blob/master/rust/src/smb/smb.rs#L2271

victorjulien:
Merged as part of #6383, thanks!

PR #6194, opened 2021-06-11
Smtp rfc2231 4386 v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4386

Describe changes:
- Move `FindMimeHeaderTokenRestrict` from C to rust
- mime :  handles multiple sections for a parameter as per RFC2231

Also fixes the evasions where the token name is present in a value as described in rust unit tests
- `attachment; dummy="filename=wrong"; filename=real`
- `attachment; notafilename=wrong; filename=good;`

suricata-verify-pr: 509
https://github.com/OISF/suricata-verify/pull/509

Modifies #6177 by using more idiomatic rust as suggested

PR #6351, opened 2021-09-07
Smtp rfc2231 4386 v4

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4386

Describe changes:
- Move `FindMimeHeaderTokenRestrict` from C to rust
- mime :  handles multiple sections for a parameter as per RFC2231

Also fixes the evasions where the token name is present in a value as described in rust unit tests
- `attachment; dummy="filename=wrong"; filename=real`
- `attachment; notafilename=wrong; filename=good;`

suricata-verify-pr: 509
https://github.com/OISF/suricata-verify/pull/509

Modifies #6194 with rebase to get QA happier

### DIFF
--- a/rust/src/mime/mod.rs
+++ b/rust/src/mime/mod.rs
@@ -34,39 +34,39 @@ pub struct MIMEHeaderTokens<'a> {
 }
 
 pub fn mime_parse_value_delimited(input: &[u8]) -> IResult<&[u8], &[u8]> {
-    let (i2, _) = tag!(input, "\"")?;
-    let (i3, value) = take_until!(i2, "\"")?;
-    let (i4, _) = tag!(i3, "\"")?;
-    return Ok((i4, value));
+    let (input, _) = tag!(input, "\"")?;
+    let (input, value) = take_until!(input, "\"")?;
+    let (input, _) = tag!(input, "\"")?;
+    return Ok((input, value));
 }
 
 pub fn mime_parse_header_token(input: &[u8]) -> IResult<&[u8], MIMEHeaderToken> {
     // maybe only U+0020 space and U+0009 tab
-    let (i1, _) = take_while!(input, |ch: u8| ch.is_ascii_whitespace())?;
-    let (i2, name) = take_until!(i1, "=")?;
-    let (i3, _) = tag!(i2, "=")?;
-    let (i4, value) = alt!(
-        i3,
+    let (input, _) = take_while!(input, |ch: u8| ch.is_ascii_whitespace())?;
+    let (input, name) = take_until!(input, "=")?;
+    let (input, _) = tag!(input, "=")?;
+    let (input, value) = alt!(
+        input,
         mime_parse_value_delimited | complete!(take_until!(";")) | rest
     )?;
-    let (i5, _) = opt!(i4, complete!(tag!(";")))?;
-    return Ok((i5, MIMEHeaderToken { name, value }));
+    let (input, _) = opt!(input, complete!(tag!(";")))?;
+    return Ok((input, MIMEHeaderToken { name, value }));
 }
 
 fn mime_parse_header_tokens(input: &[u8]) -> IResult<&[u8], MIMEHeaderTokens> {
-    let (mut i2, _) = take_until_and_consume!(input, ";")?;
+    let (mut input, _) = take_until_and_consume!(input, ";")?;
     let mut tokens = Vec::new();
-    while i2.len() > 0 {
-        match mime_parse_header_token(i2) {
+    while input.len() > 0 {
+        match mime_parse_header_token(input) {
             Ok((rem, t)) => {
                 tokens.push(t);
                 // should never happen
-                debug_validate_bug_on!(i2.len() == rem.len());
-                if i2.len() == rem.len() {
+                debug_validate_bug_on!(input.len() == rem.len());
+                if input.len() == rem.len() {
                     //infinite loop
                     return Err(Err::Error((input, ErrorKind::Eof)));
                 }
-                i2 = rem;
+                input = rem;
             }
             Err(_) => {
                 // keep first tokens is error in remaining buffer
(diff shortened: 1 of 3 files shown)


suricata-qa:

Pipeline 4347

rust/src/mime/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+        Err(_) => {
+            return Err(());
+        }
+    }
+    return Err(());
+}
+
+// TODO ? export with "constants" in cbindgen
+// and use in outbuf definition for rs_mime_find_header_token
+// but other constants are now defined twice in rust and in C
```
inashivb, line 151: Yes. Perhaps we should avoid duplication of constants. I have run into bugs in the past when I duplicated them.. e.g. same const name on Rust and C side but typo in the Rust one but things don't fall off anywhere because it is defined in both the Rust and C ecosystems when it actually should.

rust/src/mime/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+const RS_MIME_MAX_TOKEN_LEN: usize = 255;
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_mime_find_header_token(
+    hinput: *const u8, hlen: u32, tinput: *const u8, tlen: u32, outbuf: &mut [u8; 255],
+    outlen: *mut u32,
+) -> bool {
+    let hbuf = std::slice::from_raw_parts(hinput, hlen as usize);
```
inashivb, line 159: Perhaps the [`build_slice`](https://github.com/OISF/suricata/blob/master/rust/src/applayer.rs#L274) macro could be used here?

rust/src/mime/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        let badending = mime_find_header_token(
+            "attachment; filename=oksofar; badending".as_bytes(),
+            "filename".as_bytes(),
+            &mut outvec,
+        );
+        assert_eq!(badending, Ok("oksofar".as_bytes()));
+
+        let missend = mime_find_header_token(
+            "attachment; filename=test".as_bytes(),
```
inashivb, line 221: Q: Is it possible to have a case like `filename=test me wrong`? Or will it come encoded?

rust/src/mime/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+
+pub fn mime_parse_value_delimited(input: &[u8]) -> IResult<&[u8], &[u8]> {
+    let (input, _) = tag!(input, "\"")?;
+    let (input, value) = take_until!(input, "\"")?;
+    let (input, _) = tag!(input, "\"")?;
+    return Ok((input, value));
+}
+
+pub fn mime_parse_header_token(input: &[u8]) -> IResult<&[u8], MIMEHeaderToken> {
+    // maybe only U+0020 space and U+0009 tab
```
inashivb, line 44: RFC2047 says  `Any amount of linear-space-white between 'encoded-word's,
           even if it includes a CRLF followed by one or more SPACEs,
           is ignored for the purposes of display.`
does that mean all except `U+000C FORM FEED` are allowed in header tokens?

rust/src/mime/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+use nom::combinator::rest;
+use nom::error::ErrorKind;
+use nom::Err;
+use nom::IResult;
+
+#[derive(Clone, Debug)]
+pub struct MIMEHeaderToken<'a> {
+    pub name: &'a [u8],
+    pub value: &'a [u8],
+}
```
inashivb, line 29: Wondering if using a [`HashMap`](https://doc.rust-lang.org/std/collections/struct.HashMap.html) makes sense here since we're only storing a name and corresponding value in this struct?
That'll probably help us find multi section tokens easier? Not sure if there will be an overhead. Just a thought. WDYT?

inashivb (commented):
@catenacyber LGTM. I have some questions inline. Thank you! :)

rust/src/mime/mod.rs
catenacyber: indeed

rust/src/mime/mod.rs
catenacyber: Yes, but why ?

rust/src/mime/mod.rs
catenacyber: ok, so changing it to use RFC `linear-space-white` definition

rust/src/mime/mod.rs
catenacyber: Nice

rust/src/mime/mod.rs
catenacyber: It looks like quite some work has been already done since last time as there were only a few constants remaining on both sides : the ones like `STREAM_MIDSTREAM` from stream.h

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6653

rust/src/mime/mod.rs
inashivb: So, I am unsure wrt the docs whether we can or cannot have spaces between the value. The current parser accepts the spaces, is that correct?

catenacyber: So, the RFC 2045 says parameters are separated by `;` and `parameter := attribute "=" value`
and    `value := token / quoted-string` and `token := 1*<any (US-ASCII) CHAR except SPACE, CTLs, or tspecials>`

So, a strict implementation should error on that pattern `filename=test me wrong`
I think as an IDS, we want to be liberal and give the maximum data to the signatures, so we want in the end `test me wrong` instead of erroring or just `test`

What do you think ?

inashivb: I see. Thank you for explaining!

catenacyber: Rather a discussion than an explanation ;-)

inashivb: Yeah. I lack practical knowledge  to decide what's better. What you say makes sense to me.
Before your explanation, I thought we should error out on this and probably log it as an anomaly event or something still giving user all the info as a part of the eve log. But, I certainly did not consider finding such issues with the rules..

PR #6353, opened 2021-09-07
Smb memcap fix/v1

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4621

Fails build on some distros because of the use of `pow` func.

### DIFF
--- a/rust/src/memcap.rs
+++ b/rust/src/memcap.rs
@@ -1,0 +1,257 @@
+/* Copyright (C) 2021 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+const BASE: isize = 2;
+
+/// enum to store currently allowed memory units in config,
+/// can be easily expanded for more units
+#[derive(Debug, Eq, PartialEq)]
+pub enum MemUnit {
+    B       = BASE.pow(0),
+    KB      = BASE.pow(10),
+    MB      = BASE.pow(20),
+    GB      = BASE.pow(30),
+    Unknown = 0,
+ }
+
+/// Helper function to retrieve memory unit from a string slice
+///
+/// Return value:
+///     MemUnit enum type
+///
+/// # Arguments
+///
+/// * `unit` - A string slice possibly containg memory unit
+///
+/// # Examples
+/// ```
+/// use suricata_rust::memcap::{self, MemUnit};
+///
+/// let unit = "KB";
+/// assert_eq!(MemUnit::KB, memcap::get_memunit(unit));
+///
+/// let unit = "mb";
+/// assert_eq!(MemUnit::MB, memcap::get_memunit(unit));
+///
+/// let unit = "gB";
+/// assert_eq!(MemUnit::GB, memcap::get_memunit(unit));
+/// ```
+pub fn get_memunit(unit: &str) -> MemUnit {
+    let unit = &unit.to_lowercase()[..];
+    match unit {
+        "b"     => { MemUnit::B }
+        "kb"    => { MemUnit::KB }
+        "mb"    => { MemUnit::MB }
+        "gb"    => { MemUnit::GB }
+        _       => { MemUnit::Unknown }
+    }
+}
+
+/// Parses memory units from human readable form to machine readable
+///
+/// Return value:
+///     Result => Ok(f64)
+///            => Err(error string)
+///
+/// # Arguments
+///
+/// * `arg` - A string slice that holds the value parsed from the config
+///
+/// # Examples
+///
+/// ```
+/// use suricata_rust::memcap::{self, MemUnit};
+///
+/// let s = "5kb";
+/// let res = (5 * 1024) as f64;
+/// assert_eq!(Ok(res), memcap::parse(s));
+///
+/// let s = "10";
+/// let res = 10 as f64;
+/// assert_eq!(Ok(res), memcap::parse(s));
+///
+/// let s = "4hb";
+/// assert_eq!(true, memcap::parse(s).is_err());
+/// ```
+pub fn parse(arg: &str) -> Result<f64, &'static str> {
+    let arg = arg.trim();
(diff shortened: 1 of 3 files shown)


rust/src/memcap.rs
```diff
@@ -0,0 +1,257 @@
...
+    let unit = unit as f64;
+    match val.parse::<f64>() {
+        Ok(fval) => { Ok(fval * unit) }
+        Err(_)   => { Err("Failed to evaluate the value") }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::memcap::{self, MemUnit};
```
jasonish, line 124: Adding a `use super::*;` will give access to the non-pub parts above if needed.

rust/src/memcap.rs
```diff
@@ -0,0 +1,257 @@
...
+/// let res = (5 * 1024) as f64;
+/// assert_eq!(Ok(res), memcap::parse(s));
+///
+/// let s = "10";
+/// let res = 10 as f64;
+/// assert_eq!(Ok(res), memcap::parse(s));
+///
+/// let s = "4hb";
+/// assert_eq!(true, memcap::parse(s).is_err());
+/// ```
```
jasonish, line 89: Ah, these will fail if MemUnit is non-pub.  Without a consumer at this time tho I still think they should be non-pub.

rust/src/memcap.rs
```diff
@@ -0,0 +1,257 @@
...
+        unit = unit_vec.last().unwrap();
+    } else if arg_vec_len == 2 {
+        val = arg_vec[0];
+        unit = arg_vec[1];
+    }
+    val = val.trim();
+    unit = unit.trim();
+    if unit.is_empty() {
+        unit = "B";
+    }
```
jasonish, line 110: Any thoughts on using nom for this, it would look something like:
```
        use nom::{
            character::complete::{alpha0, digit1, multispace0},
            sequence::preceded,
            sequence::tuple,
        };
        let parser = tuple((preceded(multispace0, digit1), preceded(multispace0, alpha0)));
        let r: nom::IResult<&str, (&str, &str)> = parser("100  kb");
        if let Ok(r) = r {
            let val = r.1.0;
            let unit = r.1.1;
        }
```
(@victorjulien @catenacyber )

rust/src/memcap.rs
```diff
@@ -0,0 +1,257 @@
...
+/// assert_eq!(Ok(res), memcap::parse(s));
+///
+/// let s = "10";
+/// let res = 10 as f64;
+/// assert_eq!(Ok(res), memcap::parse(s));
+///
+/// let s = "4hb";
+/// assert_eq!(true, memcap::parse(s).is_err());
+/// ```
+pub fn parse(arg: &str) -> Result<f64, &'static str> {
```
jasonish, line 90: I see that the C variation uses double, but I'm not sure why.  I think we should return u64 though.

jasonish (changes requested):
My main thought is that this is a size parser, not specifically memcap parser so putting it in a module named `memcap` is probably not a good idea.  However, where should it go?  Maybe a function in `conf.rs` for now as its mostly a utility function for parsing values from the configuration?

rust/src/memcap.rs
```diff
@@ -0,0 +1,257 @@
...
+/// enum to store currently allowed memory units in config,
+/// can be easily expanded for more units
+#[derive(Debug, Eq, PartialEq)]
+pub enum MemUnit {
+    B       = BASE.pow(0),
+    KB      = BASE.pow(10),
+    MB      = BASE.pow(20),
+    GB      = BASE.pow(30),
+    Unknown = 0,
+ }
```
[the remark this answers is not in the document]
inashivb: hmm.. could probably just define consts for kb, mb and gb on the top and skip these altogether.. Wasn't sure if it would be clean since I was targeting a separate module like `util-misc` for this...

rust/src/memcap.rs
```diff
@@ -0,0 +1,257 @@
...
+pub fn get_memunit(unit: &str) -> MemUnit {
+    let unit = &unit.to_lowercase()[..];
+    match unit {
+        "b"     => { MemUnit::B }
+        "kb"    => { MemUnit::KB }
+        "mb"    => { MemUnit::MB }
+        "gb"    => { MemUnit::GB }
+        _       => { MemUnit::Unknown }
+    }
+}
```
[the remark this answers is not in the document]
inashivb: I had thought that at first but I wasn't sure if my choice of making an enum is even right.

rust/src/memcap.rs
inashivb: hmm I feel ultra stupid rn. Did not even think about nom!

rust/src/memcap.rs
inashivb: Should it then be `ceil(any float memcap)`?

inashivb:
> My main thought is that this is a size parser, not specifically memcap parser so putting it in a module named `memcap` is probably not a good idea. However, where should it go? Maybe a function in `conf.rs` for now as its mostly a utility function for parsing values from the configuration?

Understood. Thank you very much! I'll wait for Victor and Philippe to respond on usage of nom and then submit a new PR.

rust/src/memcap.rs
jasonish: We're not really used to using it outside protocols parsers, but this could be a good place for it.

rust/src/memcap.rs
jasonish: Maybe I don't understand, but we also have "<integer> <unit>" right.. With unit defaulting to "bytes" if not set. Oh, we support "12.5 gb", etc..  I'd use `floor` to get the same behaviour we have in C.

I general I think all consumers of this would expect an integer type, not a float.

rust/src/smb/smb.rs
```diff
@@ -2282,11 +2283,9 @@ pub unsafe extern "C" fn rs_smb_register_parser() {
         SCLogDebug!("Rust SMB parser registered.");
         let retval = conf_get("app-layer.protocols.smb.stream-depth");
         if let Some(val) = retval {
-            let val = val.parse::<i32>().unwrap();
-            if val < 0 {
-                SCLogError!("invalid value for stream-depth");
-            } else {
-                stream_depth = val as u32;
+            match memcap::parse(val) {
```
victorjulien, line 2286: its called memcap, but its only consumer is not about memcap...

jasonish:
> Fails build on some distros because of the use of `pow` func.

This seems supported on Rust 1.48, so I wouldn't worry about it too much.  I think we'll be updating our min. Rust version to 1.48 soon.

suricata-qa:

Pipeline 4090

rust/src/memcap.rs
inashivb: This does seem to be working well to extract the values. Thank you!

inashivb: Is it OK for a value `5 1kb` to be parsed as `5B` only? That's one case I'm unable to make it work. I was earlier erroring out on that.

inashivb:
Replaced w #6364

PR #6335, opened 2021-09-03
Feature/http range/v26

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/1576

Describe changes:
- Uses a generic container (a thread-safe hash table whose key is the url/filename) to store file parts with HTTP Range header
- handles reassembly of these parts, while enforcing a global memcap

suricata-verify-pr: 517
https://github.com/OISF/suricata-verify/pull/517

Continues #6333 with fixing the null dereference, ie removing obsolete check for `c->container->files->tail` as `c->container->lastsize` gets used 

For testing : https://github.com/jasonish/suricata-http-range-test

Left TODO : implement for HTTP2

PR #6354, opened 2021-09-07
Feature/http range/v28

catenacyber:

Describe changes:

suricata-verify-pr: 531
https://github.com/OISF/suricata-verify/pull/531

Continues #6335 with 
- Implementation for HTTP2 !
- Fixing leak in HTPStateFree by calling `HTPFileCloseHandleRange` to release ownership of file to the multi-flow `HttpRangeContainerFile` structure
- renaming to `StreamingBufferConfig` as in C so that cbindgen produces consistent prototypes for functions using this type, ie `HttpRangeContainerOpenFile`
- http2: better rust lifetimes : lifetime of output is the same as input, even if the function also mutates the decoder (turning it to nil in case of a decompression error), so that we can mutably borrow the transaction and use the decompressed content...
- range: remove final zero in key as it seems useless (except for easy print debugging)
- range: post process out of order blocks :
This was a mistake in previous implementation :
 * I start range request 10-20, it is out of order
 * Then I start range request 0-10, in order
 * Then I finish range request 0-10
 * Last, I finish range request 10-20 : as it was deemed out of order at the opening, we did not try to put more content in the file. If this is the last block, we never complete the file !

For testing : https://github.com/jasonish/suricata-http-range-test


catenacyber:
Oh, I think I see the point of this SuricataContext...
To be continued...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6355

PR #6355, opened 2021-09-07
Feature/http range/v29

catenacyber:

Describe changes:

suricata-verify-pr: 531
https://github.com/OISF/suricata-verify/pull/531

Continues #6354 with 
- Using SuricataContext so as to compile suricata crate without C for unit tests

For testing : https://github.com/jasonish/suricata-http-range-test


suricata-qa:

Pipeline 4089

catenacyber:
Includes https://redmine.openinfosecfoundation.org/issues/4117 as well

suricata-qa:
Information:
|--|--|--|--|
| | tlpr1_stats_chk |
| .uptime | 1344 | 1220 | 110.16% |
| .flow.mgr.rows_maxlen | 450 | 404 | 111.39% |
| .detect.alert | 3090528 | 3947262 | 78.3% |

Pipeline 4089

victorjulien:
`try: close before free` < bad commit msg

rust/src/http2/http2.rs
```diff
@@ -167,6 +172,27 @@ impl HTTP2Transaction {
         if let Some(state) = self.de_state {
             core::sc_detect_engine_state_free(state);
         }
+        if self.file_range != std::ptr::null_mut() {
+            match unsafe { SC } {
+                None => panic!("BUG no suricata_config"),
+                Some(c) => {
+                    //TODO get a file container instead of NULL
```
victorjulien, line 179: TODO: address or remove comment if its not needed

rust/src/http2/http2.rs
```diff
@@ -350,6 +402,7 @@ pub struct HTTP2State {
     transactions: Vec<HTTP2Transaction>,
     progress: HTTP2ConnectionState,
     pub files: Files,
+    flow: *const Flow,
```
victorjulien, line 405: ideally we're not adding this if we can just pass it around as a function argument

rust/src/http2/range.rs
```diff
@@ -0,0 +1,241 @@
+/* Copyright (C) 2021 Open Information Security Foundation
```
victorjulien, line 1: Is the parsing of the content-range string the same for http1 and http2? If so I'd like to consolidate the parsing in Rust

victorjulien: whoops already done it seems, please ignore

src/app-layer-htp-file.c
```diff
@@ -161,60 +161,10 @@ int HTPFileOpen(HtpState *s, HtpTxUserData *tx, const uint8_t *filename, uint16_
...
-        pos++;
-        if (data[pos] != '*') {
-            // case with size
-            range->size = bstr_util_mem_to_pint(data + pos, len - pos, 10, &last_pos);
-        }
-    }
-
-    return 0;
+    uint32_t len = bstr_len(rawvalue);
+    return rs_http2_parse_content_range(range, bstr_ptr(rawvalue), len);
```
victorjulien, line 167: nit: I don't like that the call has 'http2' in the name

catenacyber: indeed, some leftover from development version when there were 2 parsing functions

rust/src/http2/http2.rs
catenacyber: Right

rust/src/http2/http2.rs
catenacyber: It will be much easier when we have file per tx 

Right now, I should do something like `state.files.get(STREAM_TOCLIENT)` but I do not have an easy reference to `state` in `HTTP2Transaction::free`

So, what should we do ?

catenacyber:
> `try: close before free` < bad commit msg

squashed into the right commit

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6409

PR #6358, opened 2021-09-08
Enum dir rust/v6

inashivb:

suricata-verify-pr: 529

### DIFF
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -34,8 +34,6 @@ pub const APP_LAYER_EVENT_TYPE_PACKET      : i32 = 2;
 // From stream.h.
 pub const STREAM_START:    u8 = 0x01;
 pub const STREAM_EOF:      u8 = 0x02;
-pub const STREAM_TOSERVER: u8 = 0x04;
-pub const STREAM_TOCLIENT: u8 = 0x08;
 pub const STREAM_GAP:      u8 = 0x10;
 pub const STREAM_DEPTH:    u8 = 0x20;
 pub const STREAM_MIDSTREAM:u8 = 0x40;
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -19,6 +19,7 @@
 
 use std;
 use crate::filecontainer::*;
+use crate::debug_validate_fail;
 
 /// Opaque C types.
 pub enum DetectEngineState {}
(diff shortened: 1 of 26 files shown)


rust/src/core.rs
```diff
@@ -38,6 +39,36 @@ pub const STREAM_TOCLIENT: u8 = 0x08;
...
+        if d == DIR_TOSERVER {
+            Direction::ToServer
+        } else if d == DIR_TOCLIENT {
+            Direction::ToClient
+        } else if d & DIR_TOSERVER != 0 {
+            Direction::ToServer
+        } else if d & DIR_TOCLIENT != 0 {
+            Direction::ToClient
+        } else {
+            debug_validate_fail!("Unknown direction!!");
```
catenacyber, line 65: Should we `debug_validate_fail ` if we have both `DIR_TOSERVER` and `DIR_TOCLIENT ` set ?

rust/src/core.rs
```diff
@@ -38,6 +39,36 @@ pub const STREAM_TOCLIENT: u8 = 0x08;
 pub const STREAM_GAP:      u8 = 0x10;
 pub const STREAM_DEPTH:    u8 = 0x20;
 pub const STREAM_MIDSTREAM:u8 = 0x40;
+pub const DIR_TOSERVER:    u8 = 0b0000_0100;
```
catenacyber, line 40: Why did you not use STREAM_TOSERVER ?

inashivb: I thought that keeping it might not enforce people to use Direction enum in the latest PRs. If you don't agree, I can rename these back to STREAM_TOSERVER

rust/src/core.rs
inashivb: hmm makes sense to me. @victorjulien ?

rust/src/core.rs
catenacyber: Smart :-)

rust/src/core.rs
```diff
@@ -33,11 +34,39 @@ pub const APP_LAYER_EVENT_TYPE_PACKET      : i32 = 2;
...
+impl Default for Direction {
+    fn default() -> Self { Direction::ToServer }
+}
+
+impl From<u8> for Direction {
+    fn from(d: u8) -> Self {
+        if d == DIR_TOSERVER {
+            Direction::ToServer
+        } else if d == DIR_TOCLIENT {
+            Direction::ToClient
```
jasonish, line 59: I think we can drop this as its covered but the 2 following checks.

rust/src/core.rs
jasonish: > hmm makes sense to me. @victorjulien ?

Makes sense to me as well.

rust/src/core.rs
inashivb: Of course! :no_mouth:

inashivb:
Replaced w #6362

PR #6361, opened 2021-09-08
(draft) rust: proof of concept of how memory can be calculated

jasonish:
Proof of concept how app-layer memory usage can be calculated for Rust. Due to
how Rust allocates memory I don't think it makes sense to update a counter
every time the equivalent of a malloc() is done, this could make the parsing
(nom) code cluttered with accounting code.

Instead each struct that makes up a transaction should implement sizeof methods that can
be used.  For DNS here, `sizeof()` can be called on a DNSTransaction and it
will call down to calculate the size of the transaction.

For DNS this is pretty easy as transactions are never updated, so we can just
gather the size of the transaction after creation which will be the same when
the transaction is dropped.

Protocols where transactions are update (or added to over time) will likely be
harder to provide accurate accounting for, but that will need to be done on a
per case basis.

I'm still not sure about the accounting.  For proof of concept purposes we
could use the global atomic like we have in C, but thats probably not idea
going into the future.

The use of a trait for SizeOf allows us to provide a default implementation for
structs that don't have any dynamically allocated data, and it also leaves us
open to creating a macro to derive implementation for us.. For example:

#[derive(Debug, PartialEq, SizeOf)]
pub struct DNSTransaction {
...
}

which I think could work for many of the struct that need to be accounted for
(but I haven't tried this yet).

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -115,6 +115,36 @@ pub const DNS_RCODE_BADTRUNC: u16 = 22;
 
 static mut ALPROTO_DNS: AppProto = ALPROTO_UNKNOWN;
 
+/// Trait for objects to implement methods that calculate the size of that object.
+pub trait SizeOf {
+    /// Returns the sizeof the object itself, plus the inner data.
+    fn sizeof(&self) -> usize {
+        std::mem::size_of_val(self) + self.sizeof_inner()
+    }
+
+    /// Just return the size of data that can be variable such as vectors. If a struct
+    /// has no dynamically allocated fields this default implementation will do.
+    fn sizeof_inner(&self) -> usize {
+        0
+    }
+}
+
+/// SizeOf implementation for Options whre the value implements SizeOf.
+impl<T> SizeOf for Option<T> where T: SizeOf {
+    fn sizeof_inner(&self) -> usize {
+        self.as_ref().map_or(0, SizeOf::sizeof_inner)
+    }
+}
+
+/// SizeOf implementation for vectors where the values implement SizeOf.
+impl<T> SizeOf for Vec<T> where T: SizeOf {
+    fn sizeof_inner(&self) -> usize {
+        // When calculating the size of a vector we take the full size of each
+        // element. Not just the inner.
+        self.iter().map(SizeOf::sizeof).sum()
+    }
+}
+
 #[derive(Debug, PartialEq, AppLayerEvent)]
 pub enum DNSEvent {
     MalformedData,
@@ -141,6 +171,12 @@ pub struct DNSQueryEntry {
     pub rrclass: u16,
 }
 
+impl SizeOf for DNSQueryEntry {
+    fn sizeof_inner(&self) -> usize {
+        self.name.len()
+    }
+}
+
 #[derive(Debug,PartialEq)]
 pub struct DNSRDataSOA {
     /// Primary name server for this zone
@@ -159,6 +195,12 @@ pub struct DNSRDataSOA {
     pub minimum: u32,
 }
 
+impl SizeOf for DNSRDataSOA {
+    fn sizeof_inner(&self) -> usize {
+        self.mname.len() + self.rname.len()
+    }
+}
+
 #[derive(Debug,PartialEq)]
 pub struct DNSRDataSSHFP {
     /// Algorithm number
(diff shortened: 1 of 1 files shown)


rust/src/dns/dns.rs
```diff
@@ -115,6 +115,36 @@ pub const DNS_RCODE_BADTRUNC: u16 = 22;
 
 static mut ALPROTO_DNS: AppProto = ALPROTO_UNKNOWN;
 
+/// Trait for objects to implement methods that calculate the size of that object.
+pub trait SizeOf {
+    /// Returns the sizeof the object itself, plus the inner data.
+    fn sizeof(&self) -> usize {
+        std::mem::size_of_val(self) + self.sizeof_inner()
```
inashivb, line 122: This is awesome. :)
I have a question. How do we calculate the size of opaque type items with no explicit type from C that is used in tx? e.g. `DetectEngineState`

jasonish: You don't, at least not simply and accurately.  I see 2 options:
- Hardcode a size of known datatypes from C, these would have to be reviewed.
- Or define similar functions on the C side and call into them.
Or you just wrap it up into the extra overhead over and above the memcap that is likely to exist for any parser, and then memcap is close, but maybe 2-3% wrong.  At least within a specific Suricata version the margin of error will be constant (I don't think any of these C structs grow in size in response to specific network traffic).

rust/src/dns/dns.rs
```diff
@@ -115,6 +115,36 @@ pub const DNS_RCODE_BADTRUNC: u16 = 22;
 
 static mut ALPROTO_DNS: AppProto = ALPROTO_UNKNOWN;
 
+/// Trait for objects to implement methods that calculate the size of that object.
+pub trait SizeOf {
```
victorjulien, line 119: if we define a trait like this, do we risk name collisions with future rust features?

rust/src/dns/dns.rs
```diff
@@ -141,6 +171,12 @@ pub struct DNSQueryEntry {
     pub rrclass: u16,
 }
 
+impl SizeOf for DNSQueryEntry {
+    fn sizeof_inner(&self) -> usize {
+        self.name.len()
```
victorjulien, line 176: iirc a vector can also tell you how much space it reserved, right?

victorjulien: also, would it be possible to somehow handle every Vec<u8> in a generic way, so w/o needing to create small util funcs for each field?

rust/src/dns/dns.rs
jasonish: These name collisions happen more with keywords.  Traits are name spaced like function names and very few are brought into the global name space (https://doc.rust-lang.org/std/prelude/index.html), but even then its not a compile error to redeclare a trait, the scoping rules will determine what has precedence, so its not something I'd worry about until it became an issue.

rust/src/dns/dns.rs
jasonish: > iirc a vector can also tell you how much space it reserved, right?

Yeah, typically they double in size every time they need to grow.  So we could use the `capacity()` method instead.

jasonish: > also, would it be possible to somehow handle every `Vec<u8>` in a generic way, so w/o needing to create small util funcs for each field?

Its actually a util method per struct, not field.  But this is where the derive crate would come into play.  It could look at the data structure and generate the sizeof methods for most of these.

sharksforarms:
This crate might be helpful/relevant: https://crates.io/crates/memuse

jasonish:
> This crate might be helpful/relevant: https://crates.io/crates/memuse

Any idea why this crate when deepsize and heapsize already exist? I steered clear of using an off-the-shelf one as while reading more about this it seems common to hit point where domain specific knowledge of the program is required to properly account for how data is used.  I guess the biggest case of this is the Servo project which dropped heapsize in favour of one they just keep internal to Servo now.

sharksforarms:
> 
> 
> > This crate might be helpful/relevant: https://crates.io/crates/memuse
> 
> Any idea why this crate when deepsize and heapsize already exist? I steered clear of using an off-the-shelf one as while reading more about this it seems common to hit point where domain specific knowledge of the program is required to properly account for how data is used. I guess the biggest case of this is the Servo project which dropped heapsize in favour of one they just keep internal to Servo now.

I'm not sure. I saw it on my timeline and thought it may be relevant to this thread, maybe for implementation ideas/patterns or other!

Looking at the github issues of the `memuse` project, it seems like the author also realized this and would like to integrate features from these crates into `memuse`.

suricata-qa:
WARNING:
|--|--|--|--|
| | tlpr1_stats_chk |
| .flow.spare | 5951235 | 2210021 | 269.28% |
| .flow.memuse | 1145338368 | 536294848 | 213.57% |

Pipeline 4513

victorjulien:
I think it would be cool if this would be developed a bit further, a look into how derive could help

catenacyber:
Closing as it needs rebase

PR #6364, opened 2021-09-09
fix: SMB stream depth setting in Rust

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4621

Fails build on some distros with old Rust.

### DIFF
--- a/rust/src/conf.rs
+++ b/rust/src/conf.rs
@@ -21,6 +21,69 @@ use std::os::raw::c_int;
 use std::ffi::{CString, CStr};
 use std::ptr;
 use std::str;
+use nom::{
+    character::complete::{multispace0, not_line_ending},
+    sequence::{preceded, tuple},
+    number::complete::double,
+    combinator::verify,
+};
+
+const BASE: u64       = 2;
+const BYTE: u64       = BASE.pow(0);
+const KILOBYTE: u64   = BASE.pow(10);
+const MEGABYTE: u64   = BASE.pow(20);
+const GIGABYTE: u64   = BASE.pow(30);
+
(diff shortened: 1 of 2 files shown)


victorjulien:
PR still talks about memcap.

Also, what is the idea of the Rust errors? It seems that a basic config string parser shouldn't lead to us bumping MSRV

victorjulien:
Can you also explain what this does in the git commits?

inashivb:
> PR still talks about memcap.
> 
Sorry. I forgot to change the branch name. Came from there. I'll change it.

> Also, what is the idea of the Rust errors? It seems that a basic config string parser shouldn't lead to us bumping MSRV
Jason [mentioned](https://github.com/OISF/suricata/pull/6353#issuecomment-914544102) in my last PR that we are going to update our MSRV soon so I don't have to worry about these errors but if it is not happening very soon, I can change the use of the function that leads to the error.

victorjulien:
Looking at the error are we really bumping MSRV because we're unwilling to use a constant value for megabyte, kilobyte, etc?

inashivb:
> Looking at the error are we really bumping MSRV because we're unwilling to use a constant value for megabyte, kilobyte, etc?

I don't think so. I think Jason meant that we are going to do it anyway so this error will be gone with the bump.

inashivb:
Replaced w #6365

jasonish:
I thought I tried `.pow` with 1.48 and it worked, but it appears you need 1.50 at least.  So we'll have to drop is use for another year or 2.

PR #6357, opened 2021-09-07
plugins: use plugin api for internal filetypes - v3

jasonish:
This PR combines Jeff's work to make the built-in syslog eve filetype register
via the new plugin file type.

After Jeff's work I added some changes, mainly renaming so that built-ins
using the API didn't have to be named with "plugin" or internal "plugin"
applying the idea that the plugin API is more of exposing internal registration
methods.

Other internal eve file types could be registered this way as well which might
cleanup some of the rather large abstraction functions we have over file types.

Previous PR: https://github.com/OISF/suricata/pull/6161

Changes from previous PR:
- Remove all the other plugin cleanups and include refactoring making this
  change easier to review.

Jeff's original PR: https://github.com/OISF/suricata/pull/6133

PR #6369, opened 2021-09-10
plugins: use plugin api for internal file types - v4

jasonish:
Minor fixup to previous PR (https://github.com/OISF/suricata/pull/6357) which
included an rogue file.

This PR combines Jeff's work to make the built-in syslog eve filetype register
via the new plugin file type.

After Jeff's work I added some changes, mainly renaming so that built-ins
using the API didn't have to be named with "plugin" or internal "plugin"
applying the idea that the plugin API is more of exposing internal registration
methods.

Other internal eve file types could be registered this way as well which might
cleanup some of the rather large abstraction functions we have over file types.

Previous PR: https://github.com/OISF/suricata/pull/6161

Changes from previous PR:
- Remove all the other plugin cleanups and include refactoring making this
  change easier to review.

Jeff's original PR: https://github.com/OISF/suricata/pull/6133


victorjulien:
Merged in #6395, thanks!

PR #6375, opened 2021-09-15
Http2 upgrade disabled 4555 v1

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4555

Describe changes:
- Do not try protocol change about HTTP2 upgrade is HTTP2 is disabled in yaml configuration

So, we will stay in `HTP_STREAM_TUNNEL` state, and libhtp will just skip the data...

How can we test this ?

### DIFF
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -965,6 +965,10 @@ static AppLayerResult HTPHandleResponseData(Flow *f, void *htp_state,
                     if (bstr_cmp_c(h->value, "h2c") != 0) {
                         break;
                     }
+                    if (!AppLayerProtoDetectConfProtoDetectionEnabled("tcp", "http2")) {
+                        // if HTTP2 is disabled, keep the HTP_STREAM_TUNNEL mode
+                        break;
+                    }
                     uint16_t dp = 0;
                     if (tx->request_port_number != -1) {
                         dp = (uint16_t)tx->request_port_number;
@@ -974,8 +978,7 @@ static AppLayerResult HTPHandleResponseData(Flow *f, void *htp_state,
                     // During HTTP2 upgrade, we may consume the HTTP1 part of the data
                     // and we need to parser the remaining part with HTTP2
                     if (consumed > 0 && consumed < input_len) {
-                        SCReturnStruct(
-                                APP_LAYER_INCOMPLETE(consumed, input_len - consumed));
+                        SCReturnStruct(APP_LAYER_INCOMPLETE(consumed, input_len - consumed));
                     }
                     SCReturnStruct(APP_LAYER_OK);
                 }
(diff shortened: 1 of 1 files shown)


victorjulien:
> How can we test this ?

I'd say SV tests that check that:
- we don't have http2 records
- stream inspection in the http2 part of the stream still works
- http1 tx is still logged at the time of the upgrade (so validate that http1 tx is closed still)

src/app-layer-htp.c
```diff
@@ -959,23 +959,25 @@ static AppLayerResult HTPHandleResponseData(Flow *f, void *htp_state,
...
-                            AppLayerRequestProtocolChange(hstate->f, dp, ALPROTO_HTTP2);
-                            // During HTTP2 upgrade, we may consume the HTTP1 part of the data
-                            // and we need to parser the remaining part with HTTP2
-                            if (consumed > 0 && consumed < input_len) {
-                                SCReturnStruct(
-                                        APP_LAYER_INCOMPLETE(consumed, input_len - consumed));
-                            }
-                            SCReturnStruct(APP_LAYER_OK);
-                        }
+                    if (h == NULL) {
```
victorjulien, line 962: `if (h == NULL || bstr_cmp_c(h->value, "h2c") != 0) {` to compress things a bit?

src/app-layer-htp.c
```diff
@@ -965,6 +965,10 @@ static AppLayerResult HTPHandleResponseData(Flow *f, void *htp_state,
                     if (bstr_cmp_c(h->value, "h2c") != 0) {
                         break;
                     }
+                    if (!AppLayerProtoDetectConfProtoDetectionEnabled("tcp", "http2")) {
```
victorjulien, line 968: `AppLayerProtoDetectConfProtoDetectionEnabled` should not be used in the packet runtime.

src/app-layer-htp.c
catenacyber: Right :-)

src/app-layer-htp.c
catenacyber: Using `AppLayerProtoDetectGetProtoName` then ?

victorjulien: I would say some bool/flag set during registration, so we don't call routines that aren't optimized for the packet runtme

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6385

catenacyber:
> > How can we test this ?
> 
> I'd say SV tests that check that:
> 
> * we don't have http2 records
> * stream inspection in the http2 part of the stream still works
> * http1 tx is still logged at the time of the upgrade (so validate that http1 tx is closed still)

Thanks Victor
Done in https://github.com/OISF/suricata-verify/pull/541

src/app-layer-htp.c
catenacyber: Do you mean to create a new bool ?
Or is there one ?

(Did not see that before next PR.)

PR #6376, opened 2021-09-16
 PostgreSQL: add initial support (#4241) - v1

jufajardini:
Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/4241

Previous PR: #6281 

Describe changes:
- Add upgrade to TLS to our rust side. - https://github.com/OISF/suricata/commit/af931d7e3306f9b2222a6bf77dae11f4a47cc9fc
- Parser now upgrades to TLS when it parses an `SSL Accepted`
- main loops for parsing request and response are now shorter, due to the extraction of util functions
- Cleaned up TransactionState StateProgress superposition of values
- Transactions are split as expected, now, respecting the concept of having only one request per transaction
- This version has a less verbose logger. For this, I commented out the parsers for DataRow, and simplified what the logger shows for `ParameterStatus` and `RowDescription`, too.
- Using Option for the return of `find_or_create_tx` to prevent it from panicking. Not sure this is the best approach.

Considerations:
- Commit history needs A LOT of work, but leaving it as is, for now, to make diff checks easier (hopefully).
- found a situation where PGSQL isn't detected when there's an SSL handshake upgraded to TLS. Not sure why it doesn't trigger in this case, but works in others, though: https://git.lekensteyn.nl/peter/wireshark-notes/commit/tls/pgsql-ssl.pcapng?id=836b6f746df24aa04fa29b71806d8d0e496c2a68

Mostly tested with the pcaps available here:
https://drive.google.com/drive/folders/1BPJICt_ybkofDUPrMdGglvfQeefFCQF_?usp=sharing


rust/src/pgsql/parser.rs
```diff
@@ -0,0 +1,2089 @@
...
+// TODO - do I really need this?
+impl fmt::Display for SSLResponseMessage {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            SSLResponseMessage::SSLAccepted => write!(f, "SSL Accepted"),
+            SSLResponseMessage::SSLRejected => write!(f, "SSL Rejected"),
+            _ => write!(f, "Invalid server response"),
+        }
+    }
+}
```
jasonish, line 139: No, I don't think so.  You'd have to match on the variants in your logger without this though.  But thats not a bad thing.  This invokes the formatter which is slow, and creates a string, just to have it dereferenced in the logger.  So may as well just match in the logger and use static strings.

Or even better, create an `impl` for `SSLResponseMessage` and add a `to_str()` method that returns the pretty string as a static.

victorjulien:
I would like to see far fewer commits. Also I would like to see the PR to have a Suricata-Verify PR reference that adds tests for the various types of logs.

victorjulien:
@catenacyber can you assist @jufajardini with the CIFuzz builds?

rust/src/pgsql/parser.rs
```diff
@@ -0,0 +1,2089 @@
...
+
+impl fmt::Display for PgsqlBEMessage {
+    fn fmt(&self, f:&mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl PgsqlBEMessage {
+    pub fn message_type_to_str(&self) -> &str {
```
jasonish, line 194: Just name this `to_str()`, I don't think there is any `to_str()` trait to worry about.

jasonish (changes requested):
I agree with what Victor said about the commits.  Also please `rustfmt` the new files and give `clippy` a run for some low hanging very minor issues.

Thanks!

rust/src/pgsql/logger.rs
```diff
@@ -0,0 +1,432 @@
...
+use crate::pgsql::parser::*;
+
+fn log_pgsql(tx: &PgsqlTransaction, js: &mut JsonBuilder) -> Result<(), JsonError> {
+    js.set_uint("transaction_id", tx.tx_id.into())?;
+    if !tx.requests.is_empty() {
+        if tx.requests.len() > 1 {
+            js.open_array("requests")?;
+            for request in &tx.requests {
+                let jbr = log_request(&request)?;
+                js.append_object(&jbr)?;
```
jasonish, line 34: Should be able to avoid new object allocation, and just build onto the `js`. If not, please comment why.

rust/src/pgsql/logger.rs
```diff
@@ -0,0 +1,432 @@
...
+            TerminationMessage{
+                identifier: _,
+                length: _,
+            }) =>
+        {
+            js.set_string("message_type", "termination_token")?;
+        }
+    }
+    js.close()?;
+    js.close()?;
```
jasonish, line 135: The double close kind of looks odd.  But not always a sign of an issue, so just add comments as to what you are closing.

jasonish (commented):
No extra comments on the logging yet.  Its very verbose, which is really awesome.  Anyways, better to be too verbose for now.  I'd like to scare up some PostgreSQL traffic before making some decisions here.

catenacyber:
> @catenacyber can you assist @jufajardini with the CIFuzz builds?

With this PR, `suricata --list-app-layer-protos` prints notices log about postgresql registration and it should not

Output is 
```
./src/suricata --list-app-layer-protos
[259] 16/9/2021 -- 22:32:56 - (app-layer-pgsql.c:47) <Notice> (RegisterPgsqlParsers) -- Registering Rust pgsql parser.
[259] 16/9/2021 -- 22:32:56 - (pgsql.rs:757) <Notice> (suricata_rust::pgsql::pgsql::rs_pgsql_register_parser) -- Rust pgsql parser registered.
=========Supported App Layer Protocols=========
http
ftp
smtp
tls
ssh
imap
smb
dcerpc
dns
nfs
ntp
ftp-data
tftp
ike
krb5
dhcp
snmp
sip
rfb
mqtt
pgsql
rdp
http2
```

I guess it should be SCLogDebug instead of SCLogNotice

jufajardini:
> @catenacyber can you assist @jufajardini with the CIFuzz builds?

Thanks! Is this what I must do? `./configure --enable-fuzztargets` ?

jufajardini:
> > @catenacyber can you assist @jufajardini with the CIFuzz builds?
> 
> With this PR, `suricata --list-app-layer-protos` prints notices log about postgresql registration and it should not
> 
> Output is
> 
> ```
> ./src/suricata --list-app-layer-protos
> [259] 16/9/2021 -- 22:32:56 - (app-layer-pgsql.c:47) <Notice> (RegisterPgsqlParsers) -- Registering Rust pgsql parser.
> [259] 16/9/2021 -- 22:32:56 - (pgsql.rs:757) <Notice> (suricata_rust::pgsql::pgsql::rs_pgsql_register_parser) -- Rust pgsql parser registered.
> =========Supported App Layer Protocols=========
> http
> ftp
> smtp
> tls
> ssh
> imap
> smb
> dcerpc
> dns
> nfs
> ntp
> ftp-data
> tftp
> ike
> krb5
> dhcp
> snmp
> sip
> rfb
> mqtt
> pgsql
> rdp
> http2
> ```
> 
> I guess it should be SCLogDebug instead of SCLogNotice

Ugh, my bad, I didn't realize this one wasn't allowed. Should have foreseen that one...

catenacyber:
> Thanks! Is this what I must do? ./configure --enable-fuzztargets ?

If you want to build the fuzz targets, yes :-)
You can also use the wrapper `./src/tests/fuzz/oss-fuzz-configure.sh `

See src/tests/fuzz/README 

Then if you want to fuzz pgsql only, you can `cp src/fuzz_applayerparserparse src/fuzz_applayerparserparse_pgsql` and then run `fuzz_applayerparserparse_pgsql`(or you can run fuzz_applayerparserparse with env variable `FUZZ_APPLAYER=31` or something like that)

rust/src/pgsql/parser.rs
jufajardini: Thanks, I shall do the same for the other `fmt::Display` that are there, then, too.

PR #6378, opened 2021-09-17
Backports 6.0.4 b2/v1

inashivb:
Test filestore-v2.7-stream-depth seems to fail (Sub test 2) for these commits. It was recently updated with an `lt-version: 7` (https://github.com/OISF/suricata-verify/commit/d3cf564bb18e7f5107dddffd127976db00d2c8dd).
I need help figuring if the tests need an update here or the code?
Commit https://github.com/inashivb/suricata/commit/1dbf8b402d3caeebfd233d9f48bc90a30549ff9d seems to be causing this failure.

### DIFF
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -60,6 +60,7 @@ pub struct AppLayerTxData {
     /// track file open/logs so we can know how long to keep the tx
     pub files_opened: u32,
     pub files_logged: u32,
+    pub files_stored: u32,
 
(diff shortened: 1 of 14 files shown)


victorjulien:
The SV test will had to be adjusted. These code changes cause a change in file output that we consider an improvement/fix. The test was updated to account for their when master was updated. So now we'll need the same for 6.

inashivb:
Replaced w #6379

PR #6380, opened 2021-09-17
ci: use quiet option w suricata-verify

inashivb:


victorjulien:
Merged as part of #6383, thanks!

PR #6346, opened 2021-09-06
base64: use Rust base64 encoder - v3

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/6342

Changes from last PR:
- Fix C buffer length accounting.

PR #6423, opened 2021-09-29
base64: use Rust base64 encoded - v4

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/6346

Changes from previous PR:
- Rebase.

Add methods `jb_set_base64` and `jb_append_base64` to JsonBuilder to set/append byte array values as base64. This avoids the intermediate buffer that we currently use, and the Rust base64 encode we use can encode directly to the underlying JsonBuilder buffer (internally it appears to use a 1024 byte intermediate buffer that is flushed to the output string when full).

Replace calls to the C Bse64Encode to a Rust version.

Would be good to have a performance test done. I haven't done my own micro-benchmarks yet, but may do so soon.


src/output-json-alert.c
```diff
@@ -423,11 +422,7 @@ static void AlertJsonTunnel(const Packet *p, JsonBuilder *js)
 static void AlertAddPayload(AlertJsonOutputCtx *json_output_ctx, JsonBuilder *js, const Packet *p)
 {
     if (json_output_ctx->flags & LOG_JSON_PAYLOAD_BASE64) {
-        unsigned long len = BASE64_BUFFER_SIZE(p->payload_len);
-        uint8_t encoded[len];
-        if (Base64Encode(p->payload, p->payload_len, encoded, &len) == SC_BASE64_OK) {
-            jb_set_string(js, "payload", (char *)encoded);
-        }
+        jb_set_base64(js, "payload", p->payload, p->payload_len);
```
jlucovsky, line 425: Will this do the correct thing if `payload==NULL` or `payload_len == 0`?

suricata-qa:
Information:
|--|--|--|--|
| | tlpr1_stats_chk |
| .flow.mgr.rows_maxlen | 450 | 364 | 123.63% |

Pipeline 4511

catenacyber:
HTTP2 uses some custom `http2_decode_base64`.
Could this be replaced as well ?

jasonish:
> HTTP2 uses some custom `http2_decode_base64`. Could this be replaced as well ?

I think so.  I plan to rebase this for testing soon, will look at converting that usage as well.

jasonish:
> > HTTP2 uses some custom `http2_decode_base64`. Could this be replaced as well ?
> 
> I think so. I plan to rebase this for testing soon, will look at converting that usage as well.

PR will come, but it appears to drop right in: https://github.com/OISF/suricata/commit/2a622750bc13f8f39139d00f201316828c33c11e

jasonish:
Continued here: https://github.com/OISF/suricata/pull/6752 with http2 converted.

catenacyber:
Thanks, commit looks good with so much red :-)

PR #6425, opened 2021-09-30
pcre: creates a match structure per match run

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4720

Describe changes:
- Creates a `pcre2_match_data` for each call of `DetectPcrePayloadMatch`

I am not sure this is the problem.

But if there are two threads which can call concurrently `DetectPcrePayloadMatch` with the same `DetectPcreData` (ie SigMatchData context), that is indeed a problem

TODO 
- check this is the bug
- check if there are other places with the same problem
- assess performance impact

### DIFF
--- a/src/detect-pcre.c
+++ b/src/detect-pcre.c
@@ -82,10 +82,11 @@ static int pcre2_use_jit = 1;
 /* \brief Helper function for using pcre2_match with/without JIT
  */
 static inline int DetectPcreExec(DetectEngineThreadCtx *det_ctx, DetectPcreData *pd,
-        const char *str, const size_t strlen, int start_offset, int options)
+        const char *str, const size_t strlen, int start_offset, int options,
+        pcre2_match_data *match)
 {
     return pcre2_match(pd->parse_regex.regex, (PCRE2_SPTR8)str, strlen, start_offset, options,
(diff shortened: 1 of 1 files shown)


src/detect-pcre.c
```diff
@@ -199,7 +200,8 @@ int DetectPcrePayloadMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
     }
 
     /* run the actual pcre detection */
-    ret = DetectPcreExec(det_ctx, pe, (char *)ptr, len, start_offset, 0);
+    pcre2_match_data *match = pcre2_match_data_create_from_pattern(pe->parse_regex.regex, NULL);
```
victorjulien, line 203: you can use thread specific data with `DetectRegisterThreadCtxFuncs`, `DetectThreadCtxGetKeywordThreadCtx`. See `src/detect-lua.c` for an example implementation.

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6427

src/detect-pcre.c
catenacyber: Thanks Victor.
I gathered from your answer that
- this was indeed one bug
- the performance impact could be avoided

PR #6428, opened 2021-09-30
http2: enable by default

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4721

Describe changes:
- Enables HTTP2 by default

It looks too simple. Did I miss something ?

### DIFF
--- a/suricata.yaml.in
+++ b/suricata.yaml.in
@@ -286,10 +286,8 @@ outputs:
         - ssh
         - mqtt:
             # passwords: yes           # enable output of passwords
-        # HTTP2 logging. HTTP2 support is currently experimental and
-        # disabled by default. To enable, uncomment the following line
-        # and be sure to enable http2 in the app-layer section.
-        #- http2
+        # HTTP2 logging. HTTP2 support is now enabled by default
+        - http2
         - stats:
             totals: yes       # stats for all threads merged together
             threads: no       # per thread stats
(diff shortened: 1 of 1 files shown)


jasonish:
> It looks too simple. Did I miss something ?

Is it still enabled by default if this section is missing from the configuration file?

catenacyber:
> > It looks too simple. Did I miss something ?
> 
> Is it still enabled by default if this section is missing from the configuration file?

I think it does per
`AppLayerProtoDetectConfProtoDetectionEnabled` which defaults to enabled cf 

```
        node = ConfGetNode(param);
        if (node == NULL) {
            SCLogDebug("Entry for %s not found.", param);
            goto enabled;
        }
```

jasonish (approved):
LGTM.

victorjulien:
So http2 was already enabled by default, just not in the default config?

catenacyber:
> So http2 was already enabled by default, just not in the default config?

Indeed.
I guess I had understood that not in the default config means not enabled by default.
But that is the case for other protocols such as dnp3, right ? (I mean dnp3 is enabled by default, just not in the default config)

jasonish:
DNP3 and ENIP are in error as well.  If left out of the config they end up enabled.  I wonder if we should fix this for 6.0.4, or do it 7.  I'm thinking 6.0.4.

catenacyber:
So, should we create another ticket for this bug ?

catenacyber:
Created https://redmine.openinfosecfoundation.org/issues/4739

catenacyber:
Cf https://github.com/OISF/suricata/pull/6454

PR #6444, opened 2021-10-05
github-ci: use sccache for gcc in commits workflow

jasonish:
This was done for Rust already, so use sccache for gcc as well.  Ccache should
be been used before as it was in the path, but a test run shows that sccache is
being used for sure here.

https://github.com/jasonish/suricata/runs/3805214195?check_suite_focus=true#step:12:1

Redmine issue: https://redmine.openinfosecfoundation.org/issues/3905

### DIFF
--- a/.github/workflows/commits.yml
+++ b/.github/workflows/commits.yml
@@ -23,7 +23,6 @@ jobs:
                 build-essential \
                 autoconf \
                 automake \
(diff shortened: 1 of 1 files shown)


jasonish:
See https://github.com/OISF/suricata/pull/6445.

PR #6446, opened 2021-10-05
fuzz: target must use the rules it parsed

catenacyber:

Describe changes:
- Make fuzz target cover more code

I noticed this with coverage report of `rs_mqtt_tx_has_type` being not covered

`DetectEngineReloadThreads` does not work for the fuzz targets as there is `no_of_detect_tvs = 0` as we did not register real threads and slots.

So, we force the flow worker module to use the newly detect engine context with all it needs


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6447

PR #6447, opened 2021-10-05
Fuzz usedectx 4125 v2

catenacyber:

Describe changes:
- Make fuzz target cover more code
- Fixes a leak quickly identified by the fuzz target

I noticed this with coverage report of `rs_mqtt_tx_has_type` being not covered

`DetectEngineReloadThreads` does not work for the fuzz targets as there is `no_of_detect_tvs = 0` as we did not register real threads and slots.

So, we force the flow worker module to use the newly detect engine context with all it needs

Modifies #6446 by adding the latest commit fixing the quickly found leak
Now falling back to known bugs

### DIFF
--- a/src/detect-engine-threshold.c
+++ b/src/detect-engine-threshold.c
@@ -727,8 +727,14 @@ void ThresholdHashAllocate(DetectEngineCtx *de_ctx)
  */
 void ThresholdContextDestroy(DetectEngineCtx *de_ctx)
 {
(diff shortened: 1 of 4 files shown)


catenacyber:
https://redmine.openinfosecfoundation.org/issues/4619 gets easily hit

victorjulien:
Merged in #6449, thanks!

PR #6455, opened 2021-10-06
queue.h: wrap the system sys/queue.h

jasonish:
Instead of using local implementations for the queue.h macro,
wrap the system provided queue.h and then adding missing
features as needed.

The idea is that Suricata when integrated with another library
that includes sys/queue.h can look at the same source of truth
for these macros.

But not all operating systems include a queue.h with the same
features, and some don't include it at all, like Windows. So
on Windows this will be a full implementation of all the queue.h
features Suricata needs.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -185,6 +185,7 @@
     AC_CHECK_HEADERS([libgen.h])
     AC_CHECK_HEADERS([mach/mach.h])
     AC_CHECK_HEADERS([stdatomic.h])
+    AC_CHECK_HEADERS([sys/queue.h])
 
     AC_CHECK_HEADERS([sys/socket.h net/if.h sys/mman.h linux/if_arp.h], [], [],
     [[#ifdef HAVE_SYS_SOCKET_H
--- a/src/queue.h
+++ b/src/queue.h
@@ -32,406 +32,49 @@
  *	@(#)queue.h	8.5 (Berkeley) 8/20/94
  */
 
-#ifndef	_SYS_QUEUE_H_
-#define	_SYS_QUEUE_H_
+// Disable clang-formatting as these implementations are copied from other
+// sources.
+//
+// clang-format off
 
-/*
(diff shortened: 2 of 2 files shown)


jasonish:
cc: @lukashino

victorjulien:
This works for all my tests, nice work!

The only thing I see is that the scan-build suppressions I had added are gone
```
  CC       log-pcap.o
log-pcap.c:887:24: warning: Use of memory after it is freed [unix.Malloc]
            if (remove(pf->filename) != 0) {
                       ^~~~~~~~~~~~
1 warning generated.

  CC       runmodes.o
runmodes.c:482:13: warning: Use of memory after it is freed [unix.Malloc]
        if (output->output_ctx != NULL && output->output_ctx->DeInit != NULL)
            ^~~~~~~~~~~~~~~~~~
1 warning generated.
```
I can live with this, but it would be nice if we could do something here.

jasonish:
> This works for all my tests, nice work!
> 
> The only thing I see is that the scan-build suppressions I had added (in [bbc0220](https://github.com/OISF/suricata/commit/bbc02205fb8c398f2d9890254c209e52aea831fd)) are gone, so these scan-build false positives are back:
> 
> ```
>   CC       log-pcap.o
> log-pcap.c:887:24: warning: Use of memory after it is freed [unix.Malloc]
>             if (remove(pf->filename) != 0) {
>                        ^~~~~~~~~~~~
> 1 warning generated.
> 
>   CC       runmodes.o
> runmodes.c:482:13: warning: Use of memory after it is freed [unix.Malloc]
>         if (output->output_ctx != NULL && output->output_ctx->DeInit != NULL)
>             ^~~~~~~~~~~~~~~~~~
> 1 warning generated.
> ```
> 
> I can live with this, but it would be nice if we could do something here.

I can't seem to suppress these from the calling code.  One thought is if in __clang_analyzer__ mode, then don't include `sys/queue.h` to force it to use our local versions that have these modifications made.

PR #6406, opened 2021-09-23
quic: Add QUIC App Layer - v8

sharksforarms:
Previous PR: https://github.com/OISF/suricata/pull/6334

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3440

Describe changes:
- Parses gquic version Q043-Q046 and logs a CYU hash
- Added a mix of unit and suricata-verify pcap tests
- Fails gracefully with unknown QUIC versions
- Other versions can be added later w/ pcaps + testing
- Logs a cyu hash and string to eve.json
  - Example ```{"timestamp":"2020-09-02T15:58:23.191651+0000","flow_id":1554154336021667,"pcap_cnt":1,"event_type":"quic","src_ip":"10.0.2.15","src_port":42855,"dest_ip":"172.217.2.109","dest_port":443,"proto":"UDP","quic":{"version":1362113590,"cyu":[{"hash":"a46560d4548108cf99308319b3b85346","string":"46,PAD-SNI-STK-VER-CCS-NONC-AEAD-UAID-SCID-TCID-PDMD-SMHL-ICSL-NONP-PUBS-MIDS-SCLS-KEXS-XLCT-CSCT-COPT-CCRT-IRTT-CFCW-SFCW"}]}}
```

References:
   - https://docs.google.com/document/d/1WJvyZflAO2pq77yOLbp9NsGjC1CHetAXV8I0fQe-B_U/edit
   - https://docs.google.com/document/d/1g5nIXAIkN_Y-7XJW5K45IblHd_L2f5LTaDUDwvZ5L6g/edit
   - https://www.slideshare.net/shigeki_ohtsu/quic-overview
   - https://tools.ietf.org/html/draft-ietf-quic-transport-17#section-19.8
   - https://github.com/salesforce/GQUIC_Protocol_Analyzer/blob/master/src/gquic-protocol.pac
   - cyu: https://engineering.salesforce.com/gquic-protocol-analysis-and-fingerprinting-in-zeek-a4178855d75f

suricata-verify-pr: 525

PR #6461, opened 2021-10-08
quic: Add QUIC App Layer - v9

sharksforarms:
Previous PR: https://github.com/OISF/suricata/pull/6406

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3440

Describe changes:
- Parses gquic version Q043-Q046 and logs a CYU hash
- Added a mix of unit and suricata-verify pcap tests
- Fails gracefully with unknown QUIC versions
- Other versions can be added later w/ pcaps + testing
- Logs a cyu hash and string to eve.json
Example:
```
{"timestamp":"2020-09-02T15:58:23.191651+0000","flow_id":1554154336021667,"pcap_cnt":1,"event_type":"quic","src_ip":"10.0.2.15","src_port":42855,"dest_ip":"172.217.2.109","dest_port":443,"proto":"UDP","quic":{"version":1362113590,"cyu":[{"hash":"a46560d4548108cf99308319b3b85346","string":"46,PAD-SNI-STK-VER-CCS-NONC-AEAD-UAID-SCID-TCID-PDMD-SMHL-ICSL-NONP-PUBS-MIDS-SCLS-KEXS-XLCT-CSCT-COPT-CCRT-IRTT-CFCW-SFCW"}]}}
```

References:
   - https://docs.google.com/document/d/1WJvyZflAO2pq77yOLbp9NsGjC1CHetAXV8I0fQe-B_U/edit
   - https://docs.google.com/document/d/1g5nIXAIkN_Y-7XJW5K45IblHd_L2f5LTaDUDwvZ5L6g/edit
   - https://www.slideshare.net/shigeki_ohtsu/quic-overview
   - https://tools.ietf.org/html/draft-ietf-quic-transport-17#section-19.8
   - https://github.com/salesforce/GQUIC_Protocol_Analyzer/blob/master/src/gquic-protocol.pac
   - cyu: https://engineering.salesforce.com/gquic-protocol-analysis-and-fingerprinting-in-zeek-a4178855d75f

suricata-verify-pr: 525


sharksforarms:
Hm, I still don't see fuzzing enabled in the CI.

sharksforarms:
> > Looking at the output, I wonder if we should log it in the way it appears in Wireshark: `Q046` instead of some numeric version of that like `1362113590`
> 
> Along these lines I also feel the `version` detection is a bit odd.
> 
> I would expect to be able to do something like `quic.version:Q046;` or perhaps `quic.version; content:"Q046";`. To allow of the numeric comparison we could have a `quic.version.raw`, which would expose the raw version bytes and one could match on these with `content`, but also `byte_test`, etc.

Just getting to this now,

In QUIC the version is bytes (https://www.rfc-editor.org/rfc/rfc8999.html#name-version) and might not be valid ASCII.

Is this still how we want to proceed? quic.version defaults to a ascii string and quic.version.raw would be the raw bytes?

jasonish:
It looks like there are no guarantees that the version is a printable string, so I don't think it should be logged as such. The number, as-is is probably the most future proof.

victorjulien:
> It looks like there are no guarantees that the version is a printable string, so I don't think it should be logged as such. The number, as-is is probably the most future proof.

I don't like it. I think doing what we do for TLS is better: we print the known names, and have a fall back otherwise 

https://github.com/OISF/suricata/blob/master/src/output-json-tls.c#L166
https://github.com/OISF/suricata/blob/master/src/app-layer-ssl.c#L394

jasonish:
> I don't like it. I think doing what we do for TLS is better: we print the known names, and have a fall back otherwise

Yeah, that could work with a translation table, and not just converting the bytes to a string.  And I guess just log the version as hex perhaps if not found in the table, as that might provide a clue as well rather than just a number.

sharksforarms:
> 
> 
> > I don't like it. I think doing what we do for TLS is better: we print the known names, and have a fall back otherwise
> 
> Yeah, that could work with a translation table, and not just converting the bytes to a string. And I guess just log the version as hex perhaps if not found in the table, as that might provide a clue as well rather than just a number.

Known versions might be tedious to update, what about bytes -> string if it's all ascii?

jasonish:
> Known versions might be tedious to update, what about bytes -> string if it's all ascii?

What I'd like to avoid is one case where the datatype is a string, and in another case its a number.  From what I've seen on these versions, a hex string might make more sense the raw string. But maybe we need 2 fields.

This only addresses the logging, I'm not sure about content matching on it.  Probably as a buffer where I could match on the hex.

sharksforarms:
In terms of next steps/version for this PR, how should I proceed?

catenacyber:
ping @jasonish @victorjulien ?

victorjulien:
Will push an updated branch soon.

victorjulien:
Continued in #6794

PR #6472, opened 2021-10-12
app-layer: clean and "close" all txs if protocol reaches error state

catenacyber:

Describe changes:

I do not see any easy way to get a pcap where this code will get used
That means a pcap with an app-layer protocol with 2 or more alive transactions, and then reaching an error state.
The best I see right now is to add a dummy patch for the HTTP2 parser to return `AppLayerResult::err()` in some specific case and then create a pcap with multiple HTTP2 requests/streams, no answers, and then triggering the dummy condition...
But we do not want a dummy patch...

### DIFF
--- a/src/flow.c
+++ b/src/flow.c
@@ -1131,7 +1131,7 @@ uint8_t FlowGetDisruptionFlags(const Flow *f, uint8_t flags)
     TcpSession *ssn = f->protoctx;
     TcpStream *stream = flags & STREAM_TOSERVER ? &ssn->client : &ssn->server;
 
-    if (stream->flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) {
+    if (stream->flags & (STREAMTCP_STREAM_FLAG_DEPTH_REACHED | STREAMTCP_FLAG_APP_LAYER_DISABLED)) {
         newflags |= STREAM_DEPTH;
     }
     /* todo: handle pass case (also for UDP!) */


src/flow.c
```diff
@@ -1131,7 +1131,7 @@ uint8_t FlowGetDisruptionFlags(const Flow *f, uint8_t flags)
     TcpSession *ssn = f->protoctx;
     TcpStream *stream = flags & STREAM_TOSERVER ? &ssn->client : &ssn->server;
 
-    if (stream->flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) {
+    if (stream->flags & (STREAMTCP_STREAM_FLAG_DEPTH_REACHED | STREAMTCP_FLAG_APP_LAYER_DISABLED)) {
```
inashivb, line 1134: Q: what happens if UDP parser gets into an error state?

catenacyber: Glad you ask.
I started discussing it in redmine cf https://redmine.openinfosecfoundation.org/issues/4318#note-4

Let's continue here :
I did not find any such thing as an error state for UDP (which seems relevant to the nature of UDP and TCP)

Furthermore, if I have this scenario over UDP :
- first request ok, creating first transaction
- second request, invalid returning a parser error
- first response, completing the first transaction

We see that having an error on the second request should not make us clean the first request... right ?

inashivb: I'm not sure. While I understand the out of order nature, don't many protocols have an "ID" even for UDP to match a request against the correct response even if they arrived weirdly? From what I understand that should be considered a complete tx, I know that all protocols do not follow this.. Do I make any sense to you?

catenacyber: > I'm not sure. While I understand the out of order nature, don't many protocols have an "ID" even for UDP to match a request against the correct response even if they arrived weirdly? 

Yes, even so...

> From what I understand that should be considered a complete tx, I know that all protocols do not follow this.. Do I make any sense to you?

Yes, even so :
I do not think it makes sens to have such a thing an error state for protocols over UDP.
Or maybe, just dismiss the app-layer protocol detected if we only get errors out of  `AppLayerHandleUdp` ?

catenacyber: @jasonish do you have an opinion on this ? (error state for protocols over UDP)

victorjulien:
Would a HTTP1 pipelining do? a couple of HTTP1.1 requests followed by an error condition?

catenacyber:
> Would a HTTP1 pipelining do? a couple of HTTP1.1 requests followed by an error condition?

Indeed, but the difficulty is how to put the HTTP1 parser in an error condition ?..
Do you see an easy way ?
Otherwise I can dig into it

victorjulien:
Looks like if libhtp returns `HTP_STREAM_ERROR` we should get there. Not sure if that is easy though

victorjulien:
> > Looks like if libhtp returns `HTP_STREAM_ERROR` we should get there. Not sure if that is easy though
> 
> Fuzzing manages to cover some cases... But they should be unreachable... So that means we have more bugs, and that we cannot craft a case out of this because the test will break once we fix the bugs...

If no traffic condition ever should be able to put the parser in an error state then we may have to resort to doing a hack of some kind. `GET /suricata::error HTTP/1.1...`, where this would only be enabled if some configure option is set?

catenacyber:
> If no traffic condition ever should be able to put the parser in an error state 

This is for the HTTP1 parser...
There may be another protocol parser who is able to get in an error state...

catenacyber:
SSH record with 0 length should do the trick 🎊

catenacyber:
> SSH record with 0 length should do the trick 🎊

But SSH will not have multiple txs to close 😢

PR #6476, opened 2021-10-14
rust/app-layer: generic tx iterator

jasonish:
Create a generate TX iterator that app-layer parsers can use
by only providing a `get_id` method on transactions and
implementing a small FFI shim.

This also updates DNS to use an iterator as an example.


suricata-qa:
WARNING:
|--|--|--|--|
| | tlpr1_stats_chk |

Pipeline 4588

PR #6477, opened 2021-10-14
Range valid 4754 v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4754
https://redmine.openinfosecfoundation.org/issues/4757

Describe changes:
- Consider invalid a range where the start is after the end (otherwise we get malloc problems)
- Fixes the closing case for a range with overlap and expected new bytes but we did not get the new bytes...


victorjulien:
Merged in #6478, thanks!

PR #4766, opened 2020-04-01
doc/devguide: Submission and style

jlucovsky:
Continuation of #4759

This commit adds code submission and coding style guidelines to the
devguide. Most of the material is a straight port from the wiki but
there have been some content modifications and additions.

Describe changes:
- Add TODO for empty sections

PR #6500, opened 2021-10-20
flow/bypass: use_cnt desync'd on bypassed flows

victorjulien:
Locally bypassed flows had unsafe updates to `Flow::use_cnt` leading to a race
issue. For a packet it would do the flow lookup, attach the flow to the packet,
increment the `use_cnt`. Then it would detect that the flow is in the bypass
state, and unlock it while holding a reference (so alos not decrementing the
`use_cnt`). When the packet was then returned to the packet pool, the flow would
be disconnected from the packet, which would decrement `use_cnt` without holding
the flow lock.

This patch addresses this issue by disconnecting the flow from the packet
immediately when the bypassed state is detected. This moves the `use_cnt`
decrement to within the lock.

Bug: #4766.

https://redmine.openinfosecfoundation.org/issues/4766

### DIFF
--- a/src/flow-worker.c
+++ b/src/flow-worker.c
@@ -217,15 +217,23 @@ static inline TmEcode FlowUpdate(ThreadVars *tv, FlowWorkerThreadData *fw, Packe
     int state = p->flow->flow_state;
     switch (state) {
 #ifdef CAPTURE_OFFLOAD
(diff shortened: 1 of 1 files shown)


victorjulien:
Part of #6538

PR #6504, opened 2021-10-21
pcrexform: use substring and not whole match

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4771

Describe changes:
- pcrexform: use substring and not whole match

We need a test that would have failed...


victorjulien:
Can you create a test? It seems substring capture works for me.

PR #6508, opened 2021-10-22
fix krb5_msg_type: Unable to detect AS-REQ, TGS-REQ and KRB-ERROR

zer1t0:

https://redmine.openinfosecfoundation.org/issues/4529

Related SV test: https://github.com/OISF/suricata-verify/pull/572

The problem is that the krb5_msg_type doesn't trigger properly for:
- AS-REQ (krb5_msg_type: 10): This rule matches the KRB-ERROR message that follows an AS-REQ, which is not the expected behaviour. 
- TGS-REQ (krb5_msg_type: 12): Same behaviour as AS-REQ.
-  KRB-ERROR (krb5_msg_type: 30) It's not triggered at all.

To solve the problem the following actions were taken in `KRB5State.parse`:
- Add transaction when parsing AS-REQ message
- Add transaction when parsing TGS-REQ message
- Change the message type of KRB-ERROR to MessageType::KRB_ERROR

suricata-verify-pr: 572

### DIFF
--- a/rust/src/krb/krb5.rs
+++ b/rust/src/krb/krb5.rs
@@ -68,6 +68,9 @@ pub struct KRB5Transaction {
     /// Error code, if request has failed
     pub error_code: Option<ErrorCode>,
 
+    /// Message type of request. For using in responses.
+    pub req_type: Option<MessageType>,
+
     /// The internal transaction id
     id: u64,
 
@@ -128,6 +131,9 @@ impl KRB5State {
                         if let Ok((_,kdc_rep)) = res {
                             let mut tx = self.new_tx();
                             tx.msg_type = MessageType::KRB_AS_REP;
+                            if self.req_id > 0 {
+                                tx.req_type = Some(MessageType(self.req_id.into()));
+                            }
                             tx.cname = Some(kdc_rep.cname);
                             tx.realm = Some(kdc_rep.crealm);
                             tx.sname = Some(kdc_rep.ticket.sname);
@@ -157,6 +163,9 @@ impl KRB5State {
                         if let Ok((_,kdc_rep)) = res {
                             let mut tx = self.new_tx();
                             tx.msg_type = MessageType::KRB_TGS_REP;
+                            if self.req_id > 0 {
+                                tx.req_type = Some(MessageType(self.req_id.into()));
+                            }
                             tx.cname = Some(kdc_rep.cname);
                             tx.realm = Some(kdc_rep.crealm);
                             tx.sname = Some(kdc_rep.ticket.sname);
@@ -178,6 +187,9 @@ impl KRB5State {
                         let res = krb5_parser::parse_krb_error(i);
                         if let Ok((_,error)) = res {
                             let mut tx = self.new_tx();
+                            if self.req_id > 0 {
+                                tx.req_type = Some(MessageType(self.req_id.into()));
+                            }
                             tx.msg_type = MessageType::KRB_ERROR;
                             tx.cname = error.cname;
                             tx.realm = error.crealm;
@@ -245,6 +257,7 @@ impl KRB5Transaction {
             sname: None,
             etype: None,
             error_code: None,
+            req_type: None,
             id: id,
             de_state: None,
             events: std::ptr::null_mut(),
--- a/rust/src/krb/krb5.rs
+++ b/rust/src/krb/krb5.rs
@@ -178,7 +178,7 @@ impl KRB5State {
                         let res = krb5_parser::parse_krb_error(i);
                         if let Ok((_,error)) = res {
                             let mut tx = self.new_tx();
-                            tx.msg_type = MessageType(self.req_id as u32);
+                            tx.msg_type = MessageType::KRB_ERROR;
                             tx.cname = error.cname;
                             tx.realm = error.crealm;
                             tx.sname = Some(error.sname);
--- a/rust/src/krb/krb5.rs
+++ b/rust/src/krb/krb5.rs
@@ -140,6 +140,16 @@ impl KRB5State {
                         self.req_id = 0;
                     },
                     12 => {
+                        let req = krb5_parser::parse_tgs_req(i);
+                        if let Ok((_,kdc_req)) = req {
+                            let mut tx = self.new_tx();
+                            tx.msg_type = MessageType::KRB_TGS_REQ;
+                            tx.cname = kdc_req.req_body.cname;
+                            tx.realm = Some(kdc_req.req_body.realm);
+                            tx.sname = kdc_req.req_body.sname;
+                            tx.etype = None;
+                            self.transactions.push(tx);
+                        };
                         self.req_id = 12;
                     },
                     13 => {
--- a/rust/src/krb/krb5.rs
+++ b/rust/src/krb/krb5.rs
@@ -132,6 +132,8 @@ impl KRB5State {
                             let mut tx = self.new_tx();
                             tx.msg_type = MessageType::KRB_AS_REP;
                             if self.req_id > 0 {
+                                // set request type only if previous message
+                                // was a request
                                 tx.req_type = Some(MessageType(self.req_id.into()));
                             }
                             tx.cname = Some(kdc_rep.cname);
@@ -164,6 +166,8 @@ impl KRB5State {
                             let mut tx = self.new_tx();
                             tx.msg_type = MessageType::KRB_TGS_REP;
                             if self.req_id > 0 {
+                                // set request type only if previous message
+                                // was a request
                                 tx.req_type = Some(MessageType(self.req_id.into()));
                             }
                             tx.cname = Some(kdc_rep.cname);
@@ -188,6 +192,8 @@ impl KRB5State {
                         if let Ok((_,error)) = res {
                             let mut tx = self.new_tx();
                             if self.req_id > 0 {
+                                // set request type only if previous message
+                                // was a request
                                 tx.req_type = Some(MessageType(self.req_id.into()));
                             }
                             tx.msg_type = MessageType::KRB_ERROR;
--- a/rust/src/krb/krb5.rs
+++ b/rust/src/krb/krb5.rs
@@ -111,6 +111,16 @@ impl KRB5State {
                 if hdr.class != BerClass::Application { return 0; }
                 match hdr.tag.0 {
                     10 => {
+                        let req = krb5_parser::parse_as_req(i);
+                        if let Ok((_,kdc_req)) = req {
+                            let mut tx = self.new_tx();
+                            tx.msg_type = MessageType::KRB_AS_REQ;
+                            tx.cname = kdc_req.req_body.cname;
+                            tx.realm = Some(kdc_req.req_body.realm);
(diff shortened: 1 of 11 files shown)


rust/src/krb/log.rs
```diff
@@ -24,8 +24,12 @@ fn krb5_log_response(jsb: &mut JsonBuilder, tx: &mut KRB5Transaction) -> Result<
 {
     match tx.error_code {
         Some(c) => {
-            jsb.set_string("msg_type", "KRB_ERROR")?;
-            jsb.set_string("failed_request", &format!("{:?}", tx.msg_type))?;
+            jsb.set_string("msg_type", &format!("{:?}", tx.msg_type))?;
+            if let Some(req_type) = tx.req_type {
+                jsb.set_string("failed_request", &format!("{:?}", req_type))?;
+            } else {
+                jsb.set_string("failed_request", "UNKNOWN")?;
```
catenacyber, line 31: How can we get here ?

rust/src/krb/krb5.rs
```diff
@@ -130,13 +146,26 @@ impl KRB5State {
                         self.req_id = 0;
                     },
                     12 => {
+                        let req = krb5_parser::parse_tgs_req(i);
```
catenacyber, line 149: I do not know much about Kerberos.
What does this 12 mean ?
Any link to a RFC ?

rust/src/krb/krb5.rs
```diff
@@ -111,13 +114,26 @@ impl KRB5State {
...
+                        };
                         self.req_id = 10;
                     },
                     11 => {
                         let res = krb5_parser::parse_as_rep(i);
                         if let Ok((_,kdc_rep)) = res {
                             let mut tx = self.new_tx();
                             tx.msg_type = MessageType::KRB_AS_REP;
+                            if self.req_id > 0 {
+                                tx.req_type = Some(MessageType(self.req_id.into()));
```
catenacyber, line 135: Could you please add a comment about this ?
Why using the condition `if self.req_id > 0`
My understanding is that what got parsed as a request id, is not in this case a request id but a request type...
Why did we parse it as a request id in the first place ?

rust/src/krb/krb5.rs
```diff
@@ -158,7 +187,10 @@ impl KRB5State {
                         let res = krb5_parser::parse_krb_error(i);
                         if let Ok((_,error)) = res {
                             let mut tx = self.new_tx();
-                            tx.msg_type = MessageType(self.req_id as u32);
+                            if self.req_id > 0 {
```
catenacyber, line 190: Why only if `self.req_id > 0` ?

catenacyber (commented):
Thanks for this work.
I left some questions

rust/src/krb/krb5.rs
zer1t0: Hi,

You are totally correct, the req_id is the request type, however that was the name of the variable in the code. I wanted to do as few changes in the code as I could so I keep this variable name (I also needed to introduce a field to keep track of request_type in responses, so I already use the word `req_type`, more accurate). But yeah, it requires a rename, 
 probably in a improvement PR.

Related to the condition `if self.req_id > 0`: in the code self.req_id is only set different to zero for requests, so you only get the request type when the previous one was a request, in other case, you keep the default None Option.

rust/src/krb/krb5.rs
zer1t0: in the code self.req_id is only set different to zero for requests, so you only get the request type when the previous one was a request, in other case, you keep the default None Option.

rust/src/krb/log.rs
zer1t0: You arrive to this conditional when a KRB-ERROR message is logged, if the previous request type is set, then you return the request type. However, in case of the request type being None (because start capturing in the middle of an UDP connection, for example) you return the UNKNOWN value.

zer1t0:
> 

Hi, I try to answer your questions. Keep in mind that this is a fix request so I tried to do just fix changes, without varying the style of the code. I think you point some code improvement issues that requires consideration, but those require another kind of PR.

rust/src/krb/log.rs
catenacyber: Thanks. Maybe this is worth a comment in the code

rust/src/krb/krb5.rs
catenacyber: Could you create a redmine ticket to track this improvement ?

rust/src/krb/krb5.rs
catenacyber: Thanks for the explanation. This naming deserves a comment

rust/src/krb/krb5.rs
catenacyber: ok

rust/src/krb/krb5.rs
zer1t0: Here it is https://redmine.openinfosecfoundation.org/issues/4950

rust/src/krb/krb5.rs
zer1t0: Done in https://github.com/OISF/suricata/pull/6508/commits/abbef13b12b58e8b6c2fb7fff6b589d97cfacde5

rust/src/krb/log.rs
zer1t0: Done in https://github.com/OISF/suricata/pull/6508/commits/4642ecf72cf88f99df193b1a029f0f2ffb17850a

rust/src/krb/krb5.rs
```diff
@@ -111,13 +114,28 @@ impl KRB5State {
...
                 match hdr.tag.0 {
                     10 => {
+                        let req = krb5_parser::parse_as_req(i);
+                        if let Ok((_,kdc_req)) = req {
+                            let mut tx = self.new_tx();
+                            tx.msg_type = MessageType::KRB_AS_REQ;
+                            tx.cname = kdc_req.req_body.cname;
+                            tx.realm = Some(kdc_req.req_body.realm);
+                            tx.sname = kdc_req.req_body.sname;
+                            tx.etype = None;
```
catenacyber, line 124: Is not this redundant ?

catenacyber: I mean is `tx.etype` not None before that ?

catenacyber:
Changes look good to me...
I guess some commits can be squashed together for the final version
Let's wait for CI

catenacyber:
Do you understand why CI fails with suricate-verify ?

zer1t0:
> Do you understand why CI fails with suricate-verify ?

It is failing in `krb5-request-frag-log` test with error:
```
===> krb5-request-frag-log: Sub test #1: FAIL : expected 2 matches; got 4 for filter {'count': 2, 'match': {'event_type': 'alert', 'alert.signature_id': 1}}
===> krb5-request-frag-log: Sub test #2: FAIL : expected 2 matches; got 4 for filter {'count': 2, 'match': {'event_type': 'alert', 'alert.signature_id': 2}}
```

It seems these rules are matching many more times than expected:
```
alert krb5 any any -> any any (krb5.sname; content:"test"; sid:1;)
alert krb5 any any -> any any (krb5.cname; content:"user01"; sid:2;)
```

That is normal with the fix, since now these rules are also matching with kerberos requests (AS-REQ and TGS-REQ), so the number of matches increases.

rust/src/krb/krb5.rs
zer1t0: Yes, you are right, it is None previously too.

catenacyber:
Ok, so could you change the S-V test so that they the right number of matches ?

catenacyber:
And thanks for the explanation :-)

zer1t0:
> Ok, so could you change the S-V test so that they the right number of matches ?

Done in https://github.com/OISF/suricata-verify/pull/572/commits/f02f87fdd5b6fb811a9de6f437fb2472830510c5

catenacyber:
Rerunning CI...

catenacyber:
@jasonish how do you feel about backports for this ?
I guess this is a feature addition and we do not want to back port it.
But then, we will need the Suricata-verify test to tell 2 alerts up to version 6, and 4 alerts from version 7 onwards...

jasonish:
> @jasonish how do you feel about backports for this ? I guess this is a feature addition and we do not want to back port it. But then, we will need the Suricata-verify test to tell 2 alerts up to version 6, and 4 alerts from version 7 onwards...

This reads more like a bug than a feature to me.

catenacyber:
Said like that, it may sound like a bug, but this is because we were only parsing half the messages... So ?

zer1t0:
It is a bug since it is expected to match the message you are referring to in the `krb5_msg_type`, but this doesn't happen in half of the cases, not even in the krb5_msg_type [documentation](https://suricata.readthedocs.io/en/suricata-6.0.4/rules/kerberos-keywords.html#krb5-msg-type) example.

catenacyber (approved):
Looks good to me.

@jasonish do you know what is the process to merge this ?
That is this fix needs to be merged in master, master6, master5 and Suricata-verify.
The CI will be red when we land one fix until we merge the other ones

victorjulien (changes requested):
Can you update the git commits to prefix them with `krb5`?

victorjulien:
> Looks good to me.
> 
> @jasonish do you know what is the process to merge this ? That is this fix needs to be merged in master, master6, master5 and Suricata-verify. The CI will be red when we land one fix until we merge the other ones

wrt backports I would like to understand better how much more log volume this will generate for the new transactions that are generated.

zer1t0:
> > Looks good to me.
> > @jasonish do you know what is the process to merge this ? That is this fix needs to be merged in master, master6, master5 and Suricata-verify. The CI will be red when we land one fix until we merge the other ones
> 
> wrt backports I would like to understand better how much more log volume this will generate for the new transactions that are generated.

Right now only the responses generate a transaction. With the fix, also the requests will generate a transaction, so we would have twice as many transactions. Since Kerberos is just a authentication protocol where each connection usually exchanges between 2 and 4 messages (2 requests and 2 responses), it shouldn't affect negatively to the performance.

victorjulien:
> > > Looks good to me.
> > > @jasonish do you know what is the process to merge this ? That is this fix needs to be merged in master, master6, master5 and Suricata-verify. The CI will be red when we land one fix until we merge the other ones
> > 
> > 
> > wrt backports I would like to understand better how much more log volume this will generate for the new transactions that are generated.
> 
> Right now only the responses generate a transaction. With the fix, also the requests will generate a transaction, so we would have twice as many transactions. Since Kerberos is just a authentication protocol where each connection usually exchanges between 2 and 4 messages (2 requests and 2 responses), it shouldn't affect negatively to the performance.

I've found a pcap where I have real krb5 traffic. In master it generates 12k records, in this branch 24k. So this is significant increase in log volume. I don't think we should do that in a backport. For master its fine.

victorjulien:
Btw please submit a new PR against master when implementing changes.

zer1t0:
New PR: https://github.com/OISF/suricata/pull/6926

PR #6509, opened 2021-10-22
fix: smb dce_iface doesn't match rpc requests/responses

zer1t0:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4767](https://redmine.openinfosecfoundation.org/issues/4767)

The bug:
The smb dce_iface keyword must match for all those dcerpc requests and responses sent in the context of the given interface. They are not matching due to two problems:
- In `rs_smb_tx_get_dce_iface` the `x.req_cmd` is erroneously compared with 1. https://github.com/OISF/suricata/blob/master/rust/src/smb/detect.rs#L175
- In `smb_write_dcerpc_record`, the current bind interfaces are deleted by any non bind message. https://github.com/OISF/suricata/blob/master/rust/src/smb/dcerpc.rs#L307

Related SV test: https://github.com/OISF/suricata-verify/pull/569

Describe changes:
- compare `x.req_cmd` with the  DCERPC_TYPE_REQUEST constant
- protect the `state.dcerpc_ifaces ` from arbitrary writings

PR #6511, opened 2021-10-22
fix: dcerpc dce_iface doesn't match the rpc requests/reponses

zer1t0:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4769

SV test: https://github.com/OISF/suricata-verify/pull/571

The bug:
The dcerpc dce_iface keyword just match the packet following the bind. Only the next request after the rpc is sent will match. However the expected behaviour it that all the rpc requests/responses sent under the context of the given interface would match.

This is happening because the interfaces are been deleted from dcerpc state after the first request in the [handle_bind_cache](https://github.com/OISF/suricata/blob/master/rust/src/dcerpc/dcerpc.rs#L546) function. Not sure why this function was added, since the bindack is the context of the state, not a cached value. My guess, a misinterpretation how the protocol works, since I've check this function was added after moving the code to rust (maybe following the example of the smb handling, in which also [deletes the bind state erroneously](https://github.com/OISF/suricata/pull/6509)).

Anyway, the fact is that the messages are not matching so I removed the function.

As collateral effect the SV test dcerpc-iface-01 is now failing, since the dce_iface now matches both the request and response (not only the first request after bind).


inashivb (commented):
Hi @zer1t0 !
Thank you very much for you contributions to our project! :)
Could you please share a link to the documentation where this behavior is explained? I sure have understood it wrong and am unable to pinpoint the correct behavior in the Open Group docs.

zer1t0:
Hi @inashivb!
Thank you for your work in this great project too!!

Related to the documentation, it is quite abstract and difficult to follow in my opinion. Generally I learn more by inspecting protocol packets with Wireshark.

But related to documentation. In the Open Group c706 the following is indicated:

In 2.2.1 Binding-related Operations, indicates that one category of binding operations are those that "operations that establish internal call routing information for the server." (The other are to establish the protocol which is not relevant here). And the following statement can be found:

Operations in the second category establish a set of mappings that the server can use to route calls internally to the appropriate manager routine. This routing is based on the **interface** and version, **operation** and any **object** requested by the call.

It indicates that server routes (to call methods) are based on the operation, interface and object.
- **Operation**: To indicate the method to call, and operation number is specified as indicated in the second step of 2.3.3.2 (*Client Binding Steps*).
- **Interface**: An interface is a set of remotely callable operations offered by a server and invokable by clients. (2.1.1.1)
- **Object**: Is the manager that implements the interface, as stated in section *Interface and Manager Selection* of 2.3.3.3. It is not mandatory, can be nil.

To call a method, a client must send a **request message** as defined in 2.6.4.9, that contains these identifiers:
- **opnum**: The opnum field identifies the operation being invoked within the interface.
- **p_cont_id** (Context ID in *Wireshark*) : The p_cont_id field holds a presentation context identifier that identifies the data representation and interface, as defined in 12.6.3.4 (*Context Identifiers*).
- **object**: The object field is contained if the PFC_OBJECT_UUID is set. (Could be interesting to create a keyword *dce_object* for matching this UUID)

Therefore, to get the correct method to invoke, the **server must map the context to the correct interface**. This is negotiated by the bind request

Interfaces are first negotiated using the bind message (12.6.4.3), contained in the p_context_elem array.  Then they are accepted or rejected using the bind_ack message (12.6.4.4).

Once these contexts are established, both client and server can use the context id, which is the index of the p_context_elem array, to refer the interface they are using.

Moreover, in the middle of the connection, the context can be changed with the alter_context message.

This is way suricata shouldn't delete the bindack attribute, that contains the contexts, used by [match_backuuid](https://github.com/OISF/suricata/blob/a87c7e5c08b027d104bfa00c6f7858398c2d63a6/rust/src/dcerpc/detect.rs#L103). This is the only way to know the interface a request message is referring to.

inashivb:
> Hi @inashivb! Thank you for your work in this great project too!!
> 
> Related to the documentation, it is quite abstract and difficult to follow in my opinion. Generally I learn more by inspecting protocol packets with Wireshark.
> 
> But related to documentation. In the Open Group c706 the following is indicated:
> 
> In 2.2.1 Binding-related Operations, indicates that one category of binding operations are those that "operations that establish internal call routing information for the server." (The other are to establish the protocol which is not relevant here). And the following statement can be found:
> 
> Operations in the second category establish a set of mappings that the server can use to route calls internally to the appropriate manager routine. This routing is based on the **interface** and version, **operation** and any **object** requested by the call.
> 
> It indicates that server routes (to call methods) are based on the operation, interface and object.
> 
>     * **Operation**: To indicate the method to call, and operation number is specified as indicated in the second step of 2.3.3.2 (_Client Binding Steps_).
> 
>     * **Interface**: An interface is a set of remotely callable operations offered by a server and invokable by clients. (2.1.1.1)
> 
>     * **Object**: Is the manager that implements the interface, as stated in section _Interface and Manager Selection_ of 2.3.3.3. It is not mandatory, can be nil.
> 
> 
> To call a method, a client must send a **request message** as defined in 2.6.4.9, that contains these identifiers:
> 
>     * **opnum**: The opnum field identifies the operation being invoked within the interface.
> 
>     * **p_cont_id** (Context ID in _Wireshark_) : The p_cont_id field holds a presentation context identifier that identifies the data representation and interface, as defined in 12.6.3.4 (_Context Identifiers_).
> 
>     * **object**: The object field is contained if the PFC_OBJECT_UUID is set. (Could be interesting to create a keyword _dce_object_ for matching this UUID)
> 
> 
> Therefore, to get the correct method to invoke, the **server must map the context to the correct interface**. This is negotiated by the bind request
> 
> Interfaces are first negotiated using the bind message (12.6.4.3), contained in the p_context_elem array. Then they are accepted or rejected using the bind_ack message (12.6.4.4).
> 
> Once these contexts are established, both client and server can use the context id, which is the index of the p_context_elem array, to refer the interface they are using.
> 
> Moreover, in the middle of the connection, the context can be changed with the alter_context message.
> 
> This is way suricata shouldn't delete the bindack attribute, that contains the contexts, used by [match_backuuid](https://github.com/OISF/suricata/blob/a87c7e5c08b027d104bfa00c6f7858398c2d63a6/rust/src/dcerpc/detect.rs#L103). This is the only way to know the interface a request message is referring to.

Thank you very much for the explanation and the s-v test. I think this makes sense to me. I'll also ask @victorjulien to chime in as the same behavior is going to be reflected in smb/dcerpc.

A few commit message change requests from my end:
1. You gave a really good description in this PR, it would be great if the same description was added to the commit message.
2. Please precede the commit message with the relevant part of the code e.g. `dcerpc: avoid delete the rpc state interface context`.
3. Please add the ticket number (redmine) at the end of commit body.
A very good example: https://github.com/OISF/suricata/pull/6564/commits/27f27f91f339d93f30399b7242c08619e636e51e

If your patch is approved by Victor as well, please fix the failing dce-iface test or let me know if you wouldn't have time to do that. Thank you very much for putting effort into this!

Note: We'll approve our CI builds in a while, I'll keep an eye if something other than the known s-v test failure jumps up.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -543,19 +543,6 @@ impl DCERPCState {
         None
     }
 
-    pub fn handle_bind_cache(&mut self, call_id: u32, is_response: bool) {
-        if self.clear_bind_cache == true {
-            self.bind = None;
-            self.bindack = None;
-        }
-        if self.prev_tx_call_id == call_id && is_response == true {
-            self.clear_bind_cache = true;
```
inashivb, line 552: Perhaps, we can get rid of this field from DCERPCState struct as it does not seem to be used elsewhere.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -543,19 +543,6 @@ impl DCERPCState {
...
-        } else {
-            self.clear_bind_cache = false;
-        }
-        self.prev_tx_call_id = call_id;
```
inashivb, line 556: This too.

zer1t0:
Done.
- Changed commit name, added description and reference to ticket
- Fixed the SV test dcerpc-iface-01
- Removed fields `prev_tx_call_id` and `clear_bind_cache` from state

zer1t0:
Any news related to this?

victorjulien:
Our policy is to close a PR and open a new one to address issues or do rebases. Otherwise things get messy. Was planning to review and test soon, but will now wait for a new clear PR.

PR #6563, opened 2021-11-02
detect/secresult: convert unittest to FAIL/PASS APIs

TheKharleeci:

Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/4055

Ticket: #4055

Previous PR: #6562

Describe changes:

- Update detect-rfb-secresult unit tests to use FAIL/PASS APIs
- Update Copyright Year

### DIFF
--- a/src/detect-rfb-secresult.c
+++ b/src/detect-rfb-secresult.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2020 Open Information Security Foundation
+/* Copyright (C) 2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -268,14 +268,13 @@ void DetectRfbSecresultFree(DetectEngineCtx *de_ctx, void *de_ptr)
  */
 static int RfbSecresultTestParse01 (void)
 {
-    DetectRfbSecresultData *de = NULL;
-    de = DetectRfbSecresultParse("fail");
-    if (de) {
-        DetectRfbSecresultFree(NULL, de);
-        return 1;
-    }
+    DetectRfbSecresultData *de = DetectRfbSecresultParse("fail");
 
-    return 0;
+    FAIL_IF_NULL(de);
+
+    DetectRfbSecresultFree(NULL, de);
+
+    PASS;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/detect-rfb-secresult.c
```diff
@@ -1,4 +1,4 @@
-/* Copyright (C) 2020 Open Information Security Foundation
+/* Copyright (C) 2021 Open Information Security Foundation
```
jufajardini, line 1: Thanks for taking the time to also update the copyright year :)

One thing is that they must be updated in such a way that it covers their existence, if I can put it that way. That means that here we should have `2020-2021`, not erasing `2020`.

src/detect-rfb-secresult.c
```diff
@@ -268,14 +268,13 @@ void DetectRfbSecresultFree(DetectEngineCtx *de_ctx, void *de_ptr)
  */
```
jufajardini, line 268: It is preferred to remove the retvals, now.
(so, deleting lines 266 and 267)

src/detect-rfb-secresult.c
```diff
@@ -286,14 +285,13 @@ static int RfbSecresultTestParse01 (void)
...
-    de = DetectRfbSecresultParse("invalidopt");
-    if (de) {
-        DetectRfbSecresultFree(NULL, de);
-        return 0;
-    }
+    DetectRfbSecresultData *de = DetectRfbSecresultParse("invalidopt");
+
+    FAIL_IF_NOT_NULL(de);
+
+    DetectRfbSecresultFree(NULL, de);
```
jufajardini, line 292: One review I received just the other day and which I believe applies here is that we can remove this `DetectRfbSecresultFree` call. The reasoning behind it is that if it is not null, we won't reach this call, and if we reach it... then it should be null, already...

jufajardini (changes requested):
Thanks for this PR, a few nit comments.

I'm not sure if it is ok or not to suppress the `rfb` here in the commit message, so that might need change...

jasonish:
> I'm not sure if it is ok or not to suppress the `rfb` here in the commit message, so that might need change...

It would be better if it was still in there: `detect-rfb-secresult: ...`

TheKharleeci:
Followed by: #6565

PR #6554, opened 2021-11-01
Delete PASS_IF macro from FAIL/PASS API - v2

jufajardini:
Link to redmine tickets:
https://redmine.openinfosecfoundation.org/issues/4795
and
https://redmine.openinfosecfoundation.org/issues/4021

Previous PR: https://github.com/OISF/suricata/pull/6546

Describe changes:
- replace `FAIL_IF` for `FAIL_IF_NOT_NULL`, and `FAIL_IF_NOT` with `FAIL_IF_NULL` calls

PR #6570, opened 2021-11-03
Delete PASS_IF macro from FAIL/PASS API - v3

jufajardini:
Link to redmine tickets:
https://redmine.openinfosecfoundation.org/issues/4795
and
https://redmine.openinfosecfoundation.org/issues/4021

Previous PR: #6554 

Describe changes:
- Remove unnecessary `DetectDsizeFree` calls where appropriate
- Undo unwelcome clang change
- Condense tests that were checking for the same `Dsize` values
- Condense tests that checked for invalid values into one

### DIFF
--- a/src/detect-dsize.c
+++ b/src/detect-dsize.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2020 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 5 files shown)


src/detect-dsize.c
```diff
@@ -896,80 +627,68 @@ static int DsizeTestParse21(void)
...
-static int DsizeTestParse23(void)
+static int DsizeTestParse11(void)
 {
-    DetectDsizeData *dd = NULL;
-    dd = DetectDsizeParse("1!");
-    if (dd) {
-        DetectDsizeFree(NULL, dd);
-        FAIL;
+    const char *strings[] = { "A", ">10<>10", "<>10", "1<>", "", " ", "2<>1", "1!", NULL };
+    for (int i = 0; strings[i]; i++) {
```
jufajardini, line 647: Not sure if this format is welcome for tests...

jufajardini (commented):
Doesn't look great for reviewing, should I make the changes more incrementally?

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.wrk.flows_evicted | 5308489 | 4693221 | 113.11% |
| .flow.mgr.rows_maxlen | 115 | 539 | 21.34% | 

Pipeline 4726

src/detect-dsize.c
jasonish: I'm OK with these "table driven tests". The only issue is that if it did fail we have no context to show which one failed. But some debug statements could easily help with that when debugging why it failed I guess.

victorjulien: I'm good with the table approach. Think its easy enough to find where it fails when it does.

An alternative approach could be something like the macros here  https://github.com/OISF/suricata/blob/master/src/tests/stream-tcp-list.c#L65

victorjulien (approved):

src/detect-dsize.c
jufajardini: Oh, I had not crossed those before, thanks!!

victorjulien:
Merged in #6581, thanks!

PR #6593, opened 2021-11-10
app-layer: reduce boilerplate code - v3

jasonish:
Builds on https://github.com/OISF/suricata/pull/6519.  While looking at making a generic function for `DetectEngineState`, it appeared that we could just include that state is the `AppLayerTxData` which allows for the removal of all boilerplate for `DetectEngineState`. This was easy because both fields are required by the app-layer API (at least in Rust).

I have this little feeling that there was a reason we couldn't put `DetectEngineState` in this common data structure, but all tests are passing.

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -118,9 +118,6 @@ typedef struct AppLayerParserProtoCtx_
     int (*StateGetEventInfo)(const char *event_name,
                              int *event_id, AppLayerEventType *event_type);
 
-    DetectEngineState *(*GetTxDetectState)(void *tx);
-    int (*SetTxDetectState)(void *tx, DetectEngineState *);
-
     AppLayerTxData *(*GetTxData)(void *tx);
     bool (*ApplyTxConfig)(void *state, void *tx, int mode, AppLayerTxConfig);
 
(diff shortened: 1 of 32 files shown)


victorjulien:
"I have this little feeling that there was a reason we couldn't put DetectEngineState in this common data structure, but all tests are passing."

I don't think there is a reason other than that we simply didn't get to it yet.

src/app-layer-parser.c
```diff
@@ -1162,16 +1147,18 @@ int AppLayerParserSupportsFiles(uint8_t ipproto, AppProto alproto)
 DetectEngineState *AppLayerParserGetTxDetectState(uint8_t ipproto, AppProto alproto, void *tx)
 {
     SCEnter();
-    DetectEngineState *s = alp_ctx.ctxs[FlowGetProtoMapping(ipproto)][alproto].GetTxDetectState(tx);
+    AppLayerTxData *d = AppLayerParserGetTxData(ipproto, alproto, tx);
```
victorjulien, line 1150: I think this should be taken one step further: get rid of `AppLayerParserGetTxDetectState`. The detect engine should have direct access to `AppLayerTxData`.

src/app-layer-parser.c
```diff
@@ -1162,16 +1147,18 @@ int AppLayerParserSupportsFiles(uint8_t ipproto, AppProto alproto)
...
     SCReturnPtr(s, "DetectEngineState");
 }
 
 int AppLayerParserSetTxDetectState(const Flow *f,
                                    void *tx, DetectEngineState *s)
 {
     SCEnter();
-    int r = alp_ctx.ctxs[f->protomap][f->alproto].SetTxDetectState(tx, s);
-    SCReturnInt(r);
+    AppLayerTxData *d = alp_ctx.ctxs[f->protomap][f->alproto].GetTxData(tx);
```
victorjulien, line 1159: Same for the setter.

catenacyber:
> +132 −780

That is how I like it

suricata-qa:
|--|--|--|--| ERROR: ASAN TEST FAIL in tlpr1_asan_suri QA test 

Pipeline 4812

suricata-qa:
ERROR: ASAN TEST FAIL in tlpr1_asan_suri QA test

ERROR: QA failed on tlpr1_asan_suri.
  

Pipeline 4812

jasonish:
Updated here: https://github.com/OISF/suricata/pull/6622

PR #6598, opened 2021-11-10
smtp: log transaction even if no email present

jasonish:
The SMTP transaction logger was not writing the log if the email
portion of the logger failed, such as in the case of STARTTLS
where this is no email decoded.

Ticket https://redmine.openinfosecfoundation.org/issues/4817

suricata-verify-pr: 581

### DIFF
--- a/src/output-json-smtp.c
+++ b/src/output-json-smtp.c
@@ -85,9 +85,8 @@ static int JsonSmtpLogger(ThreadVars *tv, void *thread_data, const Packet *p, Fl
     EveSmtpDataLogger(f, state, tx, tx_id, jb);
     jb_close(jb);
 
-    if (EveEmailLogJson(jhl, jb, p, f, state, tx, tx_id) == TM_ECODE_OK) {
-        OutputJsonBuilderBuffer(jb, jhl->ctx);
-    }
+    EveEmailLogJson(jhl, jb, p, f, state, tx, tx_id);
+    OutputJsonBuilderBuffer(jb, jhl->ctx);
 
     jb_free(jb);
 


src/output-json-smtp.c
```diff
@@ -85,9 +85,8 @@ static int JsonSmtpLogger(ThreadVars *tv, void *thread_data, const Packet *p, Fl
     EveSmtpDataLogger(f, state, tx, tx_id, jb);
     jb_close(jb);
 
-    if (EveEmailLogJson(jhl, jb, p, f, state, tx, tx_id) == TM_ECODE_OK) {
-        OutputJsonBuilderBuffer(jb, jhl->ctx);
-    }
+    EveEmailLogJson(jhl, jb, p, f, state, tx, tx_id);
```
inashivb, line 88: Do we get an empty `email: {}` in the eve in such a case now?

jasonish: No. `EveEmailLogJson` reverts the opening of that object on failure. Its like it was designed to be done this way ;)

victorjulien:
Do you have a SV test for this?

jasonish:
> Do you have a SV test for this?

Not yet, I'm not sure if the pcap I tested against can be used in a test, and the public one I found on the Wireshark wiki isn't picked up as SMTP at all, which is another issue to look at I suppose.

jasonish:
> Not yet, I'm not sure if the pcap I tested against can be used in a test, and the public one I found on the Wireshark wiki isn't picked up as SMTP at all, which is another issue to look at I suppose.

Ok, I was able to generate one that is safe to share.  S-V will be coming.

jasonish:
S-V: https://github.com/OISF/suricata-verify/pull/581

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | generic_stats_chk |
| .flow.mgr.flows_checked | 2573948 | 2151610 | 119.63% |
| .flow.mgr.flows_notimeout | 1125057 | 859269 | 130.93% |
| .flow.mgr.flows_timeout | 1448891 | 1292341 | 112.11% | 

Pipeline 4823

victorjulien:
Merged in #6629, thanks!

PR #6606, opened 2021-11-12
app-layer/template: don't always enable if unittests built

jasonish:
314ec77f88325a4e8989e898991b9af493cad3dc had the unintended side affect
of enabling the template parser and detection buffer if unittests were
enabled.

Fix this by using the new `Default` method for registering parsers.
However, the buffer still needs an explicit configuration check.

Also convert Notice debug messages to Debug to reduce output when in
unittest mode.  If we feel stronly this should still be Notice in the
template, that is a conversion we can make in the generation script when
generating a new parser.

### DIFF
--- a/src/app-layer-template.c
+++ b/src/app-layer-template.c
@@ -471,19 +471,11 @@ void RegisterTemplateParsers(void)
 {
     const char *proto_name = "template";
 
-    /* TEMPLATE_START_REMOVE */
-#ifndef UNITTESTS
-    /* Ensure template registration for unittests */
-    if (ConfGetNode("app-layer.protocols.template") == NULL) {
-        return;
-    }
-#endif
-    /* TEMPLATE_END_REMOVE */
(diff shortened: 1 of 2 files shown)


suricata-qa:
Information:
 
|--|--|--|--| 
| | generic_stats_chk |
| .flow.wrk.flows_evicted | 67793 | 77611 | 87.35% |
| .flow.mgr.flows_checked | 2524919 | 2151610 | 117.35% |
| .flow.mgr.flows_notimeout | 1106331 | 859269 | 128.75% | 

Pipeline 4831

victorjulien:
Merged through #6634, thanks. Can you still add a ticket?

jasonish:
> Merged through #6634, thanks. Can you still add a ticket?

This was only broken in master and now fixed.  Still a ticket?  Just don't want it to show up as fixed in 7 in the release notes.

PR #5502, opened 2020-10-20
detect/msg: convert to FAIL/PASS API

jufajardini:
Issue 4053. Incorporate change requests to keep code more compact and
avoid unnecessary branches.

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4053

Describe changes:
- use FAIL_IF_PASS checks for pointers
- delete no longer necessary if branches
- rearrange declarations for a more compact code

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #6612, opened 2021-11-14
detect/siggroup: convert unittests to FAIL/PASS APIs v1

ghostinthehive:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#4028](https://redmine.openinfosecfoundation.org/issues/4028)

Describe changes:
- Convert unittests `SigGroupHeadTest*` to use `FAIL/PASS` APIs
- Cleanup unittests
- Update Copyright year

### DIFF
--- a/src/detect-engine-siggroup.c
+++ b/src/detect-engine-siggroup.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2013 Open Information Security Foundation
+/* Copyright (C) 2007-2021 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -779,19 +779,15 @@ int SigAddressPrepareStage1(DetectEngineCtx *);
  */
 static int SigGroupHeadTest03(void)
 {
-    int result = 1;
-
     DetectEngineCtx de_ctx;
 
     SigGroupHeadHashInit(&de_ctx);
-
-    result &= (de_ctx.sgh_hash_table != NULL);
+    FAIL_IF_NOT(de_ctx.sgh_hash_table);
 
     SigGroupHeadHashFree(&de_ctx);
+    FAIL_IF(de_ctx.sgh_hash_table);
 
-    result &= (de_ctx.sgh_hash_table == NULL);
-
-    return result;
+    PASS;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/detect-engine-siggroup.c
```diff
@@ -1071,58 +995,35 @@ static int SigGroupHeadTest08(void)
...
+
+    s = s->next = SigInit(de_ctx, "alert tcp any any -> any any "
+                                  "(msg:\"SigGroupHead tests\"; content:\"test1\"; "
+                                  "content:\"test2\"; content:\"test3\"; sid:4;)");
+    FAIL_IF_NULL(s);
+
+    s = s->next = SigInit(de_ctx, "alert tcp any any -> any any "
+                                  "(msg:\"SigGroupHead tests\"; content:\"test1\"; "
+                                  "content:\"test2\"; content:\"test3\"; sid:5;)");
+    FAIL_IF_NULL(s);
```
ghostinthehive, line 1027: I noticed here while trying to use `DetectEngineAppendSig` the tests fail!
I don't know if its the nature of this test itself that doesn't work with `DetectEngineAppendSig` and Array Matching, or is it a significant difference between `SigInit` and `DetectEngineAppendSig` that make one valid than the other in this specific test?

src/detect-engine-siggroup.c
```diff
@@ -779,19 +779,15 @@ int SigAddressPrepareStage1(DetectEngineCtx *);
...
 static int SigGroupHeadTest03(void)
 {
-    int result = 1;
-
     DetectEngineCtx de_ctx;
 
     SigGroupHeadHashInit(&de_ctx);
-
-    result &= (de_ctx.sgh_hash_table != NULL);
+    FAIL_IF_NOT(de_ctx.sgh_hash_table);
```
jlucovsky, line 785: Although this is correct, I think it'd be clearer if the macros `FAIL_IF_NULL` and `FAIL_IF_NOT_NULL` were used.

src/detect-engine-siggroup.c
```diff
@@ -801,58 +797,35 @@ static int SigGroupHeadTest03(void)
...
-    int result = 1;
     SigGroupHead *sh = NULL;
+
     DetectEngineCtx *de_ctx = DetectEngineCtxInit();
-    Signature *prev_sig = NULL;
+    FAIL_IF_NULL(de_ctx);
 
-    if (de_ctx == NULL)
-        return 0;
+    Signature *s = DetectEngineAppendSig(de_ctx, "alert tcp any any -> any any "
```
jlucovsky, line 805: What's the reasoning for converting the test to `DetectEngineAppendSig` from `SigInit`?

src/detect-engine-siggroup.c
```diff
@@ -946,28 +895,27 @@ static int SigGroupHeadTest07(void)
...
     SigGroupHeadSetSigCnt(sh, 4);
 
-    result &= (sh->init->sig_cnt == 3);
-    result &= (SigGroupHeadContainsSigId(de_ctx, sh, 1) == 1);
-    result &= (SigGroupHeadContainsSigId(de_ctx, sh, 2) == 0);
-    result &= (SigGroupHeadContainsSigId(de_ctx, sh, 3) == 1);
-    result &= (SigGroupHeadContainsSigId(de_ctx, sh, 4) == 0);
-    result &= (SigGroupHeadContainsSigId(de_ctx, sh, 5) == 1);
+    FAIL_IF_NOT(sh->init->sig_cnt == 3);
+    FAIL_IF_NOT(SigGroupHeadContainsSigId(de_ctx, sh, 1));
```
jlucovsky, line 899: I suggest that we maintain the `== <someval>` for macros. It's important to have consistency and IMO, retaining checks with specific values is needed so use that pattern throughout.

src/detect-engine-siggroup.c
jlucovsky: It's not clear why `DetectEngineAppendSig` usage is part of this PR as opposed to maintaining the usages of `SigInit`?

ghostinthehive: It is common in the unittests I worked on and ones I've seen to use `DetectEngineAppendSig` it made the tests more compact since it directly reassigns to the latest/appended `Sig` 
Some unittests even have a mixture of both, So I don't really rely on a specific reason for this conversion _other than cleanups_ , yet this test showed a failure upon using `DetectEngineAppendSig` .
So I wonder when to consider leaving `SigInit` and when to convert it?

src/detect-engine-siggroup.c
jufajardini: Indeed, Victor has also given that feedback in some of my PRs: it is preferred to use the macros with explicit `NULL`, when dealing with pointers.

src/detect-engine-siggroup.c
jufajardini: That was a feedback from Victor (see https://github.com/OISF/suricata/pull/5502#discussion_r509016873) that I passed forward, considering, as @ghostinthehive has mentioned, making tests shorter. I'll have a look and try to figure what can be the cause for the test failure in this case!

jufajardini: I think you are on the right track! So, debugging, what I got at is that the original test would initialize the signatures following a _queue_ logic, while `DetectEngineAppendSig` appends signatures in a _stack_ logic. That caused the array misalignment. Following that,  I changed the checks to:
```    
FAIL_IF_NOT(sh->init->match_array[0] == de_ctx->sig_list->next->next->next->next);
FAIL_IF_NOT(sh->init->match_array[1] == de_ctx->sig_list->next->next);
FAIL_IF_NOT(sh->init->match_array[2] == de_ctx->sig_list);
```
And the test passed again...

Don't know what is the best approach here, but at least now I guess at least we understand the whys? :)

jufajardini (commented):
This is nit, but I'm thinking that renaming the tests could also be a good call, since you're already working them =]

src/detect-engine-siggroup.c
jufajardini: My guess is that either choice is fine. I'd say that if you do decide to keep the `DetectEngineAppendSig`, though, it's indeed better to add the comment that you've mentioned. =]

ghostinthehive:
Opened PR-v2 #6617

PR #6437, opened 2021-10-04
PostgreSQL: add initial support (#4241) - v3

jufajardini:
Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/4241

Previous PR: #6396 

Describe changes:
- tell Suricata to move on, in case it can't find a transaction
- fix SCLogNotice statements, to fix CIFuzz check failures 
- delete unneeded checks in unit tests
- fix wording in some comments
- delete unused functions in C and Rust
- clear and fix commit history
- apply rust fmt and clippy changes

suricata-verify-pr: 550
https://github.com/OISF/suricata-verify/pull/550

PR #6630, opened 2021-11-22
 PostgreSQL: add initial support (#4241) - v5

jufajardini:
Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/4241

Previous PR: [#6437](https://github.com/OISF/suricata/pull/6463)

Describe changes:
- add transaction tracking, making `pcap_cnt` show up more often (there's still a _delay_ of sorts, though. Seems to be always at least 1 pcap ahead than what one sees in Wireshark)
- address feeback from previous pr:
- clean up logger, replacing log function with `to_str()` method
- remove PgsqlTransaction from cbindgen.toml and keep `void ptr` on the C side
- remove unused/unneeded C files & functions

Must still address:
- yaml configuration to avoid logging passwords
- user guide documentation
- understand and check GAP/ incomplete messages situation, pointed out by @catenacyber 

suricata-verify-pr: 584
[OISF/suricata-verify#584](https://github.com/OISF/suricata-verify/pull/584)


jufajardini:
Must have left something slip while creating the updated tests, two are pgsql tests are indeed failing, still.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_dbg |
 

Pipeline 5032

rust/src/pgsql/pgsql.rs
```diff
@@ -0,0 +1,851 @@
...
+
+#[derive(Debug)]
+pub struct PgsqlTransaction {
+    pub tx_id: u64,
+    pub tx_state: PgsqlTransactionState,
+    pub requests: Vec<PgsqlFEMessage>,
+    pub responses: Vec<PgsqlBEMessage>,
+
+    de_state: Option<*mut core::DetectEngineState>,
+    events: *mut core::AppLayerDecoderEvents,
```
jasonish, line 47: Before merge to master I think this should be updated to be like the parsers after the very recent merge of putting these fields in the `tx_data` to remove some boilerplate.  Look at the DNS parser and see how we implement `State` and `Transaction` when can register some generic functions to handle a few things.

rust/src/pgsql/pgsql.rs
```diff
@@ -0,0 +1,851 @@
...
+            core::sc_detect_engine_state_free(state);
+        }
+    }
+}
+
+impl Drop for PgsqlTransaction {
+    fn drop(&mut self) {
+        self.free();
+    }
+}
```
jasonish, line 78: Stuff like this can just be removed with the suggested change above.  Again, this stuff just got merged in the last week or so.

rust/src/pgsql/pgsql.rs
jufajardini: Thanks, addressing that right now! :)

jasonish:
Looking at this event:
```
{
  "timestamp": "2021-09-24T13:30:17.968056-0600",
  "flow_id": 1343328945881694,
  "pcap_cnt": 53,
  "event_type": "pgsql",
  "src_ip": "192.168.1.102",
  "src_port": 41678,
  "dest_ip": "192.168.1.74",
  "dest_port": 5432,
  "proto": "TCP",
  "pgsql": {
    "tx_id": 14,
    "request": {
      "simple_query": "select * from dummy_table;"
    },
    "responses": [
      {
        "row_description": {
          "field_count": 3,
          "backend_response": "Response Ok"
        }
      },
      {
        "data_row": {
          "backend_response": "Response Ok"
        }
      },
      {
        "data_row": {
          "backend_response": "Response Ok"
        }
      },
      {
        "data_row": {
          "backend_response": "Response Ok"
        }
      },
      {
        "command_completed": "SELECT 3"
      },
      {
        "ready_for_query": {
          "transaction_status": "I"
        }
      }
    ]
  }
}
```
I'm not quite sure how to reason about the `backend_response`?  It means that the backend responded successfully?  I wonder if a `"success": true` would be better here.  My main concern is that "Response Ok" isn't the best for machine readability.

Also, it seems a `data_row` record is logged for each row returned? I'm curious what happens when I execute a query that delivers 1000s of rows?

jasonish:
I'm going to send you a pcap with a copy command followed by a simple select.  The copy request is logged, but then nothing is logged after it.

jufajardini:
> Looking at this event:
> 
> ```
> {
>   "timestamp": "2021-09-24T13:30:17.968056-0600",
>   "flow_id": 1343328945881694,
>   "pcap_cnt": 53,
>   "event_type": "pgsql",
>   "src_ip": "192.168.1.102",
>   "src_port": 41678,
>   "dest_ip": "192.168.1.74",
>   "dest_port": 5432,
>   "proto": "TCP",
>   "pgsql": {
>     "tx_id": 14,
>     "request": {
>       "simple_query": "select * from dummy_table;"
>     },
>     "responses": [
>       {
>         "row_description": {
>           "field_count": 3,
>           "backend_response": "Response Ok"
>         }
>       },
>       {
>         "data_row": {
>           "backend_response": "Response Ok"
>         }
>       },
>       {
>         "data_row": {
>           "backend_response": "Response Ok"
>         }
>       },
>       {
>         "data_row": {
>           "backend_response": "Response Ok"
>         }
>       },
>       {
>         "command_completed": "SELECT 3"
>       },
>       {
>         "ready_for_query": {
>           "transaction_status": "I"
>         }
>       }
>     ]
>   }
> }
> ```
> 
> I'm not quite sure how to reason about the `backend_response`? It means that the backend responded successfully? I wonder if a `"success": true` would be better here. My main concern is that "Response Ok" isn't the best for machine readability.
> 
> Also, it seems a `data_row` record is logged for each row returned? I'm curious what happens when I execute a query that delivers 1000s of rows?

Indeed... for the non-extended version, should we then just have something like `"data_row": "x lines logged"`?

jasonish:
> Indeed... for the non-extended version, should we then just have something like `"data_row": "x rows logged"`? Where _x_ is the number of returned data_row messages

Even for extended, this could lead too too much logging.  For now can we just count the number of rows returned then output some counter when done?  COPY operates much the same way, the data is just formatter differently.

jasonish:
> * yaml configuration to avoid logging passwords

Maybe for now just don't log the passwords.  The auth type, success/fail is all we need.

jufajardini:
> > * yaml configuration to avoid logging passwords
> 
> Maybe for now just don't log the passwords. The auth type, success/fail is all we need.

Oh. I've implemented it already. Should I drop it?

jufajardini:
> > > * yaml configuration to avoid logging passwords
> > 
> > 
> > Maybe for now just don't log the passwords. The auth type, success/fail is all we need.
> 
> Oh. I've implemented it already. Should I drop it?

this has been answered elsewhere.

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/6645

PR #6636, opened 2021-11-23
Integer overflow 4516 v6

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4516

Describe changes:
- Avoids some warnings about integers

Obviously not complete yet... But still mergeable ?

Replaces #6628 with
- CI only changing builds that support new warning flags
- commits squashing

libhtp-pr: 338

I also discovered a style issue : removing explicit useless casts such as
```
uint64_t t = 0;
uint64_t x = 1;
x = (uint64_t) t;
```
Is there some automated tool to do this ? 
(and remove the `#if 0` as well ?)

PR #6638, opened 2021-11-23
Integer overflow 4516 v7

catenacyber:

Describe changes:
- Avoids some warnings about integers

Obviously not complete yet... But still mergeable ?

Replaces #6636 with
- CI only using builds that support new warning flags (ie Fedora with latest clang)

libhtp-pr: 339

I also discovered a style issue : removing explicit useless casts such as
```
uint64_t t = 0;
uint64_t x = 1;
x = (uint64_t) t;
```
Is there some automated tool to do this ? 
(and remove the `#if 0` as well ?)

### DIFF
--- a/src/stream-tcp.c
+++ b/src/stream-tcp.c
@@ -5973,8 +5973,7 @@ invalid:
  * \param ssn TCP Session
  * \param direction direction to set the flag in: 0 toserver, 1 toclient
  */
-void StreamTcpUpdateAppLayerProgress(TcpSession *ssn, char direction,
-        const uint32_t progress)
+void StreamTcpUpdateAppLayerProgress(TcpSession *ssn, bool direction, const uint32_t progress)
 {
     if (direction) {
         ssn->server.app_progress_rel += progress;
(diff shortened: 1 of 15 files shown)


victorjulien:
Partial merge in #6642. I've left out the direction as bool commit as I would like a different solution for that.

src/stream-tcp.c
```diff
@@ -5973,8 +5973,7 @@ static inline int StreamTcpValidateAck(TcpSession *ssn, TcpStream *stream, Packe
  * \param ssn TCP Session
  * \param direction direction to set the flag in: 0 toserver, 1 toclient
  */
-void StreamTcpUpdateAppLayerProgress(TcpSession *ssn, char direction,
-        const uint32_t progress)
+void StreamTcpUpdateAppLayerProgress(TcpSession *ssn, bool direction, const uint32_t progress)
```
victorjulien, line 5976: I would like this to be solved in a different way, more like how its now done on the Rust side: with a direction enum

catenacyber: ok will do

PR #6640, opened 2021-11-23
app-layer: include decoder events in app-layer tx data

jasonish:
As most parsers use an events structure we can include it in the
tx_data structure to reduce some boilerplate/housekeeping code
in app-layer parsers.


victorjulien:
Merged in #6642, thanks!

PR #6649, opened 2021-11-25
app-layer: fix integer warnings

catenacyber:

Describe changes:
- Avoids some warnings about integers : for direction argument of `StreamTcpUpdateAppLayerProgress`

DRAFT : do not merge

@victorjulien you said :

> I would like this to be solved in a different way, more like how its now done on the Rust side: with a direction enum

But C enums are not as good as Rust enums.
So, I am not sure this is a good idea.

For instance, keeping `direction = 1 - direction;` in this patch (with direction being an enum `Direction`) does not produce warnings on my setup !

See commit https://github.com/OISF/suricata/pull/6638/commits/53aac7fd0dd4b8283f151bc223d351cd38ffda32 for the alternative

### DIFF
--- a/rust/cbindgen.toml
+++ b/rust/cbindgen.toml
@@ -77,6 +77,7 @@ include = [
     "SIPState",
     "ModbusState",
     "CMark",
+    "Direction",
 ]
 
 # A list of items to not include in the generated bindings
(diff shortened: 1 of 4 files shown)


suricata-qa:

ERROR: QA failed on tlpr1_autofp_suri.
  

Pipeline 5071

catenacyber:
See #7006

PR #6658, opened 2021-11-30
Ftp data alert json 4860 v1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4860
but also https://redmine.openinfosecfoundation.org/issues/1369

Describe changes:
- move some app-layer specific fields out of root object (into the app-layer specific object)
  - ftp-data : command and filename (for alert metadata)
  - http : xff (for alert metadata as well)

suricata-verify-pr: 590

PR #6660, opened 2021-12-02
Ftp data alert json 4860 v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4860
but also https://redmine.openinfosecfoundation.org/issues/1369

Describe changes:
- move some app-layer specific fields out of root object (into the app-layer specific object)
  - ftp-data : command and filename (for alert metadata)
  - http : xff (for alert metadata as well)

suricata-verify-pr: 590

Replaces #6658 with upgrade guide additions

### DIFF
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -519,7 +519,10 @@ static void AlertAddAppLayer(const Packet *p, JsonBuilder *jb,
             }
             break;
         case ALPROTO_FTPDATA:
+            jb_get_mark(jb, &mark);
+            jb_open_object(jb, "ftp_data");
             EveFTPDataAddMetadata(p->flow, jb);
+            jb_close(jb);
             break;
         case ALPROTO_DNP3:
             AlertJsonDnp3(p->flow, tx_id, jb);
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -441,8 +441,8 @@ static void AlertAddPayload(AlertJsonOutputCtx *json_output_ctx, JsonBuilder *js
     }
 }
 
-static void AlertAddAppLayer(const Packet *p, JsonBuilder *jb,
-        const uint64_t tx_id, const uint16_t option_flags)
+static void AlertAddAppLayer(const Packet *p, JsonBuilder *jb, const uint64_t tx_id,
+        const uint16_t option_flags, const char *xff_buffer)
 {
     const AppProto proto = FlowGetAppProtocol(p->flow);
     JsonBuilderMark mark = { 0, 0, 0 };
@@ -457,6 +457,9 @@ static void AlertAddAppLayer(const Packet *p, JsonBuilder *jb,
                 if (option_flags & LOG_JSON_HTTP_BODY_BASE64) {
                     EveHttpLogJSONBodyBase64(jb, p->flow, tx_id);
                 }
+            }
+            if (xff_buffer[0]) {
+                jb_set_string(jb, "xff", xff_buffer);
             }
             jb_close(jb);
             break;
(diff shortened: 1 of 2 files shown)


suricata-qa:
ERROR: QA failed on tlpw2_autofp_alerts_cmp.
  

Pipeline 5130

src/output-json-alert.c
```diff
@@ -715,10 +726,6 @@ static int AlertJson(ThreadVars *tv, JsonAlertLogThread *aft, const Packet *p)
             EvePacket(p, jb, 0);
         }
 
-        if (have_xff_ip && xff_cfg->flags & XFF_EXTRADATA) {
```
victorjulien, line 718: here it did not depend on `LOG_JSON_APP_LAYER` being set, so this seems like a bigger change than just logging it in a different object

catenacyber: Do you mean we want to keep `xff` in root ?

victorjulien: the point of my comment is that in this change the xff will not be logged where before it was

victorjulien:
I merged the ftp-data in #6681

src/output-json-alert.c
catenacyber: Indeed.

But beyond that, I feel unsure about where this `xff` should be : stay at the root, or move to http...
What is your feeling on this ?

catenacyber: Thinking back to this :
- `xff` is part of `http` for http and fileinfo, not for alert cf `jq 'select(.http.xff) | .event_type' tests/*/*/eve.json`
- If we want `xff` to be part of `http` object in json, I think it is expected that if we do not have `LOG_JSON_APP_LAYER`, we should not have `http` and thus we should not have `http.xff`
- If we want to keep it even without `LOG_JSON_APP_LAYER`, it is fine to leave it at the root

Anyone has an opinion where this `xff` should be ?

catenacyber: cc @victorjulien ?

victorjulien: Not sure. @jasonish?

jasonish: Tempted to leave it as is.

The xff stuff predates eve where it was just extra data on an alert, so it should probably be kept even in the absence of of metadata logging on alerts.

But I think the ideal solution would be to log a partial http metadata section if we have xff, but metadata logging is off..  So you'd end up with an alert like:
```
{
  "timestamp": "2018-02-16T20:49:12.820418+0000",
  "flow_id": 1646695893454897,
  "pcap_cnt": 9,
  "event_type": "alert",
  "src_ip": "82.165.177.154",
  "src_port": 80,
  "dest_ip": "10.16.1.11",
  "dest_port": 46676,
  "proto": "TCP",
  "tx_id": 0,
  "alert": {
    "action": "allowed",
    "gid": 1,
    "signature_id": 1,
    "rev": 1,
    "signature": "TEST RULE",
    "category": "",
    "severity": 3
  },
  "http": {
    "xff": "10.2.2.2, 10.3.3.3"
  },
  "app_proto": "http",
}
```

If metadata was enabled, you'd then see the full http section of:

```
"http": {
    "hostname": "www.testmyids.com",
    "url": "/",
    "http_user_agent": "curl/7.55.1",
    "xff": "10.2.2.2, 10.3.3.3",
    "http_content_type": "text/html",
    "http_method": "GET",
    "protocol": "HTTP/1.1",
    "status": 200,
    "length": 39
  },
```

(sample from SV http-xff-eve-reverse-extra-data)

catenacyber: Done in https://github.com/OISF/suricata/pull/6904

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6904

PR #6661, opened 2021-12-02
App maxtx 4530 v5

catenacyber:

Describe changes:

cc @satta for the MQTT changes

Replaces #6251 with making something protocol-specific instead of generic


catenacyber:
Last commit is good for merge in https://github.com/OISF/suricata/pull/6662

catenacyber:
> I like this approach even if it is more work, so lets continue with this.

How should I continue ? (beyond waiting first for merge of #6662 alone)
How can we be sure that another app-layer protocol does not create too many transactions ?
Should I add a debug validation ?

satta:
This is meant to affect only _open_ transactions, right? I.e. there can be a lot more tx total in a long-running MQTT connection, but most of they will be done already. In `mqtt.rs` transactions are removed from the `transactions` vec via `MQTTState::free_tx()` which is called via the `tx_free` callback from the MQTT `RustParser` (`rs_mqtt_state_tx_free()`). Am I correct that this is called in time from Suricata to ensure that completed tx are pruned?

catenacyber:
> This is meant to affect only _open_ transactions, right?

Right, a flow can have as many transactions as it lives, but this PR sets a maximum to the open transactions at any point in time.


> I.e. there can be a lot more tx total in a long-running MQTT connection, but most of they will be done already. In `mqtt.rs` transactions are removed from the `transactions` vec via `MQTTState::free_tx()` which is called via the `tx_free` callback from the MQTT `RustParser` (`rs_mqtt_state_tx_free()`). Am I correct that this is called in time from Suricata to ensure that completed tx are pruned?

Yes, this is the work of `AppLayerParserTransactionsCleanup`
But to know which transactions to free, the states of each transaction is checked in both direction to see that they have reached a state where they can get cleaned...

suricata-qa:

Pipeline 5129

satta:
Thanks for the clarification @catenacyber 👍🏻

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6751

PR #6668, opened 2021-12-03
Feat/appl records/v39

victorjulien:
#6666 with a profiling improvement and formatting fixups

suricata-verify-pr: 592

Some rule perf details https://github.com/OISF/suricata/pull/6666#issuecomment-985529356

PR #6669, opened 2021-12-05
Feat/appl records/v43

victorjulien:
#6668 where its all named `frame` instead of `record`.

Also includes some experiments with changing the data input to the app layer parsers from `const uint8_t *input, const uint32_t input_len` to a new `AppLayerStream` object:
* https://github.com/OISF/suricata/commit/7452bb588a2d8c49e32fc9b71b3af977591c6bdd (initial declaration and code)
* https://github.com/OISF/suricata/commit/12d29437c4a35238671432739f4291cd332ab5fb (frame API update)
* https://github.com/OISF/suricata/commit/4b126a468f8733fb78dec69ab1e7b9f1b1a01fc7 (implementation in SMB)

One of the goals is to replace the use of the `thread_local .. app_frame_*` variables in favor of something we just pass to the API.

suricata-verify-pr: 594


victorjulien:
Last added commits show how `AppLayerStream` can get rid of the thread local ptrs.

victorjulien:
Replaced by #6675

PR #6675, opened 2021-12-06
Feat/appl records/v45

victorjulien:
#6669 where `AppLayerStream` is now `StreamSlice` to better reflect what it is. Removes previous use of thread local vars to handle the records, in favor of using the `StreamSlice` data.

suricata-verify-pr: 594


src/app-layer-frames.c
```diff
@@ -0,0 +1,416 @@
...
+
+    Frame *r = FrameNew(frames, (uint32_t)ptr_offset, len);
+    if (r != NULL) {
+        r->type = frame_type;
+    }
+    return r;
+}
+
+Frame *AppLayerFrameNew2(Flow *f, const StreamSlice *stream_slice, const uint32_t frame_start_rel,
+        const uint32_t len, int dir, uint8_t frame_type)
```
jasonish, line 350: Will there just be one variation when done?

src/app-layer-frames.c
```diff
@@ -0,0 +1,416 @@
...
+    /* workarounds for many (unit|fuzz)tests not handling TCP data properly */
+#if defined(UNITTESTS) || defined(FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION)
+    if (f->protoctx == NULL)
+        return NULL;
+    if (frame_start < stream_slice->input || frame_start >= stream_slice->input + stream_slice->input_len)
+        return NULL;
+#endif
+    BUG_ON(frame_start < stream_slice->input);
+    BUG_ON(stream_slice->input == NULL);
+    BUG_ON(f->proto != IPPROTO_TCP);
```
jasonish, line 319: No UDP support? I got tripped up while looking at DNS frame support which can be TCP or UDP.

jasonish:
Looking at the logging for a crude DNS conversion I did, I see this logged:
```
"frame": {
    "type": "message",
    "id": 0,
    "offset": 0,
    "length": 34,
    "diframetion": "toserver",
    "complete": true,
    "payload": "ACAAAAEAAAEAAAAAAAADd3d3Bmdvb2dsZQNjb20AAAEAAQ==",
    "payload_printable": ". .............www.google.com.....",
    "payload_hex": "|00 20 00 00 01 00 00 01 00 00 00 00 00 00 03|www|06|google|03|com|00 00 01 00 01|"
  }
```

Out side of debug, is the `id` needed? Same with `offset`, and to a lesser extend `length`.

Also, interesting field, `diframetion`? Would the direction be needed with that other work that Eric has been doing that tells us more about the direction of an alert?

src/detect-engine.c
```diff
@@ -920,6 +920,31 @@ const char *DetectBufferTypeGetDescriptionByName(const char *name)
...
+    BUG_ON(!exists);
+    exists->mpm = true;
+    SCLogDebug("%p %s -- %d supports mpm", exists, name, exists->id);
+}
+
+void DetectEngineBufferTypeSupportsTransformations(DetectEngineCtx *de_ctx, const char *name)
+{
+    DetectBufferType *exists = DetectEngineBufferTypeLookupByName(de_ctx, name);
+    BUG_ON(!exists);
+    BUG_ON(!exists);
```
inashivb, line 1234: nit: Extra `BUG_ON`?

src/app-layer-frames.c
victorjulien: > Looking at the logging for a crude DNS conversion I did, I see this logged:
> ...
> Out side of debug, is the `id` needed? Same with `offset`, and to a lesser extend `length`.

Offset I think tells you something useful: its the start of the frame vs the start of the stream. Id I'm not sure about whether its useful.

> Also, interesting field, `diframetion`? Would the direction be needed with that other work that Eric has been doing that tells us more about the direction of an alert?

That should be `direction`. Its a artifact of `s/rec/frame/g` ;)

victorjulien:
> Looking at the logging for a crude DNS conversion I did, I see this logged:
> ...
> Out side of debug, is the `id` needed? Same with `offset`, and to a lesser extend `length`.

Offset I think tells you something useful: its the start of the frame vs the start of the stream. Id I'm not sure about whether its useful.

> Also, interesting field, `diframetion`? Would the direction be needed with that other work that Eric has been doing that tells us more about the direction of an alert?

That should be `direction`. Its a artifact of `s/rec/frame/g` ;)

src/app-layer-frames.c
victorjulien: I've focused on TCP for now, as this is where the perf gain is to be had, so blocked everything else. But I don't think its hard to extend this to UDP. The data the frames point to will be in the actual packet payload then.

src/app-layer-frames.c
victorjulien: Hope so

victorjulien:
`length` I think is useful. Think about trying to detect an issue there a record has a specific length and you want to use `frame:dns.message; bsize:666;`. Think it helps if suri just tells the rule writer/analyst what the len is.

src/app-layer-parser.c
```diff
@@ -1219,11 +1234,10 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow
 
     /* invoke the recursive parser, but only on data. We may get empty msgs on EOF */
     if (input_len > 0 || (flags & STREAM_EOF)) {
+        Setup(f, flags & (STREAM_TOSERVER | STREAM_TOCLIENT), input, input_len, &stream_slice);
```
inashivb, line 1285: Q: So, if we are at the end of a stream, we might end up creating an empty stream slice here?

src/app-layer-frames.c
```diff
@@ -0,0 +1,416 @@
...
+            // fall through
+        } else {
+            if (frames->dyn_size == 256) {
+                SCLogNotice("limit reached! 256 dynamic frames already");
+                // limit reached
+                return NULL;
+            }
+
+            /* realloc time */
+            uint16_t new_dyn_size = frames->dyn_size * 2;
```
inashivb, line 117: Q: Can it become > 256 at this stage? Say, we already have 129 dynamic frames. Is that permissible?

src/app-layer-frames.c
```diff
@@ -0,0 +1,416 @@
...
+        return &frames->dframes[dyn_cnt];
+    }
+}
+
+static void FrameClean(Frame *frame)
+{
+    memset(frame, 0, sizeof(*frame));
+}
+
+static void StreamDPUCopy(Frame *dst, Frame *src)
```
inashivb, line 144: Q: Is this a typo or does it really mean something? `DPU`

src/app-layer-frames.c
```diff
@@ -0,0 +1,416 @@
...
+            if (frame->rel_offset + frame->len <= (int32_t)slide) {
+                // remove by not incrementing 'x'
+                SCLogDebug("removing %p", frame);
+                FrameClean(frame);
+            } else {
+                Frame *nframe;
+                if (x >= FRAMES_STATIC_CNT) {
+                    nframe = &frames->dframes[x - FRAMES_STATIC_CNT];
+                } else {
+                    nframe = &frames->sframes[x];
```
inashivb, line 232: Q: Why do we set a static frame here if we are in the dynamic frame zone?

src/app-layer-frames.h
```diff
@@ -0,0 +1,94 @@
...
+#include "detect-engine-state.h"
+#include "util-file.h"
+#include "stream-tcp-private.h"
+#include "rust.h"
+#include "app-layer-parser.h"
+
+typedef int64_t FrameId;
+
+enum {
+    FRAME_FLAGE_ENDS_AT_EOF,
```
inashivb, line 37: nit: typo `FLAGE`?

suricata-qa:

Pipeline 5147

src/app-layer-frames.c
victorjulien: Kept 2 in #6684, but gave them clearer names

src/app-layer-parser.c
victorjulien: Yes but that is fine. The parser should be accepting that as an EOF notification.

src/app-layer-frames.h
victorjulien: its deliberate. The FLAGE values are set to 0-N by their position in the enum, then the define the next line turns them into proper flags. Open to suggestions on the naming.

src/app-layer-frames.c
victorjulien: the check above limits it to 256, which is arbitrary at this point. It starts at 8, and only doubles. So I don't see how we could exceed 256 here

src/app-layer-frames.c
victorjulien: left over from older naming, will fix

src/app-layer-frames.c
victorjulien: we're moving this after removing frames from our list, so we can move frames from dynamic to static

victorjulien:
Replaced by #6684. Some comments will be addressed later as I only saw some of the comments now.

PR #6674, opened 2021-12-06
Integer overflow 4527 v1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4527

Describe changes:
- Fix integer warnings in all files beginning with a (like app)

libhtp-pr: 339

There remains one warning about the use of `StreamTcpUpdateAppLayerProgress`
cf discussion in https://github.com/OISF/suricata/pull/6649

PR #6676, opened 2021-12-06
Integer overflow 4527 v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4527

Describe changes:
- Fix integer warnings in all files beginning with a (like app)

libhtp-pr: 339

There remains one warning about the use of `StreamTcpUpdateAppLayerProgress`
cf discussion in https://github.com/OISF/suricata/pull/6649

Changes #6674 to get gcc happy about `u16 + 2 < u32` where `u16 + 2` should not become a signed integer

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -642,14 +642,18 @@ static AppLayerResult FTPParseRequest(Flow *f, void *ftp_state,
                      * Min size has been checked in FTPParseRequestCommand
                      * PATH_MAX includes the null
                      */
-                    int file_name_len = MIN(PATH_MAX - 1, state->current_line_len - 5);
+                    uint32_t file_name_len = MIN(PATH_MAX - 1, state->current_line_len - 5);
+                    if (file_name_len > UINT16_MAX) {
+                        // truncate the file name if too long for util-file.h
+                        file_name_len = UINT16_MAX;
+                    }
                     data->file_name = FTPCalloc(file_name_len + 1, sizeof(char));
                     if (data->file_name == NULL) {
                         FtpTransferCmdFree(data);
                         SCReturnStruct(APP_LAYER_ERROR);
                     }
                     data->file_name[file_name_len] = 0;
-                    data->file_len = file_name_len;
+                    data->file_len = (uint16_t)file_name_len;
                     memcpy(data->file_name, state->current_line + 5, file_name_len);
                     data->cmd = state->command;
                     data->flow_id = FlowGetId(f);
@@ -1029,9 +1033,8 @@ static StreamingBufferConfig sbcfg = STREAMING_BUFFER_CONFIG_INITIALIZER;
  * \retval 1 when the command is parsed, 0 otherwise
  */
 static AppLayerResult FTPDataParse(Flow *f, FtpDataState *ftpdata_state,
-        AppLayerParserState *pstate,
-        const uint8_t *input, uint32_t input_len,
-        void *local_data, int direction)
+        AppLayerParserState *pstate, const uint8_t *input, uint32_t input_len, void *local_data,
+        uint8_t direction)
 {
     uint16_t flags = FileFlowToFlags(f, direction);
     int ret = 0;
@@ -1361,7 +1364,7 @@ uint16_t JsonGetNextLineFromBuffer(const char *buffer, const uint16_t len)
     }
 
     char *c = strchr(buffer, '\n');
-    return c == NULL ? len : c - buffer + 1;
+    return c == NULL ? len : (uint16_t)(c - buffer + 1);
 }
 
 void EveFTPDataAddMetadata(const Flow *f, JsonBuilder *jb)
--- a/src/app-layer-htp.c
+++ b/src/app-layer-htp.c
@@ -524,7 +524,7 @@ static uint32_t AppLayerHtpComputeChunkLength(uint64_t content_len_so_far, uint3
 /* below error messages updated up to libhtp 0.5.7 (git 379632278b38b9a792183694a4febb9e0dbd1e7a) */
 struct {
     const char *msg;
-    int  de;
+    uint8_t de;
 } htp_errors[] = {
     { "GZip decompressor: inflateInit2 failed", HTTP_DECODER_EVENT_GZIP_DECOMPRESSION_FAILED},
     { "Request field invalid: colon missing", HTTP_DECODER_EVENT_REQUEST_FIELD_MISSING_COLON},
@@ -547,7 +547,7 @@ struct {
 
 struct {
     const char *msg;
-    int  de;
+    uint8_t de;
 } htp_warnings[] = {
     { "GZip decompressor:", HTTP_DECODER_EVENT_GZIP_DECOMPRESSION_FAILED},
     { "Request field invalid", HTTP_DECODER_EVENT_REQUEST_HEADER_INVALID},
@@ -594,7 +594,7 @@ struct {
  *
  *  \retval id the id or 0 in case of not found
  */
-static int HTPHandleWarningGetId(const char *msg)
+static uint8_t HTPHandleWarningGetId(const char *msg)
 {
     SCLogDebug("received warning \"%s\"", msg);
     size_t idx;
@@ -618,7 +618,7 @@ static int HTPHandleWarningGetId(const char *msg)
  *
  *  \retval id the id or 0 in case of not found
  */
-static int HTPHandleErrorGetId(const char *msg)
+static uint8_t HTPHandleErrorGetId(const char *msg)
 {
     SCLogDebug("received error \"%s\"", msg);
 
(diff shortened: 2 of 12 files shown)


suricata-qa:

Pipeline 5146

src/app-layer-ftp.c
```diff
@@ -642,14 +642,18 @@ static AppLayerResult FTPParseRequest(Flow *f, void *ftp_state,
                      * Min size has been checked in FTPParseRequestCommand
                      * PATH_MAX includes the null
                      */
-                    int file_name_len = MIN(PATH_MAX - 1, state->current_line_len - 5);
+                    uint32_t file_name_len = MIN(PATH_MAX - 1, state->current_line_len - 5);
+                    if (file_name_len > UINT16_MAX) {
```
victorjulien, line 646: this is a runtime check for something that shouldn't be possible in practice, as iirc PATH_MAX is 4k. Should we use a static assertion here?
https://riptutorial.com/c/example/1842/static-assertion

src/app-layer-htp.c
```diff
@@ -524,7 +524,7 @@ static uint32_t AppLayerHtpComputeChunkLength(uint64_t content_len_so_far, uint3
 /* below error messages updated up to libhtp 0.5.7 (git 379632278b38b9a792183694a4febb9e0dbd1e7a) */
 struct {
     const char *msg;
-    int  de;
+    uint8_t de;
```
victorjulien, line 527: should this not be `unsigned int` as the values are defined in an `enum`?

src/app-layer-htp.c
```diff
@@ -1630,7 +1630,11 @@ static int HtpRequestBodyHandlePOSTorPUT(HtpState *hstate, HtpTxUserData *htud,
         }
 
         if (filename != NULL) {
-            result = HTPFileOpen(hstate, htud, filename, (uint32_t)filename_len, data, data_len,
+            if (filename_len > UINT16_MAX) {
```
victorjulien, line 1633: should this lead to an event? Also 64k seems very high still. If we consider `PATH_MAX` to be 4k for example.

victorjulien: we also have to consider that a 64k name leads to JSON records > 64k

src/app-layer-smtp.c
```diff
@@ -457,9 +457,12 @@ int SMTPProcessDataChunk(const uint8_t *chunk, uint32_t len,
             SCLogDebug("StreamTcpReassemblySetMinInspectDepth STREAM_TOSERVER %"PRIu32, depth);
             StreamTcpReassemblySetMinInspectDepth(flow->protoctx, STREAM_TOSERVER, depth);
 
+            uint16_t flen = (uint16_t)entity->filename_len;
+            if (entity->filename_len > UINT16_MAX) {
```
victorjulien, line 461: same comment as for http: this seems far to big

src/app-layer-smtp.c
```diff
@@ -1154,7 +1157,11 @@ static int SMTPParseCommandWithParam(SMTPState *state, uint8_t prefix_len, uint8
         return -1;
     memcpy(*target, state->current_line + i, spc_i - i);
     (*target)[spc_i - i] = '\0';
-    *target_len = spc_i - i;
+    if (spc_i - i > UINT16_MAX) {
```
victorjulien, line 1160: does this mean we're limiting lines to a max length now where previously we didn't?

src/app-layer-ftp.c
victorjulien: to clarify: if we have a static assert that PATH_MAX < UINT16_MAX we can get rid of a runtime check like this

catenacyber: Indeed, will do

src/app-layer-htp.c
catenacyber: An enum value such as `HTTP_DECODER_EVENT_UNKNOWN_ERROR` should be defined as an `int`(signed)
But we can safely cast it to `uint8_t` as we do not have more than 256 values. I even think that this is some kind of preprocessing that acts as `#define` 
Furthermore, we later call `HTPSetEvent` which uses `uint8_t` (because it calls `AppLayerDecoderEventsSetEventRaw` which uses `uint8_t`)

And it does not look like there is warning flag for Wimplicit-enum-int-conversion
cf https://releases.llvm.org/11.0.0/tools/clang/docs/DiagnosticsReference.html#wenum-conversion

So, I think this patch is right

src/app-layer-htp.c
catenacyber: Event seems relevant.
`PATH_MAX` seems a better limit as well.
Will do

src/app-layer-smtp.c
catenacyber: Will do

src/app-layer-smtp.c
catenacyber: > does this mean we're limiting lines to a max length now where previously we didn't?

Eheh no it does not mean that

Previously (that is currently), we are already limiting lines to a max length

But as there is an implicit cast from 'int' to 'uint16_t', if the line length is 65536, it will turn to zero (instead of saturating it to `UINT16_MAX`)
```
app-layer-smtp.c:1157:25: warning: implicit conversion loses integer precision: 'int' to 'uint16_t' (aka 'unsigned short') [-Wimplicit-int-conversion]
    *target_len = spc_i - i;
```

So, should I rather increase the sizes of `uint16_t helo_len;` and such ?

src/app-layer-htp.c
victorjulien: I think I would like us to define our own PATH_MAX(-like) limit, as I think this behaviour in suri should not depend on whatever an OS might define. Since on Linux the default seems 4k I think that might be a good value to pick.

src/app-layer-smtp.c
victorjulien: I see. No lets not allow for ridiculously long lines. But I think it should lead to another event if we reach the limit.

catenacyber: Will do

src/app-layer-htp.c
victorjulien: hmm I always thought an enum would be an unsigned int, and be 32 bits. But now I'm reading that it may actually be up to the compiler to pick a type as long as it fits? Wonder what that means for using enums in Rust - C - FFI.

catenacyber: > Wonder what that means for using enums in Rust - C - FFI.

I guess it means the enums should be defined in Rust, and exported to be used in C with cbindgen :-p

jasonish: > > Wonder what that means for using enums in Rust - C - FFI.
> 
> I guess it means the enums should be defined in Rust, and exported to be used in C with cbindgen :-p

cbindgen outputs the following, I guess in a way to fix the type:
```
enum HTTP2TransactionState {
    HTTP2StateIdle = 0,
    HTTP2StateOpen = 1,
    HTTP2StateReserved = 2,
    HTTP2StateDataClient = 3,
    HTTP2StateHalfClosedClient = 4,
    HTTP2StateDataServer = 5,
    HTTP2StateHalfClosedServer = 6,
    HTTP2StateClosed = 7,
    HTTP2StateGlobal = 8,
};
typedef uint8_t HTTP2TransactionState;
```

So that is one option.

But it looks like the representation of enums has been an issue where the compilers use a different algorithm for layout, like rustc and msvc:

https://github.com/rust-lang/rust/issues/81996

Luckily it looks like rustc, clang, gcc all agree here.

src/app-layer-ftp.c
catenacyber: What about 
```
#if PATH_MAX > UINT16_MAX
#error PATH_MAX is greater than UINT16_MAX
#endif
```
I did not know about static assertions

src/app-layer-htp.c
catenacyber: So, let's rustify all the things

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6690
Thanks for the good review

src/app-layer-htp.c
jasonish: > So, let's rustify all the things

I think it puts us on a path of one big super header file for C.  Not sure if thats really that bad of a thing tho.

catenacyber: Could not cbindgen generate one header file per rust module/directory ?

jasonish: > Could not cbindgen generate one header file per rust module/directory ?

No, thats not really how its designed.  You could call it for each file manually and put those into headers.  Then you'd have to worry about ordering of dependent data structures and so on.

src/app-layer-ftp.c
victorjulien: why not use the thing that is designed to do this, meaning static assertions?

catenacyber: > why not use the thing that is designed to do this, meaning static assertions?

For what I knew `#error` is designed to do this, I did not know about static assertions

They seem to be equivalent here :
https://wiki.sei.cmu.edu/confluence/display/c/DCL03-C.+Use+a+static+assertion+to+test+the+value+of+a+constant+expression

`static_assert` style looks more compact.
So, it is in branch v4.
Are there other comments or should I make a PR with v4 ?

PR #6677, opened 2021-12-06
detect: remove unused references from signature

catenacyber:
During thinking about https://redmine.openinfosecfoundation.org/issues/3588

Describe changes:
- Remove unused `references` from signature

So as to remove unused code and use less memory...

Am I missing something here ?

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -1331,36 +1331,6 @@ static void SigMetadataFree(Signature *s)
     SCReturn;
 }
 
-/**
- * \internal
- * \brief Free Reference list
- *
- * \param s Pointer to the signature
- */
-static void SigRefFree (Signature *s)
-{
-    SCEnter();
-
-    DetectReference *ref = NULL;
-    DetectReference *next_ref = NULL;
-
-    if (s == NULL) {
-        SCReturn;
-    }
-
-    SCLogDebug("s %p, s->references %p", s, s->references);
-
-    for (ref = s->references; ref != NULL;)   {
-        next_ref = ref->next;
-        DetectReferenceFree(ref);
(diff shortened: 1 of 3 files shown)


suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpw1_stats_chk |
| .tcp.rst | 89320 | 81110 | 110.12% | 

Pipeline 5145

jasonish:
We did talk some time ago about actually logging references in eve, but I don't see that it ever made it to a ticket.  I suppose a decision on that should be made as well.

victorjulien:
> We did talk some time ago about actually logging references in eve, but I don't see that it ever made it to a ticket. I suppose a decision on that should be made as well.

Yeah I think we need that ticket.

I'm not in favor of removing this, but I do think that adding these unconditionally to each `Signature` is a waste.

jasonish:
> > We did talk some time ago about actually logging references in eve, but I don't see that it ever made it to a ticket. I suppose a decision on that should be made as well.
> 
> Yeah I think we need that ticket.
> 
> I'm not in favor of removing this, but I do think that adding these unconditionally to each `Signature` is a waste.

https://redmine.openinfosecfoundation.org/issues/4974

One thought is that logging the rule does provided the references, but in their short form.  Logging references properly would allow us to expand them so they're usable without the reference.config file.

I wonder if it would makes sense to log some things when first seen. This could apply to rules, references and maybe some other stuff.  During a run of Suricata, the references would be logged once for that rule, then suppressed. It would be more work on the log aggregation tool I suppose. Or specific EVE logic in a SIEM.

catenacyber:
Closing in favor of the ticket to be done

PR #6683, opened 2021-12-07
smtp: sample get line processing fn

inashivb:
Sample fn to decide the direction and state of SMTP conversion to Rust.

### DIFF
--- a/rust/src/smtp/mod.rs
+++ b/rust/src/smtp/mod.rs
@@ -1,0 +1,18 @@
+/* Copyright (C) 2020 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+pub mod smtp;
--- a/rust/src/smtp/smtp.rs
+++ b/rust/src/smtp/smtp.rs
@@ -1,0 +1,165 @@
+/* Copyright (C) 2021 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+use crate::core::Direction;
+
+pub struct RustSMTPBuffer {
+    ts_buf: Vec<u8>,
+    tc_buf: Vec<u8>,
+    cur_line: Vec<u8>,
+}
+
+impl RustSMTPBuffer {
+
+    pub fn new() -> Self {
+        return Self {
+            ts_buf: Vec::new(),
+            tc_buf: Vec::new(),
+            cur_line: Vec::new(),
+        }
+    }
+
+    pub fn get_line_ts(&mut self, input: &[u8]) -> i8
+    {
+        // TODO Clear parser after integration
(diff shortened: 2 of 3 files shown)


rust/src/smtp/smtp.rs
```diff
@@ -0,0 +1,165 @@
...
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+use crate::core::Direction;
+
+pub struct RustSMTPBuffer {
+    ts_buf: Vec<u8>,
+    tc_buf: Vec<u8>,
+    cur_line: Vec<u8>,
```
victorjulien, line 22: I wonder if we can get rid of this one, and return `cur_line` from the function?

rust/src/smtp/smtp.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    pub fn new() -> Self {
+        return Self {
+            ts_buf: Vec::new(),
+            tc_buf: Vec::new(),
+            cur_line: Vec::new(),
+        }
+    }
+
+    pub fn get_line_ts(&mut self, input: &[u8]) -> i8
```
victorjulien, line 35: so instead of the `-> i8` do something like `-> Option(Vec<u8>)` here? Or maybe a slice instead of a Vec

victorjulien:
Much simpler, I like it :)

rust/src/smtp/smtp.rs
```diff
@@ -0,0 +1,165 @@
...
+            smtp_buffer.get_line_ts(buf)
+        }
+        Direction::ToClient => {
+            smtp_buffer.get_line_tc(buf)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
victorjulien, line 124: think it would be good to add a test there the input contains the end of a line, and then some more data that does not contain a line end
`EHLO [192.168.0.158]<CR><LF>more data`

Also another test to test multiple lines in one block
```
COMMAND1 data<CR><LF>
COMMAND2 data<CR><LF>
COMMAND3 data<CR><LF>
```
etc

rust/src/smtp/smtp.rs
victorjulien: Probably a `Vec<u8>` as we can split off the line from `ts_buf` or `tc_buf`

rust/src/smtp/smtp.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+
+pub unsafe extern "C" fn rs_smtp_call_getline(input: *const u8, input_len: u32, dir: u8) -> i8 {
+    if input.is_null() && input_len > 0 {
+        return -1;
+    }
+    let buf = build_slice!(input, input_len as usize);
+    let mut smtp_buffer = RustSMTPBuffer::new();
```
victorjulien, line 112: this should be passed in as an argument later

rust/src/smtp/smtp.rs
inashivb: OK. Coming up!

rust/src/smtp/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+pub mod smtp;
```
jasonish, line 18: Nothing to do with the code, but Rust style rules don't like creating the scenario of smtp/mod.rs with a contained module of the same name (smtp/smtp.rs).

The specific lint is: https://rust-lang.github.io/rust-clippy/master/

We're pretty bad for this in our code (or maybe we shouldn't consider it bad?).  But just a thought, that if this group of functions can be described better then just `smtp`, maybe they should get a different module name?

rust/src/smtp/mod.rs
```diff
@@ -0,0 +1,18 @@
+/* Copyright (C) 2020 Open Information Security Foundation
```
jasonish, line 1: 2021?

rust/src/smtp/smtp.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+        Direction::ToClient => {
+            smtp_buffer.get_line_tc(buf)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::smtp::smtp::RustSMTPBuffer;
```
jasonish, line 125: Using `use super::*;` just inside the test module is idiomatic and convenient.

rust/src/smtp/mod.rs
inashivb: Sure. I can certainly rename this one.. this is kind of not even in accordance with our existing scheme either. It's not a full parser.. and it's gonna be just some helper fns purely written in Rust.

victorjulien:
One additional thought I had: the parser (same for ftp), should ideally not do the line buffering at all. We now have the `incomplete` support in the app-layer API, which means the parser can tell the API to get called again when enough data is available. In this case it can't know much much more data is needed, so it could just return incomplete with a "needed" value of 1.

suricata-qa:

Pipeline 5305

PR #6684, opened 2021-12-07
Feat/appl records/v50

victorjulien:
#6675 rebased.
Now fully uses `StreamSlice` in API and parsers.
Fixes and cleanups.

suricata-verify-pr: 598


victorjulien:
Typo in first commit `s/enine/engine/g`

rust/src/applayer.rs
```diff
@@ -193,6 +223,48 @@ impl From<i32> for AppLayerResult {
     }
 }
 
+pub enum Frame {}
```
jasonish, line 226: I'd like to see the implementation of Frame more Rust like,if there isn't anything I'm aware of preventing it.. I quickly did this impl: https://gist.github.com/jasonish/ae033575f271f72d4d8ab36c400f340d

Ideally something like this before too many parsers are converted?

victorjulien: yeah happy to take this route

wondering still about how to deal with the `CFrame` pointer. It makes it more efficient to do something like:

* create frame
* set tx_id
* set event

As the pointer can be used directly. But after the current invocation of the parser is complete the pointer becomes invalid, so holding on to it should not be done. Then only the ID should be used. I guess the only safe way to do this is to not expose the pointer at all...

This is not different in the PR btw.

jasonish: I believe you were asking about some datatype a while ago that would make it impossible to hold onto? Makes sense here, but I don't think its possible.

rust/src/smb/smb.rs
```diff
@@ -53,6 +53,82 @@ use crate::smb::events::*;
 use crate::smb::files::*;
 use crate::smb::smb2_ioctl::*;
 
+#[repr(C)]
+pub enum SMBFrameType {
+    NBSS = 0,
+    NBSSHdr = 1,
+    NBSSData = 2,
```
jasonish, line 60: I think its time to make a derive macro for this, I don't see a way to do it purely with generics..  I'm thinking something like:
```
#[repr(C)]
#[derive(AppLayerFrame)]
pub enum SMBFramtType {
    #[frame(name = "nbss.pdu")
    NBSS,
    #[frame(name = "nbss.hdr")
    NBSSHdr,
}
```

We could get rid of the name by having a transformation method, like applayerevents use..

NbssPdu => "nbss.pdu"
Smb3Hdr => "smb3.hdr"

And do we really need the repr(C) here? It looks like we convert to u8 before passing over FFI in all cases.

rust/src/applayer.rs
victorjulien: I think in the normal case we'll have a small number of frames, so looking up by id will be a minor cost. But if we have a larger number, say a 100 or so, we'd want to avoid iterating all of them for simple operations.

victorjulien: could we somehow force a lifetime requirement on a ptr that can't be satisfied?

jasonish: Maybe.. Will need to wrap the raw pointer in a Rust struct tho that wraps the raw pointer and includes some phantom data, as raw pointers don't have a life time.

rust/src/smb/smb.rs
```diff
@@ -53,6 +53,82 @@ use crate::smb::events::*;
...
+            8 =>  Some(SMBFrameType::SMB2Data),
+            9 =>  Some(SMBFrameType::SMB3),
+            10 => Some(SMBFrameType::SMB3Hdr),
+            11 => Some(SMBFrameType::SMB3Data),
+            _ => None,
+        }
+    }
+}
+
+fn smb_frame_type_string(s: &str) -> i32 {
```
jasonish, line 92: Can we use a single datatype to represent the enum as an integer?  u8?  Looking at the derive macro here.

jasonish:
With respect to generating support code for the crate type enum...
- Here's the trait I've come up with for frame type enums: https://github.com/jasonish/suricata/commit/4da8f4c713afa24a936394bd22ee1ae5d0b2dcba
- Which simplifies `SMBFrameType` to https://github.com/jasonish/suricata/blob/aa03641b1d685af6860e54aba55a34721d6e597a/rust/src/smb/smb.rs#L57 (commit: https://github.com/jasonish/suricata/commit/aa03641b1d685af6860e54aba55a34721d6e597a)
- The commit that adds the derive macro: https://github.com/jasonish/suricata/commit/220a58a4ef0ba311963616106e5137c4b5b2bc75

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_asan |
 

Pipeline 5189

rust/src/smb/smb.rs
victorjulien: storing this as `uint8_t` on the C side.

victorjulien:
Continued in #6706

PR #6686, opened 2021-12-08
stream-tcp-reassemble: fix reassembly direction for FIN packets

amirabell:
Suricata invokes the stream reassembly logic only for the current packet
direction if the packet contains a FIN flag. However, this does not
handle the case in which the packet ACKs data from the opposing direction.
This patch forces the invocation of the stream reassembly logic
on both direction when Suricata sees a FIN packet.

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [ x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4877

Describe changes:
-Set direction to UPDATE_DIR_BOTH FIN-ACK packets
-
-

#suricata-verify-pr:https://github.com/OISF/suricata-verify/pull/599
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

Previous PR: https://github.com/OISF/suricata/pull/6685

PR #6693, opened 2021-12-09
stream-tcp-reassemble: fix reassembly direction for FIN packets

amirabell:
Suricata invokes the stream reassembly logic only for the current packet
direction if the packet contains a FIN flag. However, this does not
handle the case in which the packet ACKs data from the opposing direction.
This patch forces the invocation of the stream reassembly logic
on both direction when Suricata sees a FIN packet.

https://redmine.openinfosecfoundation.org/issues/4877

Describe changes:
-Set direction to UPDATE_DIR_BOTH FIN-ACK packets
-
-

#suricata-verify-pr:OISF/suricata-verify#599
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

Previous PR: https://github.com/OISF/suricata/pull/6686

### DIFF
--- a/src/stream-tcp-reassemble.c
+++ b/src/stream-tcp-reassemble.c
@@ -1846,7 +1846,11 @@ int StreamTcpReassembleHandleSegment(ThreadVars *tv, TcpReassemblyThreadCtx *ra_
     } else if (p->tcph->th_flags & TH_RST) { // accepted rst
         dir = UPDATE_DIR_PACKET;
     } else if ((p->tcph->th_flags & TH_FIN) && ssn->state > TCP_TIME_WAIT) {
-        dir = UPDATE_DIR_PACKET;
+        if (p->tcph->th_flags & TH_ACK) {
+            dir = UPDATE_DIR_BOTH;
+        } else {
+            dir = UPDATE_DIR_PACKET;
+        }
     } else if (ssn->state == TCP_CLOSED) {
         dir = UPDATE_DIR_BOTH;
     }


catenacyber:
@amirabell could you check the CI failures ?

amirabell:
> @amirabell could you check the CI failures ?

It seems that the errors are related to the "pcap_cnt". For instance, consider the failure on the dnp3-en-spon test (sub test 2). The event is expected at "pcap_cnt" 9 but it is generated at "pcap_cnt" 7. This is due to the fact that, with the above patch, suricata updates both directions of the stream upon receiving the first FIN-ACK (from the client in this case). 

Please, let me know if this is acceptable for you.

catenacyber:
> > @amirabell could you check the CI failures ?
> 
> It seems that the errors are related to the "pcap_cnt". For instance, consider the failure on the dnp3-en-spon test (sub test 2). The event is expected at "pcap_cnt" 9 but it is generated at "pcap_cnt" 7. This is due to the fact that, with the above patch, suricata updates both directions of the stream upon receiving the first FIN-ACK (from the client in this case).
> 
> Please, let me know if this is acceptable for you.

Thanks for the analysis.
This looks indeed like a good side-effect of the fix.

Could you make a Suricata-verify PR that will fix the tests ? 
I wonder if we should be changing `pcap_cnt` from 9 to 7 or just removing `pcap_cnt`...
That depends on wether we want to back port this fix I guess...

@jasonish how do you feel about this ?

amirabell:
> > > @amirabell could you check the CI failures ?
> > 
> > 
> > It seems that the errors are related to the "pcap_cnt". For instance, consider the failure on the dnp3-en-spon test (sub test 2). The event is expected at "pcap_cnt" 9 but it is generated at "pcap_cnt" 7. This is due to the fact that, with the above patch, suricata updates both directions of the stream upon receiving the first FIN-ACK (from the client in this case).
> > Please, let me know if this is acceptable for you.
> 
> Thanks for the analysis. This looks indeed like a good side-effect of the fix.
> 
> Could you make a Suricata-verify PR that will fix the tests ? I wonder if we should be changing `pcap_cnt` from 9 to 7 or just removing `pcap_cnt`... That depends on wether we want to back port this fix I guess...
> 
> @jasonish how do you feel about this ?

Hi @catenacyber, sure I will upload a new PR as soon as we hear from @jasonish, thanks.

amirabell:
Uploaded new suricata-verify PR: https://github.com/OISF/suricata-verify/pull/639

catenacyber:
Thanks.
The best way to test, would be to create a new PR on top of latest master, and have
`suricata-verify-pr:639` in the description (without the # that indicates a comment)
Could you do that ?

PR #6696, opened 2021-12-09
frames: trait and derive macro - v3

jasonish:
Based on https://github.com/OISF/suricata/pull/6684.

Adds an `AppLayerFrameType` trait with a derive macro for use in plain enums. As can be seen in the patch to the SMB parser, all boilerplate around the `SMBFrameType` enum can be removed.


victorjulien:
Part of #6706

PR #6694, opened 2021-12-09
Fuzz template 4125 v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4125

Describe changes:
- fuzz: enable template protocols
- dns : wrap with `HAVE_LUA` some code
- fuzz : use parsed rules in sigpcap target (as already fixed for aware target)

Replaces #6692 

suricata-verify-pr: 602
https://github.com/OISF/suricata-verify/pull/602
With https://github.com/google/oss-fuzz/pull/6990 the point is to increase fuzzing coverage

PR #6697, opened 2021-12-09
Fuzz template 4125 v4

catenacyber:

Describe changes:
- fuzz: enable template protocols
- dns : wrap with `HAVE_LUA` some code
- fuzz : use parsed rules in sigpcap target (as already fixed for aware target)
- detect: do not upgrade base64 decode when fuzzing
- detect: logs an error if a protocol is disabled 

Maybe this PR can be split in different ones.
Like the ` detect: logs an error if a protocol is disabled `, which could then have a different ticket as well ?

Replaces #6694
- fixes the bug found by CIFuzz in base64 (over allocation with a dummy signature)
- fixes the unit tests by having a special case for HTTP_ANY signatures

suricata-verify-pr: 602
https://github.com/OISF/suricata-verify/pull/602
The global point is to increase fuzzing coverage


src/detect-dns-query.c
```diff
@@ -199,6 +198,23 @@ static int PrefilterMpmDnsQueryRegister(DetectEngineCtx *de_ctx,
...
+}
+
+static int DetectEngineInspectDnsResponse(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx,
+        const struct DetectEngineAppInspectionEngine_ *engine, const Signature *s, Flow *f,
+        uint8_t flags, void *alstate, void *txv, uint64_t tx_id)
+{
+    return DetectEngineInspectGenericList(
+            de_ctx, det_ctx, s, engine->smd, f, flags, alstate, txv, tx_id);
+}
+#endif
```
jasonish, line 217: These are only used for Lua scripting then?

catenacyber: Indeed.
See also below previous comment :

> register these generic engines from here for now

suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.memuse | 541517248 | 505916608 | 107.04% | 

Pipeline 5303

catenacyber:
> fuzz : use parsed rules in sigpcap target (as already fixed for aware target)

By the way, this was what prevented fuzzing to find https://redmine.openinfosecfoundation.org/issues/4503

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6717

PR #6678, opened 2021-12-06
Dataset user v5

regit:
Rebased of #5859 on latest master.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4239

Describe changes:
- rebase on master

PR #6698, opened 2021-12-09
datasets: initialize after dropping privileges - v6

jasonish:
Alternate fix to https://redmine.openinfosecfoundation.org/issues/4239.

While testing I noticed that datasets defined within a rule didn't have the issue in the bug, just datasets defined in the `suricata.yaml` so I thought I'd try just moving where the datasets are initialized. This appears to fix the issue, and I tested that it still works with unix socket.

Ticket: https://redmine.openinfosecfoundation.org/issues/4239
Eric's PR: https://github.com/OISF/suricata/pull/6678


victorjulien:
Merged through #6707, thanks!

PR #6706, opened 2021-12-13
Feat/appl records/v57

victorjulien:
extends #6684, includes #6696

adds basic HTTP frame support

suricata-verify-pr: 608


victorjulien:
Replaced by #6713

PR #6652, opened 2021-11-29
Integrate dpdk runmode v9

lukashino:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/4507) ticket:

Describe changes:
- Integrate DPDK capture interface as an alternative to AF_PACKET or others.
- Focus implementation on workers run mode.
- Support both IDS and IPS modes.
- Follow AF_PACKET capture interface architecture and apply DPDK best practices for development.
- Update documentation to reflect new changes.
- Edit Github workflow to include build for the DPDK capture interface.

Work relies on pkg-config and requires at least DPDK 19.11. Work is focused on adding support for physical NICs. At the same time, it should be run natively (and not in docker or in virtual machines) as a primary process. To support multiple workers, the NICs need to support RSS offload.

Changelog from the previous [PR](https://github.com/OISF/suricata/pull/6611):
- created new PMD specific util files that - for now - contains setting of RSS setting on different Intel NICs. Mellanox does not require such files so far. 
- refactoring and responding to Victor's comments
- cherry-picked Victor's comments for easier debugging when RSS hash functions are not equal
- updated counters

PR #6708, opened 2021-12-14
Next/20211214/v6

victorjulien:
#6707, plus DPDK work by @lukashino #6652. The DPDK commits have been cleaned up a bit and a few compile warnings have been addressed.


amaan-fmad:
Congrats on upstreaming the DPDK work @lukashino, great work!

@victorjulien @lukashino I'd left a comment for a minor issue where the DPDK mode seems to log all received packets as dropped mistakenly (they seem to process fine, it's just the log that adds received packet counts to the dropped counter erroneously, I believe):
https://github.com/OISF/suricata/pull/6652/files#r765566286

In case the direct link doesn't work (it should once you jump to `source-dpdk.c` and click "Load diff"):
Screenshot from 2021-12-15 16-18-09

PR #6711, opened 2021-12-14
fix: dcerpc dce_iface doesn't match the rpc requests/reponses - v2

zer1t0:
Previous PR: #6511

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4769

SV test: https://github.com/OISF/suricata-verify/pull/571

The bug:
The dcerpc dce_iface keyword just match the packet following the bind. Only the next request after the rpc is sent will match. However the expected behaviour it that all the rpc requests/responses sent under the context of the given interface would match.

This is happening because the interfaces are been deleted from dcerpc state after the first request in the [handle_bind_cache](https://github.com/OISF/suricata/blob/master/rust/src/dcerpc/dcerpc.rs#L546) function. Not sure why this function was added, since the bindack is the context of the state, not a cached value. My guess, a misinterpretation how the protocol works, since I've check this function was added after moving the code to rust (maybe following the example of the smb handling, in which also [deletes the bind state erroneously](https://github.com/OISF/suricata/pull/6509)).

Anyway, the fact is that the messages are not matching so I removed the function.

As collateral effect the SV test dcerpc-iface-01 is now failing, since the dce_iface now matches both the request and response (not only the first request after bind).


inashivb:
Merged as a part of https://github.com/OISF/suricata/pull/6860. Thank you for your work! :)

PR #6717, opened 2021-12-16
Fuzz template 4125 v5

catenacyber:

Describe changes:
- fuzz: enable template protocols
- dns : wrap with `HAVE_LUA` some code
- fuzz : use parsed rules in sigpcap target (as already fixed for aware target)
- fuzz : cleans flows after each run for one target
- detect: do not upgrade base64 decode when fuzzing
- detect: logs an error if a protocol is disabled 

Replaces #6697
- adds latest commit to clean flows after each fuzzer input run


suricata-qa:

Pipeline 5335

src/detect-parse.c
```diff
@@ -1507,6 +1507,11 @@ int DetectSignatureSetAppProto(Signature *s, AppProto alproto)
         }
     }
 
+    if (AppLayerProtoDetectGetProtoName(alproto) == NULL && alproto != ALPROTO_HTTP) {
```
victorjulien, line 1510: why is there an exception for `ALPROTO_HTTP`?

src/detect-base64-decode.c
```diff
@@ -233,6 +233,9 @@ static int DetectBase64DecodeSetup(DetectEngineCtx *de_ctx, Signature *s,
         data->bytes = BASE64_DECODE_MAX;
     }
     if (data->bytes > de_ctx->base64_decode_max_len) {
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
victorjulien, line 236: should we enforce a limit for the non-fuzz case as well?

src/detect-parse.c
catenacyber: Because we can have rules with `alert http` but there is no such app-layer protocol : there is http1 and http2
So, we can improve this by testing that `alert http` rules need to have at least one of http1 or http2 enabled...
What do you think ?

src/detect-base64-decode.c
catenacyber: Well, this comes from the signature...
Maybe there is a use case to get it big.
Or we can limit it to `BASE64_DECODE_MAX`
As you feel...

victorjulien:
merged all except https://github.com/OISF/suricata/pull/6717/commits/334f142dbb603b4249e76d264e115b2df64bd337 in #6810

src/detect-parse.c
victorjulien: I don't think I understand. Shouldn't every `alproto` value lead to a non-null result for `AppLayerProtoDetectGetProtoName`?

catenacyber: > Shouldn't every alproto value lead to a non-null result for AppLayerProtoDetectGetProtoName?`

Protocols disabled in the configuration lead to a NULL value cf `AppLayerProtoDetectRegisterProtocol`
There is `AppProtoToString` which returns a non NULL value for every alproto

Does that answer your question ?

catenacyber:
Replaced with https://github.com/OISF/suricata/pull/6818

src/detect-parse.c
victorjulien: What kind of rule(+config?) case would hit this condition?

catenacyber: Rules like
```
alert ip any any -> any any (enip_command: 112; sid:2; )
alert ip any any -> any any (dnp3_func: 1; sid:3; )
```

with default config (where DNP3 and ENIP are disabled)

victorjulien: I meant the case where we the `alproto != ALPROTO_HTTP` condition.

victorjulien: Could `AppLayerProtoDetectGetProtoName` return "http_any", "http1" or "http2" depending on which is enabled?

catenacyber: Tried in https://github.com/OISF/suricata/pull/6888

PR #6726, opened 2021-12-19
output/json: Eliminate dangling XFF reference

jlucovsky:
This commit eliminates a dangling reference caused by the use of
json_object_set. This function adds a reference to the final parameter
-- in this case the object returned by json_string() whereas
json_object_set_new doesn't add the additional reference to the
final parameter.

Describe changes:
- Use `json_object_set_new` instead of `json_object_new`.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:


victorjulien:
Merged in #6727, thanks

PR #6732, opened 2021-12-22
devguide: move into userguide as last chapter - v1

jufajardini:
DevGuide now shows up as last chapter (25th) of UserGuide.

Not sure if this change is good as is, as lack of mscgen could locally break the whole userguide, now. Should I add a check to prevent building the devguide, in such cases?

Describe changes:
- bring devguide dir into userguide dir
- edit script for generating transaction diagrams to adjust location
- adjust Makefiles accordingly
- **opportunistic commit:** devguide: rust-testing: clarify path to calling tests with cargo test

html preview at: [temp-suri-user-and-devguide.surge.sh](http://temp-suri-user-and-devguide.surge.sh/)


victorjulien:
The CentOS8 builder fails because of missing `mscgen`.

jufajardini:
> The CentOS8 builder fails because of missing `mscgen`.

That made me wonder f this change is good as is, as lack of mscgen could locally break the whole userguide, then. 

Should I add a check to prevent building the devguide, in such cases?

jufajardini:
New attempt with https://github.com/OISF/suricata/pull/6917

PR #6733, opened 2021-12-22
detect: fix app-layer-protocol keyword for HTTP

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4920

Describe changes:
- Use `AppProtoEquals` for app-layer-protocol keyword to fix behavior for HTTP (any version 1 or 2)

Completes commit c8dbe24fb6202550bbca1fab452ddbe864b2c9e2
This commit missed this case, as I only looked for `git grep "alproto =="` and here we deal with alproto_tc and alproto_ts, but not alproto by itself.

Interestingly commit dd32238667f08c7211ae4fa27cfe43af7cffd52d was also about `alproto_t*` rather than directly alproto

suricata-verify-pr: 615

CI will fail as this does not fix https://redmine.openinfosecfoundation.org/issues/4921
So, 2 out of the 3 S-V tests pass

### DIFF
--- a/src/detect-app-layer-protocol.c
+++ b/src/detect-app-layer-protocol.c
@@ -43,7 +43,7 @@ static int DetectAppLayerProtocolPacketMatch(
 {
     SCEnter();
 
-    int r = 0;
+    bool r = false;
     const DetectAppLayerProtocolData *data = (const DetectAppLayerProtocolData *)ctx;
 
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 5388

victorjulien:
Merged in #6744, thanks!

PR #6734, opened 2021-12-26
detect: Avoid recomputing ntohl() in addr match

jlucovsky:
This commit makes a small optimization when comparing IPv4 and IPv6
addresses by making the host order value invariant and calculating the
value once, before entering the loop.

Describe changes:
- Move `SCNtoHL` outside of loop as it's invariant.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -1644,10 +1644,9 @@ int DetectAddressMatchIPv4(const DetectMatchAddressIPv4 *addrs,
         SCReturnInt(0);
     }
 
+    uint32_t match_addr = SCNtohl(a->addr_data32[0]);
     for (uint16_t idx = 0; idx < addrs_cnt; idx++) {
-        if (SCNtohl(a->addr_data32[0]) >= addrs[idx].ip &&
-            SCNtohl(a->addr_data32[0]) <= addrs[idx].ip2)
-        {
+        if (match_addr >= addrs[idx].ip && match_addr <= addrs[idx].ip2) {
             SCReturnInt(1);
         }
     }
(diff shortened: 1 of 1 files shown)


suricata-qa:
  

Pipeline 5426

src/detect-engine-address.c
```diff
@@ -1678,37 +1677,37 @@ int DetectAddressMatchIPv6(const DetectMatchAddressIPv6 *addrs,
...
     for (uint16_t idx = 0; idx < addrs_cnt; idx++) {
         uint16_t result1 = 0, result2 = 0;
 
         /* See if packet address equals either limit. Return 1 if true. */
-        if (SCNtohl(a->addr_data32[0]) == addrs[idx].ip[0] &&
-            SCNtohl(a->addr_data32[1]) == addrs[idx].ip[1] &&
-            SCNtohl(a->addr_data32[2]) == addrs[idx].ip[2] &&
-            SCNtohl(a->addr_data32[3]) == addrs[idx].ip[3])
-        {
+        if (match_addr[0] == addrs[idx].ip[0] && match_addr[1] == addrs[idx].ip[1] &&
```
victorjulien, line 1693: wonder if we should just use `memcmp` here, at 16 bytes its ideal for simd optimizations (though we might not be aligned correctly, but that will be handled by libc)

jlucovsky: Good point ... this can be done.

jlucovsky:
Continued in #6748

src/detect-engine-address.c
catenacyber: Do you know if it is worth a function call ?

PR #4714, opened 2020-03-20
Include devguide in distributions plus tests - v1

jasonish:
The devguide files were missing from EXTRA_DIST causing them not
to be shipped in a dist archive, as well as causing make dist to
fail if the documentation tooling was discovered during
./configure.

Update a few GitHub CI tests to check this.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR jasonish-pcap: https://buildbot.openinfosecfoundation.org/builders/jasonish-pcap/builds/455
- PR jasonish: https://buildbot.openinfosecfoundation.org/builders/jasonish/builds/812

PR #6735, opened 2021-12-27
nfs2: add unittests for nom7 parsers - v1

ghostinthehive:
partially closes task #4714

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#4714](https://redmine.openinfosecfoundation.org/issues/4714)

Describe changes:
- add unittests to `nfs2_records` nom7 parsers

### DIFF
--- a/rust/src/nfs/nfs2_records.rs
+++ b/rust/src/nfs/nfs2_records.rs
@@ -96,3 +96,211 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
     let attrs = Nfs2Attributes { atype, asize };
     Ok((i, attrs))
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::nfs::nfs2_records::*;
+    use nom7::bytes::streaming::take;
+    use nom7::IResult;
+
+    #[test]
+    fn test_nfs2_handle() {
+
+        // file_handle bytes
+        let buf: &[u8] = &[
+            0x00, 0x10, 0x10, 0x85, 0x00, 0x00, 0x03, 0xe7,
+            0x00, 0x0a, 0x00, 0x00, 0x00, 0x00, 0xb2, 0x5a,
+            0x00, 0x00, 0x00, 0x29, 0x00, 0x0a, 0x00, 0x00,
+            0x00, 0x00, 0xb2, 0x5a, 0x00, 0x00, 0x00, 0x29
+        ];
+
+        let result = parse_nfs2_handle(buf);
+        match result {
+            Ok((r, res)) => {
+                assert_eq!(r.len(), 0);
+                assert_eq!(res.value, buf);
+            }
+            Err(error) => { panic!("Parsing nfs2_handle failed: {:?}", error); }
+        }
+    }
+
+    #[test]
+    fn test_nfs2_request_lookup() {
+
+        // packet_bytes
+        let buf: &[u8] = &[
+
+        // [file_handle]
+            0x00, 0x10, 0x10, 0x85, 0x00, 0x00, 0x03, 0xe7,
+            0x00, 0x0a, 0x00, 0x00, 0x00, 0x00, 0xb2, 0x5a,
+            0x00, 0x00, 0x00, 0x29, 0x00, 0x0a, 0x00, 0x00,
+            0x00, 0x00, 0xb2, 0x5a, 0x00, 0x00, 0x00, 0x29,
+
+        // [name]
+        // name_len
+            0x00, 0x00, 0x00, 0x02,
+
+        // name_contents: (am)
+            0x61, 0x6d,
+
+        // _name_padding
+            0x00, 0x00,
+        ];
+
+        let (_, handle) = parse_nfs2_handle(buf).expect("Parsing nfs2_handle failed!");
+
+        let result = parse_nfs2_request_lookup(buf);
+        match result {
+            Ok((r, request)) => {
+                assert_eq!(r.len(), 0);
+                assert_eq!(request.handle, handle);
+                assert_eq!(request.name_vec, b"am".to_vec());
+            }
+            Err(error) => { panic!("Parsing failed: {:?}", error); }
+        }
+    }
+
+    #[test]
+    fn test_nfs2_request_read() {
+
+        // packet_bytes
+        let buf: &[u8] = &[
+
+        // [file_handle]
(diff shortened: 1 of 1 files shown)


rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,211 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+
+#[cfg(test)]
+mod tests {
+    use crate::nfs::nfs2_records::*;
+    use nom7::bytes::streaming::take;
+    use nom7::IResult;
+
+    #[test]
+    fn test_nfs2_handle() {
+
```
ghostinthehive, line 108: not sure if a separate `test_nfs2_handle()` is needed, since we already use `parse_nfs2_handle()` in later tests?

rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,211 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+        ];
+
+        let result = parse_nfs2_handle(buf);
+        match result {
+            Ok((r, res)) => {
+                assert_eq!(r.len(), 0);
+                assert_eq!(res.value, buf);
+            }
+            Err(error) => { panic!("Parsing nfs2_handle failed: {:?}", error); }
+        }
```
inashivb, line 124: Constructs like these could be an assert statement prior to other asserts. Something like:
`assert_eq!(true, result.is_ok())`

rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,211 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+            let i: IResult<&[u8], &[u8]> = take(76_usize)(input);
+            match i {
+                Ok((remainder, _input)) => {
+                        remainder
+                }
+                Err(error) => { panic!("extracting data field failed: {:?}", error); }
+            }
+        }
+
+        let expected_data = extract_data(buf);
```
inashivb, line 240: Are you trying to extract all the remaining bytes after 76 bytes here? You could do it in a one liner (within the assert statement later on) with array slices. `input[76..]`

rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,211 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+
+        // data_len: (11)
+            0x00, 0x00, 0x00, 0x0b,
+
+        // data_contents: ("the b file")
+            0x74, 0x68, 0x65, 0x20,
+            0x62, 0x20, 0x66, 0x69, 0x6c, 0x65, 0x0a,
+
+        // _data_padding: parsed as part of data_contents
+            0x00,
```
inashivb, line 223: I'm not sure about breaking down the data like this. I think the con is too many newlines and lines in general. But, a big pro is legible fields. I'll let Victor or Jason decide this.

rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,211 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+                assert_eq!(request.handle, handle);
+                assert_eq!(request.offset, 0);
+            }
+            Err(error) => { panic!("Parsing failed: {:?}", error); }
+        }
+    }
+
+    #[test]
+    fn test_nfs2_reply_read() {
+
```
inashivb, line 198: We could perhaps add the pcap data alongside the hex bytes and the source here. Excellent example: https://github.com/OISF/suricata/blob/master/rust/src/dns/dns.rs#L1013

rust/src/nfs/nfs2_records.rs
inashivb: This seems good to me since it is a separate unit too.

inashivb (changes requested):
All in all, LGTM. Good job! :)
A few suggestions and changes requested from me inline.

rust/src/nfs/nfs2_records.rs
ghostinthehive: I'll consider this with the other PR too, thanks!

rust/src/nfs/nfs2_records.rs
ghostinthehive: this seems like testing against the buffer as a whole. should I discard specifying the fields?
I understand the tests are too verbose, trying to figure out how to make it compact and clear at the same time

I found this one with comments in the middle.
https://github.com/OISF/suricata/blob/4972021cd4c362c6f02fce6c34377013a463eab0/rust/src/dns/dns.rs#L1234

rust/src/nfs/nfs2_records.rs
ghostinthehive: Now that's a savior one liner! thanks! :)
I'll adopt this from now on.

rust/src/nfs/nfs2_records.rs
jasonish: I like it.  Can help when you're debugging a test someone else wrote and you're not that familiar with the contents, as you generally can't just paste this into wireshark to get the break down.

rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,211 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+            0x00, 0x00, 0xa3, 0xe7,
+            0x38, 0x47, 0x75, 0xea,
+            0x00, 0x08, 0x64, 0x70,
+            0x38, 0x47, 0x75, 0xea,
+            0x00, 0x08, 0x64, 0x70,
+            0x38, 0x47, 0x75, 0xea,
+            0x00, 0x08, 0x64, 0x70,
+        ];
+
+        let result = parse_nfs2_attribs(buf);
```
jasonish, line 295: IMO, its perfectly acceptable to just use `.unwrap()` here if we just end up panic'ing in the error case.  The output is clear with a line number that gets you directly where you need to get.  What you have here is fine as well, its just a little more work.

ghostinthehive:
Followed by #6740

PR #6737, opened 2021-12-28
nfs3: add unittests for nom7 parsers - v1

ghostinthehive:
partially closes task #4714

Link to redmine [ticket](https://redmine.openinfosecfoundation.org/issues/4714):

Describe changes:
- add unittests to `nfs3_records`


rust/src/nfs/nfs3_records.rs
```diff
@@ -256,7 +256,7 @@ pub fn parse_nfs3_response_readdirplus_entry(
     let (i, name_contents) = take(name_len as usize)(i)?;
     let (i, _fill_bytes) = cond(name_len % 4 != 0, take(4 - (name_len % 4)))(i)?;
     let (i, _cookie) = take(8_usize)(i)?;
-    let (i, attr_value_follows) = be_u32(i)?;
+    let (i, attr_value_follows) = be_u32(i)?;   // packet with this structure?
     let (i, _attr) = cond(attr_value_follows == 1, take(84_usize))(i)?;
     let (i, handle_value_follows) = be_u32(i)?;
     let (i, handle) = cond(handle_value_follows == 1, parse_nfs3_handle)(i)?;
```
ghostinthehive, line 262: I'm not sure I can find a packet for this message type with that structure?
the `READDIR Reply` ends with an entry/ couple of entries and an `EOF` field
no handle field in this message type.

rust/src/nfs/nfs3_records.rs
```diff
@@ -301,36 +301,36 @@ pub(crate) fn many0_nfs3_response_readdirplus_entries<'a>(
...
     pub handle: Nfs3Handle<'a>,
-
-    pub cookie: u32,
+    //pub cookie: u32,
+    pub cookie: u64, // cookie is a u64 size
     pub verifier: &'a [u8],
     pub dircount: u32,
-    pub maxcount: u32,
+    //pub maxcount: u32,
 }
```
ghostinthehive, line 313: `cookie` field at `READDIR Call` is of 8byte size, and no `maxcount` field.

I left the original code as commented for reviewing.

jufajardini (commented):
Overall, it looks like a good job, to me!

One thing that I think we prefer not to do is to have a commit that changes things in the parser, while the commit message says it's just adding tests, though. But better see what Victor, Jason and Shivani have to say :)

rust/src/nfs/nfs3_records.rs
ghostinthehive: Update: looks like `readdirplus Reply` and `readdir Reply` are different messages, and I was testing `parse_nfs3_response_readdir_plus_entry()` against `readdir Reply`.

-- yet should the `readdir reply` message have a separate parser though?

TODO:
- search for a pcap with `readdirplus reply` message to test against.

inashivb: We often do not parse everything that a protocol might provide. Only the important fields and types relevant to our system and users. From what I understand, for anything that is not a particularly parsed type, we simply log the status: https://github.com/OISF/suricata/blob/master/rust/src/nfs/nfs3.rs#L261 and the common fields that come irrespective of the type of procedure and/or version https://github.com/OISF/suricata/blob/master/rust/src/nfs/log.rs#L155

jasonish:
> Overall, it looks like a good job, to me!
> 
> One thing that I think we prefer not to do is to have a commit that changes things in the parser, while the commit message says it's just adding tests, though. But better see what Victor, Jason and Shivani have to say :)

If something changes in the parser, it can be in the same commit as a test that needs to change to pass after the change in the parser, or a test that is added to verify that change.  But this commit should be isolated to its own commit to describe why the change is needed, and not bundled with other commits that are just adding tests.

rust/src/nfs/nfs3_records.rs
```diff
@@ -386,3 +388,703 @@ pub fn parse_nfs3_reply_read(i: &[u8]) -> IResult<&[u8], NfsReplyRead> {
...
+            0x00, 0x10, 0x10, 0x85, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0xa3, 0xe7, 0x38, 0x47, 0x76, 0x25,
+            0x21, 0xf9, 0x82, 0x82, 0x38, 0x47, 0x76, 0x25,
+            0x22, 0x92, 0x19, 0x00, 0x38, 0x47, 0x76, 0x25,
+            0x22, 0x92, 0x19, 0x00,
+        */
+        ];
+
+        // asserion: (1) is this a good practice?
+        // manually extract handle from packets each time we parse against it
```
inashivb, line 463: I think manual extraction will be required for making the data to match against. We have to generate expected data somehow at some point but you could save `&buf[16..]` here in `value`.

rust/src/nfs/nfs3_records.rs
```diff
@@ -386,3 +388,703 @@ pub fn parse_nfs3_reply_read(i: &[u8]) -> IResult<&[u8], NfsReplyRead> {
...
+            0x00, 0x00, 0x41, 0xed, 0x00, 0x00, 0x00, 0x02,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x60,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x10, 0x10, 0x85,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xb2, 0x5a,
+            0x38, 0x47, 0x76, 0x25, 0x20, 0x2f, 0xbf, 0x00,
+            0x38, 0x47, 0x76, 0x25, 0x25, 0x8d, 0x09, 0x82,
+            0x38, 0x47, 0x76, 0x25, 0x25, 0x8d, 0x09, 0x82 */
```
inashivb, line 529: Then, we can probably skip it altogether. ;)

inashivb (commented):
LGTM. Good job. :) Some comments inline and the same comments from #6735 + the `readdirplus` changes that might need to take effect as per our discussion.

rust/src/nfs/nfs3_records.rs
ghostinthehive: @inashivb are you suggesting to manually extract each time we need to match against  `Nfs3Handle` struct?

inashivb: @ghostinthehive if there's another and/or better way to do this, I'll be happy to learn about that. :) But for now, yes, it seems like the way to me.

rust/src/nfs/nfs3_records.rs
victorjulien: I see maxcount in my pcaps, why do you believe the maxcount is incorrect?

rust/src/nfs/nfs3_records.rs
```diff
@@ -301,36 +301,36 @@ pub(crate) fn many0_nfs3_response_readdirplus_entries<'a>(
     many0(complete(parse_nfs3_response_readdirplus_entry_cond))(input)
 }
 
+
 #[derive(Debug, PartialEq)]
 pub struct Nfs3RequestReaddirplus<'a> {
     pub handle: Nfs3Handle<'a>,
-
-    pub cookie: u32,
+    //pub cookie: u32,
```
victorjulien, line 308: cleanup: remove the comments here

rust/src/nfs/nfs3_records.rs
victorjulien: `nfs3-01` in SV has this packet type

rust/src/nfs/nfs3_records.rs
ghostinthehive: never mind those, that was when I assumed that the `**plus` record type do not exist, and I left the comments etc, as a draft, my bad!

ghostinthehive:
Followed by  #6774

PR #6740, opened 2022-01-02
nfs2: add unittests for nom7 parsers v2

ghostinthehive:
Previous PR #6735 

partially closes task #4714

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#4714](https://redmine.openinfosecfoundation.org/issues/4714)

Changes from Previous PR:
- Use slicing i.e `&input[x...]` for more compact code.
- Use `.unwrap` to just continue with the unwrapped `Ok()` result type, and panic if `Err() `.

PR #6743, opened 2022-01-04
nfs2: add unittests for nom7 parsers v3

ghostinthehive:
Previous PR #6740 

Task #4714

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#4714](https://redmine.openinfosecfoundation.org/issues/4714)

Changes from Previous PR:
- adjust formatting to be more compact.
- add checks for `Nfs2Handle` fields.

### DIFF
--- a/rust/src/nfs/nfs2_records.rs
+++ b/rust/src/nfs/nfs2_records.rs
@@ -96,3 +96,159 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
     let attrs = Nfs2Attributes { atype, asize };
     Ok((i, attrs))
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::nfs::nfs2_records::*;
+
+    #[test]
+    fn test_nfs2_handle() {
+
+        // file_handle bytes
+        let buf: &[u8] = &[
+            0x00, 0x10, 0x10, 0x85, 0x00, 0x00, 0x03, 0xe7,
+            0x00, 0x0a, 0x00, 0x00, 0x00, 0x00, 0xb2, 0x5a,
+            0x00, 0x00, 0x00, 0x29, 0x00, 0x0a, 0x00, 0x00,
+            0x00, 0x00, 0xb2, 0x5a, 0x00, 0x00, 0x00, 0x29
+        ];
+
+        let result = parse_nfs2_handle(buf).unwrap();
+        match result {
+            (r, res) => {
+                assert_eq!(r.len(), 0);
+                assert_eq!(res.value, buf);
+            }
+        }
+    }
+
+    #[test]
+    fn test_nfs2_request_lookup() {
+
+        let buf: &[u8] = &[
+        // [file_handle]
+            0x00, 0x10, 0x10, 0x85, 0x00, 0x00, 0x03, 0xe7,
+            0x00, 0x0a, 0x00, 0x00, 0x00, 0x00, 0xb2, 0x5a,
+            0x00, 0x00, 0x00, 0x29, 0x00, 0x0a, 0x00, 0x00,
+            0x00, 0x00, 0xb2, 0x5a, 0x00, 0x00, 0x00, 0x29,
+        // [name]
+            0x00, 0x00, 0x00, 0x02, 0x61, 0x6d, 0x00, 0x00,
+        ];
+
+        let (_, handle) = parse_nfs2_handle(buf).expect("Parsing nfs2_handle failed!");
+        assert_eq!(handle.value, &buf[..32]);
+
+        let result = parse_nfs2_request_lookup(buf).unwrap();
+        match result {
+            (r, request) => {
+                assert_eq!(r.len(), 0);
+                assert_eq!(request.handle, handle);
+                assert_eq!(request.name_vec, b"am".to_vec());
+            }
+        }
+    }
+
+    #[test]
+    fn test_nfs2_request_read() {
+
+        let buf: &[u8] = &[
+            0x00, 0x10, 0x10, 0x85, 0x00, 0x00, 0x03, 0xe7, // [file_handle]
+            0x00, 0x0a, 0x00, 0x00, 0x00, 0x00, 0xb2, 0x5d,
+            0x00, 0x00, 0x00, 0x2a, 0x00, 0x0a, 0x00, 0x00,
+            0x00, 0x00, 0xb2, 0x5a, 0x00, 0x00, 0x00, 0x29,
+            0x00, 0x00, 0x00, 0x00,                         // offset
+            0x00, 0x00, 0x20, 0x00,                         // count
+            0x00, 0x00, 0x20, 0x00,                         // total count
+        ];
+
+        let (_, handle) = parse_nfs2_handle(buf).expect("Parsing nfs2_handle failed!");
+        assert_eq!(handle.value, &buf[..32]);
+
+        let result = parse_nfs2_request_read(buf).unwrap();
+        match result {
+            (r, request) => {
+                assert_eq!(r.len(), 4);
+                assert_eq!(request.handle, handle);
+                assert_eq!(request.offset, 0);
+            }
+        }
+    }
+
+    #[test]
+    fn test_nfs2_reply_read() {
+
+        let buf: &[u8] = &[
+        // Status: NFS_OK - (0)
+            0x00, 0x00, 0x00, 0x00,
+        // attr_blob
+            0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x81, 0xa4,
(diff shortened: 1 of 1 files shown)


rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,159 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+    fn test_nfs2_request_lookup() {
+
+        let buf: &[u8] = &[
+        // [file_handle]
+            0x00, 0x10, 0x10, 0x85, 0x00, 0x00, 0x03, 0xe7,
+            0x00, 0x0a, 0x00, 0x00, 0x00, 0x00, 0xb2, 0x5a,
+            0x00, 0x00, 0x00, 0x29, 0x00, 0x0a, 0x00, 0x00,
+            0x00, 0x00, 0xb2, 0x5a, 0x00, 0x00, 0x00, 0x29,
+        // [name]
+            0x00, 0x00, 0x00, 0x02, 0x61, 0x6d, 0x00, 0x00,
```
ghostinthehive, line 134: IMO here we have 3 fields in one line, it's not clear what's the length of `name_contents` since we assert against it, it would take time to grab the relevant bytes and check for the value since we have a _not clear_ hidden `_name_padding` byte.

ghostinthehive:
I left different formatting for commenting structure fields ...etc, please let me know which formatting you think is better to adopt it with other tests

for me an example of https://github.com/OISF/suricata/pull/6743/commits/26143c18c14f4543985b89cd9629cff398a962d8#r777866370 is a bit vague although previously it was weird to have just one byte at a separate line, but I'm not sure how to count the trade offs?

while working with the tests I found my eye catches the fields' bytes while being separated with comments as in `test_nfs2_reply_read()` faster than  jumping around the parser structs to figure out the size of respective fields ..etc

rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,159 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+            0x00, 0x0b, 0x71, 0xb0, 0x38, 0x47, 0x71, 0xc4,
+            0x00, 0x08, 0xb2, 0x90, 0x38, 0x47, 0x75, 0xea,
+            0x00, 0x09, 0x00, 0xb0,
+        // data_len: (11)
+            0x00, 0x00, 0x00, 0x0b,
+        // data_contents: ("the b file")
+            0x74, 0x68, 0x65, 0x20,
+            0x62, 0x20, 0x66, 0x69, 0x6c, 0x65, 0x0a,
+        // _data_padding: parsed as part of data_contents
+            0x00,
```
ghostinthehive, line 198: I had to specify this `_data_padding` byte here since it confused me while testing against `data_contents` as it was parsed with it and the assertion failed due to mismatched sizes. 
`data_contents` in the parser is larger than its respective PCAP field with one byte.

jufajardini: Could this be solved if the comment line was `// data_contents: ("the b file") + _data_padding`, or do you think that wouldn't be clear enough?

rust/src/nfs/nfs2_records.rs
jufajardini: `rustfmt` suggested this style in my code: https://github.com/OISF/suricata/blob/5dec901bb7ae6a37f873c42260653f20279222cd/rust/src/pgsql/parser.rs#L1012

Although I can't say that I love it, it does allow for having portions of data identified by comments without adding too many additional lines to the tests...

rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,159 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+            0x00, 0x00, 0x00, 0x00,
+            0x00, 0x10, 0x10, 0x85,
+            0x00, 0x00, 0xa3, 0xe7,
+            0x38, 0x47, 0x75, 0xea,
+            0x00, 0x08, 0x64, 0x70,
+            0x38, 0x47, 0x75, 0xea,
+            0x00, 0x08, 0x64, 0x70,
+            0x38, 0x47, 0x75, 0xea,
+            0x00, 0x08, 0x64, 0x70,
+        ];
```
jufajardini, line 243: If `rustfmt` doesn't complain, maybe what I shared for the other buffer could be used here, as well, if reducing number of lines is a goal. ^^

rust/src/nfs/nfs2_records.rs
```diff
@@ -96,3 +96,159 @@ pub fn parse_nfs2_attribs(i: &[u8]) -> IResult<&[u8], Nfs2Attributes> {
...
+            (r, response) => {
+                assert_eq!(r.len(), 0);
+                assert_eq!(response.status, 0);
+                assert_eq!(response.attr_follows, 1);
+                assert_eq!(response.attr_blob.len(), 68);
+                assert_eq!(response.count, response.data_len);
+                assert_eq!(response.eof, false);
+                assert_eq!(response.data_len, 11);
+                assert_eq!(response.data, "the b file\n\0".as_bytes());
+                assert_eq!(response.data, &buf[76..]);
```
jufajardini, line 212: Curious: is there a reason to have this tested as well, not just the assertion on line 211?

rust/src/nfs/nfs2_records.rs
ghostinthehive: that what `rustfmt` suggested with my code, and I cannot say that I love it too :) , as I mentioned I dont know how to count the trade off with regard to this.

rust/src/nfs/nfs2_records.rs
ghostinthehive: I wouldn't mind, If possible can you share what exactly the goals with regard to formatting and style.
- what is considered (too verbose) ?
- what is considered (too vague) ?

rust/src/nfs/nfs2_records.rs
jufajardini: Jason has recommended/ asked me to run rustfmt on my pgsql code, so I think we're ok with the outcome, even if sometimes it wouldn't be our personal choice :P

rust/src/nfs/nfs2_records.rs
ghostinthehive: the string to test against had invalid characters, something like `\n - newline` '\0 - null' .. etc
I wanted to make sure that the fields have exactly the same parsable bytes. 
do you think this is redundant?

rust/src/nfs/nfs2_records.rs
jufajardini: I'll leave that to the other reviewers to give the final word. My take is: I go with rustfmt tells me to do. We enforce format style on the c side with clang-format, and if I recall correctly, we want to do the same with rust, with rustfmt, so...

Another general rule that I notice is that it is preferred to have fewer lines of code, when possible.

rust/src/nfs/nfs2_records.rs
jufajardini: If _both_ should pass and we are `asserting equal` on the same field, I _think_ we could have just one of the checks. But I don't feel confident giving the final word on this.

ghostinthehive:
@jufajardini  Thanks for your feedback, as I sort of mentioned, here is what I think about:
- "is my code readable to other people debugging after me? does it take time to catch different fields in a struct and  know the value?"
- "what is too verbose and naive?"
- "what is too compact and vague?"

Most of the unittests that already exist have small buffer with few fields so maybe I didn't think its a good fit, but please let me know the preferred format/style so I can move on with it.
_I wouldn't have a good preference since I already wrote the code so I kind of know the drill with these specific buffers_ so I need a second eye on this :)

rust/src/nfs/nfs2_records.rs
victorjulien: The issue here is that the parser isn't correct. If you look at the pcap in wireshark you'll see the `fill bytes` logic. I suspect it fills bytes to make the payload a multiple of 4.

victorjulien: https://github.com/OISF/suricata/pull/6764/commits/ddf14e51dcf5dabdb4bc07414a118530e2a1b819

victorjulien: I've pushed ddf14e5 to master, can you do a new PR where you rebase your branch to master?

ghostinthehive: Ok good, thanks!

ghostinthehive:
Followed by #6776

PR #6746, opened 2022-01-05
detect-uricontent: convert tests to suricata verify

TheKharleeci:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4911

Describe changes:
- convert tests to Suricata-verify

suricata-verify-pr: 627


victorjulien:
I think the git message should not say "convert", as it suggests it is converted to something else in the suricata code. I would say something like:

```
detect/uricontent: remove unittests

These tests are reimplemented as Suricata-Verify tests

Ticket ...
```

src/detect-uricontent.c
```diff
@@ -501,1221 +226,6 @@ static int DetectUriSigTest01(void)
...
-        printf("umatch %p or pmatch %p: ", s->sm_lists[g_http_uri_buffer_id], s->sm_lists[DETECT_SM_LIST_PMATCH]);
-        goto end;
-    }
-
-    if (    ((DetectContentData*) s->sm_lists[DETECT_SM_LIST_PMATCH]->ctx)->depth != 15 ||
-            ((DetectContentData*) s->sm_lists[DETECT_SM_LIST_PMATCH]->ctx)->offset != 5 ||
-            ((DetectContentData*) s->sm_lists_tail[g_http_uri_buffer_id]->ctx)->distance != 30 ||
-            ((DetectContentData*) s->sm_lists_tail[g_http_uri_buffer_id]->ctx)->within != 60 ||
-            ((DetectContentData*) s->sm_lists_tail[DETECT_SM_LIST_PMATCH]->ctx)->distance != 45 ||
-            ((DetectContentData*) s->sm_lists_tail[DETECT_SM_LIST_PMATCH]->ctx)->within != 70 ||
```
inashivb, line 942: @victorjulien is it ok to remove these tests? The signatures are used as a part of tests in s-v so it means they were parsed properly but here we're asserting every param in the signature. Also, it does not involve multi layer integrations testing.

victorjulien: Right, this shouldn't be removed, but should be converted to use FAIL_IF/PASS logic (in a separate commit)

src/detect-uricontent.c
```diff
@@ -501,1221 +226,6 @@ static int DetectUriSigTest01(void)
...
-    UTHFreeFlow(f);
-    UTHFreePackets(&p, 1);
-    StreamTcpFreeConfig(true);
-    PASS;
-}
-
-/**
- * \test Test content for dce sig.
- */
-static int DetectUriSigTest08(void)
```
victorjulien, line 1218: same for these rule parsing tests

src/detect-uricontent.c
inashivb: @TheKharleeci let's keep these signature parsing tests here intact (the ones that are *only* parsing signatures) even if we have used this signatures in the suricata-verify tests. Thank you!

TheKharleeci: Ok. Thank you.

TheKharleeci:
Followed by: #6804

PR #6748, opened 2022-01-06
detect: Avoid recomputing ntohl() in addr match

jlucovsky:
Continuation of #6734 

This commit makes a small optimization when comparing IPv4 and IPv6
addresses by making the host order value invariant and calculating the
value once, before entering the loop.

Describe changes:
- Move `SCNtoHL` outside of loop as it's invariant.

Updates:
- Use `memcmp` for address instead of multi-condition `if` clause

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -1644,10 +1644,9 @@ int DetectAddressMatchIPv4(const DetectMatchAddressIPv4 *addrs,
         SCReturnInt(0);
     }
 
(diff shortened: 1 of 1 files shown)


jlucovsky:
Continued in #6750

PR #6759, opened 2022-01-08
xbits: error out on unaccepted values

inashivb:
If we have unaccepted values as a part of xbits keyword, they are
silently passed when it should fail. This patch compares the total
length of the raw string with the match patterns and errors out if
anything apart from the matched patterns is present.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4786

### DIFF
--- a/src/detect-xbits.c
+++ b/src/detect-xbits.c
@@ -196,7 +196,8 @@ static int DetectXbitParse(DetectEngineCtx *de_ctx,
     DetectXbitsData *cd = NULL;
     uint8_t fb_cmd = 0;
     uint8_t hb_dir = 0;
-    int ret = 0, res = 0;
+    int ret = 0, res = 0, rawstr_len = 0;
+    int args_len = 0;
     size_t pcre2len;
     char fb_cmd_str[16] = "", fb_name[256] = "";
     char hb_dir_str[16] = "";
@@ -209,6 +210,7 @@ static int DetectXbitParse(DetectEngineCtx *de_ctx,
         return -1;
     }
     SCLogDebug("ret %d, %s", ret, rawstr);
+    rawstr_len = strlen(rawstr);
     pcre2len = sizeof(fb_cmd_str);
     res = pcre2_substring_copy_bynumber(
             parse_regex.match, 1, (PCRE2_UCHAR8 *)fb_cmd_str, &pcre2len);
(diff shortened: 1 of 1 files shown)


src/detect-xbits.c
```diff
@@ -271,6 +275,33 @@ static int DetectXbitParse(DetectEngineCtx *de_ctx,
...
+                 * up to 2n+1 values, where n = no. of capturing parentheses */
+                if (bufptr != prev_buf) { // FIXME Dangerous check? Could it be wrong in any cases?
+                    args_len += strlen(bufptr);
+                }
+                prev_buf = bufptr;
+            }
+            if (args_len != stripped_rawstr_len) {
+                SCLogError(SC_ERR_INVALID_VALUE, "Invalid values found in the xbits setting");
+                return -1;
+            }
```
inashivb, line 304: This is probably too long and seemingly dirty way to achieve the fix. Better suggestions?

src/detect-xbits.c
```diff
@@ -546,6 +577,87 @@ static int XBitsTestSig02(void)
...
+    Signature *s = DetectEngineAppendSig(de_ctx,
+            "drop http any any -> any any (msg:\"Malicious file detected\"; \
+            filesha256:blacklist; xbits:set, blocked_http, track ip_pair, \
+            expire 3600; sid:2; rev:2;");
+    FAIL_IF_NULL(s);
+
+    DetectEngineCtxFree(de_ctx);
+    XBitsTestShutdown();
+    PASS;
+}
```
inashivb, line 658: Found this test throw a SEGV in s-v but nowhere in the xbits code, during file hash processing. Gotta know better.

inashivb:
OP says on the ticket
> I suggest adding a bit of context which indicates it should be a standalone keyword in the rule and not an "option" to the xbits keyword.

but I see https://github.com/OISF/suricata/blob/master/src/detect-xbits.c#L277 I _think_ this came from `hostbits` implementation but I'm not sure if it should or should not be an acceptable command for `xbits`?

suricata-qa:
ERROR: ASAN TEST FAIL in tlpr1_asan_cfg QA test

Pipeline 5613

victorjulien:
I think especially for the git master I would like to get rid of the pcre use here, and use a simpler tokenizer instead.

victorjulien:
> I think especially for the git master I would like to get rid of the pcre use here, and use a simpler tokenizer instead.

For example like in `DetectDatasetParse` and `DetectMetadataParse`

inashivb:
Replaced w #6905

PR #6766, opened 2022-01-11
Pcap conditional v2.2.12

scottfgjordan:

Describe changes:
- Rebase
- Bugfix where pcap logs weren't being created in the default log mode (LOGMODE_COND_ALL)
- Rebased and updated Suricata Verify to use test.yaml cases. The Surivata-Verify PR doesn't include the fix described here: https://github.com/OISF/suricata-verify/pull/623, but I have tested it locally with those changes to confirm the test cases pass.

suricata-verify-pr: 633

### DIFF
--- a/src/log-pcap.c
+++ b/src/log-pcap.c
@@ -293,7 +293,7 @@ static int PcapLogCloseFile(ThreadVars *t, PcapLogData *pl)
         if (comp->format == PCAP_LOG_COMPRESSION_FORMAT_LZ4) {
             /* pcap_dump_close did not write any data because we call
              * pcap_dump_flush() after every write when writing
-	     * compressed output. */
+             * compressed output. */
             uint64_t bytes_written = LZ4F_compressEnd(comp->lz4f_context,
                     comp->buffer, comp->buffer_size, NULL);
             if (LZ4F_isError(bytes_written)) {
--- a/src/log-pcap.c
+++ b/src/log-pcap.c
@@ -26,7 +26,9 @@
  */
 
 #include "suricata-common.h"
+#include "util-buffer.h"
 #include "util-fmemopen.h"
+#include "stream-tcp-util.h"
 
 #ifdef HAVE_LIBLZ4
 #include <lz4frame.h>
@@ -189,6 +191,7 @@ typedef struct PcapLogData_ {
 
 typedef struct PcapLogThreadData_ {
     PcapLogData *pcap_log;
+    MemBuffer *buf;
 } PcapLogThreadData;
 
 /* Pattern for extracting timestamp from pcap log files. */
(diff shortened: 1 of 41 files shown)


victorjulien:
Scan-build flags 2 issues
```
  CC       stream-tcp-reassemble.o
stream-tcp-reassemble.c:283:21: warning: Result of 'calloc' is converted to a pointer of type 'uint8_t', which is incompatible with sizeof operand type 'char' [unix.MallocSizeof]
                    SCCalloc(1, sizeof(char) * TCPSEG_PKT_HDR_DEFAULT_SIZE);
                    ^~~~~~~~    ~~~~~~~~~~~~
./util-mem.h:36:18: note: expanded from macro 'SCCalloc'
#define SCCalloc calloc
                 ^~~~~~
1 warning generated.

  CC       log-pcap.o
log-pcap.c:721:9: warning: Null pointer passed to 1st parameter expecting 'nonnull' [core.NonNullParamChecker]
        SCMutexLock(&rp->tunnel_mutex);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./threads.h:121:26: note: expanded from macro 'SCMutexLock'
#define SCMutexLock(mut) pthread_mutex_lock(mut)
                         ^~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

jasonish:
Operational observation: It can take a long time for the pcap to flushed to disk in conditional mode (alerts), especially if the flow is short and the ruleset is highly tuned.  Is it possible to do a flush at the end of a flow?

src/log-pcap.c
```diff
@@ -1698,6 +1912,15 @@ static int PcapLogOpenFileCtx(PcapLogData *pl)
     return -1;
 }
 
+char *PcapLogGetFilename(void)
+{
+    /* do we have a file per thread */
+    if (pcap_file_multi != NULL) {
+        return pcap_file_multi;
+    }
+    return NULL;
```
jasonish, line 1921: Can we get this in non-multi mode as well?  With rotation, its still valuable to have no matter what, and consistent.

victorjulien:
> Operational observation: It can take a long time for the pcap to flushed to disk in conditional mode (alerts), especially if the flow is short and the ruleset is highly tuned. Is it possible to do a flush at the end of a flow?

Should this not be per alert? Seems odd to go from flow timeout code back into logging to flush a fp?

jasonish:
> > Operational observation: It can take a long time for the pcap to flushed to disk in conditional mode (alerts), especially if the flow is short and the ruleset is highly tuned. Is it possible to do a flush at the end of a flow?
> 
> Should this not be per alert? Seems odd to go from flow timeout code back into logging to flush a fp?

This would only flush what was written at the time of the alert, the flow may continue for a few packets, and its these packets that may be stuck in the file buffer until another alert triggers on this thread.  So it would be nice to have some flush that happened some time later.

As its a packet logger I think its being called into way more often than it may log, so perhaps just a last flushed variable, and then flush after some interval with no writes? Probably simpler than hooking into a flow logic.

PR #6770, opened 2022-01-11
Fix app-layer protocol detect patterns for SSLv2

giannitedesco:

The SSLv2 record format starts with 2 byte length, then it has a 1 byte
message-type field set to |01| for the client handshake, then a 2 byte
protocol version.

The SSLv3 record format (in use with modern TLS) starts with the
message-type of |16| and then the protocol, followed by the length.

This means that several of the patterns in the SSL app-layer detection
code are incorrect because they're looking in the wrong place in the
packet.

suricata-verify-pr: 555
OISF/suricata-verify#555
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -2830,9 +2830,8 @@ static int SSLRegisterPatternsForProtocolDetection(void)
     }
 
     /** SSLv3 */
-    if (AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP, ALPROTO_TLS,
-                                               "|01 03 00|", 3, 0, STREAM_TOSERVER) < 0)
-    {
+    if (AppLayerProtoDetectPMRegisterPatternCS(
+                IPPROTO_TCP, ALPROTO_TLS, "|01 03 00|", 5, 2, STREAM_TOSERVER) < 0) {
         return -1;
     }
     if (AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP, ALPROTO_TLS,
@@ -2842,9 +2841,8 @@ static int SSLRegisterPatternsForProtocolDetection(void)
     }
 
     /** TLSv1 */
-    if (AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP, ALPROTO_TLS,
-                                               "|01 03 01|", 3, 0, STREAM_TOSERVER) < 0)
-    {
+    if (AppLayerProtoDetectPMRegisterPatternCS(
+                IPPROTO_TCP, ALPROTO_TLS, "|01 03 01|", 5, 2, STREAM_TOSERVER) < 0) {
         return -1;
     }
     if (AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP, ALPROTO_TLS,
@@ -2854,9 +2852,8 @@ static int SSLRegisterPatternsForProtocolDetection(void)
     }
 
     /** TLSv1.1 */
-    if (AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP, ALPROTO_TLS,
-                                               "|01 03 02|", 3, 0, STREAM_TOSERVER) < 0)
-    {
+    if (AppLayerProtoDetectPMRegisterPatternCS(
+                IPPROTO_TCP, ALPROTO_TLS, "|01 03 02|", 5, 2, STREAM_TOSERVER) < 0) {
         return -1;
     }
     if (AppLayerProtoDetectPMRegisterPatternCS(IPPROTO_TCP, ALPROTO_TLS,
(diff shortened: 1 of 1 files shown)


giannitedesco:
@catenacyber created new repo, actually I got my remotes mixed up so I was rebasing on origin/master not upstream/master, woops :)

catenacyber:
Thanks CI looks better.
The failures are not related to the PR

catenacyber (approved):

victorjulien:
I need to look at this again. I looked earlier w/o understanding the logic and not seeing any test cases pass/fail in relation to this PR.

SSL3/TLS records can certainly start with other values than `|16|`.

victorjulien:
@catenacyber can you dig a bit deeper into this?

catenacyber:
> not seeing any test cases pass/fail in relation to this PR.

The S-V tests fail without this PR and succeed with it...

For instance, sslv2-tls1_0-client-handshake.pcap tcp payload begins with bytes 80 80 01 03 01. First 2 bytes are SSLv2, Third byte is 1:client hello, Byte 4 and 5 are request version 03.01 being TLS 1.0

catenacyber (changes requested):
@giannitedesco can you rebase both this PR and the S-V one ?
It looks like sslv2-handshake is now failing with this PR

Flow has 
```
  "app_proto": "tls",
  "app_proto_tc": "failed",
```
as in master

giannitedesco:
> @giannitedesco can you rebase both this PR and the S-V one ? It looks like sslv2-handshake is now failing with this PR
> 
> Flow has
> 
> ```
>   "app_proto": "tls",
>   "app_proto_tc": "failed",
> ```
> 
> as in master

Those are supposed to fail, there is no SSLv2 decoder. The tests are included so that they will verify the SSLv2 decoder when it is added.

Is there a way to mark tests as "ignore failures" or should I redo them to just verify the protocol detection works?

giannitedesco:
@catenacyber So I just commented out the SSLv2 protocol decode stuff, so they can be uncommented if/when that gets implemented. At least the test can verify the detection of protocol part then.

catenacyber:
> So I just commented out the SSLv2 protocol decode stuff, so they can be uncommented if/when that gets implemented. At least the test can verify the detection of protocol part then.

@inashivb is there a `skip` keyword for Suricata-verify ?

catenacyber:
> Those are supposed to fail, there is no SSLv2 decoder. The tests are included so that they will verify the SSLv2 decoder when it is added.

The commit message in this PR seems to say otherwise.
Is it fixing SSLv2 or SSLv3 detect patterns ?

giannitedesco:
> The commit message in this PR seems to say otherwise. Is it fixing SSLv2 or SSLv3 detect patterns ?

This is fixing SSLv2 detect patterns.

There are two distinct things, whether we detect an off-port flow as SSLv2 (which current code does not, and this patch fixes). But then there is decoding the SSLv2 frames, after you have identified SSL/TLS and associated he decoder to that flow.

There are actually numerous things that the SSL/TLS decoder does not support and one of them is SSLv2, which has an incompatible frame format and totally different handshake message structure. SSLv3 and TLS use the same frame formats and are broadly compatible with each other.

giannitedesco:
Anyway, adding support for decoding SSLv2 (and fragmented handshakes for that matter) is another question, and they are much bigger jobs.

catenacyber:
I see that 01 03 01 is indeed used for sslv2-tls1_0-client-handshake.pcap

But I do not see which pattern is used for sslv2-handshake.pcap
Is something missing ?

> Anyway, adding support for decoding SSLv2 (and fragmented handshakes for that matter) is another question, and they are much bigger jobs.

Ok, understood

catenacyber:
> > Flow has
> > ```
> >   "app_proto": "tls",
> >   "app_proto_tc": "failed",
> > ```
> > 

If we want to fix SSLv2 detect patterns, we should also detect the patterns to client (from server)
Here, there are only to server (from client), right ?

inashivb:
> 

There is one but seems like it is
1. bound to features (See example text.yaml in https://github.com/oiSF/suricata-verify)
 or 
 2. skipping an entire test (e.g. https://github.com/OISF/suricata-verify/blob/master/tests/dhcp-request-flood/test.yaml#L3)
 
 I tried it on specific filters and it does not seem to work that way as of now.

giannitedesco:
> > > Flow has
> > > ```
> > >   "app_proto": "tls",
> > >   "app_proto_tc": "failed",
> > > ```
> 
> If we want to fix SSLv2 detect patterns, we should also detect the patterns to client (from server) Here, there are only to server (from client), right ?

That's right, I can investigate adding the reverse direction filter, too, thanks.

giannitedesco:
> > @inashivb is there a `skip` keyword for Suricata-verify ?
> 
> There is one but seems like it is
> 
>     1. bound to features (e.g. https://github.com/OISF/suricata-verify/blob/master/tests/filestore-v2.3-fserror/test.yaml#L7)
>        or
> 
>     2. skipping an entire test (e.g. https://github.com/OISF/suricata-verify/blob/master/tests/dhcp-request-flood/test.yaml#L3)
> 
> 
> I tried it on specific filters and it does not seem to work that way as of now.

Aha, thanks, okay, for now I will forget about decoder events in the test cases and we can add tests for the decoder if and when the decoder is added in future.

catenacyber:
Thanks @giannitedesco I am closing this, waiting for a new version then

PR #6771, opened 2022-01-12
Dev/nom7 upgrade http2 v0.1

chifflier:

https://redmine.openinfosecfoundation.org/issues/4165

Describe changes:
- Rewrite HTTP/2 parsers to nom 7
- This PR aims for minimal changes, and only the HTTP/2 parser. That said, I think several functions (`detect_parse_*` for ex) could be factorized
- unit tests for cargo and suricata-verify all pass
- As for other PRs, no rustfmt, only affected files are changed

### DIFF
--- a/rust/src/http2/http2.rs
+++ b/rust/src/http2/http2.rs
@@ -24,7 +24,7 @@ use crate::applayer::{self, *};
 use crate::core::*;
 use crate::filecontainer::*;
 use crate::filetracker::*;
-use nom;
+use nom7::Err;
 use std;
 use std::ffi::CString;
 use std::fmt;
@@ -647,7 +647,7 @@ impl HTTP2State {
                         //we could set an event on remaining data
                         return HTTP2FrameTypeData::SETTINGS(set);
                     }
-                    Err(nom::Err::Incomplete(_)) => {
+                    Err(Err::Incomplete(_)) => {
                         if complete {
                             self.set_event(HTTP2Event::InvalidFrameData);
                             return HTTP2FrameTypeData::UNHANDLED(HTTP2FrameUnhandled {
@@ -738,7 +738,7 @@ impl HTTP2State {
                         self.process_headers(&hs.blocks, dir);
                         return HTTP2FrameTypeData::PUSHPROMISE(hs);
                     }
-                    Err(nom::Err::Incomplete(_)) => {
+                    Err(Err::Incomplete(_)) => {
                         if complete {
                             self.set_event(HTTP2Event::InvalidFrameData);
                             return HTTP2FrameTypeData::UNHANDLED(HTTP2FrameUnhandled {
@@ -772,7 +772,7 @@ impl HTTP2State {
                         self.process_headers(&hs.blocks, dir);
                         return HTTP2FrameTypeData::CONTINUATION(hs);
                     }
-                    Err(nom::Err::Incomplete(_)) => {
+                    Err(Err::Incomplete(_)) => {
                         if complete {
                             self.set_event(HTTP2Event::InvalidFrameData);
                             return HTTP2FrameTypeData::UNHANDLED(HTTP2FrameUnhandled {
@@ -807,7 +807,7 @@ impl HTTP2State {
                         }
                         return HTTP2FrameTypeData::HEADERS(hs);
                     }
-                    Err(nom::Err::Incomplete(_)) => {
+                    Err(Err::Incomplete(_)) => {
                         if complete {
                             self.set_event(HTTP2Event::InvalidFrameData);
                             return HTTP2FrameTypeData::UNHANDLED(HTTP2FrameUnhandled {
@@ -929,7 +929,7 @@ impl HTTP2State {
                     }
                     input = &rem[hlsafe..];
                 }
-                Err(nom::Err::Incomplete(_)) => {
+                Err(Err::Incomplete(_)) => {
                     //we may have consumed data from previous records
                     return AppLayerResult::incomplete(
                         (il - input.len()) as u32,
@@ -1035,7 +1035,7 @@ pub unsafe extern "C" fn rs_http2_probing_parser_tc(
                 }
                 return ALPROTO_HTTP2;
             }
-            Err(nom::Err::Incomplete(_)) => {
+            Err(Err::Incomplete(_)) => {
                 return ALPROTO_UNKNOWN;
             }
             Err(_) => {
--- a/rust/src/http2/huffman.rs
+++ b/rust/src/http2/huffman.rs
@@ -15,9 +15,11 @@
  * 02110-1301, USA.
  */
 
-use nom::error::ErrorKind;
-use nom::Err;
-use nom::IResult;
+use nom7::bits::streaming::take as take_bits;
+use nom7::branch::alt;
+use nom7::combinator::{complete, map_opt};
+use nom7::error::{make_error, ErrorKind};
+use nom7::{Err, IResult};
 
 fn http2_huffman_table_len5(n: u32) -> Option<u8> {
     match n {
@@ -35,9 +37,9 @@ fn http2_huffman_table_len5(n: u32) -> Option<u8> {
     }
 }
 
-named!(http2_decode_huffman_len5<(&[u8], usize), u8>,
-    complete!( map_opt!(take_bits!(5u32), http2_huffman_table_len5) )
-);
+fn http2_decode_huffman_len5(input: (&[u8], usize)) -> IResult<(&[u8], usize), u8> {
+    complete(map_opt(take_bits(5u32), http2_huffman_table_len5))(input)
+}
 
 fn http2_huffman_table_len6(n: u32) -> Option<u8> {
     match n {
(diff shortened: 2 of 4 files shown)


catenacyber:
Thanks Pierre.

> That said, I think several functions (detect_parse_* for ex) could be factorized

I understand tou are talking about the same thing as in https://redmine.openinfosecfoundation.org/issues/4112 
Am I correct ?

> , no rustfmt, only affected files are changed

All rust HTTP2 files are rustfmted, so please keep them rustfmted ;-)

Why did you call this PR "v0.1". ?
Are you planning more work on this ?

rust/src/http2/range.rs
```diff
@@ -38,52 +40,45 @@ pub struct HTTPContentRange {
 }
 
 pub fn http2_parse_content_range_star<'a>(input: &'a [u8]) -> IResult<&'a [u8], HTTPContentRange> {
-    let (i2, _) = char!(input, '*')?;
-    let (i2, _) = char!(i2, '/')?;
-    let (i2, size) = map_res!(i2, map_res!(digit1, std::str::from_utf8), i64::from_str)?;
+    let (i2, _) = char('*')(input)?;
```
catenacyber, line 43: Looks strange to me to use `char` as a function.
Is there no risk of future collision ?

rust/src/http2/range.rs
```diff
@@ -38,52 +40,45 @@ pub struct HTTPContentRange {
...
+        http2_parse_content_range_star,
+        http2_parse_content_range_def,
+    ))(i2);
 }
 
 pub fn http2_parse_check_content_range<'a>(input: &'a [u8]) -> IResult<&'a [u8], HTTPContentRange> {
     let (rem, v) = http2_parse_content_range(input)?;
     if v.start > v.end {
-        return Err(Err::Error((rem, ErrorKind::Verify)));
+        return Err(Err::Error(make_error(rem, ErrorKind::Verify)));
```
catenacyber, line 81: Do we need the 3 `Err` `Err::Error` and `make_error` ?

rust/src/http2/http2.rs
```diff
@@ -647,7 +647,7 @@ impl HTTP2State {
                         //we could set an event on remaining data
                         return HTTP2FrameTypeData::SETTINGS(set);
                     }
-                    Err(nom::Err::Incomplete(_)) => {
+                    Err(Err::Incomplete(_)) => {
```
catenacyber, line 650: Why not keep `nom7::Err::Incomplete(_)` ?

rust/src/http2/huffman.rs
```diff
@@ -482,23 +484,44 @@ fn http2_huffman_table_len30(n: u32) -> Option<u8> {
...
+        http2_decode_huffman_len20,
+        http2_decode_huffman_len21,
+        http2_decode_huffman_len21,
+        http2_decode_huffman_len22,
+        http2_decode_huffman_len23,
+        http2_decode_huffman_len24,
+        http2_decode_huffman_len25,
+        http2_decode_huffman_len26,
+        http2_decode_huffman_len27,
+        alt((
```
catenacyber, line 521: Why another alt ?

rust/src/http2/parser.rs
```diff
@@ -344,39 +357,38 @@ fn http2_parse_headers_block_indexed<'a>(
...
+        return Err(Err::Error(make_error(i2, ErrorKind::LengthValue)));
     }
-    let (i3, data) = take!(i2, stringlen as usize)?;
+    let (i3, data) = take(stringlen as usize)(i2)?;
     if huffslen.0 == 0 {
         return Ok((i3, data.to_vec()));
     } else {
-        let (_, val) = bits!(data, many0!(huffman::http2_decode_huffman))?;
+        let (_, val) =
+            bits::<_, _, Error<(&[u8], usize)>, _, _>(many0(huffman::http2_decode_huffman))(data)?;
```
catenacyber, line 391: this annotation is not so nice.
How did you find it ?

catenacyber:
Thanks for this work.
Looks good to upgrade nom

chifflier:
> Thanks Pierre.
> 
> > That said, I think several functions (detect_parse_* for ex) could be factorized
> 
> I understand tou are talking about the same thing as in https://redmine.openinfosecfoundation.org/issues/4112 Am I correct ?

It could be a solution, yes. Since most of this functions differ only by a tag and the return value, these could also be passed as arguments of a function.

> 
> > , no rustfmt, only affected files are changed
> 
> All rust HTTP2 files are rustfmted, so please keep them rustfmted ;-)
> 

I did :) My comment was not really relevant for http2, I took it from my other nom7 PRs

> Why did you call this PR "v0.1". ? Are you planning more work on this ?

No particular intention, I always use numbers in case changes are requested and I submit a new version.

rust/src/http2/range.rs
chifflier: You mean because `char` is a rust keyword? I don't think there will be problems, it is part of the language. You can even use reserved keywords using the raw identifiers notation (see https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html).

Note that the previous code was also using keywords as variable names in other places (see `u64` in `detect_parse_u64`)

rust/src/http2/range.rs
chifflier: That part of nom errors also irritates me ...
- The first `Err` is the `Result` variant, so required
- `Err::Error` is the nom error variant (could be `Error`, or `Failure` to stop without possible recovery), it is needed
- `make_error` is a convenience function to create a generic nom error. This is not required but only useful for genericity and to ease future code updates. It can be removed it to return only the tuple.

rust/src/http2/http2.rs
chifflier: To simplify later updates by regrouping all namespaces at top of file, so if we change back the name from `nom7` to `nom` this will be much easier.
Note: the `nom7` name is only a trick used to simplify the transition

rust/src/http2/huffman.rs
chifflier: This is detailed in the comment above the big `alt` statement: nom has implemented the `Alt` trait (used internally by the `alt` function) only for lists up to size 20, and this list has 21 elements :)
Using nested `alt` is the proposed workaround from the nom documentation (https://docs.rs/nom/7.1.0/nom/branch/fn.alt.html):
```
It takes as argument a tuple of parsers. There is a maximum of 21 parsers. If you need more, it is possible to nest them in other alt calls, like this: alt(parser_a, alt(parser_b, parser_c))

```

https://github.com/chifflier/suricata/blob/dev/nom7-upgrade-http2-v0.1/rust/src/http2/huffman.rs#L497-L499

rust/src/http2/parser.rs
chifflier: Agreed, it is neither nice nor readable.

The rust compiler is lost in generic types in the `bits` combinator, and fails to infer the error type correctly. Here, the solution is to explicit the error type while letting the compiler infer other generic parameters (`_`).
This problem did not exist with macro (but they had other problems), and seem to happen only with the `bits` combinator. The only way I found to avoid this is to create a function containing the subparser, so the error type is given in the function declaration and the compiler infers the rest - but this is not great, because it prevents inlining the parser ...

The annotation is similar to the one used in nom's documentation: https://docs.rs/nom/7.1.0/nom/bits/fn.bits.html

rust/src/http2/range.rs
catenacyber: ok fair enough

> see u64 in detect_parse_u64

I feel the shame

rust/src/http2/range.rs
catenacyber: Well, I let you make the PR to nom to get something better ;-)

Should we macro it meanwhile ?

rust/src/http2/http2.rs
catenacyber: I was rather afraid of collisions about `Err` from something else... What do you think ?

rust/src/http2/huffman.rs
catenacyber: Thanks Pierre for the comment I had not seen 🙄

rust/src/http2/parser.rs
catenacyber: Maybe this deserves a comment...

catenacyber:
Looks good to me, just waiting for your last comments on my questions

rust/src/http2/http2.rs
chifflier: In my experience, rust handles this quite well: it resolves the type `Err::Incomplete` as a variant from the nom type, without conflict with the builtin variant `Result::Err`, so the notation will not create any conflict.
The only possible error would be to explicitly import several types `Err` in the `use` statements, but the compiler detects that and will require to change one of the names.

rust/src/http2/range.rs
chifflier: Indeed, we can shorten the notation using a macro.
Do you want me to push commits for this?

catenacyber: If you feel it is idiomatic

catenacyber (approved):

rust/src/http2/parser.rs
jasonish: What prevents the inlining? Its used enough that a helper function could clean it up a little.

rust/src/http2/range.rs
jasonish: I've got used to this pattern the more I've worked with nom7.  I'm not sure a macro would help much.

jasonish (approved):
Looks good for a direct port to nom 7.

victorjulien:
Merged in #6802, thanks!

rust/src/http2/range.rs
chifflier: It's up to you! Personally, I am now used to this syntax and agree with Jason, so I do not think a macro would really help. In fact, I find it confusing since it tends to hide types and makes resolving compiler errors a bit more difficult.

rust/src/http2/parser.rs
chifflier: Nothing prevents inlining (and we can use the `#[inline]` attribute to force it). The size of the bits seems to vary, so it is possible to write a function that would take some parameters as arguments and return a combinator.
If that's ok for you, I will push the patch to the end of this PR.

PR #6620, opened 2021-11-19
detect: file_data keyword works on nfs protocol

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4839

Describe changes:
- Allows nfs protocol on file_data keyword

PR #6777, opened 2022-01-13
Email log backport 6.0.5/v2

inashivb:
https://redmine.openinfosecfoundation.org/issues/4931
https://redmine.openinfosecfoundation.org/issues/4894

suricata-verify-pr: 632

### DIFF
--- a/src/detect-file-data.c
+++ b/src/detect-file-data.c
@@ -113,6 +113,10 @@ void DetectFiledataRegister(void)
             ALPROTO_HTTP2, HTTP2StateDataServer);
     DetectAppLayerInspectEngineRegister2("file_data", ALPROTO_HTTP, SIG_FLAG_TOCLIENT,
             HTP_RESPONSE_BODY, DetectEngineInspectBufferHttpBody, NULL);
+    DetectAppLayerMpmRegister2(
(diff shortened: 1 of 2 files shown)


victorjulien:
The nfs file_data commit is a feature that should not be backported.

inashivb:
> The nfs file_data commit is a feature that should not be backported.

hmm I'm just noticing you conversation w Philippe about it on the issue. @catenacyber is https://github.com/OISF/suricata/pull/6620/files fixing the leak in master somehow? Could you please tell what's the right way to approach the issue?

catenacyber:
> is https://github.com/OISF/suricata/pull/6620/files fixing the leak in master somehow?

Yes, it does the fix.
The bug is due to having `file.data` and later a `nfs` keyword in a signature.
The fix in master is to allow such a signature.

For the backport, I think Victor wants to error on such a signature to avoid the leak...
And yes, I am not sure what is the right way to achieve this...

jlucovsky:
> > is https://github.com/OISF/suricata/pull/6620/files fixing the leak in master somehow?
> 
> Yes, it does the fix. The bug is due to having `file.data` and later a `nfs` keyword in a signature. The fix in master is to allow such a signature.
> 
> For the backport, I think Victor wants to error on such a signature to avoid the leak... And yes, I am not sure what is the right way to achieve this...

I've been looking at the same issue with backports to 5.0.x.

Perhaps we could flag the issue when parsing the rule and invalidate it?

PR #6773, opened 2022-01-12
App maxtx 4530 v7

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4530

Describe changes:
- limits the number of active transactions per flow for HTTP2 and MQTT
- fixes MQTT transaction completion

This is to avoid DOS by quadratic complexity for protocols looping over the whole list of transactions looking for a
specific transation with an identifier, so that a new PDU gets processed within that transaction.

There may be other protocols with this problem (like Modbus)

Replaces #6751 with 
- fixing nits from code review
- making the maxtx configurable for MQTT

There is still the documentation about the Suricata.yaml added settings to do

PR #6783, opened 2022-01-13
App maxtx 4530 v8

catenacyber:

Describe changes:

Replaces #6773 with 
- implementing the maxtx for FTP
- user guide documentation for added suricata.yaml fields


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6788

PR #6791, opened 2022-01-14
util: fix int warnings

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4516

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` in all util files

PR #6792, opened 2022-01-14
Warnint util 4516 v2

catenacyber:

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` in all util files

Modifies #6791 
- also removes warnings from unit tests
- fixes radix tree loop (using uint, we can no longer have condition >= 0)

### DIFF
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -448,7 +448,7 @@ static MimeDecField * MimeDecFillField(MimeDecEntity *entity, uint8_t *name,
         /* convert to lowercase and store */
         uint32_t u;
         for (u = 0; u < nlen; u++)
-            name[u] = tolower(name[u]);
+            name[u] = u8_tolower(name[u]);
 
         field->name = (uint8_t *)name;
         field->name_len = nlen;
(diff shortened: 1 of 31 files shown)


src/util-decode-mime.c
```diff
@@ -687,7 +687,11 @@ static uint8_t * GetFullValue(DataValue *dv, uint32_t *len)
 static inline uint8_t * FindBuffer(const uint8_t *src, uint32_t len, const uint8_t *find, uint32_t find_len)
 {
     /* Use utility search function */
-    return BasicSearchNocase(src, len, find, find_len);
+    if (find_len > UINT16_MAX) {
+        // TODO rewrite all this to have app-layer events
+        find_len = UINT16_MAX;
+    }
+    return BasicSearchNocase(src, len, find, (uint16_t)find_len);
```
victorjulien, line 694: I wonder if we should just update `BasicSearchNocase` to accept `uint32_t`?

catenacyber: `len` is `uint32_t`(haystack)
The needle is `uint16_t`
That looks reasonable...

suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.memuse | 524447808 | 553066368 | 94.83% | 

Pipeline 5699

src/util-decode-mime.c
victorjulien: ok, makes sense. Can you add a ticket for tracking the events setting?

victorjulien:
replaced by #6820

src/util-decode-mime.c
catenacyber: Or should this mime code get rustified first ?

victorjulien: For master that would be nice. Would we need a backport for this int handling though?

catenacyber: I do not know.
It will be a big work I guess...
I am not sure it is worth it...

PR #6794, opened 2022-01-14
Quic/v10

victorjulien:
#6461 rebased and adjusted to recent api changes.

Changed `quic.version` to be a sticky buffer.

suricata-verify-pr: 647

TODO this essentially creates a tx per packet, even if most packets are uninteresting. Should probably not do that.

PR #6796, opened 2022-01-16
Quic/v11

victorjulien:
#6794 with
- formatting fixups
- sni logging and new `quic.sni` keyword
- reduced tx creation/logging: only "long header" frames currently lead to a tx

suricata-verify-pr: 653

cc: @sharksforarms


jasonish:
Only comments around logging are the CYU.  Does it need to be a list? Would be nice to show this in an S-V test.
And should the CYU be optional much like the ja3 is for TLS?

victorjulien:
> Only comments around logging are the CYU. Does it need to be a list? Would be nice to show this in an S-V test. And should the CYU be optional much like the ja3 is for TLS?

It's stored as a list as well, and inspected as such, but I'm not sure really. The pcaps we've seen don't have multiple. @sharksforarms ?

sharksforarms:
> > Only comments around logging are the CYU. Does it need to be a list? Would be nice to show this in an S-V test. And should the CYU be optional much like the ja3 is for TLS?
> 
> It's stored as a list as well, and inspected as such, but I'm not sure really. The pcaps we've seen don't have multiple. @sharksforarms ?

Previous discussion/change here:
https://github.com/OISF/suricata/pull/5574#discussion_r527012309

I don't have any pcaps to backup this theory.

victorjulien:
replaced by #6813

PR #6800, opened 2022-01-17
App layer error close txs 4318 v2

catenacyber:

Describe changes:

The best way I see to test this is SMB
But I am not sure how to get multiple live transactions over SMB

Then to get an error, simply have a NBSS message with length 4, and just `\xFFSMB` so the SMB parser will fail decode the SMB1 header (not sure it should put the whole parser in an error state by the way, as we could go on with the next NBSS message)

Modifies #6472 by checking `STREAMTCP_FLAG_APP_LAYER_DISABLED` against `ssn->flags` instead of `stream->flags`
Is there a way (other than rust) to prevent this kind of mistake ?

### DIFF
--- a/src/flow.c
+++ b/src/flow.c
@@ -1131,7 +1131,8 @@ uint8_t FlowGetDisruptionFlags(const Flow *f, uint8_t flags)
     TcpSession *ssn = f->protoctx;
     TcpStream *stream = flags & STREAM_TOSERVER ? &ssn->client : &ssn->server;
 
-    if (stream->flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) {
+    if ((stream->flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) ||
+            (ssn->flags & STREAMTCP_FLAG_APP_LAYER_DISABLED)) {
         newflags |= STREAM_DEPTH;
     }
     /* todo: handle pass case (also for UDP!) */


suricata-qa:
Information:
 

ERROR: QA failed on tlpw1_files_md5.
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.memuse | 524834368 | 553066368 | 94.9% | 

Pipeline 5748

victorjulien:
> Modifies #6472 by checking STREAMTCP_FLAG_APP_LAYER_DISABLED against ssn->flags instead of stream->flags
Is there a way (other than rust) to prevent this kind of mistake ?

There is a cocci check for flags, but I'm not sure its working currently.

catenacyber:
> There is a cocci check for flags, but I'm not sure its working currently.

Do you mean qa/coccinelle/struct-flags.py ?

I do not get what ie does at first sight...

src/flow.c
```diff
@@ -1131,7 +1131,8 @@ uint8_t FlowGetDisruptionFlags(const Flow *f, uint8_t flags)
     TcpSession *ssn = f->protoctx;
     TcpStream *stream = flags & STREAM_TOSERVER ? &ssn->client : &ssn->server;
 
-    if (stream->flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) {
+    if ((stream->flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) ||
+            (ssn->flags & STREAMTCP_FLAG_APP_LAYER_DISABLED)) {
```
victorjulien, line 1135: I think this makes sense, but I feel like the `STREAM_DEPTH` flags name is now getting confusing. Since we're using it as a "stream is interupted/disrupted/truncated" flag, I think we should also rename it. How about `STREAM_DISRUPTED`?

victorjulien (changes requested):
see inline comments

src/flow.c
catenacyber: ok doing

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6887

victorjulien:
> > There is a cocci check for flags, but I'm not sure its working currently.
> 
> Do you mean qa/coccinelle/struct-flags.py ?
> 
> I do not get what ie does at first sight...

Here the script checks that flags set (and checked?) on the `flags` field all start with `STREAMTCP_INIT_`

```c
typedef struct TcpStreamCnf_ {
    ...
    /* coccinelle: TcpStreamCnf:flags:STREAMTCP_INIT_ */
    uint8_t flags;
```
The .py script actually generates the cocci script.

catenacyber:
Oh, so coccinnelle should have caught this but it did not work :-/

PR #6807, opened 2022-01-18
Warnint util 4516 v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4516

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` in all files except d*

Extends #6792 : now only d* files remain with warnings...

Thoughts about the commit history ?

PR #6820, opened 2022-01-19
Warnint util 4516 v4

catenacyber:

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` in all files

Extends #6792 : complete with CI check...

Replaces #6807 with completion for all files

Thoughts about the commit history ?
This PR can be split into smaller PRs as some points need a careful review (detect-content.c for instance)

### DIFF
--- a/src/util-fix_checksum.c
+++ b/src/util-fix_checksum.c
@@ -54,7 +54,6 @@ FixChecksum(uint16_t sum, uint16_t old, uint16_t new)
 
     l = sum + old - new;
     l = (l >> 16) + (l & 65535);
-    l = l & 65535;
 
-    return l;
+    return (uint16_t)l;
 }
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -851,7 +851,7 @@ jobs:
           chmod 755 $HOME/.cargo/bin/cbindgen
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - run: ./autogen.sh
-      - run: LIB_FUZZING_ENGINE="fail_to_onefile_driver" CC=clang-10 CXX=clang++-10 CFLAGS="-fprofile-arcs -ftest-coverage -g -fno-strict-aliasing -fsanitize=address -fno-omit-frame-pointer -fPIC -Wno-unused-parameter -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION=1" CXXFLAGS="-fprofile-arcs -ftest-coverage -g -fno-strict-aliasing -fsanitize=address -fno-omit-frame-pointer -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION=1 -stdlib=libc++" ac_cv_func_malloc_0_nonnull=yes ac_cv_func_realloc_0_nonnull=yes ./configure --with-gnu-ld --enable-fuzztargets --disable-shared --enable-gccprotect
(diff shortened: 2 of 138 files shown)


suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.memuse | 534901888 | 505591168 | 105.8% | 

Pipeline 5805

src/util-fix_checksum.c
```diff
@@ -54,7 +54,6 @@ FixChecksum(uint16_t sum, uint16_t old, uint16_t new)
 
     l = sum + old - new;
     l = (l >> 16) + (l & 65535);
-    l = l & 65535;
 
-    return l;
+    return (uint16_t)l;
```
victorjulien, line 58: @jasonish any thoughts here? IIRC this func was introduced by you?

catenacyber: This is an easy one, without knowing the function's purpose
`l = l & 65535; return l;` does the same thing as  `return (uint16_t)l;` (right ?)

victorjulien: I guess I'm not actually sure what C is doing (and/or expected to do here), esp since this code IIRC comes from OpenBSD. They seem to know a thing or 2 about this stuff.

jasonish: > `l = l & 65535; return l;` does the same thing as `return (uint16_t)l;` (right ?)

A quick test shows that they do the same thing, but its not nearly as clear to me as the original form.

catenacyber: Do you prefer `l = l & 65535; (uint16_t)return l;` ? (even if it is redundant)

catenacyber:
Any thoughts on the CIFuzz find ?
Should we apply this patch ?
```
diff --git a/src/defrag.h b/src/defrag.h
index 771616e4d..9c8e6f1d8 100644
--- a/src/defrag.h
+++ b/src/defrag.h
@@ -45,7 +45,7 @@ typedef struct Frag_ {
     uint16_t offset;            /**< The offset of this fragment, already
                                  *   multiplied by 8. */
 
-    uint16_t len;               /**< The length of this fragment. */
+    uint32_t len;               /**< The length of this fragment. */
```

src/util-fix_checksum.c
jasonish: I think so, at least its clear.. Would this work as well?
```
return (uint16_t)(l & 65535);
```

catenacyber: ok for me

victorjulien:
> Any thoughts on the CIFuzz find ? Should we apply this patch ?
> 
> ```
> diff --git a/src/defrag.h b/src/defrag.h
> index 771616e4d..9c8e6f1d8 100644
> --- a/src/defrag.h
> +++ b/src/defrag.h
> @@ -45,7 +45,7 @@ typedef struct Frag_ {
>      uint16_t offset;            /**< The offset of this fragment, already
>                                   *   multiplied by 8. */
>  
> -    uint16_t len;               /**< The length of this fragment. */
> +    uint32_t len;               /**< The length of this fragment. */
> ```

Can you do a PR with this so we can review it properly? I think adding new patch suggestions in PR comments isn't very effective for review/discussion.

catenacyber:
> 
> Can you do a PR with this so we can review it properly? I think adding new patch suggestions in PR comments isn't very effective for review/discussion.

Done in https://github.com/OISF/suricata/pull/7006

But I was rather under the impression that we should split this big PR into smaller PRs...

victorjulien:
> > Can you do a PR with this so we can review it properly? I think adding new patch suggestions in PR comments isn't very effective for review/discussion.
> 
> Done in #7006
> 
> But I was rather under the impression that we should split this big PR into smaller PRs...

Sure you can split PRs or do PRs for smaller changes, right? Suggesting code changes in comments instead of a PR just seems to be an anti-pattern when the PR is to review code changes in git.

PR #6835, opened 2022-01-21
ci: adds scorecard analysis GitHub workflow

catenacyber:
None

Describe changes:
- ci: adds scorecard analysis GitHub workflow

cf https://github.blog/2022-01-19-reducing-security-risk-oss-actions-opensff-scorecards-v4/

### DIFF
--- a/.github/workflows/scorecards-analysis.yml
+++ b/.github/workflows/scorecards-analysis.yml
@@ -1,0 +1,55 @@
+name: Scorecards supply-chain security
+on:
+  # Only the default branch is supported.
+  branch_protection_rule:
+  schedule:
+    - cron: '28 22 * * 6'
+  push:
+    branches: [ master ]
+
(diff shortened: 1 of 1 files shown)


catenacyber:
For this to work, it requires a Personal Access Token (PAT) for authentication and save the token value as a repository secret...

victorjulien:
> For this to work, it requires a Personal Access Token (PAT) for authentication and save the token value as a repository secret...

Any idea how the PAT should be generated given that this is an "organization" account? I can create one from my account of course, but won't rather do it correctly from the start.

victorjulien:
> For this to work, it requires a Personal Access Token (PAT) for authentication and save the token value as a repository secret...

Any idea how the PAT should be generated given that this is an "organization" account? I can create one from my account of course, but won't rather do it correctly from the start.

PR #6840, opened 2022-01-21
pgsql: fix defect found by coverity

jufajardini:
Pgsql was using bitwise operations to assign password output config to
its context flags, but mixing that with logic negation of the default
value, resulting in the expressions having a constant value as result.

Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5007

Describe changes:
- use bitwise operator, instead of logical
- also change the default value in the code to be 0, to better reflect that the actual default is `disabled`

suricata-verify-pr: 656
https://github.com/OISF/suricata-verify/pull/656

### DIFF
--- a/src/output-json-pgsql.c
+++ b/src/output-json-pgsql.c
@@ -47,7 +47,7 @@
 #include "output-json-pgsql.h"
 #include "rust.h"
 
-#define PGSQL_LOG_PASSWORDS BIT_U32(1)
+#define PGSQL_LOG_PASSWORDS BIT_U32(0)
 
 typedef struct OutputPgsqlCtx_ {
     uint32_t flags;
@@ -103,10 +103,10 @@ static void JsonPgsqlLogParseConfig(ConfNode *conf, OutputPgsqlCtx *pgsqllog_ctx
         if (ConfValIsTrue(query)) {
             pgsqllog_ctx->flags |= PGSQL_LOG_PASSWORDS;
         } else {
-            pgsqllog_ctx->flags &= !PGSQL_LOG_PASSWORDS;
+            pgsqllog_ctx->flags &= ~PGSQL_LOG_PASSWORDS;
         }
     } else {
-        pgsqllog_ctx->flags &= !PGSQL_LOG_PASSWORDS;
+        pgsqllog_ctx->flags &= ~PGSQL_LOG_PASSWORDS;
     }
 }
 
--- a/rust/src/pgsql/logger.rs
+++ b/rust/src/pgsql/logger.rs
@@ -24,7 +24,7 @@ use crate::pgsql::parser::*;
 use crate::pgsql::pgsql::*;
 use std;
 
-pub const PGSQL_LOG_PASSWORDS: u32 = BIT_U32!(1);
+pub const PGSQL_LOG_PASSWORDS: u32 = BIT_U32!(0);
 
 fn log_pgsql(tx: &PgsqlTransaction, flags: u32, js: &mut JsonBuilder) -> Result<(), JsonError> {
     js.set_uint("tx_id", tx.tx_id)?;


src/output-json-pgsql.c
```diff
@@ -47,7 +47,7 @@
 #include "output-json-pgsql.h"
 #include "rust.h"
 
-#define PGSQL_LOG_PASSWORDS BIT_U32(1)
+#define PGSQL_LOG_PASSWORDS BIT_U32(0)
```
catenacyber, line 50: I think we can get rid of this line as cbindgen now export constants

jufajardini: Not managing to make it work, failed at finding similar code in suri that could guide me...

jufajardini: Could it it that it's not working because it uses the macros on the rust side, when declaring the constants?

jufajardini: (if I expand the value instead of using the macros, I manage to make the code compile. Otherwise, I didn't find a way to make it work...)

catenacyber: cf `APP_LAYER_PARSER_EOF_TS` defined in rust, used in C

victorjulien (changes requested):
See Philippe's comments

suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.memuse | 545016768 | 505591168 | 107.8% | 

Pipeline 5867

src/output-json-pgsql.c
catenacyber: cbindgen should put it in rust/gen/rust-bindings.h 
cf commit 87d9c44ec504f459578fdb551b16a1015b8239c7

It looks like it works if you do
```
-pub const PGSQL_LOG_PASSWORDS: u32 = BIT_U32!(1);
+pub const PGSQL_LOG_PASSWORDS: u32 = 2;
```

@jasonish do you know why cbindgen does not work with this macro ?
And there are other examples of constants defined twice. (like `FLOW_DIR_REVERSED`)
Should we have a CI check for this ?

victorjulien (approved):
Going to merge as-is. We can dedup the flags separately.

victorjulien:
Merged in #6854, thanks!

catenacyber:
> Going to merge as-is. We can dedup the flags separately.

Created https://redmine.openinfosecfoundation.org/issues/5012

src/output-json-pgsql.c
jufajardini: > It looks like it works if you do
> 
> ```
> -pub const PGSQL_LOG_PASSWORDS: u32 = BIT_U32!(1);
> +pub const PGSQL_LOG_PASSWORDS: u32 = 2;
> ```

Yeah, I noticed that, as well. That was the only way to make it work.

jasonish: > cbindgen should put it in rust/gen/rust-bindings.h cf commit [87d9c44](https://github.com/OISF/suricata/commit/87d9c44ec504f459578fdb551b16a1015b8239c7)
> 
> It looks like it works if you do
> 
> ```
> -pub const PGSQL_LOG_PASSWORDS: u32 = BIT_U32!(1);
> +pub const PGSQL_LOG_PASSWORDS: u32 = 2;
> ```
> 
> @jasonish do you know why cbindgen does not work with this macro ? 

Cbindgen is a rather simple parser that doesn't expand macros like this. There is experimental support for it, but you have to be using the nightly rust compiler, so its not an option for us.

> And there are other examples of constants defined twice. (like `FLOW_DIR_REVERSED`) Should we have a CI check for this ?

I'm not sure what CI would be checking for? Duplicate constants?

catenacyber: > Cbindgen is a rather simple parser that doesn't expand macros like this. There is experimental support for it, but you have to be using the nightly rust compiler, so its not an option for us.

Do you mean Cbindgen launches an external processus rustc nightly ? Or That we have to compile cbindgen with nightly

> I'm not sure what CI would be checking for? Duplicate constants?

Indeed

jasonish: > Do you mean Cbindgen launches an external processus rustc nightly ? Or That we have to compile cbindgen with nightly

It uses rustc, so nightly has to be set as your default toolchain, or the default rustc in your path.

catenacyber: So, what is best ? Should we avoid macros in const definitions ? So that they can be used in C ?

victorjulien: The last is closest to the `BIT_U32` macro, but overall I see this as a step back :-/

jufajardini: So what should be the preferable approach in this case, for now? Leave the duplicates, or keep it just Rust, with the one that is closest to the macro?

catenacyber: I just created https://github.com/eqrion/cbindgen/issues/735

catenacyber: I created https://redmine.openinfosecfoundation.org/issues/5012, not first priority

PR #6843, opened 2022-01-21
(draft) windows: allow libpcap - v1

jasonish:
Allow the Windows build to succeed with libpcap from msys2. This won't allow
for live capture, is useful for reading capture files, or in the future,
libsuricata.

Known issue: If you attempt to listen on a live interface, Suricata will
segfault.  I need to detect that its not available earlier in the startup
sequence.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1456,8 +1456,8 @@ jobs:
       - name: Running suricata-verify
         run: python3 ./suricata-verify/run.py -q
 
-  windows-msys2-mingw64:
-    name: Windows MSYS2 MINGW64
+  windows-msys2-mingw64-npcap:
+    name: Windows MSYS2 MINGW64 (NPcap)
     runs-on: windows-latest
     needs: [prepare-deps]
     defaults:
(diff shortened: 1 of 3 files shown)


victorjulien:
To avoid the "live" issue would it be enough to simply not register the packet source in `TmModuleReceivePcapRegister`

suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.memuse | 537868288 | 505591168 | 106.38% | 

Pipeline 5865

jasonish:
> To avoid the "live" issue would it be enough to simply not register the packet source in `TmModuleReceivePcapRegister`?

Actually there isn't an issue.  The NPCap build fails in the same way when the argument to `--pcap` can't be resolved.  The only real different between libpcap and npcap is that with libpcap no argument can be resolved to a device, while with npcap, a valid argument can.

My initial test that we could use `pcap_activate` was wrong as well, it appears to be in the msys2 libpcap.  I think the best option now is that if OS is windows, and not wpcap, cause the PCAP_DEV mode early and error out before doing much of anything as it will be clear to the user.

And then probably some ./configure output which pcap is being used..

So new PR will be on the way..

jasonish:
Continued with https://github.com/OISF/suricata/pull/6869.

PR #6851, opened 2022-01-23
Rust overflows v1

catenacyber:
None

Describe changes:

cc @satta for the mqtt commit

If I understand correctly , mqtt properties are part of a bigger message.
So, they have a length, and we should limit to parse what is within that length, and not what is after (even if it is still in the MQTT message)

### DIFF
--- a/rust/src/http2/parser.rs
+++ b/rust/src/http2/parser.rs
@@ -530,11 +530,11 @@ fn http2_parse_var_uint(input: &[u8], value: u64, max: u64) -> IResult<&[u8], u6
     for i in 0..varia.len() {
         varval += ((varia[i] & 0x7F) as u64) << (7 * i);
     }
-    varval += (finalv as u64) << (7 * varia.len());
-    if varval < max {
-        // this has overflown u64
+    if (finalv as u64) << (7 * varia.len()) > u64::MAX - varval {
+        // this would overflow u64
         return Ok((i3, 0));
     }
+    varval += (finalv as u64) << (7 * varia.len());
     return Ok((i3, varval));
 }
 
--- a/rust/src/mqtt/parser.rs
+++ b/rust/src/mqtt/parser.rs
@@ -106,13 +106,13 @@ fn parse_properties(input: &[u8], precond: bool) -> IResult<&[u8], Option<Vec<MQ
     // parse properties length
     match parse_mqtt_variable_integer(input) {
         Ok((rem, mut proplen)) => {
-            if proplen == 0 {
+            if proplen == 0 || proplen as usize > rem.len() {
                 // no properties
                 return Ok((rem, None));
             }
             // parse properties
             let mut props = Vec::<MQTTProperty>::new();
-            let mut newrem = rem;
+            let mut newrem = &rem[..proplen as usize];
             while proplen > 0 {
                 match parse_property(newrem) {
                     Ok((rem, val)) => {


satta:
Correct -- yes, it is safer here to restrict parsing to the indicated length of the properties. Thanks for spotting this.

suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpr1_stats_chk |
| .flow.memuse | 548500288 | 505591168 | 108.49% | 

Pipeline 5885

rust/src/http2/parser.rs
```diff
@@ -530,11 +530,11 @@ fn http2_parse_var_uint(input: &[u8], value: u64, max: u64) -> IResult<&[u8], u6
     for i in 0..varia.len() {
         varval += ((varia[i] & 0x7F) as u64) << (7 * i);
     }
-    varval += (finalv as u64) << (7 * varia.len());
-    if varval < max {
-        // this has overflown u64
+    if (finalv as u64) << (7 * varia.len()) > u64::MAX - varval {
+        // this would overflow u64
```
victorjulien, line 534: should this return error instead? Seems like such an overflow is not correct behavior in the protocol?

rust/src/mqtt/parser.rs
```diff
@@ -106,13 +106,13 @@ fn parse_properties(input: &[u8], precond: bool) -> IResult<&[u8], Option<Vec<MQ
     // parse properties length
     match parse_mqtt_variable_integer(input) {
         Ok((rem, mut proplen)) => {
-            if proplen == 0 {
+            if proplen == 0 || proplen as usize > rem.len() {
```
victorjulien, line 109: in what condition can this happen? Should it lead to error handling and/or setting of event?

satta: In theory, one can craft a message that has this value set to a larger value than the total length specified in the header. One could imagine raising an event here, yes.

rust/src/http2/parser.rs
catenacyber: Good idea indeed. Will look into it

rust/src/mqtt/parser.rs
catenacyber: @satta It looks like there are more problems with MQTT :
if `parse_connect` returns `Err::Incomplete` (for example because, there is not enough bytes for the properties), we will keep buffering forever because we have the full message with its length already cf `let rem = &fullrem[..len];`

Am I correct ?

I guess the solution is to use nom's `complete` (which turns Err::Incomplete into Err:Error)

Then, I guess there is more work to do to get the error to create an app-layer event...

catenacyber: I also think we can have `parse_properties` more idiomatic with nom by using `take` and `many`

rust/src/http2/parser.rs
sharksforarms: Could the `checked_add` function be used here? https://doc.rust-lang.org/std/primitive.u64.html#method.checked_add

catenacyber: looks relevant

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6864

rust/src/mqtt/parser.rs
satta: > @satta It looks like there are more problems with MQTT : if `parse_connect` returns `Err::Incomplete` (for example because, there is not enough bytes for the properties), we will keep buffering forever because we have the full message with its length already cf `let rem = &fullrem[..len];`
> 
> Am I correct ?

Hmm, I think you are right. We make sure we have the announced full length of the message (as defined in the common header `header.remaining_length`) buffered already because we try to handle buffering before trying to do any type-specific parsing in line 509ff:
```
            // We have not exceeded the maximum length limit, but still do not
            // have enough data in the input buffer to handle the full
            // message. Signal this by returning an Incomplete IResult value.
            if fullrem.len() < len {
                return Err(Err::Incomplete(Needed::Size(len - fullrem.len())));
            }
```
So if we notice later that the pre-filled buffer is not enough, we have a problem. But, maybe it helps to know that in a valid control packet ("message") we should _never_ generate `Err::Incomplete` somewhere else than in the code above since we SHOULD have buffered all the expected data to parse the whole message already. So any mismatch of expected and actually matched length will be an indication of an inconsistency in the message, and we should simply notify this and stop parsing since obviously something's amiss.

> I guess the solution is to use nom's `complete` (which turns Err::Incomplete into Err:Error)

I guess so, too. And ideally this should be done globally, whenever one of the `parse_*()` functions wants more data. What do you think?

catenacyber: Created https://redmine.openinfosecfoundation.org/issues/5018 about this

A global solution is good.
That would be wrapping the 
```
match message_type {
                MQTTTypeCode::CONNECT => match parse_connect(rem) {
```
in a `nom::complete` function

Let me know if you need more help ;-)

satta: Something like this? -> https://github.com/satta/suricata/commit/a3d26d59aa3ab25f4a16ee30be2075b0abdd41eb

catenacyber: Not sure about the `move` but the `complete` look nice :-)
Waiting for the PR ;-)

satta: Regarding the `move`: Got that from https://github.com/Geal/nom/blob/main/doc/upgrading_to_nom_5.md
Will try to get a suricata-verify case going before starting the PR.

satta: See https://github.com/OISF/suricata/pull/6988

catenacyber: > From what I could see, the original issue should not be a problem at all in the current master code, since nom7::bytes::complete::take() is used to pre-fetch a subslice of length proplen in line 116 of rust/src/mqtt/parser.rs

Hey @satta took me some time to get back to this

So, you say my patch from https://github.com/OISF/suricata/pull/6890/commits/3c265611fe830b5e7e4f395d762ba223bd316423 was enough ?

I do not think it is enough for all cases.
For instance `parse_connect` calls`parse_mqtt_string` which calls nom `length_data` which can return incomplete cf  https://docs.rs/nom/latest/nom/multi/fn.length_data.html
What do you think ?

satta: I think that the current solution wrapping everything in `complete()` also catches that and should be preferred. See my latest MR.

catenacyber: discussion continued in #7009 :-)

PR #6788, opened 2022-01-13
App maxtx 4530 v9

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4530

Describe changes:
- limits the number of active transactions per flow for HTTP2 and MQTT
- fixes MQTT transaction completion

This is to avoid DOS by quadratic complexity for protocols looping over the whole list of transactions looking for a
specific transation with an identifier, so that a new PDU gets processed within that transaction.

There may be other protocols with this problem (like Modbus)

Replaces #6783 with 
- fixing the tx count in the fuzz target

PR #6863, opened 2022-01-25
App maxtx 4530 v10

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4948

Describe changes:
- SMTP : prevents assertion from being triggered (is this assertion meaningful ?)

Replaces #6788 with 
- fixing the SMTP assertions triggered from https://redmine.openinfosecfoundation.org/issues/4948

### DIFF
--- a/doc/userguide/configuration/suricata-yaml.rst
+++ b/doc/userguide/configuration/suricata-yaml.rst
@@ -1375,6 +1375,36 @@ independent. The ``probing parsers`` will only run on the ``detection-ports``.
 SMB is commonly used to transfer the DCERPC protocol. This traffic is also handled by
 this parser.
 
+Configure HTTP2
+~~~~~~~~~~~~~~~
+
+HTTP2 has 2 parameters that can be customized.
+The point of these 2 parameters is to find a balance between the completeness
+of analysis and the resource consumption.
+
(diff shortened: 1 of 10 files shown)


catenacyber:
Looks like we need a maxtx for nfs too...

suricata-qa:

Pipeline 5920

victorjulien:
Since there is still work to be done, I've converted this to draft. Can you do a PR that does not include the Fuzz check? Then we can (hopefully) see a full CI pass on that. Then the Fuzz check can go last in when all pass.

catenacyber:
> Since there is still work to be done, I've converted this to draft. Can you do a PR that does not include the Fuzz check? Then we can (hopefully) see a full CI pass on that. Then the Fuzz check can go last in when all pass.

Done in https://github.com/OISF/suricata/pull/6881

catenacyber:
> Looks like we need a maxtx for nfs too...

And maybe one for tls... or a bigger rework of tls...

victorjulien:
> > Looks like we need a maxtx for nfs too...
> 
> And maybe one for tls... or a bigger rework of tls...

tls just uses a single fixed tx (state as tx) iirc

PR #6864, opened 2022-01-25
Rust overflows v2

catenacyber:
None

Describe changes:
- PGSQL : fixes unsigned overflow
- Adds a consistency check about timestamp microseconds

cc @satta for the mqtt commit
cc @jufajardini for the PGSQL commit

Modifies #6851 :
- adds PGSQL and timestamp microseconds commits
- MQTT : more idiomatic nom (still more work to do to avoid endless buffering with more `complete`, not as part of this PR)
- HTTP2 : sets an app-layer event in case of integer overflow


catenacyber:
Need a better fix for pgsql...

jufajardini:
> Need a better fix for pgsql...

I can look into it, thanks for pointing this bug out, @catenacyber !

jufajardini:
> > Need a better fix for pgsql...
> 
> I can look into it, thanks for pointing this bug out, @catenacyber !

The problem with pgsql is due to how the message parsers were written, they assume the first byte (the identifier) will be present). With this fix and approach, we let go of that, and then all underlying parsers fail... Bad design choice when planning the parser, I think.

The true fix to this would involve re-writing the parsers to ignore that first bit, since it has already been checked by the calling parser...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/6890

PR #6869, opened 2022-01-26
windows: build even if libpcap is not npcap - v3

jasonish:
A simplification of https://github.com/OISF/suricata/pull/6843.  We first try to link against `libwpcap` on Windows, then fall back to `libpcap`.  This will allow Suricata to build on Windows with the libpcap provided in msys2, but Suricata will lack the ability to listen on a live interface.

If a live interface is requested Suricata was not compiled against libwpcap, raise a fatal error. This is to prevent confusion when no interface name will ever resolve to an interface.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1456,8 +1456,8 @@ jobs:
       - name: Running suricata-verify
         run: python3 ./suricata-verify/run.py -q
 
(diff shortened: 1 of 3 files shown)


victorjulien (approved):
I'll run clang format on merge here

victorjulien:
Merged in #6870, thanks!

PR #6890, opened 2022-01-28
Rust overflows v3

catenacyber:
None

Describe changes:
- NFS : fixes unsigned underflow
- Adds a consistency check about timestamp microseconds

Modifies #6864 :
- removing PGSQL commit to be handled in https://redmine.openinfosecfoundation.org/issues/5016
- adding NFS commit
- MQTT : more idiomatic nom (still more work to do to avoid endless buffering with more `complete`, not as part of this PR)
- HTTP2 : sets an app-layer event in case of integer overflow

For NFS, should we do this ?
```
diff --git a/rust/src/nfs/nfs3_records.rs b/rust/src/nfs/nfs3_records.rs
index 66c02cf07..d1d1b31da 100644
--- a/rust/src/nfs/nfs3_records.rs
+++ b/rust/src/nfs/nfs3_records.rs
@@ -344,8 +344,8 @@ pub fn parse_nfs3_request_write(i: &[u8]) -> IResult<&[u8], Nfs3RequestWrite> {
     let (i, count) = be_u32(i)?;
     let (i, stable) = be_u32(i)?;
     let (i, file_len) = be_u32(i)?;
-    let (i, file_data) = take(file_len as usize)(i)?;
-    let (i, _file_padding) = cond(file_len % 4 !=0, take(4 - (file_len % 4)))(i)?;
+    let (i, file_data) = take(count as usize)(i)?;
+    let (i, _file_padding) = cond(count % 4 !=0, take(4 - (count % 4)))(i)?;
     let req = Nfs3RequestWrite {
         handle,
         offset,
```

cf https://datatracker.ietf.org/doc/html/rfc1813#section-3.3.7

### DIFF
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -295,6 +295,10 @@ impl Flow {
             let mut secs: u64 = 0;
             let mut usecs: u64 = 0;
             FlowGetLastTimeAsParts(self, &mut secs, &mut usecs);
+            if usecs > 999999 {
+                // if microseconds are inconsistent, reset them to 0
+                usecs = 0;
+            }
             std::time::Duration::new(secs, usecs as u32 * 1000)
         }
     }
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -593,11 +593,7 @@ impl NFSState {
         // for now assume that stable FILE_SYNC flags means a single chunk
         let is_last = if w.stable == 2 { true } else { false };
 
-        let mut fill_bytes = 0;
-        let pad = w.file_len % 4;
-        if pad != 0 {
-            fill_bytes = 4 - pad;
-        }
+        let fill_bytes = if w.file_len % 4 == 0 { 0 } else {4 - (w.file_len % 4) };
 
         let file_handle = w.handle.value.to_vec();
         let file_name = if let Some(name) = self.namemap.get(w.handle.value) {
@@ -648,8 +644,9 @@ impl NFSState {
         }
         if !self.is_udp {
             self.ts_chunk_xid = r.hdr.xid;
-            let file_data_len = w.file_data.len() as u32 - fill_bytes as u32;
-            self.ts_chunk_left = w.file_len as u32 - file_data_len as u32;
+            // parse_nfs3_request_write does (i, file_data) = take(file_len as usize)(i)?;
+            // so this should always be 0... maybe count rather than file_len was intended
+            self.ts_chunk_left = w.file_len as u32 - w.file_data.len() as u32;
             self.ts_chunk_fh = file_handle;
             SCLogDebug!("REQUEST chunk_xid {:04X} chunk_left {}", self.ts_chunk_xid, self.ts_chunk_left);
         }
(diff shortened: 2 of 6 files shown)


rust/src/core.rs
```diff
@@ -295,6 +295,10 @@ impl Flow {
             let mut secs: u64 = 0;
             let mut usecs: u64 = 0;
             FlowGetLastTimeAsParts(self, &mut secs, &mut usecs);
+            if usecs > 999999 {
```
victorjulien, line 298: have you observed this?

catenacyber: From fuzzing yes...
Maybe I can change fuzzpcap to produce only valid timestamps...

suricata-qa:

Pipeline 5959

rust/src/core.rs
victorjulien: this comes directly from `Flow::last_ts`, which is updated directly from `Packet::timestamp` iirc, so perhaps some lower level `DEBUG_VALIDATE_BUG_ON`s might be helpful.

victorjulien:
Merged the http2 and mqtt in #6903, thanks. I think flow and nfs need some more work/discussion

rust/src/core.rs
catenacyber: This comes indeed from `flow->lastts.tv_usec` which comes from `COPY_TIMESTAMP(&p->ts, &f->lastts);`

Then, it depends on the source.

For the fuzz targets, there is `p->ts.tv_usec = header->ts.tv_usec;` where header comes from `pcap_next_ex(pkts, &header, &pkt);
Same goes for source-pcap-file-helper.c:    `p->ts.tv_usec = h->ts.tv_usec;`

Should the sources rather do `p->ts.tv_usec = header->ts.tv_usec % 1000000;` ?
and then we can have a `DEBUG_VALIDATE_BUG_ON` later for all sources ?

rust/src/nfs/nfs.rs
```diff
@@ -648,8 +644,9 @@ impl NFSState {
         }
         if !self.is_udp {
             self.ts_chunk_xid = r.hdr.xid;
-            let file_data_len = w.file_data.len() as u32 - fill_bytes as u32;
-            self.ts_chunk_left = w.file_len as u32 - file_data_len as u32;
+            // parse_nfs3_request_write does (i, file_data) = take(file_len as usize)(i)?;
+            // so this should always be 0... maybe count rather than file_len was intended
+            self.ts_chunk_left = w.file_len as u32 - w.file_data.len() as u32;
```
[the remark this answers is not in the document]
inashivb: @catenacyber ^^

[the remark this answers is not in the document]
catenacyber: Tracked as https://redmine.openinfosecfoundation.org/issues/5145

catenacyber:
> Merged the http2 and mqtt in #6903, thanks. I think flow and nfs need some more work/discussion

NFS ticket  https://redmine.openinfosecfoundation.org/issues/5145
flow : do not see on oss-fuzz (yet ?)

So closing this PR

PR #6893, opened 2022-01-28
Datasets parsing/v1

inashivb:
If there is a space following a keyword that does not expect a value,
the rule fails to load due to improper value evaluation.
e.g. Space after `set` command
```
alert http any any -> any any (http.user_agent; dataset:set  ,ua-seen,type string,save datasets.csv; sid:1;)
```

gives error
```
[ERRCODE: SC_ERR_UNKNOWN_VALUE(129)] - dataset action "" is not supported.
```

Fix this by handling values correctly for such cases.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5019

suricata-verify-pr: 801

### DIFF
--- a/src/detect-dataset.c
+++ b/src/detect-dataset.c
@@ -39,9 +39,6 @@
 #include "util-print.h"
 #include "util-misc.h"
 
-#define PARSE_REGEX         "([a-z]+)(?:,\\s*([\\-_A-z0-9\\s\\.]+)){1,4}"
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 5958

jlucovsky:
I think we need an s-v test with this PR.

victorjulien (approved):

victorjulien:
Merged in #7210, thanks!

PR #6895, opened 2022-01-28
decode: use configurable value for packet_alert_max - v1

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4207

Describe changes:
- add a packet-alert-max config option to suricata.yaml
- dynamically allocate the array of PacketAlert inside PacketAlerts, so array size can be configurable
- this change made lots of unittests fail, so used a coccinelle script to massively change Packet declaration in unittests to call PacketGetFromAlloc

suricata-verify-pr: 694
https://github.com/OISF/suricata-verify/pull/694

PR #6896, opened 2022-01-29
Packet alerts/v3

victorjulien:
#6895 with an fix/optimization for alert queue removal/insert.

@catenacyber can you help review the `memmove` logic? The relation between array index, array size and array use cnt is easily messed up I think.

suricata-verify-pr: 694

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2019 Open Information Security Foundation
+/* Copyright (C) 2007-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -74,6 +74,31 @@ extern bool stats_decoder_events;
 extern const char *stats_decoder_events_prefix;
 extern bool stats_stream_events;
 uint8_t decoder_max_layers = PKT_DEFAULT_MAX_DECODED_LAYERS;
+uint16_t packet_alert_max = PACKET_ALERT_MAX;
+
+/**
+ * \brief Initialize PacketAlerts with dynamic alerts array size
+ *
+ */
+PacketAlert *CreatePacketAlert(void)
+{
+    PacketAlert *pa_array = NULL;
+    if (RunmodeIsUnittests()) {
+        pa_array = SCCalloc(PACKET_ALERT_MAX, sizeof(PacketAlert));
+    } else {
+        pa_array = SCCalloc(packet_alert_max, sizeof(PacketAlert));
+    }
+    BUG_ON(pa_array == NULL);
+
+    return pa_array;
+}
+
+void FreePacketAlert(PacketAlert *pa)
+{
+    if (pa != NULL) {
+        SCFree(pa);
+    }
+}
 
 static int DecodeTunnel(ThreadVars *, DecodeThreadVars *, Packet *, const uint8_t *, uint32_t,
         enum DecodeTunnelProto) WARN_UNUSED;
@@ -780,6 +805,19 @@ void DecodeGlobalConfig(void)
         } else {
             decoder_max_layers = value;
         }
+    }
+    GetPacketMaxAlertConfig();
+}
+
+void GetPacketMaxAlertConfig(void)
+{
+    intmax_t max = 0;
+    if (ConfGetInt("packet-alert-max", &max) == 1) {
+        SCLogDebug("detect->packet_alert_max set to %" PRIdMAX "", max);
+        packet_alert_max = max;
+    } else {
+        packet_alert_max = PACKET_ALERT_MAX;
+        SCLogDebug("detect.packet_alert_max set to default %d", packet_alert_max);
     }
 }
 
--- a/src/decode.h
+++ b/src/decode.h
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2020 Open Information Security Foundation
+/* Copyright (C) 2007-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -299,15 +299,20 @@ typedef struct PacketAlert_ {
 /** alert is in a frame, frame_id set */
 #define PACKET_ALERT_FLAG_FRAME 0x20
 
+extern uint16_t packet_alert_max;
 #define PACKET_ALERT_MAX 15
 
 typedef struct PacketAlerts_ {
     uint16_t cnt;
-    PacketAlert alerts[PACKET_ALERT_MAX];
+    PacketAlert *alerts;
     /* single pa used when we're dropping,
      * so we can log it out in the drop log. */
     PacketAlert drop;
 } PacketAlerts;
+
+PacketAlert *CreatePacketAlert(void);
+
+void FreePacketAlert(PacketAlert *pa);
 
 /** number of decoder events we support per packet. Power of 2 minus 1
  *  for memory layout */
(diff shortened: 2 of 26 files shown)


src/decode.h
```diff
@@ -760,11 +765,13 @@ void CaptureStatsSetup(ThreadVars *tv, CaptureStats *s);
...
  */
-#define PACKET_INITIALIZE(p) {         \
-    SCMutexInit(&(p)->tunnel_mutex, NULL); \
-    PACKET_RESET_CHECKSUMS((p)); \
-    (p)->livedev = NULL; \
-}
+#define PACKET_INITIALIZE(p)                                                                       \
+    {                                                                                              \
+        SCMutexInit(&(p)->tunnel_mutex, NULL);                                                     \
+        (p)->alerts.alerts = CreatePacketAlert();                                                  \
```
catenacyber, line 771: Could/Should we not defer the need for allocation until there truly is one alert ?

src/decode.c
```diff
@@ -781,6 +806,19 @@ void DecodeGlobalConfig(void)
...
+}
+
+void GetPacketMaxAlertConfig(void)
+{
+    intmax_t max = 0;
+    if (ConfGetInt("packet-alert-max", &max) == 1) {
+        SCLogDebug("detect->packet_alert_max set to %" PRIdMAX "", max);
+        packet_alert_max = max;
+    } else {
+        packet_alert_max = PACKET_ALERT_MAX;
```
catenacyber, line 819: is it not redundant with line 77 ?

src/decode.c
```diff
@@ -781,6 +806,19 @@ void DecodeGlobalConfig(void)
...
         }
     }
+    GetPacketMaxAlertConfig();
+}
+
+void GetPacketMaxAlertConfig(void)
+{
+    intmax_t max = 0;
+    if (ConfGetInt("packet-alert-max", &max) == 1) {
+        SCLogDebug("detect->packet_alert_max set to %" PRIdMAX "", max);
```
catenacyber, line 816: Could have only one debug line for both cases I guess

src/decode.c
```diff
@@ -74,6 +74,31 @@ extern bool stats_decoder_events;
...
+
+/**
+ * \brief Initialize PacketAlerts with dynamic alerts array size
+ *
+ */
+PacketAlert *CreatePacketAlert(void)
+{
+    PacketAlert *pa_array = NULL;
+    if (RunmodeIsUnittests()) {
+        pa_array = SCCalloc(PACKET_ALERT_MAX, sizeof(PacketAlert));
```
catenacyber, line 87: Do we have any idea what is the performance impact ?
Could/should we use a pool of preallocated buffers ?

src/detect-engine-alert.c
```diff
@@ -154,27 +154,21 @@ int PacketAlertCheck(Packet *p, uint32_t sid)
...
+    if (pos >= p->alerts.cnt) {
         SCLogDebug("removing %u failed, pos > cnt %u", pos, p->alerts.cnt);
-        return 0;
+        return;
     }
 
-    for (i = pos; i <= p->alerts.cnt - 1; i++) {
-        memcpy(&p->alerts.alerts[i], &p->alerts.alerts[i + 1], sizeof(PacketAlert));
+    const uint16_t after = (p->alerts.cnt - 1) - pos;
+    if (after) {
```
catenacyber, line 166: Is `p->alerts.alerts` somehow sorted ?
If not, it would be more efficient to just move the last alert to the removed position.
That would be like `memmove(p->alerts.alerts + pos, &p->alerts.alerts[p->alerts.cnt - 1], sizeof(PacketAlert)`

src/detect-engine-alert.c
```diff
@@ -206,14 +200,18 @@ int PacketAlertAppend(DetectEngineThreadCtx *det_ctx, const Signature *s,
...
+        /* find position to insert */
+        for (i = p->alerts.cnt - 1; i >= 0 && p->alerts.alerts[i].num > s->num; i--)
+            ;
 
         i++; /* The right place to store the alert */
 
+        const uint16_t target_pos = i + 1;
+        const uint16_t space_post_target = packet_alert_max - 1 - i;
+        const uint16_t to_move = MIN(space_post_target, (p->alerts.cnt - i));
+        memmove(p->alerts.alerts + target_pos, p->alerts.alerts + i,
```
catenacyber, line 212: Will this work correctly if `to_move == 0` ?

src/detect-engine-alert.c
```diff
@@ -206,14 +200,18 @@ int PacketAlertAppend(DetectEngineThreadCtx *det_ctx, const Signature *s,
         p->alerts.alerts[p->alerts.cnt].frame_id =
                 (flags & PACKET_ALERT_FLAG_FRAME) ? det_ctx->frame_id : 0;
     } else {
-        /* We need to make room for this s->num
-         (a bit ugly with memcpy but we are planning changes here)*/
-        for (i = p->alerts.cnt - 1; i >= 0 && p->alerts.alerts[i].num > s->num; i--) {
-            memcpy(&p->alerts.alerts[i + 1], &p->alerts.alerts[i], sizeof(PacketAlert));
-        }
+        /* find position to insert */
+        for (i = p->alerts.cnt - 1; i >= 0 && p->alerts.alerts[i].num > s->num; i--)
```
catenacyber, line 204: There is a duplicate test for `p->alerts.alerts[p->alerts.cnt - 1].num < s->num`

catenacyber: And 
`if (p->alerts.cnt == 0 || (p->alerts.cnt > 0 &&
                               p->alerts.alerts[p->alerts.cnt - 1].num < s->num)) {`
can be simplified to
`if (p->alerts.cnt == 0 || 
                               p->alerts.alerts[p->alerts.cnt - 1].num < s->num) {`
as alerts.cnt is unsigned cf `PacketAlerts_`

src/decode.c
jufajardini: Yes, you're right.

src/decode.h
jufajardini: Victor has mentioned something similar, actually, iirc. I'll have a look at how to do address that...

catenacyber:
>  can you help review the `memmove` logic? The relation between array index, array size and array use cnt is easily messed up I think.
> 

It looks dangerous indeed... I left some questions

src/decode.c
victorjulien: The packets themselves are pooled and reused, so this allocation is a once per packet step.

src/decode.h
victorjulien: The packets and their alert structures are pooled and reused all the time. We don't free them, only clear them. So allocing this at init makes sense and avoids alloc time overhead at runtime for an already expensive packet (ie an alerting packet)

src/detect-engine-alert.c
victorjulien: yes this logic is to have it sorted by Signature::num (the internal id, assigned after sort by sig priority).

src/detect-engine-alert.c
victorjulien: I'll add a BUG_ON but given the check above it should not ever be 0

catenacyber: > it should not ever be 0

It can be 0 : if `space_post_target` is 0 c-which can be the case if you have like `packet_alert_max`= 3, you have 3 alerts with num 3,5,7 and you want to add one with num 6

Than no memmove, just replace the last one


Otherwise, all is good :-)

suricata-qa:
  

Pipeline 6035

catenacyber:
> ERROR: QA failed on build_asan.
> 
> Pipeline 6035

Because of needed rebase

jufajardini:
Rebased and addressed things related to my changes on https://github.com/OISF/suricata/pull/6929

src/detect-engine-alert.c
jufajardini: > There is a duplicate test for `p->alerts.alerts[p->alerts.cnt - 1].num < s->num`

Sorry, I don't think I get this. Can you point out where?

src/detect-engine-alert.c
jufajardini: S-V tests failing with `BUG_ON(to_move == 0)` atm:
===> http2-bugfixes: FAILED: got exit code -6, expected 0
===> http2-files: FAILED: got exit code -6, expected 0
===> http-sticky-server: FAILED: got exit code -6, expected 0
===> smb2-07-frames: FAILED: got exit code -6, expected 0
===> tls13-draft28-frames: FAILED: got exit code -6, expected 0

jufajardini:
Comments addressed with https://github.com/OISF/suricata/pull/6999
Can this one be closed?

victorjulien:
Thanks @jufajardini

src/detect-engine-alert.c
catenacyber: It is tested a first time in the surrounding if. Then in the else clause, it is tested again in the loop

jufajardini: I think I see it now. Because the surrounding if has it `<` while the for checks for `>` I didn't realize it before. Thanks :)

PR #6909, opened 2022-01-31
devguide: move image generation into sphinx setup - v1

jasonish:
While ReadTheDocs can't execute arbitrary scripts, we can install
mscgen in the container that builds the docs (see .readthedocs.yaml).

Then instead of calling generate-images.sh from the Makefile, move this
into the setup function defined on conf.yaml, which will generate the
scripts as part of a normal Sphinx workflow.

This should give us an image generation solution that works on
ReadTheDocs, and locally within anyones build system provided they have
mscgen installed.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,3 +7,7 @@ python:
   # Use an empty install section to avoid RTD from picking up a non-python
   # requirements.txt file.
   install: []
+
+build:
+  apt_packages:
+    - mscgen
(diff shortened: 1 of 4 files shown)


jufajardini (commented):
That's awesome!! I did get to the mscgen package with read the docs, but I then tried way more complicated ways, for I didn't know how to make the script work from the yaml. <3

victorjulien (approved):

PR #6921, opened 2022-02-02
flow: fix timestamp in pseudo packet setup

regit:
When Suricata was creating a pseudo packet for flow cleaning
purpose, it was using the timestamp of first packet of the pcap
file when running in pcap mode.

As a result, the timestamp of events triggered during this
flow cleaning phase were completely wrong. As they did often occurs
before the start of the flow itself.

This patch fix this by using flow lastts as timestamp of the packet.
If this is not exact, this is at least really close to reality.

In the case of live traffic, this is also correct as TimeGet function
will get the time when the callback is run which can differ a lot
from the time of the event (which is before timestamp of last seen
packet).

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5037

Describe changes:
- Use lastts for pseudo packet timestamp

### DIFF
--- a/src/flow-timeout.c
+++ b/src/flow-timeout.c
@@ -246,8 +246,7 @@ static inline Packet *FlowForceReassemblyPseudoPacketSetup(Packet *p,
                                               (uint16_t *)p->tcph, 20, 0);
     }
 
-    memset(&p->ts, 0, sizeof(struct timeval));
-    TimeGet(&p->ts);
+    p->ts = f->lastts;
 
     if (direction == 0) {
         if (f->alparser && !STREAM_HAS_SEEN_DATA(&ssn->client)) {


src/flow-timeout.c
```diff
@@ -246,8 +246,7 @@ static inline Packet *FlowForceReassemblyPseudoPacketSetup(Packet *p,
                                               (uint16_t *)p->tcph, 20, 0);
     }
 
-    memset(&p->ts, 0, sizeof(struct timeval));
-    TimeGet(&p->ts);
+    p->ts = f->lastts;
```
victorjulien, line 249: The flows `lastts` is set to the timestamp of the last packet of the flow. If the `flow-timeout` value is for example 10 minutes, the pseudo packet could use a timestamp 10 minutes in the past. If it is larger, e.g. several hours, this gets worse.

victorjulien: Wrt the pcap file use case, this will need a better explanation and analysis.

victorjulien (changes requested):
Not acceptable as is, will need more discussion and analysis.

src/flow-timeout.c
regit: I've updated https://redmine.openinfosecfoundation.org/issues/5037 with a pcap and some jq commands explaining the issue.

regit: With regards to live mode, the concept is to take the timestamp of the last packet of the flow as the actual timestamp of the transaction is just before this point in time. From a conceptual point of view, Suricata timeout the flow at lastts + N secs and Suricata creates a pseudo packet to inspect the content that is in need for processing. This packet is fake but it is an implementation "limitation"/constraint that Suricata did inspect it so late. The actual application layer traffic was really closer than lastts than it is the timeout date. So it makes more sense to use lastts instead of time().

suricata-qa:
Warning: no commits in this PR have specified the following ticket(s):
- 5037 - https://redmine.openinfosecfoundation.org/issues/5037

Please update the commit(s) and submit a new PR.

suricata-qa:

Pipeline 6081

PR #6917, opened 2022-02-02
devguide: move into userguide as last chapter - v2

jufajardini:
DevGuide now shows up as last chapter (25th) of UserGuide, using the approach pointed out by @jasonish for using Sphinx build for handling the image generation script.

This fails with `make distcheck` but wasn't able to figure out how to fix that.

Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/4909

Previous PR: https://github.com/OISF/suricata/pull/6732

Describe changes:
- clean up makefiles, since script should be called from conf.py now
- pass devguide as a subdir in the userguide makefile
- opportunistic commit: devguide: rust-testing: clarify path to calling tests with cargo test

Preview at:
https://suri-rtd-test.readthedocs.io/en/latest/devguide/

PR #6922, opened 2022-02-02
devguide: move into userguide as last chapter - v3

jufajardini:
This still fails with make distcheck.
I tried adding `tools/generate-images.sh` directly as a `EXTRA_DIST`, but it still didn't work, so this version only lists the directory for the script. I noticed that when the failure occurs the build command seems to be called from a deeper subdir, somehow, and wonder if this could be affecting, even though the files should have been copied for the dist package.

I believe we didn't see this failure happening in master because the command is run for the userguide only, and therefore the script wasn't being invoked before.

Describe changes:
- added a commit to fix the build of the devguide in the dist package.

Link to redmine ticket:

Previous PR: https://github.com/OISF/suricata/pull/6917

Preview at:
https://suri-rtd-test.readthedocs.io/en/latest/devguide/


.github/workflows/builds.yml
```diff
@@ -221,6 +221,9 @@ jobs:
                 sudo \
                 which \
                 zlib-devel
+          # Packages needed for some diagrams generation.
+          dnf -y install epel-release
+          dnf -y install mscgen
           # These packages required to build the PDF.
           yum -y install \
```
inashivb, line 228: Q: why do we use `yum` for all the packages?

inashivb (commented):
Q: Do we need the static files again in devguide? Do we change something there or is it required by RTD?

jufajardini:
> Q: Do we need the static files again in devguide? Do we change something there or is it required by RTD?

I think that's something that the `conf.py` script is doing. Currently I've let the devguide in such a way that it could be built and generated independently, if wanted, and so the conf.py, among other files, stayed. If this is judged unnecessary, then I think we could/would also remove those files....

.github/workflows/builds.yml
jasonish: As of CentOS 8 (Alma, RHEL, etc) `yum` is just an alias to `dnf`.  When I wrote this, I probably still had the muscle memory from CentOS 7 days where there was only `yum`.

inashivb: oooh didn't know it was an alias now. Thank you!

jasonish (changes requested):
I think we should just focus on it being part of the main guide, and not worry about making the dev part being independently buildable - that just doubles the amount "infrastructure" we need to worry about, test, etc.

Has this been a specific request already?

jufajardini:
> I think we should just focus on it being part of the main guide, and not worry about making the dev part being independently buildable - that just doubles the amount "infrastructure" we need to worry about, test, etc.
> 
> Has this been a specific request already?

Not that I have seen anywhere. Oki, I'll submit a new PR without those - but from my local tests I think this won't solve the make distcheck thing, yet :/

jufajardini:
followed by: https://github.com/OISF/suricata/pull/6975

PR #6923, opened 2022-02-03
fuzz: test for too many open txs in a flow

catenacyber:

Describe changes:
- fuzz : finds if a flow has too many open/live transactions
  
Comes from #6863 and CI should already catch missing protocol

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -20,7 +20,7 @@
 use std;
 use std::cmp;
 use std::collections::HashMap;
-use std::ffi::{CStr, CString};
+use std::ffi::CString;
 
 use nom7::{Err, Needed};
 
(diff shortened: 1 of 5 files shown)


suricata-qa:

Pipeline 6103

victorjulien:
Should this be marked "ready for review"?

catenacyber:
> Should this be marked "ready for review"?

Yes, I pushed some commits to fix the NFS bug triggered by CIFuzz...

rust/src/nfs/nfs.rs
```diff
@@ -359,6 +350,19 @@ impl NFSState {
...
         tx.id = self.tx_id;
+        if self.transactions.len() > unsafe { NFS_MAX_TX } {
+            // set at least one another transaction to the drop state
+            for tx_old in &mut self.transactions {
+                if !tx_old.request_done || !tx_old.response_done {
+                    tx_old.request_done = true;
+                    tx_old.response_done = true;
+                    tx_old.is_file_closed = true;
+                    tx_old.tx_data.set_event(NFSEvent::TooManyTransactions as u8);
+                    self.events += 1;
```
inashivb, line 361: Q: what to we keep the events count for?

catenacyber: It is indeed unused, added a new commit to remove it.

@jasonish is there some rust tool to detect this ?

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7005

rust/src/nfs/nfs.rs
jasonish: Do you mean unused variables?  Unread vars generate a warning by default.  However, this is a read and a write of the var, so probably not.  Maybe clippy can?

catenacyber: Yes unused variables.
Oh, I see technically `+= 1` is a read and a write... But it is not so much of a read semantically...

Does clippy get run on CI somewhere ?

PR #6927, opened 2022-02-04
userguide: document frame rule keyword - v1

jufajardini:
Add a section about the frame keyword, explaining basic usages, listing
protocols that support said keyword and showing a few rule examples
taken from suricata-verify tests.

Task #4980

https://redmine.openinfosecfoundation.org/issues/4980

### DIFF
--- a/doc/userguide/rules/frame-keywords.rst
+++ b/doc/userguide/rules/frame-keywords.rst
@@ -1,0 +1,80 @@
+Frame Keywords
+**************
+
+Frame keywords will work for application layer protocols that have frame support added.
+As such, the list of available frame keywords vary in a per-protocol basis.
+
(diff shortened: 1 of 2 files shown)


doc/userguide/rules/frame-keywords.rst
```diff
@@ -0,0 +1,80 @@
+Frame Keywords
```
victorjulien, line 1: think it would be good to add a block on the shorthand notation, e.g.

with `tls.pdu`, you can use it as `alert tcp ... frame:pdu`.

jufajardini: TIL. Thanks. :P

jufajardini:
Requested change incorporated with: https://github.com/OISF/suricata/pull/6930

PR #6718, opened 2021-12-16
doc/devguide: add page on Stream Frame support - v3

jufajardini:
This document offers a perspective more useful for developers. Once Frame support has been merged, we'll also add pages for rule writers and users (Frame logging).

Previous PR: #6704 

Describe changes:
- rebase & build on top of https://github.com/OISF/suricata/pull/6713 (to benefit from `literalinclude`)
- add short note on what Frame means in Suri context
- use `literalinclude` for code snippets
- reorganize contents for more fluidity
- add a bit more explanation on `StreamSlice`
- update rules, images and other places where Frame, PDU, Record & StreamSlice were still too blurred
- remove Scope section (table of contents works for that)

TODO add pages about:
- rule writing
 - frame logging

Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/4697

Compiled version can be seen at http://stream-frame-support-v3.surge.sh/extending/detect/app-layer-frames.html

PR #6928, opened 2022-02-04
doc/devguide: add page on Frame support - v4

jufajardini:
Previous PR: https://github.com/OISF/suricata/pull/6718

Describe changes:
- Remove parts of the explanation that would make more sense from the user's perspective, or maybe for a blog post on the new feature
- Reorganize contents, give more emphasis on how to add frame support to an applayer parser
- Update code snippets
- Remove note about being a work in progress

TODO:
- once the page about the frame keyword has been merged, update this one to link to it

Link to redmine ticket: 
- https://redmine.openinfosecfoundation.org/issues/4697

Compiled version can be seen at 
https://stream-frame-support-v4.surge.sh/extending/app-layer/app-layer-frames.html


jufajardini:
Force pushed the removal of a change that should not be.

doc/devguide/extending/app-layer/app-layer-frames.rst
```diff
@@ -0,0 +1,129 @@
...
+    :start-at: pub fn new_tc
+    :lines: 1-6
+    :dedent: 4
+
+The parameters represent:
+
+- ``flow``: dedicated data type to carry specific flow-related data
+- ``stream_slice``: dedicated data type to carry stream data, shown bellow
+- ``frame_start``: a pointer to the start of the frame buffer in the stream, ``cur_i`` in the example code snippet
+- ``frame_len``: what we expect the frame length to be (the engine may need to wait until it has enough data...)
```
jufajardini, line 89: I think this may require a bit more explanation. Maybe telnet example?

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/6933

doc/devguide/extending/app-layer/app-layer-frames.rst
victorjulien: "if the length is unknown pass -1"
"length can exceed the current input"

In telnet or other protocols that search for a end of frame char we'd generally start with length unknown, then update the length when the char is found

In smb and other binary protocols where the frames have explicit length fields we set frames to match those lengths, even if the length is bigger than the current input

PR #6937, opened 2022-02-07
devguide: add page on Frame support - v6

jufajardini:
Previous PR: https://github.com/OISF/suricata/pull/6933

Describe changes:
- improve explanation about `frame_len`, with regards to frame registering

TODO:
- once the page about the frame keyword has been merged, update this one to link to it

Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/4697

Compiled version can be seen at
https://stream-frame-support-v6.surge.sh/extending/app-layer/app-layer-frames.html

PR #6946, opened 2022-02-07
devguide: add page on Frame support - v7

jufajardini:
Previous PR: https://github.com/OISF/suricata/pull/6937

Describe changes:
- Add explanation about the importance of putting thought when deciding frame types, and to following Suricata naming conventions
- further develop what is the expected outcome once frame support has been added, with rule examples
- explain that frame output has hardcoded limits
- Include code snippets for C codebase
- include examples and step for registering relevant frame callbacks when registering parser

TODO:
- once the page about the frame keyword has been merged, update this one to link to it

Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/4697

Compiled version can be seen at
https://stream-frame-support-v7.surge.sh/extending/app-layer/app-layer-frames.html


doc/devguide/extending/app-layer/app-layer-frames.rst
```diff
@@ -0,0 +1,227 @@
...
+
+``StreamSlice`` contains the input data to the parser, alongside other Stream-related data important in parsing context. Definition  is found in *applayer.rs*:
+
+.. literalinclude:: ../../../../rust/src/applayer.rs
+    :caption: rust/src/applayer.rs
+    :language: rust
+    :start-at: pub struct StreamSlice
+    :end-before: impl StreamSlice
+
+Registering relevant frame callbacks:
```
victorjulien, line 129: can we add that these are inferred from the frame enum's `AppLayerFrameType` derive statement?

victorjulien:
I wonder if we can format the literal include blocks a bit nicer, which I assume means we'll have to reformat the actual code... I think the default line length we use in the code (100ish) doesn't work so nicely in the docs.

victorjulien:
Looking good to me, just a minor suggestion other than the cosmetics question

jufajardini:
> I wonder if we can format the literal include blocks a bit nicer, which I assume means we'll have to reformat the actual code... I think the default line length we use in the code (100ish) doesn't work so nicely in the docs.

Indeed, some of the snippets for `smb` and `telnet` are a bit annoying with the need to scroll. 
I haven't found a quick way to re-format the code imported with the `literalinclude` directive, though. Do I have permission to format the original code to fit into a shorter line length?

doc/devguide/extending/app-layer/app-layer-frames.rst
```diff
@@ -0,0 +1,227 @@
...
+These are the frame registration functions highlighted above:
+
+.. literalinclude:: ../../../../rust/src/smb/smb.rs
+    :caption: rust/src/smb/smb.rs
+    :language: rust
+    :start-at: fn add_nbss_ts_frames(&mut
+    :end-before: fn add_smb1_ts_hdr_data
+    :dedent: 4
+
+Frame registering with the `telnet <https://github.com/OISF/suricata/blob/master/rust/src/telnet/telnet.rs#L171>`_ parser:
```
jufajardini, line 97: after re-working code formatting. make sure to change this, too.

doc/devguide/extending/app-layer/app-layer-frames.rst
```diff
@@ -0,0 +1,227 @@
...
+
+.. _telnet-snippet:
+
+.. literalinclude:: ../../../../rust/src/telnet/telnet.rs
+    :caption: rust/src/telnet/telnet.rs
+    :language: rust
+    :start-at: fn parse_request
+    :end-at: TelnetFrameType::Data as u8)
+    :lines: 1, 19-29
+    :dedent: 4
```
jufajardini, line 107: this is duplicated code? remove.

jasonish:
> I wonder if we can format the literal include blocks a bit nicer, which I assume means we'll have to reformat the actual code... I think the default line length we use in the code (100ish) doesn't work so nicely in the docs.

I tweaked the CSS to widen the page which works, but kind of goes against guidelines in the RTD theme which claims their wide is based on reading comprehension.

Please with it wide enoug to not scroll, you need about a 1400px width monitor to comfortably view, and this doesn't help the PDF case
Screenshot from 2022-02-08 16-22-18
.

doc/devguide/extending/app-layer/app-layer-frames.rst
```diff
@@ -0,0 +1,227 @@
...
+
+Frame registering with the `telnet <https://github.com/OISF/suricata/blob/master/rust/src/telnet/telnet.rs#L171>`_ parser:
+
+.. _telnet-snippet:
+
+.. literalinclude:: ../../../../rust/src/telnet/telnet.rs
+    :caption: rust/src/telnet/telnet.rs
+    :language: rust
+    :start-at: fn parse_request
+    :end-at: TelnetFrameType::Data as u8)
```
jasonish, line 105: I reformatted some code and these includes started to fail.  I wonder if we should drop magic start/stop comments to match on instead? It would also be a clue in the code that modifying the code might need modification of the documentation.

jufajardini: If we're ok with adding those, this is indeed better. It is how these options are usually used, I reckon. 

I do like the idea of having the failure because of that: it tells us that we may need to update the documentation...

jufajardini:
Feedback and suggestions incorporated on https://github.com/OISF/suricata/pull/6968

PR #6970, opened 2022-02-10
xbits parsing in Rust - v3

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4786

Changes since last rev:
- Fix for heap buffer overflow
- Handling of space and special cases
- cleanup of code

### DIFF
--- a/doc/userguide/rules/xbits.rst
+++ b/doc/userguide/rules/xbits.rst
@@ -8,8 +8,6 @@ Syntax::
     xbits:<set|unset|isset|isnotset|toggle>,<name>,track <ip_src|ip_dst|ip_pair>;
     xbits:<set|unset|isset|toggle>,<name>,track <ip_src|ip_dst|ip_pair> \
         [,expire <seconds>];
-    xbits:<set|unset|isset|toggle>,<name>,track <ip_src|ip_dst|ip_pair> \
-        [,expire <seconds>];
 
 Notes
 ~~~~~
@@ -21,9 +19,13 @@ Notes
    ``track ip_dst``, if you want to match on the server response,
    you check it (``isset``) with ``track ip_src``.
 
--  To not alert, use ``noalert;``
+-  To not alert, use ``xbits:noalert;``, it can only be used as a standalone
+   option like ``flowbits``.
 
-- the ``toggle`` option will flip the value of the xbits.
+-  the ``toggle`` option will flip the value of the xbits.
+
+-  the order of options is strict, therefore, using ``expire`` option before
+   ``track`` for example shall be considered invalid
 
 -  See also:
 
--- a/rust/src/detect/xbits.rs
+++ b/rust/src/detect/xbits.rs
@@ -1,0 +1,224 @@
+use std::ffi::{CStr, CString};
+use std::os::raw::{c_char, c_void};
+
+pub const DETECT_XBITS_CMD_SET: u8 = 0;
+pub const DETECT_XBITS_CMD_TOGGLE: u8 = 1;
+pub const DETECT_XBITS_CMD_UNSET: u8 = 2;
+pub const DETECT_XBITS_CMD_ISNOTSET: u8 = 3;
+pub const DETECT_XBITS_CMD_ISSET: u8 = 4;
+pub const DETECT_XBITS_CMD_NOALERT: u8 = 5;
+pub const DETECT_XBITS_CMD_MAX: u8 = 6;
+pub const DETECT_XBITS_TRACK_IPSRC: u8 = 0;
+pub const DETECT_XBITS_TRACK_IPDST: u8 = 1;
+pub const DETECT_XBITS_TRACK_IPPAIR: u8 = 2;
+pub const DETECT_XBITS_TRACK_FLOW: u8 = 3;
+pub const DETECT_XBITS_EXPIRE_DEFAULT: u32 = 30;
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct RSXBitsData {
+    cmd: u8,
+    tracker: u8,
+    expire: u32,
+    vartype: u8,
+    name: *const c_char,
+}
+
+fn get_xbit_type(str_t: &str) -> Result<u8, ()> {
+    let res = match str_t {
+        "ip_src" => DETECT_XBITS_TRACK_IPSRC,
+        "ip_dst" => DETECT_XBITS_TRACK_IPDST,
+        "ip_pair" => DETECT_XBITS_TRACK_IPPAIR,
+        "flow" => DETECT_XBITS_TRACK_FLOW,
+        _ => return Err(()),
+    };
+    Ok(res)
+}
+
+fn get_xbit_cmd(str_c: &str) -> Result<u8, ()> {
+    let res = match str_c {
+        "set" => DETECT_XBITS_CMD_SET,
+        "toggle" => DETECT_XBITS_CMD_TOGGLE,
+        "unset" => DETECT_XBITS_CMD_UNSET,
+        "isnotset" => DETECT_XBITS_CMD_ISNOTSET,
+        "isset" => DETECT_XBITS_CMD_ISSET,
+        "noalert" => DETECT_XBITS_CMD_NOALERT,
+        _ => return Err(()),
+    };
+    Ok(res)
+}
+
+fn evaluate_args(args: Vec<&str>) -> Result<(u8, CString, u8, u32, u8), ()> {
+    let tracker: Vec<&str> = args[2].trim().split(' ').collect();
+    if tracker.len() != 2 {
+        return Err(());
+    }
+    match tracker[0].trim() {
+        "track" => {}
+        _ => {
+            SCLogError!("xbits track keyword parsing failed");
+            return Err(());
+        }
+    }
+    let tracker: u8 = match get_xbit_type(tracker[1].trim()) {
+        Ok(val) => val,
+        Err(e) => {
+            SCLogError!("xbits tracker parsing failed: {:?}", e);
+            return Err(());
+        }
+    };
+    let vartype = if tracker == DETECT_XBITS_TRACK_IPPAIR {
+        11 // VAR_TYPE_IPPAIR_BIT
+    } else {
+        8 // VAR_TYPE_HOST_BIT
+    };
+    let cmd: u8 = match get_xbit_cmd(args[0].trim()) {
+        Ok(val) => val,
+        Err(_) => {
+            SCLogError!("xbits cmd parsing failed");
+            return Err(());
+        }
+    };
+    let name = CString::new(args[1].trim()).unwrap();
+    if args.len() == 3 {
+        return Ok((cmd, name, tracker, DETECT_XBITS_EXPIRE_DEFAULT, vartype));
+    }
+    let expire: Vec<&str> = args[3].trim().split(' ').collect();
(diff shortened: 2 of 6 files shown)


rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+pub const DETECT_XBITS_CMD_UNSET: u8 = 2;
+pub const DETECT_XBITS_CMD_ISNOTSET: u8 = 3;
+pub const DETECT_XBITS_CMD_ISSET: u8 = 4;
+pub const DETECT_XBITS_CMD_NOALERT: u8 = 5;
+pub const DETECT_XBITS_CMD_MAX: u8 = 6;
+pub const DETECT_XBITS_TRACK_IPSRC: u8 = 0;
+pub const DETECT_XBITS_TRACK_IPDST: u8 = 1;
+pub const DETECT_XBITS_TRACK_IPPAIR: u8 = 2;
+pub const DETECT_XBITS_TRACK_FLOW: u8 = 3;
+pub const DETECT_XBITS_EXPIRE_DEFAULT: u32 = 30;
```
inashivb, line 15: Is there a way to de duplicate these? I remember @catenacyber asking for something like that a while ago. @catenacyber Did you manage to?

rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+    if tracker.len() != 2 {
+        return Err(());
+    }
+    match tracker[0].trim() {
+        "track" => {}
+        _ => {
+            SCLogError!("xbits track keyword parsing failed");
+            return Err(());
+        }
+    }
```
inashivb, line 62: Is there a better, idiomatic way of doing this?

rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+    let tracker: u8 = match get_xbit_type(tracker[1].trim()) {
+        Ok(val) => val,
+        Err(e) => {
+            SCLogError!("xbits tracker parsing failed: {:?}", e);
+            return Err(());
+        }
+    };
+    let vartype = if tracker == DETECT_XBITS_TRACK_IPPAIR {
+        11 // VAR_TYPE_IPPAIR_BIT
```
inashivb, line 71: Is there a way to bring C constants to Rust without getter functions? Should we have a generic getter function for bringing constants from C that are not going to be brought here soon?

src/detect-xbits.c
```diff
@@ -305,25 +215,26 @@ static int DetectXbitParse(DetectEngineCtx *de_ctx,
...
-        cd->idx, fb_cmd_str, strlen(fb_name) ? fb_name : "(none)");
+    cd->idx = VarNameStoreSetupAdd(ptr->name, ptr->vartype);
+    cd->cmd = ptr->cmd;
+    cd->tracker = ptr->tracker;
+    cd->type = ptr->vartype;
+    cd->expire = ptr->expire;
 
     *cdout = cd;
+    // ptr's job is done, free it
+    rs_xbits_free(ptr, ptr->name);
```
inashivb, line 237: Should this only be freed once in the `free` function for the keyword?

victorjulien: could the Rust side just completely own `DetectXbitsData` instead of how it is here, where there is a mix?

inashivb: Indeed. I had thought about it but was not parsing and passing all the params back then and it is used at more places than just `detect-xbits`. I think it is possible with the current struct though.

src/detect-xbits.c
```diff
@@ -194,107 +194,17 @@ static int DetectXbitParse(DetectEngineCtx *de_ctx,
...
-        fb_cmd = DETECT_XBITS_CMD_TOGGLE;
-    } else {
-        SCLogError(SC_ERR_UNKNOWN_VALUE, "xbits action \"%s\" is not supported.", fb_cmd_str);
+    if (ptr == NULL) {
         return -1;
     }
-
-    switch (fb_cmd) {
+    switch (ptr->cmd) {
         case DETECT_XBITS_CMD_NOALERT: {
```
inashivb, line 203: Fix this. IT should work.

rust/src/detect/xbits.rs
jasonish: These are already being export to C with `cbindgen`, for example:
```
$ grep DETECT_XBITS_CMD_TOGGLE gen/rust-bindings.h 
#define DETECT_XBITS_CMD_TOGGLE 1
```

We're not seeing a conflic on the C side as there is no code that includes `detect-xbits.h` and `rust-bindings.h` yet.  You could try including `rust.h` to pull these in and remove them from `detect-xbits.h` and it might work out OK.  There are still some header ordering issues where you can end up with headers that are required before rust.h, these can't depend on stuff defined in Rust. This prevents us from moving some constants into Rust like the ALPROTO_ values.

rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+        "isnotset" => DETECT_XBITS_CMD_ISNOTSET,
+        "isset" => DETECT_XBITS_CMD_ISSET,
+        "noalert" => DETECT_XBITS_CMD_NOALERT,
+        _ => return Err(()),
+    };
+    Ok(res)
+}
+
+fn evaluate_args(args: Vec<&str>) -> Result<(u8, CString, u8, u32, u8), ()> {
+    let tracker: Vec<&str> = args[2].trim().split(' ').collect();
```
jasonish, line 52: You can trim and split in one line...
```
args[2].trim().split(' ').map(|s| s.trim()).collect();`
```

rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+    if tracker.len() != 2 {
+        return Err(());
+    }
+    match tracker[0].trim() {
```
jasonish, line 56: `if tracker[0] == "track"` probably makes more sense as its just a single match.

rust/src/detect/xbits.rs
jasonish: Not at this time.  That would require `bindgen` which we don't have integrated at this time.  Best to just duplicate them as constants somewhere.

rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+pub unsafe extern "C" fn rs_xbits_parse(carg: *const c_char) -> *mut c_void {
+    if carg.is_null() {
+        return std::ptr::null_mut();
+    }
+
+    let arg = match CStr::from_ptr(carg).to_str() {
+        Ok(arg) => arg,
+        _ => {
+            return std::ptr::null_mut();
+        }
```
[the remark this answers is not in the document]
jasonish: In fact, this can be tidied up a bit if there is no special error handling to
```
if let Ok(arg) = CStr::from_ptr(carg).to_str() {
        if let Ok(detect) = parse_xbits(arg) {
            return Box::into_raw(Box::new(detect));
        }
    }
    std::ptr::null_mut();
```

rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn test_good_input() {
+        assert_eq!(true, parse_xbits("noalert").is_ok());
+        assert_eq!(true, parse_xbits("set,abc,track ip_pair").is_ok());
+        assert_eq!(
+            true,
+            parse_xbits("set, abc ,track ip_pair, expire 3600").is_ok()
+        );
+        assert_eq!(
+            true,
+            parse_xbits("set  ,abc,track ip_src, expire 1234").is_ok()
```
jasonish, line 221: Should test that the parsed values are as expected.. No leading/trailing white space, etc.

rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+pub const DETECT_XBITS_CMD_MAX: u8 = 6;
+pub const DETECT_XBITS_TRACK_IPSRC: u8 = 0;
+pub const DETECT_XBITS_TRACK_IPDST: u8 = 1;
+pub const DETECT_XBITS_TRACK_IPPAIR: u8 = 2;
+pub const DETECT_XBITS_TRACK_FLOW: u8 = 3;
+pub const DETECT_XBITS_EXPIRE_DEFAULT: u32 = 30;
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct RSXBitsData {
```
jasonish, line 19: I think we should drop the `RS` prefix unless its to avoid some naming conflict.

rust/src/detect/xbits.rs
```diff
@@ -0,0 +1,224 @@
...
+        cmd: res.0,
+        name: (res.1).into_raw(),
+        tracker: res.2,
+        expire: res.3,
+        vartype: res.4,
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_xbits_parse(carg: *const c_char) -> *mut c_void {
```
jasonish, line 151: This could actually return `*const RSXBitData` which will produce a better C binding of this:
```
const struct RSXBitsData *rs_xbits_parse(const char *carg);
```
instead of this
```
void *rs_xbits_parse(const char *carg);
```

I think we have a lot of functions that return c_void when they could return something better though.

suricata-qa:
ERROR:
 ERROR: Invalid Signature config error in tlpr1_asan_cfg QA test

Pipeline 6195

src/detect-xbits.c
```diff
@@ -305,25 +215,26 @@ static int DetectXbitParse(DetectEngineCtx *de_ctx,
...
 
-    cd->idx = VarNameStoreSetupAdd(fb_name, var_type);
-    cd->cmd = fb_cmd;
-    cd->tracker = hb_dir;
-    cd->type = var_type;
-    cd->expire = expire;
-
-    SCLogDebug("idx %" PRIu32 ", cmd %s, name %s",
-        cd->idx, fb_cmd_str, strlen(fb_name) ? fb_name : "(none)");
+    cd->idx = VarNameStoreSetupAdd(ptr->name, ptr->vartype);
```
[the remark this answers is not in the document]
inashivb: @jasonish this question was missed ^^
It uses mutex lock on the C side to get the var. Is it possible to expose this fn somehow to Rust?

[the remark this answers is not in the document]
jasonish: I'm not sure it should be. Parse in Rust, do any setup/registration in C.

doc/userguide/rules/xbits.rst
```diff
@@ -21,9 +19,13 @@ Notes
    ``track ip_dst``, if you want to match on the server response,
    you check it (``isset``) with ``track ip_src``.
 
--  To not alert, use ``noalert;``
+-  To not alert, use ``xbits:noalert;``, it can only be used as a standalone
+   option like ``flowbits``.
```
jasonish, line 23: I wonder if we should allow this.  You can already simply specify `noalert;` as an option.  So I think the following 2 rules are equivalent:
```
alert tcp any any -> any any (xbits:noalert; sid:1; rev:1;)
alert tcp any any -> any any (noalert; sid:1; rev:1;)
```

inashivb: Yeah. This was more to keep consistency with `flowbits` since we have `flowbits:noalert;`..

PR #6988, opened 2022-02-12
mqtt: raise event on parse error

satta:

https://redmine.openinfosecfoundation.org/issues/5018

Describe changes:
- Remove `pub` property from some parser functions.
- Always require `complete` parsing of all messages. This basically addresses any case where crafted internal buffer lengths exceed the total length given in the `remaining_length` header entry. Also see https://github.com/OISF/suricata/pull/6851#discussion_r803741210 for a discussion of this potential issue.
- Raise Suricata app-layer event `parse_error`  in case of  any parse error.

#suricata-verify-pr: 737

### DIFF
--- a/rust/src/mqtt/parser.rs
+++ b/rust/src/mqtt/parser.rs
@@ -92,7 +92,7 @@ pub fn parse_mqtt_string_pair(i: &[u8]) -> IResult<&[u8], (String, String)> {
 // MESSAGE COMPONENTS
 
 #[inline]
-pub fn parse_property(i: &[u8]) -> IResult<&[u8], MQTTProperty> {
+fn parse_property(i: &[u8]) -> IResult<&[u8], MQTTProperty> {
     let (i, identifier) = parse_mqtt_variable_integer(i)?;
     let (i, value) = parse_qualified_property(i, identifier)?;
     Ok((i, value))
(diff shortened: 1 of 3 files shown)


victorjulien (changes requested):
The different styles around the `where` statement caught my eye. I see
```
func prototype where {

func prototype where
{

func prototype
where {

func prototype
where
{
```
Now I should read up on what `where` even does :)

satta:
> The different styles around the `where` statement caught my eye [...]

I think that's likely `rustfmt`'s doing :)

satta:
Ah, just to be sure... do you want me to change the code formatting or is the review still on?

rust/src/mqtt/parser.rs
```diff
@@ -222,99 +222,109 @@ pub fn parse_connect(i: &[u8]) -> IResult<&[u8], MQTTConnectData> {
...
-        MQTTPublishData {
-            topic,
-            message_id,
-            message: message.to_vec(),
-            properties,
-        },
-    ))
+    protocol_version: u8, has_id: bool,
+) -> impl Fn(&[u8]) -> IResult<&[u8], MQTTPublishData>
+where {
```
jasonish, line 246: Is `where` even needed here?  Seems it evaluates to an empty set of constraints.

satta: I followed the nom documentation that suggests this syntax to return a closure receiving only a `u8` slice, given a number of parameters. See https://github.com/Geal/nom/blob/main/doc/upgrading_to_nom_5.md#from-macros-to-functions.
I am not familiar with how `where` would otherwise be used but in combination with `move` it does what it needs to do here.

satta:
New PR: #7009

PR #6991, opened 2022-02-12
detect/ip_proto: Use built-in protcol mapping table.

jlucovsky:
This PR modifies the detect keyword `ip_proto` to use the built-in protocol table instead of using the underlying system's.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5072](https://redmine.openinfosecfoundation.org/issues/5072)

Describe changes:
- Add utility function to validate protocol name

suricata-verify-pr: 738
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-ipproto.c
+++ b/src/detect-ipproto.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2020 Open Information Security Foundation
+/* Copyright (C) 2007-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -38,6 +38,7 @@
 #include "detect-engine-address.h"
 
 #include "util-byte.h"
+#include "util-proto-name.h"
 #include "util-unittest.h"
 #include "util-unittest-helper.h"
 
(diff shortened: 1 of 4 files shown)


src/detect-ipproto.c
```diff
@@ -121,20 +122,23 @@ static DetectIPProtoData *DetectIPProtoParse(const char *optstr)
 
     /* Protocol name/number */
     if (!isdigit((unsigned char)*(args[1]))) {
-        struct protoent *pent = getprotobyname(args[1]);
-        if (pent == NULL) {
-            SCLogError(SC_ERR_INVALID_VALUE, "Malformed protocol name: %s",
-                       str_ptr);
+        int proto = SCGetProtoByName(args[1]);
+        if (-1 == proto) {
+            SCLogError(SC_ERR_INVALID_VALUE, "Malformed protocol name: \"%s\"", str_ptr);
```
victorjulien, line 127: Is 'Malformed' the correct term here? Could it also just be unknown

src/util-proto-name.c
```diff
@@ -196,3 +196,22 @@ bool SCProtoNameValid(uint16_t proto)
...
+ *
+ * \param proto Protocol name.
+ * \retval ret On success returns the protocol number; else -1
+ */
+int SCGetProtoByName(char *protoname)
+{
+    if (protoname) {
+        for (int i = 0; i <= 255; i++) {
+            if (known_proto[i] &&
+                    0 == strncasecmp(known_proto[i], protoname, strlen(known_proto[i]))) {
```
victorjulien, line 211: wondering if we should access any case mix here... e.g. tCp would work, and TcP... seems we should maybe stay close to what it was when using the `/etc/protocols`

src/detect-ipproto.c
```diff
@@ -121,20 +122,23 @@ static DetectIPProtoData *DetectIPProtoParse(const char *optstr)
...
-        data->proto = (uint8_t)pent->p_proto;
+        data->proto = (uint8_t)proto;
     }
     else {
         if (StringParseUint8(&data->proto, 10, 0, args[1]) <= 0) {
             SCLogError(SC_ERR_INVALID_VALUE, "Malformed protocol number: %s",
                        str_ptr);
             goto error;
         }
+        if (!SCProtoNameValid((uint16_t)data->proto)) {
```
victorjulien, line 138: will this not limit our support for pretty much no reason?
`ip_proto:144` would be rejected as bad input even though it could just work?

victorjulien (changes requested):

src/util-proto-name.c
jlucovsky: Ok ... I'll change this to not ignore case -- which is get `getprotobyname` was oing.

src/detect-ipproto.c
jlucovsky: Will update.

src/detect-ipproto.c
jlucovsky: Good point. Will update.

suricata-qa:

Pipeline 6238

jlucovsky:
Continued in #7021

PR #6992, opened 2022-02-12
nfs: Implement frames v4

ghostinthehive:
Previous PR #6976
Changes from Previous PR:
- Cleanups
- Add check over `nfs4.ops` frame:
  for both Fuzzing CI check errors:
```
thread '<unnamed>' panicked at 'range start index 8 out of range for slice of length 6', src/nfs/nfs.rs:487:67
thread '<unnamed>' panicked at 'range start index 8 out of range for slice of length 4', src/nfs/nfs.rs:487:67
```
although there is no `NFS4` records w/ a `header length < 8` that I found from the `nfs4` pcaps, add the check anyways.

Frames:
	- rpc.pdu
	- rpc.hdr
	- rpc.data
	- rpc.creds -- for rpc calls

  - NFSv2, NFSv3
	- nfs.pdu

  - NFSv4 Only Frames
	- nfs4.pdu
	- nfs4.hdr

RPC tcp/udp frames created with separate registration functions e.g:
`add_rpc_tcp_tc_frames()`
`add_rpc_udp_tc_frames()`

suricata-verify-pr: 739

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -26,6 +26,7 @@ use nom7::{Err, Needed};
 
 use crate::applayer;
 use crate::applayer::*;
+use crate::frames::*;
 use crate::core::*;
 use crate::conf::*;
 use crate::filetracker::*;
@@ -84,6 +85,22 @@ static mut ALPROTO_NFS: AppProto = ALPROTO_UNKNOWN;
  * NFSTransaction where they can be looked up per handle as part of the
  * Transaction lookup.
  */
+
+#[derive(AppLayerFrameType)]
+pub enum NFSFrameType {
+    RPCPdu,
+    RPCHdr,
+    RPCData,
+    RPCCreds, // for rpc calls
+
+    NFSPdu,
+    NFSStatus,
+
+    NFS4Pdu,
+    NFS4Hdr,
+    NFS4Ops,
+    NFS4Status,
+}
 
 #[repr(u32)]
 pub enum NFSEvent {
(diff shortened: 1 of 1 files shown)


victorjulien:
Please specify the suricata verify pr without a `#` on the line.

victorjulien:
I would like to see a update to the logic around the `pdu` frames. They are currently created only when the data is successfully parsed. For the `pdu` frames I'd like these to be created regardless of parser status (so effectively before the parser is called), then the other frames will be created only if the parser succeeded.

rust/src/nfs/nfs.rs
```diff
@@ -85,6 +86,22 @@ static mut ALPROTO_NFS: AppProto = ALPROTO_UNKNOWN;
...
  */
 
+#[derive(AppLayerFrameType)]
+pub enum NFSFrameType {
+    RPCPdu,
+    RPCHdr,
+    RPCData,
+    RPCCreds, // for rpc calls
+
+    NFSPdu,
```
victorjulien, line 96: should we also split out nfs2 and nfs3 here? I think splitting out nfs4 makes total sense, the format is very different. How do you see this for 2 vs 3?

ghostinthehive: As I went about this I had them separated at first, but then I found that they are mostly the same, some records --a good deal of them-- are even identical. As for the format, they don't have a distinguished format, there is no wrapper field that could be extracted out as a frame as in the `nfs4.ops` frame that holds all `ops` data, so for `NFS2 | NFS3` its just a blob of data --pdu-- with fields according to the records, and a `Status` for the responses.

The way I see it: If its important from the perspective of having versions show up in the frames wrt rule writing and the like, then yes we could separate it, other than that a generic `nfs` would do the work and reduce repetitiveness.

victorjulien: think this makes sense. It's a lot like how in the ssl/tls parser we treat ssl3/tls1+ all the same, while doing something separate for ssl2. Thanks.

ghostinthehive:
Followed by  #7043

PR #6980, opened 2022-02-11
Keyword/smb cmd

zer1t0:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: 
https://redmine.openinfosecfoundation.org/issues/5069

Describe changes:
- New keyword smb.cmd added, which allows to match the command value in smb message header. It works for smb1 and smb2.
- Documentation for the keyword added (created SMB keyword documentation)

Example of rule `alert smb any any -> any any (msg: "Smb command rule"; smb.cmd: 10; sid: 1;)` . More examples in documentation.

suricata-verify-pr: 733

PR #6998, opened 2022-02-14
smb: New keyword smb.version

zer1t0:

https://redmine.openinfosecfoundation.org/issues/5075

Describe changes:
- New keyword smb.version added, which allows to match the packets of specif version of smb, that can be 1 or 2.
- Documentation for the keyword added (created SMB keyword documentation)

Example of rule `alert smb any any -> any any (msg: "SMBv2 message"; smb.version: 2; sid: 1;)` . More examples in documentation.

suricata-verify-pr: 740


zer1t0:
> Do we want more flavors like SMB version 3 to be recognized here ? When there is version negotiation, what is the behavior ? (I think there is a first V1 message to negotiate V2) Is there something like V2.1 by the way ?

Hi, currently there are two versions of SMB,1 and 2, show in the header. The SMB2 protocol allows to use different dialects that can be 2.0.2, 2.1, 3.0, 3.0.2, 3.1.1, so we can talk about SMB3, which is using SMB2 with a dialect 3.x.

But for detecting the dialects, and then recognize SMB3, we would need a different keyword.

catenacyber:
> so we can talk about SMB3, which is using SMB2 with a dialect 3.x.

Ok, this is indeed confusing.

What is the logic hidden here ?
SMB0xff is V1 and SMB0xfe, SMB0xfd, SMB0xfc is V2 ?

Is this the same difference as for the Wireshark filter `smb`and `smb2` ?

I see here 
https://www.microsoft.com/security/blog/2020/05/04/mitigating-vulnerabilities-endpoint-network-stacks/
that SMB0xfc is indeed referenced as SMBv2

zer1t0:
To sum up the header codes:
- SMB0xff is SMB1 [header](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb/3c0848a6-efe9-47c2-b57a-f7e8217150b9)
- SMB0xfe is SMB2 [normal header](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/5cd64522-60b3-4f3e-a157-fe66f1228052) (can be [sync](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/fb188936-5050-48d3-b350-dc43059638a4) or [async](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/ea4560b7-90da-4803-82b5-344754b92a79))
- SMB0xfd is SMB2 [transform header](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/d6ce2327-a4c9-4793-be66-7b5bad2175fa). This is only valid for the SMB 3.x dialect family.
- SMB0xfc is SMB2 [transform compression header](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/1d435f21-9a21-4f4c-828e-624a176cf2a0) (can be [chained](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/aa880fe8-ebed-4409-a474-ec6e0ca0dbcb)  or [unchained](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/793db6bb-25b4-4469-be49-a8d7045ba3a6)). These ones requires the use of 3.1.1 dialect.

Yes, Wireshark only distinguishes between SMB1 and SMB2, since they are totally different protocols with total different message formats, types and implementation.

On the other hand SMB3 is more an extension for SMB2. When using SMB2 we can select one of the following dialects for the conversation between client and server:
- 2.0.2
-  2.1
-  3.0
-  3.0.2
-  3.1.1

We say we are using SMB3 when we select a 3.x dialect for the conversation, so you can use SMB3.0, SMB3.0.2 or SMB3.1.1. The higher you choose, the more capabilities you have, but the message syntax and message command number remains the same.

In this case, the aim of this keyword is to distinguish between SMB1 and SMB2 since they have totally different command codes for the messages (so you can use this keyword with [smb.cmd](https://github.com/OISF/suricata/pull/6980)).

For example, you can use the following rule to match the command with code 0x1:
`alert smb any any -> any any (msg: "Smb command rule"; smb.cmd: 1; sid: 1;)`

But this rule will match the command session_setup in SMB2 (and SMB3) and the command delete directory in SMB1. Therefore the intention here is to use the smb.cmd to distinguish between the SMB1 and SMB2, so you can match the exact message you are looking for (and avoid matching the other protocol). Since SMB2 and SMB3 use the same messages, we don't really care about the dialect (neither suricata or wireshark do).

Currently, Suricata only recognizes version 1 or 2, as you can see in the [SmbVerCmdStat](https://github.com/OISF/suricata/blob/b1c09369af40bb2b65825d25094d37b3515da1bb/rust/src/smb/smb.rs#L219) struct.

In case we want to only match a conversation using SMB3, we will need to being aware of the dialect established and the beginning of the conversation between client and server, but that is a different operation that matching the cmd field in the header.

catenacyber:
Thanks.

What about version negotiation ? (I think there is a first V1 message to negotiate V2)

zer1t0:
> What about version negotiation ? (I think there is a first V1 message to negotiate V2)

What do you mean exactly?

In SMB the first message sent by the client is always a SMB1 negotiation protocol request, and the client can respond with a SMB1 or SMB2 protocol response, to set the desired SMB version.

catenacyber:
Thanks for the explanation.
I guess I mean to document that the SMB1 negotiation protocol request is treated as SMB1 by this keyword even if it proposes SMB2

zer1t0:
>

zer1t0:
Moved to https://github.com/OISF/suricata/pull/7147

PR #6943, opened 2022-02-07
Packet alerts/v6

jufajardini:
Previous PR: #6931

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4207

Describe changes:
- decode: replace `BUG_ON` check with config value checks and warning, falling back to defaults
- fix typo

suricata-verify-pr: 694

PR #6999, opened 2022-02-14
Packet alerts/v7

jufajardini:
https://github.com/OISF/suricata/pull/6943 addressing comments done on #6896 

Describe changes:
- remove redundant check alerts.cnt > 0 (type is unsigned)
- add `BUG_ON` to check if we're not calling `memmove` with size 0
- use last in queue logic for alerts for same sid (assuming it is not needed to check if tx_id differs)

suricata-verify-pr: 694

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2019 Open Information Security Foundation
+/* Copyright (C) 2007-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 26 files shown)


src/decode.c
```diff
@@ -781,6 +801,22 @@ void DecodeGlobalConfig(void)
...
         }
     }
+    GetPacketMaxAlertConfig();
+}
+
+void GetPacketMaxAlertConfig(void)
+{
+    intmax_t max = 0;
+    if (ConfGetInt("packet-alert-max", &max) == 1) {
+        if (max < 0 || max > UINT8_MAX) {
```
inashivb, line 811: Do we allow a setting of 0, then? What happens in a case like that?

src/detect-engine-alert.c
```diff
@@ -189,8 +189,10 @@ int PacketAlertAppend(DetectEngineThreadCtx *det_ctx, const Signature *s,
     SCLogDebug("sid %"PRIu32"", s->id);
 
     /* It should be usually the last, so check it before iterating */
-    if (p->alerts.cnt == 0 || p->alerts.alerts[p->alerts.cnt - 1].num < s->num) {
+    /* Same signatures can generate more than one alert, if it's a diff tx */
+    if (p->alerts.cnt == 0 || p->alerts.alerts[p->alerts.cnt - 1].num <= s->num) {
```
inashivb, line 193: Q: how do we know here if the alerts came from diff txs?

src/decode.c
inashivb, line 811: Q: What happens in a case when max == 0?

jufajardini: I think we decided that if the user chose that, that was up to them... 🤔

src/detect-engine-alert.c
jufajardini: Since this wasn't checked before, I assumed it was not needed...

src/decode.c
jufajardini: change behavior to Warning and default

jufajardini:
Followed by: https://github.com/OISF/suricata/pull/7000

PR #7005, opened 2022-02-15
Fuzz maxtx v2

catenacyber:

Describe changes:
- fuzz : finds if a flow has too many open/live transactions
- nfs : prevents the creation of too many transaction
 
Comes from #6863 and CI should already catch missing protocol

Replaces #6923 with removing unused `events` variable


victorjulien (approved):

victorjulien:
Merged in #7013, thanks!

PR #6987, opened 2022-02-12
Iponly cidr issues/v15

victorjulien:
#6986, fixing IPv6 and addressing Jasons comment on a moving comment :)

suricata-verify-pr: 735

PR #7008, opened 2022-02-15
Iponly cidr issues/v17

victorjulien:
#6987 with additional fixes. In issue 5081 we track the problem of addresses with CIDRs not getting handled correctly if the address wasn't the address range start.

https://redmine.openinfosecfoundation.org/issues/5081
https://redmine.openinfosecfoundation.org/issues/5066

suricata-verify-pr: 735

### DIFF
--- a/src/detect-engine-address.c
+++ b/src/detect-engine-address.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2021 Open Information Security Foundation
+/* Copyright (C) 2007-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -388,7 +388,6 @@ bool DetectAddressListsAreEqual(DetectAddress *list1, DetectAddress *list2)
 }
 
 /**
- * \internal
  * \brief Creates a cidr ipv6 netblock, based on the cidr netblock value.
  *
  *        For example if we send a cidr of 7 as argument, an ipv6 address
@@ -404,7 +403,7 @@ bool DetectAddressListsAreEqual(DetectAddress *list1, DetectAddress *list2)
  * \param in6  Pointer to an ipv6 address structure(struct in6_addr) which will
  *             hold the cidr netblock result.
  */
-static void DetectAddressParseIPv6CIDR(int cidr, struct in6_addr *in6)
+void DetectAddressParseIPv6CIDR(int cidr, struct in6_addr *in6)
 {
     int i = 0;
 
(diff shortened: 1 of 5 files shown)


inashivb (commented):
**For Bug 5081**:
I tested parsing of IPv4 addresses and the address ranges returned in the `DetectParse` instance. They were as per the specified subnet mask and the tests passed for:
- masks 16-32

The tests rightly failed for:
- invalid masks
- invalid IP addresses
- invalid range checks

**For Bug 5066**:
I tested matches for IPv4 addresses with diff subnet masks (only in the range of 16-32).
Following are the findings:
- If a newly added IP address + mask was a subset of the root node, it would be a "best match".
e.g.
```
/** Bug #5066              
+  *                         
+  *  insert:                
+  *      - 100.117.0.0/16:  100.117.0.0  - 100.117.255.255
+  *                         
+  *  check:                 
+  *      - 100.117.54.0/28: 100.117.54.0 - 100.117.54.15  (subset)  -- Best match
+  *      - 100.117.68.64/20: 100.117.64.0 - 100.117.79.255  (subset)  -- Best match
+  */ 
```
- If the root node itself was tested for an exact match after a `SCRadixFindKeyIPV4Netblock`, it would fail. **Iff** the below test is right:
```c
     struct sockaddr_in servaddr;                            
     SCRadixNode *node = NULL;                               
                                                             
     SCLogDebug("setup tree");                               
     SCRadixTree *tree = SCRadixCreateRadixTree(free, NULL); 
                                                             
     memset(&servaddr, 0, sizeof(servaddr));                 
     FAIL_IF(inet_pton(AF_INET, "100.117.0.0", &servaddr.sin_addr) <= 0);
     SCLogDebug("add 100.117.0.0/16");                       
     SCRadixAddKeyIPV4Netblock((uint8_t *)&servaddr.sin_addr, tree, strdup("1"), 16);
     SCRadixPrintTree(tree);                                 
     SCLogDebug("find w SCRadixFindKeyIPV4Netblock 100.117.0.0/16");
     char *r = NULL;                                         
     node = SCRadixFindKeyIPV4Netblock((uint8_t *)&servaddr.sin_addr, tree, 16, (void **)&r);
     FAIL_IF_NULL(node);                                     
     SCRadixPrintNodeInfo(node, 0, NULL);                    
                                                             
     SCLogDebug("find w SCRadixFindKeyIPV4ExactMatch 100.117.0.0/16");
     node = SCRadixFindKeyIPV4ExactMatch((uint8_t *)&servaddr.sin_addr, tree, NULL);  // FAILS
     FAIL_IF_NULL(node);                                     
     SCRadixPrintNodeInfo(node, 0, NULL);      
     SCRadixPrintTree(tree);                                 
     SCRadixReleaseRadixTree(tree);                          
     PASS;  
```
- Duplicate address/masks are **reported** at the time of adding keys to the tree e.g.
```
root: 100.117.0.0/16
node: 100.117.45.64/16  -- Duplicate
```
- Disjointed sets are **correctly** not found in the tree. e.g.
```
+  *  insert:                
+  *      - 100.117.0.0/16:  100.117.0.0  - 100.117.255.255
+  *      - 100.116.0.5  (diff set)  -- No match
```
Edit:
- One combination seemed to FAIL on a test for same subnet **iff** the following test is correct
```C
     struct sockaddr_in servaddr;
     SCRadixNode *node = NULL;
                    
     SCLogDebug("setup tree");
     SCRadixTree *tree = SCRadixCreateRadixTree(free, NULL);
                    
     memset(&servaddr, 0, sizeof(servaddr));
     FAIL_IF(inet_pton(AF_INET, "1.2.3.0", &servaddr.sin_addr) <= 0);
     SCLogDebug("add 1.2.3.0/24");
     SCRadixAddKeyIPV4Netblock((uint8_t *)&servaddr.sin_addr, tree, strdup("1"), 24);
     SCRadixPrintTree(tree);
     SCLogDebug("find 1.2.3.0/24");
     char *r = NULL;
     node = SCRadixFindKeyIPV4Netblock((uint8_t *)&servaddr.sin_addr, tree, 24, (void **)&r);
     FAIL_IF_NULL(node);
     SCRadixPrintNodeInfo(node, 0, NULL);
                    
     SCLogDebug("add 1.2.3.0/18");
     FAIL_IF(inet_pton(AF_INET, "1.2.3.0", &servaddr.sin_addr) <= 0);
     SCRadixAddKeyIPV4Netblock((uint8_t *)&servaddr.sin_addr, tree, strdup("2"), 18);
     SCRadixPrintTree(tree);
     SCLogDebug("find 1.2.3.0/18");
     node = SCRadixFindKeyIPV4Netblock((uint8_t *)&servaddr.sin_addr, tree, 18, NULL);
     FAIL_IF_NULL(node);
     SCRadixPrintNodeInfo(node, 0, NULL);
                    
     SCLogDebug("find 1.2.23.5/18");
     FAIL_IF(inet_pton(AF_INET, "1.2.23.5", &servaddr.sin_addr) <= 0);
     node = SCRadixFindKeyIPV4Netblock((uint8_t *)&servaddr.sin_addr, tree, 18, NULL); // FAILS
     FAIL_IF_NULL(node);
     SCRadixPrintNodeInfo(node, 0, NULL);
                   
     SCRadixPrintTree(tree);
                   
     SCRadixReleaseRadixTree(tree);
     PASS;         
```

victorjulien:
The unittest looks like it should fail at:
```c
     node = SCRadixFindKeyIPV4ExactMatch((uint8_t *)&servaddr.sin_addr, tree, NULL);  // FAILS
     FAIL_IF_NULL(node); 
```
as the exact match treats the address as a single IP (so a `/32`) and the range in the tree isn't that, its a `/16`. Best match for the `/32` case should return the `/16`.

inashivb:
> The unittest looks like it should fail at:
> 
> ```c
>      node = SCRadixFindKeyIPV4ExactMatch((uint8_t *)&servaddr.sin_addr, tree, NULL);  // FAILS
>      FAIL_IF_NULL(node); 
> ```
> 
> as the exact match treats the address as a single IP (so a `/32`) and the range in the tree isn't that, its a `/16`. Best match for the `/32` case should return the `/16`.

Thank you for explaining! I have added one more case I had forgotten adding based on the diff IP/mask combinations you asked to test.. Could you please check that too?

victorjulien:
I think you've rediscovered bug 5081. The radix api will let you add by any key, but for it to make sense the key should be the start of the address range. For both `1.2.23.5/18` and `1.2.3.0/18` this should be `1.2.0.0`. So the test adds the wrong keys.

I suppose a future improvement would be to either block this kind of input to the radix api, or to just apply the cidr, so silently fixing it.

victorjulien:
> I think you've rediscovered bug 5081. The radix api will let you add by any key, but for it to make sense the key should be the start of the address range. For both `1.2.23.5/18` and `1.2.3.0/18` this should be `1.2.0.0`. So the test adds the wrong keys.
> 
> I suppose a future improvement would be to either block this kind of input to the radix api, or to just apply the cidr, so silently fixing it.

I'm looking at this and seeing more issues actually, where some parts of the code can feed into the radix api by string. There the key can still be wrong.

victorjulien:
Continued in #7017

PR #7015, opened 2022-02-17
detect: only apply ConfigApplyTx with app-layers

catenacyber:

Describe changes:
- detect: only apply ConfigApplyTx with app-layers

> If the scope is flow it should require a flow to be present.

How do we do that ?
I do not see `SIG_FLAG_FLOW`


victorjulien:
> I think the `SIG_MASK_REQUIRE_FLOW` can be of use here. You may need to dig around a bit to see how you can trigger it getting set.

Or how it is set in other places for inspiration.

catenacyber:
Thanks Victor, done in https://github.com/OISF/suricata/pull/7063

PR #7023, opened 2022-02-18
(draft) smb: flags fix; what to do when messages are in the wrong direction?

jasonish:
- smb: fix smb2 header flag parsing
- wip: what to do if smb in wrong direction

### DIFF
--- a/rust/src/smb/events.rs
+++ b/rust/src/smb/events.rs
@@ -27,6 +27,10 @@ pub enum SMBEvent {
     DuplicateNegotiate,
     NegotiateMalformedDialects,
     FileOverlap,
+    /// A request was seen in the to client direction.
+    RequestToClient,
+    /// A response was seen in the to server direction,
+    ResponseToServer,
 }
 
 impl SMBTransaction {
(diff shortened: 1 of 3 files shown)


victorjulien:
I suppose there are several things to consider: 

1. we can't just enter an error state on such a message: bad client or server could send one or more of such records to evade us then
2. exception to (1) is probably if both do it AND all messages are in the wrong dir?

Wrt normally processing and logging the wrong dir messages, I think I would like to avoid this as might confuse the various state tracking logic. E.g. what if the client sends a request and then answers it itself? SMB parser has "skip" logic that we should probably set on a per record basis to efficiently skip those records

jasonish:
So one idea is if we see this state we set a `reversed` flag, then process everything in reverse. Not knowing what I don't know, I don't know how invasive this would be.  But it just means we treat `toserver` messages as responses, and `toclient` messages as request.

Then if we see a conflict in that we raise an alert. Then I'm not sure what to do. As seen, if we continue excessive buffering can occur, so again, without knowing SMB that well, maybe alert and fail out is the best we can/should do.  We'll be blind to the reset of the traffic, but we alerted on the anomaly, so its not an evasion.

jasonish:
> * we can't just enter an error state on such a message: bad client or server could send one or more of such records to evade us then

Is it an evasion if we raise an alert and then error out?

victorjulien:
> > * we can't just enter an error state on such a message: bad client or server could send one or more of such records to evade us then
> 
> Is it an evasion if we raise an alert and then error out?

Technically, I think "no". But I do feel like putting an entire stream into an error state based on one parseable but "wrong" record may be a bit harsh... Looking at something similar in NFS currently.

I'd say for 5 and 6 we can take this strict approach, but for master/7 I'd like to think a bit more on how to handle such cases.

jasonish:
Another stab at better handling for now. I'm thinking this logic could be backported to 5 and 6.

PR #7026, opened 2022-02-18
configure: use classification.config gen by suricata-update

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4136

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1431,24 +1431,24 @@
     if test "$enable_suricata_update" = "yes"; then
         if test -f "$srcdir/suricata-update/setup.py"; then
           have_suricata_update="yes"
-	fi
+        fi
     fi
 
     if test "$have_suricata_update" = "yes"; then
         if test "$have_python_yaml" != "yes"; then
-	    echo ""
-	    echo "    Warning: suricata-update will not be installed as the"
-	    echo "        Python yaml module is not installed.."
-	    echo ""
+            echo ""
+            echo "    Warning: suricata-update will not be installed as the"
+            echo "        Python yaml module is not installed.."
+            echo ""
             echo "    Install the yaml module for Python ${pymv} to enable"
             echo "    suricata-update."
-	    echo
-	else
+            echo
+        else
             SURICATA_UPDATE_DIR="suricata-update"
             AC_SUBST(SURICATA_UPDATE_DIR)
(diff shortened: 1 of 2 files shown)


inashivb:
@jasonish don't know if this is blatantly wrong but thought of following https://github.com/inashivb/suricata/commit/9d7a497cc168cf3917c1fd7b68daf0fb27e22722 

Q: Should `classification.config` be allowed to load from places other than rule dir if we're enforcing rule dir?

jasonish:
My `suricata.yaml` in my dev tree now has this:
```
classification-file: /opt/suricata/7.0.0-dev/var/lib/suricata/rulesclassification.config
```
(missing slash)

jasonish (changes requested):
Path issue.

inashivb:
> My `suricata.yaml` in my dev tree now has this:
> 
> ```
> classification-file: /opt/suricata/7.0.0-dev/var/lib/suricata/rulesclassification.config
> ```
> 
> (missing slash)

Clearly didn't test it. Sorry about this. :disappointed:

norg:
I overall agree with @jasonish as long as we ensure it's documented in a way that it's easy to spot.

The only thing that I don't like about the result is the fact, that we have the config part with all 3 (classification, reference and threshold) stuck together and a user might be asking why is just one in a different location.

One might argue that threshold is commented by default.

From my experience it is not as much in use as it would be a big problem to change that part. But config files related to rules should stick together in a way?

jasonish:
> I overall agree with @jasonish as long as we ensure it's documented in a way that it's easy to spot.
> 
> The only thing that I don't like about the result is the fact, that we have the config part with all 3 (classification, reference and threshold) stuck together and a user might be asking why is just one in a different location.
> 
> One might argue that threshold is commented by default.
> 
> From my experience it is not as much in use as it would be a big problem to change that part. But config files related to rules should stick together in a way?

The way I see it, reference.config and classification.config are usually handled by us (Suricata) or rule writers. So don't really need to be in a user configuration directory.  But threshold.config is designed to be a file that is edited by the end-user, and not distributed by rule vendors or us (other than an empty stub).  

But yeah, documentation is pretty key here.

victorjulien:
Agree with Jasons suggestions.

PR #7030, opened 2022-02-18
build-info: add info about fuzztargets - v1

jufajardini:
We were missing that information from the Development information. It is something useful to know, so it makes sense that we have it there, alongside with other configure enabled options...

none

Describe changes:
- print if fuzztargets are enable, with the command `build-info`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2592,6 +2592,8 @@ Development settings:
   Unit tests enabled:                      ${enable_unittests}
   Debug output enabled:                    ${enable_debug}
   Debug validation enabled:                ${enable_debug_validation}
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -2592,6 +2592,8 @@ Development settings:
   Unit tests enabled:                      ${enable_unittests}
   Debug output enabled:                    ${enable_debug}
   Debug validation enabled:                ${enable_debug_validation}
+  Fuzz targets enabled:                    ${enable_fuzztargets}
+
```
jasonish, line 2596: nit: this adds an extra line between these 2 sections.

jufajardini:
replaced by: https://github.com/OISF/suricata/pull/7032

PR #7027, opened 2022-02-18
Next/20220218/v1

victorjulien:
Staging:
- #7020

Fixing:
- https://redmine.openinfosecfoundation.org/issues/5094

PR #7033, opened 2022-02-18
Next/20220218/v4

victorjulien:
#7027 with:
- format fixup
- nfs fixes

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -589,15 +589,21 @@ impl NFSState {
     }
 
     pub fn process_write_record<'b>(&mut self, r: &RpcPacket<'b>, w: &Nfs3RequestWrite<'b>) -> u32 {
-        // for now assume that stable FILE_SYNC flags means a single chunk
-        let is_last = if w.stable == 2 { true } else { false };
-
         let mut fill_bytes = 0;
-        let pad = w.file_len % 4;
+        let pad = w.count % 4;
         if pad != 0 {
             fill_bytes = 4 - pad;
         }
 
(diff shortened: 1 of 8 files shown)


rust/src/nfs/nfs.rs
```diff
@@ -589,15 +589,21 @@ impl NFSState {
     }
 
     pub fn process_write_record<'b>(&mut self, r: &RpcPacket<'b>, w: &Nfs3RequestWrite<'b>) -> u32 {
-        // for now assume that stable FILE_SYNC flags means a single chunk
-        let is_last = if w.stable == 2 { true } else { false };
-
         let mut fill_bytes = 0;
-        let pad = w.file_len % 4;
+        let pad = w.count % 4;
```
inashivb, line 593: I was thinking since we now already parse the padding bytes, should we use those directly instead of calculating them?

rust/src/nfs/nfs.rs
```diff
@@ -589,15 +589,21 @@ impl NFSState {
...
         if pad != 0 {
             fill_bytes = 4 - pad;
         }
 
+        if w.count == 0 || w.count < w.file_len ||
+            ((fill_bytes + w.count) as usize) < w.file_data.len() ||
```
inashivb, line 599: Since at this stage we have already established that `count >= file_len`, and in the next check, we have `(fill_bytes + w.file_len) >= w.file_data.len()`, would this be an extra check?

inashivb (commented):
Still a bit confused about relying on `count`

victorjulien:
Needs more work on nfs

PR #7035, opened 2022-02-18
rust/frames: derive direction from StreamSlice - v1

jasonish:
On the Rust side, a Frame requires a StreamSlice to be created. We can
derive the direction and retain the direction from the StreamSlice
removing the need for callers to provide the direction when operating on
the frame.

Am I missing a reason why this might not be a good idea or work in all cases?

### DIFF
--- a/rust/src/frames.rs
+++ b/rust/src/frames.rs
@@ -17,6 +17,7 @@
 
 use crate::applayer::StreamSlice;
 use crate::core::Flow;
+use crate::core::STREAM_TOSERVER;
 
 #[repr(C)]
 struct CFrame {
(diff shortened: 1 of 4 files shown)


suricata-qa:
ERROR:
 

ERROR: QA failed on tlpw2_single_alerts_cmp.

ERROR: QA failed on tlpw2_autofp_alerts_cmp.

Pipeline 6330

rust/src/frames.rs
```diff
@@ -37,19 +38,22 @@ extern {
...
 }
 
 impl Frame {
     pub fn new(
         flow: *const Flow, stream_slice: &StreamSlice, frame_start: &[u8], frame_len: i64,
-        dir: i32, frame_type: u8,
+        frame_type: u8,
     ) -> Option<Self> {
+        // Derive the direction from the stream slice, 0 for to server, 1 for to client.
+        let direction = if stream_slice.flags() & STREAM_TOSERVER != 0 { 0 } else { 1 };
```
victorjulien, line 56: should we use `core::Direction` here?

victorjulien (commented):
I like this. Minor question inline

suricata-qa:

Pipeline 6479

catenacyber (changes requested):
Needs rebase

suricata-qa:

Pipeline 6479

rust/src/frames.rs
jasonish: No, it makes the code a little more complex.  `Direction::ToServer` and `Direction::ToClient` have values of `0x04` and `0x08`, but we immeidately use this direction in `AppLayerFrameNewByRelativeOffset` which is a C function which uses values `0` and `1` for direction.

jasonish:
Rebased at https://github.com/OISF/suricata/pull/7035.

PR #7025, opened 2022-02-18
detect/ip_proto: Use built-in protcol mapping table.

jlucovsky:
Continuation of #7021 

This PR modifies the detect keyword `ip_proto` to use the built-in protocol table instead of using the underlying system's.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5072](https://redmine.openinfosecfoundation.org/issues/5072)

Describe changes:
- Add utility function to validate protocol name

Updates
- Error checking of `strdup` return values.

suricata-verify-pr: 738
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #7040, opened 2022-02-20
detect/ip_proto: Use built-in protcol mapping table. 

jlucovsky:
Continuation of #7025

This PR modifies the detect keyword `ip_proto` to use the built-in protocol table instead of using the underlying system's.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5072](https://redmine.openinfosecfoundation.org/issues/5072)

Describe changes:
- Add utility function to validate protocol name

Updates
- Remove error goto in initialization step

suricata-verify-pr: 738
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/util-proto-name.c
+++ b/src/util-proto-name.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2021 Open Information Security Foundation
+/* Copyright (C) 2007-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -26,6 +26,10 @@
 
 #include "suricata-common.h"
 #include "util-proto-name.h"
+
+#ifdef UNITTESTS
+#include "util-unittest.h"
+#endif
 
 /** Lookup array to hold the information related to known protocol
  *  values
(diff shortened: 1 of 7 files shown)


catenacyber:
I wonder if there are other such dependencies.
I remember fixing a similar problem (different values for different OSes) about layer types over NULL/loopback interface...

src/util-proto-name.c
```diff
@@ -184,6 +188,264 @@ const char *known_proto[256] = {
...
+    /* no aliases for 142-255 */
+};
+
+typedef struct ProtoNameHashEntry_ {
+    const char *name;
+    const char *alias;
+    uint8_t number;
+} ProtoNameHashEntry;
+
+HashTable *proto_ht;
```
victorjulien, line 350: I generally make such globals `static` and initialize them to `NULL`. Should we do that here too?

src/util-proto-name.c
```diff
@@ -184,6 +188,264 @@ const char *known_proto[256] = {
...
+
+static uint32_t ProtoNameHashFunc(HashTable *ht, void *data, uint16_t datalen)
+{
+    ProtoNameHashEntry *p = (ProtoNameHashEntry *)data;
+    uint32_t hash = 0;
+    int len = strlen(p->name);
```
victorjulien, line 356: also wondering why we're not using the length fields of the api here

jlucovsky:
I've got to rework this PR to hash the aliases and the names and not combine the two into a single entry. E.g., validating "ip" fails because it's an alias of "HPOPT".

Sigh.

src/util-proto-name.c
jlucovsky: Yes -- will update.

src/util-proto-name.c
```diff
@@ -184,6 +188,264 @@ const char *known_proto[256] = {
...
+    const char *name;
+    const char *alias;
+    uint8_t number;
+} ProtoNameHashEntry;
+
+HashTable *proto_ht;
+
+static uint32_t ProtoNameHashFunc(HashTable *ht, void *data, uint16_t datalen)
+{
+    ProtoNameHashEntry *p = (ProtoNameHashEntry *)data;
```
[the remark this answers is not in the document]
jlucovsky: I'll update this to use the hashing function from `src/util-hash-string.c`

jlucovsky:
Continued in #7052

PR #7043, opened 2022-02-21
nfs: Implement frames v5

ghostinthehive:
Previous PR #6992 
Changes from Previous PR:
- Adjust logic to tag `PDU` frames regardless of parser status.

Frames:
	- rpc.pdu
	- rpc.hdr
	- rpc.data
	- rpc.creds -- for rpc calls

  - NFSv2, NFSv3
	- nfs.pdu

  - NFSv4 Only Frames
	- nfs4.pdu
	- nfs4.hdr

RPC tcp/udp frames created with separate registeration functions e.g:
`add_rpc_tcp_tc_frames()`
`add_rpc_udp_tc_frames()`

suricata-verify-pr: 754

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -26,6 +26,7 @@ use nom7::{Err, Needed};
 
 use crate::applayer;
 use crate::applayer::*;
+use crate::frames::*;
 use crate::core::*;
 use crate::conf::*;
 use crate::filetracker::*;
@@ -86,6 +87,22 @@ static mut ALPROTO_NFS: AppProto = ALPROTO_UNKNOWN;
  * NFSTransaction where they can be looked up per handle as part of the
  * Transaction lookup.
  */
+
+#[derive(AppLayerFrameType)]
+pub enum NFSFrameType {
+    RPCPdu,
+    RPCHdr,
+    RPCData,
+    RPCCreds, // for rpc calls
+
+    NFSPdu,
+    NFSStatus,
+
+    NFS4Pdu,
+    NFS4Hdr,
+    NFS4Ops,
+    NFS4Status,
+}
 
 #[derive(FromPrimitive, Debug, AppLayerEvent)]
 pub enum NFSEvent {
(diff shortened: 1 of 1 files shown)


rust/src/nfs/nfs.rs
```diff
@@ -437,6 +454,104 @@ impl NFSState {
...
     }
 
+    fn add_rpc_udp_ts_pdu_frame(&mut self, flow: *const Flow, stream_slice: &StreamSlice, input: &[u8], rpc_len: i64) -> Option<Frame> {
+        let rpc_udp_ts_pdu = Frame::new_ts(flow, stream_slice, input, rpc_len, NFSFrameType::RPCPdu as u8);
+        SCLogDebug!("rpc_udp_pdu ts frame {:?}", rpc_udp_ts_pdu);
+        rpc_udp_ts_pdu
+    }
+
+    fn add_rpc_udp_ts_creds(&mut self, flow: *const Flow, stream_slice: &StreamSlice, input: &[u8], rpc_len: i64) {
+        let _rpc_udp_ts_creds = Frame::new_ts(flow, stream_slice, &input[24..], rpc_len - 32, NFSFrameType::RPCCreds as u8);
```
victorjulien, line 464: can you explain the calculation here?

rust/src/nfs/nfs.rs
```diff
@@ -437,6 +454,104 @@ impl NFSState {
...
+        SCLogDebug!("rpc_tcp_ts_creds {:?}", _rpc_tcp_ts_creds);
+    }
+
+    fn add_nfs_ts_frame(&mut self, flow: *const Flow, stream_slice: &StreamSlice, input: &[u8], nfs_req_len: i64) {
+        let _nfs_req_pdu = Frame::new_ts(flow, stream_slice, input, nfs_req_len, NFSFrameType::NFSPdu as u8);
+        SCLogDebug!("NFS Request Frame {:?}", _nfs_req_pdu);
+    }
+
+    fn add_nfs4_ts_pdu_frame(&mut self, flow: *const Flow, stream_slice: &StreamSlice, input: &[u8], nfs4_req_len: i64) {
+        let _nfs4_ts_pdu = Frame::new_ts(flow, stream_slice, input, nfs4_req_len, NFSFrameType::NFS4Pdu as u8);
```
victorjulien, line 485: you can replace `nfs4_req_len` with `input.len() as i64` here.

rust/src/nfs/nfs.rs
```diff
@@ -437,6 +454,104 @@ impl NFSState {
...
+        SCLogDebug!("NFS4 Request Frame: {:?}", _nfs4_ts_pdu);
+    }
+
+    fn add_nfs4_ts_hdr_ops_frame(&mut self, flow: *const Flow, stream_slice: &StreamSlice, input: &[u8], nfs4_req_len: i64) {
+        if nfs4_req_len > 0 {
+            let _nfs4_ts_hdr = Frame::new_ts(flow, stream_slice, input, 8, NFSFrameType::NFS4Hdr as u8);
+            SCLogDebug!("NFS4 Hdr Frame {:?}", _nfs4_ts_hdr);
+            if (nfs4_req_len - 8) > 0 {
```
victorjulien, line 493: `if nfs4_req_len > 8` ? Then it can't underflow if < 8.

rust/src/nfs/nfs.rs
```diff
@@ -530,18 +645,22 @@ impl NFSState {
...
 
     /// complete request record
-    fn process_request_record<'b>(&mut self, r: &RpcPacket<'b>) {
+    fn process_request_record<'b>(&mut self, flow: *const Flow, stream_slice: &StreamSlice, r: &RpcPacket<'b>) {
         SCLogDebug!("REQUEST {} procedure {} ({}) blob size {}",
                 r.hdr.xid, r.procedure, self.requestmap.len(), r.prog_data.len());
 
         match r.progver {
             4 => {
+                self.add_nfs4_ts_pdu_frame(flow, stream_slice, r.prog_data, r.prog_data.len() as i64);
```
victorjulien, line 654: I guess we can just merge these 2 funcs?

victorjulien (changes requested):
I think the size handling can be done a bit clearer. Overall looking quite good already.

rust/src/nfs/nfs.rs
ghostinthehive: rpc creds is of variable length, so I had to exclude all other fixed-length field to calculate creds.
should I explain this in a comment?

rust/src/nfs/nfs.rs
```diff
@@ -1082,15 +1207,17 @@ impl NFSState {
...
                         let n2 = cmp::min(n1, rec_size);
-                        return AppLayerResult::incomplete((i.len() - cur_i.len()) as u32, n2 as u32);
+                        return AppLayerResult::incomplete(stream_slice.len() - cur_i.len() as u32, n2 as u32);
                     }
 
                     // we have the full records size worth of data,
                     // let's parse it
                     match parse_rpc(&cur_i[..rec_size]) {
                         Ok((_, ref rpc_record)) => {
+                            self.add_rpc_tcp_ts_creds(flow, stream_slice, cur_i, (cur_i.len() - rpc_record.prog_data.len()) as i64);
```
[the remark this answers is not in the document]
ghostinthehive: correct, this needs some cleanups, Is it a good idea to make a vairable inside `add_rcp_ts_creds` to do the length calculation?

rust/src/nfs/nfs.rs
ghostinthehive: > `if nfs4_req_len > 8` ? Then it can't underflow if < 8.



exactly! this change was made specifically for CI Fuzzing Error checks, I have no data to back this up 
please see: https://github.com/OISF/suricata/pull/6992

I asked that how could it fail with `nfs4.ops` with `length < 8`, and didn't fail with the `nfs4.hdr` since its always of length 8.

rust/src/nfs/nfs.rs
victorjulien: we parse the creds, right? So we should know the length?

rust/src/nfs/nfs.rs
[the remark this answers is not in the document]
victorjulien: maybe easier to store the creds slice or length in the `rpc_record` and use that?

rust/src/nfs/nfs.rs
victorjulien: I'm not sure I get what you're saying. Are you saying `if nfs4_req_len > 8` doesn't work?

ghostinthehive: Oh, my bad I got that wrong :)
Yes Its a general check, so it should work, I'll work on that.

rust/src/nfs/nfs.rs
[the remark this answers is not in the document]
ghostinthehive: I guess! I'll try this out.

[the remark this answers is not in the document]
ghostinthehive: alright, since what the `nfs4.creds` care about is the `machine_name` `machine_len` of an `AUTH_UNIX` creds, 
I think I could follow the logic of checking `if creds_flavor == AUTH_UNIX` and then return the needed fields, or might as well return the `creds_buf` as a whole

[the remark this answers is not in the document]
victorjulien: not sure I understand the machine name comment. A `rpc.creds` frame should contain all parts of the creds fields. I'd think the len, flavor and then the actual data. Rule writers can then use the flavor field to check its unix, and then find the machine name easily?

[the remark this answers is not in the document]
ghostinthehive: Yes, okay good. on it.

ghostinthehive:
Followed by #7087

PR #7046, opened 2022-02-22
nfs4: add missing parsers v1

ghostinthehive:
An `NFS4` compound record is parsed as a whole -- 
if an operation in a compound record is missing a parser, the parsing will fail.

A couple of `NFS4` compound records were logged as `MalformedData` due to missing op parsers
Add missing parsers and their respective unittests if needed.

### DIFF
--- a/rust/src/nfs/nfs4_records.rs
+++ b/rust/src/nfs/nfs4_records.rs
@@ -61,6 +61,7 @@ pub enum Nfs4RequestContent<'a> {
     ExchangeId(Nfs4RequestExchangeId<'a>),
     Sequence(Nfs4RequestSequence<'a>),
     CreateSession(Nfs4RequestCreateSession<'a>),
+    ReclaimComplete(u32),
 }
 
 #[derive(Debug,PartialEq)]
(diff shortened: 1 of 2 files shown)


inashivb:
Could you please create a redmine ticket for this and link it here? :)

ghostinthehive:
> Could you please create a redmine ticket for this and link it here? :)

Sure! should I mark it as a task?

inashivb:
> > Could you please create a redmine ticket for this and link it here? :)
> 
> Sure! should I mark it as a task?

Sounds good to me.

Also, do we hav any s-v tests which logged malformed data before this PR and don't anymore?

ghostinthehive:
> > > Could you please create a redmine ticket for this and link it here? :)
> > 
> > Sure! should I mark it as a task?
> 
> Sounds good to me.
> 
> Also, do we hav any s-v tests which logged malformed data before this PR and don't anymore?

Yes, the `nfs4-01` but i dont know the PR# for that in Suricata-Verify repo to link it here

inashivb:
> > Also, do we hav any s-v tests which logged malformed data before this PR and don't anymore?
> 
> Yes, the `nfs4-01` but i dont know the PR# for that in Suricata-Verify repo to link it here

Since we already have it in the tree, perhaps good to add the filter block there which is now generated instead of malformed data block. It should ideally fail on 6 and 5 but pass with this PR. Would that work?

victorjulien:
This is nice work. I think it would be good to consider which of the parsed fields we can verify, see my work here https://github.com/OISF/suricata/pull/7049/commits/5ecb626e50048cfa43ad67cdd80307377e90926c

ghostinthehive:
Followed by #7110

PR #7048, opened 2022-02-22
smb: New keyword smb.filename v2

zer1t0:
https://redmine.openinfosecfoundation.org/issues/5082

Describe changes:
- Add new sticky buffer smb.filename to match the filenames that are being accessed by SMB through the create file request
- Add documentation for the keyword

Rule example: `alert smb any any -> any any (msg: "SMB file a.txt";smb.filename; content:"a.txt";sid:1;)`

suricata-verify-pr: 802


catenacyber:
This looks useful.

Just a comment :

> Rule example (unicode name): alert smb any any -> any any (msg: "SMB file a.txt";smb.filename; content:"a|00|.|00|t|00|x|00|t|00|";sid:1;)

This can be easily evaded if I set in my SMB header not to use unicode and I have the file name `a.txt`

cf the latest Suricon presentation about evading SMB 
https://suricon.net/wp-content/uploads/2021/11/SURICON2021-Del-ValleJacotot-Evading-Suricata-IDS.pdf

The `smb.named_pipe` keyword does not get evaded by this

zer1t0:
Nice catch.

I thought that only Unicode was allowed but (based on the documentation) seems that as you say, it is possible to use ascii filenames on SMB1, here is what I found.

- SMB2 [TREE_CONNECT](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/832d2130-22e8-4afb-aafd-b30bb0901798) Path -> Unicode
- SMB2 [CREATE](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/e8fb45c1-a03d-44ca-b7ae-47385cfd7997) Filename -> Unicode
- SMB [SMB_COM_TREE_CONNECT_ANDX](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/90bf689a-8536-4f03-9f1b-683ee4bdd67c) Path -> Unicode or ASCII (based on the SMB_FLAGS2_UNICODE flag)
- SMB [SMB_COM_NT_CREATE_ANDX](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/f2a0f032-7545-41c9-9ceb-aab39852c11a) Filename -> Unicode or ASCII (based on the SMB_FLAGS2_UNICODE flag)

How was this issue solved by `smb.named_pipe`?


### Offtopic
While I was watching the code looking for the solution to the problem, watching the `smb.named_pipe` code, I discovered that this keyword is used to match the share name `IPC` in the tree_connect request. I think here we are not using the proper terminology because IPC$ is not a named pipe, [IPC$ is a pipe share, the share used to access named_pipes](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/b1b7cc8a-4d24-4701-bc3f-220b543ceef8#gt_a49a79ea-dac7-4016-9a84-cf87161db7e3). The named pipes themselves are those "special files" under IPC$ used to connect with an specific service in the target machine, like `samr`, `lsarpc`, etc. Therefore I think the name of the keyword is misguiding and something like `smb.pipe_share` would be more accurate.

catenacyber:
> How was this issue solved for smb.named_pipe?

I do not know, you should look in the code.

There may be pcaps in Suricata-verify tests with this ascii encoding...

For the off topic part, maybe you can create a redmine ticket.
I think there are keyword aliases which should allow to use `smb.pipe_share`

doc/userguide/rules/index.rst
```diff
@@ -26,6 +26,7 @@ Suricata Rules
    snmp-keywords
    base64-keywords
    sip-keywords
+   smb-keywords
```
catenacyber, line 29: Thanks for adding the doc.

If you are interested, there is also https://redmine.openinfosecfoundation.org/issues/3030 ;-)

src/Makefile.am
```diff
@@ -292,7 +292,8 @@ noinst_HEADERS = \
 	detect-sip-stat-code.h \
 	detect-sip-stat-msg.h \
 	detect-sip-uri.h \
-	detect-smb-share.h \
+	detect-smb-filename.h \
+  detect-smb-share.h \
```
catenacyber, line 296: nit: ident is off

rust/src/smb/detect.rs
```diff
@@ -201,3 +201,20 @@ pub extern "C" fn rs_smb_tx_get_dce_iface(state: &mut SMBState,
...
     return 0;
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn rs_smb_tx_get_filename(tx: &mut SMBTransaction,
+                                                buffer: *mut *const u8,
+                                                buffer_len: *mut u32)
+                                                -> u8
+{
+    if let Some(SMBTransactionTypeData::CREATE(ref x)) = tx.type_data {
```
catenacyber, line 211: Why only create ?
Are there not commands such as get file info where this may be interesting ?

doc/userguide/rules/smb-keywords.rst
```diff
@@ -0,0 +1,16 @@
+SMB Keywords
+==============
+
+SMB keywords used in both SMB1 and SMB2 protocols.
+
+smb.filename
```
catenacyber, line 6: Could you please explain the difference with generic `file.name` keyword in the doc ?

rust/src/smb/detect.rs
zer1t0: The file name is only transferred in the create request (that allows to create or open files) to get a handle for the target file, other requests like query_info used this handle, so only create is interesting to the get the filename.

src/Makefile.am
zer1t0: Done in https://github.com/OISF/suricata/pull/7048/commits/a83f189494b78192c1f3db055aac28c92c39f841

doc/userguide/rules/smb-keywords.rst
```diff
@@ -0,0 +1,16 @@
...
+--------------
+
+SMB filename is a sticky buffer to match the filename in SMB Create requests.
+
+.. note:: Remember that SMB2 filenames are Unicode encoded.
+
+If you want to match traffic that access to file "a.txt", you could use the following rule::
+
+  alert smb any any -> any any (msg: "SMB file match";smb.filename; content:"a|00|.|00|t|00|x|00|t|00|";sid:1;)
```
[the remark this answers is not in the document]
zer1t0: I've been investigating how normalization works in named_pipe and I found that for SMB1, both the share name and the filename, in their respectives messages are parsed the same way with the `smb1_get_string` function : 
- https://github.com/OISF/suricata/blob/99b3443369703ba69d962019edeca8f1c7c7896e/rust/src/smb/smb1_records.rs#L240
- https://github.com/OISF/suricata/blob/99b3443369703ba69d962019edeca8f1c7c7896e/rust/src/smb/smb1_records.rs#L596

However, in SMB2 things change, as tree connect is normalized after parsing:
- https://github.com/OISF/suricata/blob/99b3443369703ba69d962019edeca8f1c7c7896e/rust/src/smb/smb2.rs#L460

So I included this change for SMB2 create filename in https://github.com/OISF/suricata/pull/7048/commits/27f92a144df06b3b859736b6753109de80d1f354, please take a look to it, to be sure is fine. I'm not sure if this change could bring unintended effects.

rust/src/smb/detect.rs
catenacyber: Oh thanks :-)

doc/userguide/rules/smb-keywords.rst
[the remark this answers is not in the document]
catenacyber: Does this affect logging ?
If so, it should be documented cf doc/userguide/upgrade.rst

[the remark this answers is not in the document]
zer1t0: No, it doesn't affect logging, all tests are fine and SMB create transaction the name is filtered in the same way:

https://github.com/OISF/suricata/blob/8ef066318d7c4dd9b6686dbebd621c790828c384/rust/src/smb/log.rs#L193

zer1t0:
Moved to https://github.com/OISF/suricata/pull/7194

PR #7049, opened 2022-02-22
Nfs fixes/v6

victorjulien:
https://redmine.openinfosecfoundation.org/issues/5145
https://redmine.openinfosecfoundation.org/issues/5094

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -983,6 +983,57 @@ impl NFSState {
         return AppLayerResult::ok();
     }
 
+    /// Handle partial records
+    fn parse_tcp_partial_data_ts<'b>(&mut self, base_input: &'b[u8], start_input: &'b[u8],
+            phdr: &RpcRequestPacketPartial, rec_size: usize) -> AppLayerResult {
+        let mut cur_i = start_input;
+        // special case: avoid buffering file write blobs
+        // as these can be large.
+        if rec_size >= 512 && cur_i.len() >= 44 {
+            // large record, likely file xfer
+            SCLogDebug!("large record {}, likely file xfer", rec_size);
+
+            // quick peek, are we in WRITE mode?
+            if phdr.procedure == NFSPROC3_WRITE {
+                SCLogDebug!("CONFIRMED WRITE: large record {}, file chunk xfer", rec_size);
+
+                // lets try to parse the RPC record. Might fail with Incomplete.
+                match parse_rpc(cur_i) {
+                    Ok((remaining, ref hdr)) => {
+                        match parse_nfs3_request_write(hdr.prog_data) {
+                            Ok((_, ref w)) => {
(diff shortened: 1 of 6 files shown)


rust/src/nfs/nfs.rs
```diff
@@ -647,8 +651,8 @@ impl NFSState {
         }
         if !self.is_udp {
             self.ts_chunk_xid = r.hdr.xid;
-            let file_data_len = w.file_data.len() as u32 - fill_bytes as u32;
-            self.ts_chunk_left = w.file_len as u32 - file_data_len as u32;
+            debug_validate_bug_on!(w.file_data.len() as u32 > w.count);
```
catenacyber, line 654: I think this will trigger on fuzzing input (even if I have not seen it on real pcaps)
cf code
```
    let (i, count) = be_u32(i)?;
    let (i, stable) = be_u32(i)?;
    let (i, file_len) = be_u32(i)?;
    let (i, file_data) = take(file_len as usize)(i)?;
```

So, count can be arbitrary small compared to w.file_data.len()

rust/src/nfs/nfs.rs
```diff
@@ -1033,81 +1100,31 @@ impl NFSState {
...
-                                    },
-                                }
-                            }
-                        }
-                        // make sure we pass a value higher than current input
-                        // but lower than the record size
-                        let n1 = cmp::max(cur_i.len(), 1024);
-                        let n2 = cmp::min(n1, rec_size);
-                        return AppLayerResult::incomplete((i.len() - cur_i.len()) as u32, n2 as u32);
+                        return self.parse_tcp_partial_data_ts(i, cur_i, rpc_phdr, rec_size);
```
catenacyber, line 1106: Why do not we just buffer TCP with `AppLayerResult::incomplete` ?

rust/src/nfs/nfs.rs
victorjulien: which path to this do you still see? Record parsing now does
```rust
    let (i, count) = be_u32(i)?;
    let (i, stable) = verify(be_u32, |&v| v <= 2)(i)?;
    let (i, file_len) = verify(be_u32, |&v| v <= count)(i)?;
```
So `file_len <= count`

rust/src/nfs/nfs.rs
victorjulien: the parser can handle the large records w/o buffering. Records can be several megs, which would require much buffering

victorjulien: also means data is available to the detect engine sooner

catenacyber: > Records can be several megs

That is the point I missed, thanks
Do you have such a pcap ?

victorjulien: only in tests based on private pcaps, sadly

PR #6904, opened 2022-01-31
http: move xff logging to http object

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4860
but also https://redmine.openinfosecfoundation.org/issues/1369

Describe changes:
- move some app-layer specific fields out of root object (into the app-layer specific object)
  - http : xff (for alert metadata as well)

suricata-verify-pr: 698
https://github.com/OISF/suricata-verify/pull/698

Replaces #6660 with logging `http.xff` for alerts even without `LOG_JSON_APP_LAYER`

PR #7051, opened 2022-02-22
log: rename http.xff to http.xff_header

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4860
but also https://redmine.openinfosecfoundation.org/issues/1369

Describe changes:
- http : rename http.xff to `http.xff_header` as it does not have the same meaning as `.xff`

suricata-verify-pr: 756
https://github.com/OISF/suricata-verify/pull/756

Replaces #6904 with renaming `http.xff` to `http.xff_header`

### DIFF
--- a/doc/userguide/upgrade.rst
+++ b/doc/userguide/upgrade.rst
@@ -46,6 +46,7 @@ Logging changes
 - IKEv2 Eve logging changed, the event_type has become ``ike``. The fields ``errors`` and ``notify`` have moved to
   ``ike.ikev2.errors`` and ``ike.ikev2.notify``.
 - FTP DATA metadata for alerts are now logged in ``ftp_data`` instead of root.
+- HTTP xff header is now logged as ``http.xff_header`` instead of ``http.xff`` while the root object ``.xff`` remains the same
 
 Other changes
 ~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


suricata-qa:

Pipeline 6389

jasonish:
Can you provide samples of the output? As mentioned here, https://github.com/OISF/suricata/pull/6904#issuecomment-1065600747, I don't think I'm in favour of adding a new field. I'd rather see xff gone at the root, and xff only logged in the http object like here: https://github.com/OISF/suricata/pull/6904#issuecomment-1030405503 regardless of the code path that caused it to be logged.

This lets us remove a xff at the root, which I think is a better breaking change that renaming http.xff.

catenacyber:
> Its a breaking change for some I suppose, but I think its a nice cleanup to get the xff out of the root, and just have it in the http object. But include all addresses.

Do you know where this feature comes from ?
This feature is the `xff.deployment` configuration option...
I see commit 68f43ffffb 

Maybe I can remove xff from the root, and have another value for the config option `xff.deployment` being `all` in addition to `reverse` and `forward`

jasonish:
> > Its a breaking change for some I suppose, but I think its a nice cleanup to get the xff out of the root, and just have it in the http object. But include all addresses.
> 
> Do you know where this feature comes from ? This feature is the `xff.deployment` configuration option... I see commit [68f43ff](https://github.com/OISF/suricata/commit/68f43ffffb6f964042234cc0d5fffef5d94287cd)
> 
> Maybe I can remove xff from the root, and have another value for the config option `xff.deployment` being `all` in addition to `reverse` and `forward`

Ok, I see whats going on.

The `xff` at the root is either the first or last IP in the list based on the setting of `mode`.  I didn't realize that, and perhaps it does have a valid use being different from `http.xff`.  So maybe just move it from root into alert, so its a propertry of the alert object and be done with it?

catenacyber:
> So maybe just move it from root into alert, so its a propertry of the alert object and be done with it?

Ok will do

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7148

PR #7053, opened 2022-02-22
(draft) smb: flags fix; set event when record is in the wrong direction - v2

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/7023

Changes from last PR:
- If a record is seen in the wrong direction, set an event on the frame.
- Don't return an err, let the parser continue.

TODO: Add the rules for the alerts.

Ticket: https://redmine.openinfosecfoundation.org/issues/4945

### DIFF
--- a/rust/src/smb/smb1_records.rs
+++ b/rust/src/smb/smb1_records.rs
@@ -27,6 +27,8 @@ use nom7::IResult;
 
 pub const SMB1_HEADER_SIZE: usize = 32;
 
+pub const SMB1_FLAGS_REPLY: u8 = 0x80;
+
 fn smb_get_unicode_string_with_offset(i: &[u8], offset: usize) -> IResult<&[u8], Vec<u8>, SmbError>
 {
     let (i, _) = cond(offset % 2 == 1, take(1_usize))(i)?;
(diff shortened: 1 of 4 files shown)


jasonish:
This logic could be ported to 5 and 6 without too much issue I think.

rust/src/smb/smb1_records.rs
```diff
@@ -815,6 +817,15 @@ impl<'a> SmbRecord<'a> {
     pub fn is_dos_error(&self) -> bool {
         self.flags2 & 0x4000_u16 != 0
     }
+
+    /// Returns 0 if a request or 1 if a response.
+    pub fn direction(&self) -> u8 {
```
victorjulien, line 822: think this should return `core::Direction`

victorjulien:
> > Frame event logging/inspection isn't implemented yet. Just FYI
> 
> Ok, I don't know if setting the event on the transaction would be any better, as it probably doesn't exist.

Right. I think using the frames is fine, its just a bit hard to test atm. Would be good if you could create a draft SV test for developing the event frame support against.

suricata-qa:

Pipeline 6369

jasonish:
Replaced by https://github.com/OISF/suricata/pull/7084.

jasonish:
> Right. I think using the frames is fine, its just a bit hard to test atm. Would be good if you could create a draft SV test for developing the event frame support against.

Here are 2 tests where the payload is reversed so the request is from the server, etc. One for SMB1 and another for SMB2.

PR #7063, opened 2022-02-23
detect: only apply ConfigApplyTx with app-layers

catenacyber:

Describe changes:
- detect: only apply ConfigApplyTx when relevant (like with app-layer)

Replaces #7015 with adding `SIG_MASK_REQUIRE_FLOW` set by 
```
    if (s->init_data->init_flags & SIG_FLAG_INIT_FLOW) {
        s->mask |= SIG_MASK_REQUIRE_FLOW;
        SCLogDebug("sig requires flow");
    }
```
in `SignatureCreateMask`

### DIFF
--- a/src/detect-config.c
+++ b/src/detect-config.c
@@ -288,6 +288,12 @@ static int DetectConfigSetup (DetectEngineCtx *de_ctx, Signature *s, const char 
         fd->scope = CONFIG_SCOPE_TX;
     }
 
+    if (fd->scope == CONFIG_SCOPE_TX) {
+        s->flags |= SIG_FLAG_APPLAYER;
+    } else if (fd->scope == CONFIG_SCOPE_FLOW) {
+        s->init_data->init_flags |= SIG_FLAG_INIT_FLOW;
+    }
+
     sm->ctx = (SigMatchCtx*)fd;
     SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_POSTMATCH);
 


src/detect-config.c
```diff
@@ -288,6 +288,12 @@ static int DetectConfigSetup (DetectEngineCtx *de_ctx, Signature *s, const char
         fd->scope = CONFIG_SCOPE_TX;
     }
 
+    if (fd->scope == CONFIG_SCOPE_TX) {
+        s->flags |= SIG_FLAG_APPLAYER;
+    } else if (fd->scope == CONFIG_SCOPE_FLOW) {
+        s->init_data->init_flags |= SIG_FLAG_INIT_FLOW;
```
victorjulien, line 294: this will have side effects. The purpose of this flag is to track if the `flow:` keyword is in use in the sig. I think to trigger the `SignatureCreateMask` function itself currently.

victorjulien (changes requested):
see inline comment

suricata-qa:
Information:
 
|--|--|--|--| 
| | tlpr1_stats_chk | | |
| .flow.memuse | 501681408 | 542473728 | 92.48% | 

Pipeline 6379

src/detect-config.c
catenacyber: Ok cf https://github.com/OISF/suricata/pull/7080

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7080

PR #7072, opened 2022-02-26
detect/bytemath: Handle out-of-order keywords

jlucovsky:
Issue: 5077

This commit allows byte_math keywords past <rvalue> to be in arbitrary
order to provide better compatibility with Snort's handling.

"Arbitrary order" refers to an order that is different from Snort's documented keyword order.

Describe changes:
- Reworked PCRE capture to allow arbitrary order of keywords.
- Added unittest with out of order keywords.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-bytemath.c
+++ b/src/detect-bytemath.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2020 Open Information Security Foundation
+/* Copyright (C) 2020-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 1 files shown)


src/detect-bytemath.c
```diff
@@ -263,6 +267,7 @@ static DetectByteMathData *DetectByteMathParse(DetectEngineCtx *de_ctx, const ch
     char tmp_str[128] = "";
 
     ret = DetectParsePcreExec(&parse_regex, arg, 0, 0);
+    SCLogNotice("byte_math expression: \"%s\"; ret: %d", arg, ret);
```
victorjulien, line 270: debug

victorjulien (commented):
wondering if we should get rid of pcre for parsing completely, like we're doing with xbits

src/detect-bytemath.c
jlucovsky: Will fix.

jlucovsky:
> wondering if we should get rid of pcre for parsing completely, like we're doing with xbits

Is that in progress? https://github.com/OISF/suricata/blob/master/src/detect-xbits.c#L206

catenacyber:
> > wondering if we should get rid of pcre for parsing completely, like we're doing with xbits
> 
> Is removal of pcre from xbits work in progress? https://github.com/OISF/suricata/blob/master/src/detect-xbits.c#L206

I think there are PRs from Shivani...

jlucovsky:
Continued in #7113

PR #7073, opened 2022-02-28
style: use u8_tolower everywhere

catenacyber:

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` about uses of `tolower`

(`tolower` takes an integer ans returns an integer, but really we use it for `u8`)

Should be uncontroversial part of #7006

### DIFF
--- a/src/counters.c
+++ b/src/counters.c
@@ -1024,7 +1024,7 @@ static uint32_t CountersIdHashFunc(HashTable *ht, void *data, uint16_t datalen)
     int len = strlen(t->string);
 
     for (int i = 0; i < len; i++)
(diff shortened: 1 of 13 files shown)


victorjulien:
I'm finding `style` to be confusing here. To me "style" means code style, not something that changes the code itself. I think I would prefer a commit subject prefix of something like `ints` or maybe even a generic `src`.

catenacyber:
> To me "style" means code style, not something that changes the code itself.

That is what I meant indeed

> I think I would prefer a commit subject prefix of something like `ints` or maybe even a generic `src`.

Done in https://github.com/OISF/suricata/pull/7081

PR #6887, opened 2022-01-28
App layer error close txs 4318 v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4318

Describe changes:
- app-layer: clean and "close" all txs if protocol reaches error state

The best way I see to test this is SMB
But I am not sure how to get multiple live transactions over SMB

Then to get an error, simply have a NBSS message with length 4, and just `\xFFSMB` so the SMB parser will fail decode the SMB1 header (not sure it should put the whole parser in an error state by the way, as we could go on with the next NBSS message)

Modifies #6800 by adding a commit to rename STREAM_DEPTH to STREAM_DISRUPTED

PR #7076, opened 2022-02-28
App layer error close txs 4318 v4

catenacyber:

Describe changes:

The best way I see to test this is SMB
But I am not sure how to get multiple live transactions over SMB

Then to get an error, simply have a NBSS message with length 4, and just `\xFFSMB` so the SMB parser will fail decode the SMB1 header (not sure it should put the whole parser in an error state by the way, as we could go on with the next NBSS message)

Modifies #6887 with rebase to get QA running fresh again


catenacyber:
Oh... I think the S-V test is wrong : we only want one alert about `tls.invalid_handshake_message` as there is only call to `SSLSetEvent(ssl_state, TLS_DECODER_EVENT_INVALID_HANDSHAKE_MESSAGE);`

Now on master, we get 2 : when it happens, and when the connection gets closed and the transaction gets cleaned...

PR merging this was https://github.com/OISF/suricata-verify/pull/695 commit https://github.com/OISF/suricata-verify/pull/695/commits/7bd1448960ccea61fd7afbb0996e3f1a31f9da63

See https://github.com/OISF/suricata-verify/pull/748 which seems an attempt to fix this

catenacyber:
cc @chifflier

suricata-qa:
ERROR:
 

Pipeline 6478

catenacyber:
Need to investigate cf https://redmine.openinfosecfoundation.org/issues/1983

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7243

The buggy test showed a buggy behavior with this PR.
Having an app-layer parser return an error prevented the detection being run on the transaction in both directions with this PR...

PR #7082, opened 2022-02-28
Quic ietf 4967 v2

catenacyber:

Describe changes:
- Rustfmt previous code

suricata-verify-pr: 755

This is ready for review, but not for merge yet obviously.

Replaces https://github.com/OISF/suricata/pull/7047 with comments taken into account

Still to do :
- see https://redmine.openinfosecfoundation.org/issues/4966
- Detection (Victor said I guess we'd want parity with existing tls/tcp as a first step)


victorjulien:
Not yet sure if this is a real issue or a local issue, but I'm seeing this on Debian 11
Screenshot from 2022-03-01 08-53-44

catenacyber:
How can I test this ?

@jasonish how does my Cargo.toml.in change look like ?
Should I add `, default-features = false` ? or something else ?

jasonish:
> @jasonish how does my Cargo.toml.in change look like ? Should I add `, default-features = false` ? or something else ?

Yeah, I'd see if you can get by with default-features of false. It will drop the logging and tls12 features that may not be needed for our use case.

catenacyber:
Thanks Jason

Replaced by https://github.com/OISF/suricata/pull/7095

PR #7031, opened 2022-02-18
Staging 605 20220209/v4

inashivb:
- https://github.com/OISF/suricata/pull/6957
- https://github.com/OISF/suricata/commit/6c240938b51daadb6c869d2339c2e7cc29ebde68 (to fix the builds)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) tickets:
- https://redmine.openinfosecfoundation.org/issues/4925
- https://redmine.openinfosecfoundation.org/issues/4927

suricata-verify-pr: 724

Previous PR: #6977 

Changes since v3:
- Fuzz invalid conditions fix commit fd6edc0cbc46ab63c5e7888b4981436742a0e580 added

PR #7083, opened 2022-02-28
Next/60x/20220228/v2

victorjulien:
Staging:
- #7031 w/o the nfs commit
- #7024

CIFuzz is expected to fail due to NFS.


victorjulien:
Interesting, looks like CIFuzz (undefined) found a new corner case for when an IP-only sig uses `0.0.0.5/0.0.0.5`.

victorjulien:
> Interesting, looks like CIFuzz (undefined) found a new corner case for when an IP-only sig uses `0.0.0.5/0.0.0.5`.

https://redmine.openinfosecfoundation.org/issues/5168

suricata-qa:

Pipeline 6440

inashivb:
This was added to #7089 . Thanks!

PR #7087, opened 2022-03-01
nfs: Implement frames v6

ghostinthehive:
Previous PR #7043 
Changes from Previous PR:
- Cleanups
- Refine `rpc.creds` frame length calculation

Feature [#4872](https://redmine.openinfosecfoundation.org/issues/4872)

Frames:

  - RPC Frames: Generic over TCP/UDP
       - rpc.pdu
       - rpc.hdr
       - rpc.data
       - rpc.creds -- for rpc calls

   - NFSv2, NFSv3
       - nfs.pdu
       - nfs.status -- for nfs responses

   - NFSv4 Only Frames
       - nfs4.pdu
       - nfs4.hdr
       - nfs4.ops -- for compound request/response operations
       - nfs4.status -- for nfs4 responses

RPC tcp/udp frames created with separate registeration functions e.g:
`add_rpc_tcp_tc_frames()`
`add_rpc_udp_tc_frames()`

suricata-verfiy-pr: 771

PR #7088, opened 2022-03-01
nfs: Implement frames v7

ghostinthehive:
Previous PR https://github.com/OISF/suricata/pull/7087
Changes from Previous PR:
- Fix incorrect offset

Frames:

   -  RPC Frames: Generic over TCP/UDP
       - rpc.pdu
       - rpc.hdr
       - rpc.data
       -  rpc.creds -- for rpc calls

   -  NFSv2, NFSv3
       - nfs.pdu

   - NFSv4 Only Frames
       - nfs4.pdu
       - nfs4.hdr

RPC tcp/udp frames created with separate registeration functions e.g:
`add_rpc_tcp_tc_frames()`
`add_rpc_udp_tc_frames()`

suricata-verfiy-pr: 771

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -26,6 +26,7 @@ use nom7::{Err, Needed};
 
 use crate::applayer;
 use crate::applayer::*;
+use crate::frames::*;
 use crate::core::*;
 use crate::conf::*;
 use crate::filetracker::*;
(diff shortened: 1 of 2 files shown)


jufajardini (changes requested):
Looking good to me, from what I can follow. I'm wondering if it is better to wait for the current patches that will gome to nfs/rpc to be merged, so we know how much conflict will happen here.

Also noticed that there is a difference in when the `rpc_tcp_ts_frames` and `rpc_tcp_tc_frames` are registered (to client has all frames registered by the same function, whereas with to server we first do pdu, then after successfully parsing, do the rest, right?), but wasn't able to understand why it was done differently, probably because of not understanding nfs in-depth.

One nit formatting comment inline.

ghostinthehive:
> Looking good to me, from what I can follow. I'm wondering if it is better to wait for the current patches that will gome to nfs/rpc to be merged, so we know how much conflict will happen here.
> 
> Also noticed that there is a difference in when the `rpc_tcp_ts_frames` and `rpc_tcp_tc_frames` are registered (to client has all frames registered by the same function, whereas with to server we first do pdu, then after successfully parsing, do the rest, right?), but wasn't able to understand why it was done differently, probably because of not understanding nfs in-depth.
> 
> One nit formatting comment inline.

This is actually true, I have squashed all `rpc_tc` frames not giving attention to the need of only registering the PDU first. The frames them selves needs to be separated like with `rpc_ts` `pdu` and `creds` frames
Thanks for noting that out!

ghostinthehive:
Followed by #7094

PR #7095, opened 2022-03-02
Quic ietf 4967 v3

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5143

Describe changes:
- Rustfmt previous code

suricata-verify-pr: 775
https://github.com/OISF/suricata-verify/pull/775

Replaces https://github.com/OISF/suricata/pull/7082 with
- default features false for added crate rustls
- detection for sni
- detection for ja3 string

Still to do :
- see https://redmine.openinfosecfoundation.org/issues/4966


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | tlpw1_stats_chk | | |
| .app_layer.tx.quic | 2592 | 1140 | 227.37% |
| .app_layer.error.quic.parser | 3768 | 1810 | 208.18% |
| | tlpr1_stats_chk | | |
| .app_layer.tx.quic | 13818 | 4198 | 329.16% |
| .app_layer.error.quic.parser | 1416606 | 43951 | 3223.15% | 

Pipeline 6451

rust/src/quic/frames.rs
```diff
@@ -115,13 +123,210 @@ impl fmt::Display for StreamTag {
...
+    pub ack_delay: u64,
+    pub ack_range_count: u64,
+    pub first_ack_range: u64,
+}
+
+#[derive(Debug, PartialEq)]
+pub(crate) struct Crypto {
+    pub ciphers: Vec<TlsCipherSuiteID>,
+    pub extv: Vec<QuicTlsExtension>,
+    //does not work, because of lifetime due to references to the slice used for parsing
```
victorjulien, line 138: can this comment move to `extv` just above and explain why it is like this instead of using `Vec<TlsExtension>`? Comments about why code that is in use are like they are, are more useful than comments explaining commented out things, I think.

rust/src/quic/frames.rs
```diff
@@ -115,13 +123,210 @@ impl fmt::Display for StreamTag {
...
+                        _ => {}
+                    }
+                } else {
+                    dash = true;
+                }
+                match write!(ja3, "{}", u16::from(etype)) {
+                    _ => {}
+                }
+                let mut values = Vec::new();
+                match e {
```
victorjulien, line 202: can this be broken out into a util func?

rust/src/quic/frames.rs
```diff
@@ -115,13 +123,210 @@ impl fmt::Display for StreamTag {
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct QuicTlsExtension {
+    pub etype: TlsExtensionType,
+    pub values: Vec<Vec<u8>>,
+}
+
+// get interesting stuff out of parsed tls extensions
+fn quic_get_tls_extensions(
```
victorjulien, line 182: could this func be done in a more compact way?

rust/src/quic/frames.rs
```diff
@@ -115,13 +123,210 @@ impl fmt::Display for StreamTag {
...
+                            let mut value = Vec::new();
+                            value.extend_from_slice(alpn);
+                            values.push(value);
+                        }
+                    }
+                    _ => {}
+                }
+                extv.push(QuicTlsExtension { etype, values })
+            }
+            if client {
```
victorjulien, line 221: perhaps move this into a client specific util func

rust/src/quic/frames.rs
```diff
@@ -115,13 +123,210 @@ impl fmt::Display for StreamTag {
...
+            match hs {
+                ClientHello(ch) => {
+                    let mut ja3 = String::with_capacity(256);
+                    match write!(&mut ja3, "{},", u16::from(ch.version)) {
+                        _ => {}
+                    }
+                    let mut dash = false;
+                    for c in &ch.ciphers {
+                        if dash {
+                            match write!(&mut ja3, "-") {
```
victorjulien, line 291: these kinds of statements are taking 3 lines each for no good reason. Can these be made more compact?

rust/src/quic/quic.rs
```diff
@@ -112,44 +129,146 @@ impl QuicState {
...
+        if let Some(keys) = &self.keys {
+            let hkey = if to_server {
+                &keys.remote.header
+            } else {
+                &keys.local.header
+            };
+            if framebuf.len() < 4 + hkey.sample_len() {
+                return false;
+            }
+            let mut h2 = Vec::with_capacity(hlen + header.length);
```
victorjulien, line 145: what is the max size we might alloc here?

rust/src/quic/quic.rs
```diff
@@ -112,44 +129,146 @@ impl QuicState {
...
+                    let mut output;
+                    if self.keys.is_some() {
+                        output = Vec::with_capacity(framebuf.len() + 4);
+                        if !self.decrypt(to_server, &header, framebuf, buf, hlen, &mut output) {
+                            return false;
+                        }
+                        framebuf = &output;
+                    }
+                    buf = next_buf;
+                    match QuicData::from_bytes(framebuf) {
```
victorjulien, line 214: can this func be broken up, for example handling the following in a until func? Not only is the func large, the indent levels are also getting hard to track

victorjulien (changes requested):
Some comments, mostly code tidiness suggestions. Will also give this a run on my own quic traffic.

victorjulien:
Should the `ja3` string here end in a comma?
```json
{
  "version": "1",
  "sni": "mask.icloud.com",
  "ja3": "771,51914-4865-4866-4867,64250-0-10-16-5-13-18-51-45-43-57-27-20-64250,6682-29-23-24-25,",
  "extensions": [
    {
      "type": "Grease"
    },
    {
      "type": "ServerName",
      "values": [
        "mask.icloud.com"
      ]
    },
    {
      "type": "SupportedGroups"
    },
    {
      "type": "ApplicationLayerProtocolNegotiation",
      "values": [
        "h3"
      ]
    },
    {
      "type": "StatusRequest"
    },
    {
      "type": "SignatureAlgorithms"
    },
    {
      "type": "SignedCertificateTimestamp"
    },
    {
      "type": "KeyShare"
    },
    {
      "type": "PskExchangeModes"
    },
    {
      "type": "SupportedVersions"
    },
    {
      "type": "TlsExtensionType(57 / 0x39)"
    },
    {
      "type": "TlsExtensionType(27 / 0x1b)"
    },
    {
      "type": "ServerCertificateType"
    },
    {
      "type": "Grease"
    }
  ]
}
```

victorjulien:
@jasonish any thoughts on the logging here (see example above). I find the `extentions` rather verbose, but can't immediately come up with an alternative.

Additionally I wonder how to log the unknown extensions. If we log them this way and then later add support for them its in a way a breaking output change? :)

rust/src/quic/frames.rs
catenacyber: @jasonish is there a better rustier way to build a string and write integer into it ?

rust/src/quic/quic.rs
catenacyber: Not more than the UDP payload size...
Should it be more obvious or debug tested ?

rust/src/quic/frames.rs
jasonish: You could do
```
let mut s = String::new();
s.push_str(&my_u16.to_string());
```

rust/src/quic/frames.rs
```diff
@@ -115,13 +123,210 @@ impl fmt::Display for StreamTag {
...
+                if dash {
+                    match write!(ja3, "-") {
+                        _ => {}
+                    }
+                } else {
+                    dash = true;
+                }
+                match write!(ja3, "{}", u16::from(etype)) {
+                    _ => {}
+                }
```
jasonish, line 200: The idiomatic way to avoid the result warning is:
```
let _ = write!(...
```

rust/src/quic/frames.rs
jasonish:
```
let _ = write...
```
(noted above already)

jasonish:
> @jasonish any thoughts on the logging here (see example above). I find the `extentions` rather verbose, but can't immediately come up with an alternative.

I think it looks fine as I can't really see another way.  Would we want to log all the extensions by default? Or should we have an extended mode that logs extensions which is off by default.

> Additionally I wonder how to log the unknown extensions. If we log them this way and then later add support for them its in a way a breaking output change? :)

Maybe something like:
```
"extensions": [
    {
      "type": 0,
      "name": "ServerName",
      "values": [
        "mask.icloud.com"
      ]
    },
    {
      "type": 99
    },
```
`type` is the only required field.

I also noticed that we're using names like `ServerName`, while Wireshark uses `server_name`.  Whats the most common here?

rust/src/quic/frames.rs
catenacyber: ok

rust/src/quic/frames.rs
catenacyber: ok

rust/src/quic/frames.rs
catenacyber: Thank you Jason, looks better indeed

rust/src/quic/frames.rs
catenacyber: `push_str` seems nice

rust/src/quic/quic.rs
catenacyber: ok

rust/src/quic/quic.rs
```diff
@@ -188,15 +307,28 @@ pub unsafe extern "C" fn rs_quic_probing_parser(
...
+    _flow: *const Flow, state: *mut std::os::raw::c_void, _pstate: *mut std::os::raw::c_void,
+    stream_slice: StreamSlice, _data: *const std::os::raw::c_void,
+) -> AppLayerResult {
+    let state = cast_pointer!(state, QuicState);
+    let buf = stream_slice.as_slice();
+
+    if state.parse(buf, false) {
+        return AppLayerResult::ok();
+    }
+    return AppLayerResult::err();
```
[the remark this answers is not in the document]
catenacyber: Ok. I did not know Rust was in the explicit else side

catenacyber:
> I also noticed that we're using names like ServerName, while Wireshark uses server_name. Whats the most common here?

In the RFC https://datatracker.ietf.org/doc/html/rfc3546
`server_name` is the enum value and `ServerName` the name of the structure...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7101

jasonish:
> In the RFC https://datatracker.ietf.org/doc/html/rfc3546
> `server_name` is the enum value and `ServerName` the name of the structure...

I think using a name of `server_name` is more consistent with the RFC.

rust/src/quic/quic.rs
victorjulien: yeah, I was trying to figure out where it is set and the logic is quite hard to follow but ultimately seems to come from `quic_var_uint`, so it might be higher if not properly checked?

If it shouldn't be > UDP payload size, could it be `u16`?

victorjulien:
> > In the RFC https://datatracker.ietf.org/doc/html/rfc3546
> > `server_name` is the enum value and `ServerName` the name of the structure...
> 
> I think using a name of `server_name` is more consistent with the RFC.

Wonder if we should also make things more compact, e.g. instead of `ApplicationLayerProtocolNegotiation` use `ALPN`

catenacyber:
These wordings come from crate tls-parser... 
Should Suricata define its own wording ?

rust/src/quic/quic.rs
catenacyber: It comes indeed from `quic_var_uint`
But then, we have `if header.length > rest.len() {`
before `let (mut framebuf, next_buf) = rest.split_at(header.length);

Do you want the check to happen earlier ?

jasonish:
> Should Suricata define its own wording ?

I'm not opposed to it.  An upstream library could change its naming which would affect us resulting in a breaking change.

PR #7100, opened 2022-03-03
Warnint util 4516 v7.1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4516

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` for util files

Seems to be uncontroversial part of #7006 as #7081 was
Plus fixes the new warnings introduced meanwhile, like one introduced in https://github.com/OISF/suricata/pull/7052 and the radix changes

PR #7081, opened 2022-02-28
src: use u8_tolower everywhere

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4516

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` about uses of `tolower`

(`tolower` takes an integer ans returns an integer, but really we use it for `u8`)

Should be uncontroversial part of #7006

Modifies #7073 with commit reworded

PR #7052, opened 2022-02-22
detect/ip_proto: Use built-in protcol mapping table.

jlucovsky:
Continuation of #7040

This PR modifies the detect keyword `ip_proto` to use the built-in protocol table instead of using the underlying system's.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5072](https://redmine.openinfosecfoundation.org/issues/5072)

Describe changes:
- Add utility function to validate protocol name

Updates
- Address review comments
- Rework so search for `ip` works (which is an alias for `HOPOPT`).

suricata-verify-pr: 738
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #7102, opened 2022-03-03
Warnint util 4516 v7.2

catenacyber:

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` for util files

Seems to be uncontroversial part of #7006 as #7081 was
Plus fixes the new warnings introduced meanwhile, like one introduced in https://github.com/OISF/suricata/pull/7052 and the radix changes

Replaces #7100 with CIFuzz fix for util-proto-name (aka how to make a complete loop for every uint8_t value)

### DIFF
--- a/src/util-spm-bm.c
+++ b/src/util-spm-bm.c
@@ -38,6 +38,7 @@
 #include "util-debug.h"
 #include "util-error.h"
 #include "util-memcpy.h"
+#include "util-validate.h"
 
 static int PreBmGs(const uint8_t *x, uint16_t m, uint16_t *bmGs);
 static void PreBmBc(const uint8_t *x, uint16_t m, uint16_t *bmBc);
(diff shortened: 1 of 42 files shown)


suricata-qa:

Pipeline 6463

src/util-spm-bm.c
```diff
@@ -267,11 +270,11 @@ static void PreBmGsNocase(const uint8_t *x, uint16_t m, uint16_t *bmGs)
         bmGs[i] = m;
     }
     j = 0;
-    for (i = m - 1; i >= 0; --i) {
-        if (i == -1 || suff[i] == i + 1) {
-            for (; j < m - 1 - i; ++j) {
+    for (i = m; i > 0; --i) {
```
victorjulien, line 273: this looks like the only logic change, and I have a failing test that only happens with this branch + spm-algo=bm, while it works in master or in this branch with spm-algo=hs

Will share the test offline

catenacyber: indeed, I see the bug

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7107

src/util-spm-bm.c
catenacyber: I changed every usage of `i` by 1 except the one `suff[i]`

PR #7101, opened 2022-03-03
Quic ietf 4967 v4

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4967
https://redmine.openinfosecfoundation.org/issues/5143

Describe changes:
- Parses standardized QUIC IETF v1
- bump up rust carte tls_parser already used by rep, so that it can be used by quic
- Rustfmt previous code

suricata-verify-pr: 775
https://github.com/OISF/suricata-verify/pull/775

Replaces https://github.com/OISF/suricata/pull/7095 with review taken into account

Still to do more generally about quic : see https://redmine.openinfosecfoundation.org/issues/4966

Should I tackle https://redmine.openinfosecfoundation.org/issues/5166 ? ie Support previous quic versions like Q039 and Q043
We get the right version for Q050 but rustls cannot then decrypt it...

PR #7106, opened 2022-03-04
Quic ietf 4967 v5

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5143
https://redmine.openinfosecfoundation.org/issues/5166

Describe changes:
- Rustfmt previous code

suricata-verify-pr: 775
https://github.com/OISF/suricata-verify/pull/775

Replaces https://github.com/OISF/suricata/pull/7101 with 
- Q039 support
- move earlier the check against header.length
- use Suricata's own quic_tls_extension_name (more complete including especially `quic_transport_parameters`

Still to do more generally about quic : see https://redmine.openinfosecfoundation.org/issues/4966 (frame support)


victorjulien:
Looking much better
```
 126536 "1"
  58347 "ff00001d"
  21606 "faceb002"
   2189 "Q043"
    180 "Q046"
```
No other versions than these.

victorjulien:
ja3 situation remains unchanged. A single IP in my network has 993 unique values, which seems suspicious. If it is correct, then I seriously question if we should bother with ja3 at all for quic.

catenacyber:
Oops, forgot to update S-V test with new extensions naming

rust/src/quic/frames.rs
```diff
@@ -115,13 +122,184 @@ impl fmt::Display for StreamTag {
...
+                            dash = true;
+                        }
+                        ja3.push_str(&u16::from(*c).to_string());
+                    }
+                    ja3.push_str(",");
+                    let ciphers = ch.ciphers;
+                    let extv = quic_get_tls_extensions(ch.ext, &mut ja3, true);
+                    return Ok((rest, Frame::Crypto(Crypto { ciphers, extv, ja3 })));
+                }
+                ServerHello(sh) => {
```
victorjulien, line 286: In TLS/tcp we only have `ja3` for client hello iirc, for the server side there is `ja3s`. How is that here?

catenacyber: I used only one `ja3`.
I do not understand why there should be a different naming.

The client ja3 has EllipticCurve and EllipticCurvePointFormat data while the server does not have them (as the server is not supposed to send such extensions) cf `quic_tls_ja3_client_extends`

Does that answer your question ?

catenacyber:
>  A single IP in my network has 993 unique values, which seems suspicious. If it is correct, then I seriously question if we should bother with ja3 at all for quic.

I think it is correct.
@victorjulien You can check what Wireshark outputs like `tshark -Tfields -e tls.handshake.ja3_full -r /Users/catena/Downloads/udp443.pcap | grep -v "^$"`

catenacyber:
Needs https://github.com/OISF/suricata-verify/pull/780

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | tlpr1_stats_chk | | |
| .app_layer.flow.quic | 33523 | 39556 | 84.75% |
| .app_layer.tx.quic | 106109 | 4198 | 2527.61% |
| .app_layer.error.quic.parser | 54065 | 43951 | 123.01% | 

Pipeline 6470

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7118

rust/src/quic/frames.rs
victorjulien: at least for tls ja3 is for clients, ja3s for servers

victorjulien: this is not our invention, just part of the official "standard" https://github.com/salesforce/ja3#ja3s

catenacyber: So, you want two signature keywords `ja3` and `ja3s` instead of using `toclient` or `toserver` ? Or ?

victorjulien: I want to it match how we do it for tls, unless there is a good reason not to do that.

catenacyber: Done in https://github.com/OISF/suricata/pull/7130

I find it a bit strange as a code duplication, but I do not think this reason is good enough

PR #7094, opened 2022-03-02
 nfs: Implement frames v8

ghostinthehive:
Previous PR https://github.com/OISF/suricata/pull/7088
Changes from Previous PR:

- Cleanups

Feature [#4872](https://redmine.openinfosecfoundation.org/issues/4872)

Frames:

   - RPC Frames: Generic over TCP/UDP
       - rpc.pdu
       - rpc.hdr
       - rpc.data
       - rpc.creds -- for rpc calls

   - NFSv2, NFSv3
       - nfs.pdu
       - nfs.status -- for nfs responses

   - NFSv4 Only Frames
       - nfs4.pdu
       - nfs4.hdr
       - nfs4.ops -- for compound request/response operations
       - nfs4.status -- for nfs4 responses

RPC tcp/udp frames created with separate registration functions e.g:
`add_rpc_tcp_tc_frames()`
`add_rpc_udp_tc_frames()`

suricata-verfiy-pr: 774

PR #7108, opened 2022-03-04
nfs: Implement frames v9

ghostinthehive:
Previous PR https://github.com/OISF/suricata/pull/7094
Changes from Previous PR:
- Refine `rpc.creds`:
    `rpc.creds` frame has [ creds_flavor, creds_len, creds_buff ]

    e.g: 
    a rule to check for an rpc `creds_flavor == 1` of `UNIX` goes like:
    `alert nfs any any -> any any (flow:to_server; frame:rpc.creds; content:"|00 00 00 01|"; startswith; sid:1;)`
   

Frames:

       - rpc.pdu
       - rpc.hdr
       - rpc.data
       - rpc.creds -- for rpc calls `frame: [ creds_flavor, creds_len, creds_buff ]`

   - NFSv2, NFSv3
       - nfs.pdu

   - NFSv4 Only Frames
       - nfs4.pdu
       - nfs4.hdr

RPC tcp/udp frames created with separate registration functions e.g:
`add_rpc_tcp_tc_frames()`
`add_rpc_udp_tc_frames()`

suricata-verfiy-pr: 777

### DIFF
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -26,6 +26,7 @@ use nom7::{Err, Needed};
 
 use crate::applayer;
 use crate::applayer::*;
+use crate::frames::*;
 use crate::core::*;
 use crate::conf::*;
 use crate::filetracker::*;
@@ -86,6 +87,22 @@ static mut ALPROTO_NFS: AppProto = ALPROTO_UNKNOWN;
  * NFSTransaction where they can be looked up per handle as part of the
  * Transaction lookup.
  */
+
+#[derive(AppLayerFrameType)]
+pub enum NFSFrameType {
+    RPCPdu,
+    RPCHdr,
+    RPCData,
+    RPCCreds, // for rpc calls | rpc.creds [creds_flavor + creds_len + creds]
+
+    NFSPdu,
+    NFSStatus,
+
+    NFS4Pdu,
+    NFS4Hdr,
+    NFS4Ops,
+    NFS4Status,
+}
 
 #[derive(FromPrimitive, Debug, AppLayerEvent)]
 pub enum NFSEvent {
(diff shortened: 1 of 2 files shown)


rust/src/nfs/nfs.rs
```diff
@@ -1112,7 +1228,8 @@ impl NFSState {
                     // go to the next record.
                     match parse_rpc(cur_i, true) {
                         Ok((_, ref rpc_record)) => {
-                            self.process_request_record(rpc_record);
+                            self.add_rpc_tcp_ts_creds(flow, stream_slice, &cur_i[28..], rpc_record.creds_len as i64);
```
victorjulien, line 1231: I think 28 should be a named constant

victorjulien: Lets pass the + 8 (also with named constant) in here as well instead of doing that calc in the function

victorjulien (changes requested):
Almost there I think

rust/src/nfs/nfs.rs
ghostinthehive: > 

should I go with a generic name? e.g `&cur_i[x..]`

ghostinthehive: Ok so this would be for both `rpc_tcp` | `rpc_udp`, would something like `RPC_TCP_PRE_CREDS` do it?

victorjulien: are the values the same for both tcp and udp?

ghostinthehive: no, its 28 for tcp and 24 for udp

ghostinthehive: I meant two constants with a format like `RPC_TCP_PRE_CREDS` | `RPC_UDP_PRE_CREDS`

victorjulien: yes sounds good

ghostinthehive:
Followed by #7111

PR #7117, opened 2022-03-06
Detect generic uint 4112 v1

catenacyber:

Describe changes:

Still TODO:
- remove the C version to use only rust version

### DIFF
--- a/rust/src/detect.rs
+++ b/rust/src/detect.rs
@@ -97,6 +97,32 @@ fn detect_parse_u32_start_greater(i: &str) -> IResult<&str, DetectU32Data> {
     ))
 }
 
+pub fn detect_match_u32(x: &DetectU32Data, val: u32) -> bool {
+    match x.mode {
+        DetectUintMode::DetectUintModeEqual => {
+            if val == x.value {
+                return true;
+            }
+        }
+        DetectUintMode::DetectUintModeLt => {
+            if val < x.value {
+                return true;
+            }
+        }
+        DetectUintMode::DetectUintModeGt => {
+            if val > x.value {
+                return true;
+            }
+        }
+        DetectUintMode::DetectUintModeRange => {
+            if val < x.value && val > x.valrange {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
 pub fn detect_parse_u32(i: &str) -> IResult<&str, DetectU32Data> {
     let (i, u32) = alt((
         detect_parse_u32_start_lesser,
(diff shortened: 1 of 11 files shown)


rust/src/detect.rs
```diff
@@ -0,0 +1,421 @@
...
+pub enum DetectUintMode {
+    DetectUintModeEqual,
+    DetectUintModeLt,
+    DetectUintModeGt,
+    DetectUintModeRange,
+    DetectUintModeNe,
+}
+
+#[derive(Debug)]
+pub struct DetectU32Data {
```
catenacyber, line 34: @jasonish how do you recommend doing this ?
We will need the same structure and code for u8, u16, u32 and u64

catenacyber:
cc @inashivb as well.
Replaced by https://github.com/OISF/suricata/pull/7121

rust/src/detect.rs
jasonish: I guess one could use an enum?

```
enum UintValue {
    U8(u8),
    U16(u6),
    U32(u32),
}

struct DetectUintData {
    pub val: UintValue,
    pub valrange: UintValue,
}
```
it will at least reduce non-common code that is only size specific.

catenacyber: With this, `val` could be u8 and `valrange` u16 ...
I will look into it a bit more, see how other crates handled it...

catenacyber: I see that num-integer (used by suricata) uses macros for such as case cf https://github.com/rust-num/num-integer/blob/master/src/roots.rs#L165

@jasonish is not there a derive thing like `suricata_derive::AppLayerEvent` ?

Or maybe use crate `[num](https://docs.rs/num/0.4.0/num/index.html)::[integer](https://docs.rs/num/0.4.0/num/integer/index.html)::[Integer](https://github.com/OISF/suricata/pull/7117)`?

jasonish: > I see that num-integer (used by suricata) uses macros for such as case cf https://github.com/rust-num/num-integer/blob/master/src/roots.rs#L165
> 
> @jasonish is not there a derive thing like `suricata_derive::AppLayerEvent` ?

No, we'd have to write that.

> 
> Or maybe use crate `https://docs.rs/num/0.4.0/num/integer/trait.Integer.html` ?

If these fit your needs, then I think they are good to use. We already have them in our deps.  I don't think I've used them myself though.

catenacyber: Managing something with rust generics... let's talk in next version of PR ;-)

PR #7120, opened 2022-03-06
Warnint util 4516 v8.1

catenacyber:

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` for output, flow, and app-layer files

Part of #7006 as #7107 was

Changed from #7006 to keep using `FlowState` in function prototype and using an explicit cast afterwards to get rid of warning about implicit integer conversion

### DIFF
--- a/src/flow.c
+++ b/src/flow.c
@@ -580,8 +580,7 @@ void FlowInitConfig(bool quiet)
             FatalError(SC_ERR_FATAL, "Invalid value for flow.hash-size: NULL");
         }
 
-        if (StringParseUint32(&configval, 10, strlen(conf_val),
-                                    conf_val) > 0) {
+        if (StringParseUint32(&configval, 10, (uint16_t)strlen(conf_val), conf_val) > 0) {
             flow_config.hash_size = configval;
         }
     }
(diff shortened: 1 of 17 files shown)


suricata-qa:

Pipeline 6529

src/flow.c
```diff
@@ -580,8 +580,7 @@ void FlowInitConfig(bool quiet)
             FatalError(SC_ERR_FATAL, "Invalid value for flow.hash-size: NULL");
         }
 
-        if (StringParseUint32(&configval, 10, strlen(conf_val),
-                                    conf_val) > 0) {
+        if (StringParseUint32(&configval, 10, (uint16_t)strlen(conf_val), conf_val) > 0) {
```
victorjulien, line 583: while I'm generally not a fan of `size_t`, I wonder if we should use it here as we'd most of the time pass the result of `strlen()`?

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7219

src/flow.c
catenacyber: Done in https://github.com/OISF/suricata/pull/7239

PR #7121, opened 2022-03-06
Detect generic uint 4112 v2

catenacyber:

Describe changes:

Still TODO:
- remove the C version to use only rust version

Replaces #7117 with commit fixing smb test case

### DIFF
--- a/src/detect-engine-uint.c
+++ b/src/detect-engine-uint.c
@@ -503,3 +503,217 @@ DetectU8Data *DetectU8Parse (const char *u8str)
 
     return u8d;
 }
+
+//same as u32 but with u16
+int DetectU16Match(const uint16_t parg, const DetectU16Data *du16)
+{
+    switch (du16->mode) {
+        case DETECT_UINT_EQ:
+            if (parg == du16->arg1) {
+                return 1;
+            }
+            return 0;
+        case DETECT_UINT_LT:
+            if (parg < du16->arg1) {
+                return 1;
+            }
+            return 0;
+        case DETECT_UINT_LTE:
+            if (parg <= du16->arg1) {
+                return 1;
+            }
+            return 0;
+        case DETECT_UINT_GT:
+            if (parg > du16->arg1) {
+                return 1;
+            }
+            return 0;
+        case DETECT_UINT_GTE:
+            if (parg >= du16->arg1) {
+                return 1;
+            }
+            return 0;
+        case DETECT_UINT_RA:
+            if (parg > du16->arg1 && parg < du16->arg2) {
+                return 1;
+            }
+            return 0;
+        default:
+            BUG_ON("unknown mode");
+    }
+    return 0;
+}
+
+static int DetectU16Validate(DetectU16Data *du16)
+{
+    switch (du16->mode) {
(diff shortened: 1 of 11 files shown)


suricata-qa:

Pipeline 6517

src/detect-engine-uint.c
```diff
@@ -503,3 +503,217 @@ DetectU8Data *DetectU8Parse (const char *u8str)
...
+            if (du16->arg1 == UINT16_MAX) {
+                return 1;
+            }
+            break;
+        case DETECT_UINT_RA:
+            if (du16->arg1 >= du16->arg2) {
+                return 1;
+            }
+            // we need at least one value that can match parg > du16->arg1 && parg < du16->arg2
+            if (du16->arg1 + 1 >= du16->arg2) {
```
inashivb, line 616: Check this.

src/detect-engine-uint.c
```diff
@@ -88,6 +88,16 @@ static int DetectU32Validate(DetectU32Data *du32)
                 return 1;
             }
             break;
+        case DETECT_UINT_LTE:
+            if (du32->arg1 == UINT32_MAX) {
```
inashivb, line 92: Having a really hard time understanding these validations. I know it is not wrong since it's already been there for so long but how does checking for the first arg to be `UINT32_MAX` in case of `LTE` help us validate the argument? Also, why do we not check anything w.r.t. `arg2`?

src/detect-engine-uint.c
inashivb, line 616: So, this is for cases like `dsize 1<>2`?
Q: if I have
```
arg1 = 1
arg2 = 2
```
this matches. But, we mention in https://github.com/OISF/suricata/pull/7121/commits/06eef95923158440cc951d4e888774be0309fc11 that a case like that should be invalid as it is an empty range. Should this be allowed then?
Given that we have exclusive end range checks, should we instead check for `arg1 + 2 >= arg2`. That way we ensure at least one element in the range.

src/detect-engine-uint.c
catenacyber: `DETECT_UINT_LTE` and `UINT32_MAX` comes from `<=0xffffffff` which is always true, so we do not need a rule

Arg2 is only used with ranges

src/detect-engine-uint.c
catenacyber: > this is for cases like dsize 1<>2?

Indeed

> a case like that should be invalid as it is an empty range

Indeed

> Should this be allowed then?

It is not allowed. This check precisely does that. 
In this cas arg1=1, arg1+1=2 so arg1>=arg2 is true.
`DetectU16Validate` returns 1 when the signature is invalid

inashivb: > > Should this be allowed then?
> 
> It is not allowed. This check precisely does that. In this cas arg1=1, arg1+1=2 so arg1>=arg2 is true. `DetectU16Validate` returns 1 when the signature is invalid

ooooooh. Read the status code wrong. Sorry.

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7150

PR #7131, opened 2022-03-10
ssl: first pass limit when allocating buffer for certificates

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5188

Describe changes:
- Adds a first time check before allocating much memory to store a certificate

Feedback welcome here

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -1431,6 +1431,10 @@ static int EnsureRecordSpace(SSLStateConnp *curr_connp, const uint8_t * const in
         SCLogDebug("cert_len unknown still, create small buffer to start");
         certs_len = 256;
     }
+    // Limit in a first time allocation for very large certificates
+    if (certs_len > 0x10000 && certs_len > curr_connp->trec_pos + input_len) {
+        certs_len = 0x10000;
+    }
 
     if (curr_connp->trec == NULL) {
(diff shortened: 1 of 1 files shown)


src/app-layer-ssl.c
```diff
@@ -1431,6 +1431,10 @@ static int EnsureRecordSpace(SSLStateConnp *curr_connp, const uint8_t * const in
         SCLogDebug("cert_len unknown still, create small buffer to start");
         certs_len = 256;
     }
+    // Limit in a first time allocation for very large certificates
+    if (certs_len > 0x10000 && certs_len > curr_connp->trec_pos + input_len) {
```
victorjulien, line 1435: should we first check if current input will set the limited value?

Also please use a macro with a clear name, a decimal value and a comment to explain the default

suricata-qa:

Pipeline 6537

src/app-layer-ssl.c
catenacyber: > should we first check if current input will set the limited value?

I am not sure I understand your question.
This is a first check that current input will set the limited value...

That is if not `certs_len > curr_connp->trec_pos + input_len` that means that the current input will cross the 65k boundary, and thus, we do not limit ourselves

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7139

PR #7132, opened 2022-03-10
Coverity fixes v1

catenacyber:
None, not sure it deserves one

Describe changes: patches to make coverity happy
- Prevents a double unlock (by locking again) 
- Prevents a memory leak in xbits parsing (there is no leak currently, but the code pattern may induce it later) 
- Prevents a use of uninitialized memory in debuglog (I think that this code can only be reached if `PKT_IS_IPV6` or `PKT_IS_IPV4` but this makes it robust against future changes)

After this and #7131 (which I think is a real bug this one), there will be no more outstanding issues reported by coverity

### DIFF
--- a/src/alert-debuglog.c
+++ b/src/alert-debuglog.c
@@ -178,6 +178,9 @@ static TmEcode AlertDebugLogger(ThreadVars *tv, const Packet *p, void *thread_da
     } else if (PKT_IS_IPV6(p)) {
         PrintInet(AF_INET6, (const void *)GET_IPV6_SRC_ADDR(p), srcip, sizeof(srcip));
         PrintInet(AF_INET6, (const void *)GET_IPV6_DST_ADDR(p), dstip, sizeof(dstip));
+    } else {
+        snprintf(srcip, sizeof(srcip), "unknown-ip-version");
+        snprintf(dstip, sizeof(dstip), "unknown-ip-version");
     }
 
     MemBufferWriteString(aft->buffer, "SRC IP:            %s\n"
--- a/src/detect-xbits.c
+++ b/src/detect-xbits.c
@@ -365,6 +365,10 @@ int DetectXbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawstr)
             /* modifiers, only run when entire sig has matched */
             SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_POSTMATCH);
             break;
+        default:
+            SigMatchFree(de_ctx, sm);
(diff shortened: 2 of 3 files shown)


suricata-qa:

Pipeline 6536

src/alert-debuglog.c
```diff
@@ -178,6 +178,9 @@ static TmEcode AlertDebugLogger(ThreadVars *tv, const Packet *p, void *thread_da
     } else if (PKT_IS_IPV6(p)) {
         PrintInet(AF_INET6, (const void *)GET_IPV6_SRC_ADDR(p), srcip, sizeof(srcip));
         PrintInet(AF_INET6, (const void *)GET_IPV6_DST_ADDR(p), dstip, sizeof(dstip));
+    } else {
+        snprintf(srcip, sizeof(srcip), "unknown-ip-version");
```
victorjulien, line 182: this is dead code, since there is just one path that can get us here and that checks for ip4/ip6. I prefer to assist static code analysis with assertions instead

src/detect-xbits.c
```diff
@@ -365,6 +365,10 @@ int DetectXbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawstr)
             /* modifiers, only run when entire sig has matched */
             SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_POSTMATCH);
             break;
+        default:
+            SigMatchFree(de_ctx, sm);
```
victorjulien, line 369: more dead code, we can have an assertion here if we must. But I don't like adding code that isn't exercised at all just to please the static code analyzer

src/flow-hash.c
```diff
@@ -813,6 +813,9 @@ Flow *FlowGetFlowFromHash(ThreadVars *tv, FlowLookupStruct *fls,
                 if (f->use_cnt == 0) {
                     if (prev_f == NULL) /* if we have no prev it means new_f is now our prev */
                         prev_f = new_f;
+                    // f got unlocked by TcpReuseReplace and will be unlocked again by
+                    // MoveToWorkQueue
+                    FromHashLockCMP(f);
```
victorjulien, line 818: will review this later

victorjulien (changes requested):

src/alert-debuglog.c
catenacyber: ok, was not sure it was dead code :-)

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7138

PR #7133, opened 2022-03-10
Packet alerts/v9

jufajardini:
Previous PR: https://github.com/OISF/suricata/pull/7000

Describe changes:
- address feedback from previous PRs
- Split @victorjulien commit for optimizing alert queue processing split in two, and elaborated commit messages a bit more
- instead of discarding new alerts once we reach packet_alert_max, check signature priority and keep the highest and newest
- remove duplicate checks when possible, remove needless assigns.

I think this covers @catenacyber concerns expressed in https://github.com/OISF/suricata/pull/6896#discussion_r796064516  

Redmine tickets:
and
https://redmine.openinfosecfoundation.org/issues/5178

suricata-verify-pr: 782
https://github.com/OISF/suricata-verify/pull/782

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -126,12 +126,11 @@ static int PacketAlertHandle(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det
     SCReturnInt(1);
 }
 
-
 /**
  * \brief Check if a certain sid alerted, this is used in the test functions
  *
  * \param p   Packet on which we want to check if the signature alerted or not
- * \param sid Signature id of the signature that thas to be checked for a match
+ * \param sid Signature id of the signature that has to be checked for a match
  *
  * \retval match A value > 0 on a match; 0 on no match
  */
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -154,27 +154,22 @@ int PacketAlertCheck(Packet *p, uint32_t sid)
  * \brief Remove alert from the p->alerts.alerts array at pos
  * \param p Pointer to the Packet
  * \param pos Position in the array
- * \retval 0 if the number of alerts is less than pos
- *         1 if all goes well
  */
-int PacketAlertRemove(Packet *p, uint16_t pos)
+static void PacketAlertRemove(Packet *p, uint16_t pos)
 {
-    uint16_t i = 0;
-    int match = 0;
-
-    if (pos > p->alerts.cnt) {
+    if (pos >= p->alerts.cnt) {
         SCLogDebug("removing %u failed, pos > cnt %u", pos, p->alerts.cnt);
-        return 0;
+        return;
     }
 
-    for (i = pos; i <= p->alerts.cnt - 1; i++) {
-        memcpy(&p->alerts.alerts[i], &p->alerts.alerts[i + 1], sizeof(PacketAlert));
+    const uint16_t after = (p->alerts.cnt - 1) - pos;
+    if (after) {
+        const uint16_t after_start = pos + 1;
+        memmove(p->alerts.alerts + pos, p->alerts.alerts + after_start,
+                (sizeof(PacketAlert) * after));
     }
 
-    // Update it, since we removed 1
     p->alerts.cnt--;
-
-    return match;
 }
 
 /** \brief append a signature match to a packet
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2021 Open Information Security Foundation
+/* Copyright (C) 2007-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -190,7 +190,7 @@ int PacketAlertAppend(DetectEngineThreadCtx *det_ctx, const Signature *s,
 {
     int i = 0;
 
-    if (p->alerts.cnt == PACKET_ALERT_MAX)
+    if (p->alerts.cnt == packet_alert_max)
         return 0;
 
     SCLogDebug("sid %"PRIu32"", s->id);
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -28,6 +28,7 @@
 #include "flow-private.h"
 
 #include "util-profiling.h"
+#include "util-validate.h"
 
 /** tag signature we use for tag alerts */
 static Signature g_tag_signature;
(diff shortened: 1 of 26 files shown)


src/detect-engine-alert.c
```diff
@@ -188,44 +231,54 @@ int PacketAlertRemove(Packet *p, uint16_t pos)
...
+                    p->alerts.alerts[p->alerts.cnt - 1].s->id, s->id, p->pcap_cnt);
+            PacketAlertInsertPos(det_ctx, s, p, tx_id, flags, (p->alerts.cnt - 1));
+        } else if (num_diff < -1) {
+            /* If the new signature's internal id isn't equal/adjacent to the last one from the
+             * queue, find the correct position, to keep queue sorted by rule priority */
+            int16_t i = MemMovePacketAlertQueuePos(p, s);
+            SCLogDebug("Replacing lower priority signature %" PRIu32 " with sid %" PRIu32
+                       " in packet %" PRIu64 "",
+                    p->alerts.alerts[i].s->id, s->id, p->pcap_cnt);
+            PacketAlertInsertPos(det_ctx, s, p, tx_id, flags, i);
```
jufajardini, line 275: Check if this is enough to "push" lower priority alerts out of the queue or if more work is needed here.

src/detect-engine-alert.c
```diff
@@ -357,7 +410,4 @@ void PacketAlertFinalize(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx
     if (p->flow != NULL && p->alerts.cnt > 0) {
         FlowSetHasAlertsFlag(p->flow);
     }
-
 }
-
-
```
jufajardini, line 363: clang doing

catenacyber:
I do not think you wanted `-x` when you ran your cherry-pick of the previous branch ;-)

src/detect-engine-alert.c
```diff
@@ -155,27 +155,70 @@ int PacketAlertCheck(Packet *p, uint32_t sid)
...
+/**
+ * \brief calculate correct position for alert queue insertion, do memmove if needed
+ *
+ * \param p packet
+ * \param s matched signature
+ * \retval int16_t index of position for insertion in the alert queue
+ */
+static int16_t MemMovePacketAlertQueuePos(Packet *p, const Signature *s)
+{
+    int16_t i = 0; /* signed, because we're decrementing */
```
catenacyber, line 185: I would like `I` to be unsigned as every other variable

src/detect-engine-alert.c
```diff
@@ -188,44 +231,54 @@ int PacketAlertRemove(Packet *p, uint16_t pos)
...
-         (a bit ugly with memcpy but we are planning changes here)*/
-        for (i = p->alerts.cnt - 1; i >= 0 && p->alerts.alerts[i].num > s->num; i--) {
-            memcpy(&p->alerts.alerts[i + 1], &p->alerts.alerts[i], sizeof(PacketAlert));
+        SCLogDebug("Reached packet_alert_max.");
+        /* If we reach packet_alert_max, remove lower priority
+         * rules and keep newer, higher priority ones.
+         * If Suri wants to append a signature whose priority is lower than the
+         * ones already queued and we are at packet_alert_max, it isn't queued. */
+        int16_t num_diff = s->num - p->alerts.alerts[p->alerts.cnt - 1].num;
+        if (num_diff == 0 || num_diff == -1) {
```
catenacyber, line 262: Why have `num_diff == 0` ?

src/detect-engine-alert.c
catenacyber: I think this is enough : there is no reference to dangling pointers to free

src/detect-engine-alert.c
jufajardini: because we can have the same signature appended twice to the same packet by different transactions.

jufajardini:
> I do not think you wanted `-x` when you ran your cherry-pick of the previous branch ;-)

Was trying, but not the right occasion to use it, I see :P

jufajardini:
Time to investigate the fuzz failures, now...

src/detect-engine-alert.c
catenacyber: Ok thanks, but in that case, why do you want to keep the newest ?

src/detect-engine-alert.c
jufajardini: This patch leads to 4 unittest failures:
- Test UtilActionTest15
- Test UtilActionTest18 
- Test UtilActionTest20
- Test UtilActionTest22

src/detect-engine-alert.c
jufajardini: thanks :)

src/detect-engine-alert.c
jufajardini: I assumed it would be better to have newer alerts.
It is, to be honest, unclear what is the best approach here. 
According to Victor, this part of the work will probably go to backports, and we would have something more flexible for 7. But at some point, for either approach, we may reach packet-alert-max, and then we have to decide on what to keep. 
We will have stats saying how many alerts were discarded, as well as do something about rules with the `drop` action, but... what to do when we have the same signature, triggered for the same packet, by different transactions? How to decide? Would it be good to have the perspective of threat hunters?

src/detect-engine-alert.c
jufajardini: Nope, I just replaced the conditions for the `for`

inashivb:
Just a git etiquette thing: We should probably keep the original authors as the signing person and add whatever we changed in the commit messages and not the other way around unless the change we made was too considerable and different from the original author ;)

jufajardini:
> Just a git etiquette thing: We should probably keep the original authors as the signing person and add whatever we changed in the commit messages and not the other way around unless the change we made was too considerable and different from the original author ;)

Indeed! I don't know how to do that, to be honest. I had to split the original commit in two, which probably also affected the end result. I did use cherry-pic though, I thought that would allow me to have that, but I was wrong, it seems :P

inashivb:
> > Just a git etiquette thing: We should probably keep the original authors as the signing person and add whatever we changed in the commit messages and not the other way around unless the change we made was too considerable and different from the original author ;)
> 
> Indeed! I don't know how to do that, to be honest. I had to split the original commit in two, which probably also affected the end result. I did use cherry-pic though, I thought that would allow me to have that, but I was wrong, it seems :P

Probably `git commit --amend --author=" Victor Julien <vjulien@oisf.net>"`
Cherry-pick should ideally pick the correct author. :thinking: But, I'm not sure what happened tbh. :woman_shrugging:

jufajardini:
> > > Just a git etiquette thing: We should probably keep the original authors as the signing person and add whatever we changed in the commit messages and not the other way around unless the change we made was too considerable and different from the original author ;)
> > 
> > 
> > Indeed! I don't know how to do that, to be honest. I had to split the original commit in two, which probably also affected the end result. I did use cherry-pic though, I thought that would allow me to have that, but I was wrong, it seems :P
> 
> Probably `git commit --amend --author="Victor Julien <vjulien@oisf.net>"` Cherry-pick should ideally pick the correct author. thinking But, I'm not sure what happened tbh. woman_shrugging

Ah! Thanks, fixing it now :)
Maybe some of the rebases messed up with it. Or maybe I must add some specific config line for when rebasing from within tig? More stuff to learn!

catenacyber (changes requested):
CI failures to investigate

jufajardini:
Replaced by:
https://github.com/OISF/suricata/pull/7192

PR #7139, opened 2022-03-15
ssl: first pass limit when allocating buffer for certificates

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5188

Describe changes:
- Adds a first time check before allocating much memory to store a certificate

Modifies #7131 with using a macro for the Magic number

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -1418,6 +1418,12 @@ static uint32_t GetCertsLen(SSLStateConnp *curr_connp, const uint8_t *input,
     }
 }
 
+// For certificates whose size is bigger than this,
+// we do not allocate all the required memory straight away,
+// to avoid DOS by RAM exhaustion, but we will allocate
+// this memory once a consequent part of the certificate has been seen.
+#define SSL_CERT_MAX_FIRST_ALLOC 65536 // 0x10000
+
 /** \internal
  *  \brief setup or grow the `trec` space in the connp
  */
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 6572

src/app-layer-ssl.c
```diff
@@ -1431,6 +1437,10 @@ static int EnsureRecordSpace(SSLStateConnp *curr_connp, const uint8_t * const in
         SCLogDebug("cert_len unknown still, create small buffer to start");
         certs_len = 256;
     }
+    // Limit in a first time allocation for very large certificates
```
victorjulien, line 1440: I'm struggling a bit with understanding how this will work for actual large records. Are we getting called again in the future? A casual read of the code suggests it actually expects to be called with the full record, incl the starting bytes. If we'd have a record size that is larger than the initial input size, which will almost always be true if it is more than say 4k, do we still handle things correctly? I guess I'd love to see some test cases for this condition.

victorjulien: btw my lack of understanding is more about the original code, but its relevant to judging this PR

catenacyber: > btw my lack of understanding is more about the original code

Oh, I was indeed assuming you were familiar with it...
One of my goals for this PR is to have minimal changes (until we get a rust tls parser)

> Are we getting called again in the future?

Yes ! this function `EnsureRecordSpace` gets called for every packet  which contains bytes of the certificate even if only the 3 first bytes are useful... 🙄

> A casual read of the code suggests it actually expects to be called with the full record

I do not see this... Where do you see this ?

victorjulien: I think I see my error. On consecutive calls it does use the buffered data to get the length.

catenacyber: > On consecutive calls it does use the buffered data to get the length.

Indeed...
So, do you expect something from me ?

victorjulien (approved):

victorjulien:
Merged in #7187, thanks!

PR #7141, opened 2022-03-16
Smtp c refactor baby step 1/v1

inashivb:
Commit afc873fbc8867ce1dfdbb15892bdfedaeb1a1bc3 uses a struct to pass around input. It is cleaner and probably easier to pass back and forth if needed from Rust.
Commit 00e00d63757ab26fc56ae68c4da51323319058c4 makes use of pointer arithmetic to pass around things. This proved to be a disaster in the past when functions were moved to Rust using C pointers. :woman_shrugging: 

Note: this decides how `current_line` should also be moved out cleanly.

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -241,6 +241,8 @@ SMTPConfig smtp_config = { false, { false, false, false, NULL, false, false, 0 }
     STREAMING_BUFFER_CONFIG_INITIALIZER };
 
 static SMTPString *SMTPStringAlloc(void);
+static SMTPInput *SMTPInputAlloc(void);
+static void SMTPInputFree(SMTPInput *);
 
 /**
  * \brief Configure SMTP Mime Decoder by parsing out mime section of YAML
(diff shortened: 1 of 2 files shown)


src/app-layer-smtp.c
```diff
@@ -1302,12 +1306,13 @@ static AppLayerResult SMTPParse(uint8_t direction, Flow *f, SMTPState *state,
...
             if (SMTPProcessReply(state, f, pstate, thread_data) == -1)
                 SCReturnStruct(APP_LAYER_ERROR);
-            res = SMTPGetLine(state);
+            res = SMTPGetLine(state, smtp_input);
         }
         if (res.status == 1)
             return res;
     }
 
+    SMTPInputFree(smtp_input);
```
inashivb: I wonder if this would leak memory in error and incomplete states.. Gotta check.

suricata-qa:

Pipeline 6582

victorjulien:
This can be closed in favor of #7145? I think that is a better approach.

victorjulien:
This can be closed in favor of #7145? I think that is a better approach.

PR #7145, opened 2022-03-17
smtp: bring input out of state to avoid complexity

inashivb:
Create a structure to store input and input length for SMTP which only
consists of primary data types only. This removes the need to store
input in SMTPState and opens path to pass around input and input_len
freely without having to pass the entire SMTPState whenever input is needed.

The struct was created to save the state of input as it is to be used
by SMTPGetLine function in a loop with the updated values each time.

Remove input and input_len from SMTPState and update the tests where
the input_len was checked after parsing.

The pointer approach is shown in https://github.com/OISF/suricata/pull/7141 but it can be difficult to pass things around in Rust that way as there have to be certain updates made to input/input_len.
The idea is to originally have this SMTPInput struct on the Rust side, alloc it there, do stuff there and pass back the value to C if needed.

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -241,6 +241,8 @@ SMTPConfig smtp_config = { false, { false, false, false, NULL, false, false, 0 }
     STREAMING_BUFFER_CONFIG_INITIALIZER };
 
 static SMTPString *SMTPStringAlloc(void);
+static SMTPInput *SMTPInputAlloc(void);
+static void SMTPInputFree(SMTPInput *);
 
 /**
  * \brief Configure SMTP Mime Decoder by parsing out mime section of YAML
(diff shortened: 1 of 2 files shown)


suricata-qa:

Pipeline 6591

src/app-layer-smtp.c
```diff
@@ -1282,33 +1285,43 @@ static AppLayerResult SMTPParse(uint8_t direction, Flow *f, SMTPState *state,
         SCReturnStruct(APP_LAYER_ERROR);
     }
 
-    state->input = input;
-    state->input_len = input_len;
     state->consumed = 0;
     state->direction = direction;
-    AppLayerResult res = SMTPGetLine(state);
+    smtp_input = SMTPInputAlloc();
```
victorjulien, line 1290: I would think this could be a stack allocation instead:
```c
SMTPInput smtp_input = { .data = input, .len = input_len };
```
and then pass that arround? Safes us alloc/free calls.

Am I missing a case where this wouldn't work?

inashivb: I don't think so. I just did not think at all about stack allocation. Thanks a lot!

inashivb: Sorry, didn't try before responding. We update `input->len` in `GetLine`. That does not seem to stay consistent with the updated values with stack allocation.

victorjulien: Hmm not sure I understand. If you pass around `smtp_input` you can still update its contents.

inashivb: tbh I'm not sure either.. What I see is even after updating `input.len` in `GetLine`, it gets back to the initial value in the next call to `GetLine` from `SMTPParse`. It felt like it reverted to the value with which it was initialized in the scope of `SMTPParse`. :thinking:

victorjulien: Can you share the updated code?

inashivb: Here: https://github.com/OISF/suricata/pull/7156

PR #7150, opened 2022-03-18
Detect generic uint 4112 v3

catenacyber:

Describe changes:

Still TODO:
- Commit history likely to rework...

Replaces #7121 with remove the C version to use only rust version

### DIFF
--- a/rust/src/detect.rs
+++ b/rust/src/detect.rs
@@ -16,9 +16,9 @@
  */
 
 use nom7::branch::alt;
-use nom7::bytes::complete::{is_a, tag};
+use nom7::bytes::complete::{is_a, tag, take_while};
 use nom7::character::complete::digit1;
-use nom7::combinator::{map_opt, opt, value, verify};
+use nom7::combinator::{all_consuming, map_opt, opt, value, verify};
 use nom7::error::{make_error, ErrorKind};
 use nom7::Err;
 use nom7::IResult;
@@ -202,6 +202,7 @@ pub fn detect_parse_uint<
         detect_parse_uint_start_equal,
         detect_parse_uint_start_symbol,
     ))(i)?;
+    let (i, _) = all_consuming(take_while(|c| c == ' '))(i)?;
     Ok((i, uint))
 }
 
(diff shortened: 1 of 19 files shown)


catenacyber:
> +576 −1,110

Begins to look good 😊

catenacyber:
I do not know why `sip-method` S-V test failed only once on MacOS (not failing locally)

rust/src/detect.rs
```diff
@@ -0,0 +1,317 @@
...
+    DetectUintModeNe,
+}
+
+#[derive(Debug)]
+#[repr(C)]
+pub struct DetectUintData<T> {
+    pub arg1: T,
+    pub arg2: T,
+    pub mode: DetectUintMode,
+}
```
jasonish, line 46: Nice, I didn't know this could work with cbindgen.

catenacyber: I first tried to use some kind of typedef alias like `pub type DetectUintDataU32 DetectUintData<u32>`, but making cbindgen export types caused other problems...

rust/src/detect.rs
```diff
@@ -0,0 +1,317 @@
...
+
+fn detect_parse_uint_start_equal<T: std::str::FromStr + std::cmp::PartialOrd + num::Bounded>(
```
jasonish, line 48: I wonder if a custom trait would clean up these function definitions.. Something like:
```
pub trait DetectIntType: std::str::FromStr + std::cmp::PartialOrd + num::PrimInt + num::Bounded {}

impl<T> DetectIntType for T where T: std::str::FromStr + std::cmp::PartialOrd + num::PrimInt + num::Bounded {}
```
then most of these methods can use `<T: DetectIntType>` instead of the string of traits.

suricata-qa:

Pipeline 6597

suricata-qa:

Pipeline 6601

victorjulien:
> I do not know why `sip-method` S-V test failed only once on MacOS (not failing locally)

This test has been flaky before.

victorjulien (changes requested):
Needs some further work, see inline discussion between Jason and Philippe

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7167

PR #7156, opened 2022-03-22
smtp: bring input out of state and alloc on stack

inashivb:
Stack alloc approach

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -635,25 +635,25 @@ int SMTPProcessDataChunk(const uint8_t *chunk, uint32_t len,
  * \retval -1 Either when we don't have any new lines to supply anymore or
  *            on failure.
  */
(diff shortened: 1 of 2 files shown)


src/app-layer-smtp.c
```diff
@@ -635,25 +635,25 @@ int SMTPProcessDataChunk(const uint8_t *chunk, uint32_t len,
  * \retval -1 Either when we don't have any new lines to supply anymore or
  *            on failure.
  */
-static AppLayerResult SMTPGetLine(SMTPState *state)
+static AppLayerResult SMTPGetLine(SMTPState *state, SMTPInput input)
```
victorjulien, line 638: `SMTPGetLine` gets a copy of the `SMTPInput` struct, not a reference to the one you set up originally. So any change made to it here won't be visible to the caller.

inashivb: *digging a hole to bury myself*

PR #7155, opened 2022-03-21
Next/20220321/v4

victorjulien:
Staging:
- #7109
- #7126

Additionally, removes dead code from the streaming buffer API to simplify error handling.

Replaces #7154 to fix formatting.

PR #7159, opened 2022-03-22
Ips handling/v6

victorjulien:
Draft work around improving IPS logic:

- allow dropping all packets in a flow when memcap is hit
- log drop reason in eve/drop (https://redmine.openinfosecfoundation.org/issues/5202)
- add commandline options to simulate various exceptions with pcaps
- improve stream handling around "spurious retransmissions"

suricata-verify-pr: 791

(On top of #7155)


victorjulien:
Replaced by #7166

PR #7160, opened 2022-03-22
detect-content: error on single char hex pairs - v2

jasonish:
Fix parsing of content like "|aa b cc|" which was parsed as "|aa bc|"
without error or warning. This will now fail out, requiring all hex
values to be 2 chars.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5201

suricata-verify-pr: 792

### DIFF
--- a/src/detect-content.c
+++ b/src/detect-content.c
@@ -111,6 +111,12 @@ int DetectContentDataParse(const char *keyword, const char *contentstr,
             if (str[i] == '|') {
                 bin_count++;
                 if (bin) {
+                    if (binpos > 0) {
+                        SCLogError(SC_ERR_INVALID_SIGNATURE,
+                                "Incomplete hex code in content - %s. Invalidating signature.",
+                                contentstr);
+                        goto error;
+                    }
                     bin = 0;
                 } else {
                     bin = 1;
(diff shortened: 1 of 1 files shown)


src/detect-content.c
```diff
@@ -3050,6 +3056,22 @@ static int DetectLongContentTest3(void)
     return !DetectLongContentTestCommon(sig, 1);
 }
 
+static int DetectBadBinContent(void)
+{
+    DetectEngineCtx *de_ctx = NULL;
+    de_ctx = DetectEngineCtxInit();
+    FAIL_IF_NULL(de_ctx);
+    de_ctx->flags |= DE_QUIET;
+    FAIL_IF_NOT_NULL(SigInit(
```
victorjulien, line 3065: Prefer `DetectEngineAppendSig` as this use of `SigInit` will leak memory in the success path.

src/detect-content.c
```diff
@@ -3050,6 +3056,22 @@ static int DetectLongContentTest3(void)
     return !DetectLongContentTestCommon(sig, 1);
 }
 
+static int DetectBadBinContent(void)
```
victorjulien, line 3059: should we also have a test sig for the original reporters input? `|22 2 22|`?

victorjulien (changes requested):

PR #7004, opened 2022-02-15
devguide: move into userguide as last chapter - v5

jufajardini:
Describe changes:
- remove mscgen dependency
- check diagram images in (added watermark to them)
- remove script for image generation

Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/4909

Previous PR: https://github.com/OISF/suricata/pull/6975

Preview at (it doesn't render well, but I think that's not the real end result, just some bad config in my rtd)
https://suri-rtd-test.readthedocs.io/en/docs-dev-into-userguide-v5

PR #7161, opened 2022-03-22
devguide: move into userguide as last chapter - v6

jufajardini:
Describe changes:
- update userguide readme to explain diagrams generation via script
- upload diagram images

All attempts to control/compare image changes didn't work from the workflows/github actions. That seems to be because all files have the same timestamp at that point, so comparing which one is older (image file or msc file) doesn't work. @jasonish suggested I moved forwarded without trying to fix that for now.

Link to redmine ticket:

Previous PR: https://github.com/OISF/suricata/pull/7004

Preview at (it doesn't render well, but I think that's not the real end result, just some bad config in my rtd)
https://suri-rtd-test.readthedocs.io/en/docs-dev-into-userguide-v6


jasonish:
I rendered it locally and uploaded here, turned out a bit better: https://codemonkey.net/suricata/devguide/index.html

jasonish (approved):
Looks good to me.  I do worry a bit as the dev guide grows, but maybe that should be dealt with when it does become large.

jufajardini:
> Looks good to me. I do worry a bit as the dev guide grows, but maybe that should be dealt with when it does become large.

One of those issues we celebrate when we have to solve? :P

victorjulien:
Merged in #7187, thanks!

PR #7126, opened 2022-03-08
detect: makes config keyword really require a flow

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4972

Describe changes:
- detect: only apply ConfigApplyTx when relevant (like with a flow)

Completes #7080 with checking the case for `DETECT_CONFIG` with `DETECT_SM_LIST_POSTMATCH`

PR #7180, opened 2022-03-27
detect: do not try iponly match if sig needs flow

catenacyber:

Describe changes:
- detect: only apply ConfigApplyTx when relevant (like with a flow)

Completes #7126

But How come we are executing `IPOnlyMatchPacket` with a signature that has `s->mask & SIG_MASK_REQUIRE_FLOW` ?
Because it is a postmatch keyword ?

### DIFF
--- a/src/detect-engine-iponly.c
+++ b/src/detect-engine-iponly.c
@@ -1107,6 +1107,8 @@ void IPOnlyMatchPacket(ThreadVars *tv,
                     } else if ((s->flags & (SIG_FLAG_DP_ANY|SIG_FLAG_SP_ANY)) != (SIG_FLAG_DP_ANY|SIG_FLAG_SP_ANY)) {
                         SCLogDebug("port-less protocol and sig needs ports");
                         continue;
+                    } else if ((s->mask & SIG_MASK_REQUIRE_FLOW) != 0) {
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 6669

victorjulien:
We run the ip only engine for the first packet of both flow directions.

catenacyber:
> We run the ip only engine for the first packet of both flow directions.

How come `p->flow` is NULL then ? (in `IPOnlyMatchPacket`)

victorjulien:
> > We run the ip only engine for the first packet of both flow directions.
> 
> How come `p->flow` is NULL then ? (in `IPOnlyMatchPacket`)

IP Only also runs for flowless packets.

victorjulien:
The `bt` shows `ConfigApplyTx`, which implies it needs app-layer. While it is possible to have some app-layer at this point (e.g. UDP, TCP with TFO, midstream pickups), we should probably downgrade a "IP only" rule to a regular rule if we need app-layer.

catenacyber:
> we should probably downgrade a "IP only" rule to a regular rule if we need app-layer.

Thanks very much for this idea Victor

Done in https://github.com/OISF/suricata/pull/7191

PR #7197, opened 2022-04-01
Implement transform to match encrypted indicators

satta:
- [X] I have not updated the user guide (in doc/userguide/) to reflect the changes made (and I am also not serious with this PR ofc)

A downside of rules and datasets is that they store data in cleartext, making it possible for anyone who can get access to the rule or dataset to gather the full list of indicators. It would be useful to be able to specify indicators securely in an encrypted form, then transform buffers with observed metadata to match the encrypted form. This way one would never need to expose the original indicators.
This PR introduces an additional transform implementing this behaviour so one can do, for example:
```
alert dns any any -> any any (msg:”dns test”; dns.query; rot13; dataset:isset,dns-secret; sid:1000042; rev:1;)
```

Describe changes:
- Implement transform to allow for matching of encrypted indicators.


victorjulien:
Just in time for the new release we're announcing today, `Suricata NG 2204.1`

jufajardini:
> Just in time for the new release we're announcing today, `Suricata NG 2204.1`

Will it be announced alongside the online store, finally?

catenacyber:
That was elaborate :-)

PR #7200, opened 2022-04-02
dcerpc: store consumed_bytes as i32

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44187

Describe changes:
- store `consumed_bytes` as i32 as it can grow bigger than u16

@inashivb what do you think ?
Why is not the underlying TCP buffering used with `AppLayerResult::incomplete` ? (looks like it is used a bit)

I also think that it should be `u32` instead of `u32` and using `Result<u32, ()>` to get the rust idiomatic style for functions returning either an error or a valid u32

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -301,7 +301,7 @@ pub struct DCERPCState {
     pub buffer_tc: Vec<u8>,
     pub pad: u8,
     pub padleft: u16,
-    pub bytes_consumed: u16,
+    pub bytes_consumed: i32,
     pub tx_id: u64,
     pub query_completed: bool,
     pub data_needed_for_dir: Direction,
@@ -966,7 +966,7 @@ impl DCERPCState {
 
         // Check if header data was complete. In case of EoF or incomplete data, wait for more
         // data else return error
-        if self.bytes_consumed < DCERPC_HDR_LEN && input_len > 0 {
+        if self.bytes_consumed < DCERPC_HDR_LEN.into() && input_len > 0 {
             parsed = self.process_header(buffer);
             if parsed == -1 {
                 self.extend_buffer(buffer, direction);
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 6729

inashivb:
>     * store `consumed_bytes` as i32 as it can grow bigger than u16
> 

Q: why `i32` and not `u32`? Should `consumed_bytes` not always be a positive int?

> 
> @inashivb what do you think ? Why is not the underlying TCP buffering used with `AppLayerResult::incomplete` ? (looks like it is used a bit)

I think I just did not understand it well enough to spot where it could be used..

> 
> I also think that it should be `u32` instead of `u32` and using `Result<u32, ()>` to get the rust idiomatic style for functions returning either an error or a valid u32

Is there a typo in this sentence?

catenacyber:
> why i32 and not u32? Should consumed_bytes not always be a positive int?

Yes that is what I meant with (there was indeed a typo u32/i32)

> I also think that it should be u32 instead of i32 and using Result<u32, ()> to get the rust idiomatic style for functions returning either an error or a valid u32

But currently the code uses i32 to return error with negative values...

inashivb:
> But currently the code uses i32 to return error with negative values...

OK, so I think you are right and we should use `u32` as you suggested. Regarding negative error values and bytes returned in the same var, it does seem like a bad pattern, `Result` would make sense indeed as it would be more idiomatic as you have proposed.

catenacyber:
So, what do you think ?
Is this small change good enough for itself ?
Or should we make a bigger change ? (should we use `AppLayerResult::incomplete` instead of buffering ?)
Who does what ?

inashivb:
> So, what do you think ? Is this small change good enough for itself ?

I think so, yes.

> Or should we make a bigger change ? (should we use `AppLayerResult::incomplete` instead of buffering ?) Who does what ?

Feels like it'll be more of an optimization task. So, we could do it but probably a low priority one? I cannot take it up for a while it seems. Got features and bug fixes on high priority in other areas. A ticket wouldn't hurt though. :wink:

victorjulien (approved):

victorjulien:
Merged in #7244, thanks!

PR #7201, opened 2022-04-02
dcerpc: use wrappingadd for padding parsing

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44192&q=label%3AProj-suricata

Describe changes:
- use `wrappingadd` for padding parsing

It is safe to overflow in this case as we just want the modulo

### DIFF
--- a/rust/src/dcerpc/parser.rs
+++ b/rust/src/dcerpc/parser.rs
@@ -154,7 +154,7 @@ pub fn parse_dcerpc_bindack(i: &[u8]) -> IResult<&[u8], DCERPCBindAck> {
     let (i, _assoc_group) = take(4_usize)(i)?;
     let (i, sec_addr_len) = le_u16(i)?;
     let (i, _) = take(sec_addr_len)(i)?;
-    let (i, _) = cond((sec_addr_len + 2) % 4 != 0, |b| take(4 - (sec_addr_len + 2) % 4)(b))(i)?;
+    let (i, _) = cond((sec_addr_len.wrapping_add(2)) % 4 != 0, |b| take(4 - (sec_addr_len.wrapping_add(2)) % 4)(b))(i)?;
     let (i, numctxitems) = le_u8(i)?;
     let (i, _) = take(3_usize)(i)?; // Padding
     let (i, ctxitems) = count(parse_dcerpc_bindack_result, numctxitems as usize)(i)?;


suricata-qa:

Pipeline 6728

inashivb (approved):
LGTM. Thanks! :)

rust/src/dcerpc/parser.rs
```diff
@@ -154,7 +154,7 @@ pub fn parse_dcerpc_bindack(i: &[u8]) -> IResult<&[u8], DCERPCBindAck> {
     let (i, _assoc_group) = take(4_usize)(i)?;
     let (i, sec_addr_len) = le_u16(i)?;
     let (i, _) = take(sec_addr_len)(i)?;
-    let (i, _) = cond((sec_addr_len + 2) % 4 != 0, |b| take(4 - (sec_addr_len + 2) % 4)(b))(i)?;
+    let (i, _) = cond((sec_addr_len.wrapping_add(2)) % 4 != 0, |b| take(4 - (sec_addr_len.wrapping_add(2)) % 4)(b))(i)?;
```
victorjulien, line 157: i worry we might hide a real issue here. Is having `sec_addr_len` as 65534 or 65535 even valid? Should we instead have ` verify!` on its parsing?

catenacyber: @inashivb do you know about this sec_addr field ?

inashivb: > i worry we might hide a real issue here. Is having `sec_addr_len` as 65534 or 65535 even valid?

Didn't even think about it while reviewing. But, just checked the docs and we have,
```C
 The IDL declaration of the bind_ack PDU is as follows:
    typedef struct {
        .
        .
        u_int32 assoc_group_id;     /* 20:04 returned assoc_group_id */
        port_any_t sec_addr;        /* 24:yy optional secondary address 
                                     * for process incarnation; local port
                                     * part of address only */
        .
    }
```
where, `port_any_t` is defined as

```C
typedef struct {
        u_int16 length;
        char   [size_is(length)] port_spec; /* port string spec */
        } port_any_t;
```

which is also in accordance with what we have in our code,

```rust
pub struct DCERPCBindAck {
    .                                            
    pub sec_addr_len: u16,
    .
}
```

so I'm thinking it is valid?

victorjulien: How does it related the fields before? E.g. would it make sense if it is bigger than a length field or a max pdu size minus what we already consumed?

inashivb: > How does it related the fields before? E.g. would it make sense if it is bigger than a length field or a max pdu size minus what we already consumed?

OK. I was only thinking in terms of int limits. I now understood what you meant. Indeed, it won't make sense and we should verify if it is within max PDU size limits..

catenacyber: Some remarks without knowing everything about DCERPC :
- this is over TCP (so no 65k limit from UDP)
- Fuzzer managed to reach this, meaning it had a 65k payload. If the PDU was smaller, `take(sec_addr_len)(i)?;` should have failed and there is a bigger bug like https://redmine.openinfosecfoundation.org/issues/5018

catenacyber:
Do we really want a ticket for this ?

inashivb:
> Do we really want a ticket for this ?

If you think it'll soon show up in 6 as well, yes for sure. 😉

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7330

catenacyber:
> > Do we really want a ticket for this ?
> 
> If you think it'll soon show up in 6 as well, yes for sure. 😉

Makes sense it will be needed for 6 to avoid CIFuzz issues ;-)

PR #7196, opened 2022-04-01
detect/flowbits: use strtok_r for parsing

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5154

TODOs:
- Missing test for the issue
- Handle spaces around flowbits for ORing e.g. `fb1 | fb2` since it was allowed till now (UTs fail due to this)

PR #7207, opened 2022-04-05
detect/flowbits: use strtok_r for parsing 

inashivb:
https://redmine.openinfosecfoundation.org/issues/5154

Previous PR: https://github.com/OISF/suricata/pull/7196

Changes since v1:
- Special handling for flowbits OR op with spaces
- Test for bug 5154
- Removal of dead pcre regex code

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -212,12 +212,123 @@ int DetectFlowbitMatch (DetectEngineThreadCtx *det_ctx, Packet *p,
     return 0;
 }
 
-static int DetectFlowbitParse(const char *str, char *cmd, int cmd_len, char *name,
-    int name_len)
+static int DetectFlowbitParse(
+        DetectEngineCtx *de_ctx, const char *rawstr, DetectFlowbitsData **cdout)
 {
+    bool cmd_set = false;
+    bool name_set = false;
+    uint8_t cmd = 0;
+    char name[256] = "";
+    char *val = NULL;
+    DetectFlowbitsData *cd = NULL;
+    char copy[strlen(rawstr) + 1];
+    strlcpy(copy, rawstr, sizeof(copy));
+    char *context = NULL;
+    char *token = strtok_r(copy, ",", &context);
+    while (token != NULL) {
+        while (*token != '\0' && isblank(*token)) {
+            token++;
+        }
+        if (strchr(token, '|') == NULL) {
+            val = strchr(token, ' ');
+            if (val != NULL) {
+                *val++ = '\0';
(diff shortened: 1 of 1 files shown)


inashivb:
@catenacyber could you please tell if this fuzzing error is for me to solve?

src/detect-flowbits.c
```diff
@@ -212,13 +207,124 @@ int DetectFlowbitMatch (DetectEngineThreadCtx *det_ctx, Packet *p,
...
+                *val++ = '\0';
+                while (*val != '\0' && isblank(*val)) {
+                    val++;
+                }
+            } else {
+                SCLogDebug("val %s", token);
+            }
+        }
+        if (strlen(token) == 0) {
+            goto next;
```
victorjulien, line 239: should this be an error? It suggests an empty token, right?

src/detect-flowbits.c
```diff
@@ -212,13 +207,124 @@ int DetectFlowbitMatch (DetectEngineThreadCtx *det_ctx, Packet *p,
...
+        DetectEngineCtx *de_ctx, const char *rawstr, DetectFlowbitsData **cdout)
 {
+    bool cmd_set = false;
+    bool name_set = false;
+    uint8_t cmd = 0;
+    char name[256] = "";
+    char *val = NULL;
+    DetectFlowbitsData *cd = NULL;
+    char copy[strlen(rawstr) + 1];
+    strlcpy(copy, rawstr, sizeof(copy));
```
victorjulien, line 220: wonder if things would be simpler during parsing if we would copy only the non-`is_blank()` characters here. Then during tokenizing below we could ignore any spaces related logic.

src/detect-flowbits.c
```diff
@@ -212,13 +207,124 @@ int DetectFlowbitMatch (DetectEngineThreadCtx *det_ctx, Packet *p,
...
+                "idx %" PRIu32 ", cmd %d, name %s", cd->idx, cmd, strlen(name) ? name : "(none)");
+    }
+    *cdout = cd;
+
+    return 0;
+error:
+    if (cd != NULL)
+        DetectFlowbitFree(de_ctx, cd);
+    return -1;
+#if 0
```
victorjulien, line 322: remove

victorjulien (changes requested):

suricata-qa:

Pipeline 6768

src/detect-flowbits.c
inashivb: How do we handle "key<space>value" pairs then?
e.g. `set flowbit1`.

PR #7203, opened 2022-04-04
Packet alerts/v11

jufajardini:
Previous PR: https://github.com/OISF/suricata/pull/7192

Describe changes:
- re-work commit history

Redmine tickets:
https://redmine.openinfosecfoundation.org/issues/4207
and
https://redmine.openinfosecfoundation.org/issues/5178

suricata-verify-pr: 782
https://github.com/OISF/suricata-verify/pull/782

PR #7209, opened 2022-04-05
detect/stats: log out total of discarded alerts - v1

jufajardini:
https://github.com/OISF/suricata/pull/7203 with stats log work on top
- I have NOT updated the user guide YET (follow up work)

https://redmine.openinfosecfoundation.org/issues/5179

Describe changes:
- add a counter for discarded alerts when Suri reaches packet-alert-max
- log that out in our stats

suricata-verify-pr: 804
https://github.com/OISF/suricata-verify/pull/804

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -126,12 +126,11 @@ static int PacketAlertHandle(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det
     SCReturnInt(1);
 }
 
-
 /**
  * \brief Check if a certain sid alerted, this is used in the test functions
  *
  * \param p   Packet on which we want to check if the signature alerted or not
- * \param sid Signature id of the signature that thas to be checked for a match
+ * \param sid Signature id of the signature that has to be checked for a match
  *
  * \retval match A value > 0 on a match; 0 on no match
  */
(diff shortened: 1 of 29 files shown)


src/detect-engine-alert.c
```diff
@@ -280,8 +279,10 @@ int PacketAlertAppend(DetectEngineThreadCtx *det_ctx, const Signature *s,
                        ") in packet %" PRIu64 "",
                     p->alerts.alerts[i].s->id, p->alerts.alerts[i].num, s->id, s->num, p->pcap_cnt);
             PacketAlertInsertPos(det_ctx, s, p, tx_id, flags, i);
-            // TODO log to stats signature id that was discarded/replaced
         }
+        /* Update so we can log to stats later. If we got here, some alert was
+        discarded */
+        p->alerts.discarded++;
```
jufajardini, line 285: It seems to me that we have at least two situations:
- a signature in the alert queue is replaced;
- the new match happens for a signature with a lower priority id, so it will be discarded and nothing changes in the alert queue.
Do we want to log these two as the same thing in the stats log?
If yes, then I think this patch is enough. If not, then more work has to be done.

jufajardini:
Closing this for now, as the underlying approach may change, which may or may not affect how we save some stats - and also affect which work should be merged before we could have this one working, too.

PR #7216, opened 2022-04-08
ike: fix integer underflow in parse_proposal

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44459

Describe changes:
- ike: fix integer underflow in parse_proposal

cc @chifflier

### DIFF
--- a/rust/src/ike/parser.rs
+++ b/rust/src/ike/parser.rs
@@ -304,7 +304,7 @@ pub fn parse_proposal(i: &[u8]) -> IResult<&[u8], ProposalPayload> {
     let (i, number_transforms) = be_u8(i)?;
     let (i, spi) = take(spi_size as usize)(i)?;
     let (i, payload_data) = cond(
-        (start_i.len() as i16 - 4) - spi_size as i16 >= 0,
+        (start_i.len() - 4) >= spi_size.into(),
         |b| take((start_i.len() - 4) - spi_size as usize)(b)
         )(i)?;
     let payload = ProposalPayload {


rust/src/ike/parser.rs
```diff
@@ -304,7 +304,7 @@ pub fn parse_proposal(i: &[u8]) -> IResult<&[u8], ProposalPayload> {
     let (i, number_transforms) = be_u8(i)?;
     let (i, spi) = take(spi_size as usize)(i)?;
     let (i, payload_data) = cond(
-        (start_i.len() as i16 - 4) - spi_size as i16 >= 0,
+        (start_i.len() - 4) >= spi_size.into(),
```
victorjulien, line 307: can we be sure `start_i.len()` >= 4? IIRC it will be `usize` so -4 might underflow it?

victorjulien: answering my own question: yes we can be sure as we took 4 u8's.

victorjulien: so is the issue here that we might be called with a larger packet than `i16` can express?

catenacyber: >so is the issue here that we might be called with a larger packet than i16 can express?

Indeed, I can craft you a pcap with a UDP packet whose payload length is 0xFFFF if you need :-p

The issue here is not to shrink the integer type when it is not needed
like what was fixed in commit 6f03ee2e47d331aa29524d81777c970415b817f1

victorjulien (approved):

suricata-qa:

Pipeline 6858

victorjulien (approved):

victorjulien:
Merged in #7244, thanks!

PR #7217, opened 2022-04-08
dns: store transaction in a VecDeque

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44307

Describe changes:
- dns: store transaction in a VecDeque

@jasonish what do you think ?
Could you test the perf of this ?

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -311,7 +311,7 @@ pub struct DNSState {
     pub tx_id: u64,
 
     // Transactions.
-    pub transactions: Vec<DNSTransaction>,
+    pub transactions: VecDeque<DNSTransaction>,
 
     pub events: u16,
 
@@ -322,7 +322,8 @@ pub struct DNSState {
 
 impl State<DNSTransaction> for DNSState {
     fn get_transactions(&self) -> &[DNSTransaction] {
-        &self.transactions
+        let (r, _) = self.transactions.as_slices();
+        return r;
     }
 }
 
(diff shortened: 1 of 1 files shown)


rust/src/dns/dns.rs
```diff
@@ -322,7 +322,8 @@ pub struct DNSState {
 
 impl State<DNSTransaction> for DNSState {
     fn get_transactions(&self) -> &[DNSTransaction] {
-        &self.transactions
+        let (r, _) = self.transactions.as_slices();
+        return r;
```
jasonish, line 326: This would only return a portion of the transactions. You probably have to use `make_contiguous`.  Can you test that with your fuzzing?

suricata-qa:

Pipeline 6857

rust/src/dns/dns.rs
catenacyber: Strange that it did not fail CI if it was really needed...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7238

rust/src/dns/dns.rs
catenacyber: By the way, `make_contiguous` needs to make `self` mutable...

jasonish: > Strange that it did not fail CI if it was really needed...

If all removes are from the front, and given all adds are to the back, in most cases I think we'd be in a state where we might only use one half?

Still worth figuring out if we're to use a VecDeque, as mentioned I see it having a negligible affect in normal conditions, however the fuzz appears to produce a worst case which would be nice to fixup.

PR #7218, opened 2022-04-08
template: convert to SCLogDebug

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44427

Describe changes:
- template: convert to SCLogDebug

Is there a reason it was `SCLogNotice` ?

### DIFF
--- a/src/app-layer-template.c
+++ b/src/app-layer-template.c
@@ -18,7 +18,6 @@
 /*
  * TODO: Update \author in this file and app-layer-template.h.
  * TODO: Implement your app-layer logic with unit tests.
- * TODO: Remove SCLogNotice statements or convert to debug.
  */
 
 /**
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 6856

jasonish:
> Is there a reason it was `SCLogNotice` ?

Mostly to aid in development, it helps to be a little more verbose during initial development. I think this partly comes from the last dev training as well, it helps trace the execution while getting started.

src/app-layer-template.c
```diff
@@ -490,7 +486,7 @@ void RegisterTemplateParsers(void)
 
     if (AppLayerParserConfParserEnabled("tcp", proto_name)) {
 
-        SCLogNotice("Registering Template protocol parser.");
+        SCLogDebug("Registering Template protocol parser.");
```
jasonish, line 489: I think this one should at least stay.

jasonish (changes requested):
I think its help to at least leave the registration one during development, so you know all is good.  And in that case we should probably keep the related TODO around.

catenacyber:
Thanks Jason, continued in https://github.com/OISF/suricata/pull/7237

PR #7222, opened 2022-04-08
eve/stats: Fix ASAN-detected overflow

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5198](https://redmine.openinfosecfoundation.org/issues/5198)

The buffer overflow occurred due to the stats output not being set up with a valid context useful for threaded output. The stats output is created by the main thread - before the stats thread or any other Suricata thread is created.

The hash table was introduced to provide unique mappings for the thread to thread-slot relationship to guarantee that any thread using `LogFileEnsureExists` uses the proper slot.

Describe changes:
- Hashtable for thread/file slot mappings
- Ensure slot exists for stats thread

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/output-json-common.c
+++ b/src/output-json-common.c
@@ -53,7 +53,7 @@ OutputJsonThreadCtx *CreateEveThreadCtx(ThreadVars *t, OutputJsonCtx *ctx)
         goto error;
     }
 
-    thread->file_ctx = LogFileEnsureExists(ctx->file_ctx, t->id);
+    thread->file_ctx = LogFileEnsureExists(ctx->file_ctx);
     if (!thread->file_ctx) {
         goto error;
     }
@@ -119,7 +119,7 @@ TmEcode JsonLogThreadInit(ThreadVars *t, const void *initdata, void **data)
     }
 
     thread->ctx = ((OutputCtx *)initdata)->data;
-    thread->file_ctx = LogFileEnsureExists(thread->ctx->file_ctx, t->id);
+    thread->file_ctx = LogFileEnsureExists(thread->ctx->file_ctx);
     if (!thread->file_ctx) {
         goto error_exit;
     }
--- a/src/output-json-stats.c
+++ b/src/output-json-stats.c
@@ -395,7 +395,7 @@ static OutputInitResult OutputStatsLogInitSub(ConfNode *conf, OutputCtx *parent_
         return result;
     }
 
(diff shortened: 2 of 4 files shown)


suricata-qa:

Pipeline 6849

victorjulien:
I'm not too excited about adding another thread hash to essentially assign another id? Why not use the logic in `TmThreadsRegisterThread`?

jlucovsky:
> I'm not too excited about adding another thread hash to essentially assign another id? Why not use the logic in `TmThreadsRegisterThread`?

With 16 worker threads, there are approximately 475 calls to `LogFileEnsureExists` ... each of those calls must first locate the slot. Using a hash table for the slot to thread mapping makes is O(1) whereas a linear scan to find the slot would be O(n).

victorjulien:
> > I'm not too excited about adding another thread hash to essentially assign another id? Why not use the logic in `TmThreadsRegisterThread`?
> 
> With 16 worker threads, there are approximately 475 calls to `LogFileEnsureExists` ... each of those calls must first locate the slot. Using a hash table for the slot to thread mapping makes is O(1) whereas a linear scan to find the slot would be O(n).

I'm not questioning the value of a hash, just the logic of adding a new hash table that essentially looks up by a unique identifier. (OS thread id here, vs having a internal thread id)

jlucovsky:
> > > I'm not too excited about adding another thread hash to essentially assign another id? Why not use the logic in `TmThreadsRegisterThread`?
> > 
> > 
> > With 16 worker threads, there are approximately 475 calls to `LogFileEnsureExists` ... each of those calls must first locate the slot. Using a hash table for the slot to thread mapping makes is O(1) whereas a linear scan to find the slot would be O(n).
> 
> I'm not questioning the value of a hash, just the logic of adding a new hash table that essentially looks up by a unique identifier. (OS thread id here, vs having a internal thread id)

The stats output is initialized before any threads are created so there's not an internal thread id available. The code in 6.0.x and master use the internal thread id value for the slot -- O(1). The buffer overflow detected by ASAN was due to the stats thread using a file context that wasn't prepared properly (`LogFileEnsureExists` wasn't used). To reliably set up the stats output, I converted to the hash table since the internal thread id wasn't available in 100% of the cases.

PR #7234, opened 2022-04-14
Ips handling/v17

victorjulien:
continuation of #7166

PR #7243, opened 2022-04-15
App layer error close txs 4318 v5

catenacyber:

Describe changes:

I am more and more convinced that there is no good way to test this as app-layer parsers should not return error...

Modifies #7076 by using a flag on the directional stream instead of the bidirectional session, so as to still have detection run for both directions


catenacyber:
Converting to draft as I am not sure this is right... (because we are missing the good test I guess)

Maybe `AppLayerParserTransactionsCleanup` should take into account the direction, so that if the tcp session is disrupted, `xx_disrupt_flags = FlowGetDisruptionFlags(f, xx);` gets the disrupted flags in the current direction, but not in the opposite direction, so that we run detection for both directions, but only once...

victorjulien:
For testing, using this (draft) work will be useful I think https://github.com/OISF/suricata/pull/7234/commits/ff3219240a0d0eb6ddcd53b0c147836516834158

suricata-qa:

Pipeline 6928

suricata-qa:
Warning: no commits in this PR have specified the following ticket(s):
- 4318 - https://redmine.openinfosecfoundation.org/issues/4318

Please update the commit(s) and submit a new PR.

PR #7268, opened 2022-04-19
6.0.x - 5018 - mqtt - backport - v1/

jasonish:
Minimal backport of https://redmine.openinfosecfoundation.org/issues/5018 to 6.0.x.  Passes S-V test from https://github.com/OISF/suricata-verify/pull/737.

@satta @catenacyber Is there more to this that should be considered for 6.0.x?

Ticket: https://redmine.openinfosecfoundation.org/issues/5059


satta:
It's pretty minimal indeed, but should address the issue in that ticket. The original patch went a bit beyond that though and added a global `complete` that wraps more potential cases, not just properties parsing. Is there a reason why this patch is kept so small?

jasonish:
> It's pretty minimal indeed, but should address the issue in that ticket. The original patch went a bit beyond that though and added a global `complete` that wraps more potential cases, not just properties parsing. Is there a reason why this patch is kept so small?

The backport from master to 6.0.x is quite large given the nom7 changes and the move from macros to functions.  I wasn't fully clear on what all needed to be moved back.

victorjulien:
Merged in #7270, thanks!

PR #7278, opened 2022-04-20
6.0.x: app-layer: better warning message when enabling by default

jasonish:
The warning message suggests that the configuration section doesn't
exist if when it does, but the "enabled" flag is not set. Clarify the
warning message that the enable status is not set.


suricata-qa:

Pipeline 7059

PR #7212, opened 2022-04-06
userguide: document packet alert queue - v1

jufajardini:
This work goes with 
- packet alert max and packet alert queue patches (https://github.com/OISF/suricata/pull/7211)
- stats log about discarded alerts (https://github.com/OISF/suricata/pull/7209)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
- https://redmine.openinfosecfoundation.org/issues/5179
- https://redmine.openinfosecfoundation.org/issues/5178
- https://redmine.openinfosecfoundation.org/issues/4207

Describe changes:
- Add a section about packet alert max config
- Add subsection about packet alert queue behavior (especially when an overflow happens)
- Add subsection about discarded alerts in case of packet alert queue overflow

PR #7211, opened 2022-04-06
Packet alerts/v12

jufajardini:
Previous PR: https://github.com/OISF/suricata/pull/7203

Describe changes:
- rename functions to follow naming conventions
- remove commit with changes to unit tests, since it has been cherry-picked already

Redmine tickets:
https://redmine.openinfosecfoundation.org/issues/4207
and
https://redmine.openinfosecfoundation.org/issues/5178

suricata-verify-pr: 782
https://github.com/OISF/suricata-verify/pull/782

PR #7284, opened 2022-04-21
Alert queue optimizations - v1

jufajardini:
Related to (and building upon work from) https://github.com/OISF/suricata/pull/7211

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4943

Describe changes:
- move pre-processing of the alert queue (PacketAlert) to the `DetectEngineThreadCtx`. This queue grows dynamically as needed. This means appending and sorting the alerts queue from the Detect Engine Thread context, and passing this queue to `PacketAlertFinalize` to apply thresholding, action and alert handling before storing the final PacketAlerts queue in the Packet structure, which will also enforce packet_alert_max limits as defined in our configuration (or default values)
- remove PacketAlert functions that become obsolete with this patch
- add stats about discarded alerts (when we reach packet max alert and the matched signature has a lower priority) and suppressed alerts (when the rule matches, but it's a `noalert` rule or otherwise doesn't reach the threshold to be appended)

Leaving this as draft because I don't know if it will be considered a good idea to be addressing so many different things in a single PR, but local unittests and respective SV tests pass, and commit history is somewhat ready, too, if approved.

Once this work gets approved, then we can see if more work needs to be done in #7212 

suricata-verify-pr: 808
https://github.com/OISF/suricata-verify/pull/808


src/detect.c
```diff
@@ -828,12 +833,17 @@ static DetectRunScratchpad DetectRunSetup(
 
 #ifdef UNITTESTS
     p->alerts.cnt = 0;
+    p->alerts.discarded = 0;
+    p->alerts.suppressed = 0;
```
jufajardini, line 837: Wasn't sure if these should stay here or go with the others (outside unittests definition)

src/detect.c
```diff
@@ -1631,7 +1728,7 @@ static void DetectRunFrames(ThreadVars *tv, DetectEngineCtx *de_ctx, DetectEngin
                     det_ctx->frame_id = frame->id;
                     SCLogDebug(
                             "%p/%" PRIi64 " sig %u (%u) matched", frame, frame->id, s->id, s->num);
-                    PacketAlertAppend(det_ctx, s, p, 0, alert_flags); // TODO tx id frame field
```
jufajardini, line 1634: Is this TODO about tx_id frame field still important? mentioning so it won't get lost...

src/detect.c
```diff
@@ -1001,6 +1017,72 @@ static int RuleMatchCandidateTxArrayExpand(DetectEngineThreadCtx *det_ctx, const
...
+
+/** \internal
+ */
+static void AlertQueueExpand(DetectEngineThreadCtx *det_ctx)
+{
+    det_ctx->alert_queue_capacity *= 2;
+    void *tmp_queue = SCRealloc(
+            det_ctx->alert_queue, (size_t)(sizeof(PacketAlert) * det_ctx->alert_queue_capacity));
+    if (tmp_queue == NULL) {
+        return;
```
jufajardini, line 1047: Should we throw some error here?

jufajardini (commented):
Added a few inline questions about things I wasn't sure.

src/detect-engine-alert.c
```diff
@@ -316,36 +238,42 @@ void PacketAlertFinalize(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx
...
         if (res == 0 || res == 2 || (s->flags & SIG_FLAG_NOALERT)) {
-            PacketAlertRemove(p, i);
-
-            if (p->alerts.cnt == 0)
+            /* we will not copy this to the AlertQueue */
+            p->alerts.suppressed++;
+        } else if (p->alerts.cnt < packet_alert_max) {
+            memcpy(&p->alerts.alerts[p->alerts.cnt], &det_ctx->alert_queue[i], sizeof(PacketAlert));
+            SCLogDebug("Appending sid %" PRIu32 " alert to Packet::alerts at pos %u", s->id, i);
+            if (has_pass_action) {
```
victorjulien, line 269: could we just do `if(PacketTestAction(p, ACTION_PASS)) {` here?

src/detect.c
victorjulien: are they reset elsewhere as well?

src/detect.c
```diff
@@ -1001,6 +1017,72 @@ static int RuleMatchCandidateTxArrayExpand(DetectEngineThreadCtx *det_ctx, const
...
+    }
+    det_ctx->alert_queue = tmp_queue;
+    SCLogDebug("Alert queue size doubled. %u, bytes: %" PRIuMAX "", det_ctx->alert_queue_capacity,
+            (uintmax_t)(sizeof(PacketAlert) * det_ctx->alert_queue_capacity));
+}
+
+/**
+ * \brief Append signature to local packet alert queue for later preprocessing
+ */
+void DetectEngineThreadCtxAlertQueueAppend(DetectEngineThreadCtx *det_ctx, const Signature *s,
```
victorjulien, line 1057: Just `AlertQueueAppend`?

src/detect.c
```diff
@@ -1019,6 +1101,21 @@ DetectRunTxSortHelper(const void *a, const void *b)
         return s0->id > s1->id ? 1 : -1;
 }
 
+/** \internal
+ * \brief sort helper for sorting alerts by priority
+ *
+ * The id field is set from Signature::num, so we have higher priority (lower ids) first
+ */
+static int AlertQueueSortHelper(const void *a, const void *b)
```
victorjulien, line 1109: should maybe make sure that if they are equal we sort by PacketAlert::txid?

src/detect.c
victorjulien: ya I think so

victorjulien (changes requested):
Looks pretty good already. Much simpler.

src/detect.c
jufajardini: okey dokey!

src/detect.c
jufajardini: Can it be made into a ticket?

src/decode.c
```diff
@@ -74,6 +74,26 @@ extern bool stats_decoder_events;
 extern const char *stats_decoder_events_prefix;
 extern bool stats_stream_events;
 uint8_t decoder_max_layers = PKT_DEFAULT_MAX_DECODED_LAYERS;
+uint16_t packet_alert_max = PACKET_ALERT_MAX;
+
+/**
+ * \brief Initialize PacketAlerts with dynamic alerts array size
+ *
+ */
+PacketAlert *CreatePacketAlert(void)
```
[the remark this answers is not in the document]
jufajardini: thanks. when work is lost between changes x_x

src/detect.c
jufajardini: In `PACKET_RELEASE_REFS(p)`

src/detect-engine-alert.c
jufajardini: Likely. Sometimes, it's hard to see simple things xD I'll try it out.

src/detect.c
jufajardini: I think we should, I was thinking something like what we have in https://github.com/OISF/suricata/blob/master/src/detect.c#L992

jufajardini: Decided to go with a warning, for now.

src/detect.c
jufajardini: left a comment around the same place, for now

suricata-qa:
ERROR:
 

Pipeline 7137

jufajardini:
Feedback incorporated in: https://github.com/OISF/suricata/pull/7290

PR #7288, opened 2022-04-21
6.0.x: scripts/bundle: use git instead of tar.gz

jasonish:
To better fit with our current CI processes, use git to clone the
suricata-update and libhtp dependencies.  The requirements.txt file has
been modified to take a repo URL and a `-b` command line option for tag
or branch.

For the 6.0.x branch we will use the libhtp 0.5.x branch and the
suricata-update 1.2.4 tag.

Also allows for repo and branch names to be overrided with environment
variables:
- SU_REPO
- SU_BRANCH
- LIBHTP_REPO
- LIBHTP_BRANCH

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,69 @@
+6.0.5 -- 2022-04-21
+
+Security #4888: ftp: SEGV at flow cleanup due to protocol confusion
+Security #5026: ftp: GetLine function buffers data indefinitely if 0x0a was not found in the frag'd input
+Security #5027: smtp: GetLine function buffers data indefinitely if 0x0a was not found int the frag'd input
+Bug #4467: dataset file not written when run as user
+Bug #4630: Protocol detection : confusion with SMB in midstream
+Bug #4677: Configuration test mode succeeds when reference.config file contains invalid content
+Bug #4744: Warn if Absent app-layer protocol is always enabled by default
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -0,0 +1,7 @@
+# Specify libhtp and suricata-update requirements.
+#
+# Format:
+#
+#   name {repo} {branch|tag}
+libhtp https://github.com/OISF/libhtp 0.5.x
```
victorjulien, line 6: should this be 0.5.40?

jasonish: So leave as-is.  Updating will become part of the versioning process I guess.  So its pinned to a tag in a Suricata tag.

victorjulien: never really looked into how, but for libhtp we specify the version only here https://github.com/OISF/libhtp/blob/0.5.x/VERSION

jasonish: > can we use this file as the source of the htp version check in configure?

We can certainly use some `sed` to extract it. But how do the version check if its simply `0.5.x`? Just allow any libhtp version that matches 0.5.?

victorjulien: possibly something like what we have here? https://github.com/OISF/suricata/blob/master/configure.ac#L1526

jasonish:
Replaced by https://github.com/OISF/suricata/pull/7361.

PR #7291, opened 2022-04-22
cbindgen: handle version to stderr change

victorjulien:


jasonish:
Shouldn't be needed now. A fresh build of `cbindgen` pulls in https://github.com/clap-rs/clap/releases/tag/v3.1.12 which has this issue fixed via https://github.com/clap-rs/clap/pull/3646.

But it probably doesn't hurt to have it either.

PR #7306, opened 2022-04-25
ftp: truncate first segment if over max length - v3

jasonish:
The first segment was not limited to the configured maximum line length
allowing it to be up to 65k. This could result in the next input length
being negative, which while handled properly by the code, did trigger a
debug validation assertion.

The fix is to be consistent and apply the limit to the first segment as
well, which does ensure the input_len could never be less than 0.

Ticket https://redmine.openinfosecfoundation.org/issues/5281

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -387,6 +387,9 @@ static int FTPGetLineForDirection(FtpState *state, FtpLineState *line_state)
         }
     }
 
+    /* Should be guaranteed by the caller. */
+    DEBUG_VALIDATE_BUG_ON(state->input_len <= 0);
+
     uint8_t *lf_idx = memchr(state->input, 0x0a, state->input_len);
 
     if (lf_idx == NULL) {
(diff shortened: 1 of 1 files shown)


suricata-qa:
Information:
 

|--|--|--|--| 
| | tlpr1_stats_chk | | |
| .flow.memuse | 509338688 | 539921728 | 94.34% | 

Pipeline 7170

src/app-layer-ftp.c
```diff
@@ -397,13 +400,18 @@ static int FTPGetLineForDirection(FtpState *state, FtpLineState *line_state)
          * if we see fragmentation then it's definitely something you
          * should alert about */
         if (line_state->current_line_db == 0) {
-            line_state->db = FTPMalloc(state->input_len);
+            int32_t input_len = state->input_len;
```
catenacyber, line 403: Why is `input_len` signed ?

jasonish: Only because its source, `state->input_len` is in32_t.

catenacyber: Indeed, but why is `state->input_len` signed in the first place ?

jasonish: > Indeed, but why is `state->input_len` signed in the first place ?

```
 ¯\_(ツ)_/¯
```

jasonish: > So, I take this is not the good opportunity fo fix this... Should I create a ticket for it for later ?

Yeah, I'd rather keep the fix as focused as possible for now.

catenacyber: https://redmine.openinfosecfoundation.org/issues/5311

victorjulien:
Merged in #7349, thanks!

PR #7309, opened 2022-04-26
dns: frame support - v3

jasonish:
- frames (rust): method to create frames
- dns: add pdu frame

Ticket: https://redmine.openinfosecfoundation.org/issues/4984

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -25,6 +25,7 @@ use std::collections::VecDeque;
 use crate::applayer::*;
 use crate::core::{self, *};
 use crate::dns::parser;
+use crate::frames::Frame;
 
 use nom7::{Err, IResult};
 use nom7::number::streaming::be_u16;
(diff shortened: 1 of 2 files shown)


rust/src/dns/dns.rs
```diff
@@ -492,8 +519,11 @@ impl DNSState {
             SCLogDebug!("[request] Have {} bytes, need {} to parse",
                         cur_i.len(), size + 2);
             if size > 0 && cur_i.len() >= size + 2 {
-                let msg = &cur_i[0..(size + 2)];
-                if self.parse_request(&msg[2..]) {
+                let msg = &cur_i[2..(size + 2)];
+                if cfg!(not(test)) {
```
victorjulien, line 523: is there a more graceful way to handle this? Ideally we'd not need it at all

jasonish: > is there a more graceful way to handle this? Ideally we'd not need it at all

We could take the C functions in this code path and convert them to function pointers set via `SuricataContext` then make these calls no-ops when called from tests.  Not  sure that qualifies as more graceful though.

victorjulien: Can we hide the `if cfg!(not(test)) {` inside the `Frame` implementation somehow, so it won't pollute the parser logic?

jasonish: Yeah, makes sense.

jasonish:
https://github.com/OISF/suricata/pull/7316

PR #7313, opened 2022-04-26
app-layer: more generic state trait; dns: use vecdeque - v1

jasonish:
First, update the state trait to be more generic and update all parsers. This
allows for a VecDeque to be used as the transaction collection.

Second, update DNS to use a VecDeque.

Ticket: https://redmine.openinfosecfoundation.org/issues/5278
Ticket: https://redmine.openinfosecfoundation.org/issues/5277

### DIFF
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -1,4 +1,4 @@
-/* Copyright (C) 2017-2021 Open Information Security Foundation
+/* Copyright (C) 2017-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -535,14 +535,17 @@ pub trait Transaction {
 }
 
 pub trait State<Tx: Transaction> {
-    fn get_transactions(&self) -> &[Tx];
+    /// Return the number of transactions in the state's transaction collection.
+    fn get_transaction_count(&self) -> usize;
+
+    /// Return a transaction by its index in the container.
+    fn get_transaction_by_index(&self, index: usize) -> Option<&Tx>;
 
     fn get_transaction_iterator(&self, min_tx_id: u64, state: &mut u64) -> AppLayerGetTxIterTuple {
         let mut index = *state as usize;
-        let transactions = self.get_transactions();
-        let len = transactions.len();
+        let len = self.get_transaction_count();
         while index < len {
-            let tx = &transactions[index];
+            let tx = self.get_transaction_by_index(index).unwrap();
             if tx.id() < min_tx_id + 1 {
                 index += 1;
                 continue;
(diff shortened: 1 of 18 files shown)


jasonish:
This ticket is currently marked for backport.  If everythign looks OK, I'll take care of the 6.0 backport. Should determine its priority for a 5.0 backport as it may be a little more work, though could do a one-off for DNS.

rust/src/applayer.rs
```diff
@@ -535,14 +535,17 @@ pub trait Transaction {
 }
 
 pub trait State<Tx: Transaction> {
-    fn get_transactions(&self) -> &[Tx];
+    /// Return the number of transactions in the state's transaction collection.
+    fn get_transaction_count(&self) -> usize;
```
victorjulien, line 539: in the C API we have a `StateGetTxCnt` callback (which is misnamed. It should be `StateGetTxMaxId`). How does it relate to this function/trait logic?

catenacyber:
This also solves https://redmine.openinfosecfoundation.org/issues/5271 ;-)

catenacyber:
Should other protocols like mqtt, http2, dcerpc, rdp, pgsql be part of this PR ?

suricata-qa:
Information:
 

|--|--|--|--| 
| | tlpr1_stats_chk | | |
| .flow.memuse | 542447488 | 498197568 | 108.88% | 

Pipeline 7190

rust/src/applayer.rs
```diff
@@ -535,14 +535,17 @@ pub trait Transaction {
...
+    fn get_transaction_by_index(&self, index: usize) -> Option<&Tx>;
 
     fn get_transaction_iterator(&self, min_tx_id: u64, state: &mut u64) -> AppLayerGetTxIterTuple {
         let mut index = *state as usize;
-        let transactions = self.get_transactions();
-        let len = transactions.len();
+        let len = self.get_transaction_count();
         while index < len {
-            let tx = &transactions[index];
+            let tx = self.get_transaction_by_index(index).unwrap();
```
catenacyber, line 548: Should we rather have an iterator ?

catenacyber:
> > Should other protocols like mqtt, http2, dcerpc, rdp, pgsql be part of this PR ?
> 
> Yeah, I can take care of that as far of this, once some level of approval has been reached.

I personally approve this approach :-)

rust/src/applayer.rs
jasonish: I don't think we can make it generic, at least not without a bunch more work.. The iterator returned from VecDeque and Vec are different types, and I couldn't find away to be generic over both of them.

I did try using `.iter()` and `.iter().enumerate()` on a local DNS transaction iterator for a purely idiomatic Rust approach expecting a tiny noticeable performance different.  `.enumerate()` to get the index from the iterator instead of manually updating it was about 1000 times slower than using the index approach.  And I didn't notice a performance difference between an `.iter()` and manually updating the index var, or using the purely indexed approach here.  I was hoping to, to justify the extra work, but I can't justify it, and the added complexity right now.

victorjulien:
I like it. Are you doing a new PR with the additional protocols? Or a follow up PR after merge?

PR #7304, opened 2022-04-25
Codeql v4

catenacyber:
Link to redmine ticket:
None

Describe changes:

Updated lgtm.yml and codeql.yml so :
-Suricata will works on LGTM
-Codeql workflow for C will successfully run

Fixes 18 warnings found by CodeQL

Replaces #7299 by trying to fix compilation on windows where `fchmod` does not exist

PR #7331, opened 2022-04-27
Codeql 5307 v1

catenacyber:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5307
https://redmine.openinfosecfoundation.org/issues/5308
https://redmine.openinfosecfoundation.org/issues/5310

Describe changes:

Updated lgtm.yml and codeql.yml so :
-Suricata will works on LGTM
-Codeql workflow for C will successfully run

Fixes 18 warnings found by CodeQL

Replaces #7304 with commit reworded and tickets

### DIFF
--- a/src/util-print.c
+++ b/src/util-print.c
@@ -292,3 +292,10 @@ const char *PrintInet(int af, const void *src, char *dst, socklen_t size)
     }
     return NULL;
 }
+
+void PrintHexString(char *str, size_t size, uint8_t *buf, size_t buf_len)
+{
+    for (size_t x = 0; x < buf_len; x++) {
+        snprintf(str + 2 * x, size - 2 * x, "%02x", buf[x]);
+    }
+}
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,55 @@
+name: "CodeQL"
+
+on:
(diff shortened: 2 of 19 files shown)


suricata-qa:
ERROR:
 

Pipeline 7199

src/util-print.c
```diff
@@ -292,3 +292,10 @@ const char *PrintInet(int af, const void *src, char *dst, socklen_t size)
     }
     return NULL;
 }
+
+void PrintHexString(char *str, size_t size, uint8_t *buf, size_t buf_len)
+{
+    for (size_t x = 0; x < buf_len; x++) {
+        snprintf(str + 2 * x, size - 2 * x, "%02x", buf[x]);
```
victorjulien, line 299: can we underflow size here?

in general I think its good that we're addressing the bad snprintf patterns, but I feel we're replacing it with another bad one. Can we use a rust func?

Otherwise I think the better pattern is
```
for ( ) {
   char tmp[3];
   snprintf(tmp, sizeof(tmp) ....
   strlcat(str, size, tmp);
}
```
This avoids any buffer size tricks and leaves all validation to the intended funcs

catenacyber:
Replaced by #7344

PR #7343, opened 2022-04-29
Alert queue optimizations - v6

jufajardini:
Previous PR: https://github.com/OISF/suricata/pull/7326

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4943
https://redmine.openinfosecfoundation.org/issues/5179

Describe changes:
- move `WARN_ONCE` back to its original file
- bring related documentation commits
- extract logic for applying `Signature` actions to `Flow` to dedicated fuction
- extract logic for assigning `Signature` fields to `PacketAlert` to dedicated function

suricata-verify-pr: 808
https://github.com/OISF/suricata-verify/pull/808


jlucovsky (commented):
This looks really good.

The only question I have is regarding the configuration variable. In what situations should the default value be changed? Can we offer guidance "In <these situations>, consider increasing (or decreasing> the value". What are "these situations"?

catenacyber:
That is a good question.

Is there a log message when the threshold is reached ?
This way, the people needing this will know ;-)

jufajardini:
> > That is a good question.
> > Is there a log message when the threshold is reached ? This way, the people needing this will know ;-)
> 
> We have the stats counter that informs about `alert_queue_overflow` and will show alerts that were discarded due to queue size. And we have a log message when the queue capacity is increase, but we left that for debug level only.

So _maybe_ we could add a note saying that if folks are seeing high numbers for that counter, they could consider increasing `packet-alert-max`? Don't know what would be an advisable new size, though.

jufajardini:
Followed by: https://github.com/OISF/suricata/pull/7347

PR #7342, opened 2022-04-28
Codeql 5307 v2

catenacyber:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5307
https://redmine.openinfosecfoundation.org/issues/5308
https://redmine.openinfosecfoundation.org/issues/5310

Describe changes:

Updated lgtm.yml and codeql.yml so :
-Suricata will works on LGTM
-Codeql workflow for C will successfully run

Fixes 18 warnings found by CodeQL

Replaces #7331 using rust instead of snprintf

PR #7344, opened 2022-04-29
Codeql 5307 v3

catenacyber:
Link to redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5307
https://redmine.openinfosecfoundation.org/issues/5308
https://redmine.openinfosecfoundation.org/issues/5310

Describe changes:

Updated lgtm.yml and codeql.yml so :
-Suricata will works on LGTM
-Codeql workflow for C will successfully run

Fixes 18 warnings found by CodeQL

Replaces #7342 by isolating rust unsafe
And adding a commit to get better compile times

### DIFF
--- a/rust/src/common.rs
+++ b/rust/src/common.rs
@@ -1,5 +1,6 @@
 use std::ffi::CString;
 use std::os::raw::c_char;
+use super::build_slice;
 
 pub mod nom7 {
     use nom7::bytes::streaming::{tag, take_until};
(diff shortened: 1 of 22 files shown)


suricata-qa:
Information:
 

|--|--|--|--| 
| | tlpr1_stats_chk | | |
| .flow.memuse | 527584448 | 498197568 | 105.9% | 

Pipeline 7240

rust/src/common.rs
```diff
@@ -121,3 +122,23 @@ pub fn to_hex(input: &[u8]) -> String {
...
 }
+
+fn print_hex(islice: &[u8], oslice: &mut [u8]) {
+    static CHARS: &'static [u8] = b"0123456789abcdef";
+
+    for i in 0..islice.len() {
+        oslice[2*i] = CHARS[(islice[i] >>  4) as usize];
+        oslice[2*i+1] = CHARS[(islice[i] & 0xf) as usize];
+    }
+    oslice[2*islice.len()] = 0;
```
victorjulien, line 133: is this to essentially turn `oslice` into a nul terminated string? If the goal is to create a c-string I wonder if it would be better go all the way and do something that returns a CString? (cc @jasonish )

src/util-print.c
```diff
@@ -292,3 +294,9 @@ const char *PrintInet(int af, const void *src, char *dst, socklen_t size)
     }
     return NULL;
 }
+
+void PrintHexString(char *str, size_t size, uint8_t *buf, size_t buf_len)
+{
+    DEBUG_VALIDATE_BUG_ON(size < 2 * buf_len);
+    rs_to_hex((uint8_t *)str, size, buf, buf_len);
```
victorjulien, line 301: or alternatively we could do the nul termination here...

guess its also a bit sketchy that we use uint8_t as a c string

rust/src/common.rs
victorjulien: a `jb_set_hex` could replace all but the `datasets.c` use it seems, and would leave things in Rust

catenacyber: Doing `jb_set_hex` in next version.
There is also `OutputFilestoreFinalizeFiles` which uses `PrintHexString`...

src/util-print.c
catenacyber: Waiting for Jason to comment on this null termination...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7354

rust/src/common.rs
jasonish: This function looks like its solely designed to be used from C. So I think this makes sense as the result is being placed into a provided buffer.

If we were to build a CString, Rust would need to allocate and return it.  Then a call would have to be made back to Rust to free it.

I wonder if this should just be one function all within the `extern` function, to make it more clear that its really for C callers only.

PR #7350, opened 2022-04-30
configure.ac: handle undefined LEVEL1_DCACHE_LINESIZE

satta:

Describe changes:
- On some platforms (riscv64, s390x) this value is `undefined` as returned from `getconf`. We also need to handle this to avoid using the string `undefined` blindly in further #defines, which would otherwise cause compile errors.


catenacyber (commented):
Maybe the commit title is not clear, but the code is good to get Suricata to compile on some systems

jasonish (approved):

victorjulien:
Merged in #7367, thanks Sascha!

PR #7359, opened 2022-05-02
Ips handling/v24

victorjulien:
Implements a new way of dealing with memcap hits: a per subsystem `memcap-policy` which can be `drop-packet`, `drop-flow`, `pass-packet`, `pass-flow`, `bypass` or `ignore`. Where `ignore` is essentially the current behavior. Looking for a better name and general feedback on the approach.

In https://github.com/OISF/suricata-verify/pull/818 there are tests for most options in combination with stream reassembly memcap failure simulation.

continues work in #7234 
(includes #7356)

suricata-verify-pr: 818


src/util-memcap.c
```diff
@@ -0,0 +1,95 @@
...
+    const char *value_str = NULL;
+    if ((ConfGetValue(option, &value_str)) == 1 && value_str != NULL) {
+        if (strcmp(value_str, "drop-flow") == 0) {
+            policy = MEMCAP_POLICY_DROP_FLOW;
+            SCLogNotice("%s: %s", option, value_str);
+        } else if (strcmp(value_str, "pass-flow") == 0) {
+            policy = MEMCAP_POLICY_PASS_FLOW;
+            SCLogNotice("%s: %s", option, value_str);
+        } else if ((strcmp(value_str, "bypass") == 0) ||
+                   (strcmp(value_str, "bypass-flow") == 0)) {
```
jasonish, line 76: Both names required? Or can we just use one?

src/stream-tcp.c
```diff
@@ -462,6 +463,9 @@ void StreamTcpInitConfig(bool quiet)
             stream_config.flags |= STREAMTCP_INIT_FLAG_INLINE;
         }
     }
+    stream_config.ssn_memcap_policy = MemcapPolicyParse("stream.memcap-policy");
+    stream_config.reassembly_memcap_policy = MemcapPolicyParse("stream.reassembly.memcap-policy");
```
jasonish, line 467: Or should policies be in their own configuration file section:
```
policy:
  stream-memcap: ...
  stream-reassembly-memcap: ...
```

jasonish:
> On further thought I think the MemcapPolicy thing could even be a more general ExceptionPolicy, where possible exceptions include memcap hits, other resource limit hits, app errors and probably more

Yes, I agree.  I also think it would be more user friendly if this exception policy was configured in one place, rather than having it sprinkled among the various modules.

victorjulien:
> > On further thought I think the MemcapPolicy thing could even be a more general ExceptionPolicy, where possible exceptions include memcap hits, other resource limit hits, app errors and probably more
> 
> Yes, I agree. I also think it would be more user friendly if this exception policy was configured in one place, rather than having it sprinkled among the various modules.

Tough one. I feel like it makes sense to have them close to the memcap settings, but I also see the point of grouping them. Ultimately I feel most ppl would never touch these. They would get the implied values (`ignore` for IDS, `drop-flow` for IPS). The actual settings would be available for expert tuning.

src/util-memcap.c
victorjulien: Can do one. We use `bypass` in various places in config, but here I used `flow-(packet|flow)` pattern, so it seemed to make sense to have both.

jasonish: Being a new configuration option, I'd rather not see an alias right from the start.

suricata-qa:
ERROR:
 

|--|--|--|--| 
| | tlpr1_stats_chk | | |
| .app_layer.error.ftp-data.parser | 0 | 395 | 0.0% | 

Pipeline 7353

src/util-memcap.c
victorjulien: went with just `bypass`

PR #7360, opened 2022-05-02
dcerpc: use vecdeque for transactions

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/5321

Missed in previous batch of conversions.
Also fixes copyright year screwups in previous batch.


victorjulien:
Merged in #7367, thanks!

PR #7362, opened 2022-05-02
6.0.x: ftp: truncate first segment if over max length

jasonish:
The first segment was not limited to the configured maximum line length
allowing it to be up to 65k. This could result in the next input length
being negative, which while handled properly by the code, did trigger a
debug validation assertion.

The fix is to be consistent and apply the limit to the first segment as
well, which does ensure the input_len could never be less than 0.

Issue: https://redmine.openinfosecfoundation.org/issues/5281

(cherry picked from commit 9645285dff9eb8313db573d8603162a708736236)

suricata-verify-pr: 819

### DIFF
--- a/src/app-layer-ftp.c
+++ b/src/app-layer-ftp.c
@@ -371,6 +371,9 @@ static int FTPGetLineForDirection(FtpState *state, FtpLineState *line_state)
         }
     }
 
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged in #7369, thanks!

PR #7376, opened 2022-05-04
rules: use primary default-rule-path if set on command line

jasonish:
When reloading rules, respect `--set default-rule-path=...` from the
command line if set.

Previously the rule reload would always take the default-rule-path from
the configuration file, even if overrided on the command line.

This is a rather narrow fix for the broader issue discussed in issue 1911, but
I think thats better suited for a much larger fix, which I've started to
discuss in https://redmine.openinfosecfoundation.org/issues/4782.

Issue: https://redmine.openinfosecfoundation.org/issues/1911

### DIFF
--- a/src/detect-engine-loader.c
+++ b/src/detect-engine-loader.c
@@ -68,16 +68,22 @@ char *DetectLoadCompleteSigPath(const DetectEngineCtx *de_ctx, const char *sig_f
         return NULL;
     }
 
-    if (strlen(de_ctx->config_prefix) > 0) {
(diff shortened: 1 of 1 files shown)


suricata-qa:
ERROR:
 

|--|--|--|--| 
| | tlpr1_stats_chk | | |
| .app_layer.error.ftp-data.parser | 0 | 395 | 0.0% | 

Pipeline 7312

victorjulien:
Fixed up clang-format in staging.

victorjulien:
Merged as part of #7423

PR #7389, opened 2022-05-08
Make the droped packet know the associated sig

nowaits:
Currently if the packet flow has already has FLOW_ACTION_DROP flag, the packet will also be taged drop, but the packet has no way to know the associated sig. This patch will make it know.

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -342,7 +342,7 @@ static void SignalHandlerUnexpected(int sig_num, siginfo_t *info, void *context)
 
 terminate:
     // Propagate signal to watchers, if any
-    kill(0, sig_num);
+    kill(getpid(), sig_num);
 }
 #undef UNW_LOCAL_ONLY
 #endif /* HAVE_LIBUNWIND */


src/flow.h
```diff
@@ -499,6 +501,8 @@ typedef struct Flow_
 
     struct timeval startts;
 
+    const Signature *sig_drop;
```
victorjulien, line 504: This isn't safe. During a rule reload rules are freed, so this will become a dangling pointer. A simpler approach would be to just store the sid and simply log that.

victorjulien (changes requested):

victorjulien:
I did not see an entry in our contribution agreement db? Did you sign it? If not, could you have a look at https://suricata.io/contribution-agreements/ Thanks!

nowaits:
> I did not see an entry in our contribution agreement db? Did you sign it? If not, could you have a look at https://suricata.io/contribution-agreements/ Thanks!
When trying to sign it, i got the error `There was an error trying to send your message. Please try again later.`

src/flow.h
nowaits: I have modified the code according your advice. Can you check it?

nowaits:
> I did not see an entry in our contribution agreement db? Did you sign it? If not, could you have a look at https://suricata.io/contribution-agreements/ Thanks!

I have signed the agreement.

jufajardini (commented):
Thanks for your interest in contributing to Suricata! :)

I have added a few inline comments (which may need answers from @victorjulien ...)

When you address new requested changes, can you make sure to submit those in a new PR?
(more details about our contribution process can be found at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing)

PR #7393, opened 2022-05-09
Ips handling/v31

victorjulien:
Continuation #7359.

Implements a more general "exception policy" that can be used to handle memcap hits and other errors.

Implement this for:
- stream memcap hits
- app errors
- defrag memcap hit
- flow memcap hit

Adds error simulation commandline options.

suricata-verify-pr: 822


src/app-layer.c
```diff
@@ -620,6 +618,12 @@ static int TCPProtoDetect(ThreadVars *tv,
         }
     }
     SCReturnInt(0);
+parser_error:
+    ExceptionPolicyApply(p, g_applayerparser_error_policy, PKT_DROP_REASON_APPLAYER_ERROR);
+    SCReturnInt(-1);
+detect_error:
+    DisableAppLayer(tv, f, p);
+    SCReturnInt(-2);
```
jasonish, line 626: So if a parser enters an error state, a policy is applied, perhaps event to the flow.  On detect error, we disable the parser, will any policy be applied to those packets that will now no longer be parsed?

jasonish (commented):
Just some questions.  I see a couple WIP commits that need to be removed (not clear why they are WIP).  Then I guess documentation on policy fields, as well as these new command line options.

src/app-layer.c
victorjulien: policy for protocol detection exceptions isn't implemented. This change is just more clearly separate the errors by parser and pd

victorjulien: so the above should change the existing pd exception behavior in any way

src/decode-events.c
```diff
@@ -802,6 +802,10 @@ const struct DecodeEvents_ DEvents[] = {
             "stream.pkt_retransmission",
             STREAM_PKT_RETRANSMISSION,
     },
+    {
+            "stream.pkt_spurious_retransmission",
+            STREAM_PKT_SPURIOUS_RETRANSMISSION,
```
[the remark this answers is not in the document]
victorjulien: yes, disabled by default though. This is quite common. Might not even warrant an event, but I feel if suri classifies it as something we should be able to match on it

victorjulien:
Replaced by #7463

PR #7402, opened 2022-05-11
detect: introduce "like" ip-only signature type - v1

jasonish:
Rules that look like they should be IP-only but contain a negated rule
address are now marked with an LIKE_IPONLY flag. This is so they are
treated like IPONLY rules with respect to flow action, but don't
interfere with other IPONLY processing like using the radix tree.

Ticket: https://redmine.openinfosecfoundation.org/issues/5361

suricata-verify-pr: 824


suricata-qa:
ERROR:
 

Pipeline 7473

PR #7409, opened 2022-05-13
Clean shorterintegers

catenacyber:
None yet, draft pull request to discuss, should I create one ?

Describe changes:
- Use shorter integers for array accesses

As found by CodeQL with https://github.com/github/codeql/pull/8994

@HansmannThibaut could you run the query on this branch and give me the new results ? (util-crypt.c was removed)

This was inspired by 2d765d6c686449e78e29759b07c4852ebab3c46e where one of the root cause for the buffer overflow in `SetupU8Hash` was that we were using an `int` as an index to an array of size 256

More work to do : especially see if `mpm_matcher` can be an `uint8_t`

This may interact with https://redmine.openinfosecfoundation.org/issues/4516 (combination of fixes may expose new problems)


suricata-qa:
ERROR:
 

Pipeline 7461

src/util-decode-mime.c
```diff
@@ -950,8 +950,8 @@ static int IsIpv6Host(const uint8_t *urlhost, uint32_t len)
     char tempIp[MAX_IP6_CHARS + 1];
 
     /* Cut off at '/'  */
-    uint32_t i = 0;
-    for (i = 0; i < len && urlhost[i] != 0; i++) {
+    uint8_t i = 0;
+    for (i = 0; i < len && urlhost[i] != 0 && i < UINT8_MAX; i++) {
```
victorjulien, line 954: I feel this is an anti-pattern. We have a length field, but we ignore it?

src/detect-filestore.c
```diff
@@ -224,7 +224,7 @@ static int DetectFilestorePostMatch(DetectEngineThreadCtx *det_ctx,
     else
         flags |= STREAM_TOSERVER;
 
-    for (uint16_t u = 0; u < det_ctx->filestore_cnt; u++) {
+    for (uint8_t u = 0; u < det_ctx->filestore_cnt; u++) {
```
victorjulien, line 227: `det_ctx->filestore_cnt` is `uint16_t`, does this change make sense?

src/util-decode-mime.c
catenacyber: > I feel this is an anti-pattern. We have a length field, but we ignore it?

How so ?
`i < len` does not look to me like ignoring the length field...

src/detect-filestore.c
catenacyber: This PR changes it to uint8_t in detect.h DetectEngineThreadCtx_ definition precisely because `DETECT_FILESTORE_MAX` (size of filestore array whose `filestore_cnt` is an index) is lesser than 256

src/util-decode-mime.c
victorjulien: whoops misread the diff

catenacyber:
Current status :
- managed to add custom codeQL query to CI
- now need to get CI green again

suricata-qa:
ERROR:
 

Pipeline 7905

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7562

PR #7441, opened 2022-05-30
conf: remove ConfGetValue

jasonish:
All uses of ConfGetValue are satisfied by ConfGet


catenacyber:
I love dead code removal 👍 

By the way, is there a plan to get rid of C dependencies such as libyaml to use rust instead ?

jasonish:
> By the way, is there a plan to get rid of C dependencies such as libyaml to use rust instead ?

Yeah, in time at least.  That is how I found this.  I have a YAML loader in Rust that deals with our quirks of includes and such, the main issue would be updating all the C code to use the new API, in particular all the usages that iterate over the linked lists.

victorjulien (approved):

victorjulien:
Merged as part of #7455, thanks!

PR #7447, opened 2022-05-31
rust: make suricata context const

catenacyber:
None, should there be one to add more `const` everywhere ?

Describe changes:
- rust: make suricata context const, ie read only

### DIFF
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -219,7 +219,7 @@ extern {
 
 pub static mut SC: Option<&'static SuricataContext> = None;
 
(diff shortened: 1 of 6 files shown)


jasonish (changes requested):
I like this, however it looks like 2 rust functions need to have their signatures updated as well.

catenacyber:
> I like this, however it looks like 2 rust functions need to have their signatures updated as well.

Yes, trying a new version against CI to see if warnings disappear

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7457

PR #7450, opened 2022-05-31
events: reset logged event

catenacyber:
I should create one once this bug is acknowledged

Describe changes:
- events: reset logged event

I guess this needs a S-V test also

### DIFF
--- a/src/app-layer-events.c
+++ b/src/app-layer-events.c
@@ -122,8 +122,10 @@ void AppLayerDecoderEventsSetEventRaw(AppLayerDecoderEvents **sevents, uint8_t e
 
 void AppLayerDecoderEventsResetEvents(AppLayerDecoderEvents *events)
 {
-    if (events != NULL)
+    if (events != NULL) {
         events->cnt = 0;
+        events->event_last_logged = 0;
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


victorjulien:
Not sure I understand the bug. Can you do a SV test to show it?

suricata-qa:
ERROR:
 

Pipeline 7617

catenacyber:
> Not sure I understand the bug. Can you do a SV test to show it?

I got this with my pcap for Suricon ;-)

I try to reproduce with another from scratch pcap
It seems it depends on the packet recycling stuff
When reading a pcap, which packets get `PACKET_RECYCLE` from another ?

catenacyber:
To try to phrase the bug :
When a packet that had an event logged gets recycled, if it gets a new event, this new event does not get logged, because only  events->cnt was reset and not events->event_last_logged and `event_last_logged` gets used in output-json-anomaly.c by `AnomalyAppLayerDecoderEventJson`

victorjulien:
That sounds like a bug indeed. Essentially in the packet flow we take a packet from the pool, use it, then return it. Before returning we clean it (PACKET_RECYCLE). Sounds like its not fully cleaned.

Perhaps we need a set of debug validation checks to see if a packet we get from the pool is clean in the ways we expect.

suricata-qa:
ERROR:
 ERROR: Invalid Signature config error in tlpw1_cfg QA test

ERROR: QA failed on tlpw1_cfg.
  

Pipeline 7668

catenacyber:
> Perhaps we need a set of debug validation checks to see if a packet we get from the pool is clean in the ways we expect.

Well, if we know what we expect, we can reset it properly in the first place, no ?

PR #7457, opened 2022-06-01
rust: make suricata context const

catenacyber:
None, should there be one to add more `const` everywhere ?

Describe changes:
- rust: make suricata context const, ie read only

#7447 without the force-push fixing the compilation warning

### DIFF
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -219,7 +219,7 @@ extern {
 
 pub static mut SC: Option<&'static SuricataContext> = None;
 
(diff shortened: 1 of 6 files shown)


victorjulien (approved):
General improvement doesn't need a ticket I think.

victorjulien:
Merged in #7467, thanks!

catenacyber:
> General improvement doesn't need a ticket I think.

Maybe we should create a ticket to make other structures with function pointers const...

PR #7469, opened 2022-06-02
detect/alert: don't lose 'drop' alert - v1

jufajardini:
This is a first attempt at ensuring we don't lose a drop alert, even if the corresponding alert was discarded due to alert queue expansion failure (or just discarded in general). 

Leaving as draft both because I'm not certain this solution isn't too messy/ugly, and because I've included work from 5319, just to make sure that the corresponding sv test would work, and therefore test that behavior.

Describe changes:
- First process all alerts in the queue, then sort them and copy the ones that should stay to Packet::alerts
- split 'PacketAlertFinalize()' into two functions. The loop that process all 'PacketAlert's is now in 'PacketAlertQueueFinalize()' while each individual 'PacketAlert' is process within 'PacketAlertFinalize()'
- add a flag to indicate if there was a realloc failure
- check for 'PASS' and 'DROP' actions explicitly after processing alerts' queue
- always use a function to set 'Packet::alerts::drop'

suricata-verify-pr: 841
https://github.com/OISF/suricata-verify/pull/841

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -172,9 +172,26 @@ static inline void RuleActionToFlow(const uint8_t action, Flow *f)
     }
 }
 
+/** \internal
+ */
+static inline PacketAlert PacketAlertSet(
+        DetectEngineThreadCtx *det_ctx, const Signature *s, uint64_t tx_id, uint8_t alert_flags)
+{
+    PacketAlert pa;
+    pa.num = s->num;
+    pa.action = s->action;
+    pa.s = (Signature *)s;
+    pa.flags = alert_flags;
+    /* Set tx_id if the frame has it */
+    pa.tx_id = (tx_id == UINT64_MAX) ? 0 : tx_id;
+    pa.frame_id = (alert_flags & PACKET_ALERT_FLAG_FRAME) ? det_ctx->frame_id : 0;
+    return pa;
+}
+
 /** \brief Apply action(s) and Set 'drop' sig info,
  *         if applicable */
-static void PacketApplySignatureActions(Packet *p, const Signature *s, const uint8_t alert_flags)
+static void PacketApplySignatureActions(
+        DetectEngineThreadCtx *det_ctx, Packet *p, const Signature *s, const uint8_t alert_flags)
 {
     SCLogDebug("packet %" PRIu64 " sid %u action %02x alert_flags %02x", p->pcap_cnt, s->id,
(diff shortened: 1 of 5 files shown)


suricata-qa:
ERROR:
 

Pipeline 7665

src/detect-engine-alert.c
```diff
@@ -172,19 +172,34 @@ static inline void RuleActionToFlow(const uint8_t action, Flow *f)
     }
 }
 
+/** \internal
+ */
+static inline PacketAlert PacketAlertSet(
+        DetectEngineThreadCtx *det_ctx, const Signature *s, uint64_t tx_id, uint8_t alert_flags)
+{
+    PacketAlert pa;
```
victorjulien, line 180: if we init `pa` as `PacketAlert pa = { s->num, s->action, s, ... }` it would give a compile error if a new field was added that isn't set here. Would that make sense to avoid future issues?

src/detect-engine-alert.c
```diff
@@ -320,73 +328,139 @@ static inline void FlowApplySignatureActions(
...
+        FlowApplySignatureActions(p, pa, s, pa->flags);
+
+        /* set actions on packet */
+        PacketApplySignatureActions(det_ctx, p, s, pa->flags);
+    }
+
+    /* Thresholding removes this alert */
+    if (res == 0 || res == 2 || (s->flags & SIG_FLAG_NOALERT)) {
+        /* we will not copy this to the AlertQueue */
+        ret = 0;
```
victorjulien, line 372: just `return 1;` here?

src/detect-engine-alert.c
```diff
@@ -320,73 +328,139 @@ static inline void FlowApplySignatureActions(
...
+    } else {
+        ret = 1;
```
victorjulien, line 374: `return 1;`?

then we can get rid of `ret`

src/detect.h
```diff
@@ -1119,6 +1119,7 @@ typedef struct DetectEngineThreadCtx_ {
     uint16_t alert_queue_size;
     uint16_t alert_queue_capacity;
     PacketAlert *alert_queue;
+    bool is_expand_failure;
```
victorjulien, line 1122: name is very generic while it is about `alert_queue` expansion

jufajardini:
Comments addressed by: https://github.com/OISF/suricata/pull/7476

PR #7351, opened 2022-05-02
Warnint util 4516 v11.1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4516

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` for defrag

cc @jasonish as you own that code, right ?
cc @jufajardini for the decode.c commit

Part of #7006 as #7298 was

Then remains only detect files to fix

PR #7470, opened 2022-06-03
Warnint util 4516 v12.1

catenacyber:

Describe changes:
- Fix integer warnings `-Wimplicit-int-conversion` for detect*
- Adds ci check for this warning

Final part of #7006 after #7351

Then remains only detect files to fix

Should this be split in multiple commits ?

### DIFF
--- a/src/detect-app-layer-event.c
+++ b/src/detect-app-layer-event.c
@@ -145,7 +145,7 @@ static DetectAppLayerEventData *DetectAppLayerEventParsePkt(const char *arg,
 {
     int event_id = 0;
     int r = AppLayerGetPktEventInfo(arg, &event_id);
-    if (r < 0) {
+    if (r < 0 || r > UINT8_MAX) {
         SCLogError(SC_ERR_INVALID_SIGNATURE, "app-layer-event keyword "
                    "supplied with packet based event - \"%s\" that isn't "
                    "supported yet.", arg);
@@ -155,7 +155,7 @@ static DetectAppLayerEventData *DetectAppLayerEventParsePkt(const char *arg,
     DetectAppLayerEventData *aled = SCCalloc(1, sizeof(DetectAppLayerEventData));
     if (unlikely(aled == NULL))
         return NULL;
-    aled->event_id = event_id;
+    aled->event_id = (uint8_t)event_id;
     *event_type = APP_LAYER_EVENT_TYPE_PACKET;
 
     return aled;
@@ -231,7 +231,11 @@ static int DetectAppLayerEventParseAppP2(DetectAppLayerEventData *data,
             return -3;
         }
     }
-    data->event_id = event_id;
+    if (event_id > UINT8_MAX) {
+        SCLogWarning(SC_ERR_INVALID_SIGNATURE, "app-layer-event keyword's id has invalid value");
+        return -4;
+    }
+    data->event_id = (uint8_t)event_id;
 
     return 0;
 }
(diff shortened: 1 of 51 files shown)


suricata-qa:
ERROR:
 

ERROR: QA failed on tlpr1_asan_suri.
  

Pipeline 7671

src/detect-content.c
```diff
@@ -506,36 +506,62 @@ void DetectContentPropagateLimits(Signature *s)
...
                         if (depth && depth > offset_plus_pat) {
-                            uint16_t dist = 0;
+                            int32_t dist = 0;
                             if (cd->flags & DETECT_CONTENT_DISTANCE && cd->distance > 0) {
                                 dist = cd->distance;
                                 SCLogDebug("distance to add: %u. depth + dist %u", dist, depth + dist);
                             }
                             SCLogDebug("depth %u + cd->within %u", depth, cd->within);
-                            depth = cd->depth = depth + cd->within + dist;
+                            if (depth + cd->within + dist < 0 ||
```
victorjulien, line 523: I would like these to go into their own commit

src/detect-app-layer-event.c
```diff
@@ -231,7 +231,11 @@ static int DetectAppLayerEventParseAppP2(DetectAppLayerEventData *data,
             return -3;
         }
     }
-    data->event_id = event_id;
+    if (event_id > UINT8_MAX) {
+        SCLogWarning(SC_ERR_INVALID_SIGNATURE, "app-layer-event keyword's id has invalid value");
```
victorjulien, line 235: new error/warning condition, would like a separate commit here

catenacyber:
QA triggers debug assertion at  detect-engine-content-inspection.c:531

(oops)

src/detect.c
```diff
@@ -1150,7 +1149,8 @@ static bool DetectRunTxInspectRule(ThreadVars *tv,
                 TRACE_SID_TXS(s->id, tx, "engine %p match %d", engine, match);
                 if (engine->stream) {
                     can->stream_stored = true;
-                    can->stream_result = match;
+                    // TODO change Callback prototype ?
```
victorjulien, line 1152: yes I think that makes sense, also in a separate commit

src/detect.h
```diff
@@ -1021,7 +1021,7 @@ typedef struct HttpReassembledBody_ {
 typedef struct SignatureNonPrefilterStore_ {
     SigIntId id;
     SignatureMask mask;
-    uint8_t alproto;
+    AppProto alproto;
```
victorjulien, line 1024: does this change the size of the struct? guess not if `AppProto` is uint16_t, but if it is an enum it would

victorjulien (changes requested):
Requested separate commits in a few places, the rest can be in a single commit

src/detect-content.c
catenacyber: ok, looks relevant

src/detect-app-layer-event.c
catenacyber: ok

src/detect.c
catenacyber: that makes a big commit...

catenacyber: (done in next version)

src/detect.h
catenacyber: `typedef uint16_t AppProto;` ;-)

PR #7483, opened 2022-06-05
Add rules for ms-msdt

r00tik:
Add rules for download http files which contains ms-msdt


victorjulien:
Hi @r00tik these rules files are meant only to contain rules that match on suricata's generated events, in this case by the http parser. They are not meant for general detection rules. Perhaps they would fit https://github.com/OISF/suricata-trafficid @jasonish any thoughts on that?

jasonish:
Doesn't really fit with trafficid either. Seems more like what you might find in ET. They have rule 2036726 which is for the same CVE, is there much difference between this rules and the ET one?

victorjulien:
Right, yeah ET might be a better fit.

PR #7497, opened 2022-06-07
smb: use default stream-depth 0 by default

catenacyber:
None, should there be ?

Describe changes:
- smb: use default stream-depth 0 by default

cc @inashivb

### DIFF
--- a/rust/src/smb/smb.rs
+++ b/rust/src/smb/smb.rs
@@ -2449,8 +2449,8 @@ pub unsafe extern "C" fn rs_smb_register_parser() {
                 Ok(retval) => { stream_depth = retval as u32; }
                 Err(_) => { SCLogError!("Invalid depth value"); }
             }
-            AppLayerParserSetStreamDepth(IPPROTO_TCP as u8, ALPROTO_SMB, stream_depth);
         }
+        AppLayerParserSetStreamDepth(IPPROTO_TCP as u8, ALPROTO_SMB, stream_depth);
         let retval = conf_get("app-layer.protocols.smb.max-read-size");
         if let Some(val) = retval {
             match get_memval(val) {
--- a/src/app-layer-smb.c
+++ b/src/app-layer-smb.c
@@ -32,8 +32,6 @@
 static StreamingBufferConfig sbcfg = STREAMING_BUFFER_CONFIG_INITIALIZER;
 static SuricataFileContext sfc = { &sbcfg };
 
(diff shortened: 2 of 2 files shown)


rust/src/smb/smb.rs
```diff
@@ -2449,8 +2449,8 @@ pub unsafe extern "C" fn rs_smb_register_parser() {
                 Ok(retval) => { stream_depth = retval as u32; }
                 Err(_) => { SCLogError!("Invalid depth value"); }
             }
-            AppLayerParserSetStreamDepth(IPPROTO_TCP as u8, ALPROTO_SMB, stream_depth);
         }
+        AppLayerParserSetStreamDepth(IPPROTO_TCP as u8, ALPROTO_SMB, stream_depth);
```
inashivb, line 2453: What happens if we do not set a stream depth?

inashivb (approved):
I think this is consistent w what the behavior used to be before I introduced https://github.com/OISF/suricata/commit/e5c948df87541509e34445d80c5ef0242da3326f

Asked a question inline. If the answer to that suggests that my above mentioned commit caused a change in behavior then I think we do need a bug ticket.

rust/src/smb/smb.rs
catenacyber: Before https://github.com/OISF/suricata/commit/e5c948df87541509e34445d80c5ef0242da3326f , without setting a stream depth, the default of 0 was used for smb.

After https://github.com/OISF/suricata/commit/e5c948df87541509e34445d80c5ef0242da3326f , without setting a stream depth, the default of 100000 was used for smb as is for other protocols.

This PR resumes the previous behavior

catenacyber:
Shivani, you can test with S-V test for https://github.com/OISF/suricata/pull/7498 if you wish

rust/src/smb/smb.rs
inashivb: Thank you. Then, I changed the behavior of the proto and we should probably have a bug ticket?

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7503

rust/src/smb/smb.rs
catenacyber: Done

Do you know if it needs backports ?

inashivb: Doesn't seem like it.

PR #7505, opened 2022-06-08
dns: remove unused events field from state

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47827

Describe changes:
- remove unused events field from state

cc @jasonish 

This was not found by rust compiler because field was public and could have been used by some external user of the crate, right ?

### DIFF
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -323,8 +323,6 @@ pub struct DNSState {
     // Transactions.
     pub transactions: VecDeque<DNSTransaction>,
 
-    pub events: u16,
-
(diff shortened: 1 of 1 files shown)


catenacyber:
Do we need ticket/backports for this ?

jasonish:
> This was not found by rust compiler because field was public and could have been used by some external user of the crate, right ?

Thats correct.  We could use `pub(crate)` more I think.

jasonish (approved):
Looks good to me.

victorjulien:
Merged in #7511, thanks!

PR #7513, opened 2022-06-09
App layer error close txs 4318 v6

catenacyber:

Describe changes:

Still need to test this, and check current app-layer parser returning error

Modifies #7243 with adding ticket number in message


catenacyber:
One example where I am not sure how it should be dealt with :

if a HTP2 client does not send the magic banner, the app-layer parser returns an error.
Should it return an error ? or Should it try to parse further frames ?

It looks like we have two cases here :
- both client and server were recognized using the app-layer protocol
- or client and server do not talk the same protocol

PS : the banner can be split over multiple TCP packets, and the server side may have created multiple transactions before getting the error

catenacyber:
Another case : SSH where a record has its length <= 1

suricata-qa:

Pipeline 7759

catenacyber:
Not sure of this

PR #7515, opened 2022-06-09
detect/dcerpc: apply dcerpc to smb as well

victorjulien:
So 'alert dcerpc' also matches if the DCERPC is over SMB.

Bug: #5208.

suricata-verify-pr: 816

### DIFF
--- a/src/app-layer-protos.h
+++ b/src/app-layer-protos.h
@@ -91,6 +91,8 @@ static inline bool AppProtoEquals(AppProto sigproto, AppProto alproto)
         case ALPROTO_HTTP:
             return (alproto == ALPROTO_HTTP1) || (alproto == ALPROTO_HTTP2) ||
                    (alproto == ALPROTO_HTTP);
+        case ALPROTO_DCERPC:
+            return (alproto == ALPROTO_DCERPC || alproto == ALPROTO_SMB);
     }
     return (sigproto == alproto);
 }


victorjulien:
This piggy backs on @catenacyber work on http1/http2 protocol logic. Philippe does this look sufficient to overload `dcerpc` to mean "either dcerpc or smb" for rule matching purposes?

inashivb:
> Hmm wonder if this would allow `alert dcerpc ... smb.share; ...)`?

Just wondering if it should. :thinking: Should we not fallback to treating the traffic as `smb` if there is a rule keyword present that demands `smb` inspection? That will also check the `dcerpc` keywords if present in the rule by treating them as DCERPC/SMB..

The way I'm thinking is:
DCERPC: could be standalone or over smb => should be checked for both if only dce keywords are available in the rule.
SMB: could be standalone or have dcerpc as a part of it => should be checked if either of the keywords are present.

Above comments bc `dcerpc` could be a part of `smb` traffic but `smb` cannot be a part of `dcerpc` traffic.

victorjulien:
Right, no I should have been more clear. I think smb keywords should not be usable in a `alert dcerpc` case.

catenacyber:
Should we allow `alert smb (dcerpc.keyword : value ; smb.keyword : value) ` ?

By the way, your patch fixes some exceptional case in detect.c with ALPROTO_DCERPC and the use of SIG_FLAG_INIT_DCERPC

victorjulien:
> Should we allow `alert smb (dcerpc.keyword : value ; smb.keyword : value) ` ?

I think so, ya.

> By the way, your patch fixes some exceptional case in detect.c with ALPROTO_DCERPC and the use of SIG_FLAG_INIT_DCERPC

Checking

catenacyber:
So, smb keywords should check if `s->alproto == ALPROTO_DCERPC` before calling `DetectSignatureSetAppProto(s, ALPROTO_SMB)` to forbid `alert dcerpc ... smb.share; ...)`

This looks like the best solution to me

victorjulien:
Posting a new PR in a bit.

suricata-qa:

Pipeline 7783

victorjulien:
replaced by #7522

PR #7509, opened 2022-06-08
detect/bytemath: Support out of order options

jlucovsky:
This PR replaces the draft pr #7484.

The intent of this PR is to support arbitrary order options for `byte_math`. During the investigation phase, it was suggested that the parser be converted to Rust for simplification. Thus, the PCRE based parser for `byte_math` was removed and replaced with a Rust based parser. @jasonish's prototype rulers parser work helped guide this PR.

The C unittests were retained and many were added to the Rust based parser.

Issue: 5077

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5077](https://redmine.openinfosecfoundation.org/issues/5077)

Describe changes:
- Converts the PCRE based parser to Rust.
- Adds unit tests to the new Rust modules
- Removes the PCRE parser from detect-bytemath.c
- Adjusts the C source modules to refer to the Rust definitions

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #7516, opened 2022-06-09

jlucovsky:
Continuation of #7509

The intent of this PR is to support arbitrary order options for `byte_math`. During the investigation phase, it was suggested that the parser be converted to Rust for simplification. Thus, the PCRE based parser for `byte_math` was removed and replaced with a Rust based parser. @jasonish's prototype rulers parser work helped guide this PR.

Issue: 5077

Describe changes:

Updates
- Rebase
#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/rust/src/detect_parser/byte_math.rs
+++ b/rust/src/detect_parser/byte_math.rs
@@ -1,0 +1,1007 @@
+/* Copyright (C) 2022 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+// Author: Jeff Lucovsky <jeff@lucovsky.org>
+
+use crate::detect_parser::error::RuleParseError;
+use crate::detect_parser::parser::{parse_token, take_until_whitespace};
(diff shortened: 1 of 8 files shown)


jasonish:
I think this is good, and would approve. But I wonder if we should revist the discussion of the name spacing of public functions?

All the `extern` functions here are now part of a public interface whether we like it or not, even if considered private, they become visible to anything linking with Suricata.  I'd like to see us moving away from the `rs_` prefix for extern Rust functions, and adopt are C conventions.  When calling a method with a C signature, does it really matter if its Rust or C? Does the user need to know? I don't think so.  Then we also want a prefix to avoid name collisions.  This would result in changes like:
- `rs_bytemath_parse` to `ScByteMathParse`
- `rs_bytemath_free` to `ScByteMathFree`

to go one step further, even the datatypes become part of the *public* interface.  So should `DetectByteMathData` become `ScDetectByteMathData`?

jlucovsky:
> I think this is good, and would approve. But I wonder if we should revist the discussion of the name spacing of public functions?
> 
> All the `extern` functions here are now part of a _public_ interface whether we like it or not, even if considered private, they become visible to anything linking with Suricata. I'd like to see us moving away from the `rs_` prefix for `extern` Rust functions, and adopt our C conventions. When calling a method with a C signature, does it really matter if its Rust or C? Does the user need to know? I don't think so. Then we also want a prefix to avoid name collisions. This would result in changes like:
> 
> * `rs_bytemath_parse` to `ScByteMathParse`
> * `rs_bytemath_free` to `ScByteMathFree`
> 
> to go one step further, even the datatypes become part of the _public_ interface. So should `DetectByteMathData` become `ScDetectByteMathData`?

The `const` elements are also part of the interface then -- should we adopt the same notational changes for those?

jasonish:
> The `const` elements are also part of the interface then -- should we adopt the same notational changes for those?

As in should `Const` become part of the name of `const` elements?  I'm not sure, means changing a name if changing something from const to non-const and vice-versa.  But either way, it changes the function signature so is breaking.

suricata-qa:

Pipeline 7781

jlucovsky:
Continued in #7525

PR #7519, opened 2022-06-09
Detect generic uint 4112 v10

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4112

Describe changes:
- Use generic integer functions for more keywords

Continues https://github.com/OISF/suricata/pull/7461

What may remain
```
> git grep -Eow "[A-Z_]*_LT" | cut -d: -f2 | sort | uniq
ADDRESS_LT
DATAREP_OP_LT
DETECT_BYTETEST_OP_LT
DETECT_IPPROTO_OP_LT
DETECT_ITYPE_LT
DETECT_TLS_VALIDITY_LT
DETECT_UINT_LT
FLOWINT_MODIFIER_LT
PORT_LT
PROCEDURE_LT
SEQ_LT
```

PR #7520, opened 2022-06-09
Detect generic uint 4112 v11

catenacyber:

Describe changes:
- Use generic integer functions for more keywords

Continues https://github.com/OISF/suricata/pull/7519 with more keywords

What may remain cf ` git grep -Eow "[A-Z_]*_LT" | cut -d: -f2 | sort | uniq`

Ok now:
- DETECT_UINT_LT : this is the standard to use
- ADDRESS_LT, PORT_LT : can be ruled out because logic seems specific
- SEQ_LT : can be ruled out (not detection, but stream-engine)
- DETECT_BYTETEST_OP_LT: can be ruled out, byte test does other things such as bitwise operations
- DETECT_TLS_VALIDITY_LT: uses time_t instead of unsigned integer...

I do not know:
- FLOWINT_MODIFIER_LT: not sure as there are other modifiers such as FLOWINT_MODIFIER_ISSET in the same enum
- DETECT_IPPROTO_OP_LT : not sure as there is a special handling to create a 256-sized array with all matching protocols... Nice feature by the way
- DATAREP_OP_LT: not sure id there is much to gain here

And look for uses of set_uint in loggers to see if we can easily add new keywords

### DIFF
--- a/src/detect-nfs-procedure.c
+++ b/src/detect-nfs-procedure.c
@@ -146,7 +146,7 @@ static int DetectNfsProcedureMatch (DetectEngineThreadCtx *det_ctx,
  */
 static DetectU32Data *DetectNfsProcedureParse(const char *rawstr)
 {
-    return DetectU32Parse(rawstr);
+    return rs_detect_u32_parse_inclusive(rawstr);
 }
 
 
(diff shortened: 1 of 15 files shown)


suricata-qa:

Pipeline 7779

src/detect-nfs-procedure.c
```diff
@@ -146,7 +146,13 @@ static int DetectNfsProcedureMatch (DetectEngineThreadCtx *det_ctx,
  */
 static DetectU32Data *DetectNfsProcedureParse(const char *rawstr)
 {
-    return DetectU32Parse(rawstr);
+    DetectU32Data *r = DetectU32Parse(rawstr);
+    // this range used to be inclusive
```
victorjulien, line 150: should this be an argument to `DetectU32Parse`? e.g. `DetectU32Parse(rawstr, DETECT_U32_INCLUSIVE)`

catenacyber: I wonder...

Do we prefer 
- consistency in version 7 for the use of ranges (meaning '<>' is always an exclusive range in a signature)
- consistency for a ruleset when upgrading from 6 to 7 (meaning no need to change the rules when upgrading)

victorjulien: I think this is too subtle a change to impose on users. It's more something to change if we create a new version of a keyword under a new name.

suricata-qa:

Pipeline 7785

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7526

src/detect-nfs-procedure.c
catenacyber: Ok, wonder if we should document the inconsistency for these nfs ranges

PR #4376, opened 2019-11-18
Optimize/output/v2

victorjulien:
Describe changes:
- Various optimizations to output framework.
- @jasonish can you check they changes to the 'root loggers'?
- fixes #4374. Unix socket was broken.

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR victorjulien-pcap: https://buildbot.openinfosecfoundation.org/builders/victorjulien-pcap/builds/340
- PR victorjulien: https://buildbot.openinfosecfoundation.org/builders/victorjulien/builds/341

PR #7523, opened 2022-06-10
stream: tcp: Handle retransmitted SYN with TSval

victorjulien:
For connections that use TCP timestamps for which the first SYN packet
does not reach the server, any replies to retransmitted SYNs will be
tropped.

This is happening in StateSynSentValidateTimestamp, where the timestamp
value in a SYN-ACK packet must match the one from the SYN packet.
However, since the server never received the first SYN packet, it will
respond with an updated timestamp from any of the following SYN packets.

The timestamp value inside suricata is not being updated at any time
which should happen. This patch fixes that problem.

Bug: #4376.

Signed-off-by: Michael Tremer <michael.tremer@ipfire.org>

suricata-verify-pr: 849

Rebased and fixed the ticket reference in the commit, as well as removing a reference to commit not in our tree.

### DIFF
--- a/src/stream-tcp.c
+++ b/src/stream-tcp.c
@@ -1647,6 +1647,24 @@ static int StreamTcpPacketStateSynSent(ThreadVars *tv, Packet *p,
                     "ssn->client.last_ack %"PRIu32"", ssn,
                     ssn->client.isn, ssn->client.next_seq,
                     ssn->client.last_ack);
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 7795

victorjulien:
Merged in #7531, thanks @mtremer !

PR #7526, opened 2022-06-10
Detect generic uint 4112 v12

catenacyber:

Describe changes:
- Use generic integer functions for more keywords

Continues https://github.com/OISF/suricata/pull/7520 with keeping nfs ranges inclusive

What may remain cf ` git grep -Eow "[A-Z_]*_LT" | cut -d: -f2 | sort | uniq`

Ok now:
- DETECT_UINT_LT : this is the standard to use
- ADDRESS_LT, PORT_LT : can be ruled out because logic seems specific
- SEQ_LT : can be ruled out (not detection, but stream-engine)
- DETECT_BYTETEST_OP_LT: can be ruled out, byte test does other things such as bitwise operations
- DETECT_TLS_VALIDITY_LT: uses time_t instead of unsigned integer...

I do not know:
- FLOWINT_MODIFIER_LT: not sure as there are other modifiers such as FLOWINT_MODIFIER_ISSET in the same enum
- DETECT_IPPROTO_OP_LT : not sure as there is a special handling to create a 256-sized array with all matching protocols... Nice feature by the way
- DATAREP_OP_LT: not sure id there is much to gain here

And look for uses of set_uint in loggers to see if we can easily add new keywords

### DIFF
--- a/rust/cbindgen.toml
+++ b/rust/cbindgen.toml
@@ -109,6 +109,7 @@ exclude = [
     "free",
     "IPPROTO_TCP",
     "IPPROTO_UDP",
+    "SRepCatGetByShortname",
 ]
 
 # Types of items that we'll generate. If empty, then all types of item are emitted.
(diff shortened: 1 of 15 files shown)


suricata-qa:

Pipeline 7789

victorjulien:
Merged in #7603, thanks. Nice job on these cleanups :)

PR #7528, opened 2022-06-10
(draft) - rust configuration library - v8

jasonish:
Create a PR just for preview, and trigger QA.

This replaces libyaml with a Rust module using rust-yaml to load the yaml.
While the rust-yaml scanner is used for the low level parsing, a custom loader
(document builder) was required to handle the includes. This is largely a copy
of the original from rust-yaml with some logic added for includes.

To get rid of libyaml, a converter was added to convert from a Rust Yaml
datastructure to our configuration layout.

There's a bigger picture idea here which has some ideas starting to being
implemented in the config crate.  Like a YAML document with all the defaults.
The idea being that a minimal user provided configuration gets every other
field filled in with defaults, so something like "--dump-config" shows an
accurate picture of the running config, and not just the user config. This is
all very much a work in progress.

A benefit of including all the defaults as a hard coded YAML document is the
possibility for type checking the user provided configuration. As YAML directly
convertible to JSON (at least after handling includes) we could even check with
jsonschema.

Issue: https://redmine.openinfosecfoundation.org/issues/4780


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | generic_suri | | |
 

Pipeline 7792

rust/config/src/ffi.rs
```diff
@@ -0,0 +1,392 @@
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn ScConfGet(key: *const c_char, vptr: *mut *const c_char) -> c_int {
+    let key = match CStr::from_ptr(key).to_str() {
+        Ok(key) => key,
+        Err(_) => return -1,
+    };
+
+    let root = GLOBAL.read().unwrap();
```
victorjulien, line 145: Does this `GLOBAL` imply we have one single config? No multi-instance? Wonder if we should consider this for the lib and multi-tenancy/multi-instance type scenarios we discussed. This would of course require all the callers to be aware of a config pointer/object, so quite intrusive.

jasonish: Both are possible. This is a place holder for the global configuration that we already have, so the API wouldn't change that much.  Its not really being used right now as I'm converting to the C data structures we have, but if we did convert all the C configuration code to use the APIs provided by the configuration crate, this is where we'd put that global stuff to avoid passing it around, as you said, it would be quite intrusive to get rid of the global state.

One idea could be that the lib exposes no concept of a global configuration, and this is done by Suricata the application, with suitable wrappers around the config api to provide a global state, or work our way out of a global state.

But I think thats all more intrusive than the first goal I'm trying to hit here, which is to get a better picture of the running configuration and hardcoded defaults to the users.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | generic_suri | | |
 

Pipeline 7811

victorjulien:
Nice! Any thoughts on how to make sure any built-in default goes through the proper API?

jasonish:
> Nice! Any thoughts on how to make sure any built-in default goes through the proper API?

Sort of.  There is no way to enforce it of course, but we can try to make a convention of sorts.  Take a look here: https://github.com/OISF/suricata/blob/master/src/app-layer-ftp.c#L142, if the configuration value isn't present we use a hard-coded default.  An alternative is to error our with a fatal error if the configuration key doesn't exist.  As S-V has a lot of tests with incomplete tests, I doubt we'd ever end up with something getting into master without its default value being added to `default.yaml`.

I'm not sure how this works for libsuricata tho.  You'd have to provide a default configuration for every component you use.  Maybe a move to the builder pattern is something to consider for components.  You could have `CreateNewFtpAppPlayerParseFromConfig(configuration_node)` as one builder, and then a purely programmatic builder for library users who may be getting configuration from other sources.  I don't think it would be too hard to prototype this idea with FTP as a C example, or DNS or DHCP as a Rust example

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | generic_suri | | |
 

Pipeline 7813

jasonish:
> Loosely related question : I am getting a bit confused about `outputs.0.eve-log = enabled` when I read in the yaml `outputs.eve-log = enabled`
> 
> How does it get used with `--set ` ? Can we set `outputs.*.eve-log.types.alert.payload = false` ?

Unfortunately setting items in lists is a little tricky and is not handled very well, even today.  Given the YAML is:
```
outputs:
  - eve-log:
      enaled: yes
```
This expands out like `outputs[0].eve-log.enabled`, not `outputs.eve-log.enabled`.  I think the answer is to come up with an expression syntax, perhaps like the one you mention above with the wildcard that is run after the configuration is loaded, and does a pattern match of sorts. Maybe we can take ideas from jsonpath here?  Anyways, this work does nothing to improse this issue, but should not be changing its current behaviour either.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | generic_suri | | |
 

Pipeline 7813

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | generic_suri | | |
 

Pipeline 7813

suricata-qa:
ERROR:
 

ERROR: QA failed on build_dbg.
  

Pipeline 11348

ct0br0:
Last QA msg is from testing new wording. Catching on the panic.

jasonish:
Closing for now.  Working on a new version that uses `serde`.

PR #7541, opened 2022-06-14

jlucovsky:
Continuation of #7539 

The intent of this PR is to support arbitrary order options for `byte_math`. During the investigation phase, it was suggested that the parser be converted to Rust for simplification. Thus, the PCRE based parser for `byte_math` was removed and replaced with a Rust based parser. @jasonish's prototype rulers parser work helped guide this PR.

Issue: 5077

Describe changes:

Updates
- Add `Drop` implementation for better memory handling -- revert deferred memory allocation change from #7539 

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/rust/src/detect_parser/byte_math.rs
+++ b/rust/src/detect_parser/byte_math.rs
@@ -1,0 +1,1067 @@
+/* Copyright (C) 2022 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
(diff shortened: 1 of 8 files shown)


suricata-qa:
ERROR:
 

ERROR: QA failed on smb_files_sha256.

Pipeline 7832

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1067 @@
...
+                            return Err(make_error(format!(
+                                "invalid rvalue value: must be between {} and {}: {}",
+                                1,
+                                u32::MAX,
+                                val
+                            )));
+                        }
+                    }
+                    ResultValue::String(val) => {
+                        byte_math.rvalue_str = CString::new(val).unwrap().into_raw();
```
jasonish, line 248: I think we should turn this into an error that is returned.  While I think its impossible for this to error out under normal use, fuzzing might find away to cause this to fail, then we'd need to fix it anyways.

jasonish (changes requested):
I think just that error handling thing mentioned, otherwise this is looking good to me now.

rust/src/detect_parser/byte_math.rs
jlucovsky: Ok ... your same comment applies to how `result` is built too.

jlucovsky:
Continued in #7552

PR #7550, opened 2022-06-16
fuzz: remove check about max transactions

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44798 and other

Describe changes:
- removes fuzzer's check about too many transactions

HTTP1 and other can indeed pipeline many transactions.
The protocols for which it caused a timeout now have their limit.
I do not think this check is relevant...
Ideas ?


suricata-qa:

Pipeline 7881

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7755

PR #7540, opened 2022-06-14
Notification of fully initialised Suricata v2

rmcconnell-r7:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5384

Replaces: #7508

Describe changes:
- Updated code to comply with standards
- Added systemd-notify doc to index list
- Updated TmThreadWaitOnThreadRunning function to ignore (not wait for) completed threads

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #7551, opened 2022-06-16
Notification of fully initialised Suricata v3

rmcconnell-r7:

- [ ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5384

Replaces: #7540 

Describe changes:
- Improved systemd-notify.rst document
- Updated Fedora-36 workflow to include libsystemd-devel libraries and check for compilation against libsystemd

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -31,6 +31,10 @@
 #include <signal.h>
 #endif
 
+#if HAVE_LIBSYSTEMD
+#include <systemd/sd-daemon.h>
+#endif
+
 #include "suricata.h"
 #include "decode.h"
 #include "feature.h"
@@ -452,6 +456,13 @@ static void GlobalsDestroy(SCInstance *suri)
 static TmEcode SendOSSpecificNotification(void)
 {
     SCLogNotice("All threads in running state.");
+
+#if HAVE_LIBSYSTEMD
+    if(sd_notify(0, "READY=1") < 0) {
+        SCLogWarning(SC_ERR_SYSCALL, "failed to send sd_notify()");
+        return TM_ECODE_FAILED;
+    }
+#endif
 
     return TM_ECODE_OK;
 }
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -458,7 +458,7 @@ static TmEcode SendOSSpecificNotification(void)
     SCLogNotice("All threads in running state.");
 
 #if HAVE_LIBSYSTEMD
-    if(sd_notify(0, "READY=1") < 0) {
+    if (sd_notify(0, "READY=1") < 0) {
         SCLogWarning(SC_ERR_SYSCALL, "failed to send sd_notify()");
         return TM_ECODE_FAILED;
     }
@@ -2960,7 +2960,7 @@ int SuricataMain(int argc, char **argv)
     TmThreadContinueThreads();
 
     /* Must ensure all threads are fully operational before contunuing with init process */
-    if(TmThreadWaitOnThreadRunning() != TM_ECODE_OK) {
+    if (TmThreadWaitOnThreadRunning() != TM_ECODE_OK) {
         exit(EXIT_FAILURE);
     }
 
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -444,6 +444,18 @@ static void GlobalsDestroy(SCInstance *suri)
     suri->pid_filename = NULL;
 }
 
+/**
+ * \brief Used to send OS specific notification of running threads
+ *
+ * \retval TmEcode TM_ECODE_OK on success; TM_ECODE_FAILED on failure.
+ */
+static TmEcode SendOSSpecificNotification(void)
+{
+    SCLogNotice("All threads in running state.");
+
+    return TM_ECODE_OK;
+}
+
 /** \brief make sure threads can stop the engine by calling this
  *  function. Purpose: pcap file mode needs to be able to tell the
  *  engine the file eof is reached. */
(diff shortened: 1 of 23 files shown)


src/tm-threads.c
```diff
@@ -1772,10 +1775,64 @@ void TmThreadWaitForFlag(ThreadVars *tv, uint32_t flags)
...
+                SCMutexUnlock(&tv_root_lock);
+
+                SCLogError(SC_ERR_THREAD_INIT,
+                        "thread \"%s\" failed to "
+                        "start: flags %04x",
+                        tv->name, SC_ATOMIC_GET(tv->flags));
+                return TM_ECODE_FAILED;
+            }
+
+            if (TmThreadsCheckFlag(tv, THV_RUNNING_DONE)) {
```
victorjulien, line 1807: if you check for `THV_RUNNING_DONE|THV_RUNNING` you can avoid the check on line 1812 and reduce level of indent which is always nice

src/tm-threads.c
```diff
@@ -2148,6 +2205,7 @@ void TmThreadsInitThreadsTimestamp(const struct timeval *ts)
     struct timeval systs;
     gettimeofday(&systs, NULL);
     SCMutexLock(&thread_store_lock);
+
```
victorjulien, line 2208: please avoid whitespace only changes

src/tm-threads.c
```diff
@@ -1772,10 +1775,64 @@ void TmThreadWaitForFlag(ThreadVars *tv, uint32_t flags)
...
+            if (TmThreadsCheckFlag(tv, THV_RUNNING_DONE)) {
+                tv = tv->next;
+                continue;
+            }
+
+            if (!(TmThreadsCheckFlag(tv, THV_RUNNING))) {
+                SCMutexUnlock(&tv_root_lock);
+
+                gettimeofday(&cur_ts, NULL);
+                if ((cur_ts.tv_sec - start_ts.tv_sec) > 60) {
```
jlucovsky, line 1816: Would prefer a `#define` instead of hard-coding `60` here.

src/suricata.c
```diff
@@ -2936,6 +2959,14 @@ int SuricataMain(int argc, char **argv)
     /* Un-pause all the paused threads */
     TmThreadContinueThreads();
 
+    /* Must ensure all threads are fully operational before contunuing with init process */
```
jlucovsky, line 2962: nit: typo `contunuing`

src/tm-threads.c
```diff
@@ -1772,10 +1775,64 @@ void TmThreadWaitForFlag(ThreadVars *tv, uint32_t flags)
...
+                    SCLogError(SC_ERR_THREAD_INIT,
+                            "thread \"%s\" failed to "
+                            "start in time: flags %04x",
+                            tv->name, SC_ATOMIC_GET(tv->flags));
+                    return TM_ECODE_FAILED;
+                }
+
+                /* sleep a little to give the thread some
+                 * time to start running */
+                SleepUsec(100);
```
jlucovsky, line 1826: Same comment re: using a `#define` instead of `100`.

victorjulien: this is a common pattern in similar loops, fyi

src/tm-threads.c
```diff
@@ -427,7 +427,11 @@ static void *TmThreadsSlotVar(void *td)
 
     StatsSetupPrivate(tv);
 
-    TmThreadsSetFlag(tv, THV_INIT_DONE);
+    // Each 'worker' thread uses this func to process/decode the packet read.
+    // Each decode method is different to receive methods in that they do not
+    // enter infinite loops. They use this as the core loop. As a result, at this
+    // point the worker threads can be considered both initialized and running.
+    TmThreadsSetFlag(tv, THV_INIT_DONE | THV_RUNNING);
```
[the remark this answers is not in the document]
rmcconnell-r7: In this exact instance you are correct, they mean the same thing for worker threads. The worker threads do not enter an infinite loop, thus ``INIT_DONE=RUNNING`` in this instance. The ``RUNNING`` is designed for the init thread to ensure all threads are fully operational, not only initialised. Mainly to ensure all PPT threads are in a ``RUNNING`` state i.e. just before entering their ``while()``.

src/tm-threads.c
rmcconnell-r7: I followed previous examples, see ``TmThreadWaitOnThreadInit``. I am more than happy to ``#define`` if that is how design will be going forward?

src/tm-threads.c
rmcconnell-r7: Would you rather me change these to ``#define`` or leave them? My implementation follows current implmentation.

victorjulien: I don't see it as something we need to do as part of the PR

src/tm-threads.c
jlucovsky: Leave as is but please add a comment as to why 60 seconds is the correct value to use here.

src/tm-threads.c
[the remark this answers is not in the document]
jlucovsky: The steps between `INIT_DONE` and `RUNNING` are not part of the thread initialization proper.

My observation is that since both `RUNNING` and `INIT_DONE` are set at the same time, there's no distinction between the two states. If there were, `RUNNING` would be deferred until the thread is fully operational.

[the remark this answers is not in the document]
rmcconnell-r7: Yes for the worker threads that is the case. However if you see any of the packet processing threads/methods they set``RUNNING`` later in the process. The main aim here is to ensure they are fully operational. And to ensure they are in a capturing state before notifying suricata is operational.

rmcconnell-r7:
Replaced by: #7564

PR #7555, opened 2022-06-17
nfs4: fix write record handling

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5280#change-23168

Describe changes:
- nfs4: fix write record handling

See also commit 4418fc1b02f47533439fe00789d9c850a24271b2

For what I understand, with NFS4 (instead of NFS3), we always have complete data with these compound records PDU...

And I also understand that with a `WRITEv4` we cannot know if there are more data to be written afterwards...

Thoughts ?

### DIFF
--- a/rust/src/nfs/nfs4.rs
+++ b/rust/src/nfs/nfs4.rs
@@ -52,6 +52,11 @@ impl NFSState {
             fill_bytes = 4 - pad;
         }
 
+        // linux defines a max of 1mb. Allow several multiples.
+        if w.write_len == 0 || w.write_len > 16777216 {
+            return;
+        }
+
         let file_handle = fh.to_vec();
         let file_name = if let Some(name) = self.namemap.get(fh) {
             SCLogDebug!("WRITE name {:?}", name);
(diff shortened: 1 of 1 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_asan | | |
 

Pipeline 7887

rust/src/nfs/nfs4.rs
```diff
@@ -52,6 +52,11 @@ impl NFSState {
             fill_bytes = 4 - pad;
         }
 
+        // linux defines a max of 1mb. Allow several multiples.
+        if w.write_len == 0 || w.write_len > 16777216 {
```
jlucovsky, line 56: If the limit is `1MB` why compare against `16MB`?

catenacyber: eheh copied pasted from nfs.rs
@victorjulien do you recall if it is 1 or 16 ?

victorjulien: Linux is 1MiB but I didn't find much info about other implementations, so I thought I'd take some multiple.

victorjulien (approved):

PR #7552, opened 2022-06-16
detect/bytemath: Support out of order options

jlucovsky:
Continuation of #7541 

The intent of this PR is to support arbitrary order options for `byte_math`. During the investigation phase, it was suggested that the parser be converted to Rust for simplification. Thus, the PCRE based parser for `byte_math` was removed and replaced with a Rust based parser. @jasonish's prototype rulers parser work helped guide this PR.

The C unittests were retained and many were added to the Rust based parser.

Issue: 5077

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5077](https://redmine.openinfosecfoundation.org/issues/5077)

Describe changes:
- Converts the PCRE based parser to Rust.
- Adds unit tests to the new Rust modules
- Removes the PCRE parser from detect-bytemath.c
- Adjusts the C source modules to refer to the Rust definitions

Updates
- Handle errors from `CString::new` 

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #7565, opened 2022-06-22

jlucovsky:
Continuation of #7552 

The intent of this PR is to support arbitrary order options for `byte_math`. During the investigation phase, it was suggested that the parser be converted to Rust for simplification. Thus, the PCRE based parser for `byte_math` was removed and replaced with a Rust based parser. @jasonish's prototype rules parser work helped guide this PR.

Issue: 5077

Describe changes:

Updates
- Reword message when conversion of `CString` fails

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/rust/src/detect_parser/byte_math.rs
+++ b/rust/src/detect_parser/byte_math.rs
@@ -1,0 +1,1083 @@
+/* Copyright (C) 2022 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+// Author: Jeff Lucovsky <jlucovsky@oisf.net>
+
+use crate::detect_parser::error::RuleParseError;
+use crate::detect_parser::parser::{parse_token, take_until_whitespace};
+use std::ffi::{CStr, CString};
+use std::os::raw::c_char;
+
+use nom7::bytes::complete::tag;
+use nom7::character::complete::multispace0;
+use nom7::sequence::preceded;
+use nom7::{Err, IResult};
+use std::str;
+
+pub const DETECT_BYTEMATH_FLAG_RELATIVE: u32 = 0x01;
+pub const DETECT_BYTEMATH_FLAG_STRING: u32 = 0x02;
+pub const DETECT_BYTEMATH_FLAG_BITMASK: u32 = 0x04;
+pub const DETECT_BYTEMATH_FLAG_ENDIAN: u32 = 0x08;
+pub const DETECT_BYTEMATH_FLAG_RVALUE_VAR: u32 = 0x10;
+
+// Ensure required values are provided
+const DETECT_BYTEMATH_FLAG_NBYTES: u32 = 0x1;
+const DETECT_BYTEMATH_FLAG_OFFSET: u32 = 0x2;
+const DETECT_BYTEMATH_FLAG_OPER: u32 = 0x4;
+const DETECT_BYTEMATH_FLAG_RVALUE: u32 = 0x8;
+const DETECT_BYTEMATH_FLAG_RESULT: u32 = 0x10;
+const DETECT_BYTEMATH_FLAG_REQUIRED: u32 = DETECT_BYTEMATH_FLAG_RESULT
+    | DETECT_BYTEMATH_FLAG_RVALUE
+    | DETECT_BYTEMATH_FLAG_NBYTES
+    | DETECT_BYTEMATH_FLAG_OFFSET
+    | DETECT_BYTEMATH_FLAG_OPER;
+
+// operator: +, -, /, *, <<, >>
+pub const DETECT_BYTEMATH_OPERATOR_NONE: u8 = 1;
+pub const DETECT_BYTEMATH_OPERATOR_PLUS: u8 = 2;
+pub const DETECT_BYTEMATH_OPERATOR_MINUS: u8 = 3;
+pub const DETECT_BYTEMATH_OPERATOR_DIVIDE: u8 = 4;
+pub const DETECT_BYTEMATH_OPERATOR_MULTIPLY: u8 = 5;
+pub const DETECT_BYTEMATH_OPERATOR_LSHIFT: u8 = 6;
+pub const DETECT_BYTEMATH_OPERATOR_RSHIFT: u8 = 7;
+
+// endian <big|little|dce>
+pub const DETECT_BYTEMATH_ENDIAN_NONE: u8 = 0;
+pub const DETECT_BYTEMATH_ENDIAN_BIG: u8 = 1;
+pub const DETECT_BYTEMATH_ENDIAN_LITTLE: u8 = 2;
+pub const DETECT_BYTEMATH_ENDIAN_DCE: u8 = 3;
+pub const DETECT_BYTEMATH_ENDIAN_DEFAULT: u8 = DETECT_BYTEMATH_ENDIAN_BIG;
+
+pub const DETECT_BYTEMATH_BASE_HEX: u8 = 16;
+pub const DETECT_BYTEMATH_BASE_DEC: u8 = 10;
+pub const DETECT_BYTEMATH_BASE_OCT: u8 = 8;
+pub const DETECT_BYTEMATH_BASE_NONE: u8 = 0;
+pub const DETECT_BYTEMATH_BASE_DEFAULT: u8 = DETECT_BYTEMATH_BASE_DEC;
+
+// Fixed position parameter count: bytes, offset, oper, rvalue, result
+// result is not parsed with the fixed position parameters as it's
+// often swapped with optional parameters
+pub const DETECT_BYTEMATH_FIXED_PARAM_COUNT: usize = 5;
+// Optional parameters: endian, relative, string, dce, bitmask
+pub const DETECT_BYTEMATH_MAX_PARAM_COUNT: usize = 10;
+
+#[derive(Debug)]
+enum ResultValue {
+    Numeric(u64),
+    String(String),
+}
+
+#[derive(Debug)]
+#[repr(C)]
+pub struct DetectByteMathData {
+    rvalue_str: *const c_char,
+    result: *const c_char,
+    rvalue: u32,
+    flags: u32,
+    offset: i32,
+    bitmask_val: u32,
+    bitmask_shift_count: u16,
+    id: u16,
+    local_id: u8,
+    nbytes: u8,
+    oper: u8,
+    endian: u8, // big, little, dce
+    base: u8,   // From string or dce
+}
+
+impl Drop for DetectByteMathData {
+    fn drop(&mut self) {
+        unsafe {
+            if !self.result.is_null() {
+                let _ = CString::from_raw(self.result as *mut c_char);
+            }
+            if !self.rvalue_str.is_null() {
+                let _ = CString::from_raw(self.rvalue_str as *mut c_char);
+            }
+        }
+    }
+}
+
+impl Default for DetectByteMathData {
+    fn default() -> Self {
+        DetectByteMathData {
+            local_id: 0,
+            flags: 0,
+            nbytes: 0,
+            offset: 0,
+            oper: 0,
+            rvalue_str: std::ptr::null_mut(),
+            rvalue: 0,
+            result: std::ptr::null_mut(),
+            endian: DETECT_BYTEMATH_ENDIAN_DEFAULT,
+            base: DETECT_BYTEMATH_BASE_DEFAULT,
+            bitmask_val: 0,
+            bitmask_shift_count: 0,
+            id: 0,
+        }
+    }
+}
+
+impl DetectByteMathData {
+    pub fn new() -> Self {
+        Self {
+            ..Default::default()
+        }
+    }
+}
+
+fn get_string_value(value: &str) -> Result<u8, ()> {
+    let res = match value {
+        "hex" => DETECT_BYTEMATH_BASE_HEX,
+        "oct" => DETECT_BYTEMATH_BASE_OCT,
+        "dec" => DETECT_BYTEMATH_BASE_DEC,
+        _ => return Err(()),
+    };
+
+    Ok(res)
+}
+
+fn get_oper_value(value: &str) -> Result<u8, ()> {
+    let res = match value {
+        "+" => DETECT_BYTEMATH_OPERATOR_PLUS,
+        "-" => DETECT_BYTEMATH_OPERATOR_MINUS,
+        "/" => DETECT_BYTEMATH_OPERATOR_DIVIDE,
+        "*" => DETECT_BYTEMATH_OPERATOR_MULTIPLY,
+        "<<" => DETECT_BYTEMATH_OPERATOR_LSHIFT,
+        ">>" => DETECT_BYTEMATH_OPERATOR_RSHIFT,
+        _ => return Err(()),
+    };
+
+    Ok(res)
+}
+
+fn get_endian_value(value: &str) -> Result<u8, ()> {
+    let res = match value {
+        "big" => 1,
+        "little" => 2,
+        _ => return Err(()),
+    };
+
+    Ok(res)
+}
+
+// Parsed as a u64 for validation with u32 {min,max}
+fn parse_rvalue(input: &str) -> IResult<&str, ResultValue, RuleParseError<&str>> {
+    let (input, rvalue) = parse_token(input)?;
+    if let Ok(val) = rvalue.parse::<u64>() {
+        Ok((input, ResultValue::Numeric(val)))
+    } else {
+        Ok((input, ResultValue::String(rvalue.to_string())))
+    }
+}
+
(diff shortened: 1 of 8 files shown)


catenacyber:
By the way @jlucovsky I see that the multiplication is missing in the current regexp

as in 
```
-    ",\\s*(oper)\\s*([-+\\/]{1}|<<|>>)\\s*"                                                        \
+    ",\\s*(oper)\\s*([-+\\/\\*]{1}|<<|>>)\\s*"                                                    \
```

catenacyber (changes requested):
- CI : green but needs rebase
- Code : Will do right now
- Commits segmentation : ok
- Commit message : 
Please add that this commit fixes uses of multiplication in byte math
I do not think that we capitalize the title ie s/Convert/convert
I think the template is `Ticket: #5077`, at the end of the commit message
- Git ID set : ok
- CLA : ok ;-)
- Doc update : none needed
- Redmine ticket : I think the template is `Ticket: #5077`, at the end of the message
- Rustfmt : `rustfmt rust/src/detect_parser/*.rs` looks good :-)
- Tests : I guess I would like to see a SV test about the fix of use of multiplication in bytemath
- Dependencies added: none

rust/src/detect_parser/error.rs
```diff
@@ -0,0 +1,37 @@
+/* Copyright (C) 2022 Open Information Security Foundation
```
catenacyber, line 1: I would have chosen `detect` instead of `detect_parser` and moved detect.rs into it (in its own commit)

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+use std::ffi::{CStr, CString};
+use std::os::raw::c_char;
+
+use nom7::bytes::complete::tag;
+use nom7::character::complete::multispace0;
+use nom7::sequence::preceded;
+use nom7::{Err, IResult};
+use std::str;
+
+pub const DETECT_BYTEMATH_FLAG_RELATIVE: u32 = 0x01;
```
catenacyber, line 31: Should we use `BIT_U8` ?

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+pub const DETECT_BYTEMATH_FLAG_STRING: u32 = 0x02;
+pub const DETECT_BYTEMATH_FLAG_BITMASK: u32 = 0x04;
+pub const DETECT_BYTEMATH_FLAG_ENDIAN: u32 = 0x08;
+pub const DETECT_BYTEMATH_FLAG_RVALUE_VAR: u32 = 0x10;
+
+// Ensure required values are provided
+const DETECT_BYTEMATH_FLAG_NBYTES: u32 = 0x1;
```
catenacyber, line 38: I would have gone with a struct with 5 booleans

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+const DETECT_BYTEMATH_FLAG_RVALUE: u32 = 0x8;
+const DETECT_BYTEMATH_FLAG_RESULT: u32 = 0x10;
+const DETECT_BYTEMATH_FLAG_REQUIRED: u32 = DETECT_BYTEMATH_FLAG_RESULT
+    | DETECT_BYTEMATH_FLAG_RVALUE
+    | DETECT_BYTEMATH_FLAG_NBYTES
+    | DETECT_BYTEMATH_FLAG_OFFSET
+    | DETECT_BYTEMATH_FLAG_OPER;
+
+// operator: +, -, /, *, <<, >>
+pub const DETECT_BYTEMATH_OPERATOR_NONE: u8 = 1;
```
catenacyber, line 50: Can you make it an enum as `SSHConnectionState` ?

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+pub const DETECT_BYTEMATH_OPERATOR_MULTIPLY: u8 = 5;
+pub const DETECT_BYTEMATH_OPERATOR_LSHIFT: u8 = 6;
+pub const DETECT_BYTEMATH_OPERATOR_RSHIFT: u8 = 7;
+
+// endian <big|little|dce>
+pub const DETECT_BYTEMATH_ENDIAN_NONE: u8 = 0;
+pub const DETECT_BYTEMATH_ENDIAN_BIG: u8 = 1;
+pub const DETECT_BYTEMATH_ENDIAN_LITTLE: u8 = 2;
+pub const DETECT_BYTEMATH_ENDIAN_DCE: u8 = 3;
+pub const DETECT_BYTEMATH_ENDIAN_DEFAULT: u8 = DETECT_BYTEMATH_ENDIAN_BIG;
```
catenacyber, line 63: Can you make it an enum, and use the `default` trait ?

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+        _ => return Err(()),
+    };
+
+    Ok(res)
+}
+
+fn get_endian_value(value: &str) -> Result<u8, ()> {
+    let res = match value {
+        "big" => 1,
+        "little" => 2,
```
catenacyber, line 171: Is not `dce` missing ?

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+        _ => return Err(()),
+    };
+
+    Ok(res)
+}
+
+// Parsed as a u64 for validation with u32 {min,max}
+fn parse_rvalue(input: &str) -> IResult<&str, ResultValue, RuleParseError<&str>> {
+    let (input, rvalue) = parse_token(input)?;
```
catenacyber, line 180: You can use nom7 `map_res` and `alt`

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+    // Inner utility function for easy error creation.
+    fn make_error(reason: String) -> nom7::Err<RuleParseError<&'static str>> {
+        Err::Error(RuleParseError::InvalidByteMath(reason))
+    }
+    let (_, values) = nom7::multi::separated_list1(
+        tag(","),
+        preceded(multispace0, nom7::bytes::complete::is_not(",")),
+    )(input)?;
+
+    if values.len() < DETECT_BYTEMATH_FIXED_PARAM_COUNT
```
catenacyber, line 198: This may be done with nom7 `verify`
But I do not know if we get a good error this way

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+        || values.len() > DETECT_BYTEMATH_MAX_PARAM_COUNT
+    {
+        return Err(make_error(format!("Incorrect argument string; at least {} values must be specified but no more than {}: {:?}",
+            DETECT_BYTEMATH_FIXED_PARAM_COUNT, DETECT_BYTEMATH_MAX_PARAM_COUNT, input)));
+    }
+
+    let mut required_flags: u32 = 0;
+    let mut byte_math = DetectByteMathData::new();
+    for value in &values[0..] {
```
catenacyber, line 207: `&values[0..]` looks oddly overcomplicated

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+        let (val, name) = take_until_whitespace(value)?;
+        match name.trim() {
+            "oper" => {
+                byte_math.oper = match get_oper_value(val.trim()) {
```
catenacyber, line 211: I would love to remove multiple uses of `trim`...

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+                    }
+                    _ => {
+                        return Err(make_error(
+                            "parse string not safely convertible to C".to_string(),
+                        ));
+                    }
+                }
+            }
+            "rvalue" => {
+                let (_, res) = match parse_rvalue(val.trim()) {
```
catenacyber, line 237: Would be good if we could make `res = parse_rvalue(val.trim())?`; to avoid the lines handling error

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+                    Ok(val) => val,
+                    Err(_) => {
+                        return Err(make_error(format!("invalid rvalue value: {}", val)));
+                    }
+                };
+                match res {
+                    ResultValue::Numeric(val) => {
+                        if val > u32::MIN.into() && val <= u32::MAX.into() {
```
catenacyber, line 245: I think this should have been done in a nom7 `verify` by `parse_rvalue`

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+                            return Err(make_error(
+                                "parse string not safely convertible to C".to_string(),
+                            ))
+                        }
+                    },
+                }
+                required_flags |= DETECT_BYTEMATH_FLAG_RVALUE;
+            }
+            "endian" => {
+                if 0 != (byte_math.flags & DETECT_BYTEMATH_FLAG_ENDIAN) {
```
catenacyber, line 271: Why check for `endian` and not for multiple `oper` ?

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+            "dce" => {
+                if 0 != (byte_math.flags & DETECT_BYTEMATH_FLAG_ENDIAN) {
+                    return Err(make_error("endianess already set".to_string()));
+                }
+                byte_math.flags |= DETECT_BYTEMATH_FLAG_ENDIAN;
+                byte_math.endian = DETECT_BYTEMATH_ENDIAN_DCE;
+            }
+            "bitmask" => {
+                let val = val.trim();
+                let trimmed = if val.starts_with("0x") || val.starts_with("0X") {
```
catenacyber, line 303: use nom7 `opt(tag)` so skip these

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+    if (required_flags & DETECT_BYTEMATH_FLAG_REQUIRED) != DETECT_BYTEMATH_FLAG_REQUIRED {
+        return Err(make_error(format!(
+            "required byte_math parameters missing: \"{:?}\"",
+            input
+        )));
+    }
+
+    match byte_math.oper {
+        DETECT_BYTEMATH_OPERATOR_LSHIFT | DETECT_BYTEMATH_OPERATOR_RSHIFT => {
+            if byte_math.nbytes > 4 {
```
catenacyber, line 356: also check if value is 0 ?

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+    };
+    match parse_bytemath(arg) {
+        Ok((_, detect)) => return Box::into_raw(Box::new(detect)) as *mut DetectByteMathData,
+        Err(_) => return std::ptr::null_mut(),
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn ScByteMathFree(ptr: *mut DetectByteMathData) {
+    if !ptr.is_null() {
```
catenacyber, line 387: Do we need the null ptr check ?

rust/src/detect_parser/byte_math.rs
```diff
@@ -0,0 +1,1083 @@
...
+        let _ = Box::from_raw(ptr as *mut DetectByteMathData);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    // structure equality only used by test cases
+    impl PartialEq for DetectByteMathData {
+        fn eq(&self, other: &Self) -> bool {
```
catenacyber, line 397: There must be a rust trait for this

catenacyber:
Looks a great feature overall.
I left some rust comments.

I think it would be useful for other keywords to have a generic function that breaks out a string into the list of sub-keyword and optional) values with spaces trimmed (and maybe check that these keywords are unique)

rust/src/detect_parser/byte_math.rs
jlucovsky: `dce` is a separate keyword that implies endianness. The `endian` keyword options are only `big` and `little`

rust/src/detect_parser/byte_math.rs
jlucovsky: `nbytes` has already been validated to be in the range `[1, 10]` inclusive so the only check needed here is against the maximum value.

rust/src/detect_parser/byte_math.rs
jlucovsky: Will update to use u8.

rust/src/detect_parser/byte_math.rs
jlucovsky: Is there an advantage to using a `struct` over flags?

rust/src/detect_parser/byte_math.rs
jlucovsky: Will do.

rust/src/detect_parser/byte_math.rs
catenacyber: I am not sure about this : this was rather to start a discussion.
Within a struct, we see that these booleans are related...

Anyways, this is not at all critical for me

rust/src/detect_parser/byte_math.rs
catenacyber: From some point of view, as `DETECT_BYTEMATH_ENDIAN_DCE` is part of the same enum as `DETECT_BYTEMATH_ENDIAN_BIG`, I think it would make sense to add support for `endian dce`

What do you think ?

rust/src/detect_parser/byte_math.rs
catenacyber: Indeed, maybe a comment can be added ;-)

rust/src/detect_parser/error.rs
jlucovsky: Good idea.

rust/src/detect_parser/byte_math.rs
jlucovsky: That's not a bad idea but it is also deviates from the Snort syntax.

I don't mind doing it but it'd force rule writers to create 2 rules if they support it.

catenacyber: >  it'd force rule writers to create 2 rules if they support it.

My idea was rather to give them the choice between two syntaxes : old one, and one who seems more logic to me ;-)

rust/src/detect_parser/byte_math.rs
jlucovsky: Will do.

rust/src/detect_parser/byte_math.rs
jlucovsky: Technically -- no, the caller could do that but a check for non-NULL is required somewhere. I opted to put it into the lowest-level function so others wouldn't have to do it.

rust/src/detect_parser/byte_math.rs
jlucovsky: `trim` is invoked for different objects ... the bytemath keyword and the bytemath keyword value.

rust/src/detect_parser/byte_math.rs
jlucovsky: I can make it an enum but AFAIK using `default` with enums is an experimental feature.

rust/src/detect_parser/byte_math.rs
jlucovsky: Ok ... I'm going to leave this as is.

rust/src/detect_parser/byte_math.rs
catenacyber: Yes the trimming of the value is done in every case of the switch.
It would be good to have that preprocessed ;-)

rust/src/detect_parser/byte_math.rs
jlucovsky: I'm going to keep the current code.

rust/src/detect_parser/byte_math.rs
jlucovsky: Will do.

rust/src/detect_parser/byte_math.rs
jlucovsky: Ok. Will add.

rust/src/detect_parser/byte_math.rs
jlucovsky: I'll extend the duplicate check to all of the keywords.

rust/src/detect_parser/byte_math.rs
jlucovsky: I'm going to leave this as is to retain the error message.

rust/src/detect_parser/byte_math.rs
jasonish: You can usually dervide `Eq` and/or `PartialEq` for this.  Not always possible though.

rust/src/detect_parser/byte_math.rs
jasonish: I'm not sure.  I think the idea of `parse_rvalue` is to parse a string or a number, where parsing the largest possible number makes sense.  This is a likely candidate for re-use, so further validation on bounds, when it is a number should probably be done outside of `parse_rvalue`.

And if we want to log the friendliest error messages possible, the chaining with `verify` can get quite ugly.

Post 7.0 I see breaking out common rule parsing stuff to its own crate where we can generic-fy the patterns we see come up.

rust/src/detect_parser/byte_math.rs
jlucovsky: I'm going to leave this as is for now.

jlucovsky:
Continued in #7694

PR #7563, opened 2022-06-20
src: rework includes as per cppclean

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
Is there already one @jasonish ?

Describe changes:
- Rework includes as per cppclean 

Still to do : add `cppclean` as part of CI

I used this patched version https://github.com/myint/cppclean/pull/165

There are still exceptions :
- About Suricata : When "rust.h" or "suricata-common.h" gets included, and when includes is rust.h or rust-context.h
- About cppclean 
  - when using macros such as `TAILQ` with including queue.h (or red and black tree)
  - when using the array size, such as ``uint8_t plugin_v[PLUGIN_VAR_SIZE];  in decode.h needing suricata-plugin.h

The result is less includes from header file to another, nothing is done for C files including a useless header

PR #7570, opened 2022-06-27
Include less v2

catenacyber:
Should I add one ?

Describe changes:
- Rework includes as per cppclean 

Still to do : add `cppclean` as part of CI

I used this patched version https://github.com/myint/cppclean/pull/165

There are still exceptions :
- About Suricata : When "rust.h" or "suricata-common.h" gets included, and when includes is rust.h or rust-context.h
- About cppclean 
  - when using macros such as `TAILQ` with including queue.h (or red and black tree)
  - when using the array size, such as ``uint8_t plugin_v[PLUGIN_VAR_SIZE];  in decode.h needing suricata-plugin.h

The result is less includes from header file to another, nothing is done for C files including a useless header

Modifies #7563 with compilation fixed for Victor's options and adding one commit removing 3 unused header files

This PR adds about one hundred lines `#include "detect-engine-build.h"` because it is being used in util tests for `SigGroupBuild` and such... Any thoughts on that ?


suricata-qa:
ERROR:
 

Pipeline 7997

catenacyber:
@jasonish I would like your thoughts before rebasing (as it will likely be painful)

jasonish:
I do like where this is going: removals from .h files, and more added to the .c files makes sense.

The Rust headers are super includes by nature, not much we can do about that I don't think.

Is there much manual work involved? Or is it just run the tool?

catenacyber:
> Is there much manual work involved? Or is it just run the tool?

The tool just spots useless includes in header files.
Then, you have to manually put this include in the C files needing it...

PR #7578, opened 2022-06-28
smtp: adds server side detection

catenacyber:

Describe changes:
- smtp : adds server side detection

Draft for feedback ;-)

The most special trick is that the (server) probing parser waits for the client side to have seen some data to take a definitive positive decision.
So that If it looks like a SMTP server (it could be a FTP server), let's see if the client looks like SMTP or FTP or something unknown...

Still to do :
- S-V tests upgrades
- FTP server side detection

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -1618,6 +1618,50 @@ static int SMTPStateGetEventInfoById(int event_id, const char **event_name,
     return 0;
 }
 
+static AppProto SMTPServerProbingParser(
+        Flow *f, uint8_t direction, const uint8_t *input, uint32_t len, uint8_t *rdir)
+{
+    // another check for minimum length
+    if (len < 5) {
+        return ALPROTO_UNKNOWN;
+    }
+    // begins by 220
+    if (input[0] != '2' || input[1] != '2' || input[2] != '0') {
+        return ALPROTO_FAILED;
+    }
+    // followed by space or hypen
+    if (input[3] != ' ' && input[3] != '-') {
+        return ALPROTO_FAILED;
+    }
+    AppProto r = ALPROTO_UNKNOWN;
+    if (f->alproto_ts == ALPROTO_SMTP ||
+            (f->todstbytecnt > 4 && f->alproto_ts == ALPROTO_UNKNOWN)) {
+        // only validates SMTP if client side was SMTP
+        // or if client side is unknown despite having received bytes
+        r = ALPROTO_SMTP;
+    }
+    uint8_t state = 0; // domain
+    for (uint32_t i = 4; i < len; i++) {
+        switch (state) {
+            case 0:
+                if (isalpha(input[i]) || isdigit(input[i]) || input[i] == '.' || input[i] == '-') {
+                    // basic domain validation : continue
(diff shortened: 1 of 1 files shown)


src/app-layer-smtp.c
```diff
@@ -1635,6 +1682,18 @@ static int SMTPRegisterPatternsForProtocolDetection(void)
...
         return -1;
     }
+    if (!AppLayerProtoDetectPPParseConfPorts(
+                "tcp", IPPROTO_TCP, "smtp", ALPROTO_SMTP, 0, 5, NULL, SMTPServerProbingParser)) {
+        return;
+    }
+    if (AppLayerProtoDetectPMRegisterPatternCSwPP(IPPROTO_TCP, ALPROTO_SMTP, "220 ", 4, 0,
+                STREAM_TOCLIENT, SMTPServerProbingParser, 5, 5) < 0) {
+        return -1;
+    }
```
catenacyber, line 1692: We could have both or either : generic detection whatever the ports, or only on configured ports

It depends on what we want

src/app-layer-smtp.c
```diff
@@ -1618,6 +1618,53 @@ static int SMTPStateGetEventInfoById(int event_id, const char **event_name,
...
+        // or if client side is unknown despite having received bytes
+        r = ALPROTO_SMTP;
+    }
+    uint8_t state = 0; // domain
+    for (uint32_t i = 4; i < len; i++) {
+        switch
+            state
+            {
+                case 0:
+                    if (isalpha(input[i]) || isdigit(input[i]) || input[i] == '.' ||
```
catenacyber, line 1649: Is there already a function to validate a domain name ?

src/app-layer-smtp.c
```diff
@@ -1618,6 +1618,50 @@ static int SMTPStateGetEventInfoById(int event_id, const char **event_name,
...
+    }
+    AppProto r = ALPROTO_UNKNOWN;
+    if (f->alproto_ts == ALPROTO_SMTP ||
+            (f->todstbytecnt > 4 && f->alproto_ts == ALPROTO_UNKNOWN)) {
+        // only validates SMTP if client side was SMTP
+        // or if client side is unknown despite having received bytes
+        r = ALPROTO_SMTP;
+    }
+    uint8_t state = 0; // domain
+    for (uint32_t i = 4; i < len; i++) {
```
victorjulien, line 1644: would this consider `220-.-.-.\n` to be SMTP?

suricata-qa:
ERROR:
 ERROR: Invalid Signature config error in tlpr1_asan_cfg QA test

Pipeline 8012

src/app-layer-smtp.c
catenacyber: The advantage is that `(PROFTP)` is not recognized as a valid SMTP banner ;-)

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7588

src/app-layer-smtp.c
victorjulien: I don't recall one... seems non-trivial too.

catenacyber: Should something smarter be done ?

PR #7598, opened 2022-07-01
Detect filestore 5408 v1

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5408

Describe changes:
- Prevents segfault with forced filestore by using the flow's proto instead of the packet's one
- adds forced filestore to the fuzzing configuration

suricata-verify-pr: 867
https://github.com/OISF/suricata-verify/pull/867

### DIFF
--- a/src/output-file.c
+++ b/src/output-file.c
@@ -97,9 +97,9 @@ int OutputRegisterFileLogger(LoggerId id, const char *name, FileLogger LogFunc,
 static void CloseFile(const Packet *p, Flow *f, File *file)
 {
     DEBUG_VALIDATE_BUG_ON((file->flags & FILE_LOGGED) != 0);
-    void *txv = AppLayerParserGetTx(p->proto, f->alproto, f->alstate, file->txid);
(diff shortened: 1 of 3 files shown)


catenacyber:
PS : Found another dangerous pattern with `git grep "f->alproto" | grep "p->proto"` beyond the one being triggered by the S-V test

victorjulien:
> PS : Found another dangerous pattern with `git grep "f->alproto" | grep "p->proto"` beyond the one being triggered by the S-V test

Did you fix this one yet?

catenacyber:
> Did you fix this one yet?

yes it is the one in output-file.c

victorjulien:
While this fixes the immediate issue of the crash, I worry about the effects of actually triggering logging in the context of an ICMP packet. For example the EVE file logger uses the `Packet` pointer to create the EVE header, which will then be the ICMP packet.

I wonder if we should just skip file/filedata APIs completely for these packets and let flow timeout worry about the completeness of logging.

Still think these changes make sense, just fear side effects that need some additional thinking.

suricata-qa:

Pipeline 8087

PR #6291, opened 2021-08-11
fix: closing bracket was at wrong place

Shchelk:
This bug was found by svace static code analyzer

warn: OP_PRECEDENCE_ASSIGN_CMP
msg: Logic operator precedence in assignment can produce an unexpected result

PR #7614, opened 2022-07-06
threads: cleaner code with one instruction per line

catenacyber:
None

Describe changes:
- better code style in threads manager

Replaces #6291


suricata-qa:

Pipeline 8135

catenacyber:
> The change is OK but there are other occurrences in the source module that follow the pattern you're changing.
> 
> Suggest those are changed as part of this PR.

Where do you see another one ?
`git grep if | grep " = " | grep TM_ECODE` yields no result
`git grep -A1 "!= TM_ECODE_OK" | grep TM_ECODE_FAILED` yields no result neither

There are like 2 bad patterns fixed here :
- assigning and checking on the same line
- checking `r != TM_ECODE_OK` only to check `r == TM_ECODE_FAILED` instead of just checking the second case

victorjulien (approved):

victorjulien:
Merged in #7653, thanks!

PR #7623, opened 2022-07-12
stream/event: Add event when stream reassembly depth has been reached

jlucovsky:
This PR adds a new stream event triggered when the stream reassembly depth has been reached.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3512](https://redmine.openinfosecfoundation.org/issues/3512)

Describe changes:
- Define event
- Add sample rule utilizing event
- Raise event when the stream reassembly depth is first reached on a stream.

suricata-verify-pr: 879
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/stream-tcp-reassemble.c
+++ b/src/stream-tcp-reassemble.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2021 Open Information Security Foundation
+/* Copyright (C) 2007-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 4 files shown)


suricata-qa:

Pipeline 8223

src/stream-tcp-reassemble.c
```diff
@@ -728,10 +728,14 @@ int StreamTcpReassembleHandleSegmentHandleData(ThreadVars *tv, TcpReassemblyThre
 
     /* If we have reached the defined depth for either of the stream, then stop
        reassembling the TCP session */
+    bool depth_reached = (stream->flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED);
     uint32_t size = StreamTcpReassembleCheckDepth(ssn, stream, TCP_GET_SEQ(p), p->payload_len);
     SCLogDebug("ssn %p: check depth returned %"PRIu32, ssn, size);
 
     if (stream->flags & STREAMTCP_STREAM_FLAG_DEPTH_REACHED) {
+        if (!depth_reached)
```
victorjulien, line 736: I don't understand this, if we get here multiple times the counter would also be incremented more than once per stream?

jlucovsky: The thinking is that one event per stream instead of once per occurrence.

What's your preference?

victorjulien: The logic is supposed to be once per stream both for the event and the counter increment.

jlucovsky: Will adjust to only count and raise the event the first time the limit is reached.

victorjulien: It *should* already work that way. If not, there is a bug in the existing code.

jlucovsky:
Continued in #7680

PR #7625, opened 2022-07-13
file/swf: Use lzma-rs decompression instead of libhtp.

cccs-rtmorti:
Use the lzma-rs crate for decompressing swf/lzma files instead of
the lzma decompressor in libhtp. This decouples suricata from libhtp
except for actual http parsing, and means libhtp no longer has to
export a lzma decompression interface.

- [ X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ X] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [ X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Reference [libhtp-rs PR comment](https://github.com/OISF/suricata/pull/7438#discussion_r887130753) suggesting this change.

Describe changes:
- Create a lzma.rs file which exports a lzma decompression function interface.
- In util-file-swf-decompression.c, use this interface instead of the lzma decompression interface presented by libhtp.
- Update error types / constants to reflect lzma-rs error conditions / results.


victorjulien:
Thanks Todd. How does this deal with the issue of lzma preallocating some attacker controlled u32 value? The reason we vendored the lzma code was to address this.

cccs-rtmorti:
> Thanks Todd. How does this deal with the issue of lzma preallocating some attacker controlled u32 value? The reason we vendored the lzma code was to address this.

We actually sent the lzma-rs project some patches to add limits to prevent this kind of attack. It was part of our assessment for whether or not we could use this crate in libhtp-rs for lzma decompression.

https://github.com/gendx/lzma-rs/commit/fcbb11fab6f8389067a266f40a74d15f02bf30ea

jasonish:
Looks like we could drop the check for zlib in configure.ac?

catenacyber:
> Looks like we could drop the check for zlib in configure.ac?

Why ? zlib will still be used in libhtp

jasonish:
> Why ? zlib will still be used in libhtp

libhtp has its own check for zlib which in the end results in `suricata` dynamically linking with it anyways.

cccs-rtmorti:
> Looks like we could drop the check for zlib in configure.ac?

util-file-swf-decompression still uses zlib to decompress CWS compressed swf archives. This PR only changes the lzma decompression path. zlib is still directly used for the CWS path, and this file includes zlib.h.

src/util-file-swf-decompression.c
```diff
@@ -32,9 +32,9 @@
 #include "util-misc.h"
 #include "util-print.h"
 
-#include <zlib.h>
+#include "rust.h"
 
-#include <htp/lzma/LzmaDec.h>
+#include <zlib.h>
```
jasonish, line 37: Can the zlib.h include be removed?

cccs-rtmorti: I don't think so. It is still used in the CWS decompression path.

jasonish: Ok, make sense.  Even if removed from this file, its pulled in via one of the other includes, so we shouldn't remove it.

catenacyber:
Reposting as it is lost in GitHub flow :

@victorjulien if I remember correctly, you mentioned that this whole code (SWF decompression) should be deprecated/removed at some point.

Should it be removed now ?

victorjulien:
> Reposting as it is lost in GitHub flow :
> 
> @victorjulien if I remember correctly, you mentioned that this whole code (SWF decompression) should be deprecated/removed at some point.
> 
> Should it be removed now ?

I don't think so, but it should probably not be enabled (at runtime) by default as flash is supposed to be a thing of the past.

catenacyber:
> > if I remember correctly, you mentioned that this whole code (SWF decompression) should be deprecated/removed at some point.
> > Should it be removed now ?
> 
> I don't think so, but it should probably not be enabled (at runtime) by default as flash is supposed to be a thing of the past.

@cccs-rtmorti could you add that to your PR ? That is change the default suricata.yaml config option for `swf-decompression` fields

cccs-rtmorti:
> @cccs-rtmorti could you add that to your PR ? That is change the default suricata.yaml config option for `swf-decompression` fields

Yeah, no problem.

PR #7634, opened 2022-07-20
6.0.x: Fix read the docs

jasonish:
6.0.6 was tagged and push with the addition of our requirements.txt file for handling Suricata requirements which is not compatible with Python requirements.txt, and Readthedocs tries to use this by default.  We addressed this in master with a `.readthedocs.yml` file, but that never made it into `master-6.0.x`, resulting in the documentation for `suricata-6.0.6` failing to build.

Current readthedocs is pinned to 6.0.5.

We could leave it pinned at 6.0.5 until 6.0.7.  Or make a new tag, `suricata-6.0.6-rtd` which would be this PR minus to version change commit to have it more up to date.  I don't think anything can be done at the RTD site, but will take another look.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@
 *.rules
 *.pcap
 *.pcapng
-/*.yaml
+/suricata.yaml
 Makefile
 .deps/
 .libs/
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,9 @@
+# Required by Read The Docs
+version: 2
+
+python:
+  version: "3.8"
+
+  # Use an empty install section to avoid RTD from picking up a non-python
+  # requirements.txt file.
+  install: []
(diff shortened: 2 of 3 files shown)


jasonish:
Ok, temporary fix was to point Readthedocs at `/threshold.conf` via the readthedocs settings control panel.  As this file is comments only, its accepted cleanly as a Python style `requirements.txt` file.

catenacyber:
CIFuzz failure is commit df2cbd6517dc751715213a4839581fb98c77af0d which was not backported

jasonish:
> CIFuzz failure is commit [df2cbd6](https://github.com/OISF/suricata/commit/df2cbd6517dc751715213a4839581fb98c77af0d) which was not backported

Is there a ticket for this one?

catenacyber:
> Is there a ticket for this one?

I do not think so cf https://github.com/OISF/suricata/pull/6890

PR #7635, opened 2022-07-21
Document rules internals - v1

jufajardini:
While advancing on the task of better understanding and describing how Suricata inspects packets and trying to explain to Jason what I did understand, we thought that adding this documentation could be a good intermediary step...

https://redmine.openinfosecfoundation.org/issues/5449

Describe changes:
Added a page that explains how rules are prioritized by Suri, as well
as what main different types of inspection happen and what elements are
involved when ordering rules.

### DIFF
--- a/doc/userguide/rules/rules-internals.rst
+++ b/doc/userguide/rules/rules-internals.rst
@@ -1,0 +1,151 @@
+Rules Processing
+================
+
+.. toctree::
+
+Suricata rules have many elements that influence how they are processed by
+Suricata and matched against network traffic.
+
+This section explains some key aspects of how Suricata handles rules internally,
+so it can be easier to understand/predict how different rules may interact in
+specific scenarios.
+
+**This material is intended for:** rule writers; developers.
+
+Possible questions one should be better equipped to answer after reading this
+document:
+
+- What happens if two rules have the same priority value (the keyword)?
+- What type of rules will be evaluated first, given a set of rules?
+- How does Suricata decide what rule is "more important" when matching traffic?
+
+.. note::
+
+    Throughout this documentation and for Suricata, the terms "rule" and
+    "signature" are mainly interchangeable, unless context indicates otherwise.
+
+Overview
+--------
+
+Rules are provided to Suricata via rules files. Starting from those, the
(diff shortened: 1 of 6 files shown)


doc/userguide/rules/rules-internals.rst
```diff
@@ -0,0 +1,151 @@
...
+--------
+
+Rules are provided to Suricata via rules files. Starting from those, the
+Detection Engine loader will:
+
+#. load all the signatures;
+#. check for validity (non-existing keywords; duplicated sid etc);
+#. report stats on loaded, good and bad signatures;
+#. sort all valid signatures and store them in a list in the Detection Engine
+   Contect (``DetectEngineCtx``) structure, taking into consideration several
```
jmtaylor90, line 37: Is this Detect Engine Context?

jmtaylor90:
This is great, thank you!

doc/userguide/rules/rules-internals.rst
```diff
@@ -0,0 +1,151 @@
...
+Inspection Flow
+---------------
+
+Once it is time to inspect network traffic against the loaded rules, the
+Detection Engine will match against - if applicable:
+
+#. IPOnly rules (those where packet payload isn't inspected);
+#. Packet/payload related rules;
+#. Frame keywords;
+#. Application layer protocol transactions.
```
victorjulien, line 136: This deserves a bit more explanation, as it has its own logic. Each parser has its own state machine, where it uses a per direction parsing state "progress". For each progress value keywords can be registered. E.g. http has a value "request line available" for which we registered keywords like http.uri, http.method, etc. If in the traffic we reach this state, the sigs with those keywords may be already evaluated, even if they have a lower prio than say a body inspecting sig.

victorjulien (changes requested):
Nice work. Some inline comments.

doc/userguide/rules/rules-internals.rst
jufajardini: Yes, that. Thanks!

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/7655

PR #5458, opened 2020-09-30
Http2 pcapfixes v5

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3955 
https://redmine.openinfosecfoundation.org/issues/3956
https://redmine.openinfosecfoundation.org/issues/3972
https://redmine.openinfosecfoundation.org/issues/3989

Describe changes:
- Runs probing parser for protocol found by pattern on other direction
- Rerun probing parser for protocol found by pattern on other direction if probing parser failed in the first try
- HTTP2 : Uses variable integers to read headers length 
- HTTP2 : better handle transaction state and adds exceptions to stream id reuse event as authorized per RFC
- HTTP2 : keep track of dynamic headers table size

Modifies #5455 by adding the commit fixing the tracking of dynamic headers table size https://redmine.openinfosecfoundation.org/issues/3989

suricata-verify-pr: 332

PR #7639, opened 2022-07-25
detect/alert: ensure reject action is applied to packet/flow - v1

jufajardini:
Bug 5458 states that the reject action is no longer working. While SV
tests that use the reject action still pass, it indeed seems that a
regression has happened with commit aa93984b7e58d3d8c, because while the
function that applies rule actions to the flow (RuleActionToFlow) does
check for the reject action, the newly added function PacketApply
SignatureActions only checks for ACTION_DROP or ACTION_PASS when
deciding to call RuleActionToFlow.

Bug #5458

Describe changes:
- also check for `ACTION_REJECT_ANY` when checking for `ACTION_DROP` when applying signature actions to the packet.

suricata-verify-pr: 888
https://github.com/OISF/suricata-verify/pull/888

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -179,7 +179,7 @@ static void PacketApplySignatureActions(Packet *p, const Signature *s, const uin
     SCLogDebug("packet %" PRIu64 " sid %u action %02x alert_flags %02x", p->pcap_cnt, s->id,
             s->action, alert_flags);
 
-    if (s->action & ACTION_DROP) {
+    if (s->action & (ACTION_DROP | ACTION_REJECT_ANY)) {
         PacketDrop(p, PKT_DROP_REASON_RULES);
 
         if (p->alerts.drop.action == 0) {


suricata-qa:

Pipeline 8312

src/detect-engine-alert.c
```diff
@@ -179,7 +179,7 @@ static void PacketApplySignatureActions(Packet *p, const Signature *s, const uin
     SCLogDebug("packet %" PRIu64 " sid %u action %02x alert_flags %02x", p->pcap_cnt, s->id,
             s->action, alert_flags);
 
-    if (s->action & ACTION_DROP) {
+    if (s->action & (ACTION_DROP | ACTION_REJECT_ANY)) {
```
victorjulien, line 182: will this not just treat rejects as drop?

tugtugtug: My local fix is to add `PACKET_UPDATE_ACTION(p, s->action);` below the `PacketDrop`

jufajardini: > will this not just treat rejects as drop?

Oh, sorry, I misunderstand how the `reject` action works >__<'

jufajardini: > My local fix is to add `PACKET_UPDATE_ACTION(p, s->action);` below the `PacketDrop`

thanks! :)

jufajardini: > My local fix is to add `PACKET_UPDATE_ACTION(p, s->action);` below the `PacketDrop`

Do you mean like this?
```diff
- if (s->action & (ACTION_DROP) {
-        PacketDrop(p, PKT_DROP_REASON_RULES);
-
+ if (s->action & (ACTION_DROP | ACTION_REJECT_ANY)) {
+        PacketDrop(p, PKT_DROP_REASON_RULES);
+        PACKET_UPDATE_ACTION(p, s->action);
```
If that is the case, I wonder if we could call `PACKET_UPDATE_ACTION` before the checks, since it would be called in both arms...

tugtugtug: > > My local fix is to add `PACKET_UPDATE_ACTION(p, s->action);` below the `PacketDrop`
> 
> Do you mean like this?
> 
> ```diff
> - if (s->action & (ACTION_DROP) {
> -        PacketDrop(p, PKT_DROP_REASON_RULES);
> -
> + if (s->action & (ACTION_DROP | ACTION_REJECT_ANY)) {
> +        PacketDrop(p, PKT_DROP_REASON_RULES);
> +        PACKET_UPDATE_ACTION(p, s->action);
> ```
> 
> If that is the case, I wonder if we could call `PACKET_UPDATE_ACTION` before the checks, since it would be called in both arms...

No, if you take a look at the PacketDrop implementation, it calls PACKET_SET_ACTION, which resets the action to DROP.
Ideally, I guess PacketDrop should really call PACKET_UPDATE_ACTION, and then you can move the PACKET_UPDATE_ACTION back to the top of the function.

victorjulien:
I think I would like the solution to be something along the lines of adding a `action` or `aux_action` argument to `PacketDrop`, and then call things like `PacketDrop(p, s->action, ...);` and inside `PacketDrop` something like `PACKET_UPDATE_ACTION(p, aux_action);`.

tugtugtug:
> I think I would like the solution to be something along the lines of adding a `action` or `aux_action` argument to `PacketDrop`, and then call things like `PacketDrop(p, s->action, ...);` and inside `PacketDrop` something like `PACKET_UPDATE_ACTION(p, aux_action);`.

in addition, should a unittest be added to cover this case?

victorjulien:
Wonder if we could also create a offline debug facility for this. Perhaps libnet can also craft a pcap or some other file based output? Need to look into this. Would be great if SV could catch this type of regression.

jufajardini:
> Wonder if we could also create a offline debug facility for this. Perhaps libnet can also craft a pcap or some other file based output? Need to look into this. Would be great if SV could catch this type of regression.

I saw that we have a stats counter `ips.rejected` but was unable to see that being logged, yet... Wanted to use it to help craft SV tests for this one.

victorjulien:
> > Wonder if we could also create a offline debug facility for this. Perhaps libnet can also craft a pcap or some other file based output? Need to look into this. Would be great if SV could catch this type of regression.
> 
> I saw that we have a stats counter `ips.rejected` but was unable to see that being logged, yet... Wanted to use it to help craft SV tests for this one.

That is only used by a select group of IPS modes (there is a ticket to expand it). Perhaps it makes sense to add dedicated "reject" counters. But in general all offline testing is tricky, as the libnet based reject implementation cannot work in offline mode. Perhaps we need another `--simulate-...` option to trigger some pseudo reject behavior that we can test from SV.

victorjulien:
(side note: I added a task for our qalab to add live test as well. This will take some time)

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/7657

PR #7642, opened 2022-07-26
Dhcp leasetime 5435 v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5435

Describe changes:
- Adds `dhcp.leasetime` integer keyword

suricata-verify-pr: 873
https://github.com/OISF/suricata-verify/pull/873

Follow up on https://github.com/OISF/suricata/pull/7566 to have more keywords, looking at integers

Replaces #7611 with style improved, and added warning with keyword when signature parsing fails

Should all DHCP keywords be put in one PR ?

PR #7611, opened 2022-07-04
dhcp: adds leasetime keyword

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5435

Describe changes:
- Adds `dhcp.leasetime` integer keyword

suricata-verify-pr: 873
https://github.com/OISF/suricata-verify/pull/873

Follow up on https://github.com/OISF/suricata/pull/7566 to have more keywords, looking at integers

Replaces #7609 to run again with updated S-V

PR #7645, opened 2022-07-26
Detect keyword warning v1

catenacyber:
None

Describe changes:
- Logs a warning when parsing a signature keyword fails
- removes unused warning and error codes

suricata-verify-pr: 891
https://github.com/OISF/suricata-verify/pull/891

Comes from #7642


suricata-qa:

Pipeline 8322

src/detect-parse.c
```diff
@@ -818,17 +818,17 @@ static int SigParseOptions(DetectEngineCtx *de_ctx, Signature *s, char *optstr,
...
         if (setup_ret == -2) {
+            SCLogDebug("\"%s\" failed to setup", st->name);
             enum DetectKeywordId idx = SigTableGetIndex(st);
             if (de_ctx->sm_types_silent_error[idx] == false) {
                 de_ctx->sm_types_silent_error[idx] = true;
                 return -1;
             }
             return -2;
         }
+        SCLogWarning(SC_ERR_INVALID_RULE_ARGUMENT, "\"%s\" failed to setup", st->name);
```
victorjulien, line 831: this is not very helpful wrt user feedback. It doesn't give enough context about the issue.

Also, this should be an error, not a warning, as the rule will fail to parse

catenacyber: Ok for error vs warning.

> this is not very helpful wrt user feedback. It doesn't give enough context about the issue.

I am not sure what you mean.
This brings more context about the issue than what is currently being done, as suggested in https://github.com/OISF/suricata/pull/7611#discussion_r929672047

When a signature fails to parse, we know already which one, we may have a log error from the keyword.
With this PR, we will know every time which keyword caused the failure.

Am I missing something ?

victorjulien: It just adds another line of output, which has some useful info (keyword name), but makes errors even more verbose and noisy. I would like a solution where we don't add more error/warning lines for a single issue. I kind of dislike the current minimum of 2 lines (keyword issues specific error, then parser issues a generic one which includes the sig). Ideally we go from 2 to 1, not from 2 to 3 :)

catenacyber: Ok, got it

The solution will not be a one-liner then :-/

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7651

PR #7657, opened 2022-07-28
detect/alert: ensure reject action is applied to packet/flow - v2

jufajardini:
Previous PR: #7639 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5458

Describe changes:
- remove unused macro `PACKET_DROP`
- update `PacketDrop()` to receive action as a parameter, so it can also update the packet with the action that triggered the drop it will also call `PACKET_UPDATE_ACTION` instead of `PACKET_SET_ACTION` now

Missing: unittests or sv tests, still not sure how to go about them, and as per Victor's comments on the last PR, sv tests may require some more investigation.

Thoughts...

Many places call `PacketDrop` in contexts where we don't have a rule nor is it clear what action resulted in the drop. I passed `ACTION_DROP` in all those cases. Wondering if we should have something more specific/descriptive?

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -180,7 +180,7 @@ static void PacketApplySignatureActions(Packet *p, const Signature *s, const uin
             s->action, alert_flags);
 
     if (s->action & ACTION_DROP) {
-        PacketDrop(p, PKT_DROP_REASON_RULES);
+        PacketDrop(p, s->action, PKT_DROP_REASON_RULES);
 
         if (p->alerts.drop.action == 0) {
             p->alerts.drop.num = s->num;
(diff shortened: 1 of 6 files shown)


src/detect-engine-alert.c
```diff
@@ -179,7 +179,8 @@ static void PacketApplySignatureActions(Packet *p, const Signature *s, const uin
     SCLogDebug("packet %" PRIu64 " sid %u action %02x alert_flags %02x", p->pcap_cnt, s->id,
             s->action, alert_flags);
 
-    if (s->action & ACTION_DROP) {
+    if (s->action & (ACTION_DROP | ACTION_REJECT_ANY)) {
+        /* PacketDrop will update the packet action, too */
         PacketDrop(p, s->action, PKT_DROP_REASON_RULES);
```
jufajardini, line 184: From what  I understood, once we have `PacketDrop` updating the Packet with the proper action, this approach would work, so I kept this.

suricata-qa:
ERROR:
 

Pipeline 8384

victorjulien:
I think the PR should target master before master-6.0.x, unless there is a good reason for just looking at 6

jufajardini:
> I think the PR should target master before master-6.0.x, unless there is a good reason for just looking at 6

Ok! No, there isn't :D:

PR #7661, opened 2022-07-29
detect/alert: ensure reject action is applied to packet/flow - v1

jufajardini:

Describe changes:
- update PacketDrop() to receive action as a parameter, so it can also update the packet with the action that triggered the drop it will also call PACKET_UPDATE_ACTION instead of PACKET_SET_ACTION now

Missing: unittests or sv tests, still not sure how to go about them, and as per Victor's comments on the last PR, sv tests may require some more investigation.

Thoughts...

Many places call PacketDrop in contexts where we don't have a rule nor is it clear what action resulted in the drop. I passed ACTION_DROP in all those cases. Wondering if we should have something more specific/descriptive?

### DIFF
--- a/src/decode.h
+++ b/src/decode.h
@@ -922,12 +922,22 @@ static inline void PacketSetAction(Packet *p, const uint8_t a)
 
 #define PACKET_TEST_ACTION(p, a) (p)->action &(a)
 
-static inline void PacketDrop(Packet *p, enum PacketDropReason r)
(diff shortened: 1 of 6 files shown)


suricata-qa:
ERROR:
 

Pipeline 8405

jufajardini:
Replaced by https://github.com/OISF/suricata/pull/7682

PR #7665, opened 2022-07-31
detect: checks config keyword in cleanup

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5456

Describe changes:
- detect: checks config keyword in cleanup

### DIFF
--- a/src/app-layer-parser.c
+++ b/src/app-layer-parser.c
@@ -1023,7 +1023,8 @@ void AppLayerParserTransactionsCleanup(Flow *f)
 
         if (txd && logger_expectation != 0) {
             LoggerId tx_logged = GetTxLogged(txd);
-            if (tx_logged != logger_expectation) {
+            if (tx_logged != logger_expectation &&
+                    !(txd->config.log_flags & BIT_U8(CONFIG_TYPE_TX))) {
                 SCLogDebug("%p/%"PRIu64" skipping: logging not done: want:%"PRIx32", have:%"PRIx32,
                         tx, i, logger_expectation, tx_logged);
                 skipped = true;


suricata-qa:

Pipeline 8427

src/app-layer-parser.c
```diff
@@ -1023,7 +1023,8 @@ void AppLayerParserTransactionsCleanup(Flow *f)
 
         if (txd && logger_expectation != 0) {
             LoggerId tx_logged = GetTxLogged(txd);
-            if (tx_logged != logger_expectation) {
+            if (tx_logged != logger_expectation &&
```
victorjulien, line 1026: could we move the cost of this to the "config" keyword execution? Mark a tx as logged there?

catenacyber: We can. I do not know if we should.

This was not the logic which was used previously cf `OutputTxLog`.
There seems to be something more complex with DNS unidirectional transactions
Like if a DNS request is marked as not log-worthy, the reply will not be logged either...

I can do a big simplification/removal of code ;-)
But then, this code may then come back if we want to have more features in this config keyword...

So, what do you want ?

catenacyber: friendly ping @victorjulien ?

suricata-qa:

Pipeline 8427

src/app-layer-parser.c
catenacyber: friendly ping @victorjulien ?

catenacyber: @jasonish can I do a big removal of `AppLayerTxConfig` ?

jasonish: > @jasonish can I do a big removal of `AppLayerTxConfig` ?

Not used anymore?  I'm not up to date on the context here.

catenacyber: > No, can't remove.

Why ? 🤔 
This is a structure with only one field `log_flags: u8` and this bitflag is only used for one bit `CONFIG_TYPE_TX` 

> Mark a tx as logged there?

To do that, we should use `logged.flags` instead of `config.log_flags`
This way, the whole `AppLayerTxConfig` gets unused...

victorjulien: I need to look more into the issue, but it was added to be extended with more features so I'm not inclined to remove it.

catenacyber:
New solution proposed in https://github.com/OISF/suricata/pull/8418

PR #7666, opened 2022-07-31
Http2 noalloc header compare 5454 v1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5454

Describe changes:
- http2: remove `to_vec` in comparisons (no need to alloc)

@jasonish does clippy catch this ?

PR #7671, opened 2022-08-01
Http2 noalloc header compare 5454 v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5454

Describe changes:
- http2: remove `to_vec` in comparisons (no need to alloc)

@jasonish does clippy catch this ?

Replaces #7666 with version compatible with rust MSRV 1.48 and adds a commit for CI to test MSRV

### DIFF
--- a/rust/src/http2/decompression.rs
+++ b/rust/src/http2/decompression.rs
@@ -168,13 +168,13 @@ impl HTTP2DecoderHalf {
     pub fn http2_encoding_fromvec(&mut self, input: &Vec<u8>) {
         //use first encoding...
         if self.encoding == HTTP2ContentEncoding::HTTP2ContentEncodingUnknown {
-            if *input == "gzip".as_bytes().to_vec() {
+            if &input[..] == "gzip".as_bytes() {
                 self.encoding = HTTP2ContentEncoding::HTTP2ContentEncodingGzip;
                 self.decoder = HTTP2Decompresser::GZIP(GzDecoder::new(HTTP2cursor::new()));
-            } else if *input == "deflate".as_bytes().to_vec() {
+            } else if &input[..] == "deflate".as_bytes() {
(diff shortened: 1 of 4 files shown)


rust/src/http2/decompression.rs
```diff
@@ -168,13 +168,13 @@ impl HTTP2DecoderHalf {
     pub fn http2_encoding_fromvec(&mut self, input: &Vec<u8>) {
```
jasonish, line 168: Clippy will usually complain about `&Vec` in signatures, especially if a `&[u8]` could be used instead.

suricata-qa:

Pipeline 8437

rust/src/http2/decompression.rs
catenacyber: I meant : does clippy complain about running `to_vec` to do a comparison ?

jasonish: > I meant : does clippy complain about running `to_vec` to do a comparison ?

Not that I can tell.

catenacyber: Should it not ?
Should we ask someone ?

jasonish: I'm not seeing a lint that would be called out for this, or one of our supressiosn is also suppressing this.  I suppose we could hunt that down and contribute a lint?  But I think looking at the function signature first, and not using `&Vec` would have resulted in a better choice for the comparison to begin with.

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7685

rust/src/http2/decompression.rs
catenacyber: Here is a minimal reproducer

```
fn main() {
    let a = vec![0x33, 0x34, 0x35];
// bad
    if a == "345".as_bytes().to_vec() {
        println!("Hello, world!");
    }
// good
    if &a == "345".as_bytes() {
        println!("Hello, world!");
    }
}
```

PR #7674, opened 2022-08-02
Bug 4786/v4

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5207

**Note:** This branch has a known failure and is just for an intermediary review. The work is yet to be completed and tested.
Known issue is commented where it happens in the code.

Major changes that come along with this branch:
All *bits shall follow the same syntax which was one of the complaints and motivation behind this work.
The syntax shall be
`xbits:set,abc,expire 1000, track ip_dst;`
`hostbits:set,abc,expire 1000, track ip_dst;`
`flowbits:set,abc;`

This makes a huge difference in the currently implemented `hostbits` syntax and will break any existing rules that use it. No rules were found in ET rulesets that used `hostbits`.
The syntax currently in `master` for `hostbits` is:
```
hostbits:set,abc,src
```
which is an equivalent in terms of functionality to the syntax proposed in this PR:
```
hostbits:set,abc,expire 300, track ip_src;
```


rust/src/rule_parser/xbits.rs
```diff
@@ -0,0 +1,323 @@
...
+                Err(_) => return Err(()),
+            };
+            return Ok(DetectBitsData {
+                idx: 0,
+                cmd: cmd_name.0,
+                name: (cmd_name.1).into_raw(),
+                tracker: 0,
+                expire: 0,
+                vartype: VAR_TYPE_FLOW_BIT,
+                fb_names: (cmd_name.2).as_mut_ptr(),
```
inashivb, line 205: Here, `cmd_name.2` is a vector of raw C strings and holds all the pointers to strings that it is supposed to.
Although, after this stage, on the C side, a pointer is received but it seems to not point to anything at all.
It was expected to be a `*mut *mut c_char`.
I'm not sure how should I try to traverse or reown the strings to read inside C so that may be very wrong but right now, `fb_names` seems to be dropped much earlier than when it should. I tried to `forget` about the vector pointer but could not make it work properly.

src/detect-flowbits.c
```diff
@@ -213,118 +167,68 @@ int DetectFlowbitMatch (DetectEngineThreadCtx *det_ctx, Packet *p,
...
+                return -1;
             }
-        }
+            break;
     }
 
-    return 1;
+    if (cd->or_list_size != 0) {
+        for (int i = 0; i < cd->or_list_size; i++) {
+            int leq = VarNameStoreSetupAdd(cd->fb_names[i], VAR_TYPE_FLOW_BIT);
```
inashivb, line 206: This is where the heap use after free error arises.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_asan | | |
 

Pipeline 8456

rust/src/rule_parser/xbits.rs
```diff
@@ -0,0 +1,323 @@
...
+#[repr(C)]
+#[derive(Debug)]
+pub struct DetectBitsData {
+    idx: u32,
+    cmd: u8,
+    tracker: u8,
+    expire: u32,
+    vartype: u8,
+    name: *mut c_char,
+    fb_names: *mut *mut c_char,
```
victorjulien, line 41: some of the complexity seems to come from a desire to make this data structures accessible to C, meaning we need C types etc. Have you considered making it opague to C and just access it through some utility accessor functions? Then we can have simpler logic in Rust using native rust types.

src/detect-flowbits.c
```diff
@@ -334,7 +238,7 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
     sm->type = DETECT_FLOWBITS;
     sm->ctx = (SigMatchCtx *)cd;
```
inashivb, line 239: Q1: If I were to make the struct opaque to C, how would I set sigmatch context here?
Q2: SigMatchCtx seems to have just one member `int foo`,  does it matter what i assign to it?
Q3: What part of the code should I look at to understand the importance of context?

victorjulien: I think it might be helpful to check how `DetectHTTP2sizeUpdateSetup` and the functions below it handle this. Essentially from C there is only a void pointer.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_asan | | |
 

Pipeline 8456

suricata-qa:
ERROR:
 

Pipeline 10596

PR #7668, opened 2022-07-31
Ike v1 logging tx transforms 5455 v2

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5455

Describe changes:
- ikev1 : log fields from transaction instead of fields from state

ping @satta @fhonza as you have worked on ikev1

Replaces #7667 with S-V passing by keeping in the state the one transform chosen by the server

PR #7675, opened 2022-08-02
Ike v1 logging tx transforms 5455 v3

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5455

Describe changes:
- ikev1 : log fields from transaction instead of fields from state

Replaces #7668 with using u64 instead of usize

### DIFF
--- a/rust/src/ike/ikev1.rs
+++ b/rust/src/ike/ikev1.rs
@@ -40,27 +40,27 @@ pub struct IkeV1Header {
 pub struct Ikev1ParticipantData {
     pub key_exchange: String,
     pub nonce: String,
-    pub vendor_ids: HashSet<String>,
-    /// nested Vec, outer Vec per Proposal/Transform, inner Vec has the list of attributes.
-    pub transforms: Vec<Vec<SaAttribute>>,
+    pub nb_transforms: u64,
+    pub transform: Vec<SaAttribute>,
 }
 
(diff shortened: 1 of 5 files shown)


rust/src/ike/ikev1.rs
```diff
@@ -40,27 +40,27 @@ pub struct IkeV1Header {
 pub struct Ikev1ParticipantData {
     pub key_exchange: String,
     pub nonce: String,
-    pub vendor_ids: HashSet<String>,
-    /// nested Vec, outer Vec per Proposal/Transform, inner Vec has the list of attributes.
-    pub transforms: Vec<Vec<SaAttribute>>,
+    pub nb_transforms: u64,
```
victorjulien, line 43: so what happens if we actually get some really high number in here? the u64 suggests we'd handle it but given the loops of the actual transforms vec I'm not so sure

suricata-qa:

Pipeline 8457

rust/src/ike/ikev1.rs
catenacyber: To reach it, someone needs to send at least U64_MAX bytes of data...

The worst that would happen then, is that we would stop logging the transform chosen by the server (which is controlled by the attacker in this scenario)

victorjulien:
Merged in #7713, thanks!

PR #7682, opened 2022-08-02
detect/alert: ensure reject action is applied to packet/flow - v2

jufajardini:
Previous PR: #7661 

Changes from last PR:
- Added two unittests to check for correct packet-action flags after signature matching

Thoughts:
- should more unittests be added?
- Many places call PacketDrop in contexts where we don't have a rule nor is it clear what action resulted in the drop. I passed ACTION_DROP in all those cases. Wondering if we should have something more specific/descriptive?

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -23,6 +23,8 @@
 #include "detect-engine-tag.h"
 
 #include "decode.h"
+#include "decode-icmpv4.h"
+#include "decode-ipv4.h"
 
 #include "flow.h"
 #include "flow-private.h"
@@ -31,7 +33,10 @@
 #include "util-exception-policy.h"
 #endif
 
+#include "util-mpm-ac.h"
 #include "util-profiling.h"
+#include "util-unittest.h"
+#include "util-unittest-helper.h"
 
 /** tag signature we use for tag alerts */
 static Signature g_tag_signature;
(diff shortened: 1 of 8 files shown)


suricata-qa:

Pipeline 8465

src/detect-engine-alert.c
```diff
@@ -183,15 +188,17 @@ static void PacketApplySignatureActions(Packet *p, const Signature *s, const uin
     SCLogDebug("packet %" PRIu64 " sid %u action %02x alert_flags %02x", p->pcap_cnt, s->id,
             s->action, alert_flags);
 
-    if (s->action & ACTION_DROP) {
-        PacketDrop(p, PKT_DROP_REASON_RULES);
+    if (s->action & (ACTION_DROP | ACTION_REJECT_ANY)) {
```
victorjulien, line 191: looking at how `s->action` is set up, this is a bit redundant. For reject we always also set `ACTION_DROP`. There is no harm in showing the reject here, but I would like 2 additions:
1. a comment stating this
2. a `DEBUG_VALIDATE_BUG_ON` that makes sure `ACTION_DROP` is always set in this branch

src/detect-engine-alert.c
```diff
@@ -415,4 +422,63 @@ void PacketAlertFinalize(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx
 
 }
 
+#ifdef UNITTESTS
```
victorjulien, line 425: can these be moved into (a new) `src/tests/detect-engine-alert.c`

src/detect-engine-threshold.c
```diff
@@ -300,7 +300,7 @@ static inline void RateFilterSetAction(Packet *p, PacketAlert *pa, uint8_t new_a
             pa->flags |= PACKET_ALERT_RATE_FILTER_MODIFIED;
             break;
         case TH_ACTION_DROP:
-            PacketDrop(p, PKT_DROP_REASON_RULES_THRESHOLD);
+            PacketDrop(p, new_action, PKT_DROP_REASON_RULES_THRESHOLD);
```
victorjulien, line 303: new_action seems to have a different set of flags. It has `TH_ACTION_DROP`, but is treated here as if it has `ACTION_DROP`. Even if these would be the same a safer approach would be to just pass `ACTION_DROP` here explicitly.

victorjulien: Btw the reject case below should also act like reject+drop, can you review that it does?

victorjulien (changes requested):
See inline comments

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/7723

PR #7683, opened 2022-08-03
Fix dns in tcp flow not detected

zagge-cgeo:

https://redmine.openinfosecfoundation.org/issues/4759

Describe changes:
This fixes issue 4759. The problem is, that when tcp rules are active
a tcp dns connection first the server to client is inspected
and nothing found.
Afterwards the full flow is marked as processed and therefore the dns
query in TCP is not found.
This patch does not mark the flow as processed in case of a server to client connection


victorjulien:
Many Suricata-Verify tests are failing, so it seems that the patch has side effects.

victorjulien:
replaced by #7693

zagge-cgeo:
As far as I can see it is not a problem of the code but the test. The test that failed has more then ALPROTO_MAXTS dns transactions in the failed test case. So it has to fail. Before the change the transactions where not found and gone unnoticed.

PR #7693, opened 2022-08-04
Fix dns in tcp flow not detected

zagge-cgeo:

https://redmine.openinfosecfoundation.org/issues/4759

Describe changes:
Make suricata-verify pass

This replaces https://github.com/OISF/suricata/pull/7683


victorjulien:
Fuzz finds a way to create lots of transactions that are never closed, this is introduced by the changes here.

catenacyber:
> Make suricata-verify pass

Which test are you talking about ?

zagge-cgeo:
> > Make suricata-verify pass
> 
> Which test are you talking about ?

I'm talking about the unit tests run by suricata-verify that where failing in the first version of the patch 

https://github.com/OISF/suricata/pull/7683/checks

zagge-cgeo:
Forcing dns=11
Too many open transactions for protocol dns
Assertion failure: dns

These is now failing as there is a limit of 4096 transactions in the unit tests. However the testcase create more than that number of transactions in the failing test. The test did not fail before as all transactions where ignored and none of them matched the rule in tcp.

Do you have an idea on how this could be fixed?

catenacyber:
> > > Make suricata-verify pass
> > 
> > Which test are you talking about ?
> 
> I'm talking about the unit tests run by suricata-verify that where failing in the first version of the patch
> 

Ok thanks, usually we describe what the PR does + in another line the changes from last PR

And I guess the elements from the redmine ticket can be put into a new Suricata-Verify test.

I can reproduce the bug.
But I wonder what the right fix is.

> Fix dns in tcp flow not detected

The flow is detected as DNS, but the signature does not trigger...

Why do you say transactions are bidirectional For DNS over TCP ?
I do not see so...

And I have no idea why adding a tls signature influences here...

catenacyber:
Interesting fact: the alert happens 2 packets later than the logging when there is not the tls rule...

zagge-cgeo:
> > > > Make suricata-verify pass
> > > 
> > > 
> > > Which test are you talking about ?
> > 
> > 
> > I'm talking about the unit tests run by suricata-verify that where failing in the first version of the patch
> 
> Ok thanks, usually we describe what the PR does + in another line the changes from last PR
> 
> And I guess the elements from the redmine ticket can be put into a new Suricata-Verify test.
> 
> I can reproduce the bug. But I wonder what the right fix is.
> 
> > Fix dns in tcp flow not detected
> 
> The flow is detected as DNS, but the signature does not trigger...
> 
> Why do you say transactions are bidirectional For DNS over TCP ? I do not see so...
> 
> And I have no idea why adding a tls signature influences here...

It is not the tls signature. Any signature that uses tcp will result in the behaviour. 

The problem happens when there is any rule using tcp, the dns package is only looked at when the content of the request is not yet available and therefore it is not analysed afterwards as it is marked as process already.

catenacyber:
Thanks for this analysis.
So, it looks like the problem is broader, and is the fact that detection does not happen when logging does...

catenacyber:
So, I understand this far :
- packet 4 : dns query
- packet 5 : ack of dns query and logging, no detection as packet is to client + BUG : transaction gets cleaned up by `AppLayerParserTransactionsCleanup` cf `if (is_unidir && tx_skipped && !inspected) {`
- packet 7 : running detection as packet is to server

catenacyber:
Stranger results with
```
alert dns any any -> any any (msg:"DNS standard opcode"; dns.opcode: 0; id:1; rev:1;)
alert dns any any -> any any (msg:".com in DNS query"; dns.query; content:".com"; sid:2; rev:1;)
```

No alerts at all...

victorjulien:
Could be a case of a missing `AppLayerParserTriggerRawStreamReassembly` after the TCP record has been received.

catenacyber:
@jasonish I think your commit 60ebc27c4eb755800e6d3f4ec1a5d55a5230a214 is responsible for this bug.

For unidirectional transactions, we may run detection on the other side than the transaction.

For instance (cf previous rules posted), we may have dns rules in both directions, a packet to client asking a dns query, and thus we will run `DetectRunTx` which will update `new_detect_flags |= APP_LAYER_TX_INSPECTED_FLAG;` for the tocliendt direction.

And then `AppLayerParserTransactionsCleanup` cleans up the transaction as detection has been run on one side, even if it was not the relevant side. But there is no way currently for unidirectional transactions to tell their direction...

jasonish:
> @jasonish I think your commit [60ebc27](https://github.com/OISF/suricata/commit/60ebc27c4eb755800e6d3f4ec1a5d55a5230a214) is responsible for this bug.
> 
> For unidirectional transactions, we may run detection on the other side than the transaction.
> 
> For instance (cf previous rules posted), we may have dns rules in both directions, a packet to client asking a dns query, and thus we will run `DetectRunTx` which will update `new_detect_flags |= APP_LAYER_TX_INSPECTED_FLAG;` for the tocliendt direction.
> 
> And then `AppLayerParserTransactionsCleanup` cleans up the transaction as detection has been run on one side, even if it was not the relevant side. But there is no way currently for unidirectional transactions to tell their direction...

So we should be looking at a fix that is outside of DNS and generic to all unidirectional transactions?

catenacyber:
> So we should be looking at a fix that is outside of DNS and generic to all unidirectional transactions?

I think so indeed

jasonish:
> > So we should be looking at a fix that is outside of DNS and generic to all unidirectional transactions?
> 
> I think so indeed

Have you looked deep enough for a possible fix? I don't want to repeat what work might have already been done.

catenacyber:
Just reverted the culprit commit as premature optimisation.

To keep it optimized, I think you need to implement a trait for unidirectional transaction to tell its direction. I did not write anything like it yet

jasonish:
> Just reverted the culprit commit as premature optimisation.
> 
> To keep it optimized, I think you need to implement a trait for unidirectional transaction to tell its direction. I did not write anything like it yet

Ok. So each unidirectional parser needs an update and not just the outer transaction handling.

catenacyber:
> Ok. So each unidirectional parser needs an update and not just the outer transaction handling.

I see 3 ways :
- revert the optimization (is it really helpful ?)
- change every unidirectional parser so that a transaction has the direction and use it to set `inspected = true;` 
- Think more and find a better way :-p

jasonish:
> * revert the optimization (is it really helpful ?)

I don't think its an optimization, but instead fixes a real issue.  Of course, 2 years ago we weren't reliable at putting ticket numbers in commits, but I think this is it: https://redmine.openinfosecfoundation.org/issues/3877

catenacyber:
Indeed you know better, that is why I called on you Jason ;-)

rust/src/dns/dns.rs
```diff
@@ -773,13 +776,17 @@ pub unsafe extern "C" fn rs_dns_parse_response_tcp(flow: *const core::Flow,
...
                                                  -> std::os::raw::c_int
 {
     // This is a stateless parser, just the existence of a transaction
-    // means its complete.
+    // means its complete. However this is only true for UDP and not TCP
+    let tx = cast_pointer!(_tx, DNSTransaction);
     SCLogDebug!("rs_dns_tx_get_alstate_progress");
+    if tx.is_tcp && _direction & core::STREAM_TOCLIENT != 0 && !tx.response.is_some() {
+        return 0;
+    }
```
jasonish, line 789: I think this needs to be more complete, then fuzzing appears to pass.  See https://github.com/OISF/suricata/pull/7825/files#diff-fe45d440e3be98030915c1946205744d6a8cd791fad6223d1926c67705a91be4R787 for what I mean, feel free to integrate that into your PR.

jasonish (changes requested):
I'd also see if we can drop the tcp flag, given how this parser determines if its a request or response, I don't think it should matter.

rust/src/dns/dns.rs
zagge-cgeo: I integate the change and it works also without the tcp flag.

However some logging output is missing when analyzing udp dns packages. It no longer logs the dns traffic. Output like the following is missing with the updated patch:

{"timestamp":"2018-10-21T01:41:56.742793+0000","flow_id":1939439079413129,"pcap_cnt":1,"event_type":"dns","src_ip":"192.168.1.18","src_port":1027,"dest_ip":"192.168.1.1","dest_port":53,"proto":"UDP","pkt_src":"wire/pcap","dns":{"type":"query","id":17024,"rrname":"time.windows.com","rrtype":"A","tx_id":0}}
{"timestamp":"2018-10-21T01:41:56.743525+0000","flow_id":1939439079413129,"pcap_cnt":2,"event_type":"dns","src_ip":"192.168.1.18","src_port":1027,"dest_ip":"192.168.1.1","dest_port":53,"proto":"UDP","pkt_src":"wire/pcap","dns":{"version":2,"type":"answer","id":17024,"flags":"8180","qr":true,"rd":true,"ra":true,"rrname":"time.windows.com","rrtype":"A","rcode":"NOERROR","answers":[{"rrname":"time.windows.com","rrtype":"CNAME","ttl":3600,"rdata":"time.microsoft.akadns.net"}],"grouped":{"CNAME":["time.microsoft.akadns.net"]}}}

Any idea why that is happening?

catenacyber: I think it is the bug described in https://github.com/OISF/suricata/pull/7693#issuecomment-1226958253 which needs a bigger fix, not only for DNS.

A suricata-verify pull request would be nice to have all these expected behaviors

zagge-cgeo: It seems that I was wrong. The logs are created when including https://github.com/OISF/suricata/pull/7825/files#diff-fe45d440e3be98030915c1946205744d6a8cd791fad6223d1926c67705a91be4R787.
Seems I had not used the correct patch when I tested. I will update my pull request.

When I remove the tcp flag the logs no longer work.

jasonish: > When I remove the tcp flag the logs no longer work.

Ok.  It is ok to leave the tcp check in there.  I think this gets us to having dns behave better with other protocols, but as mentioned, there are some other issues at play here that likely need some looking at outside just dns.

Do you plan to do a new PR with this change?

zagge-cgeo: > Do you plan to do a new PR with this change?

The new pull request can be found here https://github.com/OISF/suricata/pull/7835

PR #7371, opened 2022-05-03
python: remove use of distutils - v2

jasonish:
Don't use distutils to install Python code, instead just use make.

Distutils is being deprecated and setuptools insists on working within the
scope of the system and not within our prefix, not allowing 2 versions of
Suricata-Update to be installed in different prefixes, among various other
issues.

As we have little to gain from using setuptools, and its less work to maintain
a Makefile than fight setuptools, just use make to install our Python code.

Requires a modification to Suricata-Update: https://github.com/OISF/suricata-update/pull/315

Recommended to backport to 6.0.x, but not 5.0.x.

Issue: https://redmine.openinfosecfoundation.org/issues/5313

suricata-update-pr: 315

PR #7698, opened 2022-08-05
python: remove use of distutils - v3

jasonish:
Don't use distutils to install Python code, instead just use make.

Distutils is being deprecated and setuptools insists on working within the
scope of the system and not within our prefix, not allowing 2 versions of
Suricata-Update to be installed in different prefixes, among various other
issues.

As we have little to gain from using setuptools, and its less work to maintain
a Makefile than fight setuptools, just use make to install our Python code.

Requires a modification to Suricata-Update: https://github.com/OISF/suricata-update/pull/315

Recommended to backport to 6.0.x, but not 5.0.x.

Issue: https://redmine.openinfosecfoundation.org/issues/5313

suricata-update-pr: 315

Previous PR: https://github.com/OISF/suricata/pull/7371
Changes from last PR:
- Portable usage of `install` command, tested on FreeBSD.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -117,42 +117,6 @@
        pymv="$($HAVE_PYTHON -c 'import sys; print(sys.version_info[[0]]);')"
     fi
 
-    # Check for python-distutils (setup).
-    have_python_distutils="no"
-    if test "x$enable_python" = "xyes"; then
-        AC_MSG_CHECKING([for python-distutils])
-	if $HAVE_PYTHON -c "import distutils; from distutils.core import setup" 2>/dev/null; then
-	   AC_MSG_RESULT([yes])
-	   have_python_distutils="yes"
-	else
-	   AC_MSG_RESULT([no])
-	fi
-    fi
-    AM_CONDITIONAL([HAVE_PYTHON_DISTUTILS],
-	[test "x$have_python_distutils" = "xyes"])
-    if test "$have_python_distutils" = "no"; then
-       echo ""
-       echo "    Warning: Python distutils not found. Python tools will"
-       echo "        not be installed."
-       echo ""
-       echo "    Install the distutils module for Python ${pymv} to enable"
-       echo "    the Python tools."
-       echo ""
-    fi
-
-    # Check for python-yaml.
-    have_python_yaml="no"
-    if test "x$enable_python" = "xyes"; then
-        AC_MSG_CHECKING([for python-yaml])
-	if $HAVE_PYTHON -c "import yaml" 2>/dev/null; then
-	   have_python_yaml="yes"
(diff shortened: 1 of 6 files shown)


victorjulien:
Several of my test runners fail in testing the `suricata-update -V` command where they previously succeeded. It fails to find `python`. I suppose I can force `python3 suricata-update` but this feels like a regression.

Examples of OS': Rocky 8, Alma 8, Ubuntu 20.04, 22.04, Debian 11, Fedora 34/35, or just about all we care about.

CentOS 7 fails to find `pyyaml`.

jasonish:
> Several of my test runners fail in testing the `suricata-update -V` command where they previously succeeded. It fails to find `python`. I suppose I can force `python3 suricata-update` but this feels like a regression.
> 
> Examples of OS': Rocky 8, Alma 8, Ubuntu 20.04, 22.04, Debian 11, Fedora 34/35, or just about all we care about.
> 

Ah, I think that is one thing using setuptools did for us, I'll have to look into that.  These all Docker as well?  I find often `python` works on a real install, but not a Docker install.  Probably something involved with the dependencies that get installed on a proper install vs a container image.

> CentOS 7 fails to find `pyyaml`.

Is pyyaml installed?  I guess there is a different issue here.  We'll continue to install suricata-update even if pyyaml is installed: https://github.com/OISF/suricata/pull/7371#discussion_r864639626

victorjulien:
> > Several of my test runners fail in testing the `suricata-update -V` command where they previously succeeded. It fails to find `python`. I suppose I can force `python3 suricata-update` but this feels like a regression.
> > Examples of OS': Rocky 8, Alma 8, Ubuntu 20.04, 22.04, Debian 11, Fedora 34/35, or just about all we care about.
> 
> Ah, I think that is one thing using setuptools did for us, I'll have to look into that. These all Docker as well? I find often `python` works on a real install, but not a Docker install. Probably something involved with the dependencies that get installed on a proper install vs a container image.
> 
> > CentOS 7 fails to find `pyyaml`.
> 
> Is pyyaml installed? I guess there is a different issue here. We'll continue to install suricata-update even if pyyaml is installed: [#7371 (comment)](https://github.com/OISF/suricata/pull/7371#discussion_r864639626)

`yum -y install gcc libpcap-devel pcre-devel pcre2-devel libyaml-devel file-devel zlib-devel jansson-devel nss-devel libcap-ng-devel libnet-devel tar make libnetfilter_queue-devel lua-devel jq git python3 python36-PyYAML cargo rustc xz-devel`

Yes, see the install line. This works in all branches except the PR.

jasonish:
Hopefully these issues are addressed here: https://github.com/OISF/suricata/pull/7704

PR #7709, opened 2022-08-09
rust: pin serde to 1.0.142 - v1

jasonish:
Pin serde to 1.0.142 for now as 1.0.143 fails to build on Rust 1.48.

We don't depend on this directly, but instead via sawp-modbus.


victorjulien:
Upstream serde fixed things it seems, and we bumped MSRV. So closing this.

PR #6483, opened 2021-10-18
bittorrent-dht: add bittorrent-dht app layer v4

abungay:
Parses and logs the bittorrent-dht protocol. 

Feature: #3086

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3086

Describe changes:
- Ensures BitTorrent DHT traffic does not get marked as app_proto: failed
- Parses BitTorrent DHT packets and logs data to eve.json

Changes since last PR:
- rebase to master, fix merge conflicts

Reference BitTorrent DHT specification:
https://www.bittorrent.org/beps/bep_0005.html

TODO after merging PR:
- Update wiki for SID ranges at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/AppLayer for the newly introduced bittorrent-dht SID range (start=2271000, end=2271999)

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

Previous PR: https://github.com/OISF/suricata/pull/5942

suricata-verify-pr: 560

PR #7715, opened 2022-08-10
bittorrent-dht: add bittorrent-dht app layer - v6

jasonish:
Parses and logs the bittorrent-dht protocol. 

Feature: https://redmine.openinfosecfoundation.org/issues/3086

Describe changes:
- Ensures BitTorrent DHT traffic does not get marked as app_proto: failed
- Parses BitTorrent DHT packets and logs data to eve.json

Changes since last PR:
- rebase to master, fix merge conflicts
- set as unidirectional
- fixes for updated generic event handling
- convert to registering a pattern for protocol detection
- rename event type to bittorrent_dht (to avoid underscores in key names)

Reference BitTorrent DHT specification:
https://www.bittorrent.org/beps/bep_0005.html

TODO after merging PR:
- Update wiki for SID ranges at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/AppLayer for the newly introduced bittorrent-dht SID range (start=2271000, end=2271999)

Previous PR: https://github.com/OISF/suricata/pull/6483

suricata-verify-pr: 904


jasonish:
Out of draft, ready for another round of review.

jufajardini (dismissed):
doc changes look good to me :)

jasonish (commented):
There is a discrepancy in the logging.  For alert and metadata we use `bittrorrent_dht` as part of our effort to not have hyphens as key names to be more friendly to SIEMs as hyphens don't make valid object names in most programming languages.

However, in the stats this protocol still shows up as `bittorrrent-dht` but `bittorrent_dht` is used for the event_type and protocol object name.  `ftp-data` has the same issue as well.

I'll attempt to address this in a new version of this PR.

jufajardini:
> There is a discrepancy in the logging. For alert and metadata we use `bittrorrent_dht` as part of our effort to not have hyphens as key names to be more friendly to SIEMs as hyphens don't make valid object names in most programming languages.
> 
> However, in the stats this protocol still shows up as `bittorrrent-dht` but `bittorrent_dht` is used for the event_type and protocol object name. `ftp-data` has the same issue as well.
> 
> I'll attempt to address this in a new version of this PR.

This sounds a bit like - https://redmine.openinfosecfoundation.org/issues/5494 
Would it make sense to have an umbrella ticket to check for other similar cases?

jasonish:
> > There is a discrepancy in the logging. For alert and metadata we use `bittrorrent_dht` as part of our effort to not have hyphens as key names to be more friendly to SIEMs as hyphens don't make valid object names in most programming languages.
> > However, in the stats this protocol still shows up as `bittorrrent-dht` but `bittorrent_dht` is used for the event_type and protocol object name. `ftp-data` has the same issue as well.
> > I'll attempt to address this in a new version of this PR.
> 
> This sounds a bit like - https://redmine.openinfosecfoundation.org/issues/5494 Would it make sense to have an umbrella ticket to check for other similar cases?

I think these are different enough.  The tls stuff looks like more like ordinary typos.  The problem with protocol names is that in the configuration and rule syntax I think we want `bittorrent-dht` as hypens are more user friendly.  And we have some code to parse names as such and the inverse to print them out.  However, underscores are much more computer friendly, so in the json we want to use them instead of hyphens.

victorjulien:
A few hours later
```
      1 {"transaction_id":"fn\u0000\u0000","error":{"num":202,"msg":"Server Error"}}
      5 {"transaction_id":"pn\u0000\u0000","error":{"num":202,"msg":"Server Error"}}
  49042 {"transaction_id":""}
```

jasonish:
> After running this for a while on my network, while downloading a iso and seeding another, I see nothing but
> 
> `{"transaction_id":""}`. Am I missing something?

Need more test pcaps for sure. The one we have appears to be crafted as well. 

I wonder if the rather small pattern match is picking up other traffic.

jasonish:
These `empty` fields are fixed in https://github.com/OISF/suricata/pull/7731, however, there is still more to do as discussed in that PR.

PR #7701, opened 2022-08-08
Tls random/v2

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5190

suricata-verify-pr: 892

Previous PR: https://github.com/OISF/suricata/pull/7647

PR #7719, opened 2022-08-12
Tls random/v3

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5190

suricata-verify-pr: 906

Previous PR: https://github.com/OISF/suricata/pull/7701


src/detect-tls-random.c
```diff
@@ -0,0 +1,296 @@
...
+
+        if (flow_flags & STREAM_TOSERVER) {
+            random_buf = ssl_state->server_connp.random[0];
+        } else {
+            random_buf = ssl_state->client_connp.random[0];
+        }
+
+        /* First four bytes of random represent current UTC in the
+         * Unix epoch format so for now, first byte should never be 0 */
+        if (random_buf == 0) {
```
inashivb, line 280: This maybe incorrect. Check again.

src/detect-tls-random.c
```diff
@@ -0,0 +1,296 @@
...
+        uint8_t random_buf;
+        const SSLState *ssl_state = (SSLState *)f->alstate;
+
+        if (flow_flags & STREAM_TOSERVER) {
+            random_buf = ssl_state->server_connp.random[0];
+        } else {
+            random_buf = ssl_state->client_connp.random[0];
+        }
+
+        /* First four bytes of random represent current UTC in the
```
inashivb, line 278: not the first four

src/detect-tls-random.c
```diff
@@ -0,0 +1,296 @@
...
+         * Unix epoch format so for now, first byte should never be 0 */
```
victorjulien, line 207: but what if they are? We can't match on it then? Think this needs a flag or some other way to see if the field is filled.

victorjulien: the field is under full attacker control, so no assumptions about the values of the bytes are valid

victorjulien (changes requested):
see comment

PR #7720, opened 2022-08-12
doc/conf: fix sphinx language setting

inashivb:
`sphinx-build 5.1.1` and above throws a warning which is treated as an
error while building.

```
Invalid configuration value found: 'language = None'. Update your configuration to a valid language code. Falling back to 'en' (English).
```


jufajardini (commented):
The change is very welcome, imo, but I think the error message breaks our commit body length guidelines.

victorjulien:
> The change is very welcome, imo, but I think the error message breaks our commit body length guidelines.

I think for exact messages pasted into the commits its fine.

PR #7724, opened 2022-08-15
detect/parse: test sig parsing for more actions - v1

jufajardini:
Our unittests were only covering sig parsing for alert actions. As in
environments without LibNet the reject action will not work, we must
ensure that our parser properly fails in such cases, instead of silently
accepting an unsupported action.

Added tests for the reject and drop action.

https://redmine.openinfosecfoundation.org/issues/5496

Describe changes:
- add unittest to check that the signature parser properly assesses action validity with and without libnet present
- added also a test to check the parsing for the drop action, since we only had tests with the alert action

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -4229,6 +4229,38 @@ static int SigParseBidirWithSameSrcAndDest02(void)
     PASS;
 }
 
+static int SigParseTestActionReject(void)
+{
+    DetectEngineCtx *de_ctx = DetectEngineCtxInit();
+    FAIL_IF_NULL(de_ctx);
+
+    Signature *sig = SigInit(
+            de_ctx, "reject tcp 1.2.3.4 any -> !1.2.3.4 any (msg:\"SigParseTest01\"; sid:1;)");
+#ifdef HAVE_LIBNET11
+    FAIL_IF_NULL(sig);
+#else
+    FAIL_IF_NOT_NULL(sig);
+#endif
+
+    SigFree(de_ctx, sig);
+    DetectEngineCtxFree(de_ctx);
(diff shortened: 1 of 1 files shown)


src/detect-parse.c
```diff
@@ -4229,6 +4229,38 @@ static int SigParseBidirWithSameSrcAndDest02(void)
     PASS;
 }
 
+static int SigParseTestActionReject(void)
+{
+    DetectEngineCtx *de_ctx = DetectEngineCtxInit();
+    FAIL_IF_NULL(de_ctx);
+
+    Signature *sig = SigInit(
```
victorjulien, line 4237: new code should use DetectEngineAppendSig

src/detect-parse.c
```diff
@@ -4229,6 +4229,38 @@ static int SigParseBidirWithSameSrcAndDest02(void)
...
+            de_ctx, "reject tcp 1.2.3.4 any -> !1.2.3.4 any (msg:\"SigParseTest01\"; sid:1;)");
+#ifdef HAVE_LIBNET11
+    FAIL_IF_NULL(sig);
+#else
+    FAIL_IF_NOT_NULL(sig);
+#endif
+
+    SigFree(de_ctx, sig);
```
victorjulien, line 4245: then this call is unnecessary

victorjulien (changes requested):
see comments

src/detect-parse.c
jufajardini: When one forgets to learn from past mistakes x____x

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/7725

PR #7727, opened 2022-08-17
stream: add exception policy for midstream flows - v1

jufajardini:
This allows to set a midstream-policy that can:
- fail closed (stream.midstream-policy=drop-flow)
- fail open (stream.midstream-policy=bypass)
- do nothing (default behavior)

Usage and behavior:

If Suricata is configured to inspect midstream traffic
(stream.midstream=true), then it is possible to define a midstream
policy to fully DROP or PASS a midstream flow, when Suricata identifies
one, or just ignore it and keep the default behavior (that is, inspect
it).

https://redmine.openinfosecfoundation.org/issues/5468

Describe changes:
- add a `stream.midstream-policy` option that allows for controlling if a flow should be fully dropped or bypassed. It is also possible to just ignore the policy (keeps normal behavior) if it is a midstream flow (in IPS mode)
- This change assumes that this policy should only be applied if Suricata is configured with `stream.midstream=true`

Missing: documentation (will start working on it following this PR)

suricata-verify-pr: 907
https://github.com/OISF/suricata-verify/pull/907

### DIFF
--- a/src/stream-tcp.c
+++ b/src/stream-tcp.c
@@ -468,6 +468,7 @@ void StreamTcpInitConfig(bool quiet)
     stream_config.ssn_memcap_policy = ExceptionPolicyParse("stream.memcap-policy", true);
     stream_config.reassembly_memcap_policy =
             ExceptionPolicyParse("stream.reassembly.memcap-policy", true);
+    stream_config.midstream_policy = ExceptionPolicyParse("stream.midstream-policy", true);
 
     if (!quiet) {
         SCLogConfig("stream.\"inline\": %s",
@@ -927,6 +928,9 @@ static int StreamTcpPacketStateNone(ThreadVars *tv, Packet *p,
         if (!stream_config.midstream && stream_config.async_oneside == FALSE)
             return 0;
 
+        /* Drop reason will only be used if midstream policy is set to fail closed */
+        ExceptionPolicyApply(p, stream_config.midstream_policy, PKT_DROP_REASON_FLOW_DROP);
+
         if (ssn == NULL) {
             ssn = StreamTcpNewSession(p, stt->ssn_pool_id);
             if (ssn == NULL) {
@@ -1087,6 +1091,9 @@ static int StreamTcpPacketStateNone(ThreadVars *tv, Packet *p,
     } else if (p->tcph->th_flags & TH_ACK) {
         if (!stream_config.midstream)
             return 0;
+
+        /* Drop reason will only be used if midstream policy is set to fail closed */
+        ExceptionPolicyApply(p, stream_config.midstream_policy, PKT_DROP_REASON_FLOW_DROP);
 
         if (ssn == NULL) {
             ssn = StreamTcpNewSession(p, stt->ssn_pool_id);
--- a/src/stream-tcp.h
+++ b/src/stream-tcp.h
@@ -67,6 +67,7 @@ typedef struct TcpStreamCnf_ {
 
     enum ExceptionPolicy ssn_memcap_policy;
     enum ExceptionPolicy reassembly_memcap_policy;
+    enum ExceptionPolicy midstream_policy;
 
     StreamingBufferConfig sbcnf;
 } TcpStreamCnf;


src/stream-tcp.c
```diff
@@ -927,6 +928,9 @@ static int StreamTcpPacketStateNone(ThreadVars *tv, Packet *p,
         if (!stream_config.midstream && stream_config.async_oneside == FALSE)
             return 0;
 
+        /* Drop reason will only be used if midstream policy is set to fail closed */
+        ExceptionPolicyApply(p, stream_config.midstream_policy, PKT_DROP_REASON_FLOW_DROP);
```
jufajardini, line 932: I was wondering if we would want to create a specific drop reason for this policy

jufajardini:
Trying to understand how to figure out the CodeQL issue...

src/stream-tcp.c
victorjulien: yeah probably. Something like `PKT_DROP_REASON_STREAM_MIDSTREAM`?

victorjulien (changes requested):
It feels a bit weird to have to enable midstream tracking support in order to apply a policy that would most likely lead to not tracking it after all? I think the option should probably not depend on the `stream.midstream` option.

src/stream-tcp.c
jufajardini: works for me!

jufajardini:
> It feels a bit weird to have to enable midstream tracking support in order to apply a policy that would most likely lead to not tracking it after all? I think the option should probably not depend on the `stream.midstream` option.

Fair! I wasn't sure about what should the approach be for that...

Should the stream.midstream-policy configuration option be added to the yaml file?

src/stream-tcp.c
```diff
@@ -927,6 +928,9 @@ static int StreamTcpPacketStateNone(ThreadVars *tv, Packet *p,
         if (!stream_config.midstream && stream_config.async_oneside == FALSE)
             return 0;
```
jufajardini, line 929: If we want the midstream-policy to be non `midstream=true` dependant, then I'm thinking of adding a check for `midstream_policy == EXCEPTION_POLICY_IGNORE` here and with the other check. 

Is that acceptable?

src/stream-tcp.c
jufajardini: If we do add this reason, do we want to propagate that to all packet drops of a given flow, or is logging the first one with that enough?

njlavigne:
> It feels a bit weird to have to enable midstream tracking support in order to apply a policy that would most likely lead to not tracking it after all? I think the option should probably not depend on the `stream.midstream` option.

Agree, and I think this point is important because this midstream exception policy will be most useful to IPS users needing to block midstream flows instead of track them (i.e. with midstream disabled).

njlavigne:
> This allows to set a midstream-policy that can:
> 
> * fail closed (stream.midstream-policy=drop-flow)
> * fail open (stream.midstream-policy=bypass)
> * do nothing (default behavior)

Are you additionally planning to support "reject" as an action here?  This would be highly relevant for IPS use cases.

src/stream-tcp.c
victorjulien: just the first one

src/stream-tcp.c
victorjulien: yeah, but mixing with `EXCEPTION_POLICY_PASS` should also be supported, as that still does stream reassembly and app-layer parsing/logging, just no detection (like `pass` rule keyword).

victorjulien:
> > This allows to set a midstream-policy that can:
> > 
> > * fail closed (stream.midstream-policy=drop-flow)
> > * fail open (stream.midstream-policy=bypass)
> > * do nothing (default behavior)
> 
> Are you additionally planning to support "reject" as an action here? This would be highly relevant for IPS use cases.

Can you open a ticket for this? It would be a generic option for all exception policies.

njlavigne:
> > > This allows to set a midstream-policy that can:
> > > 
> > > * fail closed (stream.midstream-policy=drop-flow)
> > > * fail open (stream.midstream-policy=bypass)
> > > * do nothing (default behavior)
> > 
> > 
> > Are you additionally planning to support "reject" as an action here? This would be highly relevant for IPS use cases.
> 
> Can you open a ticket for this? It would be a generic option for all exception policies.

https://redmine.openinfosecfoundation.org/issues/5503

jufajardini:
> > It feels a bit weird to have to enable midstream tracking support in order to apply a policy that would most likely lead to not tracking it after all? I think the option should probably not depend on the `stream.midstream` option.
> 
> Agree, and I think this point is important because this midstream exception policy will be most useful to IPS users needing to block midstream flows instead of track them (i.e. with midstream disabled).

Thanks for the input and explanation! Working on that...

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/7748

PR #7728, opened 2022-08-17
WIP: Dataset ops v2.4

regit:
Update of #7687 

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
- https://redmine.openinfosecfoundation.org/issues/5184
- https://redmine.openinfosecfoundation.org/issues/5383

Describe changes:
- Rebase
- Address comments on documentation

suricata-verify-pr: 908

PR #7687, opened 2022-08-03
Dataset ops v2.3

regit:
Update of #7626 with a rebase and windows build fix.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) tickets:
* https://redmine.openinfosecfoundation.org/issues/5184
* https://redmine.openinfosecfoundation.org/issues/5383

Describe changes:
- rebase
- fix windows build

suricata-verify-pr: 880

PR #7732, opened 2022-08-18
Dataset ops v2.5

regit:
Update of #7687 (and #7728)

- https://redmine.openinfosecfoundation.org/issues/5184
- https://redmine.openinfosecfoundation.org/issues/5383

Describe changes:
- Rebase
- Address comments on documentation
- Fix MacOSX build

suricata-verify-pr: 908


regit:
I don't get what is happening there: suricata-verify passes locally with the same branch as the one pointed by the PR.

doc/userguide/rules/datasets.rst
```diff
@@ -183,6 +227,10 @@ md5
   in the file as hex encoded string
 sha256
   in the file as hex encoded string
+ipv4
+  in the file as string
+ip
+  in the file as string, it can be IPv6 or IPv4 address (standard notation or IPv4 in IPv6 one)
```
victorjulien, line 233: "IPv4 in IPv6 one" -> not sure what is meant here?

doc/userguide/rules/ipaddr.rst
```diff
@@ -0,0 +1,29 @@
+IP Addresses Match
```
victorjulien, line 1: I think this needs integration with regular address matching logic in rules. Might confuse ppl otherwise.

src/datasets-ipv4.h
```diff
@@ -0,0 +1,39 @@
+/* Copyright (C) 2017-2019 Open Information Security Foundation
```
victorjulien, line 1: 2022

src/datasets.c
```diff
@@ -57,6 +60,10 @@ enum DatasetTypes DatasetGetTypeFromString(const char *s)
         return DATASET_TYPE_SHA256;
     if (strcasecmp("string", s) == 0)
         return DATASET_TYPE_STRING;
+    if (strcasecmp("ipv4", s) == 0)
+        return DATASET_TYPE_IPV4;
+    if (strcasecmp("ip", s) == 0)
```
victorjulien, line 65: is `ip` for ipv6 or mixed?

src/datasets.c
```diff
@@ -156,6 +163,217 @@ static int ParseRepLine(const char *in, size_t ins, DataRepType *rep_out)
...
+    /* Checking IPv6 case */
+    char *got_colon = strchr(line, ':');
+    if (got_colon) {
+        if (inet_pton(AF_INET6, line, in6) != 1) {
+            FatalErrorOnInit(SC_ERR_FATAL, "dataset data parse failed %s/%s: %s", set->name,
+                    set->load, line);
+            return -1;
+        }
+        /* IPv4 in IPv6 notation needs transformation to internal Suricata storage */
+#ifndef OS_WIN32
```
victorjulien, line 255: can we hide this kind of ifdeffery?

victorjulien: we do address parsing elsewhere w/o needing this?

src/detect-ipaddr.c
```diff
@@ -0,0 +1,154 @@
...
+    BUG_ON(g_dest_ipaddr_buffer_id < 0);
+
+    DetectBufferTypeSupportsPacket(KEYWORD_NAME_DST);
+
+    DetectPktMpmRegister(KEYWORD_NAME_DST, 2, PrefilterGenericMpmPktRegister, GetDataDst);
+
+    DetectPktInspectEngineRegister(
+            KEYWORD_NAME_DST, GetDataDst, DetectEngineInspectPktBufferGeneric);
+
+    /* NOTE: You may want to change this to SCLogNotice during development. */
```
victorjulien, line 96: nit: remove comment

src/detect-ipaddr.c
```diff
@@ -0,0 +1,154 @@
...
+ *  \brief get the data to inspect from the buffer
+ *
+ *  \retval buffer or NULL in case of error
+ */
+static InspectionBuffer *GetDataDst(DetectEngineThreadCtx *det_ctx,
+        const DetectEngineTransforms *transforms, Packet *p, const int list_id)
+{
+    InspectionBuffer *buffer = InspectionBufferGet(det_ctx, list_id);
+    if (buffer->inspect == NULL) {
+        InspectionBufferSetup(det_ctx, list_id, buffer, p->dst.address.address_un_data8, 16);
```
victorjulien, line 145: so the buffer size (bsize keyword) will be 16 even for ipv4?

victorjulien: also matching on this for ipv4, would this work?
`ip.src; content:"|00 00 00 00 00 00|";`

src/runmode-unix-socket.c
```diff
@@ -762,6 +762,100 @@ TmEcode UnixSocketDatasetRemove(json_t *cmd, json_t* answer, void *data)
...
+        return TM_ECODE_FAILED;
+    }
+
+    Dataset *set = DatasetFind(set_name, t);
+    if (set == NULL) {
+        json_object_set_new(answer, "message", json_string("set not found or wrong type"));
+        return TM_ECODE_FAILED;
+    }
+
+    THashCleanup(set->hash);
```
victorjulien, line 804: is this safe on a in use set? what happens if entries are in use?

victorjulien (changes requested):

src/datasets.c
regit: Mixed.

src/datasets.c
regit: I tried to find and did not find it.

src/detect-ipaddr.c
regit: yes, ipv4 storage in ip set takes as much as IPv6 one.

regit: it is a binary field so the content match should work.

doc/userguide/rules/datasets.rst
regit: This is notation like 192.168.1.2 is ::FFFF:192.168.1.2 (https://www.ibm.com/docs/en/zos/2.2.0?topic=addresses-ipv4-mapped-ipv6)

src/datasets.c
victorjulien: did you look at `detect-engine-address.c`

src/detect-ipaddr.c
victorjulien: I don't agree with this. Having 16 bytes padded with 12 0 bytes in case of IPv4 makes no sense.

regit: Does not make sense for matching or for memory usage ?

doc/userguide/rules/datasets.rst
jufajardini: Maybe we can add a note/explanation, for those who are unfamiliar with this notation?

src/detect-ipaddr.c
victorjulien: For matching. We shouldn't expose nonsensical data to the rule lang. I think the simple solution is just to expose 4 bytes for IPv4, 16 for IPv6.

src/datasets.c
regit: Indeed it can be used.

src/detect-ipaddr.c
regit: I found a way to do that. It will be in the next MR.

src/runmode-unix-socket.c
regit: THashCleanup is going to move the data to spare if the reference counter is not null. So we should not touch any data that is in use. Potentially we could have some left over entries but this is not a big deal for this feature.

PR #7760, opened 2022-08-24
exceptions: error out when invalid policy is used - v1

jufajardini:
Before, if an invalid value was passed as exception policy, Suricata
would log a warning and set the exception policy to "ignore". This is a
very different result, than, say, dropping or bypassing a midstream flow.

Task #5504

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5504

Describe changes:
- instead of a config log informing that the exception policy will be set to `ignore`, fail at init, indicating the error, so it can be fixed asap
<Error> (ExceptionPolicyParse) -- [ERRCODE: SC_ERR_INVALID_ARGUMENT(13)] - "flow-drop" is not a valid exception policy value,  please check Suricata documentation for available options.

PR #5504, opened 2020-10-21
detect: filesha1: convert unittests to use PASS/FAIL API

Sylfrena:
Currently, unit tests return integer values to denote whether the test fails or passes on a certain condition. `Pass` and `Fail` are denoted by returning 0 and 1 respectively.

Modify `detect-filesha1.c` test to replace these integer values with the PASS/FAIL API.

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/

Fixes ~#4037~

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4037?tab=properties

Describe changes:
- Removed comments describing the initial test methods.
- Replaces 1, 0 values with PASS/FAIL macros respectively.

Changes in v2:
- Modify code to make changes only to unit tests.

PR #7766, opened 2022-08-25
exceptions: error out when invalid policy is used - v2

jufajardini:
Before, if an invalid value was passed as exception policy, Suricata
would log a warning and set the exception policy to "ignore". This is a
very different result, than, say, dropping or bypassing a midstream flow.

Task #5504

https://redmine.openinfosecfoundation.org/issues/5504

Previous PR: #7760

Changes from previous PR: 
- Make the error message more informative and user friendly:
`<Error> (ExceptionPolicyParse) -- [ERRCODE: SC_ERR_INVALID_ARGUMENT(13)] - "flow-drop" is not a valid exception policy value. Valid options are drop-flow, pass-flow, bypass, drop-packet or pass-packet. For more infomation, please check the Suricata documentation.`

### DIFF
--- a/src/util-exception-policy.c
+++ b/src/util-exception-policy.c
@@ -88,7 +88,11 @@ enum ExceptionPolicy ExceptionPolicyParse(const char *option, const bool support
             policy = EXCEPTION_POLICY_IGNORE;
             SCLogConfig("%s: %s", option, value_str);
         } else {
-            SCLogConfig("%s: ignore", option);
+            FatalErrorOnInit(SC_ERR_INVALID_ARGUMENT,
+                    "\"%s\" is not a valid exception policy value. Valid options are drop-flow, "
+                    "pass-flow, bypass, drop-packet or pass-packet. For more infomation, please "
+                    "check the Suricata documentation.",
+                    value_str);
         }
 
         if (!support_flow) {


src/util-exception-policy.c
```diff
@@ -88,7 +88,11 @@ enum ExceptionPolicy ExceptionPolicyParse(const char *option, const bool support
             policy = EXCEPTION_POLICY_IGNORE;
             SCLogConfig("%s: %s", option, value_str);
         } else {
-            SCLogConfig("%s: ignore", option);
+            FatalErrorOnInit(SC_ERR_INVALID_ARGUMENT,
+                    "\"%s\" is not a valid exception policy value. Valid options are drop-flow, "
+                    "pass-flow, bypass, drop-packet or pass-packet. For more infomation, please "
+                    "check the Suricata documentation.",
```
victorjulien, line 94: thinking about this again, I wonder if we should reference the docs. Its not something we really do (much) elsewhere. Why would we do so here? Or, should we do so everywhere?

jufajardini: I think that... if the information given can be considered enough for most users to understand what they have to next, then, indeed referring to the documentation may be too much.

On the other hand, it is a way of telling them that we do have more detailed information on topics that may need elaboration. What is a good balance? 🤔

victorjulien: Also wondering if we should then mention a specific doc/url, instead of generally saying "go read the docs".

In reality most ppl will just google it most likely.

jufajardini: I'll remove that bit :P

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/7772

PR #7795, opened 2022-08-30
make copyrights current

inashivb:
Done w regex and substitution
Missing rust


src/app-layer-dnp3-objects.c
```diff
@@ -1,4 +1,4 @@
-/* Copyright (C) 2015 Open Information Security Foundation
+/* Copyright (C) 2022 Open Information Security Foundation
```
victorjulien, line 1: this type of update is wrong, should become `2015-2022`

PR #7797, opened 2022-08-30
Minor copyright/v2

inashivb:
Previous PR: https://github.com/OISF/suricata/pull/7795

Changes since last PR:
- Fix copyrights for non range years
- Make updates to the entire codebase
- clang fmt changes done separately as they seemed intrusive


rust/cbindgen.toml
```diff
@@ -10,7 +10,7 @@ language = "C"
 
 # An optional string of text to output at the beginning of the generated file
 # default: doesn't emit anything
-header = """/* Copyright (C) 2019 Open Information Security Foundation
+header = """/* Copyright (C) 2019-2022 Open Information Security Foundation
```
inashivb, line 13: Should probably be 2022?

inashivb:
@jufajardini could you please help review this?

rust/cbindgen.toml
jufajardini: Makes sense to me!

rust/src/detect.rs
```diff
@@ -1,4 +1,4 @@
-/* Copyright (C) 2022 Open Information Security Foundation
+/* Copyright (C) 2022-2022 Open Information Security Foundation
```
jufajardini, line 1: Another exception case! :P

jufajardini (changes requested):
Found some files that were created on 2022 and which won't need the copyright update either, commented on one of them :)

catenacyber:
Did you use a script to create this and modify > 1k files ?
Can it be run in CI ?

inashivb:
> Did you use a script to create this and modify > 1k files ? 

Yeah. Kind of. I used a grep + sed oneliner.

> Can it be run in CI ?

I was thinking that maybe we could do this the start of every year. Now that all the files are current, what should we add in the CI?
Or do you mean using CI as an indicator to do this kind of an update. That's actually much more reliable than remembering to do this every year indeed.
Did I get you right?

rust/src/detect.rs
inashivb: Thanks. :)
Fixed in the new rev.

catenacyber:
> > Did you use a script to create this and modify > 1k files ?
> 
> Yeah. Kind of. I used a grep + sed oneliner. But, there are some corner cases.
> 

Can you put this one-liner in the commit description ?

> > Can it be run in CI ?
> 
> I was thinking that maybe we could do this the start of every year. Now that all the files are current, what should we add in the CI? Or do you mean using CI as an indicator to do this kind of an update. That's actually much more reliable than remembering to do this every year indeed. Did I get you right?

Should every file have an updated copyright to 2022 ? Or just until the last year it was modified ?

inashivb:
> Can you put this one-liner in the commit description ?

Yes. I'll write down a script for the series of commands. 
 
> Should every file have an updated copyright to 2022 ? Or just until the last year it was modified ?

hmmm i thought it was always supposed to be current. i am not very sure. No one seemed to mention this either.

jufajardini:
> > Can you put this one-liner in the commit description ?
> 
> Yes. I'll write down a script for the series of commands.
> 
> > Should every file have an updated copyright to 2022 ? Or just until the last year it was modified ?
> 
> hmmm i thought it was always supposed to be current. i am not very sure. No one seemed to mention this either.

I had a feeling that we did the "update-when-change" thing because it was more practical, only...

inashivb:
> I had a feeling that we did the "update-when-change" thing because it was more practical, only...

In that case, let's take other folks' opinion in. Maybe this is not needed at all then.

PR #7798, opened 2022-08-31
divert changes for FreeBSD

glebius:
You may skip the patch that removes SO_BROADCAST. I'm 99% sure this setsockopt() is useless, but I haven't tested.

The second patch is necessary to work on FreeBSD 14 and beyond without a warning message.

### DIFF
--- a/src/source-ipfw.c
+++ b/src/source-ipfw.c
@@ -334,7 +334,11 @@ TmEcode ReceiveIPFWThreadInit(ThreadVars *tv, const void *initdata, void **data)
 
     IPFWMutexInit(nq);
     /* We need a divert socket to play with */
+#ifdef PF_DIVERT
+    if ((nq->fd = socket(PF_DIVERT, SOCK_RAW, 0)) == -1) {
+#else
     if ((nq->fd = socket(PF_INET, SOCK_RAW, IPPROTO_DIVERT)) == -1) {
+#endif
         SCLogError(SC_ERR_IPFW_SOCK,"Can't create divert socket: %s", strerror(errno));
         SCReturnInt(TM_ECODE_FAILED);
     }
--- a/src/source-ipfw.c
+++ b/src/source-ipfw.c
@@ -323,7 +323,6 @@ TmEcode ReceiveIPFWLoop(ThreadVars *tv, void *data, void *slot)
 TmEcode ReceiveIPFWThreadInit(ThreadVars *tv, const void *initdata, void **data)
 {
     struct timeval timev;
-    int flag;
     IPFWThreadVars *ntv = (IPFWThreadVars *) initdata;
     IPFWQueueVars *nq = IPFWGetQueue(ntv->ipfw_index);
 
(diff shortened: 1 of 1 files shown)


glebius:
I have communicated about the changes to Nick Rogness via email.

src/source-ipfw.c
```diff
@@ -350,15 +353,6 @@ TmEcode ReceiveIPFWThreadInit(ThreadVars *tv, const void *initdata, void **data)
         SCReturnInt(TM_ECODE_FAILED);
     }
 
-    /* set SO_BROADCAST on the divert socket, otherwise sendto()
```
victorjulien, line 353: @ndenev git shows me you explicitly added this a long time ago in 29b69fb026a9dd06951c94b28d7bd38539e30fe9. Any thoughts on whether its still useful?

ndenev: I vaguely remember needing this to make it work back then, but it might no longer be needed as Gleb is saying.

glebius: Could this be related to some other operating system, forked off FreeBSD? E.g. Mac OS X, DragonflyBSD? My analysis shows that it was never needed.

victorjulien (approved):

src/source-ipfw.c
ndenev: It was on FreeBSD, don’t really remember if I was running CURRENT or most likely 9-STABLE.

This is what I found (internet archives are better than my memory :)) https://oisf-devel.openinfosecfoundation.narkive.com/IQOKQqo8/suricata-ipfw-ips-mode-on-freebsd-and-broadcast-packets

glebius: Thanks Nikolay! Victor, can you please skip the first patch from me, and just commit the one that changes the socket syscall to use PF_DIVERT?

P.S. You made me puzzled with this SO_BROADCAST, I'm going to dig out the truth :)

glebius: Ok, I did very thorough code review and I can not see: 
- how SO_BROADCAST may have any affect on a divert(4) socket
- how divert(4) socket may return EACCES on a packet due to packet being a broadcast

This is true both for modern FreeBSD 14 and to FreeBSD 9. So, after all I written a small program to test that. Program requires an interface configured to 10.0.0.x/24 and ipfw default to allow ruleset. It works same on FreeBSD 9 and FreeBSD 14 - successfully sends broadcast to the network.

So, guys, it is up to you whether to merge the commit that removes SO_BROADCAST or not. But please merge the one that supports new API of the socket call. Thanks!

```
#include <sys/types.h>
#include <sys/socket.h>
#include <netinet/in.h>
#include <netinet/ip.h>
#include <netinet/ip_icmp.h>
#include <arpa/inet.h>

#include <stdio.h>
#include <errno.h>
#include <err.h>

int
main(int argc, char *argv[])
{
        static char out[IP_MAXPACKET];
        struct sockaddr_in sin = {
                .sin_family = AF_INET,
                .sin_addr = INADDR_ANY,
                .sin_port = htons(8668)
        };
        struct ip *ip;
        struct icmp *icp;
        ssize_t sent;
        int s;

        if ((s = socket(PF_INET, SOCK_RAW, IPPROTO_DIVERT)) < 0)
                err(1, "socket");

        if (bind(s, (struct sockaddr *)&sin, sizeof(sin)) < 0)
                err(1, "bind");

        ip = (struct ip *)out;
        icp = (struct icmp *)(ip + 1);

        *ip = (struct ip ){
                .ip_v = IPVERSION,
                .ip_hl = sizeof(struct ip) >> 2,
                .ip_ttl = 128,
                .ip_p = IPPROTO_ICMP,
                .ip_len = htons(sizeof(*ip) + sizeof(*icp)),
                .ip_dst = inet_addr("10.0.0.255"),
        };

        *icp = (struct icmp ){
                .icmp_type = ICMP_ECHO,
                .icmp_seq = htons(1),
                .icmp_id = htons(1),
        };

        sent = sendto(s, &out, sizeof(*ip) + sizeof(*icp), 0,
            (struct sockaddr *)&sin, sizeof(sin));
        printf("rv = %d, errno %d\n", sent, errno);
}
```

victorjulien:
First part is merged in #7839, thanks a lot @glebius. I'm inclined to also take the broadcast patch although we'll not be able to test it.

PR #7799, opened 2022-08-31
Minor copyright/v3

inashivb:
Previous PR: https://github.com/OISF/suricata/pull/7797

Changes since v2:
- Fix corner case of files created in 2022 being updated by the regex to 2022-2022


jufajardini (approved):
Looks good to me! :D

(no more being considered mean for nitting PRs for copyright years :D )

suricata-qa:

Pipeline 8904
WARNING: THERE IS A KNOWN BAD BASELINE WITH PACKET DROPS. bE MINDFUL OF ANY RESULTS.

inashivb:
Seems like the verdict is that we only update the copyrights to the last modified date. So, this is useless.

catenacyber:
So, you just need to improve your script with some `git log -- filename` magic ;-)

inashivb:
> So, you just need to improve your script with some `git log -- filename` magic ;-)

I was thinking how about a git hook that updates copyrights in all modified files before pushing?
Been thinking a while that we should probably have certain git hooks in the repo that get installed for every developer.. Common stuff like this and clang-format, commit build etc

catenacyber:
Good idea.

The developer has to manually install them at some point, ie `git clone` will not modify any files in your .git/hooks/ directory (because that would be RCE)

inashivb:
> Good idea.
> 
> The developer has to manually install them at some point, ie `git clone` will not modify any files in your .git/hooks/ directory (because that would be RCE)

huh. did not think of it that way. indeed.. thank you!

PR #7821, opened 2022-09-01
detect: option to run flowbits on stream

regit:
When a flowbit is set on application layer tests, this has
for consequence that there is no point in checking it per packet.
This patch adds the `data` option to the flowbits keyword so
the evaluation is done at the stream level and not at the packet
one.

As a result, using the data option is preventing the bug
to appear.

Ticket: #2836

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2836

Describe changes:
- Add data option to flowbits

suricata-verify-pr: 924

### DIFF
--- a/src/detect-flowbits.c
+++ b/src/detect-flowbits.c
@@ -46,7 +46,7 @@
 #include "util-unittest.h"
 #include "util-debug.h"
 
-#define PARSE_REGEX         "^([a-z]+)(?:,\\s*(.*))?"
+#define PARSE_REGEX "^([a-z]+)(?:,\\s*([^,]*))?(?:,\\s*([a-z]+))*"
 static DetectParseRegex parse_regex;
 
 #define MAX_TOKENS 100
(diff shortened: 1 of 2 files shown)


suricata-qa:

Pipeline 8977

jufajardini (commented):
Doc change looks good to me! :)

src/detect-flowbits.c
```diff
@@ -340,7 +358,10 @@ int DetectFlowbitSetup (DetectEngineCtx *de_ctx, Signature *s, const char *rawst
         case DETECT_FLOWBITS_CMD_ISNOTSET:
         case DETECT_FLOWBITS_CMD_ISSET:
             /* checks, so packet list */
-            SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_MATCH);
+            if (on_data)
+                SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_PMATCH);
```
victorjulien, line 362: this won't work and will trigger https://github.com/OISF/suricata/blob/master/src/detect-engine-content-inspection.c#L651

victorjulien: the PMATCH list is only for matches that can be evaluated by `DetectEngineContentInspection`

victorjulien: gdb will tell you `DetectFlowbitMatchIsnotset` will not be called

victorjulien:
I wonder if a better solution would be to add more classes of "bits": txbits, statebits, framebits, etc, where the execution and lifetime would align better with the type of data that is inspected.

PR #7823, opened 2022-09-02
Rust mimetype

regit:
This patchset proposes an alternative to filemagic as it has multiple issues:

- filemagic is known for bad performance
- database is not consistent between system
- results are sometime too detailed (do we care about size of image)

This patchet uses a [rust crate](https://docs.rs/tree_magic_mini/latest/tree_magic_mini/) to extract the mime type instead of the magic:

- it is really faster than filemagic
- mimetype database is included in the crate
- mimetype are easier to predict

It thus fixes most of the issues present with filemagic and thus is a nice alternative.

Describe changes:
- add mimetype computation
- add mimetype to fileinfo
- add file.mimetype keyword

### DIFF
--- a/doc/userguide/rules/file-keywords.rst
+++ b/doc/userguide/rules/file-keywords.rst
@@ -31,6 +31,15 @@ Example::
 
   fileext:"jpg";
 
+file.mime_type
+--------------
+
+Sticky buffer to match on the file mime type.
+
+Example::
+
+  file.mime_type; content:"application/x-executable";
+
 filemagic
 ---------
 
--- a/etc/schema.json
+++ b/etc/schema.json
@@ -122,6 +122,9 @@
                     "magic": {
                         "type": "string"
                     },
+                    "mimetype": {
+                        "type": "string"
+                    },
                     "md5": {
                         "type": "string"
                     },
@@ -1248,6 +1251,9 @@
                     "type": "boolean"
                 },
                 "magic": {
+                    "type": "string"
+                },
+                "mimetype": {
                     "type": "string"
                 },
                 "md5": {
--- a/rust/Cargo.toml.in
+++ b/rust/Cargo.toml.in
@@ -56,6 +56,7 @@ md-5 = "~0.10.1"
 regex = "~1.5.5"
 lazy_static = "~1.4.0"
 base64 = "~0.13.0"
+tree_magic_mini = "~3.0.3"
 
 suricata-derive = { path = "./derive" }
 
--- a/rust/src/filemimetype.rs
+++ b/rust/src/filemimetype.rs
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2022 Open Information Security Foundation
+/* Copyright (C) 2022 Open Information Security Foundation
  *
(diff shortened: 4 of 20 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_asan | | |
 

Pipeline 8992

victorjulien:
The crate repo shows limited signs of activity, with the last commit about a year ago? https://github.com/mbrubeck/tree_magic/
@jasonish any thoughts on the crate?

victorjulien:
Moved to draft as it needs some serious patching up as CI shows.

jasonish:
Is the database included, from the tree_magic readme:

> By default, tree_magic_mini will attempt to load the shared MIME info database from the standard locations at runtime.

Embedding the database requires the `tree_magic_db` feature, but the database is GPL and warned about in the README. Without embedding it, we still have the issue of inconsistency, or the magic file not even existing on a system.

The performance increase would be nice.

I've always thought we should roll this ourselves, with a limit magic database.  I think Snort does this.  Even with this crate, we'd probably need to roll our own database due to the GPL.

jasonish:
Just some followup... In the current version of this PR I verified that the embedded/bundled magic database is not being used.  Instead if checks the following paths:
```
"/usr/share/mime/magic",
"/usr/local/share/mime/magic",
"/$HOME/.local/share/mime/magic",
```
If it can't load magic from those file, it does have some builtin's.. For example, it will detect a plain text file as `text/plain`, or a PDF as `application/octet-stream`, but with magic databases it would use the pdf mimetype.

The questionable files are the magic database, which come from freedesktop.org shared-mime-info project which is licensed with the GPLv2+, https://freedesktop.org/wiki/Software/shared-mime-info/, but being data files and not code, I'm not sure what this means.

If we were to include these data files, but *not* embed them in the binary, then we would need to modify `tree_magic_mini` to support loading the database from arbitrary locations, as it currently uses a hardcoded list.

regit:
Should we make the build only if an enable-flag is passed ? This would avoid the GPL issue.

jasonish:
> Should we make the build only if an enable-flag is passed ? This would avoid the GPL issue.

I'm not sure I like the idea of something I see as an obvious enhancement over an existing feature being gated by a license flag.

jasonish:
Ooops, didn't mean to close.

norg:
Just to give you an idea on the impact, we did some performance testing with that feature compared to file-magic:

34 Gbit/s Traffic on 40G with baseline t-rex traffic an 4.19 kernel after 10min running:

`force-mimetype: false` and `file-magic: false` 4.0 % Drop

`force-mimetype: true` and `file-magic: false` 4.1 % Drop

`force-mimetype: false` and `file-magic: true` 43 % Drop

So for setups that use this a lot it could be a big improvement.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_asan | | |
 

Pipeline 8992

victorjulien:
I've created ticket https://redmine.openinfosecfoundation.org/issues/5894

However I agree with Jason that this approach is not viable due to licensing restrictions.

PR #7833, opened 2022-09-06
exceptions: add reject action support to exception policy - v1

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/projects/suricata/issues/5503

Describe changes:
- This enables the usage of 'reject' as an exception policy. As for both IPS and IDS modes the intended result of sending a reject packet is to reject the related flow, this will effectively mean setting the reject action to the packet that triggered the exception condition, and then dropping the associated flow.

suricata-verify-pr: 928
https://github.com/OISF/suricata-verify/pull/928


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | tlpr1_stats_chk | | |
| .flow.memuse | 562615488 | 625215168 | 111.13% | 

Pipeline 9031

src/util-exception-policy.c
```diff
@@ -32,6 +32,10 @@ void ExceptionPolicyApply(Packet *p, enum ExceptionPolicy policy, enum PacketDro
         switch (policy) {
             case EXCEPTION_POLICY_IGNORE:
                 break;
+            case EXCEPTION_POLICY_REJECT:
+                SCLogDebug("EXCEPTION_POLICY_REJECT");
+                PacketDrop(p, ACTION_REJECT, drop_reason);
```
victorjulien, line 37: will calling `PacketDrop` twice in this path have any side effects?

jufajardini: I've considered that... Since now `PacketDrop` uses `PACKET_UPDATE_ACTION` and this one will not override the existing action or drop reason, my understanding is that it should be ok...

victorjulien:
Merged in #7839, thanks!

PR #7837, opened 2022-09-08
Smb file multiflow 4861 v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4861

Describe changes:
- smb : handle multi-stream file transfers

This is a draft for feedback.

Based on a private pcap for testing, this uses SMB2_FILE_ENDOFFILE_INFO cf Wireshark filter `smb2.file_info.infolevel == 0x14 to start using multi-stream logic

Questions :
- How good is this `SMB2_FILE_ENDOFFILE_INFO` assumption ?
- How much should we use ippair logic ?
ippair is currently not used by the http range logic
ippair can improve the locking mechanisms
But then, it is harder to have a global memcap on the buffering done for multi-stream smb files...
This is the big question.
I am not sure about how much code HTTP range and smb should share...
Should multi-files share one memcap for instance ?
- Should we support parallel multiple multi-streams files transfers ?
That is between 2 IPs, multiple files (with their SMB guid) concurrently being transferred, each over multiple streams
- Should we handle reads as writes ? If so, I would need a pcap (cf SMB2_FILE_ENDOFFILE_INFO assumption)
- This draft is SMB2 only, would we want SMB1 ?
 
Also, the current multi-stream logic will not log a file until it is complete.
Because there may be a new flow coming that will complete the file.
And because to log it, we have to move its ownership from the global hash table to the transaction which will end up freeing it...
Thoughts about this ?

### DIFF
--- a/rust/src/smb/files.rs
+++ b/rust/src/smb/files.rs
@@ -22,8 +22,10 @@ use crate::filecontainer::*;
 
 use crate::smb::smb::*;
 
+use std::os::raw::c_uchar;
+
 /// File tracking transaction. Single direction only.
-#[derive(Default, Debug)]
+#[derive(Debug)]
 pub struct SMBTransactionFile {
     pub direction: Direction,
     pub fuid: Vec<u8>,
@@ -33,14 +35,29 @@ pub struct SMBTransactionFile {
     /// after a gap, this will be set to a time in the future. If the file
     /// receives no updates before that, it will be considered complete.
     pub post_gap_ts: u64,
+    pub file_range: *mut HttpRangeContainerBlock,
+    pub multi: bool,
 }
 
 impl SMBTransactionFile {
     pub fn new() -> Self {
         return Self {
             file_tracker: FileTransferTracker::new(),
-            ..Default::default()
+            file_range: std::ptr::null_mut(),
+            multi: false,
+            post_gap_ts: 0,
+            direction: Direction::default(),
+            fuid: Vec::new(),
+            file_name: Vec::new(),
+            share_name: Vec::new(),
         }
+    }
+}
+
+impl Drop for SMBTransactionFile {
+    fn drop(&mut self) {
+        // should have been already closed
+        debug_validate_bug_on!(!self.file_range.is_null());
     }
 }
 
@@ -55,6 +72,13 @@ pub fn filetracker_newchunk(ft: &mut FileTransferTracker, files: &mut FileContai
                     chunk_size, 0, is_last, xid); }
         None => panic!("no SURICATA_SMB_FILE_CONFIG"),
     }
+}
+
+// Defined in app-layer-htp-range.h
+extern "C" {
+    pub fn HttpRangeAppendData(
+        c: *mut HttpRangeContainerBlock, data: *const c_uchar, data_len: u32,
+    ) -> std::os::raw::c_int;
 }
 
 impl SMBState {
@@ -122,7 +146,7 @@ impl SMBState {
 
     // update in progress chunks for file transfers
     // return how much data we consumed
-    pub fn filetracker_update(&mut self, direction: Direction, data: &[u8], gap_size: u32) -> u32 {
+    pub fn filetracker_update(&mut self, direction: Direction, data: &[u8], gap_size: u32, eof: bool) -> u32 {
         let mut chunk_left = if direction == Direction::ToServer {
             self.file_ts_left
         } else {
@@ -175,6 +199,32 @@ impl SMBState {
                     }
 
                     let file_data = &data[0..data_to_handle_len];
+                    if !tdf.file_range.is_null() {
+                        unsafe {
+                            HttpRangeAppendData(tdf.file_range, file_data.as_ptr(), data_to_handle_len as u32);
+                        }
+                        if chunk_left == 0 || eof {
+                            let added = if let Some(c) = unsafe { SC } {
+                                let added = (c.HTPFileCloseHandleRange)(
+                                    files,
+                                    flags,
+                                    tdf.file_range,
+                                    std::ptr::null_mut(),
+                                    0,
+                                );
+                                (c.HttpRangeFreeBlock)(tdf.file_range);
+                                added
+                            } else {
+                                false
+                            };
+                            tdf.file_range = std::ptr::null_mut();
+                            if added {
+                                tx.tx_data.incr_files_opened();
+                            }
+                        }
+                    }
+
+                    //TODOsmbmulti5 use eof ?
                     let cs = tdf.file_tracker.update(files, flags, file_data, gap_size);
                     cs
                 } else {
--- a/rust/src/smb/smb.rs
+++ b/rust/src/smb/smb.rs
@@ -593,8 +593,9 @@ impl SMBTransaction {
     }
 
     pub fn free(&mut self) {
-        debug_validate_bug_on!(self.tx_data.files_opened > 1);
-        debug_validate_bug_on!(self.tx_data.files_logged > 1);
+        // there can be a multi file + a regular one logged
+        debug_validate_bug_on!(self.tx_data.files_opened > 2);
+        debug_validate_bug_on!(self.tx_data.files_logged > 2);
     }
 }
 
@@ -756,6 +757,8 @@ pub struct SMBState<> {
     pub ssn2vec_map: HashMap<SMBCommonHdr, Vec<u8>>,
     /// map guid to filename
     pub guid2name_map: HashMap<Vec<u8>, Vec<u8>>,
+    /// map guid to file size
+    pub guid2eof_map: HashMap<Vec<u8>, u64>,
     /// map ssn key to read offset
     pub ssn2vecoffset_map: HashMap<SMBCommonHdr, SMBFileGUIDOffset>,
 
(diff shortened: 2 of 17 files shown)


rust/src/smb/files.rs
```diff
@@ -33,17 +35,32 @@ pub struct SMBTransactionFile {
     /// after a gap, this will be set to a time in the future. If the file
     /// receives no updates before that, it will be considered complete.
     pub post_gap_ts: u64,
+    pub file_range: *mut HttpRangeContainerBlock,
```
catenacyber, line 38: @jasonish can I keep the derive Default and have `file_range` being null ?

rust/src/smb/files.rs
```diff
@@ -33,17 +35,32 @@ pub struct SMBTransactionFile {
...
+            fuid: Vec::new(),
+            file_name: Vec::new(),
+            share_name: Vec::new(),
         }
     }
 }
 
+impl Drop for SMBTransactionFile {
+    fn drop(&mut self) {
+        // should have been already closed
```
catenacyber, line 59: By the way, maybe we should try that for HTTP ranges

rust/src/smb/files.rs
```diff
@@ -175,6 +199,32 @@ impl SMBState {
...
+                                false
+                            };
+                            tdf.file_range = std::ptr::null_mut();
+                            if added {
+                                tx.tx_data.incr_files_opened();
+                            }
+                        }
+                    }
+
+                    //TODOsmbmulti5 use eof ?
```
catenacyber, line 227: @victorjulien who would know if we should use `stream_slice.flags() & STREAM_EOF` in this SMB code to properly close the file ? (somehow unrelated to this PR)

rust/src/smb/smb2.rs
```diff
@@ -264,7 +268,19 @@ pub fn smb2_read_response_record<'b>(state: &mut SMBState, r: &Smb2Record<'b>)
     }
 }
 
-pub fn smb2_write_request_record<'b>(state: &mut SMBState, r: &Smb2Record<'b>)
+// Defined in app-layer-smb.h
```
catenacyber, line 271: No declaration in app-layer-smb.h, but code in  app-layer-smb.c

rust/src/smb/smb2.rs
```diff
@@ -449,6 +501,16 @@ pub fn smb2_request_record<'b>(state: &mut SMBState, r: &Smb2Record<'b>)
                             tx.vercmd.set_smb2_cmd(SMB2_COMMAND_SET_INFO);
                             true
                         }
+                        Smb2SetInfoRequestData::EOF(ref eof) => {
+                            if let Some(_fname) = state.guid2name_map.get(rd.guid) {
+                                // gets removed on file close
+                                //TODOsmbmulti5 by the way guid2name_map never removed and not limited !
```
catenacyber, line 507: Should I create a ticket for `guid2name_map` : stateful structure forever growing without limits ?

src/app-layer-smb.c
```diff
@@ -36,11 +39,145 @@ static SuricataFileContext sfc = { &sbcfg };
...
+                if (key->filename == NULL) {
+                    SCFree(key);
+                    goto end;
+                }
+                key->filename_len = name_len;
+                memcpy(key->filename, filename, name_len);
+            }
+            goto end;
+        } else {
+            // different guid : keep old one TODOsmbmulti6 : handle multimulti ?
```
catenacyber, line 85: Do we want to handle multiple files, each over multiple streams ?

catenacyber: For one ip pair

src/app-layer-smb.c
```diff
@@ -36,11 +39,145 @@ static SuricataFileContext sfc = { &sbcfg };
...
+        goto end;
+    }
+    fcr.start = offset;
+    fcr.size = key->filesize;
+    if (offset + rlen > key->filesize) {
+        fcr.end = key->filesize;
+    } else {
+        fcr.end = offset + rlen;
+    }
+    // TODOsmbmulmti4 should we only rely on ippair and have nothing global ?
```
catenacyber, line 150: That is the big question for me...

src/ippair.c
```diff
@@ -773,6 +773,22 @@ static IPPair *IPPairGetUsedIPPair(void)
     return NULL;
 }
 
+int GetFlowAddresses(const Flow *f, Address *ip_src, Address *ip_dst)
```
catenacyber, line 776: Lost inlining because it is now used by 2 files...

suricata-qa:
ERROR:
 

ERROR: QA failed on smb_files_sha256.
  

Pipeline 9048

rust/src/smb/files.rs
```diff
@@ -57,6 +74,13 @@ pub fn filetracker_newchunk(ft: &mut FileTransferTracker, files: &mut FileContai
     }
 }
 
+// Defined in app-layer-htp-range.h
+extern "C" {
+    pub fn HttpRangeAppendData(
+        c: *mut HttpRangeContainerBlock, data: *const c_uchar, data_len: u32,
```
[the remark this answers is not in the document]
victorjulien: Think that makes sense.

rust/src/smb/files.rs
victorjulien: Not sure. A test case would be helpful.

rust/src/smb/smb.rs
```diff
@@ -593,8 +593,9 @@ impl SMBTransaction {
     }
 
     pub fn free(&mut self) {
-        debug_validate_bug_on!(self.tx_data.files_opened > 1);
-        debug_validate_bug_on!(self.tx_data.files_logged > 1);
+        // there can be a multi file + a regular one logged
```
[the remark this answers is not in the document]
victorjulien: I think you could have special logic for a tx that tracks a multi-stream file?

rust/src/smb/smb2.rs
victorjulien: Yes. We need a review of all the HashMaps in SMB and NFS state

src/app-layer-smb.c
victorjulien: I don't know for sure, but I would be quite surprised if there could just be a single multi stream file transfer per ip-pair. So this probably has to be a list.

Also wondering if such a list could be a Rust implementation. Guess it would need a void ptr in the key.

src/app-layer-smb.c
victorjulien: Not sure. What do the SMB protocol docs have to say about this?

src/ippair.c
victorjulien: Not critical I think

victorjulien:
Would like to see `smb: support multi-stream file transfers` commit message updated to explain the implementation choices, like using the ip-pair table, as well as file lifecycle, relationship with transactions, limits management, etc.

catenacyber:
Thanks for the review, will dive into it another day

rust/src/smb/files.rs
jasonish: > @jasonish can I keep the derive Default and have `file_range` being null ?

No, I do not believe so.

src/ippair.c
catenacyber: ok

src/app-layer-smb.c
catenacyber: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb/44c3cf8d-0931-4923-8fdc-738537ba70ba

>   The FileId MUST be unique for a file on a given object store.

So, look like it does not depend on the client, only the server...

But SMB2 is less clear 
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/f1d9b40d-e335-45fc-9d0b-199a31ede4c3

So, it is maybe best to not use ippair, and use a global hash table as for HTTP ranges, with the key having the server address and the file unique id...

src/app-layer-smb.c
catenacyber: Solved by dropping use of ippair and only using global hash table as for HTTP ranges (see below)

rust/src/smb/smb.rs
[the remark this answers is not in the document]
catenacyber: good idea, will do that

rust/src/smb/files.rs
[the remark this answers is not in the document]
catenacyber: will do

catenacyber:
> Would like to see `smb: support multi-stream file transfers` commit message updated to explain the implementation choices, like using the ip-pair table, as well as file lifecycle, relationship with transactions, limits management, etc.

Good idea, I would like to settle first on these implementation choices ;-)

victorjulien:
> > Would like to see `smb: support multi-stream file transfers` commit message updated to explain the implementation choices, like using the ip-pair table, as well as file lifecycle, relationship with transactions, limits management, etc.
> 
> Good idea, I would like to settle first on these implementation choices ;-)

As it will also help reviewers, I'd argue it needs to be part of draft PRs as well.

src/app-layer-smb.c
victorjulien: Might be good to look at other tools to see how they solve this. Zeek and Wireshark come to mind. Samba might also be interesting to look at.

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7863

PR #7458, opened 2022-06-01
security: prevents process creation

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5373

Describe changes:
- optionally calls `setrlimit(RLIMIT_NPROC, 0)` to prevent process creation by Suricata process

Modifies #7448 by excluding windows which does not have setrlimit

PR #7842, opened 2022-09-09
security: prevents process creation

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5373

Describe changes:
- optionally calls `setrlimit(RLIMIT_NPROC, 0)` to prevent process creation by Suricata process

Modifies #7458 by taking review comments into account

### DIFF
--- a/doc/userguide/upgrade.rst
+++ b/doc/userguide/upgrade.rst
@@ -37,6 +37,12 @@ Major changes
 ~~~~~~~~~~~~~
 - Upgrade of PCRE1 to PCRE2. See :ref:`pcre-update-v1-to-v2` for more details.
 
+Security changes
+~~~~~~~~~~~~~~~~
+- Process creation by Suricata is now disabled by default, this can be configured with `security.limit-noproc` (for LUA scripts for instance)
+  This is done through the call of `setrlimit(RLIMIT_NPROC` cf `man setrlimit`.
+  This prevents potential exploits against Suricata to fork a new process, even if it does not prevent the call of `exec`.
+
 Removals
 ~~~~~~~~
 - The libprelude output plugin has been removed.
--- a/configure.ac
+++ b/configure.ac
@@ -209,6 +209,7 @@
     AC_CHECK_FUNCS([gethostname inet_ntoa uname])
     AC_CHECK_FUNCS([gettimeofday clock_gettime utime strptime tzset localtime_r])
     AC_CHECK_FUNCS([socket setenv select putenv dup2 endgrent endpwent atexit munmap])
+    AC_CHECK_FUNCS([setrlimit])
 
     AC_CHECK_FUNCS([fwrite_unlocked])
 
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -30,6 +30,12 @@
 #if HAVE_SIGNAL_H
 #include <signal.h>
 #endif
+#ifndef OS_WIN32
+#ifdef HAVE_SYS_RESOURCE_H
+// setrlimit
+#include <sys/resource.h>
+#endif
+#endif
 
 #include "suricata.h"
 #include "decode.h"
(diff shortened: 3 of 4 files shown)


doc/userguide/upgrade.rst
```diff
@@ -37,6 +37,12 @@ Major changes
 ~~~~~~~~~~~~~
 - Upgrade of PCRE1 to PCRE2. See :ref:`pcre-update-v1-to-v2` for more details.
 
+Security changes
+~~~~~~~~~~~~~~~~
+- Process creation by Suricata is now disabled by default, this can be configured with `security.limit-noproc` (for LUA scripts for instance)
+  This is done through the call of `setrlimit(RLIMIT_NPROC` cf `man setrlimit`.
```
victorjulien, line 43: typo: missing `)`

doc/userguide/upgrade.rst
```diff
@@ -37,6 +37,12 @@ Major changes
 ~~~~~~~~~~~~~
```
victorjulien, line 37: would expect doc about this topic also outside of an 'upgrade' doc

doc/userguide/upgrade.rst
```diff
@@ -37,6 +37,12 @@ Major changes
...
 - Upgrade of PCRE1 to PCRE2. See :ref:`pcre-update-v1-to-v2` for more details.
 
+Security changes
+~~~~~~~~~~~~~~~~
+- Process creation by Suricata is now disabled by default, this can be configured with `security.limit-noproc` (for LUA scripts for instance)
```
victorjulien, line 42: the feature is disabled by default, but enabled in the new default config. Upgrade users would not get this unless they update their yaml

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | tlpw1_stats_chk | | |
| .tcp.rst | 103033 | 125983 | 122.27% |
| | tlpr1_stats_chk | | |
| .flow.memuse | 562615488 | 609251008 | 108.29% | 

Pipeline 9082

suricata-qa:

Pipeline 9098

doc/userguide/upgrade.rst
catenacyber: Well `setrlimit(RLIMIT_NPROC)` is incorrect because setrlimit has 2 arguments...

Will rephrase

doc/userguide/upgrade.rst
catenacyber: @jufajardini should we have a security.rst in the user guide ? 
Or a Config hardening ?

doc/userguide/upgrade.rst
catenacyber: What do we want ?
I would have started with disabled by default, but I remember you were thinking otherwise, right ?

doc/userguide/upgrade.rst
victorjulien: btw I think `cf` should also be expanded. I took me a while to get used to this acronym, so others may be confused as well.

doc/userguide/upgrade.rst
victorjulien: Yes :)

doc/userguide/upgrade.rst
victorjulien: I think the behavior makes sense, but the statement here is a bit incomplete. Upgrading users might expect the setting to be enabled, while they still need to make a config adjustment for it to be enabled.

doc/userguide/upgrade.rst
jufajardini: (TIL config hardening :o )
I think I prefer the idea of a config hardening chapter/section

doc/userguide/upgrade.rst
jufajardini: Would it help if we rephrased to something like:
"Process creation by Suricata is now disabled by default. This must be configured with `security.limit-noproc`  "
?

doc/userguide/upgrade.rst
catenacyber: Ok, I thought latin confer was universal ;-)

doc/userguide/upgrade.rst
catenacyber: Proposing something in next PR

doc/userguide/upgrade.rst
catenacyber: New version coming

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7858

PR #7384, opened 2022-05-05
flow: optionally use vlan tracking also for ICMPv4

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5330

Describe changes:
- Use configuration value of `vlan.use-for-tracking` also for ICMPv4 flows

That is use util function `CmpVlanIds` as is done in other cases

Replaces #7382 with commit reworded

PR #7859, opened 2022-09-14
flow: optionally use vlan tracking also for ICMPv4

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5330

Describe changes:
- Use configuration value of `vlan.use-for-tracking` also for ICMPv4 flows

That is use util function `CmpVlanIds` as is done in other cases

Replaces #7384 with rebase to get Victor's attention

libhtp-pr: 366

### DIFF
--- a/src/flow-hash.c
+++ b/src/flow-hash.c
@@ -374,28 +374,20 @@ static inline int FlowCompareICMPv4(Flow *f, const Packet *p)
         /* first check the direction of the flow, in other words, the client ->
          * server direction as it's most likely the ICMP error will be a
          * response to the clients traffic */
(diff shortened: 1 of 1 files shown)


victorjulien:
I find the commit message confusing. To me it indicates adding some optional behavior, but it seems that in fact it fixes the vlan behavior not being optional like in the other protocol cases.

So I would suspect some thing like:

```
flow/icmpv4: fix vlan.use-for-tracking

For ICMPv4 error messages the vlan id's were always considered, even if the 'vlan.use-for-tracking' option was
disabled.
```

catenacyber:
> I find the commit message confusing. To me it indicates adding some optional behavior, but it seems that in fact it fixes the vlan behavior not being optional like in the other protocol cases.
> 

I would say it fixes the vlan behavior so that it can be optional (as per the config option)

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7861

PR #7885, opened 2022-09-20
6.0.x: content: don't error out on incomplete hex - v1

jasonish:
Before 6.0.6 if hex content was incomplete, Suricata didn't error out.
With 6.0.6 incomplete hex was detected and errored on which is a
breaking change in a release branch.  Instead, only emit a warning
unless strict content checking has been requested.

To enable strict behaviour on incomplete content hex in a rule,
"--strict-rule-keywords=content" can be used on the command line.

Issue: https://redmine.openinfosecfoundation.org/issues/5546

For now there is no accompanying master commit, as the change in
behaviour for a major version is OK.

### DIFF
--- a/src/detect-content.c
+++ b/src/detect-content.c
@@ -23,6 +23,7 @@
  * Simple content match part of the detection engine.
  */
 
+#include "detect-engine-register.h"
 #include "suricata-common.h"
 #include "decode.h"
 #include "detect.h"
@@ -79,8 +80,8 @@ void DetectContentRegister (void)
  *  \retval -1 error
  *  \retval 0 ok
  */
-int DetectContentDataParse(const char *keyword, const char *contentstr,
-        uint8_t **pstr, uint16_t *plen)
+int DetectContentDataParse(
+        const char *keyword, const char *contentstr, uint8_t **pstr, uint16_t *plen, bool *warning)
 {
     char *str = NULL;
     size_t slen = 0;
@@ -112,9 +113,18 @@ int DetectContentDataParse(const char *keyword, const char *contentstr,
                 bin_count++;
                 if (bin) {
                     if (binpos > 0) {
-                        SCLogError(SC_ERR_INVALID_SIGNATURE,
-                                "Incomplete hex code in content - %s. Invalidating signature.",
+                        SCLogWarning(SC_ERR_INVALID_SIGNATURE,
+                                "Incomplete hex code in content - %s. Invalidating signature. "
+                                "This will become an error in 7.0.",
                                 contentstr);
+                        /* Prior to 6.0.6, incomplete hex was silently
+                           ignored. With 6.0.6 this turned into an
+                           error with -T. For 6.0.7, make the error
+                           non-fatal unless strict content parsing is
+                           enabled. */
+                        if (warning != NULL && !SigMatchStrictEnabled(DETECT_CONTENT)) {
+                            *warning = true;
+                        }
                         goto error;
                     }
                     bin = 0;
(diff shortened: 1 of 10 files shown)


src/detect-content.c
```diff
@@ -79,8 +80,8 @@ void DetectContentRegister (void)
  *  \retval -1 error
  *  \retval 0 ok
  */
-int DetectContentDataParse(const char *keyword, const char *contentstr,
-        uint8_t **pstr, uint16_t *plen)
+int DetectContentDataParse(
+        const char *keyword, const char *contentstr, uint8_t **pstr, uint16_t *plen, bool *warning)
```
jasonish, line 84: An output var just seemed more appropriate here than changing the return value.

src/detect-fileext.c
```diff
@@ -149,7 +149,7 @@ static DetectFileextData *DetectFileextParse (DetectEngineCtx *de_ctx, const cha
 
     memset(fileext, 0x00, sizeof(DetectFileextData));
 
-    if (DetectContentDataParse("fileext", str, &fileext->ext, &fileext->len) == -1) {
+    if (DetectContentDataParse("fileext", str, &fileext->ext, &fileext->len, NULL) == -1) {
```
jasonish, line 152: I guess these other users of content parsing should respect the same behaviour.

jasonish:
Another issue is that when not strict, you have no context what rule is causing the error.  The similar error for classtype has access to `de_ctc` to display more context, however that is usually gone a few calls before parsing the context.

src/detect-content.c
```diff
@@ -369,6 +381,8 @@ int DetectContentSetup(DetectEngineCtx *de_ctx, Signature *s, const char *conten
 
 error:
     DetectContentFree(de_ctx, cd);
+    if (warning)
+        return 0;
```
victorjulien, line 385: will this not load the rule, but minus the broken content condition?

so effectively turning
`alert ... content:"|aa a aa|"; dsize:10;`
into
`alert ... dsize:10;`
?

victorjulien: Looks like we should either return -2 or -3 here (not sure which though):
```c
    int ret = SigParse(de_ctx, sig, sigstr, dir, &parser);                                               
    if (ret == -3) {                                                                                     
        de_ctx->sigerror_silent = true;                                                                  
        de_ctx->sigerror_ok = true;                                                                      
        goto error;                                                                                      
    }                                                                                                    
    else if (ret == -2) {                                                                                
        de_ctx->sigerror_silent = true;                                                                  
        goto error;                                                                                      
    } else if (ret < 0) {                                                                                
        goto error;                                                                                      
    }                                                                                                    
```

src/detect-content.c
```diff
@@ -112,9 +113,18 @@ int DetectContentDataParse(const char *keyword, const char *contentstr,
                 bin_count++;
                 if (bin) {
                     if (binpos > 0) {
-                        SCLogError(SC_ERR_INVALID_SIGNATURE,
-                                "Incomplete hex code in content - %s. Invalidating signature.",
+                        SCLogWarning(SC_ERR_INVALID_SIGNATURE,
```
victorjulien, line 116: I think this should only be a warning if we're not in strict mode. If we are in strict mode it should be an error like it was.

src/detect-content.c
jasonish: Actually, its a -4...  New.. Log the error, but don't error out.  An option we don't have it.  Which handles the previous case as well.

ie)
```
        de_ctx->sigerror_ok = true;                                                                      
```

jasonish:
Updated at https://github.com/OISF/suricata/pull/7887.

PR #7887, opened 2022-09-20
6.0.x: content: don't error out on incomplete hex - v2

jasonish:
Before 6.0.6 if hex content was incomplete, Suricata didn't error out.
With 6.0.6 incomplete hex was detected and errored on which is a
breaking change in a release branch. Instead, only emit a warning
unless strict content checking has been requested.

To enable strict behaviour on incomplete content hex in a rule,
"--strict-rule-keywords=content" can be used on the command line.

Issue: https://redmine.openinfosecfoundation.org/issues/5546

For now there is no accompanying master commit, as the change in
behaviour for a major version is OK.

Previous PR: https://github.com/OISF/suricata/pull/7885

Changes from last PR:
- Better error handing.  In non-strict mode an error with context will
  be displayed, but not result in a test failure unless strict mode is
  requested.
- Log error if strict mode is enabled, not a warning.
- Add this warn/error checking to all users of content parsing.

suricata-verify-pr: 946

### DIFF
--- a/src/detect-content.c
+++ b/src/detect-content.c
@@ -23,6 +23,7 @@
  * Simple content match part of the detection engine.
  */
 
+#include "detect-engine-register.h"
 #include "suricata-common.h"
 #include "decode.h"
 #include "detect.h"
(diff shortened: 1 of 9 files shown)


suricata-qa:

Pipeline 9372

PR #7918, opened 2022-09-23
eve/schema: check that each array has at least one element

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5167

Describe changes:
- eve/schema : forbid empty arrays

Replaces https://github.com/OISF/suricata-verify/pull/897 (yes replacing S-V PR with suricata one)

### DIFF
--- a/etc/schema.json
+++ b/etc/schema.json
@@ -106,6 +106,7 @@
         },
         "files": {
             "type": "array",
+            "minItems": 1,
             "items": {
                 "type": "object",
                 "optional": true,
@@ -148,6 +149,7 @@
                     },
                     "sid": {
                         "type": "array",
+                        "minItems": 1,
                         "items": {
                             "type": "integer"
                         }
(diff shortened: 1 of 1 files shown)


catenacyber:
ping @jasonish ?

victorjulien:
This looks like it fails on `dns-lua-rules`, which has this record:
```json
{
  "query": [],
  "answer": {
    "version": 2,
    "type": "answer",
    "id": 36146,
    "flags": "81a0",
    "qr": true,
    "rd": true,
    "ra": true,
    "rrname": "www.suricata-ids.org",
    "rrtype": "A",
    "rcode": "NOERROR"
  }
}
```

catenacyber:
Nice catch, fix in https://github.com/OISF/suricata/pull/8231

catenacyber:
Why did not CI catch this ? is S-V run with lua support ?

jasonish:
> Why did not CI catch this ? is S-V run with lua support ?

On a quick look, Debian 9 with Luajit does an S-V test in GitHub CI.  But we should enable this more explicitly in some newer builds.

PR #7919, opened 2022-09-23
mqtt: limit maximum number of live transactions

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5430

Describe changes:
- mqtt: limit maximum number of live transactions

### DIFF
--- a/rust/src/mqtt/mqtt.rs
+++ b/rust/src/mqtt/mqtt.rs
@@ -37,6 +37,8 @@ const MQTT_CONNECT_PKT_ID: u32 = std::u32::MAX;
 // this value, it will be truncated. Default: 1MB.
 static mut MAX_MSG_LEN: u32 = 1048576;
 
+const MQTT_MAX_TX: usize = 4096;
+
 static mut ALPROTO_MQTT: AppProto = ALPROTO_UNKNOWN;
 
 #[derive(FromPrimitive, Debug)]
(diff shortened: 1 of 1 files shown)


catenacyber:
Another option would be to prevent a transaction to be pushed into the transactions list/Vec
But that would require changing the 31 occurrences of `self.transactions.push`

Another option would be to free directly the transactions instead of marking them as complete, but they would not get logged...

This is the best I have come up with... Not sure if it is good enough for 60x

jasonish:
> Another option would be to free directly the transactions instead of marking them as complete, but they would not get logged...

This is what we used to do with DNS, and I think its a fair emergency strategy along with a flooded alert which you have.

rust/src/mqtt/mqtt.rs
```diff
@@ -52,6 +54,7 @@ pub enum MQTTEvent {
     MissingMsgId,
     UnassignedMsgtype,
     MalformedTraffic,
+    FloodedFlushed,
```
jasonish, line 57: Should be `TooManyTransactions` as thats what we already have in master with more or less the same meaning.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .capture.kernel_drops | 37050 | 89591 | 241.81% |
| .ftp.memuse | 0 | 56 | - | 

Pipeline 9437

catenacyber:
> > Another option would be to free directly the transactions instead of marking them as complete, but they would not get logged...
> 
> This is what we used to do with DNS, and I think its a fair emergency strategy along with a flooded alert which you have.

You do not have the flooded alert, if you free the alert, because you do not have time to run detection...

Closing in favor of Jeff's alternative MR

PR #7920, opened 2022-09-23
Smb file multiflow 4861 v4.2

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4861

Describe changes:
- smb : handle multi-stream file transfers

Rebase of #7865 now that CI works, so that we get real bugs from it

This is a draft for feedback.

Questions :
- Should we handle reads as writes ?
- This draft is SMB2 only, would we want SMB1 ?
Style questions :
- Should we move `ContainerTHashTable ContainerSmbHt` to app-layer-smb.c ?
- Should app-layer-htp-range.h be renamed util-file-range.h ?
- Should we make `HTPFileCloseHandleRange` generic ?

_Blocker question_
Also, the current multi-stream logic will not log a file until it is complete.
Because there may be a new flow coming that will complete the file.
And because to log it, we have to move its ownership from the global hash table to the transaction which will end up freeing it...
Thoughts about this @victorjulien ?
I now think the ownership should not be moved back to a transaction and the logging happen on its own...

Could logging be called from `HttpRangeContainersTimeoutHash` ? which gets called by `FlowManager`
logging being `OutputFileLogFfc`

TODOs:
- There are other SMB tickets to create based on the comments
- Fix bugs that will be found by this CI run ;-p

PR #7928, opened 2022-09-26
Smb file multiflow 4861 v5

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4861

Describe changes:
- smb : handle multi-stream file transfers

Rebase of #7920 
- fuzz target calls `AppLayerParserParse` with `STREAM_EOF` before calling `FlowFree` (is it ok ?)
- doc improvements

This is a draft for feedback.

Questions :
- Should we handle reads as writes ?
- This draft is SMB2 only, would we want SMB1 ?
Style questions :
- Should we move `ContainerTHashTable ContainerSmbHt` to app-layer-smb.c ?
- Should app-layer-htp-range.h be renamed util-file-range.h ?
- Should we make `HTPFileCloseHandleRange` generic ?

_Blocker question_
Also, the current multi-stream logic will not log a file until it is complete.
Because there may be a new flow coming that will complete the file.
And because to log it, we have to move its ownership from the global hash table to the transaction which will end up freeing it...
Thoughts about this @victorjulien ?
I now think the ownership should not be moved back to a transaction and the logging happen on its own...

Could logging be called from `HttpRangeContainersTimeoutHash` ? which gets called by `FlowManager`
logging being `OutputFileLogFfc`

TODOs:
- There are other SMB tickets to create based on the comments
- get the answers to the questions above and act on them

### DIFF
--- a/doc/userguide/file-extraction/file-extraction.rst
+++ b/doc/userguide/file-extraction/file-extraction.rst
@@ -175,3 +175,18 @@ Updating Filestore Configuration
 .. toctree::
 
    config-update
+
+File extraction over multiple flows
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Protocols such as HTTP and SMB allow to transfer a file using multiple flows.
+For example in HTTP, this is done with the `Range` header in requests.
+
+Suricata can manage to recombine the parts of files seen in the multiple flows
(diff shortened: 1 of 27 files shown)


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7929

PR #7476, opened 2022-06-03
detect/alert: don't lose 'drop' alert - v2

jufajardini:
Previous PR: #7469 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5180

Describe changes:
- Use a more descriptive name for the flag that indicates the alert queue reallocation failure
- make PacketAlertSet more human-error proof
- make PacketAlertFinalize more compact

Still as draft due to the Feat 5319 commits

suricata-verify-pr: 842
https://github.com/OISF/suricata-verify/pull/842

PR #7969, opened 2022-10-04
detect/alert: don't lose 'drop' alert - v3

jufajardini:

This work is sort of a continuation of #7476 but with some smaller optimizations to functions in detect-engine-alert and detect-engine-threshold.

This is a draft because it is failing the SV test `exception-policy-default-01`, showing that the way I'm trying to post-process the Pass action isn't enough. I have tried a few things to try to fix/improve that, to no avail.

Describe changes:

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -177,9 +177,26 @@ static inline void RuleActionToFlow(const uint8_t action, Flow *f)
     }
 }
 
+/** \internal
+ */
+static inline PacketAlert PacketAlertSet(
+        DetectEngineThreadCtx *det_ctx, const Signature *s, uint64_t tx_id, uint8_t alert_flags)
+{
+    PacketAlert pa = { s->num, s->action, alert_flags, s, tx_id, 0 };
+    pa.num = s->num;
+    pa.action = s->action;
+    pa.s = (Signature *)s;
+    pa.flags = alert_flags;
+    /* Set tx_id if the frame has it */
+    pa.tx_id = (tx_id == UINT64_MAX) ? 0 : tx_id;
+    pa.frame_id = (alert_flags & PACKET_ALERT_FLAG_FRAME) ? det_ctx->frame_id : 0;
+    return pa;
+}
+
 /** \brief Apply action(s) and Set 'drop' sig info,
  *         if applicable */
-static void PacketApplySignatureActions(Packet *p, const Signature *s, const uint8_t alert_flags)
+static void PacketApplySignatureActions(
+        DetectEngineThreadCtx *det_ctx, Packet *p, const Signature *s, const uint8_t alert_flags)
 {
     SCLogDebug("packet %" PRIu64 " sid %u action %02x alert_flags %02x", p->pcap_cnt, s->id,
             s->action, alert_flags);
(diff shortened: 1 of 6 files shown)


src/detect-engine-threshold.h
```diff
@@ -30,6 +30,15 @@
...
 
+typedef enum SigThresholdResults_ {
+    THRESHOLD_DONT_ALERT = 0,
+    THRESHOLD_ALERT = 1,
+    THRESHOLD_SILENT_MATCH = 2,
+    THRESHOLD_SUPPRESSED = 0,
+    THRESHOLD_NOT_SUPPRESSED = 1,
+    THRESHOLD_SUPPRESS_NEED_ACTIONS = 2,
+} SigThresholdResults;
+
```
jufajardini, line 41: Rethink/decide on these variant names

src/detect-engine-threshold.c
```diff
@@ -317,18 +329,20 @@ static inline void RateFilterSetAction(Packet *p, PacketAlert *pa, uint8_t new_a
...
+ *
+ * \param lookup_tsh Current threshold entry
+ * \param td Threshold settings
+ * \param packet_time used to compare against previous detection and to set timeouts
+ *
+ * \retval bool true if threshold reached for this entry
+ * \retval bool false if threshold not reached
+ *
+ */
+static bool IsThresholdReached(
```
jufajardini, line 342: Changed this to bool as it seemed that it would be more evident

src/detect-engine-alert.c
```diff
@@ -338,73 +336,135 @@ static inline void FlowApplySignatureActions(
...
+            } else {
+                break;
+            }
+        }
+        if (PacketTestAction(p, ACTION_DROP)) {
+            if (p->alerts.drop.s->num > pass_sid_num) {
+                // if this should be a pass, then ignore drop action
+                p->alerts.drop.action = 0;
+                PacketPass(p);
+            }
```
jufajardini, line 461: This isn't enough. From my investigations, that's because both the packet and the flow, at this point, have already had the drop action set to them. 
I have tried to clear the action with a bitwise operation, but it's not enough, and triggers a bug in libhtp, so I am a bit blocked now.
But overall I am not sure if this is a good approach.

src/detect-engine-alert.c
```diff
@@ -338,73 +336,135 @@ static inline void FlowApplySignatureActions(
...
+ * \brief Check the threshold of the sigs that match, set actions, break on pass action
+ *        This function iterates the packet alerts array, removing those that didn't match
  *        the threshold, and those that match after a signature with the action "pass".
  *        The array is sorted by action priority/order
  * \param de_ctx detection engine context
  * \param det_ctx detection engine thread context
  * \param p pointer to the packet
  */
-void PacketAlertFinalize(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx, Packet *p)
+void PacketAlertQueueFinalize(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx, Packet *p)
```
jufajardini, line 394: I have noticed that we call PacketAlertQueueFinalize even when there are no alerts in the packet, so I'm thinking maybe we can optimize that, too.

src/detect-engine-alert.c
```diff
@@ -338,73 +336,135 @@ static inline void FlowApplySignatureActions(
...
 {
     SCEnter();
 
-    /* sort the alert queue before thresholding and appending to Packet */
-    qsort(det_ctx->alert_queue, det_ctx->alert_queue_size, sizeof(PacketAlert),
-            AlertQueueSortHelper);
-
```
jufajardini, line 355: It was suggested that we removed the qsort from here and called it after the alerts had been processed, but this leads to the current issue we're having with the sv test that is failing (exception-policy-default-01), that happens because the pass action is the last to be processed with the new approach, while in master it is processed first (so the flow and packet don't get the flag to be dropped set, and then things work fine).

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .app_layer.tx.ftp | 819 | 647 | 79.0% |
| .tcp.synack | 6726744 | 7541502 | 112.11% |
| .tcp.rst | 4910373 | 5330470 | 108.56% |
| .app_layer.error.ftp-data.gap | 1 | 0 | - |
| .app_layer.error.snmp.parser | 0 | 1 | - | 

Pipeline 9865

jufajardini:
Replaced by https://github.com/OISF/suricata/pull/7976, but comments that are here have not been addressed yet.

PR #7971, opened 2022-10-05
base64: add new mode as per RFC 4648

inashivb:
As per RFC 4648,
Implementations MUST reject the encoded data if it contains characters outside the base alphabet when interpreting base-encoded data, unless the specification referring to this document explicitly states otherwise.

Add a new mode `BASE64_MODE_RFC4648`, and handle input strictly as per the specification.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5223

Draft bc
1. Trying to figure out if `sid:3` should alert with the fix (it does with this patch)
2. Needs more testing, only been tested on a small dataset


victorjulien:
Can you add SV tests?

I would also like to see a code comment explaining the different modes with examples similar to how is done in the RFC 4648
image

This should probably be part of unittests as well.

PR #7973, opened 2022-10-05
rust: fix all clippy lints that could be done automatically - v3

jasonish:
A commit for clippy --fix run per lint.

I'm not sure about the following ones as they are purely style, but are considered best practice in the Rust community:
- needless returns
- assign op pattern
- let and return

In fact most of these are stylistic.  But I'm finding its become more and more normal for rust projects to be clippy clean by default, as well as rustfmt clean, but we'll get to that one later.


jasonish:
Hmm... No idea why CI isn't running.

PR #7976, opened 2022-10-05
detect/alert: don't lose 'drop' alert - v4

jufajardini:

This work is sort of a continuation of https://github.com/OISF/suricata/pull/7969 but with some smaller optimizations to functions in detect-engine-alert and detect-engine-threshold.

This is a draft because it is failing the SV test exception-policy-default-01, showing that the way I'm trying to post-process the Pass action isn't enough. I have tried a few things to try to fix/improve that, to no avail.

Changes from previous PR:
Trying to fix a bug that is being reported by the CIFuzz (https://github.com/OISF/suricata/actions/runs/3186046936/jobs/5196260225#step:5:468)

Describe changes:
- extract packet alerts processing to functions
- move packet alert queue sorting to after the alerts are processed

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -177,9 +177,26 @@ static inline void RuleActionToFlow(const uint8_t action, Flow *f)
     }
 }
 
+/** \internal
+ */
+static inline PacketAlert PacketAlertSet(
+        DetectEngineThreadCtx *det_ctx, const Signature *s, uint64_t tx_id, uint8_t alert_flags)
+{
+    PacketAlert pa = { s->num, s->action, alert_flags, s, tx_id, 0 };
+    pa.num = s->num;
+    pa.action = s->action;
+    pa.s = (Signature *)s;
+    pa.flags = alert_flags;
+    /* Set tx_id if the frame has it */
+    pa.tx_id = (tx_id == UINT64_MAX) ? 0 : tx_id;
+    pa.frame_id = (alert_flags & PACKET_ALERT_FLAG_FRAME) ? det_ctx->frame_id : 0;
+    return pa;
+}
+
 /** \brief Apply action(s) and Set 'drop' sig info,
  *         if applicable */
-static void PacketApplySignatureActions(Packet *p, const Signature *s, const uint8_t alert_flags)
+static void PacketApplySignatureActions(
+        DetectEngineThreadCtx *det_ctx, Packet *p, const Signature *s, const uint8_t alert_flags)
 {
     SCLogDebug("packet %" PRIu64 " sid %u action %02x alert_flags %02x", p->pcap_cnt, s->id,
             s->action, alert_flags);
@@ -190,9 +207,7 @@ static void PacketApplySignatureActions(Packet *p, const Signature *s, const uin
         PacketDrop(p, s->action, PKT_DROP_REASON_RULES);
 
         if (p->alerts.drop.action == 0) {
-            p->alerts.drop.num = s->num;
-            p->alerts.drop.action = s->action;
-            p->alerts.drop.s = (Signature *)s;
+            p->alerts.drop = PacketAlertSet(det_ctx, s, 0, alert_flags);
         }
         if ((p->flow != NULL) && (alert_flags & PACKET_ALERT_FLAG_APPLY_ACTION_TO_FLOW)) {
             RuleActionToFlow(s->action, p->flow);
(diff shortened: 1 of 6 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128100 | 104117 | 81.28% |
| .flow.spare | 2202676 | 1998034 | 90.71% |
| .flow.memuse | 557245568 | 630915328 | 113.22% | 

Pipeline 9898

src/detect-engine-alert.c
```diff
@@ -338,71 +336,136 @@ static inline void FlowApplySignatureActions(
...
-                    if (smd->is_last)
-                        break;
-                    smd++;
+    if (max_pos > 0) {
+        PacketAlert tmp_queue[max_pos];
+        uint16_t tmp_size = 0;
+        bool is_action_pass = false;
+        uint32_t pass_sid_num = 0;
+
+        while (i < max_pos) {
```
victorjulien, line 407: could this be a for loop? It seems we unconditionally increment i on every iteration. In a for loop it would also be easier to not use `if (res == 0) { suppress } else { no suppress }` but instead something like
```c
if (res == 0) {
   suppress
   continue;
}
no suppress
```

src/detect-engine-alert.c
```diff
@@ -338,71 +336,136 @@ static inline void FlowApplySignatureActions(
...
 {
     SCEnter();
 
-    /* sort the alert queue before thresholding and appending to Packet */
-    qsort(det_ctx->alert_queue, det_ctx->alert_queue_size, sizeof(PacketAlert),
-            AlertQueueSortHelper);
-
-    uint16_t i = 0;
+    int i = 0;
     uint16_t max_pos = det_ctx->alert_queue_size;
```
victorjulien, line 399: can be `const`?

src/detect-engine-alert.c
```diff
@@ -338,71 +336,136 @@ static inline void FlowApplySignatureActions(
...
-            SCLogDebug("Appending sid %" PRIu32 " alert to Packet::alerts at pos %u", s->id, i);
-
-            if (PacketTestAction(p, ACTION_PASS)) {
-                /* Ok, reset the alert cnt to end in the previous of pass
-                 * so we ignore the rest with less prio */
-                break;
+        if (is_action_pass) {
+            // this means we got a 'PASS' action for this Packet, pass_sid_num
+            // has the related signature's internal id
+            uint16_t j;
```
victorjulien, line 449: would prefer j to be initialized to 0 here and not in the for statement, as its used outside of the for loop as well

src/detect-engine-threshold.c
```diff
@@ -317,18 +329,20 @@ static inline void RateFilterSetAction(Packet *p, PacketAlert *pa, uint8_t new_a
...
+ *
+ * \retval bool true if threshold reached for this entry
+ * \retval bool false if threshold not reached
+ *
+ */
+static bool IsThresholdReached(
+        DetectThresholdEntry *lookup_tsh, const DetectThresholdData *td, struct timeval packet_time)
 {
-    int ret = 0;
+    int ret = false;
```
victorjulien, line 345: `bool`

src/detect-engine-threshold.h
```diff
@@ -30,6 +30,15 @@
 #include "ippair.h"
 #include "host-storage.h"
 
+typedef enum SigThresholdResults_ {
```
victorjulien, line 33: would be good to have comments for each value explaining what they mean

victorjulien (changes requested):
Not sure about the CI failures, needs checking.

Otherwise, see inline comments.

src/detect-engine-alert.c
```diff
@@ -338,71 +336,136 @@ static inline void FlowApplySignatureActions(
...
+
+        /* set actions on the flow */
+        FlowApplySignatureActions(p, pa, s, pa->flags);
+
+        /* set actions on packet */
+        PacketApplySignatureActions(det_ctx, p, s, pa->flags);
+    }
+
+    /* Thresholding removes this alert */
+    if (res == 0 || res == 2 || (s->flags & SIG_FLAG_NOALERT)) {
```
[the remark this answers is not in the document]
jufajardini: > Not sure about the CI failures, needs checking.
> 
> Otherwise, see inline comments.

Thanks for the review, sorry for submitting a not-so-thorough work. 

I'll submit a new one incorporating the feedback and adding the comments I from the previous PR where I'm stuck or have doubts.

src/detect-engine-alert.c
```diff
@@ -338,71 +336,136 @@ static inline void FlowApplySignatureActions(
...
+        while (i < max_pos) {
+            const Signature *s = de_ctx->sig_array[det_ctx->alert_queue[i].num];
+            uint8_t res = PacketAlertFinalize(de_ctx, det_ctx, p, &det_ctx->alert_queue[i], s);
+
+            /* Thresholding removes this alert */
+            if (res == 0) {
+                /* we will not copy this to the AlertQueue */
+                p->alerts.suppressed++;
+                SCLogDebug("Suppressing sid %" PRIu32 " alert from alerts' queue", s->id);
+            } else if (res == 1) {
```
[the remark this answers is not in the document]
jufajardini: From what I understand, the issue here is that if res is THRESHOLD_SILENT_MATCH, then we won't add that to the alert queue, that's why we need the comparison

jufajardini:
Feedback incorporated in: https://github.com/OISF/suricata/pull/7993

PR #7978, opened 2022-10-07
Base64 decode 5223/v2

inashivb:
As per RFC 4648,
Implementations MUST reject the encoded data if it contains characters outside the base alphabet when interpreting base-encoded data, unless the specification referring to this document explicitly states otherwise.

Add a new mode `BASE64_MODE_RFC4648`, and handle input strictly as per the specification.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5223

Previous PR: https://github.com/OISF/suricata/pull/7971

Changes since v1:
- Cleaned up existing tests
- Added test vectors for both RFCs
- Added code comments

suricata-verify-pr: 958


inashivb:
Seems like I did not build it with ASAN while testing. Sending new PR

PR #7993, opened 2022-10-12
detect/alert: don't lose 'drop' alert - v5

jufajardini:

This work is sort of a continuation of https://github.com/OISF/suricata/pull/7969 but with some smaller optimizations to functions in detect-engine-alert and detect-engine-threshold.

This is a draft because it is failing the SV test exception-policy-default-01, showing that the way I'm trying to post-process the Pass action isn't enough. I have tried a few things to try to fix/improve that, to no avail.

Previous PR: #7976

Changes from previous PR:
- improve code documentation
- simplify processing loop in `PacketAlertQueueFinalize`
- fix wrong type declarations in a few places
- reduce number of enum variants and comment each one
- clean up commit history a bit

Describe changes:
- extract packet alerts processing to functions
- move packet alert queue sorting to after the alerts are processed
- only process alerts queue if there are actual alerts for the packet

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -177,9 +177,20 @@ static inline void RuleActionToFlow(const uint8_t action, Flow *f)
     }
 }
 
+/** \internal
+ */
+static inline PacketAlert PacketAlertSet(
+        DetectEngineThreadCtx *det_ctx, const Signature *s, uint64_t tx_id, uint8_t alert_flags)
+{
+    PacketAlert pa = { s->num, s->action, alert_flags, s, (tx_id == UINT64_MAX) ? 0 : tx_id,
+        (alert_flags & PACKET_ALERT_FLAG_FRAME) ? det_ctx->frame_id : 0 };
+    return pa;
+}
+
 /** \brief Apply action(s) and Set 'drop' sig info,
  *         if applicable */
-static void PacketApplySignatureActions(Packet *p, const Signature *s, const uint8_t alert_flags)
+static void PacketApplySignatureActions(
+        DetectEngineThreadCtx *det_ctx, Packet *p, const Signature *s, const uint8_t alert_flags)
 {
     SCLogDebug("packet %" PRIu64 " sid %u action %02x alert_flags %02x", p->pcap_cnt, s->id,
             s->action, alert_flags);
@@ -190,9 +201,7 @@ static void PacketApplySignatureActions(Packet *p, const Signature *s, const uin
         PacketDrop(p, s->action, PKT_DROP_REASON_RULES);
 
         if (p->alerts.drop.action == 0) {
-            p->alerts.drop.num = s->num;
-            p->alerts.drop.action = s->action;
-            p->alerts.drop.s = (Signature *)s;
+            p->alerts.drop = PacketAlertSet(det_ctx, s, 0, alert_flags);
         }
         if ((p->flow != NULL) && (alert_flags & PACKET_ALERT_FLAG_APPLY_ACTION_TO_FLOW)) {
             RuleActionToFlow(s->action, p->flow);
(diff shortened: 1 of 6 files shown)


src/detect-engine-threshold.h
```diff
@@ -30,6 +30,14 @@
 #include "ippair.h"
 #include "host-storage.h"
 
+typedef enum SigThresholdResults_ {
+    THRESHOLD_DONT_ALERT = 0,     // Rule doesn't match
+    THRESHOLD_ALERT = 1,          // Rule match
+    THRESHOLD_SILENT_MATCH = 2,   // Rule match, alert is suppressed, rule actions apply
+    THRESHOLD_SUPPRESSED = 0,     // Alert suppressed due to thresholding
+    THRESHOLD_NOT_SUPPRESSED = 1, // Rule match, alert not suppressed
+} SigThresholdResults;
```
jufajardini, line 39: Adding comments has helped in combing through the variants, but not sure if now they're good enough.

src/detect-engine-alert.c
```diff
@@ -337,72 +336,141 @@ static inline void FlowApplySignatureActions(
...
+                    p->alerts.alerts[j] = tmp_queue[j];
+                } else {
+                    break;
+                }
+            }
+            if (PacketTestAction(p, ACTION_DROP)) {
+                if (p->alerts.drop.s->num > pass_sid_num) {
+                    // if this should be a pass, then ignore drop action
+                    p->alerts.drop.action = 0;
+                    PacketPass(p);
```
jufajardini, line 467: This isn't enough. From my investigations, that's because both the packet and the flow, at this point, have already had the drop action set to them.
I have tried to clear the action with a bitwise operation, but it's not enough, and triggers a bug in libhtp, so I am a bit blocked now.
But overall I am not sure if this is a good approach.

src/detect-engine-alert.c
```diff
@@ -337,72 +336,141 @@ static inline void FlowApplySignatureActions(
...
+    uint16_t max_pos = det_ctx->alert_queue_size;
 
-        /* Thresholding removes this alert */
-        if (res == 0 || res == 2 || (s->flags & SIG_FLAG_NOALERT)) {
-            /* we will not copy this to the AlertQueue */
-            p->alerts.suppressed++;
-        } else if (p->alerts.cnt < packet_alert_max) {
-            p->alerts.alerts[p->alerts.cnt] = det_ctx->alert_queue[i];
+    /* If there are alerts for this packet, process them */
+    if (max_pos > 0) {
```
jufajardini, line 406: I have noticed that we call PacketAlertQueueFinalize even when there are no alerts in the packet, so I'm thinking maybe we can optimize that, too.

Did that here, also because the CI Fuzz found a bug caused when max_pos is zero (no alerts for packet).

src/detect-engine-alert.c
```diff
@@ -337,72 +336,141 @@ static inline void FlowApplySignatureActions(
...
+                KEYWORD_PROFILING_END(det_ctx, smd->type, 1);
+                if (smd->is_last)
+                    break;
+                smd++;
+            }
+        }
 
-    /* sort the alert queue before thresholding and appending to Packet */
-    qsort(det_ctx->alert_queue, det_ctx->alert_queue_size, sizeof(PacketAlert),
-            AlertQueueSortHelper);
```
jufajardini, line 354: It was suggested that we removed the qsort from here and called it after the alerts had been processed, but this leads to the current issue we're having with the sv test that is failing (exception-policy-default-01), that happens because the pass action is the last to be processed with the new approach, while in master it is processed first (so the flow and packet don't get the flag to be dropped set, and then things work fine).

jufajardini (commented):
The CI failures from last PR are related to `exception-policy-default-01`. The checks fail there because the drop and alert are still going through, even after the pass.

I have a comment talking more about that on the PR.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128100 | 104040 | 81.22% |
| .flow.spare | 2202676 | 2003196 | 90.94% |
| .flow.memuse | 557245568 | 661795968 | 118.76% | 

Pipeline 9992

src/detect-engine-alert.c
victorjulien: Where has the drop already been set?

src/detect-engine-threshold.h
victorjulien: would like to see this split into 2 enums. One for the top 3 values, another for the bottom 2

src/detect-engine-alert.c
victorjulien: I think we need to do things in the correct order. If that requires sorting, so be it.

src/detect-engine-alert.c
victorjulien: Worth looking into, but lets do that after this work has been finalized to not intro more moving parts.

src/detect-engine-alert.c
jufajardini: in `AlertQueueAppend()` -> https://github.com/OISF/suricata/blob/master/src/detect-engine-alert.c#L278
doing that there to prevent losing the drop in case there's a failure when trying to append, or to prevent it not being set due to max alert-que size.

src/detect-engine-alert.c
jufajardini: roger!

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128100 | 104040 | 81.22% |
| .flow.spare | 2202676 | 2003196 | 90.94% |
| .flow.memuse | 557245568 | 661795968 | 118.76% | 

Pipeline 9992

jufajardini:
Closing this for now as it's stale since issues with higher priority got in the way. I'll come back to this in the near future.

PR #7999, opened 2022-10-13
loggers: share logger id's to allow for more loggers - v2

jasonish:
As we've hit the maximum number of transaction loggers for the current method
of tx logging, we either need to restructure or reduce the number of loggers.
It looks like all JSON tx loggers can share the same ID, which seems to be the
least intrusive approach at this time.

This will allow the bittorrent parser work, and other parsers to proceed.


src/suricata-common.h
```diff
@@ -445,38 +445,19 @@ typedef enum {
...
-    LOGGER_JSON_DNS,
-    LOGGER_JSON_HTTP,
-    LOGGER_JSON_SMTP,
-    LOGGER_JSON_TLS,
-    LOGGER_JSON_NFS,
-    LOGGER_JSON_TFTP,
-    LOGGER_JSON_FTP,
+    LOGGER_JSON_TX,
+
+    /* DNP3 loggers. These ones don't fit the common model of a
```
victorjulien, line 450: so how does it work? They use the tx logging api which expects this id logic to work?

jasonish: Actually not sure, I took a shot based on your suggestion and it passes all tests so far.  I should take a closer look. It doesn't work for DNP3 which relies on 2 loggers being registered for one protocol, where the current logic does fail.

victorjulien: could it be that dnp3 registers 2 loggers for the same tx?

victorjulien: one for both directions from what it looks like

jasonish: > could it be that dnp3 registers 2 loggers for the same tx?

Yes, that is what it does, to get around the fact that it correlates requests and responses into the same tx, but logs as if it was unidirectional.  We have better support/examples for this now, its just a non-trivial refactor of dnp3, but actively looking at it.

victorjulien: Wonder if this is in fact a valid use case that we should support?

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128100 | 103927 | 81.13% | 

Pipeline 10001

src/suricata-common.h
jasonish: > Wonder if this is in fact a valid use case that we should support?

While I can't think of any reason why we would need it right now, a quick way to do it would be: https://github.com/OISF/suricata/commit/6619ef501b5b322f79a96cc295fb863cb36d6cb2

But I think any protocol that would need this could also be done with the transaction per direction?

Some time refactoring DNP3 was not successful, but it should be in the longer term.  So it might just be best to leave these DNP3 IDs in there so its clear, and hopefully remove them in the future.

jasonish:
DNP3 is addressed here: https://github.com/OISF/suricata/pull/8006

PR #8006, opened 2022-10-14
eve: share logger id to allow for more loggers - v4

jasonish:
As we've hit the maximum number of transaction loggers for the current method
of tx logging, we either need to restructure or reduce the number of loggers.
It looks like all JSON tx loggers can share the same ID, which seems to be the
least intrusive approach at this time.

This will allow the bittorrent parser work, and other parsers to proceed.

Previous PR: https://github.com/OISF/suricata/pull/7999

Changes from last PR:
- Convert DNP3 into a unidirectional parser to allow for a single logger to be
  registered.


src/detect-dnp3.c
```diff
@@ -443,11 +438,10 @@ static int DetectDNP3ObjMatch(DetectEngineThreadCtx *det_ctx,
...
-    if (flags & STREAM_TOSERVER) {
-        objects = &tx->request_objects;
-    }
-    else if (flags & STREAM_TOCLIENT) {
-        objects = &tx->response_objects;
+    if (flags & STREAM_TOSERVER && tx->is_request) {
+        objects = &tx->objects;
+    } else if (flags & STREAM_TOCLIENT && !tx->is_request) {
+        objects = &tx->objects;
     }
```
jasonish, line 445: Its code like this that fails when `APP_LAYER_PARSER_OPT_UNIDIR_TXS` is set for the protocol.  When set the direction logic appears to be inversed. The detect function is never called with `STREAM_TOSERVER` for a request transaction, and vice-versa, as these detect functions are being called the TCP ACK.  Running in IPS mode makes the direction logic correct, which I'm still trying to figure out.  I believe the progress logic is correct at https://github.com/OISF/suricata/pull/8006/files#diff-e6b6ada9b3a17879876cc1c3a1db5f162e3f834cc731bf3eab4d02bde39ae424R1437

However, the parser, logging and detect seems to work find without setting the `APP_LAYER_PARSER_OPT_UNIDIR_TXS`, so it might not be needed.

src/app-layer-dnp3.c
```diff
@@ -1626,6 +1588,12 @@ void RegisterDNP3Parsers(void)
         AppLayerParserRegisterTxDataFunc(IPPROTO_TCP, ALPROTO_DNP3,
             DNP3GetTxData);
         AppLayerParserRegisterStateDataFunc(IPPROTO_TCP, ALPROTO_DNP3, DNP3GetStateData);
+#if 0
+        /* While this parser is now fully unidirectional. setting this
+         * flag breaks detection at this time. */
+        AppLayerParserRegisterOptionFlags(
+                IPPROTO_TCP, ALPROTO_DNP3, APP_LAYER_PARSER_OPT_UNIDIR_TXS);
+#endif
```
jasonish, line 1596: See https://github.com/OISF/suricata/pull/8006/files#r995980177

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128100 | 103980 | 81.17% | 

Pipeline 10007

victorjulien:
Merged in #8097, thanks!

I left out the DNP3 patch as I had no time yet to review that.

jasonish:
Rebased: https://github.com/OISF/suricata/pull/8101

PR #8014, opened 2022-10-15
c: Use WARN_UNUSED for ByteExtract* functions

hsadia538:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues/3658) ticket:

**Draft PR for discussion and feedback**

Describe changes:
- Adding WARN_UNUSED for all ByteExtract functions in `util-byte.h`
- Fix warnings generated in code due to the macro

### DIFF
--- a/src/app-layer-enip-common.c
+++ b/src/app-layer-enip-common.c
@@ -91,10 +91,12 @@ static int ENIPExtractUint32(uint32_t *res, const uint8_t *input, uint16_t *offs
         return 0;
     }
 
-    ByteExtractUint32(res, BYTE_LITTLE_ENDIAN, sizeof(uint32_t),
-            (const uint8_t *) (input + *offset));
-    *offset += sizeof(uint32_t);
-    return 1;
+    if(ByteExtractUint32(res, BYTE_LITTLE_ENDIAN, sizeof(uint32_t),
+            (const uint8_t *) (input + *offset)) != -1)
+    {
+        *offset += sizeof(uint32_t);
+        return 1;
+    }
 }
 
 /**
--- a/src/util-byte.h
+++ b/src/util-byte.h
@@ -94,7 +94,7 @@ void BytesToStringBuffer(const uint8_t *bytes, size_t nbytes, char *outstr, size
  * \return n Number of bytes extracted on success
  * \return -1 On error
  */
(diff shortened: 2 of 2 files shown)


src/app-layer-enip-common.c
```diff
@@ -91,10 +91,12 @@ static int ENIPExtractUint32(uint32_t *res, const uint8_t *input, uint16_t *offs
         return 0;
     }
 
-    ByteExtractUint32(res, BYTE_LITTLE_ENDIAN, sizeof(uint32_t),
-            (const uint8_t *) (input + *offset));
-    *offset += sizeof(uint32_t);
-    return 1;
+    if(ByteExtractUint32(res, BYTE_LITTLE_ENDIAN, sizeof(uint32_t),
```
hsadia538, line 94: The warnings are generated due to the output of this function being never actually used in the call.

The purpose of the caller function `ENIPExtractUint32` seems to get satisfied even without a return value being called and used from the `ByteExtractUint32` function.

The only thing that is making sense to me is adding an `if{}` like above and function returning success if bytes get extracted i.e `-1` is not returned.

Is this the right approach?

victorjulien: Doesn't look like it is. In other anomaly conditions, we return `0` (e.g. not enough input bytes), so if `ByteExtractUint32` fails we should probably do the same.

src/util-byte.h
```diff
@@ -464,7 +464,7 @@ int WARN_UNUSED StringParseI16RangeCheck(
  * \param len Number of bytes to extract (23 max or 0 for unbounded)
  * \param str String to extract from
  *
- * \return n Number of bytes extracted on success
+ * \return n Number of bytes extracted on su./configccess
```
victorjulien, line 467: accidental change by the looks of it

victorjulien (changes requested):

src/util-byte.h
hsadia538: I think so, I will resolve this in next PR

src/app-layer-enip-common.c
hsadia538: I did not quite understand I think. You are suggesting that here we should return 0 if `ByteExtractUint32 ` fails?

inashivb: Yes

hsadia538:
Followed by: #8022

PR #8019, opened 2022-10-16
rust: fix lint warnings about if same then else -v5

Rai0Catodic0:
Ticket: [4609](https://redmine.openinfosecfoundation.org/issues/4609)

Make sure these boxes are signed before submitting your Pull Request -- thank you.

[ :white_check_mark: ]  I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
    [  :white_check_mark: ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
 - [ ]  I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Previous PR: #8010 

Describe changes:
- Fixed redundant if condition

PR #8010, opened 2022-10-15
rust: fix lint warnings about if same then else -v4

Rai0Catodic0:
Ticket: [4609](https://redmine.openinfosecfoundation.org/issues/4609)

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [ :white_check_mark: ] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ :white_check_mark: ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Previous PR: #8009

Describe changes:
- Fix commit message
- convert SClogNotice to SCLogDebug

PR #8007, opened 2022-10-14
rust: fix more clippy lints with clippy --fix - v6

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/7975

Changes from previous PR:
- Fix some code behind feature flags.
- In GitHub-CI, run clippy with --all-features
- Fixes for needless returns and let and return have been reverted due to style preferences.

PR #8021, opened 2022-10-17
rust: fix lint warnings about if same then else -v6

Rai0Catodic0:
Ticket: [4609](https://redmine.openinfosecfoundation.org/issues/4609)

[ :white_check_mark: ]  I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
    [  :white_check_mark: ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
 - [ ]  I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Previous PR: #8019  

Describe changes:
- Squash 2 commits

### DIFF
--- a/rust/src/http2/range.rs
+++ b/rust/src/http2/range.rs
@@ -131,11 +131,11 @@ pub fn http2_range_open(
     tx: &mut HTTP2Transaction, v: &HTTPContentRange, flow: *const Flow,
     cfg: &'static SuricataFileContext, dir: Direction, data: &[u8],
 ) {
-    if v.end <= 0 || v.size <= 0 {
-        // skipped for incomplete range information
-        return;
-    } else if v.end == v.size - 1 && v.start == 0 {
+    if (v.end <= 0 || v.size <= 0)
+        // skipped for incomplete range information 
+    || (v.end == v.size - 1 && v.start == 0)
         // whole file in one range
+    {
         return;
     }
     let (_, flags) = tx.files.get(dir);
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -54,7 +54,6 @@
 #![allow(clippy::module_inception)]
 #![allow(clippy::needless_range_loop)]
 #![allow(clippy::enum_variant_names)]
-#![allow(clippy::if_same_then_else)]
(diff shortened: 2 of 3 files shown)


jufajardini:
Sorry to be jumping here later on the thread, but I've noticed that the CI is failing with many warnings for this same lint. I tried to look into previous PRs but couldn't find anything commenting that the others shouldn't be addressed. Am I missing something?

jufajardini:
Hey @Rai0Catodic0 so I checked things better, as Shivani mentioned here: https://github.com/OISF/suricata/pull/8010#issuecomment-1279670796
your PR should only be merged after #8007 gets merged, so the CI check for cargo clippy is updated and then we shouldn't the failures that are happening now.

In short: I don't think you have to fix the warnings that are showing in that CI check :)

rust/src/http2/range.rs
```diff
@@ -131,11 +131,11 @@ pub fn http2_range_open(
...
 ) {
-    if v.end <= 0 || v.size <= 0 {
-        // skipped for incomplete range information
-        return;
-    } else if v.end == v.size - 1 && v.start == 0 {
+    if (v.end <= 0 || v.size <= 0)
+        // skipped for incomplete range information 
+    || (v.end == v.size - 1 && v.start == 0)
         // whole file in one range
+    {
```
jasonish, line 138: There is nothing technically wrong with this, but after looking a little closer, I think there is a better way that makes this more redable.

Instead of a combined if/else statement, just conver to 2 if statements on their own.  It'll pass the lint and be more readable.

Rai0Catodic0: Do you mean like this: 
``` rust
if v.end <= 0 || v.size <= 0 {
       // skipped for incomplete range information 
       return;
 } if v.end == v.size - 1 && v.start == 0{
       // whole file in one range
       return:
}
```

jufajardini: Yes, but with a line break between each if :P

Rai0Catodic0: ok, Thanks :)

jufajardini:
Closing this in favour of newer versions.

PR #8038, opened 2022-10-19
runmodes: enable IPS in DPDK runmode v1

lukashino:
Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5580#change-24800) :ticket::

Describe changes:
- prototype changes of RunModeSetLiveCaptureAutoFp and RunModeSetLiveCaptureWorkers functions to move the IPS enable logic out of suricata.c file
- added support for enabling IPS support in DPDK mode

PR #8041, opened 2022-10-19
runmodes: enable IPS in DPDK runmode v2

lukashino:
Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5580#change-24800) :ticket::

Follow up of #8038 

Describe changes:
- refactor

### DIFF
--- a/src/runmode-dpdk.c
+++ b/src/runmode-dpdk.c
@@ -1412,8 +1412,8 @@ int RunModeIdsDpdkWorkers(void)
     TimeModeSetLive();
 
     InitEal();
-    ret = RunModeSetLiveCaptureWorkers(ParseDpdkConfigAndConfigureDevice, DPDKConfigGetThreadsCount,
-            "ReceiveDPDK", "DecodeDPDK", thread_name_workers, NULL);
+    ret = RunModeSetLiveCaptureWorkers(ParseDpdkConfigAndConfigureDevice, NULL,
+            DPDKConfigGetThreadsCount, "ReceiveDPDK", "DecodeDPDK", thread_name_workers, NULL);
     if (ret != 0) {
         FatalError(SC_ERR_FATAL, "Unable to start runmode");
     }
--- a/src/runmode-dpdk.c
+++ b/src/runmode-dpdk.c
@@ -45,6 +45,8 @@
 #include "util-dpdk-ice.h"
 #include "util-dpdk-ixgbe.h"
 #include "util-time.h"
+#include "util-conf.h"
+#include "suricata.h"
 
 #ifdef HAVE_DPDK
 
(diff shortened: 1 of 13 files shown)


src/runmode-dpdk.c
```diff
@@ -1335,6 +1337,85 @@ static int DeviceConfigure(DPDKIfaceConfig *iconf)
...
+            } else {
+                has_ids = true;
+            }
+        } else {
+            has_ids = true;
+        }
+    }
+
+    if (has_ids && has_ips) {
+        for (int ldev = 0; ldev < nlive; ldev++) {
```
victorjulien, line 1382: is there a point to this loop if we'll unconditionally `FatalError` right after it?

lukashino: I thought we can let the user know which interfaces are set to the IDS/TAP mode before we exit Suri

victorjulien: could this be done during the initial parsing instead of during an additional loop?

lukashino: I think unless the user would specify the overall mode outright (e.g. a command line parameter or a global config entry), we need to do a two-way pass to determine what is the overall runmode of Suricata. If the first interface would be ids and then the other two interfaces would be paired in IPS mode then you would not be able to warn the user about the first interface in IDS mode. 
Minor optimization:
I could break from the first loop as soon as I would encounter both runmodes set at the same time (`has_ips && has_ids` condition) but I thought that would unnecessarily clutter the code and "performance" is not an issue here.

victorjulien: Can't we do something like this?
```diff
diff --git a/src/runmode-dpdk.c b/src/runmode-dpdk.c
index a0d1d2871d..9afd188c99 100644
--- a/src/runmode-dpdk.c
+++ b/src/runmode-dpdk.c
@@ -577,6 +577,9 @@ static int ConfigSetCopyIface(DPDKIfaceConfig *iconf, const char *entry_str)
     SCReturnInt(0);
 }

+static DpdkCopyModeEnum g_copy_mode = DPDK_COPY_MODE_NONE;
+static bool g_copy_mode_set = false;
+
 static int ConfigSetCopyMode(DPDKIfaceConfig *iconf, const char *entry_str)
 {
     SCEnter();
@@ -604,6 +607,15 @@ static int ConfigSetCopyMode(DPDKIfaceConfig *iconf, const char *entry_str)
         iconf->copy_mode = DPDK_COPY_MODE_IPS;
     }

+    if (!g_copy_mode_set) {
+        g_copy_mode_set = true;
+        g_copy_mode = iconf->copy_mode;
+    } else {
+        if (g_copy_mode != iconf->copy_mode) {
+            FatalError(SC_ERR_DPDK_CONF, "can't mix copy modes");
+        }
+    }
+
     SCReturnInt(0);
 }
```

lukashino: Possibly yes - would maybe keep it in the "enable IPS function API" to follow the same convention as AFP or Netmap. 

I wrote the original PR so that **all** erroneous interfaces are printed out in a single Suricata run. This approach exits after the first error.

lukashino: I think printing out all incorrect interfaces would be more user-friendly although the logic is a bit more complex. 
But if you want to go with this solution I'll redo it.

victorjulien: For DPDK I prefer a harsh approach, to keep complexity low and behavior clear. The AFP/Netmap changes are fine, we can do further cleanups there later.

lukashino:
Continues in #8057

PR #8049, opened 2022-10-20
detect-replace: Convert unittests to FAIL/PASS API

Rai0Catodic0:
Ticket: [4054](https://redmine.openinfosecfoundation.org/issues/4054)

- [:white_check_mark:] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [:white_check_mark:  ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Previous PR: #8046 

Describe changes:
- Fix commit message 
- Fix wrong use of macro

PR #8054, opened 2022-10-21
detect-replace: Convert unittests to FAIL/PASS API -v4

Rai0Catodic0:
Ticket: [4054](https://redmine.openinfosecfoundation.org/issues/4054)

- [ :white_check_mark: ] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ :white_check_mark: ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/

Previous PR:  #8049 

Describe changes:
-  Started implementing changes on int DetectReplaceLongPatternMatchTest
- Fix code style errors

### DIFF
--- a/src/detect-replace.c
+++ b/src/detect-replace.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2011-2014 Open Information Security Foundation
+/* Copyright (C) 2011-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -276,15 +276,11 @@ int DetectReplaceLongPatternMatchTest(uint8_t *raw_eth_pkt, uint16_t pktsize,
     DecodeEthernet(&th_v, &dtv, p, GET_PKT_DATA(p), pktsize);
 
     DetectEngineCtx *de_ctx = DetectEngineCtxInit();
-    if (de_ctx == NULL) {
-        goto end;
-    }
+    FAIL_IF_NULL(de_ctx);
     de_ctx->flags |= DE_QUIET;
 
     de_ctx->sig_list = SigInit(de_ctx, sig);
-    if (de_ctx->sig_list == NULL) {
-        goto end;
-    }
+    FAIL_IF_NULL(DetectEngineAppendSig(de_ctx, sig));
     de_ctx->sig_list->next = NULL;
 
     if (de_ctx->sig_list->sm_lists_tail[DETECT_SM_LIST_PMATCH]->type == DETECT_CONTENT) {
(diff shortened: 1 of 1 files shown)


src/detect-replace.c
```diff
@@ -276,15 +276,11 @@ int DetectReplaceLongPatternMatchTest(uint8_t *raw_eth_pkt, uint16_t pktsize,
     DecodeEthernet(&th_v, &dtv, p, GET_PKT_DATA(p), pktsize);
```
Rai0Catodic0, line 277: a few if statements in the test have printfs and debug prints. I'm not sure how I should deal with then, any suggestions

src/detect-replace.c
```diff
@@ -276,15 +276,11 @@ int DetectReplaceLongPatternMatchTest(uint8_t *raw_eth_pkt, uint16_t pktsize,
...
-        goto end;
-    }
+    FAIL_IF_NULL(de_ctx);
     de_ctx->flags |= DE_QUIET;
 
     de_ctx->sig_list = SigInit(de_ctx, sig);
-    if (de_ctx->sig_list == NULL) {
-        goto end;
-    }
+    FAIL_IF_NULL(DetectEngineAppendSig(de_ctx, sig));
```
Rai0Catodic0, line 283: Is it right to use the DetectEngineAppendSig function here?

src/detect-replace.c
```diff
@@ -276,15 +276,11 @@ int DetectReplaceLongPatternMatchTest(uint8_t *raw_eth_pkt, uint16_t pktsize,
...
     de_ctx->sig_list->next = NULL;
 
     if (de_ctx->sig_list->sm_lists_tail[DETECT_SM_LIST_PMATCH]->type == DETECT_CONTENT) {
```
Rai0Catodic0, line 286: this if is an example of what I asked earlier.

Rai0Catodic0:
Hi Reviewer. there are a few parts of the code that I wasn't able to write a comment on because I didn't make any changes on them and I didn't know that GitHub wouldn't allow me to comment on them, sorry about that. 
I asked if a should use the DetectEngineAppendSig function and that brought me to another question. 
at the end of the test function, there are a few functions that I believe are related to cleaning up resources after the test. 
with the addition of the FAIL/PASS macros, what should be the approach to that part of the code? 

Thanks for the review and sorry for the big text :)

src/detect-replace.c
jufajardini: It's ok to leave them there, thanks for asking :)

src/detect-replace.c
jufajardini: From my understanding, it would be, yes.

src/detect-replace.c
jufajardini: I'm rethinking what I had asked, sorry again about that. Let's leave this function untouched. There are many places where the if statements are not meant to mean to be an instant fail, but rather sort of debug what is happening along the way.

I have still answered your other questions, so you know what would be the approach, if you were to proceed with this patch :)

jufajardini (changes requested):
Thanks for this work :)

Thinking again about the test you were converting, I don't think it's a good idea to convert it, after all.

Therefore, let's just remove the changes you've started to make on it, and then, from my end, this contribution would be ready to go.

Thanks for your patience ^^

src/detect-replace.c
victorjulien: `DetectEngineAppendSig` is preferred, but needs to be used instead of `SigInit`

victorjulien (changes requested):

Rai0Catodic0:
Closing in favour of a newer PR: #8078

PR #8074, opened 2022-10-24
Simple errors/v7

victorjulien:
For discussion. Implement a errno style error scheme which much fewer codes, and reduction of number of SCLogError statements.

See for example https://github.com/OISF/suricata/commit/e14be6628d061f956a8966e7075dac21bfb4a560 for this could look like.

https://redmine.openinfosecfoundation.org/issues/3160

CI expected to fail due to check that looks for a removed code.


jasonish:
I like the use of `sc_errno` as a means of error propagation as used in C, and if we are to use it, I'm in favour of a limited subset of what we have now instead of the constant addition of new entries.

What I'm not convinced of is keeping the error code in the log message.  Typically errno in a C program is used for internal error propagation, not for reporting to the outside world.  That is typically done via `strerror`, `perror`, etc.  Sure these may include the error code in them, (ie: `Failure when trying to get MTU via ioctl for 'eno1': No such device (19)`) but usually doesn't carry much meaning I don't think.

Maybe there is value there and I'm just not seeing it?  I've already discussed this a little here: https://redmine.openinfosecfoundation.org/issues/3160#note-3

For example, 
```
19/11/2022 -- 17:06:17 - <Warning> - [ERRCODE: SC_ERR_SYSCALL(50)] - Failure when trying to get MTU via ioctl for 'eno1': No such device (19)
```
vs
```
19/11/2022 -- 17:06:17 - <Warning> - Failure when trying to get MTU via ioctl for 'eno1': No such device (19)
```

jufajardini:
> I like the use of `sc_errno` as a means of error propagation as used in C, and if we are to use it, I'm in favor of a limited subset of what we have now instead of the constant addition of new entries.
> 
> What I'm not convinced of is keeping the error code in the log message. Typically errno in a C program is used for internal error propagation, not for reporting to the outside world. That is typically done via `strerror`, `perror`, etc. Sure these may include the error code in them, (ie: `Failure when trying to get MTU via ioctl for 'eno1': No such device (19)`) but usually don't carry much meaning I don't think.
> 
> Maybe there is value there and I'm just not seeing it? I've already discussed this a little here: https://redmine.openinfosecfoundation.org/issues/3160#note-3
> 
> For example,
> 
> ```
> 19/11/2022 -- 17:06:17 - <Warning> - [ERRCODE: SC_ERR_SYSCALL(50)] - Failure when trying to get MTU via ioctl for 'eno1': No such device (19)
> ```
> 
> vs
> 
> ```
> 19/11/2022 -- 17:06:17 - <Warning> - Failure when trying to get MTU via ioctl for 'eno1': No such device (19)
> ```
> 
> I think I'm indifferent to it being there, but if we're breaking the numbers and code names, this would be the time to consider this as well.

(two cents) Right now, the second example (without the error code) looks more legible to me. I think that error codes make sense to be present if they could be used by someone trying to find more information, but if the error message already says enough, it probably just makes the message more cluttered...

victorjulien:
Continued in #8224

PR #8099, opened 2022-10-27
github-ci: non-root builder - v1

jasonish:
All the GitHub CI jobs run as root inside a container. This means the
testing is done in a different environment than a developer typically
uses, running as a user.

Add a job that does the build as a non-root user.

### DIFF
--- a/.github/workflows/scripts/fedora-non-root.sh
+++ b/.github/workflows/scripts/fedora-non-root.sh
@@ -1,0 +1,47 @@
+#! /usr/bin/env bash
+#
+# Helper script for Fedora build as a non-root user.
+#
+# We break the build up itno parts that need to be called individually
+# to avoid outputting too much data in a single step so we can see the
+# output in the UI.
+
+set -e
+set -x
+
+export PATH="$HOME/.cargo/bin:$PATH"
+
+case "$1" in
+    cbindgen)
+        # Setup cbindgen.
+        mkdir -p $HOME/.cargo/bin
+        cp prep/cbindgen $HOME/.cargo/bin
+        chmod 755 $HOME/.cargo/bin/cbindgen
+        ;;
+    autogen)
+        ./autogen.sh
+        ;;
+    configure)
+        ac_cv_func_realloc_0_nonnull="yes" \
+            ac_cv_func_malloc_0_nonnull="yes" \
+            LDFLAGS="-fsanitize=address" \
+            CC="clang" \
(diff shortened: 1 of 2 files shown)


jasonish:
The CI failure is showing a current problem with git master under these conditions:
- building with asan
- running as a non root user
- using a `suricata.yaml` generated as of commit a003640ecf190b02ab29f90c266d8b2a77dcbb78 or newer.

With the up to date configuration file, `security.limit-noproc` is set to true, however the leak sanitizer appears require the ability to create a thread when Suricata exits to do its job.

As `security.limit-noproc` has no effect when running as root, it is no seen in the other ci jobs which use the github ci default docker user, which is root.

Work-around: delete the `security.limit-noproc` line in `suricata.yaml` or set to `no`.

Related ticket: https://redmine.openinfosecfoundation.org/issues/5621

.github/workflows/scripts/fedora-non-root.sh
```diff
@@ -0,0 +1,47 @@
+#! /usr/bin/env bash
+#
+# Helper script for Fedora build as a non-root user.
+#
+# We break the build up itno parts that need to be called individually
```
jlucovsky, line 5: nit: s/itno/into

jlucovsky (commented):
Have we considered pre-building images and using the github container registry to reduce CI time?

E.g., create an image based on `fedora:36` but with the `dnf` packages.

jasonish:
> Have we considered pre-building images and using the github container registry to reduce CI time?
> 
> E.g., create an image based on `fedora:36` but with the `dnf` packages.

I've looked before but I don't see any easy way.  I think it would first involve having another project that pre-builds the containers and pushes them.  Then the Suricata actions depends on those? Could run into sync issues when system dependencies change?  Or is it simpler than this?

jasonish:
Replaced by https://github.com/OISF/suricata/pull/8179.

jasonish:
> > Have we considered pre-building images and using the github container registry to reduce CI time?
> > E.g., create an image based on `fedora:36` but with the `dnf` packages.
> 
> I've looked before but I don't see any easy way. I think it would first involve having another project that pre-builds the containers and pushes them. Then the Suricata actions depends on those? Could run into sync issues when system dependencies change? Or is it simpler than this?

For a Fedora 36 build I played with caching the downloaded RPMs using the github actions cache module.  It lowered the `dnf install` from 2m35s to 22s.  Caching containers could save that extra 22s but at the expense of higher maintenance complexity I think.

I previously didn't cache the downloaded RPMs (or .deb) as older versions of the cache action didn't support files outside of the working directory, but it does now.  So I'll start caching these as we update jobs.

PR #8133, opened 2022-11-01
Smtp server detection 1125 v7

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/1125

Describe changes:
- smtp : adds server side detection
- ftp : adds server side detection

The most special trick is that the (server) probing parser waits for the client side to have seen some data to take a definitive positive decision.
So that If it looks like a SMTP server (it could be a FTP server), let's see if the client looks like SMTP or FTP or something unknown...

Modifies #8119 by reusing many1

@ct0br0 on your provided pcap with this PR, `jq 'select(.event_type=="flow") | .app_proto' log/eve.json` only gives me smtp...

PR #8135, opened 2022-11-01
Smtp server detection 1125 v8

catenacyber:

Describe changes:
- smtp : adds server side detection
- ftp : adds server side detection
- protocoldetect: use stricter max depth

The most special trick is that the (server) probing parser waits for the client side to have seen some data to take a definitive positive decision.
So that If it looks like a SMTP server (it could be a FTP server), let's see if the client looks like SMTP or FTP or something unknown...

Modifies #8133 by adding last commit to restrict probing parser max depth to the probing parsers with a matching pattern

Before that last commit, S-V test `protocol-change-failed` was failing, because
- Suricata recognized SMTP the same
- Suricata did not start parsing server-side until it had pm_ctx->pp_max_len = 36 bytes, and thus when it started parsing, client and server were out-of-sync...

### DIFF
--- a/rust/src/util.rs
+++ b/rust/src/util.rs
@@ -18,7 +18,103 @@
 use std::ffi::CStr;
 use std::os::raw::c_char;
 
+use nom7::bytes::complete::take_while1;
+use nom7::character::complete::char;
+use nom7::character::{is_alphabetic, is_alphanumeric};
+use nom7::combinator::verify;
+use nom7::multi::many1_count;
+use nom7::IResult;
+
 #[no_mangle]
 pub unsafe extern "C" fn rs_check_utf8(val: *const c_char) -> bool {
     CStr::from_ptr(val).to_str().is_ok()
+}
+
+fn is_alphanumeric_or_hyphen(chr: u8) -> bool {
+    return is_alphanumeric(chr) || chr == b'-';
+}
+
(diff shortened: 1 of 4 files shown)


catenacyber:
@ct0br0 do I understand correctly that QA says that there are no changes for SMTP, but a big drop for FTP ?

ct0br0:
The file-checksum list has SMTP files missing, but other tests appear to hav FTP counts lower and higher error counts.

catenacyber:
For TREX_GENERIC_stats_chk, how is it possible to have `.app_layer.flow.ftp = 0` and `.app_layer.tx.ftp == 55103` ?

catenacyber:
So, I understand the SMTP deviation

With the PR, if the server banner does not match the probing parser (because of not using a valid domain name), so when client-side SMTP is recognized, SMTP server banner is not parsed first, and is thus never parsed as such because of `state->parser_state |= SMTP_PARSER_STATE_FIRST_REPLY_SEEN;` only happening if `(state->cmds_idx == state->cmds_cnt)`

Solution is to relax probing parser in the case `f->alproto_ts == ALPROTO_SMTP`

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8141

PR #8147, opened 2022-11-03
rust: fix warnings on rustc 1.67.0-nightly

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None 

Describe changes:
- rust: fix warnings on rustc 1.67.0-nightly

PR #8156, opened 2022-11-06
rust: explicitly convert from Direct to u8 , otherwise when add serde_…

jiamo:
We are writing custom app protocol in rust. When we need add serde_json package. It always got error such like 
>note: multiple `impl`s satisfying `u8: PartialEq<_>` found in the following crates: `core`, `serde_json`:
And we think it is good do the conversion explicitly. Follow the suggestion from https://github.com/rust-lang/rust/issues/103968. 

Describe changes:
- explicitly comvert Direction to u8

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -1258,10 +1258,10 @@ pub unsafe extern "C" fn rs_dcerpc_get_tx_cnt(vtx: *mut std::os::raw::c_void) ->
 pub unsafe extern "C" fn rs_dcerpc_get_alstate_progress(tx: *mut std::os::raw::c_void, direction: u8
                                                  )-> std::os::raw::c_int {
     let tx = cast_pointer!(tx, DCERPCTransaction);
-    if direction == Direction::ToServer.into() && tx.req_done {
+    if direction == Direction::ToServer as u8 && tx.req_done {
         SCLogDebug!("tx {} TOSERVER progress 1 => {:?}", tx.call_id, tx);
         return 1;
-    } else if direction == Direction::ToClient.into() && tx.resp_done {
+    } else if direction == Direction::ToClient as u8 && tx.resp_done {
         SCLogDebug!("tx {} TOCLIENT progress 1 => {:?}", tx.call_id, tx);
         return 1;
     }
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -42,7 +42,7 @@ pub const DIR_BOTH:        u8 = 0b0000_1100;
 const DIR_TOSERVER:        u8 = 0b0000_0100;
 const DIR_TOCLIENT:        u8 = 0b0000_1000;
 
-#[repr(C)]
+#[repr(u8)]
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum Direction {
     ToServer = 0x04,
--- a/rust/src/dns/dns.rs
+++ b/rust/src/dns/dns.rs
@@ -950,7 +950,7 @@ pub unsafe extern "C" fn rs_dns_probe_tcp(
         } else {
             Direction::ToClient
         };
-        if (direction & DIR_BOTH) != dir.into() {
+        if (direction & DIR_BOTH) != dir as u8 {
             *rdir = dir as u8;
         }
         return ALPROTO_DNS;
--- a/rust/src/http2/detect.rs
+++ b/rust/src/http2/detect.rs
@@ -681,7 +681,7 @@ pub unsafe extern "C" fn rs_http2_tx_get_status(
 pub unsafe extern "C" fn rs_http2_tx_get_cookie(
     tx: &mut HTTP2Transaction, direction: u8, buffer: *mut *const u8, buffer_len: *mut u32,
 ) -> u8 {
-    if direction == Direction::ToServer.into() {
+    if direction == Direction::ToServer as u8 {
         if let Ok(value) = http2_frames_get_header_value(tx, Direction::ToServer, "cookie") {
             *buffer = value.as_ptr(); //unsafe
             *buffer_len = value.len() as u32;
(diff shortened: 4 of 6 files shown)


victorjulien:
Minor request: please update the commit message to have a short subject line followed by a new line, then the body with the explanation. See further https://github.com/OISF/suricata/blob/6d3140bc012b093a361506b5c4f21cc9ae8878e8/doc/userguide/devguide/codebase/contributing/code-submission-process.rst#commits

@jasonish can you check the code?

jiamo:
> Minor request: please update the commit message to have a short subject line followed by a new line, then the body with the explanation. See further https://github.com/OISF/suricata/blob/6d3140bc012b093a361506b5c4f21cc9ae8878e8/doc/userguide/devguide/codebase/contributing/code-submission-process.rst#commits
> 
> @jasonish can you check the code?

Thanks. I rewrite the message. By the way `cargo clippy --all-features --fix` in the work flow failed. I think it is not caused by my commit.

catenacyber:
> Thanks. I rewrite the message. By the way `cargo clippy --all-features --fix` in the work flow failed. I think it is not caused by my commit.

Indeed, cf https://github.com/OISF/suricata/pull/8147 for the Rust check failure

alexsunrise:
> Minor request: please update the commit message to have a short subject line followed by a new line, then the body with the explanation. See further https://github.com/OISF/suricata/blob/6d3140bc012b093a361506b5c4f21cc9ae8878e8/doc/userguide/devguide/codebase/contributing/code-submission-process.rst#commits
> 
> @jasonish can you check the code?

better to avoid to use `.into()` when your members to CR, personal suggestion.

catenacyber:
> when your members to CR

What does this mean ?

alexsunrise:
> > when your members to CR
> 
> What does this mean ?

I mean when your guys to code review, it should be avoid using some code like `.into()`, like `if direction == Direction::ToServer.into()` change to `if direction == Direction::ToServer as u8` or `if direction == u8::from(Direction::ToServe)`. I think it will be better.

victorjulien:
> > > when your members to CR
> > 
> > 
> > What does this mean ?
> 
> I mean when your guys to code review, it should be avoid using some code like `.into()`, like `if direction == Direction::ToServer.into()` change to `if direction == Direction::ToServer as u8` or `if direction == u8::from(Direction::ToServe)`. I think it will be better.

The `as u8` is a pattern I believe clippy warns against.

Can you explain why you believe one is better than the other?

jiamo:
This pull request come from I am implementing some custom protocol in rust and need json functions. 
I add serde_json for the purpose. see https://github.com/serde-rs/json/issues/948
I got error because serde_json add `PartialEq` implement for `u8` which cause the error.
So I think we can just explicitly do convert process. The benefit only exists the suricata rust code don't want to conflict another package.   

> Can you explain why you believe one is better than the other?

jasonish:
The code seems find, but I'm not happy with the "look".  I've replicated the issues by adding serde and a simple serializable struct.  So I still want to look further into this, but it is something we need to resolve. I see serde entering our code at some point or another.

jasonish (changes requested):
Rebase, and use as u8.  That should be clippy clean and look nicer as well.

catenacyber:
> Rebase, and use as u8. That should be clippy clean and look nicer as well.

I think the downside of `as u8` is that this is an explicit cast that is losing precision.
So if by mistake we do a u16 as u8 when we should not, there will be no warning, and it can get nasty to debug...

jiamo:
Happy to see you will go deeper into it.  While @catenacyber said  `as u8` will be `nasty to debug` . Looking forward a more elegant way to do it. 


> The code seems find, but I'm not happy with the "look". I've replicated the issues by adding serde and a simple serializable struct. So I still want to look further into this, but it is something we need to resolve. I see serde entering our code at some point or another.

jasonish:
> > Happy to see you will go deeper into it. While @catenacyber said `as u8` will be `nasty to debug` . Looking forward a more elegant way to do it.
> > > The code seems find, but I'm not happy with the "look". I've replicated the issues by adding serde and a simple serializable struct. So I still want to look further into this, but it is something we need to resolve. I see serde entering our code at some point or another.
> 
> I wonder if `#[repr(u8)]` would cause any issues with the ffi. It'll catch those cast errors.

`#[repr(u8)]` does seem to do the right thing with respect to FFI and cbindgen.

jiamo:
Do you mean 
```
#[repr(u8)]
#[derive(Debug, PartialEq, Eq, Clone, Copy)]
pub enum Direction {
    ToServer = 0x04,
    ToClient = 0x08,
}
```
The problem is still here after change like that.

> I wonder if `#[repr(u8)]` would cause any issues with the ffi. It'll catch those cast errors.

jasonish:
> Do you mean
> 
> ```
> #[repr(u8)]
> #[derive(Debug, PartialEq, Eq, Clone, Copy)]
> pub enum Direction {
>     ToServer = 0x04,
>     ToClient = 0x08,
> }
> ```
> 
> The problem is still here after change like that.
> 
> > I wonder if `#[repr(u8)]` would cause any issues with the ffi. It'll catch those cast errors.

You still need to use `as u8`, but at least you can't accidentally stick a non-u8 value in the enum.

jiamo:
> Rebase, and use as u8. That should be clippy clean and look nicer as well.

Done. and change to `#[repr(u8)]`.

jiamo:
> @jiamo would you rebase and submit a new version of this PR ?

rebase done.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -1258,10 +1258,10 @@ pub unsafe extern "C" fn rs_dcerpc_get_tx_cnt(vtx: *mut std::os::raw::c_void) ->
 pub unsafe extern "C" fn rs_dcerpc_get_alstate_progress(tx: *mut std::os::raw::c_void, direction: u8
                                                  )-> std::os::raw::c_int {
     let tx = cast_pointer!(tx, DCERPCTransaction);
-    if direction == Direction::ToServer.into() && tx.req_done {
+    if direction == Direction::ToServer as u8 && tx.req_done {
```
catenacyber, line 1261: These changes do not seem needed

Could you just change `[repr(C)]` to `[repr(u8)]` ?

catenacyber:
Closing as stale, you can open a new rebased PR with comments taken into account.

PR #8161, opened 2022-11-11
Ci warnint v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None

Describe changes:
- Makes CI error on integer warnings !
- Fixes one newly introduced integer warning (by eb155639c6293e5fd412a8268a2c1a6ca03ebe45)

libhtp-pr: 369

Modifies https://github.com/OISF/suricata/pull/8144 by adding debug validations in defrag code...

PR #8168, opened 2022-11-14
Ci warnint v4

catenacyber:
None

Describe changes:
- Makes CI error on integer warnings !
- Fixes one newly introduced integer warning (by eb155639c6293e5fd412a8268a2c1a6ca03ebe45)

libhtp-pr: 369

Modifies https://github.com/OISF/suricata/pull/8161 by fixing comparison direction (like > instead of <=)

### DIFF
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -297,8 +297,8 @@ Defrag4Reassemble(ThreadVars *tv, DefragTracker *tracker, Packet *p)
     rp->recursion_level = p->recursion_level;
 
     int fragmentable_offset = 0;
-    int fragmentable_len = 0;
-    int hlen = 0;
+    uint16_t fragmentable_len = 0;
+    uint16_t hlen = 0;
     int ip_hdr_offset = 0;
 
     RB_FOREACH(frag, IP_FRAGMENTS, &tracker->fragment_tree) {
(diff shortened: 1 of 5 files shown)


catenacyber:
Why is there no CI here ?

suricata-qa:
ERROR:
 

Pipeline 10608

suricata-qa:

Pipeline 10611

src/defrag.c
```diff
@@ -297,8 +297,8 @@ Defrag4Reassemble(ThreadVars *tv, DefragTracker *tracker, Packet *p)
     rp->recursion_level = p->recursion_level;
 
     int fragmentable_offset = 0;
-    int fragmentable_len = 0;
-    int hlen = 0;
+    uint16_t fragmentable_len = 0;
+    uint16_t hlen = 0;
     int ip_hdr_offset = 0;
```
victorjulien, line 302: Wonder why this needs to stay `int` here, but can be made `uint16_t` in IPv6. In general I'm not clear why it needs to be signed and why it needs to be bigger than `uint16_t`. Does it take the offset from the whole packet? Not just the IP part?

@jasonish any idea?

victorjulien: I think a previous version of the PR did this, and showed issues in QAlab runs.

catenacyber: I do not think previous versions of the PR did this. I can try in a next one.
I just fixed the warnings, this `ip_hdr_offset` is not cast back to another uint16, just used as a pointer offset and in an addition for `SET_PKT_LEN`

For IPv6, it is also used in `unfragmentable_len` computation which is an uint16 (and does not exist in IP4)

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8173

PR #8178, opened 2022-11-16
Add derive to implement many common State functions

exlavasage:
Describe changes:
- Add derive to implement many common State functions, not useable for all State, but handles the general case
- Use AppLayerState where possible

### DIFF
--- a/rust/derive/src/utils.rs
+++ b/rust/derive/src/utils.rs
@@ -15,6 +15,21 @@
  * 02110-1301, USA.
  */
 
+/// If we're being used from within Suricata we have to reference the internal name space with
+/// "crate", but if we're being used by a library or plugin user we need to reference the
+/// Suricata name space as "suricata". Check the CARGO_PKG_NAME environment variable to
+/// determine what identifier to setup.
+pub fn crate_id() -> syn::Ident {
+    let is_suricata = std::env::var("CARGO_PKG_NAME")
+        .map(|var| var == "suricata")
+        .unwrap_or(false);
+    if is_suricata {
+        syn::Ident::new("crate", proc_macro2::Span::call_site())
+    } else {
+        syn::Ident::new("suricata", proc_macro2::Span::call_site())
+    }
+}
+
 /// Transform names such as "OneTwoThree" to "one_two_three".
 pub fn transform_name(name: &str, delim: char) -> String {
     let mut out = String::new();
--- a/rust/derive/src/utils.rs
+++ b/rust/derive/src/utils.rs
@@ -1,0 +1,60 @@
+/* Copyright (C) 2021 Open Information Security Foundation
+ *
+ * You can copy, redistribute or modify this Program under the terms of
+ * the GNU General Public License version 2 as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * version 2 along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ */
+
+/// Transform names such as "OneTwoThree" to "one_two_three".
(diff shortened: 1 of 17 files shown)


rust/derive/src/utils.rs
```diff
@@ -0,0 +1,75 @@
...
+        .unwrap_or(false);
+    if is_suricata {
+        syn::Ident::new("crate", proc_macro2::Span::call_site())
+    } else {
+        syn::Ident::new("suricata", proc_macro2::Span::call_site())
+    }
+}
+
+/// Transform names such as "OneTwoThree" to "one_two_three".
+pub fn transform_name(name: &str, delim: char) -> String {
```
jasonish, line 34: The move and the change of behaviour exist in one commit where it should probably be 2 and document why the change is needed in the commit message.

rust/derive/src/utils.rs
```diff
@@ -0,0 +1,75 @@
+/* Copyright (C) 2021 Open Information Security Foundation
```
jasonish, line 1: New file, so 2022.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -188,14 +188,14 @@ pub struct DCERPCTransaction {
 
 impl Transaction for DCERPCTransaction {
     fn id(&self) -> u64 {
-        // need +1 to match state.tx_id
-        self.id + 1
+        self.id
```
jasonish, line 191: Change of behaviour not addressed in git commits.

jasonish (changes requested):
Just a very light review so far.

- There appears to be some CI issues.
- Please create a ticket here: https://redmine.openinfosecfoundation.org/issues/4153
- I'm also undecided if putting this much code behind a derive is a good thing. It does make the code harder to reason about when reading.  I wonder if some simpler, more foundational functions make sense to be derived, then use generics for more of it, like has somewhat been started already.

exlavasage:
Going back through these. Thoughts on if/how this should be done?

jasonish:
> Going back through these. Thoughts on if/how this should be done?

My hesitation at the moment is its too soon to macro this up. I'm currently playing with some panic unhandling for parsers, and would like to abstract that away from the parsers, and at least at this time I think that would be harder if we had put this behind a derive macro as the patterns are still evolving

exlavasage:
> > Going back through these. Thoughts on if/how this should be done?
> 
> My hesitation at the moment is its too soon to macro this up. I'm currently playing with some panic unhandling for parsers, and would like to abstract that away from the parsers, and at least at this time I think that would be harder if we had put this behind a derive macro as the patterns are still evolving.

@jasonish Any feelings on retrying this? Currently writing a new Rust parser is a lot of boiler plate code and the standard for a lot of the common functions has changed, so you can't refer to other parsers that are wrong (including the template sometimes).

The derive can always be optional/overridable for parser that need to support some of them specially.

jasonish:
> > > Going back through these. Thoughts on if/how this should be done?
> > 
> > 
> > My hesitation at the moment is its too soon to macro this up. I'm currently playing with some panic unhandling for parsers, and would like to abstract that away from the parsers, and at least at this time I think that would be harder if we had put this behind a derive macro as the patterns are still evolving.
> 
> @jasonish Any feelings on retrying this? Currently writing a new Rust parser is a lot of boiler plate code and the standard for a lot of the common functions has changed, so you can't refer to other parsers that are wrong (including the template sometimes).
> 
> The derive can always be optional/overridable for parser that need to support some of them specially.

Probably worth looking at again once 8 is out.  @catenacyber What do you think? You've been doing a lot with Rust parsers these days as well.

jasonish:
> > > > Going back through these. Thoughts on if/how this should be done?
> > > 
> > > 
> > > My hesitation at the moment is its too soon to macro this up. I'm currently playing with some panic unhandling for parsers, and would like to abstract that away from the parsers, and at least at this time I think that would be harder if we had put this behind a derive macro as the patterns are still evolving.
> > 
> > 
> > @jasonish Any feelings on retrying this? Currently writing a new Rust parser is a lot of boiler plate code and the standard for a lot of the common functions has changed, so you can't refer to other parsers that are wrong (including the template sometimes).
> > The derive can always be optional/overridable for parser that need to support some of them specially.
> 
> Probably worth looking at again once 8 is out. @catenacyber What do you think? You've been doing a lot with Rust parsers these days as well.

edit: once 8 is out

catenacyber:
Maybe there is something good to do indeed

PR #8179, opened 2022-11-16
github-ci: non-root builder - v2

jasonish:
All the GitHub CI jobs run as root inside a container. This means the
testing is done in a different environment than a developer typically
uses, running as a user.

Add a job that does the build as a non-root user.

Previous PR: https://github.com/OISF/suricata/pull/8099


victorjulien:
Merged in #8217, thanks!

PR #8188, opened 2022-11-18
afpacket/netmap: warn about mixed ips, ids/tap deprecation - v1

jasonish:
Suricata already logs if AF_PACKET or Netmap are running in a mixed IPS
and IDS/TAP mode.  As the behavior is undefined when these modes are
mixed, it is best to deprecate and to not allow this behavior. For now
warn that it will be unsupported and fail in Suricata 8.

Ticket: https://redmine.openinfosecfoundation.org/issues/5587

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -738,7 +738,9 @@ int AFPRunModeIsIPS()
     }
 
     if (has_ids && has_ips) {
-        SCLogInfo("AF_PACKET mode using IPS and IDS mode");
+        SCLogWarning(SC_ERR_INVALID_ARGUMENT,
+                "AF_PACKET using both IPS and TAP/IDS mode, this will not "
+                "be allowed in Suricata 8 due to undefined behavior.");
         for (ldev = 0; ldev < nlive; ldev++) {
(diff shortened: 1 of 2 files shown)


src/runmode-af-packet.c
```diff
@@ -738,7 +738,9 @@ int AFPRunModeIsIPS()
     }
 
     if (has_ids && has_ips) {
-        SCLogInfo("AF_PACKET mode using IPS and IDS mode");
+        SCLogWarning(SC_ERR_INVALID_ARGUMENT,
+                "AF_PACKET using both IPS and TAP/IDS mode, this will not "
+                "be allowed in Suricata 8 due to undefined behavior.");
```
victorjulien, line 743: I think it makes sense to reference the ticket id, so we can add more info after the fact or address questions.

jasonish: Full link?  Or Just the "See ticket #xxxx for more details.".

jasonish:
https://github.com/OISF/suricata/pull/8190

src/runmode-af-packet.c
victorjulien: Ah, the latter I guess. As this seems to be an existing pattern in the code.

PR #8189, opened 2022-11-18
decode-udp: Add exception for invalid length on padded packets v1

lukashino:
If the original packet is shorter than Ethernet's minimal transmission unit (64B), the transmitting device pads the packet to the given size. Padding is added to the IP layer but the UDP layer remains unaffected. As a result, the UDP header length does not reach the end of the packet as there is still padding after the UDP layer.

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5693) ticket:

Describe changes:
- add header length check exception

suricata-verify-pr: 1009

### DIFF
--- a/src/decode-udp.c
+++ b/src/decode-udp.c
@@ -56,7 +56,9 @@ static int DecodeUDPPacket(ThreadVars *t, Packet *p, const uint8_t *pkt, uint16_
         return -1;
     }
 
-    if (unlikely(len != UDP_GET_LEN(p))) {
+    if (unlikely(len != UDP_GET_LEN(p) &&
+                 // avoid flagging IP padded packets to the minimal Ethernet unit as invalid HLEN
+                 (p->ethh != NULL && p->pktlen > ETHERNET_PKT_MIN_LEN))) {
         ENGINE_SET_INVALID_EVENT(p, UDP_HLEN_INVALID);
         return -1;
     }
--- a/src/decode-ethernet.h
+++ b/src/decode-ethernet.h
@@ -25,6 +25,7 @@
 #define __DECODE_ETHERNET_H__
 
 #define ETHERNET_HEADER_LEN           14
+#define ETHERNET_PKT_MIN_LEN          60 // Min. transmission unit on Ethernet is 64B (4B for FCS)
 
 /* Cisco Fabric Path / DCE header length. */
 #define ETHERNET_DCE_HEADER_LEN       (ETHERNET_HEADER_LEN + 2)


src/decode-udp.c
```diff
@@ -56,7 +56,9 @@ static int DecodeUDPPacket(ThreadVars *t, Packet *p, const uint8_t *pkt, uint16_
         return -1;
     }
 
-    if (unlikely(len != UDP_GET_LEN(p))) {
+    if (unlikely(len != UDP_GET_LEN(p) &&
+                 // avoid flagging IP padded packets to the minimal Ethernet unit as invalid HLEN
+                 (p->ethh != NULL && p->pktlen > ETHERNET_PKT_MIN_LEN))) {
```
lukashino, line 61: I did not want to assume that underlying L2 is always Ethernet so I added `p->ethh` to be more certain. But not sure if that is the best solution :shrug:

victorjulien: I think this should be addressed on a completely different level.

We know:
1. the packet size we got from capture
2. packet type (ethernet)
3. size of ethernet header
4. size of IP packet
5. size of UDP header
6. UDP length field
 
So based on this we can calculate:
- trailing ethernet data
- trailing IP data

UDP length logic should be based on IP size, not be aware of any lower layers.

lukashino: I am not sure this information would help you with this check. 

I interpret this check as whatever payload length is after the IP layer must equal to the length of the UDP layer. This check assumes that whatever is after the IP header must be part of the UDP layer. 
So when I have a packet with a length of 150 bytes, ETH could be 14, IP 20, I expect to find a UDP layer length equal to 116 bytes in the UDP header. IP length does not count for trailing Ethernet data. 
If we assume there should be nothing after the UDP layer (so everything should be encapsulated within UDP layer) then this check makes sense to me. 
However, when I transmit a packet that is less than 60 bytes long over Ethernet then the NIC/kernel pads the IP layer with extra 0 to 60 bytes. This length is adjusted in the IP layer but neither in the Ethernet nor in the UDP layer. 
So a packet with ETH 14 bytes, IP 20 bytes, and UDP 8 bytes will have the space after the UDP layer padded with 0, setting the IP length to 46 bytes.  The padding is not set as ETH trailer and neither in UDP layer length (in the UDP layer). 

So you can calculate trailing IP data but I do not see how this would help you in this check and I think that then defies the purpose of this check. You also can not assume each packet must be at least 60 bytes long as this minimum probably only applies to Ethernet and not PPP/FrameRelay etc. packets. 

So unless there is a solution that I don't see I would either remove the check completely or leave it as is.

victorjulien: From what it looks like to me, this is simply trailing IP data. In the packet we have 60 bytes total, 14 bytes ethernet header, 46 bytes IP. 26 bytes are IP data (46 - 20 for the IP hdr), but the UDP layer only consumes 8 bytes. So there are 18 bytes not consumed that are part of the IP packet. We should probably expose those to the detection engine as "ip padding data" or "ip trailing data". Regardless, the ethernet layer is not relevant to the code here. The code should only care about how much IP data we get, then how much we consume in UDP (and the same is likely true for TCP and possibly others?)

lukashino: Considering your comment:
- the assumption about the length of the L4 layer being equal to the payload length of the L3 layer is not correct
- then we can assume even bigger packets (150 byte example in my previous comment) breaking this assumption are actually correct and should not be marked as invalid packets
- the check is based on this assumption and if this assumption is not valid then the check should be removed. 
- we can call everything after the first L4 layer and within IP payload length as IP trailing data

As you mentioned, storing these data in a separate buffer, I've thought, is that really necessary? Would the payload matching keyword match on the whole packet?

victorjulien: Payload matching trusts the TCP/UDP length logic, so the trailing data would be excluded from it. I think this is correct, as we should be able to match on the end of the TCP/UDP payload w/o rule writers having to worry about possible trailing data. I think the UDP decoder should store a end of data offset that the IP decoders then can use to know that there is trailing data. A special keyword can be added to expose it to the rule lang.

lukashino:
Continues in #8211

PR #8205, opened 2022-11-23
Bug 4786/v5

inashivb:


src/detect-flowbits.c
```diff
@@ -207,125 +170,86 @@ int DetectFlowbitMatch (DetectEngineThreadCtx *det_ctx, Packet *p,
...
             }
-        }
+            break;
     }
 
-    return 1;
+    if (or_list_size != 0) {
+        for (uint32_t i = 0; i < or_list_size; i++) {
+            int leq = VarNameStoreSetupAdd(rs_xbits_get_or_list_item(cd, i), VAR_TYPE_FLOW_BIT);
+            rs_xbits_set_or_list_id_at(cd, i, leq);
```
inashivb, line 225: @jasonish need a little help here.. Made all the args `const void *` to send over to rust and to keep compatibility with the C fn calls but I also want to modify certain fields in Rust on the struct.
Here, I get the error for discarding const qualifier bc I try to get it as mut on the Rust side to modify it.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_asan | | |
 

Pipeline 10911

suricata-qa:
ERROR:
 

Pipeline 10911

src/detect-flowbits.c
jasonish: I think the answer is to not pass them as `const`, which might require API changes elsewhere as well.

inashivb:
Needs a lot more work.

PR #8208, opened 2022-11-25
Next/20221124/v3

victorjulien:
Staging:
- #8202 
- #8199 
- file handling fixes
- smb handling fixes

suricata-verify-pr: 1014

PR #8202, opened 2022-11-22
smb: do not use tree id to match request and response

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5161

Describe changes:
- smb2: do not use tree id to match request and response for access to `ssn2vecoffset_map`

I wonder if this should be completed for ssn2vec_map as well

suricata-verify-pr: 1014
https://github.com/OISF/suricata-verify/pull/1014

PR #8199, opened 2022-11-22
readthedocs: enable all formats - v1

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/5654

PR #8209, opened 2022-11-25
Next/20221124/v4

victorjulien:
Staging:
- #8202 
- #8199 
- file handling fixes
- smb handling fixes

suricata-verify-pr: 1014

replaces #8208 with a fuzz fixup

### DIFF
--- a/src/util-file.h
+++ b/src/util-file.h
@@ -61,6 +61,10 @@ typedef struct SCMd5 SCMd5;
 // to be used instead of PATH_MAX which depends on the OS
 #define SC_FILENAME_MAX 4096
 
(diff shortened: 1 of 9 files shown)


src/util-file.h
```diff
@@ -61,6 +61,10 @@ typedef struct SCMd5 SCMd5;
 // to be used instead of PATH_MAX which depends on the OS
 #define SC_FILENAME_MAX 4096
 
+#define FILEDATA_CONTENT_LIMIT            100000
```
catenacyber, line 64: Why did this one move ?
It is also duplicately defined in app-layer-smtp.c

PR #8211, opened 2022-11-25
decode-udp:  Allow shorter UDP packets than the remaining payload length

lukashino:
Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5693) ticket:

Follow-up of #8189 

Describe changes:
- allow a UDP layer of any size to be shorter than the remaining payload size

### DIFF
--- a/src/decode-ethernet.h
+++ b/src/decode-ethernet.h
@@ -25,6 +25,7 @@
 #define __DECODE_ETHERNET_H__
 
 #define ETHERNET_HEADER_LEN           14
+#define ETHERNET_PKT_MIN_LEN          60 // Min. transmission unit on Ethernet is 64B (4B for FCS)
 
 /* Cisco Fabric Path / DCE header length. */
 #define ETHERNET_DCE_HEADER_LEN       (ETHERNET_HEADER_LEN + 2)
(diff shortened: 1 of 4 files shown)


src/decode-ethernet.h
```diff
@@ -25,6 +25,7 @@
 #define __DECODE_ETHERNET_H__
 
 #define ETHERNET_HEADER_LEN           14
+#define ETHERNET_PKT_MIN_LEN          60 // Min. transmission unit on Ethernet is 64B (4B for FCS)
```
victorjulien, line 28: left over from previous round

src/decode-events.c
```diff
@@ -266,10 +266,6 @@ const struct DecodeEvents_ DEvents[] = {
             "decoder.udp.hlen_too_small",
             UDP_HLEN_TOO_SMALL,
     },
-    {
```
victorjulien, line 269: how will this work wrt loading an old rule? Test with `rules/decoder-events.rules`

lukashino: Good catch, it fails to start and complains about using an invalid keyword. 
I can create a deprecation notice for the keyword in 7.0.0 and silently ignore the keyword in 6.0.9?

lukashino:
Continues in #8213

PR #8223, opened 2022-11-30
Fuzz fixes/v3

victorjulien:
Fuzz fixes for smb and stream.

### DIFF
--- a/rust/src/smb/files.rs
+++ b/rust/src/smb/files.rs
@@ -89,6 +89,34 @@ impl SMBState {
         return tx_ref.unwrap();
     }
 
+    /// get file tx for a open file. Returns None if a file for the fuid exists,
+    /// but has already been closed.
+    pub fn get_file_tx_by_fuid_with_open_file(&mut self, fuid: &[u8], direction: Direction)
+        -> Option<&mut SMBTransaction>
+    {
+        let f = fuid.to_vec();
+        for tx in &mut self.transactions {
+            let found = match tx.type_data {
+                Some(SMBTransactionTypeData::FILE(ref mut d)) => {
+                    direction == d.direction && f == d.fuid && !d.file_tracker.is_done()
+                },
+                _ => { false },
+            };
+
+            if found {
+                SCLogDebug!("SMB: Found SMB file TX with ID {}", tx.id);
+                if let Some(SMBTransactionTypeData::FILE(ref mut d)) = tx.type_data {
+                    tx.tx_data.update_file_flags(self.state_data.file_flags);
+                    d.update_file_flags(tx.tx_data.file_flags);
+                }
+                return Some(tx);
+            }
+        }
+        SCLogDebug!("SMB: Failed to find SMB TX with FUID {:?}", fuid);
+        return None;
+    }
+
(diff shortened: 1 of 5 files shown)


suricata-qa:

Pipeline 10997

rust/src/smb/files.rs
```diff
@@ -89,6 +89,34 @@ impl SMBState {
         return tx_ref.unwrap();
     }
 
+    /// get file tx for a open file. Returns None if a file for the fuid exists,
+    /// but has already been closed.
+    pub fn get_file_tx_by_fuid_with_open_file(&mut self, fuid: &[u8], direction: Direction)
+        -> Option<&mut SMBTransaction>
+    {
+        let f = fuid.to_vec();
+        for tx in &mut self.transactions {
```
catenacyber, line 98: Do we really need this loop over transactions ?

rust/src/smb/smb2.rs
```diff
@@ -297,7 +297,7 @@ pub fn smb2_write_request_record<'b>(state: &mut SMBState, r: &Smb2Record<'b>)
             };
 
             let mut set_event_fileoverlap = false;
-            let found = match state.get_file_tx_by_fuid(&file_guid, Direction::ToServer) {
```
catenacyber, line 300: Do we still need simple function `get_file_tx_by_fuid` ?

src/tests/fuzz/fuzz_sigpcap.c
```diff
@@ -163,6 +163,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     p->ts.tv_sec = header->ts.tv_sec;
     p->ts.tv_usec = header->ts.tv_usec % 1000000;
     p->datalink = pcap_datalink(pkts);
+    p->pkt_src = PKT_SRC_WIRE;
```
catenacyber, line 166: Should this be done for fuzz_predefpcap_aware.c as well ? 
Looks done already for fuzz_sigpcap_aware.c

rust/src/smb/smb2.rs
victorjulien: Yes it's used by the close command and one other case.

src/tests/fuzz/fuzz_sigpcap.c
victorjulien: Yeah probably. If I run anything with `--enable-debug` it reaches an assertion otherwise.

catenacyber: Ok, I never ran fuzzing with debug (only debug validation)

victorjulien: It's just when I'm running the reproducer, then I have debug on.

rust/src/smb/files.rs
inliniac: Goal is to look up a tx in the tx vector, so I think so.

victorjulien: > Goal is to look up a tx in the tx vector, so I think so.

Wrong account, wh00ps

PR #8227, opened 2022-11-30
http: logs custom headers in a subobject

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5320

Describe changes:
- http: logs custom headers in a subobject

To avoid collisions, for instance for `content_range`

With this PR, the output will be
```
    "content_range": {
      "raw": "bytes 10-20/69",
      "start": 10,
      "end": 20,
      "size": 69
    },
    "custom": {
      "content_range": "bytes 10-20/69"
    }
```

instead of 
```
    "content_range": {
      "raw": "bytes 10-20/69",
      "start": 10,
      "end": 20,
      "size": 69
    },
   "content_range": "bytes 10-20/69"
```

suricata-verify-pr: 1023

### DIFF
--- a/etc/schema.json
+++ b/etc/schema.json
@@ -1685,6 +1685,10 @@
                     },
                     "additionalProperties": false
                 },
+                "custom": {
+                    "type": "object",
+                    "additionalProperties": true
+                },
                 "http2": {
                     "type": "object",
                     "properties": {
--- a/src/output-json-http.c
+++ b/src/output-json-http.c
@@ -269,6 +269,7 @@ static void EveHttpLogJSONCustom(LogHttpFileCtx *http_ctx, JsonBuilder *js, htp_
     char *c;
     HttpField f;
 
+    jb_open_object(js, "custom");
     for (f = HTTP_FIELD_ACCEPT; f < HTTP_FIELD_SIZE; f++)
     {
         if ((http_ctx->fields & (1ULL<<f)) != 0)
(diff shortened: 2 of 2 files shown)


catenacyber:
This bug is only present if http custom logging is enabled in suricata.yaml ;-)

jasonish:
> "custom": {

Would "headers" be a better name? Or would that conflict in other ways?

catenacyber:
> > "custom": {
> 
> Would "headers" be a better name? Or would that conflict in other ways?

There is already `request_headers` and `response_headers` (maybe from HTTP2)
So, it would look confusing, would it not ?

suricata-qa:

Pipeline 11014

jasonish:
> > > "custom": {
> > 
> > 
> > Would "headers" be a better name? Or would that conflict in other ways?
> 
> There is already `request_headers` and `response_headers` (maybe from HTTP2) So, it would look confusing, would it not ?

Hmm.. Sounds confusing.  So we could end up with the same headers in `request_headers` and `custom`?  I just think `custom` is a bad name and don't want to see it propogated into our logging.

catenacyber:
`custom_headers` then ?

jasonish:
> `custom_headers` then ?

If the header is request, can we just put it in `request_headers`, and if a response header, put it in `response_headers`? Probably shouldn't add a third object for headers if we can avoid it.

I'm thinking we should maybe look at whats best for the overall format of http logs.

catenacyber:
Okay...

So, this means we will change eve.json output from
`"content_range": "bytes 10-20/69"` to
```
{
"name": "content_range",
"value": "bytes 10-20/69",
}
```

Is this okay ?

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8229

PR #8229, opened 2022-12-02
http: logs custom headers in a subobject

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5320

Describe changes:
- http: logs custom headers in a subobject

To avoid collisions, for instance for `content_range`

With this PR, the output will be
```
    "content_range": {
      "raw": "bytes 10-20/69",
      "start": 10,
      "end": 20,
      "size": 69
    },
    "response_headers": [
      {
        "name": "Content-Range",
        "value": "bytes 10-20/69"
      }
    ]
```

instead of 
```
    "content_range": {
      "raw": "bytes 10-20/69",
      "start": 10,
      "end": 20,
      "size": 69
    },
   "content_range": "bytes 10-20/69"
```

suricata-verify-pr: 1026

Replaces #8227 with using `request_headers` json object

### DIFF
--- a/doc/userguide/output/eve/eve-json-format.rst
+++ b/doc/userguide/output/eve/eve-json-format.rst
@@ -286,7 +286,7 @@ suricata.yaml file the following fields are (can) also included:
 * "status": HTTP status code
 * "protocol": Protocol / Version of HTTP (ex: HTTP/1.1)
 * "http_method": The HTTP method (ex: GET, POST, HEAD)
-* "http_refer": The referrer for this action
+* "http_refer": The referer for this action
 
 In addition to the extended logging fields one can also choose to enable/add
 from more than 50 additional custom logging HTTP fields enabled in the
@@ -318,7 +318,7 @@ suricata.yaml file. The additional fields can be enabled as following:
               allow, connection, content-encoding, content-language,
               content-length, content-location, content-md5, content-range,
               content-type, date, etags, expires, last-modified, link, location,
-              proxy-authenticate, referrer, refresh, retry-after, server,
+              proxy-authenticate, referer, refresh, retry-after, server,
               set-cookie, trailer, transfer-encoding, upgrade, vary, warning,
               www-authenticate, x-flash-version, x-authenticated-user]
 
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -192,7 +192,7 @@ last_modified           last-modified
 link                    link
 location                location
 proxy_authenticate      proxy-authenticate
-referrer                referrer
+referer                 referer
 refresh                 refresh
 retry_after             retry-after
 server                  server
--- a/src/output-json-http.c
+++ b/src/output-json-http.c
@@ -138,7 +138,7 @@ struct {
     const char *config_field;
     const char *htp_field;
     uint32_t flags;
-} http_fields[] =  {
+} http_fields[] = {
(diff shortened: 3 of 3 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .capture.kernel_drops | 0 | 57235 | - |
| .flow.end.tcp_state.syn_sent | 0 | 123 | - |
| .flow.end.tcp_state.syn_recv | 0 | 1 | - |
| .flow.end.tcp_state.fin_wait1 | 0 | 23 | - |
| .flow.end.tcp_state.fin_wait2 | 0 | 10 | - |
| .flow.end.tcp_state.time_wait | 0 | 10 | - |
| .flow.end.tcp_state.last_ack | 0 | 10 | - |
| .flow.end.tcp_state.close_wait | 0 | 20 | - |
| .tcp.segment_from_pool | 15728353 | 16523457 | 105.06% |
| .app_layer.error.http.parser | 0 | 3 | - |
| .app_layer.error.ftp.gap | 0 | 2 | - |
| .app_layer.error.smtp.gap | 0 | 10 | - |
| .app_layer.error.dcerpc_tcp.parser | 0 | 3 | - | 

Pipeline 11035

jasonish:
This makes sense to me.  It woldn't duplicate a logged header right?  Like when we log one by default, then add it to custom?

The name, value setup is pre-existing right?  I ask cause I quite like Caddy log server format, but its probably too late for us: https://gist.github.com/jasonish/b23ba7bf9a8097172807d3c2541258df

catenacyber:
> It woldn't duplicate a logged header right?

It should not, but please review the code, as there may be a combination of configuration options I did not take into account...

> The name, value setup is pre-existing right? 

Yes it is in schema.json

The Caddy log server format was in the previous PR in the `custom` object :-p

jasonish:
> > It woldn't duplicate a logged header right?
> 
> It should not, but please review the code, as there may be a combination of configuration options I did not take into account...
> 
> > The name, value setup is pre-existing right?
> 
> Yes it is in schema.json
> 
> The Caddy log server format was in the previous PR in the `custom` object :-p

Note how each value in Caddy is an array.  Anyways, best for us to consolidate on one format, that used mixed formats.

catenacyber:
> Note how each value in Caddy is an array

Yeah, I had noticed, this is the golang way...
Not sure I like it for my part :-p

Anyways, let me know how this PR is

jasonish:
This is much better than current master IMO.  One quirk.. Given the following configuration:
```
            custom: ["accept"]
            # set this value to one and only one from {both, request, response}
            # to dump all HTTP headers for every HTTP request and/or response
            dump-all-headers: both
```
in master, I get "accept" and dump-all-headers.  With this PR, the `custom` seems to take precedence over `dump-all-headers`. Is that intentional?

catenacyber:
This config should print a new warning.
How would you handle it ? Fatal error ?

jasonish:
> This config should print a new warning.
> 
> How would you handle it ? Fatal error ?

As dump-all-headers is just a superset of custom it should silently win.

jasonish:
I noticed that current master does seem to silently prefer dump-all-headers.  This branch does log a warning but prefers custom.  I guess a warning is fine, but dump-all-headers should be preferred to keep compatibility, and I think it makes more sense.  My preference is to silently prefer it like mast currently does as well.

catenacyber:
Ok done in https://github.com/OISF/suricata/pull/8241

PR #8256, opened 2022-12-08
fuzz: fix use of uninitialized value

catenacyber:
None cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54111

Describe changes:
- fuzz targets: fix use of uninitialized value

### DIFF
--- a/src/tests/fuzz/fuzz_sigpcap_aware.c
+++ b/src/tests/fuzz/fuzz_sigpcap_aware.c
@@ -157,7 +157,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     // loop over packets
     r = FPC_next(&pkts, &header, &pkt);
     p = PacketGetFromAlloc();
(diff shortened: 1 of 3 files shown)


src/tests/fuzz/fuzz_sigpcap_aware.c
```diff
@@ -157,7 +157,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     // loop over packets
     r = FPC_next(&pkts, &header, &pkt);
     p = PacketGetFromAlloc();
-    if (header.ts.tv_sec >= INT_MAX - 3600) {
+    if (v header.ts.tv_sec >= INT_MAX - 3600) {
```
victorjulien, line 160: this looks like a paste error

while at it, could you add a check that makes sure tv_usec >= 0 (for each of these pcap targets)? I've seen rust overflow panics still due to negative values

catenacyber: Ok thanks

PR #8117, opened 2022-10-28
detect-detection: Convert unittests to new FAIL/PASS API

hsadia538:
- [X] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Previous PR: https://github.com/OISF/suricata/pull/8115

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/4019) ticket:

Describe new changes:

- Remove extra files

PR #8262, opened 2022-12-08
detect-detection: Convert unittests to new FAIL/PASS API

hsadia538:

Previous PR: https://github.com/OISF/suricata/pull/8117

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/4019) ticket:

Draft PR to gather feedback

Describe Changes:

- Remove `drops` and `alerts` counter and replace with `FAIL/PASS*` macro

### DIFF
--- a/src/detect-detection-filter.c
+++ b/src/detect-detection-filter.c
@@ -284,14 +284,14 @@ static void DetectDetectionFilterFree(DetectEngineCtx *de_ctx, void *df_ptr)
  */
 static int DetectDetectionFilterTestParse01 (void)
 {
-    DetectThresholdData *df = NULL;
-    df = DetectDetectionFilterParse("track by_dst,count 10,seconds 60");
-    if (df && (df->track == TRACK_DST) && (df->count == 10) && (df->seconds == 60)) {
-        DetectDetectionFilterFree(NULL, df);
-        return 1;
-    }
+    DetectThresholdData *df = DetectDetectionFilterParse("track by_dst,count 10,seconds 60");
+    FAIL_IF_NULL(df);
+    FAIL_IF_NOT(df->track == TRACK_DST);
+    FAIL_IF_NOT(df->count == 10);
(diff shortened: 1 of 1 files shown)


hsadia538:
Based on feedback on the last PR, I have removed the counters for drops and alerts and replaced them with FAIL/PASS macro. I have two queries/confusions regarding the tests in the code:

- For `DetectDetectionFilterTestSig1` the test passes only when I change all alert conditions to `FAIL_IF` meaning all the 8 checks should not generate an alert. But in the code, it says there should be an alert match of 4 which is implemented via the counter(old code). The test fails if I try to do the same by using FAIL/PASS API.
- For `DetectDetectionFilterTestSig3` The test is supposed to generate an alert after encountering 2 or more matches within 60 seconds window but when I try to use FAIL_IF_NOT(alert) for matches after the timer is set to 200 seconds, the test fails.

I suppose the issue is that I am not putting the right condition as to when it should or shouldn't generate an alert and that's why it fails. But after trying multiple combinations based on my understanding of the rule I am really confused.

src/detect-detection-filter.c
```diff
@@ -395,64 +384,55 @@ static int DetectDetectionFilterTestParse06 (void)
...
+            "alert tcp any any -> any 80 (msg:\"detection_filter Test\"; detection_filter: "
+            "track by_dst, count 4, seconds 60; sid:1;)");
+    FAIL_IF_NULL(s);
 
     SigGroupBuild(de_ctx);
     DetectEngineThreadCtxInit(&th_v, (void *)de_ctx, (void *)&det_ctx);
 
     SigMatchSignatures(&th_v, de_ctx, det_ctx, p);
-    alerts = PacketAlertCheck(p, 1);
+    FAIL_IF(PacketAlertCheck(p, 10));
```
victorjulien, line 411: sid 10? rule is sid 1

hsadia538: fixed it. It was failing because of this only. Thank you!

PR #8266, opened 2022-12-09
exceptions: add exception policy master switch - v1

jufajardini:
https://redmine.openinfosecfoundation.org/issues/5219

Describe changes:
- add a master switch for the exception policies, allowing one to set a single policy to all traffic exceptions
- document said feature
- add note about that in the upgrading section

The current behavior considers that the default behavior for 7 will still be to have this disabled as built-in, but enabled via config file.

Jason has argued in the ticket that we might as well introduce this as a breaking change from 7, already, and just document it. This PR is also to gather feedback on the current approach, in case we want to switch for the directly breaking one.

suricata-verify-pr: 1036
https://github.com/OISF/suricata-verify/pull/1036

### DIFF
--- a/doc/userguide/configuration/exception-policies.rst
+++ b/doc/userguide/configuration/exception-policies.rst
@@ -66,10 +66,10 @@ are:
 - ``pass-flow``: disable payload and packet detection; stream reassembly,
   app-layer parsing and logging still happen.
 - ``pass-packet``: disable detection, still does stream updates and app-layer
-  parsing (depeding on which policy triggered it).
+  parsing (depending on which policy triggered it).
 - ``ignore``: do not apply exception policies (default behavior).
 
-The *Drop*, *pass* and *reject* are similar to the rule actions described in :ref:`rule
+The *drop*, *pass* and *reject* are similar to the rule actions described in :ref:`rule
 actions<suricata-yaml-action-order>`.
 
 Command-line Options for Simulating Exceptions
--- a/doc/userguide/configuration/exception-policies.rst
+++ b/doc/userguide/configuration/exception-policies.rst
@@ -16,6 +16,28 @@ simulate failures or errors and understand Suricata behavior under such conditio
 Exception Policies
 ------------------
 
+Master Switch
+~~~~~~~~~~~~~
+
+In IPS mode, the default behavior for all exception policies is to drop packets
+and/or flows. It is possible to disable this default, by setting the exception
+policies "master switch" yaml config option, to ``disabled``:
+
+::
+
+   # Define a common behavior for all exception policies
+   # the default is drop-packet/drop-flow. To fallback to old behavior (setting
+   # each of them individually, or ignoring all), disable it.
+   # Possible values: auto (drop-packet/drop-flow), perfomance (bypass),
+   # disabled
+   exception-policy-master-switch: disabled
+
+This value will be overwritten by specific exception policies whose settings are
+also defined in the yaml file.
+
+Specific settings
+~~~~~~~~~~~~~~~~~
+
 Exception policies are implemented for:
 
 .. list-table:: Exception Policy configuration variables
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -125,6 +125,7 @@
 #include "util-device.h"
 #include "util-dpdk.h"
 #include "util-ebpf.h"
+#include "util-exception-policy.h"
 #include "util-host-os-info.h"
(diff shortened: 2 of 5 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128237 | 104100 | 81.18% |
| .capture.kernel_drops | 0 | 513 | 0.00 | 

Pipeline 11120

doc/userguide/configuration/exception-policies.rst
```diff
@@ -16,6 +16,28 @@ simulate failures or errors and understand Suricata behavior under such conditio
...
+policies "master switch" yaml config option, to ``disabled``:
+
+::
+
+   # Define a common behavior for all exception policies
+   # the default is drop-packet/drop-flow. To fallback to old behavior (setting
+   # each of them individually, or ignoring all), disable it.
+   # Possible values: auto (drop-packet/drop-flow), perfomance (bypass),
+   # disabled
+   exception-policy-master-switch: disabled
```
victorjulien, line 33: wonder if we should just accept the exception-policy values here, in addition to "auto"

src/suricata.c
```diff
@@ -2678,6 +2679,8 @@ int PostConfLoadedSetup(SCInstance *suri)
 
     MacSetRegisterFlowStorage();
 
+    setMasterExceptionPolicy();
```
victorjulien, line 2682: style: `SetMasterExceptionPolicy`

src/util-exception-policy.c
```diff
@@ -73,6 +75,83 @@ void ExceptionPolicyApply(Packet *p, enum ExceptionPolicy policy, enum PacketDro
...
 
+/* Should I make a separate enum for this? With a separate function?
+   This seems to make sense in my head right now, as then things would be more logically containted,
+   and then I could call the master-switch function from within the ExceptionPolicyParse function...
+ */
+
+void setMasterExceptionPolicy()
+{
+    const char *switch_str = NULL;
+    if (EngineModeIsIPS()) {
```
victorjulien, line 86: this shouldn't be specific to IPS mode. Performance mode (or bypass mode) would be useful for IDS mode too

src/util-exception-policy.c
```diff
@@ -73,6 +75,83 @@ void ExceptionPolicyApply(Packet *p, enum ExceptionPolicy policy, enum PacketDro
...
+        }
+    }
+}
+
+/** brief Set a master Exception Policy, if one has been defined
+ *
+ *  \retval true if the master Exception Policy has been set
+ *  \retval false if the master switch is disabled and no policy was set
+ */
+static bool getMasterExceptionPolicy(const char *option, enum ExceptionPolicy *policy)
```
victorjulien, line 124: might also extend the ExceptionPolicy to include a EXCEPTION_POLICY_NOT_SET and just return that

victorjulien (changes requested):

doc/userguide/configuration/exception-policies.rst
victorjulien: then it could also just be called `exception-policy` in the root namespace?

src/util-exception-policy.c
jufajardini: Ok! So, only the drop portion would be IPS-only?

victorjulien: yeah

doc/userguide/configuration/exception-policies.rst
jufajardini: > then it could also just be called `exception-policy` in the root namespace?

I'm having a bit of trouble understanding this one. 
Do you mean that then it could be something like:
`exception-policy: <mode>`?
In the yaml configuration?

src/util-exception-policy.c
jufajardini: Can you think of any differences between EXCEPTION_POLICY_IGNORE and what EXCEPTION_POLICY_NOT_SET would do? I'm thinking that it's the same, since we use 'ignore' when the user doesn't want to apply exception policies...

doc/userguide/configuration/exception-policies.rst
jufajardini: > wonder if we should just accept the exception-policy values here, in addition to "auto"

Do I interpret it right, that we'd also have just one enum for this?

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/8273

PR #5747, opened 2021-01-19
PostgreSQL: initial support

jufajardini:
- pgsql/parser: add startup message structure, basic nom parsers and corresponding unittests
- pgsql/parser: add parsers for the startup parameters "user" and "database" and corresponding unittests

Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4241?tab=properties

Previous (draft) PR: 
https://github.com/OISF/suricata/pull/5721

Describe changes:
- pgsql/parser: parse a simple startup message with mandatory 'user' and 'database' fields, in this order. If there are more parameters, those are stored in a data field, which isn't processed yet.
- There's no further check regarding packet size, parser read known parameters and stores the rest.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #8268, opened 2022-12-09
radix: fix ipv6 address parsing warning

victorjulien:
The check meant to see if the ip address part of the ip/cidr combo was more specific than needed wasn't fully implemented, leading to warnings being issued on completely valid and correct input.

This patch implements the same logic as in IPv4. If the ip address as specified is different from the ip after the mask has been applied, a warning is displayed.

Bug: #5747.
(cherry picked from commit 991f9fde3292d45eefcfe9e01ef8480e5864977b)


suricata-qa:

Pipeline 11128

PR #8172, opened 2022-11-16
devguide: add page on installing suricata from git

mrbazzan:
Continuation from #8128 

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#5585](https://redmine.openinfosecfoundation.org/issues/5585)

Describe changes:
- Incorporated review from #8128 

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #8269, opened 2022-12-09
devguide/install: add note about ubuntu version - v1

jufajardini:
We want to make it clear with which system the instructions for installing from were tested with.

Addressing https://github.com/OISF/suricata/pull/8172#issuecomment-1342909401


victorjulien:
Merged in #8314, thanks!

PR #8279, opened 2022-12-12
exceptions: add exception policy master switch - v3

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: 5219

Previous PR: https://github.com/OISF/suricata/pull/8273

Changes from previous PR:
- remove 'performance' option
- add 'pass' as a possible option for IDS mode
- make 'auto' have different behaviors in IPS (drop) and IDS (disabled), defined when parsing the config
- check if the passed exception-policy is valid at parsing time to prevent weird behaviors/errors with the master switch

suricata-verify-pr: 1039
https://github.com/OISF/suricata-verify/pull/1039

PR #8273, opened 2022-12-11
exceptions: add exception policy master switch - v2.1

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: 5219

Previous PR: #8266 

Changes from previous PR:
- use the same Exception Policies enum for the master switch
- simplify the config set name to ``exception-policy``
- accept all exception policy values, in addition to "auto", "performance" and "disabled"
- simplify how the master switch is processed
- accept performance policy for IDS mode as well

suricata-verify-pr: 1037
https://github.com/OISF/suricata-verify/pull/1037

PR #8280, opened 2022-12-13
exceptions: add exception policy master switch - v3.1

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5219](https://redmine.openinfosecfoundation.org/issues/5219)

Previous PR: https://github.com/OISF/suricata/pull/8273

This is the same branch as PR https://github.com/OISF/suricata/pull/8279, but I had used the wrong SV PR and therefore messed up with the CI checks

Changes from previous PR:
- remove 'performance' option
- add 'pass' as a possible option for IDS mode
- make 'auto' have different behaviors in IPS (drop) and IDS (disabled), defined when parsing the config
- check if the passed exception-policy is valid at parsing time to prevent weird behaviors/errors with the master switch

suricata-verify-pr: 1039
https://github.com/OISF/suricata-verify/pull/1039

### DIFF
--- a/src/util-exception-policy.c
+++ b/src/util-exception-policy.c
@@ -27,12 +27,24 @@
 #include "stream-tcp-reassemble.h"
 #include "action-globals.h"
 
+enum ExceptionPolicy g_eps_master_switch = EXCEPTION_POLICY_NOT_SET;
+
+void SetMasterExceptionPolicy()
+{
+    g_eps_master_switch = ExceptionPolicyParse("exception-policy", true);
+}
+
+static enum ExceptionPolicy GetMasterExceptionPolicy(const char *option)
+{
+    return g_eps_master_switch;
+}
+
 void ExceptionPolicyApply(Packet *p, enum ExceptionPolicy policy, enum PacketDropReason drop_reason)
 {
     SCLogDebug("start: pcap_cnt %" PRIu64 ", policy %u", p->pcap_cnt, policy);
     if (EngineModeIsIPS()) {
         switch (policy) {
-            case EXCEPTION_POLICY_IGNORE:
+            case EXCEPTION_POLICY_NOT_SET:
                 break;
             case EXCEPTION_POLICY_REJECT:
                 SCLogDebug("EXCEPTION_POLICY_REJECT");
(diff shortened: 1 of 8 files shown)


src/util-exception-policy.c
```diff
@@ -69,36 +81,86 @@ void ExceptionPolicyApply(Packet *p, enum ExceptionPolicy policy, enum PacketDro
...
-            policy = EXCEPTION_POLICY_DROP_PACKET;
-            SCLogConfig("%s: %s", option, value_str);
+            if (is_ips_mode) {
+                policy = EXCEPTION_POLICY_DROP_PACKET;
+                SCLogConfig("%s: %s", option, value_str);
+            } else {
+                policy = EXCEPTION_POLICY_NOT_SET;
+                SCLogConfig("%s: %s not a valid config in IDS mode. Setting it to disabled)",
+                        option, value_str);
+            }
```
jufajardini, line 139: I'm not happy with this solution, as it duplicates a lot of code. The alternative could be.... to have a separate function to parse the exception policy master switch?

src/util-exception-policy.c
```diff
@@ -69,36 +81,86 @@ void ExceptionPolicyApply(Packet *p, enum ExceptionPolicy policy, enum PacketDro
...
+                }
+                /* fall through */
+            case EXCEPTION_POLICY_PASS_PACKET:
+                SCLogDebug("EXCEPTION_POLICY_PASS_PACKET");
+                DecodeSetNoPayloadInspectionFlag(p);
+                DecodeSetNoPacketInspectionFlag(p);
+                break;
+            default:
+                break;
+        }
```
jufajardini, line 105: This looks ugly to me, due to the code repetition. But I couldn't quickly come up with a better alternative - other than maybe having two separate functions, but in the end, they'd still have pretty similar code...

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128237 | 103870 | 81.0% |
| .capture.kernel_drops | 0 | 74581 | 0.01 | 

Pipeline 11168

src/util-exception-policy.c
victorjulien: these can just be merged? Also reject is valid for non-ips

src/util-exception-policy.c
```diff
@@ -69,36 +81,86 @@ void ExceptionPolicyApply(Packet *p, enum ExceptionPolicy policy, enum PacketDro
...
-            policy = EXCEPTION_POLICY_REJECT;
-            SCLogConfig("%s: %s", option, value_str);
-        } else if (strcmp(value_str, "ignore") == 0) { // TODO name?
-            policy = EXCEPTION_POLICY_IGNORE;
+            if (is_ips_mode) {
+                policy = EXCEPTION_POLICY_REJECT;
+                SCLogConfig("%s: %s", option, value_str);
+            } else {
+                policy = EXCEPTION_POLICY_NOT_SET;
+                SCLogConfig("%s: %s not a valid config in IDS mode. Setting it to disabled)",
```
victorjulien, line 149: reject is valid for IDS mode

src/util-exception-policy.c
victorjulien: could some something like
```c
policy = SetIPSOption(option, value_str);
...
enum ExceptionPolicy SetIPSOption(const char *option, const char *value_str, enum ExceptionPolicy p)
{
    if (!EngineModeIsIPS()) {
         SCLogConfig("%s: %s not a valid config in IDS mode. Setting it to disabled)", option, value_str);
         return EXCEPTION_POLICY_NOT_SET;
    }
    return p;
}
```

src/util-exception-policy.c
jufajardini: 🤔

src/util-exception-policy.c
jufajardini: Are the `DROP` options the only ones not available in IDS, then?

jufajardini:
Feedback incorporated in https://github.com/OISF/suricata/pull/8287

PR #8293, opened 2022-12-16
rfb: add unittests

hsadia538:

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5741) ticket:

Describe changes:
- Add tests for `parse_request`, and `parse_response` functions in `rfb.rs`.
- Format `rfb.rs` file

### DIFF
--- a/rust/src/rfb/parser.rs
+++ b/rust/src/rfb/parser.rs
@@ -24,6 +24,7 @@ use nom7::combinator::map_res;
 use nom7::number::streaming::*;
 use nom7::*;
 
+#[derive(Copy, Clone, PartialOrd, PartialEq, Eq, Debug)]
 pub enum RFBGlobalState {
     TCServerProtocolVersion,
     TCSupportedSecurityTypes,
--- a/rust/src/rfb/rfb.rs
+++ b/rust/src/rfb/rfb.rs
@@ -1,4 +1,4 @@
-/* Copyright (C) 2020-2021 Open Information Security Foundation
+/* Copyright (C) 2020-2022 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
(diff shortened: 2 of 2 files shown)


rust/src/rfb/rfb.rs
```diff
@@ -633,3 +633,86 @@ pub unsafe extern "C" fn rs_rfb_register_parser() {
...
+
+        assert_eq!(
+            r,
+            AppLayerResult {
+                status: -1,
+                consumed: 0,
+                needed: 0
+            }
+        );
+    }
```
jufajardini, line 689: nit: line break

rust/src/rfb/rfb.rs
```diff
@@ -633,3 +633,86 @@ pub unsafe extern "C" fn rs_rfb_register_parser() {
...
+    #[test]
+    fn test_request_response_events() {
+        let mut init_state = RFBState::new();
```
jufajardini, line 692: I find it cool how you were able to find a valid buffer to test so many States, but I'm not sure about testing the request and response functions in the same unit test. Let's try to get feedback from some of the others about this bit :)

jufajardini (commented):
Good work!

One minor comment, and something that we need input from other mentors about...

rust/src/rfb/rfb.rs
hsadia538: > I find it cool how you were able to find a valid buffer to test so many States, but I'm not sure about testing the request and response functions in the same unit test. Let's try to get feedback from some of the others about this bit :)

The alternative would be to test either just one request or response in one test and add multiple tests to test each state.
I am testing both functions in one right now mainly because the input to one function is the output of the other so they are interlinked and thought maybe it will be good if the flow is tested. I myself felt that it might be too much for a unit test but wasn't sure about the right approach here.

jufajardini: Your thinking is reasonable, in my opinion! I just don't think I can decide on that one alone...

rust/src/rfb/rfb.rs
```diff
@@ -614,22 +650,97 @@ pub unsafe extern "C" fn rs_rfb_register_parser() {
...
+        ok_state = parser::RFBGlobalState::TSClientInit;
+        assert_eq!(init_state.state, ok_state);
+
+        init_state.parse_request(&buf[35..36]);
+        ok_state = parser::RFBGlobalState::TCServerInit;
+        assert_eq!(init_state.state, ok_state);
+
+        init_state.parse_response(&buf[36..90]);
+        ok_state = parser::RFBGlobalState::Message;
+        assert_eq!(init_state.state, ok_state);
```
inashivb, line 744: Nice work, Haleema! :)
Could you please tell the motivation behind a unit test testing the internal chained states of the parser?
It seems to me that the parser expects an input, runs it in a loop and moves from one state to another as a part of processing the same request.
Passing the parser an input from the middle of a request to validate a state looks like an invalid input to me. But, Idk anything about RFB so maybe it does make sense.. 🤷🏽‍♀️ and hence the question about your motivation. 😉

rust/src/rfb/parser.rs
```diff
@@ -24,6 +24,7 @@ use nom7::combinator::map_res;
 use nom7::number::streaming::*;
 use nom7::*;
 
+#[derive(Copy, Clone, PartialOrd, PartialEq, Eq, Debug)]
```
inashivb, line 27: Which operations does this PR do to use these?

hsadia538: I was getting errors for PartialEq and Debug.

rust/src/rfb/rfb.rs
hsadia538: I think I did this to try to test multiple states in both requests as I wasn't sure how the states are getting updated.
Further going into detail I found that the output of ` parse_request `will go as input to `parse_response` and so on.
The intention for testing the chained states was to test the flow from start to end. 
I was trying to think of it as 1 whole function instead of multiple states and was focusing on the flow.
Let me know what you think about this.

Also, I may still have some gaps in understanding but from what I have understood,when we are in `parse_request` function within the loop, once there is a match between the buffer state, the state gets updated and after that, there will be no match in the same function. The changed state will now have a match in `parse_response` function so the loop won't change the state again and the function just returns it.

rust/src/rfb/parser.rs
jufajardini: Following up on Shivani's question - do we need all of those?

rust/src/rfb/rfb.rs
jufajardini: I _think_  that if your intention is the test the whole flow the best place for that would be Suricata-verify tests...

jufajardini (commented):
Went back to this after a conversation with Haleema, hope this helps with deciding on how to move on :)

rust/src/rfb/rfb.rs
hsadia538: So you are suggesting that I close this PR and write some s-v tests for these functions?

jufajardini: For the purpose of what you were trying to do with this unittest here, yes. If you want to break down the test into smaller units, that is probably also fine and welcome, I would say. But let's try to keep those simpler :)

rust/src/rfb/rfb.rs
```diff
@@ -633,3 +633,86 @@ pub unsafe extern "C" fn rs_rfb_register_parser() {
...
+            r,
+            AppLayerResult {
+                status: -1,
+                consumed: 0,
+                needed: 0
+            }
+        );
+    }
+    #[test]
+    fn test_request_response_events() {
```
jufajardini, line 691: Adding some feedback after discussing this with Victor :)
So, he sees this as testing the state machine for the RFB app-proto, and that's a positive thing, and this is the right place for it. We just need some adjustments, to make it easier to understand ;)
1) Change the test name to something like `test_state_machine`
2) Add a comment explaining what's the test purpose - since it's a more extensive one and its workings are not immediately clear

rust/src/rfb/rfb.rs
```diff
@@ -633,3 +633,86 @@ pub unsafe extern "C" fn rs_rfb_register_parser() {
...
+            0x2e, 0x6c, 0x6f, 0x63, 0x61, 0x6c, 0x64, 0x6f, 0x6d, 0x61, 0x69,
+            0x6e, /* Server framebuffer parameters */
+        ];
+        init_state.parse_response(&buf[0..12]);
+        let mut ok_state = parser::RFBGlobalState::TSClientProtocolVersion;
+        assert_eq!(init_state.state, ok_state);
+
+        init_state.parse_request(&buf[0..12]);
+        ok_state = parser::RFBGlobalState::TCSupportedSecurityTypes;
+        assert_eq!(init_state.state, ok_state);
```
jufajardini, line 716: 3) In situations where you use the same buffer as input for `parse_response` and `parse_request`, add a comment explaining why, as it's not so expected that the same buffer can act as a response and as a request.

jufajardini (changes requested):

rust/src/rfb/rfb.rs
hsadia538: In real-time, they are going to be two separate streams of values. I used the same one here because both corresponded to the protocol version. Is that okay?

jufajardini: Can you elaborate a bit more? I'm not sure I understand what you mean

hsadia538: In real-time, there will be two separate buffers e.g `buff1 = [0x52, 0x46, 0x42, 0x20, 0x30, 0x30, 0x33, 0x2e, 0x30, 0x30, 0x38, 0x0a] and buff2 = [0x52, 0x46, 0x42, 0x20, 0x30, 0x30, 0x33, 0x2e, 0x30, 0x30, 0x38, 0x0a]`
 Buff
Here I am passing the same because of similar values.

jufajardini: I see. So does this mean that this portion of the message header is the same for both request and response? If some, I guess that's alright - so long as you add a comment explaining this.

Is it the same thing that is happening from `15..31`?

hsadia538: yes

jufajardini: Then, same answer :P

hsadia538:
followed by #8392

PR #5762, opened 2021-01-21
Http2 keywords 4067 v8

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/4067

Describe changes:
- Adds signature protocol extending flow protocol (one signature protocol matches multiple flow protocols cf `AppProtoEquals`)
- Adds `ALPROTO_HTTP_ANY` (as such a signature protocol) for signatures to match on both HTTP1 and HTTP2
- Rename `ALPROTO_HTTP` to `ALPROTO_HTTP1` (and `ALPROTO_HTTP` is now for any version)
- `http.uri` keyword now matches on HTTP2 traffic (as well as `http.uri.raw`)
- `http.user_agent` keyword now matches on HTTP2 traffic
- `http.stat_msg` keyword now matches on HTTP2 traffic

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):

suricata-verify-pr: 384

Makes S-V test in https://github.com/OISF/suricata-verify/pull/384 pass

What remains to be done :
- rename `alproto` to `sig_alproto`. Thoughts about a better name ?
- Commits squashing  (commit-check)
- CI (commit-check)
- Do all the keywords

Modifies #5669 by 
- adding a first commit so that `StringToAppProto` recognizes "ftp-data" as `ALPROTO_FTPDATA` (ans thus we can have signatures with this protocol)

PR #8296, opened 2022-12-16
detect/pcre: don't use JIT if disabled

victorjulien:
If we determined not to use JIT at start up, really don't use JIT.

Bug: #5762. https://redmine.openinfosecfoundation.org/issues/5762


suricata-qa:

Pipeline 11235

PR #8287, opened 2022-12-14
exceptions: add exception policy master switch - v4

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5219](https://redmine.openinfosecfoundation.org/issues/5219)

Previous PR: https://github.com/OISF/suricata/pull/8280

Changes from previous PR:
- clean up code, removing redundancy
- allow reject policy for IDS mode
- remove usage of  'disabled' in favour of the already existing 'ignore'

suricata-verify-pr: 1040
https://github.com/OISF/suricata-verify/pull/1040

PR #8310, opened 2022-12-20
exceptions: add exception policy master switch - v5

jufajardini:
Previous PR: #8287 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5219

Describe changes:
- make `SCLogConfig` message more informative
- bring back config warning about unset exception-policy master switch that had gotten lost among PRs
- rebase and resolve conflicts

suricata-verify-pr: 1040
https://github.com/OISF/suricata-verify/pull/1040

### DIFF
--- a/doc/userguide/configuration/exception-policies.rst
+++ b/doc/userguide/configuration/exception-policies.rst
@@ -66,10 +66,10 @@ are:
 - ``pass-flow``: disable payload and packet detection; stream reassembly,
   app-layer parsing and logging still happen.
 - ``pass-packet``: disable detection, still does stream updates and app-layer
-  parsing (depeding on which policy triggered it).
+  parsing (depending on which policy triggered it).
 - ``ignore``: do not apply exception policies (default behavior).
 
-The *Drop*, *pass* and *reject* are similar to the rule actions described in :ref:`rule
+The *drop*, *pass* and *reject* are similar to the rule actions described in :ref:`rule
 actions<suricata-yaml-action-order>`.
 
 Command-line Options for Simulating Exceptions
--- a/doc/userguide/configuration/exception-policies.rst
+++ b/doc/userguide/configuration/exception-policies.rst
@@ -16,6 +16,45 @@ simulate failures or errors and understand Suricata behavior under such conditio
 Exception Policies
 ------------------
 
+.. _master-switch:
+
+Master Switch
+~~~~~~~~~~~~~
+
+It is possible to set all configuration policies via what we call "master
+switch". This offers a quick way to define what the engine should do in case of
+traffic exceptions, while still allowing for the flexibility of indicating a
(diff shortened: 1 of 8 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128237 | 104014 | 81.11% |
| .capture.kernel_drops | 0 | 603564 | 0.10 | 

Pipeline 11269

victorjulien:
I'm a bit confused about SV passing, while it doesn't pass locally. Locally it fails on `detect-udp-flow-rule-01`, giving 2 alerts instead of 4.

The test uses `--simulate-ips` and the default yaml, which would mean drop-flow. It also triggers app-layer errors on packets 6 and 7, and invokes the exception policy? I see no alerts after that, but do see some sip records. Very confusing.

jufajardini:
> I'm a bit confused about SV passing, while it doesn't pass locally. Locally it fails on `detect-udp-flow-rule-01`, giving 2 alerts instead of 4.
> 
> The test uses `--simulate-ips` and the default yaml, which would mean drop-flow. It also triggers app-layer errors on packets 6 and 7, and invokes the exception policy? I see no alerts after that, but do see some sip records. Very confusing.

I'll try to debug that one... Tests pass locally for me, with or without `exception-policy: auto` in the yaml...

jufajardini:
@victorjulien I checked and there are no updates to this from my further work. Do you see anything here that needs to be changed?

victorjulien:
Hmm I see a failure in a private test that uses `--set stream.midstream=true`... do we still have a conflict between that setting and the policy?

victorjulien:
> Hmm I see a failure in a private test that uses `--set stream.midstream=true`... do we still have a conflict between that setting and the policy?

The failure is that the flow is dropped where I expect it to not be dropped.

jufajardini:
> Hmm I see a failure in a private test that uses `--set stream.midstream=true`... do we still have a conflict between that setting and the policy?

We do, I haven't addressed that bug yet. Should I bump its priority?

jufajardini:
> > Hmm I see a failure in a private test that uses `--set stream.midstream=true`... do we still have a conflict between that setting and the policy?
> 
> The failure is that the flow is dropped where I expect it to not be dropped.

Is this because it's a midstream flow? I should probably go address that issue, now, then...

jufajardini:
> > Hmm I see a failure in a private test that uses `--set stream.midstream=true`... do we still have a conflict between that setting and the policy?
> 
> The failure is that the flow is dropped where I expect it to not be dropped.

I've put this work together with the config one: https://github.com/OISF/suricata/pull/8409

victorjulien:
Merged in https://github.com/OISF/suricata/pull/8430, thanks!

PR #8206, opened 2022-11-24
runmodes: dpdk: Move IPS check outside of the core suricata file and add checksum offload to DPDK v3

lukashino:
This is a follow-up of PR #8204

Changes
- Adds a fix to [this Redmine ticket](https://redmine.openinfosecfoundation.org/issues/5706) 

Redmine tickets:
https://redmine.openinfosecfoundation.org/issues/5580
https://redmine.openinfosecfoundation.org/issues/5553
https://redmine.openinfosecfoundation.org/issues/5706

PR #8329, opened 2022-12-27
runmodes: dpdk: Move IPS check outside of the core suricata file and add checksum offload to DPDK v5

lukashino:
This is a follow-up of PR #8206 

Changes
- Rebased to the newest master

Redmine tickets:
https://redmine.openinfosecfoundation.org/issues/5580
https://redmine.openinfosecfoundation.org/issues/5553
https://redmine.openinfosecfoundation.org/issues/5706


lukashino:
With this PR runmode initialization could be `suricata/runmodes.c` (core Suricata) files independent if function `RunModeGetConfOrDefault` would be less strict with capture_mode check and instead of FatalError it would write Error or nothing and return a NULL value (caller would handle further behavior). 
Should I change that or is this behavior good enough (currently `runmodes.c` contain a list of modes supporting IPS mode)?

lukashino:
Sorry, figured out the change mentioned in the comment above could be made without any problems and proceeded with it in #8330

PR #8334, opened 2022-12-29
Rust alproto 3524 v1

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3524

Describe changes:
- move ALPROTO enum to rust
- fixes some new clang warnings by the way (set but not used)

What do you think of this ?


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128237 | 104010 | 81.11% |
| .capture.kernel_drops | 0 | 88952 | 0.02 | 

Pipeline 11332

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 128237 | 104013 | 81.11% |
| .capture.kernel_drops | 0 | 129531 | 0.02 | 

Pipeline 11333

jasonish:
Looks good on a first look.  Good choice on the second cbindgen config.

One thing on the back of my mind is dynamically adding new protocols, if we get there in plugins.  Would that render this enum obsolete at some point?

catenacyber:
> Looks good on a first look. Good choice on the second cbindgen config.
> 
> One thing on the back of my mind is dynamically adding new protocols, if we get there in plugins. Would that render this enum obsolete at some point?

Indeed, dynamically adding new protocols does not seem to match with using a rust enum...

So, should I make these simply constants ?

Is there another way in Rust to have these global variables be defined during some initialization and be constant afterwards ?

jasonish:
> > Looks good on a first look. Good choice on the second cbindgen config.
> > One thing on the back of my mind is dynamically adding new protocols, if we get there in plugins. Would that render this enum obsolete at some point?
> 
> Indeed, dynamically adding new protocols does not seem to match with using a rust enum...
> 
> So, should I make these simply constants ?
> 
> Is there another way in Rust to have these global variables be defined during some initialization and be constant afterwards ?

There is lazy_static: https://docs.rs/lazy_static/latest/lazy_static/

However, not sure if that'll play with cbindgen.

To be fair, the C way isn't that compatible with dynamic protocol registration either.. Which will probably come with 8.

IMO, I'm not sure if we should do this for 7, so it will be revisited for 8 as well.

victorjulien:
I had already set the 8 milestone on this PR. Think we can merge it in 8 even if perhaps we'll make it dynamic later on.

catenacyber:
So, what is the conclusion ?
Should I close this PR ? (until we start work on 8)

victorjulien:
We can merge in 8, look at dynamic id's later.

PR #8342, opened 2023-01-04
Pr 8341 and payload len fix/v1

inashivb:
https://redmine.openinfosecfoundation.org/issues/5379
https://redmine.openinfosecfoundation.org/issues/5693

suricata-verify-pr: 1055

### DIFF
--- a/src/decode-udp.c
+++ b/src/decode-udp.c
@@ -56,11 +56,6 @@ static int DecodeUDPPacket(ThreadVars *t, Packet *p, const uint8_t *pkt, uint16_
         return -1;
     }
 
-    if (unlikely(len != UDP_GET_LEN(p))) {
-        ENGINE_SET_INVALID_EVENT(p, UDP_HLEN_INVALID);
-        return -1;
-    }
-
     SET_UDP_SRC_PORT(p,&p->sp);
     SET_UDP_DST_PORT(p,&p->dp);
 
--- a/src/decode-udp.c
+++ b/src/decode-udp.c
@@ -56,11 +56,16 @@ static int DecodeUDPPacket(ThreadVars *t, Packet *p, const uint8_t *pkt, uint16_
         return -1;
     }
 
+    if (unlikely(UDP_GET_LEN(p) < UDP_HEADER_LEN)) {
+        ENGINE_SET_INVALID_EVENT(p, UDP_PKT_TOO_SMALL);
+        return -1;
+    }
+
     SET_UDP_SRC_PORT(p,&p->sp);
     SET_UDP_DST_PORT(p,&p->dp);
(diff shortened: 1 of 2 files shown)


inashivb:
Wait a second. This is incomplete. The OP had reported this issue with TCP as well so that needs test and fixing too..

src/detect-engine-event.c
```diff
@@ -160,6 +160,9 @@ static DetectEngineEventData *DetectEngineEventParse (const char *rawstr)
 
     if (de->event == STREAM_REASSEMBLY_OVERLAP_DIFFERENT_DATA) {
         StreamTcpReassembleConfigEnableOverlapCheck();
+    } else if (de->event == UDP_HLEN_INVALID) {
+        SCLogWarning("Rule uses decode-event \"udp.hlen_invalid\" which will be deprecated in "
```
lukashino, line 164: Just a note: Suricata 8.0 was not agreed upon, I just put it there as a safe milestone when to remove the keyword. So @victorjulien needs to confirm this. Alternatively, the keyword can be used for a case that Shivani fixed.

lukashino (approved):
Looks good to me

src/decode-udp.c
```diff
@@ -56,16 +56,16 @@ static int DecodeUDPPacket(ThreadVars *t, Packet *p, const uint8_t *pkt, uint16_
         return -1;
     }
 
-    if (unlikely(len != UDP_GET_LEN(p))) {
-        ENGINE_SET_INVALID_EVENT(p, UDP_HLEN_INVALID);
+    if (unlikely(UDP_GET_LEN(p) < UDP_HEADER_LEN)) {
+        ENGINE_SET_INVALID_EVENT(p, UDP_PKT_TOO_SMALL);
```
victorjulien, line 60: the issue here is that the length field isn't correct, not that the packet is too small?

inashivb: Yes. HLEN_INVALID also did not seem to fit.. I think in my head, I went with this bc the length received from the pkt is smaller than 8bytes..
New error code?

victorjulien: The pkt too small class of events is meant to indicate the input data is too small, not that a length field is wrong

inashivb: Indeed. That was an incorrect event. So is UDP_HLEN_INVALID and UDP_HLEN_TOO_SMALL, would you agree? I'm thinking these indicate that the packet came with an invalid or too small header length. This event is however about the entire length of the packet.

Should we make a new event for this case?

victorjulien: yeah I think so, something like `UDP_LEN_INVALID` or something

PR #8345, opened 2023-01-05
github-ci: check for new authors

jasonish:
Add a GitHub CI job that checks the commits in a pull requests for new
authors that have not been seen before. If a new author is found, the
job will contain some error annotations with the new authors and a
comment will be added to the pull request.

Example pull request with new authors:
https://github.com/jasonish/suricata/pull/148

Job annotations:
https://github.com/jasonish/suricata/actions/runs/3848850381

### DIFF
--- a/.github/workflows/authors.yml
+++ b/.github/workflows/authors.yml
@@ -1,0 +1,57 @@
+name: New Author Check
+
+on:
+  - pull_request
+
+jobs:
+  check-id:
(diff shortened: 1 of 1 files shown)


jasonish:
Turns out a pull request is not run in the correct context to have enough permissions to leave a comment on the PR, at least not easily.  Need to look at work-around here: https://stackoverflow.com/a/71683208

PR #8361, opened 2023-01-10
github-ci: Check for new authors - v8

jasonish:
Add a GitHub CI job that checks the commits in a pull requests for new
authors that have not been seen before. If a new author is found, the
job will error out with the unknown authors listed as error annotations.

Note the list of previous authors is taken from the master branch, so
there is a chance an author might not be detected as an existing author.

This is a job that simply fails that has the unfortunate affect of making the
PR look like it failed.  The various continue on fail options has more to do
with running subsequent jobs in a work flow, but not actually letting a work
flow fail for informational purposes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,3 +111,4 @@ A: If you really think it is, we can discuss how to improve it. But don't come t
 __Q: do you require signing of a contributor license agreement?__
 
 A: Yes, we do this to keep the ownership of Suricata in one hand: the Open Information Security Foundation. See http://suricata.io/about/open-source/ and http://suricata.io/about/contribution-agreement/
+
--- a/.github/workflows/authors.yml
+++ b/.github/workflows/authors.yml
@@ -1,0 +1,34 @@
+name: New Author Check
+
+on:
+  - pull_request
+
+jobs:
+  check-id:
+    name: New Author Check
+    runs-on: ubuntu-latest
+    steps:
+      - run: sudo apt -y install git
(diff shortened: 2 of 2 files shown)


jasonish:
The new authors can be seen by going to the failed job and dropping down the "Check new authors" step, for example: https://github.com/OISF/suricata/actions/runs/3884516524/jobs/6627182072

Or viewing the job summary: https://github.com/OISF/suricata/actions/runs/3884516524

README.md
```diff
@@ -111,3 +111,4 @@ A: If you really think it is, we can discuss how to improve it. But don't come t
 __Q: do you require signing of a contributor license agreement?__
 
 A: Yes, we do this to keep the ownership of Suricata in one hand: the Open Information Security Foundation. See http://suricata.io/about/open-source/ and http://suricata.io/about/contribution-agreement/
+
```
jlucovsky, line 114: nit: not sure why the blank line at the file's end was added.

jlucovsky: > nit: not sure why the blank line at the file's end was added.

nvm; this is your "dummy commit" :-)

victorjulien:
Something doesn't seem right about this, I wanted to merge it here https://github.com/OISF/suricata/pull/8411 but got
image

jasonish:
> Something doesn't seem right about this, I wanted to merge it here #8411 but got image

I'll try using raw git commands instead of the checkout action.  But if that fails, there's probably not much we can do about this.

PR #8362, opened 2023-01-10
rfb: add RFB frames v9

hsadia538:
- [X] I have read the contributing guide lines at https://suricata.readthedocs.io/en/latest/devguide/codebase/contributing/contribution-process.html
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5717

Previous PR: https://github.com/OISF/suricata/pull/8358

Describe changes:

- Fix format issues

suricata-verify-pr: 1059

PR #8364, opened 2023-01-11
rfb: add RFB frames v10

hsadia538:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5717

Previous PR: https://github.com/OISF/suricata/pull/8362

Describe changes:

- Update commit message

suricata-verify-pr: 1059

### DIFF
--- a/rust/src/rfb/rfb.rs
+++ b/rust/src/rfb/rfb.rs
@@ -17,13 +17,13 @@
 
 // Author: Frank Honza <frank.honza@dcso.de>
 
-use std;
-use std::ffi::CString;
-use crate::core::{ALPROTO_UNKNOWN, AppProto, Flow, IPPROTO_TCP};
+use super::parser;
 use crate::applayer;
 use crate::applayer::*;
+use crate::core::{AppProto, Flow, ALPROTO_UNKNOWN, IPPROTO_TCP};
 use nom7::Err;
-use super::parser;
+use std;
+use std::ffi::CString;
 
 static mut ALPROTO_RFB: AppProto = ALPROTO_UNKNOWN;
 
@@ -87,7 +87,7 @@ pub struct RFBState {
     state_data: AppLayerStateData,
     tx_id: u64,
     transactions: Vec<RFBTransaction>,
-    state: parser::RFBGlobalState
+    state: parser::RFBGlobalState,
 }
 
 impl State<RFBTransaction> for RFBState {
@@ -98,7 +98,6 @@ impl State<RFBTransaction> for RFBState {
     fn get_transaction_by_index(&self, index: usize) -> Option<&RFBTransaction> {
         self.transactions.get(index)
     }
-
 }
 
 impl Default for RFBState {
@@ -113,7 +112,7 @@ impl RFBState {
             state_data: AppLayerStateData::new(),
             tx_id: 0,
             transactions: Vec::new(),
-            state: parser::RFBGlobalState::TCServerProtocolVersion
+            state: parser::RFBGlobalState::TCServerProtocolVersion,
         }
     }
 
@@ -185,7 +184,10 @@ impl RFBState {
                             }
                         }
                         Err(Err::Incomplete(_)) => {
-                            return AppLayerResult::incomplete(consumed as u32, (current.len() + 1) as u32);
+                            return AppLayerResult::incomplete(
+                                consumed as u32,
+                                (current.len() + 1) as u32,
+                            );
                         }
                         Err(_) => {
                             return AppLayerResult::err();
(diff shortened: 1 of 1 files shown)


rust/src/rfb/rfb.rs
```diff
@@ -170,6 +179,10 @@ impl RFBState {
                             consumed += current.len() - rem.len();
                             current = rem;
 
+                            if let Some(frame) = &pdu {
```
catenacyber, line 182: Should not there be a new frame at each iteration of the loop ?

rust/src/rfb/rfb.rs
```diff
@@ -231,7 +248,11 @@ impl RFBState {
                         consumed += current.len() - rem.len();
                         current = rem;
 
-                            self.state = parser::RFBGlobalState::TCSecurityResult;
```
catenacyber, line 229: This line seems oof in this commit...

catenacyber:
Is there not a header frame like `RFB 3.8` something ?

catenacyber:
Is there some doc to update ?

rust/src/rfb/rfb.rs
hsadia538: I am registering a PDU frame at the start of both `parse_request` and `parse_response` functions.
It is going to match either one of the match criteria and would go to the other function.

hsadia538:
> Is there not a header frame like `RFB 3.8` something ?

There are a bunch of initialization messages including the handshake etc but no set header for the protocol that's why there is no header frame.

hsadia538:
> Is there some doc to update ?

Which document are you referring to?

rust/src/rfb/rfb.rs
catenacyber: But `parse_request`. does a loop that can create multiple transactions.

If so, there should me multiple frames, right ?

rust/src/rfb/rfb.rs
catenacyber: But with a different indentation ;-)

rust/src/rfb/rfb.rs
hsadia538: Maybe I am still missing something but as far as what I have understood from the working of the function. Once it matches one of the states in `parse_request` the state changes. .e.g once it matches `TSClientProtocolVersion`, the state changes to `TCServerSecurityType` at line 188. once we have this new state. it will only match something in `parse_response` function and will not be parsed in the same in `parse_request`.

Once it goes to `parse_response` a new frame for pdu gets registered.

hsadia538:
> > Which document are you referring to?
> 
> @jufajardini is there a doc about frames and every value supported (like rfb.pdu added here)

this is the documentation related to frames: https://suricata.readthedocs.io/en/latest/devguide/extending/app-layer/app-layer-frames.html?highlight=frame

as a standard(per doc) we use pdu,headers,request,response etc based on what we find in packets but there is no strict rule as to what can be registered as a frame.

rust/src/rfb/rfb.rs
catenacyber: So, does the parser need the `loop` at all ? ie Can it loop more than once ?

hsadia538: I might need more insights into this as well

rust/src/rfb/rfb.rs
hsadia538: it's due to rustfmt , earlier the file did not seem to be formatted with it.

jufajardini:
> > Which document are you referring to?
> 
> @jufajardini is there a doc about frames and every value supported (like rfb.pdu added here)

What Haleema said! Because frames are added in part to ease rule writer's lives, and because app-layer protos vary so much among each other, we've defined the bare minimum, and then other frames can be added as it makes sense - per protocol, and per usefulness. This is one of the aspects where we welcome feedback the most - especially from rule writers. We want to know, from their perspective, which (per proto) fields they would be interested in being able to more easily check with the rule language.

rust/src/rfb/rfb.rs
jufajardini: Sorry for taking so long to answer this, I didn't realize it had a pending discussion going on. I am not familiar with RFB, but from what I can see, it seems that it expects to be able to parse more than one message at once - although I don't understand when would that happen. This seems necessary as of now due to the function logic, though, as we are not returning AppLayerResult::ok() when we successfully parse a message, but only once the input is totally consumed 🤔 

I think the simplest solution to this would be to register the frame as a whole once we have the Ok matches, maybe check how it's done with SMB - which also has it created inside a loop, only that there it's a `while`...

hsadia538: So you are saying at every enum match for an RFB state, I should register the PDU frame instead of at the start?

hsadia538: Also in this scenario I can directly specify the length instead of updating it later as the frame is getting at the same place now e.g
```
` parser::RFBGlobalState::TCServerInit => {
                    match parser::parse_server_init(current) {
                        Ok((rem, request)) => {
                            consumed += current.len() - rem.len();
                            current = rem;
```
**REGISTER FRAME HERE**
```
                            if let Some(frame) = &pdu {
                                frame.set_len(flow, consumed as i64); // THIS GETS REMOVED
                            }`
```

jufajardini: Yes, I think we could try that approach, to prevent having the pdu overwritten in case the parser is indeed able to parse more than one message in one call. (I suspect the loop is there to ensure we have enough data to process one single message, but I'm not sure and the way it's implemented, IMO, leaves room for both scenarios, so better safe than sorry?)

hsadia538:
followed by #8416

rust/src/rfb/rfb.rs
catenacyber: I think one call to the rfb or other app-layer parser should be able to register multiple PDU frames, right @victorjulien ?

PR #8383, opened 2023-01-16
Mqtt frames v3

hsadia538:

Previous PR: #8382 

Describe new changes:
- Fix check rust CI failure

suricata-verify-pr: 1065


hsadia538:
I see checks failing due to suricata verify tests `mqtt-limit-x`.

jufajardini:
> I see checks failing due to suricata verify tests `mqtt-limit-x`.

This will require investigation, to understand how adding frames has impacted on that. When you run SV locally, do these tests also fail? Can you check what's the error output?

jufajardini (commented):
Except for the CI failure - which must be investigated -, these additions look good to me, kudos! :)

While reviewing the SV tests that go along with this contribution, I wondered if it would/could be useful for rulewriters to be able to easily access the MQTT message length as a frame type... (which really is just me wondering what other useful info we could get readily available via frames) 🤔

hsadia538:
> > I see checks failing due to suricata verify tests `mqtt-limit-x`.
> 
> This will require investigation, to understand how adding frames has impacted on that. When you run SV locally, do these tests also fail? Can you check what's the error output?

The fail locally as well and I see the following error logged in the output for mqtt-limit-x:

`thread '<unnamed>' panicked at 'range end index 100009 out of range for slice of length 65536', library/core/src/slice/index.rs:73:5
note: run with` `RUST_BACKTRACE=1` `environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5`

hsadia538:
> Except for the CI failure - which must be investigated -, these additions look good to me, kudos! :)
> 
> While reviewing the SV tests that go along with this contribution, I wondered if it would/could be useful for rulewriters to be able to easily access the MQTT message length as a frame type... (which really is just me wondering what other useful info we could get readily available via frames) 🤔

When you say MQTT message length as a frame type , you mean defining a frame type with called "len" maybe and then making a frame for whole length of the packet? but isn't it the length that corresponds to pdu as well?

jufajardini:
> > Except for the CI failure - which must be investigated -, these additions look good to me, kudos! :)
> > While reviewing the SV tests that go along with this contribution, I wondered if it would/could be useful for rulewriters to be able to easily access the MQTT message length as a frame type... (which really is just me wondering what other useful info we could get readily available via frames) thinking
> 
> When you say MQTT message length as a frame type , you mean defining a frame type with called "len" maybe and then making a frame for whole length of the packet? but isn't it the length that corresponds to pdu as well?

No, more like a frame allowing rule writers to check what's the value of the message length. Maybe that's just useless 😅

hsadia538:
> No, more like a frame allowing rule writers to check what's the value of the message length. Maybe that's just useless 😅

I think I am not understanding it correctly. We have a message length field as part of mqtt header as well.

jufajardini:
> > No, more like a frame allowing rule writers to check what's the value of the message length. Maybe that's just useless sweat_smile
> 
> I think I am not understanding it correctly. We have a message length field as part of mqtt header as well.

Yes, as I said, it's probably useless, in reality. But just to clarify, as far as I understand, having the length in the mqtt header doesn't mean it's easily available from the rule language, unless we register a sticky buffer (or a frame type) for it. So, as of now, someone who wanted to check that, would have to check for that content at the right offset from the header start, in order to access it.

(does it make sense to you?)

Don't worry about the `length` frame idea, I don't think it was good. ;)

jufajardini:
> > > I see checks failing due to suricata verify tests `mqtt-limit-x`.
> > 
> > 
> > This will require investigation, to understand how adding frames has impacted on that. When you run SV locally, do these tests also fail? Can you check what's the error output?
> 
> The fail locally as well and I see the following error logged in the output for mqtt-limit-x:
> 
> `thread '<unnamed>' panicked at 'range end index 100009 out of range for slice of length 65536', library/core/src/slice/index.rs:73:5 note: run with` `RUST_BACKTRACE=1` `environment variable to display a backtrace fatal runtime error: failed to initiate panic, error 5`

Sorry, I missed this reply when I checked this last. Shall we check this together on our next call? I think this might be easier. What I am imagining is that somehow the way that things are happening now is invading/consuming data from some other mqtt field. But we'll have to look more carefully...

hsadia538:
> > > > I see checks failing due to suricata verify tests `mqtt-limit-x`.
> > > 
> > > 
> > > This will require investigation, to understand how adding frames has impacted on that. When you run SV locally, do these tests also fail? Can you check what's the error output?
> > 
> > 
> > The fail locally as well and I see the following error logged in the output for mqtt-limit-x:
> > `thread '<unnamed>' panicked at 'range end index 100009 out of range for slice of length 65536', library/core/src/slice/index.rs:73:5 note: run with` `RUST_BACKTRACE=1` `environment variable to display a backtrace fatal runtime error: failed to initiate panic, error 5`
> 
> Sorry, I missed this reply when I checked this last. Shall we check this together on our next call? I think this might be easier. What I am imagining is that somehow the way that things are happening now is invading/consuming data from some other mqtt field. But we'll have to look more carefully...

Sure

hsadia538:
followed by #8506

PR #8326, opened 2022-12-26
Multipart mime 3487 v9

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3487

Describe changes:
- convert HTTP to use new rust mime parser
- convert SMTP to use new rust mime parser

Follows #8299 with needed rebase

Still to do : review !
Should I list my questions ?

suricata-verify-pr: 1035

PR #8384, opened 2023-01-16
Multipart mime 3487 v11

catenacyber:

Describe changes:

Follows #8326 with more tolerant base64 decoding

Still to do : 
- get answers for https://github.com/OISF/suricata/pull/8326 questions
- wait for QA rebase with bigger `request-body-limit`

suricata-verify-pr: 1035

QA SURI_TLPW1_files_sha256 showed a difference in SMTP extracted files.
There were at least two problems :
- spaces in base64 lines are ignored/skipped in current implementation : strict rust base64 crate decoding errored on them
- base64 lines do not always complete blocks. Using a [DecoderReader](https://docs.rs/base64/latest/base64/read/struct.DecoderReader.html#) and providing it with input slices each time we get one was not possible to me without adding a `get_mut` method to base64 crate (doing as is done for the HTTP2 decompression with a cursor).

Custom base64 decoding was easy

QA SURI_TLPW1_files_sha256 will still show a difference about HTTP files, because there are files bigger than default suricata.yaml `request-body-limit` and the files are not "chunked" the same way (this PR seems more complete than current master). Solution would be to have QA use a higher `request-body-limit` to have complete files (but it will need a QA rebase because hashes will be updated)


suricata-qa:
ERROR:
 

Pipeline 11631

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8385

PR #8400, opened 2023-01-18
detect-lua: Convert unittests to FAIL/PASS API

Rai0Catodic0:
Ticket: [4051](https://redmine.openinfosecfoundation.org/issues/4051)

- [ :white_check_mark: ] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [ :white_check_mark: ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/

Describe changes:
- Convert tests to use the FAIL/PASS macro

### DIFF
--- a/src/detect-lua.c
+++ b/src/detect-lua.c
@@ -1197,7 +1197,6 @@ static int LuaMatchTest01(void)
         "end\n"
         "return 0\n";
     char sig[] = "alert http any any -> any any (flow:to_server; lua:unittest; sid:1;)";
-    int result = 0;
     uint8_t httpbuf1[] =
         "POST / HTTP/1.1\r\n"
         "Host: www.emergingthreats.net\r\n\r\n";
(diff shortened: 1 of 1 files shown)


src/detect-lua.c
```diff
@@ -1951,64 +1780,39 @@ static int LuaMatchTest03a(void)
...
-        printf("buffer mismatch: ");
-        goto end;
-    }
+    FAIL_IF(memcmp(fv->data.fv_str.value, "2", 1) != 0);
 
-    result = 1;
-end:
     if (de_ctx != NULL)
-        DetectEngineCtxFree(de_ctx);
+        DetectEngineCtxFree(de_ctx); // should this be kept ?
```
Rai0Catodic0, line 1809: Hi, sorry for taking so long to send a new pr.
I have a question about this section if I remove these two lines (1808,1809) half of the tests fail. I didn't quite understand why, did I made some mistake previous to this and it should be fix ? 

Thanks : )

jufajardini: Hello, we understand that volunteer work can be like that, sometimes. If something is urgent, we would probably have kindly told you that someone on the team would be taking it over, and thanked you for your efforts up to that point. That said, thanks for getting back to it! :)

About your question, now:
I think my previous feedback was misleading, I apologize. We must still call `DetectEngineCtxFree(de_ctx)` and the following clean-up routines. We just don't need the `if (de_ctx != NULL)` check, as this part will have been covered already if we reach this point in the test. Does that make sense?

src/detect-lua.c
```diff
@@ -1243,87 +1242,48 @@ static int LuaMatchTest01(void)
     StreamTcpInitConfig(true);
 
     DetectEngineCtx *de_ctx = DetectEngineCtxInit();
-    if (de_ctx == NULL) {
-        goto end;
-    }
+    FAIL_IF_NULL(de_ctx);
     de_ctx->flags |= DE_QUIET;
 
     s = DetectEngineAppendSig(de_ctx, sig);
```
jufajardini, line 1248: When possible, we now prefer to declare the variable in its first usage. So, instead of having `Signature *s = NULL;` on line 1212, you'd declare it here, directly. 

This also applies to Packet variables.

src/detect-lua.c
```diff
@@ -1243,87 +1242,48 @@ static int LuaMatchTest01(void)
...
-        goto end;
-    }
+    FAIL_IF(fv->data.fv_str.value_len != 1);
 
-    result = 1;
-end:
-    if (alp_tctx != NULL)
-        AppLayerParserThreadCtxFree(alp_tctx);
-    if (de_ctx != NULL)
-        DetectEngineCtxFree(de_ctx);
```
jufajardini, line 1318: After all `FAIL`  checks have been passed, we still have to call `AppLayerParserThreadCtxFree(alp_tctx);` and `DetectEngineCtxFree(de_ctx);` as answered elsewhere. 

That's because in this case, we know the test passed, and therefore we have to call all the cleaning routines.

src/detect-lua.c
```diff
@@ -1243,87 +1242,48 @@ static int LuaMatchTest01(void)
...
+    FAIL_IF(memcmp(fv->data.fv_str.value, "2", 1) != 0);
 
     StreamTcpFreeConfig(true);
-    FLOW_DESTROY(&f);
```
jufajardini, line 1321: This one also stays, I reckon.

jufajardini (changes requested):
Hello, I have answered your question and pointed out a couple of other places where similar situations are probably leading to the tests' failures. See if addressing these solves the issue! :)

jufajardini:
I have just noticed that there is a work in progress to clean up most of the tests in `detect-lua.c` and convert them to Suricata-verify tests. I sincerely apologise for the duplicated efforts this has caused. This likely happened due to the fact that sometimes in the past have created tickets that would list several files to be worked on - which greatly removed visibility of what was really assigned to someone or not.
This is the ticket I am referring to: https://redmine.openinfosecfoundation.org/issues/4911. As one of the mentors, I am sorry for not noticing this when the work on this task started. 
I recommend that any work here is put to a halt while I figure out what tests would be covered by the SV ones, and if anything would be left to be converted to the FAIL/PASS API. (Here's the PR where the work is happening: https://github.com/OISF/suricata-verify/pull/1086)

Rai0Catodic0:
Closed in favour of a newer PR: #8517 8517

PR #8401, opened 2023-01-18
pcap-log: fix output filenames when reading from pcap files - v1

jasonish:
Remove early opening of output files if running in an offline mode, as
we don't yet know the timestamp to use.

Prevents the first pcap files being opened with a timestamp of 0,
bringing us back to the same behvaviour of pcap logging in 6.0.

Other commits:
- log-pcap: display mininum limit on error
- log-pcap: fix typo in multi-mode error message

Issue: https://redmine.openinfosecfoundation.org/issues/5374

suricata-verify-pr: 1066

### DIFF
--- a/src/log-pcap.c
+++ b/src/log-pcap.c
@@ -1106,11 +1106,16 @@ static TmEcode PcapLogDataInit(ThreadVars *t, const void *initdata, void **data)
 #endif /* INIT_RING_BUFFER */
     }
 
-    if (pl->mode == LOGMODE_MULTI) {
-        PcapLogOpenFileCtx(td->pcap_log);
-    } else {
-        if (pl->filename == NULL) {
-            PcapLogOpenFileCtx(pl);
+    /* Don't early initialize output files if in a PCAP file mode. */
+    bool is_offline =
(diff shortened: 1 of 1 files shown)


src/log-pcap.c
```diff
@@ -1106,11 +1106,16 @@ static TmEcode PcapLogDataInit(ThreadVars *t, const void *initdata, void **data)
...
     }
 
-    if (pl->mode == LOGMODE_MULTI) {
-        PcapLogOpenFileCtx(td->pcap_log);
-    } else {
-        if (pl->filename == NULL) {
-            PcapLogOpenFileCtx(pl);
+    /* Don't early initialize output files if in a PCAP file mode. */
+    bool is_offline =
+            RunmodeGetCurrent() == RUNMODE_PCAP_FILE || RunmodeGetCurrent() == RUNMODE_UNIX_SOCKET;
```
victorjulien, line 1111: elsewhere we use `IsRunModeOffline(RunmodeGetCurrent())`

jasonish: > elsewhere we use `IsRunModeOffline(RunmodeGetCurrent())`

Fixed here: https://github.com/OISF/suricata/pull/8403

PR #8415, opened 2023-01-20
netmap: Fixup issues with v14+ backport

jlucovsky:
This commit reduces the changes associated with adding the v14 api to 6.0.x

During the preparation of this commit, issues in the original backport were corrected

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5744](https://redmine.openinfosecfoundation.org/issues/5744)

Describe changes:
- Failure to release a lock under error conditions
- Typo in an CPP ifdef
- Incorrect target for goto statement in an error handling case.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/runmode-netmap.c
+++ b/src/runmode-netmap.c
@@ -53,7 +53,7 @@
 #include "util-ioctl.h"
 #include "util-byte.h"
 
-#ifdef HAVE_NETMAP
+#if HAVE_NETMAP && USE_NEW_NETMAP_API
 #define NETMAP_WITH_LIBS
 #include <net/netmap_user.h>
 #endif /* HAVE_NETMAP */
@@ -70,14 +70,11 @@ const char *RunModeNetmapGetDefaultMode(void)
 void RunModeIdsNetmapRegister(void)
 {
 #if HAVE_NETMAP
-    SCLogInfo("Using netmap version %d ["
 #if USE_NEW_NETMAP_API
-              "new"
-#else
-              "legacy"
-#endif
+    SCLogInfo("Using netmap version %d"
               " API interfaces]",
             NETMAP_API);
+#endif
     RunModeRegisterNewRunMode(RUNMODE_NETMAP, "single",
             "Single threaded netmap mode",
             RunModeIdsNetmapSingle);
@@ -127,13 +124,15 @@ static int ParseNetmapSettings(NetmapIfaceSettings *ns, const char *iface,
         }
     }
 
+#ifdef USE_NEW_NETMAP_API
     /* we will need the base interface name for later */
     char base_name[IFNAMSIZ];
     strlcpy(base_name, ns->iface, sizeof(base_name));
     if (strlen(base_name) > 0 &&
-            (base_name[strlen(base_name) - 1] == '^' || base_name[strlen(base_name) - 1] == '*')) {
-        base_name[strlen(base_name) - 1] = '\0';
     }
+#else
+    char *base_name = ns->iface;
+#endif
 
     /* prefixed with netmap or vale means it's not a real interface
      * and we don't check offloading. */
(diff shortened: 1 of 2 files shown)


src/runmode-netmap.c
```diff
@@ -127,13 +124,15 @@ static int ParseNetmapSettings(NetmapIfaceSettings *ns, const char *iface,
         }
     }
 
+#ifdef USE_NEW_NETMAP_API
     /* we will need the base interface name for later */
     char base_name[IFNAMSIZ];
     strlcpy(base_name, ns->iface, sizeof(base_name));
     if (strlen(base_name) > 0 &&
-            (base_name[strlen(base_name) - 1] == '^' || base_name[strlen(base_name) - 1] == '*')) {
```
bmeeks8, line 134: This change results in a syntax error during compilation because removing line 134 deletes the opening brace "{" for the if() statement. Furthermore, I don't believe lines 134 and 135 should be deleted.

src/runmode-netmap.c
```diff
@@ -53,7 +53,7 @@
 #include "util-ioctl.h"
 #include "util-byte.h"
 
-#ifdef HAVE_NETMAP
+#if HAVE_NETMAP && USE_NEW_NETMAP_API
```
bmeeks8, line 56: This change will produce a later compilation error as both the legacy netmap API code and the new v14 multiple host stack rings code needs the <netmap_user.h> header included. Both also need the NETMAP_WITH_LIBS conditional defined.

suricata-qa:
Information:
 

ERROR: QA failed on SURI_TLPR1_alerts_cmp.
|--|--|--|--| 
| .capture.kernel_drops | 0 | 2692666 | 0.46 | 

Pipeline 11806

victorjulien:
Given the ongoing issues I wonder if we should simply revert the v14 backport.

jlucovsky:
> Given the ongoing issues I wonder if we should simply revert the v14 backport.

I support the revert.

bmeeks8:
> Given the ongoing issues I wonder if we should simply revert the v14 backport.

Granted there does seem to be a potential issue with the v14 API on FreeBSD at least. But with that said, the current netmap API used in Suricata 6.0.8 and earlier has been marked as deprecated by upstream. The legacy API in Suricata does not, and will never, support multiple host stack rings based on what I read in the upstream include file (specifically in <libnetmap.h>. So if the v14 API code is reverted, and netmap support stays essentially at the same API level as existed in 6.0.8 and earlier, then Suricata 6.x will be dependent upon deprecated netmap code.

Also, does not version 7.0 of Suricata contain more or less the same new netmap v14 code that is being discussed in this pull request? Would the intent be to remove the new netmap code from 7.x as well, or revert just the backport to the 6.x branch?

I personally suspect the issue referenced in [Redmine Issue 5744](https://redmine.openinfosecfoundation.org/issues/5744) is a thread lock of some type. As evidenced by the flow manager issue with FreeBSD that was identified and corrected in summer of 2021, FreeBSD is much more picky about thread ownership when locking/unlocking a mutex than other operating systems such as Linux. Also, there are currently open issues against at least one Intel driver with respect to netmap lockups in FreeBSD.

My vote would be to expend some energy testing fairly heavily on current FreeBSD versions to see if the bug can be reliably replicated. It seems to have surfaced for the user posting in [Issue 5744](https://redmine.openinfosecfoundation.org/issues/5744) with high traffic loads such as when performing Internet speed tests at gigabit and higher speeds. More sets of eyes looking at the new v14 API code paying particular attention to potential thread locking issues may be beneficial to locating the bug.

src/runmode-netmap.c
jlucovsky: Will fix.

jlucovsky:
Continued in #8432

PR #8417, opened 2023-01-20
ftp: add events for request/response command being too long - v1

jasonish:
This adds events for when a FTP request or response command is too long
requiring truncation.

I wanted to use the Rust "generics" for the boiler plate of defining app-layer
events, however the way we export enums from Rust to C wasn't very compatible
with the auto-naming of app-layer events into a scheme that fits our current
convention.

To fix this, I modified the cbindgen configuration to do enum naming from
idiomatic Rust names to C style names, which are more to our naming convention
for C enums. As this renamed all export enum fields the commit is quite large.

Once that was done, the FTP events could be added with Rust.

Finally I went looking for SID range documentation in-tree and didn't find any,
and I feel it should be there.

suricata-verify-pr: 1072

### DIFF
--- a/rules/Makefile.am
+++ b/rules/Makefile.am
@@ -7,6 +7,7 @@ dhcp-events.rules \
 dnp3-events.rules \
 dns-events.rules \
 files.rules \
+ftp-events.rules \
 http-events.rules \
 http2-events.rules \
 ipsec-events.rules \
--- a/rules/README.md
+++ b/rules/README.md
@@ -1,0 +1,34 @@
+# Suricata Reserved SID Allocations
+
+Unless otherwise noted, each component or protocol is allocated 1000
+signature IDs.
+
+## Components
+
+| Component         | Start   | End     |
+| ----------------- | ------- | ------- |
+| Decoder           | 2200000 | 2200999 |
+| Stream            | 2210000 | 2210999 |
+| Generic App-Layer | 2260000 | 2260999 |
+
+## App-Layer Protocols
+
+| Protocol | Start   | End     |
+| -------- | ------- | ------- |
+| SMTP     | 2220000 | 2220999 |
+| HTTP     | 2221000 | 2221999 |
+| NTP      | 2222000 | 2222999 |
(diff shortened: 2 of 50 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .app_layer.error.tls.parser | 931 | 1145 | 122.99% |
| .app_layer.flow.http | 3118951 | 3340800 | 107.11% |
| .app_layer.flow.tls | 510462 | 623355 | 122.12% |
| .app_layer.flow.ssh | 8937 | 11032 | 123.44% |
| .app_layer.flow.telnet | 1624 | 2334 | 143.72% |
| .app_layer.error.ftp-data.gap | 0 | 1 | - |
| .capture.kernel_drops | 0 | 492621 | 0.08 | 

Pipeline 11818

victorjulien:
While it makes the code look nicer and more idiomatic, I think this has a important drawback: which is that something like `git grep SSH_CONNECTION_STATE_BANNER_DONE` will now not give a quick view into where things happen for that state, but instead we'll need to grep twice (once for ` SSH_CONNECTION_STATE_BANNER_DONE` and once for `BannerDone`) and there will be no single place where the two are obviously mapped.

jasonish:
> While it makes the code look nicer and more idiomatic, I think this has a important drawback: which is that something like `git grep SSH_CONNECTION_STATE_BANNER_DONE` will now not give a quick view into where things happen for that state, but instead we'll need to grep twice (once for ` SSH_CONNECTION_STATE_BANNER_DONE` and once for `BannerDone`) and there will be no single place where the two are obviously mapped.

Hmm.. Good point.  I'll probably have to back peddle on this and use a C implementation of `get_event_info` and `get_event_info_by_id` as the alert name generation doesn't work well with enum's with a stutter in them.... Or what might have been simpler is just a Rust method for each event type exported to C to not have to worry about exporting the enum at all.  Thats what I think I'll go with.

jasonish:
Replaced by https://github.com/OISF/suricata/pull/8425.

PR #8441, opened 2023-01-25
Http2 range leak 5808 v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5808

Describe changes:
- Fix leak with HTTP2 range files
- re-rustfmt http2 code (I should make a CI check)


victorjulien:
I merged the fix in #8459, thanks. Wrt the reformat, I would like to do a full reformat as soon as 8dev opens and then enforce it from then.

PR #8472, opened 2023-01-27
draft: Backport/bug/5793 udp invalid event v1

lukashino:
If the packet is shorter than IP payload length we no longer flag it as an invalid UDP packet. UDP packet can be therefore shorter than IP payload. Keyword "udp.hlen_invalid" became outdated as we no longer flag short UDP packets as invalid. The keyword's evaluation remains the same.

Inspired by commit: d18e52ed93e996bc0335d4a100b2ac7f12b3848d

Redmine ticket: #5793

suricata-verify-pr: 1078


inashivb:
Pulled in https://github.com/OISF/suricata/pull/8475 Thank you! :)

PR #8491, opened 2023-01-30
rules/decoder: fix sid for udp.len_invalid rule

inashivb:


victorjulien:
Does the fact that this passes w/o SV test mean that we have no SV test that checks for this rule?

inashivb:
> Does the fact that this passes w/o SV test mean that we have no SV test that checks for this rule?

oh. we do. but it's not merged yet. https://github.com/OISF/suricata-verify/pull/1099
will fix..

victorjulien:
Merged in #8501, thanks!

PR #8499, opened 2023-01-30
Next/60x/20230130/v5

victorjulien:
Staging:
- #8490 
- #8492 

suricata-verify-pr: 1101

PR #8498, opened 2023-01-30
requirements: suricata-update 1.2.7

jasonish:
Use Suricata-Update 1.2.7.

PR #8492, opened 2023-01-30
fuzz: checks pcap timestamp consistency (master-6.0.x)

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55433&q=label%3AProj-suricata

Describe changes:
- fuzz: checks pcap timestamp consistency

No cherry-picking but taking what is what is good in the current master version

PR #8490, opened 2023-01-30
Udp len invalid/60x/v3

inashivb:
Previous PR: https://github.com/OISF/suricata/pull/8475

suricata-verify-pr: 1100

Changes since v2:
- Update the sid in the rules to the next expected in the file

PR #8500, opened 2023-01-30
Next/60x/20230130/v6

victorjulien:
Staging:
- #8490 
- #8492
- #8498 
- #8499

suricata-verify-pr: 1101

### DIFF
--- a/src/detect-engine-event.c
+++ b/src/detect-engine-event.c
@@ -110,6 +110,14 @@ static int DetectEngineEventMatch (DetectEngineThreadCtx *det_ctx,
     SCReturnInt(0);
 }
 
+static bool OutdatedEvent(const char *raw)
+{
+    if (strcmp(raw, "decoder.udp.hlen_invalid") == 0) {
+        return true;
+    }
+    return false;
+}
+
 /**
  * \brief This function is used to parse decoder events options passed via decode-event: keyword
  *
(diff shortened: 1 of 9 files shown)


suricata-qa:

Pipeline 12153

src/detect-engine-event.c
```diff
@@ -173,8 +173,13 @@ static DetectEngineEventData *DetectEngineEventParse (const char *rawstr)
     }
 
     if (OutdatedEvent(rawstr)) {
-        SCLogWarning(
-                SC_WARN_DEPRECATED, "decode-event keyword no longer supports event \"%s\"", rawstr);
+        if (SigMatchStrictEnabled(DETECT_DECODE_EVENT)) {
+            SCLogError(SC_WARN_DEPRECATED, "decode-event keyword no longer supports event \"%s\"",
+                    rawstr);
+            goto error;
```
inashivb, line 179: Do we want existing rules to start failing in strict mode?

src/detect-engine-event.c
```diff
@@ -173,8 +173,13 @@ static DetectEngineEventData *DetectEngineEventParse (const char *rawstr)
...
+        } else {
+            SCLogNotice("decode-event keyword no longer supports event \"%s\"", rawstr);
```
inashivb, line 181: Also, is the Notice intentional?

src/detect-engine-event.c
victorjulien: yes, it only happens with `--strict-rule-keywords=decode-event`, not with regular `-T`

src/detect-engine-event.c
victorjulien: Yeah, wanted to reduce the severity as warning is known to be treated similar to error by some. In 7 it is still a warning.

PR #8513, opened 2023-02-02
Mqtt frames v7

hsadia538:

Previous PR: #8508 

Describe new changes:
- Fix check rust CI Failure.

suricata-verify-pr: 1065


jufajardini:
CI Failures for Debian 9 and CentOS 7 are unrelated to PR changes

rust/src/mqtt/mqtt.rs
```diff
@@ -589,6 +624,64 @@ impl MQTTState {
...
+    ) {
+        let hdr = stream_slice.as_slice();
+        let hdr_length = input.skipped_length - msg.header.remaining_length as usize;
+        let _mqtt_hdr = Frame::new(
+            flow,
+            stream_slice,
+            hdr,
+            hdr_length as i64,
+            MQTTFrameType::Header as u8,
+        );
```
jufajardini, line 656: I'm curious to see what we will see in these frames.

Analizing the code some more, it seems to me that the Truncated message will have all header info, but no payload parsed by Suricata.
So, I'm thinking that maybe this is a special Frame type where folks only have access to info in the header, or, if we really wanted to give some insight into the (partial) truncated data, we could maybe use the remainder stream of the parsed message to register the data frame, and in that case the length would be mqtt max msg length.

rust/src/mqtt/mqtt.rs
```diff
@@ -589,6 +624,64 @@ impl MQTTState {
...
+
+            let rem_length = current.len() - hdr_length;
+            let data = &hdr[hdr_length..current.len()];
+            let _mqtt_data = Frame::new(
+                flow,
+                stream_slice,
+                data,
+                rem_length as i64,
+                MQTTFrameType::Data as u8,
+            );
```
jufajardini, line 670: Following my previous comment, I think that if we really want to offer any insight into this truncated data, we would have to pass `rem` (https://github.com/OISF/suricata/blob/master/rust/src/mqtt/mqtt.rs#L457) to `mqtt_frames_trunc` (cf https://github.com/OISF/suricata/blob/master/rust/src/mqtt/parser.rs#L663)... 🤔

rust/src/mqtt/mqtt.rs
jufajardini: Now, one of the questions is what is interesting/important for us to expose to `mqtt` frames...

jufajardini (changes requested):
I'm curious to what the SV tests for truncated mqtt messages will show in the frames. I've left some comments with my considerations as to why...

Apart from that, a couple of (possibly nit) inline comments.

rust/src/mqtt/mqtt.rs
```diff
@@ -589,6 +624,64 @@ impl MQTTState {
         tx.tx_data.set_event(event as u8);
         self.transactions.push_back(tx);
     }
+
+    fn mqtt_frames(&mut self, flow: *const Flow, stream_slice: &StreamSlice, input: &MQTTMessage) {
```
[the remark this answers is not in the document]
hsadia538: I will try to come up with something for this

rust/src/mqtt/mqtt_property.rs
```diff
@@ -24,7 +24,7 @@ use nom7::IResult;
 
 // TODO: It might be useful to also add detection on property presence and
 // content, e.g. mqtt.property: AUTHENTICATION_METHOD.
-#[derive(Debug, PartialEq, PartialOrd)]
+#[derive(Debug, PartialEq, Eq, PartialOrd)]
```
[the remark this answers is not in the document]
hsadia538: this was added due to cargo clippy when I was having the issue. I will try to remove it and see if it works this time.

rust/src/mqtt/mqtt.rs
hsadia538: so basically as far as I have understood when the length of parsed input message > max_mg_len then there are two conditions

1. Where it is greater than the current buffer meaning data exceeds the boundary of parsed as well as the current buffer. here I am trying to parse header and whatever remaining length is in current buffer I call it data
2. Where it exceeds the parsed message buffer but still lies within the boundary of the current overall buffer. here I use the skipped length to first extract the header and then remain portion of skipped length is data

hsadia538: > Now, one of the questions is what is interesting/important for us to expose to `mqtt` frames...

All of this brings me back to your last comment ^

rust/src/mqtt/mqtt.rs
hsadia538: This might work too. but I am passing the current buffer which also represents the full input buffer. I have based my length calculations on this.
Do you still want me to look into this?

rust/src/mqtt/mqtt.rs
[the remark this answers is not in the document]
hsadia538: how about `mqtt_hdr_data_frames` ?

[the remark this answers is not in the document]
jufajardini: To me, this gives the idea of `header data` frames, instead of header and data frames 🤔  not sure if this is premature optimization or not.

rust/src/mqtt/mqtt.rs
jufajardini: From running the tests locally, your approach seems to be working. So I think we can move on ;)

rust/src/mqtt/mqtt.rs
jufajardini: I have left a review on the tests PR that adds some input into this convo.

rust/src/mqtt/mqtt.rs
hsadia538: Based on your comment on the test frames PR , I will make a new PR for frames to add trunc_data frames

rust/src/mqtt/mqtt.rs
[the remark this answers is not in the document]
hsadia538: how about `mqtt_hdr_and_data_frames` ?

hsadia538:
followed by #8540

PR #8540, opened 2023-02-13
Mqtt frames v8

hsadia538:

Previous PR: #8508 

Describe new changes:
- Add` trunc.data` frames to cater for truncated packets

suricata-verify-pr: 1123

### DIFF
--- a/rust/src/mqtt/mqtt.rs
+++ b/rust/src/mqtt/mqtt.rs
@@ -481,12 +481,24 @@ impl MQTTState {
                             current.len()
                         );
                         if trunc.skipped_length >= current.len() {
-                            self.mqtt_hdr_and_data_frames_trunc(flow, &stream_slice, trunc, current, &msg);
+                            self.mqtt_hdr_and_data_frames_trunc(
+                                flow,
+                                &stream_slice,
+                                trunc,
+                                current,
+                                &msg,
+                            );
                             self.skip_request = trunc.skipped_length - current.len();
                             self.handle_msg(msg, true);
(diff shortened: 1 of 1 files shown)


rust/src/mqtt/mqtt.rs
```diff
@@ -41,6 +42,14 @@ static mut MQTT_MAX_TX: usize = 1024;
 
 static mut ALPROTO_MQTT: AppProto = ALPROTO_UNKNOWN;
 
+#[derive(AppLayerFrameType)]
+pub enum MQTTFrameType {
+    Pdu,
+    Header,
+    Data,
+    TruncData,
```
inashivb, line 50: Following the discussion that happened privately, this data frame type does not seem to be necessary as we cannot annotate something that does not exist. Even if the data is truncated, the parsed parts can be tagged with `Pdu`, so, we can get rid of this. Make sure to add tests corresponding to truncated data. e.g.

```
aaaaaaaaaaaaaaabbbbbbbbbb 
```
where `a..a` == data exactly the length limit
`b..b` == extra data that should ideally be truncated by the parser.
So, you have three tests.
1. Well within the boundary.
check for say first 4B of data in `Pdu`, it should match.
2. Boundary test. Should match.
check for all `aaaa` exactly up to the length limit.
3. Post boundary test. Should not match.
check for `aaaa...bbb` i.e. good data + truncated data

inashivb (changes requested):

hsadia538:
followed by #8730

PR #8565, opened 2023-03-01
detect/iponly: improve performance of IPOnlyMatchPacket

JustinAzoff:

https://redmine.openinfosecfoundation.org/issues/4578

Describe changes:
- Reduce the size of the SigNumArray bitsets
- remove unused DetectEngineIPOnlyThreadCtx

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/detect-engine-iponly.c
+++ b/src/detect-engine-iponly.c
@@ -875,14 +875,24 @@ error:
  */
 void IPOnlyInit(DetectEngineCtx *de_ctx, DetectEngineIPOnlyCtx *io_ctx)
 {
-    io_ctx->tree_ipv4src = SCRadixCreateRadixTree(SigNumArrayFree,
-                                                  SigNumArrayPrint);
-    io_ctx->tree_ipv4dst = SCRadixCreateRadixTree(SigNumArrayFree,
-                                                  SigNumArrayPrint);
-    io_ctx->tree_ipv6src = SCRadixCreateRadixTree(SigNumArrayFree,
-                                                  SigNumArrayPrint);
-    io_ctx->tree_ipv6dst = SCRadixCreateRadixTree(SigNumArrayFree,
-                                                  SigNumArrayPrint);
+    io_ctx->tree_ipv4src = SCRadixCreateRadixTree(SigNumArrayFree, SigNumArrayPrint);
+    io_ctx->tree_ipv4dst = SCRadixCreateRadixTree(SigNumArrayFree, SigNumArrayPrint);
+    io_ctx->tree_ipv6src = SCRadixCreateRadixTree(SigNumArrayFree, SigNumArrayPrint);
+    io_ctx->tree_ipv6dst = SCRadixCreateRadixTree(SigNumArrayFree, SigNumArrayPrint);
+
+    io_ctx->sig_mapping = SCCalloc(1, de_ctx->sig_array_len * sizeof(uint32_t));
+    if (io_ctx->sig_mapping == NULL) {
+        FatalError("Unable to allocate iponly signature tracking area");
+    }
+    io_ctx->sig_mapping_size = 0;
+}
+
+SigIntId IPOnlyTrackSigNum(DetectEngineIPOnlyCtx *io_ctx, SigIntId signum)
+{
+    SigIntId loc = io_ctx->sig_mapping_size;
+    io_ctx->sig_mapping[loc] = signum;
+    io_ctx->sig_mapping_size++;
+    return loc;
 }
 
 /**
@@ -942,6 +952,10 @@ void IPOnlyDeinit(DetectEngineCtx *de_ctx, DetectEngineIPOnlyCtx *io_ctx)
     if (io_ctx->tree_ipv6dst != NULL)
         SCRadixReleaseRadixTree(io_ctx->tree_ipv6dst);
     io_ctx->tree_ipv6dst = NULL;
+
+    if (io_ctx->sig_mapping != NULL)
+        SCFree(io_ctx->sig_mapping);
+    io_ctx->sig_mapping = NULL;
 }
 
 /**
@@ -1026,20 +1040,18 @@ void IPOnlyMatchPacket(ThreadVars *tv,
     for (u = 0; u < src->size; u++) {
         SCLogDebug("And %"PRIu8" & %"PRIu8, src->array[u], dst->array[u]);
 
-        /* The final results will be at io_tctx */
-        io_tctx->sig_match_array[u] = dst->array[u] & src->array[u];
+        uint8_t bitarray = dst->array[u] & src->array[u];
 
         /* We have to move the logic of the signature checking
          * to the main detect loop, in order to apply the
          * priority of actions (pass, drop, reject, alert) */
-        if (io_tctx->sig_match_array[u] != 0) {
+        if (bitarray) {
             /* We have a match :) Let's see from which signum's */
-            uint8_t bitarray = io_tctx->sig_match_array[u];
             uint8_t i = 0;
 
             for (; i < 8; i++, bitarray = bitarray >> 1) {
                 if (bitarray & 0x01) {
-                    Signature *s = de_ctx->sig_array[u * 8 + i];
+                    Signature *s = de_ctx->sig_array[io_ctx->sig_mapping[u * 8 + i]];
 
                     if ((s->proto.flags & DETECT_PROTO_IPV4) && !PKT_IS_IPV4(p)) {
                         SCLogDebug("ip version didn't match");
@@ -1548,10 +1560,13 @@ void IPOnlyAddSignature(DetectEngineCtx *de_ctx, DetectEngineIPOnlyCtx *io_ctx,
     if (!(s->flags & SIG_FLAG_IPONLY))
         return;
 
+    SigIntId mapped_signum = IPOnlyTrackSigNum(io_ctx, s->num);
+    SCLogDebug("Adding IPs from rule: %" PRIu32 " (%s) as %" PRIu32 " mapped to %" PRIu32 "\n",
+            s->id, s->msg, s->num, mapped_signum);
     /* Set the internal signum to the list before merging */
-    IPOnlyCIDRListSetSigNum(s->cidr_src, s->num);
+    IPOnlyCIDRListSetSigNum(s->cidr_src, mapped_signum);
 
-    IPOnlyCIDRListSetSigNum(s->cidr_dst, s->num);
+    IPOnlyCIDRListSetSigNum(s->cidr_dst, mapped_signum);
 
     /**
      * ipv4 and ipv6 are mixed, but later we will separate them into
@@ -1560,8 +1575,8 @@ void IPOnlyAddSignature(DetectEngineCtx *de_ctx, DetectEngineIPOnlyCtx *io_ctx,
     io_ctx->ip_src = IPOnlyCIDRItemInsert(io_ctx->ip_src, s->cidr_src);
     io_ctx->ip_dst = IPOnlyCIDRItemInsert(io_ctx->ip_dst, s->cidr_dst);
 
-    if (s->num > io_ctx->max_idx)
-        io_ctx->max_idx = s->num;
+    if (mapped_signum > io_ctx->max_idx)
+        io_ctx->max_idx = mapped_signum;
 
     /** no longer ref to this, it's in the table now */
     s->cidr_src = NULL;
--- a/src/detect-engine-iponly.c
+++ b/src/detect-engine-iponly.c
@@ -896,25 +896,6 @@ SigIntId IPOnlyTrackSigNum(DetectEngineIPOnlyCtx *io_ctx, SigIntId signum)
 }
 
 /**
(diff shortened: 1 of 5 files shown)


JustinAzoff:
Testing this - and even reproducing the issue is a little tricky.

For the issues to occur you need at least 2 ip only rules:

```
# One that will always match on src due to using any or home_net
alert ip any any -> [list of ips] any
# One that will always match on dst using any or home_net
alert ip any [list of ips] -> $HOME_NET any
```

If you don't have at least one rule of each shape, this statement will trigger and it will never hit the problematic code:

```
    if (src == NULL || dst == NULL)
        SCReturn;
```

If you tested ip only rules with only one of the two "shapes" of rules, you'd never hit the performance issue because src or dest will always be NULL.

The other requirement to trigger the issue is a large number of rules.

The bitsets are created with the internal suricata signums. This means that if you have a large number of rules and the above 2 rules, the bitsets do not look like this:

```
1
2
```

or in binary

```
01
10
```
but instead, they potentially look like

```
30001
30002
```
or in binary

```
000000000000000000(30000 0s)01
000000000000000000(30000 0s)10
```

To easily reproduce this issue I wrote a little script to generate 40,000 rules, then 4 ip only rules.  Depending on the order of the rules and the sids, the rules either end up with signums of [0,1,2,3] or [40001,40002,40003,40004].  If the ip only rules have a signum of 40000, the bitset ends up 5,000 bytes wide instead of only one byte.

[break_iponly.py.txt](https://github.com/OISF/suricata/files/10863543/break_iponly.py.txt)
[generated.rules.txt](https://github.com/OISF/suricata/files/10863544/generated.rules.txt)
[reordered.rules.txt](https://github.com/OISF/suricata/files/10863545/reordered.rules.txt)

With those 2 rulesets running against a pcap of 1,000,000 syn packets you will see performance like this:

```
 hyperfine -m 3 -L rules generated.rules,reordered.rules './src/suricata -c suricata.yaml -S {rules} -r ~/pcaps/scan.pcap'
Benchmark 1: ./src/suricata -c suricata.yaml -S generated.rules -r ~/pcaps/scan.pcap
  Time (mean ± σ):     18.791 s ±  0.183 s    [User: 19.641 s, System: 0.995 s]
  Range (min … max):   18.640 s … 18.995 s    3 runs
 
Benchmark 2: ./src/suricata -c suricata.yaml -S reordered.rules -r ~/pcaps/scan.pcap
  Time (mean ± σ):      8.737 s ±  0.100 s    [User: 9.516 s, System: 0.837 s]
  Range (min … max):    8.642 s …  8.842 s    3 runs
 
Summary
  './src/suricata -c suricata.yaml -S reordered.rules -r ~/pcaps/scan.pcap' ran
    2.15 ± 0.03 times faster than './src/suricata -c suricata.yaml -S generated.rules -r ~/pcaps/scan.pcap'
```

With the patch, the two rulesets run with the same performance.

There's a further improvement that can be made to increase the size of the bitsets from uint8's to uint32 or uint64s, which will further reduce the number of loop iterations inside of IPOnlyMatchPacket.

src/detect.h
```diff
@@ -716,6 +711,8 @@ typedef struct DetectEngineIPOnlyCtx_ {
     /* Used to build the radix trees */
     IPOnlyCIDRItem *ip_src, *ip_dst;
     uint32_t max_idx;
+    uint32_t *sig_mapping;
```
victorjulien, line 714: can you add a description of what the mapping is?

src/detect-engine-iponly.c
```diff
@@ -883,25 +883,17 @@ void IPOnlyInit(DetectEngineCtx *de_ctx, DetectEngineIPOnlyCtx *io_ctx)
                                                   SigNumArrayPrint);
     io_ctx->tree_ipv6dst = SCRadixCreateRadixTree(SigNumArrayFree,
                                                   SigNumArrayPrint);
+
+    io_ctx->sig_mapping = SCCalloc(1, de_ctx->sig_array_len * sizeof(uint32_t));
```
victorjulien, line 887: unchecked alloc will fail CI

src/detect-engine-iponly.c
```diff
@@ -1023,23 +1008,22 @@ void IPOnlyMatchPacket(ThreadVars *tv,
         SCReturn;
 
     uint32_t u;
+    uint8_t bitarray;
     for (u = 0; u < src->size; u++) {
         SCLogDebug("And %"PRIu8" & %"PRIu8, src->array[u], dst->array[u]);
 
-        /* The final results will be at io_tctx */
-        io_tctx->sig_match_array[u] = dst->array[u] & src->array[u];
+        bitarray = dst->array[u] & src->array[u];
```
victorjulien, line 1015: declare here

src/detect.h
```diff
@@ -1137,9 +1134,6 @@ typedef struct DetectEngineThreadCtx_ {
      * prototype held by DetectEngineCtx. */
     SpmThreadCtx *spm_thread_ctx;
 
-    /** ip only rules ctx */
-    DetectEngineIPOnlyThreadCtx io_ctx;
```
victorjulien, line 1141: nice cleanup

src/detect-engine-iponly.c
```diff
@@ -883,25 +883,17 @@ void IPOnlyInit(DetectEngineCtx *de_ctx, DetectEngineIPOnlyCtx *io_ctx)
...
-    /* initialize the signature bitarray */
-    io_tctx->sig_match_size = de_ctx->io_ctx.max_idx / 8 + 1;
-    io_tctx->sig_match_array = SCMalloc(io_tctx->sig_match_size);
-    if (io_tctx->sig_match_array == NULL) {
-        exit(EXIT_FAILURE);
-    }
-
-    memset(io_tctx->sig_match_array, 0, io_tctx->sig_match_size);
+    SigIntId loc = io_ctx->sig_mapping_size;
+    io_ctx->sig_mapping[loc] = signum;
```
victorjulien, line 894: could just be `io_ctx->sig_mapping[io_ctx->sig_mapping_size++] = signum;`?

src/detect-engine-iponly.c
```diff
@@ -1023,23 +1008,22 @@ void IPOnlyMatchPacket(ThreadVars *tv,
...
-        if (io_tctx->sig_match_array[u] != 0) {
+        if (bitarray) {
             /* We have a match :) Let's see from which signum's */
-            uint8_t bitarray = io_tctx->sig_match_array[u];
             uint8_t i = 0;
 
             for (; i < 8; i++, bitarray = bitarray >> 1) {
                 if (bitarray & 0x01) {
-                    Signature *s = de_ctx->sig_array[u * 8 + i];
+                    Signature *s = de_ctx->sig_array[io_ctx->sig_mapping[u * 8 + i]];
```
victorjulien, line 1026: I wonder if it makes sense to avoid this additional indirection. Perhaps we can just store the `Signature` pointers themselves. Takes twice the space, but gets rid of this double lookup.

victorjulien (commented):
Thanks for looking into this. I guess another, perhaps simpler approach would to handle this during `Signature::num` assignment, which depends on the sort order of the sigs in the list after loading them. If we'd have all ip only sigs in the beginning of the list, they'd all get id's starting at 0 and incrementing w/o gaps. Could even have a max_ip_only_id or something to cap the size of the array.

src/detect-engine-iponly.c
JustinAzoff: it could, but then it would have to

```
return io_ctx->sig_mapping_size-1
```
to return the index that was allocated.

src/detect-engine-iponly.c
JustinAzoff: I don't see why not..  I'm not sure about the performance impact, but the result could be easier to read with out the extra lookup

src/detect.h
JustinAzoff: No problem.. also, that should just be defined as a SigIntId right? I realized that and the alloc call should probably be SigIntId and not use uint32 directly.

JustinAzoff:
Hi,

I did also have the idea to sort rules internally to ensure that ip only rules were first, but I'm not familiar with how that works and if there could be any adverse affects.  The end result would be the same, but it seemed like a fishy design to have the iponly engine have a hard dependency on something elsewhere sorting signatures in a specific way.

src/detect-engine-iponly.c
JustinAzoff: changed to

```
    io_ctx->sig_mapping = SCCalloc(1, de_ctx->sig_array_len * sizeof(uint32_t));
    if (io_ctx->sig_mapping == NULL) {
        FatalError("Unable to allocate iponly signature tracking area");
    }
```

src/detect.h
JustinAzoff: Was sort of an accident :-)  Once I noticed I could remove io_tctx->sig_match_array and use a local variable instead, I noticed there were no other references to the tctx.  I wasn't sure if I could actually remove it until I tried it and everything still worked.

src/detect.h
JustinAzoff: Added comment

```
+    /* Used to map large signums to smaller values to compact the bitsets
+     * stored in the radix trees */
```

Still a little fuzzy on if i should change the uint32_t to be SigIntId or not.

src/detect-engine-iponly.c
JustinAzoff: Oh, i see what you mean like this?

```
SigIntId loc = io_ctx->sig_mapping_size;
io_ctx->sig_mapping[io_ctx->sig_mapping_size++] = signum;
return loc;
```

I had split it up because I had it inlined the first time but it was slightly wrong and it was returning the wrong value.

jlucovsky:
My experiences with the PR also demonstrate a performance increase(`suricata.rules` is built 

The following measurements were made with the changes from the PR:
```
Benchmark 1: ./src/suricata -c suricata.yaml -S /home/jlucovsky/generated.rules -r ~/scan.pcap -l /tmp/ll --runmode=single
  Time (mean ± σ):     25.662 s ±  0.147 s    [User: 22.361 s, System: 5.659 s]
  Range (min … max):   25.504 s … 25.796 s    3 runs

Benchmark 2: ./src/suricata -c suricata.yaml -S ~/reordered.rules -r ~/scan.pcap -l /tmp/ll --runmode=single
  Time (mean ± σ):     25.554 s ±  0.134 s    [User: 22.264 s, System: 5.765 s]
  Range (min … max):   25.410 s … 25.675 s    3 runs

Benchmark 3: ./src/suricata -c suricata.yaml -S ~/suricata.rules -r ~/scan.pcap -l /tmp/ll --runmode=single
  Time (mean ± σ):     87.893 s ±  0.415 s    [User: 82.921 s, System: 7.720 s]
  Range (min … max):   87.457 s … 88.284 s    3 runs

Summary
  './src/suricata -c suricata.yaml -S ~/reordered.rules -r ~/scan.pcap -l /tmp/ll --runmode=single' ran
    1.00 ± 0.01 times faster than './src/suricata -c suricata.yaml -S /home/jlucovsky/generated.rules -r ~/scan.pcap -l /tmp/ll --runmode=single'
    3.44 ± 0.02 times faster than './src/suricata -c suricata.yaml -S ~/suricata.rules -r ~/scan.pcap -l /tmp/ll --runmode=single'
    ```

For comparison purposes, these results are from using `master`:
```
Benchmark 1: ./src/suricata -c suricata.yaml -S /home/jlucovsky/generated.rules -r ~/scan.pcap -l /tmp/ll --runmode=single
  Time (mean ± σ):     45.062 s ±  0.655 s    [User: 42.254 s, System: 6.184 s]
  Range (min … max):   44.326 s … 45.579 s    3 runs

Benchmark 2: ./src/suricata -c suricata.yaml -S ~/reordered.rules -r ~/scan.pcap -l /tmp/ll --runmode=single
  Time (mean ± σ):     25.613 s ±  0.158 s    [User: 22.321 s, System: 5.559 s]
  Range (min … max):   25.450 s … 25.766 s    3 runs

Benchmark 3: ./src/suricata -c suricata.yaml -S ~/suricata.rules -r ~/scan.pcap -l /tmp/ll --runmode=single
  Time (mean ± σ):     116.036 s ±  3.008 s    [User: 111.718 s, System: 8.465 s]
  Range (min … max):   112.570 s … 117.957 s    3 runs

Summary
  './src/suricata -c suricata.yaml -S ~/reordered.rules -r ~/scan.pcap -l /tmp/ll --runmode=single' ran
    1.76 ± 0.03 times faster than './src/suricata -c suricata.yaml -S /home/jlucovsky/generated.rules -r ~/scan.pcap -l /tmp/ll --runmode=single'
    4.53 ± 0.12 times faster than './src/suricata -c suricata.yaml -S ~/suricata.rules -r ~/scan.pcap -l /tmp/ll --runmode=single'
    ```

victorjulien:
Merged in #8713, thanks a lot @JustinAzoff ! Thanks for the perf numbers @jlucovsky

PR #8570, opened 2023-03-03
config: put version in configuration as a proper value - v1

jasonish:
Adds a new field, "suricata-version" to the configuration file with
the major and minor version of the Suricata that generated the
configuration file.

This may be useful in the future for presenting warnings about
important changes, or even providing different defaults based on what
the user might expect.

Example of whats added to `suricata.yaml`:

```
# This configuration file generated by Suricata 7.0.0-rc2-dev.
suricata-version: "7.0"
```

Ticket: https://redmine.openinfosecfoundation.org/issues/5822


victorjulien:
Merged in #8572, thanks!

PR #8573, opened 2023-03-06
schema: fix json schema - v1

jasonish:
- Fix engines section: it was not valid jsonschema but this wasn't being picked
  up by the Rust validator.

- Fix optional: optional is not a valid json field. Instead required elements
  must be listed in an array.


victorjulien:
Merged in #8587, thanks!

PR #8578, opened 2023-03-08
[draft] app-layer: mark unidirectional transactions with a direction - v1

jasonish:
This PR pulls in a commit from @regit that cleans the flow flags to just to
direction before passing to StateGetProgress is that is what parsers are
inspecting.

Then at the app-layer, mark each unidirectional transaction with its direction
so its not considered for inspection in the other direction, which may cause it
to be disposed of too early.

Issue: https://redmine.openinfosecfoundation.org/issues/4759 (and others)

suricata-verify-pr: 1140

### DIFF
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -122,6 +122,10 @@ pub struct AppLayerTxData {
     /// STREAM_TOSERVER|STREAM_TOCLIENT: files possible in both dirs
     pub file_tx: u8,
 
+    /// The direction of the transaction. 0 for transactions that are
+    /// made up of request and replies.
+    direction: u8,
+
     /// detection engine flags for use by detection engine
     detect_flags_ts: u64,
     detect_flags_tc: u64,
@@ -157,6 +161,7 @@ impl AppLayerTxData {
             files_stored: 0,
             file_flags: 0,
             file_tx: 0,
+	    direction: 0,
             detect_flags_ts: 0,
             detect_flags_tc: 0,
             de_state: std::ptr::null_mut(),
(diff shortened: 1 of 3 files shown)


rust/src/applayer.rs
```diff
@@ -122,6 +122,10 @@ pub struct AppLayerTxData {
     /// STREAM_TOSERVER|STREAM_TOCLIENT: files possible in both dirs
     pub file_tx: u8,
 
+    /// The direction of the transaction. 0 for transactions that are
+    /// made up of request and replies.
+    direction: u8,
```
jasonish, line 127: This could probably be encoded into the flags below to remove the new field, but I wonder if this would be useful to have.

src/detect.c
```diff
@@ -1256,13 +1267,17 @@ static DetectTransaction GetDetectTx(const uint8_t ipproto, const AppProto alpro
...
+        SCLogDebug("%" PRIu64 " tx already fully inspected for %s. Flags %016" PRIx64, tx_id,
+                flow_flags & STREAM_TOSERVER ? "toserver" : "toclient", detect_flags);
         return no_tx;
     }
 
+    if (txd && txd->direction) {
+        if (txd->direction != (flow_flags & (STREAM_TOSERVER | STREAM_TOCLIENT))) {
+            return no_tx;
+        }
+    }
```
jasonish, line 1279: Needs comment: Don't inspect unidirectional transactions in the reverse direction.

src/app-layer-parser.c
```diff
@@ -1503,6 +1504,25 @@ int AppLayerParserParse(ThreadVars *tv, AppLayerParserThreadCtx *alp_tctx, Flow
...
+                AppLayerGetTxIterTuple ires =
+                        IterFunc(f->proto, f->alproto, f->alstate, min_id, cur_tx_cnt, &state);
+                if (ires.tx_ptr == NULL) {
+                    break;
+                }
+                AppLayerTxData *txd = AppLayerParserGetTxData(f->proto, alproto, ires.tx_ptr);
+                txd->direction = flags & (STREAM_TOSERVER | STREAM_TOCLIENT);
+                min_id += 1;
+            }
+        }
```
jasonish, line 1525: As one call to the parser may create multiple transactions, which need this loop here.  The alternative is to have the parserset the direction on each unidirectional transaction, but my attempt here was to remove that responsibility from the parser writer.

jasonish:
Replaced by https://github.com/OISF/suricata/pull/8601 where the creation of unidirectional transactions is the job of the app-layer, not be "guessed" by the app-layer framework.

PR #8585, opened 2023-03-10
smb/ntlmssp:the negotiate flags num is little endian

tianjinshan:
Ticket: https://redmine.openinfosecfoundation.org/issues/5783

To get the 6th bit of integer, have to pass 3*8+6 bits in memory。I have tested in master-5.0.x, it's looks like:
let (i, nego_flags) = bits!(i, tuple!(take_bits!(u32, 30),take_bits!(u8,1),take_bits!(u8,1)))?;
I have updated the unitest(smb::ntlmssp_records::tests::test_parse_auth_nego_flags). The Suricata-Verify test is "tests/smb2-ntlmssp-negotiateflags/", no need to change.

Signed-off-by: tianjinshan tianjinshan@gmail.com

### DIFF
--- a/rust/src/smb/ntlmssp_records.rs
+++ b/rust/src/smb/ntlmssp_records.rs
@@ -69,7 +69,7 @@ pub struct NTLMSSPAuthRecord<'a> {
 }
 
 fn parse_ntlm_auth_nego_flags(i: &[u8]) -> IResult<&[u8], (u32, u8, u8)> {
-    bits(tuple((take_bits(25u8), take_bits(1u8), take_bits(6u32))))(i)
+    bits(tuple((take_bits(30u32), take_bits(1u8), take_bits(1u8))))(i)
 }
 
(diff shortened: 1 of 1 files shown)


tianjinshan:
> 

I think you are right, the purpose of using take_bits is to get reminder of payload, but the reminder looks like unused any more, so your opinion is more readable. My changes are as small as possible to avoid introducing bugs.

I have only changed the file "rust/src/smb/ntlmssp_records.rs", but the check result is about rust/src/core.rs, it's confusing...

catenacyber:
@tianjinshan do you want to submit another PR with the use of `le_u32` instead of `take_bits` ?
Otherwise, I can do it to fix the bug....

tianjinshan:
> 
le_u32 maybe a suitable modification, you can do it. i will close my PR.

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8673

Thanks @tianjinshan :-)

PR #8596, opened 2023-03-14
output/eve: add 'verdict' field to 'alert' and 'drop' events - v7

jufajardini:
https://redmine.openinfosecfoundation.org/issues/5464

Previous PR: https://github.com/OISF/suricata/pull/8559

Changes from last PR:
- remove global variable for saving info on `alert.action` setting
- incorporate other feedback from previous PR
 
suricata-verify-pr: 1145
https://github.com/OISF/suricata-verify/pull/1145

### DIFF
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -264,6 +264,19 @@ enabled, then the log gets more verbose.
 
 By using ``custom`` it is possible to select which TLS fields to log.
 
+Drops
+~~~~~
+
+Drops are event types logged when the engine drops a packet.
+
+Config::
+
+    - drop:
+        alerts: yes      # log alerts that caused drops
+        flows: all       # start or all: 'start' logs only a single drop
+                         # per flow direction. All logs each dropped pkt.
+
+
 Date modifiers in filename
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/decode.c
+++ b/src/decode.c
@@ -1,4 +1,4 @@
-/* Copyright (C) 2007-2022 Open Information Security Foundation
+/* Copyright (C) 2007-2023 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
--- a/src/decode.c
+++ b/src/decode.c
@@ -819,6 +819,24 @@ const char *PacketDropReasonToString(enum PacketDropReason r)
     return NULL;
 }
 
+/** \brief Decide Packet's final verdict based on packet action, return it as a string */
+const char *PacketActionVerdictToString(const Packet *p)
+{
+    /* Reject is valid in both IDS and IPS */
+    if (PacketCheckAction(p, ACTION_REJECT_ANY)) {
+        return "reject";
+    } else if (EngineModeIsIPS()) {
+        /* Verdicts will be reject, drop, or accept */
+        if (PacketCheckAction(p, ACTION_DROP)) {
+            return "drop";
+        }
+        return "accept";
+    }
+
+    /* If we're in IDS mode and action isn't reject, we won't log verdict */
+    return NULL;
+}
+
 /* TODO drop reason stats! */
 void CaptureStatsUpdate(ThreadVars *tv, CaptureStats *s, const Packet *p)
 {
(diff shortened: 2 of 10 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .flow.end.tcp_state.last_ack | 1004 | 257 | 25.6% |
| .tcp.rst | 145769 | 113307 | 77.73% |
| .flow.end.tcp_state.fin_wait2 | 156072 | 127814 | 81.89% | 

Pipeline 12746

src/decode.c
```diff
@@ -819,6 +819,24 @@ const char *PacketDropReasonToString(enum PacketDropReason r)
     return NULL;
 }
 
+/** \brief Decide Packet's final verdict based on packet action, return it as a string */
+const char *PacketActionVerdictToString(const Packet *p)
+{
+    /* Reject is valid in both IDS and IPS */
+    if (PacketCheckAction(p, ACTION_REJECT_ANY)) {
+        return "reject";
```
victorjulien, line 827: I'd like to rethink the reject here. I feel it takes something very simple (was a packet accepted or dropped) and makes it complicated and hard to reason about. Perhaps we need to split it out. E.g.

IPS:
```json
"verdict": {
     "verdict": "drop"
     "active-response": "reject"
}
```
and IDS:
```json
"verdict": {
     "active-response": "reject"
}
```
(I don't like the verdict.verdict here, need a better name. But not enough coffee yet to come up with one)

jasonish: I think just as much information is provided with

IPS:

```
"verdict": "drop"
```

IDS:

```
"verdict": "reject"
```

victorjulien: I disagree. Is a packet dropped as well as rejected in IPS mode? What about IDS mode?

jasonish: Lets enumerate the actions, what do we have:
- drop
- reject
- drop-and-reject
Anything else?

Maybe a new noun is needed:
- verdict: the decision made, this could simply be allow/deny
- action: this is the action taken.. drop, reject, both, ?

jufajardini: Pondering on this  to try to come up with a suggestion...

jufajardini: I think we have two aspects to consider:
- the initial problem - having only `action` leading to ambiguous results (when, say, a packet triggers signatures with conflicting/different actions and therefore `action` isn't informative enough of the final action taken on the packet). 
- And: that `reject` has different results in `IPS` and `IDS` and, moreover, that in `IPS` `reject` also means dropping the packet.

The `verdict` field seems to only be insufficient in `reject`'s case, when I can't see how to solve this with a single word. And even then, given reject's implications, I think this still should require going to the documentation.

I think that we should avoid using the `action` noun, considering we use it for the signatures. 

However I look at this, I end up feeling that we'll be redundant and lack a bit of info, at the same time, when it comes to `reject`... >_<

victorjulien: How about:
```json
"verdict": {
   "action": "reject",
   "ips-verdict": "drop",
   "active-response": {
       "name": "tcp-reset"
       "direction": "toserver"
   }
}
```

jasonish: I think we need to provide example rules, and modes that lead to what the output would look like, as I do find it a bit hard to reason about.

If we take a court as an example, verdict is indepedent of action.  The verdict might be guilty, not-guilty.  Action could be nothing, jail, jail+fine, etc.

So in the example above, I believe verdict is `reject` (from the rule), and the action is drop+reject. So I think `ips-verdict` doesn't have the base name.  Yes, the verdict was drop, but did we drop it?

jufajardini: Do you think that we should change the words used, or that the documentation/ rule examples/ SV tests showcasing could be enough?

doc/userguide/output/eve/eve-json-output.rst
```diff
@@ -264,6 +264,19 @@ enabled, then the log gets more verbose.
 
 By using ``custom`` it is possible to select which TLS fields to log.
 
+Drops
+~~~~~
+
+Drops are event types logged when the engine drops a packet.
```
inashivb, line 270: Is it like `event_type == drop`? or an alert indicating that the packet was dropped?

jufajardini: Sorry for not seeing this question before >___<'
This is the for `event_type == drop`, which Suri issues when a packet is dropped (caused by a rule or another reason).

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .flow.end.tcp_state.last_ack | 1004 | 257 | 25.6% |
| .tcp.rst | 145769 | 113307 | 77.73% |
| .flow.end.tcp_state.fin_wait2 | 156072 | 127814 | 81.89% | 

Pipeline 12746

PR #8599, opened 2023-03-15
smtp: sync in-code defaults with configuration file - v1

jasonish:
Currently the default suricata.yaml sets some values that do not
reflect the default values in the code. As most users are probably
using a default suricata.yaml, make the code have the same defaults as
in suricata.yaml:

- mime.decode-mime: false -> true
- mime.decode-base64: false -> true
- mime.decode-quoted-printable: false -> true
- mime.extract-urls: false -> true

I guess the question is if its OK for a simple configuration of:

```
app-layer:
  protocols:
    smtp:
      enabled: yes
```

to acquire these new defaults.

Ticket: https://redmine.openinfosecfoundation.org/issues/5823

### DIFF
--- a/etc/schema.json
+++ b/etc/schema.json
@@ -1338,7 +1338,10 @@
                 },
                 "has_exe_url": {
                     "type": "boolean"
-                }
+                },
+		"message_id": {
+		    "type": "string"
+		}
             },
             "additionalProperties": false
         },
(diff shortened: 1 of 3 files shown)


victorjulien (approved):

victorjulien:
Merged in #8605, thanks!

PR #8603, opened 2023-03-17
Base64 data issue/v3

inashivb:
Previous PR: #8580 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5885

Changes since v2:
- Unit tests for RFC 4648 and RFC2045
- Fix TODO about skipping over any invalid char in case of RFC2045

### DIFF
--- a/src/util-base64.c
+++ b/src/util-base64.c
@@ -348,6 +348,9 @@ static int B64TestVectorsRFC4648(void)
     const char *src9 = "Zm$9vYm.Fy";
     const char *fin_str9 = "f";
 
+    const char *src10 = "Y21Wd2IzSjBaVzFoYVd4bWNtRjFaRUJoZEc4dVoyOTJMbUYxOmpqcHh4b3Rhb2w%3D";
+    const char *fin_str10 = "cmVwb3J0ZW1haWxmcmF1ZEBhdG8uZ292LmF1:jjpxxotaol";
+
     TEST_RFC4648(src1, fin_str1, ASCII_BLOCK * 2, strlen(fin_str1), strlen(src1), BASE64_ECODE_OK);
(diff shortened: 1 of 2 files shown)


inashivb:
Q: If the following string was to be decoded as per RFC 2045 i.e. skip over any invalid data.

Base64 encoded string: `Y21Wd2IzSjBaVzFoYVd4bWNtRjFaRUJoZEc4dVoyOTJMbUYxOmpqcHh4b3Rhb2w%5d`
Len: 66

The decoded string is as follows:
Decoded string: `cmVwb3J0ZW1haWxmcmF1ZEBhdG8uZ292LmF1:jjpxxotaol9t`
Len: 49

(ref: [Decoded data](https://cyberchef.io/#recipe=From_Base64('A-Za-z0-9%2B/%3D',true)&input=WTIxV2QySXpTakJhVnpGb1lWZDRiV050UmpGYVJVSm9aRWM0ZFZveU9USk1iVVl4T21wcWNIaDRiM1JoYjJ3JTVk))

but, the number of decoded bytes logged by our decoder is 48.
Reason:
Out of the 66 bytes, 65 bytes are valid characters and they make 16 perfect b64 blocks. This leaves the one byte at the end: `d`. So, the decoder takes it and adds 3 padding bytes to it and decodes it correctly to the `Decoded string` mentioned above. But, actually we only had 48 properly decoded ASCII bytes. The `t` in the end of the `Decoded string` is a result of 1 byte from the encoded string + 3 padding bytes.

Do you think the decoded bytes should have indicated 49 instead? Or is 48 the right value in this case?

inashivb:
Replaced w https://github.com/OISF/suricata/pull/8651

PR #8606, opened 2023-03-18
 nfq: fix potential NULL deref

ProjectMutilation:
Added some asserts statements for NFQGetQueue so that
 it is not possible to return NULL.
 Also deleted useless check of returned value of
 NFQGetQueue() in NFQInitThread()
 Found by Security Code with Svace static analyzer
Signed-off-by: Maxim Korotkov <m.korotkov@securitycode.ru>

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/6385

Describe changes:
- added check of ptr and added return error code if check is fail

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/source-nfq.c
+++ b/src/source-nfq.c
@@ -569,10 +569,6 @@ static TmEcode NFQInitThread(NFQThreadVars *t, uint32_t queue_maxlen)
     struct timeval tv;
     int opt;
     NFQQueueVars *q = NFQGetQueue(t->nfq_index);
-    if (q == NULL) {
-        SCLogError("no queue for given index");
-        return TM_ECODE_FAILED;
-    }
     SCLogDebug("opening library handle");
     q->h = nfq_open();
     if (q->h == NULL) {
@@ -915,13 +911,10 @@ int NFQParseAndRegisterQueues(const char *queues)
  *  \param number idx of the queue in our array
  *
  *  \retval ptr pointer to the NFQThreadVars at index
- *  \retval NULL on error
  */
 void *NFQGetQueue(int number)
 {
-    if (unlikely(number < 0 || number >= receive_queue_num || g_nfq_q == NULL))
-        return NULL;
-
+    BUG_ON(number < 0 || number >= receive_queue_num || g_nfq_q == NULL);
     return (void *)&g_nfq_q[number];
 }
 


src/source-nfq.c
```diff
@@ -536,6 +536,9 @@ static int NFQCallBack(struct nfq_q_handle *qh, struct nfgenmsg *nfmsg,
     if (ret == -1) {
 #ifdef COUNTERS
         NFQQueueVars *q = NFQGetQueue(ntv->nfq_index);
+        if (!q) {
+            return -1;
```
jlucovsky, line 540: The `Packet` object obtained from `PacketGetFromQueueOrAlloc()` must be returned on this (and the other new error return that you added).

Check the other error cases -- I think `TmqhOutputPacketpool` will do it or maybe even calling `NFQReleasePacket`.

ProjectMutilation:
```
        if (q) {
            q->errs++;
            q->pkts++;
            q->bytes += GET_PKT_LEN(p);
        }
```
is it okay solution? I'm not sure what you want from me

jlucovsky: Yes, this seems ok for this change and the second change in the PR.

src/source-nfq.c
```diff
@@ -536,9 +536,11 @@ static int NFQCallBack(struct nfq_q_handle *qh, struct nfgenmsg *nfmsg,
     if (ret == -1) {
 #ifdef COUNTERS
         NFQQueueVars *q = NFQGetQueue(ntv->nfq_index);
```
victorjulien, line 538: `NFQGetQueue` should probably not be able to return NULL instead, adding some asserts. The checks that are done in the function can probably move to some startup routine.

jlucovsky: @victorjulien I agree but that might be outside the scope of this PR. Perhaps a tracking issue for removing the possibility of a `NULL` return value would be a reasonable option?

victorjulien: I don't see it as out of scope. I don't see the point of adding runtime complexity for a very theoretical issue, essentially just to satisfy a code checker we have no access to.

ProjectMutilation: I agree that it's theoretical issue, but the probability exists, may be NFQGetQueue() must not be able to return a null pointer?

catenacyber (changes requested):
> NFQGetQueue should probably not be able to return NULL instead, adding some asserts

src/source-nfq.c
```diff
@@ -915,13 +911,12 @@ int NFQParseAndRegisterQueues(const char *queues)
...
  *
  *  \retval ptr pointer to the NFQThreadVars at index
- *  \retval NULL on error
  */
 void *NFQGetQueue(int number)
 {
-    if (unlikely(number < 0 || number >= receive_queue_num || g_nfq_q == NULL))
-        return NULL;
-
+    assert(number < 0);
```
catenacyber, line 917: It can remain one line with `||`
And it looks like you are asserting the negation of what should be asserted, right ?

And I think we would like to use our macro `BUG_ON` rather than `assert`

jlucovsky: Something like
```
BUG_ON(number > 0 && number < receive_queue_num && g_nfq_q != NULL)
```

ProjectMutilation: Sorry, I really forgot to invert the conditions, but think there should be a BUG_ON for each condition that defines exactly what happened

ProjectMutilation:
```
        #define BUG_ON(x) do {      \
            if (((x))) {            \
                fprintf(stderr, "BUG at %s:%d(%s)\n", __FILE__, __LINE__, __func__);    \
                fprintf(stderr, "Code: '%s'\n", xstr((x)));                             \
                exit(EXIT_FAILURE); \
            }                       \
        } while(0)
     ```
     I may be wrong, but it seems there is no need to invert the condition
     We want detect if (number < 0 || number >= receive_queue_num) and if g_nfq_q is NULL

jlucovsky: You're correct --- `BUG_ON` uses the condition `x` to trigger the issue.

Whether a single `BUG_ON` or not is used depends on context. In this case, I'd recommend a single `BUG_ON` statement.

jufajardini:
Should this be rebased and re-submitted in a new PR, so we can make sure CI is good and this can be considered for merge?

jufajardini:
And shouldn't there be a redmine ticket for this?

ProjectMutilation:
> And shouldn't there be a redmine ticket for this?

Nope, there is no existing ticket for this. 
I can see if it is possible to accept the patch on the last commit in the dev branch  in the near future if the patch is not rejected

jufajardini:
> > And shouldn't there be a redmine ticket for this?
> 
> Nope, there is no existing ticket for this. I can see if it is possible to accept the patch on the last commit in the dev branch in the near future if the patch is not rejected

My bad, I meant: should we create a ticket for it? We've been trying to have tickets mentioned in commits, unless changes are really trivial...

victorjulien (approved):

victorjulien:
We'll need a new, clean PR for this to get merged, per our contribution guidelines. History of the PR is too messy this way.

PR #8613, opened 2023-03-20
Bundle Cargo.lock

jasonish:
We can pin versions used by git checkouts by bundling a Cargo.lock, however we have to bundled a `Cargo.lock.in` and not just a `Cargo.lock` to satisfy automake and `make distcheck`.

So updating `Cargo.lock` means updating it, then copying it to `Cargo.lock.in` which the `make update-lock` helper is for.

Should prevent issues like https://redmine.openinfosecfoundation.org/issues/5914 coming up a random.  It won't prevent them for good though, but we can then address them on our schedule.


victorjulien:
Passes my QA, so looking good. Would be good to have a test release to make sure there are no surprises.

victorjulien:
Merged in #8635, thanks!

PR #8626, opened 2023-03-24
exception/policy: use pkt action if no flow support - v1

jufajardini:
Defrag memcap and flow memcap do not support flow action for the exception policies, as there is no flow when the exception condition is hit. In such cases, the exception policy must be considered for the packet only, when that makes sense, or should be ignored, in case of `bypass`.

Bug #TBD

didn't create one yet, as not confirmed it's a bug

Describe changes:
- Add checks to `ExceptionPolicyParse` to ensure that if an exception policy with flow action is selected when that's not valid, we'll fall back to packet actions, when possible.

suricata-verify-pr: 1155
https://github.com/OISF/suricata-verify/pull/1155

### DIFF
--- a/src/util-exception-policy.c
+++ b/src/util-exception-policy.c
@@ -150,10 +150,24 @@ enum ExceptionPolicy ExceptionPolicyParse(const char *option, const bool support
         }
 
         if (!support_flow) {
-            if (policy == EXCEPTION_POLICY_DROP_FLOW || policy == EXCEPTION_POLICY_PASS_FLOW ||
-                    policy == EXCEPTION_POLICY_BYPASS_FLOW) {
-                SCLogWarning("flow actions not supported for %s, defaulting to \"ignore\"", option);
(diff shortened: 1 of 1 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | build_asan | | |
 

Pipeline 12827

jufajardini:
> Seems to work as it should. Should probably also update the comment(s) in the configuration, for example:
> 
> ```
> # The memcap-policy value can be "drop-flow", "pass-flow", "bypass",
> # "drop-packet", "pass-packet", "reject" or "ignore" (which is the default).
> defrag:
> ```
> 
> to not include the policies that are not supported.

🙇🏽  thanks for the reminder!

Next PR will also address the CI failures, which didn´t fail locally, for some reason...

jufajardini:
Feedback incorporated in https://github.com/OISF/suricata/pull/8629

PR #8627, opened 2023-03-24
conf: deprecate duplicate includes - v1

jasonish:
Deprecates duplicate "include" keyword at the same level in a configuration
file by displaying a warnings.

Adds support for "include" to be an array.

The following will now output an error:

```
include: one.yaml
include: two.yaml
```

The following will not be processed as you might expect it to:

```
include:
  - one.yaml
  - two.yaml
```

and s asimple include will continue to work:

```
include: one.yaml
```


suricata-qa:

Pipeline 12830

victorjulien:
CIFuzz finds a way to SEGV this, needs investigation.

jasonish:
Replaced by https://github.com/OISF/suricata/pull/8627.

PR #8629, opened 2023-03-24
exception/policy: use pkt action if no flow support - v2

jufajardini:
Defrag memcap and flow memcap do not support flow action for the exception policies, as there is no flow when the exception condition is hit. In such cases, the exception policy must be considered for the packet only, when that makes sense, or should be ignored, in case of bypass.

Bug #5940

https://redmine.openinfosecfoundation.org/issues/5940

Describe changes from last PR:
- Try to fix error with `default` branch of the switch-case
- update documentation (including in `suricata.yaml`) to better reflect what are the possible usages and supported actions for each exception policy

suricata-verify-pr: 1155
https://github.com/OISF/suricata-verify/pull/1155

### DIFF
--- a/src/util-exception-policy.c
+++ b/src/util-exception-policy.c
@@ -150,10 +150,25 @@ enum ExceptionPolicy ExceptionPolicyParse(const char *option, const bool support
         }
 
         if (!support_flow) {
-            if (policy == EXCEPTION_POLICY_DROP_FLOW || policy == EXCEPTION_POLICY_PASS_FLOW ||
-                    policy == EXCEPTION_POLICY_BYPASS_FLOW) {
-                SCLogWarning("flow actions not supported for %s, defaulting to \"ignore\"", option);
-                policy = EXCEPTION_POLICY_NOT_SET;
+            switch (policy) {
+                case EXCEPTION_POLICY_DROP_FLOW:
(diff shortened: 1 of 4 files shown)


suricata-qa:

Pipeline 12831

src/util-exception-policy.c
```diff
@@ -150,10 +150,25 @@ enum ExceptionPolicy ExceptionPolicyParse(const char *option, const bool support
...
+                            option);
+                    policy = EXCEPTION_POLICY_PASS_PACKET;
+                    break;
+                case EXCEPTION_POLICY_BYPASS_FLOW:
+                    SCLogWarning(
+                            "flow actions not supported for %s, defaulting to \"ignore\"", option);
+                    policy = EXCEPTION_POLICY_NOT_SET;
+                    break;
+                default:
+                    /* do nothing, these are the ones that bring issues */
```
victorjulien, line 170: then better would be to add the other cases here as no-ops, so that if a new type is added a compile warning will be raise here

victorjulien (changes requested):
See one inline comment. Looks good otherwise.

src/util-exception-policy.c
jufajardini: 🙇🏽

jufajardini:
Replaced by https://github.com/OISF/suricata/pull/8631

PR #8632, opened 2023-03-27
 conf: deprecate duplicate includes - v2

jasonish:
Deprecates duplicate "include" keyword at the same level in a configuration
file by displaying a warnings.

Adds support for "include" to be an array.

The following will now output an error:

```
include: one.yaml
include: two.yaml
```

The following will not be processed as you might expect it to:

```
include:
  - one.yaml
  - two.yaml
```

and s asimple include will continue to work:

```
include: one.yaml
```

Previous PR: #8627

Changes from previous PR:
- [cifuzz] Don't attempt to include null value as file

### DIFF
--- a/src/conf-yaml-loader.c
+++ b/src/conf-yaml-loader.c
@@ -48,7 +48,7 @@ static int mangle_errors = 0;
 
 static char *conf_dirname = NULL;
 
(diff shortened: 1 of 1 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113267 | 145139 | 128.14% | 

Pipeline 12846

victorjulien:
Merged in #8638, thanks!

PR #8643, opened 2023-03-29
runmodes: Determine engine's copy-mode as early as possible

lukashino:
Configuration and behavior of HTP app layer depends on the copy mode of Suricata engine. Copy mode was set after the app layer setup. Decision of engine's copy mode operation is now made earlier.

Ticket: #5934

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5934) ticket:

Describe changes:
- engine mode decision is executed earlier

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2554,6 +2554,10 @@ int PostConfLoadedSetup(SCInstance *suri)
         SCReturnInt(TM_ECODE_FAILED);
     }
 #endif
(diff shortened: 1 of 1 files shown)


victorjulien:
QA seems very unhappy. Also, are there any tests for this issue?

lukashino:
Will need to check what's the problem. I can try to write a test too

victorjulien:
replaced by #8645

PR #8647, opened 2023-03-29
af-packet: Ignore outgoing packets on loopback interfaces

awelzel:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5955

Describe changes:

When reading packets from a loopback interface, packets are received twice: Once as outgoing and once as incoming packets.

Libpcap and a future version of Zeek's AF_PACKET plugin will ignore the outgoing packets.

With current versions of Suricata, sniffing a single http://localhost:80 request over lo minimally shows two syn packets, two synacks and twice as many packets in the stats entries than you'd see/expect when running tcpdump.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -327,6 +327,8 @@ typedef struct AFPThreadVars_
 
     int promisc;
 
+    int loopback;
+
     int down_count;
 
     uint16_t cluster_id;
@@ -898,6 +900,14 @@ static int AFPReadFromRing(AFPThreadVars *ptv)
             goto next_frame;
         }
 
+        /* Drop outgoing packets on loopback interfaces, only process incoming ones. */
+        if ( ptv->loopback ) {
+            const struct sockaddr_ll *sll =
+                    (const struct sockaddr_ll*)((uint8_t*)h.h2 + TPACKET_ALIGN(sizeof(struct tpacket2_hdr)));
+            if (sll->sll_pkttype == PACKET_OUTGOING)
+                goto next_frame;
+        }
+
         Packet *p = PacketGetFromQueueOrAlloc();
         if (p == NULL) {
             return AFPSuriFailure(ptv, h);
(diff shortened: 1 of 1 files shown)


src/source-af-packet.c
```diff
@@ -327,6 +327,8 @@ typedef struct AFPThreadVars_
 
     int promisc;
 
+    int loopback;
```
jlucovsky, line 330: Suggest this be a `bool` and be renamed to `is_loopback`

src/source-af-packet.c
```diff
@@ -988,8 +998,19 @@ static inline int AFPWalkBlock(AFPThreadVars *ptv, struct tpacket_block_desc *pb
 {
     const int num_pkts = pbd->hdr.bh1.num_pkts;
     uint8_t *ppd = (uint8_t *)pbd + pbd->hdr.bh1.offset_to_first_pkt;
+    const int loopback = ptv->loopback;
 
     for (int i = 0; i < num_pkts; ++i) {
+        /* Drop outgoing packets on loopback interfaces, only process incoming ones. */
+        if ( loopback ) {
```
victorjulien, line 1005: I'm not a fan of this check for a rare edge case on a per packet basis. I don't currently have an alternative in mind though.

awelzel: Yep. While I'd be surprised the overhead is measurable, it does stand out quite prominently.

A second `AFPWalkBlockLoopback` and  function pointer indirection  `ptv->walk_block(...)` would maybe do.

But, not sure that's worth it. I don't personally have a use-case for `lo` sniffing with Suricata (cursory google indicates there are cases of people wanting to do it, e.g. reverse proxy monitoring). In Zeek duplicated packets are flagged as retransmissions and very visible in the conn.log. Also, there was/is a discrepancy how the default libpcap based packet source behaves compared to the AF_PACKET one (and IMO they should behave the same in Zeek sniffing loopback).

Given this has not come up before (or not visible/noticable enough), maybe this is just not needed. How would you feel about a warning during source initialization with something like: "Monitoring loopback interfaces with AF_PACKET results in duplicated packets, consider using the libpcap based source instead or explain why you need to use AF_PACKET :-)" instead?

victorjulien: How about replacing this with something like
```c
for (int i = 0; i < num_pkts; ++i) {
   const struct sockaddr_ll *sll =  (const struct sockaddr_ll*)(ppd + TPACKET_ALIGN(sizeof(struct tpacket3_hdr)));
   if (unlikely(ptv->pkt_flags & sll->sll_pkttype) == 0)) {
       ppd = ppd + ((struct tpacket3_hdr *)ppd)->tp_next_offset;
       continue;
   }
```
It still has the check, but now it's more generalized. The Setup code would need to properly setup `ptv->pkt_flags`

awelzel: > It still has the check, but now it's more generalized. The Setup code would need to properly setup ptv->pkt_flags

That looks special, yes The values of `sll_pkttype` aren't suitable to be used with a bit mask, so that would need to look a bit more like the following with a few asserts about the max value of pkttype :-/
```
(ptv->pkt_flags & (1 << sll->sll_pkttype))
```

I can give that a whirl if that seems still okay.

victorjulien: Could use `ptv->pkt_flags & BIT_U32(sll->ssl_pkttype)`. But yeah lets give this a try.

awelzel:
Replaced by https://github.com/OISF/suricata/pull/8752

PR #8645, opened 2023-03-29
Draft: Backport/bug/5934 determining engine mode v2

lukashino:
Follow-up of #8643

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5934) ticket:

Describe changes:
- postpone the function call - picked the wrong commit for the previous version...

PR #8649, opened 2023-03-29
Backport/bug/5934 determining engine mode v3

lukashino:
Follow-up of #8645 
Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5934) ticket:

Changes:
- in how runmode is determined
- bug test for runmode changing during the Suricata startup

I do not think adding a SV test is viable as for the test we would need to setup an actual AFP/Netmap setup where the check for incorrectly set runmode mode would be run. (`--simulate-ips` doesn't really help here as it explicitly enables IPS mode. In the AFP/Netmap case it was checked from config but after libhtp was initialized)

I've been able to verify manually that prior to the fix, Suricata kept buffering the reassembled stream (IDS mode) and would only output an alert after some (flow) timeout (or Suricata shutdown). After the fix, Suricata properly behaves as in IPS inline mode where the alert was generated right after the packet containing the specified pattern.

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -719,7 +719,7 @@ static int AFPConfigGeThreadsCount(void *conf)
 
 int AFPRunModeIsIPS(void)
 {
-    int nlive = LiveGetDeviceCount();
+    int nlive = LiveGetDeviceNameCount();
     int ldev;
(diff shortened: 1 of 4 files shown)


lukashino:
Continues in #8655

PR #8670, opened 2023-04-03
Detect http2 4067 v1

catenacyber:

Describe changes:
- Support some HTTP keywords for HTTP2

Still S-V tests to do

`http.start` is the last one unsupported cf `git grep ALPROTO_HTTP1 src/detect-http*.c | grep DetectSignatureSetAppProto`

Should we warn on a rule `alert http` that is only for HTTP1 or HTTP2 based on its keywords ?


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113181 | 145045 | 128.15% | 

Pipeline 12981

src/detect-http-protocol.c
```diff
@@ -114,6 +114,19 @@ static InspectionBuffer *GetData(DetectEngineThreadCtx *det_ctx,
     return buffer;
 }
 
+static InspectionBuffer *GetData2(DetectEngineThreadCtx *det_ctx,
+        const DetectEngineTransforms *transforms, Flow *_f, const uint8_t _flow_flags, void *txv,
+        const int list_id)
+{
+    InspectionBuffer *buffer = InspectionBufferGet(det_ctx, list_id);
+    if (buffer->inspect == NULL) {
+        InspectionBufferSetup(det_ctx, list_id, buffer, (const uint8_t *) "HTTP/2", strlen("HTTP/2"));
```
victorjulien: maybe the compiler is smart enough to optimize it, but it seems using a runtime length func here is a bit odd.

src/detect-http-protocol.c
```diff
@@ -120,7 +120,8 @@ static InspectionBuffer *GetData2(DetectEngineThreadCtx *det_ctx,
 {
     InspectionBuffer *buffer = InspectionBufferGet(det_ctx, list_id);
     if (buffer->inspect == NULL) {
-        InspectionBufferSetup(det_ctx, list_id, buffer, (const uint8_t *) "HTTP/2", strlen("HTTP/2"));
+        InspectionBufferSetup(
+                det_ctx, list_id, buffer, (const uint8_t *)"HTTP/2", strlen("HTTP/2"));
```
victorjulien, line 124: protocol update should be in different commit it seems, also same strlen comment

src/detect-http-stat-msg.c
```diff
@@ -72,6 +72,19 @@ static InspectionBuffer *GetData(DetectEngineThreadCtx *det_ctx,
         const uint8_t _flow_flags, void *txv, const int list_id);
 static int DetectHttpStatMsgSetupSticky(DetectEngineCtx *de_ctx, Signature *s, const char *str);
 
+static InspectionBuffer *GetData2(DetectEngineThreadCtx *det_ctx,
+        const DetectEngineTransforms *transforms, Flow *_f, const uint8_t _flow_flags, void *txv,
+        const int list_id)
+{
+    InspectionBuffer *buffer = InspectionBufferGet(det_ctx, list_id);
+    if (buffer->inspect == NULL) {
+        InspectionBufferSetup(det_ctx, list_id, buffer, (const uint8_t *)"", 0);
```
victorjulien, line 81: will this match on `bsize:0;`?

src/detect-http-client-body.c
```diff
@@ -76,6 +76,219 @@ static uint8_t DetectEngineInspectBufferHttpBody(DetectEngineCtx *de_ctx,
 static int PrefilterMpmHttpRequestBodyRegister(DetectEngineCtx *de_ctx, SigGroupHead *sgh,
         MpmCtx *mpm_ctx, const DetectBufferMpmRegistery *mpm_reg, int list_id);
 
+static inline InspectionBuffer *FiledataWithXformsGetDataCallback(DetectEngineThreadCtx *det_ctx,
```
victorjulien, line 79: this looks like a copy of file data code. Can we just reuse that?

src/detect-http-protocol.c
catenacyber: What is the good pattern to do this ?

src/detect-http-protocol.c
catenacyber: > protocol update should be in different commit it seems

This is `./scripts/clang-format.sh rewrite-branch`

Redoing git history in next version...

src/detect-http-stat-msg.c
catenacyber: I do not know.
Should it ?

src/detect-http-client-body.c
catenacyber: Yes it is.
Is there something even simpler to do ? Like for detect-http-server-body.c which uses g_file_data_buffer_id directly ?

catenacyber:
Continuing in #8687

src/detect-http-client-body.c
victorjulien: Maybe, it would require a bit of investigation.

catenacyber: Ok, so going for simple reuse of code, not bigger changes in v3...

PR #8678, opened 2023-04-04
stream: improve FIN checking

victorjulien:
After recent next_seq changes, the FIN checks could be too strict leading to stalling sessions in IPS mode.

This patch requires a FIN to be >= last ack and <= next_win to be accepted.

(cherry picked from commit 39a6f411e9748e76211b76b1f45b6c1863012972)


suricata-qa:

Pipeline 13016

PR #8671, opened 2023-04-03
draft: policy: postpone setting of policy after determining engine mode v2

lukashino:
Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5960) ticket:

Describe changes:
- moved evaluation of exception policy prior to the setting of engine mode to make sure that the failing IPS test from #8668 is related to the IPS mode setting and not to finalizing LiveDevices. The expected outcome is passing the test which is in fact incorrect (and the pipeline baseline should be adjusted).

PR #8668, opened 2023-04-02
draft: policy: postpone setting of policy after determining engine mode

lukashino:
Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5960) ticket:

Describe changes:
- postpone exec policy setting

PR #8681, opened 2023-04-04
Engine mode fixes v1 

lukashino:
PR is a collection of:
- Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5957) ticket.
- Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5959) ticket.
- Link to [redmine](https://redmine.openinfosecfoundation.org/issues/5960) ticket.

PR fixes determination of engine mode. The main problem is that some parts of Suricata queries the state of the engine mode before it is determined. They are given the default engine mode state (IDS) while later Suricata determined IPS behavior. This then led to unwanted behavior in IPS mode. 

BPF contained this issue but hasn't really affected Suricata. However, the policy settings did affect Suricata run in IPS mode. This has been tested in PRs #8668 and #8671. #8687 has moved the policy setting only after DeviceFinalization but still before engine mode evaluation. This PR has not triggered different behavior in the pipeline run. #8668 has moved the policy setting after the engine mode evaluation and the behavior of Suricata in IPS mode changed. The bug is also highlighted with the addition of unknown/uninitialized engine mode in combination with the BUG_ON macro.

IPS behavior deviation is expected and I suggest baseline adjustment.

### DIFF
--- a/src/runmode-netmap.c
+++ b/src/runmode-netmap.c
@@ -205,14 +205,6 @@ static int ParseNetmapSettings(NetmapIfaceSettings *ns, const char *iface,
         ns->real = true;
     }
 
-    const char *bpf_filter = NULL;
-    if (ConfGet("bpf-filter", &bpf_filter) == 1) {
-        if (strlen(bpf_filter) > 0) {
-            ns->bpf_filter = bpf_filter;
-            SCLogInfo("%s: using command-line provided bpf filter '%s'", iface, ns->bpf_filter);
-        }
-    }
-
     if (if_root == NULL && if_default == NULL) {
         SCLogInfo("%s: unable to find netmap config for interface \"%s\" or \"default\", using "
                   "default values",
@@ -244,11 +236,19 @@ static int ParseNetmapSettings(NetmapIfaceSettings *ns, const char *iface,
 
     /* load netmap bpf filter */
     /* command line value has precedence */
-    if (ns->bpf_filter == NULL) {
-        if (ConfGetChildValueWithDefault(if_root, if_default, "bpf-filter", &bpf_filter) == 1) {
-            if (strlen(bpf_filter) > 0) {
-                ns->bpf_filter = bpf_filter;
-                SCLogInfo("%s: using bpf filter %s", iface, ns->bpf_filter);
+    const char *bpf_filter = NULL;
+    if (ConfGet("bpf-filter", &bpf_filter) == 1) {
+        if (strlen(bpf_filter) > 0) {
+            ns->bpf_filter = bpf_filter;
+            SCLogInfo("%s: using command-line provided bpf filter '%s'", iface, ns->bpf_filter);
+        }
+    } else { // reading from the file
+        if (ns->bpf_filter == NULL) {
+            if (ConfGetChildValueWithDefault(if_root, if_default, "bpf-filter", &bpf_filter) == 1) {
+                if (strlen(bpf_filter) > 0) {
+                    ns->bpf_filter = bpf_filter;
+                    SCLogInfo("%s: using file provided bpf filter %s", iface, ns->bpf_filter);
+                }
             }
         }
     }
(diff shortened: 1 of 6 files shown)


suricata-qa:
Information:
 

ERROR: QA failed on IPS_AFP_drop_chk.
|--|--|--|--| 
| | IPS_AFP_stats_chk | | |
| .flow.end.state.established | 550799 | 583199 | 105.88% |
| .flow.end.tcp_state.established | 169560 | 201960 | 119.11% | 

Pipeline 13021

src/runmode-pfring.c
```diff
@@ -323,24 +323,28 @@ static void *ParsePfringConfig(const char *iface)
...
-                    SCLogDebug("Going to use bpf filter %s",
-                               pfconf->bpf_filter);
+                    SCLogConfig("%s: using file provided bpf filter %s", pfconf->iface,
+                            pfconf->bpf_filter);
                 }
             }
         }
     }
 
+    if (pfconf->bpf_filter != NULL && EngineModeIsIPS()) {
```
victorjulien, line 344: we don't support IPS for pfring

src/suricata.c
```diff
@@ -208,13 +208,24 @@ int SuriHasSigFile(void)
...
 {
+#ifdef DEBUG
+    BUG_ON(g_engine_mode == ENGINE_MODE_UNKNOWN);
+#endif
     return (g_engine_mode == ENGINE_MODE_IPS);
 }
 
 int EngineModeIsIDS(void)
 {
+#ifdef DEBUG
```
victorjulien, line 226: replace with `DEBUG_VALIDATE_BUG_ON`?

src/suricata.c
```diff
@@ -2671,6 +2682,11 @@ int PostConfLoadedSetup(SCInstance *suri)
     RunModeEngineIsIPS(
             suricata.run_mode, suricata.runmode_custom_mode, suricata.capture_plugin_name);
 
+    if (EngineModeIsUnknown()) { // if still uninitialized the set the default
```
victorjulien, line 2685: Is this really a bug?

src/runmode-netmap.c
```diff
@@ -244,11 +236,19 @@ static int ParseNetmapSettings(NetmapIfaceSettings *ns, const char *iface,
 
     /* load netmap bpf filter */
     /* command line value has precedence */
-    if (ns->bpf_filter == NULL) {
-        if (ConfGetChildValueWithDefault(if_root, if_default, "bpf-filter", &bpf_filter) == 1) {
-            if (strlen(bpf_filter) > 0) {
-                ns->bpf_filter = bpf_filter;
-                SCLogInfo("%s: using bpf filter %s", iface, ns->bpf_filter);
+    const char *bpf_filter = NULL;
```
victorjulien, line 239: this block looks virtually the same as the afpacket one. Should we use a util func?

victorjulien (changes requested):
The commits are a bit low on explanation. The PR message is not becoming part of the git history, so please make sure to move the explanations from there into the relevant commits.

src/suricata.c
lukashino: No this is meant to ensure the default engine mode is set when it hasn't been set until now.

But the whole PR is about making sure that Suricata modules don't query engine mode while uninitialized. 
If it is not initialized by other Suricata ways (e.g. config, command-line parameter) then set the default. But after this line, the engine mode (should) never change.

src/suricata.c
lukashino: Done

src/runmode-netmap.c
lukashino: Done

lukashino:
Continues in #8684

src/runmode-pfring.c
lukashino: removed the bpf_filter test from the condition but left the IPS test to stop user from running Suricata with PF_RING and IPS mode.

PR #8687, opened 2023-04-05
Detect http2 4067 v2

catenacyber:

Describe changes:
- Support some HTTP keywords for HTTP2

`http.start` is the last one unsupported cf `git grep ALPROTO_HTTP1 src/detect-http*.c | grep DetectSignatureSetAppProto`

Should we warn on a rule `alert http` that is only for HTTP1 or HTTP2 based on its keywords ?

suricata-verify-pr: 1164
https://github.com/OISF/suricata-verify/pull/1164

Modifies #8670 with git history cleant (no fix of clang format script)

### DIFF
--- a/src/detect-http-stat-msg.c
+++ b/src/detect-http-stat-msg.c
@@ -72,6 +72,19 @@ static InspectionBuffer *GetData(DetectEngineThreadCtx *det_ctx,
         const uint8_t _flow_flags, void *txv, const int list_id);
 static int DetectHttpStatMsgSetupSticky(DetectEngineCtx *de_ctx, Signature *s, const char *str);
 
+static InspectionBuffer *GetData2(DetectEngineThreadCtx *det_ctx,
+        const DetectEngineTransforms *transforms, Flow *_f, const uint8_t _flow_flags, void *txv,
+        const int list_id)
+{
+    InspectionBuffer *buffer = InspectionBufferGet(det_ctx, list_id);
+    if (buffer->inspect == NULL) {
+        InspectionBufferSetup(det_ctx, list_id, buffer, (const uint8_t *)"", 0);
+        InspectionBufferApplyTransforms(buffer, transforms);
+    }
+
+    return buffer;
+}
+
 /**
  * \brief Registration function for keyword: http_stat_msg
  */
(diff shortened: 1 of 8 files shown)


suricata-qa:

Pipeline 13033

src/detect-http-stat-msg.c
```diff
@@ -72,6 +72,19 @@ static InspectionBuffer *GetData(DetectEngineThreadCtx *det_ctx,
         const uint8_t _flow_flags, void *txv, const int list_id);
 static int DetectHttpStatMsgSetupSticky(DetectEngineCtx *de_ctx, Signature *s, const char *str);
 
+static InspectionBuffer *GetData2(DetectEngineThreadCtx *det_ctx,
+        const DetectEngineTransforms *transforms, Flow *_f, const uint8_t _flow_flags, void *txv,
+        const int list_id)
+{
+    InspectionBuffer *buffer = InspectionBufferGet(det_ctx, list_id);
+    if (buffer->inspect == NULL) {
+        InspectionBufferSetup(det_ctx, list_id, buffer, (const uint8_t *)"", 0);
```
victorjulien, line 81: would like to see some SV tests that show how this behaves with:
- `bsize:0;`
- `http.stat_msg; content:!"nomatch"`.
(didn't check the SV tests referenced in the PR yet)

An alternative approach could be a "magic" value, like how we lot a magic URI in http tx logs when we missed the request, IIRC `/libhtp::request_not_seen` or something.

catenacyber: So, this never matches on all of the following
```
alert http2 any any -> any any (http.stat_msg; bsize: 0; sid:42;)
alert http2 any any -> any any (http.stat_msg; content: "a"; sid:43;)
alert http2 any any -> any any (http.stat_msg; bsize: 1; sid:44;)
alert http2 any any -> any any (http.stat_msg; content: !"a"; sid:45;)
```

If I put a dummy magic value, I can get 44 and 45 to trigger.
I think we do not want 44 to trigger...
Do we want 42 and 45 to trigger ?

catenacyber:
Continued in https://github.com/OISF/suricata/pull/8704

src/detect-http-stat-msg.c
victorjulien: I think 45 is the usecase  Brandon was after, esp in combination with `alert http ... ` and additional http matches.

catenacyber: Ok, is this not already filed under another ticket ?
This being the fact to that a negated match triggers on an empty buffer

catenacyber: Maybe the solution is to have an empty 0-length buffer still triggering the evaluation

victorjulien: > Maybe the solution is to have an empty 0-length buffer still triggering the evaluation

Can you do open a ticket on this?

catenacyber: https://github.com/OISF/suricata/pull/8770 solves https://redmine.openinfosecfoundation.org/issues/6025

PR #8693, opened 2023-04-06
Rfb frames v13

hsadia538:
- [X] I have read the contributing guide lines at https://suricata.readthedocs.io/en/latest/devguide/codebase/contributing/contribution-process.html
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5717

Previous PR: https://github.com/OISF/suricata/pull/8526

Describe changes:

- Rebase branch
- Update function signature in tests to reflect the changes made due to frames

suricata-verify-pr: 1117

PR #8694, opened 2023-04-07
Rfb frames v14

hsadia538:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5717

Previous PR: https://github.com/OISF/suricata/pull/8693

Describe changes:

- Rebase branch
- Update function signature in tests to reflect the changes made due to frames
- Fix commit order

suricata-verify-pr: 1117

### DIFF
--- a/rust/src/rfb/rfb.rs
+++ b/rust/src/rfb/rfb.rs
@@ -19,7 +19,8 @@
 
 use std;
 use std::ffi::CString;
-use crate::core::{ALPROTO_UNKNOWN, AppProto, Flow, IPPROTO_TCP};
+use crate::core::{ALPROTO_UNKNOWN, AppProto, Flow, IPPROTO_TCP, STREAM_START};
+use crate::frames::*;
 use crate::applayer;
 use crate::applayer::*;
 use nom7::Err;
@@ -27,6 +28,10 @@ use super::parser;
 
 static mut ALPROTO_RFB: AppProto = ALPROTO_UNKNOWN;
 
+#[derive(AppLayerFrameType)]
+pub enum RFBFrameType {
+    Pdu,
+}
 pub struct RFBTransaction {
     tx_id: u64,
     pub complete: bool,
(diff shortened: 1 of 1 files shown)


hsadia538:
I have made a single commit for both frames and changed the signature or parse_request and parse_response functions in tests. This is not ideal as a single commit should reflect one change but I did this to avoid a build dependency that will cause both to fail if I do them separately.
For the frames PR to build successfully the tests should use the updated function signature
For the tests to have the updated function signature, frames should already exist.
This PR is open to suggestions especially if there is a better way to approach this.

hsadia538:
Check rust is failing. When I checked locally it deletes the STREAM_START import to fix the code which throws `the not found in scope error.`
I am importing this to use it in the function calls. not sure why cargo is deleting it to fix the code.
image

inashivb:
> I have made a single commit for both frames and changed the signature or parse_request and parse_response functions in tests. This is not ideal as a single commit should reflect one change but I did this to avoid a build dependency that will cause both to fail if I do them separately. For the frames PR to build successfully the tests should use the updated function signature For the tests to have the updated function signature, frames should already exist. This PR is open to suggestions especially if there is a better way to approach this.

I'm unable to see the relationship b/w fn sig and frames 🤔 
Did you first change the fn signature and then tried to add frames and modify the code?

inashivb:
> Check rust is failing. When I checked locally it deletes the STREAM_START import to fix the code which throws `the not found in scope error.` I am importing this to use it in the function calls. not sure why cargo is deleting it to fix the code. 

Not sure if cargo deletes stuff like that.. This failure seems to be coming from tests which require their own set of imports because of different scope of the module. Tests are for example only run when you specifically tell cargo to run them and then they do their own setup.. Could you please give that a try?

hsadia538:
> > I have made a single commit for both frames and changed the signature or parse_request and parse_response functions in tests. This is not ideal as a single commit should reflect one change but I did this to avoid a build dependency that will cause both to fail if I do them separately. For the frames PR to build successfully the tests should use the updated function signature For the tests to have the updated function signature, frames should already exist. This PR is open to suggestions especially if there is a better way to approach this.
> 
> I'm unable to see the relationship b/w fn sig and frames 🤔 Did you first change the fn signature and then tried to add frames and modify the code?

Apologies for not being clear by fn signature I mean when I call the parse_request and parse_response functions in tests. Before frame addition, they have fewer parameters and don't require a Streamslice to be passed. the frames PR modifies it and hence creates a conflict between the two. that's why I am trying to modify both in same PR

hsadia538:
followed by #8703

PR #8716, opened 2023-04-12
pcap: fix return check

victorjulien:
The check that meant to check if pcap_dispatch processed fewer packets than the desired number was inaccurate. It would also include all errors (negative return values).

This patch considers only positive values for this check.


victorjulien:
CIFuzz fail is unrelated.

PR #5034, opened 2020-06-07
sip: extend parser over tcp v2

glongo:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata-ids.org/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3351

Describe changes:
- Patchset rebased 
- Commit https://github.com/OISF/suricata/pull/4384/commits/1ea9085ca306b5beeb3f05230de71880fa1cae6f removed, output looks correct even without it
- Added detect flags

[PRScript](https://redmine.openinfosecfoundation.org/projects/suricata/wiki/PRscript) output (if applicable):
- PR glongo: https://buildbot.openinfosecfoundation.org/builders/glongo/builds/225
- PR glongo-pcap: https://buildbot.openinfosecfoundation.org/builders/glongo-pcap/builds/89

#suricata-verify-pr: https://github.com/OISF/suricata-verify/pull/155

PR #8725, opened 2023-04-13
rust/sip: register parser for tcp v3

glongo:

https://redmine.openinfosecfoundation.org/issues/3351

Previous PR: https://github.com/OISF/suricata/pull/5034

Describe changes:
The main purpose of this patch is to register sip parser to work over tcp protocol, taking care of handling data before calling the request/response parsers.

In addition, the following minor fixes/improvements are made:

- Accept characters that are not allowed atm
- Register logger to log sip messages coming over tcp
- Define SIP_PORTS var to be used within rules
- Update JSON schema
- Set direction for tx_data

#suricata-verify-pr: https://github.com/OISF/suricata-verify/pull/1171
suricata-verify-pr: 1171

### DIFF
--- a/rust/src/sip/sip.rs
+++ b/rust/src/sip/sip.rs
@@ -20,7 +20,7 @@
 use crate::frames::*;
 use crate::applayer::{self, *};
 use crate::core;
-use crate::core::{AppProto, Flow, ALPROTO_UNKNOWN};
+use crate::core::{AppProto, Flow, ALPROTO_UNKNOWN, ALPROTO_FAILED};
 use crate::sip::parser::*;
 use nom7::Err;
 use std;
@@ -86,9 +86,9 @@ impl SIPState {
         self.transactions.clear();
     }
 
-    fn new_tx(&mut self, _direction: crate::core::Direction) -> SIPTransaction {
+    fn new_tx(&mut self, direction: crate::core::Direction) -> SIPTransaction {
         self.tx_id += 1;
-        SIPTransaction::new(self.tx_id)
+        SIPTransaction::new(self.tx_id, direction)
     }
 
     fn get_tx_by_id(&mut self, tx_id: u64) -> Option<&SIPTransaction> {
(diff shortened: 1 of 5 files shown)


rust/src/sip/sip.rs
```diff
@@ -147,6 +151,55 @@ impl SIPState {
...
+            }
+            Err(Err::Incomplete(_needed)) => {
+                let consumed = input.len();
+                let needed_estimation = input.len() + 1;
+                SCLogDebug!(
+                    "Needed: {:?}, estimated needed: {:?}",
+                    _needed,
+                    needed_estimation
+                );
+                self.set_event(SIPEvent::IncompleteData);
```
jasonish, line 181: For TCP I don't think this makes much sense. In theory, the data could trickle in one byte at a time but still be valid.

rust/src/sip/sip.rs
```diff
@@ -147,6 +151,55 @@ impl SIPState {
...
+        );
+        SCLogDebug!("ts: pdu {:?}", _pdu);
+
+        match sip_parse_request(input) {
+            Ok((_, request)) => {
+                sip_frames_ts(flow, &stream_slice, &request);
+                self.append_request(input, request);
+            }
+            Err(Err::Incomplete(_needed)) => {
+                let consumed = input.len();
```
jasonish, line 174: Consumed should only be updated after a successful parse as we're letting the engine know it can now discard that data.

glongo: What would you suggest to do here? Is `return AppLayerResult::ok()` enough?

rust/src/sip/sip.rs
glongo: Ok, let's remove it.

jasonish (changes requested):
Null comment so I can set changes requested.

rust/src/sip/sip.rs
glongo: Would you please describe what kind of behaviour have you seen?

jasonish: > Would you please describe what kind of behaviour have you seen?

Nothing specific to SIP I can pinpoint, but other protocols like DNS, in particular multiple DNS requests in TCP packets.  So just wouldn't surprise me if it happened in SIP.  But its still valid for any TCP protocol anyways, so its a case we would want to support, for TCP in general.

PR #8728, opened 2023-04-14
6.0.x - use vendored fork of der-parser - v1

jasonish:
Use a fork of der-parser that has a branch with the code that fails to compile
disabled.

Ticket: #5986

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1071,6 +1071,7 @@ jobs:
                 automake \
                 autoconf \
                 build-essential \
+                cbindgen \
                 ccache \
                 curl \
                 git \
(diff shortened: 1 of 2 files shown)


catenacyber:
Why is this a draft ?

jasonish:
> Why is this a draft ?

Its more proof of concept of how we might fork an upstream if there is a no chance of them creating a point release.

What we might want to consider:
- Do we care about Debian 10? If so, this is probably what we need to do.
- These forks of upstream projects should probably go under the OISF repo, not my personal.

catenacyber:
Ok, so POC looks good.
But we should hope not to fall into this situation, right ?
Anything we need to do, or just close ?

jasonish:
@victorjulien Did we decide to give up on Debian 10? Can't recall what the decision was there.

jasonish:
Closing for now.

PR #8736, opened 2023-04-17
ci - better vars, use bundle.sh - v6

jasonish:
- Use better variable like names for set CI parameters (example below).
- Use bundle.sh to pull in libhtp, suricata-update.  It respects these
  environment variables so the actions only need to pass the variables on.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```


jasonish:
See https://github.com/OISF/suricata/pull/8750.

PR #8745, opened 2023-04-20
Suricata library v5

amirabell:

- [x ] I have read the contributing guide lines at https://suricata.readthedocs.io/en/latest/devguide/codebase/contributing/contribution-process.html
- [x ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/

Describe changes:
- First commits to run suricata in library mode: define initial set of API and callbacks to
- invoke for corresponding EVE events
- Many more functionalities will come in following PRs.

Previous PR: https://github.com/OISF/suricata/pull/8701

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:


catenacyber:
Is there a redmine ticket ?
As I quickly glance it, the main idea is the addition of callbacks, so that other software can act at some events, do I get it right ?

amirabell:
I have not created a redmine ticket for it, but happy to do it if required. 

You got it right. There two main concepts here:
 * The worker threads are not managed by suricata but by an external client, which is in charge of the packet acquisition.
 *  As you said, the client receive callbacks for events instead of JSON logging.

catenacyber:
See https://redmine.openinfosecfoundation.org/issues/2693 and related tickets I guess

Thanks again for this interesting contribution.

I also wonder if this PR can be more rust than C...

> As you said, the client receive callbacks for events instead of JSON logging.

Trying to understand the logic, I have a few questions which may be dummy :

If it happens at logging, and if it is a replacement (instead of an addition), could we rather hack `JsonAlertLogger` ?
It looks like it is rather a new type of output : instead of file or socket, it is a callback...

Why do we need a new `AlertEvent` ? Cannot Suricata supply the data it already has, and the caller can process it in the new structure it wants

amirabell:
No dummy questions at all, it can be a bit overwhelming to go through the patch without any context.
Also notice that these are just first couple of commits. There are many more to come with additional callbacks/API/documentaiton, but I have been asked by @victorjulien  to break it down in subsequent PRs.

There are two main reason I wanted to avoid to hack the EVE logger modules:
 * I wanted to modify as little as possible of the existing suricata codebase to avoid conflicts
 * It is (as you said) a new kind of output based on callbacks. The idea is to provide to the client an object (e.g `AlertEvent`) instead of a JSON line to give the client a bit more flexibility on how to process the event without the need for JSON deserialisation, at least for some of the events. The event callbacks are split into two categories: the most important ones (e.g alert, fileinfo, flow, http) will provide an object such as `AlertEvent`. The other ones would still provide a JSON line similar to what done by the corresponding EVE logger (this second type of callback is not introduced by this PR).

amirabell:
> As a callback, is there not a third option to supply the Suricata objects directly ? (like the signature, the transaction, the flow, etc...)

They are grouped as an object to avoid having too many parameters in the callback. Or am I missing something?

> Ok, not obvious to me, what the less/most conflicting way is
Also notice there are callbacks for EVE  events but there might be callbacks for other cases (in future PRs).

catenacyber:
> They are grouped as an object to avoid having too many parameters in the callback. Or am I missing something?

I am just wondering if this should be done in libsuricata or by its caller (using the objects)

> Also notice there are callbacks for EVE events but there might be callbacks for other cases (in future PRs).

Ok, I had not noticed. What kind of cases ?

src/output-callback-alert.c
```diff
@@ -0,0 +1,184 @@
...
+            AlertAddAppLayer(p, pa->tx_id, &event);
+
+            /* TODO: Add file info? */
+
+            /* TODO: Add flow info? */
+        }
+
+        /* Invoke callback and cleanup */
+        tv->callbacks->alert.func(tv->callbacks->alert.user_ctx, &event);
+        AlertCleanupAppLayer(p, pa->tx_id, &event);
```
traceflight, line 130: Interesting work😄!  Tested and found some issues.
This function may cause segmentation fault when p->flow is null.

traceflight:
It seems packed pool are not initialized for worker threads. This will case deadlock in some case.

src/output-callback-alert.c
traceflight: Maybe we need more discussion. And I think you’re doing **nice** work. I’d like to see all your contribution.

catenacyber: @jasonish Angelo is asking how to proceed. What do you think ?

jasonish:
Rebased here: https://github.com/OISF/suricata/pull/9428

PR #8757, opened 2023-04-24
[draft] Example of catching panics in Rust

jasonish:


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 112875 | 145123 | 128.57% |
| .app_layer.error.dns_tcp.parser | 1 | 2630 | 263000.0% |
| .app_layer.error.dns_tcp.parser | 30 | 11551 | 38503.33% | 

Pipeline 13305

rust/src/dns/dns.rs
```diff
@@ -769,8 +770,10 @@ pub unsafe extern "C" fn rs_dns_parse_request(
     flow: *const core::Flow, state: *mut std::os::raw::c_void, _pstate: *mut std::os::raw::c_void,
     stream_slice: StreamSlice, _data: *const std::os::raw::c_void,
 ) -> AppLayerResult {
-    let state = cast_pointer!(state, DNSState);
-    state.parse_request_udp(flow, stream_slice);
+    let _ = catch_unwind(|| {
+        let state = cast_pointer!(state, DNSState);
+        state.parse_request_udp(flow, stream_slice);
+    });
```
inashivb, line 776: I'm trying to understand why we're using it here. I'm reading `Don't build your programs to unwind under normal circumstances. Ideally, you should only panic for programming errors or extreme problems.` at https://doc.rust-lang.org/nomicon/unwinding.html where I also see `You must absolutely catch any panics at the FFI boundary!` so I kind of see why we want to do this. But,
1. What points of panic/undefined behavior can happen on the enclosed code?
2. How do we know that the panic caused here for whatever reason would be implemented with unwinding? Or, do we not care about it as it is not incurring any runtime cost in either case?

Thank you for this! :)

victorjulien: For reference https://redmine.openinfosecfoundation.org/issues/3333

jasonish: > I'm trying to understand why we're using it here. I'm reading `Don't build your programs to unwind under normal circumstances. Ideally, you should only panic for programming errors or extreme problems.` at https://doc.rust-lang.org/nomicon/unwinding.html where I also see `You must absolutely catch any panics at the FFI boundary!` so I kind of see why we want to do this. But,
> 
> 1. What points of panic/undefined behavior can happen on the enclosed code?

Simple ones that we can catch would be programming errors like walking off the end of an array. In C this might create a segfault, but is a catchable error in Rust.  These warnings are more to prevent people from using unwinding as a flow control mechanism, or treating them like exceptions when coming from C++, Python, etc.

I also disagree about the *must* on FFI boundaries.  Currently, we crash if the Rust code panics at the FFI boundary, and that's a reasonable outcome.

> 2. How do we know that the panic caused here for whatever reason would be implemented with unwinding? Or, do we not care about it as it is not incurring any runtime cost in either case?

We don't care.  If the panic is unwindable we'll catch it and not crash.  If it's not catchable, perhaps a deep-rooted segfault, or an illegal instruction the program should abort and crash.

All the Rust web frameworks like Axum, Rocket, Warp, implement such unwinding in their request handlers, so one bad request handler won't take down the server, this is a very similar approach.

rust/src/dns/dns.rs
```diff
@@ -790,27 +795,42 @@ pub unsafe extern "C" fn rs_dns_parse_request_tcp(
...
-        return state.parse_response_tcp(flow, stream_slice);
+    match catch_unwind(|| {
+	let x: Vec<bool> = vec![];
+        let state = cast_pointer!(state, DNSState);
+        if stream_slice.is_gap() {
+            state.response_gap(stream_slice.gap_size());
+        } else if !stream_slice.is_empty() {
+            return state.parse_response_tcp(flow, stream_slice);
+        }
+	if x[99] {}
```
jasonish, line 825: Oops, I didn't mean to commit this introduce error, this is why QA is so far off.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 112875 | 145124 | 128.57% |
| .flow.memuse | 528023824 | 938040360 | 177.65% | 

Pipeline 13347

rust/src/dns/dns.rs
inashivb: Thank you for the explanation!
Do I get it right that if there is actually an error that can be unwinded, it'll affect the performance negatively?
Do we log it in stats somehow to indicate what's causing the possible slowdown?

jasonish: > Do I get it right that if there is actually an error that can be unwinded, it'll affect the performance negatively?

Yes, it should a little as it has to do some resetting of data. Plus that line that gets log that we can't control is slow.  All prints to the console are slow though.

The alternative is crashing.

victorjulien:
I think this looks good for using with big expensive calls like going into the parser. What are your thoughts about the frequent much smaller calls like into json builder? Can we wrap these at minimal cost too, or should we rework those to not be able to panic (to a reasonable level of course)?

catenacyber:
> IIRC catenacyber was seeing panics in jsonbuilder when experimenting with alloc failures.

Indeed cf https://github.com/google/oss-fuzz/pull/9902

jasonish:
> I think this looks good for using with big expensive calls like going into the parser. What are your thoughts about the frequent much smaller calls like into json builder? Can we wrap these at minimal cost too, or should we rework those to not be able to panic (to a reasonable level of course)?
> 
> IIRC @catenacyber was seeing panics in jsonbuilder when experimenting with alloc failures.

I will have to do some tests around `JsonBuilder` but its worth doing.  For something like `JsonBuilder` it makes sense to hand roll this into the API but for parsers that follow an interface for being called into, I'm hoping to abstract this out so each individual parser doesn't need to worry about it - this is something I'm currently looking at, but may be quite intrusive.

catenacyber:
For `JsonBuilder`, there should only be a few places with allocations

What is the solution ?
https://github.com/microsoft/rust_fallible_vec ?

jasonish:
> For `JsonBuilder`, there should only be a few places with allocations
> 
> What is the solution ? https://github.com/microsoft/rust_fallible_vec ?

I think the idea is for more than just allocations, just any panic that can be unwound.

As all the `jb` functions currently return an error code, we'd just wrap them and return error if anything unwindable happened, which I assume would include allocation errors.

That crate could be interesting to look at, but this unwinding is a little more generic.

catenacyber:
Indeed I get it

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113274 | 145130 | 128.12% | 

Pipeline 13683

rust/src/jsonbuilder.rs
```diff
@@ -604,7 +647,11 @@ impl JsonBuilder {
     /// than building onto the buffer.
     #[inline(always)]
     fn encode_string(&mut self, val: &str) -> Result<(), JsonError> {
-        let mut buf = vec![0; val.len() * 2 + 2];
+	let size = val.len() * 2 + 2;
+        let mut buf = Vec::new();
```
victorjulien, line 651: can this fail as well?

rust/src/jsonbuilder.rs
```diff
@@ -651,38 +699,45 @@ impl JsonBuilder {
...
-    return out;
+    return Ok(out);
 }
 
 #[no_mangle]
 pub extern "C" fn jb_new_object() -> *mut JsonBuilder {
-    let boxed = Box::new(JsonBuilder::new_object());
-    Box::into_raw(boxed)
+    match JsonBuilder::try_new_object() {
+        Ok(js) => Box::into_raw(Box::new(js)),
```
victorjulien, line 730: is `Box::new()` also doing an alloc?

suricata-qa:

Pipeline 13684

rust/src/jsonbuilder.rs
jasonish: It is.  `Box::try_new` is an experimental API that would be ideal here, but until then it should probably be wrapped in a panic handler.  I need to actually look into if Rust memory allocations panic or abort, the latter not being catchable.

jasonish:
I guess the issue of handling memory allocation errors is still in question.  Rust has these `try_reserve` APIs in many places that do this in a sane way (see the new JsonBuilder), but outside of that I don't think memory allocation errors can be reliably caught, as Rust will abort, rather than panic.  The reason being, if the system is truly out of memory, it can't do anything to return a value, or provide that state for a panic handler to unwind into a known state.

So when it comes to allocated memory for data, we should do our best with `try_reserve` and like methods, but unwinding panics would be more about programming errors, than memory errors.

suricata-qa:

Pipeline 13705

victorjulien:
@catenacyber are you able to test how this behaves in your alloc tests?

jasonish:
> @catenacyber are you able to test how this behaves in your alloc tests?

This would probably be the better one to test: https://github.com/OISF/suricata/pull/8855

I'm going to close this one and and this branch of pull requests will become more for catching panics in parsers..

PR #8759, opened 2023-04-25
smtp: return on line completion

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/6019

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -1139,6 +1139,13 @@ static int SMTPProcessRequest(SMTPState *state, Flow *f, AppLayerParserState *ps
     SCEnter();
     SMTPTransaction *tx = state->curr_tx;
 
+    /* If current input is to be discarded because it completes a long line,
+     * line's length and delimeter len are reset to 0. Skip processing this line.
+     * This line is only to get us out of the state where we should discard any
+     * data till LF. */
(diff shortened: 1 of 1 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 112875 | 144994 | 128.46% | 

Pipeline 13307

src/app-layer-smtp.c
```diff
@@ -1139,6 +1139,13 @@ static int SMTPProcessRequest(SMTPState *state, Flow *f, AppLayerParserState *ps
     SCEnter();
     SMTPTransaction *tx = state->curr_tx;
 
+    /* If current input is to be discarded because it completes a long line,
+     * line's length and delimeter len are reset to 0. Skip processing this line.
+     * This line is only to get us out of the state where we should discard any
+     * data till LF. */
+    if (line->len == 0 && line->delim_len == 0) {
+        return 0;
+    }
```
inashivb, line 1148: toserver_data_count should be incremented before this.
Also, increment it in case of long line where we just return from GetLine fn bypassing this fn.

victorjulien:
The commit message is not explaining the issue. Please update it to explain:
- the problem
- how this patch solves the issue

PR #8765, opened 2023-04-26
detect: fix break loop for timeout based rule reload - v1

jasonish:
As part of 6d8b50b748844e9de6010cde5a6b139148c0e937, the settings of
THV_CAPTURE_INJECT_PKT ended up in a location unreachable by capture
methods that did not have PktAcqBreakLoop. Refactor so this flag is
set outside of TmThreadsCaptureBreakLoop for such capture methods like
AF_PACKET.

This fixes the case where read threads won't "break" for rule reloads
until packets are seen.

Ticket: https://redmine.openinfosecfoundation.org/issues/6021

### DIFF
--- a/src/tm-threads.h
+++ b/src/tm-threads.h
@@ -253,6 +253,10 @@ static inline void TmThreadsCaptureHandleTimeout(ThreadVars *tv, Packet *p)
         tv->tmqh_out(tv, p);
 }
 
+/**
+ * Note: Only ever called if tv->break_loop is true and
+ *   tm->PktAcqBreakLoop is set.
+ */
 static inline void TmThreadsCaptureBreakLoop(ThreadVars *tv)
 {
     if (unlikely(!tv->break_loop))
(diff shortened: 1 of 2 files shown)


suricata-qa:

Pipeline 13348

src/tm-threads.h
```diff
@@ -253,6 +253,10 @@ static inline void TmThreadsCaptureHandleTimeout(ThreadVars *tv, Packet *p)
         tv->tmqh_out(tv, p);
 }
 
+/**
+ * Note: Only ever called if tv->break_loop is true and
```
victorjulien, line 257: is this is correct all the if's below could be (debug) asserts

tugtugtug: For my local fix, I simply removed the ` if (unlikely(!tv->break_loop))` in `TmThreadsCaptureBreakLoop` and changed the `} else if (detect_tvs[i]->break_loop) {` to `} else {`.

I think that is better as:
1. the logic of break loop is still within a single function.
2. there are already extra checks of `tv->break_loop` in other use cases.
3. the `if (tm->PktAcqLoop && tm->PktAcqBreakLoop)` basically ensures `tv->break_loop` is true.

victorjulien: I agree it is better to keep the capture specific logic out of the detect engine as much as possible.

PR #8767, opened 2023-04-26
App layer cleanup v1

catenacyber:

Describe changes:
- debug: use function `AppProtoToString` instead of recoding it twice
- fix setup-app-layer.py and remove unused code
- app-layer: use an array for app-layer proto strings

By the way, there is also output-json-alert.c that should be patched by setup-app-layer.py
But I plan for a bigger simplification of the transaction logging code, basically adding an array indexed by app-layer proto and having the name (of the json object), and the logging function (which takes a transaction pointer, a jsonbuilder pointer, maybe some other stuff for some protocols like state, flags... and returning a bool) 
Then we can use it in both `JsonGenericLogger` and `AlertAddAppLayer`

### DIFF
--- a/src/app-layer-protos.c
+++ b/src/app-layer-protos.c
@@ -25,122 +25,52 @@
 #include "suricata-common.h"
 #include "app-layer-protos.h"
 
-#define CASE_CODE(E)  case E: return #E
+const char *AppProtoStrings[ALPROTO_MAX] = {
+    "unknown",        // ALPROTO_UNKNOWN,
+    "http",           // ALPROTO_HTTP1
+    "ftp",            // ALPROTO_FTP
+    "smtp",           // ALPROTO_SMTP
+    "tls",            // ALPROTO_TLS
+    "ssh",            // ALPROTO_SSH
+    "imap",           // ALPROTO_IMAP
+    "jabber",         // ALPROTO_JABBER
+    "smb",            // ALPROTO_SMB
+    "dcerpc",         // ALPROTO_DCERPC
+    "irc",            // ALPROTO_IRC
+    "dns",            // ALPROTO_DNS
+    "modbus",         // ALPROTO_MODBUS
+    "enip",           // ALPROTO_ENIP
+    "dnp3",           // ALPROTO_DNP3
+    "nfs",            // ALPROTO_NFS
+    "ntp",            // ALPROTO_NTP
+    "ftp-data",       // ALPROTO_FTPDATA
+    "tftp",           // ALPROTO_TFTP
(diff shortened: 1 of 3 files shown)


src/app-layer-protos.c
```diff
@@ -151,74 +81,11 @@ AppProto StringToAppProto(const char *proto_name)
...
-    if (strcmp(proto_name, "rdp") == 0)
-        return ALPROTO_RDP;
-    if (strcmp(proto_name, "http2") == 0)
-        return ALPROTO_HTTP2;
-    if (strcmp(proto_name, "bittorrent-dht") == 0)
-        return ALPROTO_BITTORRENT_DHT;
-    if (strcmp(proto_name, "failed") == 0)
-        return ALPROTO_FAILED;
+    // We could use a Multi Pattern Matcher
+    for (size_t i = 0; i < sizeof(AppProtoStrings) / sizeof(const char *); i++) {
```
victorjulien, line 85: think `sizeof(AppProtoStrings[0])` is a common pattern to do this. IIRC we even have a macro for this? Not sure.

catenacyber: Nice, pushed a fixup commit  already ;-)

src/app-layer-protos.c
```diff
@@ -25,122 +25,52 @@
 #include "suricata-common.h"
 #include "app-layer-protos.h"
 
-#define CASE_CODE(E)  case E: return #E
+const char *AppProtoStrings[ALPROTO_MAX] = {
+    "unknown",        // ALPROTO_UNKNOWN,
+    "http",           // ALPROTO_HTTP1
```
jasonish, line 30: The case statement had the safety that a refactor/re-ordering would keep the values consistent.  How do we ensure that now?

jasonish:
The build now fails after this line from CI:
```
      - run: ./scripts/setup-app-layer.py --parser --logger --detect FooBar payload
```
with:
```
output.c:81:10: fatal error: output-json-foobar-rust.h: No such file or directory
   81 | #include "output-json-foobar-rust.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

As the generated file is, and should be `output-json-foobar.h`.

src/app-layer-protos.c
catenacyber: That is a good point.

We can have a visual verification by making this a list of tuple, AppProto + string, and debug validate that the index is the value of the AppProto.

What do you think of it ?

catenacyber:
Continues in https://github.com/OISF/suricata/pull/8772

PR #8744, opened 2023-04-20
rust: remove unused

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None, generic cleaning

Describe changes:
- rust: remove unused dead code

Found with
`git grep 'pub ' rust/src/ | cut -d: -f1 | uniq | xargs sed -i -e 's/pub /pub(crate) /'` then see rust warnings when compiling

There is more to it :
- https://redmine.openinfosecfoundation.org/issues/4517 `APP_LAYER_PARSER_NO_INSPECTION_PAYLOAD` and like should only be defined on the rust side
- There are parsed fields which get never used. (like `pad` in DCERPCState) Coming in a next PR
- dcerpc defines lots of unused constants like PFC_RESERVED_1. Should these get removed as well ? cc @inashivb ?
- `AuthenticationKerb5` and like are never constructed for pgsql. Should they be @jufajardini ?

PR #8771, opened 2023-04-27
Rust unused v2

catenacyber:
None, generic cleaning

Describe changes:

Found with

There is more to it :
- `AuthenticationKerb5` and like are never constructed for pgsql. Should they be @jufajardini ?

Follows #8744 with better commit history split in 2


suricata-qa:

Pipeline 13448

rust/src/conf.rs
```diff
@@ -103,29 +101,6 @@ impl ConfNode {
         return Self { conf }
     }
 
-    pub fn get_child_value(&self, key: &str) -> Option<&str> {
```
jasonish, line 106: Not yet, possible public API usage.

rust/src/ffi/hashing.rs
```diff
@@ -24,10 +24,6 @@ use std::os::raw::c_char;
 pub const SC_SHA1_LEN: usize = 20;
 pub const SC_SHA256_LEN: usize = 32;
 
-// Length of hex digests without trailing NUL.
-pub const SC_MD5_HEX_LEN: usize = 32;
-pub const SC_SHA256_HEX_LEN: usize = 64;
```
jasonish, line 29: Would rather see if we can remove this from `src/rust.h` instead, and have these exported to C.

rust/src/jsonbuilder.rs
```diff
@@ -328,36 +328,6 @@ impl JsonBuilder {
         }
     }
 
-    /// Add a byte array to a JSON array encoded as hex.
-    pub fn append_hex(&mut self, val: &[u8]) -> Result<&mut Self, JsonError> {
```
jasonish, line 332: Not yet, possible public API usage.

jasonish (changes requested):
I'd like to see this proceed with stuff that is only ever used internally Suricata, but still leave stuff that forms part of a public API, even if for completeness.

For example, unused code in a parser will liley never be needed for public API, or code that C bindings for code that only ever existed  for parsers is probably good to go.

rust/src/conf.rs
catenacyber: Ok

rust/src/jsonbuilder.rs
catenacyber: ok

rust/src/ffi/hashing.rs
catenacyber: Well, they are already exported to C, so we can remove them from rust.h no problem ✌️

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8783

PR #8772, opened 2023-04-27
App layer cleanup v2

catenacyber:

Describe changes:
- debug: use function `AppProtoToString` instead of recoding it twice
- fix setup-app-layer.py and remove unused code
- app-layer: use an array for app-layer proto strings

By the way, there is also output-json-alert.c that should be patched by setup-app-layer.py
But I plan for a bigger simplification of the transaction logging code, basically adding an array indexed by app-layer proto and having the name (of the json object), and the logging function (which takes a transaction pointer, a jsonbuilder pointer, maybe some other stuff for some protocols like state, flags... and returning a bool) 
Then we can use it in both `JsonGenericLogger` and `AlertAddAppLayer`

Follows #8767 by 
- fixing unit tests (handling special case for HTTP/HTTP1 not having the same string as input and output)
- using `ARRAY_SIZE`
- having better future-proof consistency of `AppProtoStrings`
- removing src/output-json-template.c


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113274 | 145157 | 128.15% | 

Pipeline 13437

jasonish:
Still some issues in the generated output.  I think I have it all fixed up, see my 2 commits here: https://github.com/jasonish/suricata/pull/174/commits

catenacyber:
> Still some issues in the generated output. I think I have it all fixed up, see my 2 commits here: https://github.com/jasonish/suricata/pull/174/commits

Oh, I pushed a fix like the first commit, but I do not see what the second is doing...

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113274 | 145162 | 128.15% | 

Pipeline 13447

jasonish:
> Oh, I pushed a fix like the first commit, but I do not see what the second is doing...

It fixed a naming issue that you have appeared to fixed since.

jasonish:
> > Oh, I pushed a fix like the first commit, but I do not see what the second is doing...
> 
> It fixed a naming issue that you have appeared to fixed since.

Or I was looking at a bad rev, anyways, not an issue anymore.

jasonish (approved):
Looks OK to me.

Not sure if the case to the array and looping is better, but does look a little easier to maintain.  The other cleanups are needed IMO.

victorjulien:
Merged in #8815, thanks!

PR #8773, opened 2023-04-27
radix: Don't leak memory when key can't be added.

jlucovsky:
Thanks to @glongo for the contributions included in the redmine ticket.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5748](https://redmine.openinfosecfoundation.org/issues/5748)

Describe changes:
- Check for duplicate keys before adding to radix tree.
-  Free key memory when key is a duplicate

```
SV_REPO=
SV_BRANCH= pr/1190
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/util-radix-tree.c
+++ b/src/util-radix-tree.c
@@ -295,16 +295,14 @@ static int SCRadixPrefixNetmaskCount(SCRadixPrefix *prefix)
  *
  * \param prefix      Pointer to the ip prefix that is being checked.
  * \param netmask     The netmask value for which we will have to return the user_data
- * \param exact_match Bool flag which indicates if we should check if the prefix
+ * \param exact_match flag which indicates if we should check if the prefix
  *                    holds proper netblock(< 32 for ipv4 and < 128 for ipv6) or not.
  *
  * \retval 1 On match.
  * \retval 0 On no match.
  */
-static int SCRadixPrefixContainNetmaskAndSetUserData(SCRadixPrefix *prefix,
-                                                     uint16_t netmask,
-                                                     int exact_match,
-                                                     void **user_data_result)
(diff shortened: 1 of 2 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113274 | 145133 | 128.13% | 

Pipeline 13446

src/util-radix-tree.c
```diff
@@ -1009,6 +1009,12 @@ SCRadixNode *SCRadixAddKeyIPV4String(const char *str, SCRadixTree *tree, void *u
         SCRadixValidateIPv4Key((uint8_t *)&ip, netmask);
 #endif
     }
+
+    if (SCRadixFindKeyIPV4Netblock((uint8_t *)&ip, tree, netmask, NULL) != NULL) {
```
victorjulien, line 1013: I don't understand why this is done here. Adding duplicates is supported, right? It returns an existing node. In for example the IP only code we can have multiple sigs include the same ip/netblock.

glongo: The idea here is to not add duplicates, that's why the check is done here.
But if we assume that having duplicate netblocks is valid, I'm afraid this approach is incorrect.

jlucovsky: > Adding duplicates is supported, right? It returns an existing node.

If the tree doesn't persist duplicates and returns an existing node, then the caller's `user` must still be handled properly.

If this is the case, then we must check if callers to `SCRadixAddKey*` can detect the case and free the `user` data, if appropriate.

jlucovsky:
Continued in #8835

PR #8621, opened 2023-03-23
Http2 header 5780 v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5780

Describe changes:
- http2: rename http2.header to http.request_header and http.response_header as it is confusing with `http.header`

Replaces #8615 with added commit to add the HTTP1 functionality

I have lots of questions...

Should this be one commit ?
Even if it is renaming + adding HTTP1 functionality

Did I get right the use of HttpHeaderBuffer ?

Is src/detect-http-header.c the right file do put it ?

I removed the validation callback as it is HTTP2-specific
Should we just remove the escaping being done in HTTP2, and not be able to match differently on a HTTP2 header name having comma in it like `x: y` versus a HTTP2 header having x as its name and y as its value ?

How do we merger with S-V having some test with rules with the keyword being deprecated ?

Do we keep the deprecated keyword in 6 ?

suricata-verify-pr: 1153
https://github.com/OISF/suricata-verify/pull/1153

PR #8775, opened 2023-04-27
Http2 header 5780 v4

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5780

Describe changes:
- http2: rename http2.header to http.request_header and http.response_header as it is confusing with `http.header`

Replaces #8621 with needed rebase (because `DetectBufferSetActiveList` wants now `de_ctx` as argument)

I have lots of questions...

Should this be one commit ?
Even if it is renaming + adding HTTP1 functionality

Did I get right the use of HttpHeaderBuffer ?

Is src/detect-http-header.c the right file do put it ?

I removed the validation callback as it is HTTP2-specific
Should we just remove the escaping being done in HTTP2, and not be able to match differently on a HTTP2 header name having comma in it like `x: y` versus a HTTP2 header having x as its name and y as its value ?

How do we merger with S-V having some test with rules with the keyword being deprecated ?

Do we keep the deprecated keyword in 6 ?

https://github.com/OISF/suricata-verify/pull/1153

```
SV_BRANCH=pr/1153
```


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113274 | 145114 | 128.11% | 

Pipeline 13443

jufajardini (commented):
Doc changes look good to me :)

src/detect-http-header.c
```diff
@@ -467,6 +467,327 @@ void DetectHttpHeaderRegister(void)
...
+
+    htp_tx_t *tx = (htp_tx_t *)cbdata->txv;
+    htp_table_t *headers;
+    if (flags & STREAM_TOSERVER) {
+        headers = tx->request_headers;
+    } else {
+        headers = tx->response_headers;
+    }
+    if (cbdata->local_id < htp_table_size(headers)) {
+        htp_header_t *h = htp_table_get_index(headers, cbdata->local_id, NULL);
```
victorjulien, line 613: `htp_table_get_index` can return `NULL` I think

doc/userguide/rules/http-keywords.rst
```diff
@@ -684,6 +684,41 @@ like ``distance``, ``offset``, ``within``, etc.
 The ``nocase`` keyword is not allowed anymore. Keep in mind that you need
 to specify a lowercase pattern.
 
+http.request_header
+-------------------
+
+Match on the name and value of a HTTP request header (HTTP1 or HTTP2).
+
+For HTTP2, name and value get concatenated by ": ", colon and space.
+Each colon in the name or the value should be escaped as a double colon "::" for detection
```
victorjulien, line 693: I don't understand this escaping logic. The normal way to escape `:` in `content` is `\:`

victorjulien (changes requested):

victorjulien:
> Is src/detect-http-header.c the right file do put it ?

Yeah seems good.

victorjulien:
> Did I get right the use of HttpHeaderBuffer ?

Looks good. Pattern is equal to http.start for example.

victorjulien:
> Should this be one commit ?
> Even if it is renaming + adding HTTP1 functionality

Looks fine as is.

src/detect-http-header.c
```diff
@@ -467,6 +467,327 @@ void DetectHttpHeaderRegister(void)
...
+                    mpm_ctx, &det_ctx->mtcu, &det_ctx->pmq, buffer->inspect, buffer->inspect_len);
+            PREFILTER_PROFILING_ADD_BYTES(det_ctx, buffer->inspect_len);
+        }
+
+        local_id++;
+    }
+}
+
+static int PrefilterMpmHttp1HeaderRegister(DetectEngineCtx *de_ctx, SigGroupHead *sgh,
+        MpmCtx *mpm_ctx, const DetectBufferMpmRegistery *mpm_reg, int list_id)
```
victorjulien, line 672: rebase hint: was renamed to `DetectBufferMpmRegistry`

catenacyber:
> Is this about the `::` escaping? I didn't immediately see how it works, but I don't really like it. Would like buffers to be as "raw" as possible. Does that help?

Yes, this is about `::` escaping.
Yes, this helps.

The reason for `::` escaping was to be able to detect HTTP2 header names having `:`

doc/userguide/rules/http-keywords.rst
catenacyber: But then, I would also need to escape backslash itself ;-)

src/detect-http-header.c
catenacyber: Well, it is not NULL-checked in detect-http-header-names.c

catenacyber:
> Old keyword test can go lt-version: 6, tests for new stuff min-version: 7

That is more complex, as this is about rules, not S-V checks...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8986

PR #8630, opened 2023-03-26
nfq: Ensure packet release function set 

jlucovsky:
Continuation of #8608 

This PR ensures that early error returns in `NFQCallBack` will have the packet's release function pointer set.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5916](https://redmine.openinfosecfoundation.org/issues/5916)

Describe changes:
- Set packet's `ReleasePacket` earlier for early error paths.

Updates:
- Moved setting of `ReleasePacket` to `NFQSetupPkt`

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #8781, opened 2023-04-28
rust: update cargo.lock: updates nom 5 to 5.1.3

jasonish:
Regenerate Cargo.lock to update dependencies. Most specifically the update to
Nom 5.1.3 to avoid breakage with future version of rustc.


victorjulien:
Merged in #8630, thanks!

PR #8783, opened 2023-04-29
Rust unused v3

catenacyber:
None, generic cleaning

Describe changes:

Found with

There is more to it :
- `AuthenticationKerb5` and like are never constructed for pgsql. Should they be @jufajardini ?

Follows #8771 with code review from @jasonish taken into account


suricata-qa:

Pipeline 13462

jufajardini:
I'll refresh my memory and notes to give an answer on the Kerb5/PGSQL question. o>

jufajardini:
@catenacyber according to Postgresql v13 docs, Kerb5 is no longer supported since  9.4 - I checked previous docs versions to find this info. (https://www.postgresql.org/docs/13/protocol-flow.html#id-1.10.5.7.3). I guess my question with regard to this is: should we be able to parse unsupported versions of messages? If yes, then the Kerb5 should stay - and then I must extend `pgsql_parse_authentication_message`. If not, then yep, to remove it.

catenacyber:
> according to Postgresql v13 docs, Kerb5 is no longer supported since 9.4 - I checked previous docs versions to find this info. (https://www.postgresql.org/docs/13/protocol-flow.html#id-1.10.5.7.3). I guess my question with regard to this is: should we be able to parse unsupported versions of messages? If yes, then the Kerb5 should stay - and then I must extend `pgsql_parse_authentication_message`. If not, then yep, to remove it.

@jufajardini I would say that this Kerb5 should be removed until implemented for real, and I am not sure this is important enough to do so...

jufajardini:
> > according to Postgresql v13 docs, Kerb5 is no longer supported since 9.4 - I checked previous docs versions to find this info. (https://www.postgresql.org/docs/13/protocol-flow.html#id-1.10.5.7.3). I guess my question with regard to this is: should we be able to parse unsupported versions of messages? If yes, then the Kerb5 should stay - and then I must extend `pgsql_parse_authentication_message`. If not, then yep, to remove it.
> 
> @jufajardini I would say that this Kerb5 should be removed until implemented for real, and I am not sure this is important enough to do so...

Ok, I'll prepare a PR to remove them for now.

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8880

PR #8797, opened 2023-05-02
log/file: Reduce log file flush frequency, if configured.

jlucovsky:
This PR provides a configurable way to reduce the number of times an output file is flushed.

By default, output files will continue to be flushed after each write. When a flush-threshold is configured, files will be flushed every `flush-threshold` bytes; the byte count since the last write is tracked and when it reaches or exceeds the threshold, the file will be flushed.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3449](https://redmine.openinfosecfoundation.org/issues/3449)

Describe changes:
- Refactor write/clearerror logic to remove duplication
- Add `flush-threshold` config setting to control `fflush` frequency
- When a threshold exists, track flush count and display (as perf) when output file is closed.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -30,6 +30,12 @@ Output types::
       # Enable for multi-threaded eve.json output; output files are amended
       # with an identifier, e.g., eve.9.json. Default: off
       #threaded: off
+      # Control for calls to flush file contents:
+      # - 0 Flush after each write
+      # - <byte-count> Flush periodically. Track bytes written; when bytes
+      #   written is equal to or greater, flush file contents. The next
+      #   flush occurs when the bytes written meets or exceeds this value.
+      #flush-threshold: 0
       #prefix: "@cee: " # prefix to prepend to each log entry
       # the following are valid when type: syslog above
       #identity: "suricata"
@@ -279,7 +285,7 @@ The example above adds epoch time to the filename. All the date modifiers from t
 C library should be supported. See the man page for ``strftime`` for all supported
 modifiers.
 
-.. _output_eve_rotate:
+.. _output_eve_threaded:
 
 Threaded file output
 ~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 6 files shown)


src/util-logopenfile.c
```diff
@@ -517,6 +506,15 @@ SCConfLogOpenGeneric(ConfNode *conf,
     if (filetype == NULL)
         filetype = DEFAULT_LOG_FILETYPE;
 
+#define DEFAULT_LOG_FLUSH_THRESHOLD 0
```
jasonish, line 509: Not used.

doc/userguide/output/eve/eve-json-output.rst
```diff
@@ -301,6 +307,32 @@ This example will cause each Suricata thread to write to its own "eve.json" file
...
+up to ``flush-threshold`` bytes unbuffered/unwritten should Suricata unexpectedly terminate.
+
+The default value is ``0``.
+
+::
+
+   outputs:
+     - eve-log:
+         filename: eve.json
+         flush-threshold: 1024
```
jasonish, line 329: As no alternatives to the default are provided, users may look at this example as a suggestion, however, given average log sizes a value of 1024 might not help much - might reduce the flush to every 2nd log record. Would 4k or 8k be more reasonable?

victorjulien:
I guess what I don't like about such approaches if that if traffic suddenly stops (common in testing for example), the last logs may be held indefinitely. A time based approach might make more sense, but is probably more complex.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .app_layer.error.snmp.parser | 1 | 0 | - | 

Pipeline 13517

doc/userguide/output/eve/eve-json-output.rst
jlucovsky: I'll update the example.

However, the timing issue (lingering data) is a problem that I have to solve if this PR is to move forward.

src/util-logopenfile.c
jlucovsky: Will remove.

victorjulien:
I've been thinking about time issues before, and been wondering if it would make sense to create "heartbeat" like pseudo packets that are flushed through on an interval of say 1 second. Then we'd be able to do housekeeping tasks, like flushing lingering log data.

jlucovsky:
> I guess what I don't like about such approaches if that if traffic suddenly stops (common in testing for example), the last logs may be held indefinitely. A time based approach might make more sense, but is probably more complex.

The bytes are always written; this PR only reduces the calls to `fflush()`.

src/util-logopenfile.c
```diff
@@ -240,43 +252,15 @@ static int SCLogFileWriteNoLock(const char *buffer, int buffer_len, LogFileCtx *
...
-            SCConfLogReopen(log_ctx);
-        }
-
-        if (log_ctx->flags & LOGFILE_ROTATE_INTERVAL) {
-            time_t now = time(NULL);
-            if (now >= log_ctx->rotate_time) {
-                SCConfLogReopen(log_ctx);
-                log_ctx->rotate_time = now + log_ctx->rotate_interval;
-            }
-        }
```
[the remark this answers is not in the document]
jlucovsky: It does not; these steps are performed in `SCLogFileWriteNoLock`.

jlucovsky:
I plan to revisit this after the 7.0 release.

The main issue is how to handle periods of inactivity and at shutdown and ensure that data that's been written is actually flushed.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .app_layer.error.snmp.parser | 1 | 0 | - | 

Pipeline 13517

jlucovsky:
Continued in #9587

PR #8808, opened 2023-05-04
Smtp condition check/v3

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/6019

Previous PR: https://github.com/OISF/suricata/pull/8807

Changes since v2:
- Update commit message to reflect the issue

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -1139,6 +1139,15 @@ static int SMTPProcessRequest(SMTPState *state, Flow *f, AppLayerParserState *ps
     SCEnter();
     SMTPTransaction *tx = state->curr_tx;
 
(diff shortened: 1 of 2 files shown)


src/app-layer-smtp.c
```diff
@@ -1139,6 +1139,15 @@ static int SMTPProcessRequest(SMTPState *state, Flow *f, AppLayerParserState *ps
     SCEnter();
     SMTPTransaction *tx = state->curr_tx;
 
+    state->toserver_data_count += (line->len + line->delim_len);
```
inashivb, line 1142: Since this gets updated with line->len and delim_len, it was probably ok at its previous place as when len and delim_len are 0, nothing is updated here anyway.

PR #8812, opened 2023-05-04
Fixups/v3

victorjulien:
Attempt to implement more fine grained error handling in streaming buffer, so the caller than know what went wrong.

Uses a mix of `sc_errno` style and direct returns where possible. Not sure I like it yet.


suricata-qa:

Pipeline 13585

src/util-streaming-buffer.c
```diff
@@ -1204,9 +1224,12 @@ static StreamingBufferRegion *BufferInsertAtRegionConsolidate(StreamingBuffer *s
     return dst;
 }
 
+/** \note sets sc_errno */
 static StreamingBufferRegion *BufferInsertAtRegionDo(StreamingBuffer *sb,
         const StreamingBufferConfig *cfg, const uint64_t offset, const uint32_t len)
 {
+    sc_errno = SC_OK;
```
jasonish, line 1231: Just a nit depending on how much like `errno` we want to be...  Functions as part of a library never (re)set errno back to 0, that is left up to the caller. Just thinking out loud with respect to a library where we should have documentation on the error handling strategy.

src/util-streaming-buffer.c
```diff
@@ -127,30 +135,35 @@ static inline int InitBuffer(StreamingBuffer *sb, const StreamingBufferConfig *c
 {
     sb->region.buf = CALLOC(cfg, 1, cfg->buf_size);
     if (sb->region.buf == NULL) {
-        return -1;
+        return SC_ENOMEM;
```
jasonish, line 138: Could `CALLOC` set `sc_errno`?  I think this is what you meant by using a mix of direct returns and sc_errno?  Here I think it would make sense to have success/fail results, then `sc_errno` could be consulted.

jasonish:
> Uses a mix of sc_errno style and direct returns where possible. Not sure I like it yet.

Yeah, I find this hard to reason about. Even if direct return values are more *ergonomic* when they can be used, they can't be used everywhere.

suricata-qa:

Pipeline 14164

src/util-streaming-buffer.c
victorjulien: I've seen examples where it was set to OK/SUCCESS by the lib call, which to me seems like a more dev friendly pattern. But we can go either way, I guess. Just don't really like having to reset it all over the call sites.

Btw, we're not really resetting it before making syscalls. I guess in an ideal case it won't ever have to be reset, as we'd only check it in case of error, and in case of error it should have been set.

src/util-streaming-buffer.c
victorjulien: Moved setting to the allocators.

victorjulien:
Merged with additional updates in #9021

PR #8820, opened 2023-05-05
smtp: return on line completion

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/6020

### DIFF
--- a/src/app-layer-smtp.c
+++ b/src/app-layer-smtp.c
@@ -1051,6 +1051,13 @@ static int SMTPProcessRequest(SMTPState *state, Flow *f,
     SCEnter();
     SMTPTransaction *tx = state->curr_tx;
 
+    /* If current input is to be discarded because it completes a long line,
+     * line's length and delimeter len are reset to 0. Skip processing this line.
+     * This line is only to get us out of the state where we should discard any
+     * data till LF. */
+    if (line->len == 0 && line->delim_len == 0) {
+        return 0;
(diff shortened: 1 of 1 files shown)


src/app-layer-smtp.c
```diff
@@ -1051,6 +1051,13 @@ static int SMTPProcessRequest(SMTPState *state, Flow *f,
     SCEnter();
     SMTPTransaction *tx = state->curr_tx;
 
+    /* If current input is to be discarded because it completes a long line,
+     * line's length and delimeter len are reset to 0. Skip processing this line.
+     * This line is only to get us out of the state where we should discard any
+     * data till LF. */
+    if (line->len == 0 && line->delim_len == 0) {
+        return 0;
+    }
```
victorjulien, line 1060: in master we also moved `toserver_data_count`, should we do that here as well?

inashivb: We didn't. Because `toserver_data_count` seems to be incremented by `line->len + line->delim_len` so, if they're both 0, `toserver_data_count` is never really updated. And if either of the len or delim_len is greater than 0 then we bypass this if block and update the `toserver_data_count` anyway later on.

That being said, something did mess up somewhere during the push because `line` does not exist in `6.0.x` so this needs a new PR. 🤔

PR #7708, opened 2022-08-09
Spelling

jsoref:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [ ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:

This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/suricata/commit/ba63450813ff7b77c5d88f36be3fe32a067561fb#commitcomment-80730680

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/suricata/commit/0de1df9b732895b346229aeee0abed3960006916

> **Warning**
> this PR is incompatible with some of the workflows in this repository. Specifically, there's one that tries to build each commit of a PR. That takes about 2 minutes per commit. It could handle not more than 180 commits assuming it did no setup work. This PR has 300+ commits. GitHub historically has had a limit of 250 commits beyond which various pieces of its PR functionality has more or less broken.

> **Note**
> I don't expect this PR to be merged as-is. If possible, I'd encourage someone to review commit by commit and consider the words being changed. In general a specific word change is either acceptable as a unit or should be rejected as a unit. It is possible to split by directory or file extension or some other algorithm, but doing that risks things where a change spans the criteria.

> **Note**
> There's code formatter which doesn't actually seem happy with files in their current state on `master`, so they're effectively. a tripping hazard for anyone who might touch those files. At the risk of upsetting the formatter check, I've chosen not to apply its suggestions. But I encourage you to review its suggestions (which sadly are not provided in the workflow output) and consider how the project would want those files to be formatted. The easiest thing to do for this PR would be to drop the changes to those files.

> **Note**
> this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

I will try to annotate this PR, but it will take some time.

#suricata-verify-pr:
#suricata-verify-repo:
#suricata-verify-branch:
#suricata-update-pr:
#suricata-update-repo:
#suricata-update-branch:
#libhtp-pr:
#libhtp-repo:
#libhtp-branch:

PR #8822, opened 2023-05-05
spelling - rebase previous work - v2

jasonish:
Rebase of https://github.com/OISF/suricata/pull/7708.

- Probably needs a once over to make sure nothing wrong is here.

I'd recommend squashing this all into one super-commit, then re-look into
enabling the GitHub action and see what else it brings up.


jasonish:
Formatting check fails as it appears to be checking commit by commit and `clang-format rewrite-branch` does not appear to work, and with 300+ commits I've opted for a single `clang-format branch`.  Squashing and adding the format should allow it to pass.

victorjulien:
Also have a rebase, but with git commit cleanup for this. Will to a PR later.

PR #8802, opened 2023-05-03
Fix in the Napatech code to work with 7.0.0

tourko:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [x] I have read the contributing guide lines at https://suricata.readthedocs.io/en/latest/devguide/codebase/contributing/contribution-process.html
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:

Describe changes:
- Added 'module' argument to the SCLog() functions used by source-napatech.c
- Set nanoseconds in the packet timestamps using SCTIME_ADD_USECS macro
- Added return value check for ByteExtractStringUint8() calls

PR #8825, opened 2023-05-06
napatech: fix timestamp and flag

Kurori:
Describe changes:
- fix function call in source-napatech.c
- set the flag for thread in util-napatech.c

These changes have been tested on suricata-7.0.0-rc1 only.

Please note that the linker parameter has not been fixed.


victorjulien:
Thanks for your contribution. Sadly it's a duplicate effort, which conflicts with #8802. As the latter is done by Napatech themselves I will go with that one.

Side note: on future PRs please make sure to leave the checkboxes in place.

Thanks again.

PR #8835, opened 2023-05-07
radix: Don't leak memory when key can't be added

jlucovsky:
Thanks to @glongo for the contributions included in the redmine ticket.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [5748](https://redmine.openinfosecfoundation.org/issues/5748)

Describe changes:
- Check for duplicate keys before adding to radix tree.
-  Free key memory when key is a duplicate

Updates:
- Additional modifications to detect existing radix tree entries. When a duplicate is detected, the existing node will be returned along with a duplicate indicator so the caller can properly handle the `user` data.
- 

```
SV_REPO=
SV_BRANCH=pr/1190
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/util-radix-tree.c
+++ b/src/util-radix-tree.c
@@ -295,16 +295,14 @@ static int SCRadixPrefixNetmaskCount(SCRadixPrefix *prefix)
  *
  * \param prefix      Pointer to the ip prefix that is being checked.
  * \param netmask     The netmask value for which we will have to return the user_data
- * \param exact_match Bool flag which indicates if we should check if the prefix
+ * \param exact_match flag which indicates if we should check if the prefix
  *                    holds proper netblock(< 32 for ipv4 and < 128 for ipv6) or not.
  *
  * \retval 1 On match.
  * \retval 0 On no match.
  */
-static int SCRadixPrefixContainNetmaskAndSetUserData(SCRadixPrefix *prefix,
-                                                     uint16_t netmask,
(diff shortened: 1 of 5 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113274 | 145097 | 128.09% | 

Pipeline 13646

src/util-radix-tree.c
```diff
@@ -963,14 +961,16 @@ SCRadixNode *SCRadixAddKeyIPV6Netblock(uint8_t *key_stream, SCRadixTree *tree,
  *
  * \retval node Pointer to the newly created node
  */
-SCRadixNode *SCRadixAddKeyIPV4String(const char *str, SCRadixTree *tree, void *user)
+SCRadixNode *SCRadixAddKeyIPV4String(
```
victorjulien, line 964: none of the users of this function seem to use the return code for anything other than a NULL check. So if we're going through the hassle of changing the function prototype, we can just switch to a `int` return code? Or perhaps a `SCError`. For some examples of that see https://github.com/OISF/suricata/pull/8812/commits/f855ee99a397da67cc37d32db15101690dadb003

jlucovsky:
Continued in #9008

PR #8843, opened 2023-05-08
6.0.x: fix read the docs - v1

jasonish:
https://redmine.openinfosecfoundation.org/issues/6059


victorjulien:
If I read this correctly we could also use `build.os` to specify a different build image? https://github.com/readthedocs/readthedocs-docker-images/issues/194

https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os

jasonish:
> If I read this correctly we could also use `build.os` to specify a different build image? [readthedocs/readthedocs-docker-images#194](https://github.com/readthedocs/readthedocs-docker-images/issues/194)
> 
> https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os

What would we set it to?

jasonish:
Is that the favoured fix projects seem to be going with?

victorjulien:
> Is that the favoured fix projects seem to be going with?

Don't know, but the suggestion is by a RtD dev, so that seems to give it some weight. Also it seems more generic than locking to a old version.

jasonish:
See https://github.com/OISF/suricata/pull/8850.

PR #8844, opened 2023-05-08
github-ci: author check github action - v10

jasonish:
New draft for testing as I can't test this within the scope of my own repo (or the repo that originates a pull request).

### DIFF
--- a/.github/workflows/authors.yml
+++ b/.github/workflows/authors.yml
@@ -1,0 +1,38 @@
+name: New Author Check
+
+on:
+  - pull_request
+
+jobs:
+  check-id:
+    name: New Author Check
+    runs-on: ubuntu-latest
+    steps:
+      - run: sudo apt -y install git
+      - run: git clone https://github.com/${{ github.repository }}
+      - run: git remote add author ${{ github.event.pull_request.head.repo.html_url }}
+        working-directory: suricata
+      - run: git fetch author
+        working-directory: suricata
+      - run: git checkout author/${{ github.event.pull_request.head.ref }}
+        working-directory: suricata
+      - name: Export known authors from master branch
+        run: git log --format="%an <%ae>" origin/master | sort | uniq > ../authors.txt
+        working-directory: suricata
+      - name: Export authors from new commits
+        run: git log --format="%an <%ae>" origin/${GITHUB_BASE_REF}... | sort | uniq > ../commit-authors.txt
+        working-directory: suricata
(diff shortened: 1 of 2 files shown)


jasonish:
@victorjulien This is re-worked to use raw git commands instead of the Git checkout action. I think the only way to really test is to add it to some staging pull requests, where the incorrect results showed up before.

jasonish:
Ok, done futzing around with this version of the PR. Unfortunately, its not simple for a GitHub action in pull-request context to add a comment to its pull-request.

.github/workflows/authors.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: git log --format="%an <%ae>" origin/${GITHUB_BASE_REF}... | sort | uniq > ../commit-authors.txt
+        working-directory: suricata
+      - name: Check new authors
+        run: |
+          touch new-authors.txt
+          while read -r author; do
+             echo "Checking author: ${author}"
+             if ! grep -q "^${author}\$" authors.txt; then
+                 echo "ERROR: ${author} NOT FOUND"
+                 echo "::error ::New author found: ${author}"
```
victorjulien, line 32: what happens if you use `::warning` here and then not `exit 1`?

jasonish: Running 2 tests with exit 1 removed:
- https://github.com/OISF/suricata/pull/8848 (::error)
- https://github.com/OISF/suricata/pull/8849 (::warning)

I believe the exit is the only way to see any notification on the main pull request page, without going to the summary and looking for annotations.

victorjulien: yeah I see, thanks

jasonish:
Here's info on creating a comment from a pull request which might be the most useful: https://stackoverflow.com/questions/69499645/how-to-securely-allow-github-actions-to-check-pr-and-post-results-in-comment/71366152#71366152

Note that it requires getting some workflows into the master branch first, so it will require a couple of stages to get working likely. But can test most of it in my own fork, *I think*.

jasonish:
I guess the question is, what type of output is desired:
- A red X beside the job check for others..
- OR.. A green X, but a comment specifying that there are new authors?

victorjulien:
The latter, I think. I don't like it that  CI fails for just having new author.

jasonish:
Continued here: https://github.com/OISF/suricata/pull/8854

PR #8847, opened 2023-05-09
(draft) jsonbuilder: better handling of memory allocation errors - v1

jasonish:
Wrap (most) growth of the internal buffer in methods that first try to reserve
the required data and return a Result if unable to do so. The idea being we
return error on memory failures rather than panic/abort.

### DIFF
--- a/rust/src/jsonbuilder.rs
+++ b/rust/src/jsonbuilder.rs
@@ -1,4 +1,4 @@
-/* Copyright (C) 2020 Open Information Security Foundation
+/* Copyright (C) 2020-2023 Open Information Security Foundation
  *
  * You can copy, redistribute or modify this Program under the terms of
  * the GNU General Public License version 2 as published by the Free
@@ -17,6 +17,7 @@
 
 #![allow(clippy::missing_safety_doc)]
 
+use std::collections::TryReserveError;
 use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::str::Utf8Error;
@@ -27,6 +28,7 @@ const INIT_SIZE: usize = 4096;
 pub enum JsonError {
     InvalidState,
     Utf8Error(Utf8Error),
+    Memory,
 }
 
 impl std::error::Error for JsonError {}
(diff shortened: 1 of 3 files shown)


rust/src/jsonbuilder.rs
```diff
@@ -98,33 +107,55 @@ pub struct JsonBuilder {
...
+        }
+        self.buf.push(ch);
+        Ok(self)
+    }
+
+    /// A wrapper around String::push_str that pre-allocates data
+    /// return an error if unable to.
+    pub fn push_str(&mut self, s: &str) -> Result<&mut Self, JsonError> {
+        if self.buf.capacity() < self.buf.len() + s.len() {
+            self.buf.try_reserve(INIT_SIZE)?;
```
jasonish, line 155: I don't like the doubling that many allocators do, but this should at least check try_reserve the max of INIT_SIZE or s.len()...

catenacyber: There is no `try_push`, right ?

catenacyber:
> Still draft, but I'd appreciate your thoughts on this.

Looks cool.
How do you know if you are complete ?
And future-proof ? (that when I add a new `jb_push_sha1` or whatever, I do not add allocations without checks)

jasonish:
> And future-proof ? (that when I add a new `jb_push_sha1` or whatever, I do not add allocations without checks)

You don't really. This is very much an active effort to opt-in to.  The only thing I can think of is making the internal buffer harder to access, so you have to think twice about it. So open to ideas...  I guess its a bit like return values in C.. Nothing forces you to, but its a good idea :)

rust/src/jsonbuilder.rs
jasonish: > There is no `try_push`, right ?

No.  But `push`, `push_str` won't allocated unless they need to, which they won't if we `try_reserve` first.

jasonish:
A more complete version is now up here: https://github.com/OISF/suricata/pull/8855

PR #8850, opened 2023-05-09
6.0.x: Fix ReadTheDocs building - v2

jasonish:
Alternate fix to https://github.com/OISF/suricata/pull/8843 by using an Ubuntu
22.04 image with updated OpenSSL.


victorjulien:
RtD fixed the issue, so not merging this. We can revisit if we have to in the future.

PR #8854, opened 2023-05-09
github-ci: new author check - v13

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/8844

For a pull request, compare the authors of the commits to the known authors in
the master branch and upload these as an artifact.

Using a workflow_run, check if this file is non-zero and if so, add a comment
to the pull request.

This can-not be demonstrated in the scope of a pull request, as
".github/workflows/authors-done.yml" needs to exist in the repo's default
branch for security reasons.

This allow the author check to not fail, but get a comment about new authors.

Example: https://github.com/jasonish/suricata/pull/177#issuecomment-1540941268

### DIFF
--- a/.github/workflows/authors-done.yml
+++ b/.github/workflows/authors-done.yml
@@ -1,0 +1,52 @@
+name: New Authors Report
+
+on:
+  workflow_run:
+    workflows: [New Authors Check]
+    types: [completed]
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Author check is complete"
+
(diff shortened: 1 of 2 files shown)


jasonish:
Question: Should we put the new authors into the comment?

victorjulien:
> Question: Should we put the new authors into the comment?

I think that would be useful, yes. It's already public anyway, so I see no harm in doing so.

jasonish:
Update PR with authors in pull request message here: https://github.com/OISF/suricata/pull/8863

PR #8855, opened 2023-05-09
jsonbuilder: attempt to handle memory allocation errors - v2

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/8847

Attempt to handle all "normal" memory allocation errors in JsonBuilder where
the Rust APIs allow us. This primarily means growing the underlying String
buffer as data is added.

As the JsonBuilder already used a Result for most methods, most users of
JsonBuilder already handle errors.

Changes from last PR:
- I think this is complete now.

Things to investigate:
- Box::new failing. This might not be catchable. Box::try_new is an
  experimental API that will allow us to handle this better.

### DIFF
--- a/rust/src/jsonbuilder.rs
+++ b/rust/src/jsonbuilder.rs
@@ -17,6 +17,7 @@
 
 #![allow(clippy::missing_safety_doc)]
 
+use base64::Engine;
 use std::cmp::max;
 use std::collections::TryReserveError;
 use std::ffi::CStr;
@@ -704,7 +705,7 @@ impl JsonBuilder {
         if self.buf.capacity() < self.buf.len() + val.len() * 2 {
             self.buf.try_reserve(val.len() * 2)?;
         }
-        base64::encode_config_buf(val, base64::STANDARD, &mut self.buf);
+        base64::engine::general_purpose::STANDARD.encode_string(val, &mut self.buf);
         Ok(self)
     }
 }
--- a/rust/src/jsonbuilder.rs
+++ b/rust/src/jsonbuilder.rs
@@ -341,7 +341,7 @@ impl JsonBuilder {
         match self.current_state() {
             State::ArrayFirst => {
                 self.push('"')?;
-                base64::encode_config_buf(val, base64::STANDARD, &mut self.buf);
+                self.encode_base64(val)?;
                 self.push('"')?;
                 self.set_state(State::ArrayNth);
                 Ok(self)
@@ -349,7 +349,7 @@ impl JsonBuilder {
             State::ArrayNth => {
                 self.push(',')?;
                 self.push('"')?;
-                base64::encode_config_buf(val, base64::STANDARD, &mut self.buf);
+                self.encode_base64(val)?;
                 self.push('"')?;
                 Ok(self)
             }
@@ -530,7 +530,7 @@ impl JsonBuilder {
         self.push('"')?;
         self.push_str(key)?;
         self.push_str("\":\"")?;
-        base64::encode_config_buf(val, base64::STANDARD, &mut self.buf);
+        self.encode_base64(val)?;
         self.push('"')?;
 
         Ok(self)
@@ -639,12 +639,12 @@ impl JsonBuilder {
     ///    existing buffer.
     #[inline(always)]
     fn encode_string(&mut self, val: &str) -> Result<(), JsonError> {
-	let mut buf = Vec::new();
+        let mut buf = Vec::new();
 
-	// Start by allocating a reasonable size buffer, it will be
-	// grown if needed.
-	buf.try_reserve(val.len() * 2 + 2)?;
-	buf.resize(val.len() * 2 + 2, 0);
+        // Start by allocating a reasonable size buffer, it will be
+        // grown if needed.
+        buf.try_reserve(val.len() * 2 + 2)?;
+        buf.resize(val.len() * 2 + 2, 0);
 
         let mut offset = 0;
         let bytes = val.as_bytes();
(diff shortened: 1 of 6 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113274 | 144976 | 127.99% | 

Pipeline 13706

catenacyber:
Have you tried running jsonbuilder unit tests with an allocation failure framework ? (ie a LD_PRELOAD overload malloc to fail, running the tests X times, X being the number of allocations, and making each allocation fail)

jasonish:
> Have you tried running jsonbuilder unit tests with an allocation failure framework ? (ie a LD_PRELOAD overload malloc to fail, running the tests X times, X being the number of allocations, and making each allocation fail)

No. Any existing examples of this?

catenacyber:
Like https://github.com/tavianator/oomify but there may be others (and it is quite easy to recode)

catenacyber:
Jason, I had created this https://redmine.openinfosecfoundation.org/issues/5851 for your info

jasonish:
> Like https://github.com/tavianator/oomify but there may be others (and it is quite easy to recode)

Quick test with `oomify` does show the `reserve` calls failing in a catchable way..

```
[jason@t470 …/suricata/jsonbuilder-alloc/rust]  jsonbuilder-alloc/v3 ❯ oomify -n20 ./target/debug/test-jsonbuilder
size: 1
size: 2
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Memory', src/bin/test-jsonbuilder.rs:10:48
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
oomify: alloc 20: ./target/debug/test-jsonbuilder exited with status 101
```
The panic there is just a result on my unwrap, so that shows that if Rust isn't killed by the kernel, the error is rippled back to the application.

catenacyber:
Where can I see test-jsonbuilder.rs source ?

jasonish:
> Where can I see test-jsonbuilder.rs source ?

Its just in my working directory, not something suitable to check in, so no PR...

```
use suricata::jsonbuilder::JsonBuilder;

fn main() {
    let mut size = 1;

    for i in 1..8 {
        println!("size: {}", size);
        let mut jb = JsonBuilder::try_new_object().unwrap();
        let mut buf = vec![0x61; size];
        jb.set_string_from_bytes("test", &buf).unwrap();
        let boxed = Box::new(jb);
        size = size * 2;
    }
}
```

rust/src/jsonbuilder.rs
```diff
@@ -138,6 +139,26 @@ impl JsonBuilder {
...
+        }
+        self.buf.push(ch);
+        Ok(self)
+    }
+
+    /// A wrapper around String::push_str that pre-allocates data
+    /// return an error if unable to.
+    pub fn push_str(&mut self, s: &str) -> Result<&mut Self, JsonError> {
+        if self.buf.capacity() < self.buf.len() + s.len() {
+            self.buf.try_reserve(max(INIT_SIZE, s.len()))?;
```
catenacyber, line 157: Should it not be `self.buf.len() + s.len()` ?

rust/src/jsonbuilder.rs
```diff
@@ -698,6 +698,15 @@ impl JsonBuilder {
         }
         Ok(())
     }
+
+    fn encode_base64(&mut self, val: &[u8]) -> Result<&mut Self, JsonError> {
+        // Before base64 encoding, make sure enough space is reserved.
+        if self.buf.capacity() < self.buf.len() + val.len() * 2 {
+            self.buf.try_reserve(val.len() * 2)?;
```
catenacyber, line 706: base64 is rather ((val.len()+2)/3)*4 no ?

catenacyber: like one byte gets encoded as 4 bytes

catenacyber:
Do not you need to check also `vec![State::None, State::ObjectFirst]` ?

catenacyber:
And make `push_state` fallible

catenacyber (changes requested):
Some nits to address ;-)

I wonder if some cases should do a big reserve before multiple push_str like `open_array` doing
```
        self.push('"')?;
        self.push_str(key)?;
        self.push_str("\":[")?;
```

rust/src/jsonbuilder.rs
```diff
@@ -635,18 +635,26 @@ impl JsonBuilder {
     /// The string is encoded into an intermediate vector as its faster
     /// than building onto the buffer.
     ///
-    /// TODO: Allocate memory in a fallible way.
+    /// TODO: Revisit this, would be nice to build directly onto the
```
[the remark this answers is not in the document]
jasonish: In a way, that one todo had to be removed, and another way came to mind at the same time.

jasonish:
> Some nits to address ;-)
> 
> I wonder if some cases should do a big reserve before multiple push_str like `open_array` doing
> 
> ```
>         self.push('"')?;
>         self.push_str(key)?;
>         self.push_str("\":[")?;
> ```

Maybe.   But as the reserve behind these does 4k size, the first that needs it will allocate enough. My first iteration did try to estimate the size and reserve at once, but it did seem a little error prone.  But not harm to estimate it, reserve once so the reserve shouldn't be triggered again during the call into jsonbuilder. Not sure if it'll have a noticeable affect other than fail a little earlier if it was going to fail.

rust/src/jsonbuilder.rs
jasonish: Ah yeah, mines too small.  `base64` has a function for this, will just use that.

jasonish:
Commented addressed (well most of them) here: https://github.com/OISF/suricata/pull/8901

PR #8863, opened 2023-05-10
github-ci: new author check - v14

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/8854

Changes from previous PR:
- Display the new authors in a pull request comment

Example: https://github.com/jasonish/suricata/pull/177#issuecomment-1542662114

### DIFF
--- a/.github/workflows/authors-done.yml
+++ b/.github/workflows/authors-done.yml
@@ -1,0 +1,54 @@
+name: New Authors Report
+
+on:
+  workflow_run:
+    workflows: [New Authors Check]
+    types: [completed]
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


inashivb:
I totally forgot why we were doing this.. 🤔  GitHub has the `First Time Contributor` tag for the first timers.. so..what does this help us with?

jasonish:
> I totally forgot why we were doing this.. thinking GitHub has the `First Time Contributor` tag for the first timers.. so..what does this help us with?

Victor and I did talk about this the other day.  I think the main issue is that the PR can contain commits by other authors as well, which this GitHub feature doesn't pick up as far as I know, so this will pick up new authors in a PR that contains commits by a mix of authors, which is rare, but I believe has happened.

victorjulien:
Merged in #8879, thanks!

PR #8864, opened 2023-05-11
Output alert applayer v1

catenacyber:
Continuation of https://github.com/OISF/suricata/pull/8772

Describe changes:
- Fix setup-app-layer script so that it adds app-layer metadata to alerts
- Adds ftp metadata to alerts
- Adds tftp metadata to alerts

```
SV_BRANCH=pr/1196
```

### DIFF
--- a/rust/src/modbus/log.rs
+++ b/rust/src/modbus/log.rs
@@ -27,7 +27,6 @@ pub extern "C" fn rs_modbus_to_json(tx: &mut ModbusTransaction, js: &mut JsonBui
 
 /// populate a json object with transactional information, for logging
 fn log(tx: &ModbusTransaction, js: &mut JsonBuilder) -> Result<(), JsonError> {
-    js.open_object("modbus")?;
     js.set_uint("id", tx.id)?;
 
     if let Some(req) = &tx.request {
@@ -42,7 +41,6 @@ fn log(tx: &ModbusTransaction, js: &mut JsonBuilder) -> Result<(), JsonError> {
         js.close()?;
     }
 
-    js.close()?;
     Ok(())
 }
 
--- a/rust/src/quic/logger.rs
+++ b/rust/src/quic/logger.rs
@@ -89,7 +89,6 @@ fn quic_tls_extension_name(e: u16) -> Option<String> {
 }
 
 fn log_template(tx: &QuicTransaction, js: &mut JsonBuilder) -> Result<(), JsonError> {
-    js.open_object("quic")?;
     if tx.header.ty != QuicType::Short {
         js.set_string("version", String::from(tx.header.version).as_str())?;
 
@@ -144,7 +143,6 @@ fn log_template(tx: &QuicTransaction, js: &mut JsonBuilder) -> Result<(), JsonEr
         js.close()?;
     }
 
-    js.close()?;
     Ok(())
 }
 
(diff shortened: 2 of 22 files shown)


catenacyber:
Questions about this draft :
- Should we unify logging functions so that they use state ? and configuration flags ?
Looks like some protocols could get rid of state in their logging function : snmp, rfb, krb5
- What about cases where we want to log multiple objects ? (such as `smtp` and `email`)

TODOs : cf output.c

There are more protocols which currently miss this (beyond template, ftp and tftp being fixed here) : pgsql, dcerpc, dhcp, krb5

suricata-qa:
ERROR:
 

Pipeline 13735

catenacyber:
> There are more protocols which currently miss this (beyond template, ftp and tftp being fixed here) : pgsql, dcerpc, dhcp, krb5

I guess tickets should be created for these (including ftp and tftp)

suricata-qa:
ERROR:
 

Pipeline 13737

catenacyber:
@jasonish what do you think about the Check Rust failure ?

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .tcp.rst | 113274 | 144861 | 127.89% | 

Pipeline 13738

catenacyber:
Also, one next step is to have `JsonGenericLogger` to replace all `JsonBitTorrentDHTLogger` (and other), and get rid of `output-json-bittorrent-dht.c`

Only difficulty there is to know if we have TCP or UDP or both for OutputGenericLogInitSub

jasonish:
> @jasonish what do you think about the Check Rust failure ?

Looks like GitHub changed the format the code is checked out, in such a way we can no longer use git tools like `git diff`.

catenacyber:
> Looks like GitHub changed the format the code is checked out, in such a way we can no longer use git tools like `git diff`.

Ok, but I also do not get anything locally running `cargo clippy --all-features --fix --allow-no-vcs`

And this PR is only removing some lines of rust code...

jasonish:
Fix: https://github.com/OISF/suricata/pull/8866

jasonish:
> * What about cases where we want to log multiple objects ? (such as `smtp` and `email`)

By this do you mean that there is not a 1:1 relation between protocol and name and objects logged?

catenacyber:
> By this do you mean that there is not a 1:1 relation between protocol and name and objects logged?

Yes there is.

I mean to ask : should these cases be handled in the generic fashion proposed in this PR ? Or just have some separate logic..

jasonish:
> I mean to ask : should these cases be handled in the generic fashion proposed in this PR ? Or just have some separate logic..

I think some separate logic may make the initial versions of this PR easier to reason about?  We're shooting for post 7.0 stuff here right?

catenacyber:
> We're shooting for post 7.0 stuff here right?

Not necessarily : this is code simplification, a fix for the script to setup an app layer, and it is leading to adding missing features (adding ftp metadata to alerts)
Do not you think it can be merged already ? (after some cleanups in this PR)

catenacyber:
Need to rebase after https://github.com/OISF/suricata/pull/8867 to get CI green

catenacyber:
Replaced by #8872

jasonish:
> Do not you think it can be merged already ? (after some cleanups in this PR)

Perhaps, I wasn't looking at it in that frame of mind as the issue is targetted for 8.0.

PR #8877, opened 2023-05-12
dpdk: support capture timeout support

victorjulien:
To not inject too many packets, simulate a 100ms timeout based on af-packet's 100ms poll() timeout.

Bug: #6064.

replaces #8874, limiting the number of timeout packets


suricata-qa:

Pipeline 13766

PR #8880, opened 2023-05-14
Rust unused v4

catenacyber:
None, generic cleaning

Describe changes:

Found with

There is more to it :

Follows #8783 with needed rebase + removing some more pgsql lines of unused code (cc @jufajardini )


suricata-qa:

Pipeline 13785

rust/src/core.rs
```diff
@@ -24,12 +24,14 @@ use crate::debug_validate_fail;
 /// Opaque C types.
 pub enum DetectEngineState {}
 pub enum AppLayerDecoderEvents {}
-pub enum AppLayerParserState {}
 
-// From app-layer-events.h
-pub type AppLayerEventType = std::os::raw::c_int;
-pub const APP_LAYER_EVENT_TYPE_TRANSACTION : i32 = 1;
-pub const APP_LAYER_EVENT_TYPE_PACKET      : i32 = 2;
+#[repr(C)]
```
jasonish, line 28: If this is used by C, should we use `#[repr(i32)]`?

jasonish: Just want to make sure there are no mismatched types anywhere.

catenacyber: Currently, there is a mismatch (int in C, i32 in rust) , this PR fixes it

jufajardini (commented):
Checking this today for the pgsql parts, sorry for the delay.

rust/src/frames.rs
```diff
@@ -100,13 +99,6 @@ impl Frame {
         };
     }
 
-    #[allow(clippy::not_unsafe_ptr_arg_deref)]
-    pub fn set_tx(&self, flow: *const Flow, tx_id: u64) {
-        unsafe {
-            AppLayerFrameSetTxIdById(flow, self.direction(), self.id, tx_id);
-        };
-    }
-
```
jufajardini, line 109: Shouldn't we leave this, as we're planning to expand the usage and coverage of frames for other protocols?

catenacyber: There are argument both ways.

Pro-remove: Unused code is untested and is or will get broken soon

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/8976

rust/src/pgsql/pgsql.rs
```diff
@@ -119,6 +118,7 @@ pub enum PgsqlStateProgress {
     CommandCompletedReceived,
     ErrorMessageReceived,
     ConnectionTerminated,
+    #[cfg(test)]
```
jufajardini, line 121: Not sure I understand this here...

jufajardini (commented):
Overall, pgsql changes make sense, to me. Left one inline comment for something I didn't quite understand, and a comment unrelated to pgsql where I'm also unsure of the results. :P

rust/src/pgsql/pgsql.rs
catenacyber: `#cfg` in rust applies only to the next one, so just `UnknownState`

rust/src/frames.rs
jufajardini: Ok, the only thing I can think of for keeping it would be that if we don't have that, it is less likely that people adding frames will think of using something like that?
Would using `#[allow(dead_code)]` be acceptable? I'm guessing not, right.

catenacyber: Making these `public` makes them not dead_code, even if unused.
We should make most `pub(crate)` or even `pub(self)`

PR #8883, opened 2023-05-14
detect: Fix FTP bounce detection on little-endian platforms

coledishington:
The argument ip_orig of DetectFtpbounceMatchArgs() is in network byte order but is compared to the ip parsed from the FTP payload that was converted to host byte order.

Fix the FTP bounce ip address comparison by converting ip_orig to host byte order.

- [✓] I have read the contributing guide lines at https://suricata.readthedocs.io/en/latest/devguide/codebase/contributing/contribution-process.html

No redmine ticket due to simplicity of fix.

Describe changes:
Fix FTP bounce IP address comparison between FTP payload IP address, in host byte-order, and original IP address, in network byte order.

### DIFF
--- a/src/detect-ftpbounce.c
+++ b/src/detect-ftpbounce.c
@@ -141,7 +141,7 @@ static int DetectFtpbounceMatchArgs(
             }
             if (noctet == 4) {
                 /* Different IP than src, ftp bounce scan */
-                ip = SCByteSwap32(ip);
+                ip = htonl(ip);
 
                 if (ip != ip_orig) {
                     SCLogDebug("Different ip, so Matched ip:%d <-> ip_orig:%d",


catenacyber:
Interesting, would you have a suricata-verify test for this ?

src/detect-ftpbounce.c
```diff
@@ -117,6 +117,7 @@ static int DetectFtpbounceMatchArgs(
     /* Search for the first IP octect(Skips "PORT ") */
     while (i < payload_len && !isdigit((unsigned char)c[i])) i++;
 
+    ip_orig = SCByteSwap32(ip_orig);
```
victorjulien, line 120: this unconditionally swaps the bytes, won't this break things on little endian? ntohl might be more appropriate

victorjulien: `SCNtohl` actually

coledishington: I changed the ip address from host to network order but it didn't look like there was a Suricata specific htonl.

victorjulien:
> > Interesting, would you have a suricata-verify test for this ?
> 
> Does suricata-verify run on multiple architectures? This error does not occur for x86_64 since the bug is architecture specific.

It's run regularly on x86, x86_64, arm7 and arm8, so all little endian. Adhoc we run against a big endian VM. The code in question has no test coverage at all currently. If it had the byteswap issue should have been caught in the initial version of this PR.

src/detect-ftpbounce.c
catenacyber: @coledishington could you use `SCNtohl` ?

catenacyber (changes requested):
Thanks for this find, looking forward to having a test

coledishington:
> Thanks for this find, looking forward to having a test

Submitted a test (https://github.com/OISF/suricata-verify/pull/1212) for FTP bounce bug fixed in this patch. I tested before and after the patch on x86_64 and mips64.

On x86_64, the test passed before and after the patch as the platform is little-endian,and  thus this patch makes no difference for this target.

On mips64, the test failed before the patch and succeeded afterwards.

Thanks

src/detect-ftpbounce.c
coledishington: @catenacyber Yep, the functions htonl() to ntohl()/SCNtohl() can be used interchangeably although for the sake of readability I choose to use htonl() rather than ntohl()/SCNtohl(), as the parsed FTP IP address is being translated from host-order to network-order.

I can use SCNtohl() in the patch if Suricata prefers SCNtohl() for all byte-order conversions, rather than just network-order to host-order.

Thanks

coledishington:
> Thanks @coledishington would you submit a new version of the PR using this suricata-verify test and using `SCNtohl` ? (see the template proposed)

Submitted the new pull request over https://github.com/OISF/suricata/pull/8998.

PR #8885, opened 2023-05-15
flowworker: simplify pseudo packet use

victorjulien:
Pseudo packets originating in the flow worker do not need to leave the flow worker. Putting those in the ThreadVars::decode_pq will make them be evaluated by the next steps in the pipeline, but those will all ignore pseudo packets.

Instead, this patch returns them to the packet pool, while still honoring the IPS verdict logic.


suricata-qa:

Pipeline 13793

victorjulien:
Merged in #8888

PR #8845, opened 2023-05-09
smb: fix SMB_COM_WRITE_ANDX record parser - v2

b1tg:
Bug:  https://redmine.openinfosecfoundation.org/issues/6008

Previous PR: https://github.com/OISF/suricata/pull/8839

Changes from last PR:
- remove TODO & add comments

SV_BRANCH=pr/1195

PR #6008, opened 2021-03-29
proflite: new light weigth packet profiling

victorjulien:
Rebased #5728

Can be enabled/disabled over unix-socket.

There are several tracepoints that can be selected:

    suricatasc -c "profile-tracepoint-entry packetpool_get"
    suricatasc -c "profile-tracepoint-exit packetpool_return"

The tracepoints are:

    packetpool_get, packetpool_return,
    flowworker_enter, flowworker_exit,
    flowworker_pre_inject (as flowworker_exit but excludes flow
                           house keeping tasks)
    flowworker_applayer_enter, flowworker_applayer_end,
    flowworker_detect_enter, flowworker_detect_end,
    flowworker_output_start, flowworker_output_end,

By selecting packetpool_get as entry and packetpool_return as exit the
entire lifetime of a packet is accounted.

Features can be enabled by:

    suricatasc -c "profile-feature-enable any"
    suricatasc -c "profile-feature-enable only_http"

Features:

    all - special counters for all packets
    tcp - tcp packets with break out per SYN/FIN/RST flags
    app_http / only_app_http
    app_dns / only_app_dns
    app_ftp / only_app_ftp
    app_dcerpc / only_app_dcerpc
    any - enable all counters

PR #8889, opened 2023-05-16
smb: fix SMB_COM_WRITE_ANDX record parser - v3

b1tg:
Bug: #6008

ticket:
https://redmine.openinfosecfoundation.org/issues/6008

Previous PR: https://github.com/OISF/suricata/pull/8845

Changes from last PR:
- use smaller blob for test
- do not parse one byte padding which is possible missing
- update SV

SV_BRANCH=pr/1203

### DIFF
--- a/rust/src/smb/smb1_records.rs
+++ b/rust/src/smb/smb1_records.rs
@@ -21,6 +21,8 @@ use crate::smb::smb::*;
 use crate::smb::smb_records::*;
 use nom7::bytes::streaming::{tag, take};
 use nom7::combinator::{complete, cond, peek, rest, verify};
+use nom7::error::{make_error, ErrorKind};
+use nom7::Err;
 use nom7::multi::many1;
 use nom7::number::streaming::{le_u8, le_u16, le_u32, le_u64};
 use nom7::IResult;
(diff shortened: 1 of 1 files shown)


b1tg:
>  Or do you have another test case that shows another problem ?

@catenacyber 

In the origin packet, data_length == bcc == 20, if we use a proxy to change data_length to 17, Windows still accept it and write 17 bytes to file, but the original `parse_smb1_write_andx_request_record` will take 3 bytes padding to make record.data inconsistent with it in Windows.

```rust
#[test]
fn test_parse_smb1_write_andx_request_record_padding() {
    let data = hex::decode("0eff000000054000000000ff00000008001400000011003f000000000014004142434445464748494a4b4c4d4e4f5051520a0a").unwrap();
    let result = parse_smb1_write_andx_request_record(&data, SMB1_HEADER_SIZE);
	assert!(result.is_ok());
    let record = result.unwrap().1;
    assert_eq!(record.offset, 0);
    assert_eq!(record.len, 17);
    assert_eq!(record.data.len(), 17);
    // assert_eq!(record.data, b"DEFGHIJKLMNOPQR\n\n"); // origin `parse_smb1_write_andx_request_record` output this
    assert_eq!(record.data, b"ABCDEFGHIJKLMNOPQ"); // Windows server write this 17 bytes to file
}

```

pcap added to https://redmine.openinfosecfoundation.org/issues/6008

victorjulien:
Is that pcap also part of a SV test? If not, could you add it as well? Thanks!

catenacyber:
@b1tg running smb_bug_padding.pcap through suricata with master,
then `jq 'select(.event_type=="fileinfo") | .fileinfo.size' log/eve.json  

I see both 20 and 17
That looks like the expected result, so no need to patch, right ?

I think there is some post-processing after unit parsing to limit the file data length...
And the root cause of this complexity is that we do not want to wait for the full NBSS/SMB write record before streaming the data into the file... For reference, you can look at https://redmine.openinfosecfoundation.org/issues/5786 and https://redmine.openinfosecfoundation.org/issues/5770

b1tg:
Yes, smb_bug_padding.pcap include two tcp stream, i use a proxy to change data_length as 17 in the second stream, the issue here is not the fileinfo.size, but the file data which is not correct parsed.

catenacyber:
Ok, so Suricata and Wireshark show "DEFGHIJKLMNOPQR\n\n" but Windows gets "ABCDEFGHIJKLMNOPQ"...
Do I get it right ?

If so, there are 2 bugs : the high_offset one and this one. I guess each bug deserves its own commit/ticket/test...

b1tg:
Yes, and there is another case that may need to be considered, If proxy change data_length to 0x3c, Windows write follow data to file, i guess this can also be a possible evasion
```
00 14 00 41 42 43 44 45 46 47 48 49 4A 4B 4C 4D
4E 4F 50 51
```

catenacyber:
> If proxy change data_length to 0x3c

Not sure I understand this one.

The data gets bounded by the NBSS record, right ?

b1tg:
I did some reverse engineer on srv.sys , the follow code might be helpful to understand.

```cpp
__int64 __fastcall SrvSmbWriteAndX(__int64 this, __int64 a2, __int64 a3, __int64 a4)
  // [...]
  data_offset = *(unsigned __int16 *)(smb_record_data + 0x17);
  data_offset1 = data_offset;
  unknow = *(_QWORD *)(this + 184);
  smb_record_end_offset = *(_DWORD *)(unknow + 0x20);
  // Windows accept 0x3C as data_offset
  if ( smb_record_end_offset < data_offset || data_offset < 0x3C )// check data_offset
  {
    v10 = 2645;
    goto LABEL_189;
  }
  v17 = smb_record_end_offset - data_offset;
  data_length_real = *(unsigned __int16 *)(smb_record_data + 0x15);
  if ( data_length_real >= v17 )
    data_length_real = v17;
  data_ptr = *(_QWORD *)(this + 208) + data_offset;// pointer to data

```

catenacyber (changes requested):
Thanks for these patches.
Would you split the commit in 2, one for each bug fix ? (the offset one, and the padding one)

b1tg:
> Is that pcap also part of a SV test? If not, could you add it as well? Thanks!

Added in https://github.com/OISF/suricata-verify/pull/1211

b1tg:
use https://github.com/OISF/suricata/pull/8905

PR #8911, opened 2023-05-23
configure: Require c11 as the minimum C standard

jlucovsky:
This commit modifies the configure script to check for C11 by comparing the __STDC_VERSION__ with 201112L

An error message is issued if the C version is older than the C11 standard version.

Issue: 6029

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6029](https://redmine.openinfosecfoundation.org/issues/6029)

Describe changes:
- Updated `configure` to check `__STDC_VERSION__` against the C11 value.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -129,6 +129,14 @@
     AM_CONDITIONAL([HAVE_FETCH_COMMAND], [test "x$HAVE_WGET" != "xno" || test "x$HAVE_CURL" != "xno"])
     AM_CONDITIONAL([HAVE_WGET_COMMAND], [test "x$HAVE_WGET" != "xno"])
 
+    # We require c11+
+    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#if __STDC_VERSION__ >= 201112L
+                                         int c11_supported;
+                                         #endif
+                                       ]])],
+                                       [],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -129,6 +129,14 @@
     AM_CONDITIONAL([HAVE_FETCH_COMMAND], [test "x$HAVE_WGET" != "xno" || test "x$HAVE_CURL" != "xno"])
     AM_CONDITIONAL([HAVE_WGET_COMMAND], [test "x$HAVE_WGET" != "xno"])
 
+    # We require c11+
+    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#if __STDC_VERSION__ >= 201112L
```
victorjulien, line 133: this is largely a duplicate of what is done around line 290

victorjulien (changes requested):
`util-atomic.h` has code for when we don't support c11. The PR should also remove that I think.

jlucovsky:
Continued in #8924

PR #8914, opened 2023-05-24
Fix midstream exception policy bug 5825 - v1

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5825

Notes:
- **need feedback** on behavior when `drop` values are passed in `IDS mode`. We currently adjust these to `ignore`, but the new approach would mean changing this behavior to erroring out in such cases. 
- this PR contains exception policy stats work that are here only for debugging, they're a WIP but lower priority, and have already received feedback

Describe changes:
- add function to parse midstream exception policy config values
- Introduce erroring out when existing, but invalid config values are passed
- Update Exception Policy docs to include a matrix with all scenarios (for IDS and IPS)

```
SV_REPO=pr/1216
```

SV PR: https://github.com/OISF/suricata-verify/pull/1216

Matrix with Exception Policy  scenarios for Midstream:
image

### DIFF
--- a/doc/userguide/configuration/exception-policies.rst
+++ b/doc/userguide/configuration/exception-policies.rst
@@ -67,28 +67,26 @@ Exception policies are implemented for:
      - Expected behavior
    * - stream.memcap
      - memcap-policy
-     - If a stream memcap limit is reached, call the memcap policy on the packet
-       and flow.
+     - If a stream memcap limit is reached, apply the memcap policy to the packet and/or
+       flow.
    * - stream.midstream
      - midstream-policy
-     - If a session is picked up midstream, call the memcap policy on the packet
-       and flow.
+     - If a session is picked up midstream, apply the midstream policy to the flow.
    * - stream.reassembly.memcap
      - memcap-policy
-     - If stream reassembly reaches memcap limit, call the memcap policy on the
-       packet and flow.
+     - If stream reassembly reaches memcap limit, apply memcap policy to the
+       packet and/or flow.
    * - flow.memcap
      - memcap-policy
      - Apply policy when the memcap limit for flows is reached and no flow could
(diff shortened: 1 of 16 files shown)


doc/userguide/configuration/exception-policies.rst
```diff
@@ -131,11 +131,11 @@ midstream pick-ups enabled or not and the various exception policy values:
...
    * - Drop-flow
-     - Not valid.*
-     - Not valid.*
+     - Drop isn't valid in IDS mode, so the policy is set to `ignore`.
+     - Drop isn't valid in IDS mode, so the policy is set to `ignore`.
    * - Drop-packet
-     - Not valid.*
-     - Not valid.*
+     - Drop isn't valid in IDS mode, so the policy is set to `ignore`.
+     - Drop isn't valid in IDS mode, so the policy is set to `ignore`.
```
jufajardini: This part may need to be changed (or, rather, this commit dropped) if we decide that we will change Suri's current behavior for these.

src/util-exception-policy.c
```diff
@@ -303,6 +220,76 @@ enum ExceptionPolicy ExceptionPolicyParse(const char *option, const bool support
     return policy;
 }
 
+/**
+ * \brief Evaluate whether the provided exception policy configuration
+ * setting is valid in the given midstream policy scenario, and decide on the
+ * final exception policy value based on that.
+ */
```
jufajardini, line 227: This is still (also) a WIP, subject to refining - and possibly re-think how will this logic work, if we will take advantage of the existing `ExceptionPolicyParse` or not.

jufajardini:
Closing, as received feedback elsewhere already. Work continues.

PR #8916, opened 2023-05-24
jsonbuilder: attempt to handle memory allocation errors - v5

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/8855

Changes from last PR:
- Fix per-commit check
- Remove base64 update, will do separate PR

### DIFF
--- a/rust/src/jsonbuilder.rs
+++ b/rust/src/jsonbuilder.rs
@@ -17,6 +17,8 @@
 
 #![allow(clippy::missing_safety_doc)]
 
+use std::cmp::max;
+use std::collections::TryReserveError;
 use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::str::Utf8Error;
@@ -27,6 +29,7 @@ const INIT_SIZE: usize = 4096;
 pub enum JsonError {
     InvalidState,
     Utf8Error(Utf8Error),
+    Memory,
 }
 
 impl std::error::Error for JsonError {}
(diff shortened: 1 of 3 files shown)


suricata-qa:

Pipeline 14035

rust/src/jsonbuilder.rs
```diff
@@ -651,38 +705,62 @@ impl JsonBuilder {
...
                 );
-                self.buf.push_str(&error);
+                self.push_str(&error)?;
             }
         }
         Ok(())
     }
+
+    fn encode_base64(&mut self, val: &[u8]) -> Result<&mut Self, JsonError> {
+        let encoded_len = ((val.len() / 4) * 3) + 2;
```
catenacyber, line 715: This looks wrong : if val.len() == 1, this gives 2 instead of 4

This should rather be `4*((val.len()+2)/3)`, no ?

rust/src/jsonbuilder.rs
```diff
@@ -602,17 +646,27 @@ impl JsonBuilder {
     ///
     /// The string is encoded into an intermediate vector as its faster
     /// than building onto the buffer.
+    ///
+    /// TODO: Revisit this, would be nice to build directly onto the
+    ///    existing buffer.
```
catenacyber, line 651: I think this is worth revisiting this part.
Note that std::str::from_utf8 error case seems unreachable (as you encode the string correctly directly), so you should be able to push onto the self.buf directly

catenacyber (changes requested):
Still the base64 length looks suspicious + `encode_string` could/should be simplified

rust/src/jsonbuilder.rs
jasonish: Fixed.

jasonish:
> Still the base64 length looks suspicious + `encode_string` could/should be simplified

I feel touching encode_string here is out of scope, unless the changes added within scope are wrong.

catenacyber:
> > Still the base64 length looks suspicious + `encode_string` could/should be simplified
> 
> I feel touching encode_string here is out of scope, unless the changes added within scope are wrong.

I think you can have simpler changes with refactoring the `encode_string` code, which makes it easier to check that all allocations are checked, but I leave you final judge

jasonish:
Encoded length fixed here: https://github.com/OISF/suricata/pull/8927

PR #8893, opened 2023-05-17
Output alert applayer v4

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None, preliminary work for https://redmine.openinfosecfoundation.org/issues/5053 and app-layer plugins
Continuation of https://github.com/OISF/suricata/pull/8772

Describe changes:
- Fix setup-app-layer script so that it adds app-layer metadata to alerts
- Adds ftp metadata to alerts
- Adds tftp metadata to alerts
- Adds krb5 metadata to alerts
- Removes tx_id from ssh and http2 logging

Continues #8884 by removing tx_id from ssh and http2 logging

```
SV_BRANCH=pr/1196
```
https://github.com/OISF/suricata-verify/pull/1196

Still to do : 
- Create tickets for missing protocols : pgsql, dcerpc, dhcp, 
- Any ideas about the commit segmentation ?

PR #8922, opened 2023-05-25
Output alert applayer v5

catenacyber:
Continuation of https://github.com/OISF/suricata/pull/8772
https://redmine.openinfosecfoundation.org/issues/5977

Describe changes:
- Fix setup-app-layer script so that it adds app-layer metadata to alerts
- Adds ftp metadata to alerts
- Adds tftp metadata to alerts
- Adds krb5 metadata to alerts
- Removes tx_id from ssh and http2 logging

Continues #8893 with rebase and adding ticket number to krb5 commit

```
SV_BRANCH=pr/1196
```
https://github.com/OISF/suricata-verify/pull/1196

Still to do : 
- Create tickets for missing protocols : pgsql, dcerpc, dhcp, 
- Any ideas about the commit segmentation ?


src/output-json-mqtt.c
```diff
@@ -89,8 +81,10 @@ static int JsonMQTTLogger(ThreadVars *tv, void *thread_data,
         return TM_ECODE_FAILED;
     }
 
-    if (!rs_mqtt_logger_log(state, tx, thread->mqttlog_ctx->flags, js))
+    jb_open_object(js, "mqtt");
+    if (!rs_mqtt_logger_log(tx, thread->mqttlog_ctx->flags, js))
         goto error;
```
victorjulien, line 86: no close on error?

src/output.c
```diff
@@ -54,39 +54,32 @@
 #include "log-httplog.h"
 #include "output-json-http.h"
 #include "output-json-dns.h"
-#include "output-json-modbus.h"
 #include "log-tlslog.h"
 #include "log-tlsstore.h"
 #include "output-json-tls.h"
-#include "output-json-ssh.h"
 #include "log-pcap.h"
+// for SSHTxLogCondition
```
victorjulien, line 61: lets put these comments after the include on the same line

src/output.c
```diff
@@ -1032,6 +1025,109 @@ void OutputRegisterRootLoggers(void)
...
+    }
+
+    JsonBuilder *js = CreateEveHeader(p, LOG_DIR_PACKET, al->name, NULL, thread->ctx);
+    if (unlikely(js == NULL)) {
+        return TM_ECODE_FAILED;
+    }
+
+    jb_open_object(js, al->name);
+    if (!al->log(tx, js)) {
+        goto error;
```
victorjulien, line 1044: missing close on error?

src/output.c
```diff
@@ -1092,37 +1196,119 @@ void OutputRegisterLoggers(void)
...
-    JsonBitTorrentDHTLogRegister();
+    if (ConfGetNode("app-layer.protocols.bittorrent-dht") != NULL) {
+        /* Register as an eve sub-module. */
+        OutputRegisterTxSubModule(LOGGER_JSON_TX, "eve-log", "JsonBitTorrentDHTLog",
+                "eve-log.bittorrent-dht", OutputBitTorrentDHTLogInitSub, ALPROTO_BITTORRENT_DHT,
+                JsonGenericLogger, JsonLogThreadInit, JsonLogThreadDeinit, NULL);
+    }
+}
+
+static AppLayerLogger alert_applayer_loggers[ALPROTO_MAX] = {
```
victorjulien, line 1266: I don't really like this idea of handling these per protocol things be in a central table. Seems an anti-pattern wrt trying to make it as much a plugin as possible.

src/output.h
```diff
@@ -208,4 +208,12 @@ void OutputLoggerExitPrintStats(ThreadVars *, void *);
 void OutputSetupActiveLoggers(void);
 void OutputClearActiveLoggers(void);
 
+typedef struct AppLayerLogger {
+    AppProto proto;
+    const char *name;
+    bool (*log)(void *tx, struct JsonBuilder *jb);
```
victorjulien, line 214: func ptr should follow FuncName scheme.

src/output-json-dns.c
```diff
@@ -304,6 +304,21 @@ JsonBuilder *JsonDNSLogAnswer(void *txptr, uint64_t tx_id)
     }
 }
 
+bool AlertJsonDns(void *txptr, JsonBuilder *js)
```
victorjulien, line 307: should this return false if both calls return NULL?

victorjulien (changes requested):
Main issue is the central table, see inline.

rust/src/modbus/log.rs
```diff
@@ -27,7 +27,6 @@ pub extern "C" fn rs_modbus_to_json(tx: &mut ModbusTransaction, js: &mut JsonBui
```
victorjulien, line 27: this commit does a lot more than the commit message indicates

src/output.c
```diff
@@ -1092,37 +1196,119 @@ void OutputRegisterLoggers(void)
...
+    { ALPROTO_SMTP, NULL, NULL }, // special: logs both smtp and email fields
+    { ALPROTO_TLS, "tls", JsonTlsLogJSONExtended },
+    { ALPROTO_SSH, "ssh", rs_ssh_log_json },
+    { ALPROTO_IMAP, NULL, NULL },   // protocol detection only
+    { ALPROTO_JABBER, NULL, NULL }, // no parser, no logging
+    { ALPROTO_SMB, NULL, NULL },    // special: uses state
+    { ALPROTO_DCERPC, NULL, NULL }, // TODO missing
+    { ALPROTO_IRC, NULL, NULL },    // no parser, no logging
+    { ALPROTO_DNS, "dns", AlertJsonDns },
+    { ALPROTO_MODBUS, "modbus", (bool (*)(void *tx, struct JsonBuilder *jb))rs_modbus_to_json },
```
victorjulien, line 1279: lines like this are... yikes. Can we have something cleaner, like a typedef or something?

suricata-qa:

Pipeline 14069

src/output-json-mqtt.c
jasonish: No, the error handler just free's it.  Its not used anymore.

src/output.c
jasonish: Likewise here, just free'd on error, not used anymore.

catenacyber:
> Main issue is the central table, see inline.

Thanks for the review, sorry for taking long to respond.
I answer inline

src/output-json-mqtt.c
catenacyber: Indeed as Jason said.

To be more complete :
- When we log a MQTT event, if we fail to log the MQTT data, we do not log anything at all
- When we log an alert, if we fail to log the MQTT data, we still log the alert

src/output.c
catenacyber: ok

src/output.h
catenacyber: ok

rust/src/modbus/log.rs
catenacyber: Indeed, so, any ideas about the commit segmentation ? 
One big commit ? One commit per protocol ? (but if so, the first commit introducing the generic changes should also be applied on some protocol, otherwise, commit-check will fail because of unused functions)

Should I split this PR ?

src/output-json-dns.c
catenacyber: Likely so, this is a behavioral change

src/output.c
catenacyber: typedef seems good indeed

src/output.c
catenacyber: The use of this central table is indeed the central point, and it is meant to allow the use of app-layer plugins

What do we want ?
I think we want
- to avoid duplicating boilerplate code (and remove C file, to get better rust stats)
- that every app-layer protocols log their metadata for alerts, and for their own events

The solution currently used is to have
1.  For alerts, output-json-alert.c has a big switch with boilerplate wrapper code calling hardcoded app-layer logging functions, and this is missing for several protocols

So, the use of a central table, for alerts, 
- makes it easy to sport the missing protocols (and to add them, as is done in this PR for krb5 for instance)
- removes the boilerplate code (taking the mark, trying to log, closing or restoring the mark...)
- makes it easy to extend by plugins (that "just" need to extend the array)

2. For app-layer events, we have registration of boilerplate-wrapper functions that are registered in a central table
 `static OutputTxLogger *list[ALPROTO_MAX]` in output-tx.c 

Here, this PR 
- removes boilerplate code for simple logger (not ones using state, or flags...), like `JsonLogThreadInit` already exists but is not used by every protocol and `JsonGenericLogger` is introduced making use of the new table for simple loggers

So @victorjulien what do you think ?
It looks like a design flaw to have different wrapper functions for logging alert and app-layer metadata (causing not to have app-layer metadata for every protocol/alert)
Most of the protocols/cases are simple, but some require more complexity

Should I make a first part of this PR with the uncontroversial changes ? (first commits, use of `JsonLogThreadInit`)

catenacyber:
Continues in https://github.com/OISF/suricata/pull/8961

src/output.c
victorjulien: I think a centralized store doesn't have to use a table of hardcoded entries, but could also be more dynamic, like with many other parts of the engine. The parsers could then all a registration function that adds their entry into the store.

catenacyber: Well, not sure to see the point but this is quite easy to add :-)

src/output-json-dns.c
jasonish: > Likely so, this is a behavioral change

Not a big one right?  It would allow the caller to reset to mark to prevent an empty `dns` object from being logger in alert.

catenacyber: Does not look a big one 

#9005 is the part without behavior changes

Then I intend to do the behavioral changes about alerts : addition of protocols missing alert metadata (like krb5) + behavioral change for dns alert metadata

PR #8924, opened 2023-05-25
configure: Require c11 as the minimum C standard

jlucovsky:
Continuation of #8911 
This commit modifies the configure script to check for C11 by comparing the __STDC_VERSION__ with 201112L

An error message is issued if the C version is older than the C11 standard version.

Issue: 6029

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6029](https://redmine.openinfosecfoundation.org/issues/6029)

Describe changes:
- Updated `configure` to check `__STDC_VERSION__` against the C11 value.

Updates:
- Separate c11 from tls checks 

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -292,9 +292,18 @@
     AC_MSG_RESULT(ok)
 
     # check if our target supports c11
-    AC_MSG_CHECKING(for c11 support)
+    AC_MSG_CHECKING(for required c11 support)
+    # We require c11+ now
+    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#if __STDC_VERSION__ >= 201112L
+                                         int c11_supported;
+                                         #endif
+                                       ]])],
+                                       [AC_MSG_RESULT([yes])],
+                                       [AC_MSG_ERROR([C11 support is required but not available.])])
+
     OCFLAGS=$CFLAGS
     CFLAGS="-std=c11"
+    AC_MSG_CHECKING(for tls support)
     AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <stdlib.h>]],
                 [[ static _Thread_local int i; i = 1; i++; ]])],
             AC_MSG_RESULT([yes])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -292,9 +292,18 @@
...
+    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#if __STDC_VERSION__ >= 201112L
+                                         int c11_supported;
+                                         #endif
+                                       ]])],
+                                       [AC_MSG_RESULT([yes])],
+                                       [AC_MSG_ERROR([C11 support is required but not available.])])
+
     OCFLAGS=$CFLAGS
     CFLAGS="-std=c11"
+    AC_MSG_CHECKING(for tls support)
```
jasonish, line 306: Is `tls` support required as well?  `-std=c11` is only added if this `tls` support check passes as well. This is the reason we see the build failure on CentOS 7.  It passes the C11 test, but not the TLS test.

jasonish: CentOS 7 uses gcc 4.8.5 which passes the `__STDC_VERSION__` check, as well as any test for `-std=c11` passing.  However it does miss some C11 features like TLS, so the test should probably be about checking required C11 features, and not the actual C version.

jasonish:
FYI, I am prepping updates to the CentOS 7 job to use a newer gcc.

configure.ac
victorjulien: Guess the question now is if there is a good reason to force this issue, and I think my initial idea is "no". Lets reconsider in 8.

jasonish: > Guess the question now is if there is a good reason to force this issue, and I think my initial idea is "no". Lets reconsider in 8.

For 7, probably not.  For 8 it might be nice to enforce it sooner than later. I plan CentOS 7 RPMs for Suri 7, but requiring a newer gcc is not an issue there. But people building from source on CentOS 7 its a few more steps to get the newer compiler installed.

victorjulien: I guess I'm not really in a hurry even for 8 as I don't think we're really solving any issue other than a bit of house keeping.

jlucovsky: Should we retarget the redmine issue for a later release or should I address the issue (adding c11 iff tls is present)?

victorjulien: Lets close and revisit later when we're well into the 8 cycle.

jlucovsky:
> Lets close and revisit later when we're well into the 8 cycle.

Closing for now.

PR #8933, opened 2023-05-26
userguide: add quic eve doc; reference and schema - v1

jasonish:
Show how we can generate a table for protocol documentation. For now just QUIC
as its documented, but we could add the table for other protocols even if the
documentation is not complete, the description will just be empty.

Also add an appendix:
- Complete EVE index
- JSON schema


jasonish:
PDF Example:
Screenshot-2023-05-26_11-21

jasonish:
PGSql example, for a proto that exists in the schema, but has no descriptions:
https://jasonish-suricata.readthedocs.io/en/evedoc-v1/output/eve/eve-json-format.html?highlight=pgsql#field-reference

victorjulien:
Nice work. Can we shorten displayed names somehow?
e.g.
```
long.name | string | long name
\ hash | string | sha256 of long.name
```
Not sure I like this example, but wondering if there is a scheme that might work.

victorjulien:
Side note: the example shows how inconsistent the fields in the table are documented. We should probably provide guidance on content and formatting.

victorjulien:
in the table, what is the diff between `string` and `string[]`?

jasonish:
> Side note: the example shows how inconsistent the fields in the table are documented. We should probably provide guidance on content and formatting.

Have specific examples?  Given hardly any are documented at all.

jasonish:
> in the table, what is the diff between `string` and `string[]`?

String and array of strings.

jasonish:
> Nice work. Can we shorten displayed names somehow? e.g.
> 
> ```
> long.name | string | long name
> \ hash | string | sha256 of long.name
> ```
> 
> Not sure I like this example, but wondering if there is a scheme that might work.

Could do something better.  This current script is really dumb, just flattens, there is no relation between one field or another.

jasonish:
> Nice work. Can we shorten displayed names somehow? e.g.
> 
> ```
> long.name | string | long name
> \ hash | string | sha256 of long.name
> ```
> 
> Not sure I like this example, but wondering if there is a scheme that might work.

Played with this a little by throwing out sorting, and and bringing back my friend recursion, just to fight rendering in a ReST table which is limited, so I didn't get very far.

I can include *raw* HTML tables into the ReST which opens up possibilities, but they don't render in the PDF.  I've attached an image of something I was trying to go for..

In the end, to fit within Sphinx it might make sense to document each and every object in its own section, it might be quite verbose, but at least you can find stuff.

One of the problems with what you propose is Ctrl-F searchability, or at least I've found it troublesome in the past when looking for a specific key in a large nested structure.
Screenshot-2023-05-26_16-58

jufajardini (commented):
I like this!

The huge index looked difficult to digest, agreeing with some approach that would be per object.

The last table you've shared a screenshot of looked cool.

And I can see this making it easier to document the eve fields. Nice nice.

jasonish:
> The last table you've shared a screenshot of looked cool.

I believe it was generated with OpenAPI tooling, unfortunately, would be hard to adapt to Sphinx.  We can do some nice stylized HTML, however, it won't come out with the PDF.  So its a balance of what works with Sphinx, even though a richer UI might make sense, it would have to be its own page and not part of the userguide.

jasonish:
Alternate appendix listing: https://jasonish-suricata.readthedocs.io/en/evedoc-v2/appendix/eve-index.html

jufajardini:
> Alternate appendix listing: https://jasonish-suricata.readthedocs.io/en/evedoc-v2/appendix/eve-index.html



> > The last table you've shared a screenshot of looked cool.
> 
> I believe it was generated with OpenAPI tooling, unfortunately, would be hard to adapt to Sphinx. We can do some nice stylized HTML, however, it won't come out with the PDF. So its a balance of what works with Sphinx, even though a richer UI might make sense, it would have to be its own page and not part of the userguide.

Having this with a good UI could be a separate step, as nice to have, but not something mandatory for us to proceed with this path? 

The alternate appendix already improves readability somewhat, and we'd still be able to improve our documentation, by importing those into the userguide, when describing each portion, right?

jasonish:
> > Alternate appendix listing: https://jasonish-suricata.readthedocs.io/en/evedoc-v2/appendix/eve-index.html
> 
> > > The last table you've shared a screenshot of looked cool.
> > 
> > 
> > I believe it was generated with OpenAPI tooling, unfortunately, would be hard to adapt to Sphinx. We can do some nice stylized HTML, however, it won't come out with the PDF. So its a balance of what works with Sphinx, even though a richer UI might make sense, it would have to be its own page and not part of the userguide.
> 
> Having this with a good UI could be a separate step, as nice to have, but not something mandatory for us to proceed with this path?
> 
> The alternate appendix already improves readability somewhat, and we'd still be able to improve our documentation, by importing those into the userguide, when describing each portion, right?

Yeah, so just QUIC generated with the same script: https://jasonish-suricata.readthedocs.io/en/evedoc-v2/output/eve/eve-json-format.html?highlight=quic#event-type-quic

Needs:
- Command line option to replace the default top-level title with something else.  If we're just extracting "QUIC" we'd want to replace "Top Level" with "QUIC"
- Decide what heading levels to use, then make sure all places that will include this generated doc follow the levels laid out so we don't end up with the generated reference stuff starting a new section.

victorjulien:
> > > Alternate appendix listing: https://jasonish-suricata.readthedocs.io/en/evedoc-v2/appendix/eve-index.html
> > 
> > 
> > > > The last table you've shared a screenshot of looked cool.
> > > 
> > > 
> > > I believe it was generated with OpenAPI tooling, unfortunately, would be hard to adapt to Sphinx. We can do some nice stylized HTML, however, it won't come out with the PDF. So its a balance of what works with Sphinx, even though a richer UI might make sense, it would have to be its own page and not part of the userguide.
> > 
> > 
> > Having this with a good UI could be a separate step, as nice to have, but not something mandatory for us to proceed with this path?
> > The alternate appendix already improves readability somewhat, and we'd still be able to improve our documentation, by importing those into the userguide, when describing each portion, right?
> 
> Yeah, so just QUIC generated with the same script: https://jasonish-suricata.readthedocs.io/en/evedoc-v2/output/eve/eve-json-format.html?highlight=quic#event-type-quic
> 
> Needs:
> 
> * Command line option to replace the default top-level title with something else.  If we're just extracting "QUIC" we'd want to replace "Top Level" with "QUIC"
> * Decide what heading levels to use, then make sure all places that will include this generated doc follow the levels laid out so we don't end up with the generated reference stuff starting a new section.

I think this looks nice. I guess it would be helpful to have the type of the header as well. E.g. `'ja3 (object)` and `extensions[] (array of objects)`?

jasonish:
Replaced by https://github.com/OISF/suricata/pull/8971, with less programmatic like array notation.

PR #8948, opened 2023-06-01
Fix midstream exception policy bug 5825 - v2

jufajardini:
https://redmine.openinfosecfoundation.org/issues/5825

Previous (draf) PR: #8914 

Describe changes:
- split-up 'ExceptionPolicyParse' into functions with clearer responsibilities
- add 'ExceptionPolicyMidstreamParse' to try to handle the Midstream-related parsing scenarios
- midstream exception policy is now more strict with the values it accepts: with midstream enabled, only 'ignore' and 'pass-flow' are valid. 'packet' configs are never valid. 'auto' is the same as 'ignore', unless in 'IPS mode' and 'midstream: false'
- re-enable midstream exception policy
- Update documentation to include 'auto' config value

TODO:
- decide precedent for processing config values: 'IDS/IPS' or 'midstream: enabled'?
- update documentation to indicate changes in behavior (upgrade note?)
- update commit message to clearly state the behavioral changes
- double-check documentation for scenarios, see if they reflect what we want and have
- [CI-related] understand and fix scan-build found bug

SV_BRANCH=pr/1223

https://github.com/OISF/suricata-verify/pull/1223


src/defrag.c
```diff
@@ -1041,8 +1041,10 @@ Defrag(ThreadVars *tv, DecodeThreadVars *dtv, Packet *p)
 
     /* return a locked tracker or NULL */
     tracker = DefragGetTracker(tv, dtv, p);
-    if (tracker == NULL)
+    if (tracker == NULL) {
+        StatsIncr(tv, dtv->counter_defrag_max_hit);
```
victorjulien, line 1045: wrap in `if (tv != NULL && dtv != NULL) {`

src/util-exception-policy.c
```diff
@@ -173,7 +173,7 @@ enum ExceptionPolicy ExceptionPolicyParse(const char *option, const bool support
         } else {
             FatalErrorOnInit(
                     "\"%s\" is not a valid exception policy value. Valid options are drop-flow, "
-                    "pass-flow, bypass, drop-packet, pass-packet or ignore.",
+                    "pass-flow, bypass, reject, drop-packet, pass-packet or ignore.",
```
victorjulien: nit: compared to the other changes in this commit, this seems out of place. Can you move it into it's own commit?

src/util-exception-policy.c
```diff
@@ -141,80 +145,153 @@ static enum ExceptionPolicy PickPacketAction(const char *option, enum ExceptionP
...
+        policy = EXCEPTION_POLICY_DROP_FLOW;
+    }
+    return policy;
+}
+
+static enum ExceptionPolicy ExceptionPolicyMasterParse(const char *value)
+{
+    enum ExceptionPolicy policy = EXCEPTION_POLICY_NOT_SET;
+
+    if (value != NULL) {
```
victorjulien, line 197: can this ever be NULL? The only caller seems to not have a NULL value?

suricata-qa:

Pipeline 14177

suricata-qa:

Pipeline 14189

src/util-exception-policy.c
jufajardini: Removed the change from this commit. Didn't keep it in a separate commit, as this message ended up in a new function, so felt that the update could happen more naturally in that same commit.

src/util-exception-policy.c
jufajardini: You're right, we actually check for this in the function that calls this one.

jufajardini: You're right, we check for NULL in the functions that call this.

jufajardini:
Feedback incorporated in https://github.com/OISF/suricata/pull/8952

PR #8949, opened 2023-06-01
Flow drops/v4

victorjulien:
SV_BRANCH=pr/1224

https://redmine.openinfosecfoundation.org/issues/5802

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -334,7 +334,7 @@ static inline void FlowApplySignatureActions(
      * - sig is IP or PD only
      * - match is in applayer
      * - match is in stream */
-    if (s->action & (ACTION_DROP | ACTION_PASS)) {
+    if (pa->action & (ACTION_DROP | ACTION_PASS)) {
         DEBUG_VALIDATE_BUG_ON(s->type == SIG_TYPE_NOT_SET);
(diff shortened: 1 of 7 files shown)


suricata-qa:

Pipeline 14178

jufajardini (commented):
I like that we're handling the drop when updating the flow packet state, and that we're checking for the drop before updating app-layer for UDP protos. Seems like the drop handling is more contained. I saw that we are not calling 'FlowSetNoPacketInspectionFlag' and 'DecodeSetNoPacketInspectionFlag' after checking for the flow drops. Is that because we are handling the dropped flow earlier?

First of the commits seems to have a typo: "apply flow do" -> I imagine that should be "flow drop"?

victorjulien:
> I like that we're handling the drop when updating the flow packet state, and that we're checking for the drop before updating app-layer for UDP protos. Seems like the drop handling is more contained. I saw that we are not calling 'FlowSetNoPacketInspectionFlag' and 'DecodeSetNoPacketInspectionFlag' after checking for the flow drops. Is that because we are handling the dropped flow earlier?

Yeah, the detect engine checks for `FLOW_ACTION_DROP`, so the other flags aren't needed for this case.

victorjulien:
Merged in #8951

PR #8950, opened 2023-06-01
6.0.x: fix for issues 5868

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/5868

I have a feeling there is more to this than just delaying file cleanup.  With
master this appears to have been fixed with the move to per-tx file storage.

SV_BRANCH=pr/1225

### DIFF
--- a/src/flow-worker.c
+++ b/src/flow-worker.c
@@ -557,9 +557,6 @@ static TmEcode FlowWorker(ThreadVars *tv, Packet *p, void *data)
     // Outputs.
     OutputLoggerLog(tv, p, fw->output_thread);
 
-    /* Prune any stored files. */
-    FlowPruneFiles(p);
-
     /*  Release tcp segments. Done here after alerting can use them. */
     if (p->flow != NULL) {
         DEBUG_ASSERT_FLOW_LOCKED(p->flow);
@@ -582,6 +579,9 @@ static TmEcode FlowWorker(ThreadVars *tv, Packet *p, void *data)
         Flow *f = p->flow;
         FlowDeReference(&p->flow);
         FLOWLOCK_UNLOCK(f);
+
+        /* Prune any stored files. */
+        FlowPruneFiles(p);
     }
 
 housekeeping:
--- a/rust/src/nfs/nfs.rs
+++ b/rust/src/nfs/nfs.rs
@@ -175,7 +175,7 @@ pub struct NFSTransaction {
     pub file_handle: Vec<u8>,
 
     /// Procedure type specific data
-    /// TODO see if this can be an Option<Box<NFSTransactionTypeData>>. Initial
+    /// TODO see if this can be an `Option<Box<NFSTransactionTypeData>>`. Initial
     /// attempt failed.
     pub type_data: Option<NFSTransactionTypeData>,
 


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .uptime | 141 | 210 | 148.94% |
| .tcp.rst | 81627 | 89907 | 110.14% |
| .flow.spare | 2155079 | 2023729 | 93.91% |
| .flow.memuse | 579234048 | 1787525248 | 308.6% |
| .capture.kernel_drops | 0 | 8998 | 0.00 | 

Pipeline 14192

src/flow-worker.c
```diff
@@ -582,6 +579,9 @@ static TmEcode FlowWorker(ThreadVars *tv, Packet *p, void *data)
         Flow *f = p->flow;
         FlowDeReference(&p->flow);
         FLOWLOCK_UNLOCK(f);
+
+        /* Prune any stored files. */
+        FlowPruneFiles(p);
```
victorjulien, line 584: this is definitely wrong. Flow work needs to happen while we have the flow lock.

jasonish: So just move up a few lines so its within the lock?  A little confusing as in master-6.0.x it's calle1d outside of a condition that may open the lock.

norg: (just a small side note, this small change has the impact that the file is written correct:

Before
```
1634 0 -rw-r--r-- 1 root root 0 Jun  2 13:46 filestore/1e/1e3f22dd503676f6ad719c8fb48da377690e12408b0bf3712167cffd01e6388e
```
After the change from Jason:
```
1899 1012K -rw-r--r-- 1 root root 1012K Jun  2 13:50 filestore/1e/1e3f22dd503676f6ad719c8fb48da377690e12408b0bf3712167cffd01e6388e
```

So it leads to the correct path IMHO :)

norg: Based on my testing it needs to be moved to the spot between `FlowDeReference` and `FLOWLOCK_UNLOCK` since any earlier (towards the top) spot would still make it `0b` filesize.

One question, we also have the `FlowPruneFiles` in `FlowWorkerFlowTimeout`, do we need to change the position there as well?

victorjulien: So it seems this just has the same effect as removing the call completely, as `p` won't have a flow attached to it anymore after `FlowDeReference`. That suggests the fix is wrong, as file pruning needs to happen.

norg: that's correct, removing the `FlowPruneFiles(p)` call from that function also results in the file being correctly written to disk :thinking: 

Could it be that the call actually is the reason why the file is at `0b`?
Could it be that the file was written correct but is overwritten again to filesize `0b` with that call?

jasonish:
Closing. Opened https://github.com/OISF/suricata/pull/8954 for further discussion of this issue.

src/flow-worker.c
victorjulien: In master things work quite differently. File pruning is called from `AppLayerParserTransactionsCleanup` -> `AppLayerParserFileTxHousekeeping` there.

I think the focus in 6 should be on the logic in `FilePruneFile`

jasonish: Ok.. Specifically for this test, `FilePruneFile` is never called in master like it is in master-6.0.x which does seem a little odd - is it smarter about when to prune?

But in master-6.0.x, the issue is that `FILE_STORE` is not set early enough, so the file is pruned out before `FILE_STORE` is set.  #8954 does inspection earlier, setting `FILE_STORE` sooner, keeping the file around.

This test, https://github.com/OISF/suricata-verify/tree/master/tests, also shows a discrepancy between 6 and master behavior that I'm not sure should be there.  Will investigate further tho.

jasonish: Wireshark does agree with the `master` behavior of 8 files, suggesting its a little smart about when to prune.  Removing the pruning in 6 does result in 8 files as well, which seems to suggest pruning is a little too aggressive in 6.  Thoughts?

PR #8785, opened 2023-04-30
Draft changes for issue 4145

jlucovsky:
This commit is *draft* and demonstrates an approach to common registrations for file data.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4145](https://redmine.openinfosecfoundation.org/issues/4145)

Describe changes:
- Build on 4141 use common handling for all file based detection

With changes:
```
$ hyperfine -m3 'sudo rm -rf /tmp/ll/* && ./src/suricata -c suricata.yaml -l /tmp/ll -S ~/suricata.rules -r ~/TLP_WHITE-smb-16kfiles-no-close.pcap --set outputs.1.eve-log.threaded=true'
Benchmark 1: sudo rm -rf /tmp/ll/* && ./src/suricata -c suricata.yaml -l /tmp/ll -S ~/suricata.rules -r ~/TLP_WHITE-smb-16kfiles-no-close.pcap --set outputs.1.eve-log.threaded=true
  Time (mean ± σ):     1506.882 s ± 30.635 s    [User: 1522.916 s, System: 65.111 s]
  Range (min … max):   1476.483 s … 1537.747 s    3 runs
  ```

Comparative numbers from `master`:
```
$ hyperfine -m3 'sudo rm -rf /tmp/ll/* && ./src/suricata -c suricata.yaml -l /tmp/ll -S ~/suricata.rules -r ~/TLP_WHITE-smb-16kfiles-no-close.pcap --set outputs.1.eve-log.threaded=true'
Benchmark 1: sudo rm -rf /tmp/ll/* && ./src/suricata -c suricata.yaml -l /tmp/ll -S ~/suricata.rules -r ~/TLP_WHITE-smb-16kfiles-no-close.pcap --set outputs.1.eve-log.threaded=true
  Time (mean ± σ):     1555.037 s ± 25.689 s    [User: 1577.058 s, System: 62.196 s]
  Range (min … max):   1528.016 s … 1579.144 s    3 runs
  ```
### Provide values to any of the below to override the defaults.

To use a pull request use a branch name like `pr/N` where `N` is the pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=pr/389
```

PR #8964, opened 2023-06-05
Detect files : 

catenacyber:
None, preliminary work for https://redmine.openinfosecfoundation.org/issues/5053 and app-layer plugins (that can use files and their keywords)

Describe changes:
- detect/files: reuse AppLayerParserSupportsFiles instead of than relisting the protocols
- detect/files: centralize definition of protocols : so that they can be used by all file keywords

It still bothers me that `detect-filename.c` defines the `files` DetectBuffer that is used by all file keywords...

I do not know if it is a significant change but the progress for (like HTP_REQUEST_BODY for HTTP1) was not used in file name, file magic...

### DIFF
--- a/src/detect-engine-file.c
+++ b/src/detect-engine-file.c
@@ -49,6 +49,29 @@
 #include "util-profiling.h"
 #include "util-validate.h"
 
+FileAppProto file_protos_ts_static[] = {
+    { ALPROTO_HTTP1, HTP_REQUEST_BODY },
+    { ALPROTO_SMTP, 0 },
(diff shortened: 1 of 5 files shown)


victorjulien:
This looks like it might overlap with some of work by @jlucovsky

catenacyber:
> This looks like it might overlap with some of work by @jlucovsky

Jeff, is this https://github.com/OISF/suricata/pull/8785 ?

Replaced by https://github.com/OISF/suricata/pull/8968

PR #8961, opened 2023-06-05
Output alert applayer v6

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None, preliminary work for https://redmine.openinfosecfoundation.org/issues/5053 and app-layer plugins
Continuation of https://github.com/OISF/suricata/pull/8772
https://redmine.openinfosecfoundation.org/issues/5977

Describe changes:
- Fix setup-app-layer script so that it adds app-layer metadata to alerts
- Adds ftp metadata to alerts
- Adds tftp metadata to alerts
- Adds krb5 metadata to alerts
- Removes tx_id from ssh and http2 logging

Continues #8922 with review taken into account

```
SV_BRANCH=pr/1196
```
https://github.com/OISF/suricata-verify/pull/1196

Still to do : 
Any ideas about the commit segmentation ? and PR splitting ?
I guess I should make a first smaller version with uncontroversial changes

PR #8967, opened 2023-06-05
output/snmp: log version from tx

catenacyber:

Describe changes:
- output/snmp: log version from tx, and not from state

This is a functional change from #8961 

If a SNMP flow starts with a V2 version exchange, then there is a V3, we will now log V3 for the second transaction


suricata-qa:

Pipeline 14251

catenacyber:
Ok, done in https://github.com/OISF/suricata/pull/8977

PR #8974, opened 2023-06-06
Fuzz base64/v2

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/6050

### DIFF
--- a/src/tests/fuzz/fuzz_decodebase64.c
+++ b/src/tests/fuzz/fuzz_decodebase64.c
@@ -1,0 +1,55 @@
+/**
+ * @file
+ * @author Shivani Bhardwaj <shivani@oisf.net>
+ * fuzz target for DecodeBase64
+ */
+
+#include "suricata-common.h"
+#include "suricata.h"
+#include "util-base64.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+static int initialized = 0;
+
+static void ConstrainedDestinationSizeTest(const uint8_t *src, size_t len)
+{
+    for (uint8_t mode = BASE64_MODE_RELAX; mode <= BASE64_MODE_RFC4648; mode++) {
+        uint32_t dest_size = 3 * len / 4;
+        uint8_t *dest = malloc(dest_size);
+        uint32_t consumed_bytes = 0;
+        uint32_t decoded_bytes = 0;
+
+        DecodeBase64(dest, dest_size, src, len, &consumed_bytes, &decoded_bytes, mode);
+    }
+}
+
+static void AmpleDestinationSizeTest(const uint8_t *src, size_t len)
+{
+    for (uint8_t mode = BASE64_MODE_RELAX; mode <= BASE64_MODE_RFC4648; mode++) {
+        uint32_t dest_size = len;
+        uint8_t *dest = malloc(dest_size);
+        uint32_t consumed_bytes = 0;
+        uint32_t decoded_bytes = 0;
+
+        DecodeBase64(dest, dest_size, src, len, &consumed_bytes, &decoded_bytes, mode);
+    }
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    if (initialized == 0) {
+        // Redirects logs to /dev/null
+        setenv("SC_LOG_OP_IFACE", "file", 0);
+        setenv("SC_LOG_FILE", "/dev/null", 0);
+        // global init
+        InitGlobal();
+        run_mode = RUNMODE_UNITTEST;
+        initialized = 1;
+    }
+
+    ConstrainedDestinationSizeTest(data, size);
+    AmpleDestinationSizeTest(data, size);
+
+    return 0;
+}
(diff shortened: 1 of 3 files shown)


catenacyber:
@inashivb did you investigate the buffer overflow found by CIFuzz with this new target ?

src/tests/fuzz/fuzz_decodebase64.c
```diff
@@ -0,0 +1,55 @@
...
+        uint32_t dest_size = 3 * len / 4;
+        uint8_t *dest = malloc(dest_size);
+        uint32_t consumed_bytes = 0;
+        uint32_t decoded_bytes = 0;
+
+        DecodeBase64(dest, dest_size, src, len, &consumed_bytes, &decoded_bytes, mode);
+    }
+}
+
+static void AmpleDestinationSizeTest(const uint8_t *src, size_t len)
```
catenacyber, line 27: Why do we need `AmpleDestinationSizeTest` ? Is `ConstrainedDestinationSizeTest` not enough ?

src/tests/fuzz/fuzz_decodebase64.c
```diff
@@ -0,0 +1,55 @@
...
+#include "util-base64.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+
+static int initialized = 0;
+
+static void ConstrainedDestinationSizeTest(const uint8_t *src, size_t len)
+{
+    for (uint8_t mode = BASE64_MODE_RELAX; mode <= BASE64_MODE_RFC4648; mode++) {
+        uint32_t dest_size = 3 * len / 4;
```
catenacyber, line 18: I think it is `3 * ((len + 3)/ 4)`

Check with len = 1 ;-)

suricata-qa:

Pipeline 14281

src/tests/fuzz/fuzz_decodebase64.c
inashivb: Indeed. Missing a very apparent edge case.

src/tests/fuzz/fuzz_decodebase64.c
inashivb: I thought that if due to constrained size if some of the data wasn't fully decoded but was capable of hitting a Debug validation then this would help. Wdyt?

src/tests/fuzz/fuzz_decodebase64.c
inashivb: Wait. We want to constrain the space here. I think I kept this here this way because dest_size is controlled by the callers. So, what if the caller has run out of space in the dest buffer but still has a byte to decode? Would it not still make a valid test case?

inashivb:
> @inashivb did you investigate the buffer overflow found by CIFuzz with this new target ?

Just did. Seems to be on the case you pointed out. 0 bytes on destination buffer and length of the source string 1.
Please let me know what you think of what I think in the inline comments.

src/tests/fuzz/fuzz_decodebase64.c
inashivb: One example is the case you pointed out where we don't have enough dest space for decoding 1byte. While that should be checked as that may happen, it should also be checked that if there is just 1 byte and enough destination space, it actually does get decoded properly.

catenacyber: But `ConstrainedDestinationSizeTest` should always have enough space, right ?

src/tests/fuzz/fuzz_decodebase64.c
catenacyber: Does Suricata want to partially decode base64 ?

catenacyber (changes requested):
Will you do a new iteration Shivani ?

inashivb:
> Will you do a new iteration Shivani ?

yes. Philippe. I also have to answer your questions. I'm first trying to do #9160 as it is for the release. While doing that, I'm also able to find answers to some questions.

suricata-qa:

Pipeline 14281

inashivb:
Needs more work. Is stale and has conflicts now.

src/tests/fuzz/fuzz_decodebase64.c
inashivb: destination size is controlled by the caller and may not have enough space, I think.
refs:
1. https://github.com/OISF/suricata/blob/master/src/util-decode-mime.c#L1225
2. https://github.com/OISF/suricata/blob/master/src/util-base64.c#L178

inashivb: wdyt?

src/tests/fuzz/fuzz_decodebase64.c
inashivb: yes, in case we run out of space to decode more data, I think. wdyt?

catenacyber: So, `dest_size` should be random from the input ?

src/tests/fuzz/fuzz_decodebase64.c
catenacyber: I guess I would go for only one test, where the destination size is fuzzingly randomized

inashivb: That sounds a lot better. Thank you!

PR #8988, opened 2023-06-07
windows: add -lntdll to fix windows builds

jasonish:
Rust 1.70 has introduced some possible issues between LLVM and gcc
causing link errors that are fixed by explicitly adding -lntdll.
    
Thanks to https://github.com/extendr/rextendr/pull/285 for the fix.


jasonish:
This is needed in `master-6.0.x`, but cleanly cherry-picks.

victorjulien:
Merged in #8994 and #8995, thanks!

PR #8989, opened 2023-06-07
doc: add missing rule to engine-analysis section - v1

jufajardini:
I was checking the documentation for engine analysis, and noticed that the first example was missing the accompanying rule that should be associated with it.

I checked the suricata-verify repo and found a rule that seemed to be a good match for it, so cleaned it up and added it to the section.


victorjulien (approved):

victorjulien:
Merged in #8994, thanks!

PR #8999, opened 2023-06-09
rfb: ensure we do not return errors from the applayer parser for unimplemented traffic

satta:

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [#5912](https://redmine.openinfosecfoundation.org/issues/5912)

Describe changes:
- Ensure we do not return errors from the applayer parser if we simply face data we can not parse because we did not implement the necessary logic. This traffic might not be invalid, but simply different. Returning an error will cause drops in IPS mode with default configuration (see Redmine bug) so we want to avoid dropping legitimate traffic just for the sake of our laziness.
- Add event support to RFB applayer parser, as well as event rules.
- Make version string parsing more stringent to actually recognize the protocol header.


catenacyber:
Are you okay Sascha, or still pushing more commits ?

satta:
> Are you okay Sascha, or still pushing more commits ?

Yeah, sorry for the trickle of fixes, but it looks like the obvious things are addressed now. That's it AFAICS.

rust/src/rfb/rfb.rs
```diff
@@ -205,7 +215,10 @@ impl RFBState {
                             );
                         }
                         Err(_) => {
-                            return AppLayerResult::err();
+                            // We even failed to parse the protocol version.
+                            // Continue the flow but stop trying to parse the protocol.
```
catenacyber, line 219: What sense does it make to try to continue app-layer parsing / reassembly here ?
Why not return error ?

rust/src/rfb/rfb.rs
```diff
@@ -227,15 +240,26 @@ impl RFBState {
                             match chosen_security_type {
                                 2 => self.state = parser::RFBGlobalState::TCVncChallenge,
                                 1 => self.state = parser::RFBGlobalState::TSClientInit,
-                                _ => return AppLayerResult::err(),
+                                _ => {
+                                    if let Some(current_transaction) = self.get_current_tx() {
+                                        RFBState::set_event(
```
catenacyber, line 245: Maybe `set_event` should belong to tx and not state

rust/src/rfb/rfb.rs
```diff
@@ -227,15 +240,26 @@ impl RFBState {
...
+                                            current_transaction,
+                                            RFBEvent::UnimplementedSecurityType,
+                                        );
+                                    }
+                                    // We have just have seen a security type we don't know about.
+                                    // This is not bad per se, it might just mean this is a
+                                    // proprietary one not in the spec.
+                                    // Continue the flow but stop trying to map the protocol.
+                                    self.state = parser::RFBGlobalState::Skip;
```
catenacyber, line 254: See What SSH does for instance when it gets into encrypted traffic

rust/src/rfb/rfb.rs
```diff
@@ -227,15 +240,26 @@ impl RFBState {
...
+                                    return AppLayerResult::ok();
+                                }
                             }
 
                             if let Some(current_transaction) = self.get_current_tx() {
                                 current_transaction.ts_security_type_selection = Some(request);
                                 current_transaction.chosen_security_type =
                                     Some(chosen_security_type as u32);
-                            } else {
-                                return AppLayerResult::err();
```
catenacyber, line 238: debug validation that this is unreachable ?

rust/src/rfb/rfb.rs
```diff
@@ -308,20 +342,25 @@ impl RFBState {
...
-                    SCLogDebug!("Reversed traffic, expected response.");
-                    return AppLayerResult::err();
+                parser::RFBGlobalState::Skip => {
+                    // End of parseable handshake reached, skip rest of traffic
+                    return AppLayerResult::ok();
                 }
                 _ => {
-                    SCLogDebug!("Invalid state for request {}", self.state);
-                    current = b"";
+                    // We have gotten out of sync with the expected state flow.
```
catenacyber, line 359: How is this reachable ?
Which value from the rust enum is missing ?

rust/src/rfb/rfb.rs
```diff
@@ -549,10 +628,8 @@ impl RFBState {
                         Ok((_rem, request)) => {
                             if let Some(current_transaction) = self.get_current_tx() {
                                 current_transaction.tc_failure_reason = Some(request);
-                            } else {
-                                return AppLayerResult::err();
                             }
-                            return AppLayerResult::err();
```
catenacyber, line 555: oh @jasonish is there a linter to catch this ?
```
} else {
return x;
}
return x;
```

rust/src/rfb/rfb.rs
catenacyber: So, I understand that for unknown values for rust enumerations, we do not want to error, but maybe get in a state where we can no longer analyze...

But if the first pass of parsing the protocol version is unrecognized, I think this should still be an error

catenacyber (changes requested):
Looks on the good track.

I let you review the commit history

rust/src/rfb/rfb.rs
jasonish: > oh @jasonish is there a linter to catch this ?
> 
> ```
> } else {
> return x;
> }
> return x;
> ```

If the second one is dead code, rust will warn about it by default.

catenacyber: No dead code but it can be simplified to `return x` ;-)

jasonish: I don't see a lint here that looks like it would do this: https://rust-lang.github.io/rust-clippy/master/index.html#/return

rules/rfb-events.rules
```diff
@@ -0,0 +1,11 @@
+# RFB app-layer event rules.
+#
+# These SIDs fall in the 2228000+ range. See:
```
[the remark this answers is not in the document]
satta: True. BTW, 2232000 (which I picked first but is allocated to FTP, is not listed in https://redmine.openinfosecfoundation.org/projects/suricata/wiki/AppLayer. I added that.

rust/src/rfb/rfb.rs
satta: YEah, you are probably right. My approach here was to never cause RFB to be interrupted for parsing reasons, but maybe this is a bit too lenient. I'll make this an error.

rust/src/rfb/rfb.rs
satta: Well, the traffic is not encrypted... :)
I actually did that and my previous approach went into bypass and skip inspection here, but maybe someone wants to do just `content` based pattern matching on RFB traffic, and I do not want to break that. So I deliberately did not do that here.

rust/src/rfb/rfb.rs
satta: We currently have a global state (i.e. that is used for both directions), but if traffic can not be parsed as expected, we might not have advanced a state for one direction but received data in the "unexpected" direction, causing the parser to end up here. In that case I decided to stop parsing further.

satta: > How is this reachable ?
> Which value from the rust enum is missing ?

For example, any `TC*` (to-client) state that we might still be in globally, but are calling this function which handles request (to-server) traffic.

rust/src/rfb/rfb.rs
satta: There is only ever one tx per flow. I guess I can simplify that. Thanks!

rust/src/rfb/rfb.rs
catenacyber: Thanks, maybe add a comment then ;-)

rust/src/rfb/rfb.rs
catenacyber: You are right :-)

rust/src/rfb/rfb.rs
satta: Done.

rust/src/rfb/rfb.rs
satta: Done.

satta:
New PR: #9041

PR #8997, opened 2023-06-08
Fix midstream exception policy bug 5825 - v5

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5825

Previous PR: https://github.com/OISF/suricata/pull/8969

Describe changes:
- Rebase
- Check for `NULL` before incrementing stats in the right place

SV_BRANCH=pr/1230
SV PR: https://github.com/OISF/suricata-verify/pull/1230

PR #9011, opened 2023-06-12
Fix midstream exception policy bug 5825 - v6

jufajardini:
https://redmine.openinfosecfoundation.org/issues/5825

Previous PR: #8997 

Describe changes:
- handle case when exception policy values are in 'auto' or default, to ensure we wouldn't pass 'drop-flow' to exception policies that don't support flow action
- change most `Warning` log levels to `Notice`, except for when we ignore not valid config values in IDS mode (in `SetIPSOption`)
- rebase

Note: I'm not sure this will be enough to prevent the CIFuzz failure, as I could still see the output for `--list-app-layer` show, regardless of changing the log level...

```
SV_BRANCH=pr/1245
```
Suricata-verify PR https://github.com/OISF/suricata-verify/pull/1245


jufajardini:
force-pushed adjusting the log level to `info` instead of `notice`

jufajardini:
Checking all the failures...

jufajardini:
I'm thinking that the issue with us seeing that output for `--list-app-layer` could be connected to the fact that this isn't something that runs after the whole config has been read, it just checks for the `app-layers` section 🤔

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/9016

PR #9012, opened 2023-06-12
rust: fix unit test link error on Rust 1.70 - v1

jasonish:
Rust 1.70 appears to now link code on both branches of `if cfg!(test)` now causing Rust unit tests to fail as that pattern was used to disable functions only available when linked with the Suricata C code.

To work-around this issue, provide two versions of the `new` function, one for unit tests and one when running as an application.

This attempt confines the changes to `frames.rs`.  An alternative would be for all callers of the frame functions to wrap their calls in conditionals.

### DIFF
--- a/rust/src/frames.rs
+++ b/rust/src/frames.rs
@@ -17,9 +17,11 @@
 
 use crate::applayer::StreamSlice;
 use crate::core::Flow;
+#[cfg(not(test))]
 use crate::core::STREAM_TOSERVER;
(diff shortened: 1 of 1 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| .flow.memuse | 519171056 | 559056760 | 107.68% | 

Pipeline 14443

victorjulien:
Merged in #9021, thanks!

PR #9013, opened 2023-06-12
detect/http2: warning that http2.header is removed

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5780

Describe changes:
- detect/http2: warning that http2.header is removed

See https://github.com/OISF/suricata/pull/8775#issuecomment-1576215565

>> Do we keep the deprecated keyword in 6 ?
> Yes. But maybe it will be good to add a warning that it is removed in 7.

### DIFF
--- a/src/detect-http2.c
+++ b/src/detect-http2.c
@@ -806,6 +806,8 @@ static int DetectEngineInspectHttp2HeaderName(
 
 static int DetectHTTP2headerSetup(DetectEngineCtx *de_ctx, Signature *s, const char *arg)
 {
+    SCLogWarning(SC_WARN_POOR_RULE, "Keyword http2.header is removed in Suricata 7, and replaced "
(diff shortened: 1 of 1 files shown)


src/detect-http2.c
```diff
@@ -806,6 +806,8 @@ static int DetectEngineInspectHttp2HeaderName(
 
 static int DetectHTTP2headerSetup(DetectEngineCtx *de_ctx, Signature *s, const char *arg)
 {
+    SCLogWarning(SC_WARN_POOR_RULE, "Keyword http2.header is removed in Suricata 7, and replaced "
```
jasonish, line 809: Maybe `SC_WARN_DEPRECATED`?

victorjulien: indeed, that seems the correct code

catenacyber: nice

PR #9017, opened 2023-06-12
github-ci: update actions - v1

jasonish:
Update github-actions to their latest versions. Resolves some warnings that GitHub Actions was outputting on our jobs.


jasonish:
If the hash was already the most recent version, I unified to the hash.  Otherwise I used the latest version without hash.  Dependabot hasn't been the most reliable here, and I'm wondering if its more reliable with one over the other.

victorjulien:
Merged in #9035, thanks!

PR #9036, opened 2023-06-16
Multipart mime 3487 v16

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3487

Describe changes:
- convert HTTP to use new rust mime parser
- convert SMTP to use new rust mime parser

Follows #8886 with rebase and some review comments taken into account

Still to do : 
- investigate the QA SMTP changes : @ct0br0 would you send me a pcap when QA is done for this PR ?
- do a C fix for body_md5 cf https://github.com/OISF/suricata-verify/pull/1035#issuecomment-1563083933
- convert previous unit tests to SV tests

```
SV_BRANCH=pr/1035
```
https://github.com/OISF/suricata-verify/pull/1035

PR #9037, opened 2023-06-16
mime: compute full body md5

catenacyber:
Preliminary work for https://redmine.openinfosecfoundation.org/issues/3487

Describe changes:
- computes full `email.body_md5`

Needed before #9036 

```
SV_BRANCH=pr/1252
```
https://github.com/OISF/suricata-verify/pull/1252

### DIFF
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -2208,17 +2208,6 @@ static int ProcessMimeBody(const uint8_t *buf, uint32_t len,
     int body_found = 0;
     uint16_t tlen;
 
-    if (!g_disable_hashing) {
-        if (MimeDecGetConfig()->body_md5) {
-            if (state->body_begin == 1) {
-                if (state->md5_ctx == NULL) {
-                    state->md5_ctx = SCMd5New();
-                }
-            }
-            SCMd5Update(state->md5_ctx, buf, len + state->current_line_delimiter_len);
-        }
-    }
-
     /* pass empty lines on if we're parsing the body, otherwise we have no use
      * for them, and in fact they would disrupt the state tracking */
     if (len == 0) {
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 14717

src/util-decode-mime.c
```diff
@@ -2348,6 +2337,18 @@ static int ProcessMimeEntity(const uint8_t *buf, uint32_t len,
                 MAX_LINE_LEN);
     }
 
+    if ((state->state_flag != HEADER_READY && state->state_flag != HEADER_STARTED) ||
```
jasonish, line 2340: More of a question.. Why not check state >= HEADER_STARTED?

jasonish:
With respect to what is the full body MD5, I see the check for `CTNT_IS_BODYPART`.  Are we parsing the header, then taking the rest of the body verbatim for the MD5?  Or are there still parts after the header that won't be computed as part of the body?

src/util-decode-mime.c
catenacyber: Same condition (or rather contrary condition) is used below at line 2353 before calling `ProcessMimeHeaders` xor `ProcessMimeBody`

catenacyber:
> Are we parsing the header, then taking the rest of the body verbatim for the MD5? 

That is what I would expect

> Or are there still parts after the header that won't be computed as part of the body?

Today, some headers/boundaries are not part of the body MD5

jasonish:
> > Are we parsing the header, then taking the rest of the body verbatim for the MD5?
> 
> That is what I would expect

So that is what you think we should do, but are currently not?

catenacyber:
> > > Are we parsing the header, then taking the rest of the body verbatim for the MD5?
> > 
> > 
> > That is what I would expect
> 
> So that is what you think we should do, but are currently not?

Exactly

The current logic is too complex to express about what is not taken in the md5

src/util-decode-mime.c
```diff
@@ -2348,6 +2337,18 @@ static int ProcessMimeEntity(const uint8_t *buf, uint32_t len,
...
+            (state->stack->top->data->ctnt_flags & CTNT_IS_BODYPART)) {
+        if (!g_disable_hashing) {
```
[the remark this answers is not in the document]
catenacyber: ok

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9100

PR #9039, opened 2023-06-16
detect/byte-jump: Permit variable name for nbytes value

jlucovsky:
Continuation of #9028 
  
This PR adds the ability to use a variable name with the nbytes value in a `byte_jump` keyword option list.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6105](https://redmine.openinfosecfoundation.org/issues/6105)

Describe changes:
- Add logic to detect and use variable names for `nbytes` in the `byte_jump` option list
- Document variable name and add to Snort difference document.

Updates
- Fixed doc formatting
- Updated error message to use correct variable

```
SV_REPO=
SV_BRANCH=pr/1254
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/doc/userguide/rules/payload-keywords.rst
+++ b/doc/userguide/rules/payload-keywords.rst
@@ -516,7 +516,7 @@ The ``byte_jump`` keyword allows for the ability to select a ``<num of bytes>`` 
 
 Format::
 
-  byte_jump:<num of bytes>, <offset> [, relative][, multiplier <mult_value>] \
+  byte_jump:<num of bytes> | <variable-name>, <offset> [, relative][, multiplier <mult_value>] \
         [, <endian>][, string, <num_type>][, align][, from_beginning][, from_end] \
         [, post_offset <value>][, dce][, bitmask <value>];
 
@@ -524,6 +524,7 @@ Format::
 
 +-----------------------+-----------------------------------------------------------------------+
 | <num of bytes>        | The number of bytes selected from the packet to be converted          |
+|                       | or the name of a byte_extract/byte_math variable.                     |
 +-----------------------+-----------------------------------------------------------------------+
 | <offset>		| Number of bytes into the payload					|
 +-----------------------+-----------------------------------------------------------------------+
--- a/src/detect-bytejump.h
+++ b/src/detect-bytejump.h
@@ -40,9 +40,11 @@
 #define DETECT_BYTEJUMP_DCE       BIT_U16(6) /**< "dce" enabled */
 #define DETECT_BYTEJUMP_OFFSET_BE BIT_U16(7) /**< "byte extract" enabled */
 #define DETECT_BYTEJUMP_END       BIT_U16(8) /**< "from_end" jump */
+#define DETECT_BYTEJUMP_NBYTES_VAR BIT_U16(9) /**< nbytes string*/
 
 typedef struct DetectBytejumpData_ {
     uint8_t nbytes;                   /**< Number of bytes to compare */
+    uint8_t nbytes_idx;               /**< If nbytes is a string, idx of var */
     uint8_t base;                     /**< String value base (oct|dec|hex) */
     uint16_t flags;                   /**< Flags (big|little|relative|string) */
     uint32_t multiplier;              /**< Multiplier for nbytes (multiplier n)*/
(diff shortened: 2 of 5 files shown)


suricata-qa:

Pipeline 14705

doc/userguide/rules/payload-keywords.rst
```diff
@@ -516,14 +516,15 @@ The ``byte_jump`` keyword allows for the ability to select a ``<num of bytes>``
 
 Format::
 
-  byte_jump:<num of bytes>, <offset> [, relative][, multiplier <mult_value>] \
+  byte_jump:<num of bytes> | <variable-name>, <offset> [, relative][, multiplier <mult_value>] \
```
victorjulien, line 519: would it make sense to extend this to others too? multiplier, post_offset, bitmask?

src/detect-bytejump.h
```diff
@@ -40,9 +40,11 @@
 #define DETECT_BYTEJUMP_DCE       BIT_U16(6) /**< "dce" enabled */
 #define DETECT_BYTEJUMP_OFFSET_BE BIT_U16(7) /**< "byte extract" enabled */
 #define DETECT_BYTEJUMP_END       BIT_U16(8) /**< "from_end" jump */
+#define DETECT_BYTEJUMP_NBYTES_VAR BIT_U16(9) /**< nbytes string*/
 
 typedef struct DetectBytejumpData_ {
     uint8_t nbytes;                   /**< Number of bytes to compare */
+    uint8_t nbytes_idx;               /**< If nbytes is a string, idx of var */
```
victorjulien, line 47: Not sure if we can easily avoid it, but this grows the struct.

Old:

```c
struct DetectBytejumpData_ {
	uint8_t                    nbytes;               /*     0     1 */
	uint8_t                    base;                 /*     1     1 */
	uint16_t                   flags;                /*     2     2 */
	uint32_t                   multiplier;           /*     4     4 */
	int32_t                    offset;               /*     8     4 */
	int32_t                    post_offset;          /*    12     4 */

	/* size: 16, cachelines: 1, members: 6 */
	/* last cacheline: 16 bytes */
};
```

New:
```c
struct DetectBytejumpData_ {
	uint8_t                    nbytes;               /*     0     1 */
	uint8_t                    nbytes_idx;           /*     1     1 */
	uint8_t                    base;                 /*     2     1 */

	/* XXX 1 byte hole, try to pack */

	uint16_t                   flags;                /*     4     2 */

	/* XXX 2 bytes hole, try to pack */

	uint32_t                   multiplier;           /*     8     4 */
	int32_t                    offset;               /*    12     4 */
	int32_t                    post_offset;          /*    16     4 */

	/* size: 20, cachelines: 1, members: 7 */
	/* sum members: 17, holes: 2, sum holes: 3 */
	/* last cacheline: 20 bytes */
};
```

jlucovsky: There's a few things we could try:
- `nbytes` can't be more than `23` so perhaps it coult be a 5-bit value and the index 3-bits?
- `flags` could contain a check if `nbytes` is an index or not
- Perhaps `multiplier` could be smaller?

Thoughts @victorjulien ?

doc/userguide/rules/payload-keywords.rst
jlucovsky: @jmtaylor90  Thoughts on whether variables would be useful elsewhere?

`bitmask` has odd semantics so i don't think we'd want to include that

jmtaylor90: Talking with the team a bit, it would be nice to have variable use available wherever possible.

jlucovsky: I suggest we extend variable support to other elements in a different set of tickets for each of the `byte_*` keywords)

jmtaylor90: Definitely fine by me, don't want to unnecessarily scope creep on you :)

jlucovsky:
Continued in #9128

src/detect-bytejump.h
jlucovsky: I removed `nbytes_idx` and decreased the size of `multiplier` since it's value range is [1, 65535]

PR #9040, opened 2023-06-17
Filemagic/v3

victorjulien:
Replace backend of `filemagic` with implement of `file.magic`, which as a side effect fixes https://github.com/OISF/suricata-verify/pull/1253

SV_BRANCH=pr/1253

@jlucovsky  could you pick this up? I would like us to do the same for `filename` and `fileext`:
- `filename:"abc";` -> `file.name; content:"abc";`
- `fileext:pdf;` -> `file.name; content:".pdf"; endswith;` (note the added dot)

Another thing I didn't do yet is look at if negation still works. So we need more SV tests.

### DIFF
--- a/src/detect-tls.h
+++ b/src/detect-tls.h
@@ -38,8 +38,7 @@ typedef struct DetectTlsData_ {
     uint16_t ver; /** tls version to match */
     uint32_t flags; /** flags containing match variant (Negation for example) */
     char * subject; /** tls certificate subject substring to match */
-    char * issuerdn; /** tls certificate issuerDN substring to match */
-    char * fingerprint; /** tls fingerprint substring to match */
+    char *issuerdn; /** tls certificate issuerDN substring to match */
 } DetectTlsData;
 
 /* prototypes */
--- a/src/detect-content.c
+++ b/src/detect-content.c
@@ -327,10 +327,7 @@ void DetectContentPrint(DetectContentData *cd)
  */
 int DetectContentSetup(DetectEngineCtx *de_ctx, Signature *s, const char *contentstr)
 {
-    DetectContentData *cd = NULL;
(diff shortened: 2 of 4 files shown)


victorjulien:
This addresses https://redmine.openinfosecfoundation.org/issues/2836

src/detect-tls.h
```diff
@@ -38,8 +38,7 @@ typedef struct DetectTlsData_ {
     uint16_t ver; /** tls version to match */
     uint32_t flags; /** flags containing match variant (Negation for example) */
     char * subject; /** tls certificate subject substring to match */
-    char * issuerdn; /** tls certificate issuerDN substring to match */
```
victorjulien, line 41: @jeff wonder if we can do the same for subject/issuerdn as we did for fingerprint?

victorjulien:
Also needs a look into whether the match case sensitivity handling is still the same.

suricata-qa:

Pipeline 14719

src/detect-tls.h
inashivb: @jlucovsky I think you were meant to be tagged here ^^

catenacyber:
> Replace backend of filemagic with implement of file.magic

Why were they different in the first place ?

src/detect-tls.h
victorjulien: updated :)

victorjulien:
replaced by #9119

victorjulien:
> > Replace backend of filemagic with implement of file.magic
> 
> Why were they different in the first place ?

Original implementations were from before the sticky buffer push.

catenacyber:
So, the idea is to replace the backend, and make filemagic as deprecated ?

PR #9041, opened 2023-06-17
5912 rfb skip v2

satta:
Previous PR: #8999

Describe changes to previous PR:
- Move `set_event()` into `RFBTransaction`.
- Add some debug validation.
- Fix SID range for RFB event rules (2233000-2233999).
- Add comment to explain the global state situation and its impact when facing parsing errors.
- Return an error when not even the start of the transaction cannot be parsed correctly.
- Squash some cleanup commits.


rust/src/rfb/parser.rs
```diff
@@ -24,7 +24,7 @@ use nom7::*;
 use std::fmt;
 use std::str;
 
-#[derive(Debug,PartialEq)]
+#[derive(Debug,PartialEq, Eq)]
```
catenacyber: Could the commit message explain why we need this ?
Or squash it in another commit

rust/src/rfb/rfb.rs
```diff
@@ -122,6 +129,10 @@ impl RFBState {
         }
     }
 
+    fn set_event(tx: &mut RFBTransaction, event: RFBEvent) {
```
catenacyber: Could this be directly implemented in tx and not in state ?

rust/src/rfb/rfb.rs
```diff
@@ -206,8 +206,6 @@ impl RFBState {
 
                             if let Some(current_transaction) = self.get_current_tx() {
                                 current_transaction.ts_client_protocol_version = Some(request);
-                            } else {
-                                return AppLayerResult::err();
```
catenacyber, line 198: When could this have happened ? (should it be a debug validation)

rust/src/rfb/rfb.rs
```diff
@@ -320,20 +342,31 @@ impl RFBState {
...
-                    current = b"";
+                    // We have gotten out of sync with the expected state flow.
+                    // This could happen since we use a global state (i.e. that
+                    // is used for both directions), but if traffic can not be
+                    // parsed as expected elsewhere, we might not have advanced
+                    // a state for one direction but received data in the
+                    // "unexpected" direction, causing the parser to end up
+                    // here. Let's stop trying to parse the traffic but still
+                    // accept it.
+                    SCLogDebug!("Invalid state for request: {}", self.state);
```
catenacyber, line 362: Should we set an event here ?

rules/rfb-events.rules
```diff
@@ -0,0 +1,11 @@
+# RFB app-layer event rules.
```
catenacyber, line 1: This should be squashed in the commit bringing in the events

catenacyber: It should also be added to Makefile.am cf `git grep dnp3-events.rules`

catenacyber (changes requested):
Thanks Sascha.

There are a few details.
Main point is the question of using RFB-specific logic, or make it generic so that it can be reused by other protocols

CI : green
Code : a few questions
Commits segmentation : needs a bit of squashing/reordering
Commit messages : missing the ticket reference in the main commit
Git ID set : ok
CLA : ok
Doc update : none needed
Redmine ticket : needed, right ?
Rustfmt : ok, but could be done as first commit ;-)
Tests : maybe the removed test should simply be modified
Dependencies added: none

jasonish:
> Main point is the question of using RFB-specific logic, or make it generic so that it can be reused by other protocols

It might, but for the scope of this fix, and ticket 5912, we probably want to keep here, as there are probably other considerations for making it global.

rust/src/rfb/rfb.rs
satta: But it is. See https://github.com/OISF/suricata/blob/c2ee17130da59cfbd09c416f8d401ffe2d7e58ac/rust/src/rfb/rfb.rs#L98

I think this is a comment on an outdated code base: 
image

rust/src/rfb/parser.rs
satta: Will squash. This was needed for a different place in the code. Can't remember why but I get the point -- if I can't then it might be pointless to have separately.

rust/src/rfb/rfb.rs
satta: I think it should. Will change all instances of this to a (much better suited) debug validation.

rust/src/rfb/rfb.rs
```diff
@@ -239,15 +238,28 @@ impl RFBState {
...
+                                        RFBState::set_event(
+                                            current_transaction,
+                                            RFBEvent::UnimplementedSecurityType,
+                                        );
+                                    }
+                                    // We have just have seen a security type we don't know about.
+                                    // This is not bad per se, it might just mean this is a
+                                    // proprietary one not in the spec.
+                                    // Continue the flow but stop trying to map the protocol.
+                                    self.state = parser::RFBGlobalState::Skip;
```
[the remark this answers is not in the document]
satta: I would welcome that!

rules/rfb-events.rules
satta: Done

rust/src/rfb/rfb.rs
satta: Done.

rust/src/rfb/rfb.rs
satta: Can't hurt as it's indicative of some issue with basic protocol expectations.  Done.

rust/src/rfb/rfb.rs
```diff
@@ -758,31 +850,6 @@ mod test {
     use super::*;
     use crate::core::STREAM_START;
 
-    #[test]
-    fn test_error_state() {
```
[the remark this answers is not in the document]
satta: I can only imagine this was meant to test the case of invalid data trying to be parsed, ending up in an error state. We can still replicate this case: the only remaining case of this is the first communication (the server announcing its supported protocols) not being correctly parsed -- since the buffer is in no way valid RFB.
I reintroduced this test into the code, and made sure the `parse_response()` is being called, since the server speaks first in this protocol. The test now passes as expected.

satta:
Next PR: #9064

rust/src/rfb/rfb.rs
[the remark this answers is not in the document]
catenacyber: I wonder more about what becomes of this precise input.

What is the result with this input and the new code ?
A `skip` state and an event ?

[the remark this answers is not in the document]
satta: Yes. I assume this will be one of these cases in which the state machine ends up in an unexpected state.

PR #9044, opened 2023-06-20
Ci rustfmt v1

catenacyber:
None : QA

Describe changes:
- GitHub action to check that rustfmt-ed files are still rustfmt-ed
- rustfmt some more files which have a small diff


.github/workflows/formatting.yml
```diff
@@ -126,6 +126,7 @@ jobs:
           fi
         shell: bash {0}
       - run: git clone https://github.com/OISF/libhtp -b 0.5.x
+      - run: ./qa/rustfmt.sh
```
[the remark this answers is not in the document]
catenacyber: This is both formatting and rust, so might fit in both ?

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3836

qa/rustfmt.sh
```diff
@@ -0,0 +1,10 @@
+r=0
```
jasonish, line 1: Lets please make this a proper script.. Start with `#! /bin/sh`, little comment what it does.

qa/rustfmt.sh
```diff
@@ -0,0 +1,10 @@
...
+cat qa/rustfmt.txt | while read i; do
+    rustfmt $i;
```
jasonish, line 3: I'd just call `cargo fmt` then check each file.

qa/rustfmt.txt
```diff
@@ -0,0 +1,57 @@
+rust/derive/src/applayerframetype.rs
```
jasonish, line 1: Maybe build this list within the context of the `rust/` directory, as in where its valid to run `cargo fmt`.

.github/workflows/formatting.yml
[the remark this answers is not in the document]
jasonish: Fair enough, keep it in formatting.

jasonish (changes requested):
Good start.  While I'd rather just format `master` and `master-6.0.x` in one go at the same time, this is probably more realistic approach to get to whats expected of Rust projects today.

qa/rustfmt.sh
catenacyber: ok

victorjulien:
replaced by #9072

PR #9022, opened 2023-06-13
Backports 6.0.13/s/v1

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) tickets:
- https://redmine.openinfosecfoundation.org/issues/6138
- https://redmine.openinfosecfoundation.org/issues/6139 minus the commit https://github.com/OISF/suricata/pull/8965/commits/abc76e27de7184d6f64040fb53d7d0a4d2b3df17 because it seems that ANDX support was added as a part of https://github.com/oISF/suricata/commit/2d1460622411552
which seems to have been targeted only as a feature for 7. ref: https://redmine.openinfosecfoundation.org/issues/3475#note-22

PR #9047, opened 2023-06-20
smb: fix wrong data offset when wct = 12 (backport6)

catenacyber:
https://redmine.openinfosecfoundation.org/issues/6139

Describe changes:
- Backport of 000eb91078d5ca88ee93006340d7e68f97ade4bc

https://github.com/OISF/suricata-verify/pull/1260
```
SV_BRANCH=pr/1260
```

### DIFF
--- a/rust/src/smb/smb1_records.rs
+++ b/rust/src/smb/smb1_records.rs
@@ -93,7 +93,7 @@ named!(pub parse_smb1_write_andx_request_record<Smb1WriteRequestRecord>,
         >>  _padding_evasion: cond!(data_offset > 36+2*(wct as u16), take!(data_offset - (36+2*(wct as u16))))
         >>  file_data: rest
         >> (Smb1WriteRequestRecord {
-                offset: if high_offset != None { ((high_offset.unwrap() as u64) << 32)|(offset as u64) } else { 0 },
+                offset: ((high_offset.unwrap_or(0) as u64) << 32) | offset as u64,
                 len: (((data_len_high as u32) << 16) as u32)|(data_len_low as u32),
                 fid,
                 data:file_data,


suricata-qa:
Information:
 

|--|--|--|--| 
| .capture.kernel_drops | 0 | 364 | 0.00 | 

Pipeline 14751

jasonish:
Ticket 6008 in master also has commit abc76e27de7184d6f64040fb53d7d0a4d2b3df17. Is that not required in the backport?

catenacyber:
This simple commit has big impact.

The other commit you mention is more complex and has less impact

jasonish:
Maybe a comment in the ticket as to why, otherwise it looks like an incomplete backport.

catenacyber:
> Maybe a comment in the ticket as to why, otherwise it looks like an incomplete backport.

This backport is indeed incomplete, but it is better than nothing.

In fact, the ticket should have been split in 2 different issues, each being fixed by a commit.

This is the impactful fix that seems to me important to merge...

This simple commit is complete enough to fix an important bug.

jasonish:
The initial comment where this was held back: https://github.com/OISF/suricata/pull/9022#issuecomment-1590953660

victorjulien:
> The initial comment where this was held back: [#9022 (comment)](https://github.com/OISF/suricata/pull/9022#issuecomment-1590953660)

This was because the patches in part were about fixing andx support, which is really only supported properly in 7.

catenacyber:
> This was because the patches in part were about fixing andx support

I would not say so

This one commit in the PR fixes the case where the offset where to write the data in the file transferred over SMB is described as u16 instead of u32
Previously, we always took 0, if this offset was a u16, but this commit fixes by taking its value.
This fixes analysis of legit traffic.

The other commit fixes more completely the cases if you add padding in your SMB pdu (especially between the SMB "header" and the file data to write)
This is (attacker-)crafted traffic.
This is not about AndX offset being non 0. (But we may have still missed an evasion for some cases with these AndX offset being non 0 associated with padding)

catenacyber:
Rebased in https://github.com/OISF/suricata/pull/9098

PR #9056, opened 2023-06-21
detect/multi-tenant: Eliminate crashes processing tenant config

jlucovsky:
This PR eliminates crashes during parallel loading of tenant configurations by serializing processing of the .config files
- reference.config
- threshold.config
- classification.config

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6047](https://redmine.openinfosecfoundation.org/issues/6047)

Describe changes:
- Serialize .config file loading
- Use SCCalloc over SCMalloc in detect-engine.c

### Provide values to any of the below to override the defaults.

To use a pull request use a branch name like `pr/N` where `N` is the pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

PR #9060, opened 2023-06-22
detect/multi-tenant: Eliminate crashes processing tenant config

jlucovsky:
Continuation of #9056 

This PR eliminates crashes during parallel loading of tenant configurations by serializing processing of the .config files
- reference.config
- threshold.config
- classification.config

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6047](https://redmine.openinfosecfoundation.org/issues/6047)

Describe changes:
- Formatting fixups
- Removal of commit that triggered "uninitialized variable" errors.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/detect-engine.c
+++ b/src/detect-engine.c
@@ -247,11 +247,11 @@ void DetectAppLayerInspectEngineRegister2(const char *name,
         direction = 1;
     }
 
-    DetectEngineAppInspectionEngine *new_engine = SCMalloc(sizeof(DetectEngineAppInspectionEngine));
+    DetectEngineAppInspectionEngine *new_engine =
+            SCCalloc(1, sizeof(DetectEngineAppInspectionEngine));
     if (unlikely(new_engine == NULL)) {
         exit(EXIT_FAILURE);
     }
-    memset(new_engine, 0, sizeof(*new_engine));
     new_engine->alproto = alproto;
     new_engine->dir = direction;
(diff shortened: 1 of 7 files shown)


victorjulien:
there are far more call sites for `pcre2_substring_get_bynumber`, including some at packet time (pcre keyword), so if `pcre2_substring_get_bynumber` truly isn't thread safe, we need a much bigger patch.

I'm missing a good explanation of the thread safety issue, what is the issue?

victorjulien (changes requested):
see comment

jlucovsky:
> 

The `static` regex structures within the reference, threshold, and classification file processing (static `pcred2_code *regex` and `static pcred2_match_data` for example) were being used concurrently.

This PR prevents concurrent access -- access was most often through `pcre2_substring_copy_bynumber` -- by the multiple loader threads used in MT.

catenacyber (changes requested):
As Victor said 

> If this is the case I would suggest a different solution: move these pcre2 data structures into DetectEngineCtx. Of these we have an instance per tenant.

By the way, I like the commits to use SCCalloc and make tenant_id always a uint32 :-)

catenacyber:
I think this was solved with https://github.com/OISF/suricata/pull/9243

PR #9038, opened 2023-06-16
Multipart mime 3487 v17

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3487

Describe changes:
- convert HTTP to use new rust mime parser
- convert SMTP to use new rust mime parser

Follows #8886 with rebase and some review comments taken into account

Still to do by me : 
- investigate the QA SMTP changes : @ct0br0 would you send me a pcap when QA is done for this PR ?
- convert previous unit tests to SV tests
Required, but I cannot do : 
- merge the fix for body_md5 cf https://github.com/OISF/suricata/pull/9037
- squash commits

```
SV_BRANCH=pr/1252
```
https://github.com/OISF/suricata-verify/pull/1252

PR #9061, opened 2023-06-22
smtp: avoid counting last eol in file

catenacyber:
https://redmine.openinfosecfoundation.org/issues/6023

Describe changes:
- smtp: avoid counting last eol in file (in case of plaintext and not base64)

https://github.com/OISF/suricata-verify/pull/1265
```
SV_BRANCH=pr/1265
```

### DIFF
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -1120,22 +1120,13 @@ static int ProcessDecodedDataChunk(const uint8_t *chunk, uint32_t len,
                     tok = GetLine(
                             remainPtr, len - (remainPtr - (uint8_t *)chunk), &remainPtr, &tokLen);
                     if (tok != remainPtr) {
-                        /* If last token found without CR/LF delimiter, then save
-                         * and reconstruct with next chunk
-                         */
-                        if (tok + tokLen - (uint8_t *)chunk == (int)len) {
-                            PrintChars(SC_LOG_DEBUG, "LAST CHUNK LINE - CUTOFF", tok, tokLen);
-                            SCLogDebug("\nCHUNK CUTOFF CHARS: %u delim %u\n", tokLen,
-                                    len - (uint32_t)(tok + tokLen - (uint8_t *)chunk));
-                        } else {
-                            /* Search line for URL */
-                            ret = FindUrlStrings(tok, tokLen, state);
-                            if (ret != MIME_DEC_OK) {
-                                SCLogDebug("Error: FindUrlStrings() function"
-                                           " failed: %d",
-                                        ret);
-                                break;
-                            }
+                        /* Search line for URL */
+                        ret = FindUrlStrings(tok, tokLen, state);
+                        if (ret != MIME_DEC_OK) {
+                            SCLogDebug("Error: FindUrlStrings() function"
+                                       " failed: %d",
+                                    ret);
+                            break;
                         }
                     }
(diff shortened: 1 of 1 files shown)


src/util-decode-mime.c
```diff
@@ -2660,14 +2666,17 @@ static int TestDataChunkCallback(const uint8_t *chunk, uint32_t len,
...
+        }
         PrintChars(SC_LOG_DEBUG, "CHUNK", chunk, len);
 
         /* Parse each line one by one */
         remainPtr = (uint8_t *) chunk;
         do {
             tok = GetLine(remainPtr, len - (remainPtr - (uint8_t *) chunk),
                     &remainPtr, &tokLen);
-            if (tok != NULL && tok != remainPtr) {
+            if (tok != NULL && tok != remainPtr && remainPtr[-1] == LF) {
```
jasonish, line 2679: Can't say I've ever used a negative array pointer, but it kinda makes sense here.

suricata-qa:

Pipeline 14788

src/util-decode-mime.c
catenacyber: It will make more sense to remove all this code with https://github.com/OISF/suricata/pull/9038 ;-)

inashivb: I didn't know they existed in C 🥲

catenacyber: As Mark Twain said : 

> They did not know it was impossible so they did it

victorjulien: Reading this code I don't (clearly) see how we can be sure `remainPtr[-1]` accesses valid memory here.

src/util-decode-mime.c
```diff
@@ -1609,6 +1600,16 @@ static int ProcessBodyLine(const uint8_t *buf, uint32_t len,
             remaining -= tobuf;
             offset += tobuf;
         }
+        if (ret == MIME_DEC_OK) {
+            ret = ProcessDecodedDataChunk(state->data_chunk, state->data_chunk_len, state);
```
victorjulien, line 1604: what does failure mean here for the memcpy below? Should we exit the function here?

catenacyber: I think that previous failure should not prevent to save the end of line into the buffer for future addition

Failure of `ProcessDecodedDataChunk` is unimportant (just debug log)

Failure due to `DATA_CHUNK_SIZE - state->data_chunk_len` should be unreachable

victorjulien: we do pass the error to the caller, right?

src/util-decode-mime.c
catenacyber: > Reading this code I don't (clearly) see how we can be sure remainPtr[-1] accesses valid memory here.

This is just testing code (this is not used by Suricata on live traffic, and not even compiled if you do not take the unit tests), and the tests pass :-p

(By the way, these tests are responsible for having a function pointer, hence the big vulnerability that got fixed around)

I guess I should get rid of the call to `GetLine` and count `\n` more basically...

src/util-decode-mime.c
catenacyber: And the callers just do debug logging about it, right ?

catenacyber:
Continued in https://github.com/OISF/suricata/pull/9101

PR #8954, opened 2023-06-02
6.0.x - filedata - inspect http earlier - v2

jasonish:
Previous PR: https://github.com/OISF/suricata/pull/8950

This PR is for discussion of issue 5868. The issue is that `file_data` postpones inspection until the response body is complete, but by then the file has been pruned to the point where there is no data to save.  Example rule:

```
alert http any any -> any any (msg:"PDF LOGO"; flow:established,to_client; file_data; content:"PDF"; filestore; classtype:policy-violation; sid:9000000; rev:1;)
```

This behavior is expected if matching data at the end of the file, but the data is at the beginning for this test.  Removing `file_data` results in the file being completely logged as well, as it skips this optimization.

Ideally, we'd keep this optimization except when filestore is enabled, but I don't see a non-intrusive way to get that detail from `PrefilterTxHTTPFiledata`.

Git master does not see this issue file pruning does not appear to be occurring: https://github.com/OISF/suricata/pull/8950#discussion_r1214520123

Ticket: https://redmine.openinfosecfoundation.org/issues/5868
SV_BRANCH=pr/1225

PR #9063, opened 2023-06-22
detect-file-data: always inspect http body - v2

jasonish:
Run inspection even if the HTTP response body is not done. This
optimization causes file_data inspection to occur after the file has
been pruned.

Ticket: https://redmine.openinfosecfoundation.org/issues/5868

Previous PR:
- https://github.com/OISF/suricata/pull/8954

Changes from last PR:
- PR is now against master as this is now an issue on master and 6.0.x
- Just remove the optimization, instead of abort'ing in it

SV_BRANCH=pr/1225

### DIFF
--- a/src/detect-file-data.c
+++ b/src/detect-file-data.c
@@ -238,7 +238,6 @@ static InspectionBuffer *HttpServerBodyGetDataCallback(DetectEngineThreadCtx *de
 
     htp_tx_t *tx = txv;
     HtpState *htp_state = f->alstate;
-    const uint8_t flags = flow_flags;
 
     HtpBody *body = GetResponseBody(tx);
     if (body == NULL) {
@@ -260,27 +259,11 @@ static InspectionBuffer *HttpServerBodyGetDataCallback(DetectEngineThreadCtx *de
     SCLogDebug("response.body_limit %u response_body.content_len_so_far %" PRIu64
                ", response.inspect_min_size %" PRIu32 ", EOF %s, progress > body? %s",
             htp_state->cfg->response.body_limit, body->content_len_so_far,
-            htp_state->cfg->response.inspect_min_size, flags & STREAM_EOF ? "true" : "false",
-            (AppLayerParserGetStateProgress(IPPROTO_TCP, ALPROTO_HTTP1, tx, flags) >
+            htp_state->cfg->response.inspect_min_size, flow_flags & STREAM_EOF ? "true" : "false",
+            (AppLayerParserGetStateProgress(IPPROTO_TCP, ALPROTO_HTTP1, tx, flow_flags) >
                     HTP_RESPONSE_BODY)
                     ? "true"
                     : "false");
-
-    if (!htp_state->cfg->http_body_inline) {
-        /* inspect the body if the transfer is complete or we have hit
-        * our body size limit */
-        if ((htp_state->cfg->response.body_limit == 0 ||
-                    body->content_len_so_far < htp_state->cfg->response.body_limit) &&
-                body->content_len_so_far < htp_state->cfg->response.inspect_min_size &&
-                !(AppLayerParserGetStateProgress(IPPROTO_TCP, ALPROTO_HTTP1, tx, flags) >
-                        HTP_RESPONSE_BODY) &&
-                !(flags & STREAM_EOF)) {
-            SCLogDebug("we still haven't seen the entire response body.  "
-                       "Let's defer body inspection till we see the "
-                       "entire body.");
-            return NULL;
-        }
-    }
 
     /* get the inspect buffer
      *


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | SURI_TLPW2_single_stats_chk | | |
| .uptime | 996 | 1480 | 148.59% |
| .uptime | 219 | 247 | 112.79% |
| .tcp.overlap | 1124925 | 1040224 | 92.47% | 

Pipeline 14779

victorjulien:
I'm missing an analysis of the issue.

victorjulien:
I've done some analysis and what I'm coming to is the following:

there is a fundamental mismatch between how file_data works for HTTP1 and file tracking, as file_data in this case actually inspects the HTTP body, and not the extracted files. For this reason there is a `File::flags` flag `FILE_USE_DETECT` that all implementations use, except HTTP1. We've been tracking the body vs file issue here https://redmine.openinfosecfoundation.org/issues/4141

As a result of this mismatch, the file is "pruned" (meaning buffer slides forward) before the file.data inspection happens. The problem is not that the file.data happens too late, nor is the logic in removed in this PR an "optimization".

Ideally we move to `File` based inspection, but if we can't do that in time (@jlucovsky what is the status of that?) we should make sure the file prune/slide logic doesn't run too early in case of HTTP1.

jlucovsky:
> I've done some analysis and what I'm coming to is the following:
> 
> there is a fundamental mismatch between how file_data works for HTTP1 and file tracking, as file_data in this case actually inspects the HTTP body, and not the extracted files. For this reason there is a `File::flags` flag `FILE_USE_DETECT` that all implementations use, except HTTP1. We've been tracking the body vs file issue here https://redmine.openinfosecfoundation.org/issues/4141
> 
> As a result of this mismatch, the file is "pruned" (meaning buffer slides forward) before the file.data inspection happens. The problem is not that the file.data happens too late, nor is the logic in removed in this PR an "optimization".
> 
> Ideally we move to `File` based inspection, but if we can't do that in time (@jlucovsky what is the status of that?) we should make sure the file prune/slide logic doesn't run too early in case of HTTP1.

Right now it's looking good *except one s-v test fails* (http-body-inspect). I posted an update to #8785 that includes performance measurements.

jasonish:
> I've done some analysis and what I'm coming to is the following:
> 
> there is a fundamental mismatch between how file_data works for HTTP1 and file tracking, as file_data in this case actually inspects the HTTP body, and not the extracted files. For this reason there is a `File::flags` flag `FILE_USE_DETECT` that all implementations use, except HTTP1. We've been tracking the body vs file issue here https://redmine.openinfosecfoundation.org/issues/4141
> 
> As a result of this mismatch, the file is "pruned" (meaning buffer slides forward) before the file.data inspection happens. The problem is not that the file.data happens too late, nor is the logic in removed in this PR an "optimization".
> 
> Ideally we move to `File` based inspection, but if we can't do that in time (@jlucovsky what is the status of that?) we should make sure the file prune/slide logic doesn't run too early in case of HTTP1.

Understood, so it needs to be fixed from another direction. Any pointers there?  I also wonder if we could get a comment on the special logic for HTTP in `detect-file-data.c`.  Its a little interesting that removing that code didn't break anything, at least not in our basic set of tests.

jasonish:
> > I've done some analysis and what I'm coming to is the following:
> > there is a fundamental mismatch between how file_data works for HTTP1 and file tracking, as file_data in this case actually inspects the HTTP body, and not the extracted files. For this reason there is a `File::flags` flag `FILE_USE_DETECT` that all implementations use, except HTTP1. We've been tracking the body vs file issue here https://redmine.openinfosecfoundation.org/issues/4141
> > As a result of this mismatch, the file is "pruned" (meaning buffer slides forward) before the file.data inspection happens. The problem is not that the file.data happens too late, nor is the logic in removed in this PR an "optimization".
> > Ideally we move to `File` based inspection, but if we can't do that in time (@jlucovsky what is the status of that?) we should make sure the file prune/slide logic doesn't run too early in case of HTTP1.
> 
> Understood, so it needs to be fixed from another direction. Any pointers there? I also wonder if we could get a comment on the special logic for HTTP in `detect-file-data.c`. Its a little interesting that removing that code didn't break anything, at least not in our basic set of tests.

Forcing

```
    ff->flags |= FILE_USE_DETECT;
```

when http1 seems to do the trick as well.

victorjulien:
> Forcing
> 
> ```
>     ff->flags |= FILE_USE_DETECT;
> ```
> 
> when http1 seems to do the trick as well.

This uses `File::content_inspected`. How is it updated in this case?

jasonish:
> > Forcing
> > ```
> >     ff->flags |= FILE_USE_DETECT;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when http1 seems to do the trick as well.
> 
> This uses `File::content_inspected`. How is it updated in this case?

This would apply different sliding logic in `FilePruneFile` and does fix the tests - both the 0 size file StreamBufferGrow. Setting `FILE_STORE` works as well.  Anyways, I'm missing a component, here so have requested pointers.

jasonish:
https://github.com/OISF/suricata/pull/9140

PR #9069, opened 2023-06-23
af-packet: Remove support for rollover

jlucovsky:
This MR removes support for AF-Packet rollover.

Rollover support is defined [here](https://www.man7.org/linux/man-pages/man7/packet.7.html).

If either `rollover: yes` or `cluster-type: cluster_rollover` is configured, a warning message will be displayed and the clustering used will be `cluster_flow`.

The warning message:
```
Warning: af-packet: enp6s0f0: cluster_rollover deprecated; using "cluster_flow" instead. See https://redmine.openinfosecfoundation.org/issues/6128
```
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6128](https://redmine.openinfosecfoundation.org/issues/6128)

Describe changes:
- Document deprecation in upgrade notes
- Detect when rollover is used (either `rollover: yes` or `cluster-type: cluster_rollover`) and use `cluster_flow` instead

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -367,7 +367,17 @@ static void *ParseAFPConfig(const char *iface)
         SCLogDebug("Going to use cluster-id %" PRIu16, aconf->cluster_id);
     }
 
-    if (ConfGetChildValueWithDefault(if_root, if_default, "cluster-type", &tmpctype) != 1) {
+    int conf_val = 0;
+    bool rollover = false;
+    ConfGetChildValueBoolWithDefault(if_root, if_default, "rollover", &conf_val);
+    if (conf_val) {
+        SCLogWarning("%s: rollover deprecated; using \"cluster_flow\" instead. See "
+                     "https://redmine.openinfosecfoundation.org/issues/6128",
+                aconf->iface);
+        rollover = true;
+    }
+    if (rollover ||
+            ConfGetChildValueWithDefault(if_root, if_default, "cluster-type", &tmpctype) != 1) {
         /* default to our safest choice: flow hashing + defrag enabled */
         aconf->cluster_type = PACKET_FANOUT_HASH | PACKET_FANOUT_FLAG_DEFRAG;
         cluster_type = PACKET_FANOUT_HASH;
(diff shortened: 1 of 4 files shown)


suricata-qa:

Pipeline 14790

src/runmode-af-packet.c
```diff
@@ -367,19 +367,34 @@ static void *ParseAFPConfig(const char *iface)
...
+    int conf_val = 0;
+    bool rollover = false;
+    ConfGetChildValueBoolWithDefault(if_root, if_default, "rollover", &conf_val);
+    if (conf_val) {
+        SCLogWarning("%s: rollover deprecated; using \"cluster_flow\" instead. See "
+                     "https://redmine.openinfosecfoundation.org/issues/6128",
+                aconf->iface);
+        rollover = true;
+    }
+    if (rollover ||
```
victorjulien, line 379: this confuses rollover and cluster_rollover. If just rollover was set, the cluster setting shouldn't be modified I think

victorjulien (changes requested):
see inline comment

src/runmode-af-packet.c
jlucovsky: Specifying `rollover` or `cluster_rollover` results in the same cluster type flag -- `PACKET_FANOUT_ROLLOVER` to be used.

Do we want to remove `cluster_rollover` support but allow the other way of achieving the same result?

victorjulien: I see `rollover` leading to the flag `PACKET_FANOUT_FLAG_ROLLOVER`, while `cluster_rollover` has cluster_type `PACKET_FANOUT_ROLLOVER`. The former is a flag in addition to the cluster type, the latter sets the cluster type.

jlucovsky: Understood; I'll remove the changes for `rollover`

jlucovsky:
Continued in #9123

PR #9072, opened 2023-06-26
Ci rustfmt v2

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3836

Describe changes:
- GitHub action to check that rustfmt-ed files are still rustfmt-ed
- rustfmt some more files which have a small diff

#9044 with review taken into account 

@jasonish would you know if there is some Github CI way to have both master and PR branch tested : so that we can check rustfmt status of files between without keeping a list of known rustfmt-ed files ?

> Good start. While I'd rather just format `master` and `master-6.0.x` in one go at the same time, this is probably more realistic approach to get to whats expected of Rust projects today.

I am happy with the one-go approach but I guess the incremental one has more chances of getting merged...


suricata-qa:

Pipeline 14828

suricata-qa:
Information:
 
|--|--|--|--| 
| .capture.kernel_drops | 0 | 461 | 0.00 | 

Pipeline 14830

jasonish:
> @jasonish would you know if there is some Github CI way to have both master and PR branch tested : so that we can check rustfmt status of files between without keeping a list of known rustfmt-ed files ?

The `commits.yml` workflow does something like that..  But it'll just be easier if we can do it all at once.

qa/rustfmt.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/sh
+
+r=0
+# check the list of known rustfmt-ed files
+cat qa/rustfmt.txt | while read i; do
+    cargo fmt $i;
+    # the file used to be formatted and is not anymore
+    if [ $(git diff -- $i | wc -l) -gt 0 ]; then
+        git diff;
```
roligugus, line 9: This will print diffs of all files, not just the current file.

Probably meant `git diff -- $i`?

qa/rustfmt.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/sh
+
+r=0
+# check the list of known rustfmt-ed files
+cat qa/rustfmt.txt | while read i; do
+    cargo fmt $i;
+    # the file used to be formatted and is not anymore
+    if [ $(git diff -- $i | wc -l) -gt 0 ]; then
```
roligugus, line 8: Nothing wrong with this check, just FYI.

An alternative to check for differences would be `git diff --quiet -- $i` which makes `git diff` "exit with codes similar to `diff`". I.e. returns 1 if there are any differences, 0 otherwise. See `git help diff`

qa/rustfmt.sh
catenacyber: But `cargo fmt $i` only changed one file, right ?

qa/rustfmt.sh
catenacyber: Thanks, looks nicer.

What is the exact bash syntax for the test ?

roligugus: Different ways to skin that cat.

You might have seen something like:
```bash
git diff --quiet -- $i
if [ $? -ne 0 ] ; then
   # multiple commands ...
fi
```
I'd probably would do (if you don't need the exit value of the command):
```bash
if ! git diff --quiet -- $i; then
   # multiple commands ...
fi
```
Note the inversion with `! ...` as the `then` path is executed if the command returns 0. `diff`/`git diff --quiet` returns 1 (not 0 basically) if files are different.

Or you could short-circuit things
```bash
git diff --quiet -- $i || ( 
   # multiple commands ...
)
```

Or if you fancy one liners - not my personal preference if more than one command in failure path, but valid as well:
```bash
git diff --quiet -- $i || ( git diff -- $i; echo "$i needs formatting"; r=1 )
```

It's really your personal preference

qa/rustfmt.sh
roligugus: Hmm, looking at that at https://github.com/rust-lang/rustfmt#running

Seems `cargo fmt` does the whole crate/module. It might silently ignore the file?

You'd have to use `rustfmt` it seems with files. However, `rustfmt` might also check some included files if I got the description correct?! So that might check transiently included files multiple times? Ugh, that would open up "interesting behaviours"... Maybe worth trying it out?

If that is the case, it might be easier to run `cargo fmt` once and then `git diff` check for the list of files in `rustfmt.txt`?

catenacyber:
Should I close this, tracked by https://redmine.openinfosecfoundation.org/issues/3836 because the plan is to rustfmt all code when starting on 8 branch ?

qa/rustfmt.sh
catenacyber: Thanks Roland for the tips, waiting for a decision wether this is the right approach before tuning this

victorjulien:
> Should I close this, tracked by https://redmine.openinfosecfoundation.org/issues/3836 because the plan is to rustfmt all code when starting on 8 branch ?

Probably best, ya.

PR #9079, opened 2023-06-27
python: use portable python shebang

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3432

We no longer use distutils as stated in the bug but this shebang is anyway more portable.


inashivb:
Replaced w https://github.com/OISF/suricata/pull/9081

PR #9082, opened 2023-06-27
Python cleanups/v3

inashivb:
Previous PR: https://github.com/OISF/suricata/pull/9081

Changes since v2:
- update running instructions

For commit 120479a0a6bf996e2e389ea8047f469c31f0b2e3:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3432

We no longer use distutils as stated in the bug but this shebang is anyway more portable.


python/suricata/ctl/main.py
```diff
@@ -1,4 +1,5 @@
-# Copyright (C) 2018 Open Information Security Foundation
+#!/usr/bin/env python
+# Copyright (C) 2018-2023 Open Information Security Foundation
```
jasonish, line 2: As this file is never intended to be directly executed, I wonder if we should have the shebang here..

python/suricata/sc/suricatasc.py
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/python
-# Copyright(C) 2012-2020 Open Information Security Foundation
+#!/usr/bin/env python
+# Copyright(C) 2012-2023 Open Information Security Foundation
```
jasonish, line 2: Same here.

Looks like these are both executable in current install, which is probably a mistake as well.

victorjulien (changes requested):
Needs rebase.

python/suricata/ctl/main.py
inashivb: You're right. It doesn't make sense. Should remove it..

python/suricata/sc/suricatasc.py
inashivb: Sounds fair. I have removed the x bit in the install in the newer version of this PR, please check.

PR #9088, opened 2023-06-27
log-pcap: don't crash if directory unwritable and lz4 enabled - v1

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/5022

Also adds one-time error output as the error was hidden without lz4, and was now very noisy with lz4.

### DIFF
--- a/src/log-pcap.c
+++ b/src/log-pcap.c
@@ -429,14 +429,18 @@ static int PcapLogOpenHandles(PcapLogData *pl, const Packet *p)
 #ifdef HAVE_LIBLZ4
         else if (pl->compression.format == PCAP_LOG_COMPRESSION_FORMAT_LZ4) {
             PcapLogCompressionData *comp = &pl->compression;
-            if ((pl->pcap_dumper = pcap_dump_fopen(pl->pcap_dead_handle,
-                    comp->pcap_buf_wrapper)) == NULL) {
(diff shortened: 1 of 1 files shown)


victorjulien:
Is there a SV possible for this issue?

jasonish:
> Is there a SV possible for this issue?

Maybe. Will likely have to be disabled when running as root, which means when running in GitHub and GitLab CI as we have them setup now.  But appears we can do a test when running as a regular user by doing some pre-setup and creating a directory with with `000` or `444` permissions.  Possible, but more involved than the usual test.

jasonish:
Just some notes.. In a container, root has access to write to a directory with `000` modes.  While it would not have write access to an immutable directory (`chattr +i`), root in a container doesn't have enough privileges to create an immutable directory.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .app_layer.flow.rfb | 221 | 246 | 111.31% |
| .app_layer.tx.rfb | 227 | 246 | 108.37% |
| .app_layer.error.rfb.parser | 242 | 1 | 0.41% | 

Pipeline 14886

jasonish:
Our non-root build in OISF/suricata works with a test: https://github.com/jasonish/suricata/actions/runs/5404061042/jobs/9817895318#step:17:140

jasonish:
https://github.com/OISF/suricata/pull/9093

PR #9091, opened 2023-06-28
doc: quic in eve/schema

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6076

Describe changes:
- doc: quic in eve/schema

PR #9094, opened 2023-06-28
doc: quic in eve/schema

catenacyber:
https://redmine.openinfosecfoundation.org/issues/6076

Describe changes:
- doc: quic in eve/schema

#9091 with additional commit squashed in

### DIFF
--- a/etc/schema.json
+++ b/etc/schema.json
@@ -3031,15 +3031,18 @@
             "optional": true,
             "properties": {
                 "cyu": {
+                    "description": "ja3-like fingerprint",
                     "type": "array",
                     "minItems": 1,
                     "items": {
                         "type": "object",
                         "properties": {
                             "hash": {
+                                "description": "hex representation",
                                 "type": "string"
                             },
                             "string": {
+                                "description": "string representation",
                                 "type": "string"
                             }
                         },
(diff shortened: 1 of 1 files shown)


etc/schema.json
```diff
@@ -3031,34 +3031,41 @@
...
             "properties": {
                 "cyu": {
+                    "description": "ja3-like fingerprint",
                     "type": "array",
                     "minItems": 1,
                     "items": {
                         "type": "object",
                         "properties": {
                             "hash": {
+                                "description": "hex representation",
```
victorjulien, line 3041: looking at https://jasonish-suricata.readthedocs.io/en/evedoc-v2/output/eve/eve-json-format.html?highlight=quic#output-reference, I think it would be better to have something like `cyu hash hex representation`.

victorjulien: would also be good to mention this is limited to older QUIC variants (same for ua/sni)

victorjulien (changes requested):
inline comments

etc/schema.json
catenacyber: Same for UA, but SNI is also on standard version, extracted from the dedicated TLS client hello extension ;-)

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9097

PR #9105, opened 2023-06-29
Warn  integers downcast 64to32 6186 v1

catenacyber:
https://redmine.openinfosecfoundation.org/issues/6186

Describe changes:
- ci : add the `-Wshorten-64-to-32` flag to clang after `-Wimplicit-int-conversion`
- fix these warnings for all src/a*.c

```
LIBHTP_BRANCH=pr/396
```
https://github.com/OISF/libhtp/pull/396

Draft : is this the way to go ?
Make multiple PRs for this redmine ticket
I will remove the CI commit in the next version, but so, we have some red feedback to see what is still to be fixed

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -686,7 +686,7 @@ static uint32_t AppLayerProtoDetectProbingParserGetMask(AppProto alproto)
         FatalError("Unknown protocol detected - %u", alproto);
     }
 
-    SCReturnUInt(1UL << (uint32_t)alproto);
+    SCReturnUInt(1 << (uint32_t)alproto);
 }
 
 static AppLayerProtoDetectProbingParserElement *AppLayerProtoDetectProbingParserElementAlloc(void)
--- a/src/app-layer-frames.h
+++ b/src/app-layer-frames.h
@@ -60,7 +60,7 @@ typedef struct Frame {
 typedef struct Frames {
     uint16_t cnt;
     uint16_t dyn_size;     /**< size in elements of `dframes` */
-    uint32_t left_edge_rel;
+    uint64_t left_edge_rel;
     uint64_t base_id;
     Frame sframes[FRAMES_STATIC_CNT]; /**< static frames */
     Frame *dframes;                   /**< dynamically allocated space for more frames */
--- a/src/util-time.h
+++ b/src/util-time.h
@@ -53,8 +53,8 @@ typedef struct {
     {                                                                                              \
         .secs = 0, .usecs = 0                                                                      \
     }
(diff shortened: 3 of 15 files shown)


src/app-layer-detect-proto.c
```diff
@@ -686,7 +686,7 @@ static uint32_t AppLayerProtoDetectProbingParserGetMask(AppProto alproto)
         FatalError("Unknown protocol detected - %u", alproto);
     }
 
-    SCReturnUInt(1UL << (uint32_t)alproto);
+    SCReturnUInt(1 << (uint32_t)alproto);
```
catenacyber, line 689: Should `AppLayerProtoDetectProbingParserGetMask` return a u64 ?
But then flow structure needs to grow 2 u32 into 2 u64...

src/app-layer-frames.h
```diff
@@ -60,7 +60,7 @@ typedef struct Frame {
 typedef struct Frames {
     uint16_t cnt;
     uint16_t dyn_size;     /**< size in elements of `dframes` */
-    uint32_t left_edge_rel;
+    uint64_t left_edge_rel;
```
catenacyber, line 63: @victorjulien is this correct for frames ?

src/util-time.h
```diff
@@ -53,8 +53,8 @@ typedef struct {
     {                                                                                              \
         .secs = 0, .usecs = 0                                                                      \
     }
-#define SCTIME_USECS(t)          ((uint64_t)(t).usecs)
-#define SCTIME_SECS(t)           ((uint64_t)(t).secs)
```
catenacyber, line 57: @jlucovsky do you remember the reason for the cast ?

catenacyber:
Also draft to get CI feedback.
I think it is likely that these changes will break one build...

src/app-layer-frames.h
victorjulien: no it's u32 intentionally, it's relative to something, I think the stream base seq. So when used it's added to that other thing.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .app_layer.flow.rfb | 221 | 246 | 111.31% |
| .app_layer.tx.rfb | 227 | 246 | 108.37% |
| .app_layer.error.rfb.parser | 242 | 1 | 0.41% | 

Pipeline 14912

src/app-layer-detect-proto.c
catenacyber: Because we will soon have more than 32 app-layer protocols, right ?

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9117

PR #9116, opened 2023-06-30
doc/conf: remove deprecated fn call

inashivb:
https://redmine.openinfosecfoundation.org/issues/4371

The warning isn't there anymore but the fn call mentioned in the warning was deprecated long ago and still was used in our code. ref: https://github.com/sphinx-doc/sphinx/issues/7747

It has been deprecated since 1.8.0. I have 4.3.2 now.

### DIFF
--- a/doc/userguide/conf.py
+++ b/doc/userguide/conf.py
@@ -138,10 +138,7 @@ if not on_rtd:
     except:
         html_theme = 'default'
     def setup(app):
-        if hasattr(app, 'add_css_file'):
-            app.add_css_file('css/suricata.css')
-        else:
-            app.add_stylesheet('css/suricata.css')
+        app.add_css_file('css/suricata.css')
 else:
     html_context = {
         'css_files': [


inashivb:
Seems like AlmaLinux 8 still uses a version of sphinx-build that does not support the newer func. @catenacyber so reject this?

catenacyber:
As I understand, the warning was fixed with conditional use of add_css_file, but add_stylesheet is still needed for old versions.

So, we can close the PR and the ticket, right ?

inashivb:
> As I understand, the warning was fixed with conditional use of add_css_file, but add_stylesheet is still needed for old versions.
> 

Yes. You had done that. I was hoping since it was too old, I thought it could be removed but that's not the case.

> So, we can close the PR and the ticket, right ?

Yes. On it.

jasonish:
> Seems like AlmaLinux 8 still uses a version of sphinx-build that does not support the newer func. @catenacyber so reject this?

And our Debian 10 distribution builder for official releases uses an even older sphinx.

PR #9059, opened 2023-06-22
exception: master switch not applied to midstream - v1

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6169

Describe changes:
- When the midstream policy fix was merged, we ended up with a solution that would use the default (auto) value regardless of the master switch being defined or not. This led to Suri not applying the master switch config to the midstream. This was noticed when I retook the work with the master switch sv tests, which had not been merged yet.
- Renamed the function for choosing what 'auto' means, and added a brief comment to it, as I think a poor name and lack of documentation might have led to the situation just described, in the first place.

```
SV_BRANCH=pr/1264
```
https://github.com/OISF/suricata-verify/pull/1264

PR #9130, opened 2023-07-03
exception: master switch not applied to midstream - v2

jufajardini:
If an exception policy wasn't set up individually, use the GetDefault function to pick one. This will check for the master switch option and handle 'auto' cases.

Instead of deciding what the auto value should be when we are parsing the master switch, leave that for when some of the other policies is to be set via the master switch, when since this can change for specific exception policies - like for midstream, for instance.

Update exceptions policies documentation to clarify that the default configuration in IPS when midstream is enabled is `ignore`, not `drop-flow`.

Bug #6169Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6169

Previous PR: #9059 

Describe changes:
- If an exception policy wasn't set up individually, use the GetDefault function to pick one. This will check for the master switch option and handle 'auto' cases.
- Instead of deciding what the auto value should be when we are parsing the master switch, leave that for when some of the other policies is to be set via the master switch, when since this can change for specific exception policies - like for midstream, for instance.
- Update exceptions policies documentation to clarify that the default configuration in IPS when midstream is enabled is `ignore`, not `drop-flow`.
- rebase

```
SV_BRANCH=pr/1280
```
https://github.com/OISF/suricata-verify/pull/1280

### DIFF
--- a/doc/userguide/configuration/exception-policies.rst
+++ b/doc/userguide/configuration/exception-policies.rst
@@ -45,10 +45,13 @@ also defined in the yaml file.
 Auto
 ''''
 
-**In IPS mode**, the default behavior for all exception policies is to drop
-the flow, or the packet, when the flow action is not supported. It is possible
-to disable this default, by setting the exception policies' "master switch" yaml
-config option to ``ignore``.
+**In IPS mode**, the default behavior for most of the exception policies is to
+fail close. This means droping the flow, or the packet, when the flow action is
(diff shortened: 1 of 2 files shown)


suricata-qa:

Pipeline 14967

victorjulien (approved):

victorjulien:
Merged in #9143, thanks!

PR #9140, opened 2023-07-03
6.0.x: issue 5868: file_data fix for http1 response files - v4

jasonish:
Possible fix for: https://redmine.openinfosecfoundation.org/issues/5868

Previous PR: https://github.com/OISF/suricata/pull/9063

SV_BRANCH=https://github.com/OISF/suricata-verify/pull/1225

### DIFF
--- a/src/app-layer-htp-file.c
+++ b/src/app-layer-htp-file.c
@@ -103,6 +103,9 @@ int HTPFileOpen(HtpState *s, HtpTxUserData *tx, const uint8_t *filename, uint16_
 
         flags = FileFlowToFlags(s->f, STREAM_TOCLIENT);
 
+        /* Depend on the detection engine for file pruning. */
+        flags |= FILE_USE_DETECT;
+
         if ((s->flags & HTP_FLAG_STORE_FILES_TS) ||
                 ((s->flags & HTP_FLAG_STORE_FILES_TX_TS) && txid == s->store_tx_id)) {
             flags |= FILE_STORE;


victorjulien (changes requested):
This is not the correct approach.

Can't just set this flag and expect things to work.

victorjulien:
Also still missing a full analysis of the problem.

jasonish:
My analysis hasn't really changed. Given a rule like:
```
alert http any any -> any any (msg:"PDF LOGO"; flow:established,to_client; file_data; content:"PDF"; filestore; classtype:policy-violation; sid:9000000; rev:1;)
```
the file is stored, but 0 size, due to the pruning logic at https://github.com/OISF/suricata/blob/master-6.0.x/src/util-file.c#L382. Applying the pruning logic when `FILE_USE_DETECT` is set (https://github.com/OISF/suricata/blob/master-6.0.x/src/util-file.c#L358), does fix this as shown by the S-V test.

No new information here. I've asked for help/pointers to take this further given it fixes that failing test case.

victorjulien:
Analysis needs to go into the commit and needs to go deeper than describing that something doesn't work. See https://github.com/OISF/suricata/pull/9063#issuecomment-1607371005 for a start.

On the flag, it will have side effects that need to be described (e.g. different mechanics used for tracking progress). The fact that all of these things are missing suggest to me that this is more of a hack than a properly researched solution.

PR #9141, opened 2023-07-03
github-ci: debian 12, centos stream - v1

jasonish:
- Add a Debian 12 builder
- Use Debian 12 as dist builder instead of AlmaLinux 8
  - Gives us newer Sphinx
  - Avoid RHEL fall-out
- Add CentOS Stream 8 and 9 builders


victorjulien (approved):

victorjulien:
Merged in #9143, thanks!

PR #9160, opened 2023-07-06
Base64 data issue/v8

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:  https://redmine.openinfosecfoundation.org/issues/6135

SV_BRANCH=https://github.com/OISF/suricata-verify/pull/1292

Previous PR: https://github.com/OISF/suricata/pull/9159

Changes since v7:
- Fix clang formatting

### DIFF
--- a/src/util-base64.c
+++ b/src/util-base64.c
@@ -107,9 +107,9 @@ Base64Ecode DecodeBase64(uint8_t *dest, uint32_t dest_size, const uint8_t *src, 
         /* Get decimal representation */
         val = GetBase64Value(src[i]);
         if (val < 0) {
-            if ((mode == BASE64_MODE_RFC2045) && (src[i] == ' ')) {
+            if (mode == BASE64_MODE_RFC2045 && src[i] != '=') {
                 if (bbidx == 0) {
-                    /* Special case where last block of data has a leading space */
+                    /* Special case where last block of data has a leading space or invalid char */
                     leading_sp++;
                 }
                 sp++;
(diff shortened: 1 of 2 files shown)


src/util-base64.c
```diff
@@ -156,11 +156,26 @@ Base64Ecode DecodeBase64(uint8_t *dest, uint32_t dest_size, const uint8_t *src,
...
+            SCLogDebug("Destination buffer full");
+            ecode = BASE64_ECODE_BUF;
+            *consumed_bytes += leading_sp;
+            return ecode;
+        }
+        /* if the destination size is not at least 3 Bytes long, it'll give a dynamic
+         * buffer overflow while decoding, so, return and let the caller take care of the
+         * remaining bytes to be decoded which should always be < 4 at this stage */
+        if (dest_size - *decoded_bytes < 3)
+            return BASE64_ECODE_OK;
```
victorjulien, line 173: is this not an error?

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .flow.memuse | 592373632 | 519668040 | 87.73% | 

Pipeline 15021

victorjulien:
```
  CC       util-base64.o
util-base64.c:177:9: warning: Value stored to 'sp' is never read [deadcode.DeadStores]
        sp = 0;
        ^    ~
1 warning generated.
```
scan-build warning ^

victorjulien:
SV PR needs rebase?

src/util-base64.c
inashivb: My idea here was that if 2 bytes are permissible in dest_size, we return to the caller and let it take care of those two bytes in the next iteration/buffer like we do with the remaining bytes in MIME decoder. But, I certainly missed that this part is not called by MIME at all and the base64 decode keyword has no buffering mechanism.
So, yeah, this should probably be an error indeed.

PR #9166, opened 2023-07-06
doc/userguide: more eve http upgrade notes

jasonish:
Add more information with a examples of how the changes to EVE HTTP logging may affect users.

### DIFF
--- a/doc/userguide/upgrade.rst
+++ b/doc/userguide/upgrade.rst
@@ -65,9 +65,53 @@ Logging changes
 - Protocol values and their names are built into Suricata instead of using the system's ``/etc/protocols`` file. Some names and casing may have changed
   in the values ``proto`` in ``eve.json`` log entries and other logs containing protocol names and values.
   See https://redmine.openinfosecfoundation.org/issues/4267 for more information.
-- Custom logging of HTTP headers via suricata.yaml ``outputs.eve-log.types.http.custom``
-  is now done in subobjects ``response_headers`` or ``request_headers`` (as for option ``dump-all-headers``)
-  instead of at the root of the ``http`` json object (to avoid some collisions).
+- Logging of additional HTTP headers configured through the EVE
+  ``http.custom`` option will now be logged in the ``request_headers``
+  and/or ``response_headers`` respectively instead of merged into the
+  existing ``http`` object. In Suricata 6.0, a configuration like::
+
+    http:
+      custom: [Server]
+
+  would result in a log entry like::
+
+    "http": {
+      "hostname": "suricata.io",
+      "http_method": "GET",
+      "protocol": "HTTP/1/1",
+      "server": "nginx",
+      ...
+    }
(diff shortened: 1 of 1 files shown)


doc/userguide/upgrade.rst
```diff
@@ -65,9 +65,53 @@ Logging changes
...
+  logged like::
+
+    "http": {
+      "hostname": "suricata.io",
+      "http_method": "GET",
+      "protocol": "HTTP/1/1",
+      "response_headers": [
+        { "name": "Server", "value": "nginx" }
+      ]
+    }
```
jasonish, line 107: To me this reinforces why we need to do this and I don't see a sane way to maintain compatibility.  Lets take an arbitrary header name "Foo" that could exist in the request and response. 

If we do:
```
"http": {
  "foo": "value",
}
```
was the server or response?

Further, what should we do about a header that appears multiple times.  We then end up with:
```
```
"http": {
  "foo": ["value"],
}
```

(Elastic does fine with those arrays)

This may lead to one wanting:
```json
"http": {
  "request_headers": {
    "foo": ["one", "two", "three"]
  }
}
```
but this conflicts with our existing `request_headers` and `response_headers` that are already arrays.  We also can't make this objects for `dump-all-headers` as Elastic doesn't like arbitrary field growth, or certain chars in the keys.

As I mentioned over in https://redmine.openinfosecfoundation.org/issues/6173, we could make sure the "normalized" header names allowed in `custom` all have distinct names that don't overwrite any existing field name.. We should also prefix them with `request_` and `response_`.

So for the `custom: [Server]` example I give, this would mean logging like:
```json
"http": {
  "request_server": "nginx",
  ...
}```
(could also be `"request_server": ["nginx"]` to handle multiple occurrences of a header.

This does feel like a bit of a hack, and will still break many reports etc. that will need to be updated for the new field names. But perhaps a little easier to adapt to than the array approach for 7.

catenacyber:
Is this about @regit 's ticket ?

doc/userguide/upgrade.rst
catenacyber: I think this is rather Suricata doing a `join(",")` on libhtp's headers table

jasonish: So duplicate headers would be handled differently between http 1 and 2?

catenacyber: Good question, I think http2 has the same `join(",")` normalization at some point

But that needs to be tested ;-)

catenacyber: So looks like detection works on normalized with `join(",")` but logging logs only the last value 😨

jasonish:
> Is this about @regit 's ticket ?

Yeah, but add more info just to make possible breakage more clear.

jasonish:
Replaced by https://github.com/OISF/suricata/pull/9176.

PR #9172, opened 2023-07-07
Base64 data issue/v9

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:  https://redmine.openinfosecfoundation.org/issues/6135

SV_BRANCH=https://github.com/OISF/suricata-verify/pull/1295

Previous PR: https://github.com/OISF/suricata/pull/9160

Changes since v8:
- Fix code scan error
- return errcode BUF in case of small buffer

PR #9178, opened 2023-07-07
base64 data issue/v10

inashivb:
- util/base64: add test for long string w RFC4648
- util/base64: skip any invalid char for RFC2045
- util/base64: fix padding bytes for trailing data
- util/base64: check for dest buf size in last block
- util/base64: check dest buf size to hold 3Bytes
- util/base64: update test macro to use user data
- util/base64: fix tests for RFC2045

SV_BRANCH=https://github.com/OISF/suricata-verify/pull/1295

Previous PR: #9172

Changes since v9:
- Remove unneeded use of sp and leading_sp in calculating consumed bytes for RFC 4648

### DIFF
--- a/src/util-base64.c
+++ b/src/util-base64.c
@@ -107,9 +107,9 @@ Base64Ecode DecodeBase64(uint8_t *dest, uint32_t dest_size, const uint8_t *src, 
         /* Get decimal representation */
         val = GetBase64Value(src[i]);
         if (val < 0) {
-            if ((mode == BASE64_MODE_RFC2045) && (src[i] == ' ')) {
+            if (mode == BASE64_MODE_RFC2045 && src[i] != '=') {
                 if (bbidx == 0) {
-                    /* Special case where last block of data has a leading space */
+                    /* Special case where last block of data has a leading space or invalid char */
                     leading_sp++;
                 }
                 sp++;
@@ -156,7 +156,7 @@ Base64Ecode DecodeBase64(uint8_t *dest, uint32_t dest_size, const uint8_t *src, 
         }
     }
 
-    if (!valid && mode == BASE64_MODE_RFC4648) {
+    if (bbidx > 0 && bbidx < 4 && ((!valid && mode == BASE64_MODE_RFC4648))) {
         padding = B64_BLOCK - bbidx;
         *decoded_bytes += ASCII_BLOCK - padding;
         DecodeBase64Block(dptr, b64);
@@ -304,9 +304,12 @@ static int B64TestVectorsRFC2045(void)
     const char *fin_str8 = "foobar";
 
     const char *src9 = "Zm$9vYm.Fy";
-    const char *fin_str9 = "";
+    const char *fin_str9 = "foobar";
 
-    TEST_RFC2045(src1, fin_str1, strlen(fin_str1), strlen(fin_str1), strlen(src1), BASE64_ECODE_OK);
+    const char *src10 = "Y21Wd2IzSjBaVzFoYVd4bWNtRjFaRUJoZEc4dVoyOTJMbUYxOmpqcHh4b3Rhb2w%5d";
+    const char *fin_str10 = "cmVwb3J0ZW1haWxmcmF1ZEBhdG8uZ292LmF1:jjpxxotaol9t";
+
+    TEST_RFC2045(src1, fin_str1, ASCII_BLOCK * 2, strlen(fin_str1), strlen(src1), BASE64_ECODE_OK);
     TEST_RFC2045(src2, fin_str2, ASCII_BLOCK * 2, strlen(fin_str2), strlen(src2), BASE64_ECODE_OK);
     TEST_RFC2045(src3, fin_str3, ASCII_BLOCK * 2, strlen(fin_str3), strlen(src3), BASE64_ECODE_OK);
     TEST_RFC2045(src4, fin_str4, ASCII_BLOCK * 2, strlen(fin_str4), strlen(src4), BASE64_ECODE_OK);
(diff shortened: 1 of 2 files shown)


victorjulien (approved):

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .flow.memuse | 592373632 | 517625640 | 87.38% | 

Pipeline 15059

victorjulien:
Merged in #9198, thanks!

src/util-base64.c
```diff
@@ -318,8 +319,8 @@ static int B64TestVectorsRFC2045(void)
     const char *src9 = "Zm$9vYm.Fy";
     const char *fin_str9 = "foobar";
 
-    const char *src10 = "Y21Wd2IzSjBaVzFoYVd4bWNtRjFaRUJoZEc4dVoyOTJMbUYxOmpqcHh4b3Rhb2w%5d";
-    const char *fin_str10 = "cmVwb3J0ZW1haWxmcmF1ZEBhdG8uZ292LmF1:jjpxxotaol9t";
```
catenacyber: What is this test change about Shivani ?

src/util-base64.c
```diff
@@ -196,7 +196,7 @@ Base64Ecode DecodeBase64(uint8_t *dest, uint32_t dest_size, const uint8_t *src,
     {                                                                                              \
         uint32_t consumed_bytes = 0, num_decoded = 0;                                              \
         uint8_t dst[dest_size];                                                                    \
-        Base64Ecode code = DecodeBase64(dst, strlen(fin_str), (const uint8_t *)src, strlen(src),   \
+        Base64Ecode code = DecodeBase64(dst, dest_size, (const uint8_t *)src, strlen(src),         \
```
catenacyber, line 199: This looks wrong to me, it should have remained `strlen(fin_str)` and `strlen(fin_str)` should be used in ` uint8_t dst[dest_size];  `

src/util-base64.c
```diff
@@ -165,6 +165,11 @@ Base64Ecode DecodeBase64(uint8_t *dest, uint32_t dest_size, const uint8_t *src,
             ecode = BASE64_ECODE_BUF;
             return ecode;
         }
+        /* if the destination size is not at least 3 Bytes long, it'll give a dynamic
+         * buffer overflow while decoding, so, return and let the caller take care of the
+         * remaining bytes to be decoded which should always be < 4 at this stage */
+        if (dest_size - *decoded_bytes < 3)
```
catenacyber, line 171: We can still put `numDecoded_blk` bytes in it

inashivb: Indeed. But, not having 3 decoded bytes would overflow in https://github.com/inashivb/suricata/blob/adc4e7e9056c23a1145cb343deb433dc23ae0513/src/util-base64.c#L77. If we want to keep the numDecoded_blk bytes, we should handle such cases/change the decoding algorithm.
From what I understand, we always only want to decode a full block of data (i.e. 4 bytes).
Are you saying that we should decode even the partial data?

src/util-base64.c
inashivb: Could you please tell why is it wrong? I changed it because the callers are responsible for sending any dest_size, so, the tests here would also be able to replicate the same..

src/util-base64.c
inashivb: No idea. I don't even see it in this PR. Not sure how you reached here to comment on it 🤔 
It is labelled "Outdated" for me..

src/util-base64.c
```diff
@@ -246,7 +246,7 @@ static int B64DecodeCompleteStringWSp(void)
 
     const char *src = "SGVs bG8 gV29y bGQ=";
     const char *fin_str = "Hello World";
-    TEST_RFC2045(src, fin_str, strlen(fin_str) + 1, strlen(fin_str), strlen(src), BASE64_ECODE_OK);
+    TEST_RFC2045(src, fin_str, strlen(fin_str) + 3, strlen(fin_str), strlen(src), BASE64_ECODE_OK);
```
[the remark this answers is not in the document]
catenacyber: > But, 1 is needed in order to have the destination in multiple of 3 else we'll overflow while decoding..

I think we need to avoid overflow even when the destination size is not a multiple of 3

src/util-base64.c
catenacyber: Maybe wrong is not the right term.

But we want `dest_size` to be equal to `strlen(fin_str)` and so we should not use a second argument/variable

src/util-base64.c
catenacyber: > Are you saying that we should decode even the partial data?

Yes, we should be able to decode 2 bytes of base64 data even if it does not make a multiple of 3

src/util-base64.c
```diff
@@ -157,7 +157,8 @@ Base64Ecode DecodeBase64(uint8_t *dest, uint32_t dest_size, const uint8_t *src,
     }
 
     if (bbidx > 0 && bbidx < 4 && ((!valid && mode == BASE64_MODE_RFC4648))) {
-        padding = B64_BLOCK - bbidx;
+        /* Decoded bytes for 1 or 2 base64 encoded bytes is 1 */
+        padding = bbidx > 1 ? B64_BLOCK - bbidx : 2;
```
catenacyber, line 161: This commit lacks unit tests.
I think that for `bb_idx == 1`, padding needs to be `B64_BLOCK - bbidx` = 3, and thus no bytes can be decoded (we have only 6 bits)

inashivb: > (we have only 6 bits)

agreed. But, should we not have at least one decoded byte by making use of the padding? If I try any algorithms, even 1 byte does get decoded to something..
Please let me know what you think..

src/util-base64.c
catenacyber: I saw commit by commit

inashivb: oh. thanks.
So, this change was to make a string  such that after skipping all invalid chars, it is a perfectly divisible by 4 string which can be decoded to ascii chars that i can write in the test.
Earlier the len was 66 with an invalid char making it 65 even if that invalid char was skipped so I got rid of one char to be able to have a perfectly decoded string 😅

src/util-base64.c
victorjulien: I think generally we are greedy, and we want to decode as much as possible, esp is other implementations exist that do this as well.

PR #9183, opened 2023-07-07
6.0.x: dev guide fix for newer versions of sphinx - v1

jasonish:
Newer Sphinx does not allow a language of none, set to "en" like we do
for the user guide.


victorjulien (approved):

victorjulien:
Merged in #9204, thanks!

PR #9206, opened 2023-07-11
Base64 fuzz fix/v2

inashivb:

Previous PR: #9205 

Changes since v1:
- update ticket numbers in commit messages
- remove unneeded condition check

Note: s-v test is still under works. Seeing if there could be any other issues diagnosed by CI

### DIFF
--- a/src/util-base64.c
+++ b/src/util-base64.c
@@ -64,6 +64,22 @@ static inline int GetBase64Value(uint8_t c)
 }
 
 /**
+ * \brief Checks if the given char in a byte array is Base64 alphabet
+ *
+ * \param src Pointer to the byte array
+ * \param idx Index where to look at
+ *
+ * \return True if the char was Base64 alphabet, False otherwise
+ */
+bool IsBase64Alphabet(const uint8_t *src, const uint32_t idx)
+{
+    if (GetBase64Value(src[idx]) < 0) {
+        return false;
+    }
+    return true;
+}
+
+/**
  * \brief Decodes a 4-byte base64-encoded block into a 3-byte ascii-encoded block
  *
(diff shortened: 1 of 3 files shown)


src/util-base64.c
```diff
@@ -63,6 +63,22 @@ static inline int GetBase64Value(uint8_t c)
...
 
+/**
+ * \brief Checks if the given char in a byte array is Base64 alphabet
+ *
+ * \param src Pointer to the byte array
+ * \param idx Index where to look at
+ *
+ * \return True if the char was Base64 alphabet, False otherwise
+ */
+bool IsBase64Alphabet(const uint8_t *src, const uint32_t idx)
```
victorjulien, line 74: why not pass in the char itself?

src/util-decode-mime.c
```diff
@@ -1184,15 +1184,15 @@ static uint32_t ProcessBase64Remainder(
...
+        if (IsBase64Alphabet(state->bvremain, i)) {
             block[cnt++] = state->bvremain[i];
         }
     }
 
     /* if we don't have 4 bytes see if we can fill it from `buf` */
     if (buf && len > 0 && cnt != B64_BLOCK) {
         for (uint32_t i = 0; i < len && cnt < B64_BLOCK; i++) {
-            if (buf[i] != ' ') {
+            if (IsBase64Alphabet(buf, i)) {
```
victorjulien, line 1195: then the check would be `if (IsBase64Alphabet(buf[i)]) {`

src/util-base64.c
inashivb: For some reason, thought this was the C way. 😭  nvm. Will do it in the next rev. Thank you!

PR #9208, opened 2023-07-11
Base64 fuzz fix/v3

inashivb:

Previous PR: #9206

Changes since v2:
- pass char instead of pointer to the array and the index for `IsBase64Alphabet`

Note: s-v test is still under works. Seeing if there could be any other issues diagnosed by CI

### DIFF
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -1184,7 +1184,7 @@ static uint32_t ProcessBase64Remainder(
 
     /* Strip spaces in remainder */
     for (uint8_t i = 0; i < state->bvr_len; i++) {
-        if (state->bvremain[i] != ' ') {
+        if (IsBase64Alphabet(state->bvremain[i])) {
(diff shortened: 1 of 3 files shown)


src/util-decode-mime.c
```diff
@@ -1265,21 +1265,39 @@ static uint32_t ProcessBase64Remainder(
...
+ * \param len The length of the line
+ * \param offset The offset where to start reading the data from
+ * \param state The current parser state
+ * */
+static inline void B64SkipUnneededCharsLastBlk(
+        const uint8_t *buf, const uint32_t buf_len, const uint32_t offset, MimeDecParseState *state)
+{
+    for (uint32_t i = offset; i < buf_len; i++) {
+        if (IsBase64Alphabet(buf[i])) {
+            state->bvremain[state->bvr_len++] = buf[i];
```
victorjulien, line 1284: I would sleep better if this loop would have a guard against overflowing `bvremain`.

PR #9177, opened 2023-07-07
doc: add support status page - v4

jasonish:
Previous PR: #9163 

Changes from last PR:
- Incorporate feedback.

Generated at https://jasonish-suricata.readthedocs.io/en/doc-support-status-v4/support-status.html.

PR #9209, opened 2023-07-11
Doc/severity/v6

victorjulien:


victorjulien:
Includes #9177

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+backported to older versions that are still getting updates. We will
+update the vulnerabilities page and note the issue CVE in the changelog
+and commit message, but they may not trigger new releases.
+
+
+## Support Status of affected code
+
+4 levels are defined: Tier 1, Tier 2, Community and Unmaintained.
+
+These are documented in https://docs.suricata.io/en/latest/support-status.html
```
catenacyber, line 45: This is currently 404

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Severity Levels
+
+We will determine the severity of each issue, taking into account our
+experience dealing with past issues, versions affected, common defaults,
+and use cases. We use the following severity categories:
+
+* **CRITICAL** Severity. This affects Tier 1 features that are enabled by default
+where the issue disrupts availability of the service, leading to severe
+loss of visibility and/or availability. RCE, remotely triggerable traffic
```
catenacyber, line 15: Should we remove RCE if we describe more precisely remotely triggerable traffic based crashes ?

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+* **MODERATE** Severity. This includes issues like crashes or evasion in Tier 2 and
+Community features that are not enabled by default. These will in general be
+kept private until the next release, and that release will be scheduled so
+that it can roll up several such flaws at one time.
+
+* **LOW** Severity. This includes issues such as those that only affect the
+Suricata command line utilities, or unlikely configurations. These will in
+general be fixed immediately in latest development versions, and may be
+backported to older versions that are still getting updates. We will
+update the vulnerabilities page and note the issue CVE in the changelog
```
catenacyber, line 37: What is the vulnerabilities page ?

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+where the issue disrupts availability of the service, leading to severe
+loss of visibility and/or availability. RCE, remotely triggerable traffic
+based crashes, or evasions with a wide scope are considered to be in-scope
+for this severity. These issues will be kept private and will trigger a new
+release of all supported versions. We will attempt to address these as soon
+as possible.
+
+* **HIGH** Severity. This includes issues that are of a lower risk than critical,
+perhaps due to being disabled by default Tier 1 or affecting Tier 2 and
+Community features, or which are less likely to be exploitable. These issues
```
catenacyber, line 23: For examples, trigger able through rules instead of network traffic.

Other question : What about memory leaks ? 
I understand that every impact is equal : evasion, crash from NULL Deref or use after free, memory leak, quadratic complexity. What matters is the feature where the bug resides. Do,I get it right ?

SECURITY.md
jufajardini: I think that that's because it's added with this PR.

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Reporting Issues
+
+For reporting security issues, please use `security@oisf.net`.
+
+If you report a security issue to us, please share as much detail about the issue
+as possible: pcaps, attack scripts, potential fixes, etc. If you share pcaps or
+other data, please clearly state if these can (eventually) enter our public CI/QA.
+
+We assign a severity and will share our assessment with you.
```
jufajardini, line 56: nit:
```suggestion
We will assign a severity and will share our assessment with you.
```

SECURITY.md
```diff
@@ -0,0 +1,67 @@
...
+aim would be no longer than a month where this is something under our control.
+
+* **MODERATE** Severity. This includes issues like crashes or evasion in Tier 2 and
+Community features that are not enabled by default. These will in general be
+kept private until the next release, and that release will be scheduled so
+that it can roll up several such flaws at one time.
+
+* **LOW** Severity. This includes issues such as those that only affect the
+Suricata command line utilities, or unlikely configurations. These will in
+general be fixed immediately in latest development versions, and may be
```
jufajardini, line 35: Could we:
```suggestion
general be fixed as soon as possible in latest development versions, and may be
```

SECURITY.md
jufajardini, line 37: Will there always be a CVE?

jufajardini (changes requested):
To complement Philippe's review, some inline comments.

SECURITY.md
victorjulien: I feel it's a separate class of issues that we can mention separately here.

SECURITY.md
victorjulien: removed this language

SECURITY.md
victorjulien: Mem leaks eventually would affect availability or visibility (if we don't free up memcap for example).

I think indeed the location of the issue is often the most important, but probably also things like: single packet issue? Complex interaction between evil client and evil server? Etc.

SECURITY.md
victorjulien: no, update the language a bit

SECURITY.md
catenacyber: Ok, so risk = impact * probability

Impact is equivalent for most classes of bugs.

And probability = tier of bug's location * complexity for the attacker

SECURITY.md
victorjulien: not sure, seems more specific than we need to be?

catenacyber: So, I do not understand what is meant here by "RCE".

Is arbitrary file write with datasets rule RCE to you ?

Anyways, I think this deserves something more precise than an acronym (and one that is notoriously vague)

victorjulien:
replaced by #9225

PR #9162, opened 2023-07-06
output/eve: add 'verdict' field to 'alert' and 'drop' events -  v8

jufajardini:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5464

Previous PR: https://github.com/OISF/suricata/pull/8596

Sharing this now to clarify one aspect, and check if this is going on the right direction.

Describe changes from previous PR:
- add verdict as a new output event
- use flags to set configuration options for `drop` event (haven't touched the `flows all/start` one, as this was leading to failures, but can make it a TODO
- drop idea of having `alert.action` as optional
- incorporate verdict format suggestions from previous PR

TODOs:

- add more examples in the documentation
- document the json schema
- document verdict configs for suricata.yaml (userguide)

### Provide values to any of the below to override the defaults.

```
SV_BRANCH=pr/1293
```
https://github.com/OISF/suricata-verify/pull/1293

Outputs examples:
```json
# ids mode, reject, verdict event:
{
  "timestamp": "2004-05-13T10:17:08.993643+0000",
  "flow_id": 1055224944964893,
  "pcap_cnt": 6,
  "event_type": "verdict",
  "src_ip": "65.208.228.223",
  "src_port": 80,
  "dest_ip": "145.254.160.237",
  "dest_port": 3372,
  "proto": "TCP",
  "pkt_src": "wire/pcap",
  "verdict": {
    "action": "alert",
    "reject-target": "source",
    "reject": "[tcp-reset, icmp-prohib, user defined]"
  }
}

# ids mode, reject, alert event (non-interesting portions removed):
{
  "timestamp": "2004-05-13T10:17:08.993643+0000",
  "flow_id": 1055224944964893,
  "pcap_cnt": 6,
  "event_type": "alert",
  "src_ip": "65.208.228.223",
  "src_port": 80,
  "dest_ip": "145.254.160.237",
  "dest_port": 3372,
  "proto": "TCP",
  "pkt_src": "wire/pcap",
  "alert": {
    "action": "blocked",
    "gid": 1,
    "signature_id": 1000001,
    "rev": 0,
    "signature": "",
    "category": "",
    "severity": 3
  },
...
  "verdict": {
    "action": "alert",
    "reject-target": "source",
    "reject": "[tcp-reset, icmp-prohib, user defined]"
  }
}
# ids mode, alert:
{
  "timestamp": "2004-05-13T10:17:08.783340+0000",
[...]

PR #9216, opened 2023-07-12
output/eve: add 'verdict' field to 'alert' and 'drop' events - v9

jufajardini:
https://redmine.openinfosecfoundation.org/issues/5464

Previous PR: https://github.com/OISF/suricata/pull/9162

Describe changes from previous PR:
- remove changes for adding verdict as new eve type
- change packet->action to PASS when a pass rule matches for a packet (this wasn't done anywhere, so `verdict.action: pass` never showed up
- take into account proto to decide on `verdict.reject`
- make `verdict.reject` an array instead of a string (json)
- incorporate other feedbacks from previous PR

```
SV_BRANCH=pr/1308
```
https://github.com/OISF/suricata-verify/pull/1308

### DIFF
--- a/src/detect-engine-alert.c
+++ b/src/detect-engine-alert.c
@@ -416,6 +416,7 @@ void PacketAlertFinalize(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx
 
             /* pass "alert" found, we're done */
             if (pa->action & ACTION_PASS) {
+                p->action |= ACTION_PASS;
                 break;
             }
             p->alerts.cnt++;
--- a/doc/userguide/output/eve/eve-json-format.rst
+++ b/doc/userguide/output/eve/eve-json-format.rst
@@ -89,22 +89,16 @@ generated the event.
 Event type: Alert
 -----------------
 
-Field action
-~~~~~~~~~~~~
+This field contains data about a signature that matched, such as
+``signature_id`` (``sid`` in the rule) and the ``signature`` (``msg`` in the
+rule).
 
-Possible values: "allowed" and "blocked"
-
-Example:
-
-::
-
-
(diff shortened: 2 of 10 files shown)


suricata-qa:

Pipeline 15124

src/detect-engine-alert.c
```diff
@@ -416,6 +416,7 @@ void PacketAlertFinalize(DetectEngineCtx *de_ctx, DetectEngineThreadCtx *det_ctx
 
             /* pass "alert" found, we're done */
             if (pa->action & ACTION_PASS) {
+                p->action |= ACTION_PASS;
```
victorjulien, line 419: hmm we had done quite a nice cleanup before, removing the flag from the `Packet::action`. Since it is only for logging, I wonder if we can keep it out of this field to make clear it's not used otherwise.

victorjulien:
That last "empty" alert to just log verdict feels wrong to me. Maybe this is where we'd use the verdict event type.

jufajardini:
Will update the SV tests that are currently failing with windows, I think I inadvertedly removed a 
```yaml
requires:
  features:
    - LIBNET1.1
```
That must be there when we use `reject`

src/detect-engine-alert.c
jufajardini: > hmm we had done quite a nice cleanup before, removing the flag from the `Packet::action`. Since it is only for logging, I wonder if we can keep it out of this field to make clear it's not used otherwise.

Would it make sense to have a field for the verdict itself? I can try to access the last packet alert in the queue, which should be the one with the pass rule... 🤔

jufajardini:
Trying a different approach for the empty alert: https://github.com/OISF/suricata/pull/9220

PR #9217, opened 2023-07-12
util/mime: skip over any invalid char

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/6207
https://redmine.openinfosecfoundation.org/issues/6135

Previous PR: #9215

Changes since v6:
- remove unneeded fn

SV_BRANCH=https://github.com/OISF/suricata-verify/pull/1307

PR #9219, opened 2023-07-12
Inashivb/base64 fuzz fix/v8

catenacyber:

Previous PR: #9217

Changes since v6:
- Accounts for `=` in the base64 alphabet
- Adds an optional optimization commit

SV_BRANCH=https://github.com/OISF/suricata-verify/pull/1307

### DIFF
--- a/src/util-base64.c
+++ b/src/util-base64.c
@@ -64,6 +64,21 @@ static inline int GetBase64Value(uint8_t c)
 }
 
 /**
+ * \brief Checks if the given char in a byte array is Base64 alphabet
+ *
+ * \param Char that needs to be checked
+ *
+ * \return True if the char was Base64 alphabet, False otherwise
+ */
+bool IsBase64Alphabet(uint8_t encoded_byte)
+{
+    if (GetBase64Value(encoded_byte) < 0 && encoded_byte != '=') {
+        return false;
+    }
+    return true;
+}
+
+/**
  * \brief Decodes a 4-byte base64-encoded block into a 3-byte ascii-encoded block
  *
  * \param ascii the 3-byte ascii output block
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -1184,9 +1184,8 @@ static uint32_t ProcessBase64Remainder(
 
     /* Strip spaces in remainder */
     for (uint8_t i = 0; i < state->bvr_len; i++) {
-        if (IsBase64Alphabet(state->bvremain[i])) {
-            block[cnt++] = state->bvremain[i];
-        }
+        DEBUG_VALIDATE_BUG_ON(!IsBase64Alphabet(state->bvremain[i]));
+        block[cnt++] = state->bvremain[i];
     }
 
     /* if we don't have 4 bytes see if we can fill it from `buf` */
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -1184,7 +1184,7 @@ static uint32_t ProcessBase64Remainder(
 
     /* Strip spaces in remainder */
     for (uint8_t i = 0; i < state->bvr_len; i++) {
-        if (state->bvremain[i] != ' ') {
+        if (IsBase64Alphabet(state->bvremain[i])) {
             block[cnt++] = state->bvremain[i];
         }
     }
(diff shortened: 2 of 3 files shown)


src/util-decode-mime.c
```diff
@@ -1184,9 +1184,8 @@ static uint32_t ProcessBase64Remainder(
 
     /* Strip spaces in remainder */
     for (uint8_t i = 0; i < state->bvr_len; i++) {
-        if (IsBase64Alphabet(state->bvremain[i])) {
-            block[cnt++] = state->bvremain[i];
-        }
+        DEBUG_VALIDATE_BUG_ON(!IsBase64Alphabet(state->bvremain[i]));
```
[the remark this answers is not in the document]
catenacyber: This is meant to be unreachable

Every character in `bvremain` is base64

src/util-base64.c
```diff
@@ -63,6 +63,21 @@ static inline int GetBase64Value(uint8_t c)
...
+/**
+ * \brief Checks if the given char in a byte array is Base64 alphabet
+ *
+ * \param Char that needs to be checked
+ *
+ * \return True if the char was Base64 alphabet, False otherwise
+ */
+bool IsBase64Alphabet(uint8_t encoded_byte)
+{
+    if (GetBase64Value(encoded_byte) < 0 && encoded_byte != '=') {
```
inashivb, line 75: Please check the following case:
Last blk of data that is being parsed in the remainder is:
`D=`,
with this patch, we do,
D -> Base64Alphabet? Yes -> add to the buffer
= -> Base64Alphabet? Yes -> add to the buffer

Now, the buffer becomes `D=`, we save this because it's not 4 bytes, we have next chunk of data come in, say, `BLAhblah` we add it the existing buffer, and, decode/process the remainder, which now is `D=BL`.

Is this correct?

With the earlier behavior, we'd be passing `DBLA` for decoding/processing the remainder.

src/util-decode-mime.c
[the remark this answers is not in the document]
inashivb: oh. Sorry. Thought this was the CopyRemainder check..

src/util-base64.c
catenacyber: This is meant for
```
NA=
=Mg
==
```
which should decode into 42 and not into `34 03 20`
as is done for
```
NA==
Mg==
```

What is the correct result for `D=BL` ?

For me
```
D=
BLAhblah
```
And
`D=BLAhblah ` should get decoded the same way

src/util-decode-mime.c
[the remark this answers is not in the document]
catenacyber: But we check before filling bvremain, right ?

src/util-base64.c
inashivb: > For me
> D=
> BLAhblah
>
> And
> D=BLAhblah should get decoded the same way

I thought we were establishing that padding bytes `=` should only happen to the right most to be dealt with correctly? Are we not?

src/util-decode-mime.c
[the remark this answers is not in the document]
inashivb: yes, indeed. Sorry. The buf check comes later

src/util-base64.c
catenacyber: I do not think we are

`DecodeBase64` can make `padding++` and then `bbidx++` for `D=BL`

But this is another bug/feature...

victorjulien: Can we have test cases for each of the discussed scenarios?

catenacyber: @inashivb how can we test ? Just have S-V tests with all these scenarios ?..
Looks like an overkill but unit tests running multiple times DecodeBase64 as mime does not look obvious to me as well...

Scenario being 
```
NA=
=Mg
==
```

inashivb: Looking at this again, I'm still unsure what to test against.
What I think:
Case 1: For a string like
`D=BLAHblah`, we should skip over `=` as it's not acting as a padding byte here but as an invalid char instead.
This would require fixing our Base64Decode legacy code.
Q: 
Case 2: For buf input like
```
NA=
=Mg
==
```
It is fair to expect it to be decoded like: `NA==` followed by `Mg==` as the padding bytes are where they should be.

What do you think the output should be for these cases? Will create patch and/or tests accordingly.

catenacyber: Case 1 `D=BLAHblah` is beyond the scope of this PR, to see after release
Behavior was there before and will be there after

```
NA=
=Mg
==
```

should be handled as `NA==Mg==` whatever the linefeeds

catenacyber:
Replaced by #9226

victorjulien:
qa pipeline 15151 shows some alert deviations and other warnings that should be evaluated when this work is picked back up.

PR #9237, opened 2023-07-14
File data/v3

victorjulien:
https://redmine.openinfosecfoundation.org/issues/5868
https://redmine.openinfosecfoundation.org/issues/4141
https://redmine.openinfosecfoundation.org/issues/4145

### DIFF
--- a/src/detect-filename.c
+++ b/src/detect-filename.c
@@ -28,8 +28,6 @@
 #include "decode.h"
 
 #include "detect.h"
-#include "detect-parse.h"
-#include "detect-content.h"
 
 #include "detect-engine.h"
 #include "detect-engine-mpm.h"
@@ -37,6 +35,10 @@
 #include "detect-engine-file.h"
 #include "detect-engine-prefilter.h"
 #include "detect-engine-content-inspection.h"
+
+#include "detect-parse.h"
+#include "detect-content.h"
+#include "detect-file-data.h"
 
 #include "flow.h"
 #include "flow-var.h"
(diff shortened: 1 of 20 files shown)


src/detect-http-server-body.c
```diff
@@ -39,7 +39,11 @@
 #include "detect-parse.h"
 #include "detect-engine.h"
 #include "detect-engine-mpm.h"
+#include "detect-engine-mpm.h"
```
jlucovsky, line 42: Most of these added includes aren't necessary.

victorjulien: Can you do a follow up PR?

jlucovsky: Will do.

src/detect-filename.c
```diff
@@ -99,45 +101,23 @@ void DetectFilenameRegister(void)
...
-    for (int i = 0; file_protos_tc[i].alproto != ALPROTO_UNKNOWN; i++) {
-        DetectAppLayerInspectEngineRegister2("files", file_protos_tc[i].alproto, SIG_FLAG_TOCLIENT,
-                file_protos_tc[i].progress, DetectFileInspectGeneric, NULL);
-    }
-    g_file_match_list_id = DetectBufferTypeGetByName("files");
+    DetectBufferTypeSetDescriptionByName("file.name", "http user agent");
 
-    for (int i = 0; file_protos_ts[i].alproto != ALPROTO_UNKNOWN; i++) {
-        if (file_protos_ts[i].alproto == ALPROTO_HTTP2) {
-            // no filename on HTTP2 files
```
catenacyber, line 115: @jlucovsky do you know the consequences of this removal ?

Now, HTTP2 filename buffer is registered (even if it will always be empty)

And detect-parse.c has still error on ` "protocol HTTP2 doesn't support file name matching"` for a rule HTTP2 only like `alert http2 any any -> any any (file.name; content:"/a/expl/2008.mp4"; startswith; endswith; sid:24; rev:1;)`

victorjulien: if this is a (new) bug then please create a ticket for it

catenacyber: Well, I thought at first, but I am not sure this is a bug, since there is the guard anyways in detect-parse.c

Do you have an opinion here ?

PR #9150, opened 2023-07-06
Src div0 5920 v3

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5920

Describe changes:
- src: checks to avoid divisions by zero

Modifies #9147 with needed rebase after one division fixed by @jlucovsky in #9148 with https://github.com/OISF/suricata/commit/cb30b7185f6af0299f2ceab6072aa9030f6f3a5c

PR #9253, opened 2023-07-19
Src div0 5920 v4

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5920

Describe changes:
- src: checks to avoid divisions by zero

Modifies #9150 with rebase now that Suricata 7 is released and this is targeted for 7.0.1

### DIFF
--- a/src/source-dpdk.c
+++ b/src/source-dpdk.c
@@ -152,12 +152,18 @@ static void DPDKFreeMbufArray(struct rte_mbuf **mbuf_array, uint16_t mbuf_cnt, u
 static uint64_t CyclesToMicroseconds(const uint64_t cycles)
 {
     const uint64_t ticks_per_us = rte_get_tsc_hz() / 1000000;
+    if (ticks_per_us == 0) {
+        return 0;
+    }
     return cycles / ticks_per_us;
 }
 
 static uint64_t CyclesToSeconds(const uint64_t cycles)
 {
     const uint64_t ticks_per_s = rte_get_tsc_hz();
+    if (ticks_per_s == 0) {
+        return 0;
+    }
     return cycles / ticks_per_s;
 }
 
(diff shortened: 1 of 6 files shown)


src/source-dpdk.c
```diff
@@ -152,12 +152,18 @@ static void DPDKFreeMbufArray(struct rte_mbuf **mbuf_array, uint16_t mbuf_cnt, u
 static uint64_t CyclesToMicroseconds(const uint64_t cycles)
 {
     const uint64_t ticks_per_us = rte_get_tsc_hz() / 1000000;
+    if (ticks_per_us == 0) {
```
victorjulien, line 155: @lukashino is this necessary? Or can rte_get_tsc_hz return < 1000000?

suricata-qa:
Information:
 

ERROR: QA failed on SURI_TLPR1_alerts_cmp.
|--|--|--|--| 
| .uptime | 1267 | 1409 | 111.21% |
| .tcp.active_sessions | 0 | 474 | - |
| .flow.active | 0 | 98434 | - |
| .flow.get_used | 0 | 98434 | - |
| .flow.get_used_eval | 0 | 11609865 | - |
| .flow.get_used_eval_reject | 0 | 6136 | - |
| .flow.get_used_eval_busy | 0 | 2415 | - |
| .flow.get_used_failed | 0 | 1604299 | - |
| .flow.emerg_mode_entered | 0 | 2 | - |
| .flow.emerg_mode_over | 0 | 1 | - |
| .app_layer.error.tls.parser | 23942 | 29751 | 124.26% | 

Pipeline 15326

src/source-dpdk.c
lukashino: Seems to be possible only when:
- the CPU would have a clock rate less than 100MHz
- the CPU would not be well supported by DPDK and some register would return 0/random number instead ticks per second

So, assuming every CPU has a clock rate of at least 800MHz, and we will not care about unsupported CPUs then it should be fine even without the check.

victorjulien (approved):

victorjulien:
Merged in #9288, thanks!

PR #9255, opened 2023-07-19
src: remove unused headers-exported functions

catenacyber:
None : generic cleaning
https://redmine.openinfosecfoundation.org/issues/4083

Describe changes:
- remove functions defined in a header and never used (debug, api-completeness, and others)
- remove double definition of IPPairLock

Python helper script to do this :
```
import subprocess
import sys
import os

def patchFuncStatic(funcname, cfile, hfile):
    subprocess.run(["sed", "-i", "-e", "/%s/d" % funcname, hfile], stdout=subprocess.PIPE, text=True)
    f = open(cfile, "r")
    f2 = open(cfile+".new", "w")
    writing = True
    for l in f.readlines():
        if funcname in l:
            writing = False
            continue
        if writing:
            f2.write(l)
        elif l.rstrip() == '}':
            writing = True
    f.close()
    f2.close()
    os.rename(cfile+".new", cfile)

def testFuncName(funcret, funcname, hfile):
    result = subprocess.run(["git", "grep", funcname], stdout=subprocess.PIPE, text=True)
    d = dict()
    nbh = 0
    for l in result.stdout.split('\n'):
        hasf = l.split(':')[0]
        if len(hasf) == 0:
            continue
        if hasf == hfile:
            nbh = nbh + 1
            continue
        d[hasf] = d.get(hasf, 0)+1
    if nbh == 1 and len(d) == 1 and list(d.values())[0] == 1:
        print(funcret, funcname, d)
        patchFuncStatic(funcname, list(d.keys())[0], hfile)

f = open(sys.argv[1], "r")
for l in f.readlines():
    if '(' in l:
        ps = l.split('(')[0].split()
        if len(ps) == 2:
            funcret = ps[0]
            funcname = ps[1]
            if funcret != '/*' and funcret != '#define' and funcret != '*':
                if funcname[0] == '*':
                    funcname=funcname[1:]
                testFuncName(funcret,funcname,sys.argv[1])
```

Rebase of #9099


suricata-qa:

Pipeline 15324

victorjulien:
@jasonish can you review this from the pov that we're now in a stable branch? Will it affect lib users potentially?

src/conf.c
```diff
@@ -581,25 +581,6 @@ int ConfValIsFalse(const char *val)
  * \retval 1 will be returned if the name is found and was properly
  * converted to a double, otherwise 0 will be returned.
  */
-int ConfGetDouble(const char *name, double *val)
```
jasonish, line 584: These non-static `Conf` functions are somewhat public API, as plugin users can make use of them. This one is likely not used, but maybe?

src/output.c
```diff
@@ -536,46 +498,6 @@ void OutputRegisterFiledataModule(LoggerId id, const char *name,
...
 
-/**
- * \brief Register a file data output sub-module.
- *
- * This function will register an output module so it can be
- * configured with the configuration file.
- *
- * \retval Returns 0 on success, -1 on failure.
- */
-void OutputRegisterFiledataSubModule(LoggerId id, const char *parent_name,
```
jasonish, line 547: I suppose some plugin author could potentially be using any output registration functions.

src/conf.c
catenacyber: Are you sure it works ?
It looks like my compiler optimized these unused public API functions and that plugins cannot find them at runtime...

src/output.c
catenacyber: What would be a file data submodule ?

jasonish: Oh, I see we weren't using this at all.  I thought we were just removing the public exposure.

PR #9276, opened 2023-07-21
dcerpc incomplete/v4

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5699

Previous PR: #9270

Changes since v3:
- fix pre existing issue: error out on invalid header, use diff retvals (found by fuzzer post incomplete API changes)


rust/src/dcerpc/dcerpc.rs
```diff
@@ -621,7 +605,7 @@ impl DCERPCState {
                         header.rpc_vers,
                         header.rpc_vers_minor
                     );
-                    return -1;
+                    return -3;
```
jlucovsky, line 608: Please add `-3` to the return values listed in the function header comments.

rust/src/dcerpc/dcerpc.rs
```diff
@@ -1878,216 +1842,6 @@ mod tests {
...
-            0x62, 0x9f, 0x3d, 0x8e, 0x47, 0x73, 0x02, 0x00, 0x02, 0x00, 0x04, 0x5d, 0x88, 0x8a,
-            0xeb, 0x1c, 0xc9, 0x11, 0x9f, 0xe8, 0x08, 0x00, 0x2b, 0x10, 0x48, 0x60, 0x02, 0x00,
-            0x00, 0x00, 0x0f, 0x00, 0x01, 0x00, 0xc8, 0x4f, 0x32, 0x4b, 0x70, 0x16, 0xd3, 0x01,
-            0x12, 0x78, 0x5a, 0x47, 0xbf, 0x6e, 0xe1, 0x88, 0x03, 0x00, 0x00, 0x00, 0x04, 0x5d,
-            0x88, 0x8a, 0xeb, 0x1c, 0xc9, 0x11, 0x9f, 0xe8, 0x08, 0x00, 0x2b, 0x10, 0x48, 0x60,
-            0x02, 0x00, 0x00, 0x00,
-        ];
-        let mut dcerpc_state = DCERPCState::new();
-        assert_eq!(
-            AppLayerResult::ok(),
```
catenacyber, line 1942: Why are these tests removed instead of changed ?
Now we expect `incomplete` instead of `ok` for the first buffer
and `ok` for both buffers concatenated, right ?

rust/src/dcerpc/dcerpc.rs
catenacyber: I would even rather get rid of `process_header` to match directly nom  ` Err(Err::Incomplete(_)) ` to our `AppLayerResult::incomplete`

catenacyber (changes requested):
I think tests need to be adapted instead of simply removed

rust/src/dcerpc/dcerpc.rs
inashivb: I removed them bc I thought that in case of incomplete, we may not have populated the different fields that we test after checking the parser status code.
Q1: How does incomplete make our test suite strong here?
Q2: Aren't such cases better as s-v tests? (which do exist btw)

rust/src/dcerpc/dcerpc.rs
inashivb: But, this is a case when the data is there but is not compliant w the dcerpc standards so likely not dcerpc. Do you think that the condition checks for values of the necessary struct items should also happen in the nom parsers?

catenacyber: That is not what I mean.

I mean remove the definition of `process_header` and use directly its code :

ie instead of `parsed = self.process_header(cur_i);`a few lines below have

```
match parser::parse_dcerpc_header(cur_i) {
Ok((leftover_bytes, header)) => {
                if header.rpc_vers != 5
                    || (header.rpc_vers_minor != 0 && header.rpc_vers_minor != 1)
                {
    return AppLayerResult::err();
}
Err(Err::Incomplete(_)) => {
    return AppLayerResult::incomplete(0, DCERPC_HDR_LEN as u32);
}
```

rust/src/dcerpc/dcerpc.rs
catenacyber: If these unit tests are now in S-V, this is fine :-)
Maybe this should be written in the commit message.

> How does incomplete make our test suite strong here?

We tested the parsing of these blobs and we should still test it either we decide to return Ok (and buffer it) or Incomplete

rust/src/dcerpc/dcerpc.rs
inashivb: oh. I see! Thank you!

inashivb:
This one is blocked bc:
- I need to make a minimal reproducer of the pcap that causes the massive stats deviations and,
- confirm/deny whether this PR is correct
Will update once I have something working on that front.

catenacyber:
Did you get a minimal pcap reproducer for stats deviation ?
Can I help somehow here ?

inashivb:
> Did you get a minimal pcap reproducer for stats deviation ?

No. I couldn't pull out the particular full `dcerpc` stream(s) that caused this. But, I haven't checked in a while now.

> Can I help somehow here ? 

Yes, if you could tell how to go about extracting such streams from very large pcaps?
The one I'm concerned with for the stats deviation in this PR is 190GB.

Please note this is what I had tried:
Filter out all the `dcerpc` traffic with `tshark` and run Suricata with midstream enabled on it. But, my extracted pkts were broken so I did something wrong.  Victor said we shouldn't hack it and extract full streams. I was unable to figure out the correct filter for that at the time.

Challenges I had faced:
- `tshark` tries to load entire pcap at once so memory was an issue at first
- Saving the pkts extracted w filter requires more than just saving the extracted pkts

rust/src/dcerpc/dcerpc.rs
```diff
@@ -979,40 +946,38 @@ impl DCERPCState {
...
-        if (buffer.len()) < fraglen as usize {
+        if (input_len + self.bytes_consumed as usize) < fraglen as usize {
             SCLogDebug!("Possibly fragmented data, waiting for more..");
-            self.extend_buffer(buffer, direction);
-            return AppLayerResult::ok();
+            return AppLayerResult::incomplete(parsed as u32, (fraglen - parsed as u16) as u32);
         } else {
             self.query_completed = true;
         }
-        parsed = self.bytes_consumed;
```
catenacyber, line 1002: Why this removal ?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -948,28 +931,12 @@ impl DCERPCState {
...
-                if self.buffer_tc.len() + input_len > 1024 * 1024 {
-                    SCLogDebug!("DCERPC TOCLIENT stream: Buffer Overflow");
-                    return AppLayerResult::err();
-                }
-                v = self.buffer_tc.split_off(0);
-                v.extend_from_slice(cur_i);
-                v.as_slice()
-            }
-        };
+        if cur_i.len() > 1024 * 1024 {
```
catenacyber, line 934: Why ?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -948,28 +931,12 @@ impl DCERPCState {
             self.data_needed_for_dir = direction;
         }
 
-        let buffer = match direction {
```
catenacyber, line 951: self.bytes_consumed needs to reset to 0

And needs to be out of the state structure to be just a regular variable with a scope limited to this function

rust/src/dcerpc/dcerpc.rs
```diff
@@ -979,40 +946,38 @@ impl DCERPCState {
...
                 return AppLayerResult::err();
             }
             self.bytes_consumed += parsed;
+            input_len -= parsed as usize;
         }
 
         let fraglen = self.get_hdr_fraglen().unwrap_or(0);
 
-        if (buffer.len()) < fraglen as usize {
+        if (input_len + self.bytes_consumed as usize) < fraglen as usize {
```
catenacyber, line 961: cur_i.len() ?

rust/src/dcerpc/dcerpc.rs
```diff
@@ -979,40 +946,38 @@ impl DCERPCState {
...
-            return AppLayerResult::ok();
+            return AppLayerResult::incomplete(parsed as u32, (fraglen - parsed as u16) as u32);
         } else {
             self.query_completed = true;
         }
-        parsed = self.bytes_consumed;
 
         let current_call_id = self.get_hdr_call_id().unwrap_or(0);
 
         match self.get_hdr_type() {
```
catenacyber, line 970: Where does the fragment reassembly take place ?

catenacyber:
> > Can I help somehow here ?
> 
> Yes, if you could tell how to go about extracting such streams from very large pcaps? The one I'm concerned with for the stats deviation in this PR is 190GB.
> 
> Please note this is what I had tried: Filter out all the `dcerpc` traffic with `tshark` and run Suricata with midstream enabled on it. But, my extracted pkts were broken so I did something wrong. Victor said we shouldn't hack it and extract full streams. I was unable to figure out the correct filter for that at the time.
> 

I agree with Victor.
Can you have some filter based on ports ?

I also added some comments inline.
The `bytes_consumed` field seems to need more changes

ct0br0:
Extracting a pcap with ports 135, 139 and 445 from TLPR1 is around 415MB and has 4140 `.app_layer.error.dcerpc_tcp.parser` but splitting by IP only has 13 from 192.168.0.1 which isn't what I'd expect. I will see if I can find anything that may help more.

inashivb:
> Extracting a pcap with ports 135, 139 and 445 from TLPR1 is around 415MB and has 4140 `.app_layer.error.dcerpc_tcp.parser` but splitting by IP only has 13 from 192.168.0.1 which isn't what I'd expect. I will see if I can find anything that may help more.

Thanks a lot! Actually, I'm also interested in the 415MB pcap that has 4k+ dcerpc errors.

catenacyber:
Closing this PR as stale, let me know if I can help here Shivani

rust/src/dcerpc/dcerpc.rs
inashivb: This is not new, I think this was just a good to have estimated limit for the buffer.
Q: How should one determine such a limit? RFC does not indicate how big the transactions could get, the bounds are just about the header sizes et al.

rust/src/dcerpc/dcerpc.rs
inashivb: Not sure. Adding it back again to see if it was a mistake...

rust/src/dcerpc/dcerpc.rs
inashivb: I think no because the input_len at this point in case header was parsed successfully is cur_i.len() - header size. wdyt?

rust/src/dcerpc/dcerpc.rs
inashivb: applayer parser? `AppLayerParserTriggerRawStreamReassembly` is called via the rust context later on..

rust/src/dcerpc/dcerpc.rs
inashivb: indeed it seems useless to keep it in the state.. thank you!

inashivb: I was wrong. `bytes_consumed` is needed to know how many bytes the Applayer API has consumed in any subsequent call. Keeping it in the State

PR #9293, opened 2023-07-26
eve/stats: log all stats, even those that are unscoped

jlucovsky:
Make sure that all stats are logged -- even those that are not scoped (missing the `.` -- e.g., `mempressure_max` vs `decoder.pkts`.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6094](https://redmine.openinfosecfoundation.org/issues/6094)

Describe changes:
- Precalculate the statistic's short-name during registration
- Use the short-name value to determine the scope.
- Add unscoped stats to the "global" record; scoped stats are handled as they were previously.

pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/etc/schema.json
+++ b/etc/schema.json
@@ -3668,6 +3668,12 @@
                 "uptime": {
                     "type": "integer"
                 },
+                "memcap_pressure": {
+                    "type": "integer"
+                },
+                "memcap_pressure_max": {
+                    "type": "integer"
+                },
                 "app_layer": {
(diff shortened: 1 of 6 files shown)


jlucovsky:
Continued in #9294

PR #9296, opened 2023-07-27
af-packet: Error on invalid peer device setting

inashivb:
If the copy-iface was set to the same as the af-packet interface, the peer won't be set. This leads to segfault. Catch it and return properly.

### DIFF
--- a/src/source-af-packet.c
+++ b/src/source-af-packet.c
@@ -661,6 +661,10 @@ static void AFPWritePacket(Packet *p, int version)
     }
 
     /* Index of the network device */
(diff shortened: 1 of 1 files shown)


src/source-af-packet.c
```diff
@@ -661,6 +661,10 @@ static void AFPWritePacket(Packet *p, int version)
     }
 
     /* Index of the network device */
+    if (p->afp_v.peer == NULL) {
```
victorjulien, line 664: this shouldn't be a runtime check. We should fail to start up on this condition

victorjulien (changes requested):

PR #9308, opened 2023-07-31
af-packet: terminate on same iface and copyiface

inashivb:

Previous PR: https://github.com/OISF/suricata/pull/9307

Changes since v2:
- add void to function params

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -1024,6 +1024,10 @@ static TmEcode ParseInterfacesList(const int runmode, char *pcap_dev)
                 SCLogError("No interface found in config for af-packet");
                 SCReturnInt(TM_ECODE_FAILED);
             }
+            int retval = CheckAFPacketIPSDevs();
+            if (retval == -1) {
+                FatalError("af-packet IPS setting is incorrect");
+            }
         }
 #endif
 #ifdef HAVE_AF_XDP
--- a/src/util-device.c
+++ b/src/util-device.c
@@ -271,6 +271,36 @@ const char *LiveGetShortName(const char *dev)
     return live_dev->dev_short;
 }
 
+int CheckAFPacketIPSDevs(void)
+{
+    ConfNode *base = ConfGetNode("af-packet");
+    ConfNode *child;
+
+    if (base == NULL)
+        return 0;
+
+    TAILQ_FOREACH (child, &base->head, next) {
+        ConfNode *subchild;
+        const char *iface = NULL, *copyiface = NULL;
+        TAILQ_FOREACH (subchild, &child->head, next) {
+            if (!strcmp(subchild->name, "interface"))
+                iface = subchild->val;
+            if (!strcmp(subchild->name, "copy-iface"))
+                copyiface = subchild->val;
+            if (iface && copyiface && !strcmp(iface, copyiface)) {
+                SCLogError("af-packet interface and copy-iface cannot be the same");
+                return -1;
+            }
+            if (iface && copyiface) {
+                iface = NULL;
+                copyiface = NULL;
+            }
+        }
+    }
+
+    return 1;
+}
+
 int LiveBuildDeviceList(const char *runmode)
 {
     return LiveBuildDeviceListCustom(runmode, "interface");
--- a/src/util-device.h
+++ b/src/util-device.h
@@ -86,6 +86,7 @@ int LiveGetDeviceCount(void);
 const char *LiveGetDeviceName(int number);
 LiveDevice *LiveGetDevice(const char *dev);
 const char *LiveGetShortName(const char *dev);
+int CheckAFPacketIPSDevs(void);
 int LiveBuildDeviceList(const char *base);
 void LiveDeviceHasNoStats(void);
 int LiveDeviceListClean(void);


src/util-device.c
```diff
@@ -271,6 +271,36 @@ const char *LiveGetShortName(const char *dev)
...
+    ConfNode *base = ConfGetNode("af-packet");
+    ConfNode *child;
+
+    if (base == NULL)
+        return 0;
+
+    TAILQ_FOREACH (child, &base->head, next) {
+        ConfNode *subchild;
+        const char *iface = NULL, *copyiface = NULL;
+        TAILQ_FOREACH (subchild, &child->head, next) {
```
victorjulien, line 285: I don't really understand what we're looping here?

src/util-device.c
```diff
@@ -271,6 +271,36 @@ const char *LiveGetShortName(const char *dev)
     return live_dev->dev_short;
 }
 
+int CheckAFPacketIPSDevs(void)
```
victorjulien, line 274: why is this in util-device?

src/suricata.c
```diff
@@ -1024,6 +1024,10 @@ static TmEcode ParseInterfacesList(const int runmode, char *pcap_dev)
                 SCLogError("No interface found in config for af-packet");
                 SCReturnInt(TM_ECODE_FAILED);
             }
+            int retval = CheckAFPacketIPSDevs();
```
victorjulien, line 1027: can this be checked in the af-packet files instead?

victorjulien (changes requested):
wouldn't it be enough to check in `ParseAFPConfig` if `strcmp(iface, out_iface) == 0` right after
```c
    if (ConfGetChildValueWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
        if (strlen(out_iface) > 0) {
            aconf->out_iface = out_iface;
        }
    }
```
?

src/util-device.c
inashivb: From what I understand, each of the entries in a list item like

```
 - interface: wlp0s20f3
    threads: auto 
    cluster-id: 99
    cluster-type: cluster_flow
    defrag: yes
    copy-mode: ips
    copy-iface: eth2   
```

src/util-device.c
inashivb: I thought that since we're looping for ifaces and that kinda logic seemed to already be here in this file so this might be the right place.

src/suricata.c
inashivb: Yes. As per your main suggestion since all of the other things done are invalid, it should all probably be there.

inashivb:
> wouldn't it be enough to check in `ParseAFPConfig` if `strcmp(iface, out_iface) == 0` right after
> 
> ```c
>     if (ConfGetChildValueWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
>         if (strlen(out_iface) > 0) {
>             aconf->out_iface = out_iface;
>         }
>     }
> ```
> 
> ?

Please note that this works but the program exits much later than the approach in this PR.
Following are the comparisons:
```
real	1m12.404s
user	0m0.007s
sys	0m0.014s
```

vs this PR:
```
real	0m1.719s
user	0m0.026s
sys	0m0.036s
```

Is this OK?

victorjulien:
Can you explain why this happens?

inashivb:
> Can you explain why this happens?

For the following conf
```
af-packet:
  - interface: enp10s0
    cluster-id: 99
    cluster-type: cluster_flow
    defrag: yes
    copy-mode: ips
    copy-iface: enp9s0

  - interface: enp9s0
    cluster-id: 98
    cluster-type: cluster_flow
    defrag: yes
    copy-mode: ips
    copy-iface: enp9s0
```

It seems to me that **with this PR**,
1. runmode is one of the first things to be determined by Suricata
2. the corresponding runmode configuration is loaded
I tried to add the check for the interface and copy-iface being correctly set as a part of (2). This made the program exit in the very initial stages.

Note: However, now, I'm thinking this is probably wrong bc we haven't established if it's the AF_PACKET IPS mode yet.

**With your approach**, I saw that a lot of things happen before we get to checking the configuration.
1. IPS mode is set
2. MTU of the ifaces is checked to be the same
3. unix socket is enabled
4. output modules are initialized
5. filestore is set up
6. detection engine is set up
  - rules are loaded from the files
  - signatures are grouped as per type
7. af-packet now activates the IPS mode in one direction (enp10s0->enp9s0)
8. Now, the other direction is checked and it is found that the copy-iface is incorrectly set

It seems to me that this is the correct approach indeed as this happens after we have established that we're dealing with the correct modes (AF_PACKET IPS in this case).
Q: Do you think it makes sense to optimize it and check the conf for correct ifaces right after (1) instead of after going over other stuff?

victorjulien:
Thanks for the analysis, I think it explains the issue well. I still think the place I suggested is the correct place to check this, however it does seem inefficient. I wonder if it makes sense to rework the initialization to do more of the config validation work earlier. Perhaps a 2 step approach, where parsing happens early, but actual start of the runtime late.

Anyway,  I think we can take the simple approach for now, and create a future work ticket for the optimizations.

PR #9324, opened 2023-08-02
Bug 5870/v6

inashivb:

Previous PR: https://github.com/OISF/suricata/pull/9322

Changes since v5:
- add null checks

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -291,6 +291,10 @@ static void *ParseAFPConfig(const char *iface)
             if (strlen(out_iface) > 0) {
                 aconf->out_iface = out_iface;
             }
+            if (strcmp(iface, out_iface) == 0) {
+                FatalError("Invalid config: interface (%s) and copy-iface (%s) can't be the same",
+                        iface, out_iface);
+            }
         } else {
             SCLogWarning("copy-iface corresponding to %s interface is %s", iface, out_iface);
         }
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -287,8 +287,12 @@ static void *ParseAFPConfig(const char *iface)
     }
 
     if (ConfGetChildValueWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
-        if (strlen(out_iface) > 0) {
-            aconf->out_iface = out_iface;
+        if (out_iface != NULL) {
+            if (strlen(out_iface) > 0) {
+                aconf->out_iface = out_iface;
+            }
+        } else {
+            SCLogWarning("copy-iface corresponding to %s interface is %s", iface, out_iface);
         }
     }
 


inashivb:
Cases tested:
Case I: Same interface and copy-iface
```
  - interface: eth2        
    threads: auto          
    cluster-id: 98         
    cluster-type: cluster_flow
    defrag: yes            
    copy-mode: ips         
    copy-iface: eth2
```

Case II: Different interface and copy-iface
```
  - interface: eth2        
    threads: auto          
    cluster-id: 98         
    cluster-type: cluster_flow
    defrag: yes            
    copy-mode: ips         
    copy-iface: eth0
```

Case III: Missing copy-iface key
```
  - interface: eth2        
    threads: auto          
    cluster-id: 98         
    cluster-type: cluster_flow
    defrag: yes            
    copy-mode: ips       
```

Case IV: Missing value to copy-iface key (This made the existing code segfault)
```
  - interface: eth2        
    threads: auto          
    cluster-id: 98         
    cluster-type: cluster_flow
    defrag: yes            
    copy-mode: ips         
    copy-iface: 
```

src/runmode-af-packet.c
```diff
@@ -287,8 +287,16 @@ static void *ParseAFPConfig(const char *iface)
     }
 
     if (ConfGetChildValueWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
-        if (strlen(out_iface) > 0) {
-            aconf->out_iface = out_iface;
+        if (out_iface != NULL) {
```
victorjulien, line 290: think this isn't needed as it wasn't needed before?

inashivb: The following config segfaults on the strlen call if I don't do this
```
 - interface: eth2        
    threads: auto          
    cluster-id: 98         
    cluster-type: cluster_flow
    defrag: yes            
    copy-mode: ips         
    copy-iface: 
```
Should this be ignored?

victorjulien: Nope, good that you tested it. Wonder if we use this pattern in more places.

inashivb: I think it is easy to miss indeed so there may be more places. I think I can come up with a cocci script to check it. Will do so.

src/runmode-af-packet.c
```diff
@@ -287,8 +287,16 @@ static void *ParseAFPConfig(const char *iface)
...
+            if (strlen(out_iface) > 0) {
+                aconf->out_iface = out_iface;
+            }
+            if (strcmp(iface, out_iface) == 0) {
+                FatalError("Invalid config: interface (%s) and copy-iface (%s) can't be the same",
+                        iface, out_iface);
+            }
+        } else {
+            SCLogWarning("copy-iface corresponding to %s interface is %s", iface, out_iface);
```
victorjulien, line 299: using `out_iface` here makes little sense as we know it will be null.

kind of wonder if we should have `ConfGetChildValueWithDefault` check for an empty value instead?

suricata-qa:

Pipeline 15494

src/runmode-af-packet.c
inashivb: > using out_iface here makes little sense as we know it will be null.

yeah. It came out as a nice `(null)` in the message so I let it be.

> kind of wonder if we should have ConfGetChildValueWithDefault check for an empty value instead?

I did that in https://github.com/OISF/suricata/pull/9335, QA does not seem to have any issues, so, creating a new PR with this approach in mind.

PR #9330, opened 2023-08-02
doc: display version on main page of userguide - v1

jasonish:
When viewing the docs online at Readthedocs, or similar it might be immediately apparent what version of the documentation is being displayed. Display the version on the first line before the table of contents to make it clear.


jasonish:
Replaced by https://github.com/OISF/suricata/pull/9331.

PR #9331, opened 2023-08-02
doc: display version on main page of userguide - v2

jasonish:
When viewing the docs online at Readthedocs, or similar it might be immediately apparent what version of the documentation is being displayed. Display the version on the first line before the table of contents to make it clear.

Also, disable some GitHub CI workflows on documentation-only changes.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,10 @@
 name: CIFuzz
-on: [pull_request]
+
+on:
+  pull_request:
+    paths-ignore:
+      - "doc/**"
+
 permissions: read-all
 jobs:
  Fuzzing:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -3,9 +3,13 @@ name: "CodeQL"
 on:
   push:
     branches: [ master ]
+    paths-ignore:
+      - "doc/**"
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ master ]
+    paths-ignore:
+      - "doc/**"
   schedule:
     - cron: '18 21 * * 1'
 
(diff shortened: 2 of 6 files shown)


jufajardini:
Would it be possible to add the revision on that message that shows up on all pages, along with copyrights? 
image

Also thinking that maybe we should backport the devguide to 6. Or at least handle the non-existence of those document in six: https://docs.suricata.io/en/suricata-6.0.13/devguide/index.html

jasonish:
> Would it be possible to add the revision on that message that shows up on all pages, along with copyrights? image
> 
> Also thinking that maybe we should backport the devguide to 6. Or at least handle the non-existence of those document in six: https://docs.suricata.io/en/suricata-6.0.13/devguide/index.html

Not quite sure I follow... But:
- For the message I added, it doesn't look like Sphinx exposes a variable to get the git revision. When we see it in the footer on Readthedocs hosted docs, I think this is extra sugar added by them. You don't get this in the output when using the RTD theme locally.
- As the copyright is already at the bottom of each page, I'm not sure if it makes sense to add it to this line.

victorjulien (approved):

jufajardini:
> > Would it be possible to add the revision on that message that shows up on all pages, along with copyrights? image
> > Also thinking that maybe we should backport the devguide to 6. Or at least handle the non-existence of those document in six: https://docs.suricata.io/en/suricata-6.0.13/devguide/index.html
> 
> Not quite sure I follow... But:
> 
> * For the message I added, it doesn't look like Sphinx exposes a variable to get the git revision. When we see it in the footer on Readthedocs hosted docs, I think this is extra sugar added by them. You don't get this in the output when using the RTD theme locally.
> * As the copyright is already at the bottom of each page, I'm not sure if it makes sense to add it to this line.

Ah, ok, I was asking about having the version along with the copyright info, but if it's something done by readthedocs, nevermind :P

victorjulien:
Merged in #9336, thanks!

PR #9338, opened 2023-08-03
Bug 5870/v7

inashivb:

Previous PR: https://github.com/OISF/suricata/pull/9324

Changes since v6:
- add null check in `ConfGetChildValue`

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -354,6 +354,8 @@ int ConfGetChildValue(const ConfNode *base, const char *name, const char **vptr)
         return 0;
     }
     else {
+        if (node->val == NULL)
+            return 0;
         *vptr = node->val;
         return 1;
     }
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -289,6 +289,10 @@ static void *ParseAFPConfig(const char *iface)
     if (ConfGetChildValueWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
         if (strlen(out_iface) > 0) {
             aconf->out_iface = out_iface;
+            if (strcmp(iface, out_iface) == 0) {
+                FatalError("Invalid config: interface (%s) and copy-iface (%s) can't be the same",
+                        iface, out_iface);
+            }
         }
     }
 


src/conf.c
```diff
@@ -354,6 +354,8 @@ int ConfGetChildValue(const ConfNode *base, const char *name, const char **vptr)
         return 0;
     }
     else {
+        if (node->val == NULL)
+            return 0;
```
jasonish, line 358: I don't think this is correct, even tho it might not break anything.  `NULL` is a valid value for a configuration parameter:

```
some-value: ~
```

In which case I think we'd want the output pointer to be set to `NULL` given the node was found.

jasonish: Could use a unit test for this.

suricata-qa:

Pipeline 15509

src/conf.c
victorjulien: What would be the use case for allowing a NULL value?

inashivb: The way I interpret Jason's comment is:
Aim of the function: Find out whether the node exists (1) or not (0).
According to this PR's way, on returning 0, I'm claiming that the node didn't exist at all when in fact it did exist but just did not have any value assigned to it.

victorjulien: "Value" returning a NULL pointer as a valid "value" seems a bit like an anti-pattern to me, but I'm open to changing my mind if there are use cases.

inashivb: @jasonish do you have an example in mind?

jasonish: Given the YAML:

```
parent:
  child: ~
```

`child` exists with a value of NULL. Given a handle on the parent node,

```
const char *value = "bad";
ConfGetChildValue(parent, "child", &value);
```

I'd expect the function to return truthy (1) as child does exist, and have a NULL value.

jasonish: > What is the meaning of `child: ~`?

YAML for `NULL`. Other NULL values are `null` and no value, such as `value:`.

jasonish: Do we have to do anything? See this updated unit test: 

https://github.com/OISF/suricata/compare/master...jasonish:suricata:conf-get-child-value-null/v1

Maybe there is an expectation here that this function should fail out if its not a string?  Which is fair, but should be reflected with some function documentation.

victorjulien: w/o this PR, the af-packet parsing code segv's since it doesn't check the ptr

victorjulien: this can of course be fixed by adding a ptr null check, but that is what started this discussion. Why check a ptr if the function returns success.

jasonish: Valid point. From a purely YAML perspective, NULL is a valid value on success. So maybe function documentation to make it clear its more like a function named `ConfGetChildValueAsString`... And a unit test. Without that, a field that exists and is intentionally NULL shouldn't return an error IMO.

Serde makes this very clear, where the code would be something like...
- get value, returns true if value exists
- now you have to test what type of value it is, NULL, a string, a number?  Due to the enum type returned you are forced to do this extra check.

PR #9356, opened 2023-08-08
rule-reload: Release excess memory freed during engine reload

thomasjwinter:
The hot reload results in large chunks of memory being freed as the as the old signature tables are discarded. Help the memory management system along by telling to release as much memory as it can at this point.

- [X] I have read the contributing guide lines at https://docs.suricata.io/en/latest/devguide/codebase/contributing/contribution-process.html

Describe changes:
- We had issue on an embedded device of low memory on rule-reload.
- We found that while the memory has been freed, the kernel needs a kick to properly report this.
-

pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/detect-engine.c
+++ b/src/detect-engine.c
@@ -4679,6 +4679,15 @@ int DetectEngineReload(const SCInstance *suri)
     SCLogDebug("old_de_ctx should have been freed");
 
     SCLogNotice("rule reload complete");
+
+#ifdef HAVE_MALLOC_TRIM
+    /* The reload process potentially frees up large amounts of memory.
+     * Encourage the memory management system to reclaim as much as it
+     * can.
+     */
+    malloc_trim(0);
+#endif
+
     return 0;
 }
 
--- a/configure.ac
+++ b/configure.ac
@@ -219,6 +219,13 @@
             #include <sys/random.h> 
             ])
 
+    AC_CHECK_HEADERS([malloc.h])
+    AC_CHECK_DECL([malloc_trim],
+        AC_DEFINE([HAVE_MALLOC_TRIM], [1], [Use malloc_trim]),
+        [], [
+            #include <malloc.h> 
+            ])
+
     OCFLAGS=$CFLAGS
     CFLAGS=""
     AC_CHECK_FUNCS([strlcpy strlcat])
--- a/src/suricata-common.h
+++ b/src/suricata-common.h
@@ -215,6 +215,10 @@ typedef unsigned char u_char;
 #include <netdb.h>
 #endif
 
+#if HAVE_MALLOC_H
+#include <malloc.h>
+#endif
+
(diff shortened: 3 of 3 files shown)


victorjulien:
Looks like the last push to the Suricata-Verify repo causes some issues here. Can you rebase this PR?

src/detect-engine.c
```diff
@@ -87,6 +87,7 @@
 #include "runmodes.h"
 
 #include "reputation.h"
+#include <malloc.h>
```
victorjulien, line 90: this will need to be included in suricata-common.h, with a check in configure.ac to see if it is available

src/detect-engine.c
```diff
@@ -4679,6 +4680,13 @@ int DetectEngineReload(const SCInstance *suri)
     SCLogDebug("old_de_ctx should have been freed");
 
     SCLogNotice("rule reload complete");
+
+    /* The reload process potentially frees up large amounts of memory.
+     * Encourage the memory management system to reclaim as much as it
+     * can.
+     */
+    malloc_trim(0);
```
victorjulien, line 4688: we need a macro wrapper based on availability, it should live in util-mem.h

Man page says "This function is a GNU extension.", so we can't just assume that it is available as CI failures show.

victorjulien (changes requested):
Would also like to see some discussion of this mechanic: are there draw backs? Does it cause some performance impact?

thomasjwinter:
> Would also like to see some discussion of this mechanic: are there draw backs? Does it cause some performance impact?

I found this page which discusses malloc_trim and what it does which is a general Linux/glibc memory issue https://github.com/mruby/mruby/pull/5069
That page links to the following stackoverflow pages which also discuss it https://stackoverflow.com/questions/27945855/force-free-to-return-malloc-memory-back-to-os/47061458#47061458 and https://stackoverflow.com/questions/2215259/will-malloc-implementations-return-free-ed-memory-back-to-the-system/3860495#3860495

For our case, we were running on a system that produces a debug snapshots when memory is low and may reboot recover. We found that suricata appeared to use more memory than expected on rule reload but the memory was actually freed but needed the malloc_trim call to tell the kernel actually re-use it.

As for performance impact, my understanding is the kernel will do a little extra work to free the pages on the malloc_trim call then a little extra work to give a free page to suricata on the next malloc if needed. It shouldn't be a noticeable impact at all especially since rule reload doesn't happen very often compared to malloc/free calls.

catenacyber:
@thomasjwinter are you planning to work on this again ?

thomasjwinter:
I created redmine ticket https://redmine.openinfosecfoundation.org/issues/6454 and created new pull request https://github.com/OISF/suricata/pull/9756

PR #9337, opened 2023-08-03
af-xdp: detach XDP program early v4

joser12345678:
To mitigate a bug with AF_XDP sockets in high traffic scenarios, the XDP program must be detatched before the sockets are closed

Bug #6238

Make sure these boxes are signed before submitting your Pull Request -- thank you.

[JR ] I have read the contributing guide lines at https://docs.suricata.io/en/latest/devguide/codebase/contributing/contribution-process.html
[JR ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
[JR ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6238

Describe changes:
Added a RunModeAFXDPRemoveProg function to remove an AF_XDP program before closing sockets as this prevents the crash when receiving high traffic volume and shutting down suricata. Also fixed error for building and an error with an unread return value.
previous PR #9333 

### Provide values to any of the below to override the defaults.

To use a pull request use a branch name like `pr/N` where `N` is the
pull request number.

Alternatively, `SV_BRANCH` may also be a link to an
OISF/suricata-verify pull-request.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

PR #6238, opened 2021-06-29
doc/dcerpc: add proto keywords

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3029

PR #9360, opened 2023-08-08
af-xdp: detach XDP program early v5

joser12345678:
To mitigate a bug with AF_XDP sockets in high traffic scenarios, the XDP program must be detatched before the sockets are closed. This issue happens when large ammounts of traffic are sent to suricata and the XDP program is not removed before AF_XDP sockets are closed. I believe this is a race condition bug as detailed here: https://bugzilla.kernel.org/show_bug.cgi?id=217712

Further investigation shows this may be a bug exclusive to the driver/AMD processor combination.

This commit addresses the bug by ensuring the first thread to run the deinit function removes the XDP program, which fixes the bug as detailed in the bugzilla link.

Bug #6238

- [X ] I have read the contributing guide lines at https://docs.suricata.io/en/latest/devguide/codebase/contributing/contribution-process.html
- [X ] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [X ] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

https://redmine.openinfosecfoundation.org/issues/6238

Describe changes:
Added a RunModeAFXDPRemoveProg function to remove an AF_XDP program before closing sockets as this prevents the crash when receiving high traffic volume and shutting down suricata. Also fixed error for building and an error with an unread return value.
previous PR #9337 

pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/runmode-af-xdp.c
+++ b/src/runmode-af-xdp.c
@@ -62,6 +62,8 @@
 #include <linux/if_xdp.h>
 #include <linux/if_link.h>
 #include <xdp/xsk.h>
+#include <xdp/libxdp.h>
+#include <net/if.h>
 #endif
 
 const char *RunModeAFXDPGetDefaultMode(void)
@@ -385,6 +387,7 @@ int RunModeIdsAFXDPWorkers(void)
 #endif /* HAVE_AF_XDP */
     SCReturnInt(0);
 }
+
 /**
  * @}
  */
--- a/src/source-af-xdp.c
+++ b/src/source-af-xdp.c
@@ -59,10 +59,13 @@
 #include "runmodes.h"
 #include "flow-storage.h"
 #include "util-validate.h"
+#include "runmode-af-xdp.h"
 
 #ifdef HAVE_AF_XDP
-#include <xdp/xsk.h>
 #include <net/if.h>
+#include <bpf/libbpf.h>
+#include <xdp/xsk.h>
+#include <xdp/libxdp.h>
 #endif
 
 #if HAVE_LINUX_IF_ETHER_H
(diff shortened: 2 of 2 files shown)


src/source-af-xdp.c
```diff
@@ -829,10 +834,21 @@ static TmEcode ReceiveAFXDPLoop(ThreadVars *tv, void *data, void *slot)
...
 {
     AFXDPThreadVars *ptv = (AFXDPThreadVars *)data;
 
+    /*
+    * If AF_XDP is enabled, the program must be detached before the AF_XDP sockets
+    * are closed to mitigate a bug that causes an IO_PAGEFAULT in linux kernel
+    * version 5.19, unknown as of now what other versions this affects.
+    */
+    SCMutexLock(&sync_deinit);
+    RunModeAFXDPRemoveProg();
```
victorjulien, line 849: I think it would be best to now also move this whole function into this file. That way we keep the diff simpler and the logic local.

src/runmode-af-xdp.c
```diff
@@ -146,6 +148,43 @@ static TmEcode ConfigSetThreads(AFXDPIfaceConfig *aconf, const char *entry_str)
...
 }
 
+/**
+ * \brief function to unload an AF_XDP program
+ *
+ */
+void RunModeAFXDPRemoveProg(void)
+{
+    const char *live_dev = NULL;
+    if (live_dev == NULL)
```
rmcconnell-r7, line 158: What's this check for? The variable has been defined ``NULL`` and immediately checking for ``NULL``?

joser12345678: good catch. Its remnants of old code I tried. Fixing now. Will address in the next PR

PR #9374, opened 2023-08-10
conf: check if node value is Null

inashivb:
Leaving this as Draft here for discussion.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -354,6 +354,8 @@ int ConfGetChildValue(const ConfNode *base, const char *name, const char **vptr)
         return 0;
     }
     else {
+        if (node->val == NULL)
+            return 0;
         *vptr = node->val;
         return 1;
     }


jasonish (changes requested):
I think the change is ok, but should be accompanied by some function documentation as we're changing the function to one that can't differentiate between the value not existing and the value being null.

Doesn't help that the function wasn't documented in the first place.

suricata-qa:

Pipeline 15570

inashivb:
> I think the change is ok, but should be accompanied by some function documentation as we're changing the function to one that can't differentiate between the value not existing and the value being null.
> 
> Doesn't help that the function wasn't documented in the first place.

While writing the doc, I noticed that for the fn `ConfGetNode`, we have the following in place:
```
   * This function will return the value for a configuration node based
   * on the full name of the node.  It is possible that the value
   * returned could be NULL, this could happen if the requested node
   * does exist but is not a node that contains a value, but contains
   * children ConfNodes instead
```

This suggests that:
1. we already accept returning NULL values
2. we'll be breaking a pattern we do by accepting this PR
3. even if the node value is NULL, there could be children ConfNodes

I believe (3) could apply to children as well?

Does this mean that by accepting this PR, we'll be breaking patterns where a child has NULL value but may have other children? I think we might from what I understand.

Wdyt?
cc: @victorjulien

jasonish:
Ok, I'm second-guessing that this change might be OK.  It looks like I wrote most of this with the intention of NULL being an actual allowed value, so the caller could easily check if the value exists but is NULL, doesn't exist (returns error), or has a value.

This is more or less in line with what I've also seen in more modern Rust and/or Go applications. They'll add higher level functions like `MustGetValueAsString` which error out if the value does not exist or isn't convertible to a string, etc.

Given we kind of already have that pattern with `ConfGetInt`, a function named `ConfGetChildValueAsString` would actually make more sense than changing the contract of `ConfGetChildValue`.

PR #9388, opened 2023-08-15
memory/alloc: Use SCCalloc instead of malloc/memset

jlucovsky:
Continuation of #9385 

Replace `SCMalloc/memset` with `SCCalloc`

There are many places that use the following pattern:
```
     mem = SCMalloc(sizeof(foo));
     if (mem) {
        memset(&mem, 0, sizeof(foo));
    }
```

These can all be replaced by using `SCCalloc` which returns 0-filled memory.

Describe changes:
- Changed `SCMalloc` followed by `memset` to `SCCalloc` 

Updates:
- Fix issue with previous rebase.

pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -693,11 +693,11 @@ static AppLayerProtoDetectProbingParserElement *AppLayerProtoDetectProbingParser
 {
     SCEnter();
 
-    AppLayerProtoDetectProbingParserElement *p = SCMalloc(sizeof(AppLayerProtoDetectProbingParserElement));
+    AppLayerProtoDetectProbingParserElement *p =
+            SCCalloc(1, sizeof(AppLayerProtoDetectProbingParserElement));
     if (unlikely(p == NULL)) {
         exit(EXIT_FAILURE);
     }
-    memset(p, 0, sizeof(AppLayerProtoDetectProbingParserElement));
 
     SCReturnPtr(p, "AppLayerProtoDetectProbingParserElement");
 }
@@ -714,11 +714,11 @@ static AppLayerProtoDetectProbingParserPort *AppLayerProtoDetectProbingParserPor
 {
     SCEnter();
 
-    AppLayerProtoDetectProbingParserPort *p = SCMalloc(sizeof(AppLayerProtoDetectProbingParserPort));
+    AppLayerProtoDetectProbingParserPort *p =
+            SCCalloc(1, sizeof(AppLayerProtoDetectProbingParserPort));
     if (unlikely(p == NULL)) {
         exit(EXIT_FAILURE);
     }
-    memset(p, 0, sizeof(AppLayerProtoDetectProbingParserPort));
 
     SCReturnPtr(p, "AppLayerProtoDetectProbingParserPort");
 }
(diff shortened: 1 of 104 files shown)


src/app-layer-detect-proto.c
```diff
@@ -1269,10 +1268,9 @@ static int AppLayerProtoDetectPMMapSignatures(AppLayerProtoDetectPMCtx *ctx)
     int mpm_ret;
     SigIntId id = 0;
 
-    ctx->map = SCMalloc(ctx->max_sig_id * sizeof(AppLayerProtoDetectPMSignature *));
+    ctx->map = SCCalloc(1, ctx->max_sig_id * sizeof(AppLayerProtoDetectPMSignature *));
```
inashivb, line 1271: Q: Is there any difference b/w
1.
```
SCCalloc(1, ctx->max_sig_id * sizeof(AppLayerProtoDetectPMSignature *));
```
and
2.
```
SCCalloc(ctx->max_sig_id, sizeof(AppLayerProtoDetectPMSignature *));
```
? Why didn't we use (2)?

src/detect-engine-iponly.c
```diff
@@ -550,19 +549,17 @@ static void SigNumArrayPrint(void *tmp)
...
-    SigNumArray *new = SCMalloc(sizeof(SigNumArray));
+    SigNumArray *new = SCCalloc(1, sizeof(SigNumArray));
 
     if (unlikely(new == NULL)) {
         FatalError("Fatal error encountered in SigNumArrayNew. Exiting...");
     }
-    memset(new, 0, sizeof(SigNumArray));
 
-    new->array = SCMalloc(io_ctx->max_idx / 8 + 1);
+    new->array = SCCalloc(1, io_ctx->max_idx / 8 + 1);
```
inashivb, line 558: Same question as above.

inashivb (commented):
Looking good to me. A question inline about the change being done a certain way (because of a script?) when the declaration of calloc seems to be:

```
void *calloc(size_t nitems, size_t size)
```

jasonish (approved):
Looks OK to me.  I guess another pass could convert multiplication into `calloc` arguments, but that would also make it a little harder to review in a batch like this.

src/app-layer-detect-proto.c
jlucovsky: The first argument indicates the number of objects to allocate while the second indicates the object size.

Most of the `SCCalloc` calls allocate for one object so it'd be possible to add an interface, e.g., `SCCallocOne` (e.g.) and replace `SCCalloc(1, object_size)` with `SCCallocOne(object_size)`.

catenacyber: I agree with Shivani : `SCCalloc(ctx->max_sig_id, sizeof(AppLayerProtoDetectPMSignature *));` looks nicer

See `git grep 'SCCalloc(1, ' | grep ' \* '` which finds 46 occurences

catenacyber:
Nice @jlucovsky how did you find all these ? Do you have a magic bash one liner ? (that could be integrated into CI)

jlucovsky:
> Nice @jlucovsky how did you find all these ? Do you have a magic bash one liner ? (that could be integrated into CI)

```
cd /path/to/suricata/
git grep  -A 5 "\<SCMalloc\>"|grep memset
```

I ignored the usages in test-only code.

jlucovsky:
Continued in #9665

PR #9392, opened 2023-08-18
Conf null/v3

inashivb:
Previous PR: https://github.com/OISF/suricata/pull/9374

Changes since v2:
- add a `ConfGetChildValueString` fn more compliant w standards and code

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -469,6 +469,57 @@ int ConfGetChildValueIntWithDefault(const ConfNode *base, const ConfNode *dflt,
 }
 
 /**
+ * \brief Retrieve the value of a child configuration node as string.
+ *
+ * This function will return the value for a configuration node based
+ * on the full name of the node. In case the node exists but does not
+ * have a corresponding value, it is considered an error.
+ *
+ * \param base Base node to find the other nodes from.
+ * \param name Name of configuration parameter to get.
+ * \param vptr Pointer that will be set to the configuration value parameter.
+ *   Note that this is just a reference to the actual value, not a copy.
+ *
+ * \retval 1 will be returned if the name is found
+ * \retval 0 will be returned if the name is not found or has NULL value
+ */
+int ConfGetChildValueString(const ConfNode *base, const char *name, const char **vptr)
+{
+    int ret = ConfGetChildValue(base, name, vptr);
+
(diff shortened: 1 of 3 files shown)


src/conf.c
```diff
@@ -468,6 +468,57 @@ int ConfGetChildValueIntWithDefault(const ConfNode *base, const ConfNode *dflt,
     return ret;
 }
 
+/**
+ * \brief Retrieve the value of a child configuration node as string.
+ *
+ * This function will return the value for a configuration node based
+ * on the full name of the node. In case the node exists but does not
```
jasonish, line 475: In this case we're not using the full name of the node. We're just using the name relative to the `base` node being passed in.  Something better might be:
```
This function retrieves the value of a configuration node by name that exists under the provided base node.
```

src/conf.c
```diff
@@ -468,6 +468,57 @@ int ConfGetChildValueIntWithDefault(const ConfNode *base, const ConfNode *dflt,
...
+ *
+ * \retval 1 will be returned if the name is found
+ * \retval 0 will be returned if the name is not found or has NULL value
+ */
+int ConfGetChildValueString(const ConfNode *base, const char *name, const char **vptr)
+{
+    int ret = ConfGetChildValue(base, name, vptr);
+
+    if ((ret == 1 && vptr && *vptr == NULL) || (ret == 0)) {
+        /* Cannot convert NULL to string */
```
jasonish, line 491: As there no conversion going on here, a better comment would be ` Configuration node does not exist or does not have a value`.

src/conf.c
```diff
@@ -468,6 +468,57 @@ int ConfGetChildValueIntWithDefault(const ConfNode *base, const ConfNode *dflt,
...
+        return 0;
+    }
+    return 1;
+}
+
+/**
+ * \brief Retrieve the value of a child configuration node as string.
+ *
+ * This function will return the value for a configuration node based
+ * on the full name of the node.
```
jasonish, line 501: Same with respect to full name.

jasonish:
See https://github.com/OISF/suricata/pull/9393 where I added more to an existing unit test and converted it to load YAML.  Maybe cherry pick that in, and build on top of that commit to add unit test for this function.

PR #9393, opened 2023-08-18
conf: better doc and test for ConfGetChildValueWithDefault - v1

jasonish:
Add function documentation for ConfGetChildValueWithDefault.

Add more to the test:
- test with a NULL default node
- test a value that exists with a NULL value
- test for a value that does not exist
- use actual YAML for input


suricata-qa:

Pipeline 15638

PR #9395, opened 2023-08-19
config/flow: Avoid divide by zero at startup

jlucovsky:
Accommodate flow hash sizes of 0 for startup message. We don't expect the hash-size to be configured to 0, but if it is, avoid divide by 0 when displaying the config information.

Issue: 6255

Describe changes:
- Handle a flow hash-size of 0 in config startup message.

pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```


suricata-qa:

Pipeline 15662

catenacyber (changes requested):
I would rather forbid the user to put a config value of 0
See line 609-612

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9417

PR #9404, opened 2023-08-22
pgsql: add events to handle parser errors - v1

jufajardini:
https://redmine.openinfosecfoundation.org/issues/5524
https://redmine.openinfosecfoundation.org/issues/5566

Describe changes:
- Add events to pgsql, so when there's a recoverable error the parser can offer insight into what's wrong, and still parse possible following pgsql messages

Sharing as a draft as I'm not sure I fully grasp how applayer events work, so not sure I'm doing the right thing here. So far, I wasn't able to create sv tests that will showcase if `InvalidLength` and `MalformedData` work, but the SV shows a simple case for `TruncatedData` with `DataRow` messages.

TODOs:

- decide whether `TruncatedMessage` is actually needed - going with it, for now
- add more SV tests (`InvalidLength` and `MalformedMessage`)
- document
- Add descriptions of each event to the commit message

```
SV_BRANCH=https://github.com/OISF/suricata-verify/pull/1356
```

### DIFF
--- a/rust/src/pgsql/pgsql.rs
+++ b/rust/src/pgsql/pgsql.rs
@@ -23,6 +23,7 @@ use super::parser::{self, ConsolidatedDataRowPacket, PgsqlBEMessage, PgsqlFEMess
 use crate::applayer::*;
 use crate::conf::*;
 use crate::core::{AppProto, Flow, ALPROTO_FAILED, ALPROTO_UNKNOWN, IPPROTO_TCP};
+use nom7::error::{Error, ErrorKind};
 use nom7::{Err, IResult};
 use std;
 use std::collections::VecDeque;
@@ -33,6 +34,13 @@ pub const PGSQL_CONFIG_DEFAULT_STREAM_DEPTH: u32 = 0;
 static mut ALPROTO_PGSQL: AppProto = ALPROTO_UNKNOWN;
 
 static mut PGSQL_MAX_TX: usize = 1024;
+
+#[derive(Debug, PartialEq, Eq, AppLayerEvent)]
+pub enum PgsqlEvent {
+    InvalidLength,   // Can't identify the length field
+    TruncatedData,   // Failed to parse due to missing data
+    MalformedData,   // Enough data, but unexpected message format/fields
+}
 
 #[repr(u8)]
 #[derive(Copy, Clone, PartialOrd, PartialEq, Eq, Debug)]
@@ -134,6 +142,7 @@ pub struct PgsqlState {
     backend_pid: u32,
     state_progress: PgsqlStateProgress,
     tx_index_completed: usize,
+    events: u16,
 }
 
 impl State<PgsqlTransaction> for PgsqlState {
@@ -151,7 +160,7 @@ impl Default for PgsqlState {
         Self::new()
     }
 }
-    
+
 impl PgsqlState {
     pub fn new() -> Self {
         Self {
@@ -164,6 +173,17 @@ impl PgsqlState {
             backend_pid: 0,
             state_progress: PgsqlStateProgress::IdleState,
             tx_index_completed: 0,
+            events: 0,
+        }
+    }
+
+    /// Set an event on the most recent transaction
+    fn set_event(&mut self, event: PgsqlEvent) {
+        if let Some(tx) = self.transactions.back_mut() {
+            tx.tx_data.set_event(event as u8);
+            self.events += 1;
+        } else {
+            SCLogDebug!("PGSQL: trying to set event {} on non-existing transaction", event as u8);
         }
     }
 
@@ -356,9 +376,22 @@ impl PgsqlState {
                         _needed,
                         needed_estimation
                     );
+                    SCLogDebug!("PgsqlEvent::TruncatedData");
+                    self.set_event(PgsqlEvent::TruncatedData);
                     return AppLayerResult::incomplete(consumed as u32, needed_estimation as u32);
                 }
+                Err(Err::Error(Error{code:ErrorKind::Verify, ..})) => {
+                    SCLogDebug!("PgsqlEvent::InvalidLength");
+                    self.set_event(PgsqlEvent::InvalidLength);
+                    AppLayerResult::err();
+                }
+                Err(Err::Error(Error{code:ErrorKind::Switch, ..})) => {
+                    self.set_event(PgsqlEvent::MalformedData);
+                    SCLogDebug!("PgsqlEvent::MalformedData");
+                    return AppLayerResult::ok();
+                }
                 Err(_) => {
+                    SCLogDebug!("Error while parsing PGSQL request");
                     return AppLayerResult::err();
                 }
             }
@@ -505,10 +538,22 @@ impl PgsqlState {
                         needed_estimation,
                         &start
                     );
+                    self.set_event(PgsqlEvent::TruncatedData);
+                    SCLogDebug!("PgsqlEvent::TruncatedData");
                     return AppLayerResult::incomplete(consumed as u32, needed_estimation as u32);
                 }
+                Err(Err::Error(Error{code:ErrorKind::Verify, ..})) => {
+                    SCLogDebug!("PgsqlEvent::InvalidLength");
+                    self.set_event(PgsqlEvent::InvalidLength);
+                    AppLayerResult::err();
+                }
+                Err(Err::Error(Error{code:ErrorKind::Switch, ..})) => {
+                    self.set_event(PgsqlEvent::MalformedData);
+                    SCLogDebug!("PgsqlEvent::MalformedData");
+                    return AppLayerResult::ok();
+                }
                 Err(_) => {
-                    SCLogDebug!("Error while parsing PostgreSQL response");
+                    SCLogDebug!("Error while parsing PGSQL response");
                     return AppLayerResult::err();
                 }
             }
@@ -727,8 +772,8 @@ pub unsafe extern "C" fn rs_pgsql_register_parser() {
         tx_comp_st_ts: PgsqlTransactionState::RequestReceived as i32,
         tx_comp_st_tc: PgsqlTransactionState::ResponseDone as i32,
         tx_get_progress: rs_pgsql_tx_get_alstate_progress,
-        get_eventinfo: None,
-        get_eventinfo_byid: None,
+        get_eventinfo: Some(PgsqlEvent::get_event_info),
+        get_eventinfo_byid: Some(PgsqlEvent::get_event_info_by_id),
         localstorage_new: None,
         localstorage_free: None,
         get_tx_files: None,


rust/src/pgsql/pgsql.rs
```diff
@@ -164,6 +173,17 @@ impl PgsqlState {
...
+        }
+    }
+
+    /// Set an event on the most recent transaction
+    fn set_event(&mut self, event: PgsqlEvent) {
+        if let Some(tx) = self.transactions.back_mut() {
+            tx.tx_data.set_event(event as u8);
+            self.events += 1;
+        } else {
+            SCLogDebug!("PGSQL: trying to set event {} on non-existing transaction", event as u8);
```
jufajardini, line 186: I saw that most protos only add events to existing transactions. What if there's something with the very first request - which would lead to the creation of the transaction, that happens in case of successful parsing for pgsql, would it make sense to save those somewhere?

rust/src/pgsql/pgsql.rs
```diff
@@ -505,10 +538,22 @@ impl PgsqlState {
...
                 }
+                Err(Err::Error(Error{code:ErrorKind::Verify, ..})) => {
+                    SCLogDebug!("PgsqlEvent::InvalidLength");
+                    self.set_event(PgsqlEvent::InvalidLength);
+                    AppLayerResult::err();
+                }
+                Err(Err::Error(Error{code:ErrorKind::Switch, ..})) => {
+                    self.set_event(PgsqlEvent::MalformedData);
+                    SCLogDebug!("PgsqlEvent::MalformedData");
+                    return AppLayerResult::ok();
```
jufajardini, line 553: Must understand if I'm not seeing events with this one due to `UnknownMessage` type. But we only have that for BackEnd messages, so... must investigate. Personal note: tlpwhite jdbc pcap may help me here.

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .app_layer.flow.pgsql | 0 | 55 | - |
| .app_layer.tx.pgsql | 55 | 106 | 192.73% |
| .app_layer.error.pgsql.parser | 55 | 0 | - | 

Pipeline 15679

rust/src/pgsql/pgsql.rs
```diff
@@ -134,6 +142,7 @@ pub struct PgsqlState {
     backend_pid: u32,
     state_progress: PgsqlStateProgress,
     tx_index_completed: usize,
+    events: u16,
```
catenacyber, line 145: This field seems unused

rust/src/pgsql/pgsql.rs
```diff
@@ -164,6 +173,17 @@ impl PgsqlState {
             backend_pid: 0,
             state_progress: PgsqlStateProgress::IdleState,
             tx_index_completed: 0,
+            events: 0,
+        }
+    }
+
+    /// Set an event on the most recent transaction
+    fn set_event(&mut self, event: PgsqlEvent) {
```
catenacyber, line 181: Can you write this in `impl PgsqlTransaction ` ?

As it belongs to a transaction, not the state...

rust/src/pgsql/pgsql.rs
```diff
@@ -356,9 +376,22 @@ impl PgsqlState {
                         _needed,
                         needed_estimation
                     );
+                    SCLogDebug!("PgsqlEvent::TruncatedData");
+                    self.set_event(PgsqlEvent::TruncatedData);
```
catenacyber, line 380: I do not think we should set an event when a PGSQL payload is fragmented over multiple TCP packets...

rust/src/pgsql/pgsql.rs
```diff
@@ -356,9 +376,22 @@ impl PgsqlState {
...
                     return AppLayerResult::incomplete(consumed as u32, needed_estimation as u32);
                 }
+                Err(Err::Error(Error{code:ErrorKind::Verify, ..})) => {
```
catenacyber, line 383: It is not obvious to me why a `verify` error is `invalidlength`

In the future, I may add more pgsql parsing code with nom `verify` that is not about lengths

rust/src/pgsql/pgsql.rs
```diff
@@ -356,9 +376,22 @@ impl PgsqlState {
...
+                    SCLogDebug!("PgsqlEvent::InvalidLength");
+                    self.set_event(PgsqlEvent::InvalidLength);
+                    AppLayerResult::err();
+                }
+                Err(Err::Error(Error{code:ErrorKind::Switch, ..})) => {
+                    self.set_event(PgsqlEvent::MalformedData);
+                    SCLogDebug!("PgsqlEvent::MalformedData");
+                    return AppLayerResult::ok();
```
catenacyber, line 391: I wonder if it is possible to organize the code a bit differently :
1. get the PDU (or error if there is an invalid length)
2. parse the PDU (and if there is an error, still return `AppLayerResult::ok()` for next PDUs)

catenacyber:
> Sharing as a draft as I'm not sure I fully grasp how applayer events work, so not sure I'm doing the right thing here. So far, I wasn't able to create sv tests that will showcase if InvalidLength and MalformedData work, but the SV shows a simple case for TruncatedData with DataRow messages.

If you have one SV test with a triggering rule with an app-layer event, it looks good :-)
It is just that you do not have a pcap that triggers the other events...

To complete this, I think you should also add a file `rules/pgsql.rules` (and reserve a sid space in the redmine wiki) as is done for other protocols

rust/src/pgsql/pgsql.rs
catenacyber: (I think that the two tickets should have their own commit)

rust/src/pgsql/pgsql.rs
jufajardini: true that >_<'
I'll re-write this part.

rust/src/pgsql/pgsql.rs
jufajardini: Oki 🙇🏽

rust/src/pgsql/pgsql.rs
jufajardini: Not sure if this makes sense or not, but... The implementation I mostly see, the events are set in the transactions. But sometimes, there may not be a transaction, yet. Would it make any sense to have this counter here, for such cases, at least?

rust/src/pgsql/pgsql.rs
jufajardini: Connected to my wondering of what to do in case the event is the result of a parsing error, and to the possibility of not existing a transaction, I see that for there's a `set_event_notx` function (https://github.com/OISF/suricata/commit/1ba62993d5fc6ac6fa8). Wondering if this could be the answer.

rust/src/pgsql/pgsql.rs
jufajardini: Roger that on the commit per ticket part!
wrt to the re-structuring, I'm trying to visualize how to do that with pgsql, as the protocol only has pseudo-headers, and some PDUs may come with pure data, for which information on how to parse them (including length) would be present in a previous message, only...

rust/src/pgsql/pgsql.rs
catenacyber: I may create a transaction for the sole purpose of having the event...
Not sure it is the best way...

rust/src/pgsql/pgsql.rs
catenacyber: Maybe it makes sense at some point to have this counter, but the code now does not make use of it

So, I would remove it until it is really needed

rust/src/pgsql/pgsql.rs
catenacyber: I may create a transaction for the sole purpose of having the event...
Not sure it is the best way...

rust/src/pgsql/pgsql.rs
catenacyber: (As it is legit traffic)

rust/src/pgsql/pgsql.rs
catenacyber: > and some PDUs may come with pure data

Which ones ?
It looked to me that it is always one byte opcode, and 4 bytes length of PDU...

rust/src/pgsql/pgsql.rs
jasonish: This is about the only way right now. Frames will help at some point when events are frames get logged:

https://redmine.openinfosecfoundation.org/issues/5826

rust/src/pgsql/pgsql.rs
jufajardini: For some reason, I didn't see this question before T__T
When the pgsql backend is answering to a query request, for instance, a single PGSQL message may span over several PDUs, and the intermediate PGSQL messages don't come with any header info.
Taking https://github.com/OISF/suricata-verify/blob/master/tests/pgsql/pgsql-5000-query-results/input.pcap as an example, in packet no. 32, the second to last PGSQL message has an informed length of 482 bytes, but the PGSQL PDU only has 132 bytes, so the rest of the message spills onto the next PGSQL PDU, and for that one we don't have info on how to read it.
The info is actually on the same second-to-last message, which indicates that the next data will be a `Data` column with 360 bytes, but this is only present in the previous message. The last message starts in the middle of a rule signature, which is part of the `Data` result, which actually started on the previous message.
 I guess... this could become some transaction info that could be used for such cases?

catenacyber: On packet 32, I see the second to last PGSQL message has an informed length of 616 bytes (0x268)

Capture d’écran 2023-11-16 à 17 27 35

jufajardini: To me, it's 665:
image
But the overall thing is that in this case, the very last PGSQL message for this packet won't have any header info for us to process.

catenacyber: With Wireshark Version 4.0.10 I do not have the last postgresql in packet 32 but I see it well reassembled in packet 34

jufajardini: Thanks, I'll update mine, and maybe peak at how they handle this, I guess.

jufajardini: I have finally updated my Wireshark. I managed to get version Wireshark 4.0.6, but I still see errors starting with packet 32... 🤔

jufajardini: Still struggling a bit to figure out how to approach this bit for the PDU part. As I think of this, wondering if the introduction of the UnknownMessageType could be enough to handle this?
Introduced in 1ac5d97259994f73e645333 and 0fc9ade7a9f672560a276b

jufajardini: > I wonder if it is possible to organize the code a bit differently :
> 
> 1. get the PDU (or error if there is an invalid length)
> 2. parse the PDU (and if there is an error, still return `AppLayerResult::ok()` for next PDUs)

Added this to a ticket, as it doesn't look trivial, still, for now https://redmine.openinfosecfoundation.org/issues/6654)

I think that part of this is somewhat addressed by the implementation of the `UnknownMessageTypes` - although not entirely, as we're not changing how the PDU is processed, just reducing the situations when the parser returns an error.

catenacyber: > wondering if the introduction of the UnknownMessageType could be enough to handle this?

Yes but not sure it is the best way to have duplicated code to get the length everywhere...

But I guess the functionality (return AppLayerResult::ok() so as to parse next PDUs) can be achieved even if the code is duplicated...

jufajardini: Fair point, I tried to reduce that with the length function, but there's certainly a lot of room for improvement...

suricata-qa:
WARNING:
 
|--|--|--|--| 
| .app_layer.flow.pgsql | 0 | 55 | - |
| .app_layer.tx.pgsql | 55 | 106 | 192.73% |
| .app_layer.error.pgsql.parser | 55 | 0 | - | 

Pipeline 15679

PR #9406, opened 2023-08-23
Conf null/v4

inashivb:
- https://github.com/OISF/suricata/pull/9393

\+ a688944a6e73d3db429aac55e3b7058c2cf73774 &  bf564c654733b86cb9bb485a04467d51d12240bf

Previous PR: #9392 

Changes since v3:
- update function docs to better explain their use as per comments

### DIFF
--- a/src/runmode-af-packet.c
+++ b/src/runmode-af-packet.c
@@ -286,7 +286,7 @@ static void *ParseAFPConfig(const char *iface)
         }
     }
 
-    if (ConfGetChildValueWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
+    if (ConfGetChildValueStringWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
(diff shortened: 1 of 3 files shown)


suricata-qa:

Pipeline 15687

src/runmode-af-packet.c
```diff
@@ -286,7 +286,7 @@ static void *ParseAFPConfig(const char *iface)
         }
     }
 
-    if (ConfGetChildValueWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
+    if (ConfGetChildValueStringWithDefault(if_root, if_default, "copy-iface", &out_iface) == 1) {
         if (out_iface != NULL) {
```
jasonish, line 290: Is the point of this new function is that we can now skip this check for NULL?

inashivb: oh yes. Updating..

PR #9407, opened 2023-08-23
Analyze DPDK hugepages and be more user-friendly with non-existing user

lukashino:
PR is combination of:

- https://redmine.openinfosecfoundation.org/issues/6278
- https://redmine.openinfosecfoundation.org/issues/5966

Describe changes:
- analyze hugepages - for having allocated too many hugepages and too little/none
- tell the user that Suricata config suggest that it wants to be run as a custom user but no user is specified

pull request number.

```
SV_REPO=
SV_BRANCH=pr/1358
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/suricata.c
+++ b/src/suricata.c
@@ -2155,6 +2155,10 @@ static int InitRunAs(SCInstance *suri)
     }
     /* Get the suricata user ID to given user ID */
     if (suri->do_setuid == TRUE) {
+        if (suri->user_name == NULL) {
+            SCLogError("user name cannot be set to an empty value");
+            return TM_ECODE_FAILED;
+        }
         if (SCGetUserID(suri->user_name, suri->group_name,
                         &suri->userid, &suri->groupid) != 0) {
             SCLogError("failed in getting user ID");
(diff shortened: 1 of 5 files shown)


suricata-qa:

Pipeline 15688

src/suricata.c
```diff
@@ -2155,6 +2155,10 @@ static int InitRunAs(SCInstance *suri)
     }
     /* Get the suricata user ID to given user ID */
     if (suri->do_setuid == TRUE) {
+        if (suri->user_name == NULL) {
+            SCLogError("user name cannot be set to an empty value");
+            return TM_ECODE_FAILED;
+        }
```
jasonish, line 2161: Should we check for empty value as well instead of just null?
```
./src/suricata -T --user ""
```

src/suricata.c
```diff
@@ -2164,6 +2168,10 @@ static int InitRunAs(SCInstance *suri)
         sc_set_caps = TRUE;
     /* Get the suricata group ID to given group ID */
     } else if (suri->do_setgid == TRUE) {
+        if (suri->group_name == NULL) {
+            SCLogError("group name cannot be set to an empty value");
+            return TM_ECODE_FAILED;
+        }
```
jasonish, line 2174: Same here..

However, I wonder if we should move these checks into `util-privs.c` where some basic validation is already being done, so this is just an extra validation that seems to fit there.

jasonish (changes requested):
Comments inline.

src/suricata.c
lukashino: I didn't alter the logic on that one because it was already sanitized with the message "unable to get the user ID, check if user exist!! "

Albeit, it is true that this message can be more informative.

src/suricata.c
lukashino: move to the file as suggested.

src/suricata.c
lukashino: added a check for an empty user field

lukashino:
Continues in #9466

PR #9095, opened 2023-06-28
detect: adds flow integer keywords

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6164

Describe changes:
- detect: adds flow.pkts_toclient keyword

https://github.com/OISF/suricata-verify/pull/1267
```
SV_BRANCH=pr/1267
```

#9062 with commits squashed into one and using u64 for flow bytes count

PR #9416, opened 2023-08-30
detect: adds flow integer keywords

catenacyber:
https://redmine.openinfosecfoundation.org/issues/6164

Describe changes:
- detect: adds flow.pkts_toclient keyword

https://github.com/OISF/suricata-verify/pull/1267
```
SV_BRANCH=pr/1267
```

#9095 rebased to erase the status as changes requested and continue the debate

> Commit message needs a bit more explanation.

What do you want as an explanation ?
Doing git log --oneline | grep keyw , I see the latest added keyword is commit https://github.com/OISF/suricata/commit/9cb06d437644aba1733196087e7efd3e70d2afbb which does not have any more explanation than adding a keyword + reference to the redmine ticket...

> Also, can we support prefilter for the byte keywords?

Not obviously.
I think this would require to grow the size of the union PrefilterPacketHeaderValue (which may have perf impacts)
Any better ideas ?


suricata-qa:

Pipeline 15775

jufajardini:
The documentation change looks good to me. 
The non-zero bit flag... Looks ok, too, I think.

catenacyber:
> The non-zero bit flag... Looks ok, too, I think.

Oops, from another PR, they were merged

Now https://github.com/OISF/suricata/pull/9463

PR #9417, opened 2023-08-30
config/flow: fix division by zero

catenacyber:
https://redmine.openinfosecfoundation.org/issues/6255
https://redmine.openinfosecfoundation.org/issues/5920

Describe changes:
- Fixes commit 805b07fa4236 to really fix the division by zero if user supplies a configuration value of 0 for hash size

Replaces https://github.com/OISF/suricata/pull/9395 with earlier fix

### DIFF
--- a/src/flow.c
+++ b/src/flow.c
@@ -606,7 +606,7 @@ void FlowInitConfig(bool quiet)
             FatalError("Invalid value for flow.hash-size: NULL");
         }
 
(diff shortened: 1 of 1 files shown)


suricata-qa:

Pipeline 15778

jasonish:
Can we do a better user error message on 0, like we do for the NULL case above?

catenacyber:
> Can we do a better user error message on 0, like we do for the NULL case above?

Done in https://github.com/OISF/suricata/pull/9431

PR #9424, opened 2023-08-31
Multipart mime 3487 v18

catenacyber:

Describe changes:

Follows #9038 with rebase and squash

Still to do by me : 
- check the QA
- convert previous unit tests to SV tests


suricata-qa:
ERROR:
 

Pipeline 15791

rust/src/mime/mime.rs
```diff
@@ -0,0 +1,605 @@
+/* Copyright (C) 2021 Open Information Security Foundation
```
jasonish, line 1: 2023? Or did you write this initially in 2021?

rust/src/mime/mime.rs
```diff
@@ -0,0 +1,605 @@
...
+
+fn mime_parse_value_delimited(input: &[u8]) -> IResult<&[u8], &[u8]> {
+    let (input, _) = char('"')(input)?;
+    let mut escaping = false;
+    for i in 0..input.len() {
+        if input[i] == b'\\' {
+            escaping = true;
+        } else {
+            if input[i] == b'"' && !escaping {
+                return Ok((&input[i + 1..], &input[..i]));
```
jasonish, line 41: Could unicode mess with this.  I recently ran into this: https://github.com/jasonish/rust-suricatax-rule-parser/commit/d77c4e3ed20f0dce054bb9911b42c4efe13f80ff

rust/src/mime/mime.rs
```diff
@@ -0,0 +1,605 @@
...
+
+pub fn mime_parse_header_line(input: &[u8]) -> IResult<&[u8], &[u8]> {
+    let (input, name) = take_till(|ch: u8| ch == b':')(input)?;
+    let (input, _) = char(':')(input)?;
+    let (input, _) = take_while(is_mime_space)(input)?;
+    return Ok((input, name));
+}
+
+// s2 is already lower case
+pub fn rs_equals_lowercase(s1: &[u8], s2: &[u8]) -> bool {
```
jasonish, line 259: This isn't exposed to C, can drop the `rs` as its namespaced.

jasonish:
We should probably think about the name space in new code.  The `rs_` prefix has got to go for `extern` functions.  These form the externally linkable API (public or not) and should live behind a consistent prefix.  While that discussion seems a bit ongoing, `SC` is in common use now.

Also, since these Rust functions are part of the `C` API, they should probably follow the convention of `SCSomeTitleCase`.

But maybe we should wait for some further discussion on this.

rust/src/mime/mime.rs
catenacyber: I wrote it in 2021 ...

rust/src/mime/mime.rs
catenacyber: This is a buffer `&[u8]` not a string, so there is no need to worry about unicode, right ?

rust/src/mime/mime.rs
catenacyber: ok, renamed as `slice_equals_lowercase`

catenacyber:
> We should probably think about the name space in new code. The `rs_` prefix has got to go for `extern` functions. These form the externally linkable API (public or not) and should live behind a consistent prefix. While that discussion seems a bit ongoing, `SC` is in common use now.
> 
> Also, since these Rust functions are part of the `C` API, they should probably follow the convention of `SCSomeTitleCase`.
> 
> But maybe we should wait for some further discussion on this.

Let me know if I need to do more renaming :-)

catenacyber:
Replaced by #9436

PR #9428, opened 2023-09-01
libsuricata: threads, sources - v6

jasonish:
This is a rebase of https://github.com/OISF/suricata/pull/8745.

Changes:
- Scan-build fixes to example program.


jasonish:
Some very high level discussion points:

- This introduces a new logging callback. While I see its use in how Suricata is designed now, ideally we should get rid of library specific callbacks.  We should have callbacks that the eve module uses, that would be identical to the callbacks a library user would use, or even a plugin user. The eve sub-system would need some updates here.
- Sources can already be dynamically registered via a plugin. A library user should have access to the same registration functions meaning there doesn't really need to be `library` source.  Perhaps we need a more generic source for the purpose of plugins and library, I guess this is what the library source attempts to be, so this could be naming thing, but something to think about.
- Suricata proper, or a library user shouldn't be much different from each other, by that I mean Suricata proper running in `af-packet` IDS mode would be a routing that bootstraps the threads and the source no different than a library would.  Suricata mostly does this now, but defining an ergonomic API for Suricata proper and library users should be a goal
- Bringing event abstraction types like `Alert`, `Flow` is somewhat of a separate concern that should be moved out of the library effort. That would be an abstraction layer in the logging sub-system.  But I can see its desire by users extending Suricata on how it ended up here.

Ultimately this contribution was to showcase a concrete use case of Suricata as a library with the features required to meet business needs. So its a great starting point, and with 7.0 out of the way, its time to start diving in.

suricata-qa:

Pipeline 15803

suricata-qa:

Pipeline 15804

catenacyber:
What is expected from this draft ?

Review design/concepts ?

jasonish:
Closing for now, needs rebase and more work.

PR #9429, opened 2023-09-03
detect: Inline default pkt inspect engines

coledishington:
Scenarios with a small number of rules, no MPM-based rules, experienced a 6%-14% performance degradation from the commit 0965afd66 detect: pkt inspect engines
inwhich the default pkt inspect engines were converted to callbacks to simplify adding extra pkt inspect engines.

Avoid adding the default pkt inspect engines to the callback chain and instead call them directly in an inlined function within DetectRulePacketRules().

Bug: #6291

- [✓] I have read the contributing guide lines at https://docs.suricata.io/en/latest/devguide/codebase/contributing/contribution-process.html

https://redmine.openinfosecfoundation.org/issues/6291

Describe changes:
Updated from Suricata 4.0.6 to Suricata 7, this performance degradation was noticed on a setup with the following:
* 176 signatures
* 3 are inspecting packet payload
* 33 inspect application layer
* 83 are decoder event only
This performance impact was significant when running a small number of lightweight rules, but was not significant on larger (and more heavy-duty) rule sets.

The changes make the default packet inspection engines inline, like they were in Suricata 4 before extra packet inspection engines were supported.

No Suricata-verify test.

### DIFF
--- a/src/detect-engine-build.c
+++ b/src/detect-engine-build.c
@@ -1921,8 +1921,6 @@ static int SigMatchPrepare(DetectEngineCtx *de_ctx)
             SigMatch *sm = s->init_data->smlists[type];
             s->sm_arrays[type] = SigMatchList2DataArray(sm);
         }
-        /* set up the pkt inspection engines */
-        DetectEnginePktInspectionSetup(s);
 
         if (rule_engine_analysis_set) {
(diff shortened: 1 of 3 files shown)


catenacyber:
Could you investigate the CI failures with suricata-verify ?

PR #6292, opened 2021-08-12
Feature/http range/v14

victorjulien:
Continuation of #6282. Removes the per tracker mutex in favor of the THashData lock. Various other cleanups and fixes.

suricata-verify-pr: 517

PR #9430, opened 2023-09-04
packetpool: Lock return stack before checking if empty

coledishington:
Previously the return stack queue was checked before obtaining the lock. This leaves a small window where the FlowManager thread is waiting for the queue lock and another thread is populating the queue. The FlowManager thread would miss the signal as it had not started waiting for it. The FlowManager thread would be stuck forever.

The overall result of this is that there are no available packet buffers to be used for forcing closure of timed out flows, we run out of flows and traffic is blocked.

Bug: #6292

- [✓] I have read the contributing guide lines at https://docs.suricata.io/en/latest/devguide/codebase/contributing/contribution-process.html

https://redmine.openinfosecfoundation.org/issues/6292

Describe changes:
The return stack queue is checked before obtaining the lock in [src/tmqh-packetpool.c:118](https://github.com/OISF/suricata/blob/master/src/tmqh-packetpool.c#L118). This leaves a small window where the FlowManager thread is waiting for the queue lock and another thread is populating the queue. The FlowManager thread would miss the signal as it had not started waiting for it. The FlowManager thread would be stuck forever.

The overall result of this is that there are no available packet buffers to be used for forcing closure of timed out flows, we run out of flows and traffic is blocked.

No suricata-verify test.

### DIFF
--- a/src/tmqh-packetpool.c
+++ b/src/tmqh-packetpool.c
@@ -115,8 +115,8 @@ void PacketPoolWaitForN(int n)
         }
 
         /* check return stack, return to our pool and retry counting */
+        SCMutexLock(&my_pool->return_stack.mutex);
         if (my_pool->return_stack.head != NULL) {
-            SCMutexLock(&my_pool->return_stack.mutex);
             /* Move all the packets from the locked return stack to the local stack. */
             if (pp) {
                 pp->next = my_pool->return_stack.head;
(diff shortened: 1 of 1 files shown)


victorjulien:
I've looked info this, and I can't find a reason why we would involve the packet pool at all anymore. In the past it was used by the flow manager, but these days it isn't. So instead of patching the WaitForN function, I think we can simply remove it as there are no other users. Do you think I'm missing something?

coledishington:
> I've looked info this, and I can't find a reason why we would involve the packet pool at all anymore. In the past it was used by the flow manager, but these days it isn't. So instead of patching the WaitForN function, I think we can simply remove it as there are no other users. Do you think I'm missing something?

Thanks for the comment, this issue was originally seen in an earlier version of Suricata but it looked like the issue still existed. I am not sure which component of the packet pool isn't being used anymore, it looks like parts of the code still register for the handler and use functions outside of the hooks defined by tmqh_table. Is there a specific component of WaitForN() that isn't being used?

victorjulien:
See https://github.com/OISF/suricata/pull/9486 for a PR that removes the relevant code instead of "fixing" it.

coledishington:
> See #9486 for a PR that removes the relevant code instead of "fixing" it.

Thanks, will close this as not a bug.

PR #9418, opened 2023-08-30
Rust clippy172 v1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None

Describe changes:
- Fix clippy 1.72 warnings

@jasonish how can we prevent CI from going red when clippy 1.73 will be out ?

PR #9433, opened 2023-09-04
Rust clippy172 v2

catenacyber:
None

Describe changes:
- Fix clippy 1.72 warnings

https://github.com/OISF/suricata/pull/9418 with better fix for ARM where `char` is unsigned

### DIFF
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -44,7 +44,7 @@ impl StreamSlice {
     #[cfg(test)]
     pub fn from_slice(slice: &[u8], flags: u8, offset: u64) -> Self {
         Self {
-            input: slice.as_ptr() as *const u8,
+            input: slice.as_ptr(),
             input_len: slice.len() as u32,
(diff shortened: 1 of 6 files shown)


suricata-qa:

Pipeline 15821

victorjulien:
Merged in #9438, thanks!

jasonish:
> #9418 with better fix for ARM where `char` is unsigned

Not the first time we've hit an ARM issue. Should we try to get one in public CI?

victorjulien:
> > #9418 with better fix for ARM where `char` is unsigned
> 
> Not the first time we've hit an ARM issue. Should we try to get one in public CI?

Was just reading about self-hosted runners, and it seems github advises against using them in public repos for security reasons: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security

Perhaps a setup more like our qalab jobs would make more sense.

PR #9437, opened 2023-09-04
Multipart mime 3487 v20

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/3487

Describe changes:
- convert HTTP to use new rust mime parser
- convert SMTP to use new rust mime parser

Follows #9436 with fix for not taking the en d of line before the boundary as part of the file

```
SV_BRANCH=pr/1371
```
SV tests from unit tests in https://github.com/OISF/suricata-verify/pull/1371

Draft : 
- I think #9435 should get merged first (and QA rebased over it)

PR #9435, opened 2023-09-04
mime: process chunk as soon as possible

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
None yet, draft to see if it is ok

Describe changes:
- mime : process chunk right away (instead of buffering for later)

This allows to have more contents of a file just before stream depth is reached (as will be shown by QA diff on SURI_TLPW1_files_sha256)

See https://github.com/OISF/suricata/pull/9424

PR #9443, opened 2023-09-06
Multipart mime 3487 v21

catenacyber:

Describe changes:

Follows #9437 with fix for rust panic (copying slices must have the exact same size) + removing some dead code cf #9419

@jasonish any ideas on how to get a better feedback that `pub rs_mime_find_header_token` was no longer used by the C side and should be removed as dead code ?

```
SV_BRANCH=pr/1371
```
SV tests from unit tests in https://github.com/OISF/suricata-verify/pull/1371

Draft : 
- I think #9435 should get merged first (and QA rebased over it)


jasonish:
> @jasonish any ideas on how to get a better feedback that `pub rs_mime_find_header_token` was no longer used by the C side and should be removed as dead code ?

Not really.  The Rust code is just compiled into a library with externally addressable functions so I don't think we'll get a compiler to tell us.

jasonish:
Do you have scripts already for finding unused C code?

```
nm target/release/libsuricata_rust.a |awk '$2 == "T" && $3 !~ /^_/ { print $3 }'
```
will get you all externally linkable C functions from the Rust code.  There are a few in there for other libraries like Brotli

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9447

catenacyber:
> > @jasonish any ideas on how to get a better feedback that `pub rs_mime_find_header_token` was no longer used by the C side and should be removed as dead code ?
> 
> Not really. The Rust code is just compiled into a library with externally addressable functions so I don't think we'll get a compiler to tell us.

Can a compiler/other tool tell us all the "externally addressable functions" from rust ? and check if they are used from C ?

jasonish:
> Can a compiler/other tool tell us all the "externally addressable functions" from rust ? and check if they are used from C ?

I quick search tells me no.  The compiler links looks up functions it needs, it doesn't really see functions it doesn't need.  You can remove Rust from this question. As the `.a` library is the same as any other static C library, so all other tools like nm will work.

catenacyber:
I was not thinking about nm but some rust-analyzer tool ;-) (if it exists)

jasonish:
> I was not thinking about nm but some rust-analyzer tool ;-) (if it exists)

No, this is outside the scope of Rust and back in C land.

PR #9446, opened 2023-09-06
docs: check for min sphinx version; fix rtd theme - v1

jasonish:
Check for a minimum version of Sphinx before enabling the building of
documentation. Many distributions still ship with very old versions.

Ticket: https://redmine.openinfosecfoundation.org/issues/6297

Also, pin the theme when running on ReadTheDocs.  RTD changed the default theme
for projects that don't explicitly set it.

### DIFF
--- a/doc/userguide/conf.py
+++ b/doc/userguide/conf.py
@@ -143,6 +143,7 @@ if not on_rtd:
         else:
             app.add_stylesheet('css/suricata.css')
 else:
+    html_theme = 'sphinx_rtd_theme'
     html_context = {
         'css_files': [
             'https://media.readthedocs.org/css/sphinx_rtd_theme.css',
--- a/configure.ac
+++ b/configure.ac
@@ -2230,29 +2230,29 @@ fi
     fi
 
 # sphinx-build for documentation, and also check for a new enough version
-    AC_PATH_PROG(HAVE_SPHINXBUILD, sphinx-build, "no")
-    if test "$HAVE_SPHINXBUILD" != "no"; then
+    AC_PATH_PROG([SPHINX_BUILD], [sphinx-build], [no])
+    if test "$SPHINX_BUILD" != "no"; then
         MIN_SPHINX_BUILD_VERSION="3.4.3"
-        sphinx_build_version=$($HAVE_SPHINXBUILD --version 2>&1 | cut -d' ' -f2-)
+        sphinx_build_version=$($SPHINX_BUILD --version 2>&1 | cut -d' ' -f2-)
         AC_MSG_CHECKING([for sphinx-build >= $MIN_SPHINX_BUILD_VERSION])
         AS_VERSION_COMPARE([$sphinx_build_version], [$MIN_SPHINX_BUILD_VERSION],
             [
                 AC_MSG_RESULT([no, documentation will not be built])
-                HAVE_SPHINXBUILD="no"
+                SPHINX_BUILD="no"
             ],
             [], [])
-        if test "$HAVE_SPHINXBUILD" != "no"; then
-            AC_MSG_RESULT(yes)
+        if test "$SPHINX_BUILD" != "no"; then
+            AC_MSG_RESULT([yes])
         fi
     fi
 
-    if test "$HAVE_SPHINXBUILD" = "no"; then
+    if test "$SPHINX_BUILD" = "no"; then
        enable_sphinxbuild=no
        if test -e "$srcdir/doc/userguide/suricata.1"; then
            have_suricata_man=yes
        fi
     fi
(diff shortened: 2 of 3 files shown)


victorjulien:
Do we currently fail to build with any version less than 3.4.3?

jasonish:
> Do we currently fail to build with any version less than 3.4.3?

I didn't exhaustively try, but thats the oldest version I know for sure we build on.  I suspect its likely 3.4 and up..

jasonish:
> Do we currently fail to build with any version less than 3.4.3?

From the changelog it looks like 3.0 may be the oldest we build on.  Do you have a distro you want to build the docs on older than 3.4.3?  RHEL 8 and friends use a Sphinx version less than 2, as does Ubuntu 20.04 which is where we'll see this error the most I think.  Ubuntu 22.04 uses 4+.

victorjulien:
> > Do we currently fail to build with any version less than 3.4.3?
> 
> From the changelog it looks like 3.0 may be the oldest we build on. Do you have a distro you want to build the docs on older than 3.4.3? RHEL 8 and friends use a Sphinx version less than 2, as does Ubuntu 20.04 which is where we'll see this error the most I think. Ubuntu 22.04 uses 4+.

In general if we impose a specific check/limit in a stable branch, we should be liberal I think. We shouldn't unnecessarily fail on versions that do work.

jasonish:
> > > Do we currently fail to build with any version less than 3.4.3?
> > 
> > 
> > From the changelog it looks like 3.0 may be the oldest we build on. Do you have a distro you want to build the docs on older than 3.4.3? RHEL 8 and friends use a Sphinx version less than 2, as does Ubuntu 20.04 which is where we'll see this error the most I think. Ubuntu 22.04 uses 4+.
> 
> In general if we impose a specific check/limit in a stable branch, we should be liberal I think. We shouldn't unnecessarily fail on versions that do work.

Backtesting these are a pain, given most of the releases never made it in stable distributions or are not used in currently active distributions.. By a pain I mean, installing Python code that is not packaged in the distro can be a bit of a pain.

jasonish:
> > > > Do we currently fail to build with any version less than 3.4.3?
> > > 
> > > 
> > > From the changelog it looks like 3.0 may be the oldest we build on. Do you have a distro you want to build the docs on older than 3.4.3? RHEL 8 and friends use a Sphinx version less than 2, as does Ubuntu 20.04 which is where we'll see this error the most I think. Ubuntu 22.04 uses 4+.
> > 
> > 
> > In general if we impose a specific check/limit in a stable branch, we should be liberal I think. We shouldn't unnecessarily fail on versions that do work.
> 
> Backtesting these are a pain, given most of the releases never made it in stable distributions or are not used in currently active distributions.. By a pain I mean, installing Python code that is not packaged in the distro can be a bit of a pain.

Simple attempts in Python virtualenv using Sphinx 3.2 and 3.3 failed.  But mind you, using 3.4.3 failed as well. The trick is to find a distro that is using one of these versions and have all the dependencies worked out for us.  Non EOL distros I've looked at have 3.4.3 or newer, Or a much older version < 2.

victorjulien:
> > > > > Do we currently fail to build with any version less than 3.4.3?
> > > > 
> > > > 
> > > > From the changelog it looks like 3.0 may be the oldest we build on. Do you have a distro you want to build the docs on older than 3.4.3? RHEL 8 and friends use a Sphinx version less than 2, as does Ubuntu 20.04 which is where we'll see this error the most I think. Ubuntu 22.04 uses 4+.
> > > 
> > > 
> > > In general if we impose a specific check/limit in a stable branch, we should be liberal I think. We shouldn't unnecessarily fail on versions that do work.
> > 
> > 
> > Backtesting these are a pain, given most of the releases never made it in stable distributions or are not used in currently active distributions.. By a pain I mean, installing Python code that is not packaged in the distro can be a bit of a pain.
> 
> Simple attempts in Python virtualenv using Sphinx 3.2 and 3.3 failed. But mind you, using 3.4.3 failed as well. The trick is to find a distro that is using one of these versions and have all the dependencies worked out for us. Non EOL distros I've looked at have 3.4.3 or newer, Or a much older version < 2.

Ok, lets go with 3.4.3. We can always adjust if needed.

victorjulien (approved):

doc/userguide/conf.py
```diff
@@ -143,6 +143,7 @@ def setup(app):
         else:
             app.add_stylesheet('css/suricata.css')
 else:
+    html_theme = 'sphinx_rtd_theme'
```
jasonish, line 146: Will want to cherry-pick this commit back to 6.0.x otherwise the theme will change for 6.0.14. Its a clean cherry-pick, want a new PR for it though?

victorjulien: yes please

victorjulien:
Merged in #9449, thanks!

PR #9453, opened 2023-09-07
detect: SigMatchAppendSMToList can fail

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6104

Describe changes:
- fixes buffer overflow in rules parsing

Draft : all calls to `SigMatchAppendSMToList` now need to get their return value checked and handle it properly...
Is it good for now ? (100 more files to go)

PR #9455, opened 2023-09-07

catenacyber:

Describe changes:

#9453 not draft as all calls are checked

Steps to do this
1. `git grep SigMatchAppendSMToList src/*.c | cut -d: -f1 | uniq | while read i; do python3 sig.py $i; done` see script below
3. Compile and fix the errors by reusing the code of the last failure handling (ie return -1 instead of goto error) 

```sig.py
import sys

f = open(sys.argv[1], "r")
f2 = open(sys.argv[1]+".patch", "w")
for l in f.readlines():
    if "SigMatchAppendSMToList(s, sm," in l:
        l=l.replace("SigMatchAppendSMToList", "if (SigMatchAppendSMToList").replace(";", " < 0) {\n        sm->ctx = NULL;\n        SigMatchFree(de_ctx, sm);\n        sm = NULL;\n        goto error;\n    }")
    f2.write(l)
```

### DIFF
--- a/src/detect-app-layer-event.c
+++ b/src/detect-app-layer-event.c
@@ -286,12 +286,22 @@ static int DetectAppLayerEventSetup(DetectEngineCtx *de_ctx, Signature *s, const
     sm->ctx = (SigMatchCtx *)data;
 
     if (event_type == APP_LAYER_EVENT_TYPE_PACKET) {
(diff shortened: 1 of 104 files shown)


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9462

PR #9462, opened 2023-09-07

catenacyber:

Describe changes:

Replaces #9455 with fixups squashed

Steps to do this
1. `git grep SigMatchAppendSMToList src/*.c | cut -d: -f1 | uniq | while read i; do python3 sig.py $i; done` see script below
3. Compile and fix the errors by reusing the code of the last failure handling (ie return -1 instead of goto error) 

```sig.py
import sys

f = open(sys.argv[1], "r")
f2 = open(sys.argv[1]+".patch", "w")
for l in f.readlines():
    if "SigMatchAppendSMToList(s, sm," in l:
        l=l.replace("SigMatchAppendSMToList", "if (SigMatchAppendSMToList").replace(";", " < 0) {\n        sm->ctx = NULL;\n        SigMatchFree(de_ctx, sm);\n        sm = NULL;\n        goto error;\n    }")
    f2.write(l)
```

### DIFF
--- a/src/detect-app-layer-event.c
+++ b/src/detect-app-layer-event.c
@@ -286,12 +286,22 @@ static int DetectAppLayerEventSetup(DetectEngineCtx *de_ctx, Signature *s, const
     sm->ctx = (SigMatchCtx *)data;
 
     if (event_type == APP_LAYER_EVENT_TYPE_PACKET) {
-        SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_MATCH);
+        if (SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_MATCH) < 0) {
+            sm->ctx = NULL;
+            SigMatchFree(de_ctx, sm);
+            sm = NULL;
+            goto error;
+        }
     } else {
         if (DetectSignatureSetAppProto(s, data->alproto) != 0)
             goto error;
 
-        SigMatchAppendSMToList(s, sm, g_applayer_events_list_id);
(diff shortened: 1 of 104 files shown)


suricata-qa:

Pipeline 15868

src/detect-app-layer-event.c
```diff
@@ -286,12 +286,22 @@ static int DetectAppLayerEventSetup(DetectEngineCtx *de_ctx, Signature *s, const
     sm->ctx = (SigMatchCtx *)data;
 
     if (event_type == APP_LAYER_EVENT_TYPE_PACKET) {
-        SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_MATCH);
+        if (SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_MATCH) < 0) {
+            sm->ctx = NULL;
```
victorjulien, line 290: this is a very verbose pattern added a 100 times or so. Would like to see a more generalized solution.

catenacyber: Argh I was already pretty happy to have found a solution :-p

`SigMatchAppendSMToList` can now fail.
And its failure should induce its 100-ish callers to report the failure...

Or could it be that `SigMatchAppendSMToList` still returns `void` but sets a boolean like `sm->errored = true` which can be checked afterward ? (Then I guess you would not know precisely when the overflow happened)

catenacyber: Nice, on it :-)

PR #9465, opened 2023-09-07
issue 6300: correctly handle configuration includes from arrays - v1

jasonish:
Includes from an "include" array were being loaded into the wrong
parent as the logic for array handing in include context was not
updated.

If we are descending into an array in include context, pass through
the current parent so the included configuration is included where it
is expected.

Bug: https://redmine.openinfosecfoundation.org/issues/6300

SV_BRANCH=https://github.com/OISF/suricata-verify/pull/1382


suricata-qa:

Pipeline 15866

victorjulien (approved):

victorjulien:
Merged in #9492, thanks!

PR #9468, opened 2023-09-08
af-packet: handle null bpf filter - v1

jasonish:
Commit 587c326d736851ea169d500b2989c94fc6a64521 made our YAML parser
more cmopliant with respect to what should be a NULL value and now when
a key contains no value it is NULL instead of an empty string, or
something else like a literal "null" or "~".

Check that the returned bpf_filter is null before checking strlen.

Bug: https://redmine.openinfosecfoundation.org/issues/6302

### DIFF
--- a/src/util-bpf.c
+++ b/src/util-bpf.c
@@ -42,7 +42,7 @@ void ConfSetBPFFilter(
         }
     } else if (ConfGetChildValueWithDefault(if_root, if_default, "bpf-filter", bpf_filter) ==
                1) { // reading from a file
-        if (strlen(*bpf_filter) > 0) {
+        if (*bpf_filter && strlen(*bpf_filter) > 0) {
             SCLogConfig("%s: using file provided bpf filter %s", iface, *bpf_filter);
         }
     } else {


jasonish:
I guess we should re-open the discussion on `ConfGetChildValueWithDefault` and when it should return failiure.  While I don't like it, our usage patterns suggests that this should maybe return failure when the fields exists but contains a null value. There are other ways to check if a value exists and is none if needed. Thoughts?

suricata-qa:

Pipeline 15875

inashivb:
> I guess we should re-open the discussion on `ConfGetChildValueWithDefault` and when it should return failiure. While I don't like it, our usage patterns suggests that this should maybe return failure when the fields exists but contains a null value. There are other ways to check if a value exists and is none if needed. Thoughts?

I was thinking perhaps we could have a cocci script check whether we're checking the NULL values returned in such fn calls and if not, replace the call to `ConfGetChildValueWithDefault` with the fn `ConfGetChildValueString` ref: https://github.com/OISF/suricata/pull/9473/commits/6761e0aedba594351df9584dc63a6c1523742f28
Even if the value is not supposed to be string in certain cases, this will at least bring out all the unhandled parts of the code that we should manually inspect once. Wdyt? I had created a ticket for this a while back: https://redmine.openinfosecfoundation.org/issues/6249

victorjulien:
> > I guess we should re-open the discussion on `ConfGetChildValueWithDefault` and when it should return failiure. While I don't like it, our usage patterns suggests that this should maybe return failure when the fields exists but contains a null value. There are other ways to check if a value exists and is none if needed. Thoughts?
> 
> I was thinking perhaps we could have a cocci script check whether we're checking the NULL values returned in such fn calls and if not, replace the call to `ConfGetChildValueWithDefault` with the fn `ConfGetChildValueString` ref: [6761e0a](https://github.com/OISF/suricata/commit/6761e0aedba594351df9584dc63a6c1523742f28) Even if the value is not supposed to be string in certain cases, this will at least bring out all the unhandled parts of the code that we should manually inspect once. Wdyt? I had created a ticket for this a while back: https://redmine.openinfosecfoundation.org/issues/6249

The number of call sites seems limited, so a manual review/update pass is more appropriate I think.

jasonish:
A more general solution that matches our usage: https://github.com/OISF/suricata/pull/9476

PR #9456, opened 2023-09-07
detect: do not run tx detection without update

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6299

Describe changes:
- Optimization to not run transaction detection when a TCP packet did not update anything (no call to AppLayerParserParse)

Let's see what QA thinks of this

Maybe this check can go up with `if (pflow && pflow->alstate && likely(pflow->proto == p->proto)) {`

PR #9471, opened 2023-09-11
detect: do not run tx detection without update

catenacyber:

Describe changes:

Better check than #9456 as used by `AppLayerParserTransactionsCleanup`

### DIFF
--- a/etc/schema.json
+++ b/etc/schema.json
@@ -3759,6 +3759,9 @@
                                 "krb5_udp": {
                                     "$ref": "#/$defs/stats_applayer_error"
                                 },
+                                "modbus": {
+                                    "$ref": "#/$defs/stats_applayer_error"
+                                },
                                 "mqtt": {
                                     "$ref": "#/$defs/stats_applayer_error"
                                 },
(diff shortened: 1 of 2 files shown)


catenacyber:
3 S-V tests failing : tcp-fastopen, http2 upgrade (because of protocol upgrade having one packet first half HTTP1 second half HTTP2)

and `mqtt-unsub-rules` fails for rule about topicY
pcap has
- pkt 10 to server has unsubscribe topicX, unsubscribe topic Y
- pkt 11 to client tcp ack + MQTT unsubscribe ack for topicX
- pkt 12 ack to server
- pkt 13 to client MQTT unsubscribe ack for topicY
- pkt 14 ack to server

The rule is to server as sticky buffer `mqtt.unsubscribe.topic` is to server
On master, it triggers at packet 14, and not at packet 12 because we are waiting for transaction completion (not sure why)

With this PR, we skip detection at packet 14 as there was no update to server but we missed the update to the transaction completion...

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9475

PR #9477, opened 2023-09-11
Detect tcp noupdatetotx 6299 v4

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6299

Describe changes:
- Optimization to not run transaction detection when a TCP packet did not update anything (no call to AppLayerParserParse) in the packet direction

Let's see what QA thinks of this

Better check than #9475 to make unit test pass

PR #9481, opened 2023-09-13
Detect tcp noupdatetotx 6299 v5

catenacyber:

Describe changes:

Better check than #9477 with commit message including ticket number

### DIFF
--- a/etc/schema.json
+++ b/etc/schema.json
@@ -3759,6 +3759,9 @@
                                 "krb5_udp": {
                                     "$ref": "#/$defs/stats_applayer_error"
                                 },
+                                "modbus": {
+                                    "$ref": "#/$defs/stats_applayer_error"
+                                },
                                 "mqtt": {
                                     "$ref": "#/$defs/stats_applayer_error"
                                 },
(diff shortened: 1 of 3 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | IPS_AFP_stats_chk | | |
| .app_layer.tx.smtp | 74520 | 37800 | 50.72% | 

Pipeline 15916

victorjulien:
Needs investigation into the smtp issue

catenacyber:
@ct0br0 how can I reproduce the IPS stuff ? Do you have a pcap ?

catenacyber:
Investigation : 

Reproduce with smtp.pcap from https://wiki.wireshark.org/SampleCaptures

One Tx with this PR against 2 with master

This comes from the fact that the first transaction is freed when we parse `QUIT` which creates a second transaction
With this PR, the transaction is not freed because it does not have `detect_flags_tc & APP_LAYER_TX_INSPECTED_FLAG`
On master, detection to client is run first empty ack.
With this PR, detection to client is not run first empty ack as that is the purpose.
But it is not run on reply 250 when it should... because `tx_id_min` is 1, because even if DetectRunTx is not run on empty ack, `DetectRunPostRules` is run which calls AppLayerParserSetTransactionInspectId

catenacyber:
Replaced by #9560

PR #9482, opened 2023-09-13

catenacyber:

Describe changes:

Replaces #9462 with better prototype for `SigMatchAppendSMToList` to reduce code duplication

Steps to do this
1. `git grep SigMatchAppendSMToList src/*.c | cut -d: -f1 | uniq | while read i; do python3 sig.py $i; done`
3. Manual patches


src/detect-flow.c
```diff
@@ -409,14 +403,16 @@ int DetectFlowSetup (DetectEngineCtx *de_ctx, Signature *s, const char *flowstr)
...
-        SigMatchFree(de_ctx, sm);
-        sm = NULL;
+        appendsm = false;
     } else {
         s->init_data->init_flags |= SIG_FLAG_INIT_FLOW;
     }
 
-    if (sm != NULL) {
-        SigMatchAppendSMToList(s, sm, DETECT_SM_LIST_MATCH);
+    if (appendsm) {
```
catenacyber, line 411: special change here

src/detect-dsize.c
```diff
@@ -137,18 +136,13 @@ static int DetectDsizeSetup (DetectEngineCtx *de_ctx, Signature *s, const char *
 
     /* Okay so far so good, lets get this into a SigMatch
      * and put it in the Signature. */
-    sm = SigMatchAlloc();
-    if (sm == NULL){
-        SCLogError("Failed to allocate memory for SigMatch");
+    SigMatch *sm = SigMatchAppendSMToList(
```
catenacyber, line 139: Notice that we reuse `sm` for `s->init_data->dsize_sm `

src/detect-filestore.c
```diff
@@ -455,25 +448,22 @@ static int DetectFilestoreSetup (DetectEngineCtx *de_ctx, Signature *s, const ch
...
         AppLayerHtpNeedFileInspection();
     }
 
-    SigMatchAppendSMToList(s, sm, g_file_match_list_id);
-    s->filestore_ctx = (const DetectFilestoreData *)sm->ctx;
+    if (SigMatchAppendSMToList(
+                de_ctx, s, DETECT_FILESTORE, (SigMatchCtx *)fd, g_file_match_list_id) != NULL) {
+        goto error;
+    }
+    s->filestore_ctx = fd;
```
catenacyber, line 461: Notice that `fd` can be NULL and behavior should be the same as before

src/detect-pcre.c
```diff
@@ -918,12 +917,10 @@ static int DetectPcreSetup (DetectEngineCtx *de_ctx, Signature *s, const char *r
     if (sm_list == -1)
         goto error;
 
-    sm = SigMatchAlloc();
-    if (sm == NULL)
+    SigMatch *sm = SigMatchAppendSMToList(de_ctx, s, DETECT_PCRE, (SigMatchCtx *)pd, sm_list);
```
catenacyber, line 920: Notice that we use `sm` in later code

PR #9488, opened 2023-09-15
detect/threshold: implement per thread cache

victorjulien:
Thresholding often has 2 stages:

1. recording matches
2. appling an action, like suppress

E.g. with something like:
threshold:type limit, count 10, seconds 3600, track by_src; the recording state is about counting 10 first hits for an IP, then followed by the "suppress" state that might last an hour.

By_src/by_dst are expensive, as they do a host table lookup and lock the host. If many threads require this access, lock contention becomes a serious problem.

This patch adds a thread local cache to avoid the synchronization overhead. When the threshold for a host enters the "apply" stage, a thread local hash entry is added. This entry knows the expiry time and the action to apply. This way the action can be applied w/o the synchronization overhead.

A rbtree is used to handle expiration.

### DIFF
--- a/src/detect-engine-threshold.c
+++ b/src/detect-engine-threshold.c
@@ -100,6 +100,212 @@ int ThresholdIPPairHasThreshold(IPPair *pair)
     return IPPairGetStorageById(pair, ippair_threshold_id) ? 1 : 0;
 }
 
+#include "util-hash.h"
+
+typedef struct ThresholdCacheItem {
+    int8_t type; // by_src/by_dst
+    int8_t ipv;
+    int8_t retval;
+    uint32_t addr;
+    uint32_t sid;
+    SCTime_t expires_at;
+    RB_ENTRY(ThresholdCacheItem) rb;
+} ThresholdCacheItem;
+
+static thread_local HashTable *threshold_cache_ht = NULL;
+
+thread_local uint64_t cache_lookup_cnt = 0;
+thread_local uint64_t cache_lookup_notinit = 0;
+thread_local uint64_t cache_lookup_nosupport = 0;
+thread_local uint64_t cache_lookup_miss_expired = 0;
+thread_local uint64_t cache_lookup_miss = 0;
+thread_local uint64_t cache_lookup_hit = 0;
+thread_local uint64_t cache_housekeeping_check = 0;
+thread_local uint64_t cache_housekeeping_expired = 0;
+
+static void DumpCacheStats(void)
+{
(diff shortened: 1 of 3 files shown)


suricata-qa:
ERROR:
 

Pipeline 15955

suricata-qa:

Pipeline 15956

src/detect-engine-threshold.c
```diff
@@ -100,6 +100,212 @@ int ThresholdIPPairHasThreshold(IPPair *pair)
     return IPPairGetStorageById(pair, ippair_threshold_id) ? 1 : 0;
 }
 
+#include "util-hash.h"
+
+typedef struct ThresholdCacheItem {
+    int8_t type; // by_src/by_dst
```
victorjulien, line 106: rename to track

src/detect-engine-threshold.c
```diff
@@ -100,6 +100,212 @@ int ThresholdIPPairHasThreshold(IPPair *pair)
...
+            break;
+    }
+}
+
+/* hash table for threshold look ups */
+
+static uint32_t ThresholdCacheHashFunc(HashTable *ht, void *data, uint16_t datalen)
+{
+    ThresholdCacheItem *tci = data;
+    int hash = tci->ipv * tci->type + tci->addr + tci->sid;
```
victorjulien, line 179: need proper hash func

src/detect-engine-threshold.c
```diff
@@ -100,6 +100,212 @@ int ThresholdIPPairHasThreshold(IPPair *pair)
...
+        if (n) {
+            n->type = track;
+            n->ipv = 4;
+            n->retval = retval;
+            n->addr = addr;
+            n->sid = sid;
+            n->expires_at = expires;
+
+            if (HashTableAdd(threshold_cache_ht, n, 0) == 0) {
+                (void)THRESHOLD_CACHE_RB_INSERT(&threshold_cache_tree, n);
```
victorjulien, line 236: error check

jasonish:
How many items do we expect the hash table and rb tree to contain in normal operation?

victorjulien:
Want to redo this. I think the thread local cache makes sense, but will put an additional cost on the slow path. Want to improve the slow path by moving the storage out of host/ippair, and use a dedicated table for thresholds. The advantage would be that we can key it by ip(s)+sid, so thresholds for different rules can't cause contention like they do now.

suricata-qa:

Pipeline 15956

victorjulien:
replaced by #10925

PR #9494, opened 2023-09-16
github-ci: Fedora 37 to 39; use packaged cbindgen

victorjulien:


jasonish:
Fedora 39 isn't even beta yet. I would recommend this until closer to the final release. Or add it but leave 37 for now.

victorjulien:
> Fedora 39 isn't even beta yet. I would recommend this until closer to the final release. Or add it but leave 37 for now.

heh I thought i read somewhere it was out and didn't bother to check :)

PR #9498, opened 2023-09-18
rust: remove unused

catenacyber:
None, generic cleaning
https://redmine.openinfosecfoundation.org/issues/4083

Describe changes:

Found with

There is nothing more to be done :
- There are parsed fields which get never used. (like `rrclass` in DNSQueryEntry) : I feel good to keep them 

Rebase #9419

### DIFF
--- a/rust/src/dcerpc/dcerpc.rs
+++ b/rust/src/dcerpc/dcerpc.rs
@@ -38,54 +38,12 @@ pub const DCERPC_UUID_ENTRY_FLAG_FF: u16 = 0x0001;
 pub const PFC_FIRST_FRAG: u8 = 0x01;
 // Value to indicate last fragment
 pub const PFC_LAST_FRAG: u8 = 0x02;
-// Cancel was pending at sender
-pub const PFC_PENDING_CANCEL: u8 = 0x04;
-pub const PFC_RESERVED_1: u8 = 0x08;
-// supports concurrent multiplexing of a single connection.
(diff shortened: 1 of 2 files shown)


suricata-qa:
WARNING:
 
|--|--|--|--| 
| | ASAN_TLPR1_suri | | |
 

Pipeline 15992

jasonish (approved):

suricata-qa:

Pipeline 16043

rust/src/dcerpc/dcerpc.rs
```diff
@@ -38,54 +38,12 @@ pub const DCERPC_UUID_ENTRY_FLAG_FF: u16 = 0x0001;
 pub const PFC_FIRST_FRAG: u8 = 0x01;
 // Value to indicate last fragment
 pub const PFC_LAST_FRAG: u8 = 0x02;
-// Cancel was pending at sender
```
victorjulien, line 41: I don't really see defines like this as dead code, would rather have them stay. It will make things easier when support is added.

jasonish: Could we use a rather Rusty convention and prefix with with `_` so its clear to the reader they are not used? And perhaps awaiting use?

victorjulien: That would work for me

victorjulien (changes requested):
see inline comments

catenacyber:
Going on with https://github.com/OISF/suricata/pull/9842

PR #9502, opened 2023-09-19
mime: process chunk as soon as possible

catenacyber:
https://redmine.openinfosecfoundation.org/issues/6367

Describe changes:
- mime : process chunk right away (instead of buffering for later)

https://github.com/OISF/suricata-verify/pull/1390
```
SV_BRANCH=pr/1390
```

This allows to have more contents of a file just before stream depth is reached (as will be shown by QA diff on SURI_TLPW1_files_sha256)

Required for https://github.com/OISF/suricata/pull/9424

#9435 with ticket number and S-V test

### DIFF
--- a/src/util-decode-mime.c
+++ b/src/util-decode-mime.c
@@ -1386,6 +1386,9 @@ static int ProcessBase64BodyLine(const uint8_t *buf, uint32_t len,
         remaining = leftover_bytes;
         offset += consumed_bytes;
     }
+    if (ret == MIME_DEC_OK && state->data_chunk_len > 0) {
+        ret = ProcessDecodedDataChunk(state->data_chunk, state->data_chunk_len, state);
+    }
     return ret;
 }
 


suricata-qa:
ERROR:
 

Pipeline 16039

src/util-decode-mime.c
```diff
@@ -1386,6 +1386,9 @@ static int ProcessBase64BodyLine(const uint8_t *buf, uint32_t len,
         remaining = leftover_bytes;
         offset += consumed_bytes;
     }
+    if (ret == MIME_DEC_OK && state->data_chunk_len > 0) {
+        ret = ProcessDecodedDataChunk(state->data_chunk, state->data_chunk_len, state);
+    }
```
inashivb, line 1391: Would it be right to assess that the control would only come here if we have less than 4 remaining bytes? Or, there are other cases that you see that might bring the control here?

catenacyber: > Would it be right to assess that the control would only come here if we have less than 4 remaining bytes?

I am not sure chat you mean but `state->data_chunk_len` can go up to 3 kilobytes per `#define DATA_CHUNK_SIZE  3072`

inashivb: I meant that we're processing the data chunk in the loop before this statement, we also tend to return from the loop if there are less than 4B to process. So, when exactly should we come to this if statement? Only in case the `remaining` bytes are < 4 and we don't enter the loop in the first place?
We have a buffering logic for the leftover bytes to avoid decoding and processing incomplete base64 data and I think this might be affected with this statement. see `ProcessBase64Remainder` and `ProcessBase64BodyLineCopyRemainder`.
wdyt?

catenacyber: > So, when exactly should we come to this if statement?

We enter there once we finished the loop.

So, you are right, we should `break` from it instead of `return` for the `ProcessBase64BodyLineCopyRemainder`cases

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9513

PR #9503, opened 2023-09-19
Warnint 64to32 6186 v7

catenacyber:
[Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6186

Describe changes:
- fix `-Wshorten-64-to-32` warnings for some files

```
LIBHTP_BRANCH=pr/399
```
https://github.com/OISF/libhtp/pull/399

#9260 with more fixes

Still more warnings to fix and to add in CI

PR #9504, opened 2023-09-19
Warnint 64to32 6186 v8

catenacyber:
[Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6186

Describe changes:
- fix `-Wshorten-64-to-32` warnings for some files

```
LIBHTP_BRANCH=pr/399
```
https://github.com/OISF/libhtp/pull/399

#9503 with more fixes

Still more warnings to fix and to add in CI
- detect* files
- u* files

### DIFF
--- a/src/app-layer-htp-range.c
+++ b/src/app-layer-htp-range.c
@@ -139,7 +139,7 @@ static inline bool ContainerValueRangeTimeout(HttpRangeContainerFile *cu, const 
     return false;
 }
 
-static void ContainerUrlRangeUpdate(HttpRangeContainerFile *cu, uint32_t expire)
+static void ContainerUrlRangeUpdate(HttpRangeContainerFile *cu, uint64_t expire)
 {
     cu->expire = expire;
 }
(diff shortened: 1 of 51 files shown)


suricata-qa:
ERROR:
 

Pipeline 16038

src/app-layer-htp-range.c
```diff
@@ -543,14 +543,24 @@ File *HttpRangeClose(const StreamingBufferConfig *sbcfg, HttpRangeContainerBlock
             // a new range just begins where we ended, append it
             if (range->gap > 0) {
                 // if the range had a gap, begin by it
-                if (FileAppendData(c->container->files, sbcfg, NULL, range->gap) != 0) {
+                uint32_t gap = (uint32_t)range->gap;
+                if (range->gap > UINT32_MAX) {
```
inashivb, line 547: If we are downcasting `range->gap` to `uint32_t` first, under what circumstances should it be greater than `UINT32_MAX`?

inashivb (commented):
Hi, Philippe!
A question inline about a pattern I see throughout the PR.
I tried something trivial:
```c
#include <stdio.h>
#include <stdint.h>

int main()
{
    uint16_t myvar = 65535;
    printf("%d\n", myvar);
    uint8_t blah = (uint8_t)myvar;
    printf("%d\n", blah);
    if (blah > UINT8_MAX) {
        printf("YES\n");
    }
    return 0;
}
```
to understand how casting and limits work, with the max u16 value, I always get blah to be 255. Can something and if yes what would make blah cross 255?

src/app-layer-htp-range.c
catenacyber: Indeed,
looks like `range->gap > UINT32_MAX` if we had multiple gaps which combine over 4G

Let's redefine `range->gap` to be u32 in the first place

Thanks

catenacyber:
> Hi, Philippe! A question inline about a pattern I see throughout the PR. I tried something trivial:
> 
> to understand how casting and limits work, with the max u16 value, I always get blah to be 255. Can something and if yes what would make blah cross 255?

You should test myvar against 255 before casting it into `uint8_t blah`

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9512

PR #9474, opened 2023-09-11
detect/bytejump: Restrict post_offset to buffer

jlucovsky:
Restrict `post_offset` to remain in the buffer. When a negative `post_offset` value would point before the buffer beginning, treat it as though it points to the beginning of the buffer

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4624](https://redmine.openinfosecfoundation.org/issues/4624)

Describe changes:
- Changed "domatch" signature to return a bool to disambiguate return values
- Restrict `post_offset` to buffer.

### Provide values to any of the below to override the defaults.

To use a pull request use a branch name like `pr/N` where `N` is the
pull request number.

Alternatively, `SV_BRANCH` may also be a link to an
OISF/suricata-verify pull-request.

```
SV_REPO=
SV_BRANCH=pr/1384
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

PR #9506, opened 2023-09-19
detect/bytejump: Restrict post_offset to buffer start

jlucovsky:
Continuation of #9474 

Restrict `post_offset` to remain in the buffer. When a negative `post_offset` value would point before the buffer beginning, treat it as though it points to the beginning of the buffer

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [4624](https://redmine.openinfosecfoundation.org/issues/4624)

Describe changes:
- Changed "domatch" signature to return a bool to disambiguate return values
- Restrict `post_offset` to buffer.

pull request number.

```
SV_REPO=
SV_BRANCH=pr/1384
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/detect-bytejump.c
+++ b/src/detect-bytejump.c
@@ -132,10 +132,10 @@ static bool DetectBytejumpValidateNbytes(const DetectBytejumpData *data, int32_t
  *  \param m byte jump sigmatch
  *  \param payload ptr to the payload
  *  \param payload_len length of the payload
- *  \retval 1 match
- *  \retval 0 no match
+ *  \retval true match
+ *  \retval false no match
  */
-int DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
+bool DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
         const SigMatchCtx *ctx, const uint8_t *payload, uint32_t payload_len, uint16_t flags,
         int32_t nbytes, int32_t offset)
 {
@@ -148,7 +148,7 @@ int DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
     int extbytes;
 
     if (payload_len == 0) {
-        SCReturnInt(0);
+        SCReturnBool(false);
     }
 
     /* Validate the number of bytes we are testing
@@ -161,7 +161,7 @@ int DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
             SCLogDebug("Invalid byte_jump nbytes "
                        "seen in byte_jump - %d",
                     nbytes);
-            SCReturnInt(0);
+            SCReturnBool(false);
         }
     }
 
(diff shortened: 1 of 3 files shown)


suricata-qa:

Pipeline 16037

src/detect-bytejump.c
```diff
@@ -148,7 +148,7 @@ int DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
     int extbytes;
 
     if (payload_len == 0) {
```
inashivb, line 150: Could the payload ever be NULL?

src/detect-bytejump.c
```diff
@@ -220,26 +223,32 @@ int DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
...
-        SCLogDebug("NEWVAL: payload %p + %" PRIu64, payload, val);
+        jumpptr = payload + (int64_t)val;
+        SCLogDebug("NEWVAL: payload %p + %" PRIi64 " = %p\n", payload, (int64_t)val, jumpptr + val);
     } else if (flags & DETECT_BYTEJUMP_END) {
-        val = payload_len + val;
-        SCLogDebug("NEWVAL: payload %p + %" PRIu32 " - %" PRIu64, payload, payload_len, val);
+        jumpptr = payload + payload_len + (int64_t)val;
+        SCLogDebug(
+                "NEWVAL: payload %p + %" PRIu32 " + %" PRIi64, payload, payload_len, (int64_t)val);
     } else {
```
inashivb, line 237: What does this else condition mean?

src/detect-bytejump.c
```diff
@@ -250,9 +259,9 @@ int DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
 #endif /* DEBUG */
 
     /* Adjust the detection context to the jump location. */
-    det_ctx->buffer_offset = val;
```
inashivb, line 253: Would a debug validation statement here making sure that `jumpptr - payload` is never negative make sense?
Asking bc it seems to me  that `buffer_offset` is an unsigned var.

inashivb (commented):
Hi, Jeff. I looked at your PR and had some questions for my own understanding. They're inline.
~~Also, the ticket seems to have a pcap and certain rules that misbehave. Should we have s-v tests for those?~~ Whoops, seems they're already in place and I missed it.
Thank you. :)

src/detect-bytejump.c
jlucovsky: There are 2 byte extract options being checked for -- the default case covers when neither `from_beginning` nor `from_end` were specified in the rule.

src/detect-bytejump.c
jlucovsky: Good point; i'll add that and create a new PR.

jlucovsky:
Continued in #9516

jlucovsky:
Continued in #9516

PR #9508, opened 2023-09-20
mpm: thread ctx cleanups

victorjulien:
Remove unused thread ctx' from AC variants

Use single thread store in detection.

Minor cleanups.

Think this would be ok for 7, but can also wait for 8 if considered too intrusive.

### DIFF
--- a/src/app-layer-detect-proto.c
+++ b/src/app-layer-detect-proto.c
@@ -1998,7 +1998,7 @@ AppLayerProtoDetectThreadCtx *AppLayerProtoDetectGetCtxThread(void)
         for (j = 0; j < 2; j++) {
             mpm_ctx = &alpd_ctx.ctx_ipp[i].ctx_pm[j].mpm_ctx;
             mpm_tctx = &alpd_tctx->mpm_tctx[i][j];
-            mpm_table[mpm_ctx->mpm_type].InitThreadCtx(mpm_ctx, mpm_tctx);
+            MpmInitThreadCtx(mpm_tctx, mpm_ctx->mpm_type);
         }
     }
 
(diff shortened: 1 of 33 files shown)


suricata-qa:

Pipeline 16049

lukashino (commented):
Looks good though I have some questions.
Q1: Did you remove the stats? Or are those stats of unused AC variants? E.g. SCACBSPrintSearchStats. 
Q2: Just to be sure, contexts of unused AC variants were stored in det_ctx->mtcu and det_ctx->mtcs? And all that is now stored within single mtc?

victorjulien:
> Looks good though I have some questions. Q1: Did you remove the stats? Or are those stats of unused AC variants? E.g. SCACBSPrintSearchStats. 

The stats were never incremented. The print logic was hidden behind ifdefs that were never set. So just dead code.

> Q2: Just to be sure, contexts of unused AC variants were stored in det_ctx->mtcu and det_ctx->mtcs? And all that is now stored within single mtc?

No, the those are unrelated. Long time ago the idea was to have a thread ctx per mpm per thread per buffer (dns query, uri, etc). `mtcu` was meant for http.uri for example. However this turned out to be unnecessary. So I've removed all but one.

lukashino:
Ok, thanks for the explanation. I'm giving it a green light and I think it could be part of Suri 7.

victorjulien:
@jasonish any thoughts on merging this in 7? There is a small chance it would break custom code built on 7 I suppose.

victorjulien:
> Doesn't look like it would cause any issues to plugins, so it would only affect more deep integrations which would probably get rebuilt anyways. Whereas plugins have to be concerned with data structure size changing, does this change the size of `Packet`?

Guess I'm mostly concerned about the `InitThreadCtx` and `DestroyThreadCtx` function pointers. They were always set before, and now they aren't. So could lead to unexpected NULL-ptr derefs.

jasonish (approved):

victorjulien:
Merged in #9816, thanks!

PR #9522, opened 2023-09-26
Draft: reduce the number of fflush calls

jlucovsky:
DRAFT:

@victorjulien @jasonish  This is a poc for 
- Reducing the number of fflush calls
- Ensuring the file is flushed when closing
- (Temporary) Flush on pseudo-packets (in the alert log; this is done in the *wrong* function but demonstrates an action on pseudo packets in the output path).

Describe changes:
-
-
-

pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/doc/userguide/output/eve/eve-json-output.rst
+++ b/doc/userguide/output/eve/eve-json-output.rst
@@ -30,6 +30,12 @@ Output types::
       # Enable for multi-threaded eve.json output; output files are amended
       # with an identifier, e.g., eve.9.json. Default: off
       #threaded: off
+      # Control for calls to flush file contents:
+      # - 0 Flush after each write
+      # - <byte-count> Flush periodically. Track bytes written; when bytes
+      #   written is equal to or greater, flush file contents. The next
+      #   flush occurs when the bytes written meets or exceeds this value.
+      #flush-threshold: 0
       #prefix: "@cee: " # prefix to prepend to each log entry
       # the following are valid when type: syslog above
       #identity: "suricata"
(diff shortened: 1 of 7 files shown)


doc/userguide/output/eve/eve-json-output.rst
```diff
@@ -317,6 +323,32 @@ This example will cause each Suricata thread to write to its own "eve.json" file
...
+up to ``flush-threshold`` bytes unbuffered/unwritten should Suricata unexpectedly terminate.
+
+The default value is ``0``.
+
+::
+
+   outputs:
+     - eve-log:
+         filename: eve.json
+         flush-threshold: 4096
```
jasonish, line 345: I'd probably make this example value much higher.  Quite a lot of events are 4-8k.

victorjulien:
I think this still needs a mechanism for flushing things out when traffic stops completely. I think for this we'd need a "flush" packet that runs through the engine once a second or so.

jasonish:
> I think this still needs a mechanism for flushing things out when traffic stops completely. I think for this we'd need a "flush" packet that runs through the engine once a second or so.

Could this be done without modifying each packet source to implement a timeout and push through a flush packet?

victorjulien:
> > I think this still needs a mechanism for flushing things out when traffic stops completely. I think for this we'd need a "flush" packet that runs through the engine once a second or so.
> 
> Could this be done without modifying each packet source to implement a timeout and push through a flush packet?

We have a similar issue during the rule reload, and this is what the BreakLoop logic does. Essentially in workers mode the packet capture method can block the thread indefinitely and need a way to unblock.

In the case of detect reloads we have the main thread calling BreakLoop on the worker threads. So I think we can reuse that logic.

victorjulien:
I wonder if we can get rid of the buffering and with that flushing. According to `man fflush`, this is libc level buffering and we can control this per stream by changing the default block mode to unbuffered. We always write in a single `fwrite()` call. See `man setvbuf`.

PR #9525, opened 2023-09-27
quic: v2 support per rfc 9369

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4968

Describe changes:
- Adds support for quic v2 defined in rfc 9369

https://github.com/OISF/suricata-verify/pull/1395
```
SV_BRANCH=pr/1395
```

### DIFF
--- a/rust/src/quic/parser.rs
+++ b/rust/src/quic/parser.rs
@@ -44,6 +44,7 @@ impl QuicVersion {
     pub const Q044: QuicVersion = QuicVersion(0x51303434);
     pub const Q045: QuicVersion = QuicVersion(0x51303435);
     pub const Q046: QuicVersion = QuicVersion(0x51303436);
+    pub const V2: QuicVersion = QuicVersion(0x6b3343cf);
 
     fn is_gquic(&self) -> bool {
         *self == QuicVersion::Q043
@@ -61,6 +62,7 @@ impl From<QuicVersion> for String {
             QuicVersion(0x51303434) => "Q044".to_string(),
             QuicVersion(0x51303435) => "Q045".to_string(),
             QuicVersion(0x51303436) => "Q046".to_string(),
+            QuicVersion(0x6b3343cf) => "v2".to_string(),
             QuicVersion(x) => format!("{:x}", x),
         }
     }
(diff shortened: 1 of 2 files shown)


rust/src/quic/parser.rs
```diff
@@ -286,7 +288,18 @@ impl QuicHeader {
                             return Err(nom7::Err::Error(QuicError::InvalidPacket));
                         }
                     }
+                } else if version == QuicVersion::V2 {
```
victorjulien, line 291: branch looks the same as the next, is it needed?

rust/src/quic/parser.rs
```diff
@@ -286,7 +288,18 @@ impl QuicHeader {
...
+                        0x01 => QuicType::Initial,
+                        0x02 => QuicType::ZeroRTT,
+                        0x03 => QuicType::Handshake,
+                        0x00 => QuicType::Retry,
+                        _ => {
+                            return Err(nom7::Err::Error(QuicError::InvalidPacket));
+                        }
+                    }
                 } else {
+                    //if version == QuicVersion(1)
```
victorjulien, line 302: remove

rust/src/quic/parser.rs
catenacyber: Nope cf https://datatracker.ietf.org/doc/rfc9369/ 3.2.  Long Header Packet Types

Quic RFC is there to mess up with developers : there is a shift in the numbers for the quictypes...

(I think clippy catches branches with same code)

rust/src/quic/parser.rs
catenacyber: why ?
Do you want me to phrase this comment in English rather than pseudo code ?

rust/src/quic/parser.rs
victorjulien: ah yes I see it now

rust/src/quic/parser.rs
victorjulien: prefer that ya, my eyes immediately flagged it as a debug artifact

catenacyber:
Replaced by #9538

PR #9485, opened 2023-09-14
detect: SigMatchAppendSMToList can fail

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6104

Describe changes:
- fixes buffer overflow in rules parsing
- fixes memory leak in parsing `dcerpc.iface` on OOM condition (if `SigMatchAlloc` returns NULL)

Replaces #9484 with CI to go greener

Steps to do this
1. `git grep SigMatchAppendSMToList src/*.c | cut -d: -f1 | uniq | while read i; do python3 sig.py $i; done`
2. `ls src/*.patch | while read i; do mv $i `echo $i | sed s/.patch//`; done`
3. Manual patches

PR #9526, opened 2023-09-27
detect: check if signature uses too many buffers

catenacyber:

Describe changes:

#9485 for branch 7 with a simpler/hackier fix (when #9485 is the fix for 8)


victorjulien:
I think the code looks good for `master-7.0.x`, but I would like to see more of a explanation in the git message on why this approach has been taken.

catenacyber:
> I think the code looks good for `master-7.0.x`, but I would like to see more of a explanation in the git message on why this approach has been taken.

Done in https://github.com/OISF/suricata/pull/9528

PR #9528, opened 2023-09-28
detect: check if signature uses too many buffers

catenacyber:

Describe changes:

#9485 for branch 7 with a simpler/hackier fix (when #9485 is the fix for 8)

#9526 with a longer commit message to explain the choice

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -473,7 +473,8 @@ void SigMatchAppendSMToList(Signature *s, SigMatch *new, const int list)
                 s->init_data->curbuf == NULL) {
             if (SignatureInitDataBufferCheckExpand(s) < 0) {
                 SCLogError("failed to expand rule buffer array");
-                // return -1; TODO error handle
+                s->init_data->init_flags |= SIG_FLAG_INIT_OVERFLOW;
+                return;
             }
 
             /* initialize new buffer */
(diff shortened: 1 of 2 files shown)


src/detect-parse.c
```diff
@@ -1904,6 +1905,11 @@ static int SigValidate(DetectEngineCtx *de_ctx, Signature *s)
         SCReturnInt(0);
     }
 
+    if (s->init_data->init_flags & SIG_FLAG_INIT_OVERFLOW) {
+        SCLogError("rule %u tries to use too many buffers", s->id);
+        SCReturnInt(0);
+    }
+
```
inashivb, line 1912: Q: if there was a failure expanding the rule buffer array, could `s->init_data->curbuf->head` be NULL?
Asking to understand why this check wasn't added prior to the if statement before this.
Other than this, this patch makes sense to me. :)

inashivb: > I do not know. Why is it important ?
> 
> If `SigValidate` already returned that the signature is invalid for another reason, it is fine

I see. I had just thought we would want to say the exact the reason for invalid sig. If you think its fine either way, fine by me. Thanks :)

catenacyber: There could be other reasons that come later and we skip them...

victorjulien (approved):

victorjulien:
Merged in #9539, thanks!

PR #9534, opened 2023-09-28
readme: formatting, cleanups and a link to our issue tracker - v1

jasonish:

PR #9532, opened 2023-09-28
detect: remove code writing unused values

catenacyber:
Describe changes:
- remove code writing value that is unused

Coveridy ID 1546822 1546823 and 1546824

Looks like they introduced a new rule

PR #9513, opened 2023-09-22
mime: process chunk as soon as possible

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6367

Describe changes:
- mime : process chunk right away (instead of buffering for later)

https://github.com/OISF/suricata-verify/pull/1393
```
SV_BRANCH=pr/1393
```

This allows to have more contents of a file just before stream depth is reached (as will be shown by QA diff on SURI_TLPW1_files_sha256)

Required for https://github.com/OISF/suricata/pull/9424

#9502 with better patch thanks to code review

PR #9535, opened 2023-09-29
next/140/20230929/v1

victorjulien:
Staging:
 - #9513
 - #9516
 - #9532
 - #9534

SV_BRANCH=pr/1403

### DIFF
--- a/src/detect-bytejump.c
+++ b/src/detect-bytejump.c
@@ -168,9 +168,11 @@ bool DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
     /* Calculate the ptr value for the bytejump and length remaining in
      * the packet from that point.
      */
+    ptr = payload;
+    len = payload_len;
     if (flags & DETECT_BYTEJUMP_RELATIVE) {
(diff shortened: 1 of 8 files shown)


src/detect-bytejump.c
```diff
@@ -220,25 +223,31 @@ bool DetectBytejumpDoMatch(DetectEngineThreadCtx *det_ctx, const Signature *s,
...
-    if (val >= payload_len) {
-        SCLogDebug("Jump location (%" PRIu64 ") is not within "
-                   "payload (%" PRIu32 ")",
-                val, payload_len);
+    if (jumpptr < payload) {
+        jumpptr = payload;
+        SCLogDebug("jump location is before buffer start; resetting to buffer start");
+    } else if (jumpptr >= (payload + payload_len)) {
+        SCLogDebug("Jump location (%" PRIu64 ") is not within payload (%" PRIu32 ")",
+                payload_len + val, payload_len);
```
inashivb, line 250: @jlucovsky would the jump location not be based on `BEGIN`, `END` and the default case above?
I'm trying to understand why do we always print the jump location to be `payload_len + val`

PR #9552, opened 2023-10-04
Install: Be consistent with use of the "sudo" - v1

0xEniola:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5720

Describe changes:
- Added the "sudo" command to commands that require administrative privileges before they can be executed.

### DIFF
--- a/doc/userguide/install.rst
+++ b/doc/userguide/install.rst
@@ -93,15 +93,15 @@ Ubuntu/Debian
 Minimal::
 
     # Installed Rust and cargo as indicated above
-    apt-get install build-essential git libjansson-dev libpcap-dev \
+    sudo apt-get install build-essential git libjansson-dev libpcap-dev \
                     libpcre2-dev libtool libyaml-dev make pkg-config zlib1g-dev
     # On most distros installing cbindgen with package manager should be enough
-    apt-get install cbindgen # alternative: cargo install --force cbindgen
+    sudo apt-get install cbindgen # alternative: cargo install --force cbindgen
 
 Recommended::
 
     # Installed Rust and cargo as indicated above
-    apt-get install autoconf automake build-essential ccache clang curl git \
+    sudo apt-get install autoconf automake build-essential ccache clang curl git \
                     gosu jq libbpf-dev libcap-ng0 libcap-ng-dev libelf-dev \
                     libevent-dev libgeoip-dev libhiredis-dev libjansson-dev \
                     liblua5.1-dev libmagic-dev libnet1-dev libpcap-dev \
@@ -112,7 +112,7 @@ Recommended::
 
 Extra for iptables/nftables IPS integration::
 
-    apt-get install libnetfilter-queue-dev libnetfilter-queue1  \
+    sudo apt-get install libnetfilter-queue-dev libnetfilter-queue1  \
                     libnetfilter-log-dev libnetfilter-log1      \
                     libnfnetlink-dev libnfnetlink0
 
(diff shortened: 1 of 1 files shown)


jasonish:
Thanks for your contribution.  With respect to the details of the commit itself, could you make the following changes:
- Update the author line to be `First Last <email address>` so it matches the details you provided in the contribution agreement 
- Change for format of the subject to: `doc: be consistent <the rest here>`
- Include the ticket number
- See https://docs.suricata.io/en/latest/devguide/codebase/contributing/code-submission-process.html#commits for more details on the above

doc/userguide/install.rst
```diff
@@ -267,10 +267,10 @@ To use backports, the backports repository for the current stable
 distribution needs to be added to the system-wide sources list.
 For Debian 10 (buster), for instance, run the following as ``root``::
 
-    echo "deb http://http.debian.net/debian buster-backports main" > \
+    sudo echo "deb http://http.debian.net/debian buster-backports main" > \
         /etc/apt/sources.list.d/backports.list
```
jasonish, line 271: This particular command will actually fail when sudo is added, instead it should probably be:
```
echo "deb http://http.debian.net/debian buster-backports main" | sudo tee -a /etc/apt/sources.list.d/backports.list
```

jasonish:
We probably also need a `.. note::` or whatever the syntax is that `sudo` must be installed to follow the directions as-is.  If you install a minimal Debian, Ubuntu or RedHat EL like system, `sudo` often isn't there.

jasonish (changes requested):
Some updates required. Thanks!

doc/userguide/install.rst
0xEniola: You're right! Thank you for the feedback.

0xEniola:
> We probably also need a `.. note::` or whatever the syntax is that `sudo` must be installed to follow the directions as-is. If you install a minimal Debian, Ubuntu or RedHat EL like system, `sudo` often isn't there.

Excuse me, I do not understand the `..note::` part, can you please elaborate?

0xEniola:
> We probably also need a `.. note::` or whatever the syntax is that `sudo` must be installed to follow the directions as-is. If you install a minimal Debian, Ubuntu or RedHat EL like system, `sudo` often isn't there.

Installing `sudo` on a minimal system, one will have to be able to log in to root, and if the root password was not set at the time of installation, one will have to chroot into the system from a live disc/usb to set the root password.

jasonish:
> > We probably also need a `.. note::` or whatever the syntax is that `sudo` must be installed to follow the directions as-is. If you install a minimal Debian, Ubuntu or RedHat EL like system, `sudo` often isn't there.
> 
> Installing `sudo` on a minimal system, one will have to be able to log in to root, and if the root password was not set at the time of installation, one will have to chroot into the system from a live disc/usb to set the root password.

Where I often get stuck is... I install minimal system, usually in a container.  Then cut and paste a page of instructions that use sudo, but sudo isn't there.  Usually a quick `apt install sudo` fixes that but it can save someone reaching out to forums for support by having a one line item in the documentation.

0xEniola:
> > > We probably also need a `.. note::` or whatever the syntax is that `sudo` must be installed to follow the directions as-is. If you install a minimal Debian, Ubuntu or RedHat EL like system, `sudo` often isn't there.
> > 
> > Installing `sudo` on a minimal system, one will have to be able to log in to root, and if the root password was not set at the time of installation, one will have to chroot into the system from a live disc/usb to set the root password.
> 
> Where I often get stuck is... I install minimal system, usually in a container.  Then cut and paste a page of instructions that use sudo, but sudo isn't there.  Usually a quick `apt install sudo` fixes that but it can save someone reaching out to forums for support by having a one line item in the documentation.

re: a quick `apt install sudo` fixes that.

You know majority of these packages cannot be used without sudo or root privileges.

jasonish:
> You know majority of these packages cannot be used without sudo or root privileges.

Yes, but you can build, install, run Suricata-Verify, and do a lot of development before ever needing sudo. But its kind of beside the point, its about providing instructions that just work, with a note about where they might not work for you, like on a minimal install, or someone setting up their environment in docker.

0xEniola:
> > You know majority of these packages cannot be used without sudo or root privileges.
> 
> Yes, but you can build, install, run Suricata-Verify, and do a lot of development before ever needing sudo. But its kind of beside the point, its about providing instructions that just work, with a note about where they might not work for you, like on a minimal install, or someone setting up their environment in docker.

Yes I understand.

What I'm saying is, in the case a user wants to take the `sudo` route, to install `sudo` on Minimal, they'll have to boot live into their system and chroot into the system.

So I'll have to put these details into the doc then, right?

jasonish:
> > > You know majority of these packages cannot be used without sudo or root privileges.
> > 
> > 
> > Yes, but you can build, install, run Suricata-Verify, and do a lot of development before ever needing sudo. But its kind of beside the point, its about providing instructions that just work, with a note about where they might not work for you, like on a minimal install, or someone setting up their environment in docker.
> 
> Yes I understand.
> 
> What I'm saying is, in the case a user wants to take the `sudo` route, to install `sudo` on Minimal, they'll have to boot live into their system and chroot into the system.
> 
> So I'll have to put these details into the doc then, right?

No, I've never had to do that.  Usually when you don't have sudo already installed, you're dropped in as root. A container is a great example of this, or even many GitHub/Copilot dev containers.  Our instructions here will fail if I just go cut and paste them.  So just simply a note that the following instructions require sudo. Sphinx provides a markup for these types of note with something like `.. note:`

0xEniola:
> > > > You know majority of these packages cannot be used without sudo or root privileges.
> > > 
> > > 
> > > Yes, but you can build, install, run Suricata-Verify, and do a lot of development before ever needing sudo. But its kind of beside the point, its about providing instructions that just work, with a note about where they might not work for you, like on a minimal install, or someone setting up their environment in docker.
> > 
> > Yes I understand.
> > 
> > What I'm saying is, in the case a user wants to take the `sudo` route, to install `sudo` on Minimal, they'll have to boot live into their system and chroot into the system.
> > 
> > So I'll have to put these details into the doc then, right?
> 
> No, I've never had to do that.  Usually when you don't have sudo already installed, you're dropped in as root. A container is a great example of this, or even many GitHub/Copilot dev containers.  Our instructions here will fail if I just go cut and paste them.  So just simply a note that the following instructions require sudo. Sphinx provides a markup for these types of note with something like `.. note:`

OK! I understand you now.

So I should just specify at the beginning of each distro sub-heading a note that `sudo` is required to be installed for the instructions to work.

OK! On it.
Thank you for your assistance, Jason.

0xEniola:
> Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5720
> 
> Describe changes:
> 
> * Added the "sudo" command to commands that require administrative privileges before they can be executed.

PR #9553, opened 2023-10-04
Draft: JA4 for TLS and QUIC -- v3

satta:
Previous PR: #9545 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/6379

Changes to previous PR:
- Add `ja4.hash` rule keyword.
- Add reference to Redmine ticket to commit message.
- rustfmt/clang-format
- Support enable/auto/disable of JA4 for TLS in settings.
- Address buffer overread when JA4 is enabled and JA3 is disabled.
- Address infinite loop when `sigalgo_len` is `0xFFFF`.
- Proper bounds checking for ALPN protocol strings.
- Remove coverity annotations for new code sections in `app-layer-ssl.c`.
- Move maximum calculation for TLS versions into JA4 state.
- Properly handle GREASE values for ALPN strings.
- Simplify access to first value of ALPN.
- Remove unintentional curly braces in a `use` statement.
- Change and simplify passing of finished JA4 string to C.
- Explicitly use `std::mem::drop` in `SCJA4Free()`.

```
SV_BRANCH=pr/1408
```

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -143,8 +143,9 @@ enum {
     ERR_EXTRACT_VALIDITY,
 };
 
-/* JA3 fingerprints are disabled by default */
+/* JA3 and JA4 fingerprints are disabled by default */
 #define SSL_CONFIG_DEFAULT_JA3 0
+#define SSL_CONFIG_DEFAULT_JA4 0
 
 enum SslConfigEncryptHandling {
     SSL_CNF_ENC_HANDLE_DEFAULT = 0, /**< disable raw content, continue tracking */
@@ -154,10 +155,12 @@ enum SslConfigEncryptHandling {
 
 typedef struct SslConfig_ {
     enum SslConfigEncryptHandling encrypt_mode;
-    /** dynamic setting for ja3: can be enabled on demand if not explicitly
-     *  disabled. */
+    /** dynamic setting for ja3 and ja4: can be enabled on demand if not
+     *  explicitly disabled. */
     SC_ATOMIC_DECLARE(int, enable_ja3);
     bool disable_ja3; /**< ja3 explicitly disabled. Don't enable on demand. */
+    SC_ATOMIC_DECLARE(int, enable_ja4);
+    bool disable_ja4; /**< ja4 explicitly disabled. Don't enable on demand. */
 } SslConfig;
 
 SslConfig ssl_config;
@@ -691,6 +694,10 @@ static inline int TLSDecodeHSHelloVersion(SSLState *ssl_state,
     uint16_t version = (uint16_t)(*input << 8) | *(input + 1);
     ssl_state->curr_connp->version = version;
 
+    if (SC_ATOMIC_GET(ssl_config.enable_ja4) && ssl_state->curr_connp->ja4 != NULL) {
+        SCJA4SetTLSVersion(ssl_state->curr_connp->ja4, version);
+    }
+
     /* TLSv1.3 draft1 to draft21 use the version field as earlier TLS
        versions, instead of using the supported versions extension. */
     if ((ssl_state->current_flags & SSL_AL_FLAG_STATE_SERVER_HELLO) &&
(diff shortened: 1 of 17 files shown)


catenacyber:
Looks like CI found a memory leak ;-)

catenacyber (changes requested):
Thanks for the work Sascha :-)

- CI : A leak to fix
- Code : doing right now
- Commits segmentation : Ok
- Commit messages : Ok
- Git ID set : looks fine for me
- CLA : you already contributed
- Doc update : There should be documentation for the new keywords ;-)
- Redmine ticket : ok
- Rustfmt : ok
- Tests : Suricata-verify tests look fine, some comments there
- Dependencies added: none

satta:
> Looks like CI found a memory leak ;-)

Yep. Will get on it today to bring this into a better shape.

src/app-layer-ssl.c
```diff
@@ -1171,6 +1192,104 @@ static inline int TLSDecodeHSHelloExtensionEllipticCurvePF(SSLState *ssl_state,
...
+
+            if (!(HAS_SPACE(protolen)))
+                goto invalid_length;
+
+            /* we only want the first value for JA4 */
+            if (alpn_processed_len == 1) {
+                SCJA4SetALPN(ssl_state->curr_connp->ja4, (const char *)input, protolen);
+            }
+
+            alpn_processed_len += protolen;
```
catenacyber, line 1276: I think there is still an off by one overflow :
if `alpn_len` = 0xFFFF and `input_len` is 0x10000
Some protocols reach the state where `alpn_processed_len` = 0xFFFD

Then, we have a two bytes protolen
alpn_processed_len += 1 = 0xFFFE
alpn_processed_len += 1 = protolen (2) = 0

protolen = *input; -> overhead off by one

catenacyber: I would check that proto_len fits in alps_len as well

src/detect.h
```diff
@@ -284,6 +284,7 @@ typedef struct DetectPort_ {
 #define SIG_FLAG_INIT_FILEDATA              BIT_U32(9)  /**< signature has filedata keyword */
 #define SIG_FLAG_INIT_JA3                   BIT_U32(10) /**< signature has ja3 keyword */
 #define SIG_FLAG_INIT_OVERFLOW              BIT_U32(11) /**< signature has overflown buffers */
+#define SIG_FLAG_INIT_JA4                   BIT_U32(12) /**< signature has ja4 keyword */
```
catenacyber, line 287: I think we can reuse and make `SIG_FLAG_INIT_JA3 ` more generic

src/detect-ja4-hash.c
```diff
@@ -0,0 +1,192 @@
...
+
+    DetectAppLayerInspectEngineRegister2("ja4.hash", ALPROTO_QUIC, SIG_FLAG_TOSERVER, 1,
+            DetectEngineInspectBufferGeneric, Ja4DetectGetHash);
+
+    DetectBufferTypeSetDescriptionByName("ja4.hash", "TLS JA4 hash");
+
+    g_ja4_hash_buffer_id = DetectBufferTypeGetByName("ja4.hash");
+}
+
+int Ja4IsDisabled(const char *type)
```
catenacyber, line 100: Why this type arg ?

satta: I tried to stay as close as possible to JA3, see `Ja3IsDisabled()` -- and IIRC that allows to check if support is disabled fully or only no role existed that enabled it.

src/detect.h
catenacyber: I am not talking about config, rather about internal  `SIG_FLAG_INIT` to limit their consumption

src/detect-ja4-hash.c
catenacyber: But it is called only with the same argument "rule"...

src/detect.h
satta: Right, it appears to only be used in one spot 🤔

satta: Done

satta:
> > Looks like CI found a memory leak ;-)
> 
> Yep. Will get on it today to bring this into a better shape.

Memleak fixed.

src/app-layer-ssl.c
satta: I think it would also be sufficient to just ensure that `alpn_processed_len` cannot overflow until `alpn_len` is reached, for instance by making it larger than 16 bits unsigned.
Then in such cases we would just break out of the while loop before reading too far into the input. Did I miss anything?

src/detect-ja4-hash.c
satta: I came up with another way to do this.

src/app-layer-ssl.c
catenacyber: > I think it would also be sufficient to just ensure that `alpn_processed_len` cannot overflow until `alpn_len` is reached
 
That is implied by "checking that proto_len fits in alps_len"

I think that this is the solution for the SSL libraries (so that they do not read the part of the next extension as an ALPN)

satta: Thanks, done!

satta:
Next PR: #9576

PR #9557, opened 2023-10-05
detect: frees SigMatch when too many buffers

catenacyber:

Describe changes:
- detect: frees SigMatch when too many buffers

But now, instead of leaking memory, we risk use after free, as the caller is not aware these structures have been freed (see dsize for instance which will use `sm`)
But this is the only central place before some of the pointers get out of scope...

### DIFF
--- a/src/detect-parse.c
+++ b/src/detect-parse.c
@@ -474,6 +474,8 @@ void SigMatchAppendSMToList(Signature *s, SigMatch *new, const int list)
             if (SignatureInitDataBufferCheckExpand(s) < 0) {
                 SCLogError("failed to expand rule buffer array");
                 s->init_data->init_flags |= SIG_FLAG_INIT_OVERFLOW;
+                SigMatchFree(de_ctx, new);
+                sigmatch_table[new->type].Free(de_ctx, new->ctx);
                 return;
             }
 


catenacyber:
See https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=62990

victorjulien:
A workaround/hack that might work: instead of returning, set curbuf to a valid buffer (e.g. at index 0). Then the overflow flag set will trigger a cleanup a bit later?

catenacyber:
But we will leak the buffer at index 0 in this case, right ?

victorjulien:
> But we will leak the buffer at index 0 in this case, right ?

If we leak the buffer there something is missing from the cleanup code.

catenacyber:
I am not sure I understand.

`s->init_data->buffers` is full of 64 buffers that will be cleaned alright.
But we cannot add a 65th as we do not want to grow the array.
So, we leak this 65th buffer.
If we set `curbuf` to `&s->init_data->buffers[0]`, we will leak the first buffer instead of this 65th one...

Do you want to keep one spot in the buffers array, and check it sooner (before another one is added) to get a change to have a good cleanup ?

victorjulien:
> I am not sure I understand.
> 
> `s->init_data->buffers` is full of 64 buffers that will be cleaned alright. But we cannot add a 65th as we do not want to grow the array. So, we leak this 65th buffer. If we set `curbuf` to `&s->init_data->buffers[0]`, we will leak the first buffer instead of this 65th one...
> 
> Do you want to keep one spot in the buffers array, and check it sooner (before another one is added) to get a change to have a good cleanup ?

I think I don't understand what is being leaked.

But anything added to the buffers should be freed on error, right? If we can't add a new buffer (e.g. 65), then adding a match to buffer 0 (which is the wrong buffer/list, but sig will error out anyway), then the match should still be freed when we handle the error triggered by the overflow flag?

catenacyber:
Replaced by #9622

catenacyber:
> I think I don't understand what is being leaked.

It is a `SigMatch` and its `ctx` field

> But anything added to the buffers should be freed on error, right? If we can't add a new buffer (e.g. 65), then adding a match to buffer 0 (which is the wrong buffer/list, but sig will error out anyway), then the match should still be freed when we handle the error triggered by the overflow flag?

Ok, so now I understand that I did not understand the different objects in `Signature`

Next version coming

PR #9564, opened 2023-10-05
docs: update readthedocs config to new options - v1

jufajardini:
Our documentation was failing to build (https://readthedocs.org/projects/suricata/builds/22112658/), seems connected to the new way of indicating build options (cf
https://readthedocs.org/projects/suricata/builds/22112658/, https://docs.readthedocs.io/en/stable/config-file/v2.html#build, and https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os).

Describe changes:
- Added the build.os required new field 
- adjusted the way python version is passed

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,11 +1,9 @@
 # Required by Read The Docs
 version: 2
 
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.8"
+
 formats: all
-
-python:
-  version: "3.8"
-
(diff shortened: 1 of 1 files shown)


jufajardini:
> Looks like it still fails, do you have your own "instance" of the Suricata docs building on RTD?

I did setup one a while back, I'll try to debug this with it. 🙇🏽

jufajardini:
Replaced by: https://github.com/OISF/suricata/pull/9566

(never know if we have a different etiquette to follow for draft PRs... should one still create new PRs, or is it fine to force push, in such cases?

PR #9568, opened 2023-10-05
doc: Be consistent with use of the "sudo" - v2

0xEniola:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5720

Previous PR: #9552

Describe changes:
- Incorporate feedback from last PR.
- Added `Important` notes to inform users it is required to install `sudo` in order to successfully follow instructions.
- Removed the added `sudo` commands added to some instructions where `sudo` was not needed.

PR #9573, opened 2023-10-06
doc: Be consistent with use of the "sudo" - v3

0xEniola:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/5720

Previous PR: https://github.com/OISF/suricata/pull/9568

Describe changes:
- Incorporate feedback from last PR.
- Changed the `Important` directives to `note` directives.

### DIFF
--- a/doc/userguide/install.rst
+++ b/doc/userguide/install.rst
@@ -90,18 +90,20 @@ Rust support::
 Ubuntu/Debian
 """""""""""""
 
+.. note:: The following instructions require `sudo` be installed.
+
 Minimal::
 
     # Installed Rust and cargo as indicated above
-    apt-get install build-essential git libjansson-dev libpcap-dev \
+    sudo apt-get install build-essential git libjansson-dev libpcap-dev \
                     libpcre2-dev libtool libyaml-dev make pkg-config zlib1g-dev
     # On most distros installing cbindgen with package manager should be enough
-    apt-get install cbindgen # alternative: cargo install --force cbindgen
+    sudo apt-get install cbindgen # alternative: cargo install --force cbindgen
 
 Recommended::
 
     # Installed Rust and cargo as indicated above
-    apt-get install autoconf automake build-essential ccache clang curl git \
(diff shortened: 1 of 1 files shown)


doc/userguide/install.rst
```diff
@@ -90,18 +90,20 @@ Rust support::
 Ubuntu/Debian
 """""""""""""
 
+.. note:: The following instructions require `sudo` be installed.
```
jufajardini, line 93: nit: 
```suggestion
.. note:: The following instructions require `sudo` to be installed.
```

jufajardini: Same for the other notes ;)

doc/userguide/install.rst
```diff
@@ -193,6 +199,8 @@ After installing you can proceed to the :ref:`Basic setup`.
 Upgrading
 """""""""
 
+.. note:: The following instructions require `sudo` be installed.
+
```
jufajardini, line 203: As removing, upgrading or running Suricata would only be possible if one had previously installed it, I don't think we need the note about installing `sudo` or in the sections that are not about installing Suri.

jufajardini (changes requested):
CI :  running
Code :  doesn't apply
Commits segmentation :  ok
Commit author:  ok
Commit messages : ok 
Ticket number on message : ok
CLA :  (to be confirmed)
Doc update :  see inline comments ;)
Redmine ticket :  ok
Rustfmt :  doesn't apply
Tests : 
Dependencies added:  none

jufajardini:
One other thing I couldn't comment inline: for the `Debian` section, don't the lines after the `echo` command require the usage of `sudo` as well?

doc/userguide/install.rst
0xEniola: So you're suggesting the notes there are not necessary then.

You're right, still. 

OK.
Working on it now.

jufajardini: 😺

PR #9576, opened 2023-10-07
JA4 for TLS and QUIC -- v4

satta:
Previous PR: #9553 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/6379

Changes to previous PR:
- Use wider type for `alpn_processed_len` to avoid overflow.
- Unify use of `SIG_FLAG_INIT_JA{3,4}` -> `SIG_FLAG_INIT_JA`.
- Simplify checking of enabled/disabled status in JA4 rule keywords.
- Fix memory leak.

```
SV_BRANCH=pr/1411
```

### DIFF
--- a/src/app-layer-ssl.c
+++ b/src/app-layer-ssl.c
@@ -143,8 +143,9 @@ enum {
     ERR_EXTRACT_VALIDITY,
 };
 
-/* JA3 fingerprints are disabled by default */
+/* JA3 and JA4 fingerprints are disabled by default */
 #define SSL_CONFIG_DEFAULT_JA3 0
+#define SSL_CONFIG_DEFAULT_JA4 0
 
 enum SslConfigEncryptHandling {
     SSL_CNF_ENC_HANDLE_DEFAULT = 0, /**< disable raw content, continue tracking */
@@ -154,10 +155,12 @@ enum SslConfigEncryptHandling {
 
 typedef struct SslConfig_ {
     enum SslConfigEncryptHandling encrypt_mode;
-    /** dynamic setting for ja3: can be enabled on demand if not explicitly
-     *  disabled. */
+    /** dynamic setting for ja3 and ja4: can be enabled on demand if not
+     *  explicitly disabled. */
     SC_ATOMIC_DECLARE(int, enable_ja3);
     bool disable_ja3; /**< ja3 explicitly disabled. Don't enable on demand. */
+    SC_ATOMIC_DECLARE(int, enable_ja4);
+    bool disable_ja4; /**< ja4 explicitly disabled. Don't enable on demand. */
 } SslConfig;
 
 SslConfig ssl_config;
@@ -691,6 +694,10 @@ static inline int TLSDecodeHSHelloVersion(SSLState *ssl_state,
     uint16_t version = (uint16_t)(*input << 8) | *(input + 1);
     ssl_state->curr_connp->version = version;
 
+    if (SC_ATOMIC_GET(ssl_config.enable_ja4) && ssl_state->curr_connp->ja4 != NULL) {
+        SCJA4SetTLSVersion(ssl_state->curr_connp->ja4, version);
+    }
+
     /* TLSv1.3 draft1 to draft21 use the version field as earlier TLS
        versions, instead of using the supported versions extension. */
     if ((ssl_state->current_flags & SSL_AL_FLAG_STATE_SERVER_HELLO) &&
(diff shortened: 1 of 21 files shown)


src/app-layer-ssl.c
```diff
@@ -1171,6 +1193,104 @@ static inline int TLSDecodeHSHelloExtensionEllipticCurvePF(SSLState *ssl_state,
...
+
+    uint16_t alpn_len = (uint16_t)(*input << 8) | *(input + 1);
+    input += 2;
+
+    if (!(HAS_SPACE(alpn_len)))
+        goto invalid_length;
+
+    if ((ssl_state->current_flags & SSL_AL_FLAG_STATE_CLIENT_HELLO) &&
+            SC_ATOMIC_GET(ssl_config.enable_ja4)) {
+        uint32_t alpn_processed_len = 0;
```
catenacyber, line 1263: Fair enough for me.
Maybe you can put a comment about why u32 ;-)

catenacyber:
Still documentation missing for the new rules keywords ;-)

src/app-layer-ssl.c
```diff
@@ -1373,6 +1522,13 @@ static int TLSDecodeHandshakeHello(SSLState *ssl_state,
     int ret;
     uint32_t parsed = 0;
 
+    if (SC_ATOMIC_GET(ssl_config.enable_ja4) &&
+            ssl_state->current_flags & SSL_AL_FLAG_STATE_CLIENT_HELLO) {
+        ssl_state->curr_connp->ja4 = SCJA4New();
```
catenacyber, line 1527: Why is it this condition to prevent the leak ? `ssl_state->current_flags & SSL_AL_FLAG_STATE_CLIENT_HELLO`

Why not `ssl_state->curr_connp->ja4 == NULL`

src/app-layer-ssl.c
```diff
@@ -1373,6 +1522,13 @@ static int TLSDecodeHandshakeHello(SSLState *ssl_state,
...
+    } else {
+        ssl_state->curr_connp->ja4 = NULL;
```
catenacyber, line 1529: Why set it to NULL ?
Either it is NULL already, or the memory is leaking, right ?

jasonish:
Can `ja4` be moved out of `quic`? Does seem to be `quic` only, so can probably live on its own in `src/`.

satta:
> Can `ja4` be moved out of `quic`? Does seem to be `quic` only, so can probably live on its own in `src/`.

Yes, good idea! I wasn't sure if we want to have a separate directory for such shared code that will be accessed from C as well.

src/app-layer-ssl.c
satta: Done.

jasonish:
> > Can `ja4` be moved out of `quic`? Does seem to be `quic` only, so can probably live on its own in `src/`.
> 
> Yes, good idea! I wasn't sure if we want to have a separate directory for such shared code that will be accessed from C as well.

I think a top-level `ja4` is good for now with the FFI functions directly in it. I don't see it growing much bigger, it sounds like the other ja4 methods would need to be done as a plugin or something, if we even want to go there.

src/app-layer-ssl.c
satta: It's not technically required to set it to NULL and not at all meant as a signal. I was just trying to avoid uninitialized values here in case the `curr_connp` is not zeroed (which it is, I found the associated `memset`). So this can be removed as it only raises questions.

src/app-layer-ssl.c
satta: The idea is to only create a JA4 state if we have

1. JA4 effectively enabled, and
2. we are in a client hello (JA4 -- not JA4S -- is defined for the client only).

So just not having a JA4 state set yet is not enough to do the right thing. My assumption is there is only ever one handshake to be parsed in a TLS flow, and if there is a client hello, it is first.

You are right, I am not sure if `TLSDecodeHandshakeHello()` can be called multiple times in a flow for more than one client hello, so let's be defensive -- I have added an additional `NULL` check to avoid leaking memory by overwriting the pointer if it is already set, unlikely as it may be. 👍🏻

satta:
> Still documentation missing for the new rules keywords ;-)

Yep, noticed that as well. Will be added in the next PR!

satta:
Next PR: #9598

PR #9577, opened 2023-10-07
detect-ssh-proto-version: use FAIL macros in tests

I3run0:
Make sure these boxes are signed before submitting your Pull Request -- thank you.

- [X] I have read the contributing guide lines at https://docs.suricata.io/en/latest/devguide/codebase/contributing/contribution-process.html
- [X] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [X] I have updated the user guide (in doc/userguide/) to reflect the changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6337](https://redmine.openinfosecfoundation.org/issues/6337)

Describe changes:
- Updated the detect-ssh-proto-version tests to use new FAIL/PASS API

PR #9578, opened 2023-10-08
detect-ssh-proto-version: use FAIL macros in tests

I3run0:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [6337](https://redmine.openinfosecfoundation.org/issues/6337)

Previous PR: #9577

Describe changes:
- Updated the detect-ssh-proto-version tests to use new FAIL/PASS API
- Fixed clang-format error from the previous PR

### DIFF
--- a/src/detect-ssh-proto-version.c
+++ b/src/detect-ssh-proto-version.c
@@ -286,12 +286,10 @@ static int DetectSshVersionTestParse01 (void)
 {
     DetectSshVersionData *ssh = NULL;
     ssh = DetectSshVersionParse(NULL, "1.0");
-    if (ssh != NULL && strncmp((char *) ssh->ver, "1.0", 3) == 0) {
-        DetectSshVersionFree(NULL, ssh);
-        return 1;
-    }
+    FAIL_IF_NOT(ssh != NULL && strncmp((char *)ssh->ver, "1.0", 3) == 0);
+    DetectSshVersionFree(NULL, ssh);
 
-    return 0;
+    PASS;
 }
 
 /**
@@ -302,12 +300,10 @@ static int DetectSshVersionTestParse02 (void)
 {
     DetectSshVersionData *ssh = NULL;
     ssh = DetectSshVersionParse(NULL, "2_compat");
-    if (ssh->flags & SSH_FLAG_PROTOVERSION_2_COMPAT) {
-        DetectSshVersionFree(NULL, ssh);
-        return 1;
-    }
+    FAIL_IF_NOT(ssh->flags & SSH_FLAG_PROTOVERSION_2_COMPAT);
+    DetectSshVersionFree(NULL, ssh);
 
-    return 0;
+    PASS;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/detect-ssh-proto-version.c
```diff
@@ -286,12 +286,10 @@ static int DetectSshVersionTestParse01 (void)
 {
     DetectSshVersionData *ssh = NULL;
     ssh = DetectSshVersionParse(NULL, "1.0");
-    if (ssh != NULL && strncmp((char *) ssh->ver, "1.0", 3) == 0) {
-        DetectSshVersionFree(NULL, ssh);
-        return 1;
-    }
+    FAIL_IF_NOT(ssh != NULL && strncmp((char *)ssh->ver, "1.0", 3) == 0);
```
inashivb, line 289: Tip: Read out what the statement is trying to do and then convert them as is using FAIL/PASS API.

1. Let's break these conditions down for better readability. Unless needed, make such statements separate in different FAIL_IF.. calls.
2. Try to avoid double negation and keep the intention/readability of the statement clear.
3. There are more available FAIL_* macros in https://github.com/OISF/suricata/blob/master/src/util-unittest.h#L89 that may be better suited to your work here. 😉

src/detect-ssh-proto-version.c
```diff
@@ -318,27 +314,19 @@ static int DetectSshVersionTestParse03 (void)
...
-        return 0;
-    }
+    FAIL_IF_NOT_NULL(ssh);
     ssh = DetectSshVersionParse(NULL, "lalala");
+    FAIL_IF_NOT_NULL(ssh);
+
     if (ssh != NULL) {
         DetectSshVersionFree(NULL, ssh);
-        return 0;
     }
```
inashivb, line 327: Make sure to remove all the calls that are covered by the FAIL/PASS API.

inashivb (changes requested):
Hi, @I3run0 !
Kudos on your first contribution to the project. :)
I have some changes requested from my end inline.

src/detect-ssh-proto-version.c
I3run0: Perfect @inashivb! I will do that! Thank you for the tips

src/detect-ssh-proto-version.c
I3run0: How this call is covered by the FAIL/PASS API?

jufajardini: I suppose you are asking about the `DetectSshVersionFree()` call.

With the new FAIL/PASS approach, engine resources must only be freed when a test succeeds. But, since this test if for ensuring the `ssh` _is_ null, if the test succeeds, there's nothing to be freed. And if the test fails, well, we won't call freeing statements either.

If your question was about the whole if statement: all if statements should be converted to the new fail/pass API ;)

I3run0: It was about DetectSshVersionFree() call. I understand now, Thank you!

I3run0:
This PrlR it closed to update de code with the comment review

PR #9580, opened 2023-10-08
rust/doc: add docstring to rust module files - v2

0xEniola:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4584

Previous PR: #9579

Describe changes:
- Incorporate feedback from last PR

PR #9582, opened 2023-10-09
rust/doc: add docstring to rust module files - v3

0xEniola:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4584

Previous PR: #9580

Describe changes:
- Incorporated feedback from previous PR.
- Corrected typographical errors.

### DIFF
--- a/rust/src/ffi/mod.rs
+++ b/rust/src/ffi/mod.rs
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Foreign Function Interface module for hashing(sha256, sha1, md5), base64 and strings.
+
 pub mod hashing;
 pub mod base64;
 pub mod strings;
--- a/rust/src/filecontainer.rs
+++ b/rust/src/filecontainer.rs
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! This module handles file-related operations(open, append, close).
+
 use std::ptr;
 use std::os::raw::{c_void};
 
--- a/rust/src/filetracker.rs
+++ b/rust/src/filetracker.rs
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Gap handling and Chunk-based file transfer tracker module.
+
 /**
  *  \file
  *  \author Victor Julien <victor@inliniac.net>
--- a/rust/src/jsonbuilder.rs
+++ b/rust/src/jsonbuilder.rs
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+ //! JSON logger utility library module.
+
 #![allow(clippy::missing_safety_doc)]
 
 use std::cmp::max;
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Library tree module.
+
 #![cfg_attr(feature = "strict", deny(warnings))]
 
 // Allow these patterns as its a style we like.
--- a/rust/src/plugin.rs
+++ b/rust/src/plugin.rs
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Bootstrap plugin utility module.
+
 pub fn init() {
     unsafe {
         let context = super::core::SCGetContext();
--- a/rust/src/telnet/mod.rs
+++ b/rust/src/telnet/mod.rs
@@ -15,5 +15,7 @@
  * 02110-1301, USA.
  */
 
+//! telnet application layer and parser module.
+
 pub mod telnet;
 mod parser;
--- a/rust/src/util.rs
+++ b/rust/src/util.rs
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Utility module to test strings for valid UTF-8.
+
 use std::ffi::CStr;
 use std::os::raw::c_char;
 
--- a/rust/src/applayer.rs
+++ b/rust/src/applayer.rs
@@ -15,7 +15,7 @@
  * 02110-1301, USA.
  */
 
-//! Parser registration functions and common interface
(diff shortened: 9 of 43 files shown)


rust/src/lib.rs
```diff
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Library tree module.
```
inashivb, line 18: This does not sound right. `The Suricata Rust Core Library` perhaps? But, it could be misleading. @jasonish wdyt?

inashivb (changes requested):
@0xEniola thanks! :)
I'm unsure about one docstring and we'll have to wait for Jason to assist us on that. But, other than that, most things are looking OK to me.

rust/src/lib.rs
0xEniola: I thought I changed this to `Suricata Library Module`.

Seems I didn't save the file 🤦🏿‍♂️.

rust/src/ffi/mod.rs
```diff
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Foreign Function Interface module for hashing(sha256, sha1, md5), base64 and strings.
```
jasonish, line 18: I'd leave off the specifics of sha256, etc. As that information can get out of date easily.  More simply this should be something like
```
Module that exposes C bindings to the Suricata Rust library.
```
or something more generic like that.

rust/src/filecontainer.rs
```diff
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! This module handles file-related operations(open, append, close).
```
jasonish, line 18: This should explicitly be file container operations, so its not confused with OS level file operations.

rust/src/filetracker.rs
```diff
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Gap handling and Chunk-based file transfer tracker module.
+
 /**
  *  \file
  *  \author Victor Julien <victor@inliniac.net>
```
jasonish, line 22: Can we integrate this existing "module" level documentation into this new proper module level documentation?

rust/src/jsonbuilder.rs
```diff
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+ //! JSON logger utility library module.
```
jasonish, line 18:
```
JsonBuilder is a module for building JSON documents.
```

There is nothing here specific to logging, so leave that out.

rust/src/lib.rs
jasonish: This is Rust documentation entry point...  
```
//! Suricata is a network intrusion, prevention and monitoring engine.
//!
//! Suricata is a hybrid C and Rust application. What is found here are the
//! components written in Rust.
```
that could probably be expanded a lot, but that's the basics.

rust/src/plugin.rs
```diff
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Bootstrap plugin utility module.
```
jasonish, line 18: Drop Bootstrap.

rust/src/telnet/mod.rs
```diff
@@ -15,5 +15,7 @@
  * 02110-1301, USA.
  */
 
+//! telnet application layer and parser module.
```
jasonish, line 18: Telnet

rust/src/util.rs
```diff
@@ -15,6 +15,8 @@
  * 02110-1301, USA.
  */
 
+//! Utility module to test strings for valid UTF-8.
```
jasonish, line 18:
```
Utility module.
```

So it doesn't get out of date. Yes, we appear to have a few of these, but that's probably a cleanup job for another task/day.

jasonish (changes requested):
Changes requested. Thanks!

rust/src/filetracker.rs
0xEniola: So I should replace the documentation I added with the existing one, yeah?

jasonish: More like merge the 2.

0xEniola: So I'll be adding those short explanations present in the existing documentation too?

rust/src/lib.rs
0xEniola: So I should use this?

rust/src/filetracker.rs
jasonish: I think so.  If there is documentation there already, it was probably meant to be module level documentation, but not marked up correctly. But the existing documentation may be missing a title like line, which is more or less what you have been adding here.

0xEniola: I was thinking leaving it like this would be good, since it already serves as a title-like line, because the existing documentation is quite long and the issue details asked for a short description.

rust/src/lib.rs
jasonish: Yeah, I think so.

0xEniola: OK then.

Since the issue asks for a short description, I'll use the `Suricata Rust core library`.

rust/src/filetracker.rs
jasonish: Hmm, Ok. One line is better than before.

But idiomatically Rust modules are typically documented like this:

```
//! This is the brief description.
//!
//! Then one or more paragraphs providing more detail about what this module does.
```
not all of our modules have that paragraph, but some like this one do, so would be nice to incorporate it since the content is there already, its just a matter of formatting correctly for rustdoc.

0xEniola: Okay then.

I'll do this then.
Thank you for the feedback.

PR #9584, opened 2023-10-09
Filestore master6 6388 v1 (no backport)

catenacyber:

Describe changes:
- Adds some debug assertions

These flags are set by


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9585

PR #9585, opened 2023-10-09
Filestore master6 6388 v2

catenacyber:

Describe changes:
- Adds some debug assertions

These flags are set by

#9584 with clear PR history, including a commit to pin rust crate

### DIFF
--- a/src/util-file.c
+++ b/src/util-file.c
@@ -1109,7 +1109,13 @@ void FileUpdateFlowFileFlags(Flow *f, uint16_t set_file_flags, uint8_t direction
         FileContainer *ffc = AppLayerParserGetFiles(f, direction);
         if (ffc != NULL) {
             for (File *ptr = ffc->head; ptr != NULL; ptr = ptr->next) {
-                ptr->flags |= per_file_flags;
+                if ((per_file_flags & FILE_NOSTORE) && (ptr->flags & FILE_STORE)) {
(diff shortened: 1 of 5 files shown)


src/util-file.c
```diff
@@ -1107,7 +1109,13 @@ void FileUpdateFlowFileFlags(Flow *f, uint16_t set_file_flags, uint8_t direction
         FileContainer *ffc = AppLayerParserGetFiles(f, direction);
         if (ffc != NULL) {
             for (File *ptr = ffc->head; ptr != NULL; ptr = ptr->next) {
-                ptr->flags |= per_file_flags;
+                if ((per_file_flags & FILE_NOSTORE) && (ptr->flags & FILE_STORE)) {
+                    // if a rule reload removes a rule storing file
+                    // but a file was being stored...
```
catenacyber, line 1114: Do you see another context where ?
- `FileUpdateFlowFileFlags` is called
- `per_file_flags & FILE_NOSTORE` is true, meaning `set_file_flags & (FLOWFILE_NO_STORE_TS|FLOWFILE_NO_STORE_TC)` was the argument
- There are opened files in `ffc` (can happen on first packet with `stream.midstream=true`)
- This opened file is already set for store (can happen if store is forced, but then `set_file_flags` was cleared)

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9586

PR #9586, opened 2023-10-10
Filestore master6 6388 v3

catenacyber:
https://redmine.openinfosecfoundation.org/issues/6390
https://redmine.openinfosecfoundation.org/issues/6391
https://redmine.openinfosecfoundation.org/issues/6392
https://redmine.openinfosecfoundation.org/issues/6393

Describe changes:
- Adds some debug assertions

These flags are set by

#9585 with more subtickets

```
SV_BRANCH=pr/1415
```

### DIFF
--- a/src/output-filedata.c
+++ b/src/output-filedata.c
@@ -167,6 +167,7 @@ static void OutputFiledataLogFfc(ThreadVars *tv, OutputLoggerThreadData *td,
                 SCLogDebug("ff FILE_STORE not set");
                 continue;
             }
+            DEBUG_VALIDATE_BUG_ON(ff->flags & FILE_NOSTORE);
 
             /* if we have no data chunks left to log, we should still
              * close the logger(s) */
--- a/rust/Cargo.toml.in
+++ b/rust/Cargo.toml.in
@@ -22,7 +22,7 @@ decompression = ["flate2", "brotli"]
 [dependencies]
 nom = "~ 5.1.3"
 bitflags = "~1.2.1"
-byteorder = "1.3"
+byteorder = "~1.3"
 uuid = "0.8"
 crc = "1.8"
 memchr = "~ 2.3"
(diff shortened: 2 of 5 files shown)


catenacyber:
Should I exclude commit [detect: log error when too many filestore](https://github.com/OISF/suricata/pull/9586/commits/ed3d2b6d6f9d7fa55d076d0b608aad0612d5f292) from this PR ? (because it is also in master, not only in master6 as the other ones)

catenacyber:
Added https://github.com/OISF/suricata-verify/pull/1415

jasonish:

So do truncated files no longer get stored?

catenacyber:
> > Make sure we do not store a file if we do not have its full contents (and thus the naming does not match the hash)
> 
> So do truncated files no longer get stored?

Truncated files can still get stored.
How can I rephrase it ?

>  if we do not have its full contents 

I mean if there is content Suricata has seen, has hashed, but not put in the streaming buffer

src/output-filedata.c
```diff
@@ -167,6 +167,7 @@ static void OutputFiledataLogFfc(ThreadVars *tv, OutputLoggerThreadData *td,
                 SCLogDebug("ff FILE_STORE not set");
                 continue;
             }
+            DEBUG_VALIDATE_BUG_ON(ff->flags & FILE_NOSTORE);
```
catenacyber, line 170: Argh this was kind of added in master with new `FileApplyTxFlags` checking
`DEBUG_VALIDATE_BUG_ON((file->flags & (FILE_STORE | FILE_NOSTORE)) == (FILE_STORE | FILE_NOSTORE));`

And S-V test hits this condition

suricata-qa:

Pipeline 16252

victorjulien:
We need fixes towards master before stable branches.

catenacyber:
> We need fixes towards master before stable branches.

Indeed, I thought master was okay, see https://github.com/OISF/suricata/pull/9996

catenacyber:
Closing this, and will rebase or cherry-pick commits when master get its own fixes

PR #9587, opened 2023-10-10
output/eve: reduce fflush call count

jlucovsky:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: [3449](https://redmine.openinfosecfoundation.org/issues/3449)

Describe changes:
- Add EVE configuration parameter to control buffering: `buffer-size`. When 0, unbuffered I/O is used; other values are used to set the stdio buffer size. The value is `outputs.eve-log.buffer-size`
- Add Suricata configuration parameter -- `flush-interval` -- to set cadence for Suricata periodically directing detect threads to flush EVE output. To be used in conjunction with `buffer-size`. Set `logging.flush-interval` to the number of seconds Suricata should periodically cause the EVE output to be flushed. The default value is `0` which instructs Suricata to never cause the EVE output to be flushed.
- Add mechanism to periodically send pseudo packets to detect threads as a way to trigger flush. Controlled by `flush-interval`

pull request number.

```
SV_REPO=
SV_BRANCH=
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=
```

### DIFF
--- a/src/detect-engine.c
+++ b/src/detect-engine.c
@@ -2279,21 +2279,26 @@ int DetectEngineInspectPktBufferGeneric(
 }
 
 /** \internal
- *  \brief inject a pseudo packet into each detect thread that doesn't use the
- *         new det_ctx yet
+ *  \brief inject a pseudo packet into each detect thread
+ *      -that doesn't use the new det_ctx yet
+ *      -*or*, if the thread should flush its output logs.
  */
-static void InjectPackets(ThreadVars **detect_tvs,
-                          DetectEngineThreadCtx **new_det_ctx,
-                          int no_of_detect_tvs)
+static void InjectPackets(ThreadVars **detect_tvs, DetectEngineThreadCtx **new_det_ctx,
+        int no_of_detect_tvs, bool flush_logs)
 {
-    /* inject a fake packet if the detect thread isn't using the new ctx yet,
-     * this speeds up the process */
+    /* inject a fake packet if the detect thread that needs it. This function
+     * is called if
+     *  - A thread isn't using a DE ctx and should
+     *  - Or, it should process a pseudo packet and flush its output logs.
+     * to speed the process. */
     for (int i = 0; i < no_of_detect_tvs; i++) {
-        if (SC_ATOMIC_GET(new_det_ctx[i]->so_far_used_by_detect) != 1) {
+        if (flush_logs || SC_ATOMIC_GET(new_det_ctx[i]->so_far_used_by_detect) != 1) {
             if (detect_tvs[i]->inq != NULL) {
                 Packet *p = PacketGetFromAlloc();
                 if (p != NULL) {
                     p->flags |= PKT_PSEUDO_STREAM_END;
+                    if (flush_logs)
+                        p->flags |= PKT_PSEUDO_LOG_FLUSH;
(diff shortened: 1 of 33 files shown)


jasonish:
Can you add more detail to the commit that adds the new flush function pointer to each registration? To me, the act of flushing is specific to the file output, not generic, so would ideally be abstracted away from the loggers themselves that are not aware of a file, socket, etc.

jlucovsky:
> Can you add more detail to the commit that adds the new flush function pointer to each registration? To me, the act of flushing is specific to the file output, not generic, so would ideally be abstracted away from the loggers themselves that are not aware of a file, socket, etc.

Yes. Note that loggers registering with OutputRegisterPacketSubModule  use the flushing interface. The remaining changes are due to the registration internals.

suricata-qa:

Pipeline 16251

suricata-qa:

Pipeline 16273

suricata-qa:

Pipeline 16278

src/detect-engine.c
```diff
@@ -2306,6 +2311,111 @@ static void InjectPackets(ThreadVars **detect_tvs,
     }
 }
 
+/**
+ * \brief Trigger detect threads to flush their output logs
+ *
+ * This function is intended to be called at regular intervals to force
+ * buffered log data to be persisted
+ */
+void DetectEngineFlush(void)
```
jlucovsky, line 2320: Suggest a different name, e.g., "WorkerFlush", because it's not really part of the "DetectEngine"

src/suricata.c
```diff
@@ -2837,6 +2844,14 @@ static void SuricataMainLoop(SCInstance *suri)
             DetectEngineReloadSetIdle();
         }
 
+        if (flush_wait_count) {
+            if (++wait_count == flush_wait_count) {
+                SCLogDebug("Sending flush packet to threads");
+                DetectEngineFlush();
```
jlucovsky, line 2850: @victorjulien suggests moving this out of the main thread as the main thread can block for events -- e.g., rule reload.

src/output.h
```diff
@@ -87,15 +89,14 @@ extern OutputModuleList output_modules;
...
-void OutputRegisterPacketModule(LoggerId id, const char *name,
-    const char *conf_name, OutputInitFunc InitFunc, PacketLogger LogFunc,
-    PacketLogCondition ConditionFunc, ThreadInitFunc, ThreadDeinitFunc,
-    ThreadExitPrintStatsFunc);
-void OutputRegisterPacketSubModule(LoggerId id, const char *parent_name,
-    const char *name, const char *conf_name, OutputInitSubFunc InitFunc,
-    PacketLogger LogFunc, PacketLogCondition ConditionFunc,
-    ThreadInitFunc ThreadInit, ThreadDeinitFunc ThreadDeinit,
-    ThreadExitPrintStatsFunc ThreadExitPrintStats);
+void OutputRegisterPacketModule(LoggerId id, const char *name, const char *conf_name,
```
jlucovsky, line 92: Investigate whether we can reduce the parameter count through the use of a parameter type that includes the many function pointers.

src/suricata.c
jlucovsky: Will create a new thread for this function since the main thread can block (e.g., during rule reloads).

jlucovsky:
Continued in #9802

PR #9560, opened 2023-10-05
Detect tcp noupdatetotx 6299 v5.1

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6299

Describe changes:
- Optimization to not run transaction detection when a TCP packet did not update anything (no call to AppLayerParserParse) in the packet direction

Let's see what QA thinks of this

Better check than #9481 with QA deviation fixed

PR #9496, opened 2023-09-18
Detect tcp noupdatetotx 6299 v6

catenacyber:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/6299

Describe changes:
- Optimization to not run transaction detection when a TCP packet did not update anything (no call to AppLayerParserParse) in the packet direction
- Merge sorted lists should be faster than qsort
- do not store state if there is no flags set

Let's see what QA thinks of this

#9481 with new other optimizations

PR #9589, opened 2023-10-10
Detect tcp noupdatetotx 6299 v6.1

catenacyber:

Describe changes:
- Merge sorted lists should be faster than qsort
- do not store state if there is no flags set

#9496 rebased on top of #9560


catenacyber:
Replaced by https://github.com/OISF/suricata/pull/9638

PR #9581, opened 2023-10-08
  Doc: Updating the installation guide in the user guide to include instructions for the 'install-full' and 'install-conf' options v1. #6342

comfort619:
- [x] I have read the contributing guide lines at https://docs.suricata.io/en/latest/devguide/codebase/contributing/contribution-process.html
- [x] I have signed the Open Information Security Foundation contribution agreement at https://suricata.io/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/install.rst) to reflect the changes made 

Link to [redmine](https://redmine.openinfosecfoundation.org/issues/6342) ticket:

Describe changes:

- Added instructions for the 'install-full' option to the installation guide.
- Included guidance on using the 'install-conf' option for better configuration.
- Ensured that these instructions are presented clearly in the appropriate sections of the guide.
- Checked and maintained consistent formatting throughout the document.

**Testing Done:**

I've tested these changes locally to ensure that the installation guide is accurate and easy to follow.

**Related Issues:**

Ticket: 6342

**Additional Notes:**

- This PR includes updates to the user guide for improved clarity.
- It follows the project's coding and documentation standards.

I look forward to your feedback and appreciate your review of these changes.

PR #9592, opened 2023-10-11
userguide: cover install-full and install-conf -v2

comfort619:
Ticket: 6342
Link to [redmine](https://redmine.openinfosecfoundation.org/issues/6342) ticket:

Describe changes:

-Incorporate feedback from last PR.
-Previous PR: https://github.com/OISF/suricata/pull/9581

### DIFF
--- a/doc/userguide/install.rst
+++ b/doc/userguide/install.rst
@@ -165,6 +165,34 @@ Follow these steps from your Suricata directory::
     make -j8 # j is for paralleling, you may de/increase depending on your CPU
     make install # to install your Suricata compiled binary
 
+
+Auto-Setup
+""""""""""
+
+You can also use the available auto-setup features of Suricata:
+
+::
+
+    ./configure && make && sudo make install-conf
+
+*make install-conf* would do the regular "make install" and then it would automatically
+create/setup all the necessary directories and ``suricata.yaml`` for you.
+
+::
+
+    ./configure && make && make install-rules
+
+*make install-rules* would do the regular "make install" and then it would automatically
+download and set up the latest ruleset from Emerging Threats available for Suricata.
+
+::
+
+    ./configure && make && make install-full
+
+*make install-full* would combine everything mentioned above (install-conf and install-rules)
+and will present you with a ready-to-run (configured and set-up) Suricata.
+
+
 .. _install-binary-packages:
 
 Binary packages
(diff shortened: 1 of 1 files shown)


doc/userguide/install.rst
```diff
@@ -165,6 +165,34 @@ Follow these steps from your Suricata directory::
...
+*make install-rules* would do the regular "make install" and then it would automatically
+download and set up the latest ruleset from Emerging Threats available for Suricata.
+
+::
+
+    ./configure && make && make install-full
+
+*make install-full* would combine everything mentioned above (install-conf and install-rules)
+and will present you with a ready-to-run (configured and set-up) Suricata.
+
```
inashivb, line 194: I feel that `compilation` and `auto setup` should not be a part of the `Dependencies` section. Thoughts @jufajardini ?

doc/userguide/install.rst
```diff
@@ -361,6 +389,7 @@ Suricata packages.
 
     yay -S suricata
 
+
```
inashivb, line 392: nit: unneeded newline

inashivb (changes requested):
Hi @comfort619 !
Looking pretty good now. Thanks for incorporating the feedback. I have a doubt about the organization of sections, I have asked Juliana's opinion on that.
One minor change requested from me inline.

doc/userguide/install.rst
comfort619: > I feel that `compilation` and `auto setup` should not be a part of the `Dependencies` section. Thoughts @jufajardini ?

I feel same too looking at it properly again, the dependencies I think are just prerequisite for the software to function which the steps are required to meet external software and library requirements but compilation and auto set up are part of the internal actions specific to the software install process. So likely the compilation and auto set-up should be in a different header section as still relating to the install process and not under dependencies. But then I still wait @jufajardini thoughts on it

jufajardini: Good catch!


I'm thinking that... 
The `Compilation` Section shouldn't be a part of `Dependencies`, and changing that title would solve this. 
Might have been something that slipped in a past commit/review...
How does that look, to you?

jufajardini: Actually, let's have two changes here:
- Make `Compilation` a section, instead of a subsection of `Dependencies`.
- Make `Auto Setup` another section. 

Seems to make more sense in my head, right now. Any thoughts you may have on that are welcome! :)

comfort619: I think the same too. I will make the changes

PR #9562, opened 2023-10-05
Analyze DPDK hugepages and be more user-friendly with non-existing user and check for shutdown before accessing Unix sockets v6

lukashino:
Follow-up of #9549 

PR is a combination of:

https://redmine.openinfosecfoundation.org/issues/6278
https://redmine.openinfosecfoundation.org/issues/5966
https://redmine.openinfosecfoundation.org/issues/6272
https://redmine.openinfosecfoundation.org/issues/6381
https://redmine.openinfosecfoundation.org/issues/6382

Describe changes:
- add Github workflow changes (Redmine ticket 6381, 6382)

To use a pull request use a branch name like pr/N where N is the
pull request number.

Alternatively, SV_BRANCH may also be a link to an
OISF/suricata-verify pull-request.

SV_REPO=
SV_BRANCH=pr/1412
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=

PR #9596, opened 2023-10-11
Analyze DPDK hugepages and be more user-friendly with non-existing user and check for shutdown before accessing Unix sockets v7

lukashino:
Follow-up of #9562 

PR is a combination of:

https://redmine.openinfosecfoundation.org/issues/6278
https://redmine.openinfosecfoundation.org/issues/5966
https://redmine.openinfosecfoundation.org/issues/6272
https://redmine.openinfosecfoundation.org/issues/6381
https://redmine.openinfosecfoundation.org/issues/6382

Describe changes:
- PR comment of #9562 

To use a pull request use a branch name like pr/N where N is the
pull request number.

Alternatively, SV_BRANCH may also be a link to an

SV_REPO=
SV_BRANCH=pr/1417
SU_REPO=
SU_BRANCH=
LIBHTP_REPO=
LIBHTP_BRANCH=

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1961,7 +1961,7 @@ jobs:
     needs: [ prepare-deps, prepare-cbindgen ]
     strategy:
       matrix:
-        dpdk_version: [ 22.11.1, 21.11.3, 20.11.7, 19.11.14 ]
+        dpdk_version: [ 22.11.3, 21.11.5, 20.11.9, 19.11.14 ]
     steps:
 
       # Cache Rust stuff.
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1961,7 +1961,7 @@ jobs:
     needs: [ prepare-deps, prepare-cbindgen ]
     strategy:
       matrix:
-        dpdk_version: [ 22.11.3, 21.11.5, 20.11.9, 19.11.14 ]
+        dpdk_version: [ "main", 22.11.3, 21.11.5, 20.11.9, 19.11.14 ]
     steps:
 
       # Cache Rust stuff.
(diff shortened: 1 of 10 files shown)


.github/workflows/builds.yml
```diff
@@ -1961,7 +1961,7 @@ jobs:
     needs: [ prepare-deps, prepare-cbindgen ]
     strategy:
       matrix:
-        dpdk_version: [ 22.11.1, 21.11.3, 20.11.7, 19.11.14 ]
+        dpdk_version: [ "main", 22.11.3, 21.11.5, 20.11.9, 19.11.14 ]
```
jasonish, line 1964: I think we should drop main for now as its a moving target.  Could be moved to a job that is allowed to fail though.

victorjulien:
Subject suggests 3 different sets of changes, can you split this up into 3 PRs?

.github/workflows/builds.yml
victorjulien: Agree

victorjulien (changes requested):
see inline

.github/workflows/builds.yml
lukashino: In the end, I decided to omit the main because it was throwing me a warning when I was using an obsolete function and at the same time it was giving me a warning when I was using the new function because it is not part of the stable ABI. 
I guess this was not intended from the DPDK side but it would mean that in either case (with the obsolete or the new function), the job would fail, or with CFLAGS warnings disabled, we would never notice such warnings. I'll try to rethink this.

lukashino:
PR continues in 4 PRs:
https://github.com/OISF/suricata/pull/9628
https://github.com/OISF/suricata/pull/9629
https://github.com/OISF/suricata/pull/9630
https://github.com/OISF/suricata/pull/9631

